PR #3878, opened 2025-07-30
perf(l1): save chain config in vm,  cache latestblock header

edg-l:
**Motivation**

On samply, most of the work seems to be in validate_tx on the rpc.

Inside validate tx, get_block_header took 22% and latest_block_number about 3%.
This pr turns those numbers into 0.2%. Now get account info takes 93%. 

PR based on #3865

Before (pr #3865):
image

After:
image

make load-test

Prev(perf/aux_sender_nonce_mempool) gigas of first 4 blocks:
0.6, 0.5, 0.4, 0.4, 0.4

PR Gigagas of first 4 blocks:
0.7, 0.56, 0.43, 0.42,

Slightly improves ethrex transfers bench on gas benches (7.82%).

PR #3998, opened 2025-08-08
fix(l1): load Store state when exporting blocks

MegaRedHand:
**Motivation**

`ethrex export` is broken due to #3878

**Description**

This PR fixes the `Export` subcommand by adding an initial state loading from the DB to initialize the `latest_block_header` cache.


cmd/ethrex/initializers.rs
```diff
@@ -58,7 +58,7 @@ pub fn init_metrics(opts: &Options, tracker: TaskTracker) {
     tracker.spawn(metrics_api);
 }
 
-/// Opens a New or Pre-exsisting Store and loads the initial state provided by the network
+/// Opens a New or pre-existing Store and loads the initial state provided by the network
```
fmoletta:
```suggestion
/// Opens a new or pre-existing Store and loads the initial state provided by the network
```

Oppen (approved):

PR #4000, opened 2025-08-08
fix(l1): don't emit warning if we have no peers file

### LINKED ISSUE #3908: Ethrex should not log an error if the peers file is not found
`2025-07-31T15:43:20.316693Z ERROR ethrex::initializers: Could not read from peers file: No config file found: No such file or directory (os error 2)`

For every new execution the peers file will not exist and this is fine. The logs should not error for this use case.

fmoletta:
Closes #3908

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -32,7 +32,7 @@ use std::{
 };
 use tokio::sync::Mutex;
 use tokio_util::{sync::CancellationToken, task::TaskTracker};
-use tracing::{error, info, warn};
+use tracing::{debug, error, info, warn};
 use tracing_subscriber::{EnvFilter, FmtSubscriber, filter::Directive};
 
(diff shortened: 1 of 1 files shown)


MegaRedHand (approved):

cmd/ethrex/initializers.rs
```diff
@@ -237,7 +237,7 @@ pub fn get_bootnodes(opts: &Options, network: &Network, data_dir: &str) -> Vec<N
 
     match read_node_config_file(config_file) {
         Ok(ref mut config) => bootnodes.append(&mut config.known_peers),
-        Err(e) => warn!("Could not read from peers file: {e}"),
+        Err(e) => debug!("Could not read from peers file: {e}"),
```
mpaulucci, line 240: we do want to log an error if there is an actual error. But we want to handle the non existance of a file separately (it is not an error, it's a valid usecase)

PR #4010, opened 2025-08-11
fix(levm): possible bug within destroyed accounts structure

JereSalo:
**Motivation**

**Description**

- We are not clearing the `destroyed_accounts` structure at any time, I'm not completely sure if we should but I'll leave this here until I'm sure about what to do

Closes #issue_number


JereSalo:
I think I was wrong and isn't a problem.

PR #4015, opened 2025-08-11
fix: retag `ethrex:unstable` as `latest` in Makefile

MegaRedHand:
**Motivation**

Running hive tests from a new repo causes errors due to the `latest` tag for the `ethrex`

**Description**

This PR fixes this by re-tagging the newly-built "unstable" image as "latest"


mpaulucci (changes requested):
I think @ManuelBilbao had decided that `latest` is the `stable` release, which is different from `unstable`. Why can't we use `unstable` in tests?

cdiielsi (approved):

PR #4020, opened 2025-08-12
perf(l1): refactor libmdbx storage, removing unneeded rlp encoding

edg-l:
**Motivation**

 Remove unneeded RLP encoding from key/values in libmdbx storage

The database tables are completly private so we don't need to use rlp encoding at all, avoiding possible perfomance costs by storing as bytes directly where possible.

Complex types like entire blocks remain as RLP, future improvements could be made, like changing to a possibly more perfomant encoding like bincode.

I see improvements of about 6% on gas benchmarks


JereSalo:
why `perf(levm)` and not `perf(l1)`?

edg-l:
> why `perf(levm)` and not `perf(l1)`?

changed it, thanks!

Oppen (approved):

crates/storage/store_db/libmdbx.rs
```diff
@@ -1272,83 +1273,52 @@ table!(
...
+    /// Key: Account hash
+    ( StateSnapShot ) Bytes32 => AccountStateRLP
 );
 
 dupsort!(
     /// Storage Snapshot used by an ongoing sync process
-    ( StorageSnapShot ) AccountHashRLP => (AccountStorageKeyBytes, AccountStorageValueBytes)[AccountStorageKeyBytes]
+    ///
+    /// Key: Account hash
+    /// (storage key, storage value)[storage key]
```
mpaulucci:
```suggestion
    /// Value: (storage key, storage value)[storage key]
```

mpaulucci (approved):
lets coordinate the merge of this PR since it is disruptive (it has a breaking change on the db)

jrchatruc (changes requested):
Blocking merge of this PR until we coordinate it as @mpaulucci said to make sure nothing in our L2s break, etc

PR #4022, opened 2025-08-12
chore(l1): improve log message when trying to run incompatible configs

mpaulucci:
**Motivation**
Current error message was confusing.

### DIFF
--- a/crates/storage/error.rs
+++ b/crates/storage/error.rs
@@ -53,5 +53,5 @@ pub enum StoreError {
     #[error("Failed to lock database for writing")]
     LockError,
     #[error("Incompatible chain configuration")]
-    IncompatibleChainConfig(),
+    IncompatibleChainConfig,
 }
--- a/crates/storage/error.rs
+++ b/crates/storage/error.rs
@@ -52,4 +52,6 @@ pub enum StoreError {
     MempoolReadLock(String),
     #[error("Failed to lock database for writing")]
     LockError,
+    #[error("Incompatible chain configuration")]
+    IncompatibleChainConfig(),
 }
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -604,7 +604,7 @@ impl Store {
             }
             Some(_) => {
                 error!(
-                    "The chain configuration stored in the database is incompatible with the provided configuration. If you intended to switch networks, clear the database (e.g., run `ethrex removedb`) and try again."
+                    "The chain configuration stored in the database is incompatible with the provided configuration. If you intended to switch networks, choose another datadir or clear the database (e.g., run `ethrex removedb`) and try again."
                 );
                 return Err(StoreError::IncompatibleChainConfig);
             }
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -26,7 +26,7 @@ use std::{
     collections::{BTreeMap, HashMap},
     sync::RwLock,
 };
-use tracing::{info, instrument};
+use tracing::{error, info, instrument};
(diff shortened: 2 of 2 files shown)


crates/storage/error.rs
```diff
@@ -52,4 +52,6 @@ pub enum StoreError {
     MempoolReadLock(String),
     #[error("Failed to lock database for writing")]
     LockError,
+    #[error("Incompatible chain configuration")]
+    IncompatibleChainConfig(),
```
MegaRedHand: The parentheses are not needed

```suggestion
    IncompatibleChainConfig,
```

crates/storage/store.rs
```diff
@@ -604,7 +602,12 @@ impl Store {
                 info!("Received genesis file matching a previously stored one, nothing to do");
                 return Ok(());
             }
-            Some(_) => panic!("{GENESIS_DIFF_PANIC_MESSAGE}"),
+            Some(_) => {
+                error!(
+                    "The chain configuration stored in the database is incompatible with the provided configuration. If you intended to switch networks, clear the database (e.g., run `ethrex removedb`) and try again."
+                );
+                return Err(StoreError::IncompatibleChainConfig());
```
MegaRedHand: Same

```suggestion
                return Err(StoreError::IncompatibleChainConfig);
```

crates/storage/store.rs
```diff
@@ -1436,16 +1439,11 @@ mod tests {
...
-        let panic = add_initial_state_handle.await.unwrap_err().into_panic();
-        assert_eq!(
-            panic
-                .downcast_ref::<String>()
-                .expect("Failed to downcast panic message"),
-            &GENESIS_DIFF_PANIC_MESSAGE
-        );
+        let result = add_initial_state_handle.await.unwrap();
+        assert!(result.is_err());
+        assert!(matches!(result, Err(StoreError::IncompatibleChainConfig())));
```
MegaRedHand: Same

```suggestion
        assert!(matches!(result, Err(StoreError::IncompatibleChainConfig)));
```

crates/storage/store.rs
```diff
@@ -1436,16 +1439,11 @@ mod tests {
...
         let add_initial_state_handle =
             tokio::task::spawn(async move { store.add_initial_state(genesis_hive).await });
-        let panic = add_initial_state_handle.await.unwrap_err().into_panic();
-        assert_eq!(
-            panic
-                .downcast_ref::<String>()
-                .expect("Failed to downcast panic message"),
-            &GENESIS_DIFF_PANIC_MESSAGE
-        );
+        let result = add_initial_state_handle.await.unwrap();
```
MegaRedHand: I think we can further simplify this by removing the `spawn` call, which was probably used only to more easily catch the panic.

MegaRedHand (approved):
LGTM. Left some comments about style and further simplifications

crates/storage/store.rs
```diff
@@ -604,7 +602,12 @@ impl Store {
                 info!("Received genesis file matching a previously stored one, nothing to do");
                 return Ok(());
             }
-            Some(_) => panic!("{GENESIS_DIFF_PANIC_MESSAGE}"),
+            Some(_) => {
+                error!(
+                    "The chain configuration stored in the database is incompatible with the provided configuration. If you intended to switch networks, clear the database (e.g., run `ethrex removedb`) and try again."
```
JereSalo: The user could also specify another datadir without needing to remove the db, right? Do you think we should specify this here so that they know they have another alternative if they don't want to remove the current database?

mpaulucci: good idea!

PR #4025, opened 2025-08-12
perfomance logging

edg-l:
**Motivation**

Logs are processed in another thread, add_log returns instantly (using a lock free queue)
```rust
// API

pub enum Value {
    Duration(Duration),
    Float(f64),
    Int(usize),
}

// add a log to be processed
pub fn add_log(label: impl Into<Cow<'static, str>>, value: Value);

// times until the guard is dropped, useful in multiple branch returning functions
let guard = perf_logger::add_time_till_drop("label"); 

guard.wrap_return(return expr)

```

Logs format:

```
"seconds_since_start | label | value | Value type description"
```

log saved to cwd at ethrex-perf-logs.log


Arkenan:
This was a single time debug effort, we can close it now.

PR #1926, opened 2025-02-12
fix(l1): hive invalid ancestor

MarcosNicolau:
**Motivation**
Fixes hive invalid ancestor tests under `engine-cancun` with the pattern `engine-cancun/Invalid Missing .* .*CanonicalReOrg=False`.

**Description**
The following was implemented to fix the tests:
- Because `forkchoice_update` and `new_payload` require us to send the latest valid hash, we now cache the last valid hash of the chain corresponding to bad blocks, these are the `invalid_ancestors`.
- The bad blocks are cached when:
   1. When failing to add a block during syncing
   2.  When the block execution of `new_payload` fails.
- The latest valid hash was fixed on `new_payload`; this was a known issue see #982.
- Finally, when syncing, if we have a pending block from a `new_payload` request, we attach it if the parent hash of the block corresponds to the last header parent hash.

Closes #982
Advances on #1285

PR #4028, opened 2025-08-12
fix(l1): handle syncing in case of reorgs and handle the case of reorg with invalid side chain

### LINKED ISSUE #3105: Invalid Missing Ancestor Syncing ReOrG hive test is flaky
The hive test `Invalid Missing Ancestor Syncing ReOrG` is flaky and has been deactivated in the CI. We need to find the cause of the problem and reactivate it.

cdiielsi:
**Motivation**

Restoring `request_block_headers` previous to this [pr](https://github.com/lambdaclass/ethrex/pull/4213) since it brought issues in case of re org.
Fixing the Invalid Missing Ancestor Syncing ReOrG hive tests which are failing.

**Description**

Up to now, when executing blocks during syncing, if there's an error while running a batch we set only the failing block with having an invalid ancestor. This pr sets all the downloaded descendants of a failing block with having an invalid ancestor during syncing.

Also, when syncing, it drops the sync head header from the requested headers when downloading block bodies if the sync head is stored as a pending block, since we already have it and there's no need to ask for it.

On another note, it was necessary to restore for full syncing the `request_block_headers` that we had previous to merging this [pr](https://github.com/lambdaclass/ethrex/pull/4213), since there was an issue with the way the block headers were requested. The core problem with the update was that we requested the all headers left from the sync head on by number, and since the tests come with a re-org scenario and the sync head number was the same as the current head number, no headers were requested, so the fork choice update never stopped syncing.

Closes #3105


crates/networking/p2p/sync.rs
```diff
@@ -546,6 +547,12 @@ impl FullBlockSyncState {
                 .map(|(header, body)| Block { header, body });
             self.current_blocks.extend(blocks);
         }
+        self.current_blocks.push(
+            self.store
+                .get_pending_block(sync_head_header.hash())
+                .await?
+                .unwrap(),
```
cdiielsi: TODO: drop unwrap

crates/networking/p2p/sync.rs
```diff
@@ -572,14 +579,33 @@ impl FullBlockSyncState {
...
                 cancel_token.clone(),
             )
             .await
             {
                 if let Some(batch_failure) = batch_failure {
                     warn!("Failed to add block during FullSync: {err}");
+                    let blocks_with_invalid_ancestor: Vec<Block> = block_batch[block_batch
+                        .iter()
+                        .position(|x| x.hash() == batch_failure.failed_block_hash)
+                        .unwrap()
```
cdiielsi: TODO: drop unwrap

crates/networking/p2p/sync.rs
```diff
@@ -574,14 +590,34 @@ impl FullBlockSyncState {
...
+                            .set_latest_valid_ancestor(block.hash(), batch_failure.last_valid_hash)
+                            .await?;
+                    }
+
+                    for header in self.current_headers.clone() {
+                        self.store
+                            .set_latest_valid_ancestor(header.hash(), batch_failure.last_valid_hash)
+                            .await?;
+                    }
+
```
cdiielsi, line 620: Original code set the failing block as invalid regardless of the type of error. Should this be the same for it's descendants as it is now? Or should descendants be marked with invalid ancestor only on "Invalid Block" error?

crates/networking/p2p/sync.rs
```diff
@@ -574,20 +590,34 @@ impl FullBlockSyncState {
...
             {
                 if let Some(batch_failure) = batch_failure {
                     warn!("Failed to add block during FullSync: {err}");
-                    self.store
-                        .set_latest_valid_ancestor(
-                            batch_failure.failed_block_hash,
-                            batch_failure.last_valid_hash,
-                        )
-                        .await?;
+                    // Since running the batch failed we set the failing block and it's descendants with having an invalid ancestor.
```
cdiielsi: Original code set the failing block as invalid regardless of the type of error. Should this be the same for it's descendants as it is now? Or should descendants be marked with invalid ancestor only on "Invalid Block" error?

mpaulucci: yes, it seems you should only set the block as invalid if the block validation yielded that it was invalid, not with every error like network error (If that's what you're asking).

crates/networking/p2p/sync.rs
```diff
@@ -529,6 +529,16 @@ impl FullBlockSyncState {
             // We don't have enough headers to fill up a batch, lets request more
             return Ok(());
         }
+        // If we already have the sync head header block there's no need to request it.
```
mpaulucci: Syncing should not have to deal with pending blocks ideally. Why can't you sync to the block n-1 instead of to the block n, and then execute the block n?

cdiielsi: Currently when the syncing cycle starts, [here](https://github.com/lambdaclass/ethrex/blob/572230cdd74c0d6b331632178dd6262a2af1be4d/crates/networking/p2p/sync.rs#L228) the pending block's header is included in the block headers that are going to be requested when `process_incoming_headers` is executed. I thought this might be needed in the case of snap syncing, that's why I dealt with the pending block directly in process_incoming_headers in the case of full syncing and I synced up to the nth block. Should I refactor sync_cycle to change this?

mpaulucci: Yeah, that code is probably incorrect. It is not related to snap sync, it was included in this PR: https://github.com/lambdaclass/ethrex/pull/1926. I would remove it and refactor the code. Lets not ask another PR for a block if we have it stored locally.

crates/networking/p2p/peer_handler.rs
```diff
@@ -554,7 +557,72 @@ impl PeerHandler {
         Some(ret)
     }
 
-    /// given a peer id, a chunk start and a chunk limit, requests the block headers from the peer
+    /// Requests block headers from any suitable peer, starting from the `start` block hash towards either older or newer blocks depending on the order
+    /// - No peer returned a valid response in the given time and retry limits
+    ///   Since request_block_headers brought problems in cases of reorg seen in this pr, we have this other function to request block headers only for full sync.
+    pub async fn request_block_headers_full_sync(
```
gianbelinche: We should not refer to "this pr" in the code.

crates/networking/p2p/sync.rs
```diff
@@ -590,13 +576,24 @@ impl FullBlockSyncState {
...
+        // If we have the sync_head as a pending block form a new_payload request and its parent_hash matches the hash of the latest received header
+        // we set the sync_head as found. Then later, after requesting all blocks, we'll add the pending block in current_blocks for execution.
+        if let Some(block) = self.store.get_pending_block(sync_head).await? {
+            if let Some(last_header) = self.current_headers.last() {
+                if block.header.parent_hash == last_header.hash() {
+                    sync_head_found = true;
+                }
+            }
+        }
+
```
gianbelinche: I think we should not pass a bool as mut, it makes the code harder to follow, instead we should return it and update it where the function is called.
Since now the function will return 2 bools, we should document what they mean so there are no confusions

crates/networking/p2p/peer_handler.rs
cdiielsi: I meant to add the link but I forgot 🤦‍♀️

crates/networking/p2p/Cargo.toml
```diff
@@ -13,6 +13,7 @@ ethrex-storage.workspace = true
 ethrex-trie.workspace = true
 ethrex-storage-rollup.workspace = true
 ethereum-types.workspace = true
+ethrex-vm.workspace = true
```
mpaulucci: why this change?

crates/networking/p2p/peer_handler.rs
mpaulucci: It doesn't make sense to refer to `full_sync` here, this module should not know nor care about full sync. `request_block_headers_from_hash` for example makes more sense

crates/networking/p2p/sync.rs
mpaulucci: +1

crates/networking/p2p/sync.rs
```diff
@@ -590,13 +576,24 @@ impl FullBlockSyncState {
...
-        sync_head_found: bool,
+        sync_head: H256,
+        mut sync_head_found: bool,
         blockchain: Arc<Blockchain>,
         peers: PeerHandler,
         cancel_token: CancellationToken,
     ) -> Result<bool, SyncError> {
         info!("Processing incoming headers full sync");
         self.current_headers.extend(block_headers);
+        // If we have the sync_head as a pending block form a new_payload request and its parent_hash matches the hash of the latest received header
```
mpaulucci: s/form/from

crates/networking/p2p/sync.rs
```diff
@@ -590,13 +576,24 @@ impl FullBlockSyncState {
...
+        // we set the sync_head as found. Then later, after requesting all blocks, we'll add the pending block in current_blocks for execution.
+        if let Some(block) = self.store.get_pending_block(sync_head).await? {
```
mpaulucci: why do we need to deal with pending blocks at this stage?

crates/networking/p2p/sync.rs
```diff
@@ -640,7 +645,7 @@ impl FullBlockSyncState {
                 .hash()
         );
         let execution_start = Instant::now();
-        let block_batch: Vec<Block> = self
+        let mut block_batch: Vec<Block> = self
```
mpaulucci: why this?

crates/networking/p2p/sync.rs
```diff
@@ -661,20 +666,49 @@ impl FullBlockSyncState {
         // Run the batch
         if let Err((err, batch_failure)) = Syncer::add_blocks(
             blockchain.clone(),
-            block_batch,
+            block_batch.clone(),
```
mpaulucci: can't we avoid cloning blocks? This is not cheap

crates/networking/p2p/sync.rs
```diff
@@ -661,20 +666,49 @@ impl FullBlockSyncState {
...
-                        batch_failure.last_valid_hash,
-                    )
-                    .await?;
+                // Since running the batch failed we set the failing block and it's descendants with having an invalid ancestor on the following cases.
+                match err {
+                    ChainError::InvalidBlock(_)
+                    | ChainError::InvalidTransaction(_)
+                    | ChainError::EvmError(EvmError::Transaction(_))
+                    | ChainError::EvmError(EvmError::Header(_))
+                    | ChainError::EvmError(EvmError::InvalidDepositRequest) => {
```
mpaulucci: why do we need to include this now? Doesn't seem related to the actual fix

crates/networking/p2p/sync.rs
```diff
@@ -661,20 +666,49 @@ impl FullBlockSyncState {
...
+                        for block in blocks_with_invalid_ancestor {
+                            self.store
+                                .set_latest_valid_ancestor(
+                                    block.hash(),
+                                    batch_failure.last_valid_hash,
+                                )
+                                .await?;
+                        }
+                        // We also set with having an invalid ancestor all the hashes remaining which are descendants as well.
+                        for header in self.current_headers.clone() {
```
mpaulucci: do we need to clone here?

crates/networking/p2p/sync.rs
cdiielsi: The thing is, originally this boolean is set as true [here](https://github.com/lambdaclass/ethrex/blob/99ebb5fc5ab0917feb61015d82cc39d74b16714b/crates/networking/p2p/sync.rs#L392-L393) if the sync_head is between the requested headers. The problem is when running the tests the sync_head is never sent when we request the headers ([here](https://github.com/lambdaclass/ethrex/blob/99ebb5fc5ab0917feb61015d82cc39d74b16714b/crates/networking/p2p/sync.rs#L346-L347)). But we have the sync_head as pending so the boolean should be true. 

I guess that's why the sync_head was added like [this](https://github.com/lambdaclass/ethrex/pull/1926/files#diff-06c504488a1699af75adf47638c44d8ed1b42d02898180cc0b64a29b9546e5dcR194) to `block_headers` before the big changes in syncing and p2p where introduced. To sum it up, the old code added the sync_head in `block_headers` and then did [this](https://github.com/lambdaclass/ethrex/blob/99ebb5fc5ab0917feb61015d82cc39d74b16714b/crates/networking/p2p/sync.rs#L392-L393), so sync_head would be set to true in case it was saved as a pending block.

Now, this boolean is used later when calling `Syncer::add_blocks`. So what I did is, move this checks on the pending block to the `process_incoming_headers` function because I already needed to retrieve the pending block there since when requesting the block bodies it never arrives, and I thought it would be better to have all this together. So that's why I set sync_head_found as true there. And that's why I pass this bool as mut.

crates/networking/p2p/Cargo.toml
cdiielsi: This is because I'm importing ethrex_vm::EvmError; to match specific errors that should result in marking a block with having an invalid ancestor.

crates/networking/p2p/peer_handler.rs
cdiielsi: I see, yeah it makes much more sense, I'm changing this!!

crates/networking/p2p/sync.rs
cdiielsi: When analysing a previous comment I realized that I can only deal with pending block [here](https://github.com/lambdaclass/ethrex/pull/4028/files#diff-06c504488a1699af75adf47638c44d8ed1b42d02898180cc0b64a29b9546e5dcR622) and even set the boolean as true there. So I'm dropping this lines. I remember the previous disposition depended on the while loop that was commented [here](https://github.com/lambdaclass/ethrex/blob/99ebb5fc5ab0917feb61015d82cc39d74b16714b/crates/networking/p2p/sync.rs#L603) with the big changes in sync and p2p.

crates/networking/p2p/sync.rs
cdiielsi: Because [here](https://github.com/lambdaclass/ethrex/blob/99ebb5fc5ab0917feb61015d82cc39d74b16714b/crates/networking/p2p/sync.rs#L689) I use the drain method on block_batch to get all blocks that come after the one that turned out invalid, to then mark them as invalid as well. I guess I could just get the index and work with that.

crates/networking/p2p/sync.rs
cdiielsi: The thing is I need the blocks in block_batch in case of error to mark them as invalid, but in truth I only need the block hashes, so I could only clone that, I'd still be cloning but it would be cheaper. Otherwise we could change the parameter's type in `add_block` as `&[Block]` instead of `Vec<Block>` since it's used as a reference in `add_block` but that change requires using references in UpdateBatch.

crates/networking/p2p/sync.rs
cdiielsi: I selected all the `ChainError`s that I though were relevant to mark a block as invalid. I was in doubt about ChainError::InvalidBlock would always cover all the cases through the block's execution.

crates/networking/p2p/sync.rs
cdiielsi: No! Sorry! I'm dropping it!

crates/networking/p2p/sync.rs
mpaulucci: I don't see how this is related to the original comment. You are commenting about functionality, whereas OP is talking about code style, i.e don't pass a mutable argument, but instead return it as a result.

crates/networking/p2p/sync.rs
mpaulucci: All the errors should be encompassed in `ChainError::InvalidBlock`. For example, if the block has an invalid transaction, it should still be `ChainError::InvalidBlock(<invalid transaction error>)` (this is just pseudocode). This might involve changing the upstream code.

crates/networking/p2p/sync.rs
mpaulucci: yes, please avoid cloning large objects like a vector of blocks, you can clone a vector of references for example, there should be many solutions to this.

crates/networking/p2p/Cargo.toml
mpaulucci: We should avoid adding dependencies between the crates when it's not strictly necessary. I think in this case it is not necessary, specially considering it was not necessary up until now.

crates/networking/p2p/sync.rs
mpaulucci: This is a nit, it's not extremelty important. But it hurts readability when you have a mutable object like this. So one reads the variable and thinks "this is the batch of blocks" but then suddently some blocks are removed from the vector, so it's still named "block batch" but in reality it is "block batch minus some blocks that I decided to remove for some reason".

crates/networking/p2p/sync.rs
cdiielsi: Done!

crates/networking/p2p/sync.rs
cdiielsi: I understand, after the changes I made to avoid cloning the vector, there's no longer need to for it to be mut, so there's no problem.

crates/networking/p2p/sync.rs
cdiielsi: Ok! I'll have to map the error [here](https://github.com/lambdaclass/ethrex/blob/122a76e8fd40f74bfe114867783ae2ae18b7b881/crates/blockchain/blockchain.rs#L139-L140) to `ChainError::InvalidTransaction(_)` and I think that would do it!

mpaulucci: It should still be `ChainError::InvalidBlock` if a transaction inside the block is invalid

cdiielsi: Fixed in this [commit](https://github.com/lambdaclass/ethrex/pull/4028/commits/63e7aaae36e67bfea9b5239230895e364614c880)!

PR #4029, opened 2025-08-12
feat(l1): start building payloads as soon as we get an fcu

### LINKED ISSUE #3920: Handle the block building asynchronously
The spec expects us to handle payload building asynchronously. That is, we should start building on `ForkChoiceUpdated` and then finish on `GetPayload`. Currently we only start building when we receive the `GetPayload method`, which may cause nodes to be sent as empty by consensus timeouts and other similar problems. We need to have a separate task that takes care of building payloads between this message.

fmoletta:
**Motivation**
* Start the block building process as soon as we get a payload from a FCU
* Only wait for the building process to finish when we get a `GetPayload` message
* Keep latest 10 payloads in memory instead of filling up the DB with old payloads
* Continuosly rebuild payload until requested/ slot time is up
* Only remove transactions from mempool when we execute a payload

**Description**
* Add `Blockchain` methods `initiate_payload_build` & `get_payload` to start & finish a payload build process
* Add `Blockchain` field `payloads` containing a vector of payload ids and payload building tasks/ built payloads (max 10 at a time)
* Remove any instance of payloads being stored in the DB

Closes #3920


crates/blockchain/blockchain.rs
```diff
@@ -665,6 +674,14 @@ impl Blockchain {
         self.mempool.remove_transaction(hash)
     }
 
+    /// Remove all transactions in the executed block from the pool (if we have them)
```
mpaulucci, line 702: Do we really need this? See this PR: https://github.com/lambdaclass/ethrex/pull/4029

It seems the place to remove mempool tx should be the FCU call

cmd/ethrex/bench/build_block_benchmark.rs
```diff
@@ -182,44 +180,29 @@ async fn fill_mempool(b: &Blockchain, accounts: Vec<SecretKey>) {
     }
 }
 
-pub async fn bench_payload(input: &(&mut Blockchain, Block, &Store)) -> (Duration, u64) {
```
mpaulucci, line 185: looking at the name of this function, I have no idea what it does. Maybe some doc would be useful.

cmd/ethrex/bench/build_block_benchmark.rs
```diff
@@ -182,44 +180,29 @@ async fn fill_mempool(b: &Blockchain, accounts: Vec<SecretKey>) {
...
-    let (b, genesis_block, store) = input;
+pub async fn bench_payload(input: &(Arc<Blockchain>, Block, &Store)) -> (Duration, u64) {
+    let (blockchain, genesis_block, store) = input;
+    let blockchain = std::sync::Arc::new(blockchain);
```
mpaulucci: isnt this equivalent to `blockchain.clone()`?

cmd/ethrex/bench/build_block_benchmark.rs
```diff
@@ -182,44 +180,29 @@ async fn fill_mempool(b: &Blockchain, accounts: Vec<SecretKey>) {
...
-            blobs_bundle,
-            block_value,
-            requests,
-            payload,
-            ..
-        } = b.build_payload(payload.block.clone()).await.unwrap();
-        (blobs_bundle, requests, block_value, payload)
+    // This method will wait for the building process to finish and retrieve the finalized block
+    let block = {
+        let PayloadBuildResult { payload, .. } = blockchain.get_payload(payload_id).await.unwrap();
```
mpaulucci, line 199: nit: avoid unwrap

crates/blockchain/blockchain.rs
mpaulucci: (this function itself is fine, the problem is the call above)

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,69 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
 }
 
 impl Blockchain {
+    /// Attempts to fetch a payload given it's id. If the payload is still being built, it will be finished.
```
mpaulucci, line 316: Waiting for the current payload to finish is how other clients deal with this? The alternative would be to return the last finished payload (and not risk a missing block).

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,69 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+        let self_clone = self.clone();
+        let cancel_token = CancellationToken::new();
+        let cancel_token_clone = cancel_token.clone();
+        let payload_build_task = tokio::task::spawn(async move {
+            self_clone
+                .build_payload_loop(payload, cancel_token_clone)
+                .await
+        });
+        let mut payloads = self.payloads.lock().await;
+        if payloads.len() == MAX_PAYLOADS {
```
mpaulucci: nit: use `>=`. The reason is that if because of some bug the vector goes above 10, it would potentially be able to go to infinity

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,69 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+    /// Build the given payload and keep on rebuilding it until either the time slot
+    /// given by `SECONDS_PER_SLOT` is up or the `cancel_token` is cancelled
+    pub async fn build_payload_loop(
+        self: Arc<Blockchain>,
+        payload: Block,
+        cancel_token: CancellationToken,
+    ) -> Result<PayloadBuildResult, ChainError> {
+        let start = Instant::now();
+        let self_clone = self.clone();
+        const SECONDS_PER_SLOT: Duration = Duration::from_secs(12);
```
mpaulucci, line 368: this should be taken from a config value, not all networks have 12 seconds block times.

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,69 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+        // Attempt to rebuild the payload as many times within the given timeframe to maximize fee revenue
+        let mut res = self_clone.build_payload(payload.clone()).await?;
+        while start.elapsed() < SECONDS_PER_SLOT && !cancel_token.is_cancelled() {
```
mpaulucci, line 371: nit: would it be worth it to check if there are any new txs in the mempool before potentially building the same payload?

crates/networking/rpc/engine/payload.rs
```diff
@@ -681,19 +676,15 @@ async fn try_execute_payload(
...
-        Err(ChainError::WitnessGeneration(e)) => {
+        Err(e) => {
             error!("{e} for block {block_hash}");
             Err(RpcErr::Internal(e.to_string()))
         }
         Ok(()) => {
+            // Remove executed transactions from mempool
+            context
+                .blockchain
+                .remove_block_transactions_from_pool(block)?;
```
mpaulucci: see https://github.com/lambdaclass/ethrex/pull/4029/files#r2301179232

crates/storage/store_db/libmdbx.rs
```diff
@@ -602,28 +601,6 @@ impl StoreEngine for Store {
             .map_err(StoreError::from)
     }
 
-    async fn add_payload(&self, payload_id: u64, block: Block) -> Result<(), StoreError> {
```
mpaulucci, line 634: nice!

cmd/ethrex/bench/build_block_benchmark.rs
Oppen: I believe this produces an `Arc<Arc<Blockchain>>` instead. Don't know why would want that tho, cloning the original `Arc<Blockchain>` should be better.

cmd/ethrex/bench/build_block_benchmark.rs
Oppen: For the benchmark `unwrap` should be OK and desirable. Errors make the numbers less reliable, so a crash is probably for the best. That said, maybe using `expect` is better.

cmd/ethrex/bench/build_block_benchmark.rs
fmoletta: I think I might have added it when solving merge conflicts, the line makes no sense, removed it

cmd/ethrex/bench/build_block_benchmark.rs
fmoletta: Would you prefer `bench_payload_build`?

crates/blockchain/payload.rs
fmoletta: Updated: [873aead]
We now cancel the current rebuild process and return the previous payload if it is requested before time is up

crates/blockchain/payload.rs
fmoletta: geth uses 12 seconds (hardcoded) https://github.com/ethereum/go-ethereum/blob/95ab643bb8acbe5a01c6699dab4fcd0d8d533293/miner/payload_building.go#L241

MegaRedHand: This won't be a problem unless the network has > 12 seconds block times, right?

cmd/ethrex/bench/build_block_benchmark.rs
```diff
@@ -261,7 +243,7 @@ pub fn build_block_benchmark(c: &mut Criterion<GasMeasurement>) {
 
                     (block_chain, genesis.get_block(), store_with_genesis)
                 };
-                let input = (&mut blockchain, genesis_block, &store);
+                let input = (Arc::new(blockchain), genesis_block, &store);
```
MegaRedHand, line 248: Should this be `blockchain.clone()`? `blockchain` is an `Arc`

cmd/ethrex/bench/build_block_benchmark.rs
```diff
@@ -182,44 +180,28 @@ async fn fill_mempool(b: &Blockchain, accounts: Vec<SecretKey>) {
...
+    // the RPC handling. The payload is then sent to `Blockchain` to initiate the payload building
+    // Blockchain::initiate_payload_build eventually calls 'fill_transactions'
+    // which should take transactions from the previously filled mempool
     let (payload_block, payload_id) = create_payload_block(genesis_block, store).await;
-    store
-        .add_payload(payload_id, payload_block.clone())
-        .await
-        .unwrap();
-    // 2. engine_getPayload is called, this code path ends up calling Store::get_payload(id),
+    <Arc<Blockchain> as Clone>::clone(blockchain)
```
MegaRedHand: Can we do `blockchain.clone()` here?

crates/blockchain/payload.rs
```diff
@@ -40,6 +42,39 @@ use crate::{
...
+    /// If self is already a `PayloadOrTask::Payload` this is a NoOp
+    pub async fn to_payload(self) -> Result<Self, ChainError> {
+        Ok(match self {
+            PayloadOrTask::Payload(_) => self,
+            PayloadOrTask::Task(task) => PayloadOrTask::Payload(Box::new(task.finish().await?)),
+        })
+    }
+}
+
 pub struct BuildPayloadArgs {
```
MegaRedHand, line 78: Let's derive `Debug` for this

crates/blockchain/blockchain.rs
```diff
@@ -65,6 +69,9 @@ pub struct Blockchain {
     /// This does not reflect whether there is an ongoing sync process
     is_synced: AtomicBool,
     pub r#type: BlockchainType,
+    /// Mapping from a payload id to either a complete payload or a payload build task
+    /// We need to keep completed payloads around in case consensus requests them twice
+    pub payloads: Arc<TokioMutex<Vec<(u64, PayloadOrTask)>>>,
```
MegaRedHand, line 76: Can we use a `BTreeMap<u64, PayloadOrTask>` instead? For cleaning up, we could just spawn a timer that removes the payload after 12s and remove the limit on the number of payloads.

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,75 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+            .iter()
+            .position(|(id, _)| id == &payload_id)
+            .ok_or(ChainError::UnknownPayload)?;
+        let finished_payload = (payload_id, payloads.remove(idx).1.to_payload().await?);
+        payloads.insert(idx, finished_payload);
+        // Return the held payload
+        match &payloads[idx].1 {
+            PayloadOrTask::Payload(payload) => Ok(*payload.clone()),
+            // We already converted the payload into a finished version
+            _ => unreachable!(),
```
MegaRedHand:
```suggestion
            _ => unreachable!("we already converted the payload into a finished version"),
```

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,75 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+        }
+    }
+
+    /// Starts a payload build process. The built payload can be retrieved by calling `get_payload`.
+    /// The build process will run for the full block building timeslot or until `get_payload` is called
+    pub async fn initiate_payload_build(self: Arc<Blockchain>, payload: Block, payload_id: u64) {
```
MegaRedHand, line 336: This can receive a reference:

```suggestion
    pub async fn initiate_payload_build(self: &Arc<Blockchain>, payload: Block, payload_id: u64) {
```

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,75 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+        let payload_build_task = tokio::task::spawn(async move {
+            self_clone
+                .build_payload_loop(payload, cancel_token_clone)
+                .await
+        });
+        let mut payloads = self.payloads.lock().await;
+        if payloads.len() >= MAX_PAYLOADS {
+            // Remove oldest unclaimed payload
+            payloads.remove(0);
+        }
```
MegaRedHand, line 349: This could remove ongoing building tasks.

cmd/ethrex/bench/build_block_benchmark.rs
mpaulucci: Ah, I didn't realize we were in a benchmark file

crates/blockchain/payload.rs
```diff
@@ -277,6 +313,75 @@ impl From<PayloadBuildContext> for PayloadBuildResult {
...
+
+    /// Build the given payload and keep on rebuilding it until either the time slot
+    /// given by `SECONDS_PER_SLOT` is up or the `cancel_token` is cancelled
+    pub async fn build_payload_loop(
+        self: Arc<Blockchain>,
+        payload: Block,
+        cancel_token: CancellationToken,
+    ) -> Result<PayloadBuildResult, ChainError> {
+        let start = Instant::now();
+        let self_clone = self.clone();
```
MegaRedHand, line 367: We don't need this clone

crates/blockchain/payload.rs
fmoletta: Is that a problem?

crates/blockchain/payload.rs
fmoletta: Arc is already a reference, what do we gain from adding another reference on top of it?

crates/blockchain/payload.rs
fmoletta: We do need it for the tokio task to take ownership

cmd/ethrex/bench/build_block_benchmark.rs
fmoletta: Sure

crates/blockchain/payload.rs
fmoletta: We haven't needed it until now, I don't see how this PR should add it

cmd/ethrex/bench/build_block_benchmark.rs
fmoletta: blockchain is not an Arc in this context

crates/blockchain/blockchain.rs
fmoletta: Spawning a timer for each payload once it is finished in order to remove it sounds a lot more complicated than just using a Vec and checking its size. It also doesn't prevent the node from holding on to a high number of payloads if it gets spammed with fcu which can be a problem now that payloads are held in memory

crates/blockchain/blockchain.rs
fmoletta: I figured `NewPayload` was a better place as we already have the block and we don't need to fetch it, but we could also do it at `FCU`.

fmoletta: Updated!

crates/blockchain/payload.rs
fmoletta: Could be overkill? We would have to have a `last_modified` timestamp and lock it for each mempool interaction

MegaRedHand (approved):

crates/blockchain/blockchain.rs
Oppen: We could have a single task doing the cleanup every 12 seconds though, as long as it has the necessary information to make decisions.

PR #4033, opened 2025-08-12
fix(l1): byte comparison with usize wrong in 32 bit architecture

JereSalo:
**Motivation**

**Description**

- We were considering usize as u64, that's usually correct except for the case of a zkVM using 32 bit RISC-V.

### DIFF
--- a/crates/common/types/requests.rs
+++ b/crates/common/types/requests.rs
@@ -174,7 +174,6 @@ impl Deposit {
             INDEX_SIZE,
         ];
 
-        // Compare two numbers with different byte count without padding
         // Validate Offsets & Sizes
         for (i, (expected_offset, expected_size)) in
             OFFSETS.into_iter().zip(SIZES.into_iter()).enumerate()
--- a/crates/common/types/requests.rs
+++ b/crates/common/types/requests.rs
@@ -132,26 +132,26 @@ impl Deposit {
             return None;
         }
 
-        //Encoding scheme:
+        // Encoding scheme:
         //
-        //positional arguments -> 5 parameters with uint256 positional value for each -> 160b
-        //pub_key: 32b of len + 48b padded to 64b
-        //withdrawal_credentials: 32b of len + 32b
-        //amount: 32b of len + 8b padded to 32b
-        //signature: 32b of len + 96b
-        //index: 32b of len + 8b padded to 32b
+        // positional arguments -> 5 parameters with uint256 positional value for each -> 160b
+        // pub_key: 32b of len + 48b padded to 64b
+        // withdrawal_credentials: 32b of len + 32b
+        // amount: 32b of len + 8b padded to 32b
+        // signature: 32b of len + 96b
+        // index: 32b of len + 8b padded to 32b
         //
-        //-> Total len: 576 bytes
+        // -> Total len: 576 bytes
 
         const OFFSET_LEN: usize = 32;
 
-        const PUB_KEY_OFFSET: usize = 160;
-        const WITHDRAWAL_CREDENTIALS_OFFSET: usize = 256;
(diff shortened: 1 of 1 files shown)


crates/common/types/requests.rs
```diff
@@ -177,7 +177,9 @@ impl Deposit {
 
         // Compare two numbers with different byte count without padding
         let is_eq = |bytes: [u8; 32], u: usize| -> bool {
-            bytes[..24] == [0; 24] && bytes[24..] == u.to_be_bytes()
+            let be = u.to_be_bytes(); // 4 bytes on rv32, 8 on rv64/x86_64
+            let k = 32 - be.len(); // 28 on rv32, 24 on 64-bit
+            bytes[..k].iter().all(|&b| b == 0) && bytes[k..] == be
```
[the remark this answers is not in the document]
JereSalo: nah I'll stick to the original. Thanks though

ilitteri (approved):

crates/common/types/requests.rs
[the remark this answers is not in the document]
JereSalo: > "The second form pays an extra bill: it allocates Vec<u8> of length k and fills it with zeros before doing the compare. That’s O(k) work (memset) plus allocator overhead, on top of the equality check. The first form just walks the existing slice. On modern Rust/LLVM, that all(|&b| b == 0) loop gets vectorized nicely, so you’re usually memory-bandwidth bound either way—except the second one does more memory traffic and a heap call."

LeanSerra (approved):

crates/common/types/requests.rs
```diff
@@ -177,7 +177,9 @@ impl Deposit {
 
         // Compare two numbers with different byte count without padding
         let is_eq = |bytes: [u8; 32], u: usize| -> bool {
```
Oppen: Just use `u32`. Change in the argument and the constant types. Also, note all sizes happen to be under one byte, so even `u8` and last byte check would be valid. I'd leave `u32` in case that changes in the future.

JereSalo: Good call, I left it as u32. It's kind of a tradeoff because we have to cast to usize in many cases but at least it removes complexity there.
https://github.com/lambdaclass/ethrex/pull/4033/commits/cd61e738c71422a45898023899dd779a02b1d283

crates/common/types/requests.rs
```diff
@@ -177,7 +177,9 @@ impl Deposit {
...
-            bytes[..24] == [0; 24] && bytes[24..] == u.to_be_bytes()
+            let be = u.to_be_bytes(); // 4 bytes on rv32, 8 on rv64/x86_64
+            let k = 32 - be.len(); // 28 on rv32, 24 on 64-bit
```
ilitteri: Use a constant for this value

ilitteri (changes requested):
Blocking until the use of hardcoded constants in this PR is fixed.

crates/common/types/requests.rs
```diff
@@ -176,32 +176,37 @@ impl Deposit {
         ];
 
         // Compare two numbers with different byte count without padding
```
fmoletta: This comment should be removed along with the code below it

JereSalo:
> Blocking until the use of hardcoded constants in this PR is fixed.

Everything has been addressed, code is a little big longer but is more detailed too. Now for comparison we parse to u32 to make it more clear too.

crates/common/types/requests.rs
JereSalo: https://github.com/lambdaclass/ethrex/pull/4033/commits/33881dea6d03a4ecf105a933b084c655e9111f20

PR #4040, opened 2025-08-13
fix(l1,levm): usize and u64 interactions

JereSalo:
**Motivation**

**Description**

- on `u256_to_usize` function we cast to usize and if the value is greater than 32 bits like 1 << 34 it would cast the wrong number. Not a problem on 64 bit architecture but it is a problem in 32 bit ones.
- On requests consider that the usize can be 32 bits, use u32 when corresponding

I don't know if this will be merged, right now it's being used mostly for testing and it's useful to have the CI run in here to have a quick view if it breaks something or not

Closes #issue_number


JereSalo:
Closing because this branch was used for testing purposes and not for merging.

PR #4043, opened 2025-08-13
feat(l1): add more block execution breakdown metrics.

mpaulucci:
**Motivation**
To know how much time it takes to read accounts/storage.

**Description**
Also fixed the store DB job not working properly


crates/storage/store_db/libmdbx.rs
```diff
@@ -130,6 +130,7 @@ impl StoreEngine for Store {
     async fn apply_updates(&self, update_batch: UpdateBatch) -> Result<(), StoreError> {
         let db = self.db.clone();
         tokio::task::spawn_blocking(move || {
+            let _span = tracing::trace_span!("Block DB update").entered();
```
mpaulucci, line 133: the problem with this is that if you put the instrumentation outside the spawn blocking, it doesn't work. Wish there was a less hacky way to fix this.

JereSalo (approved):

cdiielsi (approved):

PR #4045, opened 2025-08-14
perf(levm): use malachite for modexp

edg-l:
Use malachite library to improve the performance of the modexp precompile. There are speedups up to 7-8x observed in the new benchmarks.


Arkenan (approved):
LGTM great PR

crates/vm/levm/Cargo.toml
```diff
@@ -27,7 +27,8 @@ p256 = { version = "0.13.2", features = [
 ] }
 sha2 = "0.10.8"
 ripemd = "0.1.3"
-num-bigint = "0.4.5"
+num-bigint = "0.4.6"
```
MegaRedHand, line 30: Do we still need `num-bigint`?

Arkenan: We'll check it in a separate PR, but it might be the case.

PR #4046, opened 2025-08-14
perf(levm): use rug instead of big int to peform modexp

Arkenan:
This is a ~100x for modexp precompile.


edg-l:
Looks like since rug uses gmp which is a C dep the TDX from l2 struggles to build

edg-l:
Looks like riscv0 target doesnt support it

 thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gmp-mpfr-sys-1.6.5/build.rs:105:9:
Cross compilation from x86_64-unknown-linux-gnu to riscv32im-risc0-zkvm-elf not supported! Use the `force-cross` feature to cross compile anyway.

crates/vm/levm/src/precompiles.rs
```diff
@@ -8,6 +8,7 @@ use ethrex_common::{
     Address, H160, H256, U256, kzg::verify_kzg_proof, serde_utils::bool, types::Fork,
     utils::u256_from_big_endian,
 };
+use rug::Integer;
```
ElFantasma: `rug::Integer` is a **signed** arbitrary-precision integer (wraps GMP's `mpz_t`), while `malachite::Natural` was unsigned. In this use case it's safe because `Integer::from_digits` always produces a non-negative value, but it's worth being aware of the semantic difference — e.g., `to_digits()` operates on the absolute value, so a negative result from some arithmetic would silently lose its sign.

Consider adding a brief comment near the `mod_exp` call or at the import noting that all values in this context are non-negative.

crates/vm/levm/src/precompiles.rs
```diff
@@ -434,15 +422,14 @@ fn get_slice_or_default(
...
-    if modulus == Natural::ZERO {
-        Natural::ZERO
-    } else if exponent == Natural::ZERO {
-        Natural::from(1_u8) % modulus
+fn mod_exp(base: Integer, exponent: Integer, modulus: Integer) -> Integer {
+    if modulus == Integer::ZERO {
+        Integer::ZERO
+    } else if exponent == Integer::ZERO {
+        Integer::from(1_u8) % modulus
     } else {
```
ElFantasma: The `.unwrap()` is safe here: `pow_mod` only returns `Err` when the exponent is negative and the base has no modular inverse (GMP's `mpz_invert` fails). Since all values come from `Integer::from_digits` (always non-negative), this can't happen.

However, note a subtle issue: when `modulus == 0`, GMP's `mpz_powm` will **abort the process** (divide-by-zero in C), not return `Err`. The `modulus == Integer::ZERO` guard above prevents this, but the `#[expect]` reason should mention that the zero-modulus case is handled above, not just that "modulus is always positive" — since this function could be called with a zero modulus if the guard were ever removed.

MegaRedHand:
We should revisit this

PR #4048, opened 2025-08-14
fix(l2): disable gpu in `make lint` target

MegaRedHand:
**Motivation**

`make lint` was failing with:

```text
warning: risc0-sys@1.4.0: Compiler family detection failed due to error: ToolNotFound: failed to find tool "nvcc": No such file or directory (os error 2)
error: failed to run custom build command for `risc0-sys v1.4.0`

Caused by:
  process didn't exit successfully: `/Users/mega/execution/ethrex/target/release/build/risc0-sys-d088ac4aaf714579/build-script-build` (exit status: 1)
  --- stdout
...(a lot of lines)

  --- stderr


  error occurred in cc-rs: failed to find tool "nvcc": No such file or directory (os error 2)
```

**Description**

This PR removes the `--all-features` flag and replaces it with `-F` and a list of all features minus the "gpu" one that caused the problem (`--all-features` is kept for L2 lint workflow). I got the list of features with `cargo add -n ethrex`.


mpaulucci (approved):

.github/workflows/pr-main_l1.yaml
ManuelBilbao, line 1: We could remove nvcc install now

.github/workflows/pr-main_l2.yaml
ManuelBilbao, line 1: Same here

Arkenan (approved):

.github/workflows/pr-main_l1.yaml
MegaRedHand: Removed

.github/workflows/pr-main_l2.yaml
MegaRedHand: Removed

PR #4049, opened 2025-08-14
Cache tx hash

rodrigo-o:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -80,7 +80,7 @@ impl Mempool {
                 self.blobs_bundle_pool
                     .lock()
                     .map_err(|error| StoreError::Custom(error.to_string()))?
-                    .remove(&tx.compute_hash());
+                    .remove(&tx.hash());
             }
 
             self.txs_by_sender_nonce
@@ -283,7 +283,7 @@ impl Mempool {
         nonce: u64,
         tx: &Transaction,
     ) -> Result<Option<H256>, MempoolError> {
-        let Some(tx_in_pool) = self.contains_sender_nonce(sender, nonce, tx.compute_hash())? else {
+        let Some(tx_in_pool) = self.contains_sender_nonce(sender, nonce, tx.hash())? else {
             return Ok(None);
         };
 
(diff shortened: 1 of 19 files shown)


crates/blockchain/mempool.rs
```diff
@@ -80,7 +80,7 @@ impl Mempool {
                 self.blobs_bundle_pool
                     .lock()
                     .map_err(|error| StoreError::Custom(error.to_string()))?
-                    .remove(&tx.compute_hash());
```
Oppen, line 83: This should not be computing the hash again in the first place, given it was the input to the function.

crates/blockchain/payload.rs
```diff
@@ -410,25 +410,22 @@ impl Blockchain {
...
-                    "Skipping transaction: {}, no gas left",
-                    head_tx.tx.compute_hash()
-                );
+                debug!("Skipping transaction: {}, no gas left", head_tx.tx.hash());
                 // We don't have enough gas left for the transaction, so we skip all txs from this account
                 txs.pop();
                 continue;
             }
 
             // TODO: maybe fetch hash too when filtering mempool so we don't have to compute it here (we can do this in the same refactor as adding timestamp)
```
Oppen, line 419: This was great advice. I think a later iteration should consider it.

crates/blockchain/mempool.rs
rodrigo-o: Good catch!

crates/blockchain/payload.rs
rodrigo-o: Agree

crates/blockchain/mempool.rs
rodrigo-o: Done here: [remove unnecesary call to tx.hash()](https://github.com/lambdaclass/ethrex/pull/4049/commits/f431488687366b41973eb56dae91d2bc5cf4889f)

crates/blockchain/payload.rs
```diff
@@ -410,25 +410,22 @@ impl Blockchain {
...
-            let tx_hash = head_tx.tx.compute_hash();
```
Oppen, line 423: This could be moved before the `gas_limit` check and the value used in all later appearances.

rodrigo-o: Done here: [Reuse already set variable](https://github.com/lambdaclass/ethrex/pull/4049/commits/005488384f1270b1b28b3cb02cb1b40aa9a614ea). As we discussed async just the reuse not the moving!

PR #4058, opened 2025-08-15
fix: redb feature

### LINKED ISSUE #4038: Use redb when redb feature is enabled
Currently the store engine is chosen like this:

```rust
pub fn open_store(data_dir: &str) -> Store {
    let path = PathBuf::from(data_dir);
    if path.ends_with("memory") {
        Store::new(data_dir, EngineType::InMemory).expect("Failed to create Store")
    } else {
        cfg_if::cfg_if! {
            if #[cfg(feature = "libmdbx")] {
                let engine_type = EngineType::Libmdbx;
            } else if #[cfg(feature = "redb")] {
                let engine_type = EngineType::RedB;
            } else {
                error!("No database specified. The feature flag `redb` or `libmdbx` should've been set while building.");
                panic!("Specify the desired database engine.");
            }
        }
        Store::new(data_dir, engine_type).expect("Failed to create Store")
    }
}

varun-doshi:
**Motivation**

Since libmdbx is a default feature, it should be afterwards in the if than redb, because redb is not enabled by default, so if a user enables it it, it probably wants to use it. in this case the user would need to disable all features and enable redb and other features. Changing the order allows the user to simply enable the redb feature without disabling all features

**Description**

the code now first checks if `redb` is enabled and as a result user does not have to disable other defaults like `libmdbx`

Closes #4038

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -100,10 +100,10 @@ pub fn open_store(data_dir: &str) -> Store {
         Store::new(data_dir, EngineType::InMemory).expect("Failed to create Store")
     } else {
         cfg_if::cfg_if! {
-            if #[cfg(feature = "libmdbx")] {
-                let engine_type = EngineType::Libmdbx;
-            } else if #[cfg(feature = "redb")] {
+          if #[cfg(feature = "redb")] {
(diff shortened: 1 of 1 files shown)


MegaRedHand:
redb was removed from the repo

PR #4063, opened 2025-08-18
refactor(l2): remove redb and libmdbx from rollup store

### LINKED ISSUE #3743: L2: remove `redb` and `libmdbx` from rollup store
In our rollup store, we currently support 3 dbs, but we are only using `SQL`, as it can be inspected it with simple queries which is useful for debugging and we don't have a need for a key value store in the rollup. Supporting 3 dbs triplicates the effort, since we have to update all of them for every minor change. This was originally implemented to replicate the L1 Store, but since we haven't used `libmdbx` or `redb` in a while, we can remove them now.

avilagaston9:
**Motivation**

We are not using `libmdbx` and `redb` for the rollup store and maintaining them is a waste of time.

**Description**

Removes `libmdbx` and `redb` from the rollup store.

Closes #3743


Makefile
```diff
@@ -11,7 +11,7 @@ build: ## 🔨 Build the client
 lint: ## 🧹 Linter check
 	# Note that we are compiling without the "gpu" feature (see #4048 for why)
 	# To compile with it you can replace '-F' with '--all-features', but you need to have nvcc installed
-	cargo clippy --all-targets -F debug,redb,risc0,rollup_storage_libmdbx,rollup_storage_redb,sp1,sync-test \
+	cargo clippy --all-targets -F debug,redb,risc0,sp1,sync-test \
```
MegaRedHand, line 14: Shouldn't this include `rollup_storage_sql`?

LeanSerra: I don't think so because `rollup_storage_sql` is a default feature

PR #4070, opened 2025-08-19
perf(levm): improve perf of calldataload, calldatacopy,  extcodecopy, codecopy, returndatacopy

edg-l:
**Motivation**

improve perf of calldataload, calldatacopy, extcodecopy, codecopy, returndatacopy

(left is this pr, right is main)

image

### DIFF
--- a/crates/vm/levm/src/opcode_handlers/environment.rs
+++ b/crates/vm/levm/src/opcode_handlers/environment.rs
@@ -204,7 +204,7 @@ impl<'a> VM<'a> {
 
         // Happiest fast path, copy without an intermediate buffer because there is no need to pad 0s and also size doesn't overflow.
         if let Some(code_offset_end) = code_offset.checked_add(size) {
-            if code_offset_end < current_call_frame.bytecode.len() {
+            if code_offset_end <= current_call_frame.bytecode.len() {
                 #[expect(unsafe_code, reason = "bounds checked beforehand")]
                 let slice = unsafe {
                     current_call_frame
--- a/crates/vm/levm/src/opcode_handlers/environment.rs
+++ b/crates/vm/levm/src/opcode_handlers/environment.rs
@@ -102,19 +102,7 @@ impl<'a> VM<'a> {
         let mut data = [0u8; 32];
         let size = 32;
 
-        // Happy fast path
-        if let Some(calldata_offset_end) = offset.checked_add(size) {
-            if calldata_offset_end <= current_call_frame.calldata.len() {
-                #[expect(unsafe_code, reason = "bounds checked beforehand")]
-                let slice = unsafe {
-                    current_call_frame
-                        .calldata
-                        .get_unchecked(offset..calldata_offset_end)
-                };
-
-                data.copy_from_slice(slice);
-            }
-        } else if offset < current_call_frame.calldata.len() {
+        if offset < current_call_frame.calldata.len() {
             let diff = current_call_frame.calldata.len().wrapping_sub(offset);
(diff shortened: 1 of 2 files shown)


crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -149,19 +152,37 @@ impl<'a> VM<'a> {
             return Ok(OpcodeResult::Continue { pc_increment: 1 });
         }
 
-        let mut data = vec![0u8; size];
+        // Happiest fast path, copy without an intermediate buffer because there is no need to pad 0s and also size doesn't overflow.
+        if let Some(calldata_offset_end) = calldata_offset.checked_add(size) {
+            if calldata_offset_end <= current_call_frame.calldata.len() {
+                #[expect(unsafe_code, reason = "bounds checked beforehand")]
+                let slice = unsafe {
```
MegaRedHand, line 159: I'm curious. How much does the unsafe contribute to the performance increase?

crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -149,19 +152,37 @@ impl<'a> VM<'a> {
             return Ok(OpcodeResult::Continue { pc_increment: 1 });
         }
 
-        let mut data = vec![0u8; size];
+        // Happiest fast path, copy without an intermediate buffer because there is no need to pad 0s and also size doesn't overflow.
```
MegaRedHand, line 155: I'm not sure we need to special-case this, since we could just try to copy `offset..diff` first and then fill the rest with zeros, which we'd skip for this case.

crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -100,17 +100,20 @@ impl<'a> VM<'a> {
...
-            .iter()
-            .skip(offset)
-            .take(32)
-            .enumerate()
-        {
-            if let Some(data_byte) = data.get_mut(i) {
-                *data_byte = *byte;
+        let size = 32;
+
+        if offset < current_call_frame.calldata.len() {
```
MegaRedHand, line 105: We should move byte-copying logic to a helper. We have the same logic in all related opcodes.

MegaRedHand (approved):
Left some comments. Since the performance increases are substantial, we can merge as-is and address those comments afterwards.

crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -149,19 +152,37 @@ impl<'a> VM<'a> {
...
+                };
+                current_call_frame.memory.store_data(dest_offset, slice)?;
+
+                return Ok(OpcodeResult::Continue { pc_increment: 1 });
+            }
+        }
 
-        if calldata_offset <= current_call_frame.calldata.len() {
-            for (i, byte) in current_call_frame
+        let mut data = vec![0u8; size];
```
Oppen, line 170: Isn't size bounded to 32? Maybe use an array instead to avoid the allocation.

Oppen: Ah no, this one is variable. I thought it was the same function as above.

Oppen (approved):

crates/vm/levm/src/opcode_handlers/environment.rs
MegaRedHand: That's only the case for `CALLDATALOAD`. This is for `CALLDATACOPY` which takes an arbitrary size.

PR #4071, opened 2025-08-19
perf(levm): improve signextend opcode performance

edg-l:
**Motivation**

Removed some unnecesary double checks.

(left is pr, right is main)
image


ManuelBilbao (approved):

crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -226,27 +225,24 @@ impl<'a> VM<'a> {
...
         #[expect(clippy::arithmetic_side_effects)]
-        let shifted_value = value_to_extend >> sign_bit_index;
-        let sign_bit = shifted_value & U256::one();
-
-        let sign_bit_mask = checked_shift_left(U256::one(), sign_bit_index)?
-            .checked_sub(U256::one())
-            .ok_or(InternalError::Underflow)?; //Shifted should be at least one
+        let sign_bit = (value_to_extend >> sign_bit_index) & U256::one();
+        #[expect(clippy::arithmetic_side_effects)]
+        let mask = (U256::one() << sign_bit_index) - U256::one();
```
Oppen: Just add a `reason` to the  `expect` clause. You can also make a block if you want to avoid having to write it twice:
```rust
        #[expect(clippy::arithmetic_side_effects, reason = "sign_bit_index max value is 31 * 8 + 7 = 255, which can't overflow")]
        let (sign_bit, mask) = (
            (value_to_extend >> sign_bit_index) & U256::one(),
            (U256::one() << sign_bit_index) - U256::one()
        );
```

Oppen: You could include the `sign_bit_index` part as well, as it's all more or less same lint and same reasons why it's safe.

Oppen: Another thing: maybe just use a const table for the masks? It's at most 255 AFAICT. Or really 31 if the argument is always full bytes. It might be worth checking.

PR #4072, opened 2025-08-19
perf(levm): improve mulmod and addmod opcodes

edg-l:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/vm/levm/src/opcode_handlers/arithmetic.rs
+++ b/crates/vm/levm/src/opcode_handlers/arithmetic.rs
@@ -148,22 +148,25 @@ impl<'a> VM<'a> {
         let [augend, addend, modulus] = *current_call_frame.stack.pop()?;
 
         if modulus.is_zero() {
-            current_call_frame.stack.push(&[U256::zero()])?;
+            current_call_frame.stack.push1(U256::zero())?;
             return Ok(OpcodeResult::Continue { pc_increment: 1 });
         }
 
         let new_augend: U512 = augend.into();
         let new_addend: U512 = addend.into();
 
-        let sum = new_augend
-            .checked_add(new_addend)
-            .ok_or(InternalError::Overflow)?;
+        #[allow(
+            clippy::arithmetic_side_effects,
+            reason = "both values come from a u256, so the product can fit in a U512"
+        )]
+        let sum = new_augend + new_addend;
+        #[allow(clippy::arithmetic_side_effects, reason = "can't overflow")]
+        let sum_mod = sum % modulus;
 
-        let sum_mod = sum
-            .checked_rem(modulus.into())
-            .ok_or(InternalError::Overflow)?
+        #[allow(clippy::expect_used, reason = "can't overflow")]
+        let sum_mod: U256 = sum_mod
             .try_into()
-            .map_err(|_err| InternalError::Overflow)?;
+            .expect("can't fail because we applied % mod where mod is a U256 value");
 
         current_call_frame.stack.push1(sum_mod)?;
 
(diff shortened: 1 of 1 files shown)


crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -148,22 +148,25 @@ impl<'a> VM<'a> {
         let [augend, addend, modulus] = *current_call_frame.stack.pop()?;
 
         if modulus.is_zero() {
-            current_call_frame.stack.push(&[U256::zero()])?;
+            current_call_frame.stack.push1(U256::zero())?;
```
Oppen, line 150: Wasn't there an optimized `push0` too?

crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -148,22 +148,25 @@ impl<'a> VM<'a> {
...
 
-        let sum = new_augend
-            .checked_add(new_addend)
-            .ok_or(InternalError::Overflow)?;
+        #[allow(
+            clippy::arithmetic_side_effects,
+            reason = "both values come from a u256, so the product can fit in a U512"
+        )]
+        let sum = new_augend + new_addend;
+        #[allow(clippy::arithmetic_side_effects, reason = "can't overflow")]
```
Oppen: I think reason here should be `non-zero modulus`, given the side-effect would be division by zero and it's checked in line 150.

crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -148,22 +148,25 @@ impl<'a> VM<'a> {
...
+        let sum_mod = sum % modulus;
 
-        let sum_mod = sum
-            .checked_rem(modulus.into())
-            .ok_or(InternalError::Overflow)?
+        #[allow(clippy::expect_used, reason = "can't overflow")]
+        let sum_mod: U256 = sum_mod
             .try_into()
-            .map_err(|_err| InternalError::Overflow)?;
+            .expect("can't fail because we applied % mod where mod is a U256 value");
```
Oppen, line 171: I think it's worth checking if using `unwrap_unchecked` gives a measurable improvement in this case. If it does, I'd replace the expect with that.

crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -185,14 +188,18 @@ impl<'a> VM<'a> {
...
+        )]
+        let product = multiplicand * multiplier;
+        #[allow(clippy::arithmetic_side_effects, reason = "can't overflow")]
+        let product_mod = product % modulus;
+
+        #[allow(clippy::expect_used, reason = "can't overflow")]
+        let product_mod: U256 = product_mod
             .try_into()
-            .map_err(|_err| InternalError::Overflow)?;
+            .expect("can't fail because we applied % mod where mod is a U256 value");
```
Oppen, line 204: Both comments apply to this block as well.

Oppen (approved):

crates/vm/levm/src/opcode_handlers/arithmetic.rs
edg-l: not that i know

crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -147,22 +147,28 @@ impl<'a> VM<'a> {
...
-            .checked_add(new_addend)
-            .ok_or(InternalError::Overflow)?;
+        #[allow(
+            clippy::arithmetic_side_effects,
+            reason = "both values come from a u256, so the product can fit in a U512"
+        )]
+        let sum = new_augend + new_addend;
+        #[allow(
+            clippy::arithmetic_side_effects,
+            reason = "can't overflow because non-zero modulus"
```
Oppen: This case was never about overflow, but avoid division by zero. The closest one-word description I could think of is "trap":
```suggestion
            reason = "can't trap because non-zero modulus"
```

JereSalo (approved):

PR #4073, opened 2025-08-19
perf(levm): improve bls12 precompile

edg-l:
**Motivation**

Improves 

- g1 msm from 18 mgas/s to 100
- g2 msm from 14 mgas/s to 143

Closes #issue_number


crates/vm/levm/src/precompiles.rs
```diff
@@ -948,28 +949,30 @@ pub fn bls12_g1msm(calldata: &Bytes, gas_remaining: &mut u64) -> Result<Bytes, V
...
-            .ok_or(InternalError::Overflow)?;
-        let scalar_offset = point_offset
-            .checked_add(128)
-            .ok_or(InternalError::Overflow)?;
-        let pair_end = scalar_offset
-            .checked_add(32)
-            .ok_or(InternalError::Overflow)?;
+        let point_offset = i * BLS12_381_G1_MSM_PAIR_LENGTH;
+        let scalar_offset = point_offset + 128;
+        let pair_end = scalar_offset + 32;
```
MauroToscano, line 973: What's the reason why this can never overflow ? It should be added in a comment

crates/vm/levm/src/precompiles.rs
```diff
@@ -948,28 +949,30 @@ pub fn bls12_g1msm(calldata: &Bytes, gas_remaining: &mut u64) -> Result<Bytes, V
...
 
-        let point = parse_g1_point(calldata.get(point_offset..scalar_offset), false)?;
-        let scalar = parse_scalar(calldata.get(scalar_offset..pair_end))?;
+        let point = parse_g1_point(&calldata[point_offset..scalar_offset], false)?;
+        let scalar = parse_scalar(&calldata[scalar_offset..pair_end])?;
 
-        let scaled_point = G1Projective::mul(point, scalar);
-        result = result.add(&scaled_point);
+        if scalar != Scalar::zero() {
```
```suggestion
        if !scalar.is_zero() {
```

crates/vm/levm/src/precompiles.rs
```diff
@@ -1180,46 +1192,50 @@ pub fn bls12_map_fp2_tp_g2(calldata: &Bytes, gas_remaining: &mut u64) -> Result<
...
-        ))?;
-    unpadded_coordinate
+
+    if coordinate_raw_bytes[0..16] != SIXTEEN_ZEROES {
+        return Err(PrecompileError::ParsingInputError.into());
+    }
+
+    coordinate_raw_bytes[16..64]
         .try_into()
         .map_err(|_| PrecompileError::ParsingInputError.into())
```
Oppen: Given the check at the beginning, this is always successful. Wouldn't an expect o even unwrap_unchecked be worth trying?

crates/vm/levm/src/precompiles.rs
```diff
@@ -1256,91 +1272,95 @@ fn parse_g1_point(
...
-                .into_option()
-                .ok_or(PrecompileError::ParsingInputError)?;
-
-            G2Projective::from(g2_affine)
-        }
-    };
+            // The crate serialize the coordinates in a reverse order
+            // https://docs.rs/bls12_381/0.8.0/src/bls12_381/g2.rs.html#401-464
+            let g2_bytes: [u8; 192] = [x_1, x_0, y_1, y_0]
+                .concat()
```
Oppen: Careful with `concat` as it introduces a vector allocation.

crates/vm/levm/src/precompiles.rs
```diff
@@ -1256,91 +1272,95 @@ fn parse_g1_point(
...
-    let scalar_bytes: [u8; 32] = scalar_raw_bytes
-        .ok_or(InternalError::Slicing)?
+#[inline]
+fn parse_scalar(scalar_raw_bytes: &[u8]) -> Result<Scalar, VMError> {
+    if scalar_raw_bytes.len() != 32 {
+        return Err(PrecompileError::ParsingInputError.into());
+    }
+
+    let bytes: [u8; 32] = scalar_raw_bytes
         .try_into()
```
Oppen: This is infallible at this point.

crates/vm/levm/src/precompiles.rs
```diff
@@ -1256,91 +1272,95 @@ fn parse_g1_point(
...
+        ]),
+        u64::from_be_bytes([
+            bytes[16], bytes[17], bytes[18], bytes[19], bytes[20], bytes[21], bytes[22], bytes[23],
+        ]),
+        u64::from_be_bytes([
+            bytes[8], bytes[9], bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15],
+        ]),
+        u64::from_be_bytes([
+            bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7],
+        ]),
```
Oppen: Given you're directly accessing the bytes by index, why not just use `scalar_raw_bytes` here? I believe the only reason the copy was needed before was the call to `reverse` which requires mutability.

Oppen:
LGTM. Did we measure again to see if there were extra improvements?

diegokingston (approved):

PR #4082, opened 2025-08-19
fix(levm): check coordinates are within field bounds in ecpairing precompile

fmoletta:
**Motivation**
A gas mismatch was found on a [sepolia transaction](https://sepolia.etherscan.io/tx/0x4355d49be46e61a53c71f45a128ebefb52cb38df08ed55833c2c162d26396819). This was due to the transaction failing at an ecpairing precompile on the network but not on levm. The difference was that geth checks that the coordinates received are within the bounds of the field, while we perform modulus on the input coordinates (see `BN254FieldElement::from_bytes_be`). This PR fixes this by checking that the input coordinates are within the field bounds before constructing the FieldElements.
It also adds some tests taken from geth to ensure that the implementation remains correct + a test for the out of bounds case

**Description**
* Check that input coordinates are within field bounds in `ecpairing` precompile
* Add tests for `ecpairing` precompile

Closes #issue_number


SDartayet (approved):
LGTM!

crates/vm/levm/src/precompiles.rs
```diff
@@ -619,22 +621,23 @@ type FirstPointCoordinates = (
 );
 
 /// Parses first point coordinates and makes verification of invalid infinite
-#[inline]
```
JereSalo, line 642: Why remove the inline?

fmoletta: Nice catch! Must have overlooked it when solving merge conflicts

PR #4083, opened 2025-08-19
refactor(l2): decouple eth_clients from SDK

avilagaston9:
**Motivation**

WIP

**Description**

Closes None


crates/networking/rpc/clients/eth/mod.rs
```diff
@@ -363,11 +339,7 @@ impl EthClient {
             value: overrides.value.unwrap_or_default(),
             from: overrides.from.unwrap_or_default(),
             gas: overrides.gas_limit,
-            gas_price: if let Some(gas_price) = overrides.max_fee_per_gas {
-                gas_price
-            } else {
-                self.get_gas_price().await?.as_u64()
-            },
+            gas_price: overrides.max_fee_per_gas.unwrap_or_default(),
```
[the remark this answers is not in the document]
avilagaston9: If `gas_price` is set to zero, then the `eth_call` ignores all `gas_price` related checks.

crates/l2/sequencer/setup.rs
```diff
@@ -69,20 +69,19 @@ pub async fn register_tdx_key(
...
-                max_priority_fee_per_gas: Some(gas_price),
-                ..Default::default()
-            },
-        )
-        .await?;
+    let verify_tx = build_generic_tx(
+        eth_client,
+        TxType::EIP1559,
+        tdx_address,
+        get_address_from_secret_key(private_key).map_err(ProofCoordinatorError::InternalError)?,
```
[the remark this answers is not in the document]
avilagaston9: It's OK. It's not an `EthClientError`.

crates/l2/sdk/src/sdk.rs
```diff
@@ -382,19 +367,55 @@ pub enum DeployError {
...
     deployer: &Signer,
     salt: &[u8],
     eth_client: &EthClient,
 ) -> Result<(H256, Address), DeployError> {
     let bytecode = hex::decode(read_to_string(contract_path)?)?;
-    deploy_contract_from_bytecode(constructor_args, &bytecode, deployer, salt, eth_client).await
+    create2_deploy_from_bytecode(constructor_args, &bytecode, deployer, salt, eth_client).await
 }
 
 /// Same as `deploy_contract`, but takes the bytecode directly instead of a path.
```
tomip01: I think the comment got outdated

```suggestion
/// Same as `create2_deploy_from_path `, but takes the bytecode directly instead of a path.
```

Or we can remove it since the funtion name already says what it is in the comment

crates/l2/sdk/src/sdk.rs
```diff
@@ -604,3 +631,456 @@ pub fn get_address_alias(address: Address) -> Address {
...
+        ));
+    }
+
+    let value = U256::from_str_radix(hex_string, 16).map_err(|_| {
+        EthClientError::Custom(
+            "Failed to parse after call, U256::from_str_radix failed.".to_owned(),
+        )
+    })?;
+    Ok(value)
+}
```
tomip01: Maybe is too nit but this could be moved to `calldata.rs` since it only parses hex strings.

Maybe is too nitpicky

crates/l2/sdk/src/sdk.rs
```diff
@@ -604,3 +631,456 @@ pub fn get_address_alias(address: Address) -> Address {
...
+    if data_end > bytes.len() {
+        return Err(EthClientError::Custom("Invalid array length".to_owned()));
+    }
+
+    // Convert the slice directly to H256 array
+    bytes[data_start..data_end]
+        .chunks_exact(32)
+        .map(|chunk| Ok(H256::from_slice(chunk)))
+        .collect()
+}
```
tomip01: same here

crates/networking/rpc/clients/eth/errors.rs
```diff
@@ -251,6 +251,7 @@ pub enum CalldataEncodeError {
     InternalError,
 }
 
+// TODO: move to L2
```
tomip01, line 256: Should we create an issue for this?

tomip01 (approved):
Looks good. Just some nits

crates/l2/sdk/src/sdk.rs
avilagaston9: Done [e87c0c0](https://github.com/lambdaclass/ethrex/pull/4083/commits/e87c0c02fa27ee70c83172304d7ae09478b15f8d)!

crates/l2/sdk/src/sdk.rs
avilagaston9: Done [e87c0c0](https://github.com/lambdaclass/ethrex/pull/4083/commits/e87c0c02fa27ee70c83172304d7ae09478b15f8d)!

crates/l2/sdk/src/sdk.rs
avilagaston9: Done [e87c0c0](https://github.com/lambdaclass/ethrex/pull/4083/commits/e87c0c02fa27ee70c83172304d7ae09478b15f8d)!

crates/networking/rpc/clients/eth/errors.rs
avilagaston9: Yes! I'm just debating whether we should delete all per-RPC method errors and keep only one `EthClientError` variant with the tuple `(method, error)`. I will create the issue as soon as it’s decided.

crates/l2/common/src/utils.rs
```diff
@@ -0,0 +1,20 @@
+use ethrex_common::Address;
+use keccak_hash::keccak;
+use secp256k1::SecretKey;
+
+pub fn get_address_from_secret_key(secret_key: &SecretKey) -> Result<Address, String> {
```
ManuelBilbao, line 5: We also have the same function in `sdk.rs`. Can we remove one of them?

crates/l2/sdk/src/calldata.rs
```diff
@@ -510,6 +511,59 @@ fn copy_into(
...
+    }
+
+    // Convert the slice directly to H256 array
+    bytes[data_start..data_end]
+        .chunks_exact(32)
+        .map(|chunk| Ok(H256::from_slice(chunk)))
+        .collect()
+}
+
+pub fn from_hex_string_to_u256(hex_string: &str) -> Result<U256, EthClientError> {
```
ManuelBilbao: Why not use `U256::from_str_radix` directly in the caller?

crates/l2/sdk/src/calldata.rs
```diff
@@ -510,6 +511,59 @@ fn copy_into(
...
+    let hex_string = hex_string.strip_prefix("0x").ok_or(EthClientError::Custom(
+        "Couldn't strip prefix from request.".to_owned(),
+    ))?;
+
+    if hex_string.is_empty() {
+        return Err(EthClientError::Custom(
+            "Failed to fetch last_committed_block. Manual intervention required.".to_owned(),
```
ManuelBilbao: This error message is misleading btw

crates/networking/rpc/clients/eth/mod.rs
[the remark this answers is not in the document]
ManuelBilbao: Are we sure about this? IIRC transactions with 0 gas price failed

crates/l2/sdk/src/sdk.rs
```diff
@@ -604,3 +633,411 @@ pub fn get_address_alias(address: Address) -> Address {
...
+    calldata.extend(vec![0; leading_zeros]);
+
+    let hex_string = client
+        .call(contract_address, calldata.into(), Overrides::default())
+        .await?;
+
+    Ok(hex_string)
+}
+
+async fn _call_variable(
```
ManuelBilbao: Where does the `_variable` come from? Also, should this be `_call_u64_variable`, accordingly to other methods?

crates/l2/sdk/src/sdk.rs
```diff
@@ -604,3 +633,411 @@ pub fn get_address_alias(address: Address) -> Address {
...
+    client: &EthClient,
+    selector: &[u8],
+    on_chain_proposer_address: Address,
+) -> Result<u64, EthClientError> {
+    let hex_string = _generic_call(client, selector, on_chain_proposer_address).await?;
+
+    let value = from_hex_string_to_u256(&hex_string)?
```
ManuelBilbao: Can't we use `u64::from_str_radix` here?

crates/l2/sdk/src/sdk.rs
ManuelBilbao, line 1: Not very fan of moving things used by the client to SDK, but we can debate it later

crates/networking/rpc/clients/eth/mod.rs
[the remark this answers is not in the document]
avilagaston9: Yes, it happens [here](https://github.com/lambdaclass/ethrex/blob/f06595020db5a6f0ebd5ce5291dfe25edac25213/crates/vm/backends/levm/mod.rs#L484).
We use `client.call()` in several places throughout the codebase without setting a `gas_price`, the change would have broken the tests otherwise.

crates/l2/common/src/utils.rs
avilagaston9: Yes! It was removed from the `SDK` and kept only here. In the PR description I clarified that I couldn't leave it only in the `SDK ` because the client needs it as well.

crates/l2/sdk/src/calldata.rs
avilagaston9: Done [f020862](https://github.com/lambdaclass/ethrex/pull/4083/commits/f02086213c02bf709e381d61abb82a81a7f65fed)!

crates/l2/sdk/src/sdk.rs
avilagaston9: Done [f020862](https://github.com/lambdaclass/ethrex/pull/4083/commits/f02086213c02bf709e381d61abb82a81a7f65fed)!

crates/l2/sdk/src/sdk.rs
avilagaston9: Done [f020862](https://github.com/lambdaclass/ethrex/pull/4083/commits/f02086213c02bf709e381d61abb82a81a7f65fed)!

crates/l2/sdk/src/sdk.rs
avilagaston9: The client can't use the `SDK` because of circular dependencies. The `SDK` is now essentially a wrapper around the client, but with additional intelligent logic. For example, methods like `deploy()` or `send_tx_bump()` are now placed in the `SDK`, but they use the client's `send_raw_tx` method under the hood.

ManuelBilbao (approved):

PR #4090, opened 2025-08-20
ci(l1,l2): separate GPU builds and use x86-64-v2 target

ManuelBilbao:
**Motivation**

**Description**

Closes #issue_number


xqft:
Can we document the performance gain by building from source for native? seems significant

PR #4091, opened 2025-08-20
fix: treat missing storage slot in pruned trie as absent instead of DB error

### LINKED ISSUE #4088: `ethrex replay`: fix error `EvmError(Custom("Database access error: DB error: Failed get storage slot"))` with ethrex + SP1
Block to reproduce: ? (modify this if we find one)

Galoretka:
**Motivation**

#4088 

**Description**

in crates/common/types/block_execution_witness.rs, ExecutionWitnessResult::get_storage_slot now returns Ok(None) when storage_trie.get(hashed_key) errors (pruned trie gaps), rather than propagating a DB error.

- Still returns Err(Database(...)) on decode failures for existing nodes.
- Aligns SP1 replay/proving behavior with LEVM expectations without altering public APIs.

Closes #4088

### DIFF
--- a/crates/common/types/block_execution_witness.rs
+++ b/crates/common/types/block_execution_witness.rs
@@ -346,17 +346,15 @@ impl ExecutionWitnessResult {
             return Ok(None);
         };
         let hashed_key = hash_key(&key);
-        if let Some(encoded_key) = storage_trie
-            .get(&hashed_key)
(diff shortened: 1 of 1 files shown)


JereSalo:
We don't want this change for now as it's been working fine lately. Thank you very much though.

PR #4092, opened 2025-08-20
perf(levm): improve blobbasefee

edg-l:
**Motivation**

Improves blobbasefee perfomance by avoiding unnecesary checks

### DIFF
--- a/crates/vm/levm/src/gas_cost.rs
+++ b/crates/vm/levm/src/gas_cost.rs
@@ -775,34 +775,47 @@ pub fn staticcall(
     calculate_cost_and_gas_limit_call(true, gas_from_stack, gas_left, call_gas_costs, 0)
 }
 
-pub fn fake_exponential(factor: U256, numerator: U256, denominator: U256) -> Result<U256, VMError> {
-    let mut i = U256::one();
+pub fn fake_exponential(factor: U256, numerator: U256, denominator: u64) -> Result<U256, VMError> {
+    if denominator == 0 {
+        return Err(InternalError::DivisionByZero.into());
+    }
+
+    if numerator.is_zero() {
+        return Ok(factor);
+    }
+
+    let mut i: u64 = 1;
     let mut output: U256 = U256::zero();
+    let denominator_u256: U256 = denominator.into();
 
     // Initial multiplication: factor * denominator
     let mut numerator_accum = factor
-        .checked_mul(denominator)
+        .checked_mul(denominator_u256)
         .ok_or(InternalError::Overflow)?;
 
-    while !numerator_accum.is_zero() {
-        // Safe addition to output
-        output = output
-            .checked_add(numerator_accum)
-            .ok_or(InternalError::Overflow)?;
+    #[expect(
+        clippy::arithmetic_side_effects,
+        reason = "division can't overflow since denominator is not 0 and i is never 0"
+    )]
+    {
+        while !numerator_accum.is_zero() {
+            // Safe addition to output
+            output = output
+                .checked_add(numerator_accum)
+                .ok_or(InternalError::Overflow)?;
 
-        // Safe multiplication and division within loop
-        numerator_accum = numerator_accum
-            .checked_mul(numerator)
-            .ok_or(InternalError::Overflow)?
-            .checked_div(denominator.checked_mul(i).ok_or(InternalError::Overflow)?)
-            .ok_or(InternalError::DivisionByZero)?;
+            // Safe multiplication and division within loop
+            numerator_accum = numerator_accum
+                .checked_mul(numerator)
+                .ok_or(InternalError::Overflow)?
+                / denominator.checked_mul(i).ok_or(InternalError::Overflow)?;
 
-        i = i.checked_add(U256::one()).ok_or(InternalError::Overflow)?;
+            // i will never realistically overflow before the other variables do
+            i = i.wrapping_add(1);
+        }
+
+        Ok(output / denominator)
     }
-
-    output
-        .checked_div(denominator)
-        .ok_or(InternalError::DivisionByZero.into())
 }
 
 pub fn sha2_256(data_size: usize) -> Result<u64, VMError> {
--- a/crates/vm/levm/src/utils.rs
+++ b/crates/vm/levm/src/utils.rs
@@ -199,7 +199,7 @@ pub fn get_base_fee_per_blob_gas(
     fake_exponential(
         MIN_BASE_FEE_PER_BLOB_GAS,
         block_excess_blob_gas.unwrap_or_default(),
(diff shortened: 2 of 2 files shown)


crates/vm/levm/src/gas_cost.rs
```diff
@@ -775,34 +775,47 @@ pub fn staticcall(
...
+
+    let mut i: u64 = 1;
     let mut output: U256 = U256::zero();
+    let denominator_u256: U256 = denominator.into();
 
     // Initial multiplication: factor * denominator
     let mut numerator_accum = factor
-        .checked_mul(denominator)
+        .checked_mul(denominator_u256)
         .ok_or(InternalError::Overflow)?;
```
Oppen, line 795: Maybe just multiply to a U512 if there is a real possibility of the value not fitting.

crates/vm/levm/src/gas_cost.rs
```diff
@@ -775,34 +775,47 @@ pub fn staticcall(
...
+            // Safe addition to output
+            output = output
+                .checked_add(numerator_accum)
+                .ok_or(InternalError::Overflow)?;
+
+            // Safe multiplication and division within loop
+            numerator_accum = numerator_accum
+                .checked_mul(numerator)
+                .ok_or(InternalError::Overflow)?
+                / denominator.checked_mul(i).ok_or(InternalError::Overflow)?;
```
Oppen: `denominator` doesn't change inside the loop, so let's just accumulate the value as addition:
```rust
let mut denominator_by_i = denominator;
// ...
while !numerator_accum.is_zero() {
    // ...
    denominator_by_i += denominator;
    // ...
}
```

crates/vm/levm/src/gas_cost.rs
```diff
@@ -775,34 +775,47 @@ pub fn staticcall(
     calculate_cost_and_gas_limit_call(true, gas_from_stack, gas_left, call_gas_costs, 0)
 }
 
-pub fn fake_exponential(factor: U256, numerator: U256, denominator: U256) -> Result<U256, VMError> {
-    let mut i = U256::one();
+pub fn fake_exponential(factor: U256, numerator: U256, denominator: u64) -> Result<U256, VMError> {
```
Oppen, line 780: Can we add a comment to this function that says what it actually computes? I believe it is this:
```suggestion
/// fake_exponential: computes ∑i=[0..N] factor * (numerator / i * denominator) ^ i
pub fn fake_exponential(factor: U256, numerator: U256, denominator: u64) -> Result<U256, VMError> {
```
Looks like a geometric series? Don't know, but it's not just an exponential and neither the code nor the name are obvious.
@JereSalo I think you're the last one to touch this function, do you have any pointers? If we know what exactly we're trying to compute there's a chance to find a better algorithm.

crates/vm/levm/src/gas_cost.rs
```diff
@@ -775,34 +775,47 @@ pub fn staticcall(
...
+                .ok_or(InternalError::Overflow)?;
+
+            // Safe multiplication and division within loop
+            numerator_accum = numerator_accum
+                .checked_mul(numerator)
+                .ok_or(InternalError::Overflow)?
+                / denominator.checked_mul(i).ok_or(InternalError::Overflow)?;
+
+            // i will never realistically overflow before the other variables do
+            i = i.wrapping_add(1);
```
Oppen: Given the above comment, why not just use `i += 1`?

crates/vm/levm/src/gas_cost.rs
JereSalo: From [EIP-4884](https://eips.ethereum.org/EIPS/eip-4844#helpers)
> Approximates factor * e ** (numerator / denominator) using Taylor expansion:

Maybe we could add a link to it or this comment

JereSalo: Fun fact, this is the same function as the one [here](https://github.com/lambdaclass/ethrex/blob/38d96c3466a2054591152b9376d0996f00f9551b/crates/common/types/block.rs#L397-L408) but this one was made specifically for LEVM because we don't allow raw arithmetic operations (that's why we return a Result instead of just the number). I believe we should change it anytime soon and use only one in the client.

crates/vm/levm/src/gas_cost.rs
JereSalo: I believe it's because of clippy constraints in LEVM, this way we don't have to put a `clippy allow` but at this point maybe it's a good idea to just get rid of these checks and have more freedom when doing things that we know shouldn't fail.

Oppen: But the `clippy::expect` is there for the whole block, so it should be allowed already.

crates/vm/levm/src/gas_cost.rs
Oppen: > From [EIP-4884](https://eips.ethereum.org/EIPS/eip-4844#helpers)
> 
> > Approximates factor * e ** (numerator / denominator) using Taylor expansion:
> 
> Maybe we could add a link to it or this comment

I suggest we add both.

crates/vm/levm/src/gas_cost.rs
JereSalo: Oh sorry, I didn't see that. I just saw the comment and that instantly popped into my mind. mb

crates/vm/levm/src/gas_cost.rs
edg-l: i prefer not to, the usual case is probably not overflowing and probably u512 are more expensive themselves than a checked mul on u256

edg-l:
> This is good. That said, there is a much better optimization. This value is:
> 
>     1. Always computed if there is a maximum fee per gas for the block (`crates/vm/levm/src/hooks/default_hook.rs:48`);
> 
>     2. Constant for a given block.
>        So the best optimization we can get is just saving the value in that preparation step and reusing it later.
>        Can we do that on a follow-up?

yes, opened a issue https://github.com/lambdaclass/ethrex/issues/4111

crates/vm/levm/src/gas_cost.rs
```diff
@@ -775,34 +775,50 @@ pub fn staticcall(
...
+            // Safe addition to output
+            output = output
+                .checked_add(numerator_accum)
+                .ok_or(InternalError::Overflow)?;
+
+            // Safe multiplication and division within loop
+            numerator_accum = numerator_accum
+                .checked_mul(numerator)
+                .ok_or(InternalError::Overflow)?
+                / denominator_by_i;
```
JereSalo, line 814: I'm in favor of replacing the checked operations for normal ones to make this prettier if you want to, we already have the clippy check disabled here though

PR #4097, opened 2025-08-20
perf(levm): make precompiles use a constant table

edg-l:
**Motivation**

Improves performance of precompile calls


crates/vm/levm/src/precompiles.rs
```diff
@@ -200,41 +200,51 @@ pub fn is_precompile(address: &Address, fork: Fork) -> bool {
...
+        precompiles[BLS12_MAP_FP_TO_G1_ADDRESS.0[19] as usize] =
+            Some(bls12_map_fp_to_g1 as PrecompileFn);
+        precompiles[BLS12_MAP_FP2_TO_G2_ADDRESS.0[19] as usize] =
+            Some(bls12_map_fp2_tp_g2 as PrecompileFn);
+
+        precompiles
     };
 
-    Ok(result)
+    let index = address.0[19] as usize;
```
Oppen: The rest of the address still needs to be validated. At the very least we need to check `&address.0[..19] == &[0u8; 19]`.

crates/vm/levm/src/precompiles.rs
```diff
@@ -200,41 +200,51 @@ pub fn is_precompile(address: &Address, fork: Fork) -> bool {
...
+
+    if index >= PRECOMPILES.len() {
+        return Err(VMError::Internal(InternalError::InvalidPrecompileAddress));
+    }
+
+    let precompile = PRECOMPILES[address.0[19] as usize]
+        .ok_or(VMError::Internal(InternalError::InvalidPrecompileAddress))?;
```
```suggestion
    let precompile = PRECOMPILES.get(index).flatten()
        .ok_or(VMError::Internal(InternalError::InvalidPrecompileAddress))?;
```

crates/vm/levm/src/precompiles.rs
Oppen: Strictly speaking, there's currently `0x0000..0xffff` reserved for precompiles, so I would do it like this to future-proof it (pseudo-rust):
```rust
if address[0..18] != [0u8; 18] {
    return Err(InvalidPrecompilaAddress);
}
let index = u16::from_bytes_be([address[18], address[19]]) as usize;
```

crates/vm/levm/src/precompiles.rs
```diff
@@ -193,41 +193,53 @@ pub fn is_precompile(address: &Address, fork: Fork) -> bool {
...
+        precompiles[BLS12_MAP_FP_TO_G1_ADDRESS.0[19] as usize] =
+            Some(bls12_map_fp_to_g1 as PrecompileFn);
+        precompiles[BLS12_MAP_FP2_TO_G2_ADDRESS.0[19] as usize] =
+            Some(bls12_map_fp2_tp_g2 as PrecompileFn);
+
+        precompiles
     };
 
-    Ok(result)
+    if address[0..18] != [0u8; 18] {
```
JereSalo, line 231: Can we use a constant instead of the raw number? I don't know if we can do PRECOMPILES.len() or something like that. Or maybe declare one PRECOMPILES_LEN outside of this and use that. But if it makes the code way uglier then it's okay to leave it like this.

crates/vm/levm/src/precompiles.rs
```diff
@@ -193,41 +193,53 @@ pub fn is_precompile(address: &Address, fork: Fork) -> bool {
...
+        return Err(VMError::Internal(InternalError::InvalidPrecompileAddress));
+    }
+    let index = u16::from_be_bytes([address[18], address[19]]) as usize;
```
JereSalo, line 235: Why are we using a `u16` instead of a `u8`? address[18] will always be zero, won't it? If so, we can make the check above be `address[0..19]`

Oppen: Because we still need to validate that byte and the valid range of precompile addresses is the whole u16, even if they don't currently all map to a precompile. I suggested doing it this way to match the spec.

PR #4098, opened 2025-08-20
perf(levm): improve load_range by returning a Bytes directly, avoding a vec

edg-l:
**Motivation**

All load range usage leads to creating a Bytes, but to do so we allocate a intermediate vec, this pr avoids this.

improve load_range by returning a Bytes directly, avoding a vec

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## Perf
 
-### 2025-08-20
-
 ### 2025-08-22
 
 - Improve Memory::load_range by returning a Bytes directly, avoding a vec allocation [#4098](https://github.com/lambdaclass/ethrex/pull/4098)


crates/vm/levm/src/memory.rs
```diff
@@ -117,9 +118,9 @@ impl Memory {
         // SAFETY: resize already makes sure bounds are correct.
         #[allow(unsafe_code)]
         unsafe {
-            Ok(buf
-                .get_unchecked(true_offset..(true_offset.wrapping_add(size)))
-                .to_vec())
+            Ok(Bytes::copy_from_slice(buf.get_unchecked(
+                true_offset..(true_offset.wrapping_add(size)),
+            )))
```
Oppen, line 122: If we're doing this we should probably just make sure `Memory` is of type `Bytes` and use the `slice` method, as this returns a reference counted slice akin to those from Go (in their case GC'd but still a pair of pointers to the old array).
Also, keep in mind `Bytes` just took ownership from the old `Vec` and converted it to a reference counted thingy, so it should be the same number of allocations with this change.

edg-l: i ran into problems trying this, Bytes doesnt impl deref mut and also i think its not a growable container like Vec its mostly read only?

Oppen: Right, might be mostly read only. There is `BytesMut` I think?

JereSalo (approved):
good one

crates/vm/levm/src/memory.rs
edg-l: we can look into this in another pr i think, since the change is more involved

crates/vm/levm/src/memory.rs
```diff
@@ -4,6 +4,7 @@ use crate::{
     constants::{MEMORY_EXPANSION_QUOTIENT, WORD_SIZE_IN_BYTES_U64, WORD_SIZE_IN_BYTES_USIZE},
     errors::{ExceptionalHalt, InternalError, VMError},
 };
+use bytes::Bytes;
 use ExceptionalHalt::OutOfBounds;
```
Oppen:
```suggestion
use ExceptionalHalt::OutOfBounds;
use bytes::Bytes;
```

PR #4099, opened 2025-08-20
fix(l2): retry sending commit every 5 min instead of waiting commit_time_ms

LeanSerra:
**Motivation**

If we set the commit_time_ms to a larger value like we would when deploying on Ethereum mainnet in case of commit tx error we would have to wait commit_time_ms until the commit is retried, instead we set it to a fixed 5 minutes to retry the commitment

**Description**

- Add a flag retry_mode to L1Committer that is set to true if `commit_next_batch_to_l1()` failed
  - In retry mode every 5 minutes we retry sending the commit tx

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -72,6 +72,7 @@ pub struct L1Committer {
     signer: Signer,
     based: bool,
     sequencer_state: SequencerState,
+    retry_mode: bool,
 }
 
 impl L1Committer {
@@ -105,6 +106,7 @@ impl L1Committer {
             signer: committer_config.signer.clone(),
             based,
             sequencer_state,
+            retry_mode: false,
         })
     }
 
@@ -548,12 +550,19 @@ impl GenServer for L1Committer {
     ) -> CastResponse {
         // Right now we only have the Commit message, so we ignore the message
         if let SequencerStatus::Sequencing = self.sequencer_state.status().await {
-            let _ = self
-                .commit_next_batch_to_l1()
-                .await
-                .inspect_err(|err| error!("L1 Committer Error: {err}"));
+            if let Err(err) = self.commit_next_batch_to_l1().await {
+                error!("L1 Committer Error: {err}");
+                self.retry_mode = true;
+            } else {
+                self.retry_mode = false;
+            }
         }
-        let check_interval = random_duration(self.commit_time_ms);
+        let check_interval = if self.retry_mode {
+            warn!("Committer is in retry mode trying to resend commit in 5 minutes");
+            random_duration(300_000)
+        } else {
+            random_duration(self.commit_time_ms)
+        };
         send_after(check_interval, handle.clone(), Self::CastMsg::Commit);
         CastResponse::NoReply
     }


ilitteri (approved):

crates/l2/sequencer/l1_committer.rs
```diff
@@ -541,19 +554,34 @@ impl GenServer for L1Committer {
...
+
+            if should_send_commitment
+                && self
+                    .commit_next_batch_to_l1()
+                    .await
+                    .inspect_err(|e| error!("L1 Committer Error: {e}"))
+                    .is_ok()
+            {
+                let last_committed_batch_timestamp = system_now_ms().unwrap_or(current_time_ms);
+                self.last_committed_batch_timestamp = last_committed_batch_timestamp;
```
ManuelBilbao: nit
```suggestion
                self.last_committed_batch_timestamp = system_now_ms().unwrap_or(current_time_ms);
```

ManuelBilbao (approved):

crates/l2/sequencer/l1_committer.rs
LeanSerra: Done in [1635340](https://github.com/lambdaclass/ethrex/pull/4099/commits/16353409276df8ebe5cf31018c3e64cdf4c1daee)

xqft (commented):
I'm not sure about this approach. If the commit transaction fails, then the committer should handle the error immediately, via 
- bumping the gas price
- resending the tx with a backoff
- creating a new batch from scratch when a number of max. tx send retries is reached

and crashing if everything fails (critical scenario potentially produced by a bug in our code)

the state about the latest commitment is in the L1 contract, it shouldn't be necessary for the committer to store it itself

crates/l2/sequencer/l1_committer.rs
```diff
@@ -541,19 +562,49 @@ impl GenServer for L1Committer {
 
     type Error = CommitterError;
 
+    // Right now we only have the Commit message, so we ignore the message
```
xqft:
```suggestion
    // Right now we only have the `Commit` message, so we ignore the `message` parameter
```

xqft: (nit)

crates/l2/sequencer/l1_committer.rs
```diff
@@ -541,19 +562,49 @@ impl GenServer for L1Committer {
...
+            let commit_time_ms: u128 = self.commit_time_ms.into();
+            let should_send_commitment =
+                current_time_ms - self.last_committed_batch_timestamp > commit_time_ms;
+
+            if should_send_commitment
+                && self
+                    .commit_next_batch_to_l1()
+                    .await
+                    .inspect_err(|e| error!("L1 Committer Error: {e}"))
+                    .is_ok()
```
xqft: you are taking advantage of short-circuit here right? I think it's correct but it's one character away (`&`) of becoming a bug (bitwise AND doesn't short-circuit), would be better to nest the if statements

xqft:
Regarding my previous comment, we talked internally and I understood the problem. This is to handle the case in which the network is congested and the committer waits (a time smaller than the commit time) for gas to be cheaper.

crates/l2/sequencer/l1_committer.rs
LeanSerra: Yes, I was not trying to be smart here but clippy complains if the `if` block is not collapsed.

LeanSerra: Ended up using allow(clippy) in [a7a6756](https://github.com/lambdaclass/ethrex/pull/4099/commits/a7a67561d6da7c1070cc97a8df17998eac9ca89a)

cmd/ethrex/l2/options.rs
```diff
@@ -165,6 +165,10 @@ impl TryFrom<SequencerOptions> for SequencerConfig {
                     .committer_opts
                     .on_chain_proposer_address
                     .ok_or(SequencerOptionsError::NoOnChainProposerAddress)?,
+                first_wake_up_time_ms: opts
+                    .committer_opts
+                    .first_wake_up_time_ms
+                    .unwrap_or(opts.committer_opts.commit_time_ms),
```
ManuelBilbao: Default should be 0 IMO

cmd/ethrex/l2/options.rs
```diff
@@ -453,6 +457,14 @@ pub struct CommitterOptions {
         help = "How often does the sequencer commit new blocks to the L1 in milliseconds."
     )]
     pub commit_time_ms: u64,
+    #[arg(
+        long = "committer.first-wake-up-time",
+        value_name = "UINT64",
+        env = "ETHREX_COMMITTER_FIRST_WAKE_UP_TIME",
+        help_heading = "L1 Committer options",
+        help = "Time to wait before the sequencer commits new blocks for the first time. Defaults to commit_time_ms. After sending the first commit `committer.commit-time` will be used."
```
ManuelBilbao:
```suggestion
        help = "Time to wait before the sequencer seals a batch when started. After committing the first batch, `committer.commit-time` will be used."
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -72,6 +77,12 @@ pub struct L1Committer {
     signer: Signer,
     based: bool,
     sequencer_state: SequencerState,
+    /// L1 Committer time to wait before checking if it should send a new batch
```
ManuelBilbao:
```suggestion
    /// Time to wait before checking if it should send a new batch
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -72,6 +77,12 @@ pub struct L1Committer {
...
+    committer_wake_up_ms: u64,
+    /// Timestamp of last successful committed batch
+    last_committed_batch_timestamp: u128,
+    /// Last succesful committed batch
```
ManuelBilbao: Nit, just to clarify
```suggestion
    /// Last succesful committed batch number
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -47,6 +50,8 @@ use spawned_concurrency::{
 const COMMIT_FUNCTION_SIGNATURE_BASED: &str =
     "commitBatch(uint256,bytes32,bytes32,bytes32,bytes32,bytes[])";
 const COMMIT_FUNCTION_SIGNATURE: &str = "commitBatch(uint256,bytes32,bytes32,bytes32,bytes32)";
+// Default wake up time for the committer to check if it should send a commit tx
```
ManuelBilbao:
```suggestion
/// Default wake up time for the committer to check if it should send a commit tx
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -124,11 +140,18 @@ impl L1Committer {
...
-        let mut l1_committer = state.start();
-        l1_committer
-            .cast(InMessage::Commit)
-            .await
-            .map_err(CommitterError::InternalError)
+        // Set the initial last committed batch
+        state.last_committed_batch = state
+            .eth_client
+            .get_last_committed_batch(state.on_chain_proposer_address)
+            .await?;
```
ManuelBilbao: Should we put this in the `::new()` function?

crates/l2/sequencer/l1_committer.rs
```diff
@@ -541,19 +564,51 @@ impl GenServer for L1Committer {
...
         if let SequencerStatus::Sequencing = self.sequencer_state.status().await {
-            let _ = self
-                .commit_next_batch_to_l1()
+            let current_last_committed_batch = self
+                .eth_client
+                .get_last_committed_batch(self.on_chain_proposer_address)
                 .await
-                .inspect_err(|err| error!("L1 Committer Error: {err}"));
+                .unwrap_or(self.last_committed_batch);
+            let Some(current_time_ms) = utils::system_now_ms() else {
```
ManuelBilbao: I'd avoid using `_ms` suffix outside of struct attributes
```suggestion
            let Some(current_time) = utils::system_now_ms() else {
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -541,19 +564,51 @@ impl GenServer for L1Committer {
...
+            // In the event that the current batch in L1 is greater than the one we have recorded we shouldn't send a new batch
+            if current_last_committed_batch > self.last_committed_batch {
+                self.last_committed_batch = current_last_committed_batch;
+                self.last_committed_batch_timestamp = current_time_ms;
+                let check_interval = random_duration(self.committer_wake_up_ms);
+                send_after(check_interval, handle.clone(), Self::CastMsg::Commit);
+                return CastResponse::NoReply;
+            }
+
+            let commit_time_ms: u128 = self.commit_time_ms.into();
```
ManuelBilbao: Same here
```suggestion
            let commit_time: u128 = self.commit_time_ms.into();
```

LeanSerra: Done in [eb3653e](https://github.com/lambdaclass/ethrex/pull/4099/commits/eb3653e05a22d73c43225c58ead4c88dfe4d3882)

crates/l2/sequencer/l1_committer.rs
LeanSerra: Done in [eb3653e](https://github.com/lambdaclass/ethrex/pull/4099/commits/eb3653e05a22d73c43225c58ead4c88dfe4d3882)

crates/l2/sequencer/l1_committer.rs
LeanSerra: Yes, I was reluctant on having an async `new()` function but we have a few of those in the code so I changed it  in [542624c](https://github.com/lambdaclass/ethrex/pull/4099/commits/542624c3d8b66ae7a050c8579b299dfaa1d68285)

cmd/ethrex/l2/options.rs
LeanSerra: Done in [ef81784](https://github.com/lambdaclass/ethrex/pull/4099/commits/ef81784bca80fba0e8f133205a11d447acf91183)

crates/l2/sequencer/l1_committer.rs
LeanSerra: Done in [5c811dc](https://github.com/lambdaclass/ethrex/pull/4099/commits/5c811dc1814bd9706fe4dba37d5660a2bab9e650)

crates/l2/sequencer/l1_committer.rs
LeanSerra: Done in [5c811dc](https://github.com/lambdaclass/ethrex/pull/4099/commits/5c811dc1814bd9706fe4dba37d5660a2bab9e650)

crates/l2/sequencer/l1_committer.rs
LeanSerra: Done in [5c811dc](https://github.com/lambdaclass/ethrex/pull/4099/commits/5c811dc1814bd9706fe4dba37d5660a2bab9e650)

cmd/ethrex/l2/options.rs
LeanSerra: Done in [5c811dc](https://github.com/lambdaclass/ethrex/pull/4099/commits/5c811dc1814bd9706fe4dba37d5660a2bab9e650)

ManuelBilbao (approved):

PR #3992, opened 2025-08-07
fix(l1): `ethrex-replay` block execution in different networks

ilitteri:
**Motivation**

The `ethrex-replay` tool is not working correctly in many aspects. From the starts block execution is not working correctly in the different networks (Hoodi, Sepolia, and Mainnet).

**Description**

This PR fixes the following related to `ethrex-replay`:
- `ExecutionWitnessResult` serialization and deserialization to make it compliant with Reth nodes.
- Ensures the use of the correct chain configuration on each command.
- The latest block is now correctly cached and labeled (by network and block number, e.g. `cache_network_number.json`).
- Ensures the replay of the last canonical block when replaying the latest block.

And the following extras:
- Improves logging.
- Renames some functions and variables.

> [!NOTE]
> - This PR ensures the correct execution of most of blocks produced by reth and ethrex clients.
> - Some block execution may fail, and their corresponding fixes should be done in other PRs.
> - To ensure block execution from other clients than the stated above, we need to support retrieving block execution data using the endpoint `eth_getProofs` (this is a WIP, we're currently using `debug_executionWitness`, but only Reth and ethrex nodes support this endpoint).

**Test it out**

```
# reth RPCs
Hoodi = http://65.108.69.58:8545
Sepolia = http://65.109.97.102:8545
Mainnet = http://157.180.1.98:8545

# Executing
cargo run --release --package ethrex-replay --bin ethrex-replay -- execute block --rpc-url http://65.108.69.58:8545 --network hoodi

# Executing (SP1)
cargo run --release --package ethrex-replay --bin ethrex-replay --features sp1 -- execute block 1033114 --rpc-url http://65.108.69.58:8545 --network hoodi

 # Proving (SP1) 
 SP1_PROVER=cuda cargo run --release --package ethrex-replay --bin ethrex-replay --features sp1,gpu -- execute block 1033114 --rpc-url http://65.108.69.58:8545 --network hoodi
```

PR #4129, opened 2025-08-21
ci(l1,l2): fix main_prover_l1 block cache

### LINKED ISSUE #4124: CI: Fix `ethrex-replay`
**Context**

The **Replaying proving** job is broken since 2ffd28ee6022cb951f3cdaf99179bec16d503d6c because the job uses a cache file whose format (serialization) is no longer valid.  

**Description**

Update the cached block with a valid one.

LeanSerra:
**Motivation**

The block was not updated with the changes in #3992

**Description**

- update with a new cache file for block https://hoodi.etherscan.io/block/974567

Closes #4124


ilitteri (approved):

.github/workflows/main_prover_l1.yaml
```diff
@@ -2,6 +2,10 @@ name: Replay proving
 on:
   push:
     branches: ["main"]
+  pull_request:
+    branches: ["**"]
+    paths:
+      - ".github/workflows/main_prover_l1.yaml"
```
mpaulucci, line 8: maybe you want to add cmd/ethrex_replay here too?

cmd/ethrex_replay/src/fetcher.rs
```diff
@@ -17,19 +17,16 @@ pub async fn get_blockdata(
...
+            Err(e) => return Err(e.into()),
+        },
         BlockIdentifier::Tag(_) => unimplemented!("Only latest block tag is supported"),
     };
 
-    info!(
-        "Retrieving execution data for block {requested_block_number} ({} block behind latest)",
-        latest_block_number - requested_block_number
-    );
-
```
ilitteri: Moving this log down in the function to the point where the error we are having does not appear anymore is not a solution, but an easy patch. We still want this log for both cached and non-cached blocks.

ilitteri (changes requested):

cmd/ethrex_replay/src/fetcher.rs
LeanSerra: Reverted those changes [6239a3b](https://github.com/lambdaclass/ethrex/pull/4129/commits/6239a3b76326328aa8f9889eefc8ed6fd3c5ee16) and [d0db9ca](https://github.com/lambdaclass/ethrex/pull/4129/commits/d0db9ca1bc0bffa6d31ec9d1bbe4837736e35e97)

.github/workflows/main_prover_l1.yaml
ilitteri: I think it's a good idea to do so. The only problem with this is that this job would require the GPU runner to step in, and that would slow down the merge queue. 

What we can do is to have a job that executes using SP1 in PRs that touch files from `cmd/ethrex_replay `. This would be a reliable barrier before the `main_prover_l1 ` workflow runs in main.

LeanSerra: I created this issue to solve this #4134

PR #4131, opened 2025-08-22
fix(l1): use proper docker image to spin up localnets.

mpaulucci:
**Motivation**
We were using the stable ethrex docker images instead of the one build locally

**Description**
- Use the local image of ethrex
- Merge many localnet makefile commands into one
- Fix passing of `el_extra_params`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ clean: clean-vectors ## 🧹 Remove build artifacts
 
 STAMP_FILE := .docker_build_stamp
 $(STAMP_FILE): $(shell find crates cmd -type f -name '*.rs') Cargo.toml Dockerfile
-	docker build -t ethrex:unstable .
+	docker build -t ethrex .
 	touch $(STAMP_FILE)
 
 build-image: $(STAMP_FILE) ## 🐳 Build the Docker image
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -62,34 +62,11 @@ checkout-ethereum-package: ## 📦 Checkout specific Ethereum package revision
...
 
 localnet: stop-localnet-silent build-image checkout-ethereum-package ## 🌐 Start local network
-	kurtosis run --enclave $(ENCLAVE) ethereum-package --args-file fixtures/network/network_params.yaml
-	docker logs -f $$(docker ps -q --filter ancestor=ethrex)
-
-localnet-snooper: stop-localnet-silent build-image checkout-ethereum-package ## 🌐 Start local network and output the JSON-RPC requests ethrex exchanges with the consensus client
-	kurtosis run --enclave $(ENCLAVE) ethereum-package --args-file fixtures/network/network_params.yaml
-	docker logs -f $$(docker ps -q --filter name=snooper-engine-3-lighthouse-ethrex)
-
-localnet-client-comparision: stop-localnet-silent build-image checkout-ethereum-package ## 🌐 Start local network
```
MegaRedHand, line 74: This removes a lot of unrelated makefile targets. Is that ok?

Makefile
```diff
@@ -25,7 +25,7 @@ clean: clean-vectors ## 🧹 Remove build artifacts
 
 STAMP_FILE := .docker_build_stamp
 $(STAMP_FILE): $(shell find crates cmd -type f -name '*.rs') Cargo.toml Dockerfile
-	docker build -t ethrex:unstable .
+	docker build -t ethrex .
```
MegaRedHand, line 28: There are some places in CI and makefile where we retag this, I think. We should clean up those

Makefile
mpaulucci: yes, I don't think we need one target per network

Makefile
mpaulucci: yes, its being worked on here: https://github.com/lambdaclass/ethrex/pull/4120

In general we shouldn't use make targets on the CI IMO

PR #3871, opened 2025-07-29
feat(levm): change feature flag to revm

shiyasmohd:
**Motivation**

Change feature flag to `revm`

**Description**

  - Rename feature flag from `levm` to `revm` in Cargo.toml
  - Update Makefile to run LEVM by default and REVM with feature flag
  - Invert EVM engine selection logic to use LEVM as default

Closes #3335

PR #4135, opened 2025-08-22
test(l1): add possibility of running Blockchain EFTests with SP1

### LINKED ISSUE #4144: run blockchain tests with SP1

JereSalo:
**Motivation**

**Description**
Run tests with SP1.

If we weren't able to generate the execution witness we won't run the test with SP1 because it'd make no sense.

Created this issue so that we can do stateless run without depending on a normal run done before, it's not essential though. https://github.com/lambdaclass/ethrex/issues/4142

Closes #4144


cmd/ef_tests/blockchain/Cargo.toml
```diff
@@ -29,6 +29,8 @@ default = ["c-kzg", "blst"]
 blst = ["ethrex-vm/blst"]
 c-kzg = ["ethrex-blockchain/c-kzg"]
 levm = []
+sp1 = ["zkvm_interface/sp1", "ethrex-prover/sp1"]
+stateless = []
```
ilitteri: Is this feature flag needed? Couldn't this be the default when `sp1` is disabled?

ilitteri (approved):

cmd/ef_tests/blockchain/Cargo.toml
JereSalo: There are 4 ways of running the tests:
- levm
- revm
- sp1 (levm)
- stateless (levm)

The difference between `stateless` and `levm` is that the former runs the levm tests normally and then it runs them in a stateless manner, generating an Execution Witness for stateless execution.

Before this PR if you wanted to run levm tests you'd run them both normally and in a stateless manner, I separated that in case that the user just wants to run them in a normal way.

Do you think we should always run stateless tests alongside normal levm tests? I separated these because I thought it made more sense but maybe the other approach is more preferable

JereSalo:
Waiting for https://github.com/lambdaclass/ethrex/pull/3871 to be merged so that we don't add conflicts to the external contributor.

PR #4138, opened 2025-08-22
feat(l1): add `prune-blocks` tool

fmoletta:
**Motivation**
Ethrex nodes can get pretty heavy. In order to lower the space of the DB when full syncing large networks this PR proposes adding a `prune-blocks` tool which will remove old blocks and their associated data.

**Description**
* Add `prune-blocks` tool

Closes #issue_number

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,8 @@ members = [
   "tooling/load_test",
   "tooling/loc",
   "tooling/archive_sync",
-  "crates/common/config", "tooling/prune-blocks",
+  "crates/common/config",
+  "tooling/prune-blocks",
 ]
 resolver = "2"
 
--- a/tooling/prune-blocks/src/README.md
+++ b/tooling/prune-blocks/src/README.md
@@ -1,6 +1,6 @@
 # Prune Blocks
 
-This tool can be used to reduce the DB size of the node by removing old blocks and their asociated data. Note that this is counter-spec and will hinder the node's ability to provide data to other nodes. It also does not performe state prunning.
+This tool can be used to reduce the DB size of the node by removing old blocks and their associated data. Note that this is counter-spec and will hinder the node's ability to provide data to other nodes. It also does not perform state pruning.
 
 ## Usage
 
@@ -8,7 +8,7 @@ The tool takes two optional arguments:
     *`datadir`: The path to the DB location, will use the default one if not provided
     *`blocks_to_keep`: The amount of latest blocks that will be kept in the DB. This value must be at least 128 and lower than the current amount of blocks in the chain.
 
-And should be ran like this:
+And should be run like this:
 
 ```bash
 cargo run --release -- --datadir DATADIR --blocks-to-keep BLOCKS_TO_KEEP


crates/storage/store_db/libmdbx.rs
```diff
@@ -322,6 +322,53 @@ impl StoreEngine for Store {
...
+        let txn = self
+            .db
+            .begin_readwrite()
+            .map_err(StoreError::LibmdbxError)?;
+        let Some(block_hash) = txn
+            .get::<CanonicalBlockHashes>(block_number)
+            .map_err(StoreError::LibmdbxError)?
+        else {
+            // Block must have been already purged
+            txn.commit().map_err(StoreError::LibmdbxError)?;
```
MegaRedHand, line 335: Do we need to commit here? Can't we just abort or discard the tx?

crates/storage/store_db/libmdbx.rs
```diff
@@ -322,6 +322,53 @@ impl StoreEngine for Store {
...
+            for (idx, tx_hash) in block_body
+                .transactions
+                .iter()
+                .map(|tx| tx.hash())
+                .enumerate()
+            {
+                txn.delete::<TransactionLocations>(tx_hash.into(), None)
+                    .map_err(StoreError::LibmdbxError)?;
+                txn.delete::<Receipts>((block_hash, idx as u64).into(), None)
+                    .map_err(StoreError::LibmdbxError)?;
```
MegaRedHand, line 355: If these aren't guaranteed to exist, should we ignore any errors then?

Cargo.toml
```diff
@@ -31,7 +31,7 @@ members = [
   "tooling/load_test",
   "tooling/loc",
   "tooling/archive_sync",
-  "crates/common/config",
+  "crates/common/config", "tooling/prune-blocks",
```
MegaRedHand: nit

```suggestion
  "crates/common/config",
  "tooling/prune-blocks",
```

tooling/prune-blocks/src/README.md
```diff
@@ -0,0 +1,15 @@
+# Prune Blocks
+
+This tool can be used to reduce the DB size of the node by removing old blocks and their asociated data. Note that this is counter-spec and will hinder the node's ability to provide data to other nodes. It also does not performe state prunning.
+
+## Usage
+
+The tool takes two optional arguments:
+    *`datadir`: The path to the DB location, will use the default one if not provided
+    *`blocks_to_keep`: The amount of latest blocks that will be kept in the DB. This value must be al least 128 and lower than the current amount of blocks in the chain.
```
MegaRedHand:
```suggestion
    *`blocks_to_keep`: The number of latest blocks that will be kept in the DB. This value must be at least 128 and lower than the current number of blocks in the chain.
```

tooling/prune-blocks/src/README.md
```diff
@@ -0,0 +1,15 @@
+# Prune Blocks
+
+This tool can be used to reduce the DB size of the node by removing old blocks and their asociated data. Note that this is counter-spec and will hinder the node's ability to provide data to other nodes. It also does not performe state prunning.
```
MegaRedHand:
```suggestion
This tool can be used to reduce the DB size of the node by removing old blocks and their associated data. Note that this is counter-spec and will hinder the node's ability to provide data to other nodes. It also does not perform state pruning.
```

tooling/prune-blocks/src/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Usage
+
+The tool takes two optional arguments:
+    *`datadir`: The path to the DB location, will use the default one if not provided
+    *`blocks_to_keep`: The amount of latest blocks that will be kept in the DB. This value must be al least 128 and lower than the current amount of blocks in the chain.
+
+And should be ran like this:
```
MegaRedHand:
```suggestion
And should be run like this:
```

crates/storage/store_db/libmdbx.rs
fmoletta: We don't get an error if they don't exist

tooling/prune-blocks/src/README.md
fmoletta: I prefer `amount` instead of `number` as number could be confused with an actual block number (aka keep blocks in store as of block number X instead of keeping X amount of blocks from the latest). Fixed the typo

crates/storage/store_db/libmdbx.rs
fmoletta: I had some problems because of this when testing it

mpaulucci:
Moving this to draft. Not sure this gives us enough benefit and the client broke once we implemented it

mpaulucci:
Not think we need this for now

PR #4145, opened 2025-08-24
perf: remove redundant clone in smoke_test

FredPhilipy:
Removes unnecessary double clone in `new_block` function.

`Blockchain::default_with_store` takes ownership of `Store`, so only one clone is needed.


mpaulucci:
hi, thx for the contribution, but its to minor to be considered

PR #2836, opened 2025-05-18
feat(l2): remove public inputs from verify() function in OnChainProposer

pedrobergamini:
**Motivation**

This PR implements the removal of public inputs from the `verifyBatch()` function in the OnChainProposer contract. Public inputs are now retrieved from the committed batch data instead of being passed as parameters.

**Description**

This PR removes public inputs from the `verifyBatch` function signature and instead reconstructs them from already-committed batch data. This simplifies the interface and reduces calldata size.

### OnChainProposer.sol
- **Modified `verifyBatch` function signature**: Removed all public input parameters
  - Removed `sp1PublicValues` parameter (SP1 now only needs `sp1ProofBytes`)
  - Removed `tdxPublicValues` parameter (TDX now only needs `tdxSignature`)
  - Note: Pico has been removed from the codebase in main branch
- **Added `_getPublicInputsFromCommitment` function**: Internal function that reconstructs 160-byte public inputs from committed batch data
  - bytes 0-32: Initial state root (from last verified batch)
  - bytes 32-64: Final state root (from current batch)
  - bytes 64-96: Withdrawals merkle root
  - bytes 96-128: Deposits log hash
  - bytes 128-160: Last block hash
- **Updated all verifiers to use reconstructed public inputs**:
  - RISC0: Uses `sha256(publicInputs)`
  - SP1: Uses `publicInputs` directly
  - TDX: Uses `publicInputs` directly

### zkVM Interface Updates
- **SP1**: Updated to use `commit_slice` for output
- **RISC0**: Updated to use `commit_slice` for output
- Both interfaces now output only the encoded result without public inputs prefix

### l1_proof_sender.rs
- Updated `VERIFY_FUNCTION_SIGNATURE` constant to match new contract signature
- Removed extraction logic for proof parts (no longer needed)
- Simplified calldata construction

### proving_systems.rs
- Updated `empty_calldata()` method for proving systems
- SP1 now only includes vkey and proof bytes
- TDX now only includes signature
- Removed Pico as it was removed from main branch

Closes #2804

Issue #2804, opened 2025-05-15
L2: remove public inputs from verify() function in OnChainProposer


xqft:
After completion of state diff check and proof of equivalence, we should have stabilized the public inputs (PIs) necessary to verify an L2 block execution proof (no more inputs should be needed for now). 

The current implementation of verify() takes 3 different fields of the public inputs for each proving backend / zkvm. After stable PIs, all PI data will be available as fields of the block commitment, so no PI argument should be necessary in the verify() function (we can retrieve PIs from the committed block).

The function will only take as arguments:
1. a proof for each backend
2. verification keys of the backends for which we didn't hardcode them yet (after merging #2708 these would be risc0 and pico).

PR #4147, opened 2025-08-25
Feat/remove public inputs clean

pedrobergamini:
### Motivation

This PR implements the removal of public inputs from the `verifyBatch()` function in the `OnChainProposer` contract. Public inputs are now retrieved from the committed batch data instead of being passed as parameters.

### Description

This PR removes public inputs from the `verifyBatch` function signature and reconstructs them from already-committed batch data. This simplifies the interface and reduces calldata size.

#### OnChainProposer.sol

- **Modified `verifyBatch` signature**: Removed the RISC0 journal parameter. Final signature:
  - `verifyBatch(uint256 batchNumber, bytes risc0BlockProof, bytes sp1ProofBytes, bytes tdxSignature)`
- **SP1**: Only `sp1ProofBytes` is provided. SP1 public values are constructed on-chain as `abi.encodePacked(uint64(publicInputs.length), publicInputs)`, and public data checks use `sp1PublicValues[8:]`.
- **TDX**: Only `tdxSignature` is provided.
- **Note**: Pico has been removed from the codebase in main.
- **Added `_getPublicInputsFromCommitment`**: Internal helper that reconstructs 256-byte public inputs from committed batch data:
  - bytes 0–32: Initial state root (from last verified batch)
  - bytes 32–64: Final state root (from current batch)
  - bytes 64–96: Withdrawals merkle root
  - bytes 96–128: Processed privileged transactions rolling hash
  - bytes 128–160: Blob versioned hash (stateDiff KZG versioned hash)
  - bytes 160–192: Last block hash
  - bytes 192–224: Chain ID (big-endian)
  - bytes 224–256: Non-privileged transactions count (set to 0)
- **Verifiers updated to use reconstructed public inputs**:
  - **RISC0**: `sha256(publicInputs)`
  - **SP1**: `sp1PublicValues = 8-byte length prefix + publicInputs`; validation over `sp1PublicValues[8:]`
  - **TDX**: `publicInputs` verified against `tdxSignature`

#### zkVM Interface Updates

- **SP1 guest**: Commits the 256-byte contract public inputs via `commit_slice(output.encode_contract_pis())`.
- **RISC0 guest**: Commits the 256-byte contract public inputs via `commit_slice(output.encode_contract_pis())`.
- **TDX**: Quote generator signs the same 256-byte contract public inputs (`encode_contract_pis()`).

#### l1_proof_sender.rs

- **Updated** `VERIFY_FUNCTION_SIGNATURE` to `"verifyBatch(uint256,bytes,bytes,bytes)"`.
- **Removed** RISC0 journal from calldata.
- **Simplified** calldata construction.

#### Proving systems

- **Updated** `empty_calldata()` for proving systems:
  - **RISC0**: Only the Groth16 seal.
  - **SP1**: Only `sp1ProofBytes`.
  - **TDX**: Only the signature.
  - Pico was removed in main.

Previous [PR](https://github.com/lambdaclass/ethrex/pull/2836)
Closes [#2804](https://github.com/lambdaclass/ethrex/issues/2804)


crates/l2/contracts/src/l1/OnChainProposer.sol
```diff
@@ -455,6 +456,70 @@ contract OnChainProposer is
...
+            publicInputs[160 + i] = bytes1(uint8(uint256(lastBlockHash) >> (8 * (31 - i))));
+        }
+
+        // Chain ID
+        bytes32 chainIdBytes = bytes32(CHAIN_ID);
+        for (uint256 i = 0; i < 32; i++) {
+            publicInputs[192 + i] = bytes1(uint8(uint256(chainIdBytes) >> (8 * (31 - i))));
+        }
+
+        // Non-privileged transactions count = 0 (bytes 224..256 already zeroed)
```
xqft: this shouldn't be fixed to zero right? the field `processedPrivilegedTransactionsRollingHash` in the batch commitment contains this value: it's the first two bytes (it's used in line 280 of this file too)

crates/l2/contracts/src/l1/OnChainProposer.sol
```diff
@@ -455,6 +456,70 @@ contract OnChainProposer is
...
+        BatchCommitmentInfo memory previousBatch = batchCommitments[lastVerifiedBatch];
+
+        bytes memory publicInputs = new bytes(256);
+
+        // Initial state root
+        bytes32 initialStateRoot = previousBatch.newStateRoot;
+        for (uint256 i = 0; i < 32; i++) {
+            publicInputs[i] = bytes1(uint8(uint256(initialStateRoot) >> (8 * (31 - i))));
+        }
+
```
xqft: would it be possible to use `bytes.concat()` for all these?

crates/l2/prover/zkvm/interface/src/io.rs
```diff
@@ -114,4 +114,29 @@ impl ProgramOutput {
...
+            self.final_state_hash.to_fixed_bytes(),
+            self.l1messages_merkle_root.to_fixed_bytes(),
+            self.privileged_transactions_hash.to_fixed_bytes(),
+            self.blob_versioned_hash.to_fixed_bytes(),
+            self.last_block_hash.to_fixed_bytes(),
+            self.chain_id.to_big_endian(),
+            [0u8; 32],
+        ]
+        .concat()
+    }
```
xqft: this is part of the other comment, I see that the reason this function exists is to hardcode the non-privileged tx count to 0

crates/l2/contracts/src/l1/based/OnChainProposer.sol
```diff
@@ -355,7 +355,7 @@ contract OnChainProposer is
 
         if (SP1VERIFIER != DEV_MODE) {
             // If the verification fails, it will revert.
-            _verifyPublicData(batchNumber, sp1PublicValues[16:]);
+            _verifyPublicData(batchNumber, sp1PublicValues[8:]);
```
xqft: the slicing shouldn't be necessary after changing the SP1 zkvm program to use `commit_slice()` instead of `commit()` (`commit()` uses bincode to serialize and it adds a prefix to the bytes which we ignore by slicing here in the contract, while `commit_slice()` write the bytes as is, without that prefix)

crates/l2/prover/zkvm/interface/risc0/src/main.rs
```diff
@@ -5,5 +5,6 @@ fn main() {
     let input: JSONProgramInput = env::read();
     let output = execution_program(input.0).unwrap();
 
-    env::commit_slice(&output.encode());
+    // Commit the 256-byte contract public inputs
```
xqft: a nit: I wouldn't add "256-byte" because public inputs may change in the future and this comment would be outdated after

(this comment applies for all cases in which the comment was added)

crates/l2/contracts/src/l1/OnChainProposer.sol
pedrobergamini: Yep sure! Sorry, done

crates/l2/contracts/src/l1/OnChainProposer.sol
pedrobergamini: Thanks for flagging this. The last 32 bytes in the contract public inputs are intentionally zero, right? The privileged transaction count is taken from a different place, it’s encoded in the first two bytes of processedPrivilegedTransactionsRollingHash (which we already read to remove pending privileged transactions). Separately, we enforce onchain that if privileged transactions have expired, then the non‑privileged count must be zero. I clarified this intent in comments and kept the last 32 bytes zero by design. Could you confirm you agree the privileged count comes from the rolling‑hash prefix and that the final 32‑byte slot should remai zero?

crates/l2/prover/zkvm/interface/risc0/src/main.rs
pedrobergamini: Done!

crates/l2/contracts/src/l1/based/OnChainProposer.sol
pedrobergamini: Done, thanks sir 🫡

crates/l2/contracts/src/l1/OnChainProposer.sol
xqft: Sorry! I mixed `privileged count` up with `non-privileged count`. Yes it's correct that the non-privileged count should be 0, but **in the case where the privileged transaction inclusion time has expired**, else the non-privileged count can be any other number, and because it's part of the public inputs of the proof, we need it for verification.

Problem here is that we don't have this number as part of the block commitment, or anywhere in the L1 (we do have the privileged count but we can't derive the non-privileged from that alone), so by removing the public inputs parameter from the verify() function we don't have a way to access this value. 

We talked internally with @avilagaston9 and the solution here should be to add the non-privileged count as a new field of the batch commitment. This would require, if I didn't miss anything:
1. Adding this new field to the `BatchCommitmentInfo` struct in Solidity
2. Adding the field as a parameter of the `commitBatch()` function
3. Update the `send_commitment()` function of the L1 commiter module in Rust to add the new field

I think this is necessary to finish this PR so those changes could be added in this same branch. If you have any questions don't hesitate on asking us!

(about the "expire time" thing: this is a mechanism in which we give the sequencer some amount of time to process privileged transactions, else it won't be able to settle non-privileged transactions, so no new transactions until it processes the pending privileged transactions, this is a way to avoid the sequencer to censor L1->L2 messages. For example, if a user deposits some ETH from L1 to L2, they know that the sequencer will include that deposit and their funds won't be lost, else the whole L2 stops)

crates/l2/contracts/src/l1/OnChainProposer.sol
xqft: thanks!!

crates/l2/contracts/src/l1/OnChainProposer.sol
pedrobergamini: Thanks for the clarifications sir! All clear, pushing the update in a sec

crates/l2/contracts/src/l1/OnChainProposer.sol
```diff
@@ -291,31 +292,33 @@ contract OnChainProposer is
...
                 RISC0_VERIFICATION_KEY,
-                sha256(risc0Journal)
+                sha256(publicInputs)
             );
         }
 
         if (SP1VERIFIER != DEV_MODE) {
             // If the verification fails, it will revert.
-            _verifyPublicData(batchNumber, sp1PublicValues[8:]);
+            _verifyPublicData(batchNumber, publicInputs);
```
avilagaston9, line 310: Given that we are constructing the `publicInputs` from the values of `BatchCommitmentInfo` there is no need to call `_verifyPublicData()` as it will always pass. We can remove the function. The only check that should remain is the `hasExpiredPrivilegedTransactions()` call.

avilagaston9: The comment applies for both `based` and `non-based` contract.

crates/l2/prover/zkvm/interface/sp1/src/main.rs
```diff
@@ -11,5 +11,6 @@ pub fn main() {
 
     let output = execution_program(input).unwrap();
 
-    sp1_zkvm::io::commit(&output.encode());
+    // Commit the contract public inputs
+    sp1_zkvm::io::commit_slice(&output.encode_contract_pis());
```
avilagaston9, line 15: Let's restore this, remove the `encode_contract_pis()` function, and update the `ProgramOutput::encode()` method to reflect the new changes instead. This change will break the proving of L1 blocks.

crates/l2/sequencer/l1_committer.rs
```diff
@@ -433,6 +433,30 @@ impl L1Committer {
...
+                .len()
+                .try_into()
+                .unwrap_or(u64::MAX);
+            total_txs = total_txs.saturating_add(txs_in_block);
+            let priv_in_block = get_block_privileged_transactions(&block_body.transactions);
+            let priv_in_block_count: u64 = priv_in_block.len().try_into().unwrap_or(u64::MAX);
+            privileged_txs = privileged_txs.saturating_add(priv_in_block_count);
+        }
+        let non_privileged_count = total_txs.saturating_sub(privileged_txs);
+
```
avilagaston9, line 459: We can add a new field, `non_privileged_count`, to our `Batch` struct and return its value as the result of [prepare_batch_from_block()](https://github.com/lambdaclass/ethrex/blob/e77ff564825296660bfee50c9623cfab2d5e6aaa/crates/l2/sequencer/l1_committer.rs#L264C14-L264C38) to avoid fetching and iterating over the blocks again.

crates/l2/sequencer/l1_committer.rs
```diff
@@ -455,12 +479,16 @@ impl L1Committer {
...
+            // Insert count before encoded blocks for based ABI
+            calldata_values.push(Value::Uint(U256::from(non_privileged_count)));
             calldata_values.push(Value::Array(
                 encoded_blocks.into_iter().map(Value::Bytes).collect(),
             ));
 
             (COMMIT_FUNCTION_SIGNATURE_BASED, calldata_values)
         } else {
+            // Append count as last arg for non-based ABI
+            calldata_values.push(Value::Uint(U256::from(non_privileged_count)));
```
avilagaston9, line 491: We can call `calldata_values.push(Value::Uint(U256::from(non_privileged_count)));` only once before line 460.

avilagaston9 (changes requested):
Hi @pedrobergamini! Thanks for your contribution. Left some comments

crates/l2/prover/zkvm/interface/sp1/src/main.rs
xqft: I would keep the commit_slice() function though, so not entirely restore the line but just change it to commit_slice(output.encode())

crates/l2/contracts/src/l1/based/OnChainProposer.sol
```diff
@@ -500,6 +514,37 @@ contract OnChainProposer is
...
+    function _getPublicInputsFromCommitment(
+        uint256 batchNumber
+    ) internal view returns (bytes memory) {
+        BatchCommitmentInfo memory currentBatch = batchCommitments[batchNumber];
+        BatchCommitmentInfo memory previousBatch = batchCommitments[lastVerifiedBatch];
+
+        // Note: the last 32 bytes encode the non-privileged transaction count.
+        // The privileged transaction count is encoded separately in the first two bytes of
+        // `processedPrivilegedTransactionsRollingHash` and is handled during verification.
+        return abi.encodePacked(
```
xqft, line 536: in the non-based version you used bytes.concat() here, we should use the same function for both implementations, I think bytes.concat() is clearer

avilagaston9:
Hi @pedrobergamini!
I’m closing this PR and continuing the work in #5804. I added you as a co-author there. Thank you for your contribution!

PR #4150, opened 2025-08-25
perf(levm): improve calldatacopy further

edg-l:
**Motivation**

Added a specific memory function to zero fill, also simplified the code + removed some unnecesary checked arithmetic.

image

### DIFF
--- a/crates/vm/levm/src/memory.rs
+++ b/crates/vm/levm/src/memory.rs
@@ -255,6 +255,29 @@ impl Memory {
 
         Ok(())
     }
+
+    #[inline(always)]
+    pub fn store_zeros(&mut self, offset: usize, size: usize) -> Result<(), VMError> {
+        if size == 0 {
+            return Ok(());
+        }
+
+        let new_size = offset.checked_add(size).ok_or(OutOfBounds)?;
+        self.resize(new_size)?;
+
+        let real_offset = self.current_base.wrapping_add(offset);
+        let mut buffer = self.buffer.borrow_mut();
+
+        // resize ensures bounds are correct
+        #[allow(unsafe_code)]
+        unsafe {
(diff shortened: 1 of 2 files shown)


crates/vm/levm/src/memory.rs
```diff
@@ -256,6 +256,29 @@ impl Memory {
...
+        }
+
+        let new_size = offset.checked_add(size).ok_or(OutOfBounds)?;
+        self.resize(new_size)?;
+
+        let real_offset = self.current_base.wrapping_add(offset);
+        let mut buffer = self.buffer.borrow_mut();
+
+        // resize ensures bounds are correct
+        #[allow(unsafe_code)]
```
MegaRedHand:
```suggestion
        #[expect(unsafe_code)]
```

crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -152,43 +152,51 @@ impl<'a> VM<'a> {
...
 
-        current_call_frame.memory.store_data(dest_offset, &data)?;
+                if copy_size > 0 {
+                    data[..copy_size].copy_from_slice(
+                        &current_call_frame.calldata[calldata_offset..calldata_offset + copy_size],
+                    );
+                }
 
-        Ok(OpcodeResult::Continue { pc_increment: 1 })
+                current_call_frame.memory.store_data(dest_offset, &data)?;
```
MegaRedHand, line 195: Did we check if changing this into a `store_data` (for the data) and a `store_zeros` (for the padding) improved things?

edg-l: no but i think it will be worse since store_zeroes also calls resize

PR #4163, opened 2025-08-25
feat(l1): added fusaka support for runners

### LINKED ISSUE #4114: Update fixtures to include fusaka
https://github.com/lambdaclass/ethrex/blob/main/cmd/ef_tests/blockchain/.fixtures_url

The fusaka tests should be configured but skipped since we don't support fusaka yet.

This ticket is to make it easy for fusaka implementers to unskip the appropriate tests. All the boilerplate should be ready for them

Also add fusaka tests to the daily Hive report

SDartayet:
**Motivation**

Adding the boilerplate in the runners to be able to execute the Fusaka tests in the future when we start implementing it

**Description**

This PR changes the fixture URLs to the most recent execution spec tests release that supports Fusaka. Additionally, it adds the parameters needed to the structs the runners use to be able to parse and execute the test; although at the moment they're skipped by default. The Osaka tests were also added to the Hive daily report.

Closes #4114

### DIFF
--- a/crates/common/types/genesis.rs
+++ b/crates/common/types/genesis.rs
@@ -188,6 +188,12 @@ pub struct ChainConfig {
     pub verkle_time: Option<u64>,
     pub osaka_time: Option<u64>,
 
+    pub bpo1_time: Option<u64>,
+    pub bpo2_time: Option<u64>,
+    pub bpo3_time: Option<u64>,
+    pub bpo4_time: Option<u64>,
+    pub bpo5_time: Option<u64>,
+
     /// Amount of total difficulty reached by the network that triggers the consensus upgrade.
     pub terminal_total_difficulty: Option<u128>,
     /// Network has already passed the terminal total difficult
@@ -198,6 +204,9 @@ pub struct ChainConfig {
     #[rkyv(with = rkyv_utils::H160Wrapper)]
     // Deposits system contract address
     pub deposit_contract_address: Address,
+
+    #[serde(default)]
+    pub enable_verkle_at_genesis: bool,
 }
 
 #[repr(u8)]
--- a/cmd/ef_tests/blockchain/.fixtures_url
+++ b/cmd/ef_tests/blockchain/.fixtures_url
@@ -1,1 +1,1 @@
-https://github.com/ethereum/execution-spec-tests/releases/download/v4.5.0/fixtures_develop.tar.gz
+https://github.com/ethereum/execution-spec-tests/releases/download/fusaka-devnet-5%40v1.1.0/fixtures_fusaka-devnet-5.tar.gz
--- a/cmd/ef_tests/blockchain/network.rs
+++ b/cmd/ef_tests/blockchain/network.rs
@@ -53,6 +53,33 @@ lazy_static! {
         prague_time: Some(0),
         ..*CANCUN_TO_PRAGUE_AT_15K_CONFIG
     };
+
+    pub static ref PRAGUE_TO_OSAKA_AT_15K_CONFIG: ChainConfig = ChainConfig {
+        osaka_time: Some(0x3a98),
+        ..*PRAGUE_CONFIG
+
+    };
+
+    pub static ref OSAKA_CONFIG: ChainConfig = ChainConfig {
+        osaka_time: Some(0),
+        ..*PRAGUE_CONFIG
+    };
+
+    pub static ref OSAKA_TO_BPO1_AT_15K_CONFIG: ChainConfig = ChainConfig {
+        bpo1_time: Some(0x3a98),
+        ..*OSAKA_CONFIG
+    };
+
+    pub static ref BPO1_TO_BPO2_AT_15K_CONFIG: ChainConfig = ChainConfig {
+        bpo2_time: Some(0x3a98),
+        ..*PRAGUE_CONFIG
+    };
+
+    pub static ref BPO2_TO_BPO3_AT_15K_CONFIG: ChainConfig = ChainConfig {
+        bpo3_time: Some(0x3a98),
+        ..*PRAGUE_CONFIG
+    };
+
 }
 
 #[derive(Debug, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
@@ -73,6 +100,11 @@ pub enum Network {
     Cancun = 11,
     CancunToPragueAtTime15k = 12,
     Prague = 13,
+    PragueToOsakaAtTime15k = 14,
+    Osaka = 15,
+    OsakaToBPO1AtTime15k = 16,
+    BPO1ToBPO2AtTime15k = 17,
+    BPO2ToBPO3AtTime15k = 18,
 }
 
 impl Network {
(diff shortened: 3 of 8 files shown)


crates/common/types/genesis.rs
```diff
@@ -188,6 +188,12 @@ pub struct ChainConfig {
     pub verkle_time: Option<u64>,
     pub osaka_time: Option<u64>,
 
+    pub bpo1_time: Option<u64>,
+    pub bpo2_time: Option<u64>,
+    pub bpo3_time: Option<u64>,
+    pub bpo4_time: Option<u64>,
+    pub bpo5_time: Option<u64>,
+
```
SDartayet, line 196: I based the parameters off of what was included in the newest fixtures, as well as [Geth's implementation](https://pkg.go.dev/github.com/ethereum/go-ethereum/params#ChainConfig)

crates/networking/rpc/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "400"]
```
SDartayet, line 2: The recursion limit setting was necessary for the Json parsing to work, otherwise some of the tests would fail

tooling/ef_tests/state_v2/src/modules/runner.rs
```diff
@@ -38,7 +38,18 @@ pub async fn run_tests(tests: Vec<Test>) -> Result<(), RunnerError> {
     let mut failing_tests = 0;
     let mut total_run = 0;
 
+    //Test with the Fusaka tests that should pass. TODO: Once we've implemented all the Fusaka EIPs this should be removed
+    //EIPs should be added as strings in the format 'eip-XXXX'
+    let fusaka_eips_to_test: Vec<&str> = vec!["eip-7939"];
```
mpaulucci, line 43: nice!

cdiielsi (approved):

tooling/ef_tests/state/report.rs
```diff
@@ -211,8 +211,7 @@ pub fn summary_for_shell(reports: &[EFTestReport]) -> String {
         },
         // NOTE: Keep in order, see the Fork Enum to check
         // NOTE: Uncomment the summaries if EF tests for those specific forks exist.
-
-        // fork_summary_shell(reports, Fork::Osaka),
+        fork_summary_shell(reports, Fork::Osaka),
```
cdiielsi, line 214: Tiny issue I just noticed, a report summary is missing for Osaka when a test fails. `writeln!(f, "{}", fork_summary_shell(&self.0, Fork::Osaka))?;` should be added [here](https://github.com/lambdaclass/ethrex/blob/c140755e0519660058d7f3acffd9bf46f401f335/tooling/ef_tests/state/report.rs#L326)

crates/networking/rpc/lib.rs
mpaulucci: Hmm it is weird for this to be on this file, without a comment (in code) explaining why it is there. Can't you add such a setting in the test code instead?

crates/common/types/genesis.rs
```diff
@@ -198,6 +204,9 @@ pub struct ChainConfig {
     #[rkyv(with = rkyv_utils::H160Wrapper)]
     // Deposits system contract address
     pub deposit_contract_address: Address,
+
+    #[serde(default)]
+    pub enable_verkle_at_genesis: bool,
```
mpaulucci, line 209: is this necessary for the tests to run?

crates/networking/rpc/lib.rs
SDartayet: Sadly I don't think it's possible, since I'm pretty sure the setting needs to be in the module doing the recursion. I could look more into it though; or alternatively add a comment explaining the reason for that setting. Not sure which you'd think is better.

tooling/ef_tests/state/report.rs
SDartayet: Solved [here](https://github.com/lambdaclass/ethrex/pull/4163/commits/3effd10d5643329a999fffea4ade0e3c3de3dc11)!

crates/common/types/genesis.rs
mpaulucci: verkle is something that is not planned for fusaka, this shouldn't be necessary

tooling/ef_tests/state/.fixtures_url
```diff
@@ -1 +1 @@
-https://github.com/ethereum/execution-spec-tests/releases/download/v4.5.0/fixtures_develop.tar.gz
+https://github.com/ethereum/execution-spec-tests/releases/download/fusaka-devnet-5%40v1.1.0/fixtures_fusaka-devnet-5.tar.gz
```
mpaulucci, line 1: Can't we use a single file for fixtures? or is there any benefit of using two?

crates/common/types/genesis.rs
SDartayet: Some of the Fusaka fixtures include it though. I added it because tests were failing due to its absence

tooling/ef_tests/state/.fixtures_url
SDartayet: Ah, I just didn't wanna mess with the code too much since it seemed potentially out of scope and I wasn't sure I wouldn't accidentally break things, but I don't think there's any reason we need two different fixtures files. If you want I can refactor it into just one in the ef_tests folder as part of this PR

tooling/ef_tests/state_v2/src/modules/runner.rs
SDartayet: Thanks!

crates/common/types/genesis.rs
mpaulucci: please exclude those test them. I don't think we should have a reference to verkle in the genesis

crates/networking/rpc/lib.rs
mpaulucci: Add a comment then explaining why did you have to add that.

tooling/ef_tests/state/.fixtures_url
mpaulucci: Ok we can do it in a follow up, please create a ticket

crates/common/types/genesis.rs
SDartayet: To clarify (not sure if it's what you meant) the `enable_verkle_at_genesis` attribute is part of the ChainConfig struct, not the Genesis one. From what I could find this is because some of the devnet tests have verkle enabled from the get go, to more easily tests things. I could exclude said tests for now though; but I think we'll end up needing it down the line anyways to run those tests

tooling/ef_tests/state/.fixtures_url
SDartayet: Done!
https://github.com/lambdaclass/ethrex/issues/4212

tooling/ef_tests/state/runner/levm_runner.rs
```diff
@@ -36,7 +36,28 @@ pub async fn run_ef_test(test: &EFTest) -> Result<EFTestReport, EFTestRunnerErro
...
     for fork in test.post.forks.keys() {
+        if fork == &Fork::Osaka
+            && !fusaka_eips_to_test.iter().any(|eip| test_eip.contains(eip))
+            && !names_of_fusaka_tests_to_run
+                .iter()
+                .any(|name| *name == test.name)
+        {
+            continue;
+        }
+        //dbg!(test.name.clone());
```
fmoletta: Please remove

fmoletta (approved):
🚀

tooling/ef_tests/state/runner/levm_runner.rs
SDartayet: Thanks for pointing it out, idk how I missed that one

crates/common/types/genesis.rs
mpaulucci: ok

mpaulucci (approved):

PR #4164, opened 2025-08-25
fix(l2): avoid blocking in `L1Committer`

MegaRedHand:
**Motivation**

The L1 Committer does some blocking computations when publishing batches to L1, which took from 100ms to 30ms according to quick measurements on an M3. This is a low number, but still degrades the tokio runtime and could cause issues elsewhere in the system.

**Description**

~This PR wraps the call to `generate_blobs_bundle` with a `spawn_blocking`, to avoid blocking the tokio runtime. Also adds a comment on another call that seems to also be blocking, taking sometimes \~40ms before reaching an `await` point.~ Changes reverted. Now we instead spawn the whole GenServer as blocking.


crates/l2/sequencer/l1_committer.rs
```diff
@@ -355,7 +355,8 @@ impl L1Committer {
                     &acc_privileged_txs,
                     acc_account_updates.clone().into_values().collect(),
                 )?;
-                generate_blobs_bundle(&state_diff)
+                spawned_rt::tasks::spawn_blocking(move || generate_blobs_bundle(&state_diff))
```
ElFantasma: Mmmhhh, can't this be avoided using [start_blocking](https://github.com/lambdaclass/spawned/blob/main/concurrency/src/tasks/gen_server.rs#L146) for the whole `GenServer` instead?

MegaRedHand: That might be a good alternative

MegaRedHand: Changed

tomip01 (approved):

avilagaston9 (approved):

PR #4175, opened 2025-08-27
perf(l1): refactor transaction broadcast filter

### LINKED ISSUE #4166: Improve send_new_pooled_tx_hashes performance
See this samply: https://share.firefox.dev/4myKYzp

The main issue is the Estalbished broadcasted_txs hashset, i think it grows quite large. This is due to keeping track of what txs have been sent to the given peer.

Maybe we can find a way to avoid needing the filtering at all.

Some ideas:
When getting a new tx in the mempool, add it to (lockless? see crossbeam queue) queue on all connected peers.
When the peer requests SendNewPooledTxHashes, simply pop from the queue, this should be way faster than filtering from the mempool.

One thing to take into account, is when a peer newly connects, in initialzie_connection, we send all the current mempool txs, we should probably not call send_new_pooled_tx_hashes but a specialzied method that simply sends all mempool tx hashes.

https://github.com/lambdaclass/ethrex/blob/7d444698ff94e416eb67d784bf9ad631e09dccf3/crates/networking/p2p/rlpx/connection/server.rs#L462-L500

azteca1998:
**Motivation**

Currently, when broadcasting transactions to our peers, we need to scan and filter the entire mempool. This is not strictly necessary.

**Description**

Make it so that we don't have to scan the entire mempool when broadcasting transactions.

Closes #4166


crates/blockchain/mempool.rs
```diff
@@ -22,7 +23,10 @@ pub struct Mempool {
     transaction_pool: RwLock<HashMap<H256, MempoolTransaction>>,
     blobs_bundle_pool: Mutex<HashMap<H256, BlobsBundle>>,
     txs_by_sender_nonce: RwLock<BTreeMap<(H160, u64), H256>>,
+
+    event_handles: EventHandlerSet<H256>,
```
Oppen, line 27: Typo?
```suggestion
    event_handlers: EventHandlerSet<H256>,
```

Arkenan (changes requested):
This PR is in the right direction, but I think it does a bit more than it needs to, as some of the messaging and locking functionality already exists in the connection genserver. That is, I would maintain most of it but change:

- I would remove the queue in the state
- Instead of adding an element to the queue when a transaction is received, I would send the transaction to all connection genservers as a cast.

ElFantasma:
Tx broadcast logic was too much complexity for the `RLPxConnection` `GenServer`s so we extracted (and centralized) it in https://github.com/lambdaclass/ethrex/pull/4259 to a new component `TxBroadcast`  that works coupled with the Mempool.

I think this PR goes in the right direction but in order to merge this with current main we'll have to move part of this behavior to `TxBroadcast`.

Anyway, we should investigate a bit to ensure it is a required feature to check that no transaction gets forwarded twice to the same peer. If that's not the case, then this behavior is not necessary.

Arkenan:
This PR is no longer relevant after the latest broadcaster refactor

PR #4176, opened 2025-08-27
fix(l2): use deserialization error in decode path

Fibonacci747:
Replace FailedToSerializeStateDiff with FailedToDeserializeStateDiff in StateDiff::decode() when validating remaining bytes before AccountStateDiff::decode(). This aligns with error semantics used throughout Decoder methods and ensures consistent error reporting during deserialization.

### DIFF
--- a/crates/l2/common/src/state_diff.rs
+++ b/crates/l2/common/src/state_diff.rs
@@ -31,8 +31,6 @@ pub const SIMPLE_TX_STATE_DIFF_SIZE: u64 = 108;
 pub enum StateDiffError {
     #[error("StateDiff failed to deserialize: {0}")]
     FailedToDeserializeStateDiff(String),
-    #[error("StateDiff failed to serialize: {0}")]
-    FailedToSerializeStateDiff(String),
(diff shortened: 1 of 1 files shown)


crates/l2/common/src/state_diff.rs
```diff
@@ -210,7 +210,7 @@ impl StateDiff {
         let mut modified_accounts = BTreeMap::new();
         for _ in 0..modified_accounts_len {
             let next_bytes = bytes.get(decoder.consumed()..).ok_or(
-                StateDiffError::FailedToSerializeStateDiff("Not enough bytes".to_string()),
+                StateDiffError::FailedToDeserializeStateDiff("Not enough bytes".to_string()),
```
MegaRedHand, line 211: We may be able to remove the error variant `FailedToSerializeStateDiff`

Fibonacci747: > We may be able to remove the error variant `FailedToSerializeStateDiff`

Yes, really, done

MegaRedHand:
Closed due to staleness

PR #4177, opened 2025-08-27
perf(levm): adjust memory resize

edg-l:
Current memory resize is overly eager, tune it to resize to multiples of 64 instead, this avoids really small allocations while not over allocating, since we require to zero the allocated bytes, it can be expensive to overallocate

**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/vm/levm/src/memory.rs
+++ b/crates/vm/levm/src/memory.rs
@@ -92,8 +92,8 @@ impl Memory {
         let real_new_memory_size = new_memory_size + self.current_base;
 
         if real_new_memory_size > buffer.len() {
-            // when resizing, resize by allocating entire pages instead of small memory sizes.
-            let new_size = real_new_memory_size.next_multiple_of(4096);
+            // when resizing, avoid really small resizes.
+            let new_size = real_new_memory_size.next_multiple_of(64);
             buffer.resize(new_size, 0);
         }
 


crates/vm/levm/src/memory.rs
```diff
@@ -92,8 +92,8 @@ impl Memory {
         let real_new_memory_size = new_memory_size + self.current_base;
 
         if real_new_memory_size > buffer.len() {
-            // when resizing, resize by allocating entire pages instead of small memory sizes.
-            let new_size = real_new_memory_size.next_multiple_of(4096);
+            // when resizing, avoid really small resizes.
+            let new_size = real_new_memory_size.next_multiple_of(64);
```
MegaRedHand, line 96: Which values did we try here? Maybe there's a better choice for this constant. Or we could also try changing it dynamically (start with 32, then 64, then 128, etc.).

edg-l: i tried, 1024, 128, 64 (and the old 4096), on the samply i saw 64 to fit well

rodrigo-o: Given that we didn't test with smaller values (and that as of now the smaller was the better) would it be useful to check for 32 or even less? just to make sure 64 is actually the sweet spot? Or we didn't have great improvements between 128 and 64?

edg-l: it didnt changem uch from 128 to 64

Issue #3629, opened 2025-07-14
L1: Use 256 bits for gas price in Legacy and EIP-2930 transactions


JereSalo:
For some reason we are using `u64` for gas price in Legacy and EIP2930 Transactions and I believe we should be using `U256` instead. This is not spec compliant

We closed this in the past because of performance reasons but we haven't yet run anything analyzing performance, so maybe it's a change worth implementing. The scenario in which gas price is more than 64 bits is very unlikely but if the cost of supporting it is negligible maybe we could implement it. I'd suggest to try change it and measure if there's any significant difference and then we'll know for sure if we want this or not

temaniarpit27:
Can I take this up?

emirongrr:
According to the Ethereum Yellow Paper (Section 4.2), the gasPrice field (Tₚ) is defined as ℕ₂₅₆, meaning it should support full 256-bit unsigned integers.

Despite the spec, clients differ in how they internally represent gas_price:

- Geth uses u64 for Tx.GasPrice
- Reth uses u128, which is much safer and future-proof. Reth even includes a comment:
- Nethermind uses u256, which aligns with the spec exactly (uint256) 
```
    /// A scalar value equal to the number of
    /// Wei to be paid per unit of gas for all computation
    /// costs incurred as a result of the execution of this transaction; formally Tp.
    /// As ethereum circulation is around 120mil eth as of 2022 that is around
    /// 120000000000000000000000000 wei we are safe to use u128 as its max number is:
    /// 340282366920938463463374607431768211455
    gas_price: u128
```
While u64 or u128 may be practically sufficient for most real-world scenarios today, they do impose artificial limits that are stricter than what the protocol allows. In practice, extremely large gas prices like 2^64 - 1 Wei have never been observed in real Ethereum blocks, as such values would imply unrealistically high transaction fees. However, if the goal is to prepare for even such extreme artificial scenarios, this value should be changed to support the full range.

JereSalo:
> According to the Ethereum Yellow Paper (Section 4.2), the gasPrice field (Tₚ) is defined as ℕ₂₅₆, meaning it should support full 256-bit unsigned integers.
> 
> Despite the spec, clients differ in how they internally represent gas_price:
> 
> * Geth uses u64 for Tx.GasPrice
> * Reth uses u128, which is much safer and future-proof. Reth even includes a comment:
> * Nethermind uses u256, which aligns with the spec exactly (uint256)
> 
> ```
[...]

PR #4179, opened 2025-08-27
test(levm,l1): run state tests as if they were blocks instead of transactions

### LINKED ISSUE #4183: State_v2: Run EFTests State tests as blocks
WE want to be able to run them with SP1, and the code that we have to do this runs blocks, so it would be best to convert these tests to blocks and run them.

JereSalo:
**Motivation**

- The final goal is to run state tests with SP1, that's why we need to run them as blocks. This is a first step in that direction.

**Description**

- Fix `sign_inplace` for Legacy Transactions.
- Remove code that checks that withdrawal request and consolidation request accounts have code. This breaks 2 tests but it's completely worth it I'd say. It is an unrealistic case that doesn't happen in mainnet nor testnets.
  - [Geth skips these tests too](https://github.com/ethereum/go-ethereum/blob/72d92698a474059f3a73798c6312699c1f210497/tests/block_test.go#L84-L85)
  - I had to remove this because in the state tests scenarios these accounts don't exist.
- I had to skip `HighGasPrice` EFTests because we need to fix the gas price for our Legacy and Type 1 transactions, it should be U256 #3629
- Now we accurately get the transaction from a test case, without doing the shady stuff we were previously doing.
  - We also sign the transaction with the private key of the sender because it has to be valid.
- Corrects all kinds of things we were doing wrong because when running the transaction only we didn't have problems. For example, make an accurate `Genesis` block. (Which wasn't very trivial I'd say)

Closes #4183


tooling/ef_tests/state_v2/src/modules/runner.rs
```diff
@@ -164,15 +166,74 @@ pub fn get_tx_from_test_case(test_case: &TestCase) -> Result<Transaction, Runner
...
+    } else if test_case.max_fee_per_blob_gas.is_some() {
+        Transaction::EIP4844Transaction(EIP4844Transaction {
+            chain_id,
+            nonce,
+            max_priority_fee_per_gas: test_case.max_priority_fee_per_gas.unwrap().as_u64(),
+            max_fee_per_gas: test_case.max_fee_per_gas.unwrap().as_u64(),
+            gas: test_case.gas,
+            to: match to {
+                TxKind::Call(to) => to,
+                TxKind::Create => return Err(RunnerError::EIP7702ShouldNotBeCreateType), //TODO: See what to do with this. Maybe we want to get rid of the error and skip the test.
```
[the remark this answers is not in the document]
JereSalo: The code is something quick that's not going to be used so being tidy here is not important, it will require more iterations and then we can make it tidier

crates/vm/backends/levm/mod.rs
```diff
@@ -255,16 +255,6 @@ impl LEVM {
             vm_type,
         )?;
 
-        // According to EIP-7002 we need to check if the WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS
```
mpaulucci, line 258: why is this removed?

JereSalo: Because these contracts do not exist in the state tests, so they will always have empty code in these scenarios and therefore block execution will fail because of this. If I added the contracts to the state it would mean that the root hash would be different than the one precalculated in the test fixture and I don't want that.

In mainnet and testnets this check only adds overhead because the contracts do have code. That's why clients like geth don't have this logic either. Not useful in real-world scenarios.

PR #4186, opened 2025-08-27
feat(l2): add support for multiple networks on a single prover

ManuelBilbao:
**Motivation**

Prover may be idle for a long time if batch generation frequency is low, wasting resources.

**Description**

Allow the prover to have multiple proof coordinator endpoints. This can be used both for multiple networks as for redundancy. Since the prover is stateless, it doesn't matter if the endpoints are from different networks.
The prover will try to get a task sequentially on each endpoint. After all endpoints are checked and proofs generated (in case there's anyone to generate), the prover will sleep `--proving-time` milliseconds and start again


crates/l2/prover/src/prover.rs
```diff
@@ -33,52 +33,61 @@ impl Prover {
...
+                    .request_new_input(endpoint)
+                    .await
+                    .inspect_err(|e| error!(%endpoint, "Failed to request new data from: {e}"))
+                else {
+                    continue;
+                };
+
+                let Some(prover_data) = prover_data else {
+                    continue;
+                };
```
LeanSerra: Nit
```suggestion
                let Ok(Some(prover_data)) = self
                    .request_new_input(endpoint)
                    .await
                    .inspect_err(|e| error!(%endpoint, "Failed to request new data from: {e}"))
                else {
                    continue;
                };
```

LeanSerra (approved):
LGTM, just a small nit

PR #4190, opened 2025-08-27
feat(l1): [EIP-7823] Set upper bounds for MODEXP

### LINKED ISSUE #4153: [EIP-7823] Set upper bounds for MODEXP
https://forkcast.org/upgrade/fusaka#eip-7823

Acceptance criteria:
- Pass all spec tests for 7823

cdiielsi:
**Motivation**

Implement EIP-7823 for Osaka fork.

**Description**

This pr implements the bound introduced in [EIP-7823](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7823.md) for the inputs of the MODEXP precompile. Since it's a backwards incompatible change it also includes the fork as a parameter for the precompile's execution.

Since this [pr](https://github.com/lambdaclass/ethrex/pull/4163) has been merged I added the eip-7823 in the fusaka_eip_for_testing vector for testing, and you can check the current pr passes the MODEXP state tests both in the ci or locally.

To test it locally you can run:

``` bash
cd tooling/ef_tests/state
cargo test --package ef_tests-state --test all --release -- --forks Prague,Cancun,Osaka --summary --paths  --tests state_tests/osaka/eip7823_modexp_upper_bounds/modexp_upper_bounds/modexp_upper_bounds.json
```

``` bash
cd tooling/ef_tests/blockchain
make test-levm
```

Closes #4153

### DIFF
--- a/tooling/ef_tests/blockchain/test_runner.rs
+++ b/tooling/ef_tests/blockchain/test_runner.rs
@@ -36,7 +36,6 @@ pub fn parse_and_execute(
     //EIPs should be added as strings in the format 'eip-XXXX'
     let fusaka_eips_to_test: Vec<&str> = vec!["eip-7823", "eip-7939"];
 
-
     //Hashes of any other tests to run, that don't correspond to an especific EIP (for examples, some integration tests)
     //We should really remove this once we're finished with implementing Fusaka, but it's a good-enough workaround to run specific tests for now
     let hashes_of_fusaka_tests_to_run: Vec<&str> = vec![


edg-l:
Before merging this i would like to review it when it's in a ready state, since I recently optimized the modexp and there is another pr like https://github.com/lambdaclass/ethrex/pull/4168/

crates/vm/levm/src/precompiles.rs
```diff
@@ -388,18 +393,31 @@ pub fn modexp(calldata: &Bytes, gas_remaining: &mut u64) -> Result<Bytes, VMErro
...
-        return Ok(Bytes::new());
-    }
-
     // The try_into are infallible and the compiler optimizes them out, even without unsafe.
     // https://godbolt.org/z/h8rW8M3c4
     let base_size = u256_from_big_endian_const::<32>(calldata[0..32].try_into()?);
     let modulus_size = u256_from_big_endian_const::<32>(calldata[64..96].try_into()?);
     let exponent_size = u256_from_big_endian_const::<32>(calldata[32..64].try_into()?);
 
+    if fork == Fork::Osaka
```
fmoletta:
```suggestion
    if fork >= Fork::Osaka
```
This will also apply for post-osaka forks

cdiielsi: You are right! Thanks! 🙏

crates/vm/levm/src/precompiles.rs
```diff
@@ -388,18 +393,31 @@ pub fn modexp(calldata: &Bytes, gas_remaining: &mut u64) -> Result<Bytes, VMErro
     let modulus_size_bytes: [u8; 32] = calldata[64..96].try_into()?;
     const ZERO_BYTES: [u8; 32] = [0u8; 32];
 
-    if base_size_bytes == ZERO_BYTES && modulus_size_bytes == ZERO_BYTES {
```
edg-l, line 391: Why did you move this check after the u256 parsing?

It's cheaper to do so before doing the u256 conversion, so if possible it should stay here. As far as I know, it shouldn't be incompatible with the new fusaka check

cdiielsi: I moved it because the tests favor the upper bound check before this one. For example, there's fork_Osaka-state_test-exp_1025_base_0_mod_0 which expects an error because of the upper bound for the exponent. But I guess I can move all the checks prior to the parsing. Is there something similar to 1024u128.to_be_bytes() I could use? Or should I hardcode an array with 1024 value?

edg-l (approved):
lgtm, a comment on the zero check order would be nice

cdiielsi:
This EIP has already been implemented in this [pr](https://github.com/lambdaclass/ethrex/pull/4206).

PR #4192, opened 2025-08-28
build(l1, l2): tag `ethrex:main` instead of `ethrex:unstable`.

mpaulucci:
**Motivation**
the term `unstable` is confusing. `main` is much clearer imo and it's the default other clients use

**Description**
- Changed the docker tag on merge from `unstable` to `main`
- Updated references to `unstable` across the codebase.

### DIFF
--- a/.github/workflows/common_hive_reports.yaml
+++ b/.github/workflows/common_hive_reports.yaml
@@ -64,8 +64,8 @@ jobs:
         if: ${{ inputs.job_type == 'main' }}
         run: |
           docker pull ghcr.io/lambdaclass/ethrex:main
-          # Tests use ethrex:latest so we retag it
-          docker tag ghcr.io/lambdaclass/ ethrex:ci
+          # Tests use ethrex:ci so we retag it
+          docker tag ghcr.io/lambdaclass/ethrex:main ethrex:ci
 
       - name: Setup Go
         uses: actions/setup-go@v5
--- a/.github/workflows/common_hive_reports.yaml
+++ b/.github/workflows/common_hive_reports.yaml
@@ -63,9 +63,9 @@ jobs:
       - name: Pull image
         if: ${{ inputs.job_type == 'main' }}
         run: |
-          docker pull ghcr.io/lambdaclass/ethrex:unstable
+          docker pull ghcr.io/lambdaclass/ethrex:main
           # Tests use ethrex:latest so we retag it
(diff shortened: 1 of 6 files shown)


.github/workflows/common_hive_reports.yaml
```diff
@@ -63,9 +63,9 @@ jobs:
       - name: Pull image
         if: ${{ inputs.job_type == 'main' }}
         run: |
-          docker pull ghcr.io/lambdaclass/ethrex:unstable
+          docker pull ghcr.io/lambdaclass/ethrex:main
           # Tests use ethrex:latest so we retag it
-          docker tag ghcr.io/lambdaclass/ethrex:unstable ethrex:ci
+          docker tag ghcr.io/lambdaclass/ ethrex:ci
```
damiramirez: Is `ethrex:main` missing after `docker tag ghcr.io/lambdaclass/` or is it not needed?

mpaulucci: yes :/. Thanks

.github/workflows/common_hive_reports.yaml
```diff
@@ -63,9 +63,9 @@ jobs:
       - name: Pull image
         if: ${{ inputs.job_type == 'main' }}
         run: |
-          docker pull ghcr.io/lambdaclass/ethrex:unstable
-          # Tests use ethrex:latest so we retag it
-          docker tag ghcr.io/lambdaclass/ethrex:unstable ethrex:ci
+          docker pull ghcr.io/lambdaclass/ethrex:main
+          # Tests use ethrex:ci so we retag it
+          docker tag ghcr.io/lambdaclass/ethrex:main ethrex:ci
```
ManuelBilbao, line 68: Do tests use `ethrex:ci`? I think hive tests are using `unstable` (now `main`) tag

mpaulucci: So, Hive uses what the client file specifies https://github.com/lambdaclass/ethrex/blob/main/.github/config/hive/clients.yaml#L4 (assuming you pass it as arg, we do). If no client file is specified, it uses the default: https://github.com/ethereum/hive/pull/1325/files#diff-88221494e3fbc0839756b9c5b02ad47942b9bbcf3970e00c335e8306c0c5009cR2

ManuelBilbao: Ah, I forgot that change

PR #4193, opened 2025-08-28
fix(l1): update references to moved folder (tooling/ef_tests)

mpaulucci:
**Motivation**
Fixing the leftovers from https://github.com/lambdaclass/ethrex/pull/4143

**Description**
- Fixes missing reference updates after moving.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -19,5 +19,5 @@ docs/
 # Local development
 hive/
 ethereum-package/
-cmd/ef_tests/blockchain/vectors
-cmd/ef_tests/state/vectors
+tooling/ef_tests/blockchain/vectors
+tooling/ef_tests/state/vectors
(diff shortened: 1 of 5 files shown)


.dockerignore
```diff
@@ -19,5 +19,5 @@ docs/
 # Local development
 hive/
 ethereum-package/
-cmd/ef_tests/blockchain/vectors
-cmd/ef_tests/state/vectors
+tooling/ef_tests/blockchain/vectors
+tooling/ef_tests/state/vectors
```
JereSalo, line 23: For now do we have to add to this state_v2 to this too?

JereSalo (approved):

cdiielsi (approved):

.dockerignore
mpaulucci: yes, but the idea is to remove state and rename state_v2 to state so it's low priority

PR #4203, opened 2025-08-29
perf(l1): experimental blob database for tries

Oppen:
We're working on writing a new database due to our current one not living up
to the required performance.
This is an **experiment** following the initial design for the MVP to see what
the integration would need and to have something to measure quickly so we have
some kind of baseline.

Currently imports 80k blocks from Hoodi in under 1h on an M1. It's still
running towards the 100k import so I can't tell the exact figures just yet.


Oppen:
This experiment is abandoned and wildly out of date.

PR #4211, opened 2025-08-29
feat(l2): configure block max gas limit

### LINKED ISSUE #2526: Support changing the gas_limit on the L2 after genesis
During the life of a blockchain, the gas_limit of the L2 could be changed. 

For example, Ethereum had 30M, and increased it to 36M recently. Other l2s may decide to increase it if hardware improves, or have some dynamic block limit if they see spare capacity.

It would be useful to have the ability to alter the gas_limit of a running l2, maybe through a config and a fast restart of the proposer

### LINKED ISSUE #4207: L2: make `gas_limit` per block configurable

avilagaston9:
> [!CAUTION]
> Merge after #4224

**Motivation**

We need a way to limit the gas used per block to avoid overloading our prover.

**Description**

- Adds a new CLI parameter: `max_gas_limit`.
- Updates the `payload_builder` to check this parameter as well.

**How to test**

Set a low `max_gas_limit`, start the L2, and watch in the monitor how the initial deposits are distributed across different blocks, ensuring that the gas used per block is lower than the configured amount.

Closes #4207 
Closes #2526

### DIFF
--- a/cmd/ethrex/l2/options.rs
+++ b/cmd/ethrex/l2/options.rs
@@ -389,7 +389,7 @@ pub struct BlockProducerOptions {
     pub elasticity_multiplier: u64,
     #[arg(
         long = "block-producer.block-gas-limit",
-        default_value = "30_000_000",
+        default_value = "30000000",
         value_name = "UINT64",
         env = "ETHREX_BLOCK_PRODUCER_BLOCK_GAS_LIMIT",
         help = "Maximum gas limit for the L2 blocks.",


tomip01 (approved):

crates/l2/sequencer/block_producer/payload_builder.rs
```diff
@@ -202,7 +239,7 @@ pub async fn fill_transactions(
         if head_tx.tx_type() == TxType::Privileged {
             let id = head_tx.nonce();
             if let Some(range) = privileged_range.as_mut() {
-                if range.end - range.start > PRIVILEGED_TX_BUDGET {
+                if range.clone().count() > PRIVILEGED_TX_BUDGET {
```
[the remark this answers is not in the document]
avilagaston9: The change was due to a bad merge. It's fixed now.

cmd/ethrex/l2/options.rs
```diff
@@ -386,6 +387,14 @@ pub struct BlockProducerOptions {
...
     )]
     pub elasticity_multiplier: u64,
+    #[arg(
+        long = "block-producer.max-gas-limit",
+        value_name = "UINT64",
+        env = "ETHREX_BLOCK_PRODUCER_MAX_GAS_LIMIT",
+        help = "Maximum gas limit for the L2 blocks.",
+        help_heading = "Block producer options"
+    )]
+    pub max_gas_limit: Option<u64>,
```
ilitteri: Why not `block_gas_limit`?

ilitteri (approved):
Left a minor comment

crates/l2/sequencer/block_producer/payload_builder.rs
```diff
@@ -115,6 +125,16 @@ pub async fn fill_transactions(
             break;
         };
 
+        // Check if we have enough gas to run more transactions within the configured max_gas_limit
+        if let Some(max_gas_limit) = max_gas_limit {
+            if context.payload.header.gas_limit - context.remaining_gas + TX_GAS_COST
+                >= max_gas_limit
+            {
```
MegaRedHand: Let's extract this part into a variable to make it easier to read:

```suggestion
        let gas_used = context.payload.header.gas_limit - context.remaining_gas;
        // Check if we have enough gas to run more transactions within the configured max_gas_limit
        if let Some(max_gas_limit) = max_gas_limit {
            if gas_used + TX_GAS_COST >= max_gas_limit {
```

Also, later we do the same computation again, so we can use the variable there too.

crates/l2/sequencer/block_producer/payload_builder.rs
```diff
@@ -136,6 +156,18 @@ pub async fn fill_transactions(
             continue;
         }
 
+        // Check if we have enough gas to run the transaction within the configured max_gas_limit
+        if let Some(max_gas_limit) = max_gas_limit {
+            if context.payload.header.gas_limit - context.remaining_gas + head_tx.tx.gas_limit()
```
MegaRedHand: Here. We could also add a method to context `context.gas_used()` that computes this.

MegaRedHand (commented):
Changes LGTM. I think we should also signal the gas limit in the header of the block. I'm not sure if we have special-cased it in the L2, but on the L1 the client sets a "desired gas limit", to which it will converge with enough time, moving the gas limit (and target) a bit on each block.

cmd/ethrex/l2/options.rs
```diff
@@ -386,6 +387,14 @@ pub struct BlockProducerOptions {
         help_heading = "Proposer options"
     )]
     pub elasticity_multiplier: u64,
+    #[arg(
+        long = "block-producer.block-gas-ceil",
+        value_name = "UINT64",
+        env = "ETHREX_BLOCK_PRODUCER_BLOCK_GAS_CEIL",
+        help = "Maximum gas ceil for the L2 blocks.",
```
MegaRedHand: We can drop the abbreviation here

```suggestion
        help = "Maximum gas ceiling for the L2 blocks.",
```

MegaRedHand (approved):
LGTM

avilagaston9:
The PR was changed to not only use a `configured_gas_limit` in `fill_transactions()`, but also to use it as the node’s `target_gas_limit`.

MegaRedHand (approved):

crates/l2/sequencer/block_producer.rs
```diff
@@ -79,6 +81,7 @@ impl BlockProducer {
             rollup_store,
             // FIXME: Initialize properly to the last privileged nonce in the chain
             last_privileged_nonce: None,
+            block_gas_limit: block_gas_limit.unwrap_or(DEFAULT_BUILDER_GAS_CEIL),
```
tomip01: Maybe I am wrong but we are always defaulting it to `DEFAULT_BUILDER_GAS_CEIL` if `None`. 

We could change from `Option<u64>` to `u64` in the `BlockProducerOptions`. And add `default_value_t = DEFAULT_BUILDER_GAS_CEIL,` in the clap options

tomip01 (commented):
One comment I just realized now

crates/l2/sequencer/block_producer.rs
avilagaston9: Done [be164a7](https://github.com/lambdaclass/ethrex/pull/4211/commits/be164a760385285e6105b06fd767532c1d8f8a0c)!

PR #4213, opened 2025-08-29
feat(l1): improve p2p peer discovery and rework snap sync

jrchatruc:
**Motivation**

This PR does two main things:

- Improves the process of finding peers by being way more aggressive in our lookup. The previous version was too passive and as a result did not find many peers in a reasonable amount of time.
- Reworks and fixes a few things in snap sync to make it work on the main networks: this was tested and ran succesfully on hoodi, sepolia and mainnet. It currently syncs hoodi in around 20 minutes in a good scenario, sepolia in around 3 and a half hours, and mainnet in ~24 hours. These times are very far from ideal and there is a ton of room for improvement here. This is a first iteration that will be improved in subsequent PRs, along with improvements to our p2p code in general (lots of bugs and performance issues were found once we started succesfully syncing big testnets). The code is also far from ideal from a readability standpoint, something we will also tackle on a future PR by rewriting this code using `spawned`.

Some of the things left to tackle on subsequent PRs are:

- Fix memory usage, currently the snap sync code does not flush data to disk frequently enough and as a result uses massive amounts of memory on big networks when syncing.
- Use a global peer scoring mechanism instead of tracking scores separately on each snap sync stage.
- Improve download task distribution to peers; the algorithm is currently too simple and becomes very slow if run on a node with high latency to most of its peers.
- Various p2p fixes: improve tx broadcast logic to decrease cpu usage when we have lots of peers, limit mempool size and other changes needed to make the node more stable.
- Optionally download blocks and receipts as well once the sync process is finished. These are not necessary for a node to be synced but they are common data that most clients usually serve by default.
- Various performance improvements to reduce time spent writing to the database, make db writes concurrent to downloads, and more.


docker-compose.yml
```diff
@@ -0,0 +1,55 @@
+services:
```
mpaulucci: what's the difference between this and `make hoodi`? See: https://github.com/lambdaclass/ethrex/blob/main/fixtures/network/hoodi.yaml

It already runs a docker instance of ethrex + a docker instance of lighthouse to be able to sync hoodi.

crates/networking/p2p/discv4/side_car.rs
```diff
@@ -0,0 +1,325 @@
+use std::{
```
mpaulucci, line 1: super nit: sidecar.rs

docker-compose.yml
jrchatruc: I'd say the difference is it does not set up kurtosis when running it, which is a pretty big requirement if people just want to run ethrex+lighthouse

mpaulucci: I would definitely not have this in the root of the repo, maybe move to tooling/sync?

crates/common/trie/verify_range.rs
```diff
@@ -132,7 +128,39 @@ pub fn verify_range(
...
+struct RangeProof<'a> {
+    node_refs: BTreeMap<H256, &'a [u8]>,
+}
+
+impl<'a> From<&'a [Vec<u8>]> for RangeProof<'a> {
+    fn from(proof: &'a [Vec<u8>]) -> Self {
+        let node_refs = proof
+            .iter()
+            .map(|node| {
+                let hash = H256::from_slice(&Keccak256::new_with_prefix(node).finalize());
```
Oppen, line 145:
```suggestion
                let hash = H256::from_slice(&Keccak256::digest(node));
```

docker-compose.yml
jrchatruc: Makes sense, moved it here [8273c3b](https://github.com/lambdaclass/ethrex/pull/4213/commits/8273c3b09372ef88090b286993bee82b0a1cca75)

crates/networking/p2p/discv4/side_car.rs
jrchatruc: The sidecar module will very soon disappear (@ElFantasma is working on integrating it to the discovery genserver instead of being a separate one), so it's probably better to just wait for his changes

tooling/sync/Makefile
```diff
@@ -186,3 +186,39 @@ start-ethrex: ## Start ethrex for the network given by NETWORK.
...
+# Use make server-sync SERVER_SYNC_BRANCH=branch_name SERVER_SYNC_NETWORK=network_name LOGS_FILE=logs_file_name  HEALING=1 MEMORY=1 SNAP=1
+# SERVER_SYNC_BRANCH is the branch to checkout before syncing, SERVER_SYNC_NETWORK is the network to sync, LOGS_FILE is the file to output logs to, HEALING enables healing mode, MEMORY uses memory datadir, SNAP enables snap sync mode.
+server-sync:
+	git fetch --all
+
+	git checkout $(SERVER_SYNC_BRANCH)
+
+	git pull
+
+	tmux kill-server || true
```
Oppen, line 214: This is a bit aggressive and breaks the workflow for tmux users. Just kill the session that ethrex controls.

crates/storage/trie_db/libmdbx_locked.rs
```diff
@@ -0,0 +1,58 @@
...
+where
+    T: Table,
+{
+    fn drop(&mut self) {
+        // The struct needs a Transaction referencing a Database object
+        unsafe {
+            drop(Box::from_raw(
+                self.db as *const Arc<Database> as *mut Arc<Database>,
+            ));
+        }
```
Oppen, line 41: I believe this still leaks the transaction object.

Oppen: Ah, my bad, that was not leaked to begin with. That said, this might use-after-free. You should manually drop the transaction first in case this was the last living reference to the database. Otherwise, the transaction might refer to the dead database on drop.

crates/storage/trie_db/libmdbx_dupsort_locked.rs
```diff
@@ -0,0 +1,71 @@
...
+where
+    T: DupSort<Key = (SK, [u8; 33]), SeekKey = SK, Value = Vec<u8>>,
+    SK: Clone + Encodable,
+{
+    fn drop(&mut self) {
+        // The struct needs a Transaction referencing a Database object
+        unsafe {
+            drop(Box::from_raw(
+                self.db as *const Arc<Database> as *mut Arc<Database>,
+            ));
```
Oppen, line 50: Same issue.

crates/storage/store_db/libmdbx.rs
```diff
@@ -1084,6 +1008,54 @@ impl StoreEngine for Store {
...
+    }
+
+    async fn write_account_code_batch(
+        &self,
+        account_codes: Vec<(H256, Bytes)>,
+    ) -> Result<(), StoreError> {
+        let account_codes = account_codes
+            .into_iter()
+            .map(|(account_hash, account_code)| (account_hash.into(), account_code.into()))
+            .collect();
```
Oppen, line 1055: I believe it's better to change the table to use `H256 => Bytes` instead of making this conversion.

crates/storage/store_db/in_memory.rs
```diff
@@ -43,8 +43,6 @@ struct StoreInner {
     invalid_ancestors: HashMap<BlockHash, BlockHash>,
     // Stores current Snap Sate
     snap_state: SnapState,
-    // Stores State trie leafs from the last downloaded tries
-    state_snapshot: BTreeMap<H256, AccountState>,
     // Stores Storage trie leafs from the last downloaded tries
     storage_snapshot: HashMap<H256, BTreeMap<H256, U256>>,
```
Oppen, line 47: Shouldn't this be gone as well?

crates/networking/rpc/admin/peers.rs
```diff
@@ -78,42 +78,42 @@ impl From<PeerData> for RpcPeer {
...
         .collect::<Vec<_>>();
     Ok(serde_json::to_value(peers)?)
 }
 
-#[cfg(test)]
-mod tests {
-    use ethrex_p2p::types::{Node, NodeRecord};
-    use rand::rngs::OsRng;
-    use secp256k1::SecretKey;
+// TODO: Adapt the test to the new P2P architecture.
```
Oppen, line 92: I think this should be fixed before merging to main.

crates/networking/p2p/utils.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::{
+    kademlia::PeerChannels,
+    rlpx::{Message, connection::server::CastMessage, snap::TrieNodes},
+};
+
+use crate::peer_handler::DumpError;
+
+/// Computes the node_id from a public key (aka computes the Keccak256 hash of the given public key)
+pub fn node_id(public_key: &H512) -> H256 {
+    keccak(public_key)
```
Oppen, line 22: I'd rather avoid depending more on the `keccak_hash` crate. It's an extra crate for essentially this:
```rust
fn keccak(data: impl AsRef<[u8]>) -> H256 {
    Keccak256::digest(data.as_ref()).0.into()
}
```

I'd rather have that in an internal helper.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -915,6 +943,13 @@ async fn handle_broadcast(
     if id != tokio::task::id() {
         match broadcasted_msg.as_ref() {
             Message::Transactions(txs) => {
+                let peer_sqrt = (state.connection_broadcast_send.receiver_count() as f64).sqrt();
+                // we want to send to sqrt(peer_count) on average
+                // sqrt(peer_count)/peer_count == 1/sqrt(peer_count)
+                let accept_prob = 1.0 / peer_sqrt;
+                if random::<f64>() < accept_prob {
+                    return Ok(());
```
Oppen, line 951: This should trigger tx hashes to be sent. The idea is that all peers receive the transactions xor their hashes.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -560,19 +589,7 @@ where
...
-    let peer_count = {
-        let table_lock = table.lock().await;
-        table_lock.count_connected_peers()
-    };
-
-    if peer_count >= MAX_PEERS_TCP_CONNECTIONS {
-        return Err(RLPxError::DisconnectSent(DisconnectReason::TooManyPeers));
-    }
-
+async fn post_handshake_checks(_table: Kademlia) -> Result<(), RLPxError> {
```
Oppen, line 592: Odd. If the method does not do anything, do we really want it?

crates/networking/p2p/network.rs
```diff
@@ -141,21 +178,273 @@ fn listener(tcp_addr: SocketAddr) -> Result<TcpListener, io::Error> {
...
+            remaining_storage_tries_state_roots = remaining_storage_tries_state_roots,
+            storage_tries_state_roots_compute_time =
+                time_taken_to_compute_storage_tries_state_roots
+                    .map(format_duration)
+                    .unwrap_or_else(|| "-".to_owned()),
+            downloaded_storage_slots = *METRICS.downloaded_storage_slots.lock().await,
+            storage_tries_tasks_queued = METRICS.storages_downloads_tasks_queued.lock().await,
+        );
+
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
Oppen, line 437: I believe this sleeps with many locks taken. Why not just use atomics for metrics?

PR #4216, opened 2025-08-29
chore(l2): optimize execution of empty batches

xqft:
**Motivation**

To optimize the execution of L2 empty batches in the prover

**Description**

- initialize block header hashes in `get_first_invalid_block_hash()` to avoid hashing twice

**Flamegraphs**
Running 7500 empty blocks

Before:
image

After:
image


crates/l2/prover/zkvm/interface/src/execution.rs
```diff
@@ -290,46 +293,48 @@ fn execute_stateless(
...
-            .map_err(StatelessExecutionError::ExecutionWitness)?;
-
-        // Update acc_account_updates
-        for account in account_updates {
-            let address = account.address;
-            if let Some(existing) = acc_account_updates.get_mut(&address) {
-                existing.merge(account);
-            } else {
-                acc_account_updates.insert(address, account);
+        if !(is_batch_empty && cfg!(feature = "l2")) {
```
MegaRedHand: Did we check skipping this doesn't also skip any important steps? The L1 EVM has some precompiles that depend on a "system contract call" being executed first thing in a block. Does L2 have anything like that?

xqft: Yes I did, ethrex L2 has no system contracts. Though reviewers should double check it of course

crates/common/types/block_execution_witness.rs
```diff
@@ -248,11 +253,26 @@ impl ExecutionWitnessResult {
...
+                .block_headers
+                .get(&block.header.number)
+                .map(|header| header.hash())
+                .ok_or(ExecutionWitnessError::Unreachable(
+                    "execution witness does not contain the block header of a block to execute"
+                        .to_string(),
+                ))?;
+            // drop Result, don't care if it was already initialized
+            let _ = block.header.hash.set(hash);
+        }
```
ilitteri: We shouldn't initialize the hashes here. We should do it at the beginning of `execute_stateless` to ensure this wanted behavior's visibility (it is kind of obscured/hidden as it is in here) and to also enforce this to be done independently of whether `get_first_invalid_block_hash ` is called or not.

crates/l2/prover/zkvm/interface/src/execution.rs
ilitteri: Try going for this pattern
```rust
if condition {
    continue; or return; or break;
}

// rest of the logic
```
instead of 
```rust
if condition {
    // rest of the logic
}
```
if `rest of the logic` is complex.

crates/l2/prover/zkvm/interface/src/execution.rs
```diff
@@ -234,6 +234,16 @@ fn execute_stateless(
...
 ) -> Result<StatelessResult, StatelessExecutionError> {
+    // if there are no transactions and no withdrawals, the batch is empty
+    let is_batch_empty = !blocks.iter().any(|block| {
+        !block.body.transactions.is_empty()
+            || !block
+                .body
+                .withdrawals
+                .as_ref()
+                .is_none_or(|withdrawals| withdrawals.is_empty())
+    });
```
ilitteri: This optimization works well, but it only applies to completely empty batches. In scenarios with low traffic, such as 1 transaction per minute, we’ll have batches that are 99% empty, as most blocks will be empty except for the block containing the single transaction. In these cases, this optimization wouldn’t take effect.

The intuition is correct, but we need to apply it at the block level to make it effective across all scenarios.

xqft: you are totally right!

crates/common/types/block_execution_witness.rs
xqft: I agree, moved the logic into its own function for `ExecutionWitnessResult`

crates/l2/prover/zkvm/interface/src/execution.rs
xqft: fixed

crates/l2/prover/zkvm/interface/src/execution.rs
xqft: done!

ilitteri (approved):

crates/l2/prover/zkvm/interface/src/execution.rs
```diff
@@ -290,46 +287,57 @@ fn execute_stateless(
...
-            if let Some(existing) = acc_account_updates.get_mut(&address) {
-                existing.merge(account);
-            } else {
-                acc_account_updates.insert(address, account);
+        let is_block_empty = !block.body.transactions.is_empty()
+            || !block
+                .body
+                .withdrawals
+                .as_ref()
+                .is_none_or(|withdrawals| withdrawals.is_empty());
```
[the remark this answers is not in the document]
xqft: correct!

crates/l2/prover/src/guest_program/src/execution.rs
```diff
@@ -243,14 +243,20 @@ fn execute_stateless(
...
     })?;
 
-    // Validate block hashes, except parent block hash (latest block hash)
+    // Hashing is an expensive operation in zkVMs, this way we avoid hashing twice
+    // (once in get_first_invalid_block_hash(), later in validate_block()).
+    wrapped_db.initialize_block_header_hashes(blocks)?;
+
+    // Validate execution witness' block hashes, except parent block hash (latest block hash).
+    // this will also initialize hashes in the `blocks` slice so that we don't hash twice
+    // (which is expensive in zkVMs).
```
avilagaston9: This is incorrect, right? We initialized them in the previous step.

avilagaston9 (approved):

crates/l2/prover/src/guest_program/src/execution.rs
xqft: yes! fixed

PR #4217, opened 2025-08-29
feat(l1): block size limit

### LINKED ISSUE #4158: [EIP-7934] RLP Execution Block Size Limit
https://forkcast.org/upgrade/fusaka#eip-7934

Acceptance criteria:
- All tests related to 7934

SDartayet:
**Description**

This PR implements [EIP-7934](https://eips.ethereum.org/EIPS/eip-7934), which will be incorporated in Fusaka. Said EIP introduces a limit on the size of the RLP encode of blocks. This PR adds the required check in block validation.

Closes #4158


crates/storage/store_db/libmdbx.rs
```diff
@@ -169,13 +169,23 @@ impl StoreEngine for Store {
 
                 tx.upsert::<Bodies>(
                     hash.into(),
-                    BlockBodyRLP::from_bytes(block.body.encode_to_vec()),
+                    BlockBodyRLP::from_bytes(
```
mpaulucci: here you are leaking an abstraction. Why can't you have `encode_to_vec` use the cached version?

crates/common/types/block.rs
```diff
@@ -36,16 +36,58 @@ use once_cell::sync::OnceCell;
...
 
     pub fn hash(&self) -> BlockHash {
         self.header.hash()
     }
+
+    // Calculate the size of the RLP encode of the block
+    // We need the RLP encode in two places: in the block validation to check its size doesn't exceed the maximum, and in the store to save it
+    // However, the store uses the encode of the body and the header, whereas for the max size we need the size of the encode of the block as a whole
+    // So we cache the body and header, and calculate the size of the block encode based off of them
+    pub fn get_rlp_encode_size(&self) -> usize {
```
mpaulucci: I would expect the block to call something like block.encode().size(), not have a special function for the size

SDartayet: The idea is to avoid reencoding the block many times, since RLP encoding is one of the more expensive parts of adding new blocks. The store saves to memory the RLP encode of the body and the header. which means caching the RLP encode of the entire block wouldn't work. So the idea was to calculate the RLP encode of the body and header (which will be stored later) and use their size to calculate the cost of the RLP encode of the entire block. It's admittedly a bit messier, but given the performance cost of RLP encoding it seemed worthwhile to avoid encoding the entire block twice.

crates/common/constants.rs
```diff
@@ -57,3 +57,9 @@ pub const GAS_PER_BLOB: u32 = 1 << 17;
 
 // Minimum base fee per blob
 pub const MIN_BASE_FEE_PER_BLOB_GAS: u64 = 1;
+
+// === EIP-7934 constants ===
+
+pub const MAX_BLOCK_SIZE: u64 = 10_485_760;
+pub const SAFETY_MARGIN: u64 = 2_097_152;
```
mpaulucci: RLP_BLOCK_SIZE_SAFETY_MARGIN maybe? SAFETY_MARGIN is too generic

crates/blockchain/error.rs
```diff
@@ -67,6 +67,8 @@ pub enum InvalidBlockError {
     BlobGasUsedMismatch,
     #[error("Invalid transaction: {0}")]
     InvalidTransaction(String),
+    #[error("Maximum block size exceeded: Maximum is {0} MiB, but block was {1} MiB")]
+    MaximumSizeExceeded(u64, u64),
```
mpaulucci: MaximumRlpSizeExceeded

mpaulucci (approved):
please create a ticket to make it more efficient

crates/common/constants.rs
SDartayet: I just followed the names they used in the EIP; but I agree RLP_BLOCK_SIZE_SAFETY_MARGIN would be more descriptive

crates/blockchain/blockchain.rs
```diff
@@ -987,7 +988,17 @@ pub fn validate_block(
...
+        let block_rlp_size = block.encode_to_vec().len();
+        if block_rlp_size > MAX_RLP_BLOCK_SIZE as usize {
+            return Err(error::ChainError::InvalidBlock(
+                InvalidBlockError::MaximumRlpSizeExceeded(
+                    MAX_RLP_BLOCK_SIZE,
+                    block_rlp_size as u64,
+                ),
+            ));
+        }
+    } else if chain_config.is_prague_activated(block.header.timestamp) {
```
fmoletta: Shouldn't this be a separate `if` block instead of an `else if`? If a block is post-osaka then it will also be post-prague and we would want to perform both Osaka and Prague specific checks

fmoletta: Aka validate the maximum rlp size & also validate the header fields

SDartayet: You're right; idk how I missed that (or the tests didn't catch it). Fixed [here](https://github.com/lambdaclass/ethrex/pull/4217/commits/7b9043a59474c8217eb1ede1c07dcd41f9790139)

PR #3763, opened 2025-07-22
perf(levm): skip `JUMPDEST` execution on `JUMP` instructions (with efficient nop-slide implementation)

azteca1998:
**Motivation**

When taking `JUMP` instructions, the target is always a `JUMPDEST`, which is a NOP.

**Description**

Optimize taken `JUMP`s by skipping all `JUMPDEST` until a non-NOP instruction is found. Since it's easy enough, also do the same when executing a `JUMPDEST` directly.

image

PR #4220, opened 2025-09-01
perf(levm): process JUMPDEST gas and pc together within the given JUMP/JUMPI opcode

edg-l:
**Motivation**

It's basically #3763 without the nop slide, see the discussion there. 
It only skips 1 jumpdest if coming from a jump, but doesn't slide if there are a lot of sequential jumpdests afterwards. 

Still has improvements overall to a lot of benches

image


crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -357,8 +360,10 @@ impl<'a> VM<'a> {
 
     // JUMPDEST operation
     pub fn op_jumpdest(&mut self) -> Result<OpcodeResult, VMError> {
-        let current_call_frame = &mut self.current_call_frame;
-        current_call_frame.increase_consumed_gas(gas_cost::JUMPDEST)?;
+        let call_frame = &mut self.current_call_frame;
+
+        call_frame.increase_consumed_gas(gas_cost::JUMPDEST)?;
```
Arkenan: While we're at it, can't we just do all in one line?

```rust
self.current_call_frame.increase_consumed_gas(gas_cost::JUMPDEST)
```

Arkenan (approved):
LGTM

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
edg-l: fixed

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -330,8 +330,11 @@ impl<'a> VM<'a> {
             .try_into()
             .map_err(|_err| ExceptionalHalt::VeryLargeNumber)?;
 
+        #[expect(clippy::arithmetic_side_effects)]
         if Self::target_address_is_valid(call_frame, jump_address_usize) {
-            call_frame.pc = jump_address_usize;
+            call_frame.increase_consumed_gas(gas_cost::JUMPDEST)?;
+
+            call_frame.pc = jump_address_usize + 1;
```
MegaRedHand, line 337: Do we handle the case where the +1 takes it out of bounds?

MegaRedHand (approved):

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
edg-l: yes `  self.bytecode.get(self.pc).copied().unwrap_or(0)`

PR #4221, opened 2025-09-01
perf(l1): refactor trie to support (db) transactions

azteca1998:
**Motivation**

A different team found out that we were not reusing database transactions between consecutive reads, and it was severely impacting performance in their use case.

**Description**

Reuse database transaction for database read operations in obvious places.

Closes #issue_number


Arkenan:
Similar features are already in main through libmdbx_locked modules.

PR #4222, opened 2025-09-01
ci(l1): fix daily hive job

mpaulucci:
**Motivation**
Hive job was not working properly since the local docker image was not loaded.

### DIFF
--- a/.github/workflows/common_hive_reports.yaml
+++ b/.github/workflows/common_hive_reports.yaml
@@ -3,10 +3,10 @@ name: Run Hive Tests
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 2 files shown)


.github/actions/build-docker/action.yml
```diff
@@ -49,9 +49,15 @@ runs:
       with:
         context: .
         file: ./Dockerfile
-        load: ${{ inputs.push == 'false' }}
```
mpaulucci, line 52: it seems this option was being ignored because we're exporting to a tar file

MegaRedHand:
Code LGTM. I started a new run to see if this fixes it: https://github.com/lambdaclass/ethrex/actions/runs/17379274505

PR #4227, opened 2025-09-01
feat(l1): add a way to resume `archive_sync` after a crash

fmoletta:
**Motivation**
Currently, archive sync always begins downloading state from the very start of the state trie (hash 0x00..). This means that in case of an unexpected crash we have no way of resuming the previous state sync, which can mean losing a lot of precious time. This PR proposes a small solution to this problem, by adding a way to resume the previously aborted archive sync via a `--continued` flag.
However, this solution is very limited, as we are not able to resume the state sync portion of archive sync (aka rebuilding the state locally) as we did not keep track of intermediate roots during the archive sync. It is only able to resume the writing of the archive node's state to files. Hence why `--continued` can only be used in combination with `--ipc_path`, `--output_dir` & `--no_sync` flags

**Description**
* Add `--continued` flag to be able to continue a previously aborted archive sync process. (Note: only usable with `--no_sync` as we are not able to resume local state rebuilding)

* (Misc) Log account processing even with `--no_sync` to avoid silent node

Closes None

### DIFF
--- a/tooling/archive_sync/README.md
+++ b/tooling/archive_sync/README.md
@@ -78,3 +78,13 @@ If we don't need the node to be synced (for example if we plan to move the state
 ```bash
  cargo run --release BLOCK_NUMBER --input_dir STATE_DUMP_DIR
 ```
+
+## Resuming archive sync after crashes
+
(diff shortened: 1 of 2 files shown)


fmoletta:
Closed in favour of #4231

PR #4229, opened 2025-09-01
fix(l1,l2): eth client support 4 byte signatures, return decoded error in `RpcErr::Revert`

### LINKED ISSUE #4021: L2: estimate_gas in EthClient can't handle 4 byte function signatures
If an eth_estimateGas call returns a 4 byte function signature as data it is currently interpreted as a string but if fails because of `Failed to slice index abi_decoded_error_data in estimate_gas`. In case the returned data can't be decoded as a string it should at least log the returned error data so if it is a 4 byte function signature the user can look it up in existing databases.

Here is a small test case of the failing code on `crates/networking/rpc/clients/eth/mod.rs`

```rust
#[cfg(test)]
mod test {
    use core::panic;

    use ethrex_common::U256;
    use serde_json::json;

    use crate::clients::{EthClientError, eth::RpcResponse};

    #[test]

    fn test_estimate_gas_revert_with_() {
        let json = json!({
            "jsonrpc": "2.0",
            "id": 1,
            "error":  {
                "code": 3,
                "message": "execution reverted",
                "data": "0xd93c0665"
            }
        });

        let as_str = json.to_string();

        let response = serde_json::from_str::<RpcResponse>(&as_str).unwrap();

        if let RpcResponse::Error(error_response) = response {
            let error_data = if let Some(error_data) = error_response.error.data {
                if &error_data == "0x" {
                    "unknown error".to_owned()
                } else {
                    let abi_decoded_error_data = hex::decode(
                        error_data
                            .strip_prefix("0x")
                            .ok_or(EthClientError::Custom(
                                "Failed to strip_prefix in estimate_gas".to_owned(),
                            ))
                            .unwrap(),
                    )
                    .map_err(|_| {
                        EthClientError::Custom("Failed to hex::decode in estimate_gas".to_owned())
                    })
                    .unwrap();
                    let string_length = U256::from_big_endian(
                        abi_decoded_error_data
                            .get(36..68)
                            .ok_or(EthClientError::Custom(
                                "Failed to slice index abi_decoded_error_data in estimate_gas"
                                    .to_owned(),
                            ))
                            .unwrap(),
                    );

                    let string_len = if string_length > usize::MAX.into() {
                        panic!("Failed to convert string_length to usize in estimate_gas");
                    } else {
                        string_length.as_usize()
                    };
                    let string_data = abi_decoded_error_data
                        .get(68..68 + string_len)
                        .ok_or(EthClientError::Custom(
                            "Failed to slice index abi_decoded_error_data in estimate_gas"
                                .to_owned(),
                        ))
                        .unwrap();
                    String::from_utf8(string_data.to_vec())
                        .map_err(|_| {
                            EthClientError::Custom(
                                "Failed to String::from_utf8 in estimate_gas".to_owned(),
                            )
                        })
                        .unwrap()
                }
            } else {
                "unknown error".to_owned()
            };

            format!("{}: {}", error_response.error.message, error_data)
        } else {
            panic!("the json was interpreted as RpcResponse::Success")
        };
    }
}

```

LeanSerra:
**Motivation**

When calling contracts with `eth_call` or `eth_estimateGas` ethrex rpc was returning an empty message from `get_message_from_revert_data`, we had some code that decoded the error from the revert data but it was only used when calling `eth_estimateGas` from our EthClient.

**Description**

- Move the code used to decode error from `EthClient::estimate_gas` to `get_message_from_revert_data`
  - Add logic to check for 4 byte function signatures, instead of trying to decode this as string, we just return the selector so that the user can look it up in existing databases
- Now that the `RpcErrorMetadata` has a message with content in it use `error_response.error.message.to_string()` in `EthClient::estimate_gas` when an error occurs

Closes #4021

### DIFF
--- a/crates/networking/rpc/clients/eth/mod.rs
+++ b/crates/networking/rpc/clients/eth/mod.rs
@@ -11,7 +11,7 @@ use crate::{
         block_identifier::{BlockIdentifier, BlockTag},
         receipt::{RpcLog, RpcReceipt},
     },
-    utils::{RpcErrorResponse, RpcRequest, RpcSuccessResponse},
+    utils::{RpcErrorResponse, RpcRequest, RpcSuccessResponse, get_message_from_revert_data},
 };
 use bytes::Bytes;
 use errors::{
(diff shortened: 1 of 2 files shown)


crates/networking/rpc/utils.rs
```diff
@@ -95,7 +96,7 @@ impl From<RpcErr> for RpcErrorMetadata {
                 data: Some(data.clone()),
                 message: format!(
                     "execution reverted: {}",
-                    get_message_from_revert_data(&data)
+                    get_message_from_revert_data(&data).unwrap_or("unknown error".to_string())
```
xqft: I would change "unknown error" to something like "tried to decode error from abi but failed: {err}" so we have more info in that case

crates/networking/rpc/clients/eth/mod.rs
```diff
@@ -297,57 +297,7 @@ impl EthClient {
...
-                    }
-                } else {
-                    "unknown error".to_owned()
-                };
-                Err(EstimateGasError::RPCError(format!(
-                    "{}: {}",
-                    error_response.error.message, error_data
-                ))
-                .into())
+                Err(EstimateGasError::RPCError(error_response.error.message.to_string()).into())
```
xqft, line 300: shouldn't we use `get_message_from_revert_data()` here to get info about the revert?

LeanSerra: the message in `error_response.error.message` is the output of calling `get_message_from_revert_data()` so its already decoded

crates/networking/rpc/utils.rs
LeanSerra: Done in [18bdedd](https://github.com/lambdaclass/ethrex/pull/4229/commits/18bdedd3540ce00a623ba143bf4e225b8d36f1d3)

crates/networking/rpc/clients/eth/mod.rs
xqft: right! missed that, sorry

xqft (approved):

avilagaston9 (approved):

PR #4231, opened 2025-09-01
feat(l1): add `--checkpoint` flag to `archive_sync` tool

fmoletta:
**Motivation**
Currenlty, we have no way of resuming an archive sync if it crashes/ we need to stop it. Our only option is to begin from the start, which can make us lose precious time in the case of unexpected crashes during long archive syncs.
This PR proposes adding a `--checkpoint` flag which will periodically write the current archive sync's checkpoint data (current root, last hash, current file, etc) to a file. And which can later be used to resume archive sync from that same checkpoint

**Description**

Closes #issue_number


mpaulucci (approved):

tooling/archive_sync/README.md
```diff
@@ -73,8 +73,16 @@ We can avoid relying on an active archive node connection once we have already p
...
+### Step 2: Run the `archive_sync` executable with `--input_dir` instead of `--ipc_path`
 
 ```bash
  cargo run --release BLOCK_NUMBER --input_dir STATE_DUMP_DIR
 ```
+
+## Resuming archive sync after a crash or manual stop
+
+In order to safely resume an archive sync process the `--checkpoint` flag can be used to provide a checkpoint file which will be periodically updated during the sync. This file can then be passed on to a second run to resume the sybc from the latest checkpoint. It can be used with any supported flag combination. The checkpoint will not store the block number so please make sure you target the same block to avoid state inconsistencies. The tool will fail if the input flags are not compatible with the checkpoint data (ie running with `--ipc_path` and then using the same checkpoint with `--input_dir`). It will also warn and request for user approval if the new run is a downgrade from the previous run which generated the checkpoint (ie, `--no_sync` flag being added or `--output_dir` flag removed) to ensure no checkpoint data is mistakenly lost. For example, you may use this flag like this:
```
cdiielsi:
```suggestion
In order to safely resume an archive sync process the `--checkpoint` flag can be used to provide a checkpoint file which will be periodically updated during the sync. This file can then be passed on to a second run to resume the sync from the latest checkpoint. It can be used with any supported flag combination. The checkpoint will not store the block number so please make sure you target the same block to avoid state inconsistencies. The tool will fail if the input flags are not compatible with the checkpoint data (ie running with `--ipc_path` and then using the same checkpoint with `--input_dir`). It will also warn and request for user approval if the new run is a downgrade from the previous run which generated the checkpoint (ie, `--no_sync` flag being added or `--output_dir` flag removed) to ensure no checkpoint data is mistakenly lost. For example, you may use this flag like this:
```

cdiielsi (approved):

tooling/archive_sync/README.md
fmoletta: Nice catch!

PR #4249, opened 2025-09-01
feat(l1, l2): add `ethrex-replay` command for extending a genesis file with new accounts

xqft:
**Motivation**

This will be useful to test run an L2 or prove a batch in a big state scenario (in which the network has a big world trie and storage tries, which we think can degrade proving performance greatly).

For now it will just add N accounts with some balance, in the future we might add bytecode and random storage.


xqft:
closed in favor of https://github.com/lambdaclass/ethrex/pull/4378

PR #4250, opened 2025-09-01
chore(l1): fix IPv6 listener

### LINKED ISSUE #4232: Add support for Ipv6 P2P connections
During snap sync with hoodi, sending the FindNode messages to our peers fails due to the error `MessageSendFailure(Os { code: 97, kind: Uncategorized, message: "Address family not supported by protocol" })`. This is likely due to IPv6 usage in `ethrex_p2p::discv4::side_car:`

fedacking:
**Motivation**

We had issues when handling discovery with IPv6 connections, because we used a TCPv4 stream. With a TCPv6 stream, it should handle both IPv4 and IPv6 connections

**Description**

- Removed superflous parameter for p2p address
- Fixed the listener to support ipv6 port
- Made the default port ipv6
- Changed documentation and makefile to new format

Closes #4232

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -295,8 +295,8 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> Node {
     let tcp_socket_addr =
         parse_socket_addr("::", &opts.p2p_port).expect("Failed to parse addr and port");
 
-    let p2p_node_ip =
-        local_ip().unwrap_or_else(|_| local_ipv6().expect("Failed to get local ip v6"));
+    let p2p_node_ip = local_ip()
+        .unwrap_or_else(|_| local_ipv6().expect("Neither ipv4 nor ipv6 local address found"));
 
     let local_public_key = public_key_from_signing_key(signer);
 


cmd/ethrex/initializers.rs
```diff
@@ -290,17 +290,14 @@ pub fn get_signer(data_dir: &str) -> SecretKey {
...
 
-    // TODO: If hhtp.addr is 0.0.0.0 we get the local ip as the one of the node, otherwise we use the provided one.
-    // This is fine for now, but we might need to support more options in the future.
-    let p2p_node_ip = if udp_socket_addr.ip() == Ipv4Addr::new(0, 0, 0, 0) {
-        local_ip().expect("Failed to get local ip")
-    } else {
-        udp_socket_addr.ip()
+    let p2p_node_ip = match local_ip() {
+        Ok(ip) => ip,
+        Err(_) => local_ipv6().expect("Failed to get local ip v6"),
```
pablodeymo: it's better to use `unwrap_or_else` instead this match

gianbelinche: Done

Oppen (approved):

cmd/ethrex/initializers.rs
Oppen: At initialization time I'd prefer for the program to crash if it can't go on. If it does that at a higher level then it's OK.

gianbelinche: If neither ipv4 nor ipv6 is available it will crash in that expect.
If we only have ipv6 we still want to use it

cmd/ethrex/initializers.rs
```diff
@@ -290,18 +290,13 @@ pub fn get_signer(data_dir: &str) -> SecretKey {
...
 
-    // TODO: If hhtp.addr is 0.0.0.0 we get the local ip as the one of the node, otherwise we use the provided one.
-    // This is fine for now, but we might need to support more options in the future.
-    let p2p_node_ip = if udp_socket_addr.ip() == Ipv4Addr::new(0, 0, 0, 0) {
-        local_ip().expect("Failed to get local ip")
-    } else {
-        udp_socket_addr.ip()
-    };
+    let p2p_node_ip =
+        local_ip().unwrap_or_else(|_| local_ipv6().expect("Failed to get local ip v6"));
```
iovoid: This error message could be better

gianbelinche: Done

jrchatruc (approved):

PR #4256, opened 2025-09-02
chore(l1): limit amount of ongoing connection attempts

fedacking:
**Motivation**

Currently, our node uses a very high amount of file descriptors for sockets, this forces our uses to change the ulimit. This PR aims to limit the amount of outgoing peer initiations

**Description**

- Changed the limit outgoing connections
- Made it so we initiate 500 peer connections at the start
- When a connection fails, we launch a new task looking for a peer
- If enough time has happened, we clear the already tried connections table


crates/networking/p2p/rlpx/initiator.rs
```diff
@@ -121,19 +143,22 @@ impl GenServer for RLPxInitiator {
...
+                    );
+                };
 
                 CastResponse::NoReply
             }
         }
     }
 }
+
+pub static INITIATOR: LazyLock<OnceCell<GenServerHandle<RLPxInitiator>>> =
```
Oppen, line 163: Isn't `LazyLock` redundant with the `OnceCell`?

PR #4258, opened 2025-09-02
ci(l1,l2): add workflow to notify failed jobs in main.

mpaulucci:
**Motivation**
Get a slack notification if a main job fails


.github/scripts/notify_workflow_failure.sh
```diff
@@ -0,0 +1,68 @@
...
+set -euo pipefail
+
+# Usage: notify_workflow_failure.sh <slack_webhook_url>
+# Expects the following env vars (provided by the caller workflow):
+#   REPO, WORKFLOW_NAME, CONCLUSION, RUN_HTML_URL, RUN_ID, HEAD_SHA, ACTOR
+
+SLACK_WEBHOOK_URL=${1:-}
+if [[ -z "${SLACK_WEBHOOK_URL}" ]]; then
+  echo "Slack webhook URL not provided; skipping notification." >&2
+  exit 0
```
MegaRedHand, line 11: Do we want this to exit with 0 here?

.github/workflows/common_failure_alerts.yaml
```diff
@@ -0,0 +1,45 @@
...
+    workflows:
+      - L1
+      - L2 (SP1 Backend)
+      - L2 (without proving)
+      - L2 Prover
+      - L2 Prover (TDX)
+      - L2 TDX build
+      - LEVM
+      - Publish Docker
+      - Replay proving
```
MegaRedHand, line 14: We should put all of these between quotes. I see the L1 item in a different color than the rest, and knowing YAML is weird, it wouldn't surprise me it interprets it in a weird way.

MegaRedHand: Also, should we include hive reports here?

mpaulucci: I've started with main jobs. We can extend to scheduled but I prefer to iterate.

fmoletta (approved):

Issue #4241, opened 2025-09-01
Improve transaction broadcasting logic


ElFantasma:
Current version just broadcast every transaction we receive to every active peer we have. This presents some issues on performance, memory and even in code responsibilities. 

We need:
- First, we have to centralize broadcast policy in a single `GenServer`. Some work was already done on this sense [here](https://github.com/lambdaclass/ethrex/blob/60161b587c2d90d681c27bf0cee72a6b602302f4/crates/networking/p2p/tx_broadcaster.rs) for transaction hashes but it needs to be improved and handle single transactions as well.
- `RLPxConnection` has a `broadcasted_tx` HashSet with all the transactions it has broadcasted, to avoid re-broadcast them. We have to remove old hashes to prevent this set to grow indefinitely. In any case, this set should be a single list in the broadcaster `GenServer` (instead of a set on every active `RLPxConnection` as it is now).
- Also, the broadcaster process should avoid reaching all the peers for all the messages. It should use sqrt(active_peers) or some other method to prevent spamming the network with broadcasts (check what does the spec says or how is it implemented in Geth)

PR #4259, opened 2025-09-02
feat(l1): improve transaction broadcasting logic

gianbelinche:
**Motivation**

Current version just broadcast every transaction we receive to every active peer we have. This presents some issues on performance, memory and even in code responsibilities.

**Description**
Reorganizes the transaction broadcast logic
When a new peer connection is initialized, we send the hashes of all the txs in the mempool.
Then every time we receive new txs, we add them to the mempool and send them to our peers, we send the txs to a sqrt of our peers, while sending only the hashes to the rest.
The mempool itself is the one responsible for knowing which txs need to be broadcasted. Every time we add a new tx to the mempool we add it to the list of pending broadcast, then once we broadcast them, we erase them. Also if a tx is removed from the mempool before being broadcasted it is also deleted from the broadcast pool.

There is one possible scenario where we could send a tx twice to the same peer, that is if after the tx left the mempool, we receive it again and add it back, we would send it again to the peers.

Closes https://github.com/lambdaclass/ethrex/issues/4241

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -170,6 +170,26 @@ impl Mempool {
         self.filter_transactions_with_filter_fn(&filter_tx)
     }
 
+    /// Gets all the transactions in the mempool
+    pub fn get_all_txs(&self) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
+        let mut txs_by_sender: HashMap<Address, Vec<MempoolTransaction>> =
+            HashMap::with_capacity(128);
+        let tx_pool = self
+            .transaction_pool
+            .read()
+            .map_err(|error| StoreError::MempoolReadLock(error.to_string()))?;
+
+        for (_, tx) in tx_pool.iter() {
+            txs_by_sender
+                .entry(tx.sender())
+                .or_insert_with(|| Vec::with_capacity(128))
+                .push(tx.clone())
+        }
+
+        txs_by_sender.iter_mut().for_each(|(_, txs)| txs.sort());
+        Ok(txs_by_sender)
+    }
+
     /// Applies the filter and returns a set of suitable transactions from the mempool.
     /// These transactions will be grouped by sender and sorted by nonce
     pub fn filter_transactions_with_filter_fn(
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -171,7 +171,7 @@ impl Mempool {
     }
 
     /// Gets all the transactions in the mempool
-    pub fn get_all_txs(&self) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
+    pub fn get_all_txs_by_sender(&self) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
         let mut txs_by_sender: HashMap<Address, Vec<MempoolTransaction>> =
             HashMap::with_capacity(128);
         let tx_pool = self
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -171,7 +171,9 @@ impl Mempool {
     }
 
     /// Gets all the transactions in the mempool
-    pub fn get_all_txs_by_sender(&self) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
+    pub fn get_all_txs_by_sender(
+        &self,
+    ) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
         let mut txs_by_sender: HashMap<Address, Vec<MempoolTransaction>> =
             HashMap::with_capacity(128);
         let tx_pool = self
--- a/crates/networking/p2p/network.rs
+++ b/crates/networking/p2p/network.rs
@@ -136,7 +136,7 @@ pub async fn start_network(context: P2PContext, bootnodes: Vec<Node>) -> Result<
                 error!("Failed to start Tx Broadcaster: {e}");
             })?;
 
-    RLPxInitiator::spawn(context.clone(), tx_broadcaster_handle.clone()) // todo add tx broadcaster handle
+    RLPxInitiator::spawn(context.clone(), tx_broadcaster_handle.clone())
         .await
         .inspect_err(|e| {
             error!("Failed to start RLPx Initiator: {e}");
--- a/crates/networking/p2p/network.rs
+++ b/crates/networking/p2p/network.rs
@@ -12,13 +12,14 @@ use crate::{
         message::Message,
         p2p::SUPPORTED_SNAP_CAPABILITIES,
     },
-    tx_broadcaster::{TxBroadcaster, TxBroadcasterError},
+    tx_broadcaster::{self, TxBroadcaster, TxBroadcasterError},
     types::{Node, NodeRecord},
 };
 use ethrex_blockchain::Blockchain;
 use ethrex_common::H256;
 use ethrex_storage::Store;
 use secp256k1::SecretKey;
+use spawned_concurrency::tasks::GenServerHandle;
 use std::{
     collections::BTreeMap,
     io,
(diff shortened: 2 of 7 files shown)


crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -493,7 +509,7 @@ async fn send_all_pooled_tx_hashes(state: &mut Established) -> Result<(), RLPxEr
         .iter()
         .any(|cap| state.capabilities.contains(cap))
     {
-        let filter = |tx: &Transaction| -> bool { !state.broadcasted_txs.contains(&tx.hash()) };
+        let filter = |_: &Transaction| -> bool { true };
```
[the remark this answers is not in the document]
gianbelinche: There is no method to just get txs from the mempool without the filter

crates/networking/p2p/network.rs
```diff
@@ -128,24 +129,30 @@ pub async fn start_network(context: P2PContext, bootnodes: Vec<Node>) -> Result<
...
+    context
+        .tracker
+        .spawn(serve_p2p_requests(context.clone(), tx_broadcaster_handle));
 
     Ok(())
 }
 
-pub(crate) async fn serve_p2p_requests(context: P2PContext) {
+pub(crate) async fn serve_p2p_requests(
+    context: P2PContext,
```
ElFantasma: Isn't `P2PContext` a good place to add the `tx_broadcaster_handle`?

I mean, that's the idea of `P2PContext`: to simplify all the handles we have to move around.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -157,6 +152,7 @@ pub enum CastMessage {
     BlockRangeUpdate,
     BroadcastMessage(task::Id, Arc<Message>),
     L2(L2Cast),
+    Transactions(Transactions),
```
ElFantasma: I don't think it's worth it to have a new message type. This seems to me as an usual `BackendMessage`.

The idea is to keep the `GenServer` interface as clear as possible: 
- `PeerMessage` when we receive a remote message from peer, 
- `BackendMessage` is when our backend requires the `RLPxConnection` to send a message to the remote peer (as this scenario)
- The rest are some special cases for periodic tasks (and probably can be simplified/removed)

ElFantasma: Also, we should document this, feel free to add a ticket regarding this.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -302,6 +305,20 @@ impl GenServer for RLPxConnection {
                         }
                     }
                 }
+                Self::CastMsg::Transactions(txs) => {
```
ElFantasma: If this was a `BackendMessage` instead, there is no need for this special case

crates/networking/p2p/rlpx/connection/server.rs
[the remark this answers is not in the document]
ElFantasma: Then we should add it: It makes no sense to retrieve all the transactions from the mempool using a filtering  method that does not filter... It even makes the code harder to understand.
Also, it is OK to retrieve all the transactions? Won't that be a lot of transactions? We are sending all the hashes from the mempool to the remote peer here.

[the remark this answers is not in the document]
gianbelinche: The idea of this function is to send all the transactions in the mempool, since we need to do that when we connect to a new peer. That is the only place this function is used, so there is no risk of sending them again.

crates/networking/p2p/rlpx/connection/server.rs
gianbelinche: It doesn't seem to be a backend message, since it is not something we want to send to the peer directly, but some logic we need to execute ourselves (which internally sends other messages to the peers)

ElFantasma:
I think the broadcasting logic should live outside `RLPxConnection`:
- On one hand `RLPxConnection` should be _only_ responsible for forwarding messages from the remote peer to the backend, and the other way around, from the backend to the remote peer.
- On the other hand, there should be a component in the backend, (probably the `TxBroadcaster` `GenServer`) that should manage the broadcasting. It should probably be coupled with the Mempool. It should have it's own ruleset to determine when and what to send to each peer.

That way the `RLPxConnection` gets simpler and easier to maintain, and if we need more complex rules for broadcasting, we just add them to the specific component.

crates/networking/p2p/rlpx/connection/server.rs
ElFantasma: Current code is doing exactly that: Sending the transactions to the remote peer (only "extra" logic is to check emptiness and a debug line)

crates/networking/p2p/network.rs
gianbelinche: We decided to not move the handle around and re add the send interval instead

crates/networking/p2p/rlpx/connection/server.rs

crates/networking/p2p/rlpx/connection/server.rs

crates/networking/p2p/rlpx/connection/server.rs
gianbelinche: Added an [issue](https://github.com/lambdaclass/ethrex/issues/4271) for the documentation

crates/blockchain/mempool.rs
```diff
@@ -170,6 +170,26 @@ impl Mempool {
         self.filter_transactions_with_filter_fn(&filter_tx)
     }
 
+    /// Gets all the transactions in the mempool
+    pub fn get_all_txs(&self) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
```
ElFantasma: Maybe rename it to provide a better description of what it returns?
(ie. `get_txs_by_sender` or `get_all_txs_by_sender`)

crates/networking/p2p/rlpx/connection/server.rs
ElFantasma, line 1: Love the removed complexity from `RLPxConnection` 😍

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -996,16 +900,6 @@ async fn handle_block_range_update(state: &mut Established) -> Result<(), RLPxEr
 
 pub(crate) fn broadcast_message(state: &Established, msg: Message) -> Result<(), RLPxError> {
```
ElFantasma, line 887: We'll need to remove the L2 broadcast feature from here as well... Maybe a new ticket for this?

crates/networking/p2p/tx_broadcaster.rs
```diff
@@ -54,17 +69,87 @@ impl TxBroadcaster {
...
+            .filter(|tx| matches!(tx.transaction(), Transaction::EIP4844Transaction { .. }))
+            .cloned()
+            .collect::<Vec<MempoolTransaction>>();
+
+        let txs_message = Message::Transactions(Transactions {
+            transactions: full_txs.clone(),
+        });
+
+        for (peer_id, mut peer_channels, capabilities) in peers {
+            if random::<f64>() < accept_prob {
```
ElFantasma: I don't like this probabilistic method to chose broadcast peers. It may end up with more or less than `sqrt(peers.len())` peers to broadcast. 
We should instead pick randomly `ceil(sqrt(peers.len()))` peers to ensure we broadcast to the exact expected amount of peers.

crates/blockchain/mempool.rs

crates/networking/p2p/tx_broadcaster.rs
gianbelinche: Done

crates/networking/p2p/rlpx/connection/server.rs
gianbelinche: Done [here](https://github.com/lambdaclass/ethrex/issues/4318)

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -454,50 +441,15 @@ where
...
-            log_peer_debug(
-                &state.node,
-                &format!("Sent {tx_count} transaction hashes to peer"),
-            );
-        }
-    }
-    Ok(())
-}
-
 async fn send_all_pooled_tx_hashes(state: &mut Established) -> Result<(), RLPxError> {
```
ElFantasma: To avoid code repetition (and potentially introducing bugs if modifying `TxBroadcaster` but not here) can we move this code to `TxBroadcaster`?
I mean, refactor the common code (eg. partition on chunks) and and use it on both cases:
- when starting a connection to send all mempool tx hashes
- on usual hashes broadcast.

ElFantasma (approved):
Good job!

PR #4262, opened 2025-09-02
feat(l1): only track state root of big accounts

iovoid:
**Motivation**

When inserting storages ranges during snap sync, we keep track of the storage root to insert to the same tree as we go from one chunk to another. We don't need to keep track if the state root is already complete.

**Description**

This PR only updates maybe_big_account_storage_state_roots if the account is, in fact, big (couldn't be fetched in a single batch).

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -1133,13 +1133,13 @@ impl Syncer {
 type StorageRoots = (H256, Vec<(NodeHash, Vec<u8>)>);
 
 fn compute_storage_roots(
-    maybe_big_account_storage_state_roots_clone: Arc<Mutex<HashMap<H256, H256>>>,
+    maybe_big_account_storage_state_roots: Arc<Mutex<HashMap<H256, H256>>>,
     store: Store,
     account_hash: H256,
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/sync.rs
```diff
@@ -1160,16 +1160,16 @@ fn compute_storage_roots(
...
 
-    maybe_big_account_storage_state_roots_clone
-        .lock()
-        .map_err(|_| SyncError::MaybeBigAccount)?
-        .insert(account_hash, computed_state_root);
-
     let account_state = store
         .get_account_state_by_acc_hash(pivot_hash, account_hash)?
         .ok_or(SyncError::AccountState(pivot_hash, account_hash))?;
     if computed_state_root == account_state.storage_root {
```
fedacking: Could we rename `computed_state_root` to `computed_storage_root`?

fedacking (commented):
LGTM small nitpick

crates/networking/p2p/sync.rs
iovoid: Done in 5b4e987

PR #4266, opened 2025-09-02
chore(l1): only validate trie in debug mode

iovoid:
**Motivation**

Currently, we reconstruct the whole state and storage tries as a validation. This is very expensive, taking several hours on mainnet.

**Description**

This moves the validations to a debug_assert, so they are only executed in debug mode.


fedacking (approved):
lgtm

jrchatruc (approved):

PR #4268, opened 2025-09-02
fix(l1): prevent panic in `eth68/Receipts` message handling

fmoletta:
**Motivation**
I ran into the following panic while running ethrex on sepolia testnet:
```
thread 'tokio-runtime-worker' panicked at crates/networking/p2p/rlpx/eth/eth68/receipts.rs:30:69:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This PR prevents this panic from happening

**Description**
* (eth68 message) `Receipts::new` no longer asumes we have at least one receipt for a given block

Closes #issue_number


crates/networking/p2p/rlpx/eth/eth68/receipts.rs
```diff
@@ -25,10 +25,15 @@ impl Receipts68 {
...
-        for r in &receipts {
-            transformed_receipts.push(vec![ReceiptWithBloom::from(&r[0])]);
-        }
+        let transformed_receipts = receipts
+            .iter()
+            .map(|receipts| {
+                receipts
+                    .first()
+                    .map(|r| vec![ReceiptWithBloom::from(r)])
+                    .unwrap_or_default()
```
MegaRedHand: Shouldn't this transform all of the receipts instead of only the first?

MegaRedHand (approved):

crates/networking/p2p/rlpx/eth/eth68/receipts.rs
fmoletta: Yes, base code looks wacky, Ill fix it

jrchatruc (approved):

PR #4274, opened 2025-09-03
ci(l1): fix daily hive report.

mpaulucci:
**Motivation**
Daily Hive report is broken

**Description**
- Fixes edge case where there are zero tests of a suite
- Revert back to stable Pectra release since the Osaka devnets only have osaka tests.

### DIFF
--- a/.github/workflows/common_hive_reports.yaml
+++ b/.github/workflows/common_hive_reports.yaml
@@ -78,7 +78,7 @@ jobs:
       - name: Set custom hive flags
         run: |
           if [[ "${{ matrix.test.simulation }}" == "ethereum/eest/consume-engine" || "${{ matrix.test.simulation }}" == "ethereum/eest/consume-rlp" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/common_hive_reports.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
       - name: Set custom hive flags
         run: |
           if [[ "${{ matrix.test.simulation }}" == "ethereum/eest/consume-engine" || "${{ matrix.test.simulation }}" == "ethereum/eest/consume-rlp" ]]; then
-            EEST_FIXTURES=$(cat tooling/ef_tests/blockchain/.fixtures_url)
+            EEST_FIXTURES="https://github.com/ethereum/execution-spec-tests/releases/download/v4.5.0/fixtures_develop.tar.gz"
```
MegaRedHand, line 81: Why not modify the fixtures version file instead?

mpaulucci: Because we're using devnet 4 as the fixture and it only contains the fusaka fixtures

PR #3760, opened 2025-07-21
fix(l2): use SP1 embedded allocator

xqft:
**Motivation**

SP1 by default uses a "[bump allocator](https://docs.succinct.xyz/docs/sp1/generating-proofs/advanced#embedded-allocator)" which never deallocs zkvm memory. If we try to prove a batch with ~1500 empty blocks, SP1 panics with an OOM error.

By enabling the `embedded-alloc` allocator the proof is generated correctly for this big batch. Tests are being made to find the new number of empty blocks which make SP1 OOM again.

PR #4275, opened 2025-09-03
feat(l1, l2): use SP1's bump allocator

ilitteri:
**Motivation**

We want to use SP1's bump allocator since it's faster than the embedded one and also enables more memory usage for the guest program in the zkVM. 
Currently, we are using the embedded allocator for two reasons: 1. It mitigates most of our allocation memory issues when executing the guest program with SP1, and 2. The bump allocator has some troubles with `rkyv` serialization.

**Description**

After reading [this discussion](https://github.com/eth-act/ere/issues/121), it seems to be a fix to the incompatibility issues we were having when using SP1's bump allocator with `rkyv` serialization.

After a few successful executions, it seems to be working.

**Test it out**

Run the following in `ethrex`'s root to periodically execute Mainnet blocks using SP1:
```
cargo r -r -p ethrex-replayer --features sp1 -- --execute --mainnet-rpc-url http://157.180.1.98:8545
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,7 +115,7 @@ spawned-rt = "0.3.0"
 lambdaworks-crypto = "0.11.0"
 tui-logger = { version = "0.17.3", features = ["tracing-support"] }
 rayon = "1.10.0"
-rkyv = { version = "0.8.10", features = ["std"] }
+rkyv = { version = "0.8.10", features = ["std", "unaligned"] }
 
 [patch.crates-io]
 secp256k1 = { git = "https://github.com/sp1-patches/rust-secp256k1", tag = "patch-0.29.1-sp1-5.0.0" }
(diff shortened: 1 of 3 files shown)


tomip01 (approved):

JereSalo (approved):

xqft (changes requested):
blocking this because it already has sufficient approvals, but we should discuss and test if this change could bring back OOM's for big batches of empty blocks (https://github.com/lambdaclass/ethrex/pull/3760), which can be generated on L2s with long batch times, like Sovra.

ilitteri:
> blocking this because it already has sufficient approvals, but we should discuss and test if this change could bring back OOM's for big batches of empty blocks (#3760), which can be generated on L2s with long batch times, like Sovra.

Based on this, I'll enable the feature `embedded` for L2.

PR #4282, opened 2025-09-03
test(l1,l2): run ef state tests with sp1

JereSalo:
**Motivation**

**Description**

- This works but of course it is a very ugly implementation of this, we may leave this like this because it's not a priority for now.


JereSalo:
We don't actually want this. Now that legacy tests were added to Blockchain tests runner we can run the blocks themselves with the blockchain runner. We can also remove the functionality of running state tests as blocks as it's no longer necessary.

PR #4284, opened 2025-09-03
feat(l1): implement osaka EIPs 7918 & 7892 + use `BlobSchedule` from EFTests

### LINKED ISSUE #4156: [EIP-7892] Blob Parameter Only ('BPO') Hardforks
https://forkcast.org/upgrade/fusaka#eip-7892

Acceptance criteria:
- Pass al spec tests related to 7892

### LINKED ISSUE #4157: [EIP-7918] Blob base fee bounded by execution cost
https://forkcast.org/upgrade/fusaka#eip-7918

Acceptance criteria:
- All spec tests related to 7918

fmoletta:
**Motivation**
Implement changes defined in [EIP-7918](https://eips.ethereum.org/EIPS/eip-7918) & [EIP-7892](https://eips.ethereum.org/EIPS/eip-7892)
Fix EFTests not using the `BlobSchedule` defined in the test files

**Description**
* Update `calc_excess_blob_gas` according to EIP
* Add `ForkBlobSchedule`s for Osaka & BPO1-BPO5 forks
* Enable EIP-7918 & EIP-7892  ef tests
* Use the `BlobSchedule` defined in each test file when running EF Tests

Closes #4157 & Closes #4156

### DIFF
--- a/tooling/ef_tests/blockchain/tests/all.rs
+++ b/tooling/ef_tests/blockchain/tests/all.rs
@@ -12,7 +12,7 @@ const SKIPPED_TESTS: &[&str] = &[
     "fork_PragueToOsaka",
     "fork_BPO0",
     "fork_BPO1",
-    "fork_BPO2"
+    "fork_BPO2",
 ];
 
 // If neither `sp1` nor `stateless` is enabled: run with whichever engine


mpaulucci (approved):

crates/common/types/genesis.rs
```diff
@@ -348,7 +403,19 @@ impl ChainConfig {
...
+        if self.is_bpo5_activated(block_timestamp) {
+            Some(self.blob_schedule.bpo5.unwrap_or_default())
+        } else if self.is_bpo4_activated(block_timestamp) {
+            Some(self.blob_schedule.bpo4.unwrap_or_default())
+        } else if self.is_bpo3_activated(block_timestamp) {
+            Some(self.blob_schedule.bpo3.unwrap_or_default())
+        } else if self.is_bpo2_activated(block_timestamp) {
+            Some(self.blob_schedule.bpo2.unwrap_or_default())
+        } else if self.is_bpo1_activated(block_timestamp) {
+            Some(self.blob_schedule.bpo1.unwrap_or_default())
```
MegaRedHand, line 424: Is it expected that these schedules could be `None` even if the bpo was activated? If it's not, I think we should return an error or log a warning. Alternatively, we could check that this doesn't happen when we load the config (i.e. no BPO has a valid activation timestamp but specifies no blob schedule).

crates/common/types/genesis.rs
```diff
@@ -110,13 +119,31 @@ pub struct BlobSchedule {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bpo1: Option<ForkBlobSchedule>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bpo2: Option<ForkBlobSchedule>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bpo3: Option<ForkBlobSchedule>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bpo4: Option<ForkBlobSchedule>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bpo5: Option<ForkBlobSchedule>,
```
MegaRedHand, line 142: We should consider turning these into some generic version like `Vec<(u64 /* activation_timestamp */, ForkBlobSchedule)>`

crates/common/types/genesis.rs
fmoletta: I agree with emitting a warning when loading the genesis file

fmoletta: Added check! [1f9acba](https://github.com/lambdaclass/ethrex/pull/4284/commits/1f9acba290d911cd53704cd358e69f7994455f80)

crates/common/types/genesis.rs
fmoletta: That would make serialization/deserialization a lot more complicated than it needs to be. We can consider it for a future PR but I don't see it as urgent as to need to add it to this PR. In comparison, geth also uses the preset bpo1-bpo5 optional fields

MegaRedHand (approved):

PR #4292, opened 2025-09-03
feat(l1): p256 verify implementation on L1

### LINKED ISSUE #4160: [EIP-7951] Precompile for secp256r1 Curve Support
https://forkcast.org/upgrade/fusaka#eip-7951

This has been implemented already for L2, it has to be moved to L1: https://github.com/lambdaclass/ethrex/blob/main/crates/vm/levm/src/l2_precompiles.rs#L54

Acceptance criteria:
- Pass al spec tests related to 7951

SDartayet:
**Description**

- Implements [EIP-7591](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7951.md), which consists of the L1 implementation for the P256 precompile that already exists on L2; with the main difference being gas cost.

- Unifies the L2 and L1 implementations of the precompile for our client, having a single function for both 

Closes #4160


crates/vm/levm/src/gas_cost.rs
```diff
@@ -151,7 +151,7 @@ pub const STATICCALL_WARM_DYNAMIC: u64 = DEFAULT_WARM_DYNAMIC;
 pub const WARM_ADDRESS_ACCESS_COST: u64 = 100;
 pub const COLD_ADDRESS_ACCESS_COST: u64 = 2600;
 pub const NON_ZERO_VALUE_COST: u64 = 9000;
-pub const BASIC_FALLBACK_FUNCTION_STIPEND: u64 = 2300;
```
SDartayet, line 154: This change is unrelated to the PR; but since the constant was unused I figured I'd remove it

crates/vm/levm/tests/p_256_verify.json
JereSalo, line 1: I think you accidentally pushed this

SDartayet: Ah, this is a file containing tests for the P256 verify. It was already there to test the L2 implementation, I updated it with the one provided in the EIP (and changed the gas costs to match)

crates/vm/levm/src/precompiles.rs
```diff
@@ -185,18 +197,26 @@ const FP2_ZERO_MAPPED_TO_G2: [u8; 256] = [
...
-    // Prague or newers forks should only use this precompiles
+    // Prague or newers forks should only use these precompiles
     // https://eips.ethereum.org/EIPS/eip-2537
     if PRECOMPILES_POST_CANCUN.contains(address) && fork < Fork::Prague {
         return false;
     }
 
-    PRECOMPILES.contains(address) || PRECOMPILES_POST_CANCUN.contains(address)
+    // P256 verify Precompile only existed on L2 before Osaka
+    if address == &P256_VERIFICATION_ADDRESS && fork < Fork::Osaka && matches!(vm_type, VMType::L1)
```
edg-l: can you reorder this if? put first the fork, then match, then address, this puts the cheaper comparison first and short circuits if possible

crates/vm/levm/src/utils.rs
```diff
@@ -616,6 +606,11 @@ impl<'a> VM<'a> {
             initial_accessed_addresses.insert(Address::from_low_u64_be(i));
         }
 
+        // Add the address for the P256 verify precompile post-Osaka
+        if self.env.config.fork >= Fork::Osaka {
```
JereSalo: I noticed you added this check in two different places, and it turns out this `initialize_substate` method wasn't removed in the PR that implemented the `initialize` method for `Substate` struct. Would you be kind enough to remove this method as it is no longer used? Thanks

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1026,104 @@ pub fn bls12_g1add(
...
+        .get(32..64)
+        .ok_or(PrecompileError::ParsingInputError)?;
+    let s = calldata
+        .get(64..96)
+        .ok_or(PrecompileError::ParsingInputError)?;
+    let x = calldata
+        .get(96..128)
+        .ok_or(PrecompileError::ParsingInputError)?;
+    let y = calldata
+        .get(128..160)
```
edg-l: you validate the length == 160, so all these gets arent needed, you should use slicing with the `reason = "bounds checked"` here.

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1026,104 @@ pub fn bls12_g1add(
...
+        {
+            return Ok(Bytes::new());
+        }
+
+        // Verify that the point formed by (x, y) is on the curve
+        let x: Option<P256FieldElement> = P256FieldElement::from_uint(x).into();
+        let y: Option<P256FieldElement> = P256FieldElement::from_uint(y).into();
+
+        let (Some(x), Some(y)) = (x, y) else {
+            return Err(InternalError::Slicing.into());
```
edg-l, line 1076: is slicing the correct error?

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1026,104 @@ pub fn bls12_g1add(
...
+
+    // Build verifier
+    let Ok(verifier) = p256::ecdsa::VerifyingKey::from_encoded_point(
+        &EncodedPoint::from_affine_coordinates(x.into(), y.into(), false),
+    ) else {
+        return Ok(Bytes::new());
+    };
+
+    // Build signature
+    let r: [u8; 32] = r.try_into().map_err(|_| InternalError::Slicing)?;
```
edg-l: you should be able to use `r.try_into()?;` here if i remember correctly.

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1026,104 @@ pub fn bls12_g1add(
...
+
+    // Verify message signature
+    let success = verifier.verify_prehash(message_hash, &signature).is_ok();
+
+    // If the verification succeeds, returns 1 in a 32-bit big-endian format.
+    // If the verification fails, returns an empty `Bytes` object.
+    if success {
+        let mut result = [0; 32];
+        result[31] = 1;
+        Ok(Bytes::from(result.to_vec()))
```
edg-l: if possible avoid the vec, i think Bytes has a from_slice or from_static method? (you can use from_static if you construct result in a const {} block as a const RESULT:..)

crates/vm/levm/tests/p_256_verify.json
mpaulucci: not sure if there type of files should be here, but if there are they should be in `fixtures`

mpaulucci: Do we really need this considering we now have EF tests?

crates/vm/levm/src/precompiles.rs
SDartayet: This was the error used by the L2 implementation (I basically just moved their code here), and I didn't change anything in this part, so presumably?

crates/vm/levm/src/precompiles.rs
SDartayet: Fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/7fdda397703a09f44a5be62a6530767eb915b92d)!

crates/vm/levm/src/precompiles.rs
SDartayet: Fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/bff430fa7271436d7b6b29ff9e273b3dc48fe437)!

crates/vm/levm/src/precompiles.rs
SDartayet: Yep it works. Fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/e3db2026cbd09601642f39f1d744f0f93e7d93e0)!

crates/vm/levm/src/precompiles.rs
SDartayet: Fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/3968fb47c1c1692a585299000074426085f0b0bc)! `from_slice` didn't exist for Bytes, so I went with the `from_static` approach. Thanks for the suggestion!

crates/vm/levm/tests/p_256_verify.json
SDartayet: I used it to check the implementation was correct, and I guess I left it in because it wasn't disruptive, but I guess we don't really need it either. I guess I'll remove it

SDartayet: [Done](https://github.com/lambdaclass/ethrex/pull/4292/commits/85961132db60329998d96503d42d414dc1b3d127)

crates/vm/levm/src/utils.rs
SDartayet: [Sorted](https://github.com/lambdaclass/ethrex/pull/4292/commits/cec557735030214f3c45be81d000540315d901b6)!

crates/vm/levm/src/precompiles.rs
```diff
@@ -185,18 +196,26 @@ const FP2_ZERO_MAPPED_TO_G2: [u8; 256] = [
 pub const G1_POINT_AT_INFINITY: [u8; 128] = [0_u8; 128];
 pub const G2_POINT_AT_INFINITY: [u8; 256] = [0_u8; 256];
 
-pub fn is_precompile(address: &Address, fork: Fork) -> bool {
+pub fn is_precompile(address: &Address, fork: Fork, vm_type: VMType) -> bool {
```
fedacking, line 199: Comment: this function could be more legible by having one structure that holds the precompiles and does the validation of if it should be used. Right now it's split between the multiple PRECOMPILES const and  the PRECOMPILES const in the execute_precompile function.

crates/vm/levm/src/precompiles.rs
```diff
@@ -208,8 +227,8 @@ pub fn execute_precompile(
 ) -> Result<Bytes, VMError> {
     type PrecompileFn = fn(&Bytes, &mut u64, Fork) -> Result<Bytes, VMError>;
 
-    const PRECOMPILES: [Option<PrecompileFn>; 18] = const {
-        let mut precompiles = [const { None }; 18];
+    const PRECOMPILES: [Option<PrecompileFn>; 512] = const {
```
MegaRedHand, line 230: Why was this bumped to 512? If it's due to an EIP, we should reference that in a comment.

crates/vm/levm/src/precompiles.rs
MegaRedHand: I agree that this could be improved. We could change this to return the precompile number in a newtype as `Option<PrecompileIndex>` (with `struct PrecompileIndex(usize)`). A simpler alternative would be to remove the `is_precompile` function altogether, and just try to run it as a precompile, which should return some sentinel value (such as `None`) if the address is not a precompile.

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1025,90 @@ pub fn bls12_g1add(
     Ok(Bytes::from(padded_result))
 }
 
+/// Signature verification in the “secp256r1” elliptic curve
+/// If the verification succeeds, returns 1 in a 32-bit big-endian format.
+/// If the verification fails, returns an empty `Bytes` object.
+/// Implemented following https://github.com/ethereum/RIPs/blob/89474e2b9dbd066fac9446c8cd280651bda35849/RIPS/rip-7212.md?plain=1#L1.
```
MegaRedHand: We should update this link. It is now an EIP, right?

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1025,90 @@ pub fn bls12_g1add(
...
+#[allow(clippy::indexing_slicing)] // The length of the calldata is checked before slicing, so it should never fail
```
MegaRedHand: We should narrow this down to the place that has this problem. Also, is it possible to use `expect` instead of `allow` here?

crates/vm/levm/src/precompiles.rs
SDartayet: The address for the P256 verify precompile is 0x100, so 256. I figured it would be best to bump it to the closest power of 2 that was greater.

crates/vm/levm/src/precompiles.rs
SDartayet: Yeah, I accidentally left that over. Fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/3d550e91117f74a6f94b1cbc6197fe4faa7a338d)!

crates/vm/levm/src/precompiles.rs
SDartayet: `is_precompile` is actually used by the VM to check if the `to` in a transaction is a precompile, so I think removing that would complicate the logic. I think if anything we should do the opposite: have `is_precompile` do the checks and remove the repeated checks in `execute_precompile`

crates/vm/levm/src/precompiles.rs
SDartayet: Didn't know about expect, but it works! Both fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/79fa5cbe4ff2301cad3684ff84208dc26bcc4563)

crates/vm/levm/src/precompiles.rs
```diff
@@ -1003,6 +1025,93 @@ pub fn bls12_g1add(
     Ok(Bytes::from(padded_result))
 }
 
+/// Signature verification in the “secp256r1” elliptic curve
+/// If the verification succeeds, returns 1 in a 32-bit big-endian format.
+/// If the verification fails, returns an empty `Bytes` object.
+/// Implemented following https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7951.md
+// The length of the calldata is checked before slicing, so it should never fail
```
MegaRedHand: We should move this comment to the `expect`.
Also, I've seen the syntax `#[expect(clippy::indexing_slicing, "length of the calldata is checked before slicing")]` somewhere, so we could try doing that instead.

SDartayet: Yep, fixed [here](https://github.com/lambdaclass/ethrex/pull/4292/commits/5c648ec581743fb1628ffd20272198e56395cfc2)!

MegaRedHand (approved):
LGTM

PR #4294, opened 2025-09-03
[WIP] Snap sync final peer score

pablodeymo:
**Motivation**

**Description**

Closes #issue_number


docker-compose.yml
```diff
@@ -0,0 +1,55 @@
+services:
```
mpaulucci: lets not have docker compose at the root of the repo. I think this was moved to tooling/sync

crates/networking/p2p/peer_handler.rs
```diff
@@ -388,56 +352,32 @@ impl PeerHandler {
...
-
-            let Some(mut free_downloader_channels) =
-                peer_channels.iter().find_map(|(peer_id, peer_channels)| {
-                    peer_id.eq(&free_peer_id).then_some(peer_channels.clone())
-                })
+            let Some((peer_id, mut peer_channel)) = self
+                .peer_scores
+                .lock()
+                .await
+                .get_peer_channel_with_highest_score(&self.peer_table, &SUPPORTED_SNAP_CAPABILITIES)
```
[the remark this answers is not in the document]
pablodeymo: good catch, bro

docker-compose.yml
fedacking: Removed, I think it got added on a merge conflict.

ElFantasma (approved):
I find all this effort a great improvement! 
But I think we should move another step forward and convert the whole peer handling logic, including kademlia table, into a `GenServer`. That way all the 
```
self.peer_scores.lock().await.do_something(...)
```
should become a cast message to the PeerHandler. Something like:
```
self.peer_handler.cast(Message::DoSomething{...}).await
```
In any case, I think it's ok to merge this, and if you agree with the GenServer approach, then we should create a ticket for that.

jrchatruc:
> I find all this effort a great improvement! But I think we should move another step forward and convert the whole peer handling logic, including kademlia table, into a `GenServer`. That way all the
> 
> ```
> self.peer_scores.lock().await.do_something(...)
> ```
> 
> should become a cast message to the PeerHandler. Something like:
> 
> ```
> self.peer_handler.cast(Message::DoSomething{...}).await
> ```
> 
> or better:
> 
> ```
> self.peer_handler.do_something(...).await
> ```
> 
> which internally creates de message and casts it
> 
> In any case, I think it's ok to merge this, and if you agree with the GenServer approach, then we should create a ticket for that.

I completely agree with this, let's create the ticket

PR #4297, opened 2025-09-03
feat: use separate send sockets for discovery with IPv4 and IPv6

Oppen:
**Motivation**

**Description**

Closes #issue_number


jrchatruc:
Closed since we will handle this differently (see #4232)

PR #4301, opened 2025-09-03
feat(l1): added the allocator jemalloc

fedacking:
**Motivation**

We want to properly monitor the ram usage and test if we have improved performance with other allocators. For this, we started with jemalloc.

**Description**

- added the jemallocator to ethrex
- bumped the jemallocator version used in ethrex-replay
- added an endpoint to our rpc which allows us to view a flamegraph of inuse memory
- gated jemalloc as well as the profiling behind 2 new feature flags jemalloc and jemalloc-profiling respectively
- logging the corrent global allocator on start


edg-l:
I believe the profiling feature should be optional and off by default, since it has performance overhead

iovoid:
Maybe malloc_conf can be removed, instead requiring the user to enable it using the env variable [MALLOC_CONF](https://jemalloc.net/jemalloc.3.html#tuning)?

mpaulucci (approved):

rodrigo-o:
> I believe the profiling feature should be optional and off by default, since it has performance overhead

Now it is, as well as jemalloc, which by default is enabled. We have a `jemalloc_profiling` feature flag as optional which adds the profiling feature to the dependency as well as set the malloc_conf 

> Maybe malloc_conf can be removed, instead requiring the user to enable it using the env variable [MALLOC_CONF](https://jemalloc.net/jemalloc.3.html#tuning)?

For now we have the `profiling` feature of the dependency gated behind the `jemalloc_profiling` flag. We could remove this and use the env variable but we need to compile the dependency always with profiling in that case (as was done before).

crates/networking/rpc/rpc.rs
```diff
@@ -69,6 +69,80 @@ use tower_http::cors::CorsLayer;
...
 
+#[cfg(feature = "jemalloc_profiling")]
+use axum::response::IntoResponse;
+
+#[cfg(feature = "jemalloc_profiling")]
+pub async fn handle_get_heap() -> Result<impl IntoResponse, (StatusCode, String)> {
+    let Some(mutex) = jemalloc_pprof::PROF_CTL.as_ref() else {
+        return Err((
+            StatusCode::NOT_IMPLEMENTED,
+            "jemalloc profiling is not available (compile with `jemalloc_profiling` and make sure env_target is not msvc)".into(),
```
MegaRedHand: We should mention somewhere (maybe in this message) that this feature only works on Linux.

rodrigo-o: Yes! I'll add it

rodrigo-o: It's already done here: [jamalloc_pprof only works on linux, make that explicit in errors and …](https://github.com/lambdaclass/ethrex/pull/4301/commits/d2141022c37b2c18c0cd62dab263226bc8fdf6ad)

I also added checks for the os

crates/networking/rpc/Cargo.toml
```diff
@@ -32,6 +32,7 @@ tokio-util = { workspace = true, features = ["codec"] }
 reqwest.workspace = true
 sha3 = "0.10.8"
 sha2.workspace = true
+jemalloc_pprof={version="0.8.0", optional = true, features=["flamegraph", "symbolize"]}
```
MegaRedHand: Nit

```suggestion
jemalloc_pprof = { version = "0.8.0", optional = true, features = ["flamegraph", "symbolize"] }
```

crates/networking/rpc/rpc.rs
```diff
@@ -171,6 +245,11 @@ pub async fn start_api(
     let cors = CorsLayer::permissive();
 
     let http_router = Router::new()
+        .route("/debug/pprof/allocs", axum::routing::get(handle_get_heap))
+        .route(
+            "/debug/pprof/allocs/flamegraph",
+            axum::routing::get(handle_get_heap_flamegraph),
+        )
```
MegaRedHand, line 252: I think we should put these behind the `jemalloc_profiling` feature, just in case.

MegaRedHand (approved):

crates/networking/rpc/rpc.rs
rodrigo-o: right now they had a stub that fails informing about the feature, but I'll take a look if its clearer and do a follow up if thats the case!

jrchatruc (approved):

PR #4324, opened 2025-09-04
feat(l1,l2): enable log levels update in runtime

### LINKED ISSUE #4299: Add the ability to change log level on runtime
It would be useful to change the log level on runtime to debug live errors that a restart might solve. Probably the solution implies using `tracing`'s `reload` [module](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html).

About how to send the update to the sequencer, an RPC call might work, maybe using the `admin` namespace. This is just an option, but other alternatives may be better

ManuelBilbao:
**Motivation**

Sometimes we want to change the log level, primarly for debugging, and restarting the node would break the case of study.

**Description**

Add a custom `admin_setLogLevel` endpoint that enables the node operator to specify a new log filter, just like with `RUST_LOG`.

How to test:
1. Run a node (e.g. `ethrex --dev`)
2. Change the log levels:
    ```
    curl localhost:8545 -H 'content-type: application/json' -d '{"jsonrpc": "2.0", "id": "1", "method": "admin_setLogLevel", "params": ["ethrex_dev::block_producer=info"]}'
     ```
3. You should now only see `ethrex_dev::block_producer` logs

> [!WARNING]
> The `admin` namespace is currently unauthenticated and cannot be turned off. Be aware of this in public nodes

Closes #4299

### DIFF
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -64,7 +64,7 @@ impl L2Command {
 
         let matches = app.try_get_matches_from(args_with_program)?;
         let init_options = Options::from_arg_matches(&matches)?;
-        l2::init_tracing(&init_options);
+        let log_filter_handler = l2::init_tracing(&init_options);
         let mut l2_options = init_options;
 
         if l2_options.node_opts.dev {
@@ -89,7 +89,7 @@ impl L2Command {
                 Some(contract_addresses.bridge_address);
             println!("Initializing L2");
         }
-        l2::init_l2(l2_options).await?;
+        l2::init_l2(l2_options, log_filter_handler).await?;
         Ok(())
     }
 }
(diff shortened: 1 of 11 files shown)


cmd/ethrex/l2/command.rs
```diff
@@ -64,15 +64,19 @@ impl L2Command {
...
             println!("Removing L1 and L2 databases...");
             remove_db(DB_ETHREX_DEV_L1, true);
             remove_db(DB_ETHREX_DEV_L2, true);
             println!("Initializing L1");
-            init_l1(crate::cli::Options::default_l1()).await?;
+            init_l1(
+                crate::cli::Options::default_l1(),
+                log_filter_handler.clone(),
+            )
+            .await?;
```
avilagaston9, line 79: Nit: It doesn't affect anything now, but I think we should simply pass `None` here

ManuelBilbao: Actually it allows to modify the L1 logs too when using `l2 --dev`

PR #4326, opened 2025-09-04
fix(l1): run tests on ARM runner

Oppen:
**Motivation**

**Description**

Closes #issue_number


Oppen:
> it seems overkill to duplicate everything to run on arm. It will consume a lot of resources.

I'm not sure what's duplicate. I'm using matrix to avoid duplicating the workflow descriptions.

> I would just add it in a single place where it has high probability of catching bugs. And maybe only on builds that are merged to main.

The motivation for this PR was that we have ARM-specific code and it made it to main with bugs because the CI doesn't look for them. If we can't trust the CI, we'll waste resources in manually testing all PRs or in chasing bugs after we ship them.
Our gold standard currently is integration tests, so that's where I'd enable it. We don't invest too much in unit tests, so if we only enable it in unit tests we're almost guaranteed to ship bugs. I left benchmarks and L2 code intentionally out.

In any case, I'd rather go the other way around and test widely by default and optimize if it shows to be a problem.

Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM rust:1.90 AS chef
 
 RUN apt-get update && apt-get install -y \
     build-essential \
+    cmake \
```
Oppen:
```suggestion
```

pablodeymo (approved):

.github/workflows/pr-main_l1.yaml
```diff
@@ -82,11 +82,13 @@ jobs:
...
-    # "Test" is a required check, don't change the name
-    name: Test
+    name: Test - ${{ matrix.runner }}
     needs: detect-changes
     if: ${{ needs.detect-changes.outputs.run_tests == 'true' }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner: [ubuntu-24.04, ubuntu-24.04-arm, macos-15]
```
ManuelBilbao: We are using Ubuntu 22.04 in release workflows to be compatible with Debian 12's glibc. Should we use the same here?

Oppen: Arguably, yes, but in practice it was marked as latest and I set it to what latest actually is, just to avoid surprise upgrades. I can change it, but it will no longer match everything else in the workflow (and I need to double-check that there are ARM builders with it).

Oppen:
```suggestion
        runner: [ubuntu-22.04, ubuntu-22.04-arm, macos-15]
```

ManuelBilbao (approved):

PR #4330, opened 2025-09-05
feat(l2): change the block timestamp precision to ms

### LINKED ISSUE #141: Make block timestamp more precise
Currently, setting the block time below 1 second results in the following error:
Image

This is because the precision of block timestamp is in second like the following:
Image

The goal is to have a more precise timestamp in `u64`.

Kanet1105:
**Motivation**
With the current timestamp precision, we cannot have sub-second block time. We understand that block time on L1 needs to be compatible with Ethereum spec, but in L2 land there are lots of chains aiming for sub-second block time mainly because of interactivity in the applications.

**Description**
Change the block timestamp precision from second to millisecond to allow sub-second block time.

Closes none

### DIFF
--- a/crates/l2/sequencer/block_producer.rs
+++ b/crates/l2/sequencer/block_producer.rs
@@ -110,6 +110,10 @@ impl BlockProducer {
         };
         let head_hash = head_header.hash();
         let head_beacon_block_root = H256::zero();
+        let timestamp: u64 = SystemTime::now()
+            .duration_since(UNIX_EPOCH)?
+            .as_millis()
+            .try_into()?;
 
         // The proposer leverages the execution payload framework used for the engine API,
         // but avoids calling the API methods and unnecesary re-execution.
@@ -120,7 +124,7 @@ impl BlockProducer {
         // Proposer creates a new payload
         let args = BuildPayloadArgs {
             parent: head_hash,
-            timestamp: SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs(),
+            timestamp,
             fee_recipient: self.coinbase_address,
             random: H256::zero(),
             withdrawals: Default::default(),


mpaulucci (approved):

MegaRedHand (changes requested):
Blocking this while we discuss internally

crates/l2/sequencer/block_producer.rs
```diff
@@ -110,6 +110,10 @@ impl BlockProducer {
         };
         let head_hash = head_header.hash();
         let head_beacon_block_root = H256::zero();
+        let timestamp: u64 = SystemTime::now()
```
ElFantasma, line 113: This timestamp in milliseconds will break fork activation checks. The `BuildPayloadArgs.timestamp` flows into `chain_config.fork(args.timestamp)` (payload.rs:132), `is_shanghai_activated()`, `is_cancun_activated()`, `is_prague_activated()`, and `is_osaka_activated()` — all of which compare against fork times defined in seconds. A millisecond value (~1.7 trillion in 2025) will always exceed any fork activation time, causing every fork to appear activated.

Also, this value ends up in the block header's `timestamp` field, which the `TIMESTAMP` opcode pushes to the EVM stack (levm `block.rs:55`). Solidity contracts universally expect `block.timestamp` in seconds — using milliseconds would break timelocks, vesting schedules, oracle staleness checks, and any time-dependent logic.

ElFantasma (commented):
Seems stalled and we should probably close this:

This change breaks fork activation logic and smart contract timestamp semantics. The `timestamp` field in `BuildPayloadArgs` flows into `chain_config.fork(args.timestamp)`, `is_shanghai_activated()`, `is_cancun_activated()`, `is_prague_activated()`, etc. — all of which compare against fork activation times defined in **Unix epoch seconds**. A millisecond timestamp (~1.7 trillion) will always be greater than any fork activation time, causing all forks to appear activated immediately regardless of actual time.

Additionally, the `TIMESTAMP` opcode (`block.rs:55`) pushes the block header timestamp directly to the EVM stack. Every deployed Solidity contract that reads `block.timestamp` expects seconds — using milliseconds would break time-based logic in virtually all contracts (timelocks, vesting, auctions, oracle staleness checks, etc.).

If sub-second block times are needed, the timestamp should remain in seconds and an alternative mechanism (e.g., a separate `block_time_ms` field for sequencer scheduling, or a sub-second counter that doesn't replace the header timestamp) should be used.

MegaRedHand:
Closing this due to staleness. A way to integrate this without breaking fork logic could be to add a new fork that changes timestamp precision. This would, however, require branches in multiple parts of the codebase.

PR #4333, opened 2025-09-05
fix(l1): multiple connection to the same peer

sfroment:
Signed-off-by: Sacha Froment <sfroment42@gmail.com>

**Motivation**
while trying to understand why I was seeing the same block being added multiple time I found that in look_for_peers we weren't checking if the peer was already connected or not

**Description**
Simply add a way to see if we are connected to the peer, and if not we try to connect other wise we do nothing

Closes #issue_number


mpaulucci:
Hi @sfroment . I think this has been fixed in https://github.com/lambdaclass/ethrex/pull/4340 . If not, please feel free to reopen de PR

Issue #4239, opened 2025-09-01
p2p: improve handling of tcp_port == 0


pablodeymo:
Some nodes have `tcp_port == 0`. In that cases, TCP connection throws an error.

We should try with the `udp_port`.

PR #4334, opened 2025-09-05
chore(l1): fix tcp port 0

gianbelinche:
**Motivation**
When connecting to a peer, sometimes the tcp port is 0, that indicates that it should be the same port as the udp port.

**Description**
When adding a node to kademlia, if the tcp port is 0, we use the udp port for the tcp port

Closes https://github.com/lambdaclass/ethrex/issues/4239

### DIFF
--- a/crates/networking/p2p/discv4/server.rs
+++ b/crates/networking/p2p/discv4/server.rs
@@ -221,7 +221,12 @@ impl DiscoveryServer {
             Entry::Occupied(_) => (),
             Entry::Vacant(entry) => {
                 let ping_hash = self.ping(&node).await?;
-                let contact = entry.insert(Contact::from(node));
+                let mut new_node = node.clone();
+                if new_node.tcp_port == 0 {
+                    // If the node does not have a TCP port, we assume it's the same as the UDP port
+                    new_node.tcp_port = new_node.udp_port;
+                }
+                let contact = entry.insert(Contact::from(new_node));
                 contact.record_sent_ping(ping_hash);
             }
         }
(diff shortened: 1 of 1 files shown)


fedacking (approved):

Oppen:
> When connecting to a peer, sometimes the tcp port is 0, that indicates that it should be the same port as the udp port.

I couldn't find mention to this in the discv4 or enr specs. Can you point to where this is specified?

gianbelinche:
> > When connecting to a peer, sometimes the tcp port is 0, that indicates that it should be the same port as the udp port.
> 
> I couldn't find mention to this in the discv4 or enr specs. Can you point to where this is specified?

As far as I know it is not present in the spec.
It was taken from our own [code](https://github.com/lambdaclass/ethrex/blob/0cae720162fd889eb18417010e5f091c6d097d35/crates/networking/p2p/types.rs#L178)
We should define it that is indeed the desired behaviour

gianbelinche:
> I think we should not allow tcp being 0. Not sure this is the proper fix

In that case we should close this PR and define wether we want to not accept a peer with tcp port 0 or just throwing an error if a peer with tcp port 0 tries to communicate via tcp (this is the default behavior prior to this pr, and it will allow a peer to use only udp discovery)

mpaulucci:
> > I think we should not allow tcp being 0. Not sure this is the proper fix
> 
> In that case we should close this PR and define wether we want to not accept a peer with tcp port 0 or just throwing an error if a peer with tcp port 0 tries to communicate via tcp (this is the default behavior prior to this pr, and it will allow a peer to use only udp discovery)

If you say that we (ethrex) are the only ones with that behaviour, then we should:
- not allow to configure ethrex to have tcp port 0
- not accept peers with tcp port 0

thoughts?

gianbelinche:
> If you say that we (ethrex) are the only ones with that behaviour, then we should:
> 
> * not allow to configure ethrex to have tcp port 0
> * not accept peers with tcp port 0
> 
> thoughts?

I couldn't find code in other nodes regarding this. All the nodes with tcp port 0 report a version of None, so I can't know which client they are from.

The second option is already covered, since if the tcp port is 0 no connection can be stablished.
I'm not sure if the first one is already covered

gianbelinche:
Closing due to the proposed changes not being specified anywhere.
The current code on main already handles not accepting a peer with tcp port 0.

PR #4172, opened 2025-08-26
feat(l1, l2): preprocess execution witness data

ilitteri:
**Motivation**

Block execution and proving replay using SP1 is currently very slow due to some computation being done inside the zkVM that can be avoided.

**Description**

This PR aims to perform said zkVM-expensive computation prior to zkVM execution.

This PR adds 3 convenient fields to `ExecutionWitnessResult`:
- `state_nodes: HashMap<H256, NodeRLP>`: A map that associates a MPT node hash to its raw RLP-encoded form. This is convenient because by having this generated before execution inside the zkVM we avoid some prover-expensive computation like hashing each node.
- `account_storage_root_hashes: HashMap<Address, H256>`:  A map that associates an account address to its storage root hash. It is convenient to have this precomputed before execution inside the zkVM because we avoid doing this association inside the zkVM (which is expensive, mainly because of the hashing and recurrent iteration over all the MPT nodes).
- `touched_account_storage_slots: HashMap<Address, Vec<H256>>`: This structure that it's only purpose is to track the accessed account storage slots. It is necessary to serve the `debug_executionWitness` requests properly.

**Samply Results**

**Before**

image

**After**

image

PR #4337, opened 2025-09-05
fix(l1, l2): remove trust assumptions from guest program inputs

tomip01:
**Motivation**

For performance reasons, [this PR](https://github.com/lambdaclass/ethrex/pull/4172) was focused on doing heavy work before pushing data to the zkVM.

Despite hashing and decoding beforehand is helpful for performance, these optimizations introduced some trust assumptions on the guest program about the correctness of the hashing and decoding computation.

**Description**

This PR aims to remove these trust assumptions by moving the computation of the convenient data structures from outside the guest program to the inside of it.

The block headers decoding, bytecode hashing, and state RLP nodes hashing computation is now part of the guest program execution (`crates/l2/prover/src/guest_program/src/execution.rs`), ensuring its correctness.

This PR also refactors the `ExecutionWitnessResult` by splitting it into two different structs with different responsibilities: `ExecutionWitness`, which essentially is an `RpcExecutionWitness` but with extra useful data like the chain config and the number of the first block to execute; and `GuestProgramState`, which is a result of a performance-friendly preprocessing of the `ExecutionWitness` **inside the zkVM** and has some other useful fields that are filled on-demand during stateless execution.

**Test it out**

Step into `ethrex`'s root directory and run, depending on your test case:

- Executing the latest mainnet block without any prover backend 
```
cargo r -r -p ethrex-replay -- block --execute --rpc-url http://157.180.1.98:8545
```
- Executing the latest mainnet block with SP1
```
cargo r -r -p ethrex-replay --features sp1 -- block --execute --rpc-url http://157.180.1.98:8545
```
- Proving the latest mainnet block with 
```
cargo r -r -p ethrex-replay --features sp1,gpu -- block --prove --rpc-url http://157.180.1.98:8545
```


crates/common/types/block_execution_witness.rs
```diff
@@ -43,17 +43,31 @@ pub struct ExecutionWitnessResult {
...
-    pub state_nodes: BTreeMap<H256, NodeRLP>,
+    /// This is computed during ExecutionWitness construction to avoid
+    /// issues of trust assumptions.
+    #[serde(skip)]
+    #[rkyv(with = rkyv::with::Skip)]
+    pub nodes_hashed: BTreeMap<H256, NodeRLP>,
+    /// This are the RLP-encoded nodes for the state trie.
+    /// They are latter encoded and moved to the state_node
+    #[rkyv(with = crate::rkyv_utils::VecVecWrapper)]
+    pub nodes: Vec<NodeRLP>,
```
ilitteri: By using `Vec<Bytes>` instead, we could avoid implementing `VecVecWrapper` and use `BytesVecWrapper`.

ilitteri (approved):

crates/l2/prover/src/guest_program/src/execution.rs
```diff
@@ -238,6 +244,45 @@ fn execute_stateless(
...
+        .drain(..)
+        .map(|node| (keccak(&node), node))
+        .collect();
+
+    // hash codes
+    db.codes_hashed = db
+        .codes
+        .drain(..)
+        .map(|code| (keccak(&code), code))
+        .collect();
```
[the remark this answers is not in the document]
tomip01: Yes, this is the intended behavior, we want to clear those fields.

crates/common/types/block_execution_witness.rs
tomip01: Done here [d69b132](https://github.com/lambdaclass/ethrex/pull/4337/commits/d69b132f8a7a480345aa063443e1d2e4034c4d71) and here [d688fa2](https://github.com/lambdaclass/ethrex/pull/4337/commits/d688fa221a67dc2eb28804263168ccd5368e381c)

tomip01: Tested with sp1 in execute mode, no significant performance impact observed.

crates/common/types/block_execution_witness.rs
```diff
@@ -391,12 +481,22 @@ impl ExecutionWitnessResult {
...
-                .map_or_else(|| block.header.hash(), |header| header.hash());
+                .unwrap_or(&block.header);
+
+            // headers hash should happen in the stateless execution
+            if header.hash.get().is_some() {
+                return Err(GuestProgramStateError::Custom(format!(
+                    "Block header hash is already set for {}",
+                    block.header.number
+                )));
+            }
```
JereSalo: Good check, if we truly want to check them all we should previously iterate over all the block headers in the witness and check this, shouldn't we?

crates/l2/prover/src/guest_program/src/execution.rs
```diff
@@ -229,17 +238,31 @@ struct StatelessResult {
     last_block_header: BlockHeader,
     last_block_hash: H256,
     non_privileged_count: U256,
+    #[cfg(feature = "l2")]
+    nodes_hashed: BTreeMap<H256, Vec<u8>>,
+    #[cfg(feature = "l2")]
+    codes_hashed: BTreeMap<H256, Vec<u8>>,
+    #[cfg(feature = "l2")]
+    parent_block_header: BlockHeader,
```
JereSalo, line 250: Could we add rustdocs all over this struct? I don't understand why some fields are compiled only with feature l2

tomip01:
> Not for necessarily tackling in this PR but could we possibly use the ExecutionWitnessRPC directly instead of the ExecutionWitness? What would it require?
> 
> The new ExecutionWitness is an RPC Witness with first_block_number and ChainConfig. We want to send these as prover input, the necessary change for us to be able to use the Witness RPC instead would be to send as prover input:
> 
> * Witness RPC
> * ChainConfig
>   I think we can get the first block number from the `Vec<Block>` that we are sending, can't we?
>   It's just for seeing if we can possibly change this (and if it's worth it), we can file an issue for this if it's something we want. Of course the RPC Witness should implement rkyv serialization as well.
> 
> I like the change that you made so now we don't have empty inputs, it was kinda confusing.

Yes, absolutely. We discussed this offline with @ilitteri  and, in a future refactor PR, we plan to see if we can consolidate into a single struct. I’ll also check what you mentioned to see if it works

crates/l2/prover/src/guest_program/src/execution.rs
tomip01: Absolutely! done [4c7c4dd](https://github.com/lambdaclass/ethrex/pull/4337/commits/4c7c4dd290a0a05981163c78772eb7fb09bc6596)

crates/common/types/block_execution_witness.rs
tomip01: Nice catch! done here [3e04ffc](https://github.com/lambdaclass/ethrex/pull/4337/commits/3e04ffc359cea9bbaf34e5de58955f811d9d7204)

ilitteri:
> Not for necessarily tackling in this PR but could we possibly use the ExecutionWitnessRPC directly instead of the ExecutionWitness? What would it require?
> 
> The new ExecutionWitness is an RPC Witness with first_block_number and ChainConfig. We want to send these as prover input, the necessary change for us to be able to use the Witness RPC instead would be to send as prover input:
> 
> * Witness RPC
> * ChainConfig
>   I think we can get the first block number from the `Vec<Block>` that we are sending, can't we?
>   It's just for seeing if we can possibly change this (and if it's worth it), we can file an issue for this if it's something we want. Of course the RPC Witness should implement rkyv serialization as well.
> 
> I like the change that you made so now we don't have empty inputs, it was kinda confusing.

As it is today, it seems obvious that `RpcExecutionWitness` and `ExecutionWitness` are pretty much the same leaving aside the extra, and probably removable, fields that `ExecutionWitness` has. The only reason we won't move forward with this is because in the near future we will "hint" the zkVM some values to improve the performance, and for this it will be useful to have a separation of the `RpcExecutionWitness` and the `ExecutionWitness` (which will be essentially an `RpcExecutionWitness` + `Hints`).

PR #4341, opened 2025-09-05
fix(l1): `ethrex_replay` temporary fix for block header hashes

tomip01:
**Motivation**

`ethrex_replay` was broken due to not finding the current block header in the `block_headers` map.

**Description**

Temporary solution for fixing it.
Initialize the header by looking it in the block

### DIFF
--- a/cmd/ethrex_replay/src/fetcher.rs
+++ b/cmd/ethrex_replay/src/fetcher.rs
@@ -52,35 +52,15 @@ pub async fn get_blockdata(
         ));
     }
 
-    debug!("Getting block data from RPC for block {requested_block_number}");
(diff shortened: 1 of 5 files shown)


ilitteri (approved):

xqft (approved):

tomip01:
> I'd add a comment like this one in order to be more clear about what we are doing: `/// Hashes headers in witness and in blocks only once if they are repeated to avoid double hashing.`

Done here [4688d09](https://github.com/lambdaclass/ethrex/pull/4341/commits/4688d097b3805cf633d6c76b5bf03debaba7f6fc)

Issue #4081, opened 2025-08-19
L1/L2: investigate possible 64-bit assumption problem in SP1


xqft:
We suspect that some problems during block execution/proving with SP1 are because we have assumptions in ethrex/levm that `usize`'s are 64 bit, whilst the SP1 zkVM is a 32 bit machine. This is a result of some blocks executing successfully with the `exec` backend but failing to do so in the SP1 zkVM.

There are two ways to approach this:
1. Find all occurrences in which we have a 64 bit assumption and fix them.
2. Wait for SP1 6.0.0, which apparently changes the architecture of the zkVM to 64 bit, as seen in the v6.0.0 of [this patch](https://github.com/sp1-patches/rust-secp256k1/compare/patch-0.29.1-sp1-4.1.0...patch-0.29.1-sp1-6.0.0), and re-execute the failing blocks.

option 2 is a quicker way to confirm that this is the reason those blocks fail to execute, but option 1 should be done nevertheless as those assumptions can be considered bugs, or at least identify all places where we make those and document them explicitly.

JereSalo:
We've ran all EFTests both blockchain and state to see if these kind of problems still existed and it seems that they have all been fixed already. Almost all tests pass and the few state tests that fail aren't because of this.
I think we can close this for now. Feel free to reopen

PR #4342, opened 2025-09-05
refactor(l1):  remove unneeded usize casts

fmoletta:
**Motivation**
Remove cases where we using `as usize` to cast `u64` values when not needed

**Description**

Contributes to #4081

### DIFF
--- a/crates/common/types/blobs_bundle.rs
+++ b/crates/common/types/blobs_bundle.rs
@@ -116,7 +116,7 @@ impl BlobsBundle {
         let max_blobs = max_blobs_per_block(fork);
         let blob_count = self.blobs.len();
 
-        if blob_count > max_blobs as usize {
+        if blob_count > max_blobs {
             return Err(BlobsBundleError::MaxBlobsExceeded);
         }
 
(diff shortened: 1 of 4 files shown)


crates/common/types/blobs_bundle.rs
```diff
@@ -196,10 +196,10 @@ impl AddAssign for BlobsBundle {
     }
 }
 
-const MAX_BLOB_COUNT: u64 = 6;
-const MAX_BLOB_COUNT_ELECTRA: u64 = 9;
+const MAX_BLOB_COUNT: usize = 6;
+const MAX_BLOB_COUNT_ELECTRA: usize = 9;
 
-fn max_blobs_per_block(fork: Fork) -> u64 {
+fn max_blobs_per_block(fork: Fork) -> usize {
```
mpaulucci, line 202: I thought we were getting rid of usize since it doesn't play well with zkevms

fmoletta: I thought the problem was because of the casts, so instead of having to to cast from a u64/u32 to a usize I left the constants as usize (as we only ever use them casted as usize) to avoid having to cast altogether. These are very small numbers so we shouldn't have a problem with size

JereSalo: For these particular cases it doesn't matter that much. The goal was indeed to get rid of usizes to avoid problems in 32 bit architecture but there are places in which it doesn't really matter. That's why we replaced some usizes for u32 but it's not very relevant

PR #4343, opened 2025-09-08
perf(levm): preload block accounts

### LINKED ISSUE #4306: Perf: Read all involved accounts for a block in a single db transaction
This is an experimental perf test idea.

When we execute a block, we need to access the `AccountInfo` for several accounts. We may not know in advance which addresses will be accessed by a smart contract, but we *do* know in advance the `from` (after sender recovery) and `to` addresses. We should be able to read all of them in a single db transaction, massively reducing overhead for blocks with a high amount of eth transactions. The case where this would theoretically be the most beneficial is for eth transfers, where most of the time is consumed by the read transactions themselves.

edg-l:
Preload all sender/to accounts from a block in batch before executing the block txs

Closes #4306

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -145,7 +145,11 @@ impl Store {
         block_hash: BlockHash,
         addresses: &BTreeSet<Address>,
     ) -> Result<BTreeMap<Address, AccountInfo>, StoreError> {
-        let state_trie = self.open_locked_state_trie(block_hash)?;
+        let Some(header) = self.get_block_header_by_hash(block_hash)? else {
+            return Ok(BTreeMap::new());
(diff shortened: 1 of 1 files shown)


crates/blockchain/vm.rs
```diff
@@ -6,7 +6,10 @@ use ethrex_common::{
 };
```
mpaulucci, line 6: What are the performance numbers that justify this added code?

edg-l:
it has improvements and regressions, i don't know if we want the pr as it is

image

image

Arkenan:
Closing, as it didn't improve performance as we expected. We might try it again in the future, but not for now.

PR #4346, opened 2025-09-08
fix(l1): prevent race conditions on getReceipts calls

avilagaston9:
**Motivation**

We are encountering the following error in our L2 integration tests:
```
2025-08-29T19:17:04.573865Z  WARN ethrex_l2::sequencer::l1_watcher: Error when getting logs from L1: eth_getLogs request error: Internal Error: Could not get receipt
```

After debugging, it turned out to be a race condition that can occur if a new block has just been marked as canonical in the store cache, but has not yet been updated in the db. If a `getReceipt` call arrives in this window, the node won't be able to find the receipts by block number.

**Description**

- Updates the engine API to identify blocks by hash instead of block number.
- Resolves the block hash by calling `get_canonical_block_hash` in the store.

Closes None

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -219,6 +219,7 @@ impl Store {
         &self,
         block_number: BlockNumber,
     ) -> Result<Option<BlockBody>, StoreError> {
+        // FIXME (#4353)
         let latest = self
             .latest_block_header
             .read()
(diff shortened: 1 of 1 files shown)


avilagaston9:
Added references to the issue in `get_receipt()` and `get_block_body()` [41219f6](https://github.com/lambdaclass/ethrex/pull/4346/commits/41219f67c3c141a23292b4349ea0fe4487e984d5)! @Oppen

PR #4347, opened 2025-09-08
feat(l1): avoid cloning the list of all block hashes

iovoid:
**Motivation**

We want to reduce RAM usage. On mainnet, the list of block hashes uses >700MB (23M blocks * 32B per hash).

**Description**

We avoid cloning this list when possible, and keep it in RAM less time if possible.

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -192,7 +192,11 @@ impl Syncer {
             debug!("Sync Log 1: In snap sync");
             debug!(
                 "Sync Log 2: State block hashes len {}",
-                block_sync_state.to_snap_block_hashes().len()
+                match block_sync_state {
+                    BlockSyncState::Full(_) => 0,
+                    BlockSyncState::Snap(ref snap_block_sync_state) =>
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/sync.rs
```diff
@@ -785,10 +780,15 @@ impl Syncer {
...
         // - Execute blocks after the pivot (like in full-sync)
-        let all_block_hashes = block_sync_state.to_snap_block_hashes();
-        let pivot_idx = all_block_hashes.len().saturating_sub(1);
+        let pivot_hash = match block_sync_state {
+            BlockSyncState::Full(_) => return Err(SyncError::NotInSnapSync),
+            BlockSyncState::Snap(snap_block_sync_state) => snap_block_sync_state
+                .block_hashes
+                .last()
+                .ok_or(SyncError::NoBlockHeaders)?,
+        };
```
Oppen, line 789: Nit: maybe this could still be a function:
```rust
fn get_snap_pivot_hash(&self) -> Result<BlockHash, SyncError> {
    // the same block of code
}
```

PR #4360, opened 2025-09-08
fix(l1):  use eth capability version when decoding messages + fix `BlockRangeUpdate` handling

### LINKED ISSUE #2785: [EIP-7642] Implement eth/69
See: https://eips.ethereum.org/EIPS/eip-7642

We should see if there are hive tests for this or if any client has implemented it already so that we can test it.

This is also being discussed in #history-expiry Discord channel of the EF

fmoletta:
**Motivation**
Being able to support eth/69 messages

**Description**
* Add missing validation to `BlockRangeUpdate` (latest_hash != zero)
* Disconnect from peers if `BlockRangeUpdate` is invalid
* Enable eth/69 as supported capability
* Use negotiated eth capability version when decoding incoming messages
* Enable hive tests for `BlockRangeUpdate`

Closes #2785

### DIFF
--- a/crates/networking/p2p/rlpx/connection/server.rs
+++ b/crates/networking/p2p/rlpx/connection/server.rs
@@ -211,7 +211,7 @@ impl GenServer for RLPxConnection {
                 log_peer_debug(&established_state.node, "Starting RLPx connection");
 
                 if let Err(reason) =
-                    initialize_connection(handle, &mut established_state, stream).await
+                    initialize_connection(handle, &mut established_state, stream, eth_version).await
                 {
                     if let Some(contact) = established_state
                         .table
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-main_l1.yaml
```diff
@@ -109,14 +109,14 @@ jobs:
     name: "Setup Hive"
     runs-on: ubuntu-latest
     env:
-      HIVE_COMMIT_HASH: 115f4d6ef1bdd2bfcabe29ec60424f6327e92f43
+      HIVE_COMMIT_HASH: 2d0e4fefb814d6815430c0f9e201fe8a045cf486
     steps:
       - uses: actions/checkout@v4
       - name: Setup Hive
         # TODO: Change repo back to https://github.com/ethereum/hive when
         # https://github.com/ethereum/hive/pull/1325 is merged
```
mpaulucci, line 117: can you remove this comment?

crates/networking/p2p/rlpx/connection/codec.rs
```diff
@@ -203,7 +213,14 @@ impl Decoder for RLPxCodec {
...
 
         let (msg_id, msg_data): (u8, _) = RLPDecode::decode_unfinished(frame_data)?;
-        Ok(Some(rlpx::Message::decode(msg_id, msg_data)?))
+        Ok(Some(rlpx::Message::decode(
+            msg_id,
+            msg_data,
+            *self
+                .eth_version
+                .read()
+                .map_err(|err| RLPxError::InternalError(err.to_string()))?,
```
mpaulucci, line 222: if we fail to decode a message, it's not necesarily an internal error, right? It could be that they sent us random bytes?

fmoletta: The mapping there is for when we fail to read from the `RwLock` that holds the `eth_version`, so it is indeed an internal error

.github/workflows/pr-main_l1.yaml
fmoletta: The whole comment or just the part mentioning the PR? As we are still using our forked hive

PR #4362, opened 2025-09-08
refactor(l2): simplify SQL tables

### LINKED ISSUE #4136: Merge rollup store SQL tables
We have a lot of key-value tables in the rollup store. This was done for compatibility with libmdbx, but now we don't support libmdbx anymore, we should merge them into more logic tables. The most visible merge is for all tables with `batch` as first column.

ManuelBilbao:
**Motivation**

We have a lot of SQL tables with a key-value format with the batch number as key.

**Description**

Merge many tables into a single one (`batches`) simplifying the codebase.

Closes #4136


crates/l2/storage/src/store_db/sql.rs
```diff
@@ -60,6 +57,16 @@ impl SQLStore {
...
             };
+
+            if let Ok(current_version) = store.get_version().await {
+                if current_version != MIGRATION_VERSION {
+                    return Err(RollupStoreError::VersionMismatch {
+                        current: current_version,
+                        expected: MIGRATION_VERSION,
+                    });
+                }
+            }
```
LeanSerra: I think we should return an error if we cant find the current version

tooling/migration/v1/migrate.sh
```diff
@@ -0,0 +1,157 @@
...
+
+batches=$(sqlite3 $db_file 'SELECT MAX(batch) from blocks')
+batch_inserts=""
+echo "Migrating $batches batches"
+
+for batch in $(seq $batches); do
+    first_block=$(sqlite3 $db_file "SELECT MIN(block_number) FROM blocks WHERE batch = $batch")
+    last_block=$(sqlite3 $db_file "SELECT MAX(block_number) FROM blocks WHERE batch = $batch")
+    batch_inserts="$batch_inserts ($batch, $first_block, $last_block, \"\"),"
+done
```
LeanSerra: This didn't take as long but we might want to log some progress too (read comment below for context)

tooling/migration/v1/revert.sh
```diff
@@ -0,0 +1,131 @@
...
+block_inserts=""
+echo "Reverting $batches batches"
+
+for batch in $(seq $batches); do
+    first_block=$(sqlite3 $db_file "SELECT first_block FROM batches WHERE number = $batch")
+    last_block=$(sqlite3 $db_file "SELECT last_block FROM batches WHERE number = $batch")
+    for block in $(seq $first_block $last_block); do
+        block_inserts="$block_inserts ($block, $batch),"
+    done
+done
```
LeanSerra: Testing this out it took a really long time (30+ minutes) trying to revert the migration of the db from `ethrex-staging-1` (~390 batches) we should at least log progress so that users don't think its stuck

crates/l2/storage/src/store_db/sql.rs
ManuelBilbao: If the DB is previous to this PR, it won't have a migration version, so this assumes it's in "version 0"

tooling/migration/v1/migrate.sh
ManuelBilbao: Moved to a Rust implementation

tooling/migration/v1/revert.sh
ManuelBilbao: Moved to a Rust implementation

docs/l2/upgrading.md
LeanSerra, line 1: This doc is outdated

tooling/migrations/src/cli.rs
```diff
@@ -141,3 +161,62 @@ async fn migrate_libmdbx_to_rocksdb(
...
+    if new_version == current_version {
+        println!("Database is already in desired version");
+        return;
+    }
+
+    let script_name = if new_version > current_version {
+        "migrate.sql"
+    } else {
+        "revert.sql"
+    };
```
LeanSerra, line 182: are these files missing for the first migration?

ManuelBilbao:
We are planning to remove SQL

PR #4367, opened 2025-09-09
feat(l2): start/pause sequencer admin endpoint

LeanSerra:
**Motivation**

We want to have more control over the sequencer, for this reason this PR implements a new admin server with an API to start/stop the sequencer commits. In the future this server can be expanded to provide a /health endpoint with information about the sequencer status.

**Description**

- This PR adds a new admin http server that starts with the sequencer and has access to the L1Committer mailbox to send start/stop messages
- Add to the L1Committer logic to handle a call message that stops the sequencer
- Add to the L1Committer logic to handle a call message that start the sequencer with an optional parameter of a start delay in ms
- Add 3 endpoints to the admin server they can be tested by running `make init-l2-dev` and then using any of the commands:
   - `/committer/start`: starts the committer `curl -i localhost:5555/committer/start`
   - `/committer/start/{delay}`: starts the committer with a delay in ms `curl -i localhost:5555/committer/start/60000`
   - `/committer/stop`: stops the committer `curl -i localhost:5555/committer/stop` 
- Add new page in docs for the admin server API


crates/l2/sequencer/mod.rs
```diff
@@ -181,19 +184,45 @@ pub async fn start_l2(
...
+            let (server_res, verifier_res) = tokio::join!(admin_server.into_future(), handle);
+            if let Err(e) = server_res {
+                error!("Admin server task error: {e}");
+            }
+            handle_verifier_result(verifier_res).await;
         }
-        Err(err) => {
-            error!("Task error: {err}");
+        (Some(handle), None) => {
+            handle_verifier_result(tokio::join!(handle).0).await;
```
tomip01, line 209: shouldn't be an error? Since `admin_server` didn't start correctly

crates/l2/sequencer/l1_committer.rs
```diff
@@ -592,10 +609,42 @@ impl L1Committer {
 
         Ok(commit_tx_hash)
     }
+
+    async fn stop_sequencer(&mut self) -> CallResponse<Self> {
```
avilagaston9: Maybe `stop_committer()` and `start_committer()`?

crates/l2/sequencer/mod.rs
LeanSerra: If the admin server didn't start correctly we just log it as an error but we still let the sequencer start. This is done because the admin server is not an essential part of the L2 and it can still work even if the server is not present.

crates/l2/sequencer/l1_committer.rs
LeanSerra: Agreed! Changed it [here](https://github.com/lambdaclass/ethrex/pull/4367/commits/1206d8a6dd78eaa5cfd2ce521f2907693c0227a1)

PR #4378, opened 2025-09-09
feat(l2): add ethrex-replay command for extending a genesis file with new accounts

xqft:
**Motivation**

This will be useful to test run an L2 or prove a batch in a big state scenario (in which the network has a big world trie and storage tries, which we think can degrade proving performance greatly).

For now it will just add N accounts with some balance, in the future we might add bytecode and random storage.

### DIFF
--- a/cmd/ethrex_replay/src/cli.rs
+++ b/cmd/ethrex_replay/src/cli.rs
@@ -969,8 +969,8 @@ impl SubcommandGenerateGenesis {
 
         let secp = Secp256k1::new();
         let mut keys_file = File::create(&self.keys_out)?;
-        let mut alloc = BTreeMap::new();
         let balance = U256::from(self.balance);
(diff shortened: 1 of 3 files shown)


cmd/ethrex_replay/src/cli.rs
```diff
@@ -916,3 +926,207 @@ pub async fn produce_custom_l2_block(
...
+
+fn public_key_to_address(public_key: &PublicKey) -> Address {
+    let public_key = public_key.serialize_uncompressed();
+    let hash = Keccak256::digest(&public_key[1..]);
+    Address::from_slice(&hash[12..])
+}
+
+#[derive(Subcommand)]
+pub enum EthrexReplayCommand {
+    #[command(
```
tomip01, line 1014: I believe from here on we may have some duplicated code from the base branch

xqft:
superseded by #4873

PR #4381, opened 2025-09-09
perf(levm): refactor substate backups

azteca1998:
**Motivation**

Currently, `Substate` backups clone everything, which is expensive.

**Description**

Implement the `Substate` as a stack of updates, which can be merged or discarded at need.

Closes #issue_number


crates/vm/levm/src/vm.rs
```diff
@@ -36,16 +37,284 @@ pub enum VMType {
...
+            fn next(&mut self) -> Option<Self::Item> {
+                let next_item = self.iter.next();
+                if next_item.is_none() {
+                    if let Some(parent) = self.parent {
+                        self.parent = parent.parent.as_deref();
+                        self.iter = parent.logs.iter();
+
+                        return self.next();
+                    }
+                }
```
MegaRedHand: Doesn't this need to be in creation order? Right now, it goes through the ones in the current substate before the parent's.

azteca1998: You're right, it's a `Vec`.

The other iterator is a `HashMap`, therefore order is irrelevant, but in this case it's important.

azteca1998: Fixed. Now they should be all extracted in the correct order.

crates/vm/levm/src/vm.rs
```diff
@@ -36,16 +37,269 @@ pub enum VMType {
     L2,
 }
 
-#[derive(Debug, Clone, Default)]
-/// Information that changes during transaction execution
+/// Information that changes during transaction execution.
+// Most fields are private by design. The backup mechanism (`parent` field) will only work properly
+// if data is append-only, therefore
```
Arkenan, line 42: nit: this comment looks unfinished?

azteca1998: Yes it is, I forgot what I wanted to write.

Oppen (approved):

PR #4388, opened 2025-09-09
feat(l1, l2): add action button to EthProofs in `BlockRunReport` slack message

ilitteri:
**Motivation**

Mainnet blocks whose numbers are multiples of 100 are proved in [EthProofs](http://ethproofs.org/). If we happen to prove one of them, it is useful to have a link to more proving stats for that block.

### DIFF
--- a/cmd/ethrex_replay/src/block_run_report.rs
+++ b/cmd/ethrex_replay/src/block_run_report.rs
@@ -89,83 +89,101 @@ impl BlockRunReport {
     }
 
     pub fn to_slack_message(&self) -> SlackWebHookRequest {
-        SlackWebHookRequest {
-            blocks: vec![
-                SlackWebHookBlock::Header {
-                    text: Box::new(SlackWebHookBlock::PlainText {
-                        text: match (&self.run_result, &self.replayer_mode) {
-                            (Ok(_), ReplayerMode::Execute) => {
-                                String::from("✅ Successfully Executed Block")
-                            }
-                            (Ok(_), ReplayerMode::ExecuteSP1) => {
-                                String::from("✅ Successfully Executed Block with SP1")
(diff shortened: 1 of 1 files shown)


cmd/ethrex_replay/src/block_run_report.rs
```diff
@@ -89,6 +89,36 @@ impl BlockRunReport {
     }
 
     pub fn to_slack_message(&self) -> SlackWebHookRequest {
```
mpaulucci, line 91: It's a bit weird to have this code here. Any reason you didn't choose to do this with GHA?

ilitteri: This is mostly useful for local proving since it takes too long to return a result, and it'd require a self-hosted runner with a GPU if we want to use GHA.

cmd/ethrex_replay/src/block_run_report.rs
```diff
@@ -89,6 +89,36 @@ impl BlockRunReport {
...
+            } else {
+                format!(
+                    "https://{}.etherscan.io/block/{}",
+                    self.network, self.number
+                )
+            },
+        }];
+
+        if let Network::PublicNetwork(_) = self.network {
+            // EthProofs only prove block numbers multiples of 100.
```
JereSalo:
```suggestion
        if let Network::PublicNetwork(PublicNetwork::Mainnet) = self.network {
            // EthProofs only prove block numbers multiples of 100.
```
I haven't seen blocks from testnets so I think this is correct, is it?

JereSalo (approved):
Nice, disregard my comment if ethProofs proves blocks from other networks

tomip01 (approved):

PR #4395, opened 2025-09-09
refactor(l1): decouple version-specific rlpx messages

### LINKED ISSUE #3032: Only use the highest shared capability
See this discussion: https://github.com/lambdaclass/ethrex/pull/2860#discussion_r2121732718

We should be able to know _before_ decoding a message, which capability it belongs to. 

So we should avoid code like this:
https://github.com/lambdaclass/ethrex/blob/15ef5f87183b0d99a91c8dde78c9b41b7aaf5c8e/crates/networking/p2p/rlpx/eth/receipts.rs#L99-L106

Instead, the caller of decode should know which version we are in and call `Receipts68::decode` or `Receipts69::decode`

This might need a refactor. This will allow the different versions to be more differenciated.

Acceptance criteria:
- Review the code to make sure we only use the highest version of a capability.
- Ignore a message when it is from a capability that has a lower version
- Refactor: I would proably avoid bundling multiple versions of a type in an enum like 
```
pub(crate) enum Receipts {
    Receipts68(Receipts68),
    Receipts69(Receipts69),
}
```
Instead, we could have both implement a common trait. Open to discussion though

fmoletta:
**Motivation**
We already know which version of messages such as `Status` and `Receipts` we are going to send/expecting to receive, so we should handle them separately instead of having unified logic for both eth/68 & eth/68 variants. This will save us from having overly-complicated logic when decoding these messages.

**Description**
* Remove `Receipts` enum message and instead use `Receipts68` & `Receipts69` directly
* Remove `Status` enum message and instead use `StatusMessage68` & `StatusMessage69` directly. Add `StatusMessage` trait to allow using pre-existing status validation logic

Closes #3032


crates/networking/p2p/rlpx/eth/status.rs
```diff
@@ -1,119 +1,13 @@
...
-        }
-    }
-
-    pub fn get_network_id(&self) -> u64 {
-        match self {
-            StatusMessage::StatusMessage68(msg) => msg.network_id,
-            StatusMessage::StatusMessage69(msg) => msg.network_id,
-        }
-    }
+pub trait StatusMessage {
```
MegaRedHand, line 5: This trait is implemented by both the 68 and 69 version status messages, but it isn't used anywhere else, right? In that case, we could drop it.

MegaRedHand (approved):

crates/networking/p2p/rlpx/eth/eth68/receipts.rs
```diff
@@ -44,10 +44,14 @@ impl Receipts68 {
...
     }
+
+    pub fn get_id(&self) -> u64 {
+        self.id
+    }
 }
 
 impl RLPxMessage for Receipts68 {
-    const CODE: u8 = 0x0F;
+    const CODE: u8 = 0x10;
```
MegaRedHand, line 54: Good catch!

crates/networking/p2p/rlpx/eth/status.rs
fmoletta: Its used so that we don't have to duplicate the logic in `validate_status`

PR #4396, opened 2025-09-09
feat(l1): replay use add block instead of stateless execution

JereSalo:
**Motivation**

**Description**

Closes #issue_number


cmd/ethrex_replay/src/cli.rs
```diff
@@ -443,6 +459,113 @@ async fn replay(cache: Cache, opts: &EthrexReplayOptions) -> eyre::Result<f64> {
...
+            let account_state_rlp_opt = guest_program
+                .state_trie
+                .as_ref()
+                .unwrap()
+                .get(&hashed_address)
+                .unwrap();
+
+            let Some(account_state_rlp) = account_state_rlp_opt else {
+                continue;
+            };
```
tomip01: You could use one less variable if you want
```suggestion
            let Some(account_state_rlp) = guest_program
                .state_trie
                .as_ref()
                .unwrap()
                .get(&hashed_address)
                .unwrap()
            else {
                continue;
            };
```

cmd/ethrex_replay/src/cache.rs
```diff
@@ -43,6 +43,29 @@ pub struct Cache {
 }
 
 impl Cache {
+    pub fn get_first_block_number(&self) -> u64 {
+        self.blocks
+            .iter()
+            .map(|block| block.header.number)
+            .min()
+            .unwrap()
```
ilitteri: Return an `eyre::Error` instead.

cmd/ethrex_replay/src/cli.rs
```diff
@@ -139,6 +152,8 @@ pub struct EthrexReplayOptions {
     pub rpc_url: Url,
     #[arg(long, group = "data_source")]
     pub cached: bool,
+    #[arg(long, help = "Execute with `add_block`, no prover backend")]
+    pub no_backend: bool,
```
ilitteri: In the future, we are going to do the "no backend" execution default, and if the user wants to run the exec prover backend, they would need to specify `--zkvm exec`. As for this PR, this means a huge refactor, so we'll leave it this way. Nevertheless, use the term "no guest program" or "no zkVM" instead of "no backend"/"no prover backend" since it requires some deeper understanding of our codebase (the other ones are more understandable for outsiders or people unrelated to our project internals).

cmd/ethrex_replay/src/cli.rs
```diff
@@ -443,6 +459,113 @@ async fn replay(cache: Cache, opts: &EthrexReplayOptions) -> eyre::Result<f64> {
     Ok(gas_used)
 }
 
+async fn replay_no_backend(cache: Cache, opts: &EthrexReplayOptions) -> eyre::Result<f64> {
```
ilitteri: Handle the `unwrap`s

ilitteri (changes requested):

crates/blockchain/vm.rs
```diff
@@ -44,16 +44,32 @@ impl StoreVmDatabase {
...
+        let result = self
+            .store
             .get_account_info_by_hash(self.block_hash, address)
-            .map_err(|e| EvmError::DB(e.to_string()))
+            .map_err(|e| EvmError::DB(e.to_string()));
+
+        #[cfg(feature = "replay")]
+        if result.is_err() {
+            return Ok(None);
+        };
```
ilitteri: We need to check this with @mpaulucci.

cmd/ethrex_replay/src/cli.rs
JereSalo: https://github.com/lambdaclass/ethrex/pull/4396/commits/9dd5334e7862bf1b1114e1250691d082bbecf72c

cmd/ethrex_replay/src/cache.rs
JereSalo: https://github.com/lambdaclass/ethrex/pull/4396/commits/9dd5334e7862bf1b1114e1250691d082bbecf72c

cmd/ethrex_replay/src/cli.rs
JereSalo: https://github.com/lambdaclass/ethrex/pull/4396/commits/9dd5334e7862bf1b1114e1250691d082bbecf72c

cmd/ethrex_replay/src/cli.rs
JereSalo: https://github.com/lambdaclass/ethrex/pull/4396/commits/9f8324e07a5dd39492638b2903a142679fb1435b

cmd/ethrex_replay/src/rpc/db.rs
JereSalo, line 1: I had another PR for this but it's easier to put this here. This fixes the keys field when using `eth_getProof`, which is important when executing with `add_block` because we use it to rebuild storage tries of all accounts involved before execution.

crates/blockchain/vm.rs
ilitteri: I think it'd be better to define a more specific (for more precision) error for this case, capture (matching it, hence the new error definition) it on the `ethrex-replay` side, and ignore it.

ilitteri: @mpaulucci suggested us to implement `VmDatabase` for a `StatelessVmDatabase` struct. I think that it's worth doing since we cannot afford having this feature flag.

JereSalo:
> Talking with @mpaulucci, we concluded that we cannot afford adding a feature flag `replay` to handle special cases for this tool.
> 
> We need to implement the trait `VmDatabase` for a `StatelessVmDatabase` struct.

But in `add_block` the database used is `StoreVmDatabase`. Maybe implementing that `StatelessVmDatabase` would add unwanted complexity.
I see if I can think of a way of solving this without a feature flag, but maybe the best approach for now is to remove that feature flag and accept that it won't work with reth nodes for now but we'll find a clean solution eventually. What do you think?

crates/common/trie/trie.rs
```diff
@@ -95,9 +95,9 @@ impl Trie {
         Ok(match self.root {
             NodeRef::Node(ref node, _) => node.get(self.db.as_ref(), Nibbles::from_bytes(path))?,
             NodeRef::Hash(hash) if hash.is_valid() => {
-                Node::decode(&self.db.get(hash)?.ok_or(TrieError::InconsistentTree)?)
-                    .map_err(TrieError::RLPDecode)?
-                    .get(self.db.as_ref(), Nibbles::from_bytes(path))?
+                let rlp = self.db.get(hash)?.ok_or(TrieError::InconsistentTree)?;
+                let node = Node::decode(&rlp).map_err(TrieError::RLPDecode)?;
+                node.get(self.db.as_ref(), Nibbles::from_bytes(path))?
```
JereSalo, line 100: This keeps the same functionality but code is clearer and I found it useful for debugging the trie. If we can keep it like this it would be great

JereSalo:
PR is working again with reth nodes without the `replay` feature flag.

cmd/ethrex_replay/src/cache.rs
```diff
@@ -46,6 +47,29 @@ pub struct Cache {
...
+        if let Some(network) = &self.network {
+            return network
+                .get_genesis()
+                .map(|genesis| genesis.config)
+                .map_err(|e| eyre::eyre!("Failed to get genesis config: {}", e));
+        }
+        Err(eyre::eyre!(
+            "Cache doesn't have network nor config, this shouldn't happen"
+        ))
+    }
```
ilitteri: If I'm not mistaken, by chaining the `if`s, you can get rid of the returns.

JereSalo: You are right! https://github.com/lambdaclass/ethrex/pull/4396/commits/301829e0f56ef56dec8885bf4be50c091b98b642

PR #4402, opened 2025-09-09
feat(replay): add `--cache-level` flag in `ethrex-replay` to decide whether to write the cache or not

ilitteri:
**Motivation**

This feature was already implemented in the `ethrex-replayer` but was not available for `ethrex-replay`. As `ethrex-replayer` already makes use of `ethrex-replay` as a library, it is almost free to add this feature to `ethrex-replay` while keeping it in `ethrex-replayer`.

**Description**

- Rename `CacheLevel::All` to `CacheLevel::On` due to the change in the context in which it is now used.
- Implements `--cache-level` arg of type `CacheLevel` with 3 variants: `on` to cache block replays, `failed` to cache only failed replays, and `off` to avoid caching.
- Refactors cache API
    - `write_cache` now derives the file name from the cache.
    - Cache file name generation is now centralized in a single function to reduce error surface.


cmd/ethrex_replay/src/cli.rs
```diff
@@ -414,6 +431,17 @@ async fn replay_block(block_opts: BlockOptions) -> eyre::Result<()> {
...
+    match opts.cache_level {
+        CacheLevel::On => write_cache(&cache, l2)?,
+        CacheLevel::Failed => {
+            if block_run_failed {
+                write_cache(&cache, l2)?;
+            }
+        }
+        CacheLevel::Off => {}
+    }
+
```
JereSalo: What if we want the cache of a block that panics in ethrex? This way we will lose the cache. I liked the other approach the most

ilitteri: In this case, we should handle the panics. I'll fix this asap.

ilitteri: Thinking it better I think that by moving this earlier in the function call we can keep both ideas.

ilitteri: Done in [7f2380b](https://github.com/lambdaclass/ethrex/pull/4402/commits/7f2380bb40db186ea27609942016ef369dbe231a). I'll file an issue for handling the panics properly.

cmd/ethrex_replay/src/fetcher.rs
```diff
@@ -104,24 +109,24 @@ pub async fn get_blockdata(
...
-    let block_cache_duration = block_cache_start_time.elapsed().unwrap_or_else(|e| {
+    let block_retrieval_duration = block_retrieval_start_time.elapsed().unwrap_or_else(|e| {
         panic!("SystemTime::elapsed failed: {e}");
     });
 
     debug!(
-        "Cached block {requested_block_number} in {}",
-        format_duration(block_cache_duration)
+        "Got block {requested_block_number} in {}",
+        format_duration(block_retrieval_duration)
```
tomip01: I think this is a bad merge conflict, we don't use the raw block anymore. The block is obtained above in the function

ilitteri: [d7f3b8b](https://github.com/lambdaclass/ethrex/pull/4402/commits/d7f3b8b2c9132fa78e064b991e815c34bee372bf)

tomip01 (approved):

PR #4405, opened 2025-09-10
docs(levm): fix minor comments in levm code

azteca1998:
**Motivation**

**Description**

Fix comments that were merged in #4381 before they could be fixed there.

Closes #issue_number


crates/vm/levm/src/vm.rs
```diff
@@ -39,7 +39,7 @@ pub enum VMType {
```
mpaulucci, line 39: `fix comments from last PR` is a useless description for someone who is going through the commits. Please update the title to make it descriptive

azteca1998: Is it better now?

MegaRedHand (approved):

PR #4407, opened 2025-09-10
fix(l1): prevent panic on oversized trie node responses

prestoalvarez:
Replace a `panic!` with safe handling when a peer returns more trie nodes than requested in storage healing. We now log a warning, increment failure counters, penalize the peer, and continue processing (extra nodes are ignored by zip truncation). Also adds `warn` to `tracing` imports.
- **Why**: Avoids a potential remote DoS where a malicious peer could crash the node by sending oversized responses. Improves robustness of sync.


mpaulucci (approved):

crates/networking/p2p/sync/storage_healing.rs
```diff
@@ -434,7 +434,19 @@ async fn zip_requeue_node_responses_score_peer(
...
+            .peer_scores
+            .lock()
+            .await
+            .record_failure(request.peer_id);
+        warn!(
+            "Peer {} sent more nodes than requested ({} > {}); ignoring extras",
+            request.peer_id,
+            nodes_size,
+            request.requests.len()
+        );
```
Oppen: Could this be a helper?
```rust
fn record_failure(&self, peer_id: H256, msg: &str) {        peer_handler


            .peer_scores
            .lock()
            .await
            .record_failure(peer_id);
        warn!(msg);
}

peer_handler.record_failure(request.peer_id,
    &format!( "Peer {} sent more nodes than requested ({} > {}); ignoring extras",
            request.peer_id,
            nodes_size,
            request.requests.len()));
```

Oppen (approved):

prestoalvarez:
@Oppen @mpaulucci

crates/networking/p2p/peer_handler.rs
```diff
@@ -149,6 +149,16 @@ impl PeerHandler {
         Self { peer_table }
     }
 
+    /// Records a peer failure with a warning message.
```
fedacking, line 152: This shouldn't be a warn, the end user doesn't care. It should be a debug, it's for developers trying to understand what's happening in peer handling.

crates/networking/p2p/peer_handler.rs
```diff
@@ -149,6 +149,16 @@ impl PeerHandler {
...
+    /// This centralizes peer penalty logic for consistency across call sites.
+    pub async fn record_failure(&self, peer_id: H256, msg: &str) {
+        warn!("{}", msg);
+        self.peer_scores
```
fedacking, line 156: Peer scoring is no longer handled through directly locking the table.

crates/networking/p2p/peer_handler.rs
```diff
@@ -149,6 +149,16 @@ impl PeerHandler {
         Self { peer_table }
     }
 
+    /// Records a peer failure with a warning message.
+    /// This centralizes peer penalty logic for consistency across call sites.
+    pub async fn record_failure(&self, peer_id: H256, msg: &str) {
+        warn!("{}", msg);
```
fedacking, line 155: The message should use the structured log mechanism.

fedacking (changes requested):
Requesting changes until merge conflicts are fixed.

crates/networking/p2p/peer_handler.rs
ElFantasma, line 155: **Bug — won't compile:** `PeerHandler` has no `peer_scores` field. The struct only has `peer_table` and `initiator`:
```rust
pub struct PeerHandler {
    pub peer_table: PeerTable,
    pub initiator: GenServerHandle<RLPxInitiator>,
}
```

All existing call sites use `peer_table.record_failure(...)` directly (e.g., line 292, 560, 722 in this file). I'd suggest dropping this wrapper and calling `peer_handler.peer_table.record_failure(&request.peer_id).await?` directly in `storage_healing.rs`, which is the established pattern.

If you do want to keep this wrapper, note that `PeerTable::record_failure` returns `Result<(), PeerTableError>` — this method silently drops that error.

crates/networking/p2p/sync/storage_healing.rs
```diff
@@ -413,7 +413,19 @@ async fn zip_requeue_node_responses_score_peer(
     }
 
     if request.requests.len() < nodes_size {
-        panic!("The node responded with more data than us!");
+        *failed_downloads += 1;
```
ElFantasma, line 416: The `record_failure` call here should match the existing pattern used elsewhere in this file (e.g., lines 306, 436, 492):
```rust
peer_handler
    .peer_table
    .record_failure(&request.peer_id)
    .await?;
```
This avoids the need for the new `PeerHandler::record_failure` wrapper (which doesn't compile anyway — see inline comment on that method).

Also, the `warn!` log can be moved inline here:
```rust
warn!(
    "Peer {} sent more nodes than requested ({} > {}); ignoring extras",
    request.peer_id, nodes_size, request.requests.len()
);
```

ElFantasma (changes requested):
This PR is stalled, but the concern is still valid. The existing pattern throughout the codebase is `self.peer_table.record_failure(&peer_id).await?` — see how it's used in the same file at lines 292, 560, 722, etc.

I'd suggest dropping the new `PeerHandler::record_failure` wrapper entirely and calling `peer_handler.peer_table.record_failure(&request.peer_id).await?` directly in `storage_healing.rs`, which matches the existing pattern at lines 306 and 492 of the same file.

Probably this should be closed and fix it in a new PR

MegaRedHand:
Closing due to staleness. The panic here no longer exists.

PR #4410, opened 2025-09-10
ci(replay): disable proving job with `ethrex_replay` until stable

tomip01:
**Motivation**

This job is failing consistently: proving with risc0 fails to compile, and with sp1 the cache is outdated.

**Description**

Remove the CI workflow triggers by temporarily commenting them in `.github/workflows/main_prover_l1.yaml` until it becomes stable.

I opened issue #4411 to re-enable it.

### DIFF
--- a/.github/workflows/pr-main_l1.yaml
+++ b/.github/workflows/pr-main_l1.yaml
@@ -232,30 +232,6 @@ jobs:
       - name: Run Hive Simulation
         run: chmod +x hive && ./hive --client-file .github/config/hive/clients.yaml --client ethrex --sim ${{ matrix.simulation }} --sim.limit "${{ matrix.test_pattern }}" --sim.parallelism 16 --sim.loglevel 1
 
-  run-replay:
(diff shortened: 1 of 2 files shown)


.github/workflows/main_prover_l1.yaml
```diff
@@ -1,94 +0,0 @@
-name: Replay proving
-on:
-  push:
-    branches: ["main"]
-  pull_request:
-    branches: ["**"]
-    paths:
-      - ".github/workflows/main_prover_l1.yaml"
```
MegaRedHand, line 8: Can't we just comment these triggers out instead? Plus, we can add a `workflow_dispatch:` trigger to manually test that this is working once we fix it.

tomip01: Nice, done here [b60a278](https://github.com/lambdaclass/ethrex/pull/4410/commits/b60a278b96b36ce2cf0e96bd4b90b77b581f658e)

PR #4416, opened 2025-09-10
Add scope for `ethrex-replay`

ilitteri:
**Motivation**

`ethrex-replay` has become big enough to have its own scope for PRs, separate from the current ones (`l1`, `l2`, and `levm`).

### DIFF
--- a/.github/workflows/pr_lint_pr_title.yml
+++ b/.github/workflows/pr_lint_pr_title.yml
@@ -43,6 +43,7 @@ jobs:
             l1
             l2
             levm
+            replay
 
           # Configure that a scope must always be provided.
           requireScope: true


.github/workflows/pr_lint_pr_title.yml
```diff
@@ -43,6 +43,7 @@ jobs:
             l1
```
mpaulucci, line 43: please add to Contributing.md

.github/workflows/pr_lint_pr_title.yml
```diff
@@ -43,6 +43,7 @@ jobs:
...
             l2
```
mpaulucci, line 44: and maybe you want to add a special label here: https://github.com/lambdaclass/ethrex/blob/main/.github/scripts/set-pr-labels.js#L22-L27

ilitteri: Thanks!

PR #4260, opened 2025-09-02
feat(l1): add rlpx spawned actor

juan518munoz:
**Motivation**

Current implementation of snap sync still functions with extensive and messy code, spawnification of parts of this code should make code clearer and more cohesive.

**Description**

Implement a new spawned actor `Downloader`, that should be responsible of handling requests made to peers of our node.

Closes #issue_number

PR #4417, opened 2025-09-10
feat(l1): add rlpx downloader (spawned actor)

juan518munoz:
Motivation

Current implementation of snap sync still functions with extensive and messy code, spawnification of parts of this code should make code clearer and more cohesive.

Description

Implement a new spawned actor `Downloader`, that should be responsible of handling requests made to peers of our node.

This PR helps as a middle step in converting the `PeerHandler` into an spawned actor as well.

Follow up of #4260

### DIFF
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -12,8 +12,9 @@ use ethrex_common::{
 };
 use ethrex_rlp::encode::RLPEncode;
 use ethrex_trie::Nibbles;
-use ethrex_trie::{Node, verify_range};
+use ethrex_trie::Node;
 use rand::seq::SliceRandom;
+use spawned_concurrency::tasks::GenServer;
 use tokio::sync::Mutex;
 
 use super::peer_score::PeerScores;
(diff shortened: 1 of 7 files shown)


crates/networking/p2p/peer_handler.rs
```diff
@@ -359,8 +341,9 @@ impl PeerHandler {
                 trace!("We didn't get a peer from the table");
                 continue;
             };
+            let available_downloader = Downloader::new(peer_id, peer_channel.clone());
```
fedacking: I think this clone is unnecessary.

crates/networking/p2p/rlpx/downloader.rs
```diff
@@ -0,0 +1,868 @@
...
+    },
+}
+
+impl GenServer for Downloader {
+    type Error = ();
+    type CallMsg = DownloaderCallRequest;
+    type CastMsg = DownloaderCastRequest;
+    type OutMsg = DownloaderCallResponse;
+
+    async fn handle_call(
```
fedacking: I don't think the call should await for long periods of time (up to the timeout here) I would change these for cast messages with a channel. On the side of the sender let's just wait for the response through the channel with a blocking recv.

juan518munoz: removed in [246946c](https://github.com/lambdaclass/ethrex/pull/4417/commits/246946cb8c2e694459a190cef29d53dc74e17e74)

crates/networking/p2p/peer_handler.rs
juan518munoz: removed in [246946c](https://github.com/lambdaclass/ethrex/pull/4417/commits/246946cb8c2e694459a190cef29d53dc74e17e74)

crates/networking/p2p/rlpx/downloader.rs
ElFantasma: I agree, we shouldn't use `call` on requests that go through the network.

gianbelinche: I guess the best option when everything is spawned is that the one who calls this is also a genserver that can pass its handle when casting the downloader, and then the downloader will cast the result back

fedacking:
Moved to draft as we consider reimplementing this

PR #4421, opened 2025-09-10
refactor(replay): rebuild storage tries before execution

JereSalo:
**Motivation**

**Description**

- When executing with stateless backend it would be nice to have the tries built before VM execution, it's more tidy.

Closes #issue_number


ilitteri:
We won't move forward with this PR as we are internally discussing with other teams whether to keep supporting the field `keys` in the `ExecutionWitness` or not. If we'd decide to keep it it'd be because of a potential optimization in the zkVM execution.

ilitteri:
Filed an issue for revisiting this PR in the near future https://github.com/lambdaclass/ethrex/issues/4437

Issue #4312, opened 2025-09-04
Investigate warning Failed to decode packet err=InvalidSize


gianbelinche:
The following warning is spammed
`2025-09-04T15:13:26.334809Z  WARN ethrex_p2p::discv4::server: Failed to decode packet err=InvalidSize`
Investigate why it happens

PR #4424, opened 2025-09-10
fix(l1): packet decoding decrease log level

gianbelinche:
**Motivation**
There are several warnings of `Failed to decode packet err=InvalidSize` and `Failed to decode packet err=HashMismatch`

**Description**
Since we have no control over what we receive on the udp socket, we can only reduce the log level to debug so that it is not spammed.

Closes https://github.com/lambdaclass/ethrex/issues/4312


gianbelinche:
With new changes in main this warn is already changed to a debug.

PR #4432, opened 2025-09-10
feat(l1): add docker compose with CL and EL

ManuelBilbao:
**Motivation**

**Description**

Add a `docker-compose.yaml` file that deploys a Consensus client (lighthouse) and ethrex with just one command:
```sh
docker compose up
```
`ETHREX_NETWORK` can be used to set the network, default is `mainnet`.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -25,6 +25,7 @@ services:
       --execution-endpoint http://ethrex:8551
       --execution-jwt /secrets/jwt.hex
       --checkpoint-sync-url https://${ETHREX_NETWORK:-mainnet}-checkpoint-sync.attestant.io
+      --checkpoint-sync-url-timeout 600
     depends_on:
       setup_jwt:
         condition: service_completed_successfully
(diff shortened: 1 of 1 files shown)


docs/l1/running.md
```diff
@@ -40,3 +40,21 @@ To sync with hoodi
...
+```
+
+Or you can set a different network:
+
+```sh
+ETHREX_NETWORK=hoodi docker compose up
+```
+
+> [!NOTE]
+> This method is currently supported only on Linux.
```
LeanSerra: Can't we add this on the docker compose file for the ethrex service 
```
platform: linux/amd64
```

(I tested it and it works but IDK if performance will be an issue with the emulation)

ManuelBilbao: It's a problem more related with Lighthouse than the ethrex platform. Do you manage to start sync with that flag?

LeanSerra: You are right I thought that it was running but lighthouse ended up timing out on the checkpoint sync. It's weird because if I run lighthouse on its own (without anything from the docker compose) it runs ok and starts syncing

ManuelBilbao: After testing a bit more, it appears to be a problem related to networking/geographics rather than macos. I'll remove the note

PR #4438, opened 2025-09-11
feat(l2): backend for OpenVM

xqft:
**Motivation**

Integrates OpenVM as a new prover backend. For now it only supports L1 proving, until we can implement blob verification using `openvm-kzg`.

Currently proving is slow as we didn't yet use the guest libraries/precompiles for crypto primitives.

**Description**

- adds an OpenVM guest program for stateless execution
- adds a new prover backend for OpenVM
- replaces `secp256k1` with `k256` in crates `ethrex-common`, `ethrex-blockchain`, `ethrex-l2-common`, `ethrex-levm` because of it being incompatible with the OpenVM target
- added a new kzg implementation using `openvm-kzg` because of the guest incompatibility with `c-kzg` and `kzg-rs`, for this it was necessary to separate implementations and add the `kzg-rs` and `openvm-kzg` feature flags in addition to the preexisting `c-kzg` as we can't no longer use `kzg-rs` by default.
- adds CI jobs for the new backend


crates/l2/prover/src/backend/mod.rs
```diff
@@ -12,6 +12,12 @@ pub mod risc0;
 #[cfg(feature = "sp1")]
 pub mod sp1;
 
+#[cfg(feature = "openvm")]
+pub mod openvm;
+
+#[cfg(all(feature = "l2", feature = "openvm"))]
+compile_error!("The OpenVM backend only supports L1 proving as of now");
```
ilitteri, line 19: What was the reason for this?

crates/vm/Cargo.toml
```diff
@@ -44,6 +44,9 @@ path = "./lib.rs"
 [features]
 default = []
 c-kzg = ["ethrex-levm/c-kzg", "ethrex-common/c-kzg"]
+kzg-rs = ["ethrex-levm/kzg-rs", "ethrex-common/kzg-rs"]
+openvm-kzg = ["ethrex-levm/openvm-kzg", "ethrex-common/openvm-kzg"]
+blst = ["revm/blst"]
```
ilitteri, line 49: Why this change?

crates/common/types/transaction.rs
```diff
@@ -1263,26 +1271,26 @@ impl Transaction {
...
+) -> Result<Address, EcdsaError> {
     // Hash message
     let payload: [u8; 32] = Keccak256::new_with_prefix(message.as_ref())
         .finalize()
         .into();
     recover_address(signature, H256::from_slice(&payload))
 }
 
-pub fn recover_address(signature: Signature, payload: H256) -> Result<Address, secp256k1::Error> {
+pub fn recover_address(signature: Signature, payload: H256) -> Result<Address, EcdsaError> {
```
ilitteri, line 1282: I think we can take advantage of the new feature flag `openvm-kzg` and have two versions for `recover_address`, this new one used under `openvm-kzg` and the old one used by default (mainly to avoid a performance regression in the L1).

@pablodeymo is already working in a definitive solution to this issue.

crates/vm/levm/src/constants.rs
```diff
@@ -69,7 +70,7 @@ pub const LAST_AVAILABLE_BLOCK_LIMIT: U256 = U256([256, 0, 0, 0]);
 
 // EIP7702 - EOA Load Code
 pub static SECP256K1_ORDER: LazyLock<U256> =
-    LazyLock::new(|| U256::from_big_endian(&secp256k1::constants::CURVE_ORDER));
+    LazyLock::new(|| U256::from_big_endian(&k256::Secp256k1::ORDER.to_be_bytes()));
```
ilitteri, line 73: Ditto.

crates/vm/levm/src/utils.rs
ilitteri, line 1: Ditto.

crates/l2/prover/src/backend/mod.rs
xqft: as I said in the PR description, openvm-kzg does not implement blob verification, which is one of the features that get enabled when proving an L2 block (blob verification is used for checking state diffs or transaction lists in rollup or validium mode respectively)

crates/vm/Cargo.toml
xqft: expanding a little bit what's in the PR description: previously we defaulted to use kzg-rs for our KZG library. We then had the c-kzg option which is more performant, and it's the crate that risc0 patches. SP1 patches kzg-rs.

Problem is that kzg-rs brings some sp1 syscall symbols that are not recognized by the openvm guest target, which means we can no longer use kzg-rs as a default, and c-kzg doesn't work in any zkvm target because it's bindings to a C library. So I also had to add a new option: [openvm-kzg](https://github.com/axiom-crypto/openvm-kzg/tree/main).

The solution I came with was to have all three options and no default, if no feature is enabled any call to the KZG library fails with a runtime error, so that it's not necessary for crates to enable/propagate a feature flag if they import ethrex-common but don't use the KZG lib.

The features are additive (doesn't matter which combination is set) and they follow a hierarchy if more than one is enabled, mostly based on performance (c-kzg > kzg-rs > openvm-kzg). A host target will use c-kzg almost always, zkvms might use different ones depending on which crate they patch (openvm doesn't patch anything so it has it's own option).

crates/common/types/transaction.rs
xqft: levm's ecrecover precompile uses k256 as per https://github.com/lambdaclass/ethrex/pull/4085/, secp256k1 was still being used for less relevant functions.

I don't think it's a good idea to use `openvm-kzg` for this, considering that the flag is meant to be used for a different operation. This would also conflict with the upcoming Zisk integration (Zisk also fails to compile with secp256k1 but won't use openvm-kzg).

considering also that Pablo is already working on this I think it's best to leave it as out of scope of this PR

crates/vm/Cargo.toml
ilitteri: Sorry, I think GitHub messed this up. My question was regarding the `blst` feature.

crates/common/types/transaction.rs
ilitteri: At the stage of the project, we cannot afford a regression in L1's performance nor in some work in progress.

I confused `k256` with `openvm-kzg`. In the diff, you are changing `recover_address` implementation to use `k256` instead of `secp256k1`. This functionality is core for L1's performance, so IMHO making a `k256` version to use under some convenient feature flag and keeping the `secp256k1` version as default (no features, this would be for L1).

But I invite others to give their opinion on this. cc @jrchatruc @mpaulucci

xqft: We talked internally, `recover_address` is not used for levm but it's currently a bottleneck of the `Transaction::sender` function, which is called frequently in the L1 client, so Ivan is right that a performance regression isn't viable in this case. We'll see how to proceed to avoid using k256 for this function

xqft:
superseded by https://github.com/lambdaclass/ethrex/pull/5412

PR #4450, opened 2025-09-12
revert(l1): remove eth/69 as supported until it is more stable.

mpaulucci:
**Description**
Apparently, the inclusion of eth/69 as a supported capability had a negative effect in the number of peers that users see in the network. Let's remove the support for now until we have made sure that it works properly, and that we have the same or more peers compared to only supporting eth/68


mpaulucci:
Moving to draft until we figure out if it's actually a problem or no

jrchatruc:
Turned out this was not actually the recent issue we encountered in snap sync, closing.

PR #4453, opened 2025-09-12
refactor(l1): make StorageTriesNodes a unique table

Oppen:
**Motivation**

Dupsort tables bring some extra cognitive complexity, as well as possibly some overhead,
and the way we're storing and indexing the nodes really doesn't make use of the dupsort
property, so we want to downgrade it to the simpler regular table kind.

**Description**

- **refactor(l1): rename Libmdbx(Locked)DupsortTrieDB to Libmdbx(Locked)PrefixedTrieDB**
- **refactor(l1): loosen bounds on Libmdbx(Locked)PrefixedTrieDB from DupSort to Table**
- **refactor(l1): make StorageTriesNodes table a regular table**


Oppen:
We're leaving libmdbx soon so there's no need to clean it up

PR #4454, opened 2025-09-12
refactor(replay): use `eth_getBlockByNumber` instead of `debug_getRawBlock`

tomip01:
**Motivation**

This pull request updates the block fetching logic in `fetcher.rs`.
Instead of relying on the debug endpoint, we now use `eth_getBlockByNumber`, which is implemented across all nodes. This makes the process more reliable and avoids depending on an endpoint that is not universally available.

**Description**

- Added a `hydrated` boolean parameter to `EthClient::get_block_by_number`, allowing callers to specify whether to retrieve full block data or just hashes. All call sites are updated to use this new parameter.
- Refactored `get_blockdata` in `cmd/ethrex_replay/src/fetcher.rs` to use `get_blockByNumber` with `hydrated=true` instead of `get_rawBlock`, and updated logic to construct the full `Block` from the response.

### DIFF
--- a/cmd/ethrex_replay/src/fetcher.rs
+++ b/cmd/ethrex_replay/src/fetcher.rs
@@ -1,13 +1,10 @@
 use std::time::{Duration, SystemTime};
 
-use ethrex_common::types::{Block, BlockBody, ChainConfig};
+use ethrex_common::types::ChainConfig;
 use ethrex_rpc::{
     EthClient,
     debug::execution_witness::execution_witness_from_rpc_chain_config,
-    types::{
-        block::BlockBodyWrapper,
-        block_identifier::{BlockIdentifier, BlockTag},
-    },
+    types::block_identifier::{BlockIdentifier, BlockTag},
 };
 use eyre::WrapErr;
 use tracing::{debug, info, warn};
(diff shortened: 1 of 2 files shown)


cmd/ethrex_replay/src/fetcher.rs
```diff
@@ -52,6 +55,47 @@ pub async fn get_blockdata(
...
+        body
+    } else {
+        return Err(eyre::eyre!("Expected full block body from RPC"));
+    };
+
+    let mut uncles = Vec::new();
+    for hash in block_body.uncles.iter() {
+        let uncle = eth_client.get_block_by_hash(*hash).await?;
+        uncles.push(uncle.header);
+    }
```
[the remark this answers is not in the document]
tomip01: Tried and tested that. No major improvements. For now I keep it as it is to be more readable

cmd/ethrex_replay/src/fetcher.rs
```diff
@@ -52,6 +55,47 @@ pub async fn get_blockdata(
...
+
+    let transactions = block_body.transactions.into_iter().map(|t| t.tx).collect();
+
+    let block = Block {
+        header: rpc_block.header,
+        body: BlockBody {
+            transactions,
+            ommers: uncles,
+            withdrawals: Some(block_body.withdrawals),
```
JereSalo: Can't we implement an `into` for this?

tomip01: I considered that, but it would require `into` to make RPC calls and become async. Would you prefer having an abstraction instead?

JereSalo: remove uncles 😎

tomip01: Done [7508499](https://github.com/lambdaclass/ethrex/pull/4454/commits/7508499a447b18e8348642f7b15114d9ff069b91)

PR #4476, opened 2025-09-15
fix(l2): proof_coordinator genserver locked by tcp listen loop

LeanSerra:
**Motivation**

Currently the proof_coordinator Genserver can only handle a single cast message `Listen { listener: Arc<TcpListener> }` as the handler for this messages starts an infinite loop that prevents other cast messages to be handled.

**Description**

Closes #issue_number

### DIFF
--- a/crates/l2/sequencer/proof_coordinator.rs
+++ b/crates/l2/sequencer/proof_coordinator.rs
@@ -1,4 +1,4 @@
-use crate::sequencer::errors::{ConnectionHandlerError, ProofCoordinatorError};
+use crate::sequencer::errors::ProofCoordinatorError;
 use crate::sequencer::setup::{prepare_quote_prerequisites, register_tdx_key};
 use crate::sequencer::utils::get_latest_sent_batch;
 use crate::{
@@ -17,17 +17,19 @@ use ethrex_metrics::metrics;
 use ethrex_rpc::clients::eth::EthClient;
 use ethrex_storage::Store;
 use ethrex_storage_rollup::StoreRollup;
+use futures::StreamExt;
 use secp256k1::SecretKey;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use spawned_concurrency::messages::Unused;
-use spawned_concurrency::tasks::{CastResponse, GenServer, GenServerHandle};
-use std::net::{IpAddr, SocketAddr};
+use spawned_concurrency::tasks::{CastResponse, GenServer, GenServerHandle, spawn_listener};
+use std::net::IpAddr;
 use std::sync::Arc;
 use tokio::{
     io::{AsyncReadExt, AsyncWriteExt},
     net::{TcpListener, TcpStream},
 };
+use tokio_stream::wrappers::TcpListenerStream;
 use tracing::{debug, error, info, warn};
 
 #[cfg(feature = "metrics")]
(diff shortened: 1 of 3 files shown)


crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -678,3 +563,34 @@ async fn send_response(
...
+        Ok(mut stream) => {
+            let mut buffer = Vec::new();
+            stream
+                .read_to_end(&mut buffer)
+                .await
+                .inspect_err(|err| error!("Failed to read from tcp stream: {err}"))
+                .ok()?;
+
+            serde_json::from_slice(&buffer)
+                .map(|data| ProofCordInMessage::Request(data, Arc::new(stream)))
```
MegaRedHand: Why do we need to wrap the stream in an `Arc` here?

MegaRedHand (approved):
LGTM

crates/l2/sequencer/proof_coordinator.rs
LeanSerra: Basically because `ProofCordInMessage` needs to implement `Clone` and `TcpStream` is not clonable.

MegaRedHand: You're right. After the question I checked in `spawned` and `GenServer` messages need to be `Clone` 😅
Well, at least until https://github.com/lambdaclass/spawned/pull/56 is merged 😎

Oppen: And are we sure there are no good reasons for that to not be `Clone`? A TCP stream is not a datagram oriented socket, meaning you could race to read and write if accessing from several tasks or interleaved inside a task.
I suspect we just want the `ProofCoordinator` to spawn a task per connection instead. That task would do the infinite loop for as long as the connection is alive, while the `ProofCoordinator` is free to handle the next cast message.

LeanSerra: You are right! I made some changes in [this](https://github.com/lambdaclass/ethrex/pull/4476/commits/a3c308bf4faed00c5fc6848b8a2c42eab69228dc) commit. 

Now we spawn a task that loops doing `listener.accept().await`. When it successfully accepts a connection it deserializes the request and then passes it and the stream over to the genserver to handle the response.
Right now I still need to wrap it in an Arc but, if the `Clone` requirements are removed from spawned, we can be 100% sure that only one task can have ownership of the stream at a given time so we have no races.

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -678,3 +560,47 @@ async fn send_response(
...
+        })?;
+    loop {
+        match listener.accept().await {
+            Ok((mut stream, _address)) => {
+                let mut buffer = Vec::new();
+                if stream
+                    .read_to_end(&mut buffer)
+                    .await
+                    .inspect_err(|err| error!("Failed to read from tcp stream: {err}"))
+                    .is_err()
```
MegaRedHand: Doing the read here means we have to wait until the client sends us the whole request. I guess this won't be exposed over the network, so we don't have to worry about a slowloris attack or something like that, but it wouldn't hurt to do this part in a spawned task. We can just spawn multiple one-shot tasks, like so:

```rust
for stream in listener.incoming() {
    spawned_rt::task::spawn(async move {
        let mut buffer = Vec::new();
        stream.read_to_end(&mut buffer).await;
        let message = serde_json::from_slice(&buffer);
        proof_coordinator.cast(message).await;
    })
}
```

LeanSerra: Done in [de46a22](https://github.com/lambdaclass/ethrex/pull/4476/commits/de46a220e19a5242797a36ec13f95b031e3c6301)

MegaRedHand (approved):
Looks better now

PR #4479, opened 2025-09-15
perf(levm): fix block hash caching

azteca1998:
**Motivation**

The original code made it so that the cache mechanism never worked properly.

**Description**

Fix the block hash caching mechanism so that the result is actually cached.

Closes #issue_number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 2025-09-15
 
-- Fix caching mechanism of block hashes [#4479](https://github.com/lambdaclass/ethrex/pull/4479)
+- Fix caching mechanism of the latest block's hash [#4479](https://github.com/lambdaclass/ethrex/pull/4479)
 
 ### 2025-09-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-09-15
+
+- Fix caching mechanism of block hashes [#4479](https://github.com/lambdaclass/ethrex/pull/4479)
+
 ### 2025-09-11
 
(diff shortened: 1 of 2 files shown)


Oppen (approved):

crates/storage/store.rs
```diff
@@ -199,14 +199,16 @@ impl Store {
...
-        if block_hash == latest.hash() {
-            return Ok(Some(latest));
+        {
+            let latest = self
+                .latest_block_header
+                .read()
+                .map_err(|_| StoreError::LockError)?;
+            if block_hash == latest.hash() {
+                return Ok(Some(latest.clone()));
+            }
```
MegaRedHand, line 209: Was the `clone()` the problem here? The new scope hides the change a bit.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-09-15
+
+- Fix caching mechanism of block hashes [#4479](https://github.com/lambdaclass/ethrex/pull/4479)
```
MegaRedHand:
```suggestion
- Fix caching mechanism of the latest block's hash [#4479](https://github.com/lambdaclass/ethrex/pull/4479)
```

MegaRedHand (approved):

crates/storage/store.rs
azteca1998: Yes, but not directly. The problem was that the `.clone()` was executed before `latest.hash()`. This made it so that the block's cached hash was updated on the clone, but not the cached block itself; which in turn meant that every time this function was called for the same block, the hash would have to be recomputed.

This change makes it so that `latest.hash()` will update the cached block's cached hash, then clone it, so that next time its hash will already be cached in the cached block.

Oppen: Can you check the same issue isn't present in other references to the cached header?

Arkenan (approved):
Great catch

Issue #1555, opened 2024-12-20
LEVM: account collision when address has storage


LeanSerra:
**Description**

Currently, we are not checking if the account has storage when doing a create, this means that if an account has no nonce, no bytecode but has storage we are not failing the contract creation.
Currently there is no easy way to get the complete storage of an account from the DB or cache.

**Failing tests**

The following tests might be failing because of this issue
- `create2collisionStorageParis.json`

damiramirez:
It seems that we have the same problem with test: `RevertInCreateInInit_Paris.json`

JereSalo:
We'll keep skipping these tests and there is no point in implementing this. The case described in the description of the issue is virtually impossible

PR #4484, opened 2025-09-15
feat(l1): update EEST to version 5.0

SDartayet:
**Motivation**

Keeping up to date with the latest EF tests for Osaka.

**Description**

This PR updates the fixture URL for the blockchain and state test runners. It also adds some necessary code for some of the tests to pass, and skips those which are poorly contructed.


mpaulucci (approved):

SDartayet:
> 4c1e3eb

Sure, added comments [here](https://github.com/lambdaclass/ethrex/pull/4484/commits/4c1e3eb5f6a954814f91556cd2bd43956daa364f)!

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -3,8 +3,18 @@ use std::path::Path;
...
+const SKIPPED_TESTS: &[&str] = &[
+    "system_contract_deployment", // Skipped because we don't have that validation in our code; also skipped because other clients do so too
+    "test_excess_blob_gas_fork_transition", // Skipped because genesis has Cancun fields when it shouldn't, so genesis loading fails
+    "test_invalid_post_fork_block_without_blob_fields", // Skipped because genesis has Cancun fields when it shouldn't, so genesis loading fails
+    "test_invalid_pre_fork_block_with_blob_fields", // Skipped because genesis has Cancun fields when it shouldn't, so genesis loading fails
+    "stTransactionTest/HighGasPriceParis",          // Skipped because it tries to
+    "dynamicAccountOverwriteEmpty_Paris", // Skipped because it fails on REVM as well. See https://github.com/lambdaclass/ethrex/issues/1555 for more details
+    "create2collisionStorageParis", // Skipped because it fails on REVM as well. See https://github.com/lambdaclass/ethrex/issues/1555 for more details
+    "RevertInCreateInInitCreate2Paris", // Skipped because it fails on REVM as well. See https://github.com/lambdaclass/ethrex/issues/1555 for more details
```
JereSalo: We probably should change these comments, we are not running with revm so we shouldn't say that we skip these tests because they fail with revm

SDartayet: What should we say then?

JereSalo: Something like "We don't support these tests as they contain impossible scenarios" maybe. If you want we can add that REVM fails these tests as well. The fact that they fail on REVM is not the reason behind why we skip them. I mean, there are some other tests that fail on REVM and not because of that we are going to skip them with LEVM.

So a possible comment would be:
`// We don't support these 3 tests as they contain impossible scenarios. See https://github.com/lambdaclass/ethrex/issues/1555`
Tune it however you like

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -9,34 +9,38 @@ const SKIPPED_TESTS: &[&str] = &[
...
     "test_invalid_post_fork_block_without_blob_fields", // Skipped because genesis has Cancun fields when it shouldn't, so genesis loading fails
     "test_invalid_pre_fork_block_with_blob_fields", // Skipped because genesis has Cancun fields when it shouldn't, so genesis loading fails
-    "stTransactionTest/HighGasPriceParis",          // Skipped because it tries to
-    "dynamicAccountOverwriteEmpty_Paris", // Skipped because it fails on REVM as well. See https://github.com/lambdaclass/ethrex/issues/1555 for more details
-    "create2collisionStorageParis", // Skipped because it fails on REVM as well. See https://github.com/lambdaclass/ethrex/issues/1555 for more details
-    "RevertInCreateInInitCreate2Paris", // Skipped because it fails on REVM as well. See https://github.com/lambdaclass/ethrex/issues/1555 for more details
+    "stTransactionTest/HighGasPriceParis", // Skipped because it sets a gas price higher than u64::MAX, which never happens in practice and most clients don't implement
+    "dynamicAccountOverwriteEmpty_Paris", // Skipped because the scenario described is extremely unlikely, since it implies doing EXTCODEHASH on an empty account that is then created
```
JereSalo: This is a red flag, saying that the scenario is extremely unlikely would mean that it could happen and if it does then LEVM would fail. My guess is that the scenario is actually impossible in public networks but I haven't seen the test fixture.
Take a deeper look at this please, if the reason why we skip this is not known yet document that instead, but we shouldn't say that we skip it because it's unlikely

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -3,10 +3,14 @@ use std::path::Path;
...
 
-#[cfg(not(feature = "revm"))]
+#[cfg(not(any(feature = "revm", feature = "sp1", feature = "stateless")))]
 const SKIPPED_TESTS: &[&str] = &[
     "system_contract_deployment",
-    "test_tx_gas_larger_than_block_gas_limit[fork_Osaka-blockchain_test-exceed_block_gas_limit_True]",
+    "stTransactionTest/HighGasPriceParis", // Skipped because it sets a gas price higher than u64::MAX, which most clients don't implement and is a virtually impossible scenario
+    "dynamicAccountOverwriteEmpty_Paris", // Skipped because the scenario described is virtually impossible
+    "create2collisionStorageParis", // Skipped because it's not worth implementing since the scenario of the test is virtually impossible. See https://github.com/lambdaclass/ethrex/issues/1555
+    "RevertInCreateInInitCreate2Paris", // Skipped because it's not worth implementing since the scenario of the test is virtually impossible. See https://github.com/lambdaclass/ethrex/issues/1555
```
MegaRedHand, line 12: It'd be good to have a brief description of the tested scenario in each one.

PR #4489, opened 2025-09-15
fix(l1, l2): fix alloy-consensus compilation by downgrading serde

xqft:
**Motivation**

https://github.com/alloy-rs/alloy/pull/2875 serde 1.0.22 breaks alloy-consensus compilation, which is a revm dependency.

this breaks some of our CI workflows

the easiest fix is to downgrade serde instead of upgrading revm.


ilitteri (dismissed):
Please file an issue in https://github.com/lambdaclass/rex

xqft:
Closing for now, the problem was in rex, in ethrex the serde version is locked to 1.0.219 (just before the breaking version), if we execute `cargo update` serde gets updated and the problem appears.

tomip01 (approved):

ilitteri (changes requested):
Blocking this PR until we decide internally that this patch is something desired.

PR #4497, opened 2025-09-16
perf(l1): cache transaction hash in HeadTransaction to avoid redundant computation

FredPhilipy:
During payload building, transaction hashes were being computed multiple times through repeated calls to `head_tx.tx.hash()`. While individual transactions have hash caching via `OnceCell<H256>`, the method call overhead and pattern matching still occurred on every access.

Resolves the TODO comment at line 516: `// TODO: maybe fetch hash too when filtering mempool so we don't have to compute it here`


mpaulucci:
Sorry, I don't think the extra complexity is worth it for a minimal benefit

PR #4500, opened 2025-09-16
perf(levm): refactor `bls12_g1add` to use lambdaworks

azteca1998:
**Motivation**

BLS12 computation is slow.

**Description**

Replace the BLS12 library with `lambdaworks`.

Closes #issue_number


crates/vm/levm/src/precompiles.rs
```diff
@@ -1007,22 +1011,90 @@ pub fn bls12_g1add(
...
+        };
+
+        let mut buffer = BytesMut::with_capacity(128);
+        buffer.put_u128(0);
+        buffer.put_u64(p2.0.representative().limbs[0]);
+        buffer.put_u64(p2.0.representative().limbs[1]);
+        buffer.put_u64(p2.0.representative().limbs[2]);
+        buffer.put_u64(p2.0.representative().limbs[3]);
+        buffer.put_u64(p2.0.representative().limbs[4]);
+        buffer.put_u64(p2.0.representative().limbs[5]);
```
Oppen: `representative` is a potentially expensive operation, as it computes a modular inverse to convert from Montgomery representation. I suggest you do it only once before serialization.

crates/vm/levm/src/precompiles.rs
```diff
@@ -1007,22 +1011,90 @@ pub fn bls12_g1add(
...
+        buffer.put_u128(0);
+        buffer.put_u64(p2.1.representative().limbs[0]);
+        buffer.put_u64(p2.1.representative().limbs[1]);
+        buffer.put_u64(p2.1.representative().limbs[2]);
+        buffer.put_u64(p2.1.representative().limbs[3]);
+        buffer.put_u64(p2.1.representative().limbs[4]);
+        buffer.put_u64(p2.1.representative().limbs[5]);
+        Ok(buffer.freeze())
```
Oppen: E.g.:
```suggestion
        let (rep_0, rep_1) = (p2.0.representative(), p2.1.representative());
        let mut buffer = BytesMut::with_capacity(128);
        buffer.put_u128(0);
        buffer.put_u64(rep_0.limbs[0]);
        buffer.put_u64(rep_0.limbs[1]);
        buffer.put_u64(rep_0.limbs[2]);
        buffer.put_u64(rep_0.limbs[3]);
        buffer.put_u64(rep_0.limbs[4]);
        buffer.put_u64(rep_0.limbs[5]);
        buffer.put_u128(0);
        buffer.put_u64(rep_1.limbs[0]);
        buffer.put_u64(rep_1.limbs[1]);
        buffer.put_u64(rep_1.limbs[2]);
        buffer.put_u64(rep_1.limbs[3]);
        buffer.put_u64(rep_1.limbs[4]);
        buffer.put_u64(rep_1.limbs[5]);
        Ok(buffer.freeze())
```

crates/vm/levm/src/precompiles.rs
```diff
@@ -1007,22 +1011,90 @@ pub fn bls12_g1add(
...
+                    let l = y01 * x10_inv;
+
+                    let x = l.square() - &p0.0 - p1.0;
+                    let y = l * (p0.0 - &x) - p0.1;
+
+                    (x, y)
+                }
+                Err(_) => unreachable!("already handled before"),
+            }
+        };
```
Oppen: This looks like something lambdaworks should be handling.

azteca1998: The unreachable? It's only reachable when `p0 == p1`, which we just handled before.
The match is legacy from before the tests were fixed, I'm thinking of changing it to just a division, which does not have a result but may panic (it won't since it's unreachable, but it'll be hidden behind `clippy::arithmetic_side_effects`).

crates/vm/levm/src/precompiles.rs
azteca1998: Yes, I'm not sure the compiler optimizes it, but I've a potentially faster way to serialize the output. I'll test the idea tomorrow.

crates/vm/levm/src/precompiles.rs
azteca1998: Fixed. Now it's hidden behind the division and depends on conditions which are impossible to meet.

crates/vm/levm/src/precompiles.rs
azteca1998: Fixed, now it takes less space and avoids the `bytes` library when populating the buffer.

edg-l (approved):

crates/vm/levm/src/precompiles.rs
Oppen: The whole block.

azteca1998: Oh, you meant the algorithm itself. It is handled, but only in projective coordinates, since it's optimized for multiple operations, but after asking about it I was told that if I only needed to add two points it is faster to do it using affine coordinates, which are not part of the library. That's the reason it's been implemented manually.

Arkenan (approved):
LGTM

PR #4514, opened 2025-09-16
fix(l1): remove weird slash from datadir path on macOS

MegaRedHand:
**Motivation**

On my machine (macOS), the default datadir was `/Users/mega/Library/Application-Support/ethrex/`, which differs from the usual default of `/Users/mega/Library/Application Support/ethrex/`.

**Description**

~This PR fixes this by removing a call to `ProjectDirs` that seems to be misplaced.~

~Note: the removed part was added in #4050~

This PR fixes this by skipping the call to `ProjectDirs` when the path is the default one.


MegaRedHand:
Superseded by #4521

PR #4515, opened 2025-09-16
fix(l1): remove second log level override from `init_tracing`

MegaRedHand:
**Motivation**

`RUST_LOG` is the common way to specify the log level for a Rust program. Currently, any global level specified by the envvar is being overriden by the value of the CLI flag `--log.level`, which defaults to `INFO`.

**Description**

This PR removes an `add_directive` call from the filter in `init_tracing`, which was causing the override. The CLI flag value will be set by the `with_default_directive` call, unless the user sets the `RUST_LOG` variable.


ManuelBilbao (approved):

SDartayet (approved):
LGTM!

PR #4516, opened 2025-09-16
fix(replay): insert addresses in keys field when using eth_getProof method

JereSalo:
**Motivation**

**Description**

- Addresses weren't being insterted with this method and in the PR that uses add_block we rely on those for rebuilding storage.


JereSalo:
This was already fixed as a part of another PR

PR #4521, opened 2025-09-16
feat!(l1): change `--datadir` to be relative to current working directory

MegaRedHand:
May supersede https://github.com/lambdaclass/ethrex/pull/4514

**Motivation**

Currently, the datadir is specified relative to the application directory (i.e. `/home/<user>/Application Support/` in macOS). This behavior is confusing, since usually path args are relative to the current working directory.

**Description**

This PR changes this behavior by removing the path-prefixing behavior. It also fixes the issue found in #4514, where any spaces in the path are replaced by slashes (even if the path is the default path).

### DIFF
--- a/cmd/ethrex/utils.rs
+++ b/cmd/ethrex/utils.rs
@@ -111,12 +111,6 @@ pub fn default_datadir() -> String {
 
 // TODO: Use PathBuf instead of strings
 pub fn init_datadir(data_dir: &str) -> String {
-    let project_dir = ProjectDirs::from("", "", data_dir).expect("Couldn't find home directory");
-    let data_dir = project_dir
-        .data_local_dir()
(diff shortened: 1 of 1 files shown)


cmd/ethrex/utils.rs
```diff
@@ -111,12 +111,6 @@ pub fn default_datadir() -> String {
 
 // TODO: Use PathBuf instead of strings
```
ManuelBilbao, line 112: Can we leverage this PR to address this?

MegaRedHand: I tried, but if I add this, the PR will balloon in diff size. Better to leave it for later.

MegaRedHand: Refactored in #4543

PR #4415, opened 2025-09-10
feat(l2): `/health` sequencer endpoint

LeanSerra:
**Motivation**

When an issue occurs on production we want an easy way to check the sequencer status

**Description**

- Add to the admin server the `/health` endpoint
  - For each component(Genserver) of the sequencer implement a call message Health and a response with information about the status of the component
     - TODO #4508 : implement it for BASED components
     - TODO #4509 :  implement it for proof coordinator
     - TODO #4510 : implement it for l1_proof_verifier
- Implement helper methods for `Signer` and `EthClient` to check their status
- Add to the admin server the `/admin/health` endpoint to check the status of the actual admin server

**How to test**

```
make init-l2-dev
```

```
curl -X GET http://localhost:5555/health
```

piping the output to jq can be used to filter components for example

```
curl -X GET http://localhost:5555/health |jq .l1_watcher
```
will show only the health status of the l1 watcher:

```
{
  "bridge_address": "0x39ba039e4862a44d6613b20373aaf90a8b3ba244",
  "check_interval": 12000,
  "l1_block_delay": 0,
  "l1_rpc_healthcheck": {
    "http://localhost:8545/": {
      "id": 1,
      "jsonrpc": "2.0",
      "result": "0x766f"
    }
  },
  "l2_rpc_healthcheck": {
    "http://localhost:1729/": {
      "id": 1,
      "jsonrpc": "2.0",
      "result": "0x14cb"
    }
  },
  "last_block_fetched": "30309",
  "max_block_step": "5000",
  "sequencer_state": "Sequencing"
}
```

PR #4522, opened 2025-09-17
feat(l2): implement /health for based components

### LINKED ISSUE #4508: L2: implement `/health` for based components
The following components from based were skipped from #4415

- `StateUpdater`
- `BlockFetcher`

We should implement a similar mechanism as in #4415 to verify the health of the components

FredPhilipy:
Implements health checks for `StateUpdater` and `BlockFetcher` components that were skipped in #4415.

- Add `CallMessage::Health` and health response structs for both components
- Modify `spawn()` methods to return `GenServerHandle` for admin server integration  
- Update admin server to include based components in `/health` endpoint
- Add integration test for health check serialization

**Health info includes:**
- ETH client connectivity status
- Component configuration (addresses, intervals)
- Current sequencer state

Closes #4508


crates/l2/sequencer/admin_server.rs
```diff
@@ -218,6 +228,28 @@ async fn health(
...
+        "block_fetcher".to_string(),
+        genserver_health(admin.block_fetcher, BlockFetcherCallMessage::Health, |msg| {
+            Some(match msg {
+                BlockFetcherOutMessage::Health(h) => h,
+                _ => return None,
+            })
+        })
+        .await,
+    );
+
```
LeanSerra: We should not insert into the response the health for based components if --based flag is not set. If we do we always get the message: `"block_fetcher": "Admin server does not have the genserver handle. Maybe it's not running?",`, which can be really confusing for operators that don't know that those components are only present  when using based sequencing and can lead to a false alarm being raised

crates/l2/tests/test_based_health.rs
LeanSerra, line 1: Let's remove this file

test_health_check.rs
LeanSerra, line 1: This shouldn't be on the root of the repo, let's remove it.

LeanSerra (changes requested):
Hi! Thanks for the contribution, functionality looks great but left some comments. Also the CI is not passing

FredPhilipy:
> Hi! Thanks for the contribution, functionality looks great but left some comments. Also the CI is not passing

Thanks, made some fix from comment and the CI looks good now.

LeanSerra (approved):
🚀

tomip01 (approved):

FredPhilipy:
Hi @LeanSerra is this good to go?

crates/l2/based/block_fetcher.rs
```diff
@@ -1,4 +1,9 @@
-use std::{cmp::min, collections::HashMap, sync::Arc, time::Duration};
+use std::{
+    cmp::min,
+    collections::{BTreeMap, HashMap},
```
ElFantasma: After rebasing onto current `main`, the `HashMap` import no longer exists in this file — it was removed. You'll only need to add `BTreeMap` (and `Serialize`) to the existing imports.

crates/l2/sequencer/mod.rs
```diff
@@ -157,8 +160,8 @@ pub async fn start_l2(
             rollup_store.clone(),
         )));
     }
-    if cfg.based.enabled {
-        let _ = StateUpdater::spawn(
+    let (state_updater, block_fetcher) = if cfg.based.enabled {
```
ElFantasma: On current `main`, `StateUpdater::spawn()` already returns `GenServerHandle<StateUpdater>` and is already wired into the admin server. The `let (state_updater, block_fetcher) = if cfg.based.enabled { ... }` restructuring will conflict heavily — after rebase, only the `BlockFetcher` handle needs to be captured and threaded through to `start_api`.

crates/l2/sequencer/admin_server.rs
```diff
@@ -1,3 +1,11 @@
+use crate::based::{
```
ElFantasma: On current `main`, `admin_server.rs` already imports `StateUpdater` and `StateUpdaterCallMessage`, and the `Admin` struct already has `state_updater: Option<GenServerHandle<StateUpdater>>`. Only the `BlockFetcher`-related imports and the block_fetcher health block in the `health()` handler are still needed after rebase.

crates/l2/sequencer/state_updater.rs
```diff
@@ -1,15 +1,15 @@
-use std::{sync::Arc, time::Duration};
+use std::{collections::BTreeMap, sync::Arc, time::Duration};
```
ElFantasma, line 1: All changes in this file are already on `main` (and the file has moved to `crates/l2/sequencer/state_updater.rs`). The current `main` version already has `CallMessage` (with `StopAt` variant), `handle_call`, `GenServerHandle` return from `spawn()`, and more. This entire file diff can be dropped after rebase.

ElFantasma (changes requested):
This PR is heavily stale (Sep 2025) and roughly half of it has already been merged into `main`. Specifically, `StateUpdater` on current `main` already has `CallMessage`, `handle_call`, returns `GenServerHandle` from `spawn()`, and is wired into the admin server — so all the `state_updater.rs` and most of the `sequencer/mod.rs` changes are already superseded.

The remaining value is the **BlockFetcher health check** (adding `CallMessage::Health`, `BlockFetcherHealth` struct, `handle_call`, returning `GenServerHandle` from `spawn()`, and wiring it into the `/health` endpoint). This part is still needed on current `main`. A rebase is required — the diff will shrink significantly and there will be conflicts in `sequencer/mod.rs` and `admin_server.rs` due to restructuring.

**crates/l2/sequencer/mod.rs** — On current `main`, `BlockFetcher::spawn()` returns `Result<(), BlockFetcherError>` and the result is discarded with `let _ =`. The PR changes it to return `GenServerHandle`, but the surrounding code structure in `start_l2` has changed significantly (e.g., `state_updater` is already handled differently). This file will need careful conflict resolution.

**crates/l2/based/block_fetcher.rs** — The `HashMap` import in the original diff adds `BTreeMap` alongside `HashMap`, but on current `main` the `HashMap` import was removed (the file only uses `std::cmp::min`). After rebase, only `BTreeMap` needs to be added.

FredPhilipy:
> This PR is heavily stale (Sep 2025) and roughly half of it has already been merged into `main`. Specifically, `StateUpdater` on current `main` already has `CallMessage`, `handle_call`, returns `GenServerHandle` from `spawn()`, and is wired into the admin server — so all the `state_updater.rs` and most of the `sequencer/mod.rs` changes are already superseded.
> 
> The remaining value is the **BlockFetcher health check** (adding `CallMessage::Health`, `BlockFetcherHealth` struct, `handle_call`, returning `GenServerHandle` from `spawn()`, and wiring it into the `/health` endpoint). This part is still needed on current `main`. A rebase is required — the diff will shrink significantly and there will be conflicts in `sequencer/mod.rs` and `admin_server.rs` due to restructuring.
> 
> **crates/l2/sequencer/mod.rs** — On current `main`, `BlockFetcher::spawn()` returns `Result<(), BlockFetcherError>` and the result is discarded with `let _ =`. The PR changes it to return `GenServerHandle`, but the surrounding code structure in `start_l2` has changed significantly (e.g., `state_updater` is already handled differently). This file will need careful conflict resolution.
> 
> **crates/l2/based/block_fetcher.rs** — The `HashMap` import in the original diff adds `BTreeMap` alongside `HashMap`, but on current `main` the `HashMap` import was removed (the file only uses `std::cmp::min`). After rebase, only `BTreeMap` needs to be added.

Thanks for the detailed review @ElFantasma! 

I've rebased onto current main and cleaned up the diff:

Dropped all StateUpdater structural changes that were already merged (spawn returning handle, CallMessage, file relocation to sequencer/)
Added Health variant to StateUpdater's existing CallMessage enum and the corresponding StateUpdaterHealth struct
Kept BlockFetcher health check as the main new addition (CallMessage, Health struct, handle_call, spawn returning handle)
Updated admin_server.rs to include both components in /health — only the new health blocks, built on top of the existing genserver_health helper
The diff went from ~162 to ~130 lines across 4 files
Should be much cleaner now. Let me know if anything else needs adjustment!

crates/l2/based/block_fetcher.rs
```diff
@@ -335,14 +357,37 @@ impl BlockFetcher {
...
+    async fn health(&self) -> CallResponse<Self> {
+        let eth_client_healthcheck = self.eth_client.test_urls().await;
+
+        CallResponse::Reply(OutMessage::Health(BlockFetcherHealth {
+            eth_client_healthcheck,
+            on_chain_proposer_address: self.on_chain_proposer_address,
+            fetch_interval_ms: self.fetch_interval_ms,
+            last_l1_block_fetched: self.last_l1_block_fetched.to_string(),
+            fetch_block_step: self.fetch_block_step.to_string(),
+            sequencer_state: format!("{:?}", self.sequencer_state.status()),
```
ElFantasma, line 370: Bug: `self.sequencer_state.status()` is an `async fn` (returns `impl Future<Output = SequencerStatus>`). Without `.await`, this formats the Future object instead of the actual status value — either a compile error or meaningless output.

Should be:
```rust
sequencer_state: format!("{:?}", self.sequencer_state.status().await),
```

See `l1_committer.rs:1401` and `l1_watcher.rs:410` for the correct pattern.

crates/l2/sequencer/state_updater.rs
```diff
@@ -297,6 +310,19 @@ impl StateUpdater {
...
+    async fn health(&self) -> CallResponse<Self> {
+        let eth_client_healthcheck = self.eth_client.test_urls().await;
+
+        CallResponse::Reply(OutMessage::Health(StateUpdaterHealth {
+            eth_client_healthcheck,
+            on_chain_proposer_address: self.on_chain_proposer_address,
+            sequencer_registry_address: self.sequencer_registry_address,
+            sequencer_address: self.sequencer_address,
+            check_interval_ms: self.check_interval_ms,
+            sequencer_state: format!("{:?}", self.sequencer_state.status()),
```
ElFantasma, line 323: Same bug as in `block_fetcher.rs`: missing `.await`.

```suggestion
            sequencer_state: format!("{:?}", self.sequencer_state.status().await),
```

ElFantasma (changes requested):
The rebase looks clean — all previously-merged StateUpdater structural changes are gone, and only the new health functionality remains. The pattern matches the existing health implementations (L1Committer, L1Watcher, etc.) correctly.

One bug in both health methods: `self.sequencer_state.status()` is missing `.await` — see inline comments.

crates/l2/based/block_fetcher.rs
FredPhilipy: SequencerState::status() is a synchronous method (uses std::sync::Mutex, not tokio::Mutex), so it returns SequencerStatus directly, not a Future. Adding .await causes a compile error: SequencerStatus is not a future.

This is consistent with all other health implementations in the codebase:

l1_committer.rs:1401 — no .await
l1_watcher.rs:410 — no .await
block_producer.rs:364 — no .await
l1_proof_sender.rs:498 — no .await
Also, within block_fetcher.rs itself, handle_cast at line 396 uses self.sequencer_state.status() without .await.

crates/l2/sequencer/state_updater.rs
FredPhilipy: Same as above — status() is sync, no .await needed.

crates/l2/based/block_fetcher.rs
ElFantasma: You are right. My comment is wrong

ElFantasma (approved):
This LGTM. Thanks for your contribution!

PR #4523, opened 2025-09-17
fix(l2): correct Current Block off-by-one, unify Current Batch as next

radik878:
The monitor displayed “Current Block” as latest + 1, which is off-by-one relative to the head semantics used across the codebase (Store::get_latest_block_number and RPC Latest). This change removes the increment so “Current Block” reflects the actual head.
Additionally, “Current Batch” in this widget represents the next batch to be committed by design (based mode spec), but was shown as “NaN” when the sequencer registry exists. We now consistently compute it as lastCommittedBatch() + 1 from L1 state already fetched above. This aligns the UI with the documented based-mode semantics and avoids misleading or placeholder values.


crates/l2/monitor/widget/chain_status.rs
```diff
@@ -119,14 +119,9 @@ impl GlobalChainStatusTable {
         let current_block = store
             .get_latest_block_number()
             .await
-            .map_err(|_| MonitorError::GetLatestBlock)?
-            + 1;
+            .map_err(|_| MonitorError::GetLatestBlock)?;
```
ElFantasma, line 122: nit: Now that the `+ 1` is removed and `current_batch` no longer has a branch, these two variables are each used only once — you could inline them directly into the `vec!` items below (e.g., `store.get_latest_block_number().await...?.to_string()` and `(last_committed_batch + 1).to_string()`). Minor though, fine either way.

ElFantasma (approved):

MegaRedHand:

PR #4525, opened 2025-09-17
feat(l2): add panic handling for exec and prove functions

### LINKED ISSUE #4452: Handle panics properly
The `exec` and `prove` functions can panic depending on the backend used for performing that run. We should catch the panics and handle them properly.

FredPhilipy:
Fixes #4452

Improves panic handling in the `exec` and `prove` functions across different backends to prevent crashes and provide meaningful error messages.

- Added panic catching with `catch_unwind` in replay functions
- Improved error messages for different backends (SP1, RISC0, Exec)
- Added proper error context for execution and proving failures
- Extracted common panic handling logic to helper functions

### DIFF
--- a/cmd/ethrex_replay/src/run.rs
+++ b/cmd/ethrex_replay/src/run.rs
@@ -37,8 +37,8 @@ pub async fn exec(backend: Backend, cache: Cache) -> eyre::Result<()> {
             let panic_msg = extract_panic_message(&panic_info);
 
             Err(eyre::Error::msg(format!(
-                "Execution panicked with backend {:?}: {}",
-                backend, panic_msg
+                "Execution panicked: {}",
+                panic_msg
             )))
         }
     }
@@ -64,19 +64,7 @@ pub async fn prove(backend: Backend, cache: Cache) -> eyre::Result<()> {
             // Try to extract meaningful error message from panic info
             let panic_msg = extract_panic_message(&panic_info);
 
-            // Provide more specific error messages based on different backends
-            let backend_name = match backend {
-                Backend::Exec => "Exec",
-                #[cfg(feature = "sp1")]
-                Backend::SP1 => "SP1",
-                #[cfg(feature = "risc0")]
-                Backend::RISC0 => "RISC0",
-            };
-
-            Err(eyre::Error::msg(format!(
-                "{} backend panicked while proving: {}",
-                backend_name, panic_msg
-            )))
+            Err(eyre::Error::msg(format!("Proving panicked: {}", panic_msg)))
         }
     }
 }


cmd/ethrex_replay/src/run.rs
```diff
@@ -22,9 +22,32 @@ pub async fn exec(backend: Backend, cache: Cache) -> eyre::Result<()> {
...
+        }
+        Err(panic_info) => {
+            // Try to extract meaningful error message from panic info
+            let panic_msg = if let Some(s) = panic_info.downcast_ref::<String>() {
+                s.clone()
+            } else if let Some(s) = panic_info.downcast_ref::<&str>() {
+                s.to_string()
+            } else {
+                format!("Unknown panic occurred during execution with backend: {:?}", backend)
+            };
```
tomip01: I think this is the function you created `extract_panic_message` but showing the `backend`.
Maybe you could modify that function to show that and use it here

cmd/ethrex_replay/src/run.rs
```diff
@@ -33,12 +56,41 @@ pub async fn prove(backend: Backend, cache: Cache) -> eyre::Result<()> {
...
+        }
+        Err(panic_info) => {
+            // Try to extract meaningful error message from panic info
+            let panic_msg = if let Some(s) = panic_info.downcast_ref::<String>() {
+                s.clone()
+            } else if let Some(s) = panic_info.downcast_ref::<&str>() {
+                s.to_string()
+            } else {
+                "Unknown panic occurred".to_string()
+            };
```
tomip01: same as above, you could use the function `extract_panic_message`

crates/l2/prover/src/backend/sp1.rs
```diff
@@ -131,3 +172,14 @@ fn to_calldata(proof: ProveOutput) -> ProofCalldata {
...
+/// Extract a meaningful error message from panic information.
+fn extract_panic_message(panic_info: &Box<dyn std::any::Any + Send>) -> String {
+    if let Some(s) = panic_info.downcast_ref::<String>() {
+        s.clone()
+    } else if let Some(s) = panic_info.downcast_ref::<&str>() {
+        s.to_string()
+    } else {
+        "Unknown panic occurred".to_string()
+    }
+}
```
tomip01: It is defined twice. Maybe you could move it to a common place

crates/l2/prover/src/lib.rs
```diff
@@ -50,3 +94,17 @@ pub fn to_batch_proof(
...
+/// and returns a readable error message.
+fn extract_panic_message(panic_info: &Box<dyn std::any::Any + Send>) -> String {
+    if let Some(s) = panic_info.downcast_ref::<String>() {
+        s.clone()
+    } else if let Some(s) = panic_info.downcast_ref::<&str>() {
+        s.to_string()
+    } else {
+        "Unknown panic occurred".to_string()
+    }
+}
```
tomip01: same

FredPhilipy:
> Looks good! I left some comments. I see you have lint failures. If you’d like to check them locally before pushing, you can run:
> 
> ```
> cargo clippy --workspace -- -D warnings
> cargo clippy -- -D warnings
> make lint
> ```

Awesome, thanks for the tips, the CI is good now. And solved all the comments

crates/l2/prover/src/lib.rs
```diff
@@ -14,27 +14,67 @@ pub async fn init_client(config: ProverConfig) {
...
+        Ok(exec_result) => exec_result,
+        Err(panic_info) => {
+            // Extract meaningful panic message
+            let panic_msg = extract_panic_message(&panic_info);
+
+            Err(Box::new(std::io::Error::other(format!(
+                "Backend {:?} panicked during execution: {}",
+                backend, panic_msg
+            ))))
+        }
```
tomip01: having the catch unwind in the `run.rs`, I think it is not necessary to have it again here and in the `sp1.rs` and `risc0.rs`

FredPhilipy: Make sense! thanks.

cmd/ethrex_replay/src/run.rs
```diff
@@ -33,12 +50,35 @@ pub async fn prove(backend: Backend, cache: Cache) -> eyre::Result<()> {
...
+            let panic_msg = extract_panic_message(&panic_info);
+
+            // Provide more specific error messages based on different backends
+            let backend_name = match backend {
+                Backend::Exec => "Exec",
+                #[cfg(feature = "sp1")]
+                Backend::SP1 => "SP1",
+                #[cfg(feature = "risc0")]
+                Backend::RISC0 => "RISC0",
+            };
```
ilitteri: `Backend` implements `Debug`, there's no need for this match, just print using `{:?}`.

ilitteri: Nevertheless, there's no value in printing which backend was used since this is already known by the user, and the block run report includes it.

ilitteri (approved):

PR #4529, opened 2025-09-17
fix(l1): remove volume for CL and EL clients in docker compose

ManuelBilbao:
**Motivation**

Changing the network without deleting the volumes cause an error as the database are not restarted.

**Description**

As this is mostly a dev setup (at least for now), removed the volumes. If the container is stopped but not removed, the data will not be lost.


ManuelBilbao:
Closing as we decided to take another approach

Issue #4494, opened 2025-09-15
Check potential blocking code


ElFantasma:
After some analysis we found that some of our tasks may be blocking the thread or taking too much time. We have to check them and refactor to avoid possible task blockers or concurrency issues:
 
- crates/networking/p2p/rlpx/connection/handshake.rs:
  `send_ack`: compress_pubkey and encode_ack_message (internally decompresses the recently compressed key)
  `send_auth`: compress_pubkey y encode_auth_message (same situation)
- crates/networking/p2p/rlpx/connection/server.rs:
  `handle_peer_message`: `GetReceipts` iterates a DB access for each block, `GetBytecodes` uses `process_byte_codes_request` (blocking) that does the same
- crates/networking/p2p/discv4/server.rs:
 `handle_find_node` just `await`s to get `kademlia`s lock and to send the UDP response; in between it performs a cuadratic search (with the lock held!) in `get_closest_nodes`
- Also in L2 we may have an issue in crates/networking/p2p/rlpx/l2/l2_connection.rs:
`should_process_new_block`, the `recover_address` call may take too long.

gianbelinche:
`crates/networking/p2p/rlpx/connection/handshake.rs:`

`send_ack` and `send_auth`: The keys being compressed and decompressed are not the same keys

MegaRedHand:
#4531 made the blocking code non-blocking by sending the processing logic to another thread. We should look for a better way to handle this.

ElFantasma:
To properly solve this, we should make CPU intensive functions async:
Iterating processes should shield between iterations, and searches should be and optimized to prevent long running operations.

MegaRedHand:
We should probably first check if each of the operations really blocks the runtime. `send_ack` and `send_auth` seem to only compress/decompress two pubkeys. This should take a really small, and more importantly, _bounded_ amount of time. In that case, the time it "blocks" the runtime is negligible, since any task switching would only add overhead.

gianbelinche:
Related to https://github.com/lambdaclass/ethrex/issues/4614

MegaRedHand:
I'll add that there are two different problems here:

1. code that is _blocking_ for the **tokio runtime** (i.e., runs for a long enough time **without yielding**)
2. and code that is _blocking_ for a **spawned `GenServer`** (i.e., runs for a long enough **total** time without returning from `handle_call`/`cast`)

PR #4531, opened 2025-09-17
fix(l1,l2): potential blocking code

gianbelinche:
**Motivation**
There are different parts of the code that could potentially block execution.

**Description**
Adds spawn blocking so that the execution is not blocked

Closes https://github.com/lambdaclass/ethrex/issues/4494

### DIFF
--- a/crates/networking/p2p/discv4/server.rs
+++ b/crates/networking/p2p/discv4/server.rs
@@ -452,10 +452,25 @@ impl DiscoveryServer {
             return;
         }
 
-        let neighbors = get_closest_nodes(node_id, table.clone());
+        let cloned_table = table.clone();
 
         drop(table);
 
+        let neighbors =
+            match tokio::task::spawn_blocking(move || get_closest_nodes(node_id, cloned_table))
+                .await
+            {
+                Ok(neighbors) => neighbors,
+                Err(err) => {
+                    debug!(
+                        received = "FindNode",
+                        to = %format!("{sender_public_key:#x}"),
+                        err = ?err,
+                        "Error getting closest nodes"
+                    );
+                    return;
+                }
+            };
         // we are sending the neighbors in 2 different messages to avoid exceeding the
         // maximum packet size
         for chunk in neighbors.chunks(8) {
--- a/crates/networking/p2p/rlpx/connection/server.rs
+++ b/crates/networking/p2p/rlpx/connection/server.rs
@@ -6,7 +6,7 @@ use std::{
 };
 
 use ethrex_blockchain::Blockchain;
-use ethrex_common::types::{MempoolTransaction, Transaction};
+use ethrex_common::types::{MempoolTransaction, Receipt, Transaction};
 use ethrex_storage::{Store, error::StoreError};
 use ethrex_trie::TrieError;
 use futures::{SinkExt as _, Stream, stream::SplitSink};
@@ -828,10 +828,19 @@ async fn handle_peer_message(state: &mut Established, message: Message) -> Resul
         }
         Message::GetReceipts(GetReceipts { id, block_hashes }) if peer_supports_eth => {
             if let Some(eth) = &state.negotiated_eth_capability {
-                let mut receipts = Vec::new();
-                for hash in block_hashes.iter() {
-                    receipts.push(state.storage.get_receipts_for_block(hash)?);
-                }
+                let cloned_state = state.clone();
+                let receipts =
+                    tokio::task::spawn_blocking(move || -> Result<Vec<Vec<Receipt>>, RLPxError> {
+                        let mut receipts = Vec::new();
+                        for hash in block_hashes.iter() {
+                            receipts.push(cloned_state.storage.get_receipts_for_block(hash)?);
+                        }
+                        Ok(receipts)
+                    })
+                    .await
+                    .map_err(|_| {
+                        RLPxError::InternalError("Get receipts for block task failed".to_string())
+                    })??;
                 let response = Receipts::new(id, receipts, eth)?;
                 send(state, Message::Receipts(response)).await?;
             }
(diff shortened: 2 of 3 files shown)


crates/networking/p2p/discv4/server.rs
```diff
@@ -452,10 +452,24 @@ impl DiscoveryServer {
             return;
         }
 
-        let neighbors = get_closest_nodes(node_id, table.clone());
+        let cloned_table = table.clone();
```
MegaRedHand, line 455: Should we move this also to the `spawn_blocking`? And does it make sense to clone it first, or just calling `get_closest_nodes` is enough?

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -841,10 +841,21 @@ async fn handle_peer_message(state: &mut Established, message: Message) -> Resul
         }
         Message::GetReceipts(GetReceipts { id, block_hashes }) if peer_supports_eth => {
             if let Some(eth) = &state.negotiated_eth_capability {
-                let mut receipts = Vec::new();
-                for hash in block_hashes.iter() {
-                    receipts.push(state.storage.get_receipts_for_block(hash)?);
-                }
+                let cloned_state = state.clone();
```
MegaRedHand: Can we clone `state.storage` here instead?

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -917,7 +928,14 @@ async fn handle_peer_message(state: &mut Established, message: Message) -> Resul
             send(state, Message::StorageRanges(response)).await?
         }
         Message::GetByteCodes(req) => {
-            let response = process_byte_codes_request(req, state.storage.clone())?;
+            let state_clone = state.clone();
```
MegaRedHand: Same here. We want to avoid copying the whole state.

crates/networking/p2p/rlpx/l2/l2_connection.rs
```diff
@@ -274,13 +274,19 @@ async fn should_process_new_block(
...
+        recover_address(msg_signature, block_hash).map_err(|e| {
+            log_peer_error(
+                &established_node,
+                &format!("Failed to recover lead sequencer: {e}"),
+            );
+            RLPxError::CryptographyError(e.to_string())
+        })
+    })
+    .await
+    .map_err(|_| RLPxError::InternalError("Recover Address task failed".to_string()))??;
```
MegaRedHand: If we're copying the node because of `log_peer_error`, we could try moving that outside the `spawn_blocking`:


```suggestion
    let recovered_lead_sequencer = tokio::task::spawn_blocking(move || recover_address(msg_signature, block_hash))
    .await
    .map_err(|_| RLPxError::InternalError("Recover Address task failed".to_string()))?
    .map_err(|e| {
        log_peer_error(
            &established_node,
            &format!("Failed to recover lead sequencer: {e}"),
        );
        RLPxError::CryptographyError(e.to_string())
    })?;
```

crates/networking/p2p/discv4/server.rs
gianbelinche: No, since we want to drop the table before calling `get_closest_nodes`.
Also we have to clone it before due to borrowing

crates/networking/p2p/rlpx/l2/l2_connection.rs
gianbelinche: Done

crates/networking/p2p/rlpx/connection/server.rs

crates/networking/p2p/rlpx/connection/server.rs

MegaRedHand (approved):

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -841,10 +841,21 @@ async fn handle_peer_message(state: &mut Established, message: Message) -> Resul
...
-                let mut receipts = Vec::new();
-                for hash in block_hashes.iter() {
-                    receipts.push(state.storage.get_receipts_for_block(hash)?);
-                }
+                let cloned_storage = state.storage.clone();
+                let receipts =
+                    tokio::task::spawn_blocking(move || -> Result<Vec<Vec<Receipt>>, RLPxError> {
+                        let mut receipts = Vec::new();
+                        for hash in block_hashes.iter() {
+                            receipts.push(cloned_storage.get_receipts_for_block(hash)?);
```
mpaulucci: shouldn't `get_receipts_for_block` be async instead of having to add `spawn_blocking` in random places?

crates/storage/store.rs
```diff
@@ -1082,7 +1082,7 @@ impl Store {
         Ok(nodes)
     }
 
-    pub fn get_receipts_for_block(
+    pub async fn get_receipts_for_block(
```
MegaRedHand: Marking it as `async` doesn't make it really async. We'd need to wrap this with `spawn_blocking` to do that.

gianbelinche: Reverted to spawn blocking

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -917,7 +928,14 @@ async fn handle_peer_message(state: &mut Established, message: Message) -> Resul
...
         Message::GetByteCodes(req) => {
-            let response = process_byte_codes_request(req, state.storage.clone())?;
+            let storage_clone = state.storage.clone();
+            let response = tokio::task::spawn_blocking(move || {
+                process_byte_codes_request(req, storage_clone.clone())
+            })
+            .await
+            .map_err(|_| {
+                RLPxError::InternalError("Failed to execute bytecode retrieval task".to_string())
+            })??;
```
tomip01: You can avoid the clone of the `storage_clone`
```suggestion
            let response =
                tokio::task::spawn_blocking(move || process_byte_codes_request(req, storage_clone))
                    .await
                    .map_err(|_| {
                        RLPxError::InternalError(
                            "Failed to execute bytecode retrieval task".to_string(),
                        )
                    })??;
```

gianbelinche:
Opened an [issue](https://github.com/lambdaclass/ethrex/issues/4614) for the get receipts for block function

PR #4536, opened 2025-09-17
feat(l2): pay base fee to coinbase

MegaRedHand:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/vm/levm/src/hooks/l2_hook.rs
+++ b/crates/vm/levm/src/hooks/l2_hook.rs
@@ -107,7 +107,19 @@ impl Hook for L2Hook {
         ctx_result: &mut ContextResult,
     ) -> Result<(), crate::errors::VMError> {
         if !vm.env.is_privileged {
-            return DefaultHook.finalize_execution(vm, ctx_result);
+            if !ctx_result.is_success() {
+                default_hook::undo_value_transfer(vm)?;
+            }
+
+            let gas_refunded: u64 = default_hook::compute_gas_refunded(vm, ctx_result)?;
+            let actual_gas_used =
+                default_hook::compute_actual_gas_used(vm, gas_refunded, ctx_result.gas_used)?;
+            default_hook::refund_sender(vm, ctx_result, gas_refunded, actual_gas_used)?;
+
+            // Different from L1, the base fee is not burned
+            pay_coinbase_with_base_fee(vm, actual_gas_used)?;
+
+            default_hook::delete_self_destruct_accounts(vm)?;
         }
 
         if !ctx_result.is_success() && vm.env.origin != COMMON_BRIDGE_L2_ADDRESS {
@@ -120,4 +132,17 @@ impl Hook for L2Hook {
 
         Ok(())
     }
+}
+
+pub fn pay_coinbase_with_base_fee(
+    vm: &mut VM<'_>,
+    gas_to_pay: u64,
+) -> Result<(), crate::errors::VMError> {
+    let coinbase_fee = U256::from(gas_to_pay)
+        .checked_mul(vm.env.gas_price)
+        .ok_or(InternalError::Overflow)?;
+
+    vm.increase_account_balance(vm.env.coinbase, coinbase_fee)?;
+
+    Ok(())
 }


crates/l2/tests/tests.rs
```diff
@@ -1474,21 +1499,19 @@ async fn test_n_withdraws(
...
-            .recoverable_fees;
+        base_fees += get_fees_details_l2(&receipt, l2_client).await.priority_fees;
     }
 
     assert_eq!(
-        fee_vault_balance_after_withdrawal,
-        fee_vault_balance_before_withdrawal + recoverable_fees,
-        "Fee vault balance didn't increase as expected after withdrawal"
+        coinbase_balance_after_withdrawal,
+        coinbase_balance_before_withdrawal + base_fees,
```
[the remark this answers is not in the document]
avilagaston9: Done [df483ca](https://github.com/lambdaclass/ethrex/pull/4536/commits/df483ca47b8eee316c6363cc8ed76d3d9192ecf3)!

crates/blockchain/blockchain.rs
```diff
@@ -120,6 +120,7 @@ impl Blockchain {
     async fn execute_block(
         &self,
         block: &Block,
+        fee_vault: Option<Address>,
```
MegaRedHand: Can we move this inside the `BlockChain` struct?

crates/blockchain/blockchain.rs
```diff
@@ -136,7 +137,7 @@ impl Blockchain {
         let vm_db = StoreVmDatabase::new(self.storage.clone(), block.header.parent_hash);
         let mut vm = self.new_evm(vm_db)?;
 
-        let execution_result = vm.execute_block(block)?;
+        let execution_result = vm.execute_block(block, fee_vault)?;
```
MegaRedHand: Then we can pass this parameter in `BlockChain::new_evm` instead, which makes more sense.

crates/networking/rpc/rpc.rs
```diff
@@ -205,6 +207,7 @@ pub async fn start_api(
     client_version: String,
     log_filter_handler: Option<reload::Handle<EnvFilter, Registry>>,
     gas_ceil: Option<u64>,
+    fee_vault: Option<Address>,
```
MegaRedHand: It's weird to have this in L1 RPC code.

MegaRedHand:
I think we can add the fee vault to the `BlockChain` struct, and make it pass it onto the VM whenever it creates one. The VM will pass the fee vault address to the L2 hook when it creates it, which can handle the basefee transfer.

crates/blockchain/blockchain.rs
```diff
@@ -73,6 +73,8 @@ pub struct Blockchain {
     /// Mapping from a payload id to either a complete payload or a payload build task
     /// We need to keep completed payloads around in case consensus requests them twice
     pub payloads: Arc<TokioMutex<Vec<(u64, PayloadOrTask)>>>,
+    /// If set, collected base fees will be sent to this address instead of being burned
```
MegaRedHand:
```suggestion
    /// (L2 only) If set, collected base fees will be sent to this address instead of being burned
```

cmd/ethrex_replay/src/cli.rs
```diff
@@ -624,7 +650,12 @@ async fn replay_block(block_opts: BlockOptions) -> eyre::Result<()> {
 
     let (eth_client, network) = setup(&opts).await?;
 
-    let cache = get_blockdata(eth_client, network.clone(), or_latest(block)?).await?;
+    #[cfg(not(feature = "l2"))]
+    let fee_vault = None;
+    #[cfg(feature = "l2")]
+    let fee_vault = get_fee_vault_address(&eth_client).await?;
```
MegaRedHand: We should probably change this to be a runtime flag, but using the `cfg!` macro instead of the attribute is a good first step. That way, the code is still checked, and unused code is removed automatically by the compiler, reducing "unused code" warnings.

```suggestion
    if cfg!(feature = "l2") {
        let fee_vault = get_fee_vault_address(&eth_client).await?;
    } else {
        let fee_vault = None;
    }
```

crates/blockchain/payload.rs
```diff
@@ -218,6 +218,7 @@ impl PayloadBuildContext {
         payload: Block,
         storage: &Store,
         blockchain_type: BlockchainType,
+        fee_vault: Option<Address>,
```
MegaRedHand: We should probably split this into `new_for_l1` and `new_for_l2`, just to keep interfaces clean:

```rust
fn new(...) { /* same as now */ }
pub fn new_for_l1(...) { new(...) }
pub fn new_for_l2(...) { new(...) }
```

We could also remove both `blockchain_type` and `fee_vault` entirely, creating the VM in the wrappers and passing it on to `new` directly.

crates/vm/levm/src/environment.rs
```diff
@@ -21,8 +21,10 @@ pub struct Environment {
     pub gas_limit: u64,
     pub config: EVMConfig,
     pub block_number: U256,
-    /// Coinbase is the block's beneficiary - the address that receives the block rewards and fees.
+    /// Coinbase is the block's beneficiary - the address that receives the block rewards (priority fees).
     pub coinbase: Address,
+    /// Fee vault is the address that receives the base fees in L2.
+    pub fee_vault: Option<Address>,
```
MegaRedHand: Can we move this to the L2 hook? Right now it's an empty struct, but it should be possible to put options inside it

tooling/ef_tests/state_v2/src/modules/block_runner.rs
```diff
@@ -144,7 +144,7 @@ pub async fn run_test(test: &Test, test_case: &TestCase) -> Result<(), RunnerErr
 
     // 3. Create Blockchain and add block.
 
-    let blockchain = Blockchain::new(store.clone(), BlockchainType::L1, false);
+    let blockchain = Blockchain::new(store.clone(), BlockchainType::L1, false, None);
```
MegaRedHand: As a heads-up, #4639 added a `BlockchainOptions` struct with these parameters, so you might need to put the `fee_vault` inside that.

fixtures/keys/private_keys.txt
```diff
@@ -12,7 +12,6 @@
 0x556a4c63c314bf2a1734f34bcc2c45941b41afe3e4827d205092f3cc8d3ab11b
 0x658edfde371c42f7dbdafdb73b9d28b87fff929753f74eb7c9888c2f25b02ea6
 0xabea088cc1a18dda221851da62d00bdc3724368e47ae0a2516727871554859bd
-0xe9ea73e0ca433882aa9d4e2311ecc4e17286121e6bd8e600e5d25d4243b2baa3
```
MegaRedHand, line 15: Why this change?

docs/CLI.md
```diff
@@ -122,32 +122,34 @@ Commands:
...
+          
           [env: ETHREX_NETWORK=]
 
       --datadir <DATABASE_DIRECTORY>
           If the datadir is the word `memory`, ethrex will use the `InMemory Engine`.
-
+          
           [env: ETHREX_DATADIR=]
-          [default: /home/runner/.local/share/ethrex]
+          [default: "/Users/gaston/Library/Application Support/ethrex"]
```
MegaRedHand: Let's keep this part as it was 😆

docs/l2/running.md
```diff
@@ -25,12 +25,17 @@ First we need to set some environment variables.
...
 	--block-producer.coinbase-address <l2-coinbase-address> \
+	--block-producer.fee-vault-address <l2-fee-vault-address>
 ```
 
 For further configuration take a look at the [CLI document](../CLI.md#ethrex-l2)
 
 This will start an ethrex l2 sequencer with the RPC server listening at `http://localhost:1729` and the proof coordinator server listening at `http://localhost:3900`
 
+> **Note:** If `--block-producer.fee-vault-address` is set, the sequencer will send collected base fees to that address instead of burning them.  
+> Be cautious: if the fee vault address is the same as the coinbase address, the coinbase balance will change in a way that differs from the standard L1 behavior, which may break assumptions about EVM compatibility.
```
MegaRedHand: Nit: the phrasing sounds kind of weird

```suggestion
> ⚠️ Caution: if the fee vault and coinbase addresses are the same, its balance will change in a way that differs from the standard L1 behavior, which may break assumptions about EVM compatibility.
```

Also, we could use a [caution blockquote here](https://github.com/orgs/community/discussions/16925)

cmd/ethrex_replay/src/cli.rs
avilagaston9: Changed to:
```Rust
let fee_vault = if cfg!(feature = "l2") {
   get_fee_vault_address(&eth_client).await?
} else {
    None
};
```
To access the variable from outside the if/else scope. [6445c83](https://github.com/lambdaclass/ethrex/pull/4536/commits/6445c832096dbb72e8fe9e85542aa5399b597577)

crates/blockchain/blockchain.rs
avilagaston9: Done [6445c83](https://github.com/lambdaclass/ethrex/pull/4536/commits/6445c832096dbb72e8fe9e85542aa5399b597577)!

crates/blockchain/payload.rs
avilagaston9: Done [9ae2402](https://github.com/lambdaclass/ethrex/pull/4536/commits/9ae24023687c8a3e40f589b88fb2a2b3bf4754ba)!

crates/vm/levm/src/environment.rs
avilagaston9: Done [65da300](https://github.com/lambdaclass/ethrex/pull/4536/commits/65da300c2fa25c3c59fb0f07151e5d7b9afc44f4)!

fixtures/keys/private_keys.txt
avilagaston9: I took the key as the default for the fee vault 😄. If the key is used for other purposes while the integration tests are running, they will fail.

docs/l2/running.md
avilagaston9: Improved [here](https://github.com/lambdaclass/ethrex/pull/4536/commits/ef38d6907ea9b70910952e93d9e2fed629adbcb9)!

docs/CLI.md
avilagaston9: 😅 Done!

crates/blockchain/payload.rs
```diff
@@ -217,11 +217,23 @@ pub struct PayloadBuildContext {
...
 
 impl PayloadBuildContext {
-    pub fn new(
+    pub fn new_for_l1(payload: Block, storage: &Store) -> Result<Self, EvmError> {
+        let vm_db = StoreVmDatabase::new(storage.clone(), payload.header.parent_hash);
+        let vm = Evm::new_for_l1(vm_db);
+        Self::new(payload, storage, vm)
+    }
+
+    pub fn new_for_l2(
```
mpaulucci: Its quite ugly having to define `new_for_l1` and `new_for_l2`, seems like this is screaming for an abstraction/interface

avilagaston9: Reused `BlockchainType` and `VMType` so that the `L2` variant holds the `FeeConfig` struct. This avoids having to pass an extra `None` parameter to each function when running in `L1` mode.

crates/l2/prover/src/guest_program/src/input.rs
```diff
@@ -16,6 +18,8 @@ pub struct ProgramInput {
     pub execution_witness: ExecutionWitness,
     /// value used to calculate base fee
     pub elasticity_multiplier: u64,
+    /// Configuration for L2 fees
+    pub fee_config: Option<FeeConfig>,
```
tomip01, line 22: Why not hide it under the `l2` flag?

docs/l2/running.md
```diff
@@ -0,0 +1,69 @@
...
+#### Run the sequencer
+
+```sh
+    ethrex l2 \
+	--network <path-to-your-genesis.json> \
+	--on_chain_proposer_address <address> \
+	--bridge_address <address> \
+	--rpc_url <rpc-url> \
+	--committer_l1_private_key <private-key> \
+	--proof_coordinator_l1_private_key \
```
tomip01:
```suggestion
	--proof_coordinator_l1_private_key <private-key> \
```

docs/l2/running.md
```diff
@@ -0,0 +1,69 @@
...
+After starting the sequencer and prover, you can verify that your L2 node is running correctly:
+
+- **Check the sequencer RPC:**
+
+  You can request the latest block number:
+
+  ```sh
+  curl http://localhost:1729 \
+  	-H 'content-type: application/json' \
+  	-d '{"jsonrpc":"2.0","method":"eth_blockNumber","id":"1","params":[]}'
```
tomip01: Maybe here we can show the `health` endpoint:
`curl http://localhost:5555/health | jq`

I use `jq` to see it more clear

crates/l2/tests/tests.rs
```diff
@@ -1592,10 +1622,17 @@ async fn test_total_eth_l2(
...
-        "Total ETH on L2 ({total_eth_on_l2}) is greater than bridge locked ETH ({bridge_locked_eth})"
-    );
+    if std::env::var("INTEGRATION_TEST_SKIP_FEE_VAULT_CHECK").is_err() {
+        assert!(
+            total_eth_on_l2 == bridge_locked_eth,
+            "Total ETH on L2 ({total_eth_on_l2}) differs from bridge locked ETH ({bridge_locked_eth})"
+        );
+    } else {
+        assert!(
+            total_eth_on_l2 < bridge_locked_eth,
```
tomip01: Shouldn't here be `<=`?

```suggestion
            total_eth_on_l2 <= bridge_locked_eth,
```

docs/l2/running.md
avilagaston9: The file was removed in main. Updated docs [0075f72](https://github.com/lambdaclass/ethrex/pull/4536/commits/0075f72c3f70523b3f63ea725102d75e4d06e641).

docs/l2/running.md
avilagaston9: The file was removed in main. Updated docs [0075f72](https://github.com/lambdaclass/ethrex/pull/4536/commits/0075f72c3f70523b3f63ea725102d75e4d06e641).

crates/l2/tests/tests.rs
avilagaston9: On each transaction, `base_fee` is burned. So `total_eth_on_l2` must always be less than `bridge_locked_eth`

crates/l2/prover/src/guest_program/src/input.rs
avilagaston9: I chose to use `Option` instead of a feature flag to avoid adding a parameter to `execute_stateless` under the feature flag as well. Just wanted to avoid increasing the use of the feature flag. But I’m open to changing it if the feature flag is better.

tomip01: I would keep it as it is! I didn't remembered that it was passed to the `execute_stateless`

crates/common/types/l2/fee_config.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use crate::rkyv_utils::{H160Wrapper, OptionH160Wrapper};
+
+#[derive(
+    Serialize, Deserialize, RDeserialize, RSerialize, Archive, Clone, Copy, Debug, Default,
+)]
+pub struct FeeConfig {
+    /// If set, the base fee is sent to this address instead of being burned.
+    #[rkyv(with=OptionH160Wrapper)]
+    pub fee_vault: Option<Address>,
```
ManuelBilbao, line 13: Would name this `base_fee_vault` to be clearer

ManuelBilbao: Also would add the `base_` prefix in CLI and other places. This would avoid confussion about what fee is it talked about

crates/blockchain/blockchain.rs
```diff
@@ -48,7 +49,7 @@ use ethrex_metrics::metrics_blocks::METRICS_BLOCKS;
 use ethrex_common::types::BlobsBundle;
 
 const MAX_PAYLOADS: usize = 10;
-const MAX_MEMPOOL_SIZE_DEFAULT: usize = 10_000;
+pub const MAX_MEMPOOL_SIZE_DEFAULT: usize = 10_000;
```
ManuelBilbao: Why this?

crates/l2/sequencer/block_producer.rs
```diff
@@ -83,9 +83,19 @@ impl BlockProducer {
...
             block_gas_limit,
         } = config;
+
+        if let Some(fee_vault) = fee_vault_address {
+            if fee_vault == coinbase_address {
+                warn!(
+                    "The coinbase address and fee vault address are the same. Coinbase balance behavior will be affected.",
+                );
+            }
+        }
```
ManuelBilbao: This will trigger clippy in Rust 1.90
```suggestion
        if fee_vault_address.is_some_and(|fee_vault| fee_vault == coinbase_address) {
            warn!(
                "The coinbase address and fee vault address are the same. Coinbase balance behavior will be affected.",
            );
        }
```

crates/l2/tests/tests.rs
```diff
@@ -79,6 +81,12 @@ const DEFAULT_PROPOSER_COINBASE_ADDRESS: Address = H160([
     0xad, 0x62, 0x0c, 0x8d,
 ]);
 
+// 0x000c0d6b7c4516a5b274c51ea331a9410fe69127
+const DEFAULT_PROPOSER_FEE_VAULT_ADDRESS: Address = H160([
```
ManuelBilbao, line 91: Do we have the PK for this? I'd put it as well

docs/getting-started/quickstart-l2.md
ManuelBilbao, line 1: This file doesn't exist anymore. Can be deleted

fixtures/genesis/l2.json
ManuelBilbao, line 1: I think this is just a reorder of items. If nothing really changed, I'd avoid it

crates/common/types/l2/fee_config.rs
avilagaston9: Yes. At the time I made this PR, it was the only fee vault we had. But in #4725 I renamed it to `base fee vault` everywhere, including the CLI.

crates/blockchain/blockchain.rs
avilagaston9: Oh, I needed it for `ethrex-replay` but it was removed now. Restored here [ca0ce92](https://github.com/lambdaclass/ethrex/pull/4536/commits/ca0ce929533fe6ed877c7b0161769bf76e746d17)!

crates/l2/sequencer/block_producer.rs
avilagaston9: Done [7172748](https://github.com/lambdaclass/ethrex/pull/4536/commits/717274827cad6dcb4e567199a136da4fc9d2b970)!

docs/getting-started/quickstart-l2.md
avilagaston9: Done!

fixtures/genesis/l2.json
avilagaston9: Done [652fa08](https://github.com/lambdaclass/ethrex/pull/4536/commits/652fa08d1f72b2225c2550a4cb4a24cac9a8a93d)!

crates/l2/tests/tests.rs
avilagaston9: Done [4eafeb4](https://github.com/lambdaclass/ethrex/pull/4536/commits/4eafeb44b102770f34871b2c81c6eb58c41f0b04)!

PR #4537, opened 2025-09-17
fix(l1): request storage peer leak

fedacking:
**Motivation**

Whenever a step of snap sync is completed, the peers should be marked as free. When requesting storage ranges, the function was ending without clearing the peers if the block was stale.

**Description**

- Changes the request_storage_ranges function to wait until all spawned tasks end and frees those peers before continuing.
- Adds utility function that frees the peers when they're supposed to be free and logs an error if it finds peers in that inconsistent state.

This pr shouldn't be merged before thorough testing on Sepolia!

### DIFF
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -1595,6 +1595,16 @@ impl PeerHandler {
                 .remove(&account_done);
         }
 
+        while completed_tasks < task_count {
+            match task_receiver.recv().await {
+                None => break,
+                Some(result) => {
+                    completed_tasks += 1;
+                    self.peer_table.free_peer(result.peer_id).await;
+                }
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


fedacking:
This has been tested in sepolia, ready for review.

jrchatruc (approved):

crates/networking/p2p/peer_handler.rs
```diff
@@ -1601,6 +1601,18 @@ impl PeerHandler {
                 .remove(&account_done);
         }
 
+        // Dropping the task sender so that the recv returns None
+        drop(task_sender);
+
+        loop {
+            match task_receiver.recv().await {
+                None => break,
```
iovoid: Could be a while-let

fedacking: Done

mpaulucci (approved):

PR #4543, opened 2025-09-17
refactor(l1): use `Path` instead of `String` for `--datadir`

MegaRedHand:
**Motivation**

We are currently using `String` when dealing with file paths. This pushes error-handling later into the node's lifecycle, which could delay some errors.

**Description**

This PR changes the CLI flag `--datadir` to receive a `PathBuf`, and also changes all uses of the flag to receive a `&Path` or similar type.


crates/networking/p2p/sync.rs
```diff
@@ -120,9 +120,9 @@ impl Syncer {
             // This won't be used
             cancel_token: CancellationToken::new(),
             blockchain: Arc::new(Blockchain::default_with_store(
-                Store::new("", EngineType::InMemory).expect("Failed to start Sotre Engine"),
+                Store::new("", EngineType::InMemory).expect("Failed to start Store Engine"),
```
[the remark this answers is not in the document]
MegaRedHand: ???????????????????????????????????????

cmd/ethrex/l2/command.rs
```diff
@@ -110,8 +110,8 @@ pub enum Command {
     },
     #[command(name = "removedb", about = "Remove the database", visible_aliases = ["rm", "clean"])]
     RemoveDB {
-        #[arg(long = "datadir", value_name = "DATABASE_DIRECTORY", default_value_t = default_datadir(), required = false)]
-        datadir: String,
+        #[arg(long = "datadir", value_name = "DATABASE_DIRECTORY", default_value = default_datadir().into_os_string(), required = false)]
```
mpaulucci, line 113: have you verified this is was is printed as defatul dir when running `ethrex --help`

cmd/ethrex/initializers.rs
```diff
@@ -140,14 +139,15 @@ pub async fn init_rpc_api(
     log_filter_handler: Option<reload::Handle<EnvFilter, Registry>>,
     gas_ceil: Option<u64>,
 ) {
+    init_datadir(&opts.datadir);
```
mpaulucci, line 142: why was this necesary?

MegaRedHand: `init_datadir` doesn't modify its argument anymore, so I made it return nothing. There were many uses that binded the return value into a variable or just passed it to a function directly, like this one.

cmd/ethrex/l2/command.rs
MegaRedHand: No, but it's printing OK. Thanks for the reminder!

```
          
          [env: ETHREX_DATADIR=]
          [default: "/Users/mega/Library/Application Support/ethrex"]
```

cmd/ethrex/initializers.rs
```diff
@@ -76,8 +76,8 @@ pub fn init_metrics(opts: &Options, tracker: TaskTracker) {
 }
 
 /// Opens a new or pre-existing Store and loads the initial state provided by the network
-pub async fn init_store(data_dir: &str, genesis: Genesis) -> Store {
-    let store = open_store(data_dir);
+pub async fn init_store(datadir: impl AsRef<Path>, genesis: Genesis) -> Store {
+    let store = open_store(datadir.as_ref());
```
gianbelinche, line 80: Why is datadir `impl AsRef<Path>` here but `&Path` in `load_store`?

MegaRedHand: `impl AsRef<Path>` is really flexible, allowing you to use `&str`, `String`, `&Path`, `PathBuf`, `&PathBuf`, etc. as a parameter. The downside is that monomorphization might increase binary size, so I tried to keep usage of it to a minimum.

I'm using it here to simplify tests, which pass a `&str` to it. Otherwise, we'd need to add a `&PathBuf::from("<path>")` on each call to it.

gianbelinche (approved):

PR #4544, opened 2025-09-17
feat(levm): ecrecover multiple versions

pablodeymo:
**Motivation**

We want to use secp256k crate original version in case we call `ecrecover` without SP1.

**Description**

This PR aims to call the right version of `ecrecover` depending on the enviroment.

Closes #issue_number


crates/vm/levm/src/precompiles.rs
```diff
@@ -295,6 +293,75 @@ pub(crate) fn fill_with_zeros(calldata: &Bytes, target_len: usize) -> Bytes {
     padded_calldata.into()
 }
 
+pub fn ecrecover(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<Bytes, VMError> {
+    // Use the ethrex implementation of ecrecover
+    // Switch to the sp1 implementation of ecrecover when we are in SP1 environment
+    
+    // TODO: change this when we have a way to detect SP1 environment
+    if cfg!(feature = "sp1") {
```
LeanSerra: I'm not 100% sure this will work in ethrex but on sp1 code they use `#[cfg(target_os = "zkvm")]`

crates/common/crypto/kzg.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn from(value: openvm_kzg::KzgError) -> Self {
+        KzgError::OpenvmKzg(value)
+    }
+}
+
+/// Verifies a KZG proof for blob committed data, using a Fiat-Shamir protocol
+/// as defined by c-kzg-4844.
+///
+/// Dispatches one of the enabled implementations following the hierarchy:
+/// c-kzg > kzg-rs
```
xqft:
```suggestion
/// c-kzg > kzg-rs > openvm-kzg
```

crates/common/crypto/kzg.rs
```diff
@@ -0,0 +1,172 @@
...
+///
+/// Different implementations exist for different targets:
+/// - Host (any, usually c-kzg as it's more performant)
+/// - SP1 Guest (kzg-rs)
+/// - Risc0 Guest (c-kzg patched)
```
xqft, line 147: would add a line for openvm here

crates/common/crypto/kzg.rs
```diff
@@ -0,0 +1,172 @@
...
+        )
+        .map_err(KzgError::from)
+    }
+    #[cfg(feature = "openvm-kzg")]
+    {
+        unimplemented!("There's no implementation of blob verification for openvm-kzg yet.");
+    }
+}
+
+/// Verifies that p(z) = y given a commitment that corresponds to the polynomial p(x) and a KZG proof
```
xqft, line 227: would add here the same description as in `verify_blob_kzg_proof`

crates/l2/common/src/utils.rs
```diff
@@ -18,3 +18,28 @@ pub fn get_address_from_secret_key(secret_key: &SecretKey) -> Result<Address, St
 
     Ok(Address::from(address_bytes))
 }
+
+#[cfg(feature = "k256")]
+pub fn get_address_from_secret_key(
+    secret_key: &k256::elliptic_curve::SecretKey<k256::Secp256k1>,
+) -> Result<Address, String> {
```
xqft: can we make something similar to the KZG and ecrover impl. so that the features are not mutually exclusive? (because of declaring 2 functions with the same name if both flags are enabled)

in some make rules and CI jobs we lint with `--all-features` so it would be best to allow for both to be set at the same time

crates/vm/levm/src/precompiles.rs
pablodeymo: changed

crates/common/crypto/kzg.rs
pablodeymo: done

crates/common/crypto/kzg.rs
pablodeymo: ok

crates/common/crypto/kzg.rs
pablodeymo: ok

crates/l2/common/src/utils.rs
pablodeymo: fixed

crates/l2/common/src/utils.rs
```diff
@@ -1,20 +1,46 @@
...
-use secp256k1::SecretKey;
 
-pub fn get_address_from_secret_key(secret_key: &SecretKey) -> Result<Address, String> {
-    let public_key = secret_key
-        .public_key(secp256k1::SECP256K1)
-        .serialize_uncompressed();
-    let hash = keccak(&public_key[1..]);
+/// secp256k1 > k256
+#[allow(clippy::needless_return)]
+pub fn get_address_from_secret_key(secret_key: &secp256k1::SecretKey) -> Result<Address, String> {
```
ilitteri: Shouldn't we enforce that one of the features is enabled?

crates/l2/Cargo.toml
```diff
@@ -15,7 +15,7 @@ tracing.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 ethereum-types.workspace = true
-ethrex-common.workspace = true
+ethrex-common = { workspace = true, features = ["c-kzg"] }
```
ilitteri: I think it'd be a good idea to clarify in a comment why `ethrex-l2` needs `c-kzg` instead of the other options.

crates/vm/levm/bench/revm_comparison/Cargo.toml
```diff
@@ -10,7 +10,7 @@ name = "revm_comparison"
 path = "src/lib.rs"
 
 [dependencies]
-ethrex-levm = { path = "../../" }
+ethrex-levm = { path = "../../", features = ["c-kzg"] }
```
ilitteri, line 14: Same here.

crates/l2/common/src/utils.rs
pablodeymo: done, thanks!

crates/l2/Cargo.toml
ilitteri: We should add the features `sp1` and `risc0`, and they should use `ethrex-common/kzg-rs` and `ethrex-common/c-kzg` respectively.

ilitteri: The default should be the one that the L1 uses (i.e. the most performant).

crates/l2/Cargo.toml
pablodeymo: fixed

crates/l2/Cargo.toml
pablodeymo: fixed

crates/vm/levm/bench/revm_comparison/Cargo.toml
pablodeymo: this will probably moved elsewhere

PR #4546, opened 2025-09-18
feat(replay): allow replay in range from some block to the latest

ilitteri:
**Motivation**

https://github.com/lambdaclass/ethrex/pull/4125

**Description**

- `--from` arg doesn't require `--to` anymore.
- If `--from` is passed alone, then the `--to` is set to the latest block.

### DIFF
--- a/cmd/ethrex_replay/src/cli.rs
+++ b/cmd/ethrex_replay/src/cli.rs
@@ -209,17 +209,7 @@ impl EthrexReplayCommand {
                     unimplemented!("cached mode is not implemented yet");
                 }
 
-                let to = if to.is_none() {
-                    let eth_client = EthClient::new(opts.rpc_url.as_str())?;
-
-                    let latest_block = eth_client.get_block_number().await?;
-
-                    Some(latest_block.as_u64())
-                } else {
-                    to
-                };
-
-                let blocks = resolve_blocks(blocks, from, to)?;
+                let blocks = resolve_blocks(blocks, from, to, opts.rpc_url.clone()).await?;
 
                 for (i, block_number) in blocks.iter().enumerate() {
                     info!(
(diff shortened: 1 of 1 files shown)


cmd/ethrex_replay/src/cli.rs
```diff
@@ -209,6 +209,16 @@ impl EthrexReplayCommand {
...
 
+                let to = if to.is_none() {
+                    let eth_client = EthClient::new(opts.rpc_url.as_str())?;
+
+                    let latest_block = eth_client.get_block_number().await?;
+
+                    Some(latest_block.as_u64())
+                } else {
+                    to
+                };
```
ManuelBilbao: In addition to Copilot's suggestion, you could simplify this with something like:
```suggestion
                let to = to.unwrap_or_else(|| fetch_latest_block_number(opts.rpc_url.as_str()).await?)
```
(Although I think it requires some extra syntax for async)

ManuelBilbao (commented):
In addition to Copilot's suggestion, you could do something like:

cmd/ethrex_replay/src/cli.rs
ilitteri: I'll try, but I think I can't put an async closure inside `unwrap_or_else `.

ilitteri: Nope, I can't use `unwrap_or_else`. I can use `unwrap_or` though, but it's eager. If we can afford this for the sake of the code's readability, it's fine for me. LMK what you think.

ilitteri: Check it out here [b690e5b](https://github.com/lambdaclass/ethrex/pull/4546/commits/b690e5b72acee15b3fce864ac7401901d255545d). I also made the code simpler by moving that logic to the blocks resolution function.

cmd/ethrex_replay/src/cli.rs
```diff
@@ -896,3 +894,9 @@ pub async fn produce_custom_l2_block(
 
     Ok(new_block)
 }
+
+async fn fetch_latest_block_number(rpc_url: Url) -> eyre::Result<u64> {
+    let eth_client = EthClient::new(rpc_url.as_str())?;
+    let latest_block = eth_client.get_block_number().await?;
+    Ok(latest_block.as_u64())
```
ilitteri:
```suggestion
    let eth_client = EthClient::new(rpc_url.as_str())?;

    let latest_block = eth_client.get_block_number().await?;

    Ok(latest_block.as_u64())
```

PR #4548, opened 2025-09-18
feat(replay): support public L2 networks blocks/batches replay

### LINKED ISSUE #4403: `ethrex-replay`: support replay for public L2 networks

tomip01:
**Motivation**

We want to run `ethrex_replay` on ethrex L2 chains. This contains the minimal changes to support them.

**Description**

- Add conditional use of crates and functions under the `l2` feature flag to solve lint problems.
- Add the required L2 fields when running under the `l2` feature flag.
- Add new network enum type `L2Chain`.
  - This is for replay only for now.
  - Contains the `chain_id` and the genesis is an empty one but with the `chain_id` setted.

**How to test**

Yo will need an L2 `RPC_URL` running ethrex

Execute a block:
```
cargo r -r -p ethrex-replay --features l2 -- l2 block --block <N> --execute --rpc-url <RPC_URL>
```

Execute a batch:
```
cargo r -r -p ethrex-replay --features l2 -- l2 batch --batch <N> --execute --rpc-url <RPC_URL>
```

Prove a block:
```
cargo r -r -p ethrex-replay --features l2 -- l2 block --block <N> --prove --rpc-url <RPC_URL>
```

Prove a batch:
```
cargo r -r -p ethrex-replay --features l2 -- l2 batch --batch <N> --prove --rpc-url <RPC_URL>
```

Closes #4403


cmd/ethrex_replay/src/cli.rs
```diff
@@ -885,12 +875,20 @@ pub async fn produce_custom_l2_block(
...
-        build_payload(blockchain.clone(), payload, store, rollup_store).await?;
+    // TODO: see if the None is correct here
+    let payload_build_result = build_payload(
+        blockchain.clone(),
+        payload,
+        store,
+        &mut None,
+        DEFAULT_BUILDER_GAS_CEIL,
+    )
+    .await?;
```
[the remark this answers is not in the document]
tomip01: Done [a13f0bb](https://github.com/lambdaclass/ethrex/pull/4548/commits/a13f0bb15e4ff242f32e51ee2a512d090284fef8)

cmd/ethrex_replay/src/fetcher.rs
```diff
@@ -143,7 +144,19 @@ pub async fn get_blockdata(
...
+    }
+    #[cfg(feature = "l2")]
+    {
+        Ok(Cache::new_for_l2(
+            vec![block],
+            witness_rpc,
+            network,
+            chain_config,
+        ))
+    }
```
ilitteri: Can't we unify this somehow? I think `network` can be derived from `chain_config.chain_id`.

I see this:

```rust
// Cache for L1
Cache::new(vec![block], witness_rpc, Some(network))

// Cache for L2
Cache:: new_for_l2(vec![block], witness_rpc, network, chain_config)

// Cache unification
Cache::new(vec![block], witness_rpc, Some(chain_config))
```

Then inside `Cache::new`, `network` is derived from `chain_config.chain_id` using `Network::try_from`.

crates/common/config/networks.rs
```diff
@@ -38,6 +38,7 @@ pub enum Network {
     PublicNetwork(PublicNetwork),
     LocalDevnet,
     LocalDevnetL2,
+    L2Chain(u64),
```
ManuelBilbao, line 41: I'm not very sure about adding this. As far as I have seen, `Network` is used for two thing in replay: to get the genesis->chain_config and the name for the cache file. The latest can be simply changed with the chain id (or it could be a mapping of common IDs to names if we want to).
About the first use, I think there's a lot of comings and goings from networks to genesis to chain_configs to chain_ids. I'd do a refactor here removing the need of a `Network` and being more specific on the type needed (mostly `ChainConfig` or chain ID I think)

ManuelBilbao: Note: I may be missing some other use case behind a feature flag, let me know if that's the case

ilitteri: Currently, deriving `Network` from the chain ID retrieved from the provided RPC URL is convenient for the L1 functionality, mostly for public networks where the user doesn't care about specific data needed for building the `ChainConfig`, they just want to replay some L1 block for some given RPC URL. This should also be the case for someone who wants to replay some block/batch from some L2.

I agree that we can find a workaround for the L2 features to avoid adding a variant to the Network enum. However, since we already have the `LocalDevnet` and `LocalDevnetL2` variants, I'm curious if there’s another reason, beyond simply adding this variant for ethrex-replay convenience (which is sufficient for me), not to include it.

tomip01: The idea behind that is that we tried to get the `Cache` as stable as possible. So for L1, just providing the name of the network will be sufficient to retrieve the `ChainConfig` and `Genesis` from our own repo. This grants us the ability to change fields from the `ChainConfig` and still not break the deserialization from old caches

tomip01: For L2 we may not have the chain config in the repo, so we made that an optional fields that is required only on L2

cmd/ethrex_replay/src/fetcher.rs
tomip01: Done [0bb1aa0](https://github.com/lambdaclass/ethrex/pull/4548/commits/0bb1aa03f84719496cf113f5a034be0b80b84ddd)

cmd/ethrex_replay/src/cache.rs
```diff
@@ -30,9 +32,7 @@ pub struct Cache {
     /// L1 network identifier.
     /// For L1 chains, this is used to retrieve the chain configuration from the repository.
     /// For L2 chains, the chain configuration is passed directly via `chain_config` instead.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[serde(flatten)]
-    pub network: Option<Network>,
+    pub network: Network,
```
ilitteri, line 36: We need to update the cache given this change.

tomip01: Done [72810eb](https://github.com/lambdaclass/ethrex/pull/4548/commits/72810ebf749e799514963dbf393a1d2927bb30d9)

PR #4558, opened 2025-09-18
fix(l1): fix snap sync edge case where we would not heal certain accounts' storages

jrchatruc:
**Motivation**

**Description**

Closes #issue_number


fedacking:
Changelog:
- Changes the "healed accounts" to include all healed accounts, regardless of if they have storage at this point.
- Made it so that if validate storage root fails the programs end.

fedacking (approved):

Issue #4271, opened 2025-09-02
Document differences between BackendMessage and PeerMessage


gianbelinche:
Document how BackendMessage and PeerMessage in `crates/networking/p2p/rlpx/connection/server.rs` work

PR #4560, opened 2025-09-18
docs(l1): document rplx messages

gianbelinche:
**Motivation**
The difference between `PeerMessage` and `BackendMessage` is not clear

**Description**
Document its differences

Closes https://github.com/lambdaclass/ethrex/issues/4271

### DIFF
--- a/crates/networking/p2p/rlpx/connection/server.rs
+++ b/crates/networking/p2p/rlpx/connection/server.rs
@@ -147,7 +147,9 @@ pub enum InnerState {
 #[derive(Clone, Debug)]
 #[allow(private_interfaces)]
 pub enum CastMessage {
+    // Received a message from the remote peer
     PeerMessage(Message),
+    // This node requests information from the remote peer
     BackendMessage(Message),
     SendPing,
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -147,7 +147,9 @@ pub enum InnerState {
 #[derive(Clone, Debug)]
 #[allow(private_interfaces)]
 pub enum CastMessage {
+    // Received a message from the remote peer
```
MegaRedHand: Can we turn this into a rustdoc comment?


```suggestion
    /// Received a message from the remote peer
```

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -147,7 +147,9 @@ pub enum InnerState {
...
     PeerMessage(Message),
+    // This node requests information from the remote peer
```
MegaRedHand: Same

crates/networking/p2p/rlpx/connection/server.rs

PR #4566, opened 2025-09-18
perf(levm): use a distinct opcode table per fork to avoid runtime checks

Oppen:
**Motivation**

**Description**

Closes #issue_number


Oppen:
There's a different implementation of this in main already.

PR #4569, opened 2025-09-18
feat(l1): make the database path-based

iovoid:
> [!WARNING]
> This PR is not ready to be merged as-is, and the design isn't final yet

**Motivation**

We want a path-based database to:
* Make reads faster: instead of going through the whole tree, we can skip straight to the value
* Reduce database size, because old nodes aren't duplicated

**Description**

Currently we are using layers of in-memory tries to handle reorgs.

### DIFF
--- a/crates/common/trie/nibbles.rs
+++ b/crates/common/trie/nibbles.rs
@@ -7,10 +7,14 @@ use ethrex_rlp::{
     structs::{Decoder, Encoder},
 };
 
+// TODO: move path-tracking logic somewhere else
+// PERF: try using a stack-allocated array
 /// Struct representing a list of nibbles (half-bytes)
 #[derive(Debug, Clone, Default)]
 pub struct Nibbles {
     data: Vec<u8>,
+    /// Parts of the path that have already been consumed (used for tracking
+    /// current position when visiting nodes). See `current()`.
     already_consumed: Vec<u8>,
 }
 


crates/common/trie/trie.rs
```diff
@@ -92,12 +92,13 @@ impl Trie {
 
     /// Retrieve an RLP-encoded value from the trie given its RLP-encoded path.
     pub fn get(&self, path: &PathRLP) -> Result<Option<ValueRLP>, TrieError> {
+        let path = Nibbles::from_bytes(path);
         Ok(match self.root {
-            NodeRef::Node(ref node, _) => node.get(self.db.as_ref(), Nibbles::from_bytes(path))?,
+            NodeRef::Node(ref node, _) => node.get(self.db.as_ref(), path)?,
             NodeRef::Hash(hash) if hash.is_valid() => {
-                Node::decode(&self.db.get(hash)?.ok_or(TrieError::InconsistentTree)?)
+                Node::decode(&self.db.get(Nibbles::default())?.ok_or(TrieError::InconsistentTree)?)
```
edg-l: Why `Nibbles::default()` here? Shouldn't it use hash? it is unused after the change besides checking if its valid.

edg-l:
Should NodeRef still store a Hash and not a path itself? Since its value used to be whats used to query the db, and now its changing to the path. I see it is now being "unused" in some match contexts

crates/common/trie/node.rs
```diff
@@ -51,23 +62,28 @@ impl NodeRef {
...
+                    }
+                    Node::Leaf(_) => {
+                        //path.extend(&node.partial);
                     }
-                    Node::Leaf(_) => {}
                 }
-
+                //println!("commit {path:?} => {node:?}");
                 let hash = hash.get_or_init(|| node.compute_hash());
-                acc.push((*hash, node.encode_to_vec()));
```
edg-l, line 70: perf nit, i think compute_hash does encode_to_vec internally, we are probably duplicating work here

iovoid:
Replaced by #4806

Issue #4238, opened 2025-09-01
Handle mempool size


ElFantasma:
Currently we are not taking care of ever growing mempool. All transactions go to the mempool and we don't remove them when they become invalid.

This causes memory usage to grow forever.

PR #4570, opened 2025-09-18
chore(l1): prune mempool

gianbelinche:
**Motivation**
Mempool grows indefinitely

**Description**
Adds a pruning mechanism which every 5 minutes prunes old txs from the mempool

Closes https://github.com/lambdaclass/ethrex/issues/4238


crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,97 @@
...
+            server.clone(),
+            InMessage::PruneMempool,
+        );
+
+        Ok(server)
+    }
+
+    pub fn prune_mempool(&self) -> Result<(), MempoolTxPrunerError> {
+        let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_micros();
+        let mempool_txs = self.blockchain.mempool.mempool_content()?;
```
Oppen: This is cloning the whole pool just for the timestamp and hash. Maybe you could use the filter method?

crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,100 @@
...
+const PRUNE_WAIT_TIME_SECS: u128 = 300; // 5 minutes
+
+// Amount of seconds between each prune
+const PRUNE_INTERVAL_SECS: u64 = 300; // 5 minutes
+
+// Number of microseconds in a second
+const MICROS_PER_SECOND: u128 = 1_000_000;
+
+#[derive(Debug, Clone)]
+pub struct MempoolTxPruner {
```
ilitteri: Lmk what you think but I'll name this `MempoolPruner` since it's simpler, and the Tx part is kind of redundant since the mempool content are only txs.

crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,100 @@
...
+    blockchain: Arc<Blockchain>,
+}
+
+#[derive(Debug, Clone)]
+pub enum InMessage {
+    PruneMempool,
```
ilitteri: This message could be also be renamed to just `Prune`.

crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,100 @@
...
+        send_interval(
+            Duration::from_secs(PRUNE_INTERVAL_SECS),
+            server.clone(),
+            InMessage::PruneMempool,
+        );
+
+        Ok(server)
+    }
+
+    pub fn prune_mempool(&self) -> Result<(), MempoolTxPrunerError> {
```
ilitteri: This may be more debatable, but we could also shorten the name.

crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,100 @@
...
+    async fn handle_cast(
+        &mut self,
+        message: Self::CastMsg,
+        _handle: &spawned_concurrency::tasks::GenServerHandle<Self>,
+    ) -> CastResponse {
+        match message {
+            Self::CastMsg::PruneMempool => {
+                debug!(received = "PruneMempool");
+                let _ = self.prune_mempool().inspect_err(|e| {
+                    debug!(error = %e, "Error pruning mempool transactions");
```
ilitteri: Shouldn't this be a warning log? It seems to me that it shouldn't occur frequently but I might be mistaken.

crates/blockchain/mempool_tx_pruner.rs
gianbelinche: Done

crates/blockchain/mempool_tx_pruner.rs
gianbelinche: Done

crates/blockchain/mempool_tx_pruner.rs
gianbelinche: Done

crates/blockchain/mempool_tx_pruner.rs
gianbelinche: Done

crates/blockchain/mempool_tx_pruner.rs
gianbelinche: Done

crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,105 @@
...
+            InMessage::Prune,
+        );
+
+        Ok(server)
+    }
+
+    pub fn prune(&self) -> Result<(), MempoolPrunerError> {
+        let now = SystemTime::now().duration_since(UNIX_EPOCH)?.as_micros();
+        let filter_fn = |tx: &MempoolTransaction| {
+            now.saturating_sub(tx.time()) > PRUNE_WAIT_TIME_SECS * MICROS_PER_SECOND
```
mpaulucci, line 61: I don't know if this is the logic that we want long term. What does the spec say regarding pruning transactions? How do other clients behave?

Also maybe we should investigate a bit which properties long running mempool txs have:
- To low of a base fee?
- Nonce to high?
- Another?

I suspect that if we have 200k txs then there might be something that we're doing wrongly, perhaps not cleaning them properly after they are included in a block for example.

crates/blockchain/mempool_tx_pruner.rs
```diff
@@ -0,0 +1,105 @@
+use std::{
```
mpaulucci, line 1: nit: seems we should create a `mempool` folder

cmd/ethrex/initializers.rs
```diff
@@ -391,6 +391,10 @@ pub async fn init_l1(
 
     let blockchain = init_blockchain(store.clone(), BlockchainType::L1, true);
 
+    MempoolPruner::spawn(blockchain.clone())
+        .await
+        .expect("Failed to start mempool pruner");
```
ilitteri, line 396: Shouldn't we propagate this error?

ilitteri (approved):

gianbelinche:
Closed in favor of https://github.com/lambdaclass/ethrex/pull/4588 due to new implementation

PR #4572, opened 2025-09-18
feat(l1): validate bytecodes at snap sync end

fedacking:
**Motivation**

We weren't validating the btytecodes in debug assert, added a validation for it.

**Description**

- Adds a validate_bytecodes function and runs it in debug assert at the end of the snap sync.

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -1470,7 +1470,7 @@ pub async fn validate_storage_root(store: Store, state_root: H256) -> bool {
 }
 
 pub async fn validate_bytecodes(store: Store, state_root: H256) -> bool {
-    info!("Starting validate_state_root");
+    info!("Starting validate_bytecodes");
     let mut is_valid = true;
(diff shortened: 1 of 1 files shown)


MegaRedHand (approved):

crates/networking/p2p/sync.rs
```diff
@@ -1467,3 +1468,28 @@ pub async fn validate_storage_root(store: Store, state_root: H256) -> bool {
...
+    info!("Starting validate_bytecodes");
+    let mut is_valid = true;
+    for (account_hash, account_state) in store
+        .iter_accounts(state_root)
+        .expect("we couldn't iterate over accounts")
+    {
+        if account_state.code_hash != *EMPTY_KECCACK_HASH
+            && !store
+                .get_account_code(account_state.code_hash)
+                .is_ok_and(|code| code.is_some())
```
MegaRedHand, line 1483: Maybe we should check that this isn't a DB error.

Issue #4419, opened 2025-09-10
Spawnify `PeerHandler`


juan518munoz:
Current `PeerHandler` would benefit from spawnification in order to tidy up its codebase.

This is also a needed step in the general spawnification of the project.

mpaulucci:
This has been done already

Issue #4245, opened 2025-09-01
Reintroduce a proper implementation for the Kademlia table


ElFantasma:
For the sake of finding peers aggressively we replaced the [Kademlia](https://github.com/lambdaclass/ethrex/blob/75ab1026b1f03b9fc8b1f2360ed37a5505b7cd44/crates/networking/p2p/kademlia.rs) table implementation for a simpler [HashMap based data structure](https://github.com/lambdaclass/ethrex/blob/60161b587c2d90d681c27bf0cee72a6b602302f4/crates/networking/p2p/kademlia.rs). 

- We should take a look to the previous implementation and re-implement required features (and check if there is some code there we can re-use)
- Also, evaluate if the Kademlia table is a good candidate to be implemented with a `spawned` `GenServer`.
- And let's be sure we are not using a bunch of `Arc<Mutex<...>>` for every field in the data structure... Worst case we should have a singe `Arc<Mutex<...>>` for the whole structure
- We should use `xor` distance instead of logarithmic distance

Issue #3963, opened 2025-08-05
Add a table for backend RLPxConnection


fedacking:
The RLPxConnection genserver handles the backend requests by mixing all of the messages into a `backend_channel: Option<mpsc::Sender<Message>>,`. This channel is then recv by those sending the backend message, but it can't discriminate who asked for that message. To do this, we need to write a table for mapping id of requests with genserverhandles that will read this. 

Right now we have a simple solution on the snap branch, but it should be generalized further.

PR #4574, opened 2025-09-18
refactor(l1): spawnify kademlia

### LINKED ISSUE #4383: Spawnify Kademlia Peer Table
Once #4352 is done we should move the access to the table of peers to be through messages to a gen server that handles adding, removing and getting good peers for our operations. This issue comes from a discussion in #4294

ElFantasma:
**Motivation**

Current kademlia module is just a peer table that requires a lot of Mutexes and locks() (and in fact it is not a kademlia implementation)

**Description**

This pr converts the PeerTable data structure into a GenServer in order to avoid all the `Mutex.lock()` required in previous version and ensure data consistency.

It is still a work in progress as I have to deal with the possible errors when accessing PeerTable server.

Closes #4383


crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -0,0 +1,1188 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Contact {
+    pub node: Node,
+    /// The timestamp when the contact was last sent a ping.
+    /// If None, the contact has never been pinged.
+    pub validation_timestamp: Option<Instant>,
+    /// The hash of the last unacknowledged ping sent to this contact, or
+    /// None if no ping was sent yet or it was already acknowledged.
+    pub ping_hash: Option<H256>,
```
gianbelinche, line 39: Maybe all of these structs (Contact, PeerData and PeerChannels) could be defined in a separate file

crates/networking/p2p/network.rs
```diff
@@ -98,8 +98,8 @@ pub enum NetworkError {
     TxBroadcasterError(#[from] TxBroadcasterError),
 }
 
-pub fn peer_table() -> Kademlia {
-    Kademlia::new()
+pub fn peer_table() -> PeerTableHandle {
+    PeerTable::spawn()
```
gianbelinche: Wouldn't this spawn multiple peer tables if used multiple times?

gianbelinche: Maybe I'm wrong but this seems as a function that you would want to use to get a reference to the global peer table, instead of just creating a new one

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -0,0 +1,1162 @@
...
+        capabilities: Vec<Capability>,
+    ) -> Result<(), PeerTableError> {
+        self.0
+            .call(CallMessage::NewConnectedPeer {
+                node,
+                channels,
+                capabilities,
+            })
+            .await
+            .map_err(|e| PeerTableError::InternalError(e.to_string()))?;
```
gianbelinche: Shouldn't all of these calls with no return be cast?

crates/networking/p2p/network.rs
ElFantasma: There is no global peer table. It's up to the user to decide if it will spawn one or many. All our "singleton" `GenServer`s behave the same way

gianbelinche: Can re rename this to something like `create_peer_table` then? So that it is clear that a new one is being generated

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: Done

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: PeerChannels should disappear soon. And Contact and PeerData are not big enough to worth another module...  They are just data holders.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -141,7 +140,7 @@ pub enum InnerState {
     HandshakeFailed,
     Initiator(Initiator),
     Receiver(Receiver),
-    Established(Established),
+    Established(Box<Established>),
```
gianbelinche, line 141: Why is it now a box?

ElFantasma: Clippy complained about the difference in size and suggested about Boxing it 🤷‍♂️

crates/networking/p2p/peer_handler.rs
```diff
@@ -279,8 +273,8 @@ impl PeerHandler {
             {
                 trace!("We received a download chunk from peer");
                 if headers.is_empty() {
-                    self.peer_table.free_peer(peer_id).await;
-                    self.peer_table.record_failure(peer_id).await;
+                    self.peer_table.free_peer(&peer_id).await?;
+                    self.peer_table.record_success(&peer_id).await?;
```
gianbelinche: This was a record failure, it is now a record success

crates/networking/p2p/network.rs
ElFantasma: It's the naming convention we used for all our `GenServer`s. We use `spawn` to make it clear we are spawning a new process.  

BTW, I removed this function from here. It doesn't make sense to have a function that just calls `PeerTable::spawn()`

crates/networking/p2p/peer_handler.rs
ElFantasma: Nice catch! Bad copy/paste probably

crates/networking/p2p/discv4/server.rs
```diff
@@ -393,134 +372,121 @@ impl DiscoveryServer {
...
+            .is_err()
+        {
+            return Ok(());
         }
 
         if let Err(err) = self.send_enr_response(hash, from).await {
             error!(sent = "ENRResponse", to = %format!("{from}"), err = ?err, "Error sending message");
-            return;
+            return Ok(());
         }
```
gianbelinche, line 437: Why are we returning Ok after errors?

ElFantasma: Return type changed, now we return a `DiscoveryServerError` on  unhandled failures, and just `Ok(())` on the happy path as well as on usual handled errors. Send failed, but we don't want to stop the DiscoveryServer.

gianbelinche: Should we leave as error! then? or change to debug?

crates/networking/p2p/sync/storage_healing.rs
```diff
@@ -287,9 +287,12 @@ pub async fn heal_storage_trie(
                     .extend(inflight_request.requests.clone());
                 peers
                     .peer_table
-                    .record_failure(inflight_request.peer_id)
-                    .await;
-                peers.peer_table.free_peer(inflight_request.peer_id).await;
+                    .record_failure(&inflight_request.peer_id)
+                    .await?;
```
Oppen: If this fails the peer isn't freed.

Oppen: We should probably have a single method `record_failure_and_free`.

crates/networking/p2p/sync/storage_healing.rs
```diff
@@ -390,26 +397,27 @@ async fn zip_requeue_node_responses_score_peer(
...
-) -> Option<Vec<NodeResponse>> {
+) -> Result<Option<Vec<NodeResponse>>, SyncError> {
     trace!(
         "We are processing the nodes, we received {} nodes from our peer",
         trie_nodes.nodes.len()
     );
     let Some(request) = requests.remove(&trie_nodes.id) else {
         info!("We received a response where we had a missing requests {trie_nodes:?}");
-        return None;
+        return Ok(None);
```
Oppen, line 403: Is it OK to return without freeing the peer here? I ask because we don't, but maybe it's intentional (e.g. we want to reuse the peer afterwards).

crates/networking/p2p/sync.rs
```diff
@@ -804,14 +805,15 @@ impl SnapBlockSyncState {
 /// Safety function that frees all peer and logs an error if we found freed peers when not expectig to
 /// Logs with where the function was when it found this error
 /// TODO: remove this function once peer table has moved to spawned implementation
-async fn free_peers_and_log_if_not_empty(peer_handler: &PeerHandler) {
-    if peer_handler.peer_table.free_peers().await != 0 {
+async fn free_peers_and_log_if_not_empty(peer_handler: &mut PeerHandler) -> Result<(), SyncError> {
+    if peer_handler.peer_table.free_peers().await? != 0 {
```
Oppen, line 809: I think we mean to log the error first.

ElFantasma: `free_peers()` return the number of non-free peers. If it is not 0 it means we ended a step without freeing all peers, and that's the error we want to log (meaning we have a "peer leak" where we endend a step and a peer is still attached to it)

crates/networking/p2p/sync/storage_healing.rs
ElFantasma: I will check this 👍

crates/networking/p2p/sync/storage_healing.rs
ElFantasma: If `record_failure` returns an error it means the `PeerTable` is not available, and the whole process should break. It's a panicking-kind failure.

But I'll analyze the idea of merging the two requests anyway, it may be more efficient.

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -0,0 +1,1127 @@
...
+            _ => unreachable!(),
+        }
+    }
+
+    /// Get list of connected peers
+    pub async fn get_connected_nodes(&mut self) -> Result<Vec<Node>, PeerTableError> {
+        match self.0.call(CallMessage::GetConnectedNodes).await? {
+            OutMessage::Nodes(nodes) => Ok(nodes),
+            _ => unreachable!(),
+        }
```
MegaRedHand: The `match` with `unreachable!` seems like a possible pain-point on `spawned`.

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -0,0 +1,1127 @@
...
+                    return None;
+                }
+                peer_data
+                    .channels
+                    .clone()
+                    .map(|peer_channels| (*node_id, peer_channels))
+            })
+            .collect();
+        peers.shuffle(&mut rand::rngs::OsRng);
+        peers.first().cloned()
```
MegaRedHand: Here we can use [`choose()`](https://rust-random.github.io/rand/rand/seq/trait.IteratorRandom.html#method.choose) to avoid the allocation and copies from the shuffle.

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -0,0 +1,1127 @@
...
+        let validation_is_stale = !contact.was_validated()
+            || contact
+                .validation_timestamp
+                .map(|ts| Instant::now().saturating_duration_since(ts) > revalidation_interval)
+                .unwrap_or(false);
+
+        let sent_ping_is_stale = contact
+            .validation_timestamp
+            .map(|ts| Instant::now().saturating_duration_since(ts) > sent_ping_ttl)
+            .unwrap_or(false);
```
MegaRedHand, line 785: This is to retry any lost pings, right? Do we stop sending pings after a few attempts?

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -0,0 +1,1127 @@
...
+        match self.0.call(CallMessage::PeerCount).await? {
+            OutMessage::PeerCount(peer_count) => Ok(peer_count),
+            _ => unreachable!(),
+        }
+    }
+
+    /// Return the amount of connected peers that matches any of the given capabilities
+    pub async fn peer_count_by_capabilities(
+        &mut self,
+        capabilities: &[Capability],
```
MegaRedHand, line 307: We receive this by reference and clone it afterwards. Is this because the clone is an internal detail?

crates/networking/p2p/peer_handler.rs
```diff
@@ -441,48 +435,54 @@ impl PeerHandler {
...
-                continue;
-            }
-            if let Some(block_headers) = tokio::time::timeout(PEER_REPLY_TIMEOUT, async move {
-                loop {
-                    match receiver.recv().await {
-                        Some(RLPxMessage::BlockHeaders(BlockHeaders { id, block_headers }))
-                            if id == request_id =>
-                        {
-                            return Some(block_headers);
+            match self.get_random_peer(&SUPPORTED_ETH_CAPABILITIES).await? {
```
MegaRedHand, line 437: `get_random_peer` doesn't check if the peer is used. What would happen if the peer is already serving one of our requests?

MegaRedHand (approved):

crates/networking/p2p/sync/storage_healing.rs
ElFantasma: Ok, answer is no: we don't have to free the peer because we don't find a `request` that matches the `response`, and that means we don't have an associated peer

crates/networking/p2p/sync/storage_healing.rs
ElFantasma: Done.

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: Yeah 😞 
Any suggestion here?
`if let` version doesn't look better:
```rust
        if let OutMessage::Nodes(nodes) = self.0.call(CallMessage::GetConnectedNodes).await? {
            Ok(nodes)
        } else {
            unreachable!()
        }
```

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: ⭐

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: Not sure; this refactor is just converting existent functionality into a GenServer-based one.
There is a couple of rewrites to improve the whole discovery/peer handling process that should address that. (See #4245, #4419, #3963)

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: Yes, the handler here hides the internal `CallMessage` construction (that requires a clone) and response conversion.

crates/networking/p2p/discv4/peer_table.rs
MegaRedHand: I don't think there's any solution from our side.

Other actor libs solve it by having each message type correspond to a `Handler<InMessage>` trait, which specifies a type for the input message and also for the output. That way, the `call` function returns the correct type. There is a tradeoff, however, since it also means having more boilerplate.

crates/networking/p2p/peer_handler.rs
ElFantasma: Yeah, that seems to be a pre-existent issue. It's [even documented](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/peer_handler.rs#L160)
Eventually our peers should be able to handle multiple inflight requests.

ElFantasma: Added an issue to review this: https://github.com/lambdaclass/ethrex/issues/4756

PR #4576, opened 2025-09-19
feat(l1): handle non-canonical headers by hash in `GetBlockHeaders`

MegaRedHand:
**Motivation**

We're currently returning empty header responses when we receive a `GetBlockHeaders` request specifying the start with the block's hash, and the hash isn't part of our canonical view of the chain.

**Description**

This PR adds support for this case by simply looking for the starting block in the DB and returning it. This stays simple enough and, though suboptimal, should make the simple case of a peer asking for a sidechain to work.

It could be further extended by looking for the block's parent through the parent hash, which should allow us to support cases where `reverse` is `true`.


crates/networking/p2p/rlpx/eth/blocks.rs
```diff
@@ -119,26 +127,48 @@ impl GetBlockHeaders {
...
                     break;
                 }
-                // TODO(#1073)
-                // Research what we should do when an error is found in a P2P request.
-                Err(err) => {
-                    error!("Error accessing DB while building header response for peer: {err}");
-                    return vec![];
+            };
+            let Some(block_header) = block_header_opt else {
+                trace!(block_number=%current_block, "Block header not found");
```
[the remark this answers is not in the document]
MegaRedHand: Done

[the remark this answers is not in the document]
MegaRedHand: Done

crates/networking/p2p/rlpx/eth/blocks.rs
```diff
@@ -119,26 +127,48 @@ impl GetBlockHeaders {
...
+                break;
+            };
+            headers.push(block_header);
+
+            // Update to next block to fetch
+            match current_block {
+                // We don't support fetching headers by hash, unless the hash is
+                // part of the canonical chain, so we break here.
+                // TODO: we could support fetching by hash in descending order,
+                // by fetching the parent of each block.
```
mpaulucci: According to https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getblockheaders-0x03 , even when fetching by hash, it should be part of the canonical chain, so maybe we should adapt the code/comments to that?

MegaRedHand: I didn't catch that part of the spec before, but Geth apparently services the request even if the hash is not part of the canonical chain:

https://github.com/ethereum/go-ethereum/blob/7e7925460507db9988088223ecd069af507815d6/eth/protocols/eth/handlers.go#L193

We can keep the logic as it was before, since it already supports returning headers in the canonical chain. I think we should add this, however, since the current behavior could cause problems in our syncing algorithm when handling reorgs in an ethrex-only network.

MegaRedHand: [b8e2005](https://github.com/lambdaclass/ethrex/pull/4576/commits/b8e20058898d04767176fcdfa5995933d0f7acaf)

PR #4579, opened 2025-09-19
pef(levm): improve instruction fetching and callerpop

edg-l:
**Motivation**

On the old gas bench callerpop improves by 100 mgas from 820 to 920~ locally on linux.

EEST benches:

image

Closes #issue_number

### DIFF
--- a/crates/vm/levm/src/call_frame.rs
+++ b/crates/vm/levm/src/call_frame.rs
@@ -10,6 +10,7 @@ use bytes::Bytes;
 use ethrex_common::{Address, U256};
 use keccak_hash::H256;
 use std::{
+    cell::OnceCell,
     collections::{BTreeMap, HashMap},
     fmt,
 };
@@ -209,6 +210,7 @@ pub struct CallFrame {
     pub pc: usize,
     /// Address of the account that sent the message
     pub msg_sender: Address,
+    pub msg_sender_u256: OnceCell<U256>,
     /// Address of the recipient of the message
     pub to: Address,
     /// Address of the code to execute. Usually the same as `to`, but can be different
@@ -307,6 +309,7 @@ impl CallFrame {
             gas_limit,
             gas_remaining: gas_limit,
             msg_sender,
+            msg_sender_u256: OnceCell::new(),
             to,
             code_address,
             bytecode: bytecode.clone(),
--- a/crates/vm/levm/src/call_frame.rs
+++ b/crates/vm/levm/src/call_frame.rs
@@ -12,7 +12,7 @@ use keccak_hash::H256;
 use std::{
     cell::OnceCell,
     collections::{BTreeMap, HashMap},
-    fmt, ptr,
+    fmt,
 };
 
 #[derive(Clone, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


azteca1998 (approved):

crates/vm/levm/src/opcodes.rs
```diff
@@ -368,14 +369,11 @@ impl<'a> OpCodeFn<'a> {
 }
 
 impl<'a> VM<'a> {
-    /// The opcode table mapping opcodes to opcode handlers for fast lookup.
-    pub(crate) const OPCODE_TABLE: [OpCodeFn<'a>; 256] = VM::build_opcode_table();
```
Oppen, line 372: Dropping the const is not really necessary.

edg-l: what do you mean? i have to mutate the table at runtime according to the fork

crates/vm/levm/src/call_frame.rs
```diff
@@ -209,6 +235,7 @@ pub struct CallFrame {
     pub pc: usize,
     /// Address of the account that sent the message
     pub msg_sender: Address,
+    pub msg_sender_u256: OnceCell<U256>,
```
JereSalo: Is this particular change beneficial enough to keep it? If so, can we add a comment saying that we keep this just for performance reasons?
- We store the U256 so that if the caller opcode is called multiple times we just convert to U256 once.
- We store it in a OnceCell because if the opcode is not called then we don't need to compute Address to U256.
Is this correct?

crates/vm/levm/src/opcodes.rs
Oppen: You can have multiple tables. The draft I shared before does that IIRC.

crates/vm/levm/src/call_frame.rs
Oppen: Why the `OnceCell`? Just initialize it at construction time.

Oppen:
Can we benchmark these optimizations individually? The aggregate is not good enough to judge whether a particular optimization carries its weight in terms of code complexity. For all we know, one of them may be worsening performance and another one carrying it.

crates/vm/levm/src/opcodes.rs
Arkenan: I agree with this. We have a set number of forks. I think we should build a const opcode table for each fork at the beginning of execution, and then at VM build time we should just match over the fork and get a pointer to the appropriate op code table. Let's test this. If it causes further improvement, we leave it, if not, we revert to the current state and merge the PR.

Arkenan (changes requested):

crates/vm/levm/src/call_frame.rs
```diff
@@ -125,6 +126,31 @@ impl Stack {
...
+        // `self.offset` is known to be within `STACK_LIMIT`.
+        #[expect(unsafe_code, reason = "next_offset == self.offset - 1 >= 0")]
+        unsafe {
+            *self
+                .values
+                .get_unchecked_mut(next_offset)
+                .0
+                .as_mut_ptr()
+                .cast() = [0u64; 4];
+        }
```
Oppen, line 147: Did we try a safe approach here? Do we have numbers to compare?

crates/vm/levm/src/opcodes.rs
edg-l: changed to const

crates/vm/levm/src/call_frame.rs
edg-l: removed the change for now, ill look into it separately

edg-l:
Seeing improvements of up to 40% still on eest after changes, on some push benches we overtake reth

crates/vm/levm/src/call_frame.rs
Arkenan: This is already how we do it in the general push case, and we did it because it improved execution times.

PR #4588, opened 2025-09-19
chore(l1): Prune mempool

gianbelinche:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -19,7 +19,7 @@ use ethrex_storage::error::StoreError;
 use std::collections::HashSet;
 
 // Max number of transactions in the mempool
-const MEMPOOL_MAX_SIZE: usize = 10_000; //TODO: Define
+const MEMPOOL_MAX_SIZE: usize = 10000; //TODO: Define
 
 #[derive(Debug, Default)]
 pub struct Mempool {
@@ -31,20 +31,26 @@ pub struct Mempool {
 }
 impl Mempool {
     pub fn new() -> Self {
-        Self::default()
+        Mempool {
+            txs_order: RwLock::new(Vec::with_capacity(MEMPOOL_MAX_SIZE)),
+            ..Default::default()
+        }
     }
 
     /// Remove the oldest transaction in the pool
-    pub fn remove_oldest_transaction(&self) -> Result<(), StoreError> {
+    pub fn remove_oldest_transaction(
+        &self,
+        tx_pool: &mut std::sync::RwLockWriteGuard<'_, HashMap<H256, MempoolTransaction>>,
+    ) -> Result<(), StoreError> {
         let oldest_entry = self
             .txs_order
-            .write()
+            .read()
             .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?
             .first()
             .cloned();
 
         if let Some(oldest_hash) = oldest_entry {
-            self.remove_transaction(&oldest_hash)?;
+            self.remove_transaction_with_lock(&oldest_hash, tx_pool)?;
         }
 
         Ok(())
@@ -56,8 +62,12 @@ impl Mempool {
         hash: H256,
         transaction: MempoolTransaction,
     ) -> Result<(), StoreError> {
-        if self.txs_by_sender_nonce.read().iter().len() >= MEMPOOL_MAX_SIZE {
-            self.remove_oldest_transaction()?;
+        let mut transaction_pool = self
+            .transaction_pool
+            .write()
+            .map_err(|error| StoreError::MempoolReadLock(error.to_string()))?;
+        if transaction_pool.len() >= MEMPOOL_MAX_SIZE {
+            self.remove_oldest_transaction(&mut transaction_pool)?;
         }
         self.txs_order
             .write()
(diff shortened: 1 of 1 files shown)


crates/blockchain/mempool.rs
```diff
@@ -18,16 +18,42 @@ use ethrex_common::{
 use ethrex_storage::error::StoreError;
 use std::collections::HashSet;
 
+// Max number of transactions in the mempool
+const MEMPOOL_MAX_SIZE: usize = 10000; //TODO: Define
```
tomip01, line 22: Should we link an issue to this?

crates/blockchain/mempool.rs
```diff
@@ -18,16 +18,42 @@ use ethrex_common::{
...
+
 #[derive(Debug, Default)]
 pub struct Mempool {
     broadcast_pool: RwLock<HashSet<H256>>,
     transaction_pool: RwLock<HashMap<H256, MempoolTransaction>>,
     blobs_bundle_pool: Mutex<HashMap<H256, BlobsBundle>>,
     txs_by_sender_nonce: RwLock<BTreeMap<(H160, u64), H256>>,
+    txs_order: RwLock<Vec<H256>>,
```
tomip01, line 30: Maybe if we want to remove by a key, should we use a structure like BTreeMap?

gianbelinche: The thing is we want to remove the oldest inserted, with a vector we know that the position 0 is the oldest.

crates/blockchain/mempool.rs
gianbelinche: https://github.com/lambdaclass/ethrex/issues/4600

crates/blockchain/mempool.rs
tomip01: You’re right, but I meant to avoid scanning the entire Vec when we want to remove the oldest element in order to retain the other values.
Maybe what I’m suggesting is just a premature optimization and not really necessary.

tomip01 (approved):

crates/blockchain/mempool.rs
```diff
@@ -108,10 +142,24 @@ impl Mempool {
...
             .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?;
+        self.remove_transaction_with_lock(hash, &mut tx_pool)?;
+        Ok(())
+    }
+
+    /// Remove a transaction from the pool with the transaction pool lock already taken
+    fn remove_transaction_with_lock(
+        &self,
+        hash: &H256,
+        tx_pool: &mut std::sync::RwLockWriteGuard<'_, HashMap<H256, MempoolTransaction>>,
```
MegaRedHand: This can be simplified to:

```suggestion
        tx_pool: &mut HashMap<H256, MempoolTransaction>,
```

crates/blockchain/mempool.rs
```diff
@@ -126,6 +174,7 @@ impl Mempool {
                 .remove(&(tx.sender(), tx.nonce()));
             tx_pool.remove(hash);
             broadcast_pool.remove(hash);
+            txs_order.retain(|h| h != hash);
```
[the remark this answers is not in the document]
MegaRedHand: I agree with this comment. Using retain would mean copying 320Kb of hashes on a full mempool. This could be OK for a one-time call, but since it can be triggered by peers multiple times per second, it could be a DoS vector.

gianbelinche:
> Let's open an issue for avoiding calls to `retain` on each new tx received by peers.

Already handled it

crates/blockchain/mempool.rs
```diff
@@ -18,16 +18,42 @@ use ethrex_common::{
...
+    ) -> Result<(), StoreError> {
+        let mut txs_order = self
+            .txs_order
+            .write()
+            .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?;
+        let oldest_entry = txs_order.first().cloned();
+
+        if let Some(oldest_hash) = oldest_entry {
+            self.remove_transaction_with_lock(&oldest_hash, tx_pool)?;
+            txs_order.remove(0);
```
MegaRedHand: On a `Vec`, this still requires copying all the transactions. Let's open an issue to handle this in a better way.

crates/blockchain/mempool.rs
MegaRedHand, line 30: Let's use a standard `Mutex` here, since we're not using calling `read` anywhere

crates/blockchain/mempool.rs
```diff
@@ -108,6 +142,23 @@ impl Mempool {
             .transaction_pool
             .write()
             .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?;
+        self.remove_transaction_with_lock(hash, &mut tx_pool)?;
+        if tx_pool.get(hash).is_some() {
```
MegaRedHand: Can this still happen after calling `remove_transaction_with_lock`?

crates/blockchain/mempool.rs

crates/blockchain/mempool.rs
gianbelinche: Reverted the change

crates/blockchain/mempool.rs
gianbelinche: With the reverted change there is a read now

crates/blockchain/mempool.rs
gianbelinche: https://github.com/lambdaclass/ethrex/issues/4605

PR #4602, opened 2025-09-22
refactor(l1,l2): use `BlockHash` type instead of `H256`

mpaulucci:
**Motivation**
Even though `BlockHash` is just a hash, lets make it explicit.

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -18,16 +18,42 @@ use ethrex_common::{
 use ethrex_storage::error::StoreError;
 use std::collections::HashSet;
 
+// Max number of transactions in the mempool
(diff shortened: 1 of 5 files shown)


MegaRedHand (approved):

ilitteri:
> I think a better approach is to use a struct instead of a type alias. Given that the type alias does not prevent anybody to call the functions with a H256 coming from, for example, a transaction hash. With type alias solution we are not using type system in our favor.

I agree with this, but in addition, I'd only go for the struct approach only if it adds some value since it'd add some complexity to the code, starting with having the `ethrex-common` crate as a dependency only for this type.

mpaulucci:
Closing for now since we did not agree on a solution

PR #4611, opened 2025-09-23
feat(l2): add verification key upgrade functions

ManuelBilbao:
**Motivation**

Currently when someone wants to upgrade the ethrex version of an L2, they have to perform an UUPS upgrade to the contract in order to upgrade the verification keys.

**Description**

Add a function to upgrade SP1 and RISC0 VKs without the need of a full upgrade to the contract

### DIFF
--- a/crates/l2/contracts/src/l1/based/OnChainProposer.sol
+++ b/crates/l2/contracts/src/l1/based/OnChainProposer.sol
@@ -230,11 +230,13 @@ contract OnChainProposer is
     /// @inheritdoc IOnChainProposer
     function upgradeSP1VerificationKey(bytes32 new_vk) public onlyOwner {
         SP1_VERIFICATION_KEY = new_vk;
+        emit VerificationKeyUpgraded("SP1", new_vk);
     }
 
     /// @inheritdoc IOnChainProposer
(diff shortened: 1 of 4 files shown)


crates/l2/contracts/src/l1/based/OnChainProposer.sol
```diff
@@ -227,6 +227,16 @@ contract OnChainProposer is
         BRIDGE = bridge;
     }
 
+    /// @inheritdoc IOnChainProposer
+    function upgradeSP1VerificationKey(bytes32 new_vk) public onlyOwner {
+        SP1_VERIFICATION_KEY = new_vk;
```
jrchatruc, line 232: Let's emit an event here after setting it (same for RISC0 and the non-based cases)

ManuelBilbao: Done [95d6580](https://github.com/lambdaclass/ethrex/pull/4611/commits/95d65801442734ec42f9e82c2874314d9b460328)

ManuelBilbao:
I created [an issue](https://github.com/lambdaclass/ethrex/issues/4618) to document this process along with other upgrades

Issue #4592, opened 2025-09-22
Refactor mempool


gianbelinche:
The mempool has currently 5 different locks, which is prone to errors.

Possible refactors:

- Spawnify mempool: This would make the code simpler to understand, though it would remove the concurrent readings we currently have with the rwLocks
- Unify all locks in one: Create a separate struct for the mempool state and lock that instead of everything separately, since we always write everything together, this shouldn't change performance. Continuing using rwLocks would still allow concurrent readings.

Issue #4605, opened 2025-09-22
Refactor removing transaction in mempool


gianbelinche:
After adding the mechanism to remove the oldest tx in the mempool, we now have to iterate through a vector to remove a tx every time we add a new tx. This is an attack vector.

PR #4615, opened 2025-09-23
refactor(l1): mempool locks

gianbelinche:
**Motivation**
The mempool has several locks being used at the same time, which is prone to concurrency issues

**Description**
Unify the mempool state in a single struct and use only one lock for it, since we write everything at the same time, this shouldn't affect performance, since we can still do concurrent readings

Closes https://github.com/lambdaclass/ethrex/issues/4592


crates/blockchain/mempool.rs
```diff
@@ -85,47 +119,50 @@ impl Mempool {
...
-            self.txs_by_sender_nonce
-                .write()
-                .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?
-                .remove(&(tx.sender(), tx.nonce()));
-            tx_pool.remove(hash);
-            broadcast_pool.remove(hash);
+            inner.txs_by_sender_nonce.remove(&(tx.sender(), tx.nonce()));
+            inner.transaction_pool.remove(hash);
+            inner.broadcast_pool.remove(hash);
+            inner.txs_order.retain(|h| h != hash);
```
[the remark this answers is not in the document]
gianbelinche: There is an issue for this https://github.com/lambdaclass/ethrex/issues/4605

crates/blockchain/mempool.rs
```diff
@@ -36,47 +68,49 @@ impl Mempool {
...
     ) -> Result<(), StoreError> {
-        self.txs_by_sender_nonce
+        let mut inner = self
+            .inner
             .write()
-            .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?
+            .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?;
+        if inner.transaction_pool.len() >= MEMPOOL_MAX_SIZE {
+            self.remove_oldest_transaction(&mut inner)?;
+        }
```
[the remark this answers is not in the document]
gianbelinche: If it fails, function ends and the tx is not added

crates/blockchain/mempool.rs
```diff
@@ -18,16 +18,48 @@ use ethrex_common::{
...
+    /// Remove the oldest transaction in the pool
+    fn remove_oldest_transaction(&self, inner: &mut MempoolInner) -> Result<(), StoreError> {
+        let oldest_entry = inner.txs_order.first().cloned();
+
+        if let Some(oldest_hash) = oldest_entry {
+            self.remove_transaction_with_lock(&oldest_hash, inner)?;
+        }
+
+        Ok(())
     }
```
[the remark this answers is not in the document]

crates/blockchain/mempool.rs
```diff
@@ -62,55 +66,49 @@ impl Mempool {
         hash: H256,
         transaction: MempoolTransaction,
     ) -> Result<(), StoreError> {
-        let mut transaction_pool = self
-            .transaction_pool
+        let mut inner = self
+            .inner
             .write()
             .map_err(|error| StoreError::MempoolWriteLock(error.to_string()))?;
```
MegaRedHand: We should move `inner.read()` and `inner.write()` calls to a helper:

```rust
let mut inner = self.write()?;
```

crates/blockchain/mempool.rs
```diff
@@ -119,62 +119,39 @@ impl Mempool {
...
+        self.remove_transaction_with_lock(hash, &mut inner)?;
         Ok(())
     }
 
     /// Remove a transaction from the pool with the transaction pool lock already taken
     fn remove_transaction_with_lock(
         &self,
         hash: &H256,
-        tx_pool: &mut HashMap<H256, MempoolTransaction>,
+        inner: &mut MempoolInner,
```
MegaRedHand: This signature looks a bit weird, since we're actually receiving two references to the same object. We can move this function to `MempoolInner`, and make it be `fn remove_transaction_with_lock(&mut self, hash: &H256) -> ...`

MegaRedHand (approved):

crates/blockchain/mempool.rs

PR #2952, opened 2025-05-28
fix(core): fix MPT branch restructuring with stateless execution

iovoid:
**Motivation**

Stateless execution used to fail on mainnet block 22567217, after failing to remove a node from the storage trie.

**Description**

While removing a node, a branch left with one child and no value is to be restructured. During stateless execution, it can happen that such child isn't known.

As [explained in the documentation](https://github.com/lambdaclass/ethrex/blob/main/crates/l2/docs/prover.md#execution-witness), this should be handled by modifying the standard MPT implementation.

While injecting possible child values (case 1) happens at the stateless client, the assumption that a missing child is a branch (case 2) has to be made at the trie.

Makes progress towards #2913

PR #4616, opened 2025-09-23
fix(replay): fix execution with eth_getProof

JereSalo:
**Motivation**

**Description**

Execution with eth_getProof had some caveats regarding the trie. For making it work we need to do 2 things:
1. Pre-inject possible nodes to the trie
2. Make a modification to the MPT in case we don't find a missing node.

The `1` was already done in the code but wasn't working properly, the problem was just that the root node wasn't being inserted into the nodes that were used for building the trie that's used for getting potential child nodes.

The 2 is a modification that we don't want in our MPT so we didn't carry along [with that](https://github.com/lambdaclass/ethrex/pull/2952). However, in execution without zkVM we have the possibility of inserting arbitrary nodes with the hashes that we want and RLP content that doesn't necessarily match to that hash, so we can "deceive" the trie into thinking it has some nodes when in fact they are just dummy nodes.
How can we use it to our advantage? We insert dummy branch nodes in every single place where a storage node is being referenced but it's missing. This way, if we want to do a trie restructuring we will find the node that otherwise we wouldn't have found, and the good thing is that we don't care about the contents of this node because they won't change, in this particular case the node will be referenced by a one nibble extension node. For more info read https://github.com/kkrt-labs/zk-pig/blob/main/docs/modified-mpt.md#modified-mpt-implementation

Closes #issue_number


cmd/ethrex_replay/src/helpers.rs
```diff
@@ -0,0 +1,36 @@
...
+) -> eyre::Result<HashSet<NodeHash>> {
+    let mut referenced_hashes: HashSet<NodeHash> = HashSet::new();
+
+    for (_node_hash, node_rlp) in nodes.iter() {
+        let node = Node::decode(node_rlp)?;
+        match node {
+            Node::Branch(node) => {
+                for choice in &node.choices {
+                    let NodeRef::Hash(hash) = *choice else {
+                        unreachable!()
```
ilitteri: Return errors instead of this.

ilitteri (approved):

cmd/ethrex_replay/src/cli.rs
```diff
@@ -534,13 +510,33 @@ async fn replay_no_zkvm(cache: Cache, opts: &EthrexReplayOptions) -> eyre::Resul
...
                 Err(_) => continue,
             };
 
+            // Fill storage trie with dummy branch nodes that have the hash of the missing nodes
+            // This is useful for eth_getProofs when we want to restructure the trie after removing a node whose sibling isn't known
+            // We assume the sibling is a branch node because we already covered the cases in which it's a Leaf or Extension node by injecting nodes in the witness.
+            // For more info read: https://github.com/kkrt-labs/zk-pig/blob/v0.8.0/docs/modified-mpt.md
+            {
+                let mut storage_nodes = storage_trie.lock().unwrap();
+                let dummy_branch = Node::from(BranchNode::default()).encode_to_vec();
```
tomip01, line 538: We can create this `dummy_branch` outside the for loop for the addresses

tomip01 (approved):
LGTM! just one comment

cmd/ethrex_replay/src/cli.rs
JereSalo: Normally I'd agree but leaving it here makes code clearer for the reader, if I move it above the for loop the code will be slightly harder to understand. Difference in performance is negligible so we can keep it as is.

cmd/ethrex_replay/src/helpers.rs
JereSalo: https://github.com/lambdaclass/ethrex/pull/4616/commits/cd843f0e8b35c3f62624ccbb08e8462259444a05

PR #4622, opened 2025-09-23
feat(l2): custom native token 

ilitteri:
**Motivation**

We want to be able to start the L2 with the native token to be one selected by the user. This means the currency on which L2 balances and fees are paid are actually ETH or an ERC20 selected.

**Description**

* Added a new `native_token_l1_address` option to `DeployerOptions` in the deployer, allowing configuration of the L1 address of the L2 native token (ETH or ERC20). This value is now written to the `.env` file after deployment.
* Updated contract deployment and initialization to pass the `native_token_l1_address` to the bridge contract, and updated the bridge initializer signature to accept this new argument.
* Refactored the Solidity `CommonBridge` contract to support deposits and withdrawals of either ETH or a configurable ERC20 as the native token. And updates the `deposit` and `claimWithdrawal` logic accordingly.
* Updated integration tests to support the new native token configuration, passing the `native_token_l1_address` through test helpers. 
* Added a new contract binary in the fixtures `NativeTokenTest.bin` is the same as the `TestToken.bin` but with more minting for enabling the deposits as they use much more. And also to not collide with the contract that is deployed for the integration tests.

**How to test**

We need to deploy the L1 token contract before starting the L2. We need two terminals

Start L1
```
cd crates/l2
make rm-db-l1 init-l1
```

Deploy the L1 token
```
rex deploy $(cat fixtures/contracts/ERC20/NativeTokenTest.bin) 0 0xbcdf20249abf0ed6d944c0288fad489e33f66b3960d9e6229c1cd214ed3bbe31
```

Add in the `deploy-l1` Makefile rule under `crates/l2` the line `--native-token-l1-address 0xb4B46bdAA835F8E4b4d8e208B6559cD267851051`

Then
```
cd crates/l2
make rm-db-l2 deploy-l1 init-l2
```

### DIFF
--- a/crates/l2/contracts/src/l1/CommonBridge.sol
+++ b/crates/l2/contracts/src/l1/CommonBridge.sol
@@ -62,11 +62,11 @@ contract CommonBridge is
     /// @notice How much of each L1 token was deposited to each L2 token.
     /// @dev Stored as L1 -> L2 -> amount
     /// @dev Prevents L2 tokens from faking their L1 address and stealing tokens
-    /// @dev The token can take the value {ETH_TOKEN} to represent ETH
+    /// @dev The token can take the value {NATIVE_TOKEN_L2} to represent the token of the L2
     mapping(address => mapping(address => uint256)) public deposits;
 
-    /// @notice Token address used to represent ETH
-    address public constant ETH_TOKEN =
+    /// @notice Token address used to represent the token of the L2
+    address public constant NATIVE_TOKEN_L2 =
         0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
 
     /// @notice Owner of the L2 system contract proxies
@@ -90,7 +90,7 @@ contract CommonBridge is
     /// @notice The L1 token address that is treated as the one to be bridged to the L2.
     /// @dev If set to address(0), ETH is considered the native token.
     /// Otherwise, this address is used for native token deposits and withdrawals.
-    address public NATIVE_TOKEN;
+    address public NATIVE_TOKEN_L1;
 
     modifier onlyOnChainProposer() {
         require(
@@ -124,7 +124,7 @@ contract CommonBridge is
 
         PRIVILEGED_TX_MAX_WAIT_BEFORE_INCLUSION = inclusionMaxWait;
 
-        NATIVE_TOKEN = _nativeToken;
+        NATIVE_TOKEN_L1 = _nativeToken;
 
         OwnableUpgradeable.__Ownable_init(owner);
         ReentrancyGuardUpgradeable.__ReentrancyGuard_init();
@@ -220,7 +220,7 @@ contract CommonBridge is
         uint256 value;
 
         // Here we define value depending on whether the native token is ETH or an ERC20
-        if (NATIVE_TOKEN == address(0)) {
+        if (NATIVE_TOKEN_L1 == address(0)) {
             require(
                 msg.value > 0,
                 "CommonBridge: the native token is ETH, msg.value must be greater than zero"
@@ -246,10 +246,14 @@ contract CommonBridge is
             value = _amount;
 
             // We lock the tokens in the bridge contract
-            IERC20(NATIVE_TOKEN).transferFrom(msg.sender, address(this), value);
+            IERC20(NATIVE_TOKEN_L1).transferFrom(
+                msg.sender,
+                address(this),
+                value
+            );
         }
 
-        deposits[ETH_TOKEN][ETH_TOKEN] += value;
+        deposits[NATIVE_TOKEN_L2][NATIVE_TOKEN_L2] += value;
 
         bytes memory callData = abi.encodeCall(
             ICommonBridgeL2.mintETH,
@@ -278,7 +282,7 @@ contract CommonBridge is
     ) external whenNotPaused {
         require(amount > 0, "CommonBridge: amount to deposit is zero");
         require(
-            tokenL1 != NATIVE_TOKEN,
+            tokenL1 != NATIVE_TOKEN_L1,
             "CommonBridge: tokenL1 is the native token address, use deposit() instead"
         );
         deposits[tokenL1][tokenL2] += amount;
@@ -377,22 +381,22 @@ contract CommonBridge is
         bytes32[] calldata withdrawalProof
     ) public override whenNotPaused {
         _claimWithdrawal(
-            ETH_TOKEN,
-            ETH_TOKEN,
+            NATIVE_TOKEN_L2,
+            NATIVE_TOKEN_L2,
             claimedAmount,
             withdrawalBatchNumber,
             withdrawalMessageId,
             withdrawalProof
         );
 
-        if (NATIVE_TOKEN == address(0)) {
+        if (NATIVE_TOKEN_L1 == address(0)) {
             (bool success, ) = payable(msg.sender).call{value: claimedAmount}(
                 ""
             );
 
             require(success, "CommonBridge: failed to send the claimed amount");
         } else {
-            IERC20(NATIVE_TOKEN).safeTransfer(msg.sender, claimedAmount);
+            IERC20(NATIVE_TOKEN_L1).safeTransfer(msg.sender, claimedAmount);
         }
     }
 
@@ -406,12 +410,12 @@ contract CommonBridge is
         bytes32[] calldata withdrawalProof
     ) public override nonReentrant whenNotPaused {
         require(
-            tokenL1 != ETH_TOKEN,
+            tokenL1 != NATIVE_TOKEN_L2,
             "CommonBridge: attempted to withdraw ETH as if it were ERC20, use claimWithdrawal()"
         );
 
         require(
-            tokenL1 != NATIVE_TOKEN,
+            tokenL1 != NATIVE_TOKEN_L1,
             "CommonBridge: attempted to withdraw the native token as if it were ERC20, use claimWithdrawal() instead"
         );
 
--- a/crates/l2/contracts/src/l2/CommonBridgeL2.sol
+++ b/crates/l2/contracts/src/l2/CommonBridgeL2.sol
@@ -12,8 +12,8 @@ contract CommonBridgeL2 is ICommonBridgeL2 {
         0x000000000000000000000000000000000000FFFE;
     address public constant BURN_ADDRESS =
         0x0000000000000000000000000000000000000000;
-    /// @notice Token address used to represent ETH
-    address public constant ETH_TOKEN =
+    /// @notice Token address used to represent the token of the L2
+    address public constant NATIVE_TOKEN_L2 =
         0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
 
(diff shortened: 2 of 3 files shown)


cmd/ethrex/l2/deployer.rs
ilitteri, line 1: Changes:
- Adds flag `--native-token-l1-address ` with `address(0)` as default (meaning ETH as native token).
- Updates `BRIDGE_INITIALIZER_SIGNATURE ` (it now requires the native token L1 address).
- Adds `NATIVE_TOKEN_DEPOSIT_SIGNATURE ` (essentially `CommonBridge`'s `deposit` function signature).
- Change initial deposits through transfers in favor of deposits through contract calls (calls to `deposit`).

crates/l2/contracts/src/l1/CommonBridge.sol
ilitteri, line 1: Changes:
- Initialization requires native token L1 address.
- `receive()` now reverts for consistency (it could work when ETH is the native token, but I opted to always revert because this could be misleading to users)
- `_deposit` is removed since it is not used more than once.
- `deposit` now handles ERC20 (as native token) deposits.
- `claimWithdrawal` now handles ERC20 (as native token) withdrawal claims.

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -194,32 +201,72 @@ contract CommonBridge is
...
-        _deposit(l2Recipient);
-    }
+    function deposit(
+        uint256 _amount,
+        address _token,
+        address l2Recipient
+    ) public payable override whenNotPaused {
+        uint256 value;
+
+        if (_token == address(0)) {
```
ilitteri: This branch is missing this check
```solidity
            require(
                _token == NATIVE_TOKEN_L1_ADDRESS,
                "CommonBridge: token address is not the native token address. Use depositERC20() instead"
            );
```

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -194,32 +201,72 @@ contract CommonBridge is
...
+        } else {
+            require(
+                msg.value == 0,
+                "CommonBridge: the native token is an ERC20, msg.value must be zero"
+            );
+
+            require(
+                _token == NATIVE_TOKEN_L1_ADDRESS,
+                "CommonBridge: token address is not the native token address. Use depositERC20() instead"
+            );
```
ilitteri: This check should be done earlier in this branch.

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -194,32 +201,72 @@ contract CommonBridge is
...
+
         _sendToL2(L2_BRIDGE_ADDRESS, sendValues);
     }
 
     receive() external payable whenNotPaused {
-        _deposit(msg.sender);
+        revert("CommonBridge: use deposit() for native token deposits");
     }
 
     function depositERC20(
```
ilitteri, line 277: We need to revert ERC20 deposit attempts when the ERC20 is the native token.

cmd/ethrex/l2/deployer.rs
```diff
@@ -977,13 +995,85 @@ async fn make_deposits(
...
+                },
+            )
+            .await?;
+            if let Err(e) = send_generic_transaction(eth_client, mint_tx, &signer).await {
+                error!(address =? signer.address(), "Failed to mint {e}");
+                continue;
+            }
+
+            let calldata = encode_calldata(
+                "approve(address,uint256)",
```
ilitteri: Make this signature a constant.

cmd/ethrex/l2/deployer.rs
```diff
@@ -977,13 +995,85 @@ async fn make_deposits(
...
+
+        let native_token_deposit_calldata =
+            encode_calldata(NATIVE_TOKEN_DEPOSIT_SIGNATURE, &calldata_values)?;
+
         let overrides = Overrides {
-            value: Some(value_to_deposit),
+            value: native_token_is_eth.then_some(value_to_deposit).or(None),
             from: Some(signer.address()),
+            nonce: if !native_token_is_eth {
+                Some(nonce + 2)
```
ilitteri, line 1080: Add a comment explaining this.

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -86,6 +87,8 @@ contract CommonBridge is
     /// @notice Deadline for the sequencer to include the transaction.
     mapping(bytes32 => uint256) public privilegedTxDeadline;
 
+    address public NATIVE_TOKEN_L1_ADDRESS;
```
ilitteri: Add documentation to this constant.

fixtures/contracts/ERC20/NativeTokenTest.bin
ilitteri, line 1: What is this contract?

ilitteri (commented):
@tomip01 I can't request changes to my own PR, but I left some comments, and the required change would be to add documentation to the new contract changes (new function arguments, new constants, etc).

cmd/ethrex/l2/deployer.rs
tomip01: Done [c3cb040](https://github.com/lambdaclass/ethrex/pull/4622/commits/c3cb0402c462e66674caaaf8d21a09616aa3fc58)

cmd/ethrex/l2/deployer.rs
tomip01: Done [e2d1899](https://github.com/lambdaclass/ethrex/pull/4622/commits/e2d1899faa9d7271754296f2f2c8e52911aea4b7)

crates/l2/contracts/src/l1/CommonBridge.sol
tomip01: Added that [here](https://github.com/lambdaclass/ethrex/pull/4622/commits/4cba04bf29c6fd5532de28548bb3e42bd1b26ae8) with  more docs and comments on the contract changes

fixtures/contracts/ERC20/NativeTokenTest.bin
tomip01: This is contract exists for testing purposes only. I think it comes very handy because we need two different contracts if we want to have an L2 with an ERC20 native token and then run the integration tests.

The integration tests deploy and make deposits of the `TestToken` ERC20. If we use this same contract as the native one, we won't be able to make deposits since we reject deposits of ERC20 being the same as the native

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -194,32 +206,77 @@ contract CommonBridge is
...
         _sendToL2(_getSenderAlias(), sendValues);
     }
 
     /// @inheritdoc ICommonBridge
-    function deposit(address l2Recipient) public payable override whenNotPaused {
-        _deposit(l2Recipient);
-    }
+    function deposit(
+        uint256 _amount,
+        address _token,
```
ricomateo: nit: since the function only allows deposits of the native token, is the `_token` parameter really needed here?

tomip01: Nice catch! Done here [5f6153a](https://github.com/lambdaclass/ethrex/pull/4622/commits/5f6153ada9595c59d554c06c61338ff84f0c707c)

cmd/ethrex/l2/deployer.rs
```diff
@@ -977,13 +998,84 @@ async fn make_deposits(
...
+        if !native_token_is_eth {
+            let mint_tx = build_generic_tx(
+                eth_client,
+                TxType::EIP1559,
+                opts.native_token_l1_address,
+                signer.address(),
+                encode_calldata("freeMint()", &[])?.into(),
+                Default::default(),
+            )
+            .await?;
```
ManuelBilbao, line 1027: This step shouldn't be needed to deploy an L2 with native token

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -86,6 +87,11 @@ contract CommonBridge is
     /// @notice Deadline for the sequencer to include the transaction.
     mapping(bytes32 => uint256) public privilegedTxDeadline;
 
+    /// @notice The L1 token address that is treated as the one to be bridged to the L2.
+    /// @dev If set to address(0), ETH is considered the native token.
+    /// Otherwise, this address is used for native token deposits and withdrawals.
+    address public NATIVE_TOKEN_L1_ADDRESS;
```
ManuelBilbao: We are in L1 here, the `_L1_ADRESS` sounds redundant

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -194,32 +206,72 @@ contract CommonBridge is
...
+
+            // We lock the tokens in the bridge contract
+            IERC20(NATIVE_TOKEN_L1_ADDRESS).transferFrom(
+                msg.sender,
+                address(this),
+                value
+            );
+        }
+
+        deposits[ETH_TOKEN][ETH_TOKEN] += value;
```
ManuelBilbao: Shouldn't this be `NATIVE_TOKEN_L1_ADDRESS`? If not, it might be better to rename `ETH_TOKEN` as it can be confussing

jrchatruc (changes requested):
PR looks good but we should make it so custom native token integration tests are run on the CI

crates/l2/contracts/src/l1/CommonBridge.sol
tomip01: Right! Done [17e43b1](https://github.com/lambdaclass/ethrex/pull/4622/commits/17e43b1137cc9f2d13c48efbd0e3bc3196b003b4)

cmd/ethrex/l2/deployer.rs
ManuelBilbao: Nvm, this is the `make_deposits`

tomip01:
> PR looks good but we should make it so custom native token integration tests are run on the CI

Done! you can see a successful run here https://github.com/lambdaclass/ethrex/actions/runs/18262702991/job/51992940038?pr=4622

crates/l2/contracts/src/l1/CommonBridge.sol
tomip01: Done here [5ba33b9](https://github.com/lambdaclass/ethrex/pull/4622/commits/5ba33b99303945465ce424dbdf7ed2b6beb5f95f)

tomip01: I actually ended changing the name here [8d91f4f](https://github.com/lambdaclass/ethrex/pull/4622/commits/8d91f4fd6f81239583f856282511335c942f674f)

crates/l2/contracts/src/l1/CommonBridge.sol
tomip01: I ended adding the suffix `_L1` to differentiate from the L2 one here [8d91f4f](https://github.com/lambdaclass/ethrex/pull/4622/commits/8d91f4fd6f81239583f856282511335c942f674f)

PR #4623, opened 2025-09-23
feat(l1, l2): new database interface

damiramirez:
**Motivation**

Our current `StoreEngine` interface has grown into a God Object with over 50 methods, making it difficult to add new database backends. Business logic is scattered across every database implementation, leading to code duplication and maintenance overhead. 

This PR addresses these issues by introducing a minimal, focused interface that makes adding new databases to ethrex trivial.

**Description**

This PR implements a fundamental redesign of the storage layer using a two-tier architecture: `Storage → Storage Backend → Database Implementation`

- `Store`
  - API exposes high-level blockchain operations
  - Underlying logic has generic optimizations like parallelizing reads and having in-memory diff-layers
  - Has knowledge of DB "shape": tables, how data is encoded/decoded.
- `StorageBackend`
  - Defines the contract for database operations with transactions
  - Supports basic DB operations for `Vec<u8>` types

New storage structure

```
crates/storage
├── api
│   ├── mod.rs
│   └── tables.rs
├── backend
│   ├── in_memory.rs
│   ├── mod.rs
│   └── rocksdb.rs
├── Cargo.toml
├── clippy.toml
├── error.rs
├── layering.rs
├── lib.rs
├── rlp.rs
├── store.rs
├── trie.rs
└── utils.rs
```

---

The new `StorageBackend` trait provides a thin abstraction over database operations:

```rust
pub type PrefixResult = Result<(Box<[u8]>, Box<[u8]>), StoreError>;

pub trait StorageBackend: Debug + Send + Sync {
    fn clear_table(&self, table: &'static str) -> Result<(), StoreError>;

    fn begin_read(&self) -> Result<Box<dyn StorageReadView + '_>, StoreError>;

    fn begin_write(&self) -> Result<Box<dyn StorageWriteBatch + 'static>, StoreError>;

    fn begin_locked(
        &self,
        table_name: &'static str,
    ) -> Result<Box<dyn StorageLockedView + 'static>, StoreError>;

    fn create_checkpoint(&self, path: &Path) -> Result<(), StoreError>;
}

pub trait StorageReadView {
    fn get(&self, table: &'static str, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError>;
    fn prefix_iterator(
        &self,
        table: &'static str,
        prefix: &[u8],
    ) -> Result<Box<dyn Iterator<Item = PrefixResult> + '_>, StoreError>;
}

pub trait StorageWriteBatch: Send {
    fn put(&mut self, table: &'static str, key: &[u8], value: &[u8]) -> Result<(), StoreError>;

    fn put_batch(
        &mut self,
        table: &'static str,
        batch: Vec<(Vec<u8>, Vec<u8>)>,
    ) -> Result<(), StoreError>;

    fn delete(&mut self, table: &'static str, key: &[u8]) -> Result<(), StoreError>;

    fn commit(&mut self) -> Result<(), StoreError>;
}

pub trait StorageLockedView: Send + Sync {
    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError>;
}
```

---
# TODO

- [ ] Benchmark performance (ethrex_benchmarks, syncing, etc)

# Further work

- Integrate with other DBs.
- Change tables to be enums instead of strings.
- Review bounds on Tx traits. Some are too restrictive for no reason, which requires synchronization in some DB implementations. We might also want to tie the lifetime of the transactions to the DB's lifetime.
- Make `StorageWriteBatch::commit` consume self. This wasn't done due to limitations related to `TrieDB`.
- Reduce the number of allocations required for using `StorageWriteBatch::put_batch` and `StorageWriteBatch::put`. Right now, they receive ownership, maybe unnecessarily. Also, review `prefix_iterator`'s return type.
- Add a `StorageReadView::get_batch` method for reading a batch of keys.
- Remove `StorageBackend::create_checkpoint` and maybe `StorageBackend::clear_table`.
- Remove `StorageLockedView`, which overlaps with `StorageReadView` in functionality.
- Reduce allocations. RocksDB gives us the option to read the data directly through `get_pinned_cf`, so we could make data return references/`Cow` to the data to account for those cases.
- Make all store methods sync, or properly implement async methods. Right now, most of them just use `tokio::task::spawn_blocking`, which should be moved to the caller side instead.

### DIFF
--- a/docs/developers/l1/storage_api.md
+++ b/docs/developers/l1/storage_api.md
@@ -10,6 +10,6 @@ This eliminates code duplication and makes adding new database backends trivial.
 
 The API differentiates between three types of database access:
 
-- Read views ([`StorageReadView`]): Read-only views of the database, with no atomicity guarantees between operations.
-- Write batches ([`StorageWriteBatch`]): Write batch functionality, with atomicity guarantees at commit time.
-- Locked views ([`StorageLockedView`]): Read-only views of a point in time (snapshots), right now it's only used during snap-sync.
+- Read views `StorageReadView`: read-only views of the database, with no atomicity guarantees between operations.
+- Write batches `StorageWriteBatch`: write batch functionality, with atomicity guarantees at commit time.
+- Locked views `StorageLockedView`: read-only views of a point in time (snapshots), right now it's only used during snap-sync.


crates/storage/backend/rocksdb.rs
```diff
@@ -0,0 +1,294 @@
...
+        opts.create_if_missing(true);
+        opts.create_missing_column_families(true);
+
+        opts.set_max_open_files(-1);
+        opts.set_max_file_opening_threads(16);
+
+        opts.set_use_fsync(false); // fdatasync
+
+        opts.enable_statistics();
+        opts.set_stats_dump_period_sec(600);
```
MegaRedHand: We'll need to resync these options with the ones in `main`. The same goes for the table options.

crates/storage/backend/rocksdb.rs
```diff
@@ -0,0 +1,294 @@
...
+        )
+        .map_err(|e| StoreError::Custom(format!("Failed to open RocksDB with all CFs: {}", e)))?;
+        Ok(Self { db: Arc::new(db) })
+    }
+
+    fn create_table(&self, _name: &'static str, _options: TableOptions) -> Result<(), StoreError> {
+        // Now we are creating the tables in the open() function
+        // Check if this function is still needed
+        Ok(())
+    }
```
MegaRedHand: I vote for removing this method. We can just assume all backends create the tables in the `open` methods.

crates/storage/backend/rocksdb.rs
```diff
@@ -0,0 +1,294 @@
...
+        let mut batch = WriteBatchWithTransaction::<true>::default();
+
+        while let Some(Ok((key, _))) = iter.next() {
+            batch.delete_cf(&cf, key);
+        }
+
+        self.db
+            .write(batch)
+            .map_err(|e| StoreError::Custom(format!("RocksDB batch write error: {}", e)))
+    }
```
MegaRedHand, line 223: This is the same as doing rw-tx -> prefix_iterator -> delete on each key. It's also only used by `clear_snap_state`. We can remove it.

crates/storage/backend/rocksdb.rs
```diff
@@ -0,0 +1,294 @@
...
+        let batch = WriteBatchWithTransaction::<true>::default();
+
+        Ok(Box::new(RocksDBRwTx {
+            db: self.db.clone(),
+            batch,
+        }))
+    }
+
+    fn begin_locked(&self, table_name: &'static str) -> Result<Box<dyn StorageLocked>, StoreError> {
+        let db = Box::leak(Box::new(self.db.clone()));
```
MegaRedHand, line 244: This comes from `main`, so no need to change it in this PR, but this leak seems unnecessary.

crates/storage/backend/rocksdb.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+/// Read-only transaction for RocksDB
+pub struct RocksDBRoTx {
+    /// Transaction
+    db: Arc<OptimisticTransactionDB<MultiThreaded>>,
+}
+
+impl StorageRoTx for RocksDBRoTx {
+    fn get(&self, table: &'static str, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError> {
```
MegaRedHand, line 273: No need to do it now, but I think we should change the table name to be an enum instead. Trading genericity for compile-time checks. We could also make it a generic argument, with the specifics of encoding/decoding the table types.

crates/storage/Cargo.toml
```diff
@@ -23,12 +23,12 @@ hex.workspace = true
 serde = { version = "1.0.203", features = ["derive"] }
 serde_json = "1.0.117"
 rocksdb = { workspace = true, optional = true }
-tokio = { workspace = true, optional = true, features = ["rt"] }
+tokio = { workspace = true, features = ["rt"] }
```
MegaRedHand, line 27: We should probably remove this dep and make every method sync.

crates/storage/trie.rs
```diff
@@ -0,0 +1,120 @@
+use crate::api::{StorageLocked, StorageRwTx};
+use ethrex_common::H256;
+use ethrex_trie::{NodeHash, TrieDB, error::TrieError};
+use std::sync::Mutex;
+
+/// StorageRwTx implementation for the TrieDB trait
+/// Wraps a transaction to allow multiple trie operations on the same transaction
+pub struct BackendTrieDB {
+    /// Read-write transaction wrapped in Mutex for interior mutability
+    tx: Mutex<Box<dyn StorageRwTx + 'static>>,
```
MegaRedHand: Review this

crates/storage/engine.rs
```diff
@@ -0,0 +1,1135 @@
...
+    rlp::{BlockBodyRLP, BlockHeaderRLP, BlockRLP},
+    store::StorageUpdates,
+    trie::{BackendTrieDB, BackendTrieDBLocked},
+    utils::{ChainDataIndex, SnapStateIndex},
+};
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub struct StoreEngine {
+    backend: Arc<dyn StorageBackend>,
```
MegaRedHand: We need to add the `TrieLayerCache` to this

crates/storage/engine.rs
```diff
@@ -0,0 +1,1135 @@
...
+        })
+        .await
+        .map_err(|e| StoreError::Custom(format!("Task panicked: {}", e)))?
+    }
+
+    /// Add a batch of blocks in a single transaction.
+    /// This will store -> BlockHeader, BlockBody, BlockTransactions, BlockNumber.
+    pub async fn add_blocks(&self, blocks: Vec<Block>) -> Result<(), StoreError> {
+        let db = self.backend.clone();
+        tokio::task::spawn_blocking(move || {
```
MegaRedHand: We need to remove this `spawn_blocking` and make the function non-async

crates/storage/engine.rs
```diff
@@ -0,0 +1,1135 @@
...
+    ) -> Result<(), StoreError> {
+        let backend = self.backend.clone();
+
+        tokio::task::spawn_blocking(move || {
+            let mut txn = backend.begin_write()?;
+            txn.put_batch(batch_ops)?;
+            txn.commit()
+        })
+        .await
+        .map_err(|e| StoreError::Custom(format!("Task panicked: {}", e)))?
```
MegaRedHand: We should remove these functions and only use `async` if the underlying DB allows it.

crates/common/trie/db.rs
```diff
@@ -12,8 +12,14 @@ pub trait TrieDB: Send + Sync {
...
         self.put_batch(vec![(key, value)])
     }
+    /// Commits any pending changes to the underlying storage
+    /// For read-only or in-memory implementations, this is a no-op
+    fn commit(&self) -> Result<(), TrieError> {
+        Ok(())
+    }
 }
 
+/// TODO: Now we have BackendTrieDB, should we remove this?
```
MegaRedHand: Yes

crates/common/trie/trie.rs
```diff
@@ -180,6 +180,9 @@ impl Trie {
             self.db.put_batch(acc)?; // we'll try to avoid calling this for every commit
         }
 
+        // Commit the underlying transaction
+        self.db.commit()?;
```
MegaRedHand, line 228: We should check if this needs to be added in some other place

crates/blockchain/smoke_test.rs
```diff
@@ -36,6 +36,8 @@ mod blockchain_integration_test {
             .await
             .unwrap();
         let retrieved_1a = store.get_block_header(1).unwrap().unwrap();
+        // Compute hash for block header
+        retrieved_1a.hash();
```
MegaRedHand: ???

crates/storage/backend/in_memory.rs
```diff
@@ -0,0 +1,216 @@
...
+type Table = BTreeMap<Vec<u8>, Vec<u8>>;
+type Database = BTreeMap<String, Table>;
+
+#[derive(Debug)]
+pub struct InMemoryBackend {
+    inner: Arc<RwLock<Database>>,
+}
+
+impl StorageBackend for InMemoryBackend {
+    fn open(_path: impl AsRef<Path>) -> Result<Self, StoreError>
```
MegaRedHand: We should move this method outside the trait.

crates/storage/store.rs
```diff
@@ -33,13 +31,11 @@ pub const MAX_SNAPSHOT_READS: usize = 100;
 
 #[derive(Debug, Clone)]
 pub struct Store {
-    pub engine: Arc<dyn StoreEngine>,
+    pub engine: Arc<StoreEngine>,
```
MegaRedHand: We should collapse both `Store` and `StoreEngine` into the same, I think. Also, all the fields in `Store` should be private.

cmd/ethrex_replay/src/cli.rs
```diff
@@ -754,34 +765,28 @@ async fn replay_no_zkvm(cache: Cache, opts: &EthrexReplayOptions) -> eyre::Resul
...
+            let mut rw_tx = in_memory_store.begin_write()?;
+            rw_tx.put_batch(storage_batch)?;
+            rw_tx.commit()?;
         }
     }
 
     // Set up store with preloaded database and the right chain config.
     let store = Store {
-        engine: Arc::new(in_memory_store),
+        engine: Arc::new(StoreEngine::new(Arc::new(in_memory_store))?),
```
MegaRedHand: This should be changed to avoid interacting with the internals of the `Store`.

crates/storage/api.rs
```diff
@@ -1,361 +1,142 @@
...
+    /// Retrieves a value by key from the locked table.
+    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError>;
+}
 
-    async fn write_account_code_batch(
-        &self,
-        account_codes: Vec<(H256, Bytes)>,
-    ) -> Result<(), StoreError>;
+/// Table names used by the storage engine.
+pub mod tables {
```
MegaRedHand: We should move this to a file.

crates/storage/api.rs
```diff
@@ -1,361 +1,142 @@
...
+    pub const HEADERS: &str = "headers";
+    pub const PENDING_BLOCKS: &str = "pending_blocks";
+    pub const TRANSACTION_LOCATIONS: &str = "transaction_locations";
+    pub const RECEIPTS: &str = "receipts";
+    pub const SNAP_STATE: &str = "snap_state";
+    pub const INVALID_CHAINS: &str = "invalid_chains";
+    pub const STATE_TRIE_NODES: &str = "state_trie_nodes";
+    pub const STORAGE_TRIE_NODES: &str = "storage_trie_nodes";
+
+    pub const LIST: [&str; 13] = [
```
MegaRedHand: Let's rename this to `TABLES`, to avoid confusion since this is being aliased to that in line 22.

crates/storage/api.rs
```diff
@@ -1,361 +1,142 @@
...
+pub trait StorageBackend: Debug + Send + Sync {
+    /// Opens a storage backend at the specified path.
+    fn open(path: impl AsRef<Path>) -> Result<Self, StoreError>
+    where
+        Self: Sized;
 
-    /// Obtain earliest block number
-    async fn get_earliest_block_number(&self) -> Result<Option<BlockNumber>, StoreError>;
+    /// Creates a new table, allowing to specify [`TableOptions`].
+    fn create_table(&self, name: &'static str, options: TableOptions) -> Result<(), StoreError>;
```
MegaRedHand: Do we want to allow arbitrary table creation? I think it's safe to assume the backend automatically configures the needed tables for `ethrex` to use. Otherwise, we'll need to find some way to allow for tuning the performance of each backend without adding too much complexity to the whole thing.

crates/storage/api.rs
```diff
@@ -1,361 +1,142 @@
...
+/// when the data comes directly from the database.
+pub type PrefixResult = Result<(Box<[u8]>, Box<[u8]>), StoreError>;
+
+/// Configuration options for table creation.
+#[derive(Debug, Clone)]
+pub struct TableOptions {
+    /// Whether the table supports duplicate keys, this means that multiple values can be stored for the same key.
+    /// This is useful for certain indexing scenarios but not supported by all backends.
+    pub dupsort: bool,
+}
```
MegaRedHand: We don't need this, if we assume the backend knows the tables `ethrex` will use. Also, `dupsort` is not really needed now, so these options would end up empty.

crates/storage/api/mod.rs
```diff
@@ -1,361 +1,142 @@
...
-    async fn write_storage_trie_nodes_batch(
-        &self,
-        storage_trie_nodes: Vec<(H256, Vec<(NodeHash, Vec<u8>)>)>,
-    ) -> Result<(), StoreError>;
+/// Locked snapshot interface for batch read operations.
+/// Provides read-only access to a specific table with a persistent snapshot.
+/// This is optimized for scenarios where many reads are performed on the same
+/// table, such as trie traversal operations.
+/// This is currently only used in snapsync stage.
+/// TODO: Check if we can remove this trait and use [`StorageRoTx`] instead.
```
MegaRedHand: Yes

crates/storage/api.rs
```diff
@@ -1,361 +1,142 @@
...
-    /// Update pending block number
-    async fn update_pending_block_number(
-        &self,
-        block_number: BlockNumber,
-    ) -> Result<(), StoreError>;
+    /// Creates a locked snapshot for a specific table.
+    ///
+    /// This provides a persistent read-only view of a single table, optimized
+    /// for batch read operations. The snapshot remains valid until dropped.
+    fn begin_locked(&self, table_name: &'static str) -> Result<Box<dyn StorageLocked>, StoreError>;
```
MegaRedHand: Maybe these three should tie `self`'s lifetime to the returned handle's.

crates/storage/api.rs
```diff
@@ -1,361 +1,142 @@
...
-        &self,
-        storage_trie_nodes: Vec<(H256, Vec<(NodeHash, Vec<u8>)>)>,
-    ) -> Result<(), StoreError>;
+/// Locked snapshot interface for batch read operations.
+/// Provides read-only access to a specific table with a persistent snapshot.
+/// This is optimized for scenarios where many reads are performed on the same
+/// table, such as trie traversal operations.
+/// This is currently only used in snapsync stage.
+/// TODO: Check if we can remove this trait and use [`StorageRoTx`] instead.
+pub trait StorageLocked: Send + Sync + 'static {
```
MegaRedHand: The `'static` here seems weird to me.

crates/storage/backend/in_memory.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        let iter = InMemoryPrefixIter {
+            results: results.into_iter(),
+        };
+        Ok(Box::new(iter))
+    }
+}
+
+impl StorageRwTx for InMemoryRwTx {
+    fn put_batch(&mut self, batch: Vec<(&str, Vec<u8>, Vec<u8>)>) -> Result<(), StoreError> {
```
MegaRedHand: This is unnecessarily generic. We should change this to be:


```suggestion
    fn put_batch(&mut self, table: &'static str, batch: Vec<(Vec<u8>, Vec<u8>)>) -> Result<(), StoreError> {
```

crates/storage/api/mod.rs
```diff
@@ -1,361 +1,142 @@
...
-        block_hash: BlockHash,
-    ) -> Result<Option<BlockNumber>, StoreError>;
-
-    /// Get the canonical block hash for a given block number.
-    fn get_canonical_block_hash_sync(
-        &self,
-        block_number: BlockNumber,
-    ) -> Result<Option<BlockHash>, StoreError>;
+    /// Commits all changes made in this transaction.
+    fn commit(&mut self) -> Result<(), StoreError>;
```
MegaRedHand, line 92: This should consume `self`.

crates/storage/api/mod.rs
```diff
@@ -1,361 +1,142 @@
...
-    /// Sets the hash of the last header downloaded during a snap sync
-    async fn set_header_download_checkpoint(&self, block_hash: BlockHash)
+    /// Stores multiple key-value pairs in the specified table within the transaction.
+    fn put_batch(&mut self, batch: Vec<(&'static str, Vec<u8>, Vec<u8>)>)
     -> Result<(), StoreError>;
 
-    /// Gets the hash of the last header downloaded during a snap sync
-    async fn get_header_download_checkpoint(&self) -> Result<Option<BlockHash>, StoreError>;
+    /// Removes a key-value pair from the specified table.
+    fn delete(&mut self, table: &'static str, key: &[u8]) -> Result<(), StoreError>;
```
MegaRedHand, line 89: I think it makes sense to have a `delete_batch` function too. Maybe even have `put` and `delete` be batch functions here, with _maybe_ a default `put_one`/`delete_one`.

crates/storage/api.rs
MegaRedHand: Update: this was done due to the `Trie` functions requiring it.

crates/storage/api.rs
MegaRedHand: Update: this was done due to the `Trie` functions requiring it.

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Uses Box<[u8]> instead of Vec<u8> to avoid unnecessary allocations
+/// when the data comes directly from the database.
+pub type PrefixResult = Result<(Box<[u8]>, Box<[u8]>), StoreError>;
+
+/// This trait provides a minimal set of operations required from a database backend.
+/// Implementations should focus on providing efficient access to the underlying storage
+/// without implementing business logic.
+pub trait StorageBackend: Debug + Send + Sync {
+    /// Removes all data from the specified table.
+    fn clear_table(&self, table: &'static str) -> Result<(), StoreError>;
```
MegaRedHand, line 34: We should change the tables to be an enum.

crates/storage/trie.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    fn make_key(&self, path: Nibbles) -> Vec<u8> {
+        apply_prefix(self.address_prefix, path).as_ref().to_vec()
+    }
+}
+
+impl TrieDB for BackendTrieDB {
+    fn flatkeyvalue_computed(&self, key: Nibbles) -> bool {
+        // TODO: this breaks TrieWrapper
+        // let key = apply_prefix(self.address_prefix, key);
```
MegaRedHand: We should check this. Just for context, this is also on the main branch.

MegaRedHand: It actually doesn't break `TrieWrapper`. The prefix is `None` in the cases this is called from it.

I changed it in [703e29e](https://github.com/lambdaclass/ethrex/pull/4623/commits/703e29ee79320bffe790be37d8a7810bcdde56a8)

crates/blockchain/smoke_test.rs
MegaRedHand: This is due to two reasons:
1. Header comparison is broken, due to the derived comparison including the `OnceCell`, which could be uninitialized.
2. With the switch in the API, now the in-memory engine _also_ RLP-encodes the headers, discarding its hash. The version in main doesn't do this, which conserves the computed hash.

Number 2 was fixed in [d5f28b1](https://github.com/lambdaclass/ethrex/pull/4623/commits/d5f28b1f0f70a9b7949cbbffc8c62b7fe1e4eb71), which also avoids hash recomputation with _any_ backend, and number 1 was left in an issue: #5125.

crates/common/rlp/decode.rs
```diff
@@ -492,13 +492,31 @@ pub fn is_encoded_as_bytes(rlp: &[u8]) -> Result<bool, RLPDecodeError> {
...
 pub fn get_rlp_bytes_item_payload(rlp: &[u8]) -> Result<&[u8], RLPDecodeError> {
     let prefix = rlp.first().ok_or(RLPDecodeError::InvalidLength)?;
     let offset: usize = (prefix - 0xb8 + 1).into();
     rlp.get(offset + 1..).ok_or(RLPDecodeError::InvalidLength)
 }
 
+/// Receives an RLP item (either string (prefix between 0x80 and 0xB8 or bytes (prefix between 0xb8 and 0xbf)) and returns its payload
+pub fn get_rlp_item_payload(rlp: &[u8]) -> Result<&[u8], RLPDecodeError> {
+    let prefix = rlp.first().ok_or(RLPDecodeError::InvalidLength)?;
+    if (RLP_NULL..=RLP_NULL+56).contains(prefix) {
```
MegaRedHand: This should be `RLP_NULL..=RLP_NULL+55` (`0x80..=0xB7`), I think. Also, I think this whole `if`-chain can be expressed better in a `match`

MegaRedHand: Maybe

```rust
/// Receives an RLP item (either string (prefix between 0x80 and 0xB8) or bytes (prefix between 0xb8 and 0xbf)) and returns its payload
pub fn get_rlp_item_payload(rlp: &[u8]) -> Result<&[u8], RLPDecodeError> {
    let (is_list, item, trailing) = decode_rlp_item(rlp)?;
    if is_list {
        return Err(RLPDecodeError::UnexpectedList);
    }
    if !trailing.is_empty() {
        return Err(RLPDecodeError::MalformedData);
    }
    Ok(item)
}
```

MegaRedHand: I think `P2PTransaction` has the same problem

MegaRedHand: I think RLP bytes are being misinterpreted here. Bytes are anything that isn't a list.

fmoletta: Addressed in #5139

crates/networking/rpc/eth/gas_price.rs
```diff
@@ -83,6 +83,8 @@ mod tests {
         assert_eq!(parsed_result, 2 * BASE_PRICE_IN_WEI);
     }
 
+    // FIXME
+    #[ignore = "fails due to the tx RLP encoding not working with empty signatures"]
```
MegaRedHand: These should be fixed by #5139

crates/storage/api/mod.rs
MegaRedHand: The tries need this to be a mutable reference to avoid moving the trait object. We can try to change it after merging.

crates/networking/rpc/eth/gas_price.rs
MegaRedHand: Fixed

crates/storage/backend/rocksdb.rs
iovoid: Also used by the FKV generato, but could be removed there too by adding the "do not insert fkvs below last_written" logic to every method that writes to the DB, or ensuring setup (genesis in fullsync, or snap in snapsync) writes FKVs.

It's probably ok to leave as a task for the future.

crates/storage/api/mod.rs
ilitteri: +1

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Creates a locked snapshot for a specific table.
+    ///
+    /// This provides a persistent read-only view of a single table, optimized
+    /// for batch read operations. The snapshot remains valid until dropped.
+    fn begin_locked(
+        &self,
+        table_name: &'static str,
+    ) -> Result<Box<dyn StorageLocked + 'static>, StoreError>;
+
+    // TODO: remove this and provide historic data via diff-layers
```
ilitteri, line 51: I'm not sure this will work, or if this is what we want for the L2. We'd need to have as many diff layers as state diffs in an L2 batch, and this cannot be fixed (it may vary a lot in the L2 configuration, even setting a high number of diff layers may not be enough).

crates/storage/Cargo.toml
```diff
@@ -24,15 +24,14 @@ serde.workspace = true
 serde_json = "1.0.117"
 rocksdb = { workspace = true, optional = true }
 rustc-hash.workspace = true
-tokio = { workspace = true, optional = true, features = ["rt"] }
-bincode = "1.3.3"
+tokio = { workspace = true, features = ["rt"] }
```
ilitteri: Is this dep still needed?

crates/storage/api/mod.rs
MegaRedHand: We'll need to persist backwards diff-layers in the DB someday, since reorgs can get really deep during non-finality, and we need a way to "go back" to previous states. Once we implement that functionality, we can use that functionality in the sequencer to avoid depending on DB-specific checkpoint functionality. The TODO is just to mark that method as temporary until we add historical diff-layers.

ilitteri (approved):

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+//! Rather than implementing business logic in each database backend, this API
+//! provides low-level primitives that higher-level code can build upon.
+//! This eliminates code duplication and makes adding new database backends trivial.
+//!
+//! The API differentiates between three types of database access:
+//!
+//! - Read-only transactions ([`StorageRoTx`])
+//! - Read-write transactions ([`StorageRwTx`])
+//! - Locked snapshots ([`StorageLocked`]): Persistent read-only views, righ now it's
+//!   only used in snapsync stage.
```
pablodeymo: We should include this documentation in the `developers` section of the book (`docs/` folder).

MegaRedHand: OK. I'll add it in another PR

MegaRedHand: Nvm. I can add it to this PR

MegaRedHand: Done

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+//! - Simple: Avoids type-system complexity and focuses on core functionality
+//!
+//! Rather than implementing business logic in each database backend, this API
+//! provides low-level primitives that higher-level code can build upon.
+//! This eliminates code duplication and makes adding new database backends trivial.
+//!
+//! The API differentiates between three types of database access:
+//!
+//! - Read transactions ([`StorageReadTx`]): Read-only views of the database,
+//!   with no atomicity guarantees between operations.
```
Oppen: If it has no atomicity between operations then it's a view, not a transaction. The name should reflect that.

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+//! - Write transactions ([`StorageWriteTx`]): Write batch functionality, with
+//!   atomicity guarantees at commit time.
+//! - Locked snapshots ([`StorageLocked`]): Read-only views of a point in time, right now it's
+//!   only used in snapsync stage.
```
Oppen: IMO this should be how most things during block execution should work.

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use crate::error::StoreError;
+use std::{fmt::Debug, path::Path};
+
+pub mod tables;
+
+/// Type alias for the result of a prefix iterator.
+/// Uses Box<[u8]> instead of Vec<u8> to avoid unnecessary allocations
```
Oppen: Doesn't `Box` cause an allocation too?

crates/storage/api/mod.rs
Oppen: +2

crates/storage/api/mod.rs
Oppen: I wouldn't persist those in the DB itself but to a different kind of file, but we can cross that bridge once we get to it.

crates/storage/api/mod.rs
Oppen: What I see is more reasons why the trie shouldn't be writing to the DB by itself.
But yes, we can do that at a later time, perfect is the enemy of good.

crates/storage/api/mod.rs
MegaRedHand: Indeed. Removed this part.

crates/storage/api/mod.rs
MegaRedHand: I think so too. I think the reason we have the "locked" version is due to an old version of the healing process. We can probably avoid the transactionality entirely in all cases.

crates/storage/api/mod.rs
MegaRedHand: I'll rename it, along with `StorageWriteTx`, since it doesn't _really_ work like a transaction.

What do you think of `StorageReadView` and `StorageWriteBatch`? And `StorageLocked` could be `StorageLockedView`/`StorageSnapshot`

MegaRedHand: Done

crates/storage/api/mod.rs
Oppen: Ah, but that's the opposite of what I'm implying. I think tries should be created based on static snapshots, rather than allowing changes under one's feet. How we implement that is indifferent, but I wouldn't want to deal with trie instance invalidation during e.g. iteration.

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+//! This eliminates code duplication and makes adding new database backends trivial.
+//!
+//! The API differentiates between three types of database access:
+//!
+//! - Read views ([`StorageReadView`]): Read-only views of the database,
+//!   with no atomicity guarantees between operations.
+//! - Write batches ([`StorageWriteBatch`]): Write batch functionality, with
+//!   atomicity guarantees at commit time.
+//! - Locked views ([`StorageLockedView`]): Read-only views of a point in time (snapshots), right now it's
+//!   only used during snap-sync.
```
Oppen: Nit:
```suggestion
//! - Read views ([`StorageReadView`]): read-only views of the database,
//!   with no atomicity guarantees between operations.
//! - Write batches ([`StorageWriteBatch`]): write batch functionality, with
//!   atomicity guarantees at commit time.
//! - Locked views ([`StorageLockedView`]): read-only views of a point in time (snapshots), right now it's
//!   only used during snap-sync.
```

crates/storage/api/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+/// without implementing business logic.
+pub trait StorageBackend: Debug + Send + Sync {
+    /// Removes all data from the specified table.
+    fn clear_table(&self, table: &'static str) -> Result<(), StoreError>;
+
+    /// Begins a new read transaction.
+    fn begin_read(&self) -> Result<Box<dyn StorageReadView + '_>, StoreError>;
+
+    /// Begins a new write transaction.
+    fn begin_write(&self) -> Result<Box<dyn StorageWriteBatch + 'static>, StoreError>;
```
Oppen: Nit:
```suggestion
    /// Opens a new read view.
    fn begin_read(&self) -> Result<Box<dyn StorageReadView + '_>, StoreError>;

    /// Creates a new write batch.
    fn begin_write(&self) -> Result<Box<dyn StorageWriteBatch + 'static>, StoreError>;
```

crates/storage/api/mod.rs
Oppen: All write functions should be batched, yes. Single-element operations are probably better off as helpers around the batch interfaces, if at all. But we can leave it for a later version.

crates/storage/backend/in_memory.rs
```diff
@@ -0,0 +1,181 @@
...
+use std::path::Path;
+use std::sync::{Arc, RwLock};
+
+type Table = BTreeMap<Vec<u8>, Vec<u8>>;
+type Database = BTreeMap<&'static str, Table>;
+
+#[derive(Debug)]
+pub struct InMemoryBackend {
+    inner: Arc<RwLock<Database>>,
+}
```
Oppen, line 15: Not sure if `InMemoryBackend` will be used in parallel real-world scenarios, but in case it is, it might be better to have per-table locks rather than a single global one, like this:
```suggestion
type Table = BTreeMap<Vec<u8>, Vec<u8>>;
type Database = BTreeMap<&'static str, RwLock<Table>>;

#[derive(Debug)]
pub struct InMemoryBackend {
    inner: Arc<Database>,
}
```

It has the complexity that the write batch needs to keep track of modified tables and lock them all before writing (to hold the atomicity of the batch).

crates/storage/backend/in_memory.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        if let Some(table_ref) = db.get_mut(table) {
+            table_ref.remove(key);
+        }
+        Ok(())
+    }
+
+    fn commit(&mut self) -> Result<(), StoreError> {
+        // We don't need to commit for in-memory backend
+        // FIXME: in-memory writes aren't atomic
```
Oppen, line 177: So, we do need to commit, we just don't just yet...

crates/storage/backend/rocksdb.rs
Oppen: Shouldn't this be using something like `drop_cf` and recreating the cf or something? Or at least calling `delete_range`.
If the issue is just avoiding the config repetition, maybe have a static table or something with it.

crates/storage/backend/rocksdb.rs
Oppen: We should be able to just use the pointer returned by Arc or a reference, right?

crates/storage/Cargo.toml
Oppen: Even without making everything sync, we might be better off having an actor accept messages and the async methods just awaiting on message passing, without having to bring the Tokio runtime to the store.

crates/storage/store.rs
```diff
@@ -15,33 +31,118 @@ use ethrex_common::{
...
+            if let Some((_, code)) = cache.pop_lru() {
+                self.cache_size
+                    .fetch_sub(code.size() as u64, std::sync::atomic::Ordering::SeqCst);
+                current_size = self.cache_size.load(std::sync::atomic::Ordering::SeqCst);
+            } else {
+                break;
+            }
+        }
+
+        cache.get_or_insert(code.hash, || code.clone());
```
Oppen, line 119: 1. All of this is happening with an exclusive lock taken, which feels like a lot;
2. `fetch_add` already returns the previous value, so it's a bit pointless to read it again just with SeqCst rather than adding again, specially considering with the lock taken nobody else should be modifying the value.

crates/storage/store.rs
```diff
@@ -15,33 +31,118 @@ use ethrex_common::{
...
 pub struct Store {
-    pub engine: Arc<dyn StoreEngine>,
-    pub chain_config: ChainConfig,
+    db_path: PathBuf,
+    backend: Arc<dyn StorageBackend>,
+    chain_config: ChainConfig,
+    trie_cache: Arc<Mutex<Arc<TrieLayerCache>>>,
+    flatkeyvalue_control_tx: std::sync::mpsc::SyncSender<FKVGeneratorControlMessage>,
+    trie_update_worker_tx: std::sync::mpsc::SyncSender<TrieUpdate>,
     /// Keeps the latest canonical block hash
```
Oppen:
```suggestion
    /// Keeps the latest canonical block header
```

crates/storage/store.rs
```diff
@@ -75,36 +176,1307 @@ pub struct AccountUpdatesList {
...
+    /// Add account code
+    pub async fn add_account_code(&self, code: Code) -> Result<(), StoreError> {
+        let hash_key = code.hash.0.to_vec();
+        let mut buf = Vec::with_capacity(
+            6 + code.bytecode.len()
+                + code
+                    .jump_targets
+                    .iter()
+                    .map(std::mem::size_of_val)
+                    .sum::<usize>(),
```
Oppen:
```suggestion
                + std::mem::size_of_val(code.jump_targets.as_slice())
```

crates/storage/store.rs
```diff
@@ -75,36 +176,1307 @@ pub struct AccountUpdatesList {
...
+        head_hash: BlockHash,
+        safe: Option<BlockNumber>,
+        finalized: Option<BlockNumber>,
+    ) -> Result<(), StoreError> {
+        let latest = self.load_latest_block_number().await?.unwrap_or(0);
+        let db = self.backend.clone();
+        tokio::task::spawn_blocking(move || {
+            let mut txn = db.begin_write()?;
+
+            if let Some(canonical_blocks) = new_canonical_blocks {
```
Oppen, line 823: Is the Option really necessary here? An empty vec has the same effect.

crates/storage/api/mod.rs
MegaRedHand: Good catch

crates/storage/backend/in_memory.rs
MegaRedHand: You're right. Removed the comment.

crates/storage/backend/in_memory.rs
MegaRedHand: I opened an issue for this: #5611

crates/storage/store.rs
MegaRedHand: In my defense, this was already the case in main 🙈 
I opened #5612 to track this

crates/storage/store.rs
MegaRedHand: You're right. Opened #5613 for this

crates/storage/store.rs
MegaRedHand: Added in e0a10f419

PR #4624, opened 2025-09-23
fix(l2): add script to read address from deployer in docker compose

tomip01:
**Motivation**

Running `docker compose up` was broken running on `crates/l2`. The problem is that the ethrex l2 container reads the `.env` at the moment of running the `docker compose up` and miss the updates on the file that writes the deployer. 

**Description**

- Deployer now creates the empty `.env` in the shared volume that the `ethrex_l2` container will read
- Add a parsing and exporting the env vars in the `ethrex_l2` container as a previous step to running the sequencer.
- Add a copy step on CI jobs to retrieve the addresses used in the deployment since they are no longer on the host machine
- Change base image of the `Dockerfile` to use a Debian as we need to use common unix binaries

**How to Test**

On main this should failed (check that you have a fresh start, removing `cmd/.env`)

```
cd crates/l2
docker compose up
```

### DIFF
--- a/crates/l2/docker-compose-l2-store.overrides.yaml
+++ b/crates/l2/docker-compose-l2-store.overrides.yaml
@@ -4,7 +4,6 @@ services:
     volumes:
       - ./store:/store
     command: >
-      l2
       --network /genesis/l2.json
       --http.addr 0.0.0.0
       --http.port 1729


Dockerfile
```diff
@@ -58,9 +58,11 @@ RUN cargo build --release $BUILD_FLAGS
 # --- Final Image ---
 # Copy the ethrex binary into a minimalist image to reduce bloat size.
 # This image must have glibc and libssl
-FROM gcr.io/distroless/cc-debian12
+FROM debian:12-slim
 WORKDIR /usr/local/bin
 
+RUN apt-get update && apt-get install -y --no-install-recommends libssl3
```
tomip01, line 64: this is because we need ssl that it is not included on the base image

crates/l2/docker-compose.yaml
```diff
@@ -51,6 +54,10 @@ services:
       - COMPILE_CONTRACTS=true
     depends_on:
       - ethrex_l1
+    entrypoint:
+      - /bin/bash
+      - -c
+      - touch /env/.env; exec ./ethrex "$0" "$@"
```
ManuelBilbao: I'd put "l2 deploy" as part of the entrypoint, as the contract deployer is always suposed to run it. This way, you can run:

```
docker compose run contract_deployer <custom_deployer_options>
```

instead of 

```
docker compose run contract_deployer l2 deploy <custom_deployer_options>
```

Also, the `exec` is not needed.

```
    entrypoint:
      - /bin/bash
      - -c
      - touch /env/.env; exec ./ethrex "$0" "$@"
    command: >
      --randomize-contract-deployment
````

File: crates/l2/docker-compose.yaml [hunk 3]
```diff
@@ -76,11 +83,13 @@ services:
...
-    env_file:
-      - ../../cmd/.env
     volumes:
       - ../../fixtures/genesis/l2.json:/genesis/l2.json
-      - ../../cmd/.env:/.env:ro
+      - env:/env/
+    entrypoint:
+      - /bin/bash
+      - -c
+      - export $$(xargs < /env/.env); exec ./ethrex "$$0" "$$@"
```
ManuelBilbao: Same here about `l2` subcommand and `exec`. Also, it's weird the double $ here but not in the previous command

tomip01: Remove the extra $ here [6921165](https://github.com/lambdaclass/ethrex/pull/4624/commits/692116519942adc2009eda3e8e0e711d2f3b29e9)

tomip01: Done the remaining here [e9ec17b](https://github.com/lambdaclass/ethrex/pull/4624/commits/e9ec17bdde41a83a832fc23c79a0b415e068cf63)

crates/l2/docker-compose.yaml
tomip01: Done [e9ec17b](https://github.com/lambdaclass/ethrex/pull/4624/commits/e9ec17bdde41a83a832fc23c79a0b415e068cf63)

ManuelBilbao: I think the "$@" is still needed. By default the command is just one argument, but the user may want to add more

ManuelBilbao (approved):

LeanSerra (approved):

PR #4625, opened 2025-09-23
feat(l1): implement eth-config

### LINKED ISSUE #3711: Implement `eth_config` endpoint
https://eips.ethereum.org/EIPS/eip-7910

Acceptance criteria:
Pass these tests: https://eest.ethereum.org/main/running_tests/execute/eth_config/

SDartayet:
**Motivation**

Achieving full Fusaka implementation before launch.

**Description**

This PR implements [EIP-7910](https://eips.ethereum.org/EIPS/eip-7910). To summarize, it consists of a new eth method, that returns the fork ID, chain ID, activation time, active precompiles and system contracts for the currently activated fork, and when applicable, the next and latest fork's the client has in its configuration. 

As part of implementing the precompiles side of things, the logic of how we check an address is a precompile was refactored, introducing a method that returns the precompiles per fork so as to more closely tie the addresses we return in the `eth-config` method and the way we check a precompile is active when executing it.

To test this PR, you can follow the instructions in [this page](https://eest.ethereum.org/main/running_tests/execute/eth_config/). Keep in mind the tests require to have whichever chain they're being run on fully synced.

Closes #3711

### DIFF
--- a/.github/workflows/pr-main_l2.yaml
+++ b/.github/workflows/pr-main_l2.yaml
@@ -155,6 +155,11 @@ jobs:
           ETHREX_DEPLOYER_RISC0_CONTRACT_ADDRESS=0x00000000000000000000000000000000000000aa \
           ETHREX_L2_VALIDIUM=${{ matrix.validium }} \
           docker compose up contract_deployer
+      
+      - name: Copy env to host
+        run: |
+          docker cp contract_deployer:/env/.env cmd/.env
+          cat cmd/.env
 
       - name: Start Sequencer
         run: |
@@ -220,6 +225,11 @@ jobs:
           ETHREX_DEPLOYER_DEPLOY_BASED_CONTRACTS=true \
           COMPILE_CONTRACTS=true \
           docker compose up contract_deployer
+      
+      - name: Copy env to host
+        run: |
+          docker cp contract_deployer:/env/.env cmd/.env
+          cat cmd/.env
 
       - name: Install rex
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-10-01
+
+- Change remaining_gas to i64, improving performance in gas cost calculations [#4684](https://github.com/lambdaclass/ethrex/pull/4684)
+
 ### 2025-09-29
 
 - Remove `OpcodeResult` to improve tight loops of lightweight opcodes [#4650](https://github.com/lambdaclass/ethrex/pull/4650)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3985,6 +3985,7 @@ dependencies = [
  "rkyv",
  "serde",
  "serde_json",
+ "sp1-prover",
  "sp1-recursion-gnark-ffi",
  "sp1-sdk",
  "thiserror 2.0.16",
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,8 +58,10 @@ RUN cargo build --release $BUILD_FLAGS
 # --- Final Image ---
 # Copy the ethrex binary into a minimalist image to reduce bloat size.
 # This image must have glibc and libssl
-FROM gcr.io/distroless/cc-debian12
+FROM debian:12-slim
 WORKDIR /usr/local/bin
+
+RUN apt-get update && apt-get install -y --no-install-recommends libssl3
 
 COPY cmd/ethrex/networks ./cmd/ethrex/networks
 COPY --from=builder /ethrex/target/release/ethrex .
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -110,6 +110,8 @@ pub fn open_store(datadir: &Path) -> Store {
                 panic!("Specify the desired database engine.");
             }
         };
+        #[cfg(feature = "metrics")]
+        ethrex_metrics::metrics_process::set_datadir_path(datadir.to_path_buf());
         Store::new(datadir, engine_type).expect("Failed to create Store")
     }
 }
--- a/cmd/ethrex/l2/options.rs
+++ b/cmd/ethrex/l2/options.rs
@@ -851,6 +851,16 @@ pub struct ProverClientOptions {
         help_heading = "Prover client options"
     )]
     pub aligned: bool,
+    #[cfg(all(feature = "sp1", feature = "gpu"))]
+    #[arg(
+        long,
+        default_value = "None",
+        value_name = "URL",
+        env = "ETHREX_SP1_SERVER",
+        help = "Url to the moongate server to use when using sp1 backend",
+        help_heading = "Prover client options"
+    )]
+    pub sp1_server: Option<Url>,
 }
 
 impl From<ProverClientOptions> for ProverConfig {
@@ -860,6 +870,8 @@ impl From<ProverClientOptions> for ProverConfig {
             proof_coordinators: config.proof_coordinator_endpoints,
             proving_time_ms: config.proving_time_ms,
             aligned_mode: config.aligned,
+            #[cfg(all(feature = "sp1", feature = "gpu"))]
+            sp1_server: config.sp1_server,
         }
     }
 }
@@ -874,6 +886,8 @@ impl Default for ProverClientOptions {
             log_level: Level::INFO,
             aligned: false,
             backend: Backend::Exec,
+            #[cfg(all(feature = "sp1", feature = "gpu"))]
+            sp1_server: None,
         }
     }
 }
--- a/crates/blockchain/metrics/metrics_process.rs
+++ b/crates/blockchain/metrics/metrics_process.rs
@@ -1,9 +1,14 @@
-use prometheus::{Encoder, Registry, TextEncoder};
-use std::sync::LazyLock;
+use prometheus::{Encoder, IntGauge, Registry, TextEncoder};
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+    sync::{LazyLock, OnceLock},
+};
 
 use crate::MetricsError;
 
 pub static METRICS_PROCESS: LazyLock<MetricsProcess> = LazyLock::new(MetricsProcess::default);
+static DATADIR_PATH: OnceLock<PathBuf> = OnceLock::new();
 
 #[derive(Debug, Clone)]
 pub struct MetricsProcess;
@@ -37,6 +42,20 @@ impl MetricsProcess {
                 })?;
         }
 
+        if let Some(path) = DATADIR_PATH.get() {
+            if let Ok(size) = directory_size(path) {
+                let gauge = IntGauge::new(
+                    "datadir_size_bytes",
+                    "Total size in bytes consumed by the configured datadir.",
+                )
+                .map_err(|e| MetricsError::PrometheusErr(e.to_string()))?;
+                let clamped = size.min(i64::MAX as u64);
+                gauge.set(clamped as i64);
+                r.register(Box::new(gauge))
+                    .map_err(|e| MetricsError::PrometheusErr(e.to_string()))?;
+            }
+        }
+
         let encoder = TextEncoder::new();
         let metric_families = r.gather();
 
@@ -48,4 +67,38 @@ impl MetricsProcess {
         let res = String::from_utf8(buffer)?;
         Ok(res)
     }
+}
+
+pub fn set_datadir_path(path: PathBuf) {
+    let _ = DATADIR_PATH.set(path);
+}
+
+fn directory_size(root: &Path) -> io::Result<u64> {
+    let mut total = 0;
+    let mut stack = vec![root.to_path_buf()];
+
+    while let Some(path) = stack.pop() {
+        let entries = match fs::read_dir(&path) {
+            Ok(entries) => entries,
+            Err(err) if err.kind() == io::ErrorKind::NotFound => continue,
+            Err(err) => return Err(err),
+        };
+
+        for entry in entries {
+            let entry = entry?;
+            let metadata = match entry.metadata() {
+                Ok(metadata) => metadata,
+                Err(err) if err.kind() == io::ErrorKind::NotFound => continue,
+                Err(err) => return Err(err),
+            };
+
+            if metadata.is_dir() {
+                stack.push(entry.path());
+            } else {
+                total += metadata.len();
+            }
+        }
+    }
+
+    Ok(total)
 }
--- a/crates/l2/docker-compose-l2-store.overrides.yaml
+++ b/crates/l2/docker-compose-l2-store.overrides.yaml
@@ -4,7 +4,6 @@ services:
     volumes:
       - ./store:/store
     command: >
-      l2
       --network /genesis/l2.json
       --http.addr 0.0.0.0
       --http.port 1729
--- a/crates/l2/docker-compose-l2-tdx.yaml
+++ b/crates/l2/docker-compose-l2-tdx.yaml
@@ -3,10 +3,10 @@ services:
   ethrex_l2:
     volumes:
       - ./tee/contracts/automata-dcap-qpl/automata-dcap-qpl-tool/target/release/automata-dcap-qpl-tool:/automata-dcap-qpl-tool:ro,exec
+      - ../../cmd/.env:/env/.env
     environment:
       - ETHREX_PROOF_COORDINATOR_TDX_PRIVATE_KEY=${ETHREX_PROOF_COORDINATOR_TDX_PRIVATE_KEY:-0x39725efee3fb28614de3bacaffe4cc4bd8c436257e2c8bb887c4b5c4be45e76d}
     command: >
-      l2
       --network /genesis/l2.json
       --http.addr 0.0.0.0
       --http.port 1729
@@ -17,3 +17,6 @@ services:
       --proof-coordinator.l1-private-key 0x39725efee3fb28614de3bacaffe4cc4bd8c436257e2c8bb887c4b5c4be45e76d
       --proof-coordinator.qpl-tool-path /automata-dcap-qpl-tool
       --no-monitor
+
+volumes:
+  env: null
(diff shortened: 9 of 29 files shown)


crates/networking/rpc/rpc.rs
```diff
@@ -560,9 +562,9 @@ mod tests {
         let result = map_http_requests(&request, context).await;
         let rpc_response = rpc_response(request.id, result).unwrap();
         let blob_schedule = serde_json::json!({
-            "cancun": { "target": 3, "max": 6, "baseFeeUpdateFraction": 3338477 },
-            "prague": { "target": 6, "max": 9, "baseFeeUpdateFraction": 5007716 },
-            "osaka": { "target": 6, "max": 9, "baseFeeUpdateFraction": 5007716 },
+            "cancun": { "baseFeeUpdateFraction": 3338477, "max": 6, "target": 3 },
+            "prague": { "baseFeeUpdateFraction": 5007716, "max": 9, "target": 6 },
+            "osaka": { "baseFeeUpdateFraction": 5007716, "max": 9, "target": 6 },
```
SDartayet: Related to the reordering of fields

crates/common/types/genesis.rs
```diff
@@ -113,9 +113,9 @@ impl TryFrom<&Path> for Genesis {
 )]
 #[serde(rename_all = "camelCase")]
 pub struct ForkBlobSchedule {
-    pub target: u32,
-    pub max: u32,
     pub base_fee_update_fraction: u64,
+    pub max: u32,
+    pub target: u32,
```
SDartayet, line 116: I reordered the fields so they'd get serialized in alphabetical order, since the spec for the methods requires them to be so

crates/vm/backends/revm/helpers.rs
```diff
@@ -122,5 +122,6 @@ pub fn fork_to_spec_id(fork: Fork) -> SpecId {
         Fork::Cancun => SpecId::CANCUN,
         Fork::Prague => SpecId::PRAGUE,
         Fork::Osaka => SpecId::OSAKA,
+        _ => SpecId::LATEST,
```
SDartayet: I wasn't sure what to do with the BPO forks given this version of REVM doesn't account for them, so I included this as a placeholder. If anyone has a better idea please let me know

MegaRedHand: We can do the same as in https://github.com/lambdaclass/ethrex/pull/4643#discussion_r2379868886 and map it to the "real" fork right before BPO forks (in this case Osaka).

crates/vm/levm/src/precompiles.rs
```diff
@@ -203,25 +108,10 @@ pub const G1_POINT_AT_INFINITY: [u8; 128] = [0_u8; 128];
...
-        return false;
-    }
-
-    PRECOMPILES.contains(address)
-        || PRECOMPILES_POST_CANCUN.contains(address)
-        || address == &P256_VERIFICATION_ADDRESS
+    (matches!(vm_type, VMType::L2) && *address == P256_VERIFICATION.address)
+        || precompiles_for_fork(fork)
+            .iter()
+            .any(|precompile| precompile.address == *address)
```
MegaRedHand: We should benchmark this and check that it doesn't tank our performance. If it ends up being an issue, I'm sure we can pre-generate a lookup table for the precompiles or something.

cc @Arkenan

crates/vm/backends/revm/helpers.rs
SDartayet: Might be worth to just wait for that PR to be merged before this one TBH, since the share a bunch of changes

crates/vm/constants.rs
MegaRedHand, line 1: We should remove this file

crates/common/types/precompile.rs
MegaRedHand, line 1: Can we keep this in the VM crate? The RPC crate already depends on the VM, so keeping it there shouldn't bother much.

crates/common/system_contracts.rs
MegaRedHand, line 1: Same for this. I think it makes sense to keep it in `ethrex-vm`

crates/common/types/genesis.rs
```diff
@@ -437,6 +447,97 @@ impl ChainConfig {
...
+            Some(Fork::BPO5)
+        } else if self.is_bpo3_activated(block_timestamp) && self.bpo4_time.is_some() {
+            Some(Fork::BPO4)
+        } else if self.is_bpo2_activated(block_timestamp) && self.bpo3_time.is_some() {
+            Some(Fork::BPO3)
+        } else if self.is_bpo1_activated(block_timestamp) && self.bpo2_time.is_some() {
+            Some(Fork::BPO2)
+        } else if self.is_osaka_activated(block_timestamp) && self.bpo1_time.is_some() {
+            Some(Fork::BPO1)
+        } else if self.is_prague_activated(block_timestamp) && self.osaka_time.is_some() {
```
MegaRedHand, line 476: Is it possible for BPO3, for example, to be configured and BPO1 and BPO2 to not be? I believe this won't work in that case, so we should check that out, or add a sanity check and warning for those cases.

crates/networking/rpc/eth/client.rs
```diff
@@ -62,3 +71,121 @@ impl RpcHandler for Syncing {
...
+        );
+        system_contracts.insert(
+            "HISTORY_STORAGE_ADDRESS".to_string(),
+            *HISTORY_STORAGE_ADDRESS,
+        );
+        system_contracts.insert(
+            "WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS".to_string(),
+            *WITHDRAWAL_REQUEST_PREDEPLOY_ADDRESS,
+        );
+    }
```
MegaRedHand, line 163: We should add a `system_contracts_for_fork`, similar to the precompiles.

crates/networking/rpc/eth/client.rs
```diff
@@ -62,3 +71,121 @@ impl RpcHandler for Syncing {
...
+        ForkId::new(
+            chain_config,
+            context
+                .storage
+                .get_block_by_number(0)
+                .await?
+                .expect("Failed to get genesis block. This should not happen.")
+                .header,
+            timestamp,
+            0,
```
MegaRedHand: Why are we passing a 0 here?

crates/networking/rpc/eth/client.rs
```diff
@@ -62,3 +71,121 @@ impl RpcHandler for Syncing {
...
+    let activation_time = chain_config.get_activation_timestamp_for_fork(fork);
+    let fork_id = if let Some(timestamp) = activation_time {
+        ForkId::new(
+            chain_config,
+            context
+                .storage
+                .get_block_by_number(0)
+                .await?
+                .expect("Failed to get genesis block. This should not happen.")
+                .header,
```
MegaRedHand: Let's move this to a variable

crates/vm/levm/src/precompiles.rs
SDartayet: Sure. Any sspecific way of benchmarking it you have in mind?

crates/common/types/precompile.rs
SDartayet: It depends on the VM crate, but I wasn't able to access stuff within LEVM from the RPC crate; hence why I moved it. It's possible I did something wrong though

crates/common/types/genesis.rs
SDartayet: As far as I can tell, that shouldn't happen. On mainnet I'm pretty sure given the EIP specification if BPO N is activated then BPO N-1 is, and it wouldn't really make much sense for testnets to not follow the same convention. IMHO it's worth revisiting if such a situation comes up in the future, but I don't see why it'd happen.

crates/networking/rpc/eth/client.rs
SDartayet: That was my bad, I assumed we should do so given we only support forks post Paris (and therefore all the relevant forks are timestamp activated). I've noticed working on another PR that's not the case, will fix now.

crates/networking/rpc/eth/client.rs
SDartayet: Sure. [Done](https://github.com/lambdaclass/ethrex/pull/4625/commits/52915dd71b5ac4b3bd9340f12f9babcc1c5ed7cf)!

crates/networking/rpc/eth/client.rs
SDartayet: Solved it [here](https://github.com/lambdaclass/ethrex/pull/4625/commits/52915dd71b5ac4b3bd9340f12f9babcc1c5ed7cf)!

crates/networking/rpc/eth/client.rs
SDartayet: I wasn't sure it was worth, since the only part of the code that checks for the fork before using them is this method. Do you still think it's worth for future proofing though?

crates/common/types/precompile.rs
```diff
@@ -0,0 +1,200 @@
...
+    BLS12_PAIRING_CHECK,
+    P256_VERIFICATION,
+];
+
+pub fn precompiles_for_fork(fork: Fork) -> Vec<Precompile<'static>> {
+    PRECOMPILES
+        .into_iter()
+        .filter(|precompile| precompile.active_since_fork <= fork)
+        .collect()
+}
```
MegaRedHand: We should avoid the allocation here:

```suggestion
pub fn precompiles_for_fork(fork: Fork) -> impl Iterator<Item = Precompile<'static>> {
    PRECOMPILES
        .into_iter()
        .filter(|precompile| precompile.active_since_fork <= fork)
}
```

crates/vm/levm/src/precompiles.rs
MegaRedHand: Nvm. The old implementation was almost the same, so it shouldn't impact performance as long as we [avoid allocating in `precompiles_for_fork`](https://github.com/lambdaclass/ethrex/pull/4625#discussion_r2382599630).

crates/vm/system_contracts.rs
```diff
@@ -0,0 +1,71 @@
+use ethrex_common::{H160, types::Fork, types::Fork::*};
+
+pub struct SystemContract<'a> {
+    pub address: H160,
+    pub name: &'a str,
+    pub active_since_fork: Fork,
+}
```
MegaRedHand: Can we avoid the lifetime here?

```suggestion
pub struct SystemContract {
    pub address: H160,
    pub name: &'static str,
    pub active_since_fork: Fork,
}
```

crates/vm/levm/src/precompiles.rs
```diff
@@ -201,26 +106,205 @@ const FP2_ZERO_MAPPED_TO_G2: [u8; 256] = [
...
-    // Prague or newers forks should only use these precompiles
-    // https://eips.ethereum.org/EIPS/eip-2537
-    if PRECOMPILES_POST_CANCUN.contains(address) && fork < Fork::Prague {
-        return false;
-    }
+pub struct Precompile<'a> {
+    pub address: H160,
+    pub name: &'a str,
+    pub active_since_fork: Fork,
+}
```
MegaRedHand: Same:


```suggestion
pub struct Precompile {
    pub address: H160,
    pub name: &'static str,
    pub active_since_fork: Fork,
}
```

crates/networking/rpc/eth/client.rs
```diff
@@ -62,3 +70,101 @@ impl RpcHandler for Syncing {
...
+    let activation_time = chain_config.get_activation_timestamp_for_fork(fork);
+    let genesis_header = context
+        .storage
+        .get_block_by_number(0)
+        .await?
+        .expect("Failed to get genesis block. This should not happen.")
+        .header;
+    let block_number = context.storage.get_latest_block_number().await?;
+    let fork_id = if let Some(timestamp) = activation_time {
+        ForkId::new(chain_config, genesis_header, timestamp, block_number).fork_hash
```
MegaRedHand, line 156: Does passing the latest block number here work? I thought the timestamp and block number need to be for the block which we want the fork ID from.

crates/networking/rpc/eth/client.rs
MegaRedHand: Not really for future-proofing. It's mostly for symmetry. Doing it this way makes it easier to understand at a glance.

crates/networking/rpc/eth/client.rs
SDartayet: I only meant for this to work properly with post-merge forks to be fair; since the EIP only requires the method to work form Cancun onwards. Probably should add a proper check and return an error for pre-Merge forks though (the EIP doesn't require the method to work for Paris and Shanghai either, but it's simple enough to return something sensible for them)

crates/vm/system_contracts.rs
SDartayet: Didn't know about the static lifetime, will fix it in a sec!

crates/networking/rpc/eth/client.rs
MegaRedHand: Returning an error for pre-Paris forks is OK. We should also add a comment here, mentioning the block number is not really correct, but being enough for post-merge forks.

MegaRedHand (approved):

crates/networking/rpc/eth/client.rs
```diff
@@ -62,3 +70,108 @@ impl RpcHandler for Syncing {
...
+        Ok(Self {})
+    }
+
+    async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        let chain_config = context.storage.get_chain_config()?;
+        let latest_block_timestamp = context
+            .storage
+            .get_block_by_number(context.storage.get_latest_block_number().await?)
+            .await?
+            .unwrap_or_default()
```
mpaulucci: what is the default here?

crates/networking/rpc/eth/client.rs
```diff
@@ -62,3 +70,108 @@ impl RpcHandler for Syncing {
...
+}
+
+impl RpcHandler for Config {
+    fn parse(_params: &Option<Vec<Value>>) -> Result<Self, RpcErr> {
+        Ok(Self {})
+    }
+
+    async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        let chain_config = context.storage.get_chain_config()?;
+        let latest_block_timestamp = context
```
mpaulucci: are you sure you need the latest block timestamp and not the current time?

SDartayet: Wouldn't the current time for configuration purposes be the timestamp of the latest block?

crates/common/types/genesis.rs
```diff
@@ -460,6 +460,97 @@ impl ChainConfig {
         self.get_fork(block_timestamp)
     }
 
+    pub fn next_fork(&self, block_timestamp: u64) -> Option<Fork> {
+        let next = if self.is_bpo5_activated(block_timestamp) {
+            None
+        } else if self.is_bpo4_activated(block_timestamp) && self.bpo5_time.is_some() {
```
mpaulucci, line 466: It seems we need to implement ordering of forks, then this would be simpler

crates/networking/rpc/eth/client.rs
SDartayet: Both blockheader and blockbody implement default, which I assume sets the default value for their respective fields (0 for numeric ones, None for options, etc). I'm not sure if that was the best way to handle the unwrap though. Would you say it's better to just use expect? Since getting the latest block number but that block number being missing shouldn't ever really happen

crates/networking/rpc/eth/client.rs
mpaulucci: not necesarily, you might be in a transition between forks. A block with the newest fork hasn't arrived yet, but the time for the old fork has passed. 

I would verify how Geth does it

crates/networking/rpc/eth/client.rs
mpaulucci: Yes, I don't think `.unwrap_or_default()` is the correct way of dealing with this. You should return an RpcError if there is no block

crates/networking/rpc/eth/client.rs
SDartayet: From what I can see in [this file](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go), Geth also gets the latest block header. The relevant lines start at 1207, and what it does is
```
var (
	c = api.b.ChainConfig()
	t = api.b.CurrentHeader().Time
)
        resp := configResponse{
	Next:    assemble(c, c.Timestamp(c.LatestFork(t)+1)),
	Current: assemble(c, c.Timestamp(c.LatestFork(t))),
	Last:    assemble(c, c.Timestamp(c.LatestFork(^uint64(0)))),
}
```

CurrentHeader() just fetches the latest blockheader, assemble is the function that builds the eth-config response object

crates/networking/rpc/eth/client.rs
SDartayet: Sure, fixed it [here](https://github.com/lambdaclass/ethrex/pull/4625/commits/3a6360cfc8809fb06090d2d024cb8e3d5d53bc91)!

crates/common/types/genesis.rs
SDartayet: I don't think just ordering of forks would suffice; unless there's an approach I'm missing. One, because just ordering wouldn't be enough to ask a fork what the fork after it would be; we'd still need to implement from<u64> for Fork or ahave an array with all the enum variants for Fork indexed by its number (or use a crate like [strum](https://crates.io/crates/strum). Two, because we'd also need to know if the fork is activated, which could only be done if the ChainConfig struct had an array for the fork timestamps we were indexing with the number assigned to the fork. 
I still think doing the necessary refactors for that would be worth it, for the sake of future maintainability reasons too; I decided not to do it to not make this PR too big since it seemed a bit out of scope.

mpaulucci: Ok, please create a follow up ticket. Not planning to work on this right now, but it's good to have one

SDartayet: [Done](https://github.com/lambdaclass/ethrex/issues/4720)

PR #4629, opened 2025-09-24
fix(l2): remove duplicate Following log in determine_new_status

Galoretka:
Remove info! log from determine_new_status for the (true, false) branch.

Why: 1) The log inside determine_new_status caused two issues:
2) Duplicate message on transition to Following (logged both in determine_new_status and update_state).
3) Log spam every tick while staying in Following, since determine_new_status is called periodically even without state changes.


ElFantasma (approved):

MegaRedHand:

PR #4630, opened 2025-09-24
chore(l1): fix memory spike when transitioning from snap to fullsync

rodrigo-o:
**Motivation**

We were seeing spikes of more than 20GB of ram transitioning from snapsync to full sync

**Description**

This PR remove the call to `into_fullsync` during the end of snap sync. That call was unnecessary, `into_fullsync` is used if we try to start a snap sync and we detect we are close to the head (<64 blocks) we move the headers to the fullsync state in that case, but in this one we already finished snapsync so we just need to start fullsync from the last known head.


jrchatruc (approved):

rodrigo-o:
Removing the into_fullsync internal for proved to remove the spikes in Sepolia. We are now running test for mainnet but the current changes should have the same effect.

crates/networking/p2p/sync.rs
```diff
@@ -307,7 +307,9 @@ impl Syncer {
         }
 
         if let SyncMode::Snap = sync_mode {
-            self.snap_sync(store, &mut block_sync_state).await?;
+            self.snap_sync(store.clone(), &mut block_sync_state).await?;
```
MegaRedHand, line 310: This store can be passed as a reference here.

rodrigo-o: Good catch, given that the store is just arcs it shouldn't impact but agree that if by reference is possible it'll be cleaner. I'll do a follow-up

PR #4635, opened 2025-09-24
feat(l2): use safe block delay for commitments and verify txs receipts

### LINKED ISSUE #4563: Wait for a safe L1 block before accepting a commit tx as confirmed
Currently the committer (and proof sender) takes a commit tx to L1 as confirmed as soon as it gets the receipt. This can cause some reorg related problems. To avoid that, wait until the L1 block where the commit was included is safe

ManuelBilbao:
**Motivation**

To add security, the operator should be able wait an amount of blocks on L1 to safely accept a receipt.

**Description**

Add a `eth.safe-block-delay` flag to set the desired delay. Remove `watcher.block-delay` in favor of the first.

Closes #4563

### DIFF
--- a/cmd/ethrex/l2/options.rs
+++ b/cmd/ethrex/l2/options.rs
@@ -5,8 +5,8 @@ use crate::{
 use clap::Parser;
 use ethrex_common::{Address, types::DEFAULT_BUILDER_GAS_CEIL};
 use ethrex_l2::{
-    BasedConfig, BlockFetcherConfig, BlockProducerConfig, CommitterConfig, EthConfig,
-    L1WatcherConfig, ProofCoordinatorConfig, SequencerConfig, StateUpdaterConfig,
+    BasedConfig, BlockFetcherConfig, BlockProducerConfig, CommitterConfig, L1WatcherConfig,
+    ProofCoordinatorConfig, SequencerConfig, StateUpdaterConfig,
     sequencer::{
         configs::{AdminConfig, AlignedConfig, MonitorConfig},
         utils::resolve_aligned_network,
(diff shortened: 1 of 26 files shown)


cmd/ethrex/l2/options.rs
```diff
@@ -253,7 +255,7 @@ pub struct EthOptions {
         help_heading = "Eth options",
         num_args = 1..
     )]
-    pub rpc_url: Vec<String>,
+    pub rpc_url: Vec<Url>,
```
avilagaston9, line 258: Good oportunity to rename this to `rpc_urls`?

crates/l2/monitor/app.rs
```diff
@@ -185,11 +186,21 @@ impl EthrexMonitorWidget {
...
+                .clone(),
+        )
+        .map_err(MonitorError::EthClientError)?;
         // TODO: De-hardcode the rollup client URL
-        let rollup_client =
-            EthClient::new("http://localhost:1729").map_err(MonitorError::EthClientError)?;
+        let rollup_client = EthClient::new(Url::parse("http://localhost:1729").map_err(|_| {
+            MonitorError::EthClientError(EthClientError::ParseUrlError(
+                "http://localhost:1729".to_string(),
+            ))
```
avilagaston9, line 201: I think you can simply do:
```suggestion
            EthClientError::ParseUrlError(
                "http://localhost:1729".to_string(),
            )?
```

docs/CLI.md
```diff
@@ -147,7 +149,7 @@ Node options:
           If the datadir is the word `memory`, ethrex will use the `InMemory Engine`.
 
           [env: ETHREX_DATADIR=]
-          [default: /home/runner/.local/share/ethrex]
+          [default: "/Users/manu/Library/Application Support/ethrex"]
```
avilagaston9, line 152: Restore this one 😄

crates/l2/sdk/src/sdk.rs
avilagaston9, line 1: Maybe we should add some `///` docs to the functions that use the client configured `safe_block_delay`

ilitteri:
Tracked to be done again in https://github.com/lambdaclass/ethrex/issues/5638

PR #4638, opened 2025-09-24
refactor(l1): enhance mempool oldest removal

gianbelinche:
**Motivation**
The removal of elements from the mempool when it is full is not efficient.

**Description**
Now the array that represent the order of insertion to the mempool is a vecdeque, which is not changed when an arbitrary element is removed from the mempool.
Only when the mempool is full, we remove from the vecdeque until we find an element present in the mempool.
Also if the vecdeque is greater than 150% of the maximum mempool size, we prune it, keeping only the txs that are actually in the mempool, preserving the order.

Closes https://github.com/lambdaclass/ethrex/issues/4605?reload=1?reload=1


edg-l (approved):

crates/blockchain/mempool.rs
```diff
@@ -70,10 +82,23 @@ impl Mempool {
...
+
+        // Prune the order queue if it has grown too much
+        if inner.txs_order.len() > MEMPOOL_PRUNE_THRESHOLD {
+            let mut new_txs_order = VecDeque::with_capacity(MEMPOOL_MAX_SIZE * 2);
+            for tx in inner.txs_order.iter() {
+                if inner.transaction_pool.contains_key(tx) {
+                    new_txs_order.push_back(*tx);
+                }
+            }
+            inner.txs_order = new_txs_order;
```
MegaRedHand: Can we use [`retain`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.retain) here?


```suggestion
            inner.txs_order.retain(|tx| inner.transaction_pool.contains_key(tx));
```

gianbelinche: I tried to, but i can't because of borrowing inner both mutably and inmutably

MegaRedHand: In that case, you can temporarily replace one of the two with an empty collection:

```rust
// NOTE: we do this to avoid borrow checker errors
let txpool = core::mem::take(inner.transaction_pool);
inner.txs_order.retain(|tx| txpool.contains_key(tx));
inner.transaction_pool = txpool;
```

MegaRedHand (approved):

Issue #4600, opened 2025-09-22
Define mempool size


gianbelinche:
Currently the mempool max size is defined as 10000, we should better define what it should be

PR #4639, opened 2025-09-24
feat(l1,l2,replay): mempool max size parameter

gianbelinche:
**Motivation**
We want to make the mempool max size a parameter

**Description**
Adds the mempool max size as a parameter to ethrex command

Closes https://github.com/lambdaclass/ethrex/issues/4600

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -20,7 +20,7 @@ use crate::{
         self,
         command::{DB_ETHREX_DEV_L1, DB_ETHREX_DEV_L2},
     },
-    utils::{self, default_datadir, get_client_version, init_datadir},
+    utils::{self, default_datadir, get_client_version, get_minimal_client_version, init_datadir},
 };
 use ethrex_config::networks::Network;
 
@@ -175,6 +175,14 @@ pub struct Options {
         help_heading = "P2P options"
     )]
     pub discovery_port: String,
+    #[arg(
+        long = "block-producer.extra-data",
+        default_value = get_minimal_client_version(),
+        value_name = "EXTRA_DATA",
+        help = "Block extra data message.",
+        help_heading = "Block producer options"
+    )]
+    pub extra_data: String,
 }
 
 impl Options {
@@ -238,6 +246,7 @@ impl Default for Options {
             dev: Default::default(),
             force: false,
             mempool_max_size: Default::default(),
+            extra_data: get_minimal_client_version(),
         }
     }
 }
--- a/cmd/ethrex/bench/build_block_benchmark.rs
+++ b/cmd/ethrex/bench/build_block_benchmark.rs
@@ -151,7 +151,7 @@ async fn create_payload_block(genesis_block: &Block, store: &Store) -> (Block, u
         gas_ceil: DEFAULT_BUILDER_GAS_CEIL,
     };
     let id = payload_args.id();
-    let block = create_payload(&payload_args, store).unwrap();
+    let block = create_payload(&payload_args, store, Bytes::new()).unwrap();
     (block, id.unwrap())
 }
 
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -134,6 +134,7 @@ pub async fn init_rpc_api(
     tracker: TaskTracker,
     log_filter_handler: Option<reload::Handle<EnvFilter, Registry>>,
     gas_ceil: Option<u64>,
+    extra_data: String,
 ) {
     init_datadir(&opts.datadir);
     // Create SyncManager
@@ -160,6 +161,7 @@ pub async fn init_rpc_api(
         get_client_version(),
         log_filter_handler,
         gas_ceil,
+        extra_data,
     );
 
     tracker.spawn(rpc_api);
@@ -422,6 +424,7 @@ pub async fn init_l1(
         log_filter_handler,
         // TODO (#4482): Make this configurable.
         None,
+        opts.extra_data.clone(),
     )
     .await;
 
--- a/cmd/ethrex/utils.rs
+++ b/cmd/ethrex/utils.rs
@@ -166,6 +166,7 @@ pub fn parse_hex(s: &str) -> eyre::Result<Bytes, FromHexError> {
     }
 }
 
+/// Returns a detailed client version string with git info.
 pub fn get_client_version() -> String {
     format!(
         "{}/v{}-{}-{}/{}/rustc-v{}",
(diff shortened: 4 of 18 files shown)


cmd/ethrex/cli.rs
```diff
@@ -229,6 +236,7 @@ impl Default for Options {
             metrics_enabled: Default::default(),
             dev: Default::default(),
             force: false,
+            mempool_max_size: Default::default(),
```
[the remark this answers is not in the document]
gianbelinche: The defaults here do not match the actual defaults from the cli

crates/blockchain/blockchain.rs
```diff
@@ -94,21 +94,26 @@ fn log_batch_progress(batch_size: u32, current_block: u32) {
...
+            mempool: Mempool::new(max_mempool_size),
             is_synced: AtomicBool::new(false),
             r#type: blockchain_type,
             payloads: Arc::new(TokioMutex::new(Vec::new())),
             perf_logs_enabled,
         }
     }
 
-    pub fn default_with_store(store: Store) -> Self {
+    pub fn default_with_store(store: Store, max_mempool_size: usize) -> Self {
```
MegaRedHand: Do we use this anywhere other than tests? We could just use a default value here instead.

crates/blockchain/blockchain.rs
```diff
@@ -94,21 +94,26 @@ fn log_batch_progress(batch_size: u32, current_block: u32) {
 }
 
 impl Blockchain {
-    pub fn new(store: Store, blockchain_type: BlockchainType, perf_logs_enabled: bool) -> Self {
+    pub fn new(
+        store: Store,
+        blockchain_type: BlockchainType,
+        perf_logs_enabled: bool,
+        max_mempool_size: usize,
```
MegaRedHand: We could turn this into a `BlockchainOptions` struct or similar, seeing how we'll probably want to add parameters to tune the mempool and others depending on available memory.

crates/blockchain/blockchain.rs
gianbelinche: We use it on tests and replay, so it shouldn't matter

gianbelinche: Done

crates/blockchain/blockchain.rs
gianbelinche: Done

crates/blockchain/blockchain.rs
```diff
@@ -75,6 +76,19 @@ pub struct Blockchain {
     pub payloads: Arc<TokioMutex<Vec<(u64, PayloadOrTask)>>>,
 }
 
+#[derive(Debug, Clone)]
+pub struct BlockchainOptions {
+    pub max_mempool_size: usize,
+}
```
JereSalo, line 83: Is it possible to add more attributes to this struct? Would that make sense? Like for example the `blockchain_type` or `perf_logs_enabled`.

crates/blockchain/blockchain.rs
```diff
@@ -94,10 +108,15 @@ fn log_batch_progress(batch_size: u32, current_block: u32) {
 }
 
 impl Blockchain {
-    pub fn new(store: Store, blockchain_type: BlockchainType, perf_logs_enabled: bool) -> Self {
+    pub fn new(
+        store: Store,
+        blockchain_type: BlockchainType,
+        perf_logs_enabled: bool,
```
MegaRedHand: We can move this into `BlockChainOpts`

crates/blockchain/blockchain.rs
MegaRedHand: I was going to suggest that too. I think it makes sense to move them.

cmd/ethrex/l2/initializers.rs
```diff
@@ -170,7 +170,14 @@ pub async fn init_l2(
...
     let rollup_store = init_rollup_store(&rollup_store_dir).await;
 
-    let blockchain = init_blockchain(store.clone(), BlockchainType::L2, true);
+    let blockchain = init_blockchain(
+        store.clone(),
+        BlockchainType::L2,
+        true,
+        ethrex_blockchain::BlockchainOptions {
+            max_mempool_size: opts.node_opts.mempool_max_size,
+        },
```
MegaRedHand: We should move this outside to a variable

crates/blockchain/blockchain.rs
tomip01, line 82: Maybe we can add here the `BlockchainType` since always the two parameters are sent together

cmd/ethrex/l2/initializers.rs
gianbelinche: Done

crates/blockchain/blockchain.rs
gianbelinche: Done

crates/blockchain/blockchain.rs
gianbelinche: Done

crates/blockchain/mempool.rs
```diff
@@ -18,21 +18,20 @@ use ethrex_common::{
...
     txs_by_sender_nonce: RwLock<BTreeMap<(H160, u64), H256>>,
     txs_order: RwLock<Vec<H256>>,
+    max_mempool_size: usize,
 }
 impl Mempool {
-    pub fn new() -> Self {
+    pub fn new(max_mempool_size: usize) -> Self {
         Mempool {
-            txs_order: RwLock::new(Vec::with_capacity(MEMPOOL_MAX_SIZE)),
+            txs_order: RwLock::new(Vec::with_capacity(max_mempool_size)),
```
MegaRedHand: We can probably reserve capacity for `txs_by_sender_nonce` and `transaction_pool` too. That won't really affect performance but can reduce noise when profiling memory, and show us if the mempool is somehow exceeding capacity due to bugs

MegaRedHand (approved):

crates/blockchain/mempool.rs
gianbelinche: Done for the pools, `txs_by_sender_nonce ` can't be done since it is a BTreeMap

crates/blockchain/mempool.rs
```diff
@@ -18,21 +18,22 @@ use ethrex_common::{
...
+            transaction_pool: RwLock::new(HashMap::with_capacity(max_mempool_size)),
+            blobs_bundle_pool: Mutex::new(HashMap::with_capacity(max_mempool_size)),
```
MegaRedHand: I think it's better if we don't reserve space for the blobs bundle. That one's less likely to fill, so we can let it find a limit automatically.

MegaRedHand (approved):

crates/blockchain/mempool.rs

PR #4641, opened 2025-09-24
refactor(replay): `ethrex-replay` improvements

ilitteri:
**Motivation**

**Description**

> [!NOTE]
> Some of these changes were inspired by [`ere`](https://github.com/eth-act/ere) and [`zkevm-benchmark-workload`](https://github.com/eth-act/zkevm-benchmark-workload).
> This PR settles the ground for adding `ere` as an alternative backend for `ethrex-replay`.

- Adds new options `--zkvm`, `--resource` (cpu, gpu), and `--action` (prove, execute).
- Adds help heading for options.
- Makes block number, batch number, tx hash positional args to avoid needing to specify them with flags (could seem redundant in some commands e.g. `ethrex-replay block --block`, `ethrex-replay batch --batch`, etc).
- Executes using CPU and no ZKVM by default.
- Removes the module `bench` as it was already replaced by the report.
- Refactors the reports (both how it is printed in the console and as slack message, se the images below).

**Old Messages/Console print**

image

image

**New**

image

image


cmd/ethrex_replay/src/cli.rs
```diff
@@ -299,42 +379,42 @@ impl EthrexReplayCommand {
...
-                    bench: false,
                     to_csv: false,
                     no_zkvm: false,
                     cache_level: CacheLevel::default(),
+                    common,
+                    slack_webhook_url: None,
                 };
 
-                let elapsed = replay_custom_l1_blocks(max(1, n_blocks), &opts).await?;
+                let elapsed = replay_custom_l1_blocks(max(1, n_blocks), opts).await?;
```
tomip01: I think this `elapsed` is now a report and the print below is no longer valid as it is not a duration

cmd/ethrex_replay/src/cli.rs
```diff
@@ -392,25 +488,29 @@ impl EthrexReplayCommand {
...
                     cached: false,
-                    bench: false,
                     to_csv: false,
                     no_zkvm: false,
                     cache_level: CacheLevel::default(),
+                    slack_webhook_url: None,
                 };
 
-                let elapsed = replay_custom_l2_blocks(max(1, n_blocks), &opts).await?;
+                let elapsed = replay_custom_l2_blocks(max(1, n_blocks), opts).await?;
```
tomip01: ditto

tomip01 (commented):
Just a few comments and I saw this: 
running `cargo r -r -p ethrex-replay -- block --action execute --resource cpu --zkvm sp1 --no-zkvm --rpc-url <RPC_URL>` will execute with no backend and print as if it was SP1.

ilitteri:
> Just a few comments and I saw this: running `cargo r -r -p ethrex-replay -- block --action execute --resource cpu --zkvm sp1 --no-zkvm --rpc-url <RPC_URL>` will execute with no backend and print as if it was SP1.

Good catch! About to push a fix to this

```sh
     Running `target/release/ethrex-replay block --action execute --resource cpu --zkvm sp1 --no-zkvm --cached`
error: the argument '--zkvm <ZKVM>' cannot be used with '--no-zkvm'
```

cmd/ethrex_replay/Makefile
```diff
@@ -1,21 +1,21 @@
...
 # Block parameters
 ifdef BLOCK_NUMBER
-REPLAY_BLOCK_ARGS = --block ${BLOCK_NUMBER}
+REPLAY_BLOCK_ARGS = ${BLOCK_NUMBER}
 endif
-REPLAY_BLOCK_ARGS += --rpc-url ${RPC_URL} --bench
+REPLAY_BLOCK_ARGS += --rpc-url ${RPC_URL}
 
 ## Execution block
 execute:
```
JereSalo, line 10: Should this be `execute-ci` given that is intended to be used there only?

cmd/ethrex_replay/src/cli.rs
```diff
@@ -639,59 +736,72 @@ async fn replay_block(block_opts: BlockOptions) -> eyre::Result<()> {
             eyre::Error::msg("no block found in the cache, this should never happen")
         })?;
 
-    let replayer_mode = replayer_mode(opts.execute, opts.no_zkvm)?;
+    let (execution_result, proving_result) = if opts.no_zkvm {
+        (replay_no_zkvm(cache.clone(), &opts).await, None)
+    } else {
+        // Always execute
+        let execution_result = exec(backend(&opts.common.zkvm)?, cache.clone()).await;
```
JereSalo, line 729: Does it make sense if execution fails instead of continuing with the proving we just return an error so that we don't waste time? Normally if execution fails then proving will also fail, right?

cmd/ethrex_replay/src/cli.rs
ilitteri: [67907fd](https://github.com/lambdaclass/ethrex/pull/4641/commits/67907fd6d8d10cfb41bf4760af305c19ec27be0e)

cmd/ethrex_replay/src/cli.rs
ilitteri: [67907fd](https://github.com/lambdaclass/ethrex/pull/4641/commits/67907fd6d8d10cfb41bf4760af305c19ec27be0e)

cmd/ethrex_replay/src/report.rs
```diff
@@ -0,0 +1,401 @@
...
+                format!("https://etherscan.io/block/{}", self.block.header.number)
+            } else {
+                format!(
+                    "https://{}.etherscan.io/block/{}",
+                    self.network, self.block.header.number
+                )
+            },
+        }];
+
+        if let Network::PublicNetwork(_) = self.network {
```
JereSalo:
```suggestion
        if let Network::PublicNetwork(PublicNetwork::Mainnet) = self.network {
```
ethproofs button should only apply for mainnet

docs/ethrex_replay/ethrex_replay.md
```diff
@@ -3,35 +3,58 @@
...
+| ------------------------------ | ----------- |
+| `ethrex-replay l2 batch`       |             |
+| `ethrex-replay l2 block`       |             |
+| `ethrex-replay l2 custom`      |             |
+| `ethrex-replay l2 transaction` |             |
+
+## Supported Clients
+
+> ✅: supported.
+> ⚠️: supported, but flacky.
```
JereSalo:
```suggestion
> ⚠️: supported, but flaky.
```

docs/ethrex_replay/ethrex_replay.md
```diff
@@ -3,35 +3,58 @@
...
+| ethrex     | ✅    | ✅      | ✅      | ✅                |
+| reth       | ✅    | ✅      | ✅      | -                 |
+| geth       | ✅    | ✅      | ✅      | -                 |
+| Nethermind | ⚠️    | ⚠️      | ⚠️      | -                 |
+| erigon     | 🔜    | 🔜      | 🔜      | -                 |
+
+## Supported zkVM Replays (execution & proving)
+
+> ✅: supported.
+> ⚠️: supported, but flacky.
```
JereSalo:
```suggestion
> ⚠️: supported, but flaky.
```

tooling/replayer/src/main.rs
```diff
@@ -3,26 +3,20 @@ use std::time::{Duration, SystemTime};
...
 pub struct Options {
-    #[arg(
-        long,
-        value_name = "URL",
-        env = "SLACK_WEBHOOK_URL",
-        help_heading = "Replayer options"
-    )]
-    pub slack_webhook_url: Option<Url>,
+    #[command(flatten)]
+    pub replayer: EthrexReplayOptions,
```
JereSalo: nice, this is way cleaner

tooling/replayer/src/main.rs
```diff
@@ -187,7 +137,7 @@ async fn main() {
             let (result, _index, _remaining_handles) = select_all(revalidation_handles).await;
             result
         } => {
-            handle_rpc_revalidation_handle_result(res, opts.hoodi_rpc_url.unwrap(), opts.slack_webhook_url.clone()).await; // TODO: change hoodi rpc to generic
+            handle_rpc_revalidation_handle_result(res, opts.hoodi_rpc_url.unwrap(), opts.replayer.slack_webhook_url.clone()).await; // TODO: change hoodi rpc to generic
```
JereSalo: This uses hoodi rpc url and inside the function it notifies about sepolia, kinda weird

cmd/ethrex_replay/src/cli.rs
ilitteri: > Normally if execution fails then proving will also fail, right?

Yes, so far this has been the case, but failures can be different depending on the zkVMs. Proving anyway ensures that if it fails for another reason, we'll know it in the report.

tooling/replayer/src/main.rs
ilitteri: Yep, we need to address this TODO.

ilitteri:
> 1. In the replayer I get this error if I run `target/release/ethrex-replayer --mainnet-rpc-url 'http://eth-mainnet-rpc:8545' --cache-level failed`
> 
> ```
> error: the following required arguments were not provided:
>   <--rpc-url <RPC_URL>|--cached>
> ```
> 
> 
> I feel like the output of replay is very verbose, maybe we can change something? For running with zkvms makes sense but if we want to run replay of loads of blocks in a matter of seconds it feels kinda spammy. I feel like GPU, CPU, RAM and Branch aren't necessary when running locally but they are important for slack notifications. Lmk what you think

1. Oh you're right, I got the same and forgot to fix it.
2. I can add a `--verbose` flag display the new print and keep the old one-liner log if not passed.

tomip01 (approved):

PR #4643, opened 2025-09-24
chore(l1): update genesis files to include Fusaka fields

### LINKED ISSUE #4598: Update testnets and mainnet genesis for fusaka
https://github.com/eth-clients/mainnet/pull/10
https://github.com/eth-clients/mainnet/pull/11 
https://github.com/eth-clients/hoodi/pull/21
https://github.com/eth-clients/sepolia/pull/111
https://github.com/eth-clients/holesky/pull/132

SDartayet:
**Description**

- Added the Fusaka timestamp and blob schedule to the genesis JSON files, for both testnets and mainnet

- Updated genesis.rs with BPO forks for future-proofing

- Updated the ForkID test cases to include de new forks. Also added test cases for hoodi given Holesky will be getting phased out.

Closes #4598


crates/vm/backends/revm/helpers.rs
```diff
@@ -122,5 +122,6 @@ pub fn fork_to_spec_id(fork: Fork) -> SpecId {
         Fork::Cancun => SpecId::CANCUN,
         Fork::Prague => SpecId::PRAGUE,
         Fork::Osaka => SpecId::OSAKA,
+        _ => SpecId::LATEST,
```
SDartayet: I wasn't sure what to do with the BPO forks given this version of REVM doesn't account for them, so I included this as a placeholder. If anyone has a better idea please let me know

cmd/ethrex/networks/mainnet/genesis.json
```diff
@@ -22,6 +22,9 @@
     "shanghaiTime": 1681338455,
     "cancunTime": 1710338135,
     "pragueTime": 1746612311,
+    "osakaTime": 1764798551,
```
MegaRedHand, line 25: [This one is still TBD](https://eips.ethereum.org/EIPS/eip-7607), right?

crates/vm/backends/revm/helpers.rs
MegaRedHand: I'm not sure. As far as I know, they receive the blob schedule parameters as part of the configuration (env?), so BPOs should just be the current fork (i.e. Osaka).

cmd/ethrex/networks/mainnet/genesis.json
SDartayet: From what I can see in the disucssion [here](https://github.com/eth-clients/mainnet/pull/11), yeah. I figured we may as well include the timestamp that's being floated and update it down the line if it changes, but I could take it out and leave it for a future PR or add a comment or open a ticket clarifying to check it in the future.

MegaRedHand: Does the "next fork" affect the fork id? If so, we might want to keep it the same as before for compatibility reasons. Otherwise, I think these values are good enough.

SDartayet: It affects the `next` field of the ForkId (given it's the timestamp of the next fork), but not the `hash` (since it only takes into account activated forks). Not sure if it's only the hash we care about for that or both fields though

MegaRedHand: Only the hash. Our peers won't mind if our fork-next is somewhere in the future.

MegaRedHand (approved):

cmd/ethrex/networks/mainnet/genesis.json
SDartayet: Then there should be no problem!

PR #4645, opened 2025-09-25
feat(l1): keep downloading big storage leaves on pivot jump during snap sync

jrchatruc:
**Motivation**

Our current snap sync implementation stops downloading an account's storage leaves if there's a pivot jump. This is a big deal because in every network there are very few contracts that have almost all of the storage; this means in big networks like Sepolia or Mainnet we currently heal most of the storage tries in their entirety.

This PR makes it so the snap sync step that downloads storage leaves keeps track of big accounts to resume their download when a pivot jump occurs.

**Description**

Closes #issue_number

### DIFF
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -12,6 +12,7 @@ use ethrex_common::{
     types::{AccountState, BlockBody, BlockHeader, Receipt, validate_block_body},
 };
 use ethrex_rlp::encode::RLPEncode;
+use ethrex_storage::Store;
 use ethrex_trie::Nibbles;
 use ethrex_trie::{Node, verify_range};
 use rand::seq::SliceRandom;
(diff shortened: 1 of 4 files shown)


fedacking (approved):
LGTM

crates/networking/p2p/peer_handler.rs
```diff
@@ -1445,27 +1475,95 @@ impl PeerHandler {
...
-
-                            let task = StorageTask {
-                                start_index: remaining_start,
-                                end_index: remaining_start + 1,
-                                start_hash,
-                                end_hash: Some(end_hash),
-                            };
-                            tasks_queue_not_started.push_back(task);
-                            task_count += 1;
+                                // TODO: DRY
```
MegaRedHand, line 1496: Please dry this

jrchatruc:
Closing since it was included in #4689

PR #4648, opened 2025-09-25
chore(l2): cache precaculated hashes for an in memory trie

xqft:
**Motivation**

The prover's guest program takes as input a list of RLP encoded nodes, hashes them and later rebuilds the trie using `from_nodes()`, but this function does not cache the hashes already calculated, leading to double hashing.

This cuts calls of `compute_hash` by half (some remain because of new nodes that appear when modifying the state trie after block execution).

flamegraphs:
before:
image
after:
image

Proving time diffs:
image


crates/common/trie/node.rs
```diff
@@ -84,6 +84,16 @@ impl NodeRef {
             NodeRef::Hash(hash) => *hash,
         }
     }
+
+    /// # Warning
+    /// will lead to errors if setting the wrong hash, use only
+    /// when the hash was correctly precalculated
+    pub fn with_hash(self, hash: NodeHash) -> Self {
+        if let NodeRef::Node(_, node_hash) = &self {
+            let _ = node_hash.set(hash);
```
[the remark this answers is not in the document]
xqft: only errs if the value was already set

crates/common/trie/node.rs
```diff
@@ -84,6 +84,16 @@ impl NodeRef {
             NodeRef::Hash(hash) => *hash,
         }
     }
+
+    /// # Warning
+    /// will lead to errors if setting the wrong hash, use only
+    /// when the hash was correctly precalculated
+    pub fn with_hash(self, hash: NodeHash) -> Self {
```
Oppen: This should have an indication in its name that it doesn't check the validity of the hash, and the warning comment should be a `SAFETY` one, so it's easier to grep for.
```suggestion
    /// # SAFETY: caller must ensure the hash is correct for the node.
    /// Otherwise, the `Trie` will silently produce incorrect results and may
    /// fail to query other nodes from the `TrieDB`.
    pub fn with_hash_unchecked(self, hash: NodeHash) -> Self {
```

crates/common/trie/trie.rs
```diff
@@ -279,7 +279,11 @@ impl Trie {
 
                         if hash.is_valid() {
                             *choice = match all_nodes.get(&hash.finalize()) {
-                                Some(rlp) => get_embedded_node(all_nodes, rlp)?.into(),
+                                Some(rlp) => {
+                                    let node_ref: NodeRef =
+                                        get_embedded_node(all_nodes, rlp)?.into();
+                                    node_ref.with_hash(hash)
```
ilitteri: It is safe to use it here since we passed the `hash.is_valid()` check, right?

xqft: `hash.is_valid()` only checks that the hash is not empty, there isn't a way to check that a hash corresponds to a preimage without hashing again

crates/common/trie/node.rs
xqft: much better! thanks

ilitteri:
Superseded by https://github.com/lambdaclass/ethrex/pull/5224

PR #4650, opened 2025-09-25
perf(levm): remove `OpcodeResult`

azteca1998:
**Motivation**

Result checking is potentially a source of performance loss in tight loops.

**Description**

Reduce the data returned in the opcodes' results to check if performance improves.


crates/vm/levm/src/opcodes.rs
```diff
@@ -358,12 +358,12 @@ impl From<Opcode> for usize {
 
 /// Represents an opcode function handler.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct OpCodeFn<'a>(fn(&'_ mut VM<'a>) -> Result<OpcodeResult, VMError>);
+pub(crate) struct OpCodeFn<'a>(fn(&'_ mut VM<'a>) -> Result<bool, VMError>);
```
Arkenan: Let's explain in the documentation comment that the boolean returned by the result is `true` if a halt was produced and the caller needs to handle it.

crates/vm/levm/src/errors.rs
```diff
@@ -218,16 +218,6 @@ pub enum DatabaseError {
...
 
-#[derive(Debug, Clone)]
-/// Note: "Halt" does not mean "Error during execution" it simply
-/// means that the execution stopped. It's not called "Stop" because
-/// "Stop" is an Opcode
-pub enum OpcodeResult {
-    Continue { pc_increment: usize },
-    SetPc { new_pc: usize },
-    Halt,
-}
```
MegaRedHand, line 229: Can we keep this as an enum? Like:

```rust
pub enum OpcodeResult {
    Continue,
    Halt,
}
```

It should yield the same performance but being more readable.

azteca1998: Fixed.

crates/vm/levm/src/opcode_handlers/arithmetic.rs
```diff
@@ -52,12 +52,12 @@ impl<'a> VM<'a> {
 
         let [dividend, divisor] = *current_call_frame.stack.pop()?;
         let Some(quotient) = dividend.checked_div(divisor) else {
-            current_call_frame.stack.push_zero()?;
-            return Ok(OpcodeResult::Continue { pc_increment: 1 });
+            current_call_frame.stack.push1(U256::zero())?;
```
edg-l: why change this? in my measurments push zero is faster

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -354,8 +351,8 @@ impl<'a> VM<'a> {
 
         current_call_frame
             .stack
-            .push1(U256::from(current_call_frame.pc))?;
+            .push1(U256::from(current_call_frame.pc.wrapping_sub(1)))?;
```
MegaRedHand, line 354: I see we're subtracting one here. Did this PR change the meaning of the program counter?

crates/vm/levm/src/opcodes.rs
```diff
@@ -357,6 +357,8 @@ impl From<Opcode> for usize {
 }
 
 /// Represents an opcode function handler.
+///
+/// The boolean return value is true if the VM's execution should be interrupted for any reason.
```
MegaRedHand: This comment is no longer needed

crates/vm/levm/src/opcode_handlers/push.rs
```diff
@@ -43,21 +34,19 @@ impl<'a> VM<'a> {
             U256::zero()
         };
 
-        current_call_frame.stack.push1(value)?;
+        call_frame.stack.push1(value)?;
 
         // The n_bytes that you push to the stack + 1 for the next instruction
```
edg-l: outdated comment i guess

edg-l (approved):

crates/vm/levm/src/opcode_handlers/push.rs
azteca1998: Not really, we're incrementing the PC by that number of bytes, but maybe it could be reworded.

It should be better now.

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
azteca1998: Internally, yes.
From an external point of view, no.

It changed in the sense that it now is pre-incremented before executing the operation handler. This has to be done since we're no longer returning the new PC value/delta but we still have to support PUSH and JUMP instructions.

The simplest way to support all that is to pre-increment the PC and override that value afterwards if necessary. Otherwise we'd need to always jump to the PC before the target instruction and would be more confusing and prone to under/overflowing (when the first instruction is a JUMPDEST).

The side effect is that the PC opcode has to subtract 1 from the internal value, which cannot underflow since we've already pre-incremented it, therefore it is known to not be zero.

crates/vm/levm/src/opcodes.rs
azteca1998: Fixed.

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
MegaRedHand: Can we add a mention of this here?

crates/vm/levm/src/vm.rs
```diff
@@ -452,21 +452,15 @@ impl<'a> VM<'a> {
 
         loop {
             let opcode = self.current_call_frame.next_opcode();
+            self.advance_pc(1)?;
```
JereSalo, line 455: maybe add a comment explaining why we pre-increment

PR #4655, opened 2025-09-25
fix(l1): preventing "Address family not supported by protocol" errors on discv4

ElFantasma:
**Motivation**

Using `UdpFramed` to send messages was causing a problem when a message send failed: A singe error was causing every other message send to fail. (See https://github.com/tokio-rs/tokio/issues/7648)
It was causing the log spam issue described in #4232 

**Description**

Moved back to use `UdpSocket` directly to send messages. It prevents the "Address family not supported by protocol" propagation of errors.

Also fixes some of the issues in #4232 (but still full IPv6 needs to be implemented):
- Comparing canonical IP addresses
- Removes a custom implementation of `to_canonical()`

### DIFF
--- a/crates/networking/p2p/discv4/server.rs
+++ b/crates/networking/p2p/discv4/server.rs
@@ -5,9 +5,10 @@ use std::{
     time::{Duration, Instant},
 };
 
+use bytes::BytesMut;
 use ethrex_common::H256;
 use ethrex_common::{H512, U256};
-use futures::{SinkExt as _, StreamExt, stream::SplitSink};
+use futures::StreamExt;
 use rand::rngs::OsRng;
 use secp256k1::SecretKey;
 use spawned_concurrency::{
@@ -81,17 +82,12 @@ pub enum OutMessage {
     Done,
 }
 
-type UdpFramedSplitSink =
-    SplitSink<UdpFramed<Discv4Codec, Arc<UdpSocket>>, (Message, std::net::SocketAddr)>;
-
 #[derive(Debug)]
 pub struct DiscoveryServer {
     local_node: Node,
     local_node_record: Arc<Mutex<NodeRecord>>,
     signer: SecretKey,
     udp_socket: Arc<UdpSocket>,
-    /// Sink end of the UdpFramed stream to send messages to peers
-    sink: Option<Arc<Mutex<UdpFramedSplitSink>>>,
     kademlia: Kademlia,
 }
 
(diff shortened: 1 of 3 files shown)


crates/networking/p2p/discv4/server.rs
```diff
@@ -246,8 +240,7 @@ impl DiscoveryServer {
                 continue;
             }
 
-            if let Err(err) = self.send_find_node(&contact.node).await {
-                error!(sent = "FindNode", to = %format!("{:#x}", contact.node.public_key), err = ?err, "Error sending message");
```
MegaRedHand, line 250: We should keep this as `debug` or `trace`

crates/networking/p2p/discv4/server.rs
```diff
@@ -353,9 +346,7 @@ impl DiscoveryServer {
         let random_pub_key = public_key_from_signing_key(&random_priv_key);
 
         let msg = Message::FindNode(FindNodeMessage::new(random_pub_key, expiration));
-        self.send(msg, node.udp_addr()).await.inspect_err(|e| {
-            error!(sent = "FindNode", to = ?node, err = ?e, "Error sending message");
```
MegaRedHand, line 357: Same

crates/networking/p2p/discv4/server.rs
```diff
@@ -475,12 +466,7 @@ impl DiscoveryServer {
         // Sending all in one packet would exceed bounds with the nodes only, weighing
         // up to 1424B.
         for chunk in neighbors.chunks(8) {
-            let _ = self
-                .send_neighbors(chunk.to_vec(), &node)
-                .await
-                .inspect_err(|e| {
-                    error!(sent = "Neighbors", to = %format!("{sender_public_key:#x}"), err = ?e, "Error sending message");
```
MegaRedHand, line 482: Same

gianbelinche (approved):

MegaRedHand (approved):
LGTM. Left minor comment.

crates/networking/p2p/discv4/server.rs
ElFantasma: I removed those because they were repeated log lines. They are being logged as `error` [here](https://github.com/lambdaclass/ethrex/blob/aada85a8cfd980efe85e6f52ce1af9d477c3cbde/crates/networking/p2p/discv4/server.rs#L520)

MegaRedHand: Oops, I didn't notice that. LGTM then

PR #4656, opened 2025-09-25
feat(l1): use descriptive error when we fail to replace a mempool tx

fmoletta:
**Motivation**
We currently return a "NonceTooLow" error when we fail to replace a mempool transaction (aka we receive a transaction with the same sender and nonce as an existing one but without a higher fee value). This doesn't make any sense and can be quite confusing. This PR adds a MempoolError enum variant for this specific case.

**Description**
* Use a specific error variant when we fail to replace a mempool transaction

Closes None

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -435,7 +435,7 @@ impl Mempool {
         };
 
         if !is_a_replacement_tx {
-            return Err(MempoolError::NonceTooLow);
+            return Err(MempoolError::UnderpricedReplacement);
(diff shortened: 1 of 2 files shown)


crates/blockchain/mempool.rs
```diff
@@ -435,7 +435,7 @@ impl Mempool {
         };
 
         if !is_a_replacement_tx {
```
MegaRedHand, line 437: Can we rename this to `is_a_valid_replacement_tx`?

PR #4657, opened 2025-09-25
perf(l1): use secondary instances for reading in RocksDB

Oppen:
- **perf(l1): use secondary replica for reading in RocksDB**
- **catch up on apply_account_updates and put_batch**
- **test: catch up on every write**
- **open column families**

**Motivation**

**Description**

Closes #issue_number


jrchatruc:
Closing as it's become too old

PR #4659, opened 2025-09-25
feat(l1): add optional metrics to docker compose

ManuelBilbao:
**Motivation**

**Description**

Added optional Prometheus/Grafana metrics to the docker compose file.
Also, remove the volumes when started, unless `KEEP_DATA=true` is set

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,6 +22,9 @@ services:
       beacon
       --http
       --http-address 0.0.0.0
+      --metrics
+      --metrics-address 0.0.0.0
+      --metrics-allow-origin http://lighthouse:5054
       --execution-endpoint http://ethrex:8551
       --execution-jwt /secrets/jwt.hex
       --checkpoint-sync-url https://${ETHREX_NETWORK:-mainnet}-checkpoint-sync.attestant.io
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -45,15 +69,112 @@ services:
...
+      apk add curl;
+      mkdir -p /metrics/grafana/data;
+      mkdir -p /metrics/grafana/provisioning/datasources;
+      mkdir -p /metrics/grafana/provisioning/dashboards/common_dashboards;
+      mkdir -p /metrics/prometheus/data;
+      chmod -R 777 /metrics/grafana/data;
+      chmod -R 777 /metrics/prometheus/data;
+      curl -L -o /metrics/grafana/provisioning/dashboards/dashboards.yaml https://raw.githubusercontent.com/lambdaclass/ethrex/refs/heads/${ETHREX_METRICS_BRANCH:-main}/metrics/provisioning/grafana/dashboards/dashboard_config_l1.yaml;
+      curl -L -o /metrics/grafana/provisioning/dashboards/common_dashboards/ethereum_metrics_exporter.json https://raw.githubusercontent.com/lambdaclass/ethrex/refs/heads/${ETHREX_METRICS_BRANCH:-main}/metrics/provisioning/grafana/dashboards/common_dashboards/ethereum_metrics_exporter.json;
+      curl -L -o /metrics/grafana/provisioning/dashboards/common_dashboards/ethrex_l1_perf.json https://raw.githubusercontent.com/lambdaclass/ethrex/refs/heads/${ETHREX_METRICS_BRANCH:-main}/metrics/provisioning/grafana/dashboards/common_dashboards/ethrex_l1_perf.json;
```
mpaulucci, line 97: why don't you use the local files?

ManuelBilbao: The idea of this `docker-compose.yaml` is to download it and run it without cloning the entire repo, as in the [docs](https://docs.ethrex.xyz/getting-started/quickstart-l1.html)

docker-compose.yaml
```diff
@@ -1,9 +1,30 @@
 services:
+  reset_volumes:
+    image: alpine
+    volumes:
+      - lighthouse:/lighthouse
+      - ethrex:/ethrex
+      - metrics:/metrics
+    command: >
+      sh -c '
+      if [ -z "${KEEP_DATA:-}" ] || [ "${KEEP_DATA:-}" == "0" ] || [ "${KEEP_DATA:-}" == "false" ];
```
mpaulucci, line 10: do we really want keep data to be opt in?

mpaulucci (approved):

docker-compose.yaml
ManuelBilbao: Ethrex does not support resuming on snapsync yet, so if the process is interrupted and then tried to run again, it will fail

PR #4663, opened 2025-09-26
ci(l2): disable replay proving with risc0

avilagaston9:
**Motivation**

Running the `Replay proving` workflow with both backends takes too much time on our self-hosted runner.

**Description**

Temporarily disable replay proving with Risc0.

Closes None

### DIFF
--- a/.github/workflows/main_prover_l1.yaml
+++ b/.github/workflows/main_prover_l1.yaml
@@ -28,6 +28,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
+        # TODO (#4669): re-enable risc0
         # backend: ["sp1", "risc0"]
         backend: ["sp1"]
     steps:


tomip01 (approved):

.github/workflows/main_prover_l1.yaml
```diff
@@ -28,7 +28,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        backend: ["sp1", "risc0"]
+        # backend: ["sp1", "risc0"]
```
ilitteri, line 32: Open an issue and add a comment explaining why that's commented.

ilitteri (changes requested):

.github/workflows/main_prover_l1.yaml
avilagaston9: Created #4669 and added a comment in [6956b6a](https://github.com/lambdaclass/ethrex/pull/4663/commits/6956b6aed2e7096957238411d64524dd21b7bd27)!

ilitteri (approved):

PR #4658, opened 2025-09-25
test(l1): add full-sync smoke test

MegaRedHand:
Blocked by #4652

**Motivation**

Full-sync appears not to be working.

**Description**

~This PR adds a test that is currently failing, related to full-sync.~ No longer fails since [9d63811](https://github.com/lambdaclass/ethrex/pull/4658/commits/9d63811632c9b4bb1732f57b948910b3ac498193). The error was due to a missing FCU.

PR #4673, opened 2025-09-26
test(l1): full Sync re org test

gianbelinche:
Blocked by https://github.com/lambdaclass/ethrex/pull/4658]
**Motivation**
We want to test a reorg where we have 3 nodes. One with the base chain, one with a side chain and a third one where we want to full sync the base chain.

**Description**
Adds a test for that scenario, this test is currently flaky, since 50% of the times it will choose the peer with the base chain, working fine, but the other 50% it will choose the peer with the side chain, where it currently fails.
Since there is always a 50% chance of failing, we run the test 30 times to make sure the probability of failure is small enough.

Closes

### DIFF
--- a/tooling/reorgs/src/main.rs
+++ b/tooling/reorgs/src/main.rs
@@ -33,6 +33,7 @@ async fn main() {
     info!("");
 
     run_test(&cmd_path, no_reorgs_full_sync_smoke_test).await;
+    run_test(&cmd_path, reorgs_full_sync_smoke_test).await;
     run_test(&cmd_path, test_one_block_reorg_and_back).await;
     run_test(&cmd_path, test_storage_slots_reorg).await;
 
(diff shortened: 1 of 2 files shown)


tooling/reorgs/src/main.rs
```diff
@@ -31,10 +32,17 @@ async fn main() {
     info!("Starting test run");
     info!("");
 
+    run_test(&cmd_path, no_reorgs_full_sync_smoke_test).await;
+    run_test(&cmd_path, test_reorg_back_to_base).await;
+    // This test is flaky 50% of the time, check that it runs correctly 30 times in a row
```
MegaRedHand, line 37: We should open an issue for this and link it, with an explanation on how it can be made to fail every time.

tooling/reorgs/src/main.rs
```diff
@@ -76,6 +87,85 @@ where
...
+    // Extend the base chain
+    let base_chain = node1.extend_chain(base_chain, 10).await;
+    // Create another chain (a fork of the first one) and extend it with a few empty blocks
+    let _ = node0.extend_chain(side_chain, 10).await;
+
+    // Try to fully sync node0 to base chain
+    node0.update_forkchoice(&base_chain).await;
+}
+
+async fn reorgs_full_sync_smoke_test(simulator: Arc<Mutex<Simulator>>) {
```
MegaRedHand: Let's rename it to `test_chain_split`, since that's the specific scenario we're testing.

tooling/reorgs/src/main.rs
gianbelinche: https://github.com/lambdaclass/ethrex/issues/4775

tooling/reorgs/src/main.rs
gianbelinche: Done

tooling/reorgs/src/main.rs
```diff
@@ -31,7 +32,16 @@ async fn main() {
     info!("Starting test run");
     info!("");
 
+    run_test(&cmd_path, no_reorgs_full_sync_smoke_test).await;
+    run_test(&cmd_path, test_reorg_back_to_base).await;
+    // This test is flaky 50% of the time, check that it runs correctly 30 times in a row
```
MegaRedHand, line 37:
```suggestion
    // This test is flaky 50% of the time, check that it runs correctly 30 times in a row
    // TODO(#4775): make it deterministic
```

Issue #2126, opened 2025-03-04
L1: fix syncing header request on side-chain scenarios


rodrigo-o:
Right now we are having 16 hive test failing in the engine-cancun suite, they are related to Syncing + reorg. The initial issue with these test is that they try to sync from the current head to the target header, which is in a completely different chain (a side-chain that forked ~5 blocks before), which ends in an interminable cycle of request to peers where we only recieve the current header (which aappears to be the way that geth answers when it knows that there is no path between the `from` header and the target for syncing.

rodrigo-o:
A quick-and-dirty workaround to look this working as expected is the following (implemented as part of this [testing PR)](https://github.com/lambdaclass/ethrex/pull/2102):

```diff
diff --git a/crates/networking/p2p/sync.rs b/crates/networking/p2p/sync.rs
index 47cb86e3f..54bc1d92e 100644
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -178,7 +178,7 @@ impl SyncManager {
         };
 
         loop {
-            debug!("Requesting Block Headers from {current_head}");
+            debug!("Requesting Block Headers from {current_head} to {sync_head}");
             // Request Block Headers from Peer
             match self
                 .peers
@@ -186,6 +186,13 @@ impl SyncManager {
                 .await
             {
                 Some(mut block_headers) => {
+                    if block_headers.len() == 1 && block_headers[0].compute_block_hash() == current_head && current_head != sync_head {
+                        warn!("Already at the current head {}, and number {} but not yet synced, we are at a fork, going back one more ancestor", current_head, block_headers[0].number);
+                        store.unset_canonical_block(block_headers[0].number)?;
+                        current_head = block_headers[0].parent_hash;
+                        continue;
+                    }
+
                     debug!(
                         "Received {} block headers| Last Number: {}",
                         block_headers.len(),
@@ -293,6 +300,7 @@ impl SyncManager {
                 self.sync_mode = SyncMode::Full;
             }
             SyncMode::Full => {
+                info!("Starting full sync of headers");
                 // full-sync: Fetch all block bodies and execute them sequentially to build the state
                 download_and_run_blocks(
                     all_block_hashes,
@@ -328,12 +336,17 @@ async fn download_and_run_blocks(
                 .drain(..block_bodies_len)
                 .zip(block_bodies.into_iter())
             {
+                info!("Processing block {}", hash);
                 let header = store
[...]

PR #4676, opened 2025-09-26
test(l1): fix hive ReOrg tests by requesting headers from newToOld

### LINKED ISSUE #4577: Fix Shanghai "Block Re-Org, Sync" hive tests
Some tests in the `engine-withdrawals` hive test suite are failing:

- Withdrawals Fork on Block 1 - 8 Block Re-Org, Sync (Paris) (ethrex)
- Withdrawals Fork on Block 8 - 10 Block Re-Org Sync (Paris) (ethrex)
- Withdrawals Fork on Canonical Block 8 / Side Block 7 - 10 Block Re-Org Sync (Paris) (ethrex)
- Withdrawals Fork on Canonical Block 8 / Side Block 9 - 10 Block Re-Org Sync (Paris) (ethrex)

These have two clients running, while the CL sends blocks to the two clients. At some point, it sends a side-chain to one of the two clients and sets that chain as canonical. That client keeps trying to ask the other one for headers until it times out.

cdiielsi:
**Motivation**

The following tests in the `engine-withdrawals` hive test suite:

- Withdrawals Fork on Block 1 - 8 Block Re-Org, Sync (Paris) (ethrex)
- Withdrawals Fork on Block 8 - 10 Block Re-Org Sync (Paris) (ethrex)
- Withdrawals Fork on Canonical Block 8 / Side Block 7 - 10 Block Re-Org Sync (Paris) (ethrex)
- Withdrawals Fork on Canonical Block 8 / Side Block 9 - 10 Block Re-Org Sync (Paris) (ethrex)

failed because when syncing we always request the block headers from our `current_head` on (OldToNew), but in these tests the request goes to a node that's on a sidechain that didn't have our `current_head` so the request returned no headers and syncing never finished.

**Description**

In this pr I tackle the case described by having a first request of headers from NewToOld from the `sync_head`, and then trying to sync, this is all covered in a new function `synced_new_to_old`. If syncing fails because we have no common ancestor within the obtained headers we go back to request headers as we did before, from OldToNew from our `current_head` and then try to sync as usual.

Closes #4577 and partially solves #2126 as a provisional solution until #4717 is solved.


crates/networking/p2p/sync.rs
```diff
@@ -336,96 +336,160 @@ impl Syncer {
...
-            else {
-                warn!("Sync failed to find target block header, aborting");
-                debug!("Sync Log 8: Sync failed to find target block header, aborting");
-                return Ok(());
+        let requested_header =
+            if let Some(sync_head_block) = store.get_pending_block(sync_head).await? {
+                sync_head_block.header.parent_hash
+            } else {
+                sync_head
             };
```
MegaRedHand: Can we do a `while` here?

crates/networking/p2p/sync.rs
```diff
@@ -336,96 +336,160 @@ impl Syncer {
...
+                .process_incoming_headers(
+                    block_headers,
+                    sync_head,
+                    true, // sync_head_found is true because of the NewToOld headers request
+                    self.blockchain.clone(),
+                    self.peers.clone(),
+                    self.cancel_token.clone(),
+                )
+                .await?;
+            Ok(())
```
MegaRedHand: Let's do an early return here.

crates/networking/p2p/sync.rs
```diff
@@ -336,96 +336,160 @@ impl Syncer {
...
-            debug!("Sync Log 1: In Full Sync");
-            debug!(
-                "Sync Log 3: State current headers len {}",
-                block_sync_state.current_headers.len()
-            );
-            debug!(
-                "Sync Log 4: State current blocks len {}",
-                block_sync_state.current_blocks.len()
-            );
+        info!("Sync Log 1: In Full Sync");
```
MegaRedHand: Let's move all of the new code to another function, to keep this one small.
Also, above the call to that function, let's add a comment with the reasoning behind the call and a ref to this PR.

crates/networking/p2p/sync.rs
```diff
@@ -336,96 +336,160 @@ impl Syncer {
...
+                    self.peers.clone(),
+                    self.cancel_token.clone(),
+                )
+                .await?;
+            Ok(())
+        } else {
+            // if found_common_ancestor is false that means we are more than 1024 blocks behind so, for now, we go back to syncing as it follows.
+            // TODO: Have full syncing always be from NewToOld
+            loop {
+                info!("Sync Log 1: In Full Sync");
```
MegaRedHand: This should still be `debug!`

crates/networking/p2p/sync.rs
cdiielsi: You mean to replace the original loop with a while? Sure! I'd replace it with `while(!sync_head_found)`

.github/workflows/pr-main_l1.yaml
```diff
@@ -191,7 +191,7 @@ jobs:
...
+            test_pattern: "engine-withdrawals/Corrupted Block Hash Payload|Empty Withdrawals|engine-withdrawals test loader|GetPayloadBodies|GetPayloadV2 Block Value|Max Initcode Size|Sync after 2 blocks - Withdrawals on Genesis|Withdraw many accounts|Withdraw to a single account|Withdraw to two accounts|Withdraw zero amount|Withdraw many accounts|Withdrawals Fork on Block 1 - 1 Block Re-Org|Withdrawals Fork on Block 1 - 8 Block Re-Org |Withdrawals Fork on Block 2|Withdrawals Fork on Block 3|Withdrawals Fork on Block 8 - 10 Block Re-Org |Withdrawals Fork on Canonical Block 8 / Side Block 7 - 10 Block Re-Org |Withdrawals Fork on Canonical Block 8 / Side Block 9 - 10 Block Re-Org"
```
mpaulucci: If you fix all the tests, you should be able to remove this completely since you will be passing all tests

crates/networking/p2p/sync.rs
MegaRedHand: No. I mean to continue fetching the `sync_head` parent with `get_pending_block` until we don't find it in the DB

crates/networking/p2p/sync.rs
```diff
@@ -336,96 +336,184 @@ impl Syncer {
...
-                block_sync_state.current_headers.len()
-            );
-            debug!(
-                "Sync Log 4: State current blocks len {}",
-                block_sync_state.current_blocks.len()
-            );
+        if !self
+            .synced_new_to_old(&mut block_sync_state, sync_head, store)
+            .await?
+        {
```
MegaRedHand: Let's do an early return here.

crates/networking/p2p/sync.rs
```diff
@@ -336,96 +336,184 @@ impl Syncer {
...
-                current_head = first_block_parent_hash;
-                continue;
-            }
+                // TODO(#2126): This is just a temporary solution to avoid a bug where the sync would get stuck
+                // on a loop when the target head is not found, i.e. on a reorg with a side-chain.
+                if first_block_hash == last_block_hash
+                    && first_block_hash == current_head
+                    && current_head != sync_head
+                {
+                    // There is no path to the sync head this goes back until it find a common ancerstor
```
MegaRedHand: Let's try removing this `if`, since it tries to solve the same issue as this PR.

crates/networking/p2p/sync.rs
```diff
@@ -620,16 +708,28 @@ impl FullBlockSyncState {
...
+        // Then we add it in current_blocks for execution.
+        let mut pending_block_to_sync = vec![];
+        let mut last_header_to_sync = sync_head;
+        while let Some(block) = self.store.get_pending_block(last_header_to_sync).await? {
+            let block_parent = block.header.parent_hash;
+            if self
+                .current_blocks
+                .last()
+                .is_some_and(|block| block.hash() == block_parent)
+                && !self.current_blocks.contains(&block)
```
MegaRedHand: This line isn't needed. If the last block is `block`'s parent, none of the rest of the blocks can be `block`, otherwise we'd have a loop.

```suggestion
```

crates/networking/p2p/sync.rs
```diff
@@ -620,16 +708,28 @@ impl FullBlockSyncState {
...
-            if let Some(last_block) = self.current_blocks.last() {
-                if last_block.hash() == block.header.parent_hash {
-                    self.current_blocks.push(block);
-                    sync_head_found = true;
-                }
+        // We check if we have pending blocks we didn't request that are needed for syncing
+        // Then we add it in current_blocks for execution.
+        let mut pending_block_to_sync = vec![];
+        let mut last_header_to_sync = sync_head;
+        while let Some(block) = self.store.get_pending_block(last_header_to_sync).await? {
```
MegaRedHand, line 725: This can potentially slow down the processing of blocks if we have a large chain of pending blocks. If we are syncing from genesis, it might take a while until this chain connects, but this loop will happen each time we process a batch of blocks. We should implement some in-memory structure to make this check easy, but for now, let's just limit this check to 32 blocks with a comment explaining this.

crates/networking/p2p/sync.rs
cdiielsi: Ok! This pr only fixes this issue if we are less than 1024 blocks behind though, otherwise I think this bug might still happen, should I remove it anyway?

crates/networking/p2p/sync.rs
cdiielsi: Sorry! I probably added it while debugging, I'm taking it away!

crates/networking/p2p/sync.rs
MegaRedHand: Actually, let's keep it, since this _could_ fix some edge cases.

MegaRedHand (approved):

PR #4681, opened 2025-09-29
fix(replay): replace unwraps with proper error handling in RpcDB

Forostovec:
**Motivation**

TODO: remove unwraps comment.

**Description**

- Replaced unsafe unwraps in RpcDB to prevent panics under RPC flakiness and align with eyre-based error propagation.
- Updated to_execution_witness to propagate errors for initial/final account fetches instead of panicking.
- Updated get_account_info to avoid panic on missing account; return default AccountInfo and cache only when present.
- Updated get_storage_value to avoid panic on missing account; return U256::zero when absent or slot not found.

### DIFF
--- a/cmd/ethrex_replay/src/rpc/db.rs
+++ b/cmd/ethrex_replay/src/rpc/db.rs
@@ -504,8 +504,11 @@ impl LevmDatabase for RpcDB {
                 self.cache.lock().unwrap().insert(address, account.clone());
                 account
             } else {
-                // If the RPC did not return the account, consider it non-existing
-                return Ok(AccountInfo::default());
+                // If the RPC did not return the account, propagate an error instead of defaulting
+                return Err(DatabaseError::Custom(format!(
+                    "RPC did not return account {}",
+                    address
+                )));
             }
         };
         if let Account::Existing {


cmd/ethrex_replay/src/rpc/db.rs
```diff
@@ -496,10 +499,14 @@ impl LevmDatabase for RpcDB {
...
-                .clone();
-            self.cache.lock().unwrap().insert(address, account.clone());
-            account
+                .cloned();
+            if let Some(account) = account.clone() {
+                self.cache.lock().unwrap().insert(address, account.clone());
+                account
+            } else {
+                // If the RPC did not return the account, consider it non-existing
+                return Ok(AccountInfo::default());
```
tomip01: I'm not entirely sure about this. What if the RPC failed for some other reason?
Maybe we should keep the behavior as it is and return a `DatabaseError`

Forostovec: > I'm not entirely sure about this. What if the RPC failed for some other reason? Maybe we should keep the behavior as it is and return a `DatabaseError`

ok

iovoid:
The replay tool now lives in a [separate repository](https://github.com/lambdaclass/ethrex-replay).

PR #4683, opened 2025-09-29
fix(replay): bring back CI benches

### LINKED ISSUE #2756: L2: Add gpu/cpu detailed information to output json in RPC bench crate
#2755 removed this because the original crate was removed `system-info-lite` was tried but returned empty values in linux. 
The objective is to have something like 
```
[{"name":"SP1, NVIDIA RTX 4000 SFF Ada Generation","unit":"Mgas/s","value":2.972603627395457e-11}]%
```
or 
```
[{"name":"SP1, Apple M1","unit":"Mgas/s","value":2.972603627395457e-11}]%
```

as the output file from the prover/bench crate

ilitteri:
**Motivation**

https://github.com/lambdaclass/ethrex/pull/4641 broke the CI by removing the `bench` module, which was used by the CI to update https://docs.ethrex.xyz/benchmarks/.

This module was removed in favor of the new `Report`, but I wasn't aware that its functionality was used for that.

**Description**

- Add logic to convert the `Report` into bench file content.
- Re-adds the bench file feature under the feature flag `ci`.
- Add caution comments to not remove this logic under any circumstances unless a refactor is made.

Closes #2756


cmd/ethrex_replay/src/cli.rs
```diff
@@ -181,6 +181,20 @@ pub struct EthrexReplayOptions {
         required = false
     )]
     pub verbose: bool,
+    // CAUTION
+    // This flag is used to create a benchmark file that is used by our CI for
+    // updating benchmarks from https://docs.ethrex.xyz/benchmarks/.
+    // Do no remove it under any circumstances, unless you are refactoring how
+    // we do benchmarks in CI.
+    #[cfg(feature = "ci")]
```
mpaulucci: do we need to have this behind a feature flag? Also, if we need to, maybe I would name it differently. It is used by `ci` but maybe something like `bench` would be more appropriate. But again, no flag seems the best solution IMO

ilitteri: I used it because it was already there. But yeah, I agree that removing it is a good idea.

ilitteri: Done in [b2a004a](https://github.com/lambdaclass/ethrex/pull/4683/commits/b2a004af1a5a8781a670cbc400267263277ae415) and [fdb0d4e](https://github.com/lambdaclass/ethrex/pull/4683/commits/fdb0d4ed769aeaff172efc93fa7477f748310922)

cmd/ethrex_replay/src/report.rs
```diff
@@ -204,6 +204,37 @@ impl Report {
             }
         }
     }
+
+    /// Convert the report to a benchmark file in JSON format.
+    ///
+    /// # CAUTION
+    ///
+    /// This functions is used to create a benchmark file that is used by our CI
```
JereSalo:
```suggestion
    /// This function is used to create a benchmark file that is used by our CI
```

cmd/ethrex_replay/src/report.rs
```diff
@@ -204,6 +204,37 @@ impl Report {
...
+    /// for updating benchmarks from https://docs.ethrex.xyz/benchmarks/.
+    ///
+    /// Do no remove it under any circumstances, unless you are refactoring how
```
JereSalo:
```suggestion
    /// Do not remove it under any circumstances, unless you are refactoring how
```

JereSalo (approved):

PR #4684, opened 2025-09-29
simple signed remaining gas

edg-l:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/vm/levm/src/gas_cost.rs
+++ b/crates/vm/levm/src/gas_cost.rs
@@ -956,11 +956,13 @@ pub fn ecpairing(groups_number: usize) -> Result<u64, VMError> {
 pub fn max_message_call_gas(current_call_frame: &CallFrame) -> Result<u64, VMError> {
     let mut remaining_gas = current_call_frame.gas_remaining;
 
-    remaining_gas = remaining_gas
-        .checked_sub(remaining_gas / 64)
-        .ok_or(InternalError::Underflow)?;
+    remaining_gas -= remaining_gas / 64;
 
-    Ok(remaining_gas)
+    if remaining_gas < 0 {
+        return Err(InternalError::Underflow.into());
+    }
+
+    Ok(remaining_gas as u64)
 }
 
 fn calculate_cost_and_gas_limit_call(
--- a/crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
+++ b/crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
@@ -162,7 +162,7 @@ impl<'a> VM<'a> {
 
         // EIP-2200
         let gas_left = self.current_call_frame.gas_remaining;
-        if gas_left <= SSTORE_STIPEND {
+        if gas_left as u64 <= SSTORE_STIPEND {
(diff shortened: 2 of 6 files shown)


crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -162,7 +162,7 @@ impl<'a> VM<'a> {
 
         // EIP-2200
         let gas_left = self.current_call_frame.gas_remaining;
-        if gas_left <= SSTORE_STIPEND {
+        if gas_left as u64 <= SSTORE_STIPEND {
```
Oppen: Maybe just change `SSTORE_STIPEND` to be `i64`.

crates/vm/levm/src/gas_cost.rs
```diff
@@ -956,11 +956,13 @@ pub fn ecpairing(groups_number: usize) -> Result<u64, VMError> {
...
 
-    remaining_gas = remaining_gas
-        .checked_sub(remaining_gas / 64)
-        .ok_or(InternalError::Underflow)?;
+    remaining_gas -= remaining_gas / 64;
 
-    Ok(remaining_gas)
+    if remaining_gas < 0 {
+        return Err(InternalError::Underflow.into());
+    }
```
JereSalo: I believe we can remove this check. We are doing `X = X - X/64` so it can't underflow

JereSalo (commented):
This will work for real use cases but gas should theoretically be 64 bits according to the protocol, so we have to decide whether to stop supporting cases in which gas is larger than 63 bits or not.

edg-l:
> This will work for real use cases but gas should theoretically be 64 bits according to the protocol, so we have to decide whether to stop supporting cases in which gas is larger than 32 bits or not.

Agree, thats why the pr is in draft yet, we need to. decide for example whether we enforce a maximum gas limit below i64::max

MegaRedHand:
[EIP-7825](https://eips.ethereum.org/EIPS/eip-7825) comes in Fusaka, limiting the gas limit per transaction to 2^24. I think it's safe to use `i64` here, even `i32`, given that a tx with a bigger gas limit never happened before and won't happen after Fusaka.

JereSalo:
> [EIP-7825](https://eips.ethereum.org/EIPS/eip-7825) comes in Fusaka, limiting the gas limit per transaction to 2^24. I think it's safe to use `i64` here, even `i32`, given that a tx with a bigger gas limit never happened before and won't happen after Fusaka.

Yeah that's right, just have to make sure to skip related tests and comment why. We should also add a comment above the `i64` variable definition justifying briefly why we use `i64` and not `u64` in case somebody doesn't know

JereSalo (approved):
Good
What if we change the rest of the gas variables to `i64` too? Would that make more sense instead of converting types every time? Just wanted to know your opinion, it's not necessary to do it in this PR

Arkenan (approved):

edg-l:
> Good What if we change the rest of the gas variables to `i64` too? Would that make more sense instead of converting types every time? Just wanted to know your opinion, it's not necessary to do it in this PR

we would need to see if its worth it

PR #4689, opened 2025-09-29
perf(l1): snap sync optimizations

fedacking:
**Motivation**

This PR aims to improve snap sync performance by

- Downloading all slots of big accounts during the initial leaves download step of snap sync
- Inserting intelligently accounts with the same state root and few (<= slots)
- Improving the performance of state trie through an ordered insertion algorithm

**Changes**

Sorted trie:
- The sorted trie uses the property that a tree can be written only once if it's written in order
- Added a single shot execution threadpool to parallelize writes and reads during the computing of tries
- Compute trie sends to write a separate task.

**Testing**

This PR is running on servers continously for testing performance.


crates/storage/trie_db/rocksdb.rs
```diff
@@ -79,6 +80,23 @@ impl TrieDB for RocksDBTrieDB {
             .write(batch)
             .map_err(|e| TrieError::DbError(anyhow::anyhow!("RocksDB batch write error: {}", e)))
     }
+
+    fn put_batch_no_alloc(&self, key_values: &[ethrex_trie::Node]) -> Result<(), TrieError> {
+        let cf = self.cf_handle()?;
+        let mut batch = rocksdb::WriteBatchWithTransaction::default();
+        let mut buffer = Vec::with_capacity(300);
```
Oppen: Use at least 532 bytes, as that's what a full branch node takes.

crates/storage/trie_db/rocksdb.rs
```diff
@@ -79,6 +80,23 @@ impl TrieDB for RocksDBTrieDB {
...
+
+        for node in key_values {
+            let db_key = self.make_key(node.compute_hash());
```
Oppen: You should probably be passing the hashes along to avoid recomputing them here.

crates/networking/p2p/utils.rs
```diff
@@ -82,6 +110,70 @@ pub fn dump_to_file(path: &Path, contents: Vec<u8>) -> Result<(), DumpError> {
...
+    pub accounts: Vec<H256>,
+    /// All slots in the trie from the accounts
+    pub storages: Vec<(H256, U256)>,
+}
+
+pub fn dump_storages_to_file(
+    path: &Path,
+    storages: Vec<AccountsWithStorage>,
+) -> Result<(), DumpError> {
+    cfg_if::cfg_if! {
```
Oppen: Please don't contribute to `cfg_if` proliferation. Just use `#[cfg]` at the blocks, or `if cfg!()` when possible to ensure the compiler checks all possibilities.

crates/networking/p2p/utils.rs
```diff
@@ -82,6 +110,70 @@ pub fn dump_to_file(path: &Path, contents: Vec<u8>) -> Result<(), DumpError> {
...
+        if #[cfg(feature = "rocksdb")] {
+            dump_to_rocks_db(
+                path,
+                accounts
+                    .into_iter()
+                    .map(|(hash, state)| (hash.0.to_vec(), state.encode_to_vec())
+                ).collect::<Vec<_>>()
+            )
```
Oppen: This is essentially cloning all the accounts state, as well as making a ton of allocations for calls to `to_vec` that don't seem necessary.

crates/networking/p2p/utils.rs
```diff
@@ -82,6 +110,70 @@ pub fn dump_to_file(path: &Path, contents: Vec<u8>) -> Result<(), DumpError> {
...
+                storages
+                    .into_iter()
+                    .flat_map(|accounts_with_slots| {
+                        accounts_with_slots.accounts.into_iter().map(|hash| {
+                            accounts_with_slots.storages.iter().map(move |(slot_hash, slot_value)| {
+                                let key = [hash.as_bytes(), slot_hash.as_bytes()].concat();
+                                (key, slot_value.encode_to_vec())
+                            }).collect::<Vec<_>>()
+                        }).collect::<Vec<_>>()
+                    }).flatten().collect::<Vec<_>>()
```
Oppen: This does many clones of big data as well I think. For each storage slot it allocates two vectors (`.concat()` and `encode_to_vec()`), possibly more internally, then for each account it allocates a new vector with all of its storage slots, and then it creates yet another big vector that needs to be alive at the same time as the others for the `.flatten().collect()` call.
I think it would be better to just have:
```rust
#[cfg(feature = "rocksdb")]
fn dump_storages_to_file(...) { ... }
#[cfg(not(feature = "rocksdb")]
fn dump_storages_to_file(...) { ... }
```
And the same for accounts.
That way, the RocksDB implementation can go directly to the specialized work without the intermediate allocs.

crates/networking/p2p/sync/code_collector.rs
```diff
@@ -39,6 +39,8 @@ impl CodeHashCollector {
         self.buffer.insert(hash);
     }
 
+    // Used depending on if the feature flag rocksdb is used
+    #[allow(dead_code)]
```
Oppen: Maybe just use the feature flag then?

fedacking: Done in [1e0910a](https://github.com/lambdaclass/ethrex/pull/4689/commits/1e0910aa32ae5a7543c5ac6818672b4962599bb1)

crates/storage/trie_db/rocksdb.rs
fedacking: Done in [ac7a8b9](https://github.com/lambdaclass/ethrex/pull/4689/commits/ac7a8b9b557612422df9d83277e10f833eeb8733).

crates/storage/trie_db/rocksdb.rs
fedacking: Done in [6d0e8d8](https://github.com/lambdaclass/ethrex/pull/4689/commits/6d0e8d87307272d4497e89bce887d4e590a552b5).

crates/networking/p2p/utils.rs
fedacking: Done in [ca3942a](https://github.com/lambdaclass/ethrex/pull/4689/commits/ca3942ae6b358fa31e5dc3bc6c3ad122ccc9b685)

crates/networking/p2p/utils.rs
fedacking: Done in [7a14f42](https://github.com/lambdaclass/ethrex/pull/4689/commits/7a14f4240500ee9aa7c7f2498e9d2ca2de04b8e8).

crates/networking/p2p/utils.rs
fedacking: Done in [7a14f42](https://github.com/lambdaclass/ethrex/pull/4689/commits/7a14f4240500ee9aa7c7f2498e9d2ca2de04b8e8)

jrchatruc (approved):

crates/common/trie/trie_sorted.rs
```diff
@@ -0,0 +1,451 @@
...
+}
+
+// The large size isn't a performance problem because we use a single instance of this
+// struct
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Clone)]
+enum CenterSideElement {
+    Branch { node: BranchNode },
+    Leaf { value: Vec<u8> },
+}
```
Oppen, line 24: Not to handle necessarily for this PR, but having `Leaf` contain a `Vec` might force unnecessary allocations, specially if we replace a `Leaf` with a `Branch` and then a `Leaf` again, we'll free the first vector and reallocate afterwards.

crates/common/trie/trie_sorted.rs
```diff
@@ -0,0 +1,451 @@
...
+            .enumerate()
+            .filter_map(|(index, child)| child.is_valid().then_some(index))
+            .collect::<Vec<_>>()
+    );
+    nodes_to_write.push((node.compute_hash(), node));
+    Ok(())
+}
+
+fn flush_nodes_to_write(
+    mut nodes_to_write: Vec<(NodeHash, Node)>,
```
Oppen, line 129: I think a comment explaining the flow of this vector is deemed. At first glance one would think we discard it, but then it's sent to `sender` to recycle.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 
 - Change remaining_gas to i64, improving performance in gas cost calculations [#4684](https://github.com/lambdaclass/ethrex/pull/4684)
 
+### 2025-09-30
+
+- Downloading all slots of big accounts during the initial leaves download step of snap sync [#4689](https://github.com/lambdaclass/ethrex/pull/4689)
+- Inserting intelligently accounts with the same state root and few (<= slots) [#4689](https://github.com/lambdaclass/ethrex/pull/4689)
```
Oppen:
```suggestion
- Downloading and inserting intelligently accounts with the same state root and few (<= slots) [#4689](https://github.com/lambdaclass/ethrex/pull/4689)
```

crates/concurrency/concurrency.rs
```diff
@@ -0,0 +1,51 @@
+use crossbeam::channel::{Sender, select_biased, unbounded};
```
Oppen, line 1: Let's add a comment explaining why we're doing our own threadpool. Explain the problems we found and that we preferred a simpler solution that helped us model correctly how things are going to work.
Otherwise future us will be tempted to delete this and use rayon, going full circle.

crates/networking/p2p/utils.rs
```diff
@@ -64,6 +72,56 @@ pub fn get_account_storages_snapshot_file(directory: &Path, chunk_index: u64) ->
...
+    path: &Path,
+    mut contents: Vec<(H256, H256, U256)>,
+) -> Result<(), rocksdb::Error> {
+    contents.sort();
+    contents.dedup_by_key(|(k0, k1, _)| {
+        let mut buffer = [0_u8; 64];
+        buffer[0..32].copy_from_slice(&k0.0);
+        buffer[32..64].copy_from_slice(&k1.0);
+        buffer
+    });
```
Oppen, line 109: Not necessary to fix now, but I believe the following should work and be more brief:
```suggestion
    contents.dedup_by_key(|(k0, k1, _)| (k0, k1));
```

crates/networking/p2p/utils.rs
```diff
@@ -64,6 +72,56 @@ pub fn get_account_storages_snapshot_file(directory: &Path, chunk_index: u64) ->
...
+    }
+    writer.finish()
+}
+
+#[cfg(feature = "rocksdb")]
+pub fn dump_storages_to_rocks_db(
+    path: &Path,
+    mut contents: Vec<(H256, H256, U256)>,
+) -> Result<(), rocksdb::Error> {
+    contents.sort();
```
Oppen, line 103: The following might be faster:
```suggestion
    contents.sort_unstable();
```

crates/storage/trie_db/rocksdb.rs
```diff
@@ -40,7 +41,7 @@ impl RocksDBTrieDB {
             .ok_or_else(|| TrieError::DbError(anyhow::anyhow!("Column family not found")))
     }
 
-    fn make_key(&self, node_hash: NodeHash) -> Vec<u8> {
+    fn make_key(&self, node_hash: &NodeHash) -> Vec<u8> {
```
Oppen, line 44: I think `NodeHash` should be copy instead. Isn't it capped to 32 bytes and the discrimnator?

crates/networking/p2p/peer_handler.rs
```diff
@@ -1279,21 +1255,62 @@ impl PeerHandler {
...
+        // pivots.
+        // At a later time we might try to also store only once and insert for all.
+        // That should help at least to do less `compute_storage_roots`, but skipping
+        // that might be problematic for healing.
+        // We can also sort by decreasing number of repetitions, so we download
+        // and settle the most common first.
+        // AFTER: try to reduce memory usage from account filtering.
+        // It currently takes about 68B per account with storages, with ~25M of them,
+        // meaning 1.7GB. Possibly several copies of this.
+        // THEN: review storage formats, maybe play with memory mapped data.
```
Oppen: This comment was for myself. Let's remove it.

Oppen: Note this PR already implements most of that.

crates/networking/p2p/peer_handler.rs
```diff
@@ -1302,10 +1319,11 @@ impl PeerHandler {
...
 
-        // 2) request the chunks from peers
-        let mut all_account_storages =
-            vec![vec![]; account_storage_roots.accounts_with_storage_root.len()];
-
+        // let all_account_hashes: Vec<H256> = account_storage_roots
+        //     .accounts_with_storage_root
+        //     .keys()
+        //     .copied()
+        //     .collect();
```
Oppen: I guess this can be removed too?

crates/networking/p2p/peer_handler.rs
```diff
@@ -1318,28 +1336,21 @@ impl PeerHandler {
...
+        // Maps storage root to vector of hashed addresses matching that root and
+        // vector of hashed storage keys and storage values.
+        let mut current_account_storages: BTreeMap<H256, AccountsWithStorage> = BTreeMap::new();
 
         debug!("Starting request_storage_ranges loop");
         loop {
-            if all_account_storages.iter().map(Vec::len).sum::<usize>() * 64 > RANGE_FILE_CHUNK_SIZE
+            if current_account_storages
+                .values()
+                .map(|accounts| 32 * accounts.accounts.len() + 32 * accounts.storages.len())
```
Oppen: There was a bug in size computation that we fixed with Javi in a different version:
```suggestion
                .map(|accounts| 32 * accounts.accounts.len() + 64 * accounts.storages.len())
```

crates/networking/p2p/peer_handler.rs
jrchatruc: removed here [c079384](https://github.com/lambdaclass/ethrex/pull/4689/commits/c07938469c4ae9e6040182a8e9f8f6ecddbb3122)

crates/networking/p2p/peer_handler.rs
jrchatruc: removed here [c079384](https://github.com/lambdaclass/ethrex/pull/4689/commits/c07938469c4ae9e6040182a8e9f8f6ecddbb3122)

crates/networking/p2p/peer_handler.rs
jrchatruc: Good catch, done here [a497870](https://github.com/lambdaclass/ethrex/pull/4689/commits/a4978705e5786544573bf3be4a3cca8e4da5d807)

crates/networking/p2p/sync.rs
```diff
@@ -1499,3 +1426,336 @@ pub async fn validate_bytecodes(store: Store, state_root: H256) -> bool {
...
+        let store_clone = store.clone();
+        let pivot_hash_moved = pivot_header.hash();
+        info!("Starting compute of account_storages_snapshot");
+        let storage_trie_node_changes = tokio::task::spawn_blocking(move || {
+            let store: Store = store_clone;
+
+            account_storages_snapshot
+                .into_par_iter()
+                .flat_map(|account_storages| {
+                    let storages: Arc<[_]> = account_storages.storages.into();
```
Oppen, line 1541: I think you'd want to do this earlier. At this point it's the same as cloning I think. But even then, `into_iter`/`into_par_iter` already move the contents, so the `clone` was not necessary either.
Let's make an issue to deal with this later.

crates/networking/p2p/sync.rs
```diff
@@ -1499,3 +1426,336 @@ pub async fn validate_bytecodes(store: Store, state_root: H256) -> bool {
...
+    db_options.create_if_missing(true);
+    let db = rocksdb::DB::open(&db_options, get_rocksdb_temp_accounts_dir(datadir))
+        .map_err(|_| SyncError::AccountTempDBDirNotFound)?;
+    let file_paths: Vec<PathBuf> = std::fs::read_dir(account_state_snapshots_dir)
+        .map_err(|_| SyncError::AccountStateSnapshotsDirNotFound)?
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|_| SyncError::AccountStateSnapshotsDirNotFound)?
+        .into_iter()
+        .map(|res| res.path())
+        .collect();
```
Oppen, line 1591: We can probably simplify this and avoid the double collect by mapping first.

Oppen (approved):

PR #4691, opened 2025-09-29
feat(replay): endless replay & only eth proofs blocks

ilitteri:
**Motivation**

We want to remove the `ethrex-replayer` internal tool. This is the first step; we still need to support requesting multiple RPCs.

**Description**

- Adds the flag `--endless` to `BlockOptions`. This allows replaying blocks endlessly (once it finishes replaying one block, it fetches the latest and starts replaying it, and so on).
- Adds the flag `--endless` to `BlocksOptions`. Similar to the above, but only works if `--from <BLOCK>` is set, replaying endlessly from `BLOCK`.
- Adds the flag `--only-eth-proofs-blocks` to `BlockOptions` and `BlocksOptions`. If set, it constrains which blocks are executed.
- Adds the above to the cache commands (the same but skips execution).

### DIFF
--- a/cmd/ethrex_replay/src/cli.rs
+++ b/cmd/ethrex_replay/src/cli.rs
@@ -254,7 +254,7 @@ pub enum CacheLevel {
     On,
 }
 
-#[derive(Parser)]
+#[derive(Parser, Clone)]
 pub struct BlockOptions {
     #[arg(
         help = "Block to use. Uses the latest if not specified.",
@@ -298,6 +298,14 @@ pub struct BlocksOptions {
         help_heading = "Command Options"
     )]
     to: Option<u64>,
+    #[arg(
+        long,
+        help = "Only fetch Ethereum proofs blocks (i.e., no L2 blocks).",
+        help_heading = "Replay Options",
+        requires = "rpc_url",
+        conflicts_with = "blocks"
+    )]
+    pub only_eth_proofs_blocks: bool,
     #[command(flatten)]
     opts: EthrexReplayOptions,
 }
@@ -341,11 +349,21 @@ impl EthrexReplayCommand {
     pub async fn run(self) -> eyre::Result<()> {
         match self {
             #[cfg(not(feature = "l2"))]
-            Self::Block(mut block_opts) => loop {
-                let _ = replay_block(block_opts).await;
+            Self::Block(block_opts) => loop {
+                let start = Instant::now();
+
+                let _ = replay_block(block_opts.clone()).await;
 
                 if !block_opts.endless {
                     break;
+                }
+
+                let elapsed = start.elapsed();
+
+                // Sleep until the next block time (12 seconds)
+                // If the replay took longer than 12 seconds, start immediately
+                if elapsed < Duration::from_secs(12) {
+                    tokio::time::sleep(Duration::from_secs(12) - elapsed).await;
                 }
             },
             #[cfg(not(feature = "l2"))]
@@ -353,6 +371,7 @@ impl EthrexReplayCommand {
                 blocks,
                 from,
                 to,
+                only_eth_proofs_blocks,
                 opts,
             }) => {
                 if opts.cached {
@@ -376,7 +395,7 @@ impl EthrexReplayCommand {
                     replay_block(BlockOptions {
                         block: Some(*block_number),
                         endless: false,
-                        only_eth_proofs_blocks: false,
+                        only_eth_proofs_blocks,
                         opts: opts.clone(),
                     })
                     .await?;
@@ -412,6 +431,7 @@ impl EthrexReplayCommand {
                 blocks,
                 from,
                 to,
+                only_eth_proofs_blocks,
                 opts,
             })) => {
                 let blocks = resolve_blocks(blocks, from, to, opts.rpc_url.clone()).await?;
@@ -423,7 +443,7 @@ impl EthrexReplayCommand {
                         eth_client.clone(),
                         network.clone(),
                         BlockIdentifier::Number(block_number),
-                        opts.only_eth_proofs_blocks,
+                        only_eth_proofs_blocks,
                     )
                     .await?;
                 }
@@ -721,7 +741,7 @@ async fn replay_transaction(tx_opts: TransactionOpts) -> eyre::Result<()> {
         eth_client,
         network,
         BlockIdentifier::Number(tx.block_number.as_u64()),
-        tx_opts.opts.only_eth_proofs_blocks,
+        false,
     )
     .await?;
 
@@ -751,7 +771,7 @@ async fn replay_block(block_opts: BlockOptions) -> eyre::Result<()> {
         eth_client,
         network.clone(),
         or_latest(block)?,
-        opts.only_eth_proofs_blocks,
+        block_opts.only_eth_proofs_blocks,
     )
     .await?;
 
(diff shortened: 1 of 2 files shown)


cmd/ethrex_replay/src/cli.rs
```diff
@@ -327,74 +356,173 @@ impl EthrexReplayCommand {
...
+
+                    // Replay endlessly
+                    Box::pin(async {
+                        Self::Block(BlockOptions {
+                            block: None,
+                            endless,
+                            only_eth_proofs_blocks,
+                            opts: opts.clone(),
+                        })
+                    })
```
tomip01: wouldn't this result in replaying `from` and then from `latest`? If there are some blocks between them, they won't be replayed

cmd/ethrex_replay/src/fetcher.rs
```diff
@@ -25,8 +25,27 @@ pub async fn get_blockdata(
...
+        info!(
+            "Latest block is {latest_block_number}, waiting for next eth proofs block ({}) in ~{}",
+            latest_block_number + blocks_left_for_next_eth_proofs_block,
+            format_duration(time_for_next_eth_proofs_block)
+        );
+
+        tokio::time::sleep(time_for_next_eth_proofs_block).await;
+
+        latest_block_number = eth_client.get_block_number().await?.as_u64();
+    }
```
JereSalo: what if we want to run only eth proofs block of blocks that already exist? Like from block 100 to 2000, that would mean running 100, 200, ..., 2000. We don't need the sleep in that scenario

JereSalo (commented):
Something is a little bit off. You can use endless both in `Block` and in `Blocks`. Then it will be the same to do `blocks --from 100 --endless` than `block --block 100 --endless`. I'd leave the endless option just for either `Block` or `Blocks`, but not both. 
For it to make the most sense I'd leave it in `blocks` parameters and I think it should only work if `to` (the end) is `None`.
Wdyt?

cmd/ethrex_replay/src/cli.rs
tomip01: Also I think we have to call `.run()`

ilitteri: You are right.

ilitteri:
> Something is a little bit off. You can use endless both in `Block` and in `Blocks`. Then it will be the same to do `blocks --from 100 --endless` than `block --block 100 --endless`. I'd leave the endless option just for either `Block` or `Blocks`, but not both. For it to make the most sense I'd leave it in `blocks` parameters and I think it should only work if `to` (the end) is `None`. Wdyt?

I agree. Endless goes better with blocks, I'll make it work just with it.

cmd/ethrex_replay/src/fetcher.rs
ilitteri: You are right. I'll take this into consideration.

cmd/ethrex_replay/src/cli.rs
```diff
@@ -327,37 +342,127 @@ impl EthrexReplayCommand {
...
+                            .await
+                        })
+                        .await?;
+
+                        let elapsed = start.elapsed();
+
+                        // Sleep until the next block time (12 seconds)
+                        if elapsed < Duration::from_secs(12) {
+                            tokio::time::sleep(Duration::from_secs(12) - elapsed).await;
+                        }
```
JereSalo: Is it possible for us to merge this with the code above this by manually setting `from` to the latest block? Logic seems similar. Not critical though.
It is like this case but with from set at runtime:
`                // * If --endless is set, we start from --from and keep checking for new blocks`

cmd/ethrex_replay/src/cli.rs
```diff
@@ -1205,10 +1375,49 @@ pub async fn produce_custom_l2_block(
...
+
+        info!(
+            "Latest block is {latest_block_number}, waiting for next eth proofs block ({}) in ~{}",
+            latest_block_number + blocks_left_for_next_eth_proofs_block,
+            format_duration(&time_for_next_eth_proofs_block)
+        );
+
+        tokio::time::sleep(time_for_next_eth_proofs_block).await;
+
+        latest_block_number = eth_client.get_block_number().await?.as_u64();
```
JereSalo, line 1394: This seems to be error prone if for some reason the latest block number is not multiple of 100, ideally it wouldn't happen but it can happen I think.

cmd/ethrex_replay/src/cli.rs
ilitteri: It's a good idea, I'll try the change and see how it goes.

cmd/ethrex_replay/src/cli.rs
ilitteri: [f7dc0ed](https://github.com/lambdaclass/ethrex/pull/4691/commits/f7dc0ed0c995e119c485fb04752d09d8420fc458)

cmd/ethrex_replay/src/cli.rs
ilitteri: [f7dc0ed](https://github.com/lambdaclass/ethrex/pull/4691/commits/f7dc0ed0c995e119c485fb04752d09d8420fc458)

cmd/ethrex_replay/src/cli.rs
```diff
@@ -284,6 +284,21 @@ pub struct BlocksOptions {
...
+        conflicts_with_all = ["blocks", "to"]
+    )]
+    pub endless: bool,
+    #[arg(
+        long,
+        help = "Only fetch Ethereum proofs blocks (i.e., no L2 blocks).",
+        help_heading = "Replay Options",
+        conflicts_with = "blocks"
+    )]
+    pub only_eth_proofs_blocks: bool,
```
JereSalo, line 300: Somewhere in here we could link their website. Maybe in the help, to give context. wdyt?
https://ethproofs.org/blocks

cmd/ethrex_replay/src/cli.rs
```diff
@@ -1205,10 +1369,49 @@ pub async fn produce_custom_l2_block(
 }
 
 #[cfg(not(feature = "l2"))]
-async fn fetch_latest_block_number(rpc_url: Url) -> eyre::Result<u64> {
+async fn fetch_latest_block_number(
```
JereSalo, line 1371: I'd add brief docs to this function explaining what it does when `only_eth_proofs_blocks` is set to true. That it gets the latest but only after being sure that there's a block multiple of 100 in between because we are interested in those only.

JereSalo (commented):
- Clap doesn't let me set `endless` without `from`
- There is something off with the `only_eth_proofs_blocks` thing. If I specify `--from <old_block> --only-eth-proofs-blocks` when fetching the `to` it will fetch the latest block number and if it's not an ethproofs block it will make me wait and it shouldn't because there already are available blocks to execute. It's just a detail though.

ilitteri:
> - Clap doesn't let me set `endless` without `from`
> 
> - There is something off with the `only_eth_proofs_blocks` thing. If I specify `--from <old_block> --only-eth-proofs-blocks` when fetching the `to` it will fetch the latest block number and if it's not an ethproofs block it will make me wait and it shouldn't because there already are available blocks to execute. It's just a detail though.

Thefirst one is a new one to me. I tried it yesterday and worked. Have you pulled the latest changes? I'll re run it anyways to double check. 

Regarding the second one, it's a known fact. I chose not to handle that special case because it makes the codebase more complex, and sooner or later waiting for the next eth proofs block will be needed.

PR #4697, opened 2025-09-29
feat(l1,l2): add tooling to migrate from libmdbx to rocksdb

ManuelBilbao:
**Motivation**

With the implementation of RocksDB as the default database, current networks need a tool to migrate its old DB.

**Description**

Create a tool for migrations. This tool supports adding new migration options in the future.


LeanSerra (commented):
Edit: All the problems were fixed by setting the migrated blocks as canonical when calling `forkchoice_update` on the new store

Let's add some kind of README with instructions on how to use the tool.
I tried migrating the database from a staging environment with ethrex version 0.0.4-alpha and I can successfully start up the sequencer but when I try to query blocks with:
```
curl -X POST http://localhost:1729 \
  -H "Content-Type: application/json" \
  --data '{
    "jsonrpc":"2.0",
    "method":"eth_getBlockByNumber",
    "params":["0x3", true],
    "id":1
  }'

{"id":1,"jsonrpc":"2.0","result":null}
``` 
I get:
`{"id":1,"jsonrpc":"2.0","result":null}`
The genesis block seems to be there (querying for 0x0 works) and also the latest  but not any of the blocks in between and block production is working (the chain is advancing)
Asking for balances seems to work, asking for contract codes also works, but querying for txs fails.

tooling/migrations/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Instructions
+
+It is recomended to backup the original database before migration even if this process does not erase the old data. To migrate a database run:
+
+```
+cargo run --release -- l2r --genesis <GENESIS_PATH> --store.old <OLD_STORAGE_PATH> --store.new <NEW_STORAGE_PATH>
+```
+
+This will output the migrated database to `<NEW_STORAGE_PATH>`. If you are migrating a db from an ethrex L2 rollup you should also copy the `rollup_store`, `rollup_store-shm` and `rollup_store-wal` files to `<NEW_STORAGE_PATH>`.
```
MegaRedHand: We should put this part about the L2 somewhere more visible. Maybe in a blockquote like:

```markdown
> [!IMPORTANT]  
> If you are migrating a db from an ethrex L2 rollup you should also copy the `rollup_store`, `rollup_store-shm` and `rollup_store-wal` files to `<NEW_STORAGE_PATH>`.

MegaRedHand (approved):

tooling/migrations/README.md
LeanSerra: Agreed! done in [49751b8](https://github.com/lambdaclass/ethrex/pull/4697/commits/49751b80acdccb22a3d3d7ff5b0847ffc3ef6967)

PR #4699, opened 2025-09-29
chore(l1,l2): remove unneeded crates

ManuelBilbao:
**Motivation**

We don't want unnecessary crates.

**Description**

Removed `rustc-hex` and `tempfile` crates

### DIFF
--- a/crates/l2/networking/rpc/signer.rs
+++ b/crates/l2/networking/rpc/signer.rs
@@ -10,7 +10,6 @@ use ethrex_common::{
 };
 use ethrex_rlp::encode::PayloadRLPEncode;
 use reqwest::{Client, StatusCode, Url};
-use rustc_hex::FromHexError;
 use secp256k1::{Message, PublicKey, SECP256K1, SecretKey};
 use serde::Serialize;
 use url::ParseError;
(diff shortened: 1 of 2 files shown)


crates/l2/networking/rpc/signer.rs
```diff
@@ -143,7 +142,7 @@ impl RemoteSigner {
                 .text()
                 .await?
                 .parse::<Signature>()
-                .map_err(SignerError::FromHexError),
+                .map_err(|_| SignerError::FromHexError),
```
MegaRedHand, line 145: That it's being used for an error variant means it's a transitive dependency. Removing it from the `Cargo.toml` won't do much for removing dependencies.

cmd/ethrex/bench/build_block_benchmark.rs
```diff
@@ -118,9 +119,9 @@ fn recover_address_for_sk(sk: &SecretKey) -> Address {
 }
 
 async fn setup_genesis(accounts: &Vec<Address>) -> (Store, Genesis) {
-    let storage_path = tempfile::TempDir::new().unwrap();
-    if std::fs::exists(&storage_path).unwrap_or(false) {
-        std::fs::remove_dir_all(&storage_path).unwrap();
+    let storage_path = Path::new(".temp");
+    if std::fs::exists(storage_path).unwrap_or(false) {
+        std::fs::remove_dir_all(storage_path).unwrap();
```
MegaRedHand, line 124: I think this is a bad idea. Using some fixed file path everywhere quickly leads to weird errors once we try to run multiple tests in parallel, or if we forget to clean it up, or if things go wrong in general (i.e. when panic or segfaults happen).

ManuelBilbao:
We'll look better solutions

PR #4701, opened 2025-09-30
fix(common): TrieIterator advance skipping equal-prefix Branch subtree

sashass1315:
**Motivation**

Nibbles::next_choice() returns None when the key is fully consumed (or at the leaf-flag). The code treated this as “nothing to push”, violating ≥ semantics for range starts and causing iteration to begin strictly after the correct first element.

**Description**

Problem: advance() skipped the entire equal-prefix subtree when the search key was exhausted at a Branch. Specifically, first_ge returned early on Node::Branch when target_nibbles.next_choice() was None, so the iterator omitted the branch’s own value (exact key match) and all of its descendants (all ≥ key).

Fix: In the Branch/None case, seed the stack with all valid children (ascending index) and the branch itself (pushed last so that after the final reverse the branch is yielded first). This ensures the branch value (== key) is produced first, followed by all ≥ descendants in order.

### DIFF
--- a/crates/common/trie/trie_iter.rs
+++ b/crates/common/trie/trie_iter.rs
@@ -48,29 +48,44 @@ impl TrieIterator {
             match &next_node {
                 Node::Branch(branch_node) => {
                     // Add all children to the stack (in reverse order so we process first child frist)
-                    let Some(choice) = target_nibbles.next_choice() else {
-                        return Ok(());
-                    };
-                    let child = &branch_node.choices[choice];
-                    // If a prefix of `key` exists under this branch, we recur to the child node, skipping
-                    // the branch itself to avoid iterating lesser keys.
-                    if child.is_valid() {
-                        first_ge(
-                            db,
-                            prefix_nibbles.append_new(choice as u8),
-                            target_nibbles,
-                            child.clone(),
-                            new_stack,
-                        )?;
-                    }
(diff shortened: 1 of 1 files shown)


Oppen:
> Nibbles::next_choice() returns None when the key is fully consumed (or at the leaf-flag). The code treated this as “nothing to push”, violating ≥ semantics for range starts and causing iteration to begin strictly after the correct first element.

There might be a naming issue at play, but that's the intended behavior. Our use case for this is just iterating over leaves, and the prefix needs to be a whole key. A valid trie in all practical cases in Ethereum never has keys pointing to branches.

sashass1315:
> > Nibbles::next_choice() returns None when the key is fully consumed (or at the leaf-flag). The code treated this as “nothing to push”, violating ≥ semantics for range starts and causing iteration to begin strictly after the correct first element.
> 
> There might be a naming issue at play, but that's the intended behavior. Our use case for this is just iterating over leaves, and the prefix needs to be a whole key. A valid trie in all practical cases in Ethereum never has keys pointing to branches.

I updated the exhausted-key-at-Branch handling to enqueue only valid children

mpaulucci:
Needs a follow up from @Oppen , setting as draft for now

ilitteri:
Closed as assumed abandoned

PR #4708, opened 2025-09-30
docs(l1): adding healing documentation

fedacking:
**Motivation**

During our development we developed the healing algorithm based on some assumptions, this adds documentation for it.

**Description**

- Uploads documentation relating to healing in snap sync and account deletion in ethereum.


docs/internal/l1/delete_accounts.md
```diff
@@ -0,0 +1,29 @@
...
+- Externaly Owned Accounts (EOA): accounts for general users to transfer eth and call contracts.
+- Contracts: which execute code and store data.
+
+Creating EOA is done through sending ETH into a new address, at which point the account is created and added into the state trie.
+
+Creating a contract can be done through the CREATE and [CREATE2](https://eips.ethereum.org/EIPS/eip-1014) opcode. Notably, those opcodes check that the account is created at an address where the code is empty and the nonce is zero, but **it doesn't check balance**. As such, a contract can be created through taking over an existing account.
+
+During the creating of a contract, the `init_code` is run which can include the [self destruct opcode](https://eips.ethereum.org/EIPS/eip-6780) that deletes the contract in the same transaction it was created. Normally, this deletes an account that was created in the same transaction (because contracts are usually created over empty accounts) but in this case the account already existed because it  already had some balance. This is the only edge case in which an account can go from existing to non-existing from one block to another after the Cancun fork.
```
MegaRedHand:
```suggestion
During the creating of a contract, the `init_code` is run which can include the [self destruct opcode](https://eips.ethereum.org/EIPS/eip-6780) that deletes the contract in the same transaction it was created. Normally, this deletes an account that was created in the same transaction (because contracts are usually created over empty accounts) but in this case the account already existed because it already had some balance. This is the only edge case in which an account can go from existing to non-existing from one block to another after the Cancun fork.
```

mpaulucci (approved):

PR #4710, opened 2025-09-30
fix(l1): use git tags for version output instead of Cargo.toml version

### LINKED ISSUE #3686: Make `--version` output consistent with release version
See https://github.com/lambdaclass/ethrex/pull/3505#discussion_r2213969532

Installing the v0.0.1-rc.1 version and looking at the version:

```sh
curl -L https://github.com/lambdaclass/ethrex/releases/download/v0.0.1-rc.1/ethrex-macos_aarch64 -o ethrex
chmod +x ethrex
./ethrex --version
```

Results in an output of:

```text
# note the v0.1.0 here
ethrex ethrex/v0.1.0-HEAD-d3aa87a/aarch64-apple-darwin/rustc-v1.87.0
```

We should ensure the two versions match.

FredPhilipy:
Closes #3686

Modified build.rs to extract version from git tags using `git describe --tags --abbrev=0` and updated get_client_version() to use it. When building from a tagged commit like v0.0.1-rc.1, the version output will now correctly show that tag instead of the hardcoded 0.1.0 from Cargo.toml. Falls back to CARGO_PKG_VERSION for untagged commits.

### DIFF
--- a/cmd/ethrex/build.rs
+++ b/cmd/ethrex/build.rs
@@ -2,6 +2,7 @@
 use std::error::Error;
 use std::fs::File;
 use std::io::BufReader;
+use std::process::Command;
 use std::{
     env, fs,
     path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


mpaulucci:
Sorry, I think the proper solution would be to update the Cargo.toml for each crate when releasing a new version.

PR #4712, opened 2025-09-30
feat(l1): run hive execute-blobs tests + add basic `getNewPayloadV5` + `getBlobsV2`

### LINKED ISSUE #3844: Add support for Blobs hive test
https://github.com/ethereum/hive/tree/master/simulators/ethereum/eest/execute-blobs

fmoletta:
**Motivation**
Be able to run hive eest/execute-blobs test suite

**Description**
* Add `run-hive-eest-blobs` makefile target
* Add `eest/execute-blobs` test suite to daily hive workflow
* Add endpoints `GetPayloadV5` and `BlobsBundleV2`
* Add `version` field to `BlobsBundle` and `wrapper_version` field to `WrappedEIP4844Transaction`
* Support cell proof verification for extended blobs

Closes #3844


crates/common/crypto/kzg.rs
```diff
@@ -15,6 +20,43 @@ impl From<kzg_rs::KzgError> for KzgError {
...
+/// as defined by EIP-7594.
+pub fn verify_cell_kzg_proof_batch(
+    blobs: &[Blob],
+    commitments: &[Commitment],
+    cell_proof: &[Proof],
+) -> Result<bool, KzgError> {
+    #[cfg(not(feature = "c-kzg"))]
+    return Err(KzgError::NotSupportedWithoutCKZG(String::from(
+        "Cell proof verification",
+    )));
```
MegaRedHand, line 46: Can we mark the whole function as `#[cfg(feature = "c-kzg")]` instead?

crates/common/crypto/kzg.rs
```diff
@@ -15,6 +20,43 @@ impl From<kzg_rs::KzgError> for KzgError {
...
+    #[cfg(feature = "c-kzg")]
+    {
+        let c_kzg_settings = c_kzg::ethereum_kzg_settings(8);
```
MegaRedHand, line 49: Should we move this to a `OnceCell` or similar?

crates/networking/rpc/engine/blobs.rs
```diff
@@ -48,23 +62,74 @@ impl RpcHandler for BlobsV1Request {
             return Err(RpcErr::TooLargeRequest);
         }
 
+        let mut res: Vec<Option<BlobAndProofV1>> = vec![None; self.blob_versioned_hashes.len()];
+
+        for blobs_bundle in context.blockchain.mempool.get_blobs_bundle_pool()? {
```
MegaRedHand, line 68: We should probably have blobs indexed by versioned hash. We should open an issue for that.

crates/common/types/blobs_bundle.rs
```diff
@@ -35,6 +35,8 @@ pub struct BlobsBundle {
     pub commitments: Vec<Commitment>,
     #[serde(with = "serde_utils::bytes48::vec")]
     pub proofs: Vec<Proof>,
+    #[serde(skip)]
+    pub version: Option<u8>,
```
MegaRedHand: Can we remove the `Option` and make it default to `0` if omitted?

crates/common/crypto/kzg.rs
fmoletta: Id rather handle this case here than on the calling code, as we do need to use this outside feature-gated code

crates/common/crypto/kzg.rs
pablodeymo: we don't need to use OnceCell since this function returns a 'static reference:
https://github.com/ethereum/c-kzg-4844/blob/7043f4d4dd7cedf510294d1336a24522c423d09d/bindings/rust/src/ethereum_kzg_settings/mod.rs#L42

crates/networking/rpc/engine/blobs.rs
fmoletta: Is it worth it though? If we did that we would have to compute the hash for each blob in a blob tx as soon as it reaches the mempool (delaying potential inclusion of said tx in a block).
We would be optimizing the `GetBlobs` rpc endpoints (which is the only point at which we need to get blobs by their hashes) at the cost of making the core mempool behaviour slower

crates/common/types/blobs_bundle.rs
fmoletta: sure! [7942f8e](https://github.com/lambdaclass/ethrex/pull/4712/commits/7942f8e570601eb8dbf80eeb2dd3825a5bc4907c)

crates/networking/rpc/engine/blobs.rs
MegaRedHand: Don't we need the versioned hash for VM execution (i.e. BLOBHASH) anyways?

crates/l2/prover/src/guest_program/src/risc0/Cargo.toml
```diff
@@ -33,7 +33,7 @@ k256 = { git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k25
 ed25519-dalek = { git = "https://github.com/risc0/curve25519-dalek", tag = "curve25519-4.1.2-risczero.0" }
 rsa = { git = "https://github.com/risc0/RustCrypto-RSA", tag = "v0.9.6-risczero.0" }
 crypto-bigint = { git = "https://github.com/risc0/RustCrypto-crypto-bigint", tag = "v0.5.5-risczero.0" }
-c-kzg = { git = "https://github.com/risc0/c-kzg-4844", tag = "c-kzg/v2.1.1-risczero.0" }
+c-kzg = { git = "https://github.com/lambdaclass/risc0-c-kzg-4844.git", tag = "c-kzg/v2.1.1-risczero.2-eip7594-default" }
```
ilitteri: 👍

MegaRedHand (approved):

crates/l2/prover/src/guest_program/src/risc0/Cargo.toml
ilitteri: Let's add a descriptive comment explaining why we are using a fork for this.

Oppen: These patches need comments explaining why we use them, specially when they imply maintaining a fork.

mpaulucci (approved):

crates/networking/rpc/engine/blobs.rs
fmoletta: BLOBHASH opcode reads a blob hash from the executed transaction, not from the mempool.
Though I now realize that we do get the blob hashes when we submit a blob tx to the mempool so we could sort them by hash without having to re-hash them with some changes to the api

crates/l2/prover/src/guest_program/src/risc0/Cargo.toml
fmoletta: Good call, I will also create an issue to check this again in the future so we can use the risc0 patch if they decide to stabilize EIP-7594 feature

fmoletta: Created https://github.com/lambdaclass/ethrex/issues/4738 & linked in toml [24d3a07](https://github.com/lambdaclass/ethrex/pull/4712/commits/24d3a079758d8080ffeb2dcf33690b55b6a3cf0a)

fmoletta: No longer needed after fix [484cb7c](https://github.com/lambdaclass/ethrex/pull/4712/commits/484cb7cbb64d69441456d58c0fd39e687b419d7a)

ilitteri (approved):

PR #4716, opened 2025-09-30
feat(replay): add `cached` feature

JereSalo:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/cmd/ethrex_replay/src/cli.rs
+++ b/cmd/ethrex_replay/src/cli.rs
@@ -148,6 +148,7 @@ pub struct EthrexReplayOptions {
     #[arg(
         long,
         group = "data_source",
+        help = "use cache as input instead of fetching from RPC",
         help_heading = "Replay Options",
         requires = "network",
         conflicts_with = "cache_level"
@@ -168,7 +169,13 @@ pub struct EthrexReplayOptions {
         help_heading = "Replay Options"
     )]
     pub cache_dir: PathBuf,
-    #[arg(long, default_value = "on", help_heading = "Replay Options")]
+    #[arg(
+        long,
+        default_value = "on",
+        help_heading = "Replay Options",
+        help = "Criteria to save a cache when fetching from RPC",
+        requires = "rpc_url"
+    )]
     pub cache_level: CacheLevel,
     #[arg(long, env = "SLACK_WEBHOOK_URL", help_heading = "Replay Options")]
     pub slack_webhook_url: Option<Url>,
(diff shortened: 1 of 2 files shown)


cmd/ethrex_replay/src/cli.rs
```diff
@@ -144,9 +144,23 @@ pub struct EthrexReplayOptions {
...
-    pub rpc_url: Url,
-    #[arg(long, group = "data_source", help_heading = "Replay Options")]
+    pub rpc_url: Option<Url>,
+    #[arg(
+        long,
+        group = "data_source",
+        help = "use cache as input instead of fetching from RPC",
+        help_heading = "Replay Options",
+        requires = "network",
+        conflicts_with = "cache_level"
```
ilitteri, line 155: `cached` should conflict with `endless` too (when used in `ethrex-replay blocks`). It is not strictly necessary because combining those will fail since `cached` needs `from` and `to` in `blocks` cmd (and those are incompatible with `endless`), but it'd be good UX to let clap enforce this and let you know in the console.

ilitteri (commented):
I think running `ethrex-replay blocks --cached --network <NETWORK>` should run all the cached blocks from `NETWORK` from the cache directory.

cmd/ethrex_replay/src/cli.rs
JereSalo: You are right! It's better to enforce it with clap
https://github.com/lambdaclass/ethrex/pull/4716/commits/b6e0d575b7d53523dd935c660c3bf76c18919fe8

cmd/ethrex_replay/src/cli.rs
```diff
@@ -956,20 +879,22 @@ async fn replay_block(block_opts: BlockOptions) -> eyre::Result<()> {
...
-                cache.delete()?;
+    // Decide whether or not to keep the cache when fetching data from RPC.
+    if !opts.cached {
+        match opts.cache_level {
+            // Cache is already saved
+            CacheLevel::On => {}
+            // Only save the cache if the block run or proving failed
+            CacheLevel::Failed => {
+                if report.execution_result.is_ok()
+                    && report.proving_result.as_ref().is_none_or(|r| r.is_ok())
```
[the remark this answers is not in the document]
JereSalo: nope, we actually want that

JereSalo:
> I think running `ethrex-replay blocks --cached --network <NETWORK>` should run all the cached blocks from `NETWORK` from the cache directory.

Good suggestion. [wdyt](https://github.com/lambdaclass/ethrex/pull/4716/commits/2e81a5409341e006284f44413d53ee95dd2c93dc)?

tomip01 (approved):

PR #4722, opened 2025-10-01
ci(l1): add workflow call to replay repo.

JereSalo:
**Motivation**

**Description**

- Call to replay workflow call in order to run a check when pushing to ethrex.

Example run: https://github.com/lambdaclass/ethrex/actions/runs/18507030861/job/52738376976

### DIFF
--- a/.github/scripts/check-hive-results.sh
+++ b/.github/scripts/check-hive-results.sh
@@ -10,6 +10,21 @@ if ! command -v jq >/dev/null 2>&1; then
   exit 1
 fi
 
+if ! command -v python3 >/dev/null 2>&1; then
+  echo "python3 is required to process Hive client logs but was not found in PATH"
+  exit 1
+fi
+
+slugify() {
+  local input="${1:-}"
+  local lowered trimmed
+  lowered="$(printf '%s' "${input}" | tr '[:upper:]' '[:lower:]')"
+  trimmed="$(printf '%s' "${lowered}" | sed -E 's/[^a-z0-9._-]+/-/g')"
+  trimmed="${trimmed#-}"
+  trimmed="${trimmed%-}"
+  printf '%s' "${trimmed}"
+}
+
 results_dir="${1:-src/results}"
 
 if [ ! -d "$results_dir" ]; then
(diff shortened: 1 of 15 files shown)


.github/workflows/pr-main_replay.yaml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  push:
+    paths:
+      - crates/**
+      - ".github/workflows/pr-main_replay.yaml"
+  workflow_dispatch:
+
+jobs:
+  workflow-call:
+    uses: lambdaclass/ethrex-replay/.github/workflows/workflow-call.yaml@add_remote_workflow
```
JereSalo, line 14: change this for main at time of merging

JereSalo:
We probably should add docs of how to fix replay CI when it fails so that the workflow is clear for every pr author

.github/workflows/pr-main_replay.yaml
```diff
@@ -0,0 +1,16 @@
+name: Workflow Call Replay
```
mpaulucci: this is not run from prs, so I would rename to main_replay.yaml

JereSalo: It is actually run from PRs, on every push to a PR so that the author knows replay is out of sync.

JereSalo:
> We want to keep ethrex-replay docs as part of the official docs.

Created [this issue](https://github.com/lambdaclass/ethrex-replay/issues/32) to address having duplicate docs.

.github/workflows/pr-main_replay.yaml
mpaulucci: Ah, I think that is probably an overkill (every push to every branch).

If you want a CI-like behaviour, why don't you specify `on pull_request`? If you have a branch but not a PR, why do you want to run it?

Why don't we start with running this on push to main first and then we see if we want to extend them to be ran on PRs? I want to be thoughtful about all the stuff that is run on a PR

mpaulucci: Add this workflow to https://github.com/lambdaclass/ethrex/blob/bb5f236e80ac718982d57de53060ae825820ecf7/.github/workflows/common_failure_alerts.yaml#L1

nit: maybe I would call the workflow just `Replay`.

JereSalo: We want to run it in PRs in order to avoid merging ethrex PRs to main without having fixed the problems first in replay (if the PR introduced any), that's the purpose of the workflow call.
You are right about the push thing, I should've specified `on pull_request` instead. My bad.

We could start off by just running it on main but it kinda kills the purpose of [the other PR in replay](https://github.com/lambdaclass/ethrex-replay/pull/29)

JereSalo: Good suggestion
https://github.com/lambdaclass/ethrex/pull/4722/commits/ff4807732fc1ad6af4a0d0484503ec6db60f9c78
https://github.com/lambdaclass/ethrex/pull/4722/commits/948bf92f22c54551ad630a60a3920505e305cb01

JereSalo:
Maybe I'm going to close this in favor of #4904 if the additional changes are acceptable (I hope so)

Edit: They weren't haha

JereSalo:
We have to see if we want to continue with this or close it.

JereSalo:
outdated

PR #4723, opened 2025-10-01
chore(l2): optimize trie hashing and `BranchNode` encoding for zkVMs

xqft:
**Motivation**

Our trie does a preorder traversal for recursive hashing, which allocates a lot of buffers (from the encoder and the keccak hasher) all the way down until it starts actually hashing a leaf. A way better approach to implement recursive hashing is to do a postorder traversal, in which we start allocating memory when we get to the lowest node that needs to be hashed, after which the memory is dealloc'd and the next (parent or sibling) node allocs again. This approach was inspired by risc0's trie, although we had a preorder traversal hashing for the `commit()` function already.

This PR also optimizes a `BranchNode` encoding by skipping our `Encoder` type and writing directly to a preallocated buffer, instead of using two buffers and copying data from one to the other (this is because the current `Encoder` can't know the length of the encoded data beforehand, but we can calculate it if we know what we are encoding). This could also be implemented for other node types, but they are the minority of node types and the perf. gains are negligible.

A third, smaller optimization is to prevent cloning the cached/computed hashes from every node.

**Description**
- adds `memoize_hashes` to both `Node` and `NodeRef` to implement postorder traversal
- adds utility functions to calculate encoded lengths of a `NodeHash` and a string of bytes
- changes `BranchNode::encode_raw()` to encode into a single buffer
- adds `compute_hash_ref` to return a reference of the cached hashed

**Testing**
This branch was used in a client snapsynced to Mainnet, running successfully from nov. 4 to nov. 5.

**Flamegraphs** block 23385900 Mainnet
this reduces cycles of `hash_no_commit` (trie hashing) by 40%, which is 5% of total cycles

before:
image

after:
image

**Proving times** RTX 4090
```
10m 07s -> 09m 52s (block 23426995 Mainnet)
16m 42s -> 16m 22s  (block 23426996 Mainnet)
```


crates/common/rlp/encode.rs
```diff
@@ -139,6 +139,24 @@ impl RLPEncode for [u8] {
...
+
+        match self.len() {
+            0 => 1,                                                  // encodes to RLP_NULL
+            1 if self[0] < 0x80 => 1,                                // `self` is its own encoding
+            1..56 => 1 + self.len(),                                 // single byte prefix
+            56..U8_MAX_PLUS_ONE => 1 + 1 + self.len(), // single byte prefix + payload len bytes
+            U8_MAX_PLUS_ONE..U16_MAX_PLUS_ONE => 1 + 2 + self.len(), // single byte prefix + payload len bytes
+            _ => {
+                // fallback if `self` is longer than 2^16 - 1 bytes
+                let payload_len_bytes = (usize::BITS - self.len().leading_zeros()) as usize / 8;
```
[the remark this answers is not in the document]
xqft: correct!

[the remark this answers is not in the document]
xqft: it is handled.

crates/common/trie/node_hash.rs
```diff
@@ -120,6 +120,14 @@ impl RLPEncode for NodeHash {
     fn encode(&self, buf: &mut dyn bytes::BufMut) {
         RLPEncode::encode(&Into::<Vec<u8>>::into(self), buf)
```
Oppen, line 121: Why do we build a vec to encode here?

crates/common/trie/rlp.rs
```diff
@@ -14,18 +15,46 @@ use crate::{Nibbles, NodeHash};
 
 impl RLPEncode for BranchNode {
     fn encode(&self, buf: &mut dyn bytes::BufMut) {
-        let mut encoder = Encoder::new(buf);
+        let value_len = <[u8] as RLPEncode>::length(&self.value);
+        let choices_len = self.choices.iter().fold(0, |acc, child| {
+            acc + RLPEncode::length(child.compute_hash_ref())
+        });
+        let payload_len = choices_len + value_len;
```
Oppen:
```suggestion
        let value_len = <[u8] as RLPEncode>::length(&self.value);
        let payload_len = self.choices.iter().fold(value_len, |acc, child| {
            acc + RLPEncode::length(child.compute_hash_ref())
        });
```
Also, strictly speaking `value_len` is always `1`, because it is impossible for a value to actually end in a branch node. We can ignore it for now.

crates/common/trie/rlp.rs
```diff
@@ -14,18 +15,46 @@ use crate::{Nibbles, NodeHash};
...
+                NodeHash::Hashed(hash) => hash.0.encode(buf),
+                NodeHash::Inline((_, 0)) => buf.put_u8(RLP_NULL),
+                NodeHash::Inline((encoded, len)) => buf.put_slice(&encoded[..*len as usize]),
+            }
+        }
+        <[u8] as RLPEncode>::encode(&self.value, buf);
+    }
+
+    // Duplicated to prealloc the buffer and avoid calculating the payload length twice
+    fn encode_to_vec(&self) -> Vec<u8> {
```
Oppen, line 35: Maybe we should make `encode_to_vec` in the generic implementation call the `length` method instead, wdyt?

Oppen (approved):

crates/common/trie/node_hash.rs
xqft: Hmm not sure, maybe to take advantage of `NodeHash::as_ref()`? might try changing it now that you mentioned it

xqft: I think I'll do it in a different PR

crates/common/trie/rlp.rs
xqft: Yes you are right! though I left it like that to not add an assumption (the trie is also used for transactions and withdrawals I think, not sure if this applies to those cases too)

crates/common/trie/rlp.rs
xqft: The thing with that is that the generic `length` encodes into a buffer (a zero capacity `Vec`) and then returns the length of that buffer. Ideally `RLPEncode` should have a way to hint what the encoded size would be to prealloc the buffer (this is done manually in `BranchNode::encode_to_vec`), and default to a non-prealloced buffer. 

If we call the current `length` we end up encoding twice

PR #4727, opened 2025-10-01
chore(l1, l2): use default options for rocksdb

jrchatruc:
**Motivation**

Our idea is to take rocksdb options from a file instead of hardcoding them. This is the first step in doing that, in order to use a file we need a change to the rust rocksdb repo [for which we've made a PR](https://github.com/rust-rocksdb/rust-rocksdb/pull/1040).

**Description**

Closes #issue_number

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -9,8 +9,8 @@ use ethrex_common::{
 };
 use ethrex_trie::{Nibbles, NodeHash, Trie};
 use rocksdb::{
-    BlockBasedOptions, BoundColumnFamily, Cache, ColumnFamilyDescriptor, MultiThreaded,
-    OptimisticTransactionDB, Options, WriteBatchWithTransaction,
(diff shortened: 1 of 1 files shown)


Oppen:
There's a missing commit that contains the "real" defaults that the RocksDB team doesn't impose due to fear of breaking existing deploys.

rodrigo-o (approved):

rodrigo-o:
> There's a missing commit that contains the "real" defaults that the RocksDB team doesn't impose due to fear of breaking existing deploys.

So this PR still needs some change @Oppen or that's for a follow-up?

ElFantasma (commented):
This PR **needs a full rebase** — the target file `crates/storage/store_db/rocksdb.rs` was moved to `crates/storage/backend/rocksdb.rs` and significantly reworked (it now includes per-column-family `BlockBasedOptions` with bloom filters, block sizes, and blob file support that didn't exist when this PR was opened).

Should we close this one?

MegaRedHand:
Closing due to staleness

PR #4728, opened 2025-10-01
feat(l1): allow overriding RocksDB config

Oppen:
- **default options**
- **fix warnings**
- **feat(l1): allow overriding RocksDB config**

**Motivation**

**Description**

Closes #issue_number


Oppen:
The string syntax is complicated and we're not sure it's even complete. Pablo is working in a more DIY version of external config.

PR #4731, opened 2025-10-01
fix(l1): added error for case where system contract code is empty

SDartayet:
**Description**

This PR adds a check for the case where a system contract code is empty. This is required to pass a few Hive tests recently added that ask to return an error in such a case.

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -383,6 +383,9 @@ pub fn generic_system_contract_levm(
 
     // This check is not necessary in practice, since contract deployment has succesfully happened in all relevant testnets and mainnet
     // However, it's necessary to pass some of the Hive tests related to system contract deployment, which is why we have it
+    // The error that should be returned for the relevant contracts is indicated in the following:
+    // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7002.md#empty-code-failure
+    // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7251.md#empty-code-failure
     if PRAGUE_SYSTEM_CONTRACTS
         .iter()
         .any(|contract| contract.address == contract_address)


crates/vm/backends/levm/mod.rs
```diff
@@ -381,6 +381,15 @@ pub fn generic_system_contract_levm(
         ..Default::default()
     };
 
+    if !(contract_address == HISTORY_STORAGE_ADDRESS.address
```
mpaulucci: would be better to have a vec of system contracts somewhere? Otherwise it will be easy to miss this when a new system contract is added?

crates/vm/levm/src/hooks/default_hook.rs
```diff
@@ -261,9 +263,8 @@ pub fn validate_min_gas_limit(vm: &mut VM<'_>) -> Result<(), VMError> {
         .checked_add(TX_BASE_COST)
         .ok_or(InternalError::Overflow)?;
 
-    let min_gas_limit = max(intrinsic_gas, floor_cost_by_tokens);
-    if vm.current_call_frame.gas_limit < min_gas_limit {
-        return Err(TxValidationError::IntrinsicGasTooLow.into());
+    if vm.current_call_frame.gas_limit < floor_cost_by_tokens {
+        return Err(TxValidationError::IntrinsicGasBelowFloorGasCost.into());
```
mpaulucci, line 267: which ones does this fix? https://hive.ethpandaops.io/#/test/generic/1759727493-afbe365eb26fd5a231ad1aa9e83cd55d

crates/vm/backends/levm/mod.rs
JereSalo: I'd personally add a comment that we have this code in order to pass the tests and that it's impossible in real scenarios. I am not in favor of adding the check but if we want it I'd just add that comment.

SDartayet: We do have a vec of system contracts. However this case is more specific; it's not any given system contract but those that have code (the function in which this check is in is already only reached when a system contract is called). We could build a vec specifically for the system contracts that do have code though (or add a has_code attribute to the existing ones)

SDartayet: Sure, done [here](https://github.com/lambdaclass/ethrex/pull/4731/commits/84ba906122912f38b694029ea2fbc638f901906a). Let me know if you think the comment is adequate

SDartayet: [Done](https://github.com/lambdaclass/ethrex/pull/4731/commits/84ba906122912f38b694029ea2fbc638f901906a)!

JereSalo: It's great, thanks

crates/vm/levm/src/hooks/default_hook.rs
JereSalo, line 1: Is it possible that the changes in this file are somehow unrelated to the PR title? Like the PR is for when the code is empty in system contract calls, this is for other scenario, isn't it?

JereSalo: Also did you add this because you saw a test that made it necessary or what?

SDartayet: Those changes were from a separate PR I made before this one. That one already got merged though, so I'm not sure why those are showing up in the diff for this one

SDartayet: Those are from [this PR](https://github.com/lambdaclass/ethrex/pull/4700)

SDartayet: Ok there, merging the latest version of main into the branch fixed it, the files changed should show up properly now

crates/vm/system_contracts.rs
```diff
@@ -69,3 +69,9 @@ pub fn system_contracts_for_fork(fork: Fork) -> impl Iterator<Item = SystemContr
         .into_iter()
         .filter(move |system_contract| system_contract.active_since_fork <= fork)
 }
+
+pub const SYSTEM_CONTRACTS_WITH_CODE: [SystemContract; 3] = [
```
JereSalo: I don't think this is an accurate name. The contracts that should be here are just 2, the EIP 7002 and EIP 7251 contracts. Idk if we can call them `PRAGUE_SYSTEM_CONTRACTS` or any name representative of that

JereSalo: For example, the beacon root contract is a system contract with code but doesn't belong to the check we are trying to make. Correct me if I'm wrong

SDartayet: Fair, I didn't know the details of the beacon roots contract. As for the rest, I'll fix it in a sec. I thought the deposits contract should be included too but I guess thinking about it now it needn't

SDartayet: [Fixed](https://github.com/lambdaclass/ethrex/pull/4731/commits/67cc166cc4311e8b5c1ac2d5456c4d062b3c1427)

JereSalo:
> Also, make sure to link the corresponding EIP for each contract that says that it should fail if there's no code https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7251.md#empty-code-failure https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7002.md#empty-code-failure

this is still missing

SDartayet:
> Adding links to EIPs

Sorry, my bad, idk how I missed that comment. Fixed it [here](https://github.com/lambdaclass/ethrex/pull/4731/commits/d9e8ef79ec16340311e4669d2b979cc52def3af1)

PR #4736, opened 2025-10-02
fix(blockchain): avoid awaiting under payloads mutex in get_payload

Forostovec:
Refactor get_payload to avoid awaiting while holding the payloads mutex. Previously, the code held the tokio::Mutex across an async join (finishing a payload build task), which could serialize concurrent engine_getPayload calls and cause avoidable stalls. The new flow removes the entry under lock, drops the lock, finishes the task, then re-acquires the lock to reinsert the finished payload at a clamped index. This preserves eviction/order semantics as much as possible while eliminating the await-under-lock hotspot.

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -317,19 +317,39 @@ impl Blockchain {
     /// Attempts to fetch a payload given it's id. If the payload is still being built, it will be finished.
     /// Fails if there is no payload or active payload build task for the given id.
     pub async fn get_payload(&self, payload_id: u64) -> Result<PayloadBuildResult, ChainError> {
-        let mut payloads = self.payloads.lock().await;
-        // Find the given payload and finish the active build process if needed
-        let idx = payloads
-            .iter()
-            .position(|(id, _)| id == &payload_id)
-            .ok_or(ChainError::UnknownPayload)?;
-        let finished_payload = (payload_id, payloads.remove(idx).1.to_payload().await?);
-        payloads.insert(idx, finished_payload);
-        // Return the held payload
-        match &payloads[idx].1 {
-            PayloadOrTask::Payload(payload) => Ok(*payload.clone()),
-            _ => unreachable!("we already converted the payload into a finished version"),
+        // Take the target entry out while holding the lock, then release the lock
+        // and await finishing the payload build outside the critical section.
+        let (idx, item) = {
+            let mut payloads = self.payloads.lock().await;
+            let idx = payloads
+                .iter()
+                .position(|(id, _)| id == &payload_id)
+                .ok_or(ChainError::UnknownPayload)?;
+            let (_, item) = payloads.remove(idx);
+            (idx, item)
+        };
+
+        // Finish the current build task outside of the mutex lock if needed
+        let finished: PayloadBuildResult = match item {
+            PayloadOrTask::Payload(p) => *p,
+            PayloadOrTask::Task(task) => task.finish().await?,
+        };
+
+        // Reinsert the finished payload; if intervening modifications changed the length,
+        // clamp the insert position to the current bounds
+        {
+            let mut payloads = self.payloads.lock().await;
+            let insert_idx = idx.min(payloads.len());
+            payloads.insert(
+                insert_idx,
+                (
+                    payload_id,
+                    PayloadOrTask::Payload(Box::new(finished.clone())),
+                ),
+            );
         }
+
+        Ok(finished)
     }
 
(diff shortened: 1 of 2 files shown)


crates/blockchain/payload.rs
```diff
@@ -317,19 +317,39 @@ impl Blockchain {
...
-            _ => unreachable!("we already converted the payload into a finished version"),
+        // Take the target entry out while holding the lock, then release the lock
+        // and await finishing the payload build outside the critical section.
+        let (idx, item) = {
+            let mut payloads = self.payloads.lock().await;
+            let idx = payloads
+                .iter()
+                .position(|(id, _)| id == &payload_id)
+                .ok_or(ChainError::UnknownPayload)?;
+            let (_, item) = payloads.remove(idx);
```
MegaRedHand: Removing the payload would make the following requests find nothing here. We should change the underlying value type in payloads to something that can be safely shared with the payloads map.

MegaRedHand (changes requested):

crates/blockchain/payload.rs
Forostovec: > Removing the payload would make the following requests find nothing here. We should change the underlying value type in payloads to something that can be safely shared with the payloads map.

done boss

MegaRedHand (approved):

crates/blockchain/payload.rs
```diff
@@ -317,18 +324,44 @@ impl Blockchain {
...
+                PayloadOrTask::Payload(p) => {
+                    // Already finished; restore and return
+                    let result = *p.clone();
+                    *entry = PayloadOrTask::Payload(Box::new(result.clone()));
+                    return Ok(result);
+                }
+                PayloadOrTask::Task(task) => {
+                    // Drop lock and finish task
+                    drop(entry);
+                    let res = task.finish().await?;
```
ElFantasma, line 352: Bug: if `task.finish()` returns `Err`, the `?` propagates immediately, but the entry was already swapped to `Pending` via `mem::replace` on line 342 (and the entry lock was dropped on line 351). The entry is now permanently stuck as `Pending` — any subsequent `get_payload` call for this `payload_id` will spin-loop forever in the `Pending` arm with no way to recover.

Suggested fix: catch the error and restore (or remove) the entry before returning:
```rust
PayloadOrTask::Task(task) => {
    drop(entry);
    match task.finish().await {
        Ok(res) => {
            let mut entry = entry_arc.lock().await;
            *entry = PayloadOrTask::Payload(Box::new(res.clone()));
            return Ok(res);
        }
        Err(e) => {
            // Remove the stuck entry from the payloads table
            let mut payloads = self.payloads.lock().await;
            payloads.retain(|(id, _)| *id != payload_id);
            return Err(e);
        }
    }
}
```

crates/blockchain/payload.rs
```diff
@@ -317,18 +324,44 @@ impl Blockchain {
...
+                    .ok_or(ChainError::UnknownPayload)?;
+                entry.clone()
+            };
+
+            // Work on the entry without removing it from the vec
+            let mut entry = entry_arc.lock().await;
+            match std::mem::replace(&mut *entry, PayloadOrTask::Pending) {
+                PayloadOrTask::Payload(p) => {
+                    // Already finished; restore and return
+                    let result = *p.clone();
```
ElFantasma, line 345: nit: `p` is already owned (taken out by `mem::replace`), so `*p.clone()` clones the inner `PayloadBuildResult` unnecessarily. Then `result.clone()` clones it again to put it back. You can avoid one allocation:
```rust
let result = *p; // move out of Box, no clone needed
*entry = PayloadOrTask::Payload(Box::new(result.clone()));
return Ok(result);
```

crates/blockchain/payload.rs
```diff
@@ -317,18 +324,44 @@ impl Blockchain {
...
+                    let res = task.finish().await?;
+                    let mut entry = entry_arc.lock().await;
+                    *entry = PayloadOrTask::Payload(Box::new(res.clone()));
+                    return Ok(res);
+                }
+                PayloadOrTask::Pending => {
+                    // Someone else is finalizing; wait and retry
+                    *entry = PayloadOrTask::Pending;
+                    drop(entry);
+                    sleep(TokioDuration::from_millis(2)).await;
```
ElFantasma, line 361: The spin-loop has no timeout or backoff. Combined with the error-path bug on line 352 (where a failed `task.finish()` leaves the entry stuck as `Pending`), this will spin forever. Even after fixing the error path, consider adding a maximum retry count or an exponential backoff to be defensive.

ElFantasma (changes requested):

PR #4740, opened 2025-10-02
chore(l1): remove libmdbx

fedacking:
**Motivation**

Libmdbx is slow and unused after rocksdb was set up by default. We should removed the unused database.

**Description**

- Removed libmdbx


cmd/ethrex/initializers.rs
```diff
@@ -40,12 +40,9 @@ use tracing_subscriber::{
 };
 
 // Compile-time check to ensure that at least one of the database features is enabled.
-#[cfg(any(
-    not(any(feature = "rocksdb", feature = "libmdbx")),
-    all(feature = "rocksdb", feature = "libmdbx")
-))]
+#[cfg(not(feature = "rocksdb"))]
```
MegaRedHand, line 43: Should we remove "rocksdb" (the feature) too?

MegaRedHand (approved):

cmd/ethrex/initializers.rs
Oppen: We're currently refactoring the interface to allow easily testing other DBs, so I think I'd rather leave it for now.

ilitteri (approved):

PR #4599, opened 2025-09-22
fix(l1): use transactional database with RocksDB

Oppen:
RocksDB is non-transactional by default, meaning writes logically
interleave rather than behaving as a unit, which is prone to errors.
The optimistic transaction mode implements ACID behavior, with conflict
resolution at commit time to improve parallelism.

PR #4743, opened 2025-10-02
chore(l1): chunk storage request during snapsync

rodrigo-o:
**Motivation**

After a couple of optimization (#4689) we detected a surge of memory consumption during storage requests

**Description**

What we saw were ~25% of the memory allocated for the request storage ranges
image

This PR chunks the storage request and removes the large memory allocations


rodrigo-o:
Here are 2 runs compared, between main (before the revert of #4689 ) and this branch

_Note: both test had reverted #4599 that was affecting the baseline memory consumption_

_Node Memory (RSS)_
| Main | This PR |
| ----- | ----- |
| Mean: 9.2GB & Max: 20GB | Mean: 9GB & Max: 16GB |
| image | image |

Also, now we don't have `storage_request_ranges` as part of the flamegraph at the peak points during the storage request

| Main | This PR |
| ----- | ----- |
| image | image |

rodrigo-o:
Closed in favor of #4821

PR #4751, opened 2025-10-02
feat(l1): add tx broadcasting time interval flag to cli

### LINKED ISSUE #3914: L1: Fix `Request Blob Pooled Transactions Multiple` hive test
This test was apparently a false positive before, which relied in a particular step to receive just 1 tx from `NewPooledTransactionHashes` and given that we were alway sending just 1 by 1 it worked, but after [this PR](https://github.com/lambdaclass/ethrex/pull/3912), we started to do the batching of transactions and it started to fail with:

`FAIL: Error executing step 4: expected 1 hashes, got 5`

This is clearly an issue with the filter logic before sending transactions and should be fixed ASAP, but we decided to comment out it in that PR fix given that it was a false positive before.

cdiielsi:
**Motivation**

We've detected that the Request Blob Pooled Transactions Multiple hive test is failing with `FAIL: Error executing step 4: expected 1 hashes, got 5` because there's a 1 second interval between broadcasting transactions that batch several (in this case 5) transactions before sending them, instead of broadcasting a transaction as soon as it arrives. To pass the test but still be able to batch transactions (since it translates in a significative improvement in performance) I'm introducing a new flag in the cli to set the interval as needed.

**Description**

This pr introduces a new flag in the cli to control the size of the broadcasting time interval for broadcasting transactions.

This pr depends on this other pr in hive to pass the ci.

Closes #3914


cmd/ethrex/cli.rs
```diff
@@ -246,6 +254,7 @@ impl Default for Options {
             dev: Default::default(),
             force: false,
             mempool_max_size: Default::default(),
+            tx_broadcasting_time_interval: Default::default(),
```
mpaulucci, line 258: isnt this 0? Do we actually want the default?

cmd/ethrex/cli.rs
```diff
@@ -175,6 +175,14 @@ pub struct Options {
         help_heading = "P2P options"
     )]
     pub discovery_port: String,
+    #[arg(
+        long = "p2p.tx-broadcasting-interval",
+        default_value = "1000",
```
mpaulucci: Can you put this as a constant instead of hardcoding it here?

cmd/ethrex/cli.rs
cdiielsi: I set the default value with the one that was used previously [here](https://github.com/lambdaclass/ethrex/pull/4751/files#diff-ff9c3e9b7d1e792454f5301f51bd83d186a25785acfed4bf60d3c7d46d948bffL38) and then with this [pr](https://github.com/lambdaclass/hive/pull/42) when hive launches ethrex I set the interval value to what I tested works.

cmd/ethrex/cli.rs
```diff
@@ -24,6 +24,9 @@ use crate::{
 };
 use ethrex_config::networks::Network;
 
+// Amount of seconds between each broadcast
+const BROADCAST_INTERVAL_MS: u64 = 1000; // 1 second
```
mpaulucci: why did you move it here?

cdiielsi: I must have misunderstood, I thought this was what you ment [here](https://github.com/lambdaclass/ethrex/pull/4751#discussion_r2407209823). How should I put this as a constant then?

mpaulucci: can't you import the constant from the place it was originally?

cdiielsi: Oh! Yes! Done [here](https://github.com/lambdaclass/ethrex/pull/4751/commits/99052261aca00dd90ee14ca9b95c3933f7c0d2ca)!

rodrigo-o:
Just for future reference. Although the hive test expects just 1 of those txs that's because of how geth and other clients work, not a spec requirement:

NewPooledTransactionHashes(0x08):
>The recommended soft limit for this message is 4096 items.

But it highlights something to have into acount, which is the added delay we have for the batching of txs. I'll create an issue and reference it here also.

TLDR: This hive test just checks how clients are supposed to work, but failing with 5 instead of 1 hash is not a violation of the spec, just the syntom of batching, which in itself could create other issues.

Issue #4668, opened 2025-09-26
Remove/justify clone() across the codebase


lferrigno:
# Campaign: Remove/justify `clone()` across the codebase

We want to reduce unnecessary clones in hot paths and make intentional clones explicit.  

## Goals
- Every `clone()` must be reviewed.
- If a clone is **removed** → commit the change and check it off here.
- If a clone is **kept** → add a comment **above the line** with  
  `// ok-clone: <reason>`.

---

## Sub-issues
- [ ] VM & execution (`crates/vm/**`)
- [ ] Sequencer / L2 core (`crates/l2/**`)
- [ ] Networking P2P (`crates/networking/p2p/**`)
- [ ] Networking RPC (`crates/networking/rpc/**`)
- [ ] Blockchain core (`crates/blockchain/**`)
- [ ] Storage & trie (`crates/storage/**`, `crates/common/trie/**`)
- [ ] Common types/serde (`crates/common/types/**`, `crates/common/serde_utils.rs`)
- [ ] CLI/tooling/docs (`cmd/**`, `tooling/**`, `docs/**`)

---

## Process
1. Locate all `clone()` calls.
2. Classify:
   - **Unnecessary** → refactor to avoid copying (`&T`, `Arc<T>`, `Cow`, `copied()`, etc).
   - **Required** → keep and mark with `// ok-clone: <reason>`.
3. Submit small PRs per subsystem.

---

## Examples

```rust
// ok-clone: needed to increment Arc refcount
let x = arc_value.clone();

// ok-clone: async move requires owned value
tokio::spawn(do_work(config.clone()));

// ok-clone: serde requires owned String
let json = serde_json::to_string(&data.clone())?;

PR #4752, opened 2025-10-02
chore(l1): remove blockchain clones

SDartayet:
**Motivation**

Keeping track of removing clones in our codebase, and removing them where possible.

**Description**

This PR goes over all the clones in the blockchain crate, removing them where possible and justifying the necessary ones. Additionally, for the ones that could potentially be removed but doing so involves some larger refactors, a `todo-clone` comment is added

Part of #4668


JereSalo (commented):
I reviewed it and everything seems to be okay.
For me some of the comments feel kind of unnecessary and the removed clones don't seem to be important either, I think we should've approached this differently, first measuring impact and then acting upon that. But if another reviewer thinks we should merge it I'm okay with it

PR #4763, opened 2025-10-03
chore(l2): remove cloning from Trie operations

xqft:
**Motivation**

**Description**

Closes #issue_number


ilitteri:
Sharing perf results on this PR

## Performance Results
### Execution Times
Average speedup: **1.138x**  
- Main branch: Min 15.591s (Block 23626696), Max 64.000s (Block 23626693), Avg 39.408s  
- Optimized: Min 13.832s (Block 23626696), Max 57.168s (Block 23626693), Avg 34.716s  
- Speedup range: 1.112x–1.171x  

| Block    | Main (s) | Optimized (s) | Speedup |
|----------|----------|---------------|---------|
| 23626690 | 56.171  | 47.969       | 1.171x |
| 23626691 | 40.272  | 34.772       | 1.158x |
| 23626692 | 20.098  | 17.716       | 1.134x |
| 23626693 | 64.000  | 57.168       | 1.120x |
| 23626694 | 29.374  | 25.339       | 1.159x |
| 23626695 | 53.333  | 47.961       | 1.112x |
| 23626696 | 15.591  | 13.832       | 1.127x |
| 23626697 | 55.857  | 50.101       | 1.115x |
| 23626698 | 19.020  | 16.461       | 1.155x |
| 23626699 | 20.094  | 17.733       | 1.133x |
| 23626700 | 59.677  | 52.827       | 1.130x |

### Proving Times
Average speedup: **1.171x**  
- Main branch: Min 392.000s (Block 23626696), Max 1750.000s (Block 23626693), Avg 1036.364s  
- Optimized: Min 327.000s (Block 23626696), Max 1514.000s (Block 23626693), Avg 891.364s  
- Speedup range: 1.124x–1.201x  

| Block    | Main (s) | Optimized (s) | Speedup |
|----------|----------|---------------|---------|
| 23626690 | 1074.000| 917.000      | 1.171x |
| 23626691 | 1127.000| 954.000      | 1.181x |
| 23626692 | 639.000 | 550.000      | 1.162x |
| 23626693 | 1750.000| 1514.000     | 1.156x |
| 23626694 | 762.000 | 640.000      | 1.191x |
| 23626695 | 1393.000| 1210.000     | 1.151x |
| 23626696 | 392.000 | 327.000      | 1.199x |
| 23626697 | 1653.000| 1471.000     | 1.124x |
| 23626698 | 556.000 | 472.000      | 1.178x |
| 23626699 | 496.000 | 413.000      | 1.201x |
| 23626700 | 1558.000| 1337.000     | 1.165x |

crates/common/trie/nibbles.rs
```diff
@@ -12,7 +12,7 @@ use ethrex_rlp::{
 /// Struct representing a list of nibbles (half-bytes)
 #[derive(Debug, Clone, Default)]
 pub struct Nibbles {
-    data: Vec<u8>,
+    pub data: Vec<u8>,
```
MegaRedHand: Let's keep this private


```suggestion
    data: Vec<u8>,
```

xqft: done

crates/common/trie/trie.rs
```diff
@@ -485,7 +497,11 @@ impl Trie {
         if self.hash_no_commit() == *EMPTY_TRIE_HASH {
             return Ok(None);
         }
-        self.root.get_node(self.db.as_ref(), Nibbles::default())
+        let node = self
+            .root
+            .get_node(self.db.as_ref(), Nibbles::default())?
+            .unwrap();
```
MegaRedHand: Is this `unwrap` intended?

crates/common/trie/trie.rs
```diff
@@ -485,7 +497,11 @@ impl Trie {
...
+        Ok(Some((*node).clone()))
```
MegaRedHand: We can use the `into_inner` method from `Arc` here

crates/common/trie/trie.rs
```diff
@@ -404,14 +416,14 @@ impl Trie {
         fn get_node_inner(
             db: &dyn TrieDB,
             current_path: Nibbles,
-            node: Node,
+            node: Arc<Node>,
```
MegaRedHand: This can be moved to be a reference, I think.

crates/common/trie/trie.rs
```diff
@@ -513,20 +529,18 @@ impl ProofTrie {
...
                 .ok_or_else(|| {
-                    TrieError::InconsistentTree(Box::new(InconsistentTreeError::RootNotFound(
-                        self.0.root.compute_hash().finalize(),
-                    )))
+                    TrieError::InconsistentTree(Box::new(InconsistentTreeError::RootNotFoundNoHash))
                 })?
-                .insert(self.0.db.as_ref(), partial_path, external_ref)?
-                .into()
+                .insert(self.0.db.as_ref(), partial_path, external_ref)?;
+            self.0.root.clear_hash();
```
MegaRedHand, line 538: This could lead to inconsistencies if we forget to call this method.

crates/common/trie/node.rs
```diff
@@ -168,31 +204,46 @@ impl Node {
...
             Node::Branch(n) => n.remove(db, path),
             Node::Extension(n) => n.remove(db, path),
             Node::Leaf(n) => n.remove(path),
+        }?;
+        match new_root {
+            Some(NodeRemoveResult::New(new_root)) => {
+                *self = new_root;
+                Ok((false, value))
+            }
+            Some(NodeRemoveResult::Mutated) => Ok((false, value)),
```
MegaRedHand: Wouldn't this need to change the root too?

crates/common/trie/node.rs
```diff
@@ -23,23 +23,53 @@ pub enum NodeRef {
...
         }
     }
 
+    pub fn get_node_mut(
+        &mut self,
+        db: &dyn TrieDB,
+        path: Nibbles,
+    ) -> Result<Option<&mut Node>, TrieError> {
+        match self {
+            NodeRef::Node(node, _) => Ok(Some(Arc::make_mut(node))),
```
MegaRedHand, line 56: Can this lead to errors? `Arc::make_mut` clones the value if there's more than one `Arc`:

> If there are other Arc pointers to the same allocation, then make_mut will [clone](https://doc.rust-lang.org/std/clone/trait.Clone.html#tymethod.clone) the inner value to a new allocation to ensure unique ownership. This is also referred to as clone-on-write.

xqft: This is taken into account in the PR description. In the case of multiple references, the behaviour would just be the same as before this PR (cloning by default), no errors. This is never the case for stateless execution, not sure about L1.

crates/common/trie/trie.rs
xqft: `into_inner` returns `None` if there's not a single strong reference to the node. This function is only used in `generate_witness_for_blocks_with_fee_configs` so it doesn't matter if a clone happens there.

Maybe it would be best to remove this function if it's only used in one place, to avoid future uses that clone the node?

crates/common/trie/trie.rs
xqft: That's true. I think we could move all the `clear_hash` calls into just two: in `Node::insert` and `Node::remove`. That way we don't forget to do it when editing any node.

crates/common/trie/trie.rs
xqft: Ended up deleting this function, see my comment below

crates/common/trie/trie.rs
xqft: Finally I deleted the function, turns out `generate_witness_for_blocks_with_fee_configs` didn't even need a `Node` just a reference to it.

crates/common/trie/trie.rs
xqft: Done! thanks

crates/common/trie/node.rs
xqft: If this matches `Mutated` it means the removal modified the root instead of yielding a new node kind, so there's nothing more to do.

I refactored this snippet a bit to make it clearer

xqft:
@MegaRedHand Me neither, I think the cloning and replacing pattern is clearer, but it's too expensive for execution in a zkVM.

About the `clear_hash`, in a comment's response I proposed moving those calls to just `Node::insert` and `Node::remove` which I think can work, that simplifies things a bit.

About testing, we ran a snapsyncing L1 from 24 oct. to 27 oct. successfully using this branch

crates/common/trie/trie.rs
xqft: Sike, reverted deletion, fixed this comment

crates/common/trie/trie.rs
xqft: Finally finally I didn't delete it, just changed it's signature to return an Arc

MegaRedHand: We can make it return a reference instead. I think returning `Arc` in the API spills unnecessary internal implementation details.

crates/common/trie/trie.rs
MegaRedHand: I think that makes sense

crates/common/trie/node.rs
```diff
@@ -83,6 +113,12 @@ impl NodeRef {
             NodeRef::Hash(hash) => *hash,
         }
     }
+
+    pub fn clear_hash(&mut self) {
```
MegaRedHand: We should document what this does and why it's needed. If we can change this to be private, even better.

crates/common/trie/node.rs
```diff
@@ -23,23 +23,53 @@ pub enum NodeRef {
...
                 .and_then(|rlp| match Node::decode(&rlp) {
-                    Ok(node) => (node.compute_hash() == hash).then_some(Ok(node)),
+                    Ok(node) => (node.compute_hash() == *hash).then_some(Ok(Arc::new(node))),
                     Err(err) => Some(Err(TrieError::RLPDecode(err))),
                 })
                 .transpose(),
         }
     }
 
+    pub fn get_node_mut(
```
MegaRedHand: Can we change this to be private? I see some uses of this could be replaced with a `get_node_ref` function that returns a non-mutable reference to the node.

Apart from that, we should add a warning to this, mentioning that modifying the node can lead to internal inconsistencies if references to the node aren't updated, and if the node's memoized hash isn't cleared.

crates/common/trie/trie_iter.rs
```diff
@@ -138,9 +139,9 @@ impl Iterator for TrieIterator {
             return None;
         };
         // Fetch the last node in the stack
-        let (mut path, next_node_ref) = self.stack.pop()?;
+        let (mut path, mut next_node_ref) = self.stack.pop()?;
         let next_node = next_node_ref
-            .get_node(self.db.as_ref(), path.clone())
+            .get_node_mut(self.db.as_ref(), path.clone())
```
MegaRedHand: This, for example, could be made into a `get_node_ref` call, since it doesn't need the mutability.

crates/common/trie/trie.rs
```diff
@@ -470,18 +482,19 @@ impl Trie {
...
+                &root_node,
                 partial_path,
             )
         } else {
             Ok(Vec::new())
         }
     }
 
-    pub fn root_node(&self) -> Result<Option<Node>, TrieError> {
+    pub fn root_node(&self) -> Result<Option<Arc<Node>>, TrieError> {
```
MegaRedHand, line 497: Can we change this to be `&Node` instead of `Arc`?

crates/common/trie/trie.rs
```diff
@@ -179,7 +191,7 @@ impl Trie {
         }
     }
 
-    pub fn get_root_node(&self, path: Nibbles) -> Result<Node, TrieError> {
+    pub fn get_root_node(&self, path: Nibbles) -> Result<Arc<Node>, TrieError> {
```
MegaRedHand, line 194: Same for this

crates/common/trie/node.rs
xqft: done! reduced visibility to pub(crate)

crates/common/trie/node.rs
xqft: reduced visibility to pub(crate) and added docs with a caution section

crates/common/trie/trie_iter.rs
xqft: done!

crates/common/trie/trie.rs
xqft: I don't think so, because the `Arc` lives until the scope of the function ends, so the reference does not live enough

crates/common/trie/trie.rs
xqft: Same as above

crates/common/trie/trie.rs
xqft: Sorry I just noticed that `insert` and `remove` do not contain info about which children needs its hash to be cleared, this info is only available to each node variant's `insert` and `remove`, and I don't think it's a good idea to pass that info up to the caller. Clearing all children would be a perf. hit

crates/common/trie/trie.rs
xqft: Addressed in a different comment below

PR #4766, opened 2025-10-03
fix(l1): add `port_publisher` section to kurtosis config file

fmoletta:
**Motivation**
Currenlt, `make localnet` fails locally with the following error: 
```
  Caused by: An error occurred getting public port spec for private port 'udp-discovery' with spec '9000/UDP' on container '/cl-1-lighthouse-geth--07e5a2e474bf40a5811c2647ec4c9acc'
  Caused by: No host machine port binding was specified for Docker port '9000/udp' which corresponds to port spec with num '9000' and protocol 'UDP'
```
This got fixed by exporting the necessary ports for the consensus and execution clients via `port_publisher`

**Description**
* Add `port_publisher` section to kurtosis config file (default.yaml)

Closes #issue_number

### DIFF
--- a/fixtures/networks/default.yaml
+++ b/fixtures/networks/default.yaml
@@ -28,6 +28,14 @@ participants:
     validator_count: 32
     # snooper_enabled: true
 
+port_publisher:
+  cl:
+    enabled: true
+    public_port_start: 33000
+  el:
+    enabled: true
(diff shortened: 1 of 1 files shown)


fmoletta:
Not needed when using OrbStack on Mac

PR #4769, opened 2025-10-03
chore(l1): remove/justify clones from common

SDartayet:
**Motivation**

Cleaning up our code and improving it's performance. This PR is part of a broader campaign to remove clones from our code.

**Description**

This PR removes most of the .clone() calls from the types and utils of the common crate. Those clones that couldn't be removed are justified with an `// ok-clone` comment

Part of #4668


crates/common/types/fork_id.rs
```diff
@@ -126,7 +126,7 @@ fn get_all_fork_id_combinations(forks: Vec<u64>, genesis_hash: BlockHash) -> Vec
             continue;
         }
         combinations.push((
-            H32::from_slice(&hasher.clone().finalize().to_be_bytes()),
+            H32::from_slice(&hasher.clone().finalize().to_be_bytes()), // ok-clone: a new hasher would need to bbe initialized for each combination anyways
```
fmoletta:
```suggestion
            H32::from_slice(&hasher.clone().finalize().to_be_bytes()), // ok-clone: a new hasher would need to be initialized for each combination anyways
```
Small typo

crates/networking/p2p/rlpx/eth/eth68/receipts.rs
```diff
@@ -26,22 +26,22 @@ impl Receipts68 {
...
+            .map(|receipts| receipts.into_iter().map(ReceiptWithBloom::from).collect())
             .collect();
         Self {
             id,
             receipts: transformed_receipts,
         }
     }
 
-    pub fn get_receipts(&self) -> Vec<Vec<Receipt>> {
+    pub fn get_receipts(&mut self) -> Vec<Vec<Receipt>> {
```
fmoletta: Maybe we should change the name to `take_receipts`? as to reflect the new behaviour

fmoletta: I would also suggest changing it into an `into_receipts` method that consumes the `Receipts` message as we don't seem to use the struct afterwards

fmoletta: Looking at this I also noticed that there is an overlooked clone at file `peer_handler.rs`  just  a few lines below where this method is called (Line 708) that could be removed similarly

crates/vm/levm/src/opcode_handlers/block.rs
```diff
@@ -141,7 +141,10 @@ impl<'a> VM<'a> {
...
             .increase_consumed_gas(gas_cost::BLOBHASH)?;
         let index = self.current_call_frame.stack.pop1()?;
-        let blob_hashes = &self.env.tx_blob_hashes;
+        let Some(blob_hashes) = self.tx.blob_versioned_hashes() else {
+            self.current_call_frame.stack.push_zero()?;
+            return Ok(OpcodeResult::Continue);
+        };
 
         let index = match u256_to_usize(index) {
             Ok(index) if index < blob_hashes.len() => index,
```
fmoletta: Maybe we could use `unwrap_or_default(&vec![])` here and let the logic below handle the case of no blobs as to not repeat the same logic (early zero return) twice

tooling/ef_tests/state/runner/levm_runner.rs
```diff
@@ -186,30 +194,46 @@ pub fn prepare_vm_for_tx<'a>(
...
             to: match test_tx.to {
                 TxKind::Call(to) => to,
                 TxKind::Create => return Err(EFTestRunnerError::EIP7702ShouldNotBeCreateType),
             },
             value: test_tx.value,
             data: test_tx.data.clone(),
-            access_list,
-            authorization_list: list,
+            access_list: access_list,
+            authorization_list: authorization_list.unwrap(),
```
fmoletta: Why was this change needed?

crates/networking/p2p/rlpx/eth/eth68/receipts.rs
SDartayet: Sure. Fixed [here](https://github.com/lambdaclass/ethrex/pull/4769/commits/c12a7ac334c326df73ce17be2eca2e956429298a)!

tooling/ef_tests/state/runner/levm_runner.rs
SDartayet: Ah, that was because I removed the blob hashes field from the environment (since it required cloning the blob hashes from the transaction). The way the runner worked was that for EIP4844 transactions, it encoded them as EIP1599 txs and then injected the blob parameters via the environment. Because this would cause tests to fail now the environment didn't have the blob hashes, I had to add some extra logic to build the test transaction as an EIP4844 one when appropriate

crates/vm/levm/src/opcode_handlers/block.rs
SDartayet: That sadly wouldn't work since the vec![] got dropped (and we need a reference). I did [this](https://github.com/lambdaclass/ethrex/pull/4769/commits/5e2084ebe95c5956a7d9180784d1881be312898a), let me know what you think

fmoletta: Looks good!

tooling/ef_tests/state/runner/levm_runner.rs
fmoletta: Makes sense, I wonder if we can keep the match/ use an if let Some so we don't have to unwrap the authorization_list

crates/common/types/fork_id.rs
```diff
@@ -230,7 +230,7 @@ mod tests {
         for test_case in test_cases {
             let fork_id = ForkId::new(
                 chain_config,
-                genesis_header.clone(),
+                genesis_header.clone(), // ok-clone: value is needed later, clone is in test
```
mpaulucci, line 233: I don't think we have to justify clones in tests

tooling/ef_tests/state/Cargo.lock
JereSalo, line 1: In main I added this to the gitignore so I think if you merge main and remove this from the remote it's not going to be pushed again

crates/common/types/transaction.rs
```diff
@@ -1464,26 +1464,10 @@ mod canonic_encoding {
...
-                P2PTransaction::EIP7702Transaction(t) => {
-                    Transaction::EIP7702Transaction(t.clone()).compute_hash()
-                }
-                P2PTransaction::PrivilegedL2Transaction(t) => {
-                    Transaction::PrivilegedL2Transaction(t.clone()).compute_hash()
-                }
+            if let P2PTransaction::PrivilegedL2Transaction(tx) = self {
+                return tx.get_privileged_hash().unwrap_or_default();
             }
+            crate::utils::keccak(self.encode_canonical_to_vec())
```
JereSalo, line 1470: Not related to this PR but we really need to do something about the `P2PTransaction` and the `Transaction` types. Both have the same `encode_canonical` implementation. Maybe we should file an issue for that if it's appropriate

crates/networking/rpc/engine/payload.rs
```diff
@@ -612,17 +612,26 @@ async fn handle_new_payload_v3(
...
+            match expected_blobs_iter.next() {
+                Some(expected_blob) if blob == expected_blob => continue,
+                Some(_) | None => {
+                    return Ok(PayloadStatus::invalid_with_err(
+                        "Invalid blob_versioned_hashes",
+                    ));
+                }
+            }
+        }
+    }
```
JereSalo, line 632: Previous implementation was more clear IMO

SDartayet: Do you think there's a way to rewrite it that doesn't require cloning? I had to rewrite it since tx.blob_versioned_hashes() cloned the blobs each time, and removing the clone there was part of the task

PR #4777, opened 2025-10-06
feat(l1): snap sync optimizations

jrchatruc:
**Motivation**

This PR reintroduces the optimizations reverted in #4689, now fixed. There were two problems with it:

- There was a bug in our new (sorted) trie insertion that meant we incorrectly reconstructed certain storage tries.
- There was a bug in our logic to keep downloading storage leaves on a pivot jump, where we would not correctly handle accounts whose storage root went back to a previous value we already had in our trie from a previous pivot. The way we handle this now is completely best effort: we try downloading storage leaves for a while and then just switch to storage healing afterwards. We will improve this when we rewrite snap sync

**Description**

Closes #issue_number


cmd/ethrex/Cargo.toml
```diff
@@ -72,7 +72,7 @@ c-kzg = [
   "ethrex-p2p/c-kzg",
 ]
 metrics = ["ethrex-blockchain/metrics", "ethrex-l2/metrics"]
-rocksdb = ["ethrex-storage/rocksdb"]
+rocksdb = ["ethrex-storage/rocksdb", "ethrex-p2p/rocksdb"]
```
mpaulucci, line 75: Can't we avoid coupling p2p with the specific db implementation?

crates/networking/p2p/sync.rs
```diff
@@ -1508,3 +1460,336 @@ pub async fn validate_bytecodes(store: Store, state_root: H256) -> bool {
...
+
+        info!("Inserting accounts into the state trie");
+
+        let store_clone = store.clone();
+        let current_state_root: Result<H256, SyncError> =
+            tokio::task::spawn_blocking(move || -> Result<H256, SyncError> {
+                let mut trie = store_clone.open_state_trie(computed_state_root)?;
+
+                for (account_hash, account) in account_states_snapshot {
+                    trie.insert(account_hash.0.to_vec(), account.encode_to_vec())?;
```
MegaRedHand, line 1512: This is missing calls to the metrics module. The "inserting accounts" progress stays at 0% even when we're past that stage.

cmd/ethrex/Cargo.toml
ilitteri: I agree. A priori is not that easy to decouple it, and we need these optimizations. I'll file an issue to tackle this.

ilitteri: @fedacking, please file the issue to tackle this.

fedacking: Done in #4807

PR #4781, opened 2025-10-06
feat(l2): implement l1 fee

### LINKED ISSUE #4696: L2: implement `l1_fee`
`l1_fee` is a dynamic fee charged to the transaction sender to cover the costs of posting the transaction’s `stateDiff` to L1.

The `l1_fee` is calculated once per batch as the rolling average of recent batches. The `OnChainProposer` must track this value and include it as a public input in each proof. On every `commitBatch()` call, the contract must calculate the value to be used for the next batch, based on the `blob_base` or `base_fee` (based).

When a user submits a transaction to L2, the `total_l1_fee` is calculated as:
```
l1_fee * state_diff_bytes_used
```

avilagaston9:
**Motivation**

WIP

**Description**

Closes #4696

### DIFF
--- a/crates/vm/levm/src/hooks/l2_hook.rs
+++ b/crates/vm/levm/src/hooks/l2_hook.rs
@@ -1,13 +1,18 @@
 use crate::{
+    call_frame::CallFrameBackup,
     errors::{ContextResult, InternalError, TxValidationError},
     hooks::{DefaultHook, default_hook, hook::Hook},
     opcodes::Opcode,
+    utils::get_account_diffs_in_tx,
     vm::VM,
 };
 
 use ethrex_common::{
     Address, H160, U256,
-    types::fee_config::{FeeConfig, OperatorFeeConfig},
+    types::{
+        account_diff::get_accounts_diff_size,
+        fee_config::{FeeConfig, L1FeeConfig, OperatorFeeConfig},
+    },
 };
 
 pub const COMMON_BRIDGE_L2_ADDRESS: Address = H160([
@@ -17,6 +22,7 @@ pub const COMMON_BRIDGE_L2_ADDRESS: Address = H160([
 
 pub struct L2Hook {
     pub fee_config: FeeConfig,
+    pub pre_execution_backup: CallFrameBackup,
 }
 
 impl Hook for L2Hook {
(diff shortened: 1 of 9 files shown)


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -252,3 +296,73 @@ fn pay_operator_fee(
...
+        return Ok(0);
+    };
+
+    let mut execution_backup = vm.current_call_frame.call_frame_backup.clone();
+    execution_backup.extend(pre_execution_backup);
+    let account_diffs_in_tx = get_account_diffs_in_tx(vm.db, execution_backup)?;
+    let account_diffs_size = get_accounts_diff_size(&account_diffs_in_tx)
+        .map_err(|e| InternalError::Custom(format!("Failed to get account diffs size: {}", e)))?;
+
+    let l1_fee = calculate_l1_fee(fee_config, account_diffs_size)?;
```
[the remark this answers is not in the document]
avilagaston9: if `vm.env.gas_price` is zero, then `L1_fee_config` was properly set to `None` and this function is never called.

crates/l2/sequencer/block_producer.rs
```diff
@@ -261,22 +281,63 @@ impl GenServer for BlockProducer {
...
+                        .inspect_err(|e| error!("Block Producer Error: {e}"));
+                }
+                send_after(
+                    Duration::from_millis(self.block_time_ms),
+                    handle.clone(),
+                    Self::CastMsg::Produce,
+                );
+                CastResponse::NoReply
+            }
+            InMessage::UpdateL1BlobBaseFee => {
```
jrchatruc: I would move this task to a different L2 genserver (maybe watcher?), because the request to get the base fee could take a while (seconds) if network conditions are not ideal or something else happens, which would in the meantime prevent block production from continuing if the block production interval is low enough

avilagaston9: Moved to watcher [6c34abb](https://github.com/lambdaclass/ethrex/pull/4781/commits/6c34abb3ac666d7c001480a03f420d5d8721850c)!

PR #4782, opened 2025-10-06
feat(l1): raise file descriptor limit from within the binary

jrchatruc:
**Motivation**

So far we have been asking people to raise the file descriptor limit on their own before running ethrex. Now we raise the file descriptor limit ourselves when starting up the L1. 

This code was taken entirely from [Parity Technologies](https://github.com/paritytech/fdlimit), with attribution included.

**Description**

Closes #issue_number

### DIFF
--- a/crates/l2/prover/src/guest_program/src/risc0/Cargo.lock
+++ b/crates/l2/prover/src/guest_program/src/risc0/Cargo.lock
@@ -1182,6 +1182,7 @@ dependencies = [
  "hex",
  "kzg-rs",
  "lazy_static",
+ "libc",
  "once_cell",
  "rayon",
  "rkyv",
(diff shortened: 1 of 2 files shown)


ilitteri (approved):

cmd/ethrex/initializers.rs
```diff
@@ -434,6 +435,8 @@ pub async fn init_l1(
         init_metrics(&opts, tracker.clone());
     }
 
+    raise_fd_limit()?;
```
MegaRedHand, line 438: Should we log and ignore any errors instead? I'm thinking this could fail with weird setups.

Oppen: I think it's better to fail quickly here than fail randomly later when we can't open fds.

Oppen (approved):

PR #4785, opened 2025-10-06
chore(l1,l2,levm): bump rust version to 1.90

ManuelBilbao:
**Motivation**

**Description**


MegaRedHand (approved):

crates/networking/p2p/types.rs
```diff
@@ -367,23 +367,6 @@ impl NodeRecord {
         Ok(())
     }
 
-    pub fn set_fork_id(&mut self, fork_id: &ForkId, signer: &SecretKey) -> Result<(), NodeError> {
```
ManuelBilbao, line 370: This function was not used anywhere

Dockerfile
```diff
@@ -58,7 +58,7 @@ RUN cargo build --release $BUILD_FLAGS
 # --- Final Image ---
 # Copy the ethrex binary into a minimalist image to reduce bloat size.
 # This image must have glibc and libssl
-FROM debian:12-slim
+FROM ubuntu:24.04
```
ManuelBilbao, line 61: Rust bump requires glibc 2.39 which isn't available in Debian image

PR #4788, opened 2025-10-06
refactor(l1): throw error on wrong version of eth capaability

### LINKED ISSUE #4760: Avoid panic when eth version is unsupported
Running `ethrex` results in some panic messages like:

```
thread 'tokio-runtime-worker' panicked at crates/networking/p2p/rlpx/eth/eth68/status.rs:51:9:
assertion `left == right` failed: only eth version 68 is supported
  left: 67
 right: 68
```

This is expected, so it should be replaced by an error and handled without messaging the user.

cdiielsi:
**Motivation**

Stop panicking when decoding a message with a wrong version of capabilities `eth68` and `eth69`.

**Description**

This pr has the `decode` method for both capabilities throw error in case the eth version of the message is wrong.
The only instance where this can happen is if the other peer isn't respecting the negotiated capabilities. This happens as seen in issue #4760. [Here](https://github.com/lambdaclass/ethrex/blob/05e0de5aaa40df9ec7088f7a20a47720c56cd350/crates/networking/p2p/rlpx/connection/server.rs#L676-L677) we can see the eth version negotiated can only be either 68 or 69. 

Closes #4760

### DIFF
--- a/crates/networking/p2p/rlpx/eth/eth68/status.rs
+++ b/crates/networking/p2p/rlpx/eth/eth68/status.rs
@@ -48,7 +48,9 @@ impl RLPxMessage for StatusMessage68 {
         let decoder = Decoder::new(&decompressed_data)?;
         let (eth_version, decoder): (u32, _) = decoder.decode_field("protocolVersion")?;
 
-        assert_eq!(eth_version, 68, "only eth version 68 is supported");
+        if eth_version != 68 {
+            return Err(RLPDecodeError::IncompatibleProtocol);
+        }
 
         let (network_id, decoder): (u64, _) = decoder.decode_field("networkId")?;
         let (total_difficulty, decoder): (U256, _) = decoder.decode_field("totalDifficulty")?;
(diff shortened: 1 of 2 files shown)


MegaRedHand (approved):
LGTM

crates/common/rlp/error.rs
```diff
@@ -16,7 +16,7 @@ pub enum RLPDecodeError {
     #[error("InvalidCompression")]
     InvalidCompression(#[from] snap::Error),
     #[error("IncompatibleProtocol")]
-    IncompatibleProtocol,
+    IncompatibleProtocol(String),
```
fmoletta, line 19: The error message is being ignored in the display, to include it change the message to "IncompatibleProtocol: {0}"

cdiielsi: Thanks!! 🙌  Fixed [here](https://github.com/lambdaclass/ethrex/pull/4788/commits/29525771050127f89f4ceb71787d95934be813ed)!!

PR #4791, opened 2025-10-07
perf(levm): remove `Result<OpcodeResult, VMError>` to avoid expensive matches

azteca1998:
**Motivation**

Matching the result is slow on tight loops.

**Description**

Remove the result so that there is no need to match anything.

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -1605,15 +1605,11 @@ async fn estimate_blob_gas(
 
     let gas_with_headroom = (blob_gas * (100 + headroom)) / 100;
 
-    // Check if we have an overflow when we take the headroom into account.
-    let gas_with_headroom_u64: u64 = gas_with_headroom
-        .try_into()
-        .map_err(|_| BlobEstimationError::OverflowError)?;
-    let blob_gas = arbitrary_base_blob_gas_price
-        .checked_add(gas_with_headroom_u64)
+    let blob_gas = U256::from(arbitrary_base_blob_gas_price)
+        .checked_add(gas_with_headroom)
         .ok_or(BlobEstimationError::OverflowError)?;
 
-    Ok(blob_gas.into())
+    Ok(blob_gas)
 }
 
 /// Regenerates state by re-applying blocks from the last known state root.
--- a/crates/networking/rpc/eth/fee_market.rs
+++ b/crates/networking/rpc/eth/fee_market.rs
@@ -1,5 +1,6 @@
 use ethrex_blockchain::payload::calc_gas_limit;
 use ethrex_common::{
+    U256,
     constants::GAS_PER_BLOB,
     types::{
         Block, BlockHeader, ELASTICITY_MULTIPLIER, Fork, ForkBlobSchedule, Transaction,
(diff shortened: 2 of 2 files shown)


crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -12,622 +26,482 @@ use crate::{
...
+        //   - Make sure the memory is expanded.
+        //   - When there is return data, preallocate it because it won't be possible while the next
+        //     call frame is active.
+        vm.current_call_frame.memory.resize(new_memory_size)?;
+
+        // Trace CALL operation.
+        let data = vm.get_calldata(args_offset, args_len)?;
+        vm.tracer.enter(
+            CallType::DELEGATECALL,
+            vm.current_call_frame.msg_sender,
```
ElFantasma: DELEGATECALL tracer parameters changed from old behavior:

- Old: `tracer.enter(DELEGATECALL, callframe.to /* current contract */, code_address /* delegate */, ...)`
- New: `tracer.enter(DELEGATECALL, msg_sender, callframe.to, ...)`

This changes `from` from the current contract to the original message sender, and `to` from the delegate's code address to the current contract. The old code had an explicit comment: *"In this trace the `from` is the current contract, we don't want the `from` to be the EOA that sent the transaction"*. This will produce different `debug_traceTransaction` output for DELEGATECALL frames.

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -12,622 +26,482 @@ use crate::{
...
+        // Resize memory: this is necessary for multiple reasons:
+        //   - Make sure the memory is expanded.
+        //   - When there is return data, preallocate it because it won't be possible while the next
+        //     call frame is active.
+        vm.current_call_frame.memory.resize(new_memory_size)?;
+
+        // Trace CALL operation.
+        let data = vm.get_calldata(args_offset, args_len)?;
+        vm.tracer.enter(
+            CallType::CALLCODE,
```
ElFantasma, line 211: Similar tracer issue: CALLCODE `to` changed from `code_address` (the contract whose code is being executed) to `callframe.to` (the current contract itself). When EIP-7702 delegation is active, these differ — the tracer loses information about which contract's code is actually running.

crates/vm/levm/src/opcodes.rs
```diff
@@ -397,217 +420,210 @@ impl<'a> VM<'a> {
...
+        opcode_table[Opcode::GAS as usize] = OpCodeFn::new::<OpGasHandler>();
+        opcode_table[Opcode::PUSH1 as usize] = OpCodeFn::new::<OpPushHandler<1>>();
+        opcode_table[Opcode::PUSH2 as usize] = OpCodeFn::new::<OpPushHandler<2>>();
+        opcode_table[Opcode::PUSH3 as usize] = OpCodeFn::new::<OpPushHandler<3>>();
+        opcode_table[Opcode::PUSH4 as usize] = OpCodeFn::new::<OpPushHandler<4>>();
+        opcode_table[Opcode::PUSH5 as usize] = OpCodeFn::new::<OpPushHandler<5>>();
+        opcode_table[Opcode::PUSH6 as usize] = OpCodeFn::new::<OpPushHandler<6>>();
+        opcode_table[Opcode::PUSH7 as usize] = OpCodeFn::new::<OpPushHandler<7>>();
+        opcode_table[Opcode::PUSH8 as usize] = OpCodeFn::new::<OpPushHandler<8>>();
+        opcode_table[Opcode::PUSH8 as usize] = OpCodeFn::new::<OpPushHandler<8>>();
```
ElFantasma: nit: PUSH8 is assigned twice (lines 442-443). This is pre-existing from the old code but worth cleaning up while refactoring.

PR #4797, opened 2025-10-07
refactor(l1): multiplex p2p requests

### LINKED ISSUE #3963: Add a table for backend RLPxConnection

ElFantasma:
**Motivation**
Currently our Peer Connections can handle only one request at a time. We want to be able to handle multiple requests and be able to redirect their responses to proper actors.

**Description**
Improve request/response handling on peer connections:

Done:
- Rename structs and functions to improve readability
- Make internal `PeerConnection` structs and enums private exposing only necessary api
- Have an internal request hashmap to direct expected responses to correct channels
- Removed receiver channel `lock()` from `PeerHandler` code and replaced by a oneshot channel
- Removed PeerChannels
- Move timeout and response handling on `PeerConnection` side.
- Remove all `mark_in_use` and `free` logic that won't be required.

To do:
- Implement a load balancer algorithm measure performance and use `PeerConnection` in a balanced fashion

Closes #3963

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ tooling/ef_tests/state/runner_v2/success_report.txt
 
 tooling/reorgs/data
 
+tooling/sync/logs/
+
 # Repos checked out by make target
 /hive/
 ethereum-package/
--- a/crates/l2/sdk/src/sdk.rs
+++ b/crates/l2/sdk/src/sdk.rs
@@ -36,10 +36,10 @@ pub use ethrex_sdk_contract_utils::*;
 
 use calldata::from_hex_string_to_h256_array;
 
-// 0xcb968a1441ffcfb88194912923c5353f8da0210b
+// 0x36664d7c5031bd965bbb405b55495a90dd780740
 pub const DEFAULT_BRIDGE_ADDRESS: Address = H160([
-    0xcb, 0x96, 0x8a, 0x14, 0x41, 0xff, 0xcf, 0xb8, 0x81, 0x94, 0x91, 0x29, 0x23, 0xc5, 0x35, 0x3f,
-    0x8d, 0xa0, 0x21, 0x0b,
+    0x36, 0x66, 0x4d, 0x7c, 0x50, 0x31, 0xbd, 0x96, 0x5b, 0xbb, 0x40, 0x5b, 0x55, 0x49, 0x5a, 0x90,
+    0xdd, 0x78, 0x07, 0x40,
 ]);
 
 // 0x000000000000000000000000000000000000ffff
@@ -75,7 +75,7 @@ pub enum SdkError {
     FailedToParseAddressFromHex,
 }
 
-/// BRIDGE_ADDRESS or 0xcb968a1441ffcfb88194912923c5353f8da0210b
+/// BRIDGE_ADDRESS or 0x36664d7c5031bd965bbb405b55495a90dd780740
 pub fn bridge_address() -> Result<Address, SdkError> {
     std::env::var("ETHREX_WATCHER_BRIDGE_ADDRESS")
         .unwrap_or(format!("{DEFAULT_BRIDGE_ADDRESS:#x}"))
--- a/tooling/sync/Makefile
+++ b/tooling/sync/Makefile
@@ -58,7 +58,7 @@ ifndef SYNC_BLOCK_NUM
 endif	
 #	samply record --unstable-presymbolicate --save-only -- 
 	mkdir -p logs
-	make start-ethrex-$(NETWORK) >> ./logs/ethrex-sync-$(NETWORK)-$(EVM).log
+	make start-ethrex >> ./logs/ethrex-sync-$(NETWORK)-$(EVM).log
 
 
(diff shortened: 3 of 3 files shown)


crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -67,21 +35,121 @@ use crate::{
...
+    }
+
+    pub async fn outgoing_request(
+        &mut self,
+        message: Message,
+        timeout: Duration,
+    ) -> Result<Message, PeerConnectionError> {
+        let id = message
+            .request_id()
+            .expect("Cannot wait on request without id");
```
mpaulucci, line 121: do we really want to panic here?

crates/networking/p2p/rlpx/message.rs
```diff
@@ -282,6 +282,39 @@ impl Message {
...
+            Message::GetTrieNodes(message) => Some(message.id),
+            Message::BlockHeaders(message) => Some(message.id),
+            Message::BlockBodies(message) => Some(message.id),
+            Message::PooledTransactions(message) => Some(message.id),
+            Message::Receipts68(message) => Some(message.id),
+            Message::Receipts69(message) => Some(message.id),
+            Message::AccountRange(message) => Some(message.id),
+            Message::StorageRanges(message) => Some(message.id),
+            Message::ByteCodes(message) => Some(message.id),
+            Message::TrieNodes(message) => Some(message.id),
```
mpaulucci, line 304: nit: cant you use `|` like you do below?

ElFantasma: How's that? `message` on each matching clause has a different type 🤔

crates/networking/p2p/rlpx/connection/server.rs
ElFantasma: Sending a request (to expect a response) using a message that does not have request id is a bug... (the task will wait a response that will never come)
I'd prefer a compile error here, to prevent the dev to introduce the bug, but since we can't with our current types, lets panic

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -173,61 +160,55 @@ impl PeerTableHandle {
...
-    pub async fn set_unwanted(&mut self, node_id: &H256) -> Result<(), PeerTableError> {
-        self.0
-            .cast(CastMessage::SetUnwanted { node_id: *node_id })
+    /// Increment the number of ongoing requests for this peer
+    pub async fn inc_requests(&mut self, node_id: H256) -> Result<(), PeerTableError> {
+        self.handle
+            .cast(CastMessage::IncRequests { node_id })
             .await?;
         Ok(())
     }
```
Oppen, line 175: Shouldn't this be managed by peer itself?

crates/networking/p2p/utils.rs
```diff
@@ -1,20 +1,12 @@
...
-use secp256k1::{PublicKey, SecretKey};
-use spawned_concurrency::error::GenServerError;
-
-use crate::peer_handler::DumpError;
-use crate::{
-    discv4::peer_table::PeerChannels,
-    rlpx::{Message, connection::server::CastMessage, snap::TrieNodes},
-};
+#[cfg(feature = "rocksdb")]
 use tracing::error;
```
Oppen, line 9: Why is this gated with `rocksdb`?

crates/networking/p2p/discv4/peer_table.rs
ElFantasma: The number of requests is used for the math to balance the load between all peers, when choosing one peer to use. If this number was on peer's processes, it would mean that the `get_best_peer` function would have to message every active peer before selecting one.

crates/networking/p2p/utils.rs
ElFantasma: To be honest: I don't know. It came from a [merge from main](https://github.com/lambdaclass/ethrex/pull/4797/commits/799e73db12078b8f7dc09fb29cb41030143497e3) commit, but it is not in the [original](https://github.com/lambdaclass/ethrex/commit/8957ff21c6036ea062e0ea952ed5251726c99281#diff-87312ae0fa6873e688d4cc92f20657040275114fff0e7b7d4c0a62cbaeef57aa) commit 🤷‍♂️

ElFantasma: Fixed, if removed it was warning on:
```
 --> crates/networking/p2p/utils.rs:9:5
  |
9 | use tracing::error;
  |     ^^^^^^^^^^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(unused_imports)]`
  ```
  But lucky for me there was an unguarded use of `tracing::error!`

crates/networking/p2p/rlpx/message.rs
pablodeymo: you are right @ElFantasma

Issue #3884, opened 2025-07-30
Polish post state storage comparisons


sofiazcoaga:
It was decided that it is not a major problem now, as post state only gets compared if expected root was not the one obtained (therefore no tests are currently failing because of this), but some post storage checks could fail for some punctual considerations: 
- The information stored in database in `current_accounts_state` and `initial_accounts_state` fields is ONLY the diff and not the full storage pre and post tx. 
- The `.json` files do not include storage slots that have the value `0`. 

Review this in more depth and consider: 
- Filtering storage slots with content `0` in the real post state storage. 
- Include missing original storage slots (that have not changed) taking them from the Genesis.

PR #4802, opened 2025-10-08
test(l1): improve post state storage comparisons for EF tests

crStiv:
- Filter out zero-value storage slots in actual storage comparison to match .json file behavior
- Include missing original storage slots from Genesis that haven't changed during transaction execution
- Only include non-zero values in AccountUpdate added_storage to maintain consistency with test expectations

These changes address GitHub issue #3884 where post state storage comparisons were failing due to:
1. JSON test files not including storage slots with value 0, but real post state containing them
2. Missing original storage slots from Genesis not being included in comparisons when accounts had empty storage


mpaulucci:
ci is very red, moving to draft

PR #4806, opened 2025-10-08
feat(l1): make the database path-based

iovoid:
**Motivation**

We want a path-based database:
* To reduce database size, because updates nodes replace old ones instead of coexisting
* As preparation for pruning, which reduces database size even further

**Description**

Support for "archive mode" (keeping indefinitely old state) is dropped.

To handle reorgs we use in-memory layers that exist on top of the db.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -126,7 +126,7 @@ Block producer options:
       --block-producer.extra-data <EXTRA_DATA>
           Block extra data message.
 
-          [default: "ethrex 0.1.0"]
+          [default: "ethrex 3.0.0"]
 ```
 
 <!-- END_CLI_HELP -->


cmd/ethrex/initializers.rs
```diff
@@ -385,6 +387,44 @@ async fn set_sync_block(store: &Store) {
     }
 }
 
+/// Rollback the chain to the latest block we have the state for.
```
mpaulucci: shouldn't this be part of the blockchain crate? It would also help us with not having to add a new dependency

crates/common/trie/node/branch.rs
```diff
@@ -100,6 +102,7 @@ impl BranchNode {
             // Insert into self
             self.update(value);
         } else {
+            // Value in branches don't happen in our use-case.
```
mpaulucci, line 105: then below it should be `unreachable!`?

MegaRedHand: It can't happen in our use-case, but a user of `ethrex-trie` can still trigger this. We should first decide if we want to support this use-case: if we don't want to, then we can mark it as `unreachable!` and add a couple of checks to avoid reaching this branch, or we can add an `unimplemented!`, but we should document the lack of support for values in branches. I think leaving it as `todo!` for now would also be OK.

cmd/ethrex/initializers.rs
MegaRedHand: We're working on a fix that includes removing this function. It's likely we'll change this function to re-execute blocks instead of deleting them from the DB, and we won't need the trie dependency.

MegaRedHand: [e88a692](https://github.com/lambdaclass/ethrex/pull/4806/commits/e88a692730fd7d778a841b935225ea3fecfbee98)

PR #4809, opened 2025-10-08
chore(l2): use SP1 patch for ecpairing precompile

xqft:
**Motivation**

Elliptic Curve pairing is one of the most expensive operations in a zkVM .We are using lambdaworks for implementing the ecpairing precompile, which is not optimized for zkVMs. We must use each zkVM pairing patch/precompile instead. 

This is true for other operations as well, like ecmul or ecadd, which can be optimized the same way in future PRs

**Flamegraphs** for block 23385900 Mainnet
ecpairing cycles are reduced from 135k to 6k, which is a 10% improvement in the total cycles.
before:
image
after:
image

### DIFF
--- a/.github/workflows/tag_release.yaml
+++ b/.github/workflows/tag_release.yaml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         platform:
           - ubuntu-22.04
-          - ubuntu-24.04-arm
+          - ubuntu-22.04-arm
           - macos-latest
         stack:
           - l1
@@ -34,7 +34,7 @@ jobs:
             os: linux
             arch: x86_64
             cpu_flags: RUSTFLAGS='-C target-cpu=x86-64-v2'
-          - platform: ubuntu-24.04-arm
+          - platform: ubuntu-22.04-arm
             os: linux
             arch: aarch64
           - platform: macos-latest
@@ -49,10 +49,10 @@ jobs:
           - platform: ubuntu-22.04
             stack: l2_gpu
             features: l2,l2-sql,sp1,risc0,gpu
-          - platform: ubuntu-24.04-arm
+          - platform: ubuntu-22.04-arm
             stack: l2
             features: l2,l2-sql,sp1
-          - platform: ubuntu-24.04-arm
+          - platform: ubuntu-22.04-arm
             stack: l2_gpu
             features: l2,l2-sql,sp1,gpu
           - stack: l2_gpu
@@ -86,7 +86,7 @@ jobs:
           ~/.sp1/bin/sp1up --version 5.0.8
 
       - name: Set up QEMU (only Linux ARM)
-        if: ${{ matrix.platform == 'ubuntu-24.04-arm' }}
+        if: ${{ matrix.platform == 'ubuntu-22.04-arm' }}
         uses: docker/setup-qemu-action@v3
         with:
           platforms: amd64
@@ -111,7 +111,7 @@ jobs:
           sub-packages: '["nvcc"]'
 
       - name: Install solc
-        if: ${{ matrix.platform != 'ubuntu-24.04-arm' }}
+        if: ${{ matrix.platform != 'ubuntu-22.04-arm' }}
         uses: lambdaclass/get-solc@master
         with:
           version: v0.8.29
@@ -123,7 +123,7 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install solc (Linux ARM)
-        if: ${{ matrix.platform == 'ubuntu-24.04-arm' }}
+        if: ${{ matrix.platform == 'ubuntu-22.04-arm' }}
         run: |
           sudo curl -L -o /usr/local/bin/solc https://github.com/nikitastupin/solc/raw/refs/heads/main/linux/aarch64/solc-v0.8.29
           sudo chmod +x /usr/local/bin/solc


crates/vm/levm/src/precompiles.rs
```diff
@@ -765,193 +760,175 @@ const ALT_BN128_PRIME: U256 = U256([
...
-            Ok(Some((first_point, second_point)))
-        }
-    }
+    Ok(())
 }
 
-/// Performs a bilinear pairing on points on the elliptic curve 'alt_bn128', returns 1 on success and 0 on failure
-pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<Bytes, VMError> {
+#[inline]
+fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<Bytes, VMError> {
```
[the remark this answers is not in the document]
xqft: good catch!

crates/vm/levm/src/precompiles.rs
```diff
@@ -765,193 +760,175 @@ const ALT_BN128_PRIME: U256 = U256([
...
 
-    let mut valid_pairs = Vec::new();
-    let mut mul: FieldElement<Degree12ExtensionField> = QuadraticExtensionFieldElement::one();
-
+    let mut batch = Vec::new();
     for input in calldata.chunks_exact(192) {
         #[expect(unsafe_code, reason = "chunks_exact ensures the conversion is valid")]
         let input: [u8; 192] = unsafe { input.try_into().unwrap_unchecked() };
+        let (g1, g2) = parse_bn254_coords(&input);
+        validate_bn254_coords(&g1, &g2).inspect(|_| batch.push((g1, g2)))?;
```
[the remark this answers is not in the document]
xqft: no it shouldn't

crates/vm/levm/src/precompiles.rs
```diff
@@ -765,193 +760,175 @@ const ALT_BN128_PRIME: U256 = U256([
...
-        let (second_point_x, second_point_y) = parse_second_point_coordinates(&input)?;
+        #[cfg(feature = "sp1")]
+        break 'inner pairing_substrate(&batch)?;
 
-        if let Some(pair) =
-            validate_pairing(first_point_x, first_point_y, second_point_x, second_point_y)?
-        {
-            valid_pairs.push(pair);
+        // default
+        break 'inner pairing_lambdaworks(&batch)?;
```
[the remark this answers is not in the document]
xqft: can't use runtime feature checking because the `pairing_substrate` depends on an optional crate, which is a compile time restriction.

it isn't too complex neither, there are already labeled scopes in this same file

crates/vm/levm/src/precompiles.rs
```diff
@@ -765,193 +760,175 @@ const ALT_BN128_PRIME: U256 = U256([
...
+                );
+                (x, y)
+            };
+            let g1 = LambdaworksG1::create_point_from_affine(g1_x, g1_y)
+                .map_err(|_| PrecompileError::InvalidPoint)?;
+            let g2 = LambdaworksG2::create_point_from_affine(g2_x, g2_y)
+                .map_err(|_| PrecompileError::InvalidPoint)?;
+            // if !g2.is_in_subgroup() {
+            //     return Err(PrecompileError::PointNotInSubgroup.into());
+            // }
```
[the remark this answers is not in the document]
xqft: good catch

crates/vm/levm/src/precompiles.rs
```diff
@@ -912,43 +801,158 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
...
+        if validate_bn254_coords(&g1, &g2)? {
+            batch.push((g1, g2));
+        }
+    }
 
-        let (first_point_x, first_point_y) = parse_first_point_coordinates(&input)?;
-
-        let (second_point_x, second_point_y) = parse_second_point_coordinates(&input)?;
+    #[allow(unreachable_code)]
+    let pairing_check = 'inner: {
```
pablodeymo: Using a label is not a recommended practice. 
I suggest to use a construction that is less error prone. For example, something that refuses to compile when using incompatible features.

Oppen: Suggested approach: let `pairing_check` be a function with the appropriate implementation for each config, e.g.:

```rust
#[cfg(feature = "sp1")]
fn pairing_check(batch: &[(G1, G2)]) -> Result<bool, VMError> {
    if batch.is_empty() {
         return Ok(true);
    }
    //  Code currently in `pairing_substrate`
}

#[cfg(not(feature = "sp1")]
fn pairing_check(batch: &[(G1, G2)]) -> Result<bool, VMError> {
    if batch.is_empty() {
         return Ok(true);
    }
    //  Code currently in `pairing_lambdaworks`
}
```

The calling code would then not need to add special cases, it would just be:
```
let pairing_check = pairing_check(&batch)?;
```

crates/vm/levm/src/precompiles.rs
```diff
@@ -912,41 +801,153 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
...
+    Ok(Bytes::from_owner(result))
+}
+
+#[cfg(feature = "sp1")]
+#[inline]
+pub fn pairing_substrate(batch: &[(G1, G2)]) -> Result<bool, VMError> {
+    use substrate_bn::{AffineG1, AffineG2, Fq, Fq2, G1 as SubstrateG1, G2 as SubstrateG2, Group};
+
+    let result = if batch.is_empty() {
+        substrate_bn::Gt::one()
```
pablodeymo: I would prefer to do an early return here to avoid the huge `else`.

crates/vm/levm/src/precompiles.rs
```diff
@@ -912,41 +801,153 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
...
+                    Fq::from_bytes_be(&g1.0.to_big_endian())
+                        .map_err(|_| InternalError::msg("failed to parse g1 x"))?,
+                    Fq::from_bytes_be(&g1.1.to_big_endian())
+                        .map_err(|_| InternalError::msg("failed to parse g1 y"))?,
+                );
+                LambdaworksG1::create_point_from_affine(g1_x, g1_y)
+                    .map_err(|_| PrecompileError::InvalidPoint)?
+            };
+            let g2 = if g2.0.is_zero() && g2.1.is_zero() && g2.2.is_zero() && g2.3.is_zero() {
+                ProjectiveG2::neutral_element()
```
pablodeymo: this for loop body it's better to check the condition of equality with neutral element (`is_neutral_element`) within the top of the loop

xqft: an early `continue;` there is tricky because we would be skipping the creation of one of the points (create_point_from_affinity), which contains implicit validation checks.

e.g. if G1 is zero and G2 is an invalid point, the precompile should revert but instead it would continue.

crates/vm/levm/src/precompiles.rs
xqft: addressed! thank you :)

crates/vm/levm/src/precompiles.rs
xqft: done!

crates/vm/levm/src/precompiles.rs
xqft: we talked internally and decided to leave it as it is, but communicate to the Lambdaworks team the wish to implement the ecpairing and other operations in Lambdaworks to avoid having too much cryptography code in LEVM

ilitteri (approved):

pablodeymo (approved):

PR #4811, opened 2025-10-08
fix(l1): fix tooling/sync sync target so it uses existing target

cdiielsi:
**Motivation**

When running `make sync` in tooling/sync you get the following error: 

```
make start-ethrex-hoodi >> ./logs/ethrex-sync-hoodi-levm.log
make[1]: *** No rule to make target `start-ethrex-hoodi'.  Stop.
make: *** [sync] Error 2
```
Because the specific `start-ethrex-hoodi` target no longer exists since this [pr](https://github.com/lambdaclass/ethrex/commit/b7b7c1bf19e873c93e1791687a7bf038fad490ae).

**Description**

Change `start-ethrex-$(NETWORK)` to `start-ethrex` in `sync` target as it was intended to be used since [pr](https://github.com/lambdaclass/ethrex/commit/b7b7c1bf19e873c93e1791687a7bf038fad490ae)

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -400,6 +400,9 @@ pub async fn init_l1(
 
     let genesis = network.get_genesis()?;
     display_chain_initialization(&genesis);
+
+    raise_fd_limit()?;
+
     let store = init_store(datadir, genesis).await;
 
     #[cfg(feature = "sync-test")]
@@ -450,8 +453,6 @@ pub async fn init_l1(
     if opts.metrics_enabled {
         init_metrics(&opts, tracker.clone());
     }
-
-    raise_fd_limit()?;
 
     if opts.dev {
         #[cfg(feature = "dev")]


tooling/sync/Makefile
```diff
@@ -58,7 +58,7 @@ ifndef SYNC_BLOCK_NUM
 endif	
 #	samply record --unstable-presymbolicate --save-only -- 
 	mkdir -p logs
-	make start-ethrex-$(NETWORK) >> ./logs/ethrex-sync-$(NETWORK)-$(EVM).log
+	make start-ethrex >> ./logs/ethrex-sync-$(NETWORK)-$(EVM).log
```
MegaRedHand, line 61: I don't usually run `make sync`, but we can also define an `ethrex-sync-%` target that does `make ethrex-sync NETWORK=$<` (not sure about specific special var), if that sounds easier.

cdiielsi: This target already works that way! Now you can run the command specifying the network using the environmental variable NETWORK, eg: `make sync NETWORK=hoodi SYNC_BLOCK_NUM=0`, otherwise mainnet is the default.

The `ethrex-sync-%` existed before, but in this this [pr](https://github.com/lambdaclass/ethrex/commit/b7b7c1bf19e873c93e1791687a7bf038fad490ae) changes were made to drop that target amongst others, and as far as I understand the thing was they just forgot tu update the `sync` target correctly.

PR #4812, opened 2025-10-08
feat(l1): download all headers from new to old as first step of full sync

### LINKED ISSUE #4717: Change full-sync to download headers from latest to genesis
Related to #2126

After the merge, clients are supposed to request headers from the latest one back to genesis. This avoids problems with reorgs or long-lived side-chains near the tip.
It also solves a DoS vector: since generating a new valid block no longer requires providing a proof-of-work, byzantine nodes could potentially give us a long chain that doesn't connect to the current latest block our consensus client gave us. By syncing latest-block first, we can easily find this block doesn't connect to our latest.

Full-sync logic currently:

1. Fetches the current head from the DB
2. Send headers request to peers (with retry and timeout), from current head ascending.
    1. If fails, returns
    2. If response is empty, retry.
3. Set current head to head’s parent if we detect a reorg (#2126).
    1. Note that this doesn’t modify the DB, so any network failure will reset the head.
4. Update the current head with the latest downloaded header.
5. `process_incoming_headers`: request block bodies, execute batch, and store in the DB.
6. If sync head found, break, else loop back to 2.

What we want:

1. Try to fetch the sync-head from the DB, and use that as our initial "tail" (the lowest header we have that connects to our sync-head).
2. Download headers from latest to genesis, in a loop:
    1. Check we don’t have the header in the DB first. If we do, update the current tail of the chain to that block's parent and `continue`. (*)
    3. Request headers from peers, descending from the current tail.
    4. Store headers directly in DB, skipping any known headers, and removing any data from side-chains we already stored, like account states or block numbers.
    5. If the chain connects to genesis, `break`.
3. Download and execute bodies from genesis to head. We can do this sequentially for the first version, and later switch to parallel downloads.

(*) Note that we'll need to add an invariant or some kind of metadata to mark a header as "we got all headers from here to genesis". This avoids the common case where we need to fetch a single block, so we fetch it, and then fetch its parent, its parent's parent, and so on, until we get all the way to genesis. For a first version, we could just have it in memory, but it should be easy to persist that information in the DB.

### LINKED ISSUE #4776: L1(FullSync): syncer stuck on side chain during full sync
As we download batches of block headers from peers we might receive a batch from a peer that is not following the canonical chain. As we have no way of knowing whether the blocks are canonic or not, we will proceed to request the block bodies for said batch and fail to do so as the blocks are not canonic. This may lead us to fail the sync and lose our peers as we asume they don't have the blocks we need.

This was encountered of fusaka-devnet-3 testnet with the following logs:

Image

Upon checking the block explorer we can see that the hashes for blocks 462 & 483 do not match the ones on the canonical chain.
The logs following the ones on the screenshot consist of an endless stream of "Looking for Peers"

fmoletta:
**Motivation**
In order to properly handle reorgs, we should download headers from the latest announced fcu head until me meet our local chain instead of requesting headers from our current head till we reach the fcu head, as our current head may not be part of the canonical chain after a reorg.
The main behaviour changes imposed by this are that
A) We must request all headers before we can begin to execute the blocks
B) We are no longer able to keep all headers in memory (in case of long syncs)

**Description**
* Download all headers from sync_head till we reach our local chain as first step of full sync
* Remove `BlockSyncState` & `FullBlockSyncState`
* [BONUS] Invalidate child block when we ask if its parent is invalid (fork choice + new payload)

**Observations & Follow-Up Work**
The current implementation downloads all headers first before processing blocks without any sort of checkpoint system to keep progress in case of unexpected failures or shutdowns. When comparing to other implementations this doesn't seem to be much of an issue as header download phase is pretty quick. Having all headers downloaded as a first step can also allow us to download headers non-sequentially (aka divide the chain into batches and fetch them in parallel). Note that this is not as high priority as it would only affect large full syncs, but it may slow down our efforts when using full sync to test vm behaviour. Potential follow up work:
* Add a way to track which headers we already downloaded if we restart a full sync after a failure (could be something as simple as looking at which header ranges we have on the DB)
* Speed up header download phase by using parallel requests for different block ranges
* Evaluate whether `SnapBlockSyncState` struct can be improved/removed

Closes #4717 + Closes #4776

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -483,7 +483,8 @@ impl Syncer {
         {
             if let Some(batch_failure) = batch_failure {
                 warn!("Failed to add block during FullSync: {err}");
-                // Since running the batch failed we set the failing block and it's descendants with having an invalid ancestor on the following cases.
+                // Since running the batch failed we set the failing block and its descendants
+                // with having an invalid ancestor on the following cases.
                 if let ChainError::InvalidBlock(_) = err {
                     let mut block_hashes_with_invalid_ancestor: Vec<H256> = vec![];
                     if let Some(index) = blocks_hashes


crates/networking/p2p/peer_handler.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
         },
```
mpaulucci, line 19: To be able to test this better:
- remove the limit here: https://github.com/lambdaclass/ethrex/blob/a00c2920361fb33d0a2c693e4808e877a6b0c974/.github/workflows/pr-main_l1.yaml#L180 
- enable cancun tests: https://github.com/lambdaclass/ethrex/blob/a00c2920361fb33d0a2c693e4808e877a6b0c974/.github/workflows/pr-main_l1.yaml#L166-L171
- Set parallelism to 4 here: https://github.com/lambdaclass/ethrex/blob/a00c2920361fb33d0a2c693e4808e877a6b0c974/.github/workflows/pr-main_l1.yaml#L228

The two last steps can be avided if you wait for https://github.com/lambdaclass/ethrex/pull/4876 and sync with main

crates/storage/store_db/rocksdb.rs
```diff
@@ -104,6 +104,11 @@ const CF_PENDING_BLOCKS: &str = "pending_blocks";
 /// - [`Vec<u8>`] = `BlockHashRLP::from(latest_valid).bytes().clone()`
 const CF_INVALID_ANCESTORS: &str = "invalid_ancestors";
 
+/// Block headers downloaded during fullsync column family: [`u8;_`] => [`Vec<u8>`]
+/// - [`u8;_`] = `block_number.to_le_bytes()`
+/// - [`Vec<u8>`] = `BlockHeaderRLP::from(block.header.clone()).bytes().clone()`
+const CF_FULLSYNC_HEADERS: &str = "fullsync_headers";
```
MegaRedHand, line 110: I think this could be merged with the main headers table. I opened #4903 to check that.

crates/networking/p2p/peer_handler.rs
mpaulucci: #4876 has been merged!

crates/networking/p2p/sync.rs
```diff
@@ -338,109 +315,220 @@ impl Syncer {
...
-            );
-            debug!(
-                "Sync Log 4: State current blocks len {}",
-                block_sync_state.current_blocks.len()
-            );
+        // Check if the sync_head is a pending block, if so, gather all pending blocks belonging to its chain
+        let mut pending_blocks = vec![];
+        while let Some(block) = store.get_pending_block(sync_head).await? {
+            if store.is_canonical_sync(block.hash())? {
+                // Ignore canonic blocks still in pending
```
mpaulucci: nit: canonical

crates/networking/p2p/peer_handler.rs
fmoletta: Merged & removed limit [b82100b](https://github.com/lambdaclass/ethrex/pull/4812/commits/b82100b6199c32b42ef730ead9e9e95a81c00aef)

crates/networking/p2p/sync.rs
fmoletta: fixed!

crates/networking/p2p/peer_handler.rs
```diff
@@ -469,11 +469,13 @@ impl PeerHandler {
                     )
                     .await
                     {
-                        if are_block_headers_chained(&block_headers, &order) {
+                        if !block_headers.is_empty()
+                            && are_block_headers_chained(&block_headers, &order)
```
MegaRedHand, line 473: This was a change in behavior from later versions we noticed. Before, empty responses were never forwarded to the user, but somewhere down the line it changed, producing errors that were previously unreachable.

crates/networking/p2p/sync.rs
```diff
@@ -338,109 +315,220 @@ impl Syncer {
...
-                && first_block_hash == current_head
-                && current_head != sync_head
-            {
-                // There is no path to the sync head this goes back until it find a common ancerstor
-                warn!("Sync failed to find target block header, going back to the previous parent");
-                current_head = first_block_parent_hash;
-                continue;
-            }
+            let first_header = block_headers.first().ok_or(SyncError::NoBlocks)?;
+            let last_header = block_headers.last().ok_or(SyncError::NoBlocks)?;
```
MegaRedHand, line 358: These errors are actually unreachable, since `request_block_headers_from_hash` never returns an empty `Vec`. I think we should change, in another PR, `request_block_headers_from_hash` to return a plain `Vec` instead of an `Option<Vec<>>`, and interpret an empty `Vec` the same as the current `None`.

crates/storage/store_db/in_memory.rs
```diff
@@ -47,6 +47,8 @@ pub struct StoreInner {
     invalid_ancestors: HashMap<BlockHash, BlockHash>,
     // Stores current Snap State
     snap_state: SnapState,
+    // Stores fetched headers during a fullsync
+    fullsync_headers: HashMap<BlockNumber, BlockHeader>,
```
edg-l, line 51: if this is expected to have less than < 10000~ it might be better to use a btree too here

PR #4813, opened 2025-10-08
fix(levm): handle storage collisions when creating contracts

JereSalo:
**Motivation**

On [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610) it's mentioned that there are a few contracts in mainnet that have storage, but zero length code and no balance. A collision with those is highly unlikely but still possible, so that's why this behavior may be necessary. Also, many tests have been done for this kind of behavior so with this fix we'll pass them.

**Description**

- Change get_account_info to get_account_state
- Use variable storage_collision for cases in which this may impede creation of an account.
- Uncomment related tests.


mpaulucci (approved):
nice! 🚀

crates/common/types/block_execution_witness.rs
```diff
@@ -357,10 +357,10 @@ impl GuestProgramState {
 
     /// Retrieves the account info based on what is stored in the state trie.
```
cdiielsi:
```suggestion
    /// Retrieves the account state from the state trie.
```

crates/l2/common/src/state_diff.rs
```diff
@@ -539,11 +539,11 @@ pub fn get_nonce_diff(
     db: &impl VmDatabase,
 ) -> Result<u16, StateDiffError> {
     // Get previous account_info either from store or cache
```
cdiielsi:
```suggestion
    // Get previous account_state either from store or cache
```

PR #4815, opened 2025-10-08
chore(l1): remove snapshot dirs when they are no longer used

fedacking:
**Motivation**

Snap sync generates some intermediate files to limit the amount of memory used during the process. These are filled with the leaves downloaded during the first step of snap sync, before healing. These are large files that remain after the node is done using them. This pr aims to delete them, and to not launch a run if the previous files are in disk.

**Description**

- Deletes snapshot files for account state, storage slots and the temporary rocksdb db used for faster insert.
- Exits the process if we find the files in disk. Example of the log in that scenario.\

`ERROR ethrex_p2p::sync: One of the folders used for temporary leaves during snap is still used. Delete them in /home/admin/.local/share/ethrex`


crates/networking/p2p/sync.rs
```diff
@@ -201,6 +201,19 @@ impl Syncer {
...
 
+        // We validate that we have the folders that are being used empty, as we currently assume
+        // they are.
+        if !validate_folders(&self.datadir) {
+            // Temp std::process::exit until #2767 is done
+            error!(
+                "One of the folders used for temporary leaves during snap is still used. Delete them in {}",
+                &self.datadir.to_str().unwrap_or_default()
+            );
+            std::process::exit(-1);
```
MegaRedHand: Should this be 1 instead? -1 would be interpreted as 255 in Linux, which could be misinterpreted as a signal kill.

PR #4829, opened 2025-10-09
refactor(l1,l2): separate binaries

ManuelBilbao:
**Motivation**

We aim to reintroduce the L2 feature flag making so that ethrex has 3 available binaries, the L1 execution client, the L2 client and the L2+GPU client. This is done for the following advantages:
- Less crates needed for the L1 as the UI, L2, zkVMs crates are excluded.
- L1, L2, y zkVMs are compiled with the optimum crates for performance and their needs.
- The L1 user only downloads a client without unneeded functionality.

**Description**

- Added feature flag l2 to ethrex to compile l2 dependencies separately
- Added feature flag l2-sql to ethrex to launch l2 with the SQL database for the rollup
- Modified CI to create 2 tar images of ethrex for testing, ethrex:main and ethrex:main-l2
- Modified lint commands in makefile to lint l1 and l2 separately. L1 can't handle the workspace flag, due to some l2 crates in the workspace.

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -163,8 +163,15 @@ pub async fn init_rpc_api(
     )
     .await;
 
+    let ws_socket_opts = if opts.ws_enabled {
+        Some(get_ws_socket_addr(opts))
+    } else {
+        None
+    };
+
     let rpc_api = ethrex_rpc::start_api(
         get_http_socket_addr(opts),
+        ws_socket_opts,
         get_authrpc_socket_addr(opts),
         store,
         blockchain,
@@ -367,6 +374,11 @@ pub fn get_authrpc_socket_addr(opts: &Options) -> SocketAddr {
 pub fn get_http_socket_addr(opts: &Options) -> SocketAddr {
     parse_socket_addr(&opts.http_addr, &opts.http_port)
         .expect("Failed to parse http address and port")
+}
+
+pub fn get_ws_socket_addr(opts: &Options) -> SocketAddr {
+    parse_socket_addr(&opts.ws_addr, &opts.ws_port)
+        .expect("Failed to parse websocket address and port")
 }
 
 #[cfg(feature = "sync-test")]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -799,6 +799,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a18ed336352031311f4e0b4dd2ff392d4fbb370777c9d18d7fc9d7359f73871"
 dependencies = [
  "axum-core 0.5.5",
+ "base64 0.22.1",
  "bytes",
  "form_urlencoded",
  "futures-util",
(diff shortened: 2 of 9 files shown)


.github/actions/build-docker/action.yml
```diff
@@ -19,6 +19,14 @@ inputs:
     description: "Whether to push the built image to the registry"
     required: false
     default: "false"
+  artifact_path:
+    description: "The name of the artifact that is going to be pushed"
+    required: false
+    default: "ethrex_image.tar"
+  build_args:
+    description: "The feature flags to be build"
```
ManuelBilbao: Actually, this should be something like "args to use in docker build", as it could be any extra parameter

.github/workflows/tag_release.yaml
ManuelBilbao, line 1: We should now build two docker tags, shouldn't we? I'd use `ethrex:<version>[-l2]` (e.g., `ethrex:4.0.0` and `ethrex:4.0.0-l2`) but we can discuss other names

crates/l2/docker-compose.yaml
```diff
@@ -14,11 +14,11 @@ services:
 
   contract_deployer:
     container_name: contract_deployer
-    image: "ethrex:main"
+    image: "ethrex:main_l2"
```
ManuelBilbao: Based on the previous comment, we should change this too

.github/workflows/tag_release.yaml
ManuelBilbao: This will be needed in the `main_docker_publish` workflow as well for latest tags

.github/actions/build-docker/action.yml
fedacking: Done in [308126a](https://github.com/lambdaclass/ethrex/pull/4829/commits/308126a52fe6e221a219c0ad7b75177598a64689)

.github/workflows/tag_release.yaml
fedacking: Done in [308126a](https://github.com/lambdaclass/ethrex/pull/4829/commits/308126a52fe6e221a219c0ad7b75177598a64689)

crates/l2/docker-compose.yaml
fedacking: Done in [308126a](https://github.com/lambdaclass/ethrex/pull/4829/commits/308126a52fe6e221a219c0ad7b75177598a64689)

cmd/ethrex/initializers.rs
```diff
@@ -193,7 +194,7 @@ pub async fn init_network(
     store: Store,
     tracker: TaskTracker,
     blockchain: Arc<Blockchain>,
-    based_context: Option<P2PBasedContext>,
+    #[cfg(feature = "l2")] based_context: Option<P2PBasedContext>,
```
ilitteri, line 204: This is not optimal and we should fix this in another PR.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -725,13 +735,17 @@ async fn exchange_hello_messages<S>(
 where
     S: Unpin + Stream<Item = Result<Message, PeerConnectionError>>,
 {
+    // This allow is because in l2 we mut the capabilities
+    // to include the l2 cap
+    #[allow(unused_mut)]
```
ilitteri, line 746:
```suggestion
    #[expect(unused_mut)]
```

ilitteri (changes requested):
Block it until next version release

PR #4831, opened 2025-10-09
fix(l1): modify Grafana dashboards so they can be filtered by instance

### LINKED ISSUE #4976: Use block number as x-axis in grafana
Some alternatives are:
- Trying push-based prometheus.
- Using block number as label.

klaus993:
**Motivation**

The Ethrex L1 Perf grafana dashboard currently doesn't let us filter by instance, it only works as a standalone dashboard for 1 ethrex instance.

**Description**

The instance filter now has a regex to avoid having to choose from each prometheus target: `server:port` and instead just filter by `server` and let it choose the port according to the different metric each pane needs to check

Closes #4976

### DIFF
--- a/metrics/provisioning/grafana/dashboards/common_dashboards/ethrex_comparison.json
+++ b/metrics/provisioning/grafana/dashboards/common_dashboards/ethrex_comparison.json
@@ -103,7 +103,9 @@
           "id": 7,
           "options": {
             "legend": {
-              "calcs": [],
+              "calcs": [
+                "mean"
+              ],
               "displayMode": "list",
               "placement": "bottom",
               "showLegend": true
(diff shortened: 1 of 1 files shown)


mpaulucci:
I see this is actively being worked on. Converting to draft, please convert to review once its ready

rodrigo-o:
Apart for adding support for multiple servers (instances) to be able to compare different runs, the dashboard had major improvements:
- Better separation of panels thanks to rows
- Moved block building related panel to a WIP section (they are not as tested/checked as the other metrics/panels)
- Reduce scrape time to 5s to be able to account for more blocks (some were missing on 12s scrape time)
- Replaced the Block time & Gas Used % panel with two different ones
  - GGas/s by block 
  - Block time by block
- made the block execution breakdown repeated by instances  

This still have some points to address in follow up PRs here are 2 issues that we need to tackle:
- #5048
- #5049 

| Main | This PR |
| ----- | ----- |
| image | image |

rodrigo-o:
This PR also added a new dashboard intended for internal use, comparing some instances of other clients in the same panel.

rodrigo-o (approved):

PR #4836, opened 2025-10-09
refactor(l1): make TrieError::InconsistentTree more specific

### LINKED ISSUE #4786: Improve `TrieError::InconsistentTree` error reporting
We currently use `InconsistentTree` whenever we don't find a node in the database. We should add some metadata to this error, like which node wasn't found. This is of special importance once we prune older states, since the cause could be that the whole state tree is missing, for example.

cdiielsi:
**Motivation**

Adding some metadata to `TrieError::InconsistentTree` to have more information when we don't find a node in the database.

**Description**

This pr introduces a new enum to make the `TrieError::InconsistentTree` more specific:

```
#[derive(Debug, Error)]
pub enum InconsistentTreeError {
    #[error("Child node of {0}, differs from expected")]
    ExtensionNodeChildDiffers(ExtensionNodeErrorData),
    #[error("No Child Node found of {0}")]
    ExtensionNodeChildNotFound(ExtensionNodeErrorData),
    #[error("Node with hash {0:#x} not found in Branch Node with hash {1:#x} using path {2:?}")]
    NodeNotFoundOnBranchNode(H256, H256, Nibbles),
    #[error("Root node with hash {0:#x} not found")]
    RootNotFound(H256),
}

#[derive(Debug)]
pub struct ExtensionNodeErrorData {
    pub node_hash: H256,
    pub extension_node_hash: H256,
    pub extension_node_prefix: Nibbles,
    pub node_path: Nibbles,
}
```

Then I got InconsistentTreeError boxed because for some error variants it surpasses the 168 bytes limit allowed by clippy on results.

Closes #4786

### DIFF
--- a/crates/networking/p2p/rlpx/connection/handshake.rs
+++ b/crates/networking/p2p/rlpx/connection/handshake.rs
@@ -1,24 +1,16 @@
-use std::{
-    collections::HashMap,
-    net::SocketAddr,
-    sync::{Arc, RwLock},
-};
-
 use super::{
     codec::RLPxCodec,
     server::{Initiator, Receiver},
 };
 #[cfg(feature = "l2")]
 use crate::rlpx::l2::l2_connection::L2ConnState;
+use crate::{log_peer_debug, log_peer_trace};
 use crate::{
     rlpx::{
         connection::server::{ConnectionState, Established},
         error::PeerConnectionError,
         message::EthCapVersion,
-        utils::{
-            compress_pubkey, decompress_pubkey, ecdh_xchng, kdf, log_peer_debug, log_peer_trace,
-            sha256, sha256_hmac,
-        },
+        utils::{compress_pubkey, decompress_pubkey, ecdh_xchng, kdf, sha256, sha256_hmac},
     },
     types::Node,
 };
@@ -37,6 +29,11 @@ use secp256k1::{
     ecdsa::{RecoverableSignature, RecoveryId},
 };
 use sha3::{Digest, Keccak256};
+use std::{
+    collections::HashMap,
+    net::SocketAddr,
+    sync::{Arc, RwLock},
+};
 use tokio::{
     io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt},
     net::{TcpSocket, TcpStream},
@@ -71,7 +68,7 @@ pub(crate) async fn perform(
             let mut stream = match tcp_stream(addr).await {
                 Ok(result) => result,
                 Err(error) => {
-                    log_peer_debug(&node, &format!("Error creating tcp connection {error}"));
+                    log_peer_debug!(&node, &format!("Error creating tcp connection {error}"));
                     // context.table.lock().await.replace_peer(node.node_id());
                     return Err(error)?;
                 }
@@ -83,7 +80,7 @@ pub(crate) async fn perform(
             let hashed_nonces: [u8; 32] =
                 Keccak256::digest([remote_state.nonce.0, local_state.nonce.0].concat()).into();
             let codec = RLPxCodec::new(&local_state, &remote_state, hashed_nonces, eth_version)?;
-            log_peer_trace(&node, "Completed handshake as initiator");
+            log_peer_trace!(&node, "Completed handshake as initiator");
             (context, node, Framed::new(stream, codec))
         }
         ConnectionState::Receiver(Receiver {
@@ -109,7 +106,7 @@ pub(crate) async fn perform(
                 peer_addr.port(),
                 remote_state.public_key,
             );
-            log_peer_trace(&node, "Completed handshake as receiver");
+            log_peer_trace!(&node, "Completed handshake as receiver");
             (context, node, Framed::new(stream, codec))
         }
         ConnectionState::Established(_) => {
--- a/crates/networking/p2p/rlpx/connection/server.rs
+++ b/crates/networking/p2p/rlpx/connection/server.rs
@@ -1,37 +1,3 @@
-use std::{
-    collections::HashMap,
-    net::SocketAddr,
-    sync::{Arc, RwLock},
-    time::Duration,
-};
-
-use ethrex_blockchain::Blockchain;
-use ethrex_common::types::MempoolTransaction;
-#[cfg(feature = "l2")]
-use ethrex_common::types::Transaction;
-use ethrex_storage::{Store, error::StoreError};
-use ethrex_trie::TrieError;
-use futures::{SinkExt as _, Stream, stream::SplitSink};
-use rand::random;
-use secp256k1::{PublicKey, SecretKey};
-use spawned_concurrency::{
-    messages::Unused,
-    tasks::{
-        CastResponse, GenServer, GenServerHandle,
-        InitResult::{self, NoSuccess, Success},
-        send_interval, spawn_listener,
-    },
-};
-use spawned_rt::tasks::BroadcastStream;
-use tokio::{
-    net::TcpStream,
-    sync::{Mutex, broadcast, oneshot},
-    task::{self, Id},
-};
-use tokio_stream::StreamExt;
-use tokio_util::codec::Framed;
-use tracing::{debug, error};
-
 #[cfg(feature = "l2")]
 use crate::rlpx::l2::{
     PERIODIC_BATCH_BROADCAST_INTERVAL, PERIODIC_BLOCK_BROADCAST_INTERVAL,
@@ -41,6 +7,7 @@ use crate::rlpx::l2::{
 };
 use crate::{
     discv4::peer_table::PeerTable,
+    log_peer_debug, log_peer_error, log_peer_trace, log_peer_warn,
     metrics::METRICS,
     network::P2PContext,
     rlpx::{
@@ -61,7 +28,6 @@ use crate::{
             SUPPORTED_ETH_CAPABILITIES, SUPPORTED_SNAP_CAPABILITIES,
         },
         snap::TrieNodes,
-        utils::{log_peer_debug, log_peer_error, log_peer_trace, log_peer_warn},
     },
     snap::{
         process_account_range_request, process_byte_codes_request, process_storage_ranges_request,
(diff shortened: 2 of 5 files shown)


crates/common/trie/trie.rs
```diff
@@ -398,10 +418,12 @@ impl Trie {
                     if partial_path.skip_prefix(&extension_node.prefix)
                         && extension_node.child.is_valid()
                     {
-                        let child_node = extension_node
-                            .child
-                            .get_node(db)?
-                            .ok_or(TrieError::InconsistentTree)?;
+                        let child_node = extension_node.child.get_node(db)?.ok_or(
+                            TrieError::InconsistentTree(format!(
+                                "Extension Node child with hash {:?} with not found in db",
```
JereSalo:
```suggestion
                                "Extension Node child with hash {:?} not found in db",
```

crates/common/trie/node/branch.rs
```diff
@@ -73,9 +79,13 @@ impl BranchNode {
...
                 // Insert into existing child and then update it
                 (choice_ref, ValueOrHash::Value(value)) => {
-                    let child_node = choice_ref
-                        .get_node(db)?
-                        .ok_or(TrieError::InconsistentTree)?;
+                    let child_node =
+                        choice_ref
+                            .get_node(db)?
+                            .ok_or(TrieError::InconsistentTree(format!(
+                                "Node with hash {:?} not found in db",
```
JereSalo: Have you checked that `{:?}` is enough for showing the hash fully on screen? I tend to use `{:#x}` because in my experience other ways of printing show only the first and last numbers.

cdiielsi: `compute_hash()` returns a `NodeHash` and as far as I've checked `{:?}` is enough for showing the hash fully on screen. Screen will show either `Hashed(...full hash...)` or `Inline((slice, len))`

crates/common/trie/error.rs
```diff
@@ -7,8 +8,12 @@ pub enum TrieError {
     RLPDecode(#[from] RLPDecodeError),
     #[error("Verification Error: {0}")]
     Verify(String),
-    #[error("Inconsistent internal tree structure")]
-    InconsistentTree,
+    #[error("Inconsistent internal tree structure: Node with hash {0:?} not found")]
+    InconsistentTree(H256),
+    #[error("Inconsistent internal tree structure: Intermediate Node with hash {0:?} not found")]
+    InconsistentTreeOnIntermediateNode(H256),
```
MegaRedHand: The name is too big. Can we rename it to `IntermediateNodeNotFound`?

cdiielsi: Done! [Here](https://github.com/lambdaclass/ethrex/pull/4836/commits/8f37de63b4a6c697d3419409b0b1d6446c3ba198)

crates/common/trie/node/branch.rs
JereSalo: Oh right. So when it's an inline node will it show it nicely too? Like the bytes of the slice will they be pretty printed or ugly? If it's ugly maybe we can manually implement Debug or Display so that the output is nice.

cdiielsi: I've just checked and it doesn't show nicely for Inline nodes, the slice is displayed as an array. I think I'll go with manually implementing Debug or Display. But first, I have a question, I was thought about changing this and directly print the hash calculated with `.compute_hash().finalize()` for each node, but that would be wrong for Inline nodes, right?

JereSalo: I think that for Inline nodes we should show the node content itself, not a hash. TBH I don't see a reason to hash the inline node as it's not indexed in the trie by hash, it's just embedded into another node.

Make sure that it's clear when displaying the node which kind of node it is. What I want to avoid is a situation in which a person confuses an Inline for a hashed node. For example if you have an inline of 31 bytes you'll have `0x123...abc` and if you print just that the person won't easily know if it's an inline or hashed (they have to measure the length), so maybe it's better to be explicit about it.

crates/common/trie/node_hash.rs
```diff
@@ -138,3 +139,26 @@ impl RLPDecode for NodeHash {
         Ok((hash, rest))
     }
 }
+
+impl fmt::Debug for NodeHash {
```
MegaRedHand: I think this is no longer needed.

cdiielsi: Thanks!! 🙌  I missed it!

crates/common/trie/error.rs
```diff
@@ -7,8 +8,14 @@ pub enum TrieError {
...
-    #[error("Inconsistent internal tree structure")]
-    InconsistentTree,
+    #[error("Inconsistent internal tree structure: Node with hash {0:?} not found")]
+    InconsistentTree(H256),
+    #[error(
+        "Inconsistent internal tree structure: Node with hash {0:?} not found from Intermediate Node with hash {1:?}"
+    )]
+    IntermediateNodeNotFound(H256, H256),
+    #[error("Root node with hash {0:#x} not found")]
+    RootNotFound(H256),
```
JereSalo: I know this comment may be kind of annoying but isn't it better if we have just one enum for trie node not found and then we have the different variants inside of it? For me it would make more sense but I prefer hearing other people's opinions. We have 3 errors for saying the same thing, node wasn't found.

MegaRedHand: I think it's a good idea, but maybe in another PR. Removing `InconsistentTree` might also be worth it, since it's really unspecific, and probably could be replaced with multiple variants on this new enum.

crates/common/trie/error.rs
```diff
@@ -7,8 +8,14 @@ pub enum TrieError {
...
     #[error("Verification Error: {0}")]
     Verify(String),
-    #[error("Inconsistent internal tree structure")]
-    InconsistentTree,
+    #[error("Inconsistent internal tree structure: Node with hash {0:?} not found")]
+    InconsistentTree(H256),
+    #[error(
+        "Inconsistent internal tree structure: Node with hash {0:?} not found from Intermediate Node with hash {1:?}"
+    )]
+    IntermediateNodeNotFound(H256, H256),
```
JereSalo: Why not call it `ChildNodeNotFound` and say:
Child node X not found from parent Y

The name `IntermediateNodeNotFound` is kinda misleading because it sounds like the intermediate node wasn't found. The description above is correct though but I believe parent and child makes more sense

cdiielsi: True! On it!

crates/blockchain/Cargo.toml
```diff
@@ -38,3 +38,4 @@ metrics = ["ethrex-metrics/transactions"]
 
 [lints.clippy]
 unwrap_used = "deny"
+result_large_err = "allow"
```
MegaRedHand: We don't want to allow this. Having large errors can lead to performance issues in frequently called functions, so a local allow makes more sense. We could also just `Box` the big `InconsistentTree` variant, to keep the error size small at no runtime cost.

crates/common/trie/node/branch.rs
```diff
@@ -67,6 +74,7 @@ impl BranchNode {
         // If path is at the end, insert or replace its own value.
         // Otherwise, check the corresponding choice and insert or delegate accordingly.
         if let Some(choice) = path.next_choice() {
+            let current_node_hash = self.compute_hash().finalize();
```
MegaRedHand: We should move this inside each `ok_or_else`, to avoid computing it in the happy-path.

crates/common/trie/error.rs
cdiielsi: Done [here](https://github.com/lambdaclass/ethrex/pull/4836/commits/5f0c9bf2b47aa7a9cbdd1b71cb584850e9bd0967)!

crates/common/trie/trie.rs
```diff
@@ -100,14 +100,15 @@ impl Trie {
...
-                    .get(Nibbles::default())?
-                    .ok_or(TrieError::InconsistentTree)?,
-            )
-            .map_err(TrieError::RLPDecode)?
-            .get(self.db.as_ref(), path)?,
+            NodeRef::Hash(hash) if hash.is_valid() => {
+                Node::decode(&self.db.get(Nibbles::default())?.ok_or_else(|| {
+                    TrieError::InconsistentTree(InconsistentTreeError::RootNotFound(
+                        hash.finalize(),
+                    ))
```
MegaRedHand: I think it makes sense to put this inside a helper method like `get_root()` or similar, that maps the error.

crates/common/trie/node/extension.rs
```diff
@@ -75,7 +88,16 @@ impl ExtensionNode {
             let branch_node = if self.prefix.at(0) == 16 {
                 match new_node.get_node(db, path.current())? {
                     Some(Node::Leaf(leaf)) => BranchNode::new_with_value(choices, leaf.value),
-                    _ => return Err(TrieError::InconsistentTree),
+                    _ => {
```
MegaRedHand: I think here we should match against all variants, with an error for each of them.

crates/common/trie/error.rs
```diff
@@ -1,18 +1,37 @@
...
+    )]
+    ExtensionNodeInsertionError(H256, H256, Nibbles, Nibbles),
+    #[error("Node with hash {0:#x} not found in Branch Node with hash {1:#x} using path {2:?}")]
+    NodeNotFoundOnBranchNode(H256, H256, Nibbles),
+    #[error(
+        "Node with hash {0:#x} not found as child of Extension Node with hash {1:#x} and prefix {2:?} using path {3:?}"
+    )]
+    NodeNotFoundOnExtensionNode(H256, H256, Nibbles, Nibbles),
+    #[error("Root node with hash {0:#x} not found")]
+    RootNotFound(H256),
```
MegaRedHand: I think we have too many fields to keep them as tuple variants. We should move to using struct variants:


```rust
    #[error(
        "Branch Node with hash {node_hash:#x} not found to insert as child of Extension Node with hash {child_hash:#x} and prefix {prefix:?} using path {path:?}"
    )]
    ExtensionNodeInsertionError { node_hash: H256, child_hash: H256, prefix: Nibbles, path: Nibbles },
    // ...
```

crates/common/trie/node/branch.rs
cdiielsi: I've tried, but there's a problem with doing so because doing self.compute_hash().finalize() inside the `ok_or_else` borrows self as inmutable inside a `match` that is doing a mutable borrow of self, and I don't know how to work around this problem.

MegaRedHand: Maybe you can manually call `drop` on the mutable reference right before doing the immutable borrow.

cdiielsi: Using `drop` didn't work, in the end we dropped the mutable reference of `self.choices[choice]` and set its new value after the match. Changes [here](https://github.com/lambdaclass/ethrex/pull/4836/commits/2b946614f986a2b64685f9f70cd13d062e279ad9).

crates/common/trie/error.rs
```diff
@@ -1,18 +1,51 @@
...
     #[error("Invalid trie input")]
     InvalidInput,
 }
+
+#[derive(Debug, Error)]
+pub enum InconsistentTreeError {
+    #[error("Child node of {0}, differs from expected")]
+    ExtensionNodeChildDiffers(Box<ExtensionNodeErrorData>),
+    #[error("No Child Node found of {0}")]
+    ExtensionNodeChildNotFound(Box<ExtensionNodeErrorData>),
```
MegaRedHand: I think it's safe to remove the `Box` in these

crates/common/trie/error.rs
```diff
@@ -1,18 +1,51 @@
...
 #[derive(Debug, Error)]
 pub enum TrieError {
     #[error(transparent)]
     RLPDecode(#[from] RLPDecodeError),
     #[error("Verification Error: {0}")]
     Verify(String),
-    #[error("Inconsistent internal tree structure")]
-    InconsistentTree,
+    #[error("Inconsistent internal tree structure: {0}")]
+    InconsistentTree(Box<InconsistentTreeError>),
```
MegaRedHand, line 14: We should add a comment explaining this `Box` was added to make the error smaller, since this variant only was bigger than 100 bytes.

cdiielsi: Done [here](https://github.com/lambdaclass/ethrex/pull/4836/commits/e7212655ef50fb20e207e3cd97d9ddd609ef01d3)!

crates/common/trie/error.rs
cdiielsi: Done [here](https://github.com/lambdaclass/ethrex/pull/4836/commits/e7212655ef50fb20e207e3cd97d9ddd609ef01d3)!

PR #4838, opened 2025-10-10
perf(l1): optimize get closest nodes

edg-l:
According to profiling, the get_closest_nodes is mainly bound by iterating the contacts btreemap.

Replaced the contacts btreemap with indexmap. Also replaced the peers one since it also beneffits from this.

Comparing profiling info, it goes from 31% of time taken to 3%.

Hoodi synced in 7mins 32seconds more or less

image

Reasoning:

Btreemap has fast lookup and slow iteration.
indexmap has fastish lookup and vec-like iteration

```
IndexMap derives a couple of performance facts directly from how it is constructed, 
which is roughly:

    A raw hash table of key-value indices, and a vector of key-value pairs.

    Iteration is very fast since it is on the dense key-values.

    Removal is fast since it moves memory areas only in the table, 
and uses a single swap in the vector.

    Lookup is fast-ish because the initial 7-bit hash lookup uses SIMD, and indices are 
densely stored. Lookup also is slow-ish since the actual key-value pairs are stored separately. (Visible when cpu caches size is limiting.)

```

introduces a new dep: https://crates.io/crates/indexmap ( 662,138,228 downloads )

### DIFF
--- a/crates/networking/p2p/discv4/peer_table.rs
+++ b/crates/networking/p2p/discv4/peer_table.rs
@@ -1,7 +1,7 @@
 use crate::{
     discv4::server::MAX_NODES_IN_NEIGHBORS_PACKET,
     metrics::METRICS,
-    rlpx::{self, connection::server::RLPxConnection, p2p::Capability},
+    rlpx::{connection::server::PeerConnection, p2p::Capability},
     types::{Node, NodeRecord},
 };
 use ethrex_common::{H256, U256};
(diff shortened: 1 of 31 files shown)


crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -606,7 +607,7 @@ impl PeerTable {
             .collect::<Vec<_>>();
 
         for contact_to_discard_id in disposable_contacts {
-            self.contacts.remove(&contact_to_discard_id);
+            self.contacts.swap_remove(&contact_to_discard_id);
```
MegaRedHand, line 554: Does indexmap have something like Vec::retain ? We could replace this loop with that

edg-l: i would prefer to look at this in another pr/change

PR #4839, opened 2025-10-10
chore(l1): memory enhancement - no transactional rocksdb + no cache

rodrigo-o:
**Motivation**

This PR reduce memory by 40~50%

**Description**

This PR sits on top of #4821 and after different testing scenarios we came up with this cut in memory:
- reverts #4599
- removes the rocksdb cache in options 

The caveat is that it appears to be 50% slower, we need to double check it.

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -250,7 +250,7 @@ impl Store {
                     cf_opts.set_compression_type(rocksdb::DBCompressionType::Lz4);
                     cf_opts.set_write_buffer_size(64 * 1024 * 1024); // 64MB
                     cf_opts.set_max_write_buffer_number(3);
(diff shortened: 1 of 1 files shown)


rodrigo-o:
Here is a table comparision of 2 runs of this PR vs main:

_Node Memory (RSS) & Host in images_
| Main | This PR #1 | This PR #2 |
| ----- | ----- | ----- |
| Snap time: 03:30 | Snap time: 05:20 | Snap time: 05:30 |
| Mean: ~15.5GB & Max: ~22GB | Mean: ~8GB & Max: ~13GB | Mean: ~9.5GB & Max: ~15GB |
| image | image | image | 

This is an enhancement of \~40\~50% in the Mean and \~30\~40% in peaks, it appears to be a noticeable regression related to time \~50% slower, but we need more test to validate it wasn't an issue of the particular runs (variance for the runs is still high)

rodrigo-o:
Closed in favor of #4853

PR #4843, opened 2025-10-13
perf(levm): remove unnecesary mul in ecpairing

### LINKED ISSUE #4817: Unnecessary multiplication in ecpairing precompile
in `precompiles.rs`:
```
    if !valid_pairs.is_empty() {
        let batch: Vec<_> = valid_pairs.iter().map(|(p1, p2)| (p1, p2)).collect();
        let pairing_result = BN254AtePairing::compute_batch(&batch)
            .map_err(|_| PrecompileError::BN254AtePairingError)?;
        mul *= pairing_result;
    }

    // Generate the result from the variable mul
    let success = mul.eq(&QuadraticExtensionFieldElement::one());
    let mut result = [0; 32];
    result[31] = u8::from(success);
    Ok(Bytes::from(result.to_vec()))
```
`mul` is initialized as
```  
    let mut mul: FieldElement<Degree12ExtensionField> = QuadraticExtensionFieldElement::one();
```
so this is always doing `1 * Gt`, we should just check that `pairing_result == QuedraticExtensionFieldElement::one()`.

edg-l:
**Motivation**

mul is initialized as

    let mut mul: FieldElement<Degree12ExtensionField> = QuadraticExtensionFieldElement::one();

so this is always doing 1 * Gt, we should just check that pairing_result == QuedraticExtensionFieldElement::one().

**Description**

Closes #4817

### DIFF
--- a/crates/vm/levm/src/precompiles.rs
+++ b/crates/vm/levm/src/precompiles.rs
@@ -934,19 +934,19 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
         }
     }
 
-    let success = if !valid_pairs.is_empty() {
+    let success = if valid_pairs.is_empty() {
+        true
+    } else {
         let batch: Vec<_> = valid_pairs.iter().map(|(p1, p2)| (p1, p2)).collect();
         let pairing_result = BN254AtePairing::compute_batch(&batch)
             .map_err(|_| PrecompileError::BN254AtePairingError)?;
         pairing_result.eq(&QuadraticExtensionFieldElement::one())
-    } else {
-        true
     };
 
     // Generate the result from the variable mul
     let mut result = [0; 32];
     result[31] = u8::from(success);
-    Ok(Bytes::from(result.to_vec()))
+    Ok(Bytes::from_owner(result))
 }
 
 /// Returns the result of Blake2 hashing algorithm given a certain parameters from the calldata.


crates/vm/levm/src/precompiles.rs
```diff
@@ -936,19 +934,16 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
...
+        pairing_result.eq(&QuadraticExtensionFieldElement::one())
+    } else {
+        true
+    };
 
     // Generate the result from the variable mul
-    let success = mul.eq(&QuadraticExtensionFieldElement::one());
     let mut result = [0; 32];
     result[31] = u8::from(success);
     Ok(Bytes::from(result.to_vec()))
```
MegaRedHand: I think we can avoid an allocation here by creating the Bytes from a static array value.

crates/vm/levm/src/precompiles.rs
```diff
@@ -936,19 +934,16 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
         }
     }
 
-    #[allow(
-        clippy::arithmetic_side_effects,
-        reason = "library will not panic on mul overflow"
-    )]
-    if !valid_pairs.is_empty() {
+    let success = if !valid_pairs.is_empty() {
```
MegaRedHand: We should invert this condition.

PR #4844, opened 2025-10-13
perf(l1): remove dynamic dispatch in favour of known Vec in rlp encoding trait

edg-l:
**Motivation**

This should enable better optimizations from the compiler since it will use a known type: Vec, instead of the dynamic fat pointer from the bufmut trait.

I looked at all the source code, and we are using a trait when we always passa Vec here. So the trait overhead is essentially useless.

**Description**

Remove the bufmut trait usage, use a Vec directly

This is the first pr im doing as i look to improve the rlp encoding/decoding, since its a known overhead now.


edg-l:
closing as it doesn-t seem to improve the performance

Arkenan:
Closing again as it doesn't seem to improve performance.

PR #4850, opened 2025-10-13
refactor(l1): improve the way we remove ONLY the storage of an account

### LINKED ISSUE #3321: Add `remove_storage` to `AccountUpdate`
After https://github.com/lambdaclass/ethrex/pull/3293

There is an edge case in which we want to remove just the storage of an account without removing the account itself. For now what we are doing is remove the whole account and then insert it again because that's what we can do with the tools that we have. The problem was handled in https://github.com/lambdaclass/ethrex/pull/3293.
I don't know how trivial it is to insert an empty trie to the account that already exists, I took a look and there doesn't seem to be a method for doing that so it will require more effort than expected.
This addition will improve `get_state_transitions()` code quality.

JereSalo:
**Motivation**

- For removing **just** the storage of an account we were doing something "dirty", which was inserting 2 updates for the same account. One with its destruction and another one with the changes, this cleans that up.

**Description**

- Add `removed_storage` attribute to `AccountUpdate`
  - If we want to remove whole storage of an account we just change the storage root of the account to the empty root.

Closes #3321


tooling/ef_tests/state/runner/revm_db.rs
```diff
@@ -198,6 +198,7 @@ impl RevmState {
                                 )
                             })
                             .collect(),
+                        removed_storage: false,
```
JereSalo, line 201: We could improve code around this so that the get state transitions of revm makes use of `removed_storage` but I consider it to be irrelevant, we are going to remove this code whenever we can anyway

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -213,11 +189,22 @@ impl GeneralizedDatabase {
...
             // 2. Storage has been updated if the current value is different from the one before execution.
             let mut added_storage = BTreeMap::new();
 
             for (key, new_value) in &new_state_account.storage {
-                let old_value = initial_state_account.storage.get(key).ok_or_else(|| { VMError::Internal(InternalError::Custom(format!("Failed to get old value from account's initial storage for address: {address}")))})?;
+                let old_value = if !removed_storage {
+                    initial_state_account.storage.get(key).ok_or_else(|| { VMError::Internal(InternalError::Custom(format!("Failed to get old value from account's initial storage for address: {address}")))})?
+                } else {
+                    // There's not an "old value" if the contract was destroyed and re-created.
+                    &ZERO_U256
```
fmoletta, line 206: Why not `U256::zero()`?

fmoletta (approved):

crates/vm/levm/src/db/gen_db.rs
JereSalo: Because it would create a reference that's dropped every time

PR #4854, opened 2025-10-13
ci(l1): make workflows run only when appropriate.

### LINKED ISSUE #3457: Avoid running some workflows on title change
There are some workflows that run on `pull_request`, but that happens when there's any change to the PR, like changing the title or opening the PR, etc.
Maybe we can change that and use the activity types defined [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request).

JereSalo:
**Motivation**

- There were some workflows that ran on simple changes like changing the title of the PR and that's pretty unnecessary. This is a simple change that will improve experience on that side at least.

**Description**

- Now on things like title change just a few appropriate workflows run

This is a very basic change that I repeated across workflows, suggestions are welcome

Closes #3457

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ tooling/ef_tests/state/runner_v2/success_report.txt
 
 tooling/reorgs/data
 
+tooling/sync/logs/
+
 # Repos checked out by make target
 /hive/
 ethereum-package/
--- a/crates/l2/sdk/src/sdk.rs
+++ b/crates/l2/sdk/src/sdk.rs
@@ -36,10 +36,10 @@ pub use ethrex_sdk_contract_utils::*;
 
 use calldata::from_hex_string_to_h256_array;
 
-// 0xcb968a1441ffcfb88194912923c5353f8da0210b
+// 0x36664d7c5031bd965bbb405b55495a90dd780740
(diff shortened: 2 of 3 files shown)


.github/workflows/daily_hive_report.yaml
```diff
@@ -10,6 +10,7 @@ on:
       - .github/workflows/daily_reports.yaml
       - .github/workflows/common_hive_reports.yaml
       - .github/scripts/publish_hive.sh
+    types: [opened, synchronize, reopened, ready_for_review]
```
mpaulucci, line 13: > There were some workflows that ran on simple changes like changing the title of the PR and that's pretty unnecessary

I don't think the default value of types includes that scenario. Where are you seeing this?

JereSalo: Oh, I'm sorry. As I mentioned in [the issue](https://github.com/lambdaclass/ethrex/issues/3457) I had experiences in the past in which I changed the title of the PR and some workflows ran but it seems that it only happened in a few workflows. I mindlessly added the types when it wasn't necessary it seems, my bad.

JereSalo:
Closing this as it wasn't necessary it seems. I made a mistake.

PR #4855, opened 2025-10-13
refactor(l1,levm): simplify vm database trait.

JereSalo:
**Motivation**

- Trait may have been unnecessarily complex when we could've used LevmDatabase trait directly instead without problems.

**Description**

- Remove VmDatabase and related structures and implement LevmDatabase in all the Storage structs we use.


crates/blockchain/Cargo.toml
```diff
@@ -12,6 +12,7 @@ ethrex-rlp.workspace = true
 ethrex-common.workspace = true
 ethrex-storage.workspace = true
 ethrex-vm.workspace = true
+ethrex-levm.workspace = true
```
mpaulucci: if we have to add this here, then the vm crate doesn't really have a reason to exist.

JereSalo: Thanks for pointing that out. I moved the `StoreVmDatabase` to the vm crate so that it lives there with the other databases that implement the `LevmDatabase` trait. I think it's tidier this way but I don't know if you think it's a good idea to have it there or not.

crates/vm/backends/levm/db/witness.rs
```diff
@@ -57,41 +58,52 @@ impl GuestProgramStateWrapper {
...
+        &self,
+        address: Address,
+    ) -> Result<ethrex_common::types::AccountState, DatabaseError> {
         self.lock_mutex()
-            .map_err(|_| EvmError::DB("Failed to lock db".to_string()))?
-            .get_account_code(code_hash)
-            .map_err(|_| EvmError::DB("Failed to get account code".to_string()))
+            .map_err(|_| DatabaseError::Custom("Failed to lock db".to_string()))?
+            .get_account_state(address)
+            .map(|opt| opt.unwrap_or_default())
```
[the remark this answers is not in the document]
JereSalo: Yup, default account state means for levm that the account is not present basically.

crates/vm/backends/levm/db/witness.rs
```diff
@@ -57,41 +58,52 @@ impl GuestProgramStateWrapper {
...
 
-    fn get_account_state(&self, address: Address) -> Result<Option<AccountState>, EvmError> {
+    fn get_storage_value(&self, address: Address, key: H256) -> Result<U256, DatabaseError> {
         self.lock_mutex()
-            .map_err(|_| EvmError::DB("Failed to lock db".to_string()))?
-            .get_account_state(address)
-            .map_err(|_| EvmError::DB("Failed to get account info".to_string()))
+            .map_err(|_| DatabaseError::Custom("Failed to lock db".to_string()))?
+            .get_storage_slot(address, key)
+            .map(|opt| opt.unwrap_or_default())
```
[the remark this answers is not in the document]
JereSalo: In ethereum -> Unset = set to zero

crates/vm/backends/levm/db/mod.rs
JereSalo, line 1: One of the core changes, the databases that implement Levm trait are now in the db folder

tooling/ef_tests/state/runner/revm_db.rs
JereSalo, line 1: Changes in this file are correct but unimportant because we don't care about the revm runner of eftests anymore. I suggest ignoring them

JereSalo:
We can close this as the impact of this PR wasn't very high and it got old. In the future we can make a similar refactor if we think we should.

PR #4857, opened 2025-10-13
chore(l2): add prover performance/optimization docs

xqft:
**Motivation**

Includes comparisons between different optimizations in proving times and zkVM cycles

### DIFF
--- a/docs/l2/bench/performance_observations.md
+++ b/docs/l2/bench/performance_observations.md
@@ -1,0 +1,62 @@
+# Notes Taken While Exploring zkVM Optimization Topics. Dec 1, 2025
+
+- **Transaction Sender Recovery:**  
+  I noticed that in zkVMs we're using `k256`. It might be worth checking whether there's a performance difference when using `secp256k1` directly. This impacts blocks mostly composed of transfers, because every 21k gas equals one transaction.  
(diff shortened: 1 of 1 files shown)


docs/l2/bench/prover_performance.md
```diff
@@ -0,0 +1,216 @@
...
+
+before
+![alt text](img/before_initial_hash_caching.png)
+
+after
+![alt text](img/after_initial_hash_caching.png)
+
+although the total proving time didn’t change that much:
+
+![alt text](img/initial_hash_caching_prove_time.png)
```
MegaRedHand, line 152: Some images aren't rendering for me:

image failed to render

xqft: image
i can see all of them, might be github not rendering them properly

docs/l2/bench/prover_performance.md
```diff
@@ -0,0 +1,216 @@
...
+            let mut option_cache = OptionCache { inner: None };
+            option_cache.init(i);
+            options.push(option_cache);
+        }
+    }
+    ```
+
+</details>
+
+in the flamegraph it’s not clear where that `memcpy` is originating (any ideas?). The calls that are on top of it are `encode_bytes()` (from our RLP lib) and some `sha3` calls.
```
tomip01, line 241: do we keep this `any ideas?` comment?

PR #4864, opened 2025-10-14
feat(l1): initial WIP Nix flake.

samoht9277:
**Description**

Enables single command execution of Ethrex binary without cloning the repo.

Running `nix build` compiles the app and produces the `./result/bin/ethrex` binary
Running `nix develop` compiles the app and enters a subshell with the compiled ethrex binary
Running `nix run` compiles the app and automatically starts the app with no shell (`nix run . -- <arg>` to pass arguments).
Running `nix shell` compiles the app and adds ethrex to your current shell (no sub-shell)

TODO: Add support for gpu builds (pass cargo features depending on platform)

**Prebuilt Binaries**
1. Add your user to the `trusted-users`
  *  `echo "trusted-users = root admin" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon`
3. Install cachix 
  * `nix profile add nixpkgs#cachix`
5. Run `cachix use ethrex`
6. Run `nix run github:lambdaclass/ethrex`


flake.nix
```diff
@@ -0,0 +1,85 @@
...
+        };
+
+        toolchain = pkgs.rust-bin.stable."1.89.0".default;
+
+        rustPlat = pkgs.makeRustPlatform {
+          cargo = toolchain;
+          rustc = toolchain;
+        };
+
+        version = "5.0.0";
```
klaus993, line 31: this will have to be updated after each release, right? maybe we could automate it in a future improvement (make the release workflow change and commit this version change in this file)

mpaulucci:
I see this is actively being worked on. Converting to draft for now

flake.nix
samoht9277: Yup. We can add a step inside the nix build workflow (.github/workflows/publish-nix.yml) to update. 

The bad part is that we need the workflow to commit the new file to main in order to get the new version working

klaus993 (approved):

flake.nix
```diff
@@ -0,0 +1,85 @@
+{
+  description = "Ethrex Nix";
+
+  inputs = {
+    nixpkgs.url     = "github:junjihashimoto/nixpkgs/60fe186313677e8dbf04f917a2b4c3d984be4efd"; #TODO: Change this to upstream nixpkgs when allowBuiltinFetchGit bug is solved (https://discourse.nixos.org/t/cannot-build-package-of-rust-application-ln-failed-to-create-symbolic-link-permission-denied/57587)
```
ElFantasma, line 5: **Supply chain concern:** Pinning `nixpkgs` to a personal fork (`junjihashimoto/nixpkgs`) means the project depends on an individual’s repo for all system packages. Even with the workaround comment, this fork won’t receive upstream security patches unless manually synced. The referenced `allowBuiltinFetchGit` bug should be checked against current upstream `nixpkgs-unstable` — it may already be fixed (this PR has been open 4+ months). If it’s fixed, switch to `NixOS/nixpkgs`. If not, document the risk more prominently and track when to switch back.

flake.nix
```diff
@@ -0,0 +1,85 @@
...
+
+  outputs = { self, nixpkgs, flake-utils, crane, rust-overlay }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ (import rust-overlay) ];
+        };
+
+        toolchain = pkgs.rust-bin.stable."1.89.0".default;
```
ElFantasma, line 24: **Version mismatch:** This pins Rust 1.89.0, but `rust-toolchain.toml` on `main` specifies 1.90.0. These should be kept in sync — ideally read from `rust-toolchain.toml` dynamically, or at minimum updated to match. Diverging toolchain versions can cause subtle compilation differences.

flake.nix
ElFantasma, line 31: **Stale version:** Hardcoded to `5.0.0`, but `main` is at v9.0.0. More importantly, `fetchFromGitHub` below fetches this specific tag — meaning `nix build` inside the repo always builds v5.0.0, never the local working tree. For local builds, use `self` or a filtered source path (e.g., `pkgs.lib.cleanSource ./.`). The `fetchFromGitHub` approach could be kept for the `apps.default` remote usage (`nix run github:lambdaclass/ethrex`), but for `nix build` in the repo it should build what’s checked out.

flake.nix
```diff
@@ -0,0 +1,85 @@
...
+          rustc = toolchain;
+        };
+
+        version = "5.0.0";
+
+        src = pkgs.fetchFromGitHub {
+          owner  = "lambdaclass";
+          repo   = "ethrex";
+          rev    = "v${version}";
+          sha256 = "sha256-ajPLu5j/sjxLEFN1F+C/f12cbDenzUapBPfStem0Tj4=";
```
ElFantasma, line 37: The `sha256` hash is tied to the v5.0.0 snapshot. Every version bump requires manually updating this hash. With the `self`-based source approach suggested above, this entire `fetchFromGitHub` block becomes unnecessary for local builds.

.github/workflows/publish-nix.yml
```diff
@@ -0,0 +1,74 @@
...
+  release:
+    types:
+      - published
+
+permissions:
+  contents: read
+
+jobs:
+  distribute-build:
+    if: ${{ github.event.release.prerelease == false }}
```
ElFantasma, line 19: The `if: ${{ github.event.release.prerelease == false }}` condition works for `release` events, but for `workflow_dispatch` events `github.event.release` is undefined, so `prerelease` is `null`. This happens to evaluate as `null == false -> true` in GitHub Actions expressions, so it works — but it’s fragile and non-obvious. Consider: `if: ${{ github.event_name == 'workflow_dispatch' || github.event.release.prerelease == false }}`

ElFantasma (changes requested):

MegaRedHand:
Closing due to staleness

PR #4867, opened 2025-10-14
feat(l1): lowmem mode

Oppen:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/docs/l1/running/configuration.md
+++ b/docs/l1/running/configuration.md
@@ -51,6 +51,12 @@ ethrex --log.level <level>
 
 Levels: `error`, `warn`, `info` (default), `debug`, `trace`
 
+## Low Memory Mode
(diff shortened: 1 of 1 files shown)


lferrigno:
The code currently copy-pastes the Store implementation because making it generic would basically require some magic. 

This should get much simpler once one of the following happens:

 - The DBInner trait is exposed and released (@Oppen  opened an issue here: https://github.com/rust-rocksdb/rust-rocksdb/issues/1046
), or

 - The store API redesign lands, which will at least reduce the amount of code per implementation.

⚠️ Note: any merge from main should manually readjust the Store copy.

Oppen:
It turns out the low-mem mode is faster, so no need for a flag, just make it default

PR #4873, opened 2025-10-14
chore(l2): estimate L2 tps

xqft:
**Motivation**

To measure the maximum transactions per second that the current L2 can withstand, using a single prover with SP1 backend running on a RTX 4090.

Multiple L2 setups have been tested.

This also adds some tools used to easily test these scenarios


tooling/load_test/src/main.rs
```diff
@@ -329,87 +494,142 @@
...
+            "Load test finished. Total elapsed time: {} seconds",
+            elapsed_time.as_secs()
+        );
+        Ok(())
+    }
+}
+
+impl SubcommandGenerateGenesis {
+    pub async fn run(self) -> eyre::Result<()> {
+        println!(
```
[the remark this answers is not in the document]
xqft: non issue

xqft:
Don't merge yet, still missing a row in the measures .md

docs/l2/bench/tps.md
tomip01, line 1: All the tables have the column `Avg. batch size (blocks)` equal to 63, should we remove it and add as a comment?

ManuelBilbao (approved):

docs/l2/bench/tps.md
xqft: yes that's better, done!

tomip01 (approved):

xqft:
Don't merge this until reviewed by @mationorato

MegaRedHand (changes requested):
https://github.com/lambdaclass/ethrex/pull/4873#issuecomment-3473509524

tooling/load_test/src/main.rs
```diff
@@ -330,87 +502,142 @@ fn parse_private_key_into_local_signer(pkey: &str) -> Signer {
...
+            let (secret_key, public_key) = secp.generate_keypair(&mut rand::thread_rng());
+            let address = public_key_to_address(&public_key);
+
+            writeln!(keys_file, "{}", hex::encode(secret_key.secret_bytes()))?;
+
+            genesis.alloc.insert(
+                address,
+                GenesisAccount {
+                    code: Bytes::new(),
+                    storage: HashMap::new(),
```
ElFantasma, line 625: `GenesisAccount.storage` is `BTreeMap<U256, U256>`, but this passes `HashMap::new()`. This won't compile.

```suggestion
                    storage: Default::default(),
```
Or explicitly `BTreeMap::new()` with the appropriate import.

tooling/load_test/src/main.rs
```diff
@@ -53,28 +71,77 @@ struct Cli {
...
+
+#[derive(Parser)]
+pub struct SubcommandGenerateGenesis {
     #[arg(
-        short = 'N',
         long,
-        default_value_t = 1000,
-        help = "Number of transactions to send for each account."
+        help = "Name of the network or genesis file. Supported: mainnet, holesky, sepolia, hoodi, local-devnet-l2. Default: mainnet",
+        value_parser = clap::value_parser!(Network),
```
ElFantasma, line 92: `clap::value_parser!(Network)` requires `Network` to implement `FromStr` (or `ValueEnum`), but `Network` only has `From<&str>`. This won't compile as-is.

Either add `impl FromStr for Network` in `ethrex-config`, or use a custom value parser here, e.g.:
```rust
value_parser = |s: &str| -> Result<Network, String> { Ok(Network::from(s)) },
```

tooling/load_test/src/main.rs
```diff
@@ -247,63 +317,161 @@ async fn load_test(
...
+    }
+
+    let interval = Duration::from_secs_f64(1.0 / rate as f64);
+    let mut ticker = tokio::time::interval(interval);
+
+    let mut nonce_map = HashMap::new();
+    let mut account_index = 0;
+    let mut tasks = JoinSet::new();
+
+    loop {
```
ElFantasma, line 352: `tps_load_test` runs an infinite `loop` and never returns `Ok(())` — it can only exit via `?` propagation or a panic. This means the caller's elapsed-time printout (line ~594) is dead code when `LoadType::Tps` is selected. Consider either:
- Adding a `--duration` flag so the loop exits after a set time, or
- Printing periodic stats inside the loop (e.g., every 10s), since the final summary will never be reached.

tooling/load_test/src/main.rs
```diff
@@ -247,63 +317,161 @@ async fn load_test(
...
+                &client,
+                TxType::EIP1559,
+                dst,
+                address,
+                calldata.into(),
+                Overrides {
+                    chain_id: Some(chain_id),
+                    value,
+                    nonce: Some(current_nonce),
+                    max_fee_per_gas: Some(i64::MAX as u64),
```
ElFantasma, line 387: nit: `i64::MAX as u64` is an unusual way to express "maximum willingness to pay". Consider `u64::MAX` directly, or better yet, fetch the current base fee and apply a multiplier — this avoids potentially overpaying if the load test is ever pointed at a real network.

crates/common/config/networks.rs
```diff
@@ -58,6 +58,7 @@ impl From<&str> for Network {
             "holesky" => Network::PublicNetwork(PublicNetwork::Holesky),
             "mainnet" => Network::PublicNetwork(PublicNetwork::Mainnet),
             "sepolia" => Network::PublicNetwork(PublicNetwork::Sepolia),
+            "local-devnet-l2" => Network::LocalDevnetL2,
```
ElFantasma, line 61: The `"local-devnet-l2"` arm is added before the `GenesisPath` fallback, which is correct. However, note the comment above says "we don't allow to manually specify the local devnet genesis" — this new arm intentionally allows it for the L2 variant. Might be worth updating that comment to clarify the distinction.

ElFantasma (changes requested):
This branch is ~4 months old — please rebase on main and verify it compiles before requesting re-review or close it if it's no longer relevant

MegaRedHand:
Closing due to staleness

PR #4883, opened 2025-10-15
test(l1): update state and blockchain tests to last version

JereSalo:
**Motivation**

Use latest fixtures

### DIFF
--- a/.github/workflows/pr-main_l1.yaml
+++ b/.github/workflows/pr-main_l1.yaml
@@ -144,9 +144,10 @@ jobs:
         include:
           - name: "Rpc Compat tests"
             simulation: ethereum/rpc-compat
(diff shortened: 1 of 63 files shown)


tooling/ef_tests/blockchain/.fixtures_url
```diff
@@ -1 +1 @@
-https://github.com/ethereum/execution-spec-tests/releases/download/v5.1.0/fixtures_develop.tar.gz
```
mpaulucci, line 1: you're only updating blockchain tests here, not state

JereSalo: I'm sorry, in the past they used the same fixture but I see now they have each one fixture. Didn't see that

JereSalo: There we go
https://github.com/lambdaclass/ethrex/pull/4883/commits/a6a6aeb4c116b16086a16a019b438bcd1f65d599

PR #4884, opened 2025-10-15
feat(l1): implement flatkeyvalue (snapshots)

### LINKED ISSUE #1997: perf: state snapshots
Having to process the state trie to query all account information, as well as having to compute the sender's public key from its address at least once per transaction, introduces significant overhead to both L1 and L2.
This must be done when receiving new data from other nodes, but it's not necessary for internal operation, given we either already verified the information or created it ourselves. We could use a flat representation of current state most of the time.
The observation is not new, and [Geth already uses state snapshots](https://blog.ethereum.org/2020/07/17/ask-about-geth-snapshot-acceleration).

While this would introduce significant complexity, it would also make a big reduction in the time we take to include transactions in the mempool, according to profiles performed on load tests for the L2.

In terms of design, we could start with something simpler than Geth, just a table on the DB with each address's data on the latest state, recomputing from the trie on reorgs (e.g. we could have a generation number of some sort to detect stale data).

### LINKED ISSUE #2409: Account state by address snapshot.
Required changes: 
- Every query for account state (e.g. balance or nonce) should go to this snapshot instead of traversing the trie. 
- When finishing a state transition for the head block, we should modify the snapshot.
- When doing a reorg, we should modify the snapshot to reflect the newly selected trie.

This will not be very performant at first. We should then add the difflayers as a separate ticket for more efficient reorgs.

### LINKED ISSUE #2410: Account storage by address and location snapshot
Every storage request should go through here.

iovoid:
**Motivation**

We want to have a way to query values directly without going through the trie (several reads).

To do this, we have a flattened key=>value mapping, which we call FlatKeyValue. This is sometimes known as "snapshots".

**Description**

A process is launched to generate this mapping in the background. We ensure `apply_updates` and this process don't conflict, and that block execution can only make use of the calculated snapshots.

Closes #1997, closes #2409, closes #2410

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -142,7 +142,7 @@ Block producer options:
       --block-producer.extra-data <EXTRA_DATA>
           Block extra data message.
 
-          [default: "ethrex 3.0.0"]
+          [default: "ethrex 4.0.0"]
 ```
 
 <!-- END_CLI_HELP -->


crates/storage/store_db/in_memory.rs
```diff
@@ -686,6 +686,10 @@ impl StoreEngine for Store {
 
         Ok(())
     }
+
+    fn generate_snapshot(&self) -> Result<(), StoreError> {
+        todo!()
```
pablodeymo: relace this `todo!` for a `Ok(())`.

crates/storage/store_db/rocksdb.rs
```diff
@@ -101,10 +101,22 @@ const CF_PENDING_BLOCKS: &str = "pending_blocks";
 /// - [`Vec<u8>`] = `BlockHashRLP::from(latest_valid).bytes().clone()`
 const CF_INVALID_ANCESTORS: &str = "invalid_ancestors";
 
-#[derive(Debug)]
+const CF_SNAPSHOTS: &str = "snapshots";
```
pablodeymo: we can use a better name for this CF. For example: `flatkeyvalue`

crates/storage/store_db/rocksdb.rs
```diff
@@ -298,11 +312,27 @@ impl Store {
...
+        let store = Self {
             db: Arc::new(db),
             trie_cache: Default::default(),
-        })
+            snapshot_pivot_tx: tx,
+        };
+        let store_clone = store.clone();
+        std::thread::spawn(move || {
+            let mut rx = rx;
+            while rx.recv() != Ok(SnapshotControlMessage::Continue) {}
```
pablodeymo: we should take in account `Err` results of this call.

crates/storage/store_db/rocksdb.rs
```diff
@@ -437,6 +467,111 @@ impl Store {
...
+
+    fn snapshoting_loop(
+        &self,
+        snapshot_pivot_rx: &mut std::sync::mpsc::Receiver<SnapshotControlMessage>,
+    ) -> Result<(), StoreError> {
+        let cf_misc = self.cf_handle(CF_MISC_VALUES)?;
+        let cf_snapshot = self.cf_handle(CF_SNAPSHOTS)?;
+
+        self.db
+            .delete_range_cf(&cf_snapshot, vec![], vec![0xff; 131])?;
```
pablodeymo: this delete should be done after snap sync, not before

crates/storage/store_db/rocksdb.rs
```diff
@@ -437,6 +467,111 @@ impl Store {
...
+            let last_written_account = last_written
+                .get(0..64)
+                .map(|v| Nibbles::from_hex(v.to_vec()))
+                .unwrap_or_default();
+            let mut last_written_storage = last_written
+                .get(66..130)
+                .map(|v| Nibbles::from_hex(v.to_vec()))
+                .unwrap_or_default();
+
+            println!("Starting snapshooting loop pivot={last_written:?} SR={state_root:x}");
```
pablodeymo: change to `debug!`

crates/storage/store_db/rocksdb.rs
```diff
@@ -440,14 +476,128 @@ impl Store {
...
+            let res = iter.try_for_each(|(path, node)| -> Result<(), StoreError> {
+                let Node::Leaf(node) = node else {
+                    return Ok(());
+                };
+                let account_state = AccountState::decode(&node.value)?;
+                let account_hash = H256::from_slice(&path.to_bytes());
+                batch.put_cf(&cf_misc, "last_written", path.as_ref());
+                batch.put_cf(&cf_snapshot, path.as_ref(), node.value);
+                ctr += 1;
+                if ctr > 10_000 {
```
edg-l, line 544: i think its possible to miss a db.write if the last loop doesnt land on a > 10_000, i.e ctr < 10_000 and we dont loop anymore, isnt the batch lost?, i dont see any other db.write outsdie a if ctr > 10_000 to handle that case

iovoid: Fixed in d692dba7

crates/common/trie/trie.rs
```diff
@@ -138,6 +151,10 @@ impl Trie {
         if !self.root.is_valid() {
             return Ok(None);
         }
+        if path.len() == 32 {
+            self.pending_removal.insert(Nibbles::from_bytes(path));
+        }
```
Oppen, line 156: This is not true for all tries (receipts, transactions and withdrawals use their index encoded as RLP as key), but also at this level `Trie` is just key-value. This should be unconditional.

crates/common/trie/trie.rs
```diff
@@ -98,6 +98,19 @@ impl Trie {
     pub fn get(&self, pathrlp: &PathRLP) -> Result<Option<ValueRLP>, TrieError> {
         let path = Nibbles::from_bytes(pathrlp);
 
+        if pathrlp.len() == 32
```
Oppen, line 101: Similarly, we shouldn't have to check length here.

crates/storage/store_db/rocksdb.rs
```diff
@@ -109,10 +109,22 @@ const CF_INVALID_ANCESTORS: &str = "invalid_ancestors";
 /// - [`Vec<u8>`] = `BlockHeaderRLP::from(block.header.clone()).bytes().clone()`
 const CF_FULLSYNC_HEADERS: &str = "fullsync_headers";
 
-#[derive(Debug)]
+pub const CF_FLATKEYVALUE: &str = "flatkeyvalue";
+
+pub const CF_MISC_VALUES: &str = "misc_values";
```
Oppen, line 114: No need for pub.

crates/storage/store_db/rocksdb.rs
```diff
@@ -448,14 +485,133 @@ impl Store {
...
+            if last_written_account > Nibbles::default() {
+                iter.advance(last_written_account.to_bytes())?;
+            }
+            let res = iter.try_for_each(|(path, node)| -> Result<(), StoreError> {
+                let Node::Leaf(node) = node else {
+                    return Ok(());
+                };
+                let account_state = AccountState::decode(&node.value)?;
+                let account_hash = H256::from_slice(&path.to_bytes());
+                batch.put_cf(&cf_misc, "last_written", path.as_ref());
```
Oppen, line 541: We may save some memory and some overhead by miving the `last_written` updates to inside the `if ctr > 10_000` cases.

crates/storage/store_db/rocksdb.rs
```diff
@@ -569,8 +746,14 @@ impl StoreEngine for Store {
             }
 
             // Single write operation
-            db.write(batch)
-                .map_err(|e| StoreError::Custom(format!("RocksDB batch write error: {}", e)))
+            let ret = db
+                .write(batch)
+                .map_err(|e| StoreError::Custom(format!("RocksDB batch write error: {}", e)));
```
Oppen, line 751: Missing comment explaining we need to unconditionally run the lines below, and that they are order sensitive.

crates/common/trie/node.rs
```diff
@@ -70,6 +70,9 @@ impl NodeRef {
                     Node::Leaf(_) => {}
                 }
                 let hash = *hash.get_or_init(|| node.compute_hash());
+                if let Node::Leaf(leaf) = node.as_ref() {
+                    acc.push((path.concat(&leaf.partial), leaf.value.clone()));
+                }
```
Oppen, line 75: We should have a comment here.

Oppen (approved):
There are a few minor comments. I'll make an issue with them. Otherwise it should be good to go.

PR #4893, opened 2025-10-15
fix(l1,l2): add feature flag to decide precompute number

tomip01:
**Motivation**

Risc0 is currently failing because precompute must be set to 0.

This changes were made by @gianbelinche

**Follow up**

Opened [this](https://github.com/lambdaclass/ethrex/issues/4905) issue to update the code once the change becomes available.


crates/common/crypto/Cargo.toml
```diff
@@ -18,3 +18,4 @@ thiserror.workspace = true
 [features]
 default = []
 c-kzg = ["dep:c-kzg"]
+risc0 = []
```
ilitteri: Let's include `c-kzg` in `risc0`.

tomip01: Done [2c5b849](https://github.com/lambdaclass/ethrex/pull/4893/commits/2c5b849fd8db10749f02e7cb046861fe33f8b399)

crates/l2/prover/src/guest_program/src/risc0/Cargo.toml
ilitteri, line 1: Don't we need to update RISC0's guest `Cargo.lock`?

tomip01: No, there is no need. I ran `cargo check` inside risc0 guest program and nothing change

PR #4896, opened 2025-10-16
feat(l1): add websocket rpc

ManuelBilbao:
**Motivation**

**Description**

Added an optional websocket server for JSON-RPC API. 3 flags are added to the CLI:
- `--ws`: enable websocket server, disabled by default.
- `--ws.addr`
- `--ws.port`

Note: WS specific methods (i.e., eth_subscribe and eth_unsubscribe) are still pending to be implemented.

**How to test:**

Start a dev node with `--ws`:
```
cargo run --release -- --dev --ws
```

With a websocket client (like [websocat](https://github.com/vi/websocat)) connect to the server and make requests:
```
websocat ws://localhost:8546
{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":83}
```

It should return something like
```
{"id":83,"jsonrpc":"2.0","result":"0x135"}
```

### DIFF
--- a/crates/networking/p2p/metrics.rs
+++ b/crates/networking/p2p/metrics.rs
@@ -175,10 +175,10 @@ impl fmt::Display for CurrentStepValue {
             CurrentStepValue::RequestingStorageRanges => write!(f, "Requesting Storage Ranges"),
             CurrentStepValue::DownloadingHeaders => write!(f, "Downloading Headers"),
             CurrentStepValue::InsertingStorageRanges => {
-                write!(f, "Inserting Storage Ranges - \x1b[31mWriting to DB\x1b[0m")
+                write!(f, "Inserting Storage Ranges - Writing to DB")
             }
             CurrentStepValue::InsertingAccountRanges => {
-                write!(f, "Inserting Account Ranges - \x1b[31mWriting to DB\x1b[0m")
+                write!(f, "Inserting Account Ranges - Writing to DB")
             }
             CurrentStepValue::InsertingAccountRangesNoDb => write!(f, "Inserting Account Ranges"),
         }
--- a/crates/networking/p2p/network.rs
+++ b/crates/networking/p2p/network.rs
@@ -352,13 +352,13 @@ pub async fn periodically_show_peer_stats_during_syncing(
                 "P2P Snap Sync:
 elapsed: {elapsed}
 {peer_number} peers.
-\x1b[93mCurrent step:\x1b[0m {current_step}
+Current step: {current_step}
 Current Header Hash: {current_header_hash:x}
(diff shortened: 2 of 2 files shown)


cmd/ethrex/cli.rs
```diff
@@ -135,6 +135,32 @@ pub struct Options {
         env = "ETHREX_HTTP_PORT"
     )]
     pub http_port: String,
+    #[arg(
+        long = "ws",
```
ilitteri: To be consistent with `p2p.enabled`
```suggestion
        long = "ws.enabled",
```

cmd/ethrex/cli.rs
```diff
@@ -135,6 +135,32 @@ pub struct Options {
...
+    pub ws_addr: String,
+    #[arg(
+        long = "ws.port",
+        default_value = "8546",
+        value_name = "PORT",
+        help = "Listening port for the websocket rpc server.",
+        help_heading = "RPC options",
+        env = "ETHREX_WS_PORT"
+    )]
+    pub ws_port: String,
```
ilitteri, line 165: Shouldn't these args require `ws`?

tomip01, line 165: Could we have only one cli arg like `ws_socket` to be `SocketAddr`?

cmd/ethrex/cli.rs
ManuelBilbao: Hm, I looked `--metrics` as reference. Geth uses `--ws` too.
I agree in rather change `--p2p.enabled` or `--metrics` to keep consistency

cmd/ethrex/cli.rs
ManuelBilbao: I followed `http` and `authrpc` convention. Think this is simpler for user

ManuelBilbao: Absolutely. [6985003](https://github.com/lambdaclass/ethrex/pull/4896/commits/698500393b67800d07277d530e24beaf85ab70bf)

cmd/ethrex/cli.rs
ilitteri: I'd go for updating `--metrics`, I'll file an issue for doing it in another PR.

ilitteri: https://github.com/lambdaclass/ethrex/issues/4902

ManuelBilbao: Done [3d10202](https://github.com/lambdaclass/ethrex/pull/4896/commits/3d10202791d73bccf4f57df61636d7f5742e153d)

tomip01 (approved):

PR #4900, opened 2025-10-16
chore(l1): load c-kzg trusted setup when initializing the node

JereSalo:
**Motivation**

- First time point evaluation precompile was called inside LEVM it took a long time to execute the block. This was because of the loading of the trusted setup. By doing it when initializing the node we avoid this.

**Description**

- Add function to warm up trusted setup and use it in the client initialization

Closes #issue_number

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3515,6 +3515,7 @@ dependencies = [
  "ethrex-blockchain",
  "ethrex-common 3.0.0",
  "ethrex-config",
+ "ethrex-crypto",
  "ethrex-dev",
  "ethrex-l2",
  "ethrex-l2-common",
--- a/cmd/ethrex/Cargo.toml
+++ b/cmd/ethrex/Cargo.toml
@@ -13,6 +13,7 @@ ethrex-config.workspace = true
 ethrex-blockchain.workspace = true
 ethrex-rpc.workspace = true
 ethrex-common.workspace = true
+ethrex-crypto.workspace = true
 ethrex-p2p.workspace = true
 ethrex-storage.workspace = true
 ethrex-vm.workspace = true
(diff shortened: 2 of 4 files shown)


edg-l:
The function called calls this, i think this the correct thing to do.

Just one questiion, is only 8 needed? where does it come from? We could also preinit 1-15


```rust
pub fn ethereum_kzg_settings(precompute: u64) -> &'static KzgSettings {
    ethereum_kzg_settings_inner(precompute).as_ref()
}

/// Returns default Ethereum mainnet KZG settings as an `Arc`.
///
/// It is useful for sharing the settings in multiple places.
///
/// Note: Precompute values 0-15 (inclusive) are supported.
pub fn ethereum_kzg_settings_arc(precompute: u64) -> Arc<KzgSettings> {
    ethereum_kzg_settings_inner(precompute).clone()
}

fn ethereum_kzg_settings_inner(precompute: u64) -> &'static Arc<KzgSettings> {
    let cache_box = match precompute {
        0 => &CACHE_0,
        1 => &CACHE_1,
        2 => &CACHE_2,
        3 => &CACHE_3,
        4 => &CACHE_4,
        5 => &CACHE_5,
        6 => &CACHE_6,
        7 => &CACHE_7,
        8 => &CACHE_8,
        9 => &CACHE_9,
        10 => &CACHE_10,
        11 => &CACHE_11,
        12 => &CACHE_12,
        13 => &CACHE_13,
        14 => &CACHE_14,
        15 => &CACHE_15,
        _ => panic!(
            "Unsupported precompute value: {precompute}. Only values 0-15 (inclusive) are supported."
        ),
    };

    cache_box.get_or_init(|| {
        let settings = KzgSettings::load_trusted_setup(
            ETH_G1_MONOMIAL_POINTS,
            ETH_G1_LAGRANGE_POINTS,
            ETH_G2_MONOMIAL_POINTS,
            precompute,
        )
        .expect("failed to load default trusted setup");
        Box::new(Arc::new(settings))
    })
}
```

edg-l (approved):

ilitteri (approved):

JereSalo:
@edg-l In the docs they suggest using 8 or 9. We've been using 8 in the client so I kept it as is because it made sense.
The methods they mention here are called inside other methods that we use within the client, like `compute_cells`.
image

PR #4904, opened 2025-10-16
fix(l1,l2): add replay feature to trie crate in order to make replay work with `eth_getProof`

JereSalo:
**Motivation**

- Fix Stateless execution of some blocks in replay. Attached to [this PR](https://github.com/lambdaclass/ethrex-replay/pull/33) in replay.

**Description**

- A modification to the trie is necessary if we want to run some blocks with eth_getProof. I tried not to modify ethrex code for achieving this but I have no choice other than doing this, I don't think the cost of making this change is high at all. Previously we executed these kind of blocks that failed with stateless execution in a non-stateless manner but that required changing even more code inside ethrex, it's just that it went unnoticed because it didn't require a feature flag. IMHO this is the best patch possible we can make, I don't love it but I don't seem to find a better alternative

These changes are the same as the ones made in #2952, this one was closed because we started using the Execution Witness but now in replay we have support for both methods: `eth_getProof` and `debug_executionWitness`

### DIFF
--- a/.github/workflows/pr-main_replay.yaml
+++ b/.github/workflows/pr-main_replay.yaml
@@ -3,7 +3,7 @@ permissions:
   contents: read
 
 on:
-  push:
+  pull_request:
     paths:
       - crates/**
       - ".github/workflows/pr-main_replay.yaml"


JereSalo:
We don't want to add `replay` feature, closing PR.

Issue #4869, opened 2025-10-14
Check Store API handling of unavailable states


MegaRedHand:
With the move to path-based state storage, we had a problem in the RPC module, where fetching old state data would result in incorrect data. This was automatically handled in the hash-based state storage implementation, since there were no key collisions. Now that we _can_ have key collisions, we'll need to explicitly handle these states.

We already patched the issue in the path-based PR in b42ff1d30576ed324143f26647f2aee302d4c9d4, but we should clean up the API and check that this is correctly handled everywhere. A better way to handle these cases could be to return a `None` in those cases, defaulting to sane values when we know an account or storage doesn't exist.

MegaRedHand:
Draft PR related to this: #4920

PR #4920, opened 2025-10-17
fix(l1): return an error when state is not available

MegaRedHand:
**Motivation**

We don't currently check if a state is available when accessing state.

**Description**

This PR adds additional runtime checks to return an error from the `Store` object whenever we try to access unavailable state.

Related to #4869


MegaRedHand:
Closing due to the rework of the Store API. We still need to work on #4869

PR #4922, opened 2025-10-17
fix(l1,l2): rename CLI arg `--metrics` to `--metrics.enabled` for naming consistency

### LINKED ISSUE #4902: Rename arg `--metrics` to `--metrics.enabled`

FredPhilipy:
Renames the `--metrics` command line argument to `--metrics.enabled` to maintain consistent dot-notation formatting with other metrics-related parameters (`--metrics.addr` and `--metrics.port`).

Closes #4902

### DIFF
--- a/tooling/sync/Makefile
+++ b/tooling/sync/Makefile
@@ -165,7 +165,7 @@ start-lighthouse: ## Start lighthouse for the network given by NETWORK.
 		--http \
 		--http-address 0.0.0.0 \
 		--http-allow-origin "*" \
-		--metrics.enabled \
+		--metrics \
 		--metrics-address 0.0.0.0 \
 		--metrics-port 5054 \
 		--datadir $(DATA_PATH)/${NETWORK}_data/lighthouse_${NODE_NAME}_$(EVM) \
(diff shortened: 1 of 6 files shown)


tooling/sync/Makefile
```diff
@@ -165,7 +165,7 @@ start-lighthouse: ## Start lighthouse for the network given by NETWORK.
 		--http \
 		--http-address 0.0.0.0 \
 		--http-allow-origin "*" \
-		--metrics \
+		--metrics.enabled \
```
ElFantasma: Bug: this `--metrics` flag belongs to the `lighthouse bn` command (see `start-lighthouse` target above), **not** ethrex. Lighthouse uses `--metrics` (with dashes: `--metrics-address`, `--metrics-port`) and doesn't accept `--metrics.enabled`. This change would break the Lighthouse invocation.

This line should remain `--metrics \`.

ElFantasma (changes requested):

tooling/sync/Makefile
FredPhilipy: Good catch @ElFantasma! — that --metrics flag is for the lighthouse bn command, not ethrex. The rename PR (--metrics → --metrics.enabled) shouldn't have touched that line. Fixed, thanks!

ElFantasma (approved):
LGTM

MegaRedHand:

PR #4926, opened 2025-10-17
chore(l1): add `block_number` and `network` labels to block metrics 

rodrigo-o:
**Motivation**

We want to have bettwr dashboards that match between them at the block level por a particular network.

**Description**

This PR adds two labels to all block-related metrics to be able to make the needed changes in grafana

Closes #issue_number


rodrigo-o:
This was a quick experiment and didn't work as expected, having labels that changed quickly is not useful.

PR #4928, opened 2025-10-17
chore(l2): optimize node RLP decoding

xqft:
**Motivation**

To minimize cloning and reallocations in node RLP decoding.

still wip, current problem to solve is the remove() call in get_embedded_nodes() for branch childs. some childs can have multiple parents and after removing once then for the next parent the node will not be present. the execution yields a inconsistent trie error.

**Description**

- Implements a new `OwnedDecoder` that moves the encoded bytes instead of cloning slices.

### DIFF
--- a/crates/common/trie/node.rs
+++ b/crates/common/trie/node.rs
@@ -4,10 +4,12 @@ mod leaf;
 
 use std::{
     array,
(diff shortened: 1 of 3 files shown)


xqft:
superseded by #5224

PR #4940, opened 2025-10-20
ci(l1): increase timeout on daily hive jobs.

mpaulucci:
**Motivation**
Jobs are taking more than the default 6hs.

**Description**
- Increase the timeout to 20hs
- Proceed with the report even if some tests didn't succeed

### DIFF
--- a/.github/workflows/common_hive_reports.yaml
+++ b/.github/workflows/common_hive_reports.yaml
@@ -141,12 +141,12 @@ jobs:
         uses: ./.github/actions/setup-rust
 
       - name: Download all results
+        continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/common_hive_reports.yaml
```diff
@@ -142,6 +146,7 @@ jobs:
           path: hive/workspace/logs
           pattern: "*_${{ inputs.job_type }}-results.zip"
           merge-multiple: true
+          if-no-artifact-found: ignore
```
MegaRedHand: This parameter doesn't exist

PR #4958, opened 2025-10-20
Fee token

tomip01:
**Motivation**

**Description**

Closes #issue_number


tomip01:
A different approach will be used

PR #4961, opened 2025-10-20
perf(levm): store valid jump targets with code

### LINKED ISSUE #4951: Jumpdest pre-processing on create

jrchatruc:
**Motivation**

Jumps are noticeably slow compared to most of the VM. This is due to lazy computation of invalid jump destinations.

**Description**

Instead of lazy computation of blocklist, do greedy computation of allowlist and store the result, fetch it with the DB.

Closes #4951 

Comparison of gigagas/s in hoodi between one node synced on this branch (orange) and three others on main:

Screenshot 2025-10-20 at 4 46 20 PM

# Missing:
- Some Revm compatibility fails to compile;
- There's some unneeded `unsafe` code, it can be converted to a simple copy, there's no evidence that overhead is problematic;
- The type needs documenting, maybe make the fields private so most users can only create through `from_bytecode`;
- A shortcut that receives the hash may be useful as well, to reuse some hashes in snap sync.
- Check there are no redundant computations as it was done in a hurry, places that seemed irrelevant to execution just compute the needed data in place;
- Tests and benchmarks need updating.

### DIFF
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -23,7 +23,7 @@ use crate::{
 pub struct Code {
     pub hash: H256,
     pub bytecode: Bytes,
-    // TODO: Consider using Arc<u16> (needs to enable serde rc feature)
+    // TODO: Consider using Arc<[u16]> (needs to enable serde rc feature)
     pub jump_targets: Vec<u16>,
 }
 
@@ -44,6 +44,7 @@ impl Code {
         let mut targets = Vec::new();
         let mut i = 0;
         while i < code.len() {
+            // TODO: we don't use the constants from the vm module to avoid a circular dependency
             match code[i] {
                 // OP_JUMPDEST
                 0x5B => {
@@ -51,6 +52,7 @@ impl Code {
                 }
                 // OP_PUSH1..32
                 c @ 0x60..0x80 => {
+                    // OP_PUSH0
                     i += (c - 0x5F) as usize;
                 }
                 _ => (),


crates/common/types/account.rs
```diff
@@ -19,11 +19,49 @@ use crate::{
     utils::keccak,
 };
 
-#[allow(unused)]
+#[derive(Clone, Default, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
Oppen: `Default` probably needs a manual implementation that sets `bytecode` and `jump_targets` to empty but `hash` to `EMPTY_KECCAK_HASH`.

Oppen: We probably also need `rkyv` implementations for efficient proving, right now it needs to recompute the jumpdests inside.

crates/common/types/account.rs
```diff
@@ -19,11 +19,49 @@ use crate::{
...
+        let mut targets = Vec::new();
+        let mut i = 0;
+        while i < code.len() {
+            match code[i] {
+                0x5B => {
+                    targets.push(i as u16);
+                }
+                c @ 0x60..0x80 => {
+                    i += (c - 0x5F) as usize;
+                }
```
Oppen, line 57: We should probably use the constants in `opcodes.rs`, it was just convenient for me to use the numbers for quick iteration.

crates/common/types/block_execution_witness.rs
```diff
@@ -151,7 +151,10 @@ impl TryFrom<ExecutionWitness> for GuestProgramState {
         let codes_hashed = value
             .codes
```
Oppen, line 153: `codes` here probably needs to be `Vec<Code>`, rather than recomputing here. This requires `rkyv` implementation.

crates/common/types/account.rs
```diff
@@ -19,11 +19,49 @@ use crate::{
     utils::keccak,
 };
 
-#[allow(unused)]
+#[derive(Clone, Default, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Code {
+    pub hash: H256,
+    pub bytecode: Bytes,
+    pub jump_targets: Vec<u16>,
```
Oppen, line 27: It may or may not be worth it to make this an `Arc<[u16]>` instead. That would make cloning much cheaper, and this is immutable once computed so it should be correct.

crates/common/types/account.rs
Oppen: Note however that would act like a hint, so it's probably correct that it's computed inside so it can be proved.

crates/storage/store_db/rocksdb.rs
```diff
@@ -740,9 +743,17 @@ impl StoreEngine for Store {
...
+                let mut buf =
+                    Vec::with_capacity(6 + code.bytecode.len() + 2 * code.jump_targets.len());
+                code.bytecode.encode(&mut buf);
+                unsafe {
+                    std::slice::from_raw_parts(
+                        code.jump_targets.as_ptr().cast(),
+                        2 * code.jump_targets.len(),
+                    )
+                }
+                .encode(&mut buf);
```
Oppen: No real need for unsafe. Just serialize the correct size header and iterate appending the `to_le_bytes`.

Oppen:
```suggestion
                ethrex_rlp::encode::encode_length(2 * code.jump_targets.len(), &mut buf);
                buf.extend(code.jump_targets.into_iter().flat_map(|t| t.to_le_bytes()));
```

crates/storage/store_db/rocksdb.rs
```diff
@@ -1108,11 +1119,18 @@ impl StoreEngine for Store {
...
+        let hash_key = code.hash.0.to_vec();
+        let mut buf = Vec::with_capacity(6 + code.bytecode.len() + 2 * code.jump_targets.len());
+        code.bytecode.encode(&mut buf);
+        unsafe {
+            std::slice::from_raw_parts(
+                code.jump_targets.as_ptr().cast(),
+                2 * code.jump_targets.len(),
+            )
+        }
+        .encode(&mut buf);
```
Oppen: Ditto.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1658,14 +1690,22 @@ impl StoreEngine for Store {
...
-            batch_ops.push((CF_ACCOUNT_CODES.to_string(), key, value));
+            let mut buf = Vec::with_capacity(6 + code.bytecode.len() + 2 * code.jump_targets.len());
+            code.bytecode.encode(&mut buf);
+            unsafe {
+                std::slice::from_raw_parts(
+                    code.jump_targets.as_ptr().cast(),
+                    2 * code.jump_targets.len(),
+                )
+            }
+            .encode(&mut buf);
```
Oppen: Ditto.

crates/common/types/account.rs
Oppen: Implementing `AsRef<Bytes>` or `AsRef<[u8]>` might also be convenient to avoid many instances of `bytecode.bytecode`.

crates/common/types/account.rs
edg-l: added a comment for the future

crates/common/types/block_execution_witness.rs
edg-l: added a todo comment for the future

crates/common/trie/trie.rs
```diff
@@ -34,10 +34,9 @@ use lazy_static::lazy_static;
 lazy_static! {
     // Hash value for an empty trie, equal to keccak(RLP_NULL)
     pub static ref EMPTY_TRIE_HASH: H256 = H256::from_slice(
-        Keccak256::new()
+        &Keccak256::new()
```
edg-l: This and other changes like this are to silence a deprecated notice from sha

edg-l:
The size diff appears large due to a cargo.lock update in the tooling/bench code adding +3,819 lines, the pr is not big

crates/common/constants.rs
```diff
@@ -35,14 +35,8 @@ pub static EMPTY_KECCACK_HASH: LazyLock<H256> = LazyLock::new(|| {
...
-pub static EMPTY_TRIE_HASH: LazyLock<H256> = LazyLock::new(|| {
-    H256::from_slice(
-        Keccak256::new()
-            .chain_update([RLP_NULL])
-            .finalize()
-            .as_slice(),
-    )
-});
+pub static EMPTY_TRIE_HASH: LazyLock<H256> =
+    LazyLock::new(|| H256::from_slice(&Keccak256::new().chain_update([RLP_NULL]).finalize()));
```
Oppen:
```suggestion
    LazyLock::new(|| H256::from_slice(&Keccak256::digest([RLP_NULL])));
```

crates/common/trie/node_hash.rs
```diff
@@ -51,12 +51,9 @@ impl NodeHash {
...
     pub fn finalize(self) -> H256 {
         match self {
-            NodeHash::Inline(_) => H256::from_slice(
-                Keccak256::new()
-                    .chain_update(self.as_ref())
-                    .finalize()
-                    .as_slice(),
-            ),
+            NodeHash::Inline(_) => {
+                H256::from_slice(&Keccak256::new().chain_update(self.as_ref()).finalize())
```
Oppen: Ditto

crates/common/trie/trie.rs
Oppen: Ditto

ilitteri:
This PR slightly improves zkVM performance for both mainnet blocks' execution and proving times. The following results were measured using SP1 as the zkVM

### Execution Times
Average speedup: **1.059x**  
- `main` (`0e63e98a`): Min 15.591s (Block 23626696), Max 64.000s (Block 23626693), Avg 39.408s  
- `perf/store_jump_targets`: Min 15.216s (Block 23626696), Max 61.000s (Block 23626693), Avg 37.397s  
- Speedup range: 1.024x–1.108x  

| Block    | Gas      | #Txs | `main` (`0e63e98a`) (s) | `perf/store_jump_targets` (s) | Speedup |
|----------|----------|------|----------|---------------|---------|
| 23626690 | 27.405.831 | 288  | 56.171  | 50.688       | 1.108x |
| 23626691 | 28.264.607 | 266  | 40.272  | 37.770       | 1.066x |
| 23626692 | 16.213.974 | 97   | 20.098  | 18.532       | 1.085x |
| 23626693 | 44.243.915 | 390  | 64.000  | 61.000       | 1.049x |
| 23626694 | 19.366.588 | 225  | 29.374  | 27.928       | 1.052x |
| 23626695 | 34.562.703 | 334  | 53.333  | 51.001       | 1.046x |
| 23626696 | 10.379.565 | 155  | 15.591  | 15.216       | 1.024x |
| 23626697 | 33.079.221 | 333  | 55.857  | 53.144       | 1.051x |
| 23626698 | 11.826.216 | 136  | 19.020  | 18.076       | 1.052x |
| 23626699 | 11.616.595 | 203  | 20.094  | 19.237       | 1.045x |
| 23626700 | 44.895.521 | 479  | 59.677  | 56.383       | 1.058x |

### Proving Times
Average speedup: **1.074x**  
- `main` (`0e63e98a`): Min 392.000s (Block 23626696), Max 1750.000s (Block 23626693), Avg 1036.364s  
- `perf/store_jump_targets`: Min 378.000s (Block 23626696), Max 1620.000s (Block 23626693), Avg 964.000s  
- Speedup range: 1.037x–1.111x  

| Block    | Gas      | #Txs  | `main` (`0e63e98a`) (s) | `perf/store_jump_targets` (s) |  Speedup |
|----------|----------|------|----------|---------------|---------|
| 23626690 | 27.405.831 | 288  | 1074.000| 1002.000     | 1.072x |
| 23626691 | 28.264.607 | 266  | 1127.000| 1047.000     | 1.077x |
| 23626692 | 16.213.974 | 97   | 639.000 | 575.000      | 1.111x |
| 23626693 | 44.243.915 | 390  | 1750.000| 1620.000     | 1.080x |
| 23626694 | 19.366.588 | 225  | 762.000 | 707.000      | 1.077x |
| 23626695 | 34.562.703 | 334  | 1393.000| 1306.000     | 1.067x |
| 23626696 | 10.379.565 | 155  | 392.000 | 378.000      | 1.037x |
| 23626697 | 33.079.221 | 333  | 1653.000| 1560.000     | 1.060x |
| 23626698 | 11.826.216 | 136  | 556.000 | 522.000      | 1.065x |
| 23626699 | 11.616.595 | 203  | 496.000 | 461.000      | 1.076x |
| 23626700 | 44.895.521 | 479  | 1558.000| 1438.000     | 1.084x |

.github/workflows/pr_perf_blocks_exec.yaml
```diff
@@ -77,9 +77,12 @@ jobs:
...
           BINS="base,head"
-          hyperfine --setup "./bin/ethrex-base removedb" -w 5 -N -r 10 --show-output --export-markdown "bench_pr_comparison.md" \
-          -L bin "$BINS" -n "{bin}" \
-          "./bin/ethrex-{bin} --network fixtures/genesis/perf-ci.json --force import ./fixtures/blockchain/l2-1k-erc20.rlp --removedb"
+          # Run hyperfine benchmark, removing the database before each run.
+          # Pipe `yes` into the remove command to automatically confirm.
+          hyperfine -w 5 -N -r 10 --show-output --export-markdown "bench_pr_comparison.md" \
+            -L bin "$BINS" -n "{bin}" \
+            --prepare "yes | ./bin/ethrex-{bin} removedb" \
+            "./bin/ethrex-{bin} --network fixtures/genesis/perf-ci.json --force import ./fixtures/blockchain/l2-1k-erc20.rlp"
```
MegaRedHand: The command was using removedb as a setup for each batch of runs, but the command was cancelled due to having no input. I changed it to use `--prepare` instead, which runs before each single run, and piped `yes` into it to automatically confirm.

MegaRedHand: This was reverted, but we should fix this if we intend to keep this benchmark.

crates/l2/prover/src/guest_program/src/execution.rs
```diff
@@ -274,7 +280,11 @@ fn execute_stateless(
     #[cfg(feature = "l2")]
     let nodes_hashed = guest_program_state.nodes_hashed.clone();
     #[cfg(feature = "l2")]
-    let codes_hashed = guest_program_state.codes_hashed.clone();
+    let codes_hashed = guest_program_state
+        .codes_hashed
+        .iter()
+        .map(|(h, c)| (*h, c.bytecode.to_vec()))
+        .collect();
```
MegaRedHand, line 292: Many of these conversions seem unneeded. We should check that they don't affect the performance of the L2.

crates/storage/store_db/rocksdb.rs
```diff
@@ -738,9 +741,16 @@ impl StoreEngine for Store {
             }
 
             for (code_hash, code) in update_batch.code_updates {
-                let code_key = code_hash.as_bytes();
-                let code_value = AccountCodeRLP::from(code).bytes().clone();
-                batch.put_cf(&cf_codes, code_key, code_value);
+                let mut buf =
+                    Vec::with_capacity(6 + code.bytecode.len() + 2 * code.jump_targets.len());
```
MegaRedHand, line 745: This needs a comment

crates/storage/store_db/rocksdb.rs
```diff
@@ -1106,11 +1116,17 @@ impl StoreEngine for Store {
...
+        let hash_key = code.hash.0.to_vec();
+        let mut buf = Vec::with_capacity(6 + code.bytecode.len() + 2 * code.jump_targets.len());
+        code.bytecode.encode(&mut buf);
+        code.jump_targets
+            .into_iter()
+            .flat_map(|t| t.to_le_bytes())
+            .collect::<Vec<u8>>()
+            .as_slice()
+            .encode(&mut buf);
+        self.write_async(CF_ACCOUNT_CODES, hash_key, buf).await
```
MegaRedHand, line 1129: We need to update the documentation of `CF_ACCOUNT_CODES`

crates/storage/store_db/rocksdb.rs
MegaRedHand: Also, can't we implement `RLPEncode` for `Code` instead? Or put this in a helper.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1135,12 +1151,26 @@ impl StoreEngine for Store {
...
+    fn get_account_code(&self, code_hash: H256) -> Result<Option<Code>, StoreError> {
         let hash_key = code_hash.as_bytes().to_vec();
-        self.read_sync(CF_ACCOUNT_CODES, hash_key)?
-            .map(|bytes| AccountCodeRLP::from_bytes(bytes).to())
-            .transpose()
-            .map_err(StoreError::from)
+        let Some(bytes) = self.read_sync(CF_ACCOUNT_CODES, hash_key)? else {
+            return Ok(None);
+        };
+        let bytes = Bytes::from_owner(bytes);
```
MegaRedHand, line 1159: No need for this variable.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1135,12 +1151,26 @@ impl StoreEngine for Store {
...
     }
 
-    fn get_account_code(&self, code_hash: H256) -> Result<Option<Bytes>, StoreError> {
+    fn get_account_code(&self, code_hash: H256) -> Result<Option<Code>, StoreError> {
         let hash_key = code_hash.as_bytes().to_vec();
-        self.read_sync(CF_ACCOUNT_CODES, hash_key)?
-            .map(|bytes| AccountCodeRLP::from_bytes(bytes).to())
-            .transpose()
-            .map_err(StoreError::from)
+        let Some(bytes) = self.read_sync(CF_ACCOUNT_CODES, hash_key)? else {
```
MegaRedHand, line 1156: We could use the `get_pinned_cf` rocksdb method here to avoid copies before deserializing.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1135,12 +1151,26 @@ impl StoreEngine for Store {
...
+            return Err(StoreError::DecodeError);
+        }
+        let code = Code {
+            hash: code_hash,
+            bytecode: Bytes::copy_from_slice(bytecode),
+            jump_targets: targets
+                .chunks_exact(2)
+                .map(|c| u16::from_le_bytes([c[0], c[1]]))
+                .collect(),
+        };
```
MegaRedHand, line 1172: This could be an `RLPDecode` implementation.

crates/storage/store.rs
```diff
@@ -468,7 +467,7 @@ impl Store {
                     account_state.code_hash = info.code_hash;
                     // Store updated code in DB
                     if let Some(code) = &update.code {
-                        self.add_account_code(info.code_hash, code.clone()).await?;
+                        self.add_account_code(code.clone()).await?;
```
MegaRedHand, line 470: Not introduced by this PR, but this clone isn't needed in the underlying `Store` implementation, since it only needs a reference for serialization.

crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -245,15 +246,23 @@ impl<'a> VM<'a> {
...
             let end = code_offset.wrapping_add(final_size);
 
             #[expect(unsafe_code, reason = "bounds checked beforehand")]
             unsafe {
-                data.get_unchecked_mut(..final_size)
-                    .copy_from_slice(current_call_frame.bytecode.get_unchecked(code_offset..end));
+                data.get_unchecked_mut(..final_size).copy_from_slice(
+                    current_call_frame
+                        .bytecode
+                        .bytecode
```
MegaRedHand, line 263: This looks kind of weird

crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -277,7 +286,8 @@ impl<'a> VM<'a> {
     pub fn op_extcodesize(&mut self) -> Result<OpcodeResult, VMError> {
         let address = word_to_address(self.current_call_frame.stack.pop1()?);
         let address_was_cold = !self.substate.add_accessed_address(address);
-        let account_code_length = self.db.get_account_code(address)?.len().into();
+        // FIXME: a bit wasteful to fetch the whole code just to get the length.
```
MegaRedHand, line 289: The only use of this opcode I've seen is to check if an account is an EOA (post-7702). If it's an EOA, then `SENDER` and `CALLER` are the same, and `EXTCODESIZE` is `23`, but later we also need to fetch the code anyways, so any cache would negate the cost of this use.

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -292,13 +291,12 @@ impl<'a> VM<'a> {
...
-            // It's a constant, therefore the conversion cannot fail.
-            *value == Opcode::JUMPDEST as u8
-                && !call_frame.jump_target_filter.is_blacklisted(jump_address)
-        })
+    fn target_address_is_valid(call_frame: &mut CallFrame, jump_address: u16) -> bool {
+        call_frame
+            .bytecode
+            .jump_targets
+            .binary_search(&jump_address)
+            .is_ok()
```
MegaRedHand, line 299: I think this should go in `Code`

crates/common/types/account.rs
```diff
@@ -19,11 +19,60 @@ use crate::{
     utils::keccak,
 };
 
-#[allow(unused)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Code {
+    pub hash: H256,
+    pub bytecode: Bytes,
+    // TODO: Consider using Arc<[u16]> (needs to enable serde rc feature)
+    pub jump_targets: Vec<u16>,
```
MegaRedHand, line 27: This needs a comment explaining the reason for using `u16` instead of bigger numbers. Linking to https://eips.ethereum.org/EIPS/eip-170 and https://eips.ethereum.org/EIPS/eip-7907. Also, EIP-7907 might break this since it bumps initcode max size to 96Kb.

crates/common/types/account.rs
```diff
@@ -19,11 +19,60 @@ use crate::{
...
+        let jump_targets = Self::compute_jump_targets(&code);
+        Self {
+            hash: keccak(code.as_ref()),
+            bytecode: code,
+            jump_targets,
+        }
+    }
+
+    fn compute_jump_targets(code: &[u8]) -> Vec<u16> {
+        debug_assert!(code.len() <= u16::MAX as usize);
```
MegaRedHand, line 43: This needs a comment

MegaRedHand (approved):
LGTM. Let's refactor and add comments in a following PR

PR #4963, opened 2025-10-20
fix(l1): validate incoming pooled tx blobs

### LINKED ISSUE #3745: Disconnect peers whenever they send invalid blob transactions in `PooledTransactions`
Currently, we are just logging the error. The response Hive expects is for the peer to disconnect:

https://github.com/ethereum/go-ethereum/blob/85ae3e16f1a3456c158ba0109dad07502c4046e9/cmd/devp2p/internal/ethtest/suite.go#L970

fmoletta:
**Motivation**
Enable hive p2p tests `TestBlobTxWithMismatchedSidecar ` & `TestBlobTxWithoutSidecar`
In order to pass these tests we have to:
* Be able to decode pooled transactions without blobs (aka plain eip4844 transactions instead of a wrapped eip4844 with its blobsbundle)
* Disconnect from peers that send transactions without blobs or blob bundles that don't match the versioned hashes

**Description**
* Handle the case of plain `Eip4844` transaction when RLP-decoding `WrappedEip4844` transactions
* Disconnect from peers that send empty/mismatched blobs

Closes #3745, part of #4941


crates/common/types/blobs_bundle.rs
```diff
@@ -169,6 +173,23 @@ impl BlobsBundle {
...
+        for (commitment, blob_versioned_hash) in
+            self.commitments.iter().zip(blob_versioned_hashes.iter())
+        {
+            if *blob_versioned_hash != kzg_commitment_to_versioned_hash(commitment) {
+                return Err(BlobsBundleError::BlobVersionedHashesError);
+            }
+        }
+        Ok(())
+    }
 }
```
ElFantasma, line 203: Another version maybe:
```suggestion
    pub fn validate_blob_commitment_hashes(&self, hashes: &[H256]) -> Result<(), BlobsBundleError> {
        if self.commitments.len() != hashes.len()
            || self
                .commitments
                .iter()
                .zip(hashes.iter())
                .any(|(commitment, hash)| *hash != kzg_commitment_to_versioned_hash(commitment))
        {
            Err(BlobsBundleError::BlobVersionedHashesError)
        } else {
            Ok(())
        }
    }
}
```

PR #4970, opened 2025-10-20
fix(l1): remove duplicate span.entered() call in apply_updates

MegaRedHand:
**Motivation**

Some of the numbers we get from grafana seem to be weird. This might have been caused by a duplicate span that was created in `apply_updates`. I suspect it was from a git conflict that was solved erroneously.

**Description**

This PR removes a duplicate span `entered()` call.

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -659,7 +659,6 @@ impl StoreEngine for Store {
                 ],
             )?;
 
-            let _span = tracing::trace_span!("Block DB update").entered();
             let mut batch = WriteBatch::default();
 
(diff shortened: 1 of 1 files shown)


crates/storage/store_db/rocksdb.rs
```diff
@@ -659,7 +659,6 @@ impl StoreEngine for Store {
                 ],
             )?;
 
-            let _span = tracing::trace_span!("Block DB update").entered();
```
MegaRedHand, line 662: Note that before calling `open_cfs` there's another line like this.

pablodeymo (approved):

rodrigo-o (approved):

PR #4973, opened 2025-10-21
feat(l1)!: allow setting a target gas limit through CLI

### LINKED ISSUE #4482: Allow configuring a desired gas limit on the block builder
Related to #2526 (this one is for L1)

Currently, we have a hardcoded desired gas limit of 30M on the block builder:

https://github.com/lambdaclass/ethrex/blob/9d61e15af6be80d7a2610c9eeea4357306e1d57e/crates/blockchain/payload.rs#L182

We should instead have this as a configurable value somewhere.

MegaRedHand:
**Motivation**

When spinning up localnets made up only of ethrex nodes, we can't use gas limits apart from the default one of 30M. We want to be able to set it through a CLI flag in those cases.

**Description**

This PR adds a flag for setting the gas limit: `--builder.gas-limit`. It also renames the flag `--block-producer.extra-data` to `--builder.extra-data`.

Closes #4482


cmd/ethrex/cli.rs
```diff
@@ -230,6 +230,14 @@ pub struct Options {
         help_heading = "Block producer options"
     )]
     pub extra_data: String,
+    #[arg(
+        long = "block-producer.gas-limit",
```
mpaulucci: nit: maybe just `producer.gas-limit`? So that the namespace is not that long

cmd/ethrex/cli.rs
```diff
@@ -230,6 +230,14 @@ pub struct Options {
...
+        default_value_t = DEFAULT_BUILDER_GAS_CEIL,
```
mpaulucci, line 235: I don't understand this? Isn't the default gas ceil different depending on the fork? I understand if you want to overwrite it but, isnt the default different for each fork?

MegaRedHand: No. The default gas limit is set by each client independently of the network. For example, geth currently has a default of 60M:

https://github.com/ethereum/go-ethereum/blob/6608a2aafd3603afe59f95fa7b2a8ec00b8eaa19/miner/miner.go#L54-L55

I think we should bump this default to 60M, but I didn't change that in this PR since it's important enough to merit its own PR and discussion.

cmd/ethrex/cli.rs
MegaRedHand: I used `block-producer` since that was the namespace used for `block-producer.extra-data`. I think `builder` would be a better namespace name.

MegaRedHand: Changed in [510f969](https://github.com/lambdaclass/ethrex/pull/4973/commits/510f969d2f5ef0fc6628562f10f4b77d2d9477fa) to `builder`.

mpaulucci: nice, even better

cmd/ethrex/initializers.rs
```diff
@@ -130,7 +130,7 @@ pub fn init_blockchain(store: Store, blockchain_opts: BlockchainOptions) -> Arc<
     Blockchain::new(store, blockchain_opts).into()
 }
 
-#[allow(clippy::too_many_arguments)]
+#[expect(clippy::too_many_arguments)]
```
ilitteri, line 133: :agite:

PR #4985, opened 2025-10-21
fix(l1): don't trigger snap sync if fcu head is already canonical

### LINKED ISSUE #4846: [Snap Sync] Ethrex returns syncing when calling FCU on the genesis block
I've seen a lot of errors in Hive since we changed to use snap sync by default:

```
unexpected status on forkchoice updated to genesis: payload_status=PayloadStatus(status=<PayloadStatusEnum.SYNCING: 'SYNCING'>, latest_valid_hash=None, validation_error=None) payload_id=None
```

The tests starts by calling fork choice updated to the genesis block (block 0). By definition, you can't sync to the genesis block, so we should not return syncing, but we should return valid.

Handling the case where we call FCU to the genesis block would solve the issue. But going one step further, if we get a FCU call to any block which we have the state for, we should not return syncing, since we are already synced to that block, we just have to set it as head.

fmoletta:
**Motivation**
In hive tests we often receive an fcu with either the genesis or last imported block as head to notify the node that it is already synced. As this is not usual in a real-case scenario, we just trigger a snap sync to the advertised head. This PR solves this by first checking if the fcu head is already part of our canonic chain before triggering a sync

**Description**
* When handling a forkchoice in snap sync mode, check that the head is not already canonical before triggering a sync, if it is already canonical, apply the forkchoice

Closes #4846

### DIFF
--- a/crates/networking/rpc/engine/fork_choice.rs
+++ b/crates/networking/rpc/engine/fork_choice.rs
@@ -215,12 +215,13 @@ async fn handle_forkchoice(
         ));
     }
 
-    // Don't trigger a sync if the block is already canonical
     if context.syncer.sync_mode() == SyncMode::Snap {
+        // Don't trigger a sync if the block is already canonical
         if context
             .storage
             .is_canonical_sync(fork_choice_state.head_block_hash)?
         {
+            // Disable snapsync mode so we can process incoming payloads
             context.syncer.disable_snap();
         } else {
             context


crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -215,7 +215,12 @@ async fn handle_forkchoice(
         ));
     }
 
-    if context.syncer.sync_mode() == SyncMode::Snap {
+    // Don't trigger a sync if the block is already canonical
```
mpaulucci: In that case, will the sync mode be set as Full somewhere?

fmoletta: The next FCU to an unknown head will trigger a sync, the syncer will detect that the sync_head is only a few blocks away and switch to full sync mode

fmoletta: Update: changed to also disable snap sync so we can process incoming new payloads right away

mpaulucci (approved):

PR #4988, opened 2025-10-22
perf/bg_trie_update update

edg-l:
Currently the upstream branch has

`trie_cache: Arc<Mutex<Arc<TrieLayerCache>>>,,`

where TrieLayerCache has the field `layers: BTreeMap<H256, Arc<TrieLayer>>,` however from the changes it suggests the author wanted to use Arc<BTreeMap<H256, Arc<TrieLayer>>> here.

And the stores have `Arc<Mutex<Arc<TrieLayerCache>>>,`

There are too many Arc layers here, it made some arc atomic swaps complex, for example to update the trielayercache one would need lock trie_cache, clone the arc, then clone the trielayrcache, then  if the layers btreemap had a surrounding arc, another btreemap clone, get the trielayer, clone it, update the trie layer, put it in the cloned btree, update the layers btreemap, then update the trielayercache

Instead i made TrieLayerCache completely clone safe and the method have all the complexity of managing the arc swaps.

So users of the cache simply store a `trie_cache: Arc<TrieLayerCache>,` and call get, put_batch, get_commitable, etc.

If this design seems good, there are still things to do:

- Some tests dont pass, however i'm not entirely sure its due to the rocksdb implementation, since tests use the memorydb which doesnt use the new background mechanism.
-

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::type_complexity)]
+
 use crate::{
     rlp::AccountCodeHashRLP,
     trie_db::{
--- a/crates/storage/trie_db/layering.rs
+++ b/crates/storage/trie_db/layering.rs
@@ -124,7 +124,7 @@ impl TrieLayerCache {
             parent_nodes
                 .unwrap_or_default()
                 .into_iter()
-                .chain((&*layer).nodes.clone().into_iter())
+                .chain(layer.nodes.clone())
                 .collect(),
         )
     }


crates/storage/trie_db/layering.rs
```diff
@@ -18,14 +21,20 @@ struct TrieLayer {
 pub struct TrieLayerCache {
     /// Monotonically increasing ID for layers, starting at 1.
     /// TODO: this implementation panics on overflow
-    last_id: usize,
-    layers: BTreeMap<H256, Arc<TrieLayer>>,
+    last_id: Arc<AtomicUsize>,
+    layers: Arc<Mutex<BTreeMap<H256, Arc<TrieLayer>>>>,
```
edg-l, line 25: i think the mutex surrounding the btreemap is worth it (as is done now), since its only used to do the get and clone ASAP to unlock again. using an Arc<Mutex<Arc<Btree>>> would be too complex IMHO

crates/storage/trie_db/layering.rs
```diff
@@ -56,55 +71,68 @@ impl TrieLayerCache {
...
-                self.last_id += 1;
-                TrieLayer {
-                    nodes: Arc::new(BTreeMap::new()),
+                let last_id = self.last_id.fetch_add(1, Ordering::Relaxed);
+                Arc::new(TrieLayer {
+                    nodes: BTreeMap::new(),
                     parent,
-                    id: self.last_id,
-                }
+                    id: last_id + 1,
```
edg-l, line 92: this +1 is to maintain the same ids used before

crates/storage/store_db/rocksdb.rs
```diff
@@ -719,8 +710,6 @@ impl Store {
             }
         }
         db.write(batch)?;
-        // Phase 3: update diff layers with the removal of bottom layer.
-        *trie_cache.lock().map_err(|_| StoreError::LockError)? = Arc::new(trie_mut);
```
Oppen, line 723: This mutation needs to happen at the end. Otherwise you lose the bottom layer BEFORE there is a backing disk layer with the data visible to users..

crates/storage/trie_db/layering.rs
Oppen: This one forces the tree traversal to be done with the mutex taken. We want to avoid locking in the read path, that's why we would rather have a single outer lock and an `Arc` clone there.

crates/storage/trie_db/layering.rs
Oppen: These kind of comments belong in the code.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::type_complexity)]
```
Oppen, line 1: Don't use global allows.

PR #4989, opened 2025-10-22
perf(l1, l2): move post-execution trie updates to the background

edg-l:
**Motivation**

This PR moves the trie portion of the `apply_updates` function to a background thread, improving block execution times. This background thread receives messages through a channel to apply new trie updates and does two things:

- First, it updates the in-memory diff layers and notifies the process that sent the message (i.e. the block production thread) so it can continue with block execution.
- Second, it performs the logic of persisting the bottom-most diff layer to disk. To do this, we first pause the ongoing thread in charge of generating the snapshots (i.e. FlatKeyValue) through a message, then persist the diff layer to disk, then notify again the thread to continue snapshot generation.

**Description**

Closes #issue_number

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -1266,5 +1266,3 @@ pub async fn regenerate_head_state(
 
     Ok(())
 }
-
-// Test


jrchatruc (approved):

crates/storage/store_db/in_memory.rs
```diff
@@ -91,10 +91,7 @@ impl StoreEngine for Store {
 
         // Store trie updates
         {
```
MegaRedHand: I think we no longer need this scope

crates/storage/trie_db/layering.rs
```diff
@@ -69,26 +64,29 @@ impl TrieLayerCache {
...
-                }
-            })
-            .nodes
-            .extend(
-                key_values
-                    .into_iter()
-                    .map(|(path, node)| (path.into_vec(), node)),
-            );
+        if self.layers.contains_key(&state_root) {
+            return;
```
MegaRedHand, line 70: We should log a warning in this case, I think.

crates/storage/trie_db/layering.rs
```diff
@@ -69,26 +64,29 @@ impl TrieLayerCache {
...
+        }
+
+        let nodes = HashMap::from_iter(
+            key_values
+                .into_iter()
+                .map(|(path, node)| (path.into_vec(), node)),
+        );
```
MegaRedHand: This can be a `collect`.

crates/storage/store_db/rocksdb.rs
```diff
@@ -753,15 +868,14 @@ impl StoreEngine for Store {
...
-                // If the channel is closed, there's nobody to notify
-                let _ = flatkeyvalue_control_tx.send(FKVGeneratorControlMessage::Continue);
-            }
-            ret
+            // Wait for an updated top layer so every caller afterwards sees a consistent view.
+            // Specifically, the next block produced MUST see this upper layer.
+            wait_for_new_layer
+                .recv()
+                .map_err(|e| StoreError::Custom(format!("recv failed: {e}")))??;
+            // After top-level is addded, we can make the rest of the changes visible.
```
MegaRedHand:
```suggestion
            // After top-level is added, we can make the rest of the changes visible.
```

crates/storage/store_db/rocksdb.rs
```diff
@@ -631,81 +775,52 @@ impl StoreEngine for Store {
...
-                    .flat_map(|(account_hash, nodes)| {
-                        nodes
-                            .into_iter()
-                            .map(move |(path, node)| (apply_prefix(Some(account_hash), path), node))
-                    })
-                    .chain(update_batch.account_updates)
-                    .collect(),
+            let (account_updates, storage_updates) = (
+                take(&mut update_batch.account_updates),
+                take(&mut update_batch.storage_updates),
```
MegaRedHand: We should destructure the `update_batch` instead of using `std::mem::take`, which leaves the old variables hanging with invalid data.

crates/storage/store_db/rocksdb.rs
```diff
@@ -607,13 +671,93 @@ impl Store {
...
+            .collect();
+        // Read-Copy-Update the trie cache with a new layer.
+        let trie = trie_cache
+            .lock()
+            .map_err(|_| StoreError::LockError)?
+            .clone();
+        let mut trie_mut = (*trie).clone();
+        trie_mut.put_batch(parent_state_root, child_state_root, new_layer);
+        *trie_cache.lock().map_err(|_| StoreError::LockError)? = Arc::new(trie_mut);
+        // Update finished, signal block production.
```
MegaRedHand:
```suggestion
        // Update finished, signal block processing.
```

crates/storage/store_db/rocksdb.rs
```diff
@@ -607,13 +671,93 @@ impl Store {
...
+        notify.send(Ok(())).map_err(|_| StoreError::LockError)?;
+
+        // Phase 2: update disk layer.
+        let trie = trie_cache
+            .lock()
+            .map_err(|_| StoreError::LockError)?
+            .clone();
```
MegaRedHand: I think here we can clone the `Arc::new(trie_mut)` in line 704 before storing it in the cache, which would avoid this lock here.

edg-l:
Fixed conflicts and review comments

MegaRedHand (approved):

PR #4990, opened 2025-10-22
perf(l1): batched reads during merkelization

iovoid:
**Motivation**

Currently when trie insertions are performed, all nodes involved are fetched from the database.

Since we know which nodes are likely to be fetched (ie those in the path to the value being changed), we can prefetch most of them.

**Description**

We can't get _every_ node that is going to be used, since there are some odd cases (during deletions we need to get siblings, and there might be unusually deep nodes) but we can catch most (~99%) of them.


crates/storage/store_db/in_memory.rs
```diff
@@ -733,6 +736,72 @@ impl StoreEngine for Store {
...
+
+    fn apply_account_updates_from_trie_batch(
+        &self,
+        state_root: H256,
+        account_updates: &[AccountUpdate],
+    ) -> Result<AccountUpdatesList, StoreError> {
+        let mut ret_storage_updates = Vec::new();
+        let mut code_updates = Vec::new();
+        let mut state_trie = self.open_state_trie(state_root)?;
+        let state_root = state_trie.hash_no_commit();
```
Oppen: This is no longer needed, you just got the root as parameter.

crates/storage/store_db/rocksdb.rs
```diff
@@ -611,6 +613,76 @@ impl Store {
...
+        }
+        let (state_trie_hash, state_updates) = state_trie.collect_changes_since_last_hash();
+
+        Ok(AccountUpdatesList {
+            state_trie_hash,
+            state_updates,
+            storage_updates: ret_storage_updates,
+            code_updates,
+        })
+    }
```
Oppen, line 830: This seems to be the same logic duplicated, why?

crates/storage/store_db/in_memory.rs
iovoid: Fixed in 5f1477137

crates/storage/store_db/rocksdb.rs
iovoid: Most of the logic in indeed duplicate, but we need the tries to be opened from RocksDBPreRead instead of through the usual APIs.

crates/storage/trie_db/rocksdb_preread.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct RocksDBPreReadTrieDB {
+    inner: Arc<RocksDBPreRead>,
+    prefix: Option<Nibbles>,
+}
+
+// TODO: tune
+const PREFETCH_DEPTH_ACCOUNT: usize = 10;
+const PREFETCH_DEPTH_STORAGE: usize = 4;
+const PREFETCH_DEPTH_STORAGE_BIG: usize = 7;
+const PREFETCH_DEPTH_BIG_COUNT: usize = 2;
```
MegaRedHand, line 36: We should tune this

crates/storage/trie_db/rocksdb_preread.rs
```diff
@@ -0,0 +1,211 @@
...
+                addr_nib.as_ref().to_vec()
+            })
+            .collect();
+        let mut reads: Vec<_> = updates
+            .par_iter()
+            .flat_map_iter(|update| {
+                let mut prefixes = HashSet::new();
+                let addr = hash_address(&update.address);
+                let addr_nib = Nibbles::from_bytes(&addr);
+                let size_heuristic = if update.added_storage.len() > PREFETCH_DEPTH_BIG_COUNT {
```
MegaRedHand: We should rename this from `size_heuristic` to `prefetch_depth` or something. We can keep the "heuristic" part as a comment.

crates/storage/trie_db/rocksdb_preread.rs
```diff
@@ -0,0 +1,211 @@
...
+                        Some(addr_nib.clone()),
+                        size_heuristic,
+                    );
+                }
+                insert_prefixes(&mut prefixes, addr_nib, None, PREFETCH_DEPTH_ACCOUNT);
+                prefixes.into_iter()
+            })
+            .collect();
+        fkv_reads.sort();
+        reads.sort();
```
MegaRedHand: Should we dedup here too?
We could try doing the sort in parallel if needed. I think rayon includes a `par_sort_unstable` or similar

crates/storage/trie_db/rocksdb_preread.rs
```diff
@@ -0,0 +1,211 @@
...
+                results.push((key, vec![]));
+            }
+        }
+        let tlc_lock = tlc.write().map_err(|_| TrieError::LockError)?;
+        let cache: HashMap<_, _> = results
+            .into_par_iter()
+            .map(|(key, value)| {
+                if let Some(value) = tlc_lock.get(state_root, &key) {
+                    (key, value)
+                } else {
```
MegaRedHand, line 141: This is confusing. We should rename `value`

crates/storage/trie_db/rocksdb_preread.rs
```diff
@@ -0,0 +1,211 @@
...
+                    (key, value)
+                }
+            })
+            .collect();
+        drop(tlc_lock);
+
+        drop(cf_misc);
+        drop(cf_nodes);
+        drop(cf_flatkeyvalue);
```
MegaRedHand: Why the explicit `drop`s?

crates/storage/trie_db/rocksdb_preread.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        let cf = self
+            .inner
+            .db
+            .cf_handle(if key.is_leaf() {
+                CF_FLATKEYVALUE
+            } else {
+                CF_TRIE_NODES
+            })
```
MegaRedHand, line 198: Does this need to have in mind the `last_computed_flatkeyvalue`?

crates/storage/trie_db/rocksdb_preread.rs
iovoid: Addressed in 657144c1f using BTreeMap

crates/storage/trie_db/rocksdb_preread.rs
iovoid: No, this is dealt with upstream. This function won't be called if the FKV isn't computed yet because `flatkeyvalue_computed` is checked first.

crates/storage/trie_db/rocksdb_preread.rs
iovoid: Because the CFs borrow from DB, and those borrows must cease to exist before we can move db into Self.

crates/storage/trie_db/rocksdb_preread.rs
iovoid: Improved in bc9f5d5a0

crates/storage/trie_db/rocksdb_preread.rs
iovoid: Done in bc9f5d5a0

fmoletta:
# Block Execution Comparison (main vs multi_get_merkelization)

Both branches where run on a server on hoodi network with grafana metrics enabled using the traget `make-hoodi-metrics-docker`. Afterwards the logs where analysed using the log_analysis tool in `tooling/log_analysis`

# Main Branch

Commit: a993586d216ca87dd4d7d3f87b5b7a2ba3985518

Grafana Block Execution Graph:
MainBlockExec

Log Analysis:
MainLogAnalysis


# Current Branch

Commit: bc9f5d5a07103220bb5ad0eca0503108430fcbcb

Grafana Block Execution Graph:

BranchBlockExec


Log Analysis:

BranchLogAnalysis

# Log Files:
[hoodi-main.log](https://github.com/user-attachments/files/23172276/hoodi-main.log)
[hoodi-branch.log](https://github.com/user-attachments/files/23172272/hoodi-branch.log)

PR #4991, opened 2025-10-22
chore(l1): update lighthouse to a version that supports fusaka

mpaulucci:
**Motivation**
Be able to run full nodes with fusaka

**Description**
- Updates lighthouse for all kurtosis actions


ManuelBilbao:
We should update lighthouse version on docs too. For example:
https://github.com/lambdaclass/ethrex/blob/d13c5a854554b90ce01719398bfdd7d8565ecb8d/docs/getting-started/README.md#L55

edg-l (approved):

PR #4993, opened 2025-10-22
perf(l1): batch blobs validate

edg-l:
**Motivation**

something i noticed in a lot of samplies is that `ethrex_common::types::blobs_bundle::BlobsBundle::validate` can take up to 20% of the runtime, depending on the blocks, which mainly calls `c_kzg::bindings::KZGSettings::verify_blob_kzg_proof`, however i noticed the existance of a function `verify_blob_kzg_proof_batch` which may speed up things.

Perf wise looks like about 10 mgas more in some blocks, not much

image

notif-11 is the pr


crates/common/crypto/kzg.rs
```diff
@@ -24,11 +24,15 @@ type Blob = [u8; BYTES_PER_BLOB];
 type Commitment = Bytes48;
 type Proof = Bytes48;
 
+#[cfg(feature = "c-kzg")]
+static KZG_TRUSTED_SETUP: LazyLock<&'static c_kzg::KzgSettings> =
+    LazyLock::new(|| c_kzg::ethereum_kzg_settings(KZG_PRECOMPUTE));
+
```
Oppen: Please don't do this. It's already internally a `OnceCell` in the library.

crates/common/crypto/kzg.rs
```diff
@@ -24,11 +24,15 @@ type Blob = [u8; BYTES_PER_BLOB];
...
 /// Ensures the Ethereum trusted setup is loaded so later KZG operations avoid the first-call cost.
 pub fn warm_up_trusted_setup() {
     #[cfg(feature = "c-kzg")]
     {
-        let _ = c_kzg::ethereum_kzg_settings(8);
+        std::hint::black_box(&KZG_TRUSTED_SETUP);
```
Oppen: This one is a good call though, we should make sure this doesn't get removed as dead code.

crates/common/crypto/kzg.rs
```diff
@@ -106,13 +106,36 @@ pub fn verify_blob_kzg_proof(
...
+
+#[cfg(feature = "c-kzg")]
+pub fn verify_kzg_proof_batch(
+    blobs: &[Blob],
+    commitments: &[Commitment],
+    cell_proof: &[Proof],
+) -> Result<bool, KzgError> {
+    {
+        // perf note: c_kzg::Blob is repr C maybe a unsafe transmute improves perf if the collect were deemed costly
+        let blobs: Vec<_> = blobs.iter().map(|x| c_kzg::Blob::new(*x)).collect();
```
MegaRedHand, line 124: We could try moving to a non-batched call in case the input has a length of 1

crates/common/crypto/kzg.rs
```diff
@@ -106,13 +106,36 @@ pub fn verify_blob_kzg_proof(
...
+            .verify_blob_kzg_proof(&blob.into(), &commitment.into(), &proof.into())
+            .map_err(KzgError::from)
+    }
+}
+
+#[cfg(feature = "c-kzg")]
+pub fn verify_kzg_proof_batch(
+    blobs: &[Blob],
+    commitments: &[Commitment],
+    cell_proof: &[Proof],
```
Oppen, line 120: Can't we just use type aliases for `Blob`, `Commitment` and `Proof` to avoid creating new vectors?

edg-l: afaik hard to do because the blob type from c_kzg is not the same as the simple slice type def we have right now

crates/common/crypto/kzg.rs
edg-l: i'm not sure the branch is worth it, also is it common to only have 1 blob?

Oppen (approved):

Issue #4956, opened 2025-10-20
Improve RLP encoding/decoding


Arkenan:
A big part of all reads is encoding and decoding. We should improve our rlp encoding and decoding of trie elements as it seems to be pretty slow and significant (e.g. more than twice the length of db access in some measurements).

The main culprit seems to be allocation.

PR #4995, opened 2025-10-22
refactor(l1)!: unify trie node encoding

iovoid:
> [!WARNING]
> This changes the format of nodes in the DB. You will need to re-sync your node.
 
**Motivation**

Currently, we have two ways of encoding a node: by specifying the node type (encode_to_vec, used for storing) and by letting the decoder infer it (encode_raw, used for hashing).

This causes confusion and prevents from making use of the RLP's API ability to decode onto an existing buffer.

Also by unifying the formats used for storing and hashing we can avoid encoding the same node twice.

**Description**

Unifies RLP encodings used for database and hashing.

Helps with #4956

### DIFF
--- a/crates/common/trie/node.rs
+++ b/crates/common/trie/node.rs
@@ -64,10 +64,10 @@ impl NodeRef {
                 let mut buf = Vec::new();
                 node.encode(&mut buf);
                 let hash = *hash.get_or_init(|| NodeHash::from_encoded(&buf));
-                acc.push((path.clone(), buf));
                 if let Node::Leaf(leaf) = node.as_ref() {
                     acc.push((path.concat(&leaf.partial), leaf.value.clone()));
                 }
+                acc.push((path.clone(), buf));
 
                 *self = hash.into();
 


crates/common/trie/node/branch.rs
```diff
@@ -248,25 +248,7 @@ impl BranchNode {
...
-                child @ NodeHash::Inline(raw) if raw.1 != 0 => {
-                    encoder = encoder.encode_raw(child.as_ref())
-                }
-                _ => encoder = encoder.encode_bytes(&[]),
-            }
-        }
-        encoder = encoder.encode_bytes(&self.value);
-        encoder.finish();
-        buf
+        NodeHash::from_encoded_raw(&self.encode_to_vec())
```
MegaRedHand: Can we rename this? `from_encoded_raw` -> `from_encoded`

crates/common/trie/rlp.rs
```diff
@@ -1,113 +1,53 @@
...
-            .encode_field(&self.value)
-            .finish()
+        let mut encoder = Encoder::new(buf);
+        for child in self.choices.iter() {
+            match child.compute_hash() {
+                NodeHash::Hashed(hash) => encoder = encoder.encode_bytes(&hash.0),
+                child @ NodeHash::Inline(raw) if raw.1 != 0 => {
+                    encoder = encoder.encode_raw(child.as_ref())
+                }
+                _ => encoder = encoder.encode_bytes(&[]),
```
MegaRedHand, line 24: Here we're storing the hash variant as RLP bytes, and the empty variant as well, but the inline variant we store as raw bytes. Is this OK? Can't it result in collisions?

crates/common/trie/rlp.rs
```diff
@@ -118,19 +58,67 @@ impl RLPEncode for Node {
...
+        let mut item;
+        // Get encoded fields
+        loop {
+            (item, decoder) = decoder.get_encoded_item()?;
+            rlp_items.push(item);
+            // Check if we reached the end or if we decoded more items than the ones we need
+            if decoder.is_done() || rlp_items.len() > 17 {
+                break;
             }
         }
```
MegaRedHand:
```suggestion
        // Check if we reached the end or if we decoded more items than the ones we need
        while !decoder.is_done() && rlp_items.len() <= 17 {
            (item, decoder) = decoder.get_encoded_item()?;
            rlp_items.push(item);
        }
```

We could also make the upper limit `< 17`, adding a check afterwards that `decoder.is_done()`.

crates/common/trie/node.rs
```diff
@@ -69,7 +61,10 @@ impl NodeRef {
                     }
                     Node::Leaf(_) => {}
                 }
-                let hash = *hash.get_or_init(|| node.compute_hash());
+                let mut buf = Vec::new();
+                node.encode(&mut buf);
+                let hash = *hash.get_or_init(|| NodeHash::from_encoded_raw(&buf));
+                acc.push((path.clone(), buf));
```
MegaRedHand: Doesn't this do the same as line 71?

crates/common/trie/rlp.rs
MegaRedHand: Nvm. This can't result in collisions, since the inline variant is stored RLP-encoded, and has a header of length less than 32 bytes.

crates/common/trie/node/branch.rs
iovoid: Done in 174067917

crates/common/trie/rlp.rs
iovoid: Turned into a while loop in 174067917.

crates/common/trie/node.rs
iovoid: Fixed in 174067917.

MegaRedHand (approved):
LGTM

jrchatruc (approved):

PR #4996, opened 2025-10-22
chore(l1): move `tooling` crates to a different workspace

### LINKED ISSUE #4845: Move `tooling` crates to a separate workspace
**Motivation**

The main `Cargo.lock` is populated with dependencies that are part of internal tooling that we don't want there.

**Description**

Move all the `tooling` crates to a workspace apart from the main one.

mpaulucci:
**Motivation**
To not have to compile all the tooling when working on the main workspace

**Description**
- Moves tooling to a separate workspace
- Removes tooling folder from Dockerfile
- Ensures tooling is still compiled and linted in the main workflow

Closes #4845


cmd/ethrex/Cargo.toml
```diff
@@ -120,5 +119,4 @@ hex = { workspace = true, optional = true }
 bytes = { workspace = true, optional = true }
 ethrex-common = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
-genesis-tool = { path = "../../tooling/genesis", optional = true }
```
mpaulucci, line 123: this was weird. Our main crate was depending on a crate from tooling. We should not do this. I moved the code to `ethrex-common` for now, but maybe there is a better place

fedacking (approved):

PR #4999, opened 2025-10-22
fix(l1): stop block building preemptively

### LINKED ISSUE #4992: `getPayload` timeouts in localnet during periods of high traffic (500Mgas or more)
We found by running localnets that when the amount of traffic goes over the node's maximum throughput, the node stops responding in time to `getPayload` requests, which cause missed slots. Ideally, we'd stop the build process and try to package the current block, but it's expected that we at least return the last built block (i.e. at least an empty block).

MegaRedHand:
**Motivation**

We were seeing some missed slots in local networks with high throughput (900Mgas limit). The cause was a blocking call being passed to `CancellationToken::run_until_cancelled`.

**Description**

This PR changes the block building process to run on a blocking task with `spawn_blocking`, and also removes the `async` tag to some non-async methods from `Store` and `Blockchain`.

Closes #4992

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -1538,7 +1538,7 @@ mod tests {
             })
             .collect();
         accounts.sort_by_key(|a| a.0);
-        let mut trie = store.open_state_trie(*EMPTY_TRIE_HASH).unwrap();
+        let mut trie = store.open_direct_state_trie(*EMPTY_TRIE_HASH).unwrap();
         for (address, state) in &accounts {
             trie.insert(address.0.to_vec(), state.encode_to_vec())
                 .unwrap();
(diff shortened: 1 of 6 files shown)


fedacking (approved):

crates/storage/store.rs
```diff
@@ -361,7 +361,7 @@ impl Store {
     /// Applies account updates based on the block's latest storage state
     /// and returns the new state root after the updates have been applied.
     #[instrument(level = "trace", name = "Trie update", skip_all)]
-    pub async fn apply_account_updates_batch(
```
mpaulucci, line 363: Im quite confused about this change (probably due to lack of knowledge). Intuitively I would expect ALL store functions to be async. And also not have to put `spawn_blocking` in places, which are code smell imo

MegaRedHand: Just for the record, I also think that `Store` functions should be `async`. However, I prefer functions to be marked as `async` **only when they are actually asynchronous**. This one didn't call any `async` functions, so the keyword only misleads. What's more, I suspect the compiler actually removes the keyword internally, and I'm upset Clippy doesn't mark cases like this 😠 

I agree that a `spawn_blocking` is a smell when intermingled with `async` code, but in this case, it's justified: building a block is mostly a synchronous task. To avoid using `spawn_blocking`, we'd need to rework the EVM itself to be `async`, or keep a dedicated threadpool for EVMs (essentially making our own `spawn_blocking`).

PR #5001, opened 2025-10-22
chore(l2): refactor `get_embedded_node` and remove hash cloning

xqft:
**Motivation**

Optimizes trie decoding in zkVM stateless execution by adding `NodeHash::finalize_mut` and `NodeRef::compute_hash_finalized` to calculate a node's finalized hash, store it and return a reference to it, instead of cloning the hash or even hashing each time if it's an inline node.

Also refactors the `get_embedded_node` function into `NodeRef::resolve_subtrie` to remove the nested function and simplify the code.

**Flamegraphs**
After all other trie zkVM optimizations, `from_nodes` is one of the most expensive calls in block execution. This PR reduces `Trie::from_nodes` from 131k to 101k, 22%. 

before:
image
after:
image

### DIFF
--- a/crates/common/trie/node_hash.rs
+++ b/crates/common/trie/node_hash.rs
@@ -60,11 +60,11 @@ impl NodeHash {
     /// changing it to the `NodeHash::Hashed` variant.
     pub fn finalize_mut(&mut self) -> &H256 {
         if let NodeHash::Inline(_) = self {
-            let hash = H256::from_slice(
+            let hash = H256(
                 Keccak256::new()
                     .chain_update(self.as_ref())
                     .finalize()
-                    .as_slice(),
+                    .into(),
             );
             *self = NodeHash::Hashed(hash);
         }


crates/common/trie/node_hash.rs
```diff
@@ -61,6 +61,24 @@ impl NodeHash {
...
+    /// Returns a reference the finalized hash, hashing `self` if it's `NodeHash::Inline`, and storing the result by
+    /// changing it to the `NodeHash::Hashed` variant.
+    pub fn finalize_mut(&mut self) -> &H256 {
+        if let NodeHash::Inline(_) = self {
+            let hash = H256::from_slice(
+                Keccak256::new()
+                    .chain_update(self.as_ref())
+                    .finalize()
+                    .as_slice(),
+            );
```
MegaRedHand: Here we can simplify to:

 ```rust
H256(Keccak256::new().chain_update(self.as_ref()).finalize().into())
```

(which would avoid copies)

Comment (xqft) [reply crates/common/trie/node_hash.rs]: done! thank you

File: crates/common/trie/node_hash.rs [hunk 2]
```diff
@@ -56,6 +56,24 @@ impl NodeHash {
         }
     }
 
+    /// Returns a reference to the finalized hash, hashing `self` if it's `NodeHash::Inline`, and storing the result by
+    /// changing it to the `NodeHash::Hashed` variant.
+    pub fn finalize_mut(&mut self) -> &H256 {
```
MegaRedHand, line 61: This function leaves the tree in an inconsistent state, right? If any function expects `compute_hash` to return the node's MPT hash, they'll receive an incorrect result. We should document that in a big warning message to avoid weird errors going forward.

xqft: You mean if there's a function that expects the node's RLP encoding (in the case it's <32 bytes) instead of its hash? Which caller could assume that? and also, how could that leave a trie in an inconsistent state?

xqft:
Superseded by #5163 (subtrie resolving will not be part of the guest program anymore)

PR #5005, opened 2025-10-22
fix(l1): exit on irrecoverable error on sync

fedacking:
**Motivation**

If we have an irrecoverable error in sync, we should exit the node. At the beginning we need to handle deleting the leaves of previous snapsync in case the node was in half of the download.

**Description**

- Made it so that when the sync_cycle function finds an irrecuperable error, it exits the node.
- Change `validate_folders` to `deletes_leaves_folder` to clean up the leaves at start in stead of stopping the node.

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -149,6 +149,7 @@ impl Syncer {
             Ok(()) => {
                 info!(
                     time_elapsed_s = start_time.elapsed().as_secs(),
+                    %sync_head,
                     "Sync cycle finished successfully",
                 );
             }
@@ -160,6 +161,7 @@ impl Syncer {
                         // We exit the node, as we can't recover this error
                         error!(
                             time_elapsed_s = start_time.elapsed().as_secs(),
+                            %sync_head,
                             %error, "Sync cycle failed, exiting as the error is irrecoverable",
                         );
                         std::process::exit(2);
@@ -168,6 +170,7 @@ impl Syncer {
                         // We do nothing, as the error is recoverable
                         error!(
                             time_elapsed_s = start_time.elapsed().as_secs(),
+                            %sync_head,
                             %error, "Sync cycle failed, retrying",
                         );
                     }


crates/networking/p2p/sync.rs
```diff
@@ -147,15 +147,49 @@ impl Syncer {
...
+                    | SyncError::BytecodeFileError
+                    | SyncError::PeerTableError(_) => {
+                        // We do nothing, as the error is recoverable
+                        error!(
+                            time_elapsed_s = start_time.elapsed().as_secs(),
+                            %error, "Sync cycle failed, exiting as the error is irrecoverable",
+                        );
+                        std::process::exit(-1);
+                    }
+                    _ => {
```
mpaulucci: Seems a bit error prone since a new SyncError will by default end here? Is this what we want? Maybe we should list all the recoverable errors instead?

fedacking: Will change it so that we have to explicitly say what category is for each error.

fedacking: Changed in [8fc6295](https://github.com/lambdaclass/ethrex/pull/5005/commits/8fc6295a7bf35909fb7c6b2d0723401c7f7839b6)

crates/networking/p2p/sync.rs
```diff
@@ -148,15 +148,59 @@ impl Syncer {
...
+                    | SyncError::PeerHandler(_)
+                    | SyncError::CorruptPath
+                    | SyncError::TrieGenerationError(_)
+                    | SyncError::AccountTempDBDirNotFound
+                    | SyncError::StorageTempDBDirNotFound
+                    | SyncError::RocksDBError(_)
+                    | SyncError::BytecodeFileError
+                    | SyncError::NoLatestCanonical
+                    | SyncError::PeerTableError(_) => {
+                        // We do nothing, as the error is recoverable
```
mpaulucci: you are exiting, I think this means the error is NOT recoverable

crates/networking/p2p/sync.rs
```diff
@@ -197,18 +241,8 @@ impl Syncer {
...
-                "One of the folders used for temporary leaves during snap is still used. Delete them in {}",
-                &self.datadir.to_str().unwrap_or_default()
-            );
-            std::process::exit(1);
-            // Cloning a single string, the node should stop after this
-            // return Err(SyncError::NotEmptyDatadirFolders(self.datadir.clone()));
-        }
-
+        // they are. If they are not empty we empty them folder
+        deletes_leaves_folder(&self.datadir);
```
mpaulucci: nit: should this be `delete_leaves_folder`?

fedacking: Changed in [5bb5c43](https://github.com/lambdaclass/ethrex/pull/5005/commits/5bb5c4303d294d3fc93d04da58acd783557c68a6)

crates/networking/p2p/sync.rs
fedacking: Fixed in [5bb5c43](https://github.com/lambdaclass/ethrex/pull/5005/commits/5bb5c4303d294d3fc93d04da58acd783557c68a6)

crates/networking/p2p/sync.rs
```diff
@@ -148,15 +148,59 @@ impl Syncer {
...
+                    | SyncError::NoBlockHeaders
+                    | SyncError::PeerHandler(_)
+                    | SyncError::CorruptPath
+                    | SyncError::TrieGenerationError(_)
+                    | SyncError::AccountTempDBDirNotFound
+                    | SyncError::StorageTempDBDirNotFound
+                    | SyncError::RocksDBError(_)
+                    | SyncError::BytecodeFileError
+                    | SyncError::NoLatestCanonical
+                    | SyncError::PeerTableError(_) => {
```
MegaRedHand: Can we put this inside a helper? Maybe as a method `SyncError::is_recoverable`

crates/networking/p2p/sync.rs
```diff
@@ -148,15 +148,59 @@ impl Syncer {
...
+                        // We exit the node, as we can't recover this error
+                        error!(
+                            time_elapsed_s = start_time.elapsed().as_secs(),
+                            %error, "Sync cycle failed, exiting as the error is irrecoverable",
+                        );
+                        std::process::exit(-1);
```
MegaRedHand: We should use a number between 1 and 128 as the exit code ([source](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#exit)).

>  126: A file to be executed was found, but it was not an executable utility.
>  127: A utility to be executed was not found.
> \>128: A command was interrupted by a signal.

fedacking: Done in [15ba627](https://github.com/lambdaclass/ethrex/pull/5005/commits/15ba6279a3945ca8b753ad83ca7ac4c4328aa79b). Should we consider an issue to make sure that we have consistent exit codes?

crates/networking/p2p/sync.rs
fedacking: Done in [15ba627](https://github.com/lambdaclass/ethrex/pull/5005/commits/15ba6279a3945ca8b753ad83ca7ac4c4328aa79b)

mpaulucci (approved):

crates/networking/p2p/sync.rs
```diff
@@ -148,15 +148,31 @@ impl Syncer {
         match self.sync_cycle(sync_head, store).await {
             Ok(()) => {
                 info!(
-                    "Sync cycle finished, time elapsed: {} secs",
-                    start_time.elapsed().as_secs()
+                    time_elapsed_s = start_time.elapsed().as_secs(),
+                    "Sync cycle finished successfully",
```
MegaRedHand, line 153: Let's add the `sync_head` (abbreviated, with `Display`) to this log.

fedacking: Done in [997571e](https://github.com/lambdaclass/ethrex/pull/5005/commits/997571e5ff106fe0ff19066ec3e63b2ca88f8886)

PR #5006, opened 2025-10-23
feat(l2): generate batch prover input at commit and store it in db

ilitteri:
**Motivation**

We want to persist batch prover inputs to avoid having to regenerate them in case of an unexpected shutdown. We also want to store the version of the binary that generated them to be able to prove yet-unproven batches from older ethrex versions.

**Description**

This PR:
- Extends the L2 store functionality to support storing and retrieving batch prover inputs.
- For the SQL store, a new table `batch_prover_input` is created with rows `prover_input` (actual rkyv-encoded `ProverInputData`), `batch_number`, and `prover_version`, being the tuple of the last two the primary key.
- For the in-memory store, a new hash map `batch_prover_input` is created with key `(batch_number, prover_version)` and value `witness` (actual rkyv-encoded `ProverInputData`).
- The committer now generates and stores the batch prover inputs in the L2 store.
- The proof coordinator now retrieves the batch prover inputs from the L2 store.
- `ProverInputData` is moved to the `ethrex-l2-common` crate, as it is now a common data structure between different crates.
- The prover client now requests batch prover input data specifying both batch number and binary version.

**Future work**

With this new feature, an older prover should be able to retrieve data from an upgraded sequencer, as long as there are batches to be proven in that version. Currently it rejects any request from a different version.

### DIFF
--- a/crates/l2/tee/quote-gen/src/main.rs
+++ b/crates/l2/tee/quote-gen/src/main.rs
@@ -3,7 +3,7 @@ mod sender;
 use configfs_tsm::create_tdx_quote;
 use ethrex_common::Bytes;
 use ethrex_common::utils::keccak;
-use ethrex_l2::sequencer::proof_coordinator::get_commit_hash;
+use ethrex_l2::sequencer::utils::get_git_commit_hash;
 use ethrex_l2_common::{
     calldata::Value,
     prover::{BatchProof, ProofCalldata, ProverType},
@@ -82,7 +82,7 @@ async fn setup(private_key: &SecretKey) -> Result<(), String> {
 #[tokio::main]
 async fn main() {
     let (private_key, _) = generate_keypair(&mut rand::rngs::OsRng);
-    let commit_hash = get_commit_hash();
+    let commit_hash = get_git_commit_hash();
     while let Err(err) = setup(&private_key).await {
         println!("Error sending quote: {}", err);
         sleep(Duration::from_millis(POLL_INTERVAL_MS)).await;
--- a/crates/l2/tee/quote-gen/src/sender.rs
+++ b/crates/l2/tee/quote-gen/src/sender.rs
@@ -4,7 +4,6 @@ use tokio::{
     net::TcpStream,
 };
 
-use ethrex_l2::constants::BIN_VERSION;
 use ethrex_l2::sequencer::proof_coordinator::ProofData;
 use ethrex_l2_common::prover::{BatchProof, ProverType};
 
@@ -16,7 +15,6 @@ const SERVER_URL_DEV: &str = "localhost:3900";
 pub async fn get_batch(commit_hash: String) -> Result<(u64, ProgramInput), String> {
     let batch = connect_to_prover_server_wr(&ProofData::BatchRequest {
         commit_hash: commit_hash.clone(),
-        version: BIN_VERSION.to_string(),
     })
     .await
     .map_err(|e| format!("Failed to get Response: {e}"))?;


crates/l2/prover/src/prover.rs
```diff
@@ -92,7 +95,7 @@ impl Prover {
 
     async fn request_new_input(&self, endpoint: &Url) -> Result<Option<ProverData>, String> {
         // Request the input with the correct batch_number
-        let request = ProofData::batch_request(self.commit_hash.clone());
+        let request = ProofData::batch_request(self.commit_hash.clone(), BIN_VERSION.to_string());
```
ManuelBilbao: I think `commit_hash` already do this job

crates/l2/sequencer/l1_committer.rs
```diff
@@ -536,6 +551,69 @@ impl L1Committer {
         ))
     }
 
+    async fn generate_and_store_batch_prover_input(
+        &self,
+        batch: &Batch,
+    ) -> Result<(), CommitterError> {
+        let blocks = self.fetch_batch_blocks(batch).await?;
```
ManuelBilbao: Can we just call `fetch_batch_blocks(batch.number, &self.store, &self.rollup_store).await` here, to avoid indirections?

crates/l2/storage/src/store.rs
```diff
@@ -355,4 +355,25 @@ impl Store {
             .delete_proof_by_batch_and_type(batch_number, proof_type)
             .await
     }
+
+    pub async fn store_prover_input_by_batch_and_version(
+        &self,
+        batch_number: u64,
+        prover_version: String,
```
ManuelBilbao: I think this can be a `&str`. This way, we can save some `.clone()`s when using SQL store. In case of InMemory store, the clone will be needed anyway, but that's not used in production

crates/l2/storage/src/store_db/sql.rs
```diff
@@ -281,6 +282,27 @@ impl SQLStore {
...
+    ) -> Result<(), RollupStoreError> {
+        let prover_input_bytes = rkyv::to_bytes::<rkyv::rancor::Error>(&prover_input)
+            .map_err(|e| {
+                RollupStoreError::Custom(format!("Failed to serialize prover input: {e}"))
+            })?
+            .to_vec();
+
+        let queries = vec![(
+            "INSERT OR REPLACE INTO batch_prover_input VALUES (?1, ?2, ?3)",
+            (batch_number, prover_version, prover_input_bytes).into_params()?,
```
ManuelBilbao, line 301: Here `prover_version` can be `&str` and it still works

crates/l2/storage/src/store_db/sql.rs
```diff
@@ -781,6 +807,47 @@ impl StoreEngineRollup for SQLStore {
...
             .transpose()
     }
+
+    async fn store_prover_input_by_batch_and_version(
+        &self,
+        batch_number: u64,
+        prover_version: String,
+        prover_input: ProverInputData,
+    ) -> Result<(), RollupStoreError> {
+        self.store_prover_input_by_batch_and_version_in_tx(
```
ManuelBilbao, line 817: Why not do `store_prover_input_by_batch_and_version_in_tx` stuff directly here?

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -296,6 +259,7 @@ impl ProofCoordinator {
         &mut self,
         stream: &mut TcpStream,
         commit_hash: String,
+        version: String,
     ) -> Result<(), ProofCoordinatorError> {
```
ManuelBilbao, line 252: We should do some refactor in this function. With this new feature, an older prover should be able to retrieve data from an upgraded sequencer, as long as there are batches to be proven in that version. Currently it rejects any request from a different version

ManuelBilbao: More on this, we should update the OCP contract, allowing to verify a batch with the older version (VK). Maybe we can do this in another PR?

crates/l2/prover/src/prover.rs
ilitteri: You're right, I've realized that when I was documenting stuff. Jijo.

crates/l2/storage/src/store_db/sql.rs
ilitteri: I followed the convention on the other `store` methods.

crates/l2/sequencer/proof_coordinator.rs
ilitteri: Yes, I think the best would be to handle it in another PR. I'll file an issue and add a future fork section in the PR's description.

crates/l2/sequencer/l1_committer.rs
ilitteri: [ac102a9](https://github.com/lambdaclass/ethrex/pull/5006/commits/ac102a94f79ab8026ea805688579c1c4e40c261e)

crates/l2/storage/src/store_db/sql.rs
ilitteri: [40c2d2c](https://github.com/lambdaclass/ethrex/pull/5006/commits/40c2d2cf8f339e49de65ca716d692f1ae503490f)

crates/l2/storage/src/store.rs
ilitteri: [40c2d2c](https://github.com/lambdaclass/ethrex/pull/5006/commits/40c2d2cf8f339e49de65ca716d692f1ae503490f)

crates/l2/prover/src/prover.rs
ilitteri: [0526c31](https://github.com/lambdaclass/ethrex/pull/5006/commits/0526c31157520edd53aa1ded28fd07b0c7b20367)

ManuelBilbao (approved):

avilagaston9 (approved):

PR #5008, opened 2025-10-23
feat(l2): revert current native token feature

tomip01:
> [!NOTE]
> #5024 re-introduces the feature with a different approach

**Motivation**

We want a new approach for having native tokens in the L2. We want to keep the balance of the accounts to remain being ETH but allow the fees of the transactions to be paid with an ERC20.

**Description**

- Reverted commit f6ccc00c8ea3eb5ac955b206e8f994429ca58580.

### DIFF
--- a/crates/l2/contracts/src/l1/CommonBridge.sol
+++ b/crates/l2/contracts/src/l1/CommonBridge.sol
@@ -88,7 +88,7 @@ contract CommonBridge is
     mapping(bytes32 => uint256) public privilegedTxDeadline;
 
     /// @dev Deprecated variable.
-	  /// @notice The L1 token address that is treated as the one to be bridged to the L2.
+    /// @notice The L1 token address that is treated as the one to be bridged to the L2.
     /// @dev If set to address(0), ETH is considered the native token.
     /// Otherwise, this address is used for native token deposits and withdrawals.
     address public NATIVE_TOKEN_L1;


crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -87,11 +86,6 @@ contract CommonBridge is
     /// @notice Deadline for the sequencer to include the transaction.
     mapping(bytes32 => uint256) public privilegedTxDeadline;
 
-    /// @notice The L1 token address that is treated as the one to be bridged to the L2.
-    /// @dev If set to address(0), ETH is considered the native token.
-    /// Otherwise, this address is used for native token deposits and withdrawals.
-    address public NATIVE_TOKEN_L1;
```
tomip01, line 93: do we deprecate this var?

cmd/ethrex/l2/deployer.rs
```diff
@@ -500,15 +490,9 @@ const INITIALIZE_ON_CHAIN_PROPOSER_SIGNATURE: &str = "initialize(bool,address,ad
...
+const BRIDGE_INITIALIZER_SIGNATURE: &str = "initialize(address,address,uint256)";
 
-// deposit(uint256 _amount, address _l2Recipient)
-const NATIVE_TOKEN_DEPOSIT_SIGNATURE: &str = "deposit(uint256,address)";
-
-// approve(address spender, uint256 amount)
-const APPROVE_SIGNATURE: &str = "approve(address,uint256)";
-
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Copy)]
```
ManuelBilbao: I'd remove `Copy` too

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -214,64 +205,28 @@ contract CommonBridge is
...
-                address(this),
-                value
-            );
-        }
-
-        deposits[NATIVE_TOKEN_L2][NATIVE_TOKEN_L2] += value;
+        _deposit(l2Recipient);
+    }
 
+    function _deposit(address l2Recipient) private {
```
ManuelBilbao: We can now have `_deposit` body directly in `deposit` and call `deposit()` on `receive()`

cmd/ethrex/l2/deployer.rs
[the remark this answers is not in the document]
tomip01: It does not break anything. We can add it in the future if we need to

crates/l2/contracts/src/l1/CommonBridge.sol
ManuelBilbao: I think we can safely this as we didn't deploy any L2 since this was added

crates/l2/contracts/src/l1/CommonBridge.sol
tomip01: Done it here [a217eff](https://github.com/lambdaclass/ethrex/pull/5008/commits/a217effea31fad20ebef312b72291ec4740aa4c9)

cmd/ethrex/l2/deployer.rs
tomip01: Done! [a217eff](https://github.com/lambdaclass/ethrex/pull/5008/commits/a217effea31fad20ebef312b72291ec4740aa4c9)

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -87,7 +87,8 @@ contract CommonBridge is
     /// @notice Deadline for the sequencer to include the transaction.
     mapping(bytes32 => uint256) public privilegedTxDeadline;
 
-    /// @notice The L1 token address that is treated as the one to be bridged to the L2.
+    /// @dev Deprecated variable.
+	  /// @notice The L1 token address that is treated as the one to be bridged to the L2.
```
ManuelBilbao: nit
```suggestion
    /// @notice The L1 token address that is treated as the one to be bridged to the L2.
```

PR #5012, opened 2025-10-23
refactor(l1): remove unused async modifiers in functions

MegaRedHand:
**Motivation**

We had some problems before due to functions marked `async` that weren't safe to use in an async context, since they blocked for most of the execution, or weren't even awaiting futures inside them.

**Description**

This PR fixes some warnings raised by the [`clippy::unused_async` lint](https://rust-lang.github.io/rust-clippy/master/index.html#unused_async) linter, which flags functions marked `async` but without any internal `await` calls.

It also refactors the "blocks" widget of the L2 monitor, since clippy now complained the types returned were too complex

Related #5662

### DIFF
--- a/crates/l2/monitor/widget/blocks.rs
+++ b/crates/l2/monitor/widget/blocks.rs
@@ -142,7 +142,7 @@ impl BlocksTable {
             })
             .collect::<Vec<_>>();
 
-        new_blocks_processed.sort_by(|a, b| b.number.cmp(&a.number));
+        new_blocks_processed.sort_by_key(|x| x.number);
 
         new_blocks_processed
     }


crates/l2/monitor/widget/blocks.rs
```diff
@@ -22,11 +22,48 @@ use crate::{
     sequencer::errors::MonitorError,
 };
 
+struct BlockEntry {
```
MegaRedHand, line 25: Clippy complained about the return types being too complex.

MegaRedHand:
I looked into the functions that were fake async functions and they all have true async functions around them, so this shouldn't bring any regression by blocking the tokio runtime (though I believe the resulting assembly will be the same as before).

crates/l2/monitor/widget/blocks.rs
```diff
@@ -100,27 +128,21 @@ impl BlocksTable {
...
+                gas_used: block.header.gas_used,
+                blob_gas_used: block.header.blob_gas_used,
+                size: block.length(),
             })
             .collect::<Vec<_>>();
 
-        new_blocks_processed.sort_by(
-            |(number_a, _, _, _, _, _, _), (number_b, _, _, _, _, _, _)| number_b.cmp(number_a),
-        );
+        new_blocks_processed.sort_by(|a, b| b.number.cmp(&a.number));
```
Oppen:
```suggestion
        new_blocks_processed.sort_by_key(|x| x.number);
```

Oppen (approved):

crates/l2/monitor/widget/blocks.rs
pablodeymo: I don't like having this refactor in the same PR removing async modifiers.
Even more, the description of the PR doesn't mention this change.

PR #5015, opened 2025-10-23
ci(l2): adds step in release workflow to compile and cache solc with lower GLIBC version.

samoht9277:
**Motivation**

This PR changes the ARM build environment from **Ubuntu 24.04** to **Ubuntu 22.04**.  
The goal is to compile Ethrex with an older **glibc**, increasing compatibility with systems running older Linux versions.

It also updates how **Solidity** is handled. Since `solc` doesn’t publish ARM binaries, we previously used a third-party ARM build that required a newer glibc, breaking builds on Ubuntu 22.04.  
Now, `solc` is compiled from source and cached. This ensures it links against the same glibc version as the 22.04 system, allowing Ethrex to build and run on environments with lower glibc versions.

**Description**

- ARM build now targets Ubuntu 22.04.  
- `solc` is built from source and cached to match system glibc.  
- Improves binary portability across older Linux distributions.


.github/workflows/tag_release.yaml
```diff
@@ -121,11 +121,31 @@ jobs:
...
-        if: ${{ matrix.platform == 'ubuntu-24.04-arm' }}
+      - name: Restore solc cache (Linux ARM)
+        if: ${{ matrix.platform == 'ubuntu-22.04-arm' }}
+        id: cache-solc
+        uses: actions/cache@v4.3.0
+        with:
+          path: ~/.cache/solc/v0.8.29
+          key: solc-${{ runner.os }}-${{ matrix.arch }}-v0-8-29
+
+      - name: Compile solc from source (Linux ARM)
```
klaus993: was this step tested in a workflow? if so, please link the workflow run. if not, we should test it beforehand.

samoht9277: There is no _real_ way to test this other than actually try this and see if it works.

I followed this guide: https://github.com/actions/cache?tab=readme-ov-file#skipping-steps-based-on-cache-hit 

I guess we can copy the workflow in a new repo and set up everything exactly like this one, and test.

samoht9277: I did test each step on an empty machine, replicating each command one by one to simulate the same behavior, and It worked with no issues.

samoht9277: Compilation was successful here: https://github.com/lambdaclass/test-solc-compile/actions/runs/18793122723

And cache also successful here: https://github.com/lambdaclass/test-solc-compile/actions/runs/18843543503

.github/workflows/tag_release.yaml
```diff
@@ -109,37 +108,33 @@ jobs:
...
-          sudo curl -L -o /usr/local/bin/solc https://github.com/nikitastupin/solc/raw/refs/heads/main/linux/aarch64/solc-v0.8.29
-          sudo chmod +x /usr/local/bin/solc
+          if [[ "${{ matrix.os }}" == "linux" ]]; then
+            sudo curl -L -o /usr/local/bin/solc https://github.com/argotorg/solidity/releases/download/v0.8.29/solc-static-linux
+            sudo chmod +x /usr/local/bin/solc
+          elif [[ "${{ matrix.os }}" == "macos" ]]; then
+            sudo curl -L -o /usr/local/bin/solc https://github.com/argotorg/solidity/releases/download/v0.8.29/solc-macos
+            sudo chmod +x /usr/local/bin/solc
+          fi
+          solc --version
```
samoht9277, line 123: Tested here: https://github.com/samoht9277/ethrex/actions/runs/18925460915


This works because ARM builds depend on `docker/setup-qemu-action` for the SP1 dependency to work. 

We can take advantage of this by using the `amd64` version of the binary on Linux (which is also compiled statically) for the ARM version as well.

PR #5030, opened 2025-10-23
test(l1): parallelize test file parsing in state_v2 runner

### LINKED ISSUE #3881: Parallelize parsing of the test files
The current active runner has been improved to parallelize the parsing of tests files, the idea would be to translate this to the new runner (might be a bit different as it is a different implementation but the current code could serve as a guide).

crStiv:
- Add rayon dependency for parallel processing
- Replace sequential directory traversal with rayon parallel iteration
- Import IndexedParallelIterator trait for collect_into_vec method
- Maintain existing file filtering and recursive directory logic

Resolves #3881

### DIFF
--- a/tooling/Cargo.lock
+++ b/tooling/Cargo.lock
@@ -3050,6 +3050,7 @@ dependencies = [
  "ethrex-vm",
  "hex",
  "prettytable-rs",
+ "rayon",
  "rustc-hash 2.1.1",
  "secp256k1 0.30.0",
  "serde",


tooling/ef_tests/state_v2/Cargo.toml
```diff
@@ -27,6 +27,7 @@ colored = "2.1.0"
 alloy-rlp = "0.3.12"
 chrono = "0.4"
 prettytable-rs = "0.10"
+rayon = "1.10.0"
```
ElFantasma: nit: `rayon` is already a workspace dependency (root `Cargo.toml` line 121). Use `rayon.workspace = true` instead of hardcoding the version — keeps it consistent with the rest of the codebase (`crates/storage/Cargo.toml`, `crates/common/Cargo.toml`, etc.).

tooling/ef_tests/state_v2/src/modules/parser.rs
```diff
@@ -70,40 +71,52 @@ pub fn parse_dir(
...
+                }
             }
-        }
-    }
-    // Up to this point the parsing of every .json file has given a Vec<Test> as a result, so we have to concat
-    // to obtain a single Vec<Test> from the Vec<Vec<Test>>.
-    Ok(tests.concat())
+            Ok(None)
+        })
+        .collect_into_vec(&mut directory_tests_results);
```
ElFantasma: The `collect_into_vec(&mut directory_tests_results)` + pre-declared `Vec` pattern is unnecessary here. You can simplify this to:

```rust
let directory_tests_results: Vec<_> = dir_entries
    .into_par_iter()
    .map(|entry| -> Result<Option<Vec<Test>>, RunnerError> {
        // ... same closure body ...
    })
    .collect();
```

This is what the v1 runner does at `state/parser.rs:71-76` for its top-level collection, and it removes the need for the `IndexedParallelIterator` import.

ElFantasma (changes requested):

PR #5037, opened 2025-10-24
fix(l2): use checkpoints to persist previous batch state

ilitteri:
**Motivation**

L2s with batches with more than 131 blocks (with at least one tx in the batch) stopped working correctly after the path-based feature was introduced.

**Description**

Extends `ethrex-storage` API with a `create_snapshot` function, which creates a checkpoint of the DB at a provided path, and it is used in the following manner:
- The node initializes with a checkpoint of the current state of the store and heals with the additional blocks.
- On each batch seal, a new checkpoint is created containing the state of the latest block from the sealed batch.
- When it's time to prepare a batch to commit, the previously created checkpoint is used for both state diff computation and witness generation.
- Checkpoints are erased once the batch they served is verified on the L1.

**Future work**

Remove checkpoints after the batch they served is verified in the L1.

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -45,7 +45,7 @@ use ethrex_rpc::{
 use ethrex_storage::EngineType;
 use ethrex_storage::Store;
 use ethrex_storage_rollup::StoreRollup;
-use ethrex_vm::BlockExecutionResult;
+use ethrex_vm::{BlockExecutionResult, Evm};
 use serde::Serialize;
 use std::{
     collections::{BTreeMap, HashMap},
@@ -495,12 +495,15 @@ impl L1Committer {
                     potential_batch_block.header.parent_hash,
                 );
 
-                // There's no specific reason to prefer a blockchain instance
-                // here, we just need one to create the EVM. We just use the
-                // current checkpoint blockchain just in case.
-                // The important thing here is to use the correct vm_db, which
-                // must be created from the checkpoint store.
-                let mut vm = one_time_checkpoint_blockchain.new_evm(vm_db).await?;
+                let fee_config = self
+                    .rollup_store
+                    .get_fee_config_by_block(block_to_commit_number)
+                    .await?
+                    .ok_or(CommitterError::FailedToGetInformationFromStorage(
+                        "Failed to get fee config for re-execution".to_owned(),
+                    ))?;
+
+                let mut vm = Evm::new_for_l2(vm_db, fee_config)?;
 
                 vm.execute_block(&potential_batch_block)?;
 


crates/blockchain/blockchain.rs
```diff
@@ -317,14 +340,39 @@ impl Blockchain {
...
+            // Later in this function we query the batch block headers by number,
+            // to do so, they need to be marked as canonical.
+            // Only users that have the state previous to the first batch block
+            // need this step.
+            if self
+                .storage
+                .get_block_header(block.header.number)?
+                .is_none()
+            {
+                apply_fork_choice(&self.storage, block.hash(), block.hash(), block.hash())
```
avilagaston9: Can't we do the calls by hash instead? I think this could be dangerous. This function is called by the `debug_executionWitness` endpoint. Let's imagine we receive a request, fetch some blocks from the db and call `generate_witness_for_blocks`. If a reorg occurs while we are generating the witness, we might retrieve a different header or maybe apply a forkchoice to invalid blocks.

ilitteri: I agree, I'll test using the hashes since they're available.

crates/blockchain/blockchain.rs
```diff
@@ -317,14 +340,39 @@ impl Blockchain {
             }
 
             // Apply account updates to the trie recording all the necessary nodes to do so
-            let (updated_trie, storage_tries_after_update) = self
+            let (storage_tries_after_update, account_updates_list) = self
```
gianbelinche, line 379: Why did this change order?

ilitteri: Sorry, I didn't understand the question.

crates/blockchain/blockchain.rs
ilitteri: I think it'd be better to tackle this in another PR.

cmd/ethrex/l2/initializers.rs
```diff
@@ -180,10 +189,18 @@ pub async fn init_l2(
         perf_logs_enabled: true,
     };
 
-    let blockchain = init_blockchain(store.clone(), blockchain_opts);
+    let blockchain = init_blockchain(store.clone(), blockchain_opts.clone());
 
     regenerate_head_state(&store, &blockchain).await?;
 
+    let (initial_checkpoint_store, initial_checkpoint_blockchain) = initialize_checkpoint(
```
avilagaston9: Why do we return these? It is confusing because `blockchain` is passed to `init_rpc_api` and `initial_checkpoint_blockchain` to the sequencer. The same happens with `initial_checkpoint_store` and `store`.

crates/blockchain/blockchain.rs
ilitteri: [8a29d11](https://github.com/lambdaclass/ethrex/pull/5037/commits/8a29d11747a4b536e055fd0fa53a4edc24e7f010)

crates/l2/sequencer/l1_committer.rs
```diff
@@ -422,13 +488,52 @@ impl L1Committer {
...
+                    one_time_checkpoint_store.clone(),
+                    potential_batch_block.header.parent_hash,
+                );
+
+                // There's no specific reason to prefer a blockchain instance
+                // here, we just need one to create the EVM. We just use the
+                // current checkpoint blockchain just in case.
+                // The important thing here is to use the correct vm_db, which
+                // must be created from the checkpoint store.
+                let mut vm = one_time_checkpoint_blockchain.new_evm(vm_db).await?;
```
avilagaston9: We need to fetch the `feeConfig` used for that block from the `rollup_store`. So this should instead be:
```suggestion
                let fee_config = self
                    .rollup_store
                    .get_fee_config_by_block(block_to_commit_number)
                    .await?
                    .ok_or(CommitterError::FailedToGetInformationFromStorage(
                        "Failed to get fee config for re-execution".to_owned(),
                    ))?;
                let mut vm = Evm::new_for_l2(vm_db, fee_config)?;
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -618,6 +734,126 @@ impl L1Committer {
...
+    }
+
+    /// Creates a checkpoint of the given store at the specified path.
+    ///
+    /// This function performs the following steps:
+    /// 1. Creates a checkpoint of the provided store at the specified path.
+    /// 2. Initializes a new store and blockchain for the checkpoint.
+    /// 3. Regenerates the head state in the checkpoint store.
+    /// 4. Validates that the checkpoint store's head block number and latest block match those of the original store.
+    async fn create_checkpoint(
```
avilagaston9, line 782: Can this reuse the one in `initializers.rs`?

crates/l2/sequencer/l1_committer.rs
ilitteri: [1948a88](https://github.com/lambdaclass/ethrex/pull/5037/commits/1948a8811bf10fefa129d5a27f4f17def0f67a09)

crates/l2/sequencer/l1_committer.rs
ilitteri: I'd do it the other way around (reusing this one in `initializers.rs`), though, I kept it this way in case we need to update some logic specific to the committer.

PR #5040, opened 2025-10-24
perf(l1): try ahash instead of fxhash

edg-l:
**Motivation**

**Description**

Closes #issue_number


edg-l:
deemed not worth it by comparisons

PR #5051, opened 2025-10-25
perf(l1): run "engine_newPayload" block execution in a worker thread

Oppen:
Extract the block execution for payloads to an OS thread, make all its
callees regular blocking functions, and removes the reliance on Tokio.

Caveats:
1. The worker is in the RPC, mostly because the Blockchain is not Clone;
2. The L2 is broken from the removal of async, fee config mostly becomes
   a mess;
3. The channel is provided by Tokio because the RPC is still async and
   shouldn't block waiting.

**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -43,7 +43,7 @@ use crate::{
 };
 
 use thiserror::Error;
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 #[derive(Debug)]
 pub struct PayloadBuildTask {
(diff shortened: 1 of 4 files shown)


crates/networking/rpc/engine/payload.rs
```diff
@@ -678,6 +679,20 @@ fn validate_block_hash(payload: &ExecutionPayload, block: &Block) -> Result<(),
...
+    ctx.block_worker_channel
+        .send((notify_send, block))
+        .map_err(|e| {
+            ChainError::Custom(format!(
+                "failed to send block execution request to worker: {e}"
+            ))
+        })?;
+    notify_recv
+        .await
+        .map_err(|e| ChainError::Custom(format!("failed to receive block execution result: {e}")))?
```
MegaRedHand, line 693: We'll need to check for edge cases like two competing blocks being received at the same time, in the future. That could lead to increased latency.

PR #5054, opened 2025-10-27
feat(l1,l2): add DB versioning

### LINKED ISSUE #5056: Add versioning to DB
We don't have an easy way to check version inconsistencies in our DB, and this causes weird errors when users update to a new version. We should add some version number to our DB and check that on startup, to better handle these cases.

MegaRedHand:
**Motivation**

We already had some changes that break the DB format. We want to easily detect these cases and tell the user how to handle them.

**Description**

This PR adds a simple versioning scheme with a single numerical version number, stored in a standalone file under `<ethrex datadir>/db.version`.

Closes #5056

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -42,6 +42,8 @@ use std::fmt::Debug;
 // TODO: use finalized hash to determine when to commit
 const COMMIT_THRESHOLD: usize = 128;
 
+const CURRENT_DB_VERSION: u64 = 1;
+
 /// Canonical block hashes column family: [`u8;_`] => [`Vec<u8>`]
 /// - [`u8;_`] = `block_number.to_le_bytes()`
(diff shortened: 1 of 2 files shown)


crates/storage/store_db/rocksdb.rs
```diff
@@ -1783,3 +1787,22 @@ fn open_cfs<'a, const N: usize>(
         .try_into()
         .map_err(|_| StoreError::Custom("Unexpected number of column families".to_string()))
 }
+
+/// Check the current DB version against the expected version
+fn ensure_correct_db_version(path: &Path) -> Result<(), StoreError> {
+    let version_file_path = path.join("db.version");
+    if !version_file_path.exists() {
+        std::fs::write(version_file_path, CURRENT_DB_VERSION.to_string())?;
```
Oppen, line 1795: This is the wrong action for missing version IMO. It should fail rather than create the file. Why? Because otherwise you claim compatibility with all past versions before we versioned, which will break.

PR #5055, opened 2025-10-27
perf(l1)!: separate Account and storage Column families

Arkenan:
The goal is to separate both the trie nodes and flat key value representation for accounts and storage, as they can be updated separately.

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -1350,11 +1350,14 @@ impl StoreEngine for Store {
     }
 
     fn get_account_code(&self, code_hash: H256) -> Result<Option<Code>, StoreError> {
-        let hash_key = code_hash.as_bytes().to_vec();
-        let Some(bytes) = self.read_sync(CF_ACCOUNT_CODES, hash_key)? else {
+        let cf = self.cf_handle(CF_ACCOUNT_CODES)?;
+        let Some(bytes) = self
+            .db
+            .get_pinned_cf(&cf, code_hash.as_bytes())
+            .map_err(|e| StoreError::Custom(format!("RocksDB read error: {}", e)))?
+        else {
             return Ok(None);
         };
-        let bytes = Bytes::from_owner(bytes);
         let (bytecode, targets) = decode_bytes(&bytes)?;
         let code = Code {
             hash: code_hash,
--- a/.github/workflows/pr-main_l1_l2_dev.yaml
+++ b/.github/workflows/pr-main_l1_l2_dev.yaml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   integration-test-l1-l2-dev:
+    if: ${{ github.event_name != 'merge_group' }}
     name: Integration Test
     runs-on: ubuntu-latest
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 2025-11-12
 
 - Separate Account and storage Column families in rocksdb [#5055](https://github.com/lambdaclass/ethrex/pull/5055)
+- Avoid copying while reading account code [#5289](https://github.com/lambdaclass/ethrex/pull/5289)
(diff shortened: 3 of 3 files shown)


crates/storage/store_db/rocksdb.rs
```diff
@@ -262,7 +277,20 @@ impl Store {
...
+                    cf_opts.set_min_write_buffer_number_to_merge(2);
+                    cf_opts.set_target_file_size_base(256 * 1024 * 1024); // 256MB
+                    cf_opts.set_memtable_prefix_bloom_ratio(0.2); // Bloom filter
+
+                    let mut block_opts = BlockBasedOptions::default();
+                    block_opts.set_block_size(16 * 1024); // 16KB
+                    block_opts.set_bloom_filter(10.0, false); // 10 bits per key
+                    cf_opts.set_block_based_table_factory(&block_opts);
+                }
+                CF_STORAGE_TRIE_NODES => {
```
MegaRedHand: Unless we have custom options for both, I suggest we do:

```suggestion
                CF_ACCOUNT_TRIE_NODES | CF_STORAGE_TRIE_NODES => {
```

crates/storage/store.rs
```diff
@@ -532,50 +532,43 @@ impl Store {
...
                 if !storage_value.is_zero() {
                     let hashed_key = hash_key(&H256(storage_key.to_big_endian()));
                     storage_trie.insert(hashed_key, storage_value.encode_to_vec())?;
                 }
             }
-            let (storage_root, new_nodes) = storage_trie.collect_changes_since_last_hash();
-            nodes.insert(H256::from_slice(&hashed_address), new_nodes);
+
+            // TODO(#5195): committing each storage trie individually is inefficient.
+            // We would benefit form a mass storage node insertion method.
```
MegaRedHand, line 559: Can't we do it like before, with `Trie::collect_changes_since_last_hash()`? We can store the changes inside a shared buffer, and dump it in the DB at the end.

crates/storage/store_db/rocksdb.rs
```diff
@@ -744,22 +775,45 @@ impl Store {
...
+            ],
+        )?;
 
         let last_written = db.get_cf(&cf_misc, "last_written")?.unwrap_or_default();
         // Commit removes the bottom layer and returns it, this is the mutation step.
         let nodes = trie_mut.commit(root).unwrap_or_default();
         for (key, value) in nodes {
             let is_leaf = key.len() == 65 || key.len() == 131;
+            // Accounts have only the account address as their path, while storage keys have
+            // the account address (32 bytes) + storage path (up to 32 bytes).
```
iovoid: Mentioning the pre-encoding sizes before code dealing with encoded sizes might generate confusion (one might think 32 + 32 and 65 are related).

crates/storage/store_db/rocksdb.rs
```diff
@@ -262,7 +277,20 @@ impl Store {
                     block_opts.set_bloom_filter(10.0, false);
                     cf_opts.set_block_based_table_factory(&block_opts);
                 }
-                CF_TRIE_NODES => {
+                CF_ACCOUNT_TRIE_NODES => {
+                    cf_opts.set_compression_type(rocksdb::DBCompressionType::Lz4);
```
iovoid: Now that #5223 is merged, this should be removed.

crates/storage/store_db/rocksdb.rs
```diff
@@ -262,7 +277,20 @@ impl Store {
                     block_opts.set_bloom_filter(10.0, false);
                     cf_opts.set_block_based_table_factory(&block_opts);
                 }
-                CF_TRIE_NODES => {
+                CF_ACCOUNT_TRIE_NODES => {
```
iovoid: If CF_STORAGE_TRIE_NODES and CF_ACCOUNT_TRIE_NODES will have the same settings, maybe they should be under the same condition.

crates/storage/store.rs
Oppen: I agree, we already have `commit_without_storing` for that.
That said, the genesis runs exactly once and I think it's quick enough as it is.

Oppen (approved):

PR #5073, opened 2025-10-27
perf(l1): optimized VM database for RocksDB

iovoid:
**Motivation**

Currently to get a storage we are:
- getting the block header to obtain the state root
- getting the account to get the storage root
- reading the value

The first step isn't needed because the state root is the same through the block, and the second isn't needed because since we're using a path-based DB we can directly get the value (without even using the trie, if the FKVs are computed).

**Description**

Implements a RocksDB-specific VM backend.


jrchatruc (changes requested):
Blocking this merge until we discuss how to properly refactor this to avoid adding a new rocksdb vm

iovoid:
Superceded by #5145

PR #5075, opened 2025-10-27
chore(l1): fixed data path for mac

fedacking:
**Motivation**

The MacOS default path includes a space in the path for "Application Support". If a user in MacOS runs the tooling links will fail due to the space in the path, and the '\' will cause issues.

**Description**

- Changes data path to be only Library, a folder that is used for theses purposes.


tooling/sync/Makefile
```diff
@@ -179,10 +179,11 @@ start-ethrex: ## Start ethrex for the network given by NETWORK.
     		--p2p.port 30303\
     		--discovery.port 30303 \
     		--metrics \
+    		--log.level debug \
```
MegaRedHand: Was this intended?

tooling/sync/Makefile
```diff
@@ -14,7 +14,7 @@ CURRENT_DATETIME = $(shell date +'%y.%m.%d-%H.%M.%S')
 BATCH_SIZE ?= 1024
 OS = $(shell uname)
 ifeq ($(OS), Darwin)
-	DATA_PATH = $(HOME)/Library/Application\ Support
+	DATA_PATH = $(HOME)/Library/Application Support
```
MegaRedHand: Don't we need to add quotes around this too?

fedacking: No, this gets added to the variable in it's entirety

tooling/sync/Makefile
fedacking: Nope removed in [51e852a](https://github.com/lambdaclass/ethrex/pull/5075/commits/51e852a46a2357bc6301e9dca3eea45a98244485)

tooling/sync/Makefile
mpaulucci: Any reason you chose this specific path? I know it's the default, but maybe for tooling we can just choose a simpler path so we don't have to do this weird logic with having different paths depending on the OS

fedacking: Changed the DATA_PATH only Library to  simplify the tooling makefile in [6e417cc](https://github.com/lambdaclass/ethrex/pull/5075/commits/6e417cca6ae9db32668c54d398f03a76eccb75f9)

mpaulucci (approved):

PR #5077, opened 2025-10-28
chore(l1): improve snapsync log

edg-l:
image

**Motivation**

More compact but it retains the info and even added some more.

### DIFF
--- a/crates/networking/p2p/network.rs
+++ b/crates/networking/p2p/network.rs
@@ -263,8 +263,6 @@ pub async fn periodically_show_peer_stats_during_syncing(
             // We round up because of the accounts whose slots get downloaded and then not used
             let storage_leaves_inserted_percentage =
                 (storage_leaves_inserted_percentage * 10.0).round() / 10.0;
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/network.rs
```diff
@@ -253,15 +254,17 @@ pub async fn periodically_show_peer_stats_during_syncing(
...
-                METRICS.storage_leaves_inserted.get() as f64 / storage_leaves_downloaded as f64
-                    * 100.0
+                storage_leaves_inserted as f64 / storage_leaves_downloaded as f64 * 100.0
             } else {
                 0.0
             };
             // We round up because of the accounts whose slots get downloaded and then not used
             let storage_leaves_inserted_percentage =
                 (storage_leaves_inserted_percentage * 10.0).round() / 10.0;
+            let storage_leaves_pending =
```
fedacking: We can't accurately calculate this with the information provided, as the ratio of downloads to insertions isn't 1 to 1.

PR #5082, opened 2025-10-28
fix(l1): headers snapsync logs

fedacking:
**Motivation**

The logs for the headers in snapsync shows inaccurate numbers due to the chunked header download. This pr aims to fix that and show the proper percentage of headers to download.

**Description**

- Change `METRICS.headers_downloaded` to an `IntCounter` instead of `AtomicU64`
- Removed `METRICS.headers_to_download` and replaced the calculation with `METRICS.sync_head_block`
- Fixed `METRICS.sync_head_block` to include the proper 

Works to improve #4950


crates/networking/p2p/sync.rs
```diff
@@ -187,6 +187,7 @@ impl Syncer {
             .get_block_number(current_head)
             .await?
             .ok_or(SyncError::BlockNumber(current_head))?;
+
```
pablodeymo: this change is not needed

fedacking: Removed in [3be355d](https://github.com/lambdaclass/ethrex/pull/5082/commits/3be355d15922a432eec13e365f755b0d19006f3f)

jrchatruc (approved):

ElFantasma (approved):

PR #5083, opened 2025-10-28
fix(l2): remove one-time checkpoint if already exists

ilitteri:
**Motivation**

In a previous PR, DB checkpoints were introduced to ensure old state availability in the current path-based fashion. Every time a batch is sealed, a checkpoint whose state is the state of the latest block of the sealed batch is created to be used in the next batch.

The checkpoint is needed in two different steps of the batch commitment: for batch preparation (this is essentially building the batch) and for witness generation. Both steps need a non-modified checkpoint, but they both need to modify the checkpoint to be able to re-execute the batch.

As batch preparation occurs before witness generation, we opted to create a one-time checkpoint out of the main checkpoint that can be modified during batch preparation if needed (sometimes the batch was already available in the DB, and there's no need to re-execute anything); then, witness generation modifies the original checkpoint as needed because it is no longer needed. 

Once the one-time checkpoint fulfills its purpose, it is removed. Currently, if batch preparation fails, the one-time checkpoint is not removed, and after retrying batch preparation, there's another attempt at creating the one-time checkpoint, which ends in an error because the directory already exists. We need to either avoid creating the one-time checkpoint again or to remove the existing one.

**Description**

Remove the existing one-time checkpoint if it already exists.

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -285,6 +285,7 @@ impl L1Committer {
                 // struct, but we need to create a one-time copy of it because
                 // we still need to use the current checkpoint store later for witness
                 // generation.
+
                 let (one_time_checkpoint_store, one_time_checkpoint_blockchain) = self
                     .create_checkpoint(&self.current_checkpoint_store, &one_time_checkpoint_path)
                     .await?;
@@ -703,7 +704,6 @@ impl L1Committer {
             "temp_checkpoint_witness_{}_{rand_suffix}",
             batch.number
         ));
-
         // We need to create a one-time checkpoint copy because if witness generation fails the checkpoint would be modified
         let (_, one_time_checkpoint_blockchain) = self
             .create_checkpoint(&self.current_checkpoint_store, &one_time_checkpoint_path)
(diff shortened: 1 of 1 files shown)


crates/l2/sequencer/l1_committer.rs
```diff
@@ -395,6 +395,14 @@ impl L1Committer {
             .checkpoints_dir
             .join(format!("temp_checkpoint_batch_{batch_number}"));
 
+        if one_time_checkpoint_path.exists() {
```
avilagaston9: Maybe we can add a comment explaining why we do this

avilagaston9:
The PR was changed to use a random `one_time_checkpoint_path` on each attempt, preventing the following error:
```
2025-10-28T17:55:42.302454Z ERROR ethrex_l2::sequencer::l1_committer: L1 Committer Error: Committer failed retrieve block from storage: Failed to open RocksDB: IO error: lock hold by current process, acquire time 1761674125 acquiring thread 50: /root/.local/share/ethrex/temp_checkpoint_batch_1/LOCK: No locks available
```

crates/l2/sequencer/l1_committer.rs
```diff
@@ -278,8 +297,27 @@ impl L1Committer {
...
+                            let _ = remove_dir_all(&one_time_checkpoint_path);
+                        }
+                    })?;
+
+                if one_time_checkpoint_path.exists() {
+                    remove_dir_all(&one_time_checkpoint_path).map_err(|e| {
+                        CommitterError::FailedToCreateCheckpoint(format!(
+                            "Failed to remove one-time checkpoint directory {one_time_checkpoint_path:?}: {e}"
+                        ))
+                    })?;
```
tomip01: If we remove the `one_time_checkpoint_path` whether or not it returns an error, we can do it in one place

avilagaston9: Done [bd44892](https://github.com/lambdaclass/ethrex/pull/5083/commits/bd448927525b16867c182517f9411ed14d73478a)!

PR #5085, opened 2025-10-28
feat(l1): add fjall store backend

MegaRedHand:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,5 +8,9 @@
 
 <!-- Link to issues: Resolves #111, Resolves #222 -->
 
+**Checklist**
+
+- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync.
+
 Closes #issue_number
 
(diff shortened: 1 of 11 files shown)


crates/storage/backend/fjall.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    fn prefix_iterator(
+        &self,
+        table: &'static str,
+        prefix: &[u8],
+    ) -> Result<Box<dyn Iterator<Item = crate::api::PrefixResult> + '_>, StoreError> {
+        self.ro_tx.prefix_iterator(table, prefix)
+    }
+}
```
Oppen: Question: does this support read-your-own-write? I think this skips the batch. I'm not sure it's a pattern we use.

MegaRedHand: It indeed skips the batch. I just checked and we don't ever use that pattern. For now, I'm going to remove the requirement `StorageRoTx` in the `StorageRwTx` trait (in the original PR). We can reintroduce it if we deem it necessary.

PR #5086, opened 2025-10-28
feat(l2): reject privileged transactions in p2p messages

tomip01:
**Motivation**

We don't want to broadcast Privileged transactions between peers as every lead sequencer can get them through the OnChainProposer contract and look for the logs.

**Description**

The P2P messages that broadcast transactions are:
- `Transactions`
  - Added a new `matches!(..)` to deny them
- `NewPooledTransactionHashes`
  - Add a new check for sending tx hashes to not be a privileged one
- `GetPooledTransactions`
  - Are currently rejected inside the function `get_p2p_transaction_by_hash`
- `PooledTransactions`: the response form the `GetPooledTransactions`
  - Added a new `matches!(..)` inside the handle method to reject adding them to the mempool

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -42,6 +42,7 @@ pub enum Transaction {
 
 /// The same as a Transaction enum, only that blob transactions are in wrapped format, including
 /// the blobs bundle.
+/// PrivilegedL2Transaction is not included as it is not expected to be sent over P2P.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum P2PTransaction {
     LegacyTransaction(LegacyTransaction),


ilitteri (approved):

jrchatruc (approved):

crates/networking/p2p/rlpx/eth/transactions.rs
```diff
@@ -295,6 +295,13 @@ impl PooledTransactions {
                     continue;
                 }
             } else {
+                if matches!(tx, P2PTransaction::PrivilegedL2Transaction(_)) {
```
MegaRedHand: It seems weird that `P2PTransaction` has a privileged variant.

crates/networking/p2p/tx_broadcaster.rs
```diff
@@ -292,6 +292,13 @@ pub async fn send_tx_hashes(
...
+                if tx.is_privileged() {
+                    debug!(
+                        peer=%format!("{:#x}", peer_id),
+                        "Not sending privileged transaction hash to peer",
+                    );
+                    continue;
+                }
                 txs_to_send.push((**tx).clone());
             }
             let hashes_message = Message::NewPooledTransactionHashes(
```
MegaRedHand, line 300: Now we need to check that we don't send an empty message. We should see if we can filter transactions before chunking them, to avoid this and also keep the chunk size consistent.

tomip01: You are right! Missed that. Changed that here [filter maps before chunking](https://github.com/lambdaclass/ethrex/pull/5086/commits/f27c4448a7eecf9d2de2124dc08597c69265db0c)

crates/networking/p2p/rlpx/eth/transactions.rs
tomip01: Removed that one here [02308eb](https://github.com/lambdaclass/ethrex/pull/5086/commits/02308eb9feb9d946f7f1f796f8925ec8d79cfd99)

crates/common/types/transaction.rs
```diff
@@ -49,7 +49,6 @@ pub enum P2PTransaction {
     EIP1559Transaction(EIP1559Transaction),
     EIP4844TransactionWithBlobs(WrappedEIP4844Transaction),
     EIP7702Transaction(EIP7702Transaction),
-    PrivilegedL2Transaction(PrivilegedL2Transaction),
```
MegaRedHand, line 52: We should probably document this change somewhere.

tomip01: Done [63675e2](https://github.com/lambdaclass/ethrex/pull/5086/commits/63675e28d1aa971c2d533b9fd70e2a4039714679)

PR #5093, opened 2025-10-29
fix(l1): fix exponential overflow in fake exponential

fedacking:
**Motivation**

Fake Exponential defined in `block.rs` can overflow. As such, we need to use the U256 data type to avoid overflows

**Description**

- Added intermediate conversions to U256 to avoid overflows
- Added test for the overflow

### DIFF
--- a/crates/common/types/block.rs
+++ b/crates/common/types/block.rs
@@ -25,7 +25,6 @@ use rkyv::{Archive, Deserialize as RDeserialize, Serialize as RSerialize};
 use serde::{Deserialize, Serialize};
 
 use std::cmp::{Ordering, max};
-use std::u64;
 
 pub type BlockNumber = u64;
 pub type BlockHash = H256;


crates/common/types/block.rs
```diff
@@ -398,14 +398,14 @@ pub fn calculate_base_fee_per_blob_gas(parent_excess_blob_gas: u64, update_fract
 // Defined in [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844)
 pub fn fake_exponential(factor: u64, numerator: u64, denominator: u64) -> u64 {
     let mut i = 1;
-    let mut output = 0;
-    let mut numerator_accum = factor * denominator;
-    while numerator_accum > 0 {
+    let mut output = U256::zero();
+    let mut numerator_accum = (U256::zero() + factor) * denominator;
```
MegaRedHand:
```suggestion
    let mut numerator_accum = U256::from(factor) * denominator;
```

crates/common/types/block.rs
```diff
@@ -398,14 +398,14 @@ pub fn calculate_base_fee_per_blob_gas(parent_excess_blob_gas: u64, update_fract
...
+    while numerator_accum > U256::zero() {
```
MegaRedHand: Since the integer is unsigned, this could be changed to:

```suggestion
    while !numerator_accum.is_zero() {
```

crates/common/types/block.rs
```diff
@@ -398,14 +398,14 @@ pub fn calculate_base_fee_per_blob_gas(parent_excess_blob_gas: u64, update_fract
...
         output += numerator_accum;
         numerator_accum = numerator_accum * numerator / (denominator * i);
         i += 1;
     }
-    output / denominator
+    (output / denominator).as_u64()
```
MegaRedHand: We should check if this fits inside a `u64`

crates/common/types/block.rs
```diff
@@ -983,4 +983,10 @@ mod test {
         let res = calc_excess_blob_gas(&parent, schedule, fork);
         assert_eq!(res, 3538944)
     }
+
+    #[test]
+    fn test_fake_exponential_overflow() {
+        // With u64 this overflows
+        fake_exponential(57532635, 3145728, 3338477);
+    }
```
MegaRedHand, line 995: Let's add a test with the maximum expected inputs here, for good measure:

```rust
fake_exponential(MIN_BASE_FEE_PER_BLOB_GAS, u64::MAX, BLOB_BASE_FEE_UPDATE_FRACTION);

File: crates/common/types/block.rs (hunk 1 above)
Comment (fedacking) [reply crates/common/types/block.rs]: Done in [7b7ec23](https://github.com/lambdaclass/ethrex/pull/5093/commits/7b7ec235b1702916f6ee75b85005e2fad7ed3f1a)

File: crates/common/types/block.rs (hunk 2 above)
Comment (fedacking) [reply crates/common/types/block.rs]: Done in [7b7ec23](https://github.com/lambdaclass/ethrex/pull/5093/commits/7b7ec235b1702916f6ee75b85005e2fad7ed3f1a)

File: crates/common/types/block.rs (hunk 3 above)
Comment (fedacking) [reply crates/common/types/block.rs]: Done in [7b7ec23](https://github.com/lambdaclass/ethrex/pull/5093/commits/7b7ec235b1702916f6ee75b85005e2fad7ed3f1a)

File: crates/common/types/block.rs [hunk 5]
```diff
@@ -398,14 +398,18 @@ pub fn calculate_base_fee_per_blob_gas(parent_excess_blob_gas: u64, update_fract
...
+    let mut output = U256::zero();
+    let mut numerator_accum = U256::from(factor) * denominator;
+    while !numerator_accum.is_zero() {
         output += numerator_accum;
         numerator_accum = numerator_accum * numerator / (denominator * i);
         i += 1;
     }
-    output / denominator
+    if (output / denominator) > U256::from(u64::MAX) {
+        u64::MAX
```
MegaRedHand, line 409: We should panic here, since I'm sure this is unreachable. In a later PR we can update this with an explanation.

crates/common/types/block.rs
MegaRedHand: This test panics.

fedacking: Created issue to define a bound #5096

jrchatruc (approved):

PR #5106, opened 2025-10-29
fix(l2): fix `ethrex-l2-common` crate compilation

ilitteri:
**Motivation**

`ethrex-l2-common` needs for the SP1 and RISC0 guests' ELF to exist to compile. `guest_program` generates said ELFs in its `build.rs`, but as `ethrex-l2-common`, it's compiled before the `build.rs` runs, causing a compilation error.

**Description**

- Embed SP1 guest program in the `guest_program` crate.
- Move logic that needs said ELFs to the sequencer (who isn't a dependency from the `guest_program` crate).
- Remove `sp1` and `risc0` features from `ethrex-l2-common` as they are not needed anymore.

### DIFF
--- a/crates/l2/sequencer/l1_proof_sender.rs
+++ b/crates/l2/sequencer/l1_proof_sender.rs
@@ -15,7 +15,7 @@ use ethrex_rpc::{
     clients::{EthClientError, eth::errors::EstimateGasError},
 };
 use ethrex_storage_rollup::StoreRollup;
-use guest_program::{ZKVM_SP1_PROGRAM_ELF, methods::ZKVM_RISC0_PROGRAM_ID};
+use guest_program::{ZKVM_RISC0_PROGRAM_VK, ZKVM_SP1_PROGRAM_ELF};
 use serde::Serialize;
 use spawned_concurrency::tasks::{
     CallResponse, CastResponse, GenServer, GenServerHandle, send_after,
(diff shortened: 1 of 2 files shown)


crates/l2/common/src/prover.rs
```diff
@@ -98,29 +84,11 @@ impl ProverType {
...
-            // other types are not supported by Aligned
-            _ => Ok(None),
-        }
-    }
-
     /// Gets the verification key or image id for this prover backend, used for
     /// proof verification. Aligned Layer uses a different vk in SP1's case.
+    ///
+    /// There's no need to have the following verifying keys here since
+    /// they are used by the deployer and they are not used by the sequencer.
```
ManuelBilbao: Can't we delete this so?

ilitteri: I think the comment is misleading, that functionality is actually used by the deployer. Unless you meant to remove that from there and use it directly in the deployer bin.

ManuelBilbao: Yes, I the comment is confusing too. Maybe something like ...?
> This function can be moved since it's used by the deployer but not the sequencer

cmd/ethrex/Cargo.toml
```diff
@@ -108,9 +108,9 @@ l2 = [
   "dep:hex",
 ]
 l2-sql = ["ethrex-storage-rollup/sql"]
-sp1 = ["ethrex-prover/sp1", "ethrex-l2-common/sp1"]
+sp1 = ["ethrex-prover/sp1"]
```
tomip01: Should this enable `ethrex-l2/sp1`?

ilitteri: Good catch.

crates/l2/common/src/prover.rs
ilitteri: I'll move the functionality there then.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -246,10 +247,30 @@ impl L1ProofSender {
...
-                )));
+            let vm_program_code = match prover_type {
+                ProverType::RISC0 => {
+                    if !cfg!(feature = "risc0") {
+                        return Err(ProofSenderError::UnexpectedError(
+                            "Trying to send RISC0 proof but RISC0 feature is disabled".to_string(),
+                        ));
+                    }
+
+                    ZKVM_RISC0_PROGRAM_ELF.to_vec()
```
gianbelinche: This shouldn't be the elf, but the vk

PR #5110, opened 2025-10-29
fix: stop payload building task when getPayload is received

### LINKED ISSUE #5011: Current block building task keeps on working when `getPayload` is received
After #4999 is merged, we no longer timeout when `getPayload` is received; however, we still leave the block building task dangling. This is not a problem, but it's a waste of CPU resources. We should add some mechanism to stop this process prematurely, maybe as part of #4998.

Lil-Duckling-22:
## Motivation

Currently, when a `getPayload` request is received, the payload building task continues running in the background even after the `CancellationToken` is cancelled, wasting CPU resources. This happens because `run_until_cancelled` only waits for cancellation but doesn't actually stop the running `spawn_blocking` task.

## Description

This PR fixes payload building task cancellation by:

- Adding `AbortHandle` to `PayloadBuildTask` for task termination
- Replacing `run_until_cancelled` with `tokio::time::timeout` to check cancellation every 100ms
- Using `abort_handle.abort()` to forcefully stop building tasks when cancellation is requested

Eliminates unnecessary CPU usage when `getPayload` is called.

Closes #5011

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -49,6 +49,7 @@ use tracing::{debug, error, warn};
 pub struct PayloadBuildTask {
     task: tokio::task::JoinHandle<Result<PayloadBuildResult, ChainError>>,
     cancel: CancellationToken,
+    abort_handle: tokio::task::AbortHandle,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


MegaRedHand:
Hello, thanks for the contribution! I'm closing this PR since it doesn't fix the issue. [According to the tokio docs](https://docs.rs/tokio/latest/tokio/task/struct.AbortHandle.html), calling `abort()` on a task spawned using `spawn_blocking` has no effect:

> Be aware that tasks spawned using [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) cannot be aborted because they are not async. If you call abort on a spawn_blocking task, then this will not have any effect, and the task will continue running normally. The exception is if the task has not started running yet; in that case, calling abort may prevent the task from starting.

PR #5123, opened 2025-10-30
fix(l2): checkpoint initialization

avilagaston9:
**Motivation**

Our current checkpoint initialization doesn't work after node restarts. We reuse the initial checkpoint if it already exists, but on subsequent runs this checkpoint can be outdated.

**Description**

- Fixes issues encountered when restarting the node.
- Moves all checkpoint-related logic to the `l1_committer`, as it is the only component that needs it.
- Now, the `l1_committer`:
    - Fetches the needed checkpoint based on the current batch.
    - If the checkpoint does not exist, creates a new store with the genesis state (this should only happen on the very first run).

Closes None

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -818,14 +818,9 @@ impl L1Committer {
             return Ok(());
         }
 
-        self.store.create_checkpoint(&new_checkpoint_path).await?;
-        let (new_checkpoint_store, new_checkpoint_blockchain) = Self::get_checkpoint_from_path(
-            self.genesis.clone(),
-            self.blockchain.options.clone(),
-            &new_checkpoint_path,
-            &self.rollup_store,
-        )
-        .await?;
+        let (new_checkpoint_store, new_checkpoint_blockchain) = self
+            .create_checkpoint(&self.store, &new_checkpoint_path, &self.rollup_store)
+            .await?;
 
         self.current_checkpoint_store = new_checkpoint_store;
 


crates/l2/sequencer/l1_committer.rs
```diff
@@ -827,15 +849,15 @@ impl L1Committer {
             let mut checkpoint_store_inner = Store::new(path, engine_type)?;
 
             checkpoint_store_inner
-                .add_initial_state(self.genesis.clone())
+                .add_initial_state(genesis.clone())
```
ManuelBilbao: Think this clone is not needed

crates/l2/sequencer/l1_committer.rs
```diff
@@ -788,14 +789,15 @@ impl L1Committer {
         // Sometimes the commit_next_batch task is retried after a failure, and in
         // that case we would try to create a checkpoint again at the same path,
         // causing a lock error under rocksdb feature.
-        if new_checkpoint_path.exists() {
-            debug!("Checkpoint at path {new_checkpoint_path:?} already exists, skipping creation");
-            return Ok(());
```
avilagaston9, line 793: If the checkpoint already exists, due to a previous restart, we still need to update `self.current_checkpoint`

crates/l2/sequencer/l1_committer.rs
avilagaston9: Nice! [7a2ebef](https://github.com/lambdaclass/ethrex/pull/5123/commits/7a2ebef58f7da7f11bf4032df04e1542f18b3c5f)

crates/l2/sequencer/l1_committer.rs
```diff
@@ -815,30 +845,51 @@ impl L1Committer {
...
 
             checkpoint_store_inner
         };
 
-        let checkpoint_blockchain = Arc::new(Blockchain::new(
-            checkpoint_store.clone(),
-            self.blockchain.options.clone(),
-        ));
+        // override blockchain options as the FeeConfig is shared under an ArcMutex
+        blockchain_opts.r#type = BlockchainType::L2(L2Config::default());
```
ilitteri, line 885: Is it ok to use `L2Config::default()` here? Shouldn't we inherit the main blockchain's `L2Config`?

avilagaston9: Yes. Inside `regenerate_head_state` we fetch the proper `fee_config` for each block.

ilitteri: I got it, we don't want to use the same `Arc<Mutex>`, so we create a new one. We should detail a little bit more the comment.

avilagaston9: Done [a2efa36](https://github.com/lambdaclass/ethrex/pull/5123/commits/a2efa36a35c64993ac8fcac3e20ea801121e77ba)!

ilitteri (approved):

PR #5133, opened 2025-10-30
chore(l2): use SP1 patched crate for ecmul precompile

xqft:
**Motivation**

The ecmul precompile represents 10% of the total proving cycles, and we are not using the corresponding SP1 patched crate for this operation that uses a zkVM precompile. 

This PR fixes that and reduces 81k cycles (86% of ecmul, 10% of total)

**Flamegraphs**
before:
image

after:
image

**Proving times**
Mainnet blocks proven in a RTX 4090
| Block number | main    | l2/ecmul_sp1 |
|---------------|---------|----------------|
| 23426993      | 13m 19s | 13m 20s        |
| 23426994      | 08m 42s | 08m 05s        |
| 23426995      | 06m 39s | 06m 35s        |
| 23426996      | 11m 24s | 10m 50s        |


crates/vm/levm/src/precompiles.rs
```diff
@@ -747,6 +748,36 @@ pub fn ecmul(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<B
     Ok(Bytes::from(out))
 }
 
+#[cfg(feature = "sp1")]
```
ilitteri: We should include `risc0` here now that it uses its own `substrate-bn` patch.

crates/vm/levm/src/precompiles.rs
```diff
@@ -706,21 +706,22 @@ pub fn ecadd(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<B
...
+    ) else {
+        return Err(InternalError::Slicing.into());
+    };
+    validate_bn254_g1_coords(&g1)?;
+    bn254_g1_mul(g1, scalar)
+}
 
-    let x = ark_bn254::Fq::from_be_bytes_mod_order(point_x);
-    let y = ark_bn254::Fq::from_be_bytes_mod_order(point_y);
+#[cfg(not(feature = "sp1"))]
```
ilitteri: Ditto.

crates/vm/levm/src/precompiles.rs
```diff
@@ -816,12 +867,12 @@ pub fn ecpairing(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
 
     let mut batch = Vec::new();
     for input in calldata.chunks_exact(192) {
-        #[expect(unsafe_code, reason = "chunks_exact ensures the conversion is valid")]
-        let input: [u8; 192] = unsafe { input.try_into().unwrap_unchecked() };
-        let (g1, g2) = parse_bn254_coords(&input);
-        if validate_bn254_coords(&g1, &g2)? {
-            batch.push((g1, g2));
-        }
```
ilitteri, line 888: Why do we remove this `unsafe`? We should double-check with @edg-l, @azteca1998, or @Oppen  if this is not counterproductive.

ilitteri: Can't we keep this behaviour when neither `sp1` nor `risc0` features are enabled?

ilitteri (changes requested):
The implementation looks good. I'll run some replays to ensure it works properly. Left some comments.

crates/vm/levm/src/precompiles.rs
xqft: The function to parse bn254 points was changed to take a slice instead of a reference to a fixed size array (so it can be reused for precompiles whose input have different sizes), so this unsafe is not necessary anymore.

The branching on the error case (which should never happen) could signify a perf. hit, I could change this to an `unreachable_unchecked!()` and it should be equivalent to before the change, but I'm not sure how impactful that can be (consider that we have other, more evident optimization opportunities, like not cloning and slicing bytes many times while parsing and converting to the crypto primitive types)

crates/vm/levm/src/precompiles.rs
xqft: done!

crates/vm/levm/src/precompiles.rs
xqft: done!

ilitteri (approved):

pablodeymo (approved):

crates/vm/levm/src/precompiles.rs
```diff
@@ -706,21 +706,23 @@ pub fn ecadd(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<B
...
+        return Err(InternalError::Slicing.into());
+    };
+    validate_bn254_g1_coords(&g1)?;
+    bn254_g1_mul(g1, scalar)
+}
 
-    let x = ark_bn254::Fq::from_be_bytes_mod_order(point_x);
-    let y = ark_bn254::Fq::from_be_bytes_mod_order(point_y);
+#[cfg(all(not(feature = "sp1"), not(feature = "risc0")))]
+#[inline]
```
```suggestion
#[inline(always)]
```

crates/vm/levm/src/precompiles.rs
```diff
@@ -747,6 +749,41 @@ pub fn ecmul(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Result<B
     Ok(Bytes::from(out))
 }
```
```suggestion
#[inline(always)]
```

PR #5137, opened 2025-10-31
feat(l1): omit copying bytes when encoding transactions

fmoletta:
**Motivation**
When encoding non-legacy transactions we need to encode the payload (txType || rlp(Transaction) as a bytes object. In order to do so we copy the payload to a `Bytes` object and then encode it in rlp, this is not needed as we can just encode the payload as bytes by invoking the implementation for [u8] directly

**Description**
* Avoid using `Bytes::copy_from_slice` when encoding transactions


crates/common/types/transaction.rs
```diff
@@ -395,7 +395,7 @@ impl RLPEncode for Transaction {
     fn encode(&self, buf: &mut dyn bytes::BufMut) {
         match self {
             Transaction::LegacyTransaction(t) => t.encode(buf),
-            tx => Bytes::copy_from_slice(&tx.encode_canonical_to_vec()).encode(buf),
+            tx => <[u8] as RLPEncode>::encode(&tx.encode_canonical_to_vec(), buf),
```
MegaRedHand, line 398: Let's add a unit test for RLP encoding/decoding of transactions first.

fmoletta: We already have `eip1559_tx_rlp_decode` & `legacy_tx_rlp_decode`

PR #5139, opened 2025-10-31
fix(l1): rlp decode impls for low size transactions

fmoletta:
**Motivation**

Transactions are encoded in the following format:

* Legacy:  rlp(Transaction) 
* Non Legacy: rlp(Payload) where Payload is a bytes object containing (TxType || rlp(Transaction))

When decoding in order to differentiate between a legacy and a non legacy transaction we check if the encoded data is a bytes object (`is_encoded_as_bytes`) by checking if the prefix is between 0xb8 and 0xbf. The problem with this is that when encoding the payload the 0xb8..0xbf prefix is only applied if the length of the payload is higher than 56 bytes. This is usually the case for most real-scenario transactions, but if the transaction payload were to be lower than 56 bytes then we would not detect it as a bytes object and would attempt and fail to decode the transaction as a legacy one.

This PR fixes this problem by:

* Changing the criteria for legacy vs non-legacy transactions to instead check if the incoming encoded data is encoded as a list (if so it will be legacy). Aka `is_encoded_as_bytes` -> `!is_encoded_as_list`
* Considering the case where the encoded payload doesn't have the bytes prefix. Aka `get_rlp_bytes_item_payload` -> `decode_rlp_item`. The latter which considers the prefix RLP_NULL+size that is used when encoding less than 56 bytes

**Description**
* Fix decoding for `Transaction` & `P2PTransaction` for cases where tx payload encoding is less than 56 bytes

Closes #issue_number

### DIFF
--- a/crates/common/types/receipt.rs
+++ b/crates/common/types/receipt.rs
@@ -227,6 +227,8 @@ impl RLPDecode for ReceiptWithBloom {
     /// A) Legacy receipts: rlp(receipt)
     /// B) Non legacy receipts: rlp(Bytes(tx_type | rlp(receipt))).
     fn decode_unfinished(rlp: &[u8]) -> Result<(Self, &[u8]), RLPDecodeError> {
+        // The minimum size for a ReceiptWithBloom is > 256 bytes (due to the Bloom type field) meaning that it is safe
+        // to check for bytes prefix to diferenticate between legacy receipts and non-legacy receipt payloads
         let (tx_type, rlp) = if is_encoded_as_bytes(rlp)? {
             let payload = get_rlp_bytes_item_payload(rlp)?;
             let tx_type = match payload.first().ok_or(RLPDecodeError::InvalidLength)? {


crates/common/types/transaction.rs
```diff
@@ -406,9 +406,9 @@ impl RLPDecode for Transaction {
     /// B) Non legacy transactions: rlp(Bytes) where Bytes represents the canonical encoding for the transaction as a bytes object.
     /// Checkout [Transaction::decode_canonical] for more information
     fn decode_unfinished(rlp: &[u8]) -> Result<(Self, &[u8]), RLPDecodeError> {
-        if is_encoded_as_bytes(rlp)? {
+        if !is_encoded_as_list(rlp)? {
```
MegaRedHand: This may be the case for `P2PTransaction` too.

crates/common/rlp/decode.rs
```diff
@@ -492,13 +492,31 @@ pub fn is_encoded_as_bytes(rlp: &[u8]) -> Result<bool, RLPDecodeError> {
...
+
 /// Receives an RLP bytes item (prefix between 0xb8 and 0xbf) and returns its payload
 pub fn get_rlp_bytes_item_payload(rlp: &[u8]) -> Result<&[u8], RLPDecodeError> {
     let prefix = rlp.first().ok_or(RLPDecodeError::InvalidLength)?;
     let offset: usize = (prefix - 0xb8 + 1).into();
     rlp.get(offset + 1..).ok_or(RLPDecodeError::InvalidLength)
 }
 
+/// Receives an RLP item (either string (prefix between 0x80 and 0xB8 or bytes (prefix between 0xb8 and 0xbf)) and returns its payload
+pub fn get_rlp_item_payload(rlp: &[u8]) -> Result<&[u8], RLPDecodeError> {
```
MegaRedHand: I think we can use `decode_rlp_item` here, instead of reimplementing the decoding.

crates/common/types/transaction.rs
fmoletta: Yes, it is also implemented there

crates/common/rlp/decode.rs
fmoletta: Updated: [34ab4b1](https://github.com/lambdaclass/ethrex/pull/5139/commits/34ab4b16f953778e51ee62384dd44f895d1d357e)

crates/common/types/transaction.rs
```diff
@@ -406,31 +405,30 @@ impl RLPDecode for Transaction {
     /// B) Non legacy transactions: rlp(Bytes) where Bytes represents the canonical encoding for the transaction as a bytes object.
     /// Checkout [Transaction::decode_canonical] for more information
     fn decode_unfinished(rlp: &[u8]) -> Result<(Self, &[u8]), RLPDecodeError> {
-        if is_encoded_as_bytes(rlp)? {
-            // Adjust the encoding to get the payload
-            let payload = get_rlp_bytes_item_payload(rlp)?;
```
MegaRedHand, line 411: I think this pattern is also used for receipt decoding. Maybe we have the same bug there too.

MegaRedHand (approved):
LGTM

crates/common/types/transaction.rs
fmoletta: Addressed [2548ae7](https://github.com/lambdaclass/ethrex/pull/5139/commits/2548ae7bc08e65f373e8311667c86f56a791c0a6)

Oppen (approved):

PR #5143, opened 2025-10-31
fix(l1): fix illegal instructions in certain ARM processors

fedacking:
**Motivation**

During ARM testing, ethrex would crash on certain ARM CPUs. The backtrace of the program was in the blake2b precompile in `crates/common/crypto/blake2f/aarch64.rs`. The specific instruction that crashed was the [XAR instruction](https://developer.arm.com/documentation/ddi0602/2022-03/SIMD-FP-Instructions/XAR--Exclusive-OR-and-Rotate-), which the documentation states that "This instruction is implemented only when FEAT_SHA3 is implemented."

**Description**

- Added check to `BLAKE2_FUNC` lazy evaluator to check for the `sha3` feature in the user's cpu at runtime.

**Notes on testing**

On an ARM CPU without the feature `sha3` the commit `fba993c9b337ab2a89664f0f39260b40e25ac4c8` includes a failing unit test. After `21fb29e439261de5c9f275f54a9de218759daa2f` the test should work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 2025-10-28
 
 - Batch BlobsBundle::validate [#4993](https://github.com/lambdaclass/ethrex/pull/4993)
+- Remove latest_block_header lock [#5050](https://github.com/lambdaclass/ethrex/pull/5050)
 
 ### 2025-10-27
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10333,17 +10333,14 @@ dependencies = [
 [[package]]
 name = "substrate-bn"
 version = "0.6.0"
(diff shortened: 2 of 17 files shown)


crates/common/crypto/blake2f/mod.rs
```diff
@@ -25,3 +27,34 @@ static BLAKE2_FUNC: LazyLock<Blake2Func> = LazyLock::new(|| {
 pub fn blake2b_f(rounds: usize, h: &mut [u64; 8], m: &[u64; 16], t: &[u64; 2], f: bool) {
     BLAKE2_FUNC(rounds, h, m, t, f)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn test_12r() {
```
Oppen: 12r?

Maybe call it `blake2b_smoke`.

fedacking: Done in [163a164](https://github.com/lambdaclass/ethrex/pull/5143/commits/163a164df24a04e2c4113b66ad74b6596eba12b5)

PR #5145, opened 2025-10-31
perf(l1): avoid opening tries

iovoid:
**Motivation**

Currently to get a storage we are:

* getting the block header to obtain the state root
* getting the account to get the storage root
* reading the value

The first step isn't needed because the state root is the same through the block, and the second isn't needed because since we're using a path-based DB we can directly get the value (without even using the trie, if the FKVs are computed).

**Description**

We make the VM database ask by state_root instead of block hash (skipping step 1), and we skip reading the account storage root when the FlatKeyValues are computed for it.

### DIFF
--- a/crates/storage/store_db/in_memory.rs
+++ b/crates/storage/store_db/in_memory.rs
@@ -739,10 +739,6 @@ impl StoreEngine for Store {
         // Silently ignoring the request to create a checkpoint is harmless
         Ok(())
     }
-
-    fn flatkeyvalue_computed(&self, _account: H256) -> Result<bool, StoreError> {
-        Ok(false)
-    }
 }
 
 impl Debug for Store {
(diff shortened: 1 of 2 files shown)


crates/storage/store_db/in_memory.rs
```diff
@@ -739,6 +739,10 @@ impl StoreEngine for Store {
         // Silently ignoring the request to create a checkpoint is harmless
         Ok(())
     }
+
+    fn flatkeyvalue_computed(&self, _account: H256) -> Result<bool, StoreError> {
+        Ok(false)
+    }
```
pablodeymo: I'd prefer to have a default implementation in the trait declaration.

jrchatruc (approved):

crates/storage/store_db/in_memory.rs
iovoid: Changed in d4234f47

PR #5147, opened 2025-10-31
perf(l1): improved discovery and peer initialization

ElFantasma:
**Motivation**

We are too aggressive on our DiscoveryServer and Initiator. On start we send up to 5000 Udp messages every 5 seconds and start up to 5000 Tcp connections every 3 seconds.
After several tests it was found that on one hand it is not necessary to be that eager, and on the other hand it caused several issues regarding the tokio runtime (ie. it was blocked for more than 10ms in several places, causing some problems when trying to synchronize from the local machine).

**Description**

This PR changes the way Lookup is performed: instead of trying a lot of connections at the same time every 3 seconds, it distributes the connection attempts in time, making it more gentle with the CPU. It also lowers the attempt rate.

- Our Discovery lookup is more frequent, but tries 1 node at a time.
- Our Connection attempt is more frequent, but tries 1 node at a time.
- We match both processes rate to avoid having both, too much pending connections, or no pending connections at all.
- Also noticed that spawning some GenServers in dedicated threads was no longer necessary, and it created a lot of threads, as every `start_on_thread` creates it's own Tokio Runtime with it's thread pool.
- Also removed a spawn_blocking inside a GenServer that may cause a contention issue.

** Pending (for a future PR) **

- Will create a command line argument to set up the lookup frequency or eagerness: #5148
- Will make the lookup frequency adjustable at runtime, depending on the amount of connected peers: #5149

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -180,7 +180,7 @@ impl L1Committer {
             Self::get_checkpoint_from_path(
                 genesis.clone(),
                 blockchain.options.clone(),
-                &checkpoints_dir.join(format!("checkpoint_batch_{}", last_committed_batch)),
+                &checkpoints_dir.join(batch_checkpoint_name(last_committed_batch)),
                 &rollup_store,
             )
             .await?;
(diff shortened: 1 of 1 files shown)


jrchatruc (approved):

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -127,7 +126,7 @@ pub struct PeerTable {
 impl PeerTable {
     pub fn spawn(target_peers: usize) -> PeerTable {
         PeerTable {
-            handle: PeerTableServer::new(target_peers).start_on_thread(),
+            handle: PeerTableServer::new(target_peers).start(),
```
ElFantasma, line 129: Reverted this as it was shown it is not affecting performance and avoids creating a new Tokio runtime for the new thread

crates/networking/p2p/discv4/server.rs
```diff
@@ -111,7 +115,7 @@ impl DiscoveryServer {
             .new_contacts(bootnodes, local_node.node_id())
             .await?;
 
-        discovery_server.start_on_thread();
+        discovery_server.start();
```
ElFantasma, line 120: Reverted this as it was shown it is not affecting performance and avoids creating a new Tokio runtime for the new thread

crates/networking/p2p/rlpx/initiator.rs
```diff
@@ -26,40 +31,32 @@ pub struct RLPxInitiator {
...
-            new_connections_per_lookup: 5000,
         }
     }
 
     pub async fn spawn(context: P2PContext) {
         info!("Starting RLPx Initiator");
         let state = RLPxInitiator::new(context);
-        let mut server = RLPxInitiator::start_on_thread(state.clone());
-        let _ = server.cast(InMessage::LookForPeers).await;
+        let mut server = RLPxInitiator::start(state.clone());
```
ElFantasma, line 50: Reverted this as it was shown it is not affecting performance and avoids creating a new Tokio runtime for the new thread

crates/networking/p2p/discv4/server.rs
```diff
@@ -80,6 +82,7 @@ pub struct DiscoveryServer {
     signer: SecretKey,
     udp_socket: Arc<UdpSocket>,
     peer_table: PeerTable,
+    find_node_message: BytesMut,
```
MegaRedHand, line 87:
```suggestion
    /// The last `FindNode` message sent, cached due to message
    /// signatures being expensive.
    find_node_message: BytesMut,
```

PR #5151, opened 2025-11-01
ci(l2): do not run pr-main_l2 workflow in the merge queue

ilitteri:
**Motivation**

The merge queue was stuck waiting for `Integration Test L2` and `Lint L2` jobs to pass. But the pr-main_l2 workflow is not triggered in the merge queue, so there isn't any job running in the merge queue to meet these GHA requirements.

**Description**

Trigger pr-main_l2 workflow in the merge queue, but skip every job to meet the GHA requirements.


ManuelBilbao (approved):

gianbelinche:
You can also do it by modifying the if at the detect changes job, instead of adding a new check in every job.

Issue #5154, opened 2025-11-01
Handle error after sealing batch and before generating witness


ManuelBilbao:
After #5152, there's an edge case where the batch is sealed and saved in the database but the witness generations failed, causing an inconsistent state

PR #5155, opened 2025-11-01
fix(2): revert sealed batch if witness generation fails

Galoretka:
Fix inconsistent state where a batch was sealed and stored but witness generation failed after #5152; in crates/l2/sequencer/l1_committer.rs wrap generate_and_store_batch_prover_input with error handling and call rollup_store.revert_to_batch(batch.number - 1) on failure to rollback the just-sealed batch, keeping DB and internal checkpoint state consistent; on success behavior is unchanged; consider future improvements like sealing only after witness generation or using a single transactional boundary for batch + witness persistence.

#5154

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -410,7 +410,10 @@ impl L1Committer {
             batch.number,
         );
 
-        self.generate_and_store_batch_prover_input(&batch).await?;
+        if let Err(e) = self.generate_and_store_batch_prover_input(&batch).await {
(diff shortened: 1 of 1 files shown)


ElFantasma (commented):
This PR is **stale and cannot merge** — the codebase has been significantly refactored since Nov 2025. The function `generate_and_store_batch_prover_input` no longer exists on `main`; batch sealing and witness generation now happen together in `produce_batch` via `seal_batch_with_prover_input`, and the `send_commitment` path only handles already-sealed batches. The fix intent (revert on witness failure) is sound but needs to be re-implemented against the current architecture.

PR #5157, opened 2025-11-03
perf(l1): optimize Nibbles::skip_prefix to avoid tail allocation and remove redundant clone

GarmashAlex:
**Motivation**

Minor inefficiencies: unnecessary clone() in get_path() and allocations in Nibbles::skip_prefix()

**Description**

- Replace self.child.clone().compute_hash() with self.child.compute_hash() in ExtensionNode::get_path() to avoid an unnecessary Arc clone. NodeRef::compute_hash(&self) already memoizes and requires no ownership.
- Reimplement Nibbles::skip_prefix() using drain(..plen) to move the matched prefix into already_consumed without allocating a new Vec for the remaining tail. Semantics are preserved: data becomes the same tail and already_consumed gains the same bytes previously appended from prefix.data.

### DIFF
--- a/crates/common/trie/node/extension.rs
+++ b/crates/common/trie/node/extension.rs
@@ -228,7 +228,7 @@ impl ExtensionNode {
             let child_node = self.child.get_node(db, path.current())?.ok_or_else(|| {
                 TrieError::InconsistentTree(Box::new(
                     InconsistentTreeError::ExtensionNodeChildNotFound(ExtensionNodeErrorData {
-                        node_hash: self.child.clone().compute_hash().finalize(),
+                        node_hash: self.child.compute_hash().finalize(),
                         extension_node_hash: self.compute_hash().finalize(),
                         extension_node_prefix: self.prefix.clone(),
                         node_path: path.current(),


crates/common/trie/nibbles.rs
```diff
@@ -93,9 +93,10 @@ impl Nibbles {
     /// If `prefix` is a prefix of self, move the offset after
     /// the prefix and return true, otherwise return false.
     pub fn skip_prefix(&mut self, prefix: &Nibbles) -> bool {
-        if self.len() >= prefix.len() && &self.data[..prefix.len()] == prefix.as_ref() {
-            self.data = self.data[prefix.len()..].to_vec();
-            self.already_consumed.extend(&prefix.data);
+        let plen = prefix.len();
+        if self.len() >= plen && &self.data[..plen] == prefix.as_ref() {
+            let consumed: Vec<u8> = self.data.drain(..plen).collect();
```
ElFantasma, line 98: nit: the intermediate `Vec<u8>` from `.collect()` can be avoided by extending directly from the drain iterator:
```rust
self.already_consumed.extend(self.data.drain(..plen));
```
`Vec::extend` accepts any `IntoIterator<Item = u8>`, so `Drain` works directly — this avoids the temporary allocation.

ElFantasma:
I'm closing this as it had no response to the feedback and another [PR](https://github.com/lambdaclass/ethrex/pull/6190) is addressing the same issue.

Feel free to reopen in case this work needs to be resumed.

PR #5158, opened 2025-11-03
perf(levm): improve compatibility of `blake2b`'s NEON implementation

azteca1998:
**Motivation**

The current implementation requires CPUs to implement SHA3 feat.

**Description**

Make it so that NEON can still be used when there's no `XAR` instruction available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-11-03
+
+- Improve `blake2b` NEON implementation compatibility [#5158](https://github.com/lambdaclass/ethrex/pull/5158)
+
 ### 2025-10-31
 
 - Improved discovery and peer initialization [#5147](https://github.com/lambdaclass/ethrex/pull/5147)


crates/common/crypto/blake2f/mod.rs
```diff
@@ -10,10 +10,12 @@ type Blake2Func = fn(usize, &mut [u64; 8], &[u64; 16], &[u64; 2], bool);
...
     #[cfg(target_arch = "aarch64")]
-    if std::arch::is_aarch64_feature_detected!("neon")
-        && std::arch::is_aarch64_feature_detected!("sha3")
-    {
-        return self::aarch64::blake2b_f;
+    if std::arch::is_aarch64_feature_detected!("neon") {
+        if std::arch::is_aarch64_feature_detected!("sha3") {
+            return self::aarch64::blake2b_f::<self::aarch64::XarSha3>;
+        } else {
+            return self::aarch64::blake2b_f::<self::aarch64::XarNeon>;
```
ElFantasma, line 17: Bug: this path selects `XarNeon` (NEON without SHA3), but `blake2b_f` still calls `veor3q_u64` in the merge step (lines 236/240 of `aarch64.rs`). `veor3q_u64` is also a SHA3 extension — it will generate an illegal instruction on CPUs that only have NEON.

Fix: replace the two `veor3q_u64(x, y, z)` calls with `veorq_u64(veorq_u64(x, y), z)` when SHA3 is unavailable (either via the same trait pattern or a separate conditional).

PR #5165, opened 2025-11-03
perf(l1): try cuckoo filter for trie layers

edg-l:
**Motivation**

Status: testing perf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 2025-10-31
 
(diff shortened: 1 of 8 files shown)


edg-l:
The capacity needs to be really large, furthermore since the cuckoo is not Clone and also has no merge method, it requires cloning it when adding a filter and then again when rebuilding the bloom. Current benchmarks don't show it improve and i think its clear since too much work around it needs to be done for it to work.

PR #5167, opened 2025-11-03
perf(l1): avoid validating node hash

iovoid:
**Motivation**

Currently when fetching a node during tree traversal, we compute it's hash to validate it. This isn't necessary when the trie is consistent.

**Description**

This adds a checked variant for use in snap sync (which requires reading from inconsistent tries) and removes hash validation from the others.

### DIFF
--- a/crates/common/trie/node.rs
+++ b/crates/common/trie/node.rs
@@ -40,7 +40,8 @@ impl NodeRef {
         }
     }
 
-    /// Gets a shared reference to the inner node, if it exists with the correct hash.
+    /// Gets a shared reference to the inner node, checking it's hash.
+    /// Returns `Ok(None)` if the hash is invalid.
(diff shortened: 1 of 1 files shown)


crates/common/trie/node.rs
```diff
@@ -25,6 +25,25 @@ pub enum NodeRef {
...
+            }
+            NodeRef::Hash(_) => db
+                .get(path)?
+                .filter(|rlp| !rlp.is_empty())
+                .map(|rlp| Ok(Arc::new(Node::decode(&rlp)?)))
+                .transpose(),
+        }
+    }
+
+    /// Gets a shared reference to the inner node.
```
pablodeymo: I would like to have an explicit doc telling the difference with `get_node()`

PR #5174, opened 2025-11-04
Fix broken link

rongyi:
fix broken link


mpaulucci:
thx but we dont accept these types of prs

rongyi:
ok, but at least you can change this, if you don't like this pr

PR #5176, opened 2025-11-04
perf(l1): avoid unnecesary work on bloom and keep a per layer bloom

edg-l:
Keeping a per layer bloom helps avoiding rehashing every time a bloom rebuild happens.
This is because the bloom merge method uses the u64 hash keys directly instead of rehashing the value of type T.

This is an improved version of https://github.com/lambdaclass/ethrex/pull/5175 where in addition to that we keep a per layer bloom to avoid extra hash work.

### DIFF
--- a/crates/storage/trie_db/layering.rs
+++ b/crates/storage/trie_db/layering.rs
@@ -1,5 +1,4 @@
 use ethrex_common::H256;
-use rayon::iter::{ParallelBridge, ParallelIterator};
 use rustc_hash::FxHashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


crates/storage/trie_db/layering.rs
```diff
@@ -10,6 +10,10 @@ struct TrieLayer {
     nodes: Arc<FxHashMap<Vec<u8>, Vec<u8>>>,
     parent: H256,
     id: usize,
+    /// Per layer bloom filter, None if the size was exceeded (exceedingly rare).
+    /// Having a bloom per layer avoids the cost of rehashing each key every time we rebuild the globam bloom,
```
Oppen:
```suggestion
    /// Having a bloom per layer avoids the cost of rehashing each key every time we rebuild the global bloom,
```

crates/storage/trie_db/layering.rs
```diff
@@ -10,6 +10,10 @@ struct TrieLayer {
...
+    /// since merge simply uses the u64 hashed keys instead of rehashing.
+    bloom: Option<qfilter::Filter>,
```
Oppen, line 15: Don't change it right now, but I think we should start referring to this as "query filter", as we're not necessarily relying on the Bloom implementation (approximate query filter is the "generic" name, but too long).

crates/storage/trie_db/layering.rs
```diff
@@ -141,22 +159,10 @@ impl TrieLayerCache {
...
-                        return None;
-                    }
-                }
-                Some(bloom)
-            })
+            .map(|entry| entry.bloom.as_ref())
             .collect();
 
-        let Some(mut ret) = blooms.pop().flatten() else {
+        let Some(mut ret) = blooms.pop().flatten().cloned() else {
```
Oppen: Just iterate the `values()`, no need to build a vector now.

PR #5177, opened 2025-11-04
perf(l1,l2): avoid redundant hashing with Code::from_hashed_bytecode and update call sites

VolodymyrBg:
**Motivation**

Unnecessary hashing work on genesis load path; noted by the file’s own TODO to add from_hashed_bytecode.

**Description**

Adds a new constructor Code::from_hashed_bytecode(hash, code) to construct Code without recomputing the keccak hash when the hash is already known and trusted. Includes a debug_assert! to validate the provided hash in debug builds.

Refactors From<GenesisAccount> for Account to avoid double hashing.

Updates call sites to use the new constructor where the hash is available:
- P2P sync bytecode download batching
- Archive sync loading
- L2 prover guest program state construction
- Test utilities (VM runner input, EF blockchain types, REVM DB runner)

This reduces unnecessary hashing on hot paths (e.g., bytecode download and witness preparation) without changing behavior.

### DIFF
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -29,12 +29,20 @@ pub struct Code {
 }
 
 impl Code {
-    // TODO: also add `from_hashed_bytecode` to optimize the download pipeline,
-    // where hash is already known and checked.
     pub fn from_bytecode(code: Bytes) -> Self {
         let jump_targets = Self::compute_jump_targets(&code);
         Self {
             hash: keccak(code.as_ref()),
+            bytecode: code,
+            jump_targets,
+        }
+    }
+
+    pub fn from_hashed_bytecode(hash: H256, code: Bytes) -> Self {
+        debug_assert_eq!(hash, keccak(code.as_ref()));
+        let jump_targets = Self::compute_jump_targets(&code);
+        Self {
+            hash,
             bytecode: code,
             jump_targets,
         }
@@ -125,13 +133,15 @@ impl Default for Code {
 
 impl From<GenesisAccount> for Account {
     fn from(genesis: GenesisAccount) -> Self {
+        let hash = code_hash(&genesis.code);
+        let code = Code::from_hashed_bytecode(hash, genesis.code);
         Self {
             info: AccountInfo {
-                code_hash: code_hash(&genesis.code),
+                code_hash: hash,
                 balance: genesis.balance,
                 nonce: genesis.nonce,
             },
-            code: Code::from_bytecode(genesis.code),
+            code,
             storage: genesis
                 .storage
                 .iter()
--- a/crates/l2/prover/src/guest_program/src/execution.rs
+++ b/crates/l2/prover/src/guest_program/src/execution.rs
@@ -93,6 +93,9 @@ pub enum StatelessExecutionError {
     InvalidParentBlockHeader,
     #[error("Failed to calculate privileged transaction hash")]
     InvalidPrivilegedTransaction,
+    #[cfg(feature = "l2")]
+    #[error("Invalid code hash: expected {0:?}, got {1:?}")]
+    InvalidCodeHash(H256, H256),
     #[error("Internal error: {0}")]
     Internal(String),
 }
--- a/crates/l2/prover/src/guest_program/src/execution.rs
+++ b/crates/l2/prover/src/guest_program/src/execution.rs
@@ -205,7 +205,7 @@ pub fn stateless_validation_l2(
         let mut guest_program_state = GuestProgramState {
             codes_hashed: codes_hashed
                 .into_iter()
-                .map(|(h, c)| (h, Code::from_bytecode(Bytes::from_owner(c))))
+                .map(|(h, c)| (h, Code::from_hashed_bytecode(h, Bytes::from_owner(c))))
                 .collect(),
             parent_block_header,
             first_block_number: initial_db.first_block_number,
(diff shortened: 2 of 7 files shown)


crates/common/types/account.rs
```diff
@@ -125,13 +133,15 @@ impl Default for Code {
 
 impl From<GenesisAccount> for Account {
     fn from(genesis: GenesisAccount) -> Self {
+        let hash = code_hash(&genesis.code);
+        let code = Code::from_hashed_bytecode(hash, genesis.code);
```
Oppen, line 137: This doesn't need changing. If you need to hash it just before calling the constructor, then just let the old constructor do it.

crates/l2/prover/src/guest_program/src/execution.rs
```diff
@@ -205,7 +205,7 @@ pub fn stateless_validation_l2(
         let mut guest_program_state = GuestProgramState {
             codes_hashed: codes_hashed
                 .into_iter()
-                .map(|(h, c)| (h, Code::from_bytecode(Bytes::from_owner(c))))
+                .map(|(h, c)| (h, Code::from_hashed_bytecode(h, Bytes::from_owner(c))))
```
Oppen: I think we should be checking that the hashes match, rather than trusting them.

crates/networking/p2p/sync.rs
```diff
@@ -952,7 +952,8 @@ impl Syncer {
                             .write_account_code_batch(
                                 code_hashes_to_download
                                     .drain(..)
-                                    .zip(bytecodes.into_iter().map(Code::from_bytecode))
+                                    .zip(bytecodes.into_iter())
+                                    .map(|(h, b)| (h, Code::from_hashed_bytecode(h, b)))
```
Oppen: This is OK because I checked the download code and it hashes itself to validate. I would like that to be expressed in the program. For now, adding a comment should suffice, later on we might want to make the type system help us (e.g. just store `Code` at the download side and not create it here).

crates/vm/levm/runner/src/input.rs
```diff
@@ -35,13 +35,14 @@ pub struct InputAccount {
 
 impl From<InputAccount> for Account {
     fn from(account: InputAccount) -> Self {
+        let hash = code_hash(&account.code);
```
Oppen, line 38: You could create the `Code` here with `Code::from_bytecode(account.code)` and then fill the hash by using the `code.hash` field.

Oppen: That said, nice catch, this duplicate hash could have been handled in the original PR but was missed.

tooling/ef_tests/blockchain/types.rs
```diff
@@ -561,13 +561,14 @@ impl From<Transaction> for EIP2930Transaction {
 
 impl From<Account> for ethrexAccount {
     fn from(val: Account) -> Self {
+        let hash = code_hash(&val.code);
```
Oppen, line 564: Ditto.

Oppen (commented):
Looks good generally, but see comments.

VolodymyrBg:
> Looks good generally, but see comments.

made changes

crates/networking/p2p/sync.rs
```diff
@@ -948,14 +948,15 @@ impl Syncer {
...
-                                    .zip(bytecodes.into_iter().map(Code::from_bytecode))
-                                    .collect(),
-                            )
-                            .await?;
+                        // Download side validates bytecode hashes; here we rely on that
+                        // and construct Code using the known hash without rehashing.
+                        let account_codes: Vec<(H256, Code)> = code_hashes_to_download
+                            .drain(..)
+                            .zip(bytecodes.into_iter())
+                            .map(|(h, b)| (h, Code::from_hashed_bytecode(h, b)))
```
ElFantasma, line 956: This change is already on `main` — the p2p sync code was updated to use `Code::from_bytecode_unchecked(code, hash)` in a prior PR. This file will conflict on rebase.

ElFantasma (commented):
This PR is **stale and largely superseded** by work already merged to `main`. The codebase now has `Code::from_bytecode_unchecked(code, hash)` (added in PR #5397) which does the same thing as the proposed `from_hashed_bytecode(hash, code)` — constructs `Code` with a pre-computed hash, skipping `keccak`. The p2p sync call sites this PR modifies were already updated to use `from_bytecode_unchecked`.

The guest program file (`crates/l2/prover/src/guest_program/src/execution.rs`) was restructured and moved to `crates/guest-program/src/common/execution.rs` (PR #5995), so that part of the diff would conflict entirely.

That said, the **intent is still valid** — a few call sites still double-hash:
- `tooling/ef_tests/blockchain/types.rs:577-581` — `code_hash()` + `Code::from_bytecode()`
- `tooling/ef_tests/state/runner/revm_db.rs:228` — `Code::from_bytecode()` when hash is already available
- `tooling/archive_sync/src/main.rs:150` — same pattern

**crates/common/types/account.rs:41** — `from_hashed_bytecode` duplicates the existing `from_bytecode_unchecked(code, hash)`. The only differences are reversed parameter order and a `debug_assert_eq!`. Rather than adding a second method, consider using the existing one at the remaining double-hash call sites, or adding the debug assert to `from_bytecode_unchecked` directly.

If we want to fix these remaining double-hash sites, a much smaller PR that uses the existing `from_bytecode_unchecked` (no new method needed) would be the right approach. 

Maybe we should close this and create a new one

MegaRedHand:

PR #5178, opened 2025-11-04
ci(l1,l2): fix apt trigger.

samoht9277:
**Motivation**

This PR fixes a CI issue that broke the `tag_latest` workflow due to a syntax error that triggered the following error:

> The workflow is requesting 'actions: write, contents: write', but is only allowed 'actions: none, contents: read'.

Now the `ethrex-apt` workflow get's triggered properly

### DIFF
--- a/.github/workflows/tag_latest.yaml
+++ b/.github/workflows/tag_latest.yaml
@@ -79,6 +79,6 @@ jobs:
       contents: write
       actions: write
     secrets:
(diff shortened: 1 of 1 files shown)


samoht9277:
Testing fix here: https://github.com/samoht9277/ethrex/actions/runs/19072665344/

samoht9277:
Tested here: https://github.com/samoht9277/ethrex/actions/runs/19077575984/job/54497681430 ✅

ilitteri (approved):

PR #5181, opened 2025-11-04
chore(l1,l2): compilation with AVX2 feature in linux x86-64

pablodeymo:
**Motivation**

**Description**

Closes #issue_number


.cargo/config.toml
```diff
@@ -1,2 +1,6 @@
 [target.x86_64-unknown-linux-gnu]
-rustflags = ["-Ctarget-cpu=x86-64-v2"]
+rustflags = [
+    "-Ctarget-cpu=x86-64-v2",
+    "-Csha2_backend=x86-avx2",
+    "-Ctarget-feature=+avx2",
```
azteca1998: Maybe just specifying `x86-64-v3` is better? According to [wikipedia](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels), it is basically what we need.

pablodeymo:
Superseded by #5189 5189

PR #5189, opened 2025-11-04
feat(l1,l2): setting avx2 flags for compilation and adding features asm to crates

pablodeymo:
**Motivation**

We want to take advantage of AVX2 processor features, and we are using `asm` features of dependencies in case it's possible.

**Description**

This PR sets compilation flag with AVX2 with the configuration:

```
target-cpu=x86-64-v3
```

At the same time, it changes Cargo.toml files in the following way:
* `c-kzg`: no default features, and add `std` and `ethereum_kzg_settings`.
* `ark-ff`: with `asm` feature.
* `sha3`: unified in Cargo.toml of the workspace and `asm` feature added.

Closes #issue_number


.cargo/config.toml
```diff
@@ -1,2 +1,5 @@
 [target.x86_64-unknown-linux-gnu]
-rustflags = ["-Ctarget-cpu=x86-64-v2"]
+rustflags = [
+    "-Ctarget-cpu=x86-64-v3",
+    "-Ctarget-feature=+avx2,+sse2,+ssse3,+sse4.1,+sse4.2,+bmi1,+lzcnt,+pclmulqdq",
```
azteca1998, line 4: The `target-feature` should already be enabled for the `x86-64-v3` target CPU.
Are they explicitly there for any reason in particular?

pablodeymo: because we found some crates are checking for the specific feature not included when we set target-cpu

PR #5190, opened 2025-11-04
refactor(l1, l2): refactored chain config

SDartayet:
**Motivation**

Improve the maintainability of our code with regards to future forks.

**Description**

The current way we handle activation timestamps and blob schedules is to have a struct field per fork. This creates maintainability issues down the line, since it requires a separate function to check if the fork is activated, a new match arm for its blobschedule, etc. This PR moves the timestamps and blob schedules into arrays, such that all that's required in the future is to modify an array containing the post-merge forks to add a new one.

Closes #issue_number

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -6,14 +6,15 @@ use std::{
 use crate::{
     constants::{
         TX_ACCESS_LIST_ADDRESS_GAS, TX_ACCESS_LIST_STORAGE_KEY_GAS, TX_CREATE_GAS_COST,
-        TX_DATA_NON_ZERO_GAS, TX_DATA_NON_ZERO_GAS_EIP2028, TX_DATA_ZERO_GAS_COST, TX_GAS_COST,
-        TX_INIT_CODE_WORD_GAS_COST,
+        TX_DATA_NON_ZERO_GAS, TX_DATA_ZERO_GAS_COST, TX_GAS_COST, TX_INIT_CODE_WORD_GAS_COST,
     },
     error::MempoolError,
 };
 use ethrex_common::{
     Address, H160, H256, U256,
-    types::{BlobsBundle, BlockHeader, ChainConfig, MempoolTransaction, Transaction, TxType},
+    types::{
+        BlobsBundle, BlockHeader, ChainConfig, Fork::*, MempoolTransaction, Transaction, TxType,
+    },
 };
 use ethrex_storage::error::StoreError;
 use std::collections::HashSet;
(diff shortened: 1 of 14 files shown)


crates/blockchain/mempool.rs
```diff
@@ -505,13 +501,12 @@ mod tests {
         Ok(store)
     }
 
-    fn build_basic_config_and_header(
```
mpaulucci, line 508: seems this should be moved to a test utils or something

crates/common/config/networks.rs
```diff
@@ -117,14 +117,19 @@ impl Network {
             }
             Network::LocalDevnet => Ok(serde_json::from_str(LOCAL_DEVNET_GENESIS_CONTENTS)?),
             Network::LocalDevnetL2 => Ok(serde_json::from_str(LOCAL_DEVNETL2_GENESIS_CONTENTS)?),
-            Network::L2Chain(chain_id) => Ok(Genesis {
-                config: ChainConfig {
-                    chain_id: *chain_id,
-                    prague_time: Some(0),
+            Network::L2Chain(chain_id) => {
```
mpaulucci, line 120: wonder if we can move this to the genesis/chain config L1 module so we don't have to leak the abstraction here.

Something like `Genesis:new(Prague)` or something

crates/common/types/genesis.rs
```diff
@@ -343,65 +232,20 @@ impl From<Fork> for &str {
...
-        self.istanbul_block.is_some_and(|num| num <= block_number)
-    }
-
-    pub fn is_london_activated(&self, block_number: BlockNumber) -> bool {
-        self.london_block.is_some_and(|num| num <= block_number)
-    }
-
-    pub fn is_eip155_activated(&self, block_number: BlockNumber) -> bool {
-        self.eip155_block.is_some_and(|num| num <= block_number)
+    pub fn is_fork_activated(&self, fork: Fork, block_timestamp: u64) -> bool {
```
mpaulucci, line 235: this change is great

crates/common/types/genesis.rs
```diff
@@ -280,54 +166,57 @@ lazy_static::lazy_static! {
...
+    Clone,
+    Copy,
+    Serialize,
+    Deserialize,
+    RSerialize,
+    RDeserialize,
+    Archive,
+)]
 pub enum Fork {
-    Frontier = 0,
```
mpaulucci, line 285: why do you have to remove these forks?

crates/common/types/genesis.rs
```diff
@@ -421,181 +265,82 @@ impl ChainConfig {
...
-            Fork::Paris
-        }
+        let Some(index) = self
+            .fork_activation_timestamps
+            .iter()
+            .rposition(|possible_time| {
+                possible_time.is_some_and(|activation_time| activation_time <= block_timestamp)
+            })
+        else {
+            return Paris;
```
mpaulucci, line 275: shouldnt we error here?

PR #5191, opened 2025-11-05
perf(l1): avoid copying ChainConfig on Store::clone

### LINKED ISSUE #5127: [PERF] Avoid copying a `ChainConfig` on each `Store::clone`
We clone the `Store` extensively throughout the codebase, since we mostly expect this to be a handle. This changed in #5071, since now the chain config (624 bytes) is copied on each clone. We should check if this affects performance.

We should also check if it makes sense to cache the chain config at this level, or if it makes more sense to cache it inside the `Blockchain` struct, for example.

crStiv:
Wraps ChainConfig in Arc to avoid copying 624 bytes on each Store::clone.

Fixes #5127


mpaulucci:
I think this looks good, but please remove the `Copy` trait from it so we don't do implicit copying.

Arkenan:
Hey, thank you very much for the contribution. At this moment we're not looking into improvements in this area and it looks like the PR has many conflicts and doesn't have recent updates. If you're still interested in adding this let us know and we can talk about starting a new PR.

PR #5194, opened 2025-11-05
perf(l1): make  bloom qfilter counting to avoid rebuilding entire filter

edg-l:
**Motivation**

The qfilter filter can keep track and count inserted items, we can use this to avoid rebuilding the entire filter. This is done by removing a count from the filter by iterating the removed layers in the `commit` method. If for some reason remove fails, we fall back to a full bloom rebuild.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -216,6 +216,7 @@ impl Blockchain {
         let queue_length_ref = &queue_length;
         let mut max_queue_length = 0;
         let (execution_result, account_updates_list) = std::thread::scope(|s| {
+            let max_queue_length_ref = &mut max_queue_length;
             let (tx, rx) = channel();
             let execution_handle = s.spawn(move || -> Result<_, ChainError> {
                 let execution_result = vm.execute_block_pipeline(block, tx, queue_length_ref)?;
@@ -233,7 +234,7 @@ impl Blockchain {
                     rx,
                     &parent_header,
                     queue_length_ref,
-                    &mut max_queue_length,
+                    max_queue_length_ref,
                 )?;
                 let merkle_end_instant = Instant::now();
                 Ok((account_updates_list, merkle_end_instant))
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -54,6 +54,14 @@ impl LEVM {
         for (tx, tx_sender) in block.body.get_transactions_with_sender().map_err(|error| {
             EvmError::Transaction(format!("Couldn't recover addresses with error: {error}"))
         })? {
+            if cumulative_gas_used + tx.gas_limit() > block.header.gas_limit {
+                return Err(EvmError::Transaction(format!(
+                    "Gas allowance exceeded. Block gas limit {} can be surpassed by executing transaction with gas limit {}",
+                    block.header.gas_limit,
+                    tx.gas_limit()
+                )));
+            }
+
             let report = Self::execute_tx(tx, tx_sender, &block.header, db, vm_type)?;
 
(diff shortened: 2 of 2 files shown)


crates/storage/trie_db/layering.rs
```diff
@@ -184,8 +144,23 @@ impl TrieLayerCache {
...
+                            tracing::warn!("TrieLayerCache: bloom.remove returned false, meaning this node wasn't inserted into the bloom, removing bloom entirely to avoid false negatives");
+                            self.bloom = None;
+                            break;
+                        }
+                    }
+                }
+
+                false
+            }
+        });
```
Oppen: You might want to use [extract_if](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extract_if) instead. That way you can use the same filter and just extract the loop for the removed layers. It's stable since 1.87 and we use 1.90 IIRC.

crates/storage/trie_db/layering.rs
```diff
@@ -184,8 +144,23 @@ impl TrieLayerCache {
...
+        self.layers.retain(|_, item| {
+            if item.id > layer.id {
+                true
+            } else {
+                if let Some(bloom) = self.bloom.as_mut() {
+                    for node in item.nodes.keys() {
+                        if !bloom.remove(node) {
+                            tracing::warn!("TrieLayerCache: bloom.remove returned false, meaning this node wasn't inserted into the bloom, removing bloom entirely to avoid false negatives");
+                            self.bloom = None;
+                            break;
```
Oppen: Maybe we should keep rebuild for this case. Rebuild has the ability to recover as the chain progresses, this one stays unfiltered (and thus slowed down) for the remaining life of the node.

crates/storage/trie_db/layering.rs
```diff
@@ -184,8 +144,23 @@ impl TrieLayerCache {
...
-        self.layers.retain(|_, item| item.id > layer.id);
-        self.rebuild_bloom(); // layers removed, rebuild global bloom filter.
+        self.layers.retain(|_, item| {
+            if item.id > layer.id {
+                true
+            } else {
+                if let Some(bloom) = self.bloom.as_mut() {
+                    for node in item.nodes.keys() {
+                        if !bloom.remove(node) {
+                            tracing::warn!("TrieLayerCache: bloom.remove returned false, meaning this node wasn't inserted into the bloom, removing bloom entirely to avoid false negatives");
```
MegaRedHand: I think this is too much info for the user.

```suggestion
                            tracing::warn!("TrieLayerCache: bloom.remove failed. Removing bloom entirely to avoid false negatives");
```

crates/storage/trie_db/layering.rs
pablodeymo: I can confirm we use rust 1.90

crates/storage/trie_db/layering.rs
```diff
@@ -135,8 +137,50 @@ impl TrieLayerCache {
         self.layers.insert(state_root, Arc::new(entry));
     }
 
-    /// Rebuilds the global bloom filter accruing all current existing layers.
+    pub fn commit(&mut self, state_root: H256) -> Option<Vec<(Vec<u8>, Vec<u8>)>> {
+        let layer = match Arc::try_unwrap(self.layers.remove(&state_root)?) {
+            Ok(layer) => layer,
+            Err(layer) => TrieLayer::clone(&layer),
+        };
```
pablodeymo: we can use 
`.unwrap_or_else(|| TrieLayer::clone(&layer))`
instead of the match

jrchatruc (changes requested):
This PR seems to have some degradation in perf over time (see image below, this branch in server notifications-4). Blocking merge until we figure out the reason.

Image

PR #5196, opened 2025-11-05
fix(l1): always use fullsync if db had data

fedacking:
**Motivation**

When the db had data, we should fullsync, not override the user's data

**Description**

- Changes the fullsync option instead of snap to be any number instead of `MIN_FULL_BLOCKS`

### DIFF
--- a/crates/networking/rpc/engine/fork_choice.rs
+++ b/crates/networking/rpc/engine/fork_choice.rs
@@ -4,7 +4,6 @@ use ethrex_blockchain::{
     payload::{BuildPayloadArgs, create_payload},
 };
 use ethrex_common::types::{BlockHeader, ELASTICITY_MULTIPLIER};
-use ethrex_p2p::sync::SyncMode;
 use serde_json::Value;
 use tracing::{info, warn};
 


crates/networking/p2p/sync.rs
```diff
@@ -285,14 +285,12 @@ impl Syncer {
...
+            // If the sync head is not 0 we search to fullsync
             if sync_head_found {
                 let latest_block_number = store.get_latest_block_number().await?;
-                if last_block_number.saturating_sub(latest_block_number) < MIN_FULL_BLOCKS as u64 {
+                if latest_block_number != 0 {
                     // Too few blocks for a snap sync, switching to full sync
-                    debug!(
-                        "Sync head is less than {MIN_FULL_BLOCKS} blocks away, switching to FullSync"
-                    );
+                    debug!("Sync head is found switching to FullSync");
```
mpaulucci: this could be an `info`. Also, comma missing.

crates/networking/p2p/sync.rs
```diff
@@ -285,14 +285,12 @@ impl Syncer {
             current_head = last_block_hash;
```
mpaulucci, line 285: Do we have to revert https://github.com/lambdaclass/ethrex/pull/4985 ?

crates/networking/p2p/sync.rs
```diff
@@ -285,14 +285,12 @@ impl Syncer {
...
             current_head_number = last_block_number;
 
-            // If the sync head is less than 64 blocks away from our current head switch to full-sync
+            // If the sync head is not 0 we search to fullsync
```
mpaulucci, line 288: IMO, if the sync head is 0 pero is less than X blocks away, we should also not do snap sync. Snap sync is for 1M blocks distance, not 100.

fedacking: Done in [e4dc36d](https://github.com/lambdaclass/ethrex/pull/5196/commits/e4dc36da0170638e368890a4601bd52156c35abf)

crates/networking/p2p/sync.rs
fedacking: Done in [e4dc36d](https://github.com/lambdaclass/ethrex/pull/5196/commits/e4dc36da0170638e368890a4601bd52156c35abf)

crates/networking/p2p/sync.rs
fedacking: Testing in [636a594](https://github.com/lambdaclass/ethrex/pull/5196/commits/636a5945cadb5db7b9ccfc6e125267b8e13da798)

crates/networking/p2p/sync.rs
```diff
@@ -285,17 +285,16 @@ impl Syncer {
...
-                    debug!(
-                        "Sync head is less than {MIN_FULL_BLOCKS} blocks away, switching to FullSync"
-                    );
-                    self.snap_enabled.store(false, Ordering::Relaxed);
-                    return self.sync_cycle_full(sync_head, store.clone()).await;
-                }
+            // If the sync head is not 0 we search to fullsync
+            let head_found = sync_head_found && store.get_latest_block_number().await? > 0;
+            // Or the head is very close to 0
+            let head_close_to_0 = last_block_number < MIN_FULL_BLOCKS;
```
MegaRedHand, line 291: ~This skips snap-syncing when we have a few blocks in the DB, while the old logic compared the distance between head and sync-head.~

Edit: this skips snap-syncing when the sync-head's block number is less than 10000

PR #5197, opened 2025-11-05
fix(l1): fix queue length log

iovoid:
**Motivation**

Currently the log incorrectly always shows max_queue_length as 0.

**Description**

The move closure takes ownership and mutates _a copy_ of the max_queue_length. This is fixed by taking the reference outside.


MegaRedHand (approved):
LGTM. What a footgun!

jrchatruc (approved):

PR #5198, opened 2025-11-05
refactor(l2): remove `COMPILE_CONTRACTS` compile flag

tomip01:
**Motivation**

We now have the `l2` again, so we could use this as whether to compile the contracts for the L2 or not.

**Description**

- Add new env var `SKIP_COMPILE_CONTRACTS`. This will as the name says, not compile the contracts. This is useful for the jobs that only checks compilation or require the `l2` flag and do not have `solc` installed.
- Remove every reference to the `COMPILE_CONTRACTS` flag


tomip01:
Close in favor of https://github.com/lambdaclass/ethrex/pull/5216

PR #5199, opened 2025-11-05
fix(l1): make function inner for InMemoryTrieDB

JereSalo:
**Motivation**

- Make necessary changes in order for ethrex-replay to work fine (we make use of these things there)

**Description**

- Added a method to `InMemoryTrieDB` to use it in ethrex-replay

### DIFF
--- a/crates/common/trie/db.rs
+++ b/crates/common/trie/db.rs
@@ -59,7 +59,7 @@ impl InMemoryTrieDB {
         }
     }
 
-    // Do not remove or make private as we use this outside of ethrex
+    // Do not remove or make private as we use this in ethrex-replay
     pub fn from_nodes(
         root_hash: H256,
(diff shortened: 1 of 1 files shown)


JereSalo:
Maybe we don't want this, so I will address Copilot comments after we decide if we actually want this or not. Otherwise we may have to solve it in a different way.

Edit: yep, I made a mistake, we didn't want this and it wasn't even necessary

ilitteri (approved):

SDartayet (approved):

JereSalo:
> Maybe it’s better to clarify in the comments that we need these methods for ethrex-replay

https://github.com/lambdaclass/ethrex/pull/5199/commits/a20302b468dc6160090f11ca45a7cce428f15298

PR #5200, opened 2025-11-05
perf(l1): batch merkelization requests if the queue is not empty

iovoid:
**Motivation**

Currently for every transaction being executed, we get the transitions and send the updates to the merkelization thread.

However this has two drawbacks:
- getting the update list isn't cheap, taking up 10% of block execution
- unbatched requests cause more work to be done, increasing total merkelization time

The merkelization queue often had dozens of requests by the end of block execution.

**Description**

We delay calculating and sending updates until the merkelizer is ready (queue is empty), which causes the changes to the batched.

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -105,6 +105,8 @@ impl LEVM {
         let mut receipts = Vec::new();
         let mut cumulative_gas_used = 0;
 
+        // Starts at 2 to account for the two precompile calls done in `Self::prepare_block`.
+        // The value itself can be safely changed.
         let mut tx_since_last_flush = 2;
 
         for (tx, tx_sender) in block.body.get_transactions_with_sender().map_err(|error| {


crates/vm/backends/levm/mod.rs
```diff
@@ -136,6 +143,9 @@ impl LEVM {
 
             receipts.push(receipt);
         }
+        if queue_length.load(Ordering::Relaxed) > 0 {
```
MegaRedHand: I find this condition weird. Why do we check the queue isn't empty?

crates/vm/backends/levm/mod.rs
```diff
@@ -105,6 +105,8 @@ impl LEVM {
         let mut receipts = Vec::new();
         let mut cumulative_gas_used = 0;
 
+        let mut tx_since_last_flush = 2;
```
MegaRedHand, line 110: Why do we start with 2?

crates/vm/backends/levm/mod.rs
```diff
@@ -620,11 +630,9 @@ pub fn extract_all_requests_levm_pipeline(
     let withdrawals_data: Vec<u8> = LEVM::read_withdrawal_requests(header, db, vm_type)?
         .output
         .into();
-    LEVM::send_state_transitions_tx(merkleizer, db, queue_length)?;
```
MegaRedHand, line 623: I think we can dedup `extract_all_requests_levm_pipeline` with `extract_all_requests_levm`

crates/vm/backends/levm/mod.rs
Oppen: To keep batching for the latter steps. We need a comment making it explicit.

crates/vm/backends/levm/mod.rs
iovoid: To account for the previous precompile calls. Documented in 2a02cc15a.

crates/vm/backends/levm/mod.rs
iovoid: Yeah this condition is wrong, fixed in 8861f88eb.

crates/vm/backends/levm/mod.rs
iovoid: Removed in 8861f88eb

PR #5203, opened 2025-11-05
perf(l1): background insertion of receipts and tx locations

iovoid:
**Motivation**

We don't need the receipts and the transaction locations to be inserted to continue with the execution of the next block.

**Description**

We move it to the background thread, with the trie commitment operations.


iovoid:
The RPC _does_, however, need the logs and receipts to correctly answer queries for the latest block.

PR #5205, opened 2025-11-05
fix(l1): build initial empty block in block building

### LINKED ISSUE #4997: Start block building process with an empty block
The spec says:

> [Client software SHOULD build the initial version of the payload which has an empty transaction set.](https://github.com/ethereum/execution-apis/blob/bbf817f7f6404fc0d270f66a3c96f8737e6d7176/src/engine/paris.md#payload-building)

We don't do that, currently. Consensus requires execution blocks to be built, even if empty, to ensure CL liveness. From what I saw, not doing so collapses local networks. We should change this to start with an empty block, since it should be simple enough.

SDartayet:
**Motivation**

Properly complying with the block building spec.

**Description**

As outlined in #4997, we should be bulding an initial empty block before filling it with transactions. This PR fixes that.

Closes #4997

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -370,8 +370,7 @@ impl Blockchain {
         let start = Instant::now();
         const SECONDS_PER_SLOT: Duration = Duration::from_secs(12);
         // Attempt to rebuild the payload as many times within the given timeframe to maximize fee revenue
-        // TODO(#4997): start with an empty block
-        let mut res = self.build_initial_payload(payload.clone())?;
+        let mut res = self.build_empty_payload(payload.clone())?;
         while start.elapsed() < SECONDS_PER_SLOT && !cancel_token.is_cancelled() {
             let payload = payload.clone();
             let self_clone = self.clone();
@@ -399,7 +398,7 @@ impl Blockchain {
         self.build_payload_inner(payload, true)
     }
 
-    pub fn build_initial_payload(&self, payload: Block) -> Result<PayloadBuildResult, ChainError> {
+    pub fn build_empty_payload(&self, payload: Block) -> Result<PayloadBuildResult, ChainError> {
         self.build_payload_inner(payload, false)
     }
 


crates/blockchain/payload.rs
```diff
@@ -371,7 +371,7 @@ impl Blockchain {
         const SECONDS_PER_SLOT: Duration = Duration::from_secs(12);
         // Attempt to rebuild the payload as many times within the given timeframe to maximize fee revenue
         // TODO(#4997): start with an empty block
```
cdiielsi: Shouldn't this TODO be dropped now?

cdiielsi (approved):

crates/blockchain/payload.rs
```diff
@@ -393,8 +393,22 @@ impl Blockchain {
...
     }
 
-    /// Completes the payload building process, return the block value
+    // We separate this into two functions, so we can build the initial empty payload and then start filling it
+    // TODO: once we implement a mechanism to gradually fill the payload with transactions we won't need this
     pub fn build_payload(&self, payload: Block) -> Result<PayloadBuildResult, ChainError> {
+        self.build_payload_inner(payload, true)
+    }
+
+    pub fn build_initial_payload(&self, payload: Block) -> Result<PayloadBuildResult, ChainError> {
```
MegaRedHand: Maybe rename this to `build_empty_payload`?

SDartayet: Sure, fixed [here](https://github.com/lambdaclass/ethrex/pull/5205/commits/5d760edcb7ea40b1e9f23417eacc09883b16dd8e)!

crates/blockchain/payload.rs
SDartayet: Sure, [fixed](https://github.com/lambdaclass/ethrex/pull/5205/commits/5d760edcb7ea40b1e9f23417eacc09883b16dd8e)!

PR #5207, opened 2025-11-05
perf(l2): avoid redundant String allocations in BlocksTable::render

Forostovec:
**Motivation**

In blocks.rs::render(), there are unnecessary allocations due to .to_string()on existing String.

**Description**

Replaced .to_string() calls on existing String values with borrowed references when building Spans in 
BlocksTable::render() (crates/l2/monitor/widget/blocks.rs).
This avoids unnecessary allocations and copies at render time. ratatui’s Span::styled takes Into<Cow<'_, str>>, so &String/&str can be used directly. Aligned with existing usage in mempool.rs.


ElFantasma:
Thanks for your contribution, but this was resolved by PR #5012, which replaced the 7-element tuple with a `BlockListItem` struct and switched `render()` to use `&item.field` references directly.

PR #5210, opened 2025-11-06
chore(l1,l2): remove dead Environment fields difficulty and block_blob_gas_used

Galoretka:
Removed unused fields difficulty and block_blob_gas_used from Environment in crates/vm/levm/src/environment.rs , and stopped populating them in:
- crates/vm/backends/levm/mod.rs::{setup_env, env_from_generic, generic_system_contract_levm}
- tooling/ef_tests/state/runner/levm_runner.rs::prepare_vm_for_tx
- tooling/ef_tests/state_v2/src/modules/runner.rs::get_vm_env_for_test

Reason: These fields are never read in opcode handlers or hooks. Post-merge execution uses prev_randao (EIP-4399) instead of difficulty. Blob fee logic relies on block_excess_blob_gas and per-tx fields; blob_gas_used is only relevant at block validation, not EVM runtime. This reduces cognitive load and avoids unnecessary environment initialization without changing execution semantics.


ElFantasma (changes requested):
The analysis is correct — `difficulty` and `block_blob_gas_used` are set but never read in LEVM opcode handlers or hooks. However, the PR is significantly stale (opened Nov 2025) and **will not compile after merge**: `crates/vm/backends/levm/mod.rs` has been refactored since then and now has 5 additional locations that set these fields (lines 338, 345, 635, 842, 849) which this PR doesn't remove. Additionally, `tooling/ef_tests/state_v2/src/modules/block_runner.rs:125` sets `difficulty` and wasn't in the original diff.

Needs a rebase against current `main` and removal of the remaining setter sites to compile.

MegaRedHand:

PR #5216, opened 2025-11-06
refactor(l2): remove `COMPILE_CONTRACTS` compile flag

### LINKED ISSUE #5098: Remove `COMPILE_CONTRACTS` env
Now that the L1 and L2 compile separately, we can get rid of the need for setting the env `COMPILE_CONTRACTS` to decide whether to compile the L2 contracts.

tomip01:
**Motivation**

The `COMPILE_CONTRACTS` flag is no longer needed as we have different binaries for the L1 and the L2.

**Description**

* Removed the `COMPILE_CONTRACTS` environment variable.
* Updated CI workflows to install `solc` v0.8.29.
* Added a Nix package for `solc` v0.8.29 to support reproducible builds in Nix environments.

Closes #5098


.github/workflows/pr-main_l1.yaml
```diff
@@ -56,6 +56,12 @@ jobs:
         with:
           components: rustfmt, clippy
 
+      - name: Install solc
```
mpaulucci: Why we need to install solc on L1? It should not be required to run L1

crates/l2/sdk/build.rs
JereSalo, line 1: If `ETHREX_SDK_OZ_CONTRACTS_DIR` changes the build script won't be triggered, right? Should we change that? Let me know if I'm wrong.
Should we add `println!("cargo::rerun-if-env-changed=ETHREX_SDK_OZ_CONTRACTS_DIR");` like the thing we had before but with `COMPILE_CONTRACTS`? I'm sure it's very unlikely for it to change and this doesn't have a real impact though haha

.github/workflows/pr-main_l1.yaml
tomip01: For L1 alone it isn’t needed, but one of the steps we run, for example `cargo check --workspace --all-targets`, ends up checking everything under `tooling/`, which in turn compiles the L2 SDK and therefore the contracts.
We talked about this a while back: I initially proposed an env var `SKIP_COMPILE_CONTRACTS` in this [PR](https://github.com/lambdaclass/ethrex/pull/5198)￼, but we agreed this approach was cleaner.

crates/l2/sdk/build.rs
tomip01: you are right!

tomip01: Done here [3b0af67](https://github.com/lambdaclass/ethrex/pull/5216/commits/3b0af6750d40dcc6f30f9ea9e6799e75bbf523f4)

.github/workflows/pr-main_l1.yaml
mpaulucci: Let me see if I understand this correctly. 

If someone downloads the repo and uses any IDE (they run cargo build --workspace by default I think), then the build will fail because the user does not have solc installed?

I think that is very bad. If that's the case, maybe we should not have `l2` flag enabled by default?

mpaulucci: Just talked to @ilitteri . `cargo build --workspace` (which is what most IDEs use) should work without having `solc` installed. This is so that new contributors are able to download the repo and contribute without having an extra dependency that is not needed when running L1. If the `l2` flag is set, it is reasonable to ask for `solc`

tomip01: Yes! I wasn't very clear with my response.
The problem is this step and the next one in the Lint for the L1:
```yaml
      - name: Run tooling cargo check
        run: cargo check --workspace --all-targets
        working-directory: tooling
```
Look at [this](https://github.com/lambdaclass/ethrex/actions/runs/19669781878/job/56335517328) run. It runs the check not in the root project but for `tooling`.

The workspace for `tooling/` does need for the L2 SDK which needs `solc`.

tomip01: Maybe we could move the `tooling` check for an L2 job but it won't run on every PR

mpaulucci: Yes, I would create a tooling workflow and move the tooling checks there.

Issue #5219, opened 2025-11-06
Rename function instrumentation to be consistent


rodrigo-o:
Right now the breakdown of time during block execution is done through an already define set of traces. The naming is not consistent and a bit confusing across them, we need to update them

PR #5220, opened 2025-11-06
feat(l1): upgrade execution breakdown pie chart and add new panels

### LINKED ISSUE #5218: Enhance grafana execution breakdown panels
Right now we have just one pie chart, and after the latest pipeline of merkleization during execution we want to be able to have more insight into the deagregation and measure it through time/block instead of just percentage in the pie chart.

rodrigo-o:
**Motivation**

We needed more granular panels to analyze  the breakdown of execution for performance decisions.

**Description**

This PR use the current instrumentation to:
- Update the pie char of execution
- Add a diff measurement between execution and merkleization to understand the work of the `execute_block_pipeline` function
- Add a deagregated series by block showing the different instrumentations available today
- Maintain all of them in the previous Block Execution Breakdown row and repeat them vertically across instances

image

**Note for running it locally:**

_After #5088 we needed to re-labeling our local instances to avoid duplicated panels that repeat across instances. This requires prometheus to be restarted to pick up the new labeling, one way to check this out is make look at intances, if `ethereum-metrics-exporte` appears there, we need to restart prometheus, only `localhost` should be there locally._

| Bad (needs restarting prometheus)| Good |
| --- | --- |
| image | image |

**Next steps:**

- Naming is not optimal, but changing it would require breaking the panel for instances not updated to this PR. We want first to test it with all instances and make sure the panels are useful and change the names afterwards, here is an issue for tracking that: #5219

Closes #5218

### DIFF
--- a/metrics/provisioning/grafana/dashboards/common_dashboards/ethrex_l1_perf.json
+++ b/metrics/provisioning/grafana/dashboards/common_dashboards/ethrex_l1_perf.json
@@ -1275,7 +1275,7 @@
             "type": "prometheus",
             "uid": "${DS_PROMETHEUS}"
           },
-          "description": "This panel show the difference between the Execution vs merkelization in the `execute_block_pipeline` function. It's show as a percentage of the whole function. If its negative, execution took more time than merkleization.",
+          "description": "This panel shows the difference between the Execution vs merkleization in the `execute_block_pipeline` function. It's shown as a percentage of the whole function. If it's negative, execution took more time than merkleization.",
           "fieldConfig": {
             "defaults": {
               "color": {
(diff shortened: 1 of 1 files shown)


crates/vm/backends/mod.rs
```diff
@@ -73,7 +73,6 @@ impl Evm {
         }
     }
 
-    #[instrument(level = "trace", name = "Block execution", skip_all)]
```
rodrigo-o, line 76: We are not instrumenting other non-pipeline related functions, if we want to do so we might need to do it in a following PR.

crates/storage/store.rs
```diff
@@ -352,7 +352,6 @@ impl Store {
 
     /// Applies account updates based on the block's latest storage state
     /// and returns the new state root after the updates have been applied.
-    #[instrument(level = "trace", name = "Trie update", skip_all)]
```
rodrigo-o, line 355: This was already instrumented at `handle_merkleization`, if it was intended to be part of the non-pipeline instrumentation, we need to do it in a follow-up.

PR #5004, opened 2025-10-22
feat(l1): admin_addPeer endpoint

cdiielsi:
**Motivation**

Include geth's [admin_addPeer](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-admin#admin-addpeer) endpoint since we found it necessary for running eip7934 hive tests with the `ethereum/eels/consume-sync ` simulation.
Besides Nethermind, Reth, Besu, Erigon and other execution clients implement it.

**Description**

This pr introduces the RPC endpoint `admin_addPeer`. Since this endpoint needed to access p2p's generic server used to establish connections, the `init_l1` function was restructure so the genserver is saved within the peer_handler struct and can later be accessed from the `RPCApiContext`.

In `admin_addPeer` the connection is tried in a period of 10 seconds waiting 100 milliseconds in between each message casted. This is because connections are asynchronous so to check if the connection with the peer was made we need to wait until it's established.

PR #5221, opened 2025-11-06
fix(l2): use dummy peer handler

avilagaston9:
**Motivation**

In #5004 , the `PeerHandler` was modified to receive the `Initiator` GenServer handle. As a result, our L2 initializer was updated to spawn the `Initiator` GenServer,  which led to the following log being repeatedly printed in our L2 node:

```
2025-11-06T14:57:09.120191Z  INFO ethrex_p2p::discv4::peer_table: Resetting list of tried peers.
```

**Description**

- Since we don't have P2P in L2, this PR updates the L2 initializer to use the `PeerHandler::dummy()` method instead.
- Even though this works for now, we should review our `init_rpc_api`, as we should be able to start the `RPC API` without using a `PeerHandler` at all. #5222 was created to review this.

Closes None

### DIFF
--- a/cmd/ethrex/l2/initializers.rs
+++ b/cmd/ethrex/l2/initializers.rs
@@ -17,7 +17,7 @@ use ethrex_p2p::{
     discv4::peer_table::PeerTable,
     network::P2PContext,
     peer_handler::PeerHandler,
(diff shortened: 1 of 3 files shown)


tomip01 (approved):

avilagaston9:
Addressed in #5253

PR #5224, opened 2025-11-06
chore(l2): implement rkyv traits for `Node`, serialize them directly into the prover guest program instead of its RLP encoding

xqft:
**Motivation**

Directly deserializing the `Node` struct in the prover's guest program allows skipping RLP decoding and resolving the subtries (this is, converting the list of nodes into the tree structure). Though this comes at a cost:

1. Now we need to encode the nodes to hash them
2. Now deserialization is a lot more expensive because rkyv calls a `memcpy` for unsized types (like smart pointers, `Arc` in our case).

Even considering these, the net change is negative: 697k cycles -> 533k cycles, 24% reduction.

This PR also removes the double hashing of nodes (previously we were hashing once at the start and then hashing again to populate the hash cache of each node)

**Description**

- implements the `rkyv` traits for `Node` and other relevant types
- changes the execution witness to contain a list of `Node` instead of encoded nodes.
- misc. needed refactors to adapt the new execution witness

**Flamegraphs**

before:
image
after:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3126,7 +3126,7 @@ dependencies = [
  "const-hex",
  "elliptic-curve",
  "ethabi",
- "generic-array 0.14.9",
+ "generic-array 0.14.7",
  "k256",
  "num_enum 0.7.5",
  "once_cell",
--- a/crates/l2/tee/quote-gen/Cargo.lock
+++ b/crates/l2/tee/quote-gen/Cargo.lock
@@ -2500,6 +2500,7 @@ dependencies = [
  "ethrex-threadpool",
  "hex",
  "lazy_static",
+ "rkyv",
  "rustc-hash",
  "serde 1.0.228",
  "serde_json",


crates/common/types/block_execution_witness.rs
```diff
@@ -44,7 +38,7 @@ pub struct GuestProgramState {
     /// program execution.
     /// The trie is built during guest program execution inside the zkVM,
     /// before the stateless validation.
-    pub state_trie: Option<Trie>,
+    pub state_trie: Trie,
```
ilitteri, line 41: Should we update the docs?

crates/common/trie/trie.rs
```diff
@@ -315,21 +316,19 @@ impl Trie {
...
+            all_nodes: &BTreeMap<H256, Node>,
+            cur_node: &Node,
         ) -> Result<Node, TrieError> {
-            let cur_node = Node::decode(cur_node_rlp)?;
-
-            Ok(match cur_node {
+            Ok(match cur_node.clone() {
                 Node::Branch(mut node) => {
                     for choice in &mut node.choices {
                         let NodeRef::Hash(hash) = *choice else {
```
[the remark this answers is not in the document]
xqft: the purpose of this function is now to embed all non-embedded nodes, regardless of the already embedded ones. If a node contains an invalid reference that problem is not in the scope of this function

crates/common/types/block_execution_witness.rs
xqft: done!

PR #5228, opened 2025-11-06
refactor(l1): change p2p_enabled flag to p2p_disabled flag

### LINKED ISSUE #4979: Replace `--p2p.enabled` flag with `--p2p.disabled` or similar
We currently have a flag to enable P2P, which is already enabled by default (essentially a no-op). We should remove it and add a similar flag to **disable** P2P.

cdiielsi:
**Motivation**

`p2p_enabled` is a no op since it's the default configuration, so there's no need for it.

**Description**

This pr replaces the `p2p_enabled` flag with a `p2p_disabled` flag in case someone wants to disable p2p. 

Closes #4979

### DIFF
--- a/crates/l2/based/README.md
+++ b/crates/l2/based/README.md
@@ -183,7 +183,6 @@ cargo run --release --manifest-path ../../Cargo.toml --bin ethrex -- l2 \
   --l1.on-chain-proposer-address $ETHREX_COMMITTER_ON_CHAIN_PROPOSER_ADDRESS \
   --l1.bridge-address $ETHREX_WATCHER_BRIDGE_ADDRESS \
   --based \
-  --p2p.enabled \
   --p2p.port 30303 \
   --discovery.port 30303
 ```


cmd/ethrex/cli.rs
```diff
@@ -285,7 +284,6 @@ impl Options {
             authrpc_addr: "localhost".into(),
             authrpc_port: "8551".into(),
             authrpc_jwtsecret: "jwt.hex".into(),
-            p2p_enabled: true,
```
cdiielsi, line 288: I opted for implementing the opposite behaviour to this and have p2p_disable be set with the default value false, but isn't it that p2p isn't necessary for L2? Shouldn't I set `p2p_disabled: true`?

SDartayet: I'm pretty sure L2 does require P2P, since the P2P module has an RLPX module related to L2 P2P networking

PR #5232, opened 2025-11-07
fix(l1): emit warning upon batch execution failure during import

fmoletta:
**Motivation**
PR #5204 added batch execution to the `import` subcommand, but omitted logging a warning when a block failed to execute within the batch.
This PR adds this log line and its associated logic

**Description**
* Emit a warning if a block fails to execute within batch execution during import

Closes #issue_number

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -587,7 +587,12 @@ pub async fn import_blocks(
                     blockchain
                         .add_blocks_in_batch(mem::take(&mut block_batch), CancellationToken::new())
                         .await
-                        .map_err(|(err, _)| err)?;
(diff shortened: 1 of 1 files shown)


cmd/ethrex/cli.rs
```diff
@@ -587,7 +587,12 @@ pub async fn import_blocks(
                     blockchain
                         .add_blocks_in_batch(mem::take(&mut block_batch), CancellationToken::new())
                         .await
-                        .map_err(|(err, _)| err)?;
+                        .inspect_err(|err| match err {
```
mpaulucci, line 590: Wonder if it isn't better just to revert https://github.com/lambdaclass/ethrex/pull/5204

Issue #4849, opened 2025-10-13
Update config to remove blob schedules that don't change blobs


mpaulucci:
https://github.com/eth-clients/hoodi/pull/23
https://github.com/eth-clients/sepolia/pull/119
https://github.com/eth-clients/holesky/pull/136

Osaka fork does not change the amount of blobs, so its been removed from the blob schedule. The idea is that the bpo forks should be different from the regular forks. We should account for this logic.

PR #5233, opened 2025-11-07
refactor(l1): refactor chainconfig

### LINKED ISSUE #4720: Improve the way we obtain the next and last forks
Stemming from [this discussion](https://github.com/lambdaclass/ethrex/pull/4625#discussion_r2391563931)

We should probably do a significant refactor to the way we handle forks. Namely, what this would involve would be:

- Having a way to, having a fork, obtain the next fork
- Having a way to obtain the latest fork out of a set
- Rather than having a separate field for the activation timestamps/block numbers of each fork, having an array that's indexed by the fork number

This would make obtaining the next fork for a given chain config, as well as the latest fork, much easier; as well as making the code more maintainable in the long run.

SDartayet:
Motivation

Improve the maintainability of our code with regards to future forks.

Description

Currently we have a lot of match statements spread throughout various functions to get the current fork, blobschedule, etc; and separate functions for checking the activation of each fork. This creates serious maintainability issues down the line. This PR centralized the match statements in two functions: one to get the activation timestamp/block for a fork, and one to get the blob schedule for a fork, and rewrites the rest of the functions accordingly. All the funcitons to check fork activation were also aggregated into a single `is_fork_activated` function that takes the fork as a parameter and checks its activation. The implementation of `get_blob_schedule_for_time` also fetches the most recent active blob schedule if the current fork specifies none, which is part of tackling #4849.

Additionally, some code related to pre-Merge logic was removed (checks for EIP 155 and EIP 2028)

Closes #4720


crates/common/types/genesis.rs
```diff
@@ -282,43 +283,68 @@ lazy_static::lazy_static! {
...
-    Petersburg = 8,
-    Istanbul = 9,
-    MuirGlacier = 10,
-    Berlin = 11,
-    London = 12,
-    ArrowGlacier = 13,
-    GrayGlacier = 14,
-    Paris = 15,
-    Shanghai = 16,
+    Homestead = 0,
```
mpaulucci: why is this change of names?

crates/common/types/genesis.rs
```diff
@@ -282,43 +283,68 @@ lazy_static::lazy_static! {
...
+    Petersburg = 7,
+    Istanbul = 8,
+    MuirGlacier = 9,
+    Berlin = 10,
+    London = 11,
+    ArrowGlacier = 12,
+    GrayGlacier = 13,
+    Paris = 14,
+    Shanghai = 15,
     #[default]
```
mpaulucci, line 303: we probably shouldn't have defaults

crates/vm/levm/src/environment.rs
```diff
@@ -63,10 +63,10 @@ impl EVMConfig {
     }
 
     pub fn new_from_chain_config(chain_config: &ChainConfig, block_header: &BlockHeader) -> Self {
-        let fork = chain_config.fork(block_header.timestamp);
+        let fork = chain_config.get_fork(block_header.timestamp);
 
         let blob_schedule = chain_config
-            .get_fork_blob_schedule(block_header.timestamp)
+            .get_current_blob_schedule(block_header.timestamp)
```
mpaulucci: `current` is deceiving if its calculated based on a timestamp

crates/common/types/genesis.rs
SDartayet: I changed the enum so the forks would be the ones mentioned in labels for the activation block numbers and timestamps. This was necessary ƒor the match statements in `fork_activation_time_or_block` and `is_fork_activated` to work properly. While I could have made it a separate enum, I think it makes more sense for the fork enums to reflect the forks as laid out in the chain config files (and in practice only the chain config struct methods and the tests use the fork enum, so it shouldn't create broader problems)

crates/common/types/genesis.rs
SDartayet: I tried to do this but it breaks other structs that have fork as a field and implement default, like parts of the LEVM runner or EVMConfig. Unless we want to remove those defaults too it might be better to leave this as is (so fork defaults aren't spread out throughout the code)

crates/vm/levm/src/environment.rs
SDartayet: Sure. Addressed [here](https://github.com/lambdaclass/ethrex/pull/5233/commits/f2a880ddacb455c03a594bb7ae7c6f8a3844347b) (alongside some fixes for other things that were failing)

crates/blockchain/constants.rs
```diff
@@ -18,9 +18,6 @@ pub const TX_ACCESS_LIST_ADDRESS_GAS: u64 = 2400;
 // Gas cost for each storage key specified on access lists
 pub const TX_ACCESS_LIST_STORAGE_KEY_GAS: u64 = 1900;
 
-// Gas cost for each non zero byte on transaction data
```
mpaulucci, line 21: this change seems unrelated to the PR

crates/common/types/genesis.rs
mpaulucci: it kinda sucks, but if you have to keep it, at least choose the (soon to be) current one, Osaka

crates/blockchain/payload.rs
```diff
@@ -553,16 +555,6 @@ impl Blockchain {
             // TODO: maybe fetch hash too when filtering mempool so we don't have to compute it here (we can do this in the same refactor as adding timestamp)
             let tx_hash = head_tx.tx.hash();
 
-            // Check whether the tx is replay-protected
```
mpaulucci, line 555: state tests don't use this?

crates/blockchain/constants.rs
SDartayet: Ah, I also took the opportunity to remove code that was only relevant for pre-Merge forks. This change is related to that. I should have clarified that in the PR description; will add it now.

crates/blockchain/payload.rs
SDartayet: They aren't failing on the CI and they aren't failing locally either, so preseumably not.

crates/common/types/genesis.rs
```diff
@@ -421,134 +449,39 @@ impl ChainConfig {
...
+            .rfind(|fork| self.is_fork_activated(*fork, block_timestamp))
+            .unwrap_or(Paris)
+    }
+
+    pub fn get_blob_schedule_for_time(&self, block_timestamp: u64) -> Option<ForkBlobSchedule> {
+        if let Some(fork_with_current_blob_schedule) = FORKS.into_iter().rfind(|fork| {
+            self.get_blob_schedule_for_fork(*fork).is_some()
+                && self.is_fork_activated(*fork, block_timestamp)
+        }) {
+            self.get_blob_schedule_for_fork(fork_with_current_blob_schedule)
```
fmoletta, line 463: Do we need to iterate all forks? Can't we just fetch the blob schedule for the active fork?

SDartayet: Ah, it's a part of tackling #4849. Recently it was decided that if a fork doesn't change the blob schedule, it won't have a blob schedule field in the genesis file, and the blob schedule from the most recent active fork should be used instead. This implements that logic; rfind short circuits as soon as it finds an element of the array that fulfills the closure conditions so this shouldn't iterate through too much of the array in practice

crates/blockchain/constants.rs
mpaulucci: ok

crates/common/types/genesis.rs
```diff
@@ -310,15 +311,43 @@ pub enum Fork {
     BPO5 = 24,
 }
 
+pub const FORKS: [Fork; 25] = [
```
mpaulucci, line 314: Why do you need this? `Fork` already has a number so you should be able to use that for ordering?

crates/common/types/genesis.rs
```diff
@@ -343,53 +372,53 @@ impl From<Fork> for &str {
...
+            Fork::BPO3 => self.blob_schedule.bpo3,
+            Fork::BPO4 => self.blob_schedule.bpo4,
+            Fork::BPO5 => self.blob_schedule.bpo5,
+            _ => None,
+        }
     }
 
-    pub fn is_eip155_activated(&self, block_number: BlockNumber) -> bool {
-        self.eip155_block.is_some_and(|num| num <= block_number)
+    pub fn is_fork_activated(&self, fork: Fork, timestamp_or_block: u64) -> bool {
```
mpaulucci, line 419: why do we need to support blocks here? Is this ever called with a block?

crates/common/types/genesis.rs
SDartayet: We do, but it helps with either iterating through forks (when, for example, we want the latest forks that fulfills a certain condition or other; we use this to get the latest scheduled fork, the current one, etc)

crates/common/types/genesis.rs
SDartayet: I don't think it ever should be but it's more of a syntax thing if anything since we want to cover the block number based forks in the match arms too (and it doesn't change functionality). If you find it better I could just make it return true for all forks pre-Merge

SDartayet: Actually I realized there is an edge case: if a call to the eth_config RPC endpoint happens before the node has synced to head, we need to check which is the currently active fork to know whether to respond or not (since the response format isn't defined for forks pre-cancun), and that requires iterating through the forks and checking which is the most recent one that's active

crates/common/types/genesis.rs
mpaulucci: you should be able to implement iterator on the other structure? I don't think we need this

crates/common/types/genesis.rs
mpaulucci: current implementation. is error prone. You should either only accept timestamps or create a type BlockNumberOrTimestamp, or something like that and deal accordingly

crates/common/types/genesis.rs
SDartayet: Not really. You can't iterate through enum variants by default. Iterating through numbers and converting them to forks wouldn't work either without implementing a from function that matches numbers to forks, but I could do that if you think it's better

crates/common/types/genesis.rs
```diff
@@ -421,134 +449,39 @@ impl ChainConfig {
...
+            FORKS.into_iter().find(|fork| {
+                *fork > current_fork && self.fork_activation_time_or_block(*fork).is_some()
+            })
         } else {
             None
-        };
-        match next {
-            Some(fork) if fork > self.fork(block_timestamp) => next,
-            _ => None,
         }
```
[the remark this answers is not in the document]
mpaulucci: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/lambdaclass/ethrex/pull/5233#discussion_r2534740808)

PR #5236, opened 2025-11-07
feat(l1): new initial_lookup_interval and lookup_interval cli args

### LINKED ISSUE #5148: Create a CLI argument to set up the Discovery and Connetion initiation Lookup Rate
We have that Interval [hardcoded](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/discv4/server.rs#L42-L43), and it can be useful to be able to parametrize it in the command line to test different configurations. Let's create a CLI arg for it

cdiielsi:
**Motivation**

It can be useful to have the `INITIAL_LOOKUP_INTERVAL` and `LOOKUP_INTERVAL` constants as parameters in the command line to test different configurations. 

**Description**

This pr introduces `initial_lookup_interval` and `lookup_interval` as new cli parameters.

Closes #5148

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -792,6 +792,7 @@ impl Blockchain {
         })
     }
 
+    #[instrument(level = "trace", name = "Block DB update", skip_all)]
     pub fn store_block(
         &self,
         block: Block,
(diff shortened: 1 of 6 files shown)


mpaulucci:
@copilot suggest better names for the cli params

mpaulucci:
After some discussion we decided to improve this in another way

PR #5243, opened 2025-11-07
perf(levm): try improve mulmod and addmod

JereSalo:
**Motivation**

**Description**

- Try see if there's any improvement if we stop using `U512` for these operations. We executed with a zkVM and we saw that it can be a bottleneck. This is experimental

WARNING: This is VERY experimental, it probably won't be merged so don't bother to review this for now haha


JereSalo:
This was a PoC, didn't continue because it wasn't a priority. For zkVMs the right approach is to use the precompile that each one offers. We already use this for ZisK.

PR #5244, opened 2025-11-08
chore(l2): include zkVM ELFs and VKs in the releases

ilitteri:
**Motivation**

We need releases of ethrex's ELFs and VKs to use independently for proving L1 blocks execution.

**Description**

- Adds a `build-ethrex-guest` workflow to run with a mattrix strategy (for sp1 and risc0) that builds the `guest_program` crate with the features `sp1` and `risc0`. We need this as a separate job and cannot reuse the builds from ubuntu-22.04 for l2_gpu because the guest program being built there is different than the one we need for statelessly verifying L1 blocks.
- Adds a `package-ethrex-guests` workflow to create `ethrex-guests.tar.gz` compressing SP1 and RISC0 elf and vks.
- RISC0's ELF file is now generated.

### DIFF
--- a/.github/workflows/tag_release.yaml
+++ b/.github/workflows/tag_release.yaml
@@ -246,7 +246,7 @@ jobs:
       - name: Archive ethrex guests
         run: |
           cd ethrex_guests/
-          tar -czvf ethrex-guests.tar.gz .
+          tar -czvf ../ethrex-guests.tar.gz .
 
       - name: Upload ethrex guests artifact
         uses: actions/upload-artifact@v4


.github/workflows/tag_release.yaml
```diff
@@ -234,7 +298,13 @@ jobs:
...
         with:
           path: ./bin
-          pattern: "ethrex*"
+          pattern: "ethrex*" # This includes the binaries and elf files
+
+      - name: Download verification keys
+        uses: actions/download-artifact@v4
+        with:
+          path: ./bin
+          pattern: "*_verification_keys" # sp1_verification_keys, risc0_verification_keys
```
ilitteri: I'm not sure if the pattern needs to be instead "\*_verification_keys/\*"

ManuelBilbao: I'd put VKs and ELFs in a .tar file so we don't add 5 more files

ilitteri: [1e5ba0](https://github.com/lambdaclass/ethrex/pull/5244/commits/e1e5ba071888241037e92ff9dde78d951184526c)

.github/workflows/tag_release.yaml
```diff
@@ -149,6 +149,114 @@ jobs:
...
+
+  # Creates ethrex-guests.tar.gz artifact including both SP1 and
+  # RISC0 elf and verification keys.
+  package-ethrex-guest:
+    needs:
+      - build-ethrex-guest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
ManuelBilbao: This is not needed

ManuelBilbao (approved):

.github/workflows/tag_release.yaml
```diff
@@ -221,8 +326,10 @@ jobs:
   finalize-release:
     needs:
       - build-ethrex
+      - build-ethrex-guest
```
tomip01, line 329: I think this is not needed since it is required in `package-ethrex-guest`

PR #5245, opened 2025-11-08
test(l1): verify we reject invalid iat claims.

### LINKED ISSUE #5074: Test jwt claims are correctly validated
It seems the crate we are using doesn't check the `iat` claim. We should test that the claims are properly validated and fail when they're invalid.

Image

We can check the jwt.io webpage to find libraries that support "iat" validation: https://www.jwt.io/libraries?programming_language=rust

lakshya-sky:
**Motivation**

jsonwebtoken crate doesn't validate `iat` claims even if we explicitly ask it, hence we needed tests.

**Description**

Added unit tests to correctly reject invalid `iat` claims. 

Closes #5074

### DIFF
--- a/crates/networking/rpc/authentication.rs
+++ b/crates/networking/rpc/authentication.rs
@@ -79,7 +79,8 @@ mod tests {
     }
 
     #[test]
-    fn test_iat_missing() {
+    fn test_iat_missing_fails() {
+        // Our `Claims` type expect `iat` so JWTs with it would simply fail to deserialize.
         let secret = Bytes::from("my_secret_key");
         let faulty_claims = FaultyClaims {
             id: None,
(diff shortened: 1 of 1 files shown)


mpaulucci (approved):
LGTM

Oppen:
Are the tests running? They should fail, given the documented behavior of the library.

lakshya-sky:
Hey @Oppen, yes tests are running here at https://github.com/lambdaclass/ethrex/actions/runs/19242301387/job/55007724747?pr=5245#step:5:843

As you can see when deserialize the claims part of the token we expect iat claim, if its not there it would throw error. When iat is available and its value is not within 60s range then we throw error. 
So its safe to assume that we are checking it correctly.

If the library were to check the iat claims it would do the same as we are doing it here.

Oppen:
> Hey @Oppen, yes tests are running here at https://github.com/lambdaclass/ethrex/actions/runs/19242301387/job/55007724747?pr=5245#step:5:843
> 
> As you can see when deserialize the claims part of the token we expect iat claim, if its not there it would throw error. When iat is available and its value is not within 60s range then we throw error. So its safe to assume that we are checking it correctly.
> 
> If the library were to check the iat claims it would do the same as we are doing it here.

Yes, a teammate showed me it's being checked externally. Not obvious, but it works.

MegaRedHand:
@lakshya-sky thanks for the contribution, but commits need to be signed. See https://github.com/lambdaclass/ethrex/blob/main/CONTRIBUTING.md#commit-signature-verification for how to do it.

lakshya-sky:
Hey @MegaRedHand, I've fixed the commits but it added two new reviewers while fixing it. Could you guys take a look again. Thanks!

PR #5248, opened 2025-11-10
perf(l1): (testing) try removing bloom filter to see if improves perf

edg-l:
**Motivation**

**Description**

Closes #issue_number


edg-l:
Note: i ran it on notif-12 mainnet, it didnt seem to increase perf and went down to about 140mgas/s

PR #5249, opened 2025-11-10
chore(levm): avoid duplicate BackupHook in L2 stateless_execute

Galoretka:
**Motivation**

Redundant BackupHook in L2 when using stateless_execute.

**Description**

This change prevents adding a second BackupHook during stateless_execute when running in L2 mode. L2 already installs BackupHook via its default hook set, so adding another caused redundant finalize_execution work and a second identical tx_backup write. Restricting the injection to VMType::L1 preserves stateless behavior for L1 while removing unnecessary duplication for L2, with no functional change to execution semantics.

### DIFF
--- a/crates/vm/levm/src/vm.rs
+++ b/crates/vm/levm/src/vm.rs
@@ -507,7 +507,9 @@ impl<'a> VM<'a> {
     /// Executes without making changes to the cache.
     pub fn stateless_execute(&mut self) -> Result<ExecutionReport, VMError> {
         // Add backup hook to restore state after execution.
-        self.add_hook(BackupHook::default());
+        if matches!(self.vm_type, VMType::L1) {
+            self.add_hook(BackupHook::default());
+        }
         let report = self.execute()?;
         // Restore cache to the state before execution.
         self.db.undo_last_transaction()?;


crates/vm/levm/src/vm.rs
```diff
@@ -507,7 +507,9 @@ impl<'a> VM<'a> {
     /// Executes without making changes to the cache.
     pub fn stateless_execute(&mut self) -> Result<ExecutionReport, VMError> {
         // Add backup hook to restore state after execution.
-        self.add_hook(BackupHook::default());
+        if matches!(self.vm_type, VMType::L1) {
```
ElFantasma, line 510: This couples `stateless_execute` to knowledge about which hooks each `VMType` installs. If someone later removes `BackupHook` from `l2_hooks()` but forgets to update this guard, L2 `stateless_execute` would silently break (no backup → `undo_last_transaction` fails).

A more resilient alternative would be to check whether a `BackupHook` is already present in `self.hooks` before adding one, e.g. via a helper like `self.has_hook::<BackupHook>()`. That said, for a codebase this size the coupling is manageable — just flagging it as something to be aware of.

ElFantasma (changes requested):
**Staleness:** This PR is from Nov 2025 and the diff targets line 507, but on current `main` `stateless_execute` is at line 662. The fix is structurally identical — just needs a rebase.

MegaRedHand:

PR #5254, opened 2025-11-10
fix(levm)!: fix jumpdests for large initcodes

iovoid:
> [!CAUTION]
> This changes the serialization of `Code`, and thus requires re-syncing.

**Motivation**

While executing the state tests in debug mode, [a debug assert](https://github.com/lambdaclass/ethrex/blob/45a9871498bd155867c8e9a9bf86e73fc634841a/crates/common/types/account.rs#L44) was triggered.

The calculation of valid jumpdests assumes the code will be less than 2^16 bytes in length, however this isn't true before Shanghai.

**Description**

This PR adds support for up to 2^32 bytes of initcode.

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -21,7 +21,9 @@ use tokio_util::sync::CancellationToken;
 use tracing::{Level, info, warn};
 
 use crate::{
-    initializers::{get_network, init_blockchain, init_store, init_tracing, load_store},
+    initializers::{
+        get_network, init_blockchain, init_store, init_tracing, load_store, regenerate_head_state,
+    },
     utils::{self, default_datadir, get_client_version, get_minimal_client_version, init_datadir},
 };
 
@@ -350,6 +352,22 @@ pub enum Subcommand {
         l2: bool,
     },
     #[command(
+        name = "import-bench",
+        about = "Import blocks to the database for benchmarking"
+    )]
+    ImportBench {
+        #[arg(
+            required = true,
+            value_name = "FILE_PATH/FOLDER",
+            help = "Path to a RLP chain file or a folder containing files with individual Blocks"
+        )]
+        path: String,
+        #[arg(long = "removedb", action = ArgAction::SetTrue)]
+        removedb: bool,
+        #[arg(long, action = ArgAction::SetTrue)]
+        l2: bool,
+    },
+    #[command(
         name = "export",
         about = "Export blocks in the current chain into a file in rlp encoding"
     )]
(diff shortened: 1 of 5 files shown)


crates/storage/store_db/rocksdb.rs
```diff
@@ -884,7 +884,7 @@ impl StoreEngine for Store {
         }
 
         for (code_hash, code) in update_batch.code_updates {
-            let mut buf = Vec::with_capacity(6 + code.bytecode.len() + 2 * code.jump_targets.len());
+            let mut buf = Vec::with_capacity(6 + code.bytecode.len() + 4 * code.jump_targets.len());
```
pablodeymo: I think it's better to use size_of to avoid hard-coding this lenght

crates/storage/store_db/rocksdb.rs
```diff
@@ -1306,8 +1306,8 @@ impl StoreEngine for Store {
             hash: code_hash,
             bytecode: Bytes::copy_from_slice(bytecode),
             jump_targets: targets
-                .chunks_exact(2)
-                .map(|c| u16::from_le_bytes([c[0], c[1]]))
+                .chunks_exact(4)
+                .map(|c| u32::from_le_bytes([c[0], c[1], c[2], c[3]]))
```
pablodeymo: Talking with @Oppen, we think that it's better to encode the jump_target in RLP, which will be smaller than 32 bits for the common case.

crates/common/types/account.rs
```diff
@@ -24,8 +24,8 @@ use crate::{
 pub struct Code {
     pub hash: H256,
     pub bytecode: Bytes,
-    // TODO: Consider using Arc<[u16]> (needs to enable serde rc feature)
-    pub jump_targets: Vec<u16>,
+    // TODO: Consider using Arc<[u32]> (needs to enable serde rc feature)
```
JereSalo, line 27: Can we add a comment here or anywhere you consider convenient about the reason why we use 32 bits? So nobody changes it to u16 back again
If you want you can link [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860) but it's not necessary

Oppen: No, it would be a good idea to link to the EIP. Also to the failing EEST test.

Oppen: When in doubt, overcommenting is better than undercommenting. Don't let school tell you otherwise.

crates/storage/store_db/rocksdb.rs
Oppen: Better is a strong word for any use of RLP. But in the tradeoff IO probably wins, and RLP is better for IO than always using all the bytes.

crates/storage/store_db/rocksdb.rs
iovoid: Improved in 9d129ce10

crates/storage/store_db/rocksdb.rs
iovoid: Done in 084c8e2a4

crates/common/types/account.rs
iovoid: Added comment in da7786879

Issue #5234, opened 2025-11-07
Print warning if we don't receive any message from CL in a while


MegaRedHand:
I remember other clients printing a warning after they go a while without receiving orders from a consensus client. We should do the same.

PR #5256, opened 2025-11-10
feat(l1): log warning if no CL messages are received in a while

SDartayet:
**Motivation**

Making it clearer when the client isn't working properly

**Description**

This PR adds an extra thread which keeps track of whether a message from the consensus layer hasn't been received in the past 30 seconds. This is done by having the auth RPC handler communicate with it when a new message is received to reset the countdown. If no messages have been received after 30 seconds, a message is printed to warn the user, along with some hints as to possible reasons why Ethrex isn't detecting the consensus client.

Closes [#5234](https://github.com/lambdaclass/ethrex/issues/5234)

### DIFF
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -308,7 +308,7 @@ pub async fn start_api(
         }
     });
 
-    let authrpc_handler = async move |ctx, auth, body| {
+    let authrpc_handler = move |ctx, auth, body| async move {
         let _ = timer_sender.send(());
         handle_authrpc_request(ctx, auth, body).await
     };


crates/networking/rpc/rpc.rs
```diff
@@ -296,7 +297,24 @@ pub async fn start_api(
...
     info!("Starting HTTP server at {http_addr}");
 
-    let authrpc_handler = |ctx, auth, body| async { handle_authrpc_request(ctx, auth, body).await };
+    let (timer_sender, mut timer_receiver) = tokio::sync::watch::channel(());
+
+    tokio::spawn(async move {
+        loop {
+            let result = timeout(Duration::from_secs(30), timer_receiver.changed()).await;
+            if result.is_err() {
+                error!("No messages from the consensus layer. Is the consensus client running?
```
SDartayet: I wasn't sure whether to make this `warn` or `error` level. I figured `error` was more appropriate for end user scenarios, since Ethrex would always need to run alongside a consensus client, but it could get cumbersome in testing scenarios where the client is run with no consensus client.

mpaulucci: warn is better

crates/networking/rpc/rpc.rs
```diff
@@ -298,7 +297,24 @@ pub async fn start_api(
...
+                 Check the auth JWT coincides with the one used by the CL and the auth RPC address and port used by it and Ethrex match."
```
mpaulucci: This is probably too much

crates/networking/rpc/rpc.rs
```diff
@@ -296,7 +297,24 @@ pub async fn start_api(
...
+                );
```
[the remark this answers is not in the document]
mpaulucci: this should be addressed?

PR #5261, opened 2025-11-10
ci(l1,l2): add a workflow for testing both L1 & L2 dev

ilitteri:
**Motivation**

L1 dev was not being tested independently of any changes in the L2.

**Description**

The integration test that aims to test ethrex L2 dev is suited for fulfilling this purpose.

### DIFF
--- a/.github/workflows/pr-main_l1_l2_dev.yaml
+++ b/.github/workflows/pr-main_l1_l2_dev.yaml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Start ethrex
         run: |
-          cargo run --release --bin ethrex -F l2,l2-sql -- l2 --dev \
+          COMPILE_CONTRACTS=1 cargo run --release --bin ethrex -F l2,l2-sql -- l2 --dev \
           --block-producer.base-fee-vault-address 0x000c0d6b7c4516a5b274c51ea331a9410fe69127 \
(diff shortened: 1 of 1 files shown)


jrchatruc (approved):

.github/workflows/pr-main_l1_l2_dev.yaml
```diff
@@ -0,0 +1,77 @@
+name: L1 & L2 Dev
```
mpaulucci: This is clearly an L2 tests, so I would call it `L2 Dev`

ilitteri:
```suggestion
name: L2 Dev
```

ilitteri: Done [3407671](https://github.com/lambdaclass/ethrex/pull/5261/commits/34076710e0bb71d5bfb42eea6ce00be26894b130).

.github/workflows/pr-main_l1_l2_dev.yaml
```diff
@@ -0,0 +1,77 @@
+name: L2 Dev
```
MegaRedHand, line 1: We should document that this is used on L1 to make sure `--dev` works correctly

PR #5265, opened 2025-11-11
docs(l1,l2): improve release process docs second step

ilitteri:
**Motivation**

The second step of the release process docs was missing some details.

**Description**

- Clarify which `Cargo.toml` file versions need to be bumped.
- Specify to run `cargo update` in the corresponding workspaces.
- Add `quote-gen` workspace to the list of workspaces to bump.

### DIFF
--- a/docs/developers/release-process.md
+++ b/docs/developers/release-process.md
@@ -31,6 +31,8 @@ After updating the version in the `Cargo.toml` files, we need to update the `Car
 - `crates/l2/prover/src/guest_program/src/risc0`
 - `crates/l2/tee/quote-gen`
 
+Then, go to the `CLI.md` file located in `docs/` and update the version of the `--builder.extra-data` flag default value to match the new version (for both ethrex and ethrex l2 sections).
+
 Finally, stage and commit the changes to the release branch.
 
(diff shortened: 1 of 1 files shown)


docs/developers/release-process.md
```diff
@@ -14,17 +14,26 @@ Examples:
...
 
-There are currently three `Cargo.lock` files that will be affected. Make sure you check them:
+- `crates/l2/prover/src/guest_program/src/sp1/Cargo.toml`
+- `crates/l2/prover/src/guest_program/src/risc0/Cargo.toml`
+- `crates/l2/tee/quote-gen/Cargo.toml`
 
-- root `Cargo.lock`
-- `sp1/Cargo.lock`
-- `risc0/Cargo.lock`
+After updating the version in the `Cargo.toml` files, we need to update the `Cargo.lock` files to reflect the new versions. Run `cargo tree` in their respective directories:
```
MegaRedHand, line 27: Does `cargo tree` update the lockfile?

ilitteri: Yes, it's even safer than `cargo update` because it doesn't update dependencies that you don't want to be updated.

ManuelBilbao:
We should also add a step changing the version in the CLI reference docs (`CLI.md` file)

PR #5266, opened 2025-11-11
perf(l1): rebuild_bloom - insert instead of merge

Arkenan:
Merging layer filters is very costly due to the cost of iterating the elements of a filter. Inserting has proven to be cheaper for the size of the filter we're using and the amount of elements we are holding. Some measurements:

- Early hoodi imports went from 8 blocks per second to about 80 (10x).
- Synced hoodi rebuild time went from ~166ms to ~27ms, (about a 6x). 

The PR also simplifies the rebuild code and avoids using rayon entirely.


crates/storage/trie_db/layering.rs
```diff
@@ -126,52 +127,28 @@ impl TrieLayerCache {
...
-                        return None;
-                    }
+        let mut new_global_filter = Self::create_filter().unwrap();
+
+        self.layers.iter().for_each(|(_, layer)| {
+            layer.nodes.iter().for_each(|(path, _)| {
+                if let Err(qfilter::Error::CapacityExceeded) = new_global_filter.insert(path) {
+                    tracing::warn!("TrieLayerCache: rebuild_bloom capacity exceeded");
+                    self.bloom = None;
+                    return;
```
MegaRedHand, line 152: Wouldn't this just return from the closure? I think we should use `for`-loops here, or change to `try_for_each` and use a `Result` here.

Arkenan: Changed to for loops. Forgot they existed for a second.

PR #5267, opened 2025-11-11
refactor(l1): unify fullsync_headers table and headers table

### LINKED ISSUE #4903: Unify `headers` and `fullsync_headers` tables
#4812 adds a new table for temporarily storing unprocessed headers in full-sync. This table could be merged with the main headers table, to avoid having to move headers to the main table during full-sync.

I think this will simplify things and remove the (wrong since snap-sync was implemented) invariant of "we have a header in DB -> we have its block in DB". An example of this invariant at play is the gas estimator:

```
2025-10-17T09:09:50.010242Z ERROR ethrex_rpc::eth::gas_tip_estimator: Block body for block number 23594386 is missing but is below the latest known block!
```

cdiielsi:
**Motivation**

**Description**

Closes #4903

### DIFF
--- a/crates/storage/store_db/in_memory.rs
+++ b/crates/storage/store_db/in_memory.rs
@@ -491,18 +491,23 @@ impl StoreEngine for Store {
         Ok(self.inner()?.headers.get(&block_hash).cloned())
     }
 
-    fn get_block_header_by_number(
+    fn get_block_headers_by_number(
         &self,
         block_number: BlockNumber,
-    ) -> Result<Option<BlockHeader>, StoreError> {
+    ) -> Result<Vec<BlockHeader>, StoreError> {
         let store = self.inner()?;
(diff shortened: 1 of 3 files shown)


crates/storage/store_db/in_memory.rs
```diff
@@ -493,6 +491,20 @@ impl StoreEngine for Store {
...
+        block_number: BlockNumber,
+    ) -> Result<Option<BlockHeader>, StoreError> {
+        let store = self.inner()?;
+        let block_header = store
+            .headers
+            .iter()
+            .filter(|(_, header)| header.number == block_number)
+            .map(|(_, header)| header)
+            .collect::<Vec<_>>();
+        Ok(Some(block_header[0].clone()))
```
cdiielsi: TODO: choose in case there are two blocks with same number

PR #5276, opened 2025-11-11
fix(l1,l2): make ethrex L1 + L2 + GPU build compile for ubuntu 22.04

ilitteri:
**Motivation**

One of the changes in a recent PR disabled the default features of the `c-kzg` crate and manually enabled a subset of them. The goal was to ensure the library imports with only 2 out of its 3 default features enabled. Unfortunately, this broke the ethrex L2 + GPU build for Ubuntu 22.04 (specifically due to a RISC0-related compilation error).
This issue went undetected because ethrex builds are only tested during release or release candidate preparation, not on every commit.

**Description**

To avoid similar regressions in the future, this PR adapts the existing GitHub Actions workflow for releases and pre-releases. The workflow will now also trigger automatically after every push to the main branch, ensuring continuous validation of key builds.

While debugging, I noticed that releases for ethrex L1 and L1 + L2 + GPU on Ubuntu 22.04 are being built with the `RUSTFLAGS='-C target-cpu=x86-64-v2'` flag, which overrides the settings in `.cargo/config.toml`. This PR removes this override to align with our configured defaults.

- Revert `c-kzg` features change: Re-enable all default features to restore compatibility without negative impacts.
- Update CI workflow: extend the release workflow to run on pushes to main for proactive build testing; and replace main_docker_publish in the process.
- Remove `RUSTFLAGS` override: eliminate the hardcoded flag from the release configuration.

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -21,7 +21,9 @@ use tokio_util::sync::CancellationToken;
 use tracing::{Level, info, warn};
 
 use crate::{
-    initializers::{get_network, init_blockchain, init_store, init_tracing, load_store},
+    initializers::{
+        get_network, init_blockchain, init_store, init_tracing, load_store, regenerate_head_state,
+    },
     utils::{self, default_datadir, get_client_version, get_minimal_client_version, init_datadir},
 };
 
@@ -350,6 +352,22 @@ pub enum Subcommand {
         l2: bool,
     },
     #[command(
+        name = "import-bench",
+        about = "Import blocks to the database for benchmarking"
+    )]
+    ImportBench {
+        #[arg(
+            required = true,
+            value_name = "FILE_PATH/FOLDER",
+            help = "Path to a RLP chain file or a folder containing files with individual Blocks"
+        )]
+        path: String,
+        #[arg(long = "removedb", action = ArgAction::SetTrue)]
+        removedb: bool,
+        #[arg(long, action = ArgAction::SetTrue)]
+        l2: bool,
+    },
+    #[command(
         name = "export",
         about = "Export blocks in the current chain into a file in rlp encoding"
     )]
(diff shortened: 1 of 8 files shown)


.github/workflows/tag_release.yaml
```diff
@@ -297,7 +302,13 @@ jobs:
 
       - name: Format name
         run: |
-          echo "TAG_VERSION=$(echo ${{ github.ref_name }} | tr -d v)" >> $GITHUB_ENV
+          # For branch builds (main) we want docker images tagged as 'main'. For tag pushes
+          # use the tag name (stripped of a leading 'v').
+          if [[ "$GITHUB_REF" == "refs/heads/main" ]]; then
```
MegaRedHand, line 305: Suggestion from copilot:

```suggestion
          if [[ ${{ github.ref_type }} == "branch" ]]; then
```

.github/workflows/tag_release.yaml
```diff
@@ -297,7 +302,13 @@ jobs:
 
       - name: Format name
         run: |
-          echo "TAG_VERSION=$(echo ${{ github.ref_name }} | tr -d v)" >> $GITHUB_ENV
+          # For branch builds (main) we want docker images tagged as 'main'. For tag pushes
```
ManuelBilbao, line 303: There's already a CI that builds the docker image for main commits. We should remove either that or this

.github/workflows/tag_release.yaml
```diff
@@ -2,8 +2,14 @@ name: Ethrex Release
...
 on:
   push:
+    # On tags, this generates a release and pushes docker images with the tag name
     tags:
       - "v*.*.*-*"
+    # On pushes to main, this silently builds and pushes docker images tagged as 'main'
+    branches:
+      - main
+      # FIXME: This is to trigger the job in the PR's CI
+      - fix_v7.0.0-rc.1
```
MegaRedHand: We could add a `workflow_dispatch` trigger, to make this run without adding patches like this

ilitteri: Yes, good idea. We have something similar in [ethrex-replay](https://github.com/lambdaclass/ethrex-replay/blob/main/.github/workflows/tag-release.yaml#L7).

.github/workflows/tag_release.yaml
ilitteri: Does this change replace entirely the `tag_latest` workflow?

ilitteri: [92079b9](https://github.com/lambdaclass/ethrex/pull/5276/commits/92079b93eec8e0a63483ddeb2380b77aba8800b5)

.github/workflows/tag_release.yaml
ilitteri: We can do that in another PR.

PR #5282, opened 2025-11-11
perf(l1): selective dirtying of trie

iovoid:
**Motivation**

Currently doing merkelization we do trie-traversal to avoid reading stale FKVs, marking the trie as dirty the first time it's changed. However, this could be done on a per-key basis.

**Description**

Keeps track of modified keys in a trie and does traversal for those.

### DIFF
--- a/.github/workflows/pr_perf_blocks_exec.yaml
+++ b/.github/workflows/pr_perf_blocks_exec.yaml
@@ -24,7 +24,7 @@ jobs:
         id: cache
         with:
           path: bin/ethrex-${{ matrix.branch }}
-          key: binary-bis-${{ github.event.pull_request[matrix.branch].sha }}
+          key: binary-${{ matrix.branch }}-${{ github.run_id }}-${{ github.run_attempt }}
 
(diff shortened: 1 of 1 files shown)


crates/common/trie/trie.rs
```diff
@@ -99,7 +100,7 @@ impl Trie {
     pub fn get(&self, pathrlp: &PathRLP) -> Result<Option<ValueRLP>, TrieError> {
         let path = Nibbles::from_bytes(pathrlp);
 
-        if pathrlp.len() == 32 && !self.dirty && self.db().flatkeyvalue_computed(path.clone()) {
```
MegaRedHand, line 102: Why did we remove this? Won't this cause some gets to fall in FKV?

pablodeymo (approved):

crates/common/trie/trie.rs
Oppen: I think this is now implied by the `contains`.

Oppen (approved):

PR #5288, opened 2025-11-12
perf(levm): cache `BLOBBASEFEE` opcode value

azteca1998:
**Motivation**

Our implementation of `BLOBBASEFEE` is slow when compared to any other client.

**Description**

The `BLOBBASEFEE` value should not change between runs with the same block header and environment, therefore it can be cached and reused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-11-12
+
+- Cache `BLOBBASEFEE` opcode value [#5288](https://github.com/lambdaclass/ethrex/pull/5288)
+
 ### 2025-11-11
 
 - Insert instead of merge for bloom rebuilds [#5223](https://github.com/lambdaclass/ethrex/pull/5223)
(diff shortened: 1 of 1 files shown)


crates/vm/levm/src/opcode_handlers/block.rs
```diff
@@ -166,8 +166,15 @@ impl<'a> VM<'a> {
         self.current_call_frame
             .increase_consumed_gas(gas_cost::BLOBBASEFEE)?;
 
-        let blob_base_fee =
-            get_base_fee_per_blob_gas(self.env.block_excess_blob_gas, &self.env.config)?;
+        let blob_base_fee = match self.blob_base_fee.get() {
```
Arkenan, line 169: nit: is this not a bit clearer if we use `OnceCell::get_or_init`?

azteca1998: Does not work because it can't handle errors. The `get_or_try_init` method is nightly-only.

Oppen: Not as much a nit, is needing the `blob_base_fee` uncommon enough to not just init it with the block, rather than using a `OnceCell`?
Besides, we already have a `&mut self` here, so we could just use an option in any case.

PR #5289, opened 2025-11-12
perf(l1): avoid copy when reading account code

iovoid:
**Motivation**

Currently we use `get_cf` for reading codes, which internally calls `get_pinned_cf` and copies the result.  This copying takes up 6% of code reading time.

**Description**

Since data needs to be deserialized first, we could use a borrowed value and avoid this copy by calling `get_pinned_cf` directly.


fmoletta (approved):

crates/storage/store_db/rocksdb.rs
```diff
@@ -1306,11 +1306,14 @@ impl StoreEngine for Store {
     }
 
     fn get_account_code(&self, code_hash: H256) -> Result<Option<Code>, StoreError> {
-        let hash_key = code_hash.as_bytes().to_vec();
-        let Some(bytes) = self.read_sync(CF_ACCOUNT_CODES, hash_key)? else {
+        let cf = self.cf_handle(CF_ACCOUNT_CODES)?;
+        let Some(bytes) = self
+            .db
+            .get_pinned_cf(&cf, code_hash.as_bytes())
+            .map_err(|e| StoreError::Custom(format!("RocksDB read error: {}", e)))?
```
Oppen, line 1313: Nit: is there no `StoreError::ReadError` or similar already?

Oppen (approved):

crates/storage/store_db/rocksdb.rs
iovoid: There is a ReadError but it's error message is "Could not open DB for reading" which implies a different kind of error, and doesn't pass-through the message.

PR #5291, opened 2025-11-12
fix(l1,l2): make ARM code a bit more portable

Oppen:
**Motivation**

Our previous solution to build errors on Linux ARM was a "make it work" approach
that duplicated all the assembly code. Making the variable parts into parameters
seems cleaner.

**Description**

- Revert "fix(l1,l2): build on ARM Linux (#5285)";
- Use a parameter to signal which kind of object file we're creating;
- Add a macro to use the proper symbol resolution syntax based on this parameter;
- Use parameters for the exported symbols, so the name mangling always matches
  the platform.


crates/common/crypto/keccak/mod.rs
```diff
@@ -1,7 +1,5 @@
-#[cfg(all(target_arch = "aarch64", target_os = "linux"))]
-std::arch::global_asm!(include_str!("keccak1600-armv8-elf.s"), options(raw));
-#[cfg(all(target_arch = "aarch64", target_os = "macos"))]
-std::arch::global_asm!(include_str!("keccak1600-armv8-macho.s"), options(raw));
+#[cfg(target_arch = "aarch64")]
+std::arch::global_asm!(include_str!("keccak1600-armv8.s"), elf=const cfg!(target_os="linux") as u32, SHA3_absorb=sym SHA3_absorb, SHA3_squeeze=sym SHA3_squeeze);
```
ElFantasma, line 2: The outer `#[cfg]` was broadened from `all(target_arch = "aarch64", target_os = "linux"/"macos")` to just `target_arch = "aarch64"`, which means this now compiles on any aarch64 platform (FreeBSD, Android, Windows, etc.). But `cfg!(target_os="linux")` only matches Linux, so on e.g. FreeBSD or Android (both ELF platforms) the `elf` flag would be `0`, triggering the Mach-O address resolution path — which will fail to assemble.

Two options:
1. Keep the broad gate but fix the flag: `elf = const cfg!(not(target_vendor = "apple")) as u32` (Mach-O is the exception, ELF is the default)
2. Restore the narrow gate: `#[cfg(all(target_arch = "aarch64", any(target_os = "linux", target_os = "macos")))]`

Option 1 is more forward-compatible.

PR #5294, opened 2025-11-12
perf(l1): avoid copying bytecode

MegaRedHand:
**Motivation**

We are spending time copying bytes from the database when reading an account's code. This isn't ideal, since RocksDB already allows us to borrow the bytes without copying them.

**Description**

This PR implementation adds an owned version of the `DBPinnableSlice` type that RocksDB has, which packs it with a reference to the DB and allows us to wrap it with `Bytes` and treat it in a generic manner.


MegaRedHand:
Since we added the code cache in #5307, this isn't needed anymore

PR #5215, opened 2025-11-06
feat(l1): add import-bench command for benchmarking block importing

fedacking:
**Motivation**

This pr aims to add a consistent command we can use to reproducible benchmark the speed of ethrex, without p2p and other concurrency problems enabled.

**Description**

- Adds an import-bench subcommand to import blocks and execute them
- Adds a makefile to autorun the bench tool
- Added documentation to the tool

**Example**

```
python3 parse_bench.py 10 13
Blocks tested 1065
Mean ggas accross multiple runs: 0.14995899843505422
Mean ggas in run: 10 0.15196807511737107
Mean ggas in run: 11 0.15213427230046964
Mean ggas in run: 12 0.14577464788732383
Mean ggas spread across blocks: 0.004736776212832552
```

PR #5299, opened 2025-11-12
chore(l1): speedup import bench

fedacking:
**Motivation**

The import bench added in #5215 had a 500 ms sleep to wait for the the commiting of the last layer of the trie. The speed of that step has been improved in #5266. We should lower the sleep to speedup testing.

**Description**

- Lowers ImportBench sleep to 100 ms


fedacking:
This may add in some edge cases noise, and it doesn't significantly speed up the test. Closing until we have the proper fix to separate the time.

PR #5305, opened 2025-11-12
refactor(levm): remove unneeded checks in swap

iovoid:
**Motivation**

Currently for swapN we do two unnecessary checks:
- `stack.len() < N`, which is also done inside `Stack::swap`
- checked add of offset (stack depth) and N, which is guaranteed to never fail as N and the offset are both bounded

**Description**

Checking that `offset + N >= STACK_DEPTH` should be enough.

### DIFF
--- a/crates/vm/levm/src/call_frame.rs
+++ b/crates/vm/levm/src/call_frame.rs
@@ -181,7 +181,7 @@ impl Stack {
         }
         #[expect(clippy::arithmetic_side_effects)]
         let index = self.offset + N;
-        
+
         if index >= self.values.len() {
             return Err(ExceptionalHalt::StackUnderflow);
         }


crates/vm/levm/src/opcode_handlers/exchange.rs
```diff
@@ -13,9 +13,6 @@ impl<'a> VM<'a> {
         let current_call_frame = &mut self.current_call_frame;
         current_call_frame.increase_consumed_gas(gas_cost::SWAPN)?;
 
-        if current_call_frame.stack.len() < N {
-            return Err(ExceptionalHalt::StackUnderflow.into());
-        }
```
MegaRedHand, line 18: This check is different from the one in `swap`, but actually wrong, since a stack length of 1 and an N of 1 would pass this check and fail inside `Stack::swap`. Writing this here so no one else goes down the rabbit hole 😞

crates/vm/levm/src/call_frame.rs
```diff
@@ -174,10 +174,7 @@ impl Stack {
 
     #[inline(always)]
     pub fn swap(&mut self, index: usize) -> Result<(), ExceptionalHalt> {
-        let index = self
-            .offset
-            .checked_add(index)
-            .ok_or(ExceptionalHalt::StackUnderflow)?;
+        let index = self.offset + index;
```
MegaRedHand: We should document that this may panic if the index is bigger than the stack limit.

edg-l (approved):

crates/vm/levm/src/call_frame.rs
iovoid: The addition can only panic if `offset+index > usize::MAX`. The swap is guaranteed to not panic by the `index >= self.values.len()` check.

iovoid: Addressed the panic in 15f5ae097, in case a future user decides to pass untrusted input.

crates/vm/levm/src/opcode_handlers/exchange.rs
azteca1998: I don't think this check was wrong, but I agree that it is redundant.

Stack length of 1 and swapping the 1st element (2nd actually since we're starting at zero) is an out of bounds access:

```
          len   <  N
Stack: [<elem 0>, nil, nil...]
            <------>
```

It would pass with N=0, which would swap an element with itself and does not make sense from a logical point of view, but it's not unsafe to execute.

crates/vm/levm/src/call_frame.rs
azteca1998: Yes, also it's complaining about potential side effects. Maybe consider using `wrapping_add`?

~~Alternatively, we could make `index` a generic constant and check it at compile-time with:~~
```
const { assert!(N < STACK_LIMIT); }
```
Edit: Nevermind you did just that.

PR #5307, opened 2025-11-12
feat(l1): code cache rocksdb

pablodeymo:
**Motivation**

Adds a code cache to rocksdb, evicting the least used entries if size goes over 64mb

### DIFF
--- a/crates/storage/store_db/rocksdb.rs
+++ b/crates/storage/store_db/rocksdb.rs
@@ -163,6 +163,10 @@ pub struct Store {
     flatkeyvalue_control_tx: std::sync::mpsc::SyncSender<FKVGeneratorControlMessage>,
     trie_update_worker_tx: TriedUpdateWorkerTx,
     last_computed_flatkeyvalue: Arc<Mutex<Vec<u8>>>,
+    /// Cache for account bytecodes, keyed by the bytecode hash.
+    /// Note that we don't remove entries on account code changes, since
+    /// those changes already affect the code hash stored in the account, and only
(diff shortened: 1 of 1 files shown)


jrchatruc (changes requested):
Blocking merge until we make sure we are properly managing any cache invalidation issues (contracts being destroyed, etc)

crates/storage/store_db/rocksdb.rs
```diff
@@ -149,13 +152,36 @@ enum FKVGeneratorControlMessage {
...
+type CodeCache = LruCache<H256, Code, FxBuildHasher>;
+
+#[derive(Debug)]
 pub struct Store {
     db: Arc<DBWithThreadMode<MultiThreaded>>,
     trie_cache: Arc<Mutex<Arc<TrieLayerCache>>>,
     flatkeyvalue_control_tx: std::sync::mpsc::SyncSender<FKVGeneratorControlMessage>,
     trie_update_worker_tx: TriedUpdateWorkerTx,
     last_computed_flatkeyvalue: Arc<Mutex<Vec<u8>>>,
+    account_code_cache: Arc<Mutex<CodeCache>>,
```
MegaRedHand, line 170:
```suggestion
    /// Cache for account bytecodes, keyed by the bytecode hash.
    /// Note that we don't remove entries on account code changes, since
    /// those changes already affect the code hash stored in the account, and only
    /// may result in this cache having useless data.
    account_code_cache: Arc<Mutex<CodeCache>>,
```

PR #5309, opened 2025-11-12
perf(l1): remove memtable prefix bloom option

MegaRedHand:
**Motivation**

**Description**

Related to #5302. We noticed a performance pessimization when re-adding this option, so here we try to remove it.


MegaRedHand:
This didn't result in noticeable improvements

PR #5311, opened 2025-11-12
feat(l1): rpc instrumentation and panels

rodrigo-o:
**Motivation**

Add rpc and engine instrumentation and panels to our dashboards

**Description**

This PR adds a couple of changes to profiling.rs and our dashboard:

- Add a simple way to time async funcrions without relying on `#instrument` (which would be more complex for every rpc)
- Add namespace as a field to the instrumentation spans
- Set a namespace for the old ones (`block_execution`)
- Make the breakdown panels work with the new changes
- Add new RPC and Engine panels

image


metrics/provisioning/grafana/dashboards/common_dashboards/ethrex_l1_perf.json
```diff
@@ -1251,14 +1239,14 @@
...
+              "expr": "(\n  sum by (function_name) (\n    increase(function_duration_seconds_sum{job=\"$job\", instance=~\"$instance(:\\\\d+)?$\", namespace!~\"rpc|engine\", function_name!~\"Block execution|Execute Block|Trie update\"}[$__range])\n  )\n)\nor\nlabel_replace(\n  sum( increase(function_duration_seconds_sum{job=\"$job\", instance=~\"$instance(:\\\\d+)?$\", namespace!~\"rpc|engine\", function_name=\"Block execution\"}[$__range]) )\n  - sum( increase(function_duration_seconds_sum{job=\"$job\", instance=~\"$instance(:\\\\d+)?$\", namespace!~\"rpc|engine\", function_name=~\"Storage read|Account read|Account code read|Block hash read\"}[$__range]) ),\n  \"function_name\",\"Block Execution (w/o reads)\",\"__name__\",\".*\"\n)\nor\nlabel_replace(\n  sum( increase(function_duration_seconds_sum{job=\"$job\", instance=~\"$instance(:\\\\d+)?$\", namespace!~\"rpc|engine\", function_name=\"Execute Block\"}[$__range]) )\n  - sum( increase(function_duration_seconds_sum{job=\"$job\", instance=~\"$instance(:\\\\d+)?$\", namespace!~\"rpc|engine\", function_name=\"Block execution\"}[$__range]) ),\n  \"function_name\",\"Merkleization Diff\",\"__name__\",\".*\"\n)",
```
rodrigo-o, line 1249: Instead of matching with the newly added `block_execution` namespace we are matching against the 2 additional ones (RPC and engine) this is to support the current running nodes when this changes are deployed to our centralized grafana instance. It avoids breaking the current Execution Breakdown for nodes that doesn'ta have this change and here is an issue to made this change in the following days: #5312

crates/networking/rpc/rpc.rs
```diff
@@ -191,7 +192,15 @@ pub trait RpcHandler: Sized {
...
         let request = Self::parse(&req.params)?;
-        request.handle(context).await
+        let namespace = match req.namespace() {
+            Ok(RpcNamespace::Engine) => "engine",
+            _ => "rpc",
+        };
+
+        record_async_duration(namespace, req.method.as_str(), async move {
+            request.handle(context).await
+        })
```
rodrigo-o, line 202: We might want to have this behind the metrics feature flag maybe? I'll take a look at how the rest of the instrumentation was working with the previous breakdown panels and in any case create an issue to follow-up on this.

crates/blockchain/metrics/profiling.rs
```diff
@@ -25,18 +29,67 @@ pub struct FunctionProfilingLayer;
 /// Wrapper around [`HistogramTimer`] to avoid conflicts with other layers
 struct ProfileTimer(HistogramTimer);
 
+/// Span extension storing the profiling namespace selected by instrumentation.
+struct Namespace(Cow<'static, str>);
+
+#[derive(Default)]
+struct NamespaceVisitor {
+    namespace: Option<Cow<'static, str>>,
```
MegaRedHand: Can we use a `&'static str` here instead of a `Cow`?

crates/blockchain/metrics/profiling.rs
```diff
@@ -54,6 +107,20 @@ where
     }
 }
 
+/// Record the duration of an async operation under the shared function profiling histogram.
+pub async fn record_async_duration<Fut, T>(namespace: &str, function_name: &str, future: Fut) -> T
```
MegaRedHand, line 126: Did you try instrumenting the future and using the profiling layer instead of this function?

crates/blockchain/metrics/profiling.rs
rodrigo-o: I guess for our current usage we could do that, but as far as I understand, the span metrics are able to receive % and ?. Thats why record_debug is in place. If we remove record_debug what will happen is that maybe others in the future might end up in the "default" namespace if they send them (%/?) or even use runtime strings. Do you think it would be better to move to `&'static str` anyway? Also we can go with just String if we want to avoid Cow but in this case it appears useful (for the record_str function)

crates/blockchain/metrics/profiling.rs
rodrigo-o: Yes I did, it is possible to do so, the issue is that as of now, the function name in the profiling layer is taken from the name function, which is always a literal string, but in this case we are using req.method as our function_name, and cant be assigned to the name in the span. There is a way to workaround that, something like:

```rust 
        let span = tracing::info_span!(
            target: "ethrex::rpc",
            "rpc_request",
            namespace = namespace,
            function_name = %req.method
        );

        request.handle(context).instrument(span).await
    }
```

But that requires us another extensions_mut at on_new_span and match both field in the visitor. We could do that anyway, I went with this because it appeared cleaner but the other might be more consistent with how we were managing the profiling layer, i guess both of them have their tradoffs.

crates/blockchain/metrics/profiling.rs
```diff
@@ -25,18 +29,67 @@ pub struct FunctionProfilingLayer;
...
+                .extensions()
+                .get::<Namespace>()
+                .map(|ns| ns.0.clone())
+                .unwrap_or_else(|| Cow::Borrowed("default"));
+
+            let function_name = span.metadata().name();
 
             let timer = METRICS_BLOCK_PROCESSING_PROFILE
-                .with_label_values(&[name])
+                .with_label_values(&[namespace.as_ref(), function_name])
```
MegaRedHand: We can avoid using `Cow` here by doing this:

```suggestion
            let extensions = span.extensions();
            let namespace = extensions
                .get::<Namespace>()
                .map(|ns| ns.0.as_ref())
                .unwrap_or_else(|| "default");

            let function_name = span.metadata().name();

            let timer = METRICS_BLOCK_PROCESSING_PROFILE
                .with_label_values(&[namespace, function_name])
```

Then we can switch to `String` in `Namespace`

rodrigo-o: Good catch, i will revert to String instead.

crates/blockchain/metrics/profiling.rs
MegaRedHand: Opened #5327 to review this and see if the profiling metrics gathering can be simplified.

crates/blockchain/metrics/profiling.rs
rodrigo-o: Done: [remove Cow given that it was already cloning](https://github.com/lambdaclass/ethrex/pull/5311/commits/39bd789e9f677f8c00dd956f13a55c3853aa9159)

PR #5318, opened 2025-11-13
refactor(l1): avoid recursion in `TrieLayerCache::commit`

MegaRedHand:
**Motivation**

We already had stack overflow errors two times, and we suspect it's caused by `TrieLayerCache::commit`, which uses recursion.

**Description**

This PR changes the recursion for a loop, which should also have performance benefits when committing multiple layers. It also will be more resilient to stack overflows.


jrchatruc:
Tested ✅

PR #5321, opened 2025-11-13
fix(l2): checkpoint creation

avilagaston9:
**Motivation**

This PR addresses two main issues:

- Sealing a batch, storing the `prover inputs`, and creating the checkpoint for that batch are currently performed non-atomically. If the node restarts during any of these three operations, the committer will end up in an invalid state.
- The batch checkpoint is created at the beginning of the building process. If an error occurs while building the batch, the `l1_committer` gets stuck with the following error:

```
2025-11-12T15:47:26.129936Z ERROR L1 Committer Error: Committer failed retrieve block from storage: Failed to create RocksDB checkpoint at "dev_ethrex_l2/checkpoint_batch_1": Invalid argument: Directory exists
```

**Description**

- Stores the batch and `prover inputs` atomically in the rollup storage.
- When the `l1_committer` encounters a batch generated in a previous iteration, it now checks whether the corresponding checkpoint exists. If it doesn't, the committer creates it by re-executing the batch.

Closes None

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -363,10 +363,7 @@ impl L1Committer {
         let batch_checkpoint_name = batch_checkpoint_name(batch.number);
         let expected_checkpoint_path = self.checkpoints_dir.join(&batch_checkpoint_name);
 
-        let current_checkpoint_path = self
-            .store
-            .get_store_directory()?
-            .join(batch_checkpoint_name);
+        let current_checkpoint_path = self.current_checkpoint_store.get_store_directory()?;
 
         if current_checkpoint_path == expected_checkpoint_path {
             info!(
@@ -381,7 +378,8 @@ impl L1Committer {
                 "Checkpoint for batch {} not found locally, generating it by re-executing the blocks in the batch",
                 batch.number
             );
-            self.generate_checkpoint_for_batch(batch).await?;
+            self.current_checkpoint_store = self.generate_checkpoint_for_batch(batch).await?;
+            return Ok(());
         }
 
         info!(
@@ -405,7 +403,10 @@ impl L1Committer {
     }
 
     /// Generate the checkpoint for the given batch by re-executing the blocks in the batch
-    async fn generate_checkpoint_for_batch(&mut self, batch: &Batch) -> Result<(), CommitterError> {
+    async fn generate_checkpoint_for_batch(
+        &mut self,
+        batch: &Batch,
+    ) -> Result<Store, CommitterError> {
         let (one_time_checkpoint_path, one_time_checkpoint_store, one_time_checkpoint_blockchain) =
             self.generate_one_time_checkpoint(batch.number).await?;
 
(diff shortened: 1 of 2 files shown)


crates/l2/sequencer/l1_committer.rs
```diff
@@ -125,10 +125,6 @@ pub struct L1Committer {
     /// It is used to ensure state availability for batch preparation and
     /// witness generation.
     current_checkpoint_store: Store,
-    /// Blockchain instance using the current checkpoint store.
-    ///
-    /// It is used for witness generation.
-    current_checkpoint_blockchain: Arc<Blockchain>,
```
avilagaston9, line 131: This was unused

crates/l2/sequencer/l1_committer.rs
```diff
@@ -337,37 +498,31 @@ impl L1Committer {
             )))?;
         let first_block_to_commit = last_block + 1;
 
-        // We need to guarantee that the checkpoint path is new
-        // to avoid causing a lock error under rocksdb feature.
-        let new_checkpoint_path = self
-            .checkpoints_dir
-            .join(batch_checkpoint_name(batch_number));
```
avilagaston9, line 344: Creating the batch checkpoint at the beginning of the execution prevents us from recovering from errors, as we can't create a checkpoint on an already used path.

crates/l2/sequencer/l1_committer.rs
```diff
@@ -321,6 +321,189 @@ impl L1Committer {
...
+
+        self.execute_batch_to_generate_checkpoint(
+            batch,
+            one_time_checkpoint_store.clone(),
+            one_time_checkpoint_blockchain,
+        )
+        .await
+        .inspect_err(|_| {
+            let _ = self.remove_one_time_checkpoint(&one_time_checkpoint_path);
+        })?;
```
tomip01, line 421: we could log the error here

crates/l2/sequencer/l1_committer.rs
```diff
@@ -321,6 +321,189 @@ impl L1Committer {
...
+        let new_checkpoint_path = self
+            .checkpoints_dir
+            .join(batch_checkpoint_name(batch.number));
+        let (new_checkpoint, _) = self
+            .create_checkpoint(
+                &one_time_checkpoint_store,
+                &new_checkpoint_path,
+                &self.rollup_store,
+            )
+            .await?;
```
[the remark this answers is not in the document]
tomip01: Agree with this, but I think is better to do it once since two lines below we do it any way

crates/l2/sequencer/l1_committer.rs
```diff
@@ -394,23 +570,30 @@ impl L1Committer {
...
+        let new_checkpoint_path = self
+            .checkpoints_dir
+            .join(batch_checkpoint_name(batch_number));
+        let (new_checkpoint_store, _) = self
+            .create_checkpoint(
+                &one_time_checkpoint_store,
+                &new_checkpoint_path,
+                &self.rollup_store,
+            )
+            .await?;
```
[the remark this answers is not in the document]
tomip01: Agree with doing something like this

PR #5322, opened 2025-11-13
feat(l1): periodically check if there is a newer release available

fmoletta:
**Motivation**
Periodically checking and informing the user if there is a newer version of ethrex available

**Description**
* Periodically query the latest relese of ethrex on GitHub and check against current running version, inform the user via log if there is a newer version available.

### DIFF
--- a/.github/workflows/pr-main_l2_prover.yaml
+++ b/.github/workflows/pr-main_l2_prover.yaml
@@ -19,7 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        backend: ["sp1", "risc0", "zisk", "openvm"]
+        # Note: openvm backend linting is disabled due to a problem in the installation https://github.com/lambdaclass/ethrex/issues/5509
+        backend: ["sp1", "risc0", "zisk"]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4


cmd/ethrex/ethrex.rs
```diff
@@ -45,22 +47,74 @@ async fn server_shutdown(
...
+        );
+    }
+}
+
+/// Checks if there is a newer ethrex verison available every hour
+async fn periodically_check_version_update() {
+    let mut interval = tokio::time::interval(Duration::from_secs(60 * 60));
+    loop {
+        interval.tick().await;
+        check_version_update().await;
```
MegaRedHand, line 121: I think this should check at the start, before waiting for a tick:

```suggestion
        check_version_update().await;
        interval.tick().await;
```

That way the user can download the newer version as soon as possible.

fmoletta: You would expect it to work like that, but when I tested it locally putting `check_version_update` first caused it to log twice in a row. I had to switch it to this for it to log once at the start and then every hour

cmd/ethrex/Cargo.toml
```diff
@@ -52,6 +51,7 @@ reqwest.workspace = true
 thiserror.workspace = true
 itertools = "0.14.0"
 url.workspace = true
+semver = "1.0.27"
```
Oppen: We're not following semver right now, so this is probably not the right tool to use.
I would just use a simple inequality check against `latest`.

fmoletta: But wouldn't that be bothersome if the user is running a newer version instead of the latest? Like a release candidate on in-development version?

Oppen: In that case, I would still not use a crate that assumes semver in a package that explicitly opts out of semver. A simple function that splits and parses integers and checks whether any is less than the latest version's one should suffice.

fmoletta: Okay, I wanted to avoid implementing a manual comparison as we already have semver as a dependency on most of our lock files that provides it, but we can add it

fmoletta: Added manual version comparison: [f637f4e](https://github.com/lambdaclass/ethrex/pull/5322/commits/f637f4e8f14f6a3b389a8da936ce75d8d4ecae3a)

cmd/ethrex/ethrex.rs
```diff
@@ -45,6 +46,80 @@ async fn server_shutdown(
...
+/// Fetches the latest release version on github
+/// Returns None if there was an error when requesting the latest version
+async fn latest_release_version() -> Option<String> {
+    #[derive(Deserialize)]
+    struct Release {
+        tag_name: String,
+    }
+    let client = reqwest::Client::new();
+    let response = client
+        .get("https://api.github.com/repos/lambdaclass/ethrex/releases/latest")
```
pablodeymo: I would put the URL in a constant in a common place

PR #5323, opened 2025-11-13
feat(l1)!: compress some tables

iovoid:
**Motivation**

We disabled compression for all tables, but some tables with less but larger elements could benefit and bring small space savings.

**Description**

This PR enables compression for codes and other, smaller, compressible tables.


Cargo.toml
```diff
@@ -121,6 +121,7 @@ tower-http = { version = "0.6.2", features = ["cors"] }
 indexmap = { version = "2.11.4" }
 k256 = "0.13.4"
 
+
```
pablodeymo:
```suggestion
```

MegaRedHand (approved):

PR #5324, opened 2025-11-13
perf(levm): specialized DUP

iovoid:
**Motivation**

Currently, DUP takes ~3% of `VM:: run_execution`.

**Description**

This brings together the get and push methods.


crates/vm/levm/src/call_frame.rs
```diff
@@ -164,6 +152,36 @@ impl Stack {
...
+            .offset
+            .checked_sub(1)
+            .ok_or(ExceptionalHalt::StackOverflow)?;
+
+        #[expect(unsafe_code, reason = "index < size, offset-1 >= 0")]
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                self.values.get_unchecked_mut(index).0.as_mut_ptr(),
+                self.values.get_unchecked_mut(self.offset).0.as_mut_ptr(),
+                4,
```
MegaRedHand: Can we use a constant for this? Otherwise, let's put it in a local variable or add a comment.

MegaRedHand: Also, I think we're missing a compile time check for `N=0`, which makes `dst` and `src` overlap.

iovoid: With N=0, `index` is `original_offset` and `self.offset` is `original_offset - 1`.

iovoid: Replaced the 4s with constants in fb564f151

MegaRedHand: Right. I mixed it up with `op_swap`, which does start at 1.

crates/vm/levm/src/call_frame.rs
```diff
@@ -11,6 +11,9 @@ use ethrex_common::{Address, U256};
 use ethrex_common::{H256, types::Code};
 use std::{collections::HashMap, fmt};
 
+// u64's that make up an U256
+const U64_PER_U256: usize = 256 / 64;
```
MegaRedHand: I'm nitpicking, but I think this is safer:

```suggestion
/// [`u64`]s that make up a [`U256`]
const U64_PER_U256: usize = U256::MAX.0.len();
```

PR #5328, opened 2025-11-13
perf(levm): add base blob fee to env to avoid recalculating it

iovoid:
**Motivation**

Currently in prepare_execution we are, for each transaction, calculating the blob base fee which is consistent across the block.

**Description**

This was already cached for `BLOBBASEFEE` in #5288, but it could be unified across all users by making part of the Environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Perf
 
-### 2025-12-13
+### 2025-11-13
 
 - Avoid recalculating blob base fee while preparing transactions [#5328](https://github.com/lambdaclass/ethrex/pull/5328)
 - Use BlobDB for account_codes column family [#5300](https://github.com/lambdaclass/ethrex/pull/5300)


CHANGELOG.md
```diff
@@ -2,8 +2,9 @@
 
 ## Perf
 
-### 2025-11-13
+### 2025-12-13
```
MegaRedHand:
```suggestion
### 2025-11-13
```

crates/vm/levm/src/environment.rs
```diff
@@ -28,6 +28,7 @@ pub struct Environment {
     pub difficulty: U256,
     pub chain_id: U256,
     pub base_fee_per_gas: U256,
+    pub base_blob_fee_per_gas: U256,
```
MegaRedHand, line 31: Are we taking into account earlier forks that didn't have blobs?

iovoid: Yes, when `block_excess_blob_gas` is empty it will default to zero, and get_base_fee_per_blob_gas will only error if `base_fee_update_fraction` is 0 which is never true since `get_blob_base_fee_update_fraction_value` will return `BLOB_BASE_FEE_UPDATE_FRACTION` (equal to 3338477).

MegaRedHand (approved):

PR #5330, opened 2025-11-13
feat(l1): check DB version compatibility upon startup

fmoletta:
**Motivation**
Add a way to track breaking changes to the DB between node restarts and require resync between incompatible versions

**Description**
* Add `STORE_SCHEMA_VERSION` numeric constant to be updated upon breaking DB changes (changes requiring a re-sync)
* Save the current store schema version when starting up the DB
* Check that the current store schema version matches the stored one when starting up the node
* Add Checklist so we remember to update `STORE_SCHEMA_VERSION` upon breaking changes


mpaulucci:
why do we need this?

cmd/ethrex/initializers.rs
```diff
@@ -395,8 +394,9 @@ pub async fn init_l1(
     ethrex_crypto::kzg::warm_up_trusted_setup();
 
     let store = init_store(datadir, genesis).await;
-    if opts.syncmode == SyncMode::Full {
-        store.generate_flatkeyvalue()?;
+
+    if let Some(latest_snap_client_version) = store.get_snap_client_version().await? {
+        info!("Latest snap sync performed at client version {latest_snap_client_version}");
```
mpaulucci: why is this useful?

Oppen: I think for troubleshooting it might be needed, as sometimes there are breaking changes in the DB that require a resync, but we currently have no obvious way to check whether resync was performed when asked for help.

crates/networking/p2p/sync.rs
```diff
@@ -288,6 +318,9 @@ impl Syncer {
 
         store.clear_snap_state().await?;
         self.snap_enabled.store(false, Ordering::Relaxed);
+        store
+            .set_snap_client_version(env!("CARGO_PKG_VERSION"))
```
iovoid: Maybe use`get_client_version` which also includes the git commit?

cmd/ethrex/initializers.rs
mpaulucci: if it's for troubleshooting then should it be `debug`?

crates/networking/p2p/sync.rs
fmoletta: While I agree with this, `get_client_version`, depends on the build script to obtain the rust version, git commit, etc. Using `get_client_version` would mean moving this build script to the common crate, which is not desirable, or importing it from cmd/ethrex crate which is also not desirable. I don't think this is as useful enough to warrant such changes

cmd/ethrex/initializers.rs
Oppen: I think info is more appropriate, specially with it being a single line at startup. It's not for diagnosing bugs but for guiding end users.

pablodeymo: I agree with @Oppen that this log should be `info` instead of `debug`. We need information easily.

.github/pull_request_template.md
```diff
@@ -8,5 +8,9 @@
 
 <!-- Link to issues: Resolves #111, Resolves #222 -->
 
+**Checklist**
+
+- [ ] Updated `STORE_SCHEMA_VERSION` if the PR includes breaking changed to the `Store` requiring a re-sync.
```
MegaRedHand:
```suggestion
- [ ] Updated `STORE_SCHEMA_VERSION` if the PR includes breaking changes to the `Store`, requiring a re-sync.
```

cmd/ethrex/initializers.rs
```diff
@@ -405,6 +405,15 @@ pub async fn init_l1(
         store.generate_flatkeyvalue()?;
     }
 
+    if let Some(latest_store_schema_version) = store.get_store_schema_version().await?
+        && latest_store_schema_version != STORE_SCHEMA_VERSION
```
MegaRedHand: This should fail if the file is not found

fmoletta: Not really, if there is no schema version then it means we are starting from a clean DB

.github/pull_request_template.md
fmoletta: nice catch!

cmd/ethrex/initializers.rs
MegaRedHand: Right. But if we are using an old ethrex without DB versioning, this would pass, right?

fmoletta: I think it is out of scope to check for compatibility prior to the introduction of a compatibility check system. It would just make code more complex as we would have to check if there is "something" in the DB to differentiate between empty DB and unversioned DB.

MegaRedHand: I propose moving the version check to `Store::new`, or even further down into the RocksDB implementation. In this line, we have already interacted with the store through `init_store` -> `Store::add_initial_state`, which can fail if we change the DB format in some way.

fmoletta: I agree, I didn't do it at first as it would mean changing a lot of signatures to return results. I don't know if I agree with linking it to RocksDB though, as that would mean managing a schema version for each individual DB backend

fmoletta: Update: Moved check to `Store::new`

cmd/ethrex/cli.rs
```diff
@@ -794,7 +794,13 @@ pub async fn export_blocks(
     last_number: Option<u64>,
 ) {
     init_datadir(datadir);
-    let store = load_store(datadir).await;
+    let store = match load_store(datadir).await {
+        Err(err) => {
+            warn!("Failed to load Store due to: {err}");
```
Oppen: Shouldn't this be `fatal!` or at least `error!`?

pablodeymo: agree

crates/storage/store.rs
```diff
@@ -1263,6 +1270,16 @@ impl Store {
...
 
+    /// Save the current store schema version
+    pub fn set_store_schema_version(&self) -> Result<(), StoreError> {
+        self.engine.set_store_schema_version()
+    }
+
+    /// Obtain the latest used store schema version
+    pub fn get_store_schema_version(&self) -> Result<Option<u64>, StoreError> {
+        self.engine.get_store_schema_version()
+    }
```
Oppen, line 1282: I think this should default to 0 rather than return an option.

cmd/ethrex/cli.rs
```diff
@@ -794,7 +794,13 @@ pub async fn export_blocks(
...
+            return;
+        }
+        Ok(store) => store,
+    };
```
pablodeymo, line 803:
```suggestion
    let Ok(store) = load_store(datadir).await else {
        error!("Failed to load Store");
        return;
    };
```

crates/storage/store_db/in_memory.rs
```diff
@@ -729,6 +729,16 @@ impl StoreEngine for Store {
...
     }
 
+    fn set_store_schema_version(&self) -> Result<(), StoreError> {
+        // This is not relevant for InMemory store as it won't be saved once execution ends
+        Ok(())
+    }
+
+    fn get_store_schema_version(&self) -> Result<Option<u64>, StoreError> {
+        // This is not relevant for InMemory store as it won't be saved once execution ends
+        Ok(None)
```
Oppen, line 739: In line with my other comment, this should return `STORE_SCHEMA_VERSION` (given it's in-memory, it will always be the current one).

crates/storage/store.rs
```diff
@@ -97,6 +97,13 @@ impl Store {
             },
         };
 
+        if let Some(latest_store_schema_version) = store.get_store_schema_version()?
+            && latest_store_schema_version != STORE_SCHEMA_VERSION
+        {
+            return Err(StoreError::IncompatibleDBVersion);
+        }
+        store.set_store_schema_version()?;
```
Oppen, line 106: To match my previous comments, setting this value should be done in initialization of the engine (where we know whether the DB exists or not), and we should only get the version here for the check.

Oppen (approved):
Left a few non-blocking comments. Feel free to address them, but I'm happy with this improvement as is.

crates/storage/store.rs
```diff
@@ -97,6 +97,13 @@ impl Store {
             },
         };
 
+        if let Some(latest_store_schema_version) = store.get_store_schema_version()?
```
pablodeymo, line 101:
```suggestion
        // Check that the DB version is the expected
        if let Some(latest_store_schema_version) = store.get_store_schema_version()?
```

cmd/ethrex/cli.rs
fmoletta: The whole error handling for this subcommand consists of `warn!` followed by `return` so I kept it consistent with the rest of the code

.github/pull_request_template.md
```diff
@@ -8,5 +8,9 @@
 
 <!-- Link to issues: Resolves #111, Resolves #222 -->
 
+**Checklist**
+
+- [ ] Updated `STORE_SCHEMA_VERSION` if the PR includes breaking changes to the `Store` requiring a re-sync.
```
pablodeymo: Please, include the location of the constant (`crates/storage/lib.rs`), to make developer life easier.

crates/storage/store.rs
fmoletta: If we are starting up the node from a clean DB (a fresh install or a run after removedb is used) then we would have no schema version, if we default to 0 then we would always fail due to incompatible version as of version 1

crates/storage/lib.rs
```diff
@@ -14,3 +14,7 @@ pub use store::{
     hash_address, hash_key,
 };
 pub use trie_db::layering::apply_prefix;
+
+/// Store Schema Version, must be updated on any breaking change
+/// An upgrade to a newer schema version invalidates currently stored data, requiring a re-sync.
+pub const STORE_SCHEMA_VERSION: u64 = 0;
```
pablodeymo: this constant should be set in 1, initially.

cmd/ethrex/cli.rs
fmoletta: Shouldn't we show the user why loading the store failed?

crates/storage/store.rs
fmoletta: If we move this behaviour to be backend-specific then shouldn't we use a versioning system for each backend instead of a global store schema version?

cmd/ethrex/cli.rs
fmoletta: Moved to `error!`

pablodeymo:
This PR should include the documentation for users in the book, stating the definition of the constant, how to update the value of the schema, etc.

crates/storage/store.rs
MegaRedHand: I propose using a JSON file instead of the DB for this. That would make it easier to inspect and make this backend-independent.

cmd/ethrex/cli.rs
pablodeymo: ok

pablodeymo (approved):
👍

cmd/ethrex/initializers.rs
```diff
@@ -96,35 +96,29 @@ pub fn init_metrics(opts: &Options, tracker: TaskTracker) {
...
-    let mut store = open_store(datadir.as_ref());
-    store
-        .add_initial_state(genesis)
-        .await
-        .expect("Failed to create genesis block");
-    store
+pub async fn init_store(datadir: impl AsRef<Path>, genesis: Genesis) -> Result<Store, StoreError> {
+    let mut store = open_store(datadir.as_ref())?;
+    store.add_initial_state(genesis).await?;
+    Ok(store)
```
ilitteri, line 102: Nit. (only if `add_initial_state` returns `StoreError`).
```suggestion
    store.add_initial_state(genesis).await
```

cmd/ethrex/initializers.rs
```diff
@@ -96,35 +96,29 @@ pub fn init_metrics(opts: &Options, tracker: TaskTracker) {
...
-    let store = open_store(datadir);
-    store
-        .load_initial_state()
-        .await
-        .expect("Failed to load store");
-    store
+pub async fn load_store(datadir: &Path) -> Result<Store, StoreError> {
+    let store = open_store(datadir)?;
+    store.load_initial_state().await?;
+    Ok(store)
```
ilitteri, line 109: Nit. (only if `load_initial_state` returns StoreError).
```suggestion
    store.load_initial_state().await
```

ilitteri (approved):

cmd/ethrex/initializers.rs
fmoletta: It doesn't return the Store :(

avilagaston9 (approved):

cmd/ethrex/initializers.rs
fmoletta: Same here

PR #5331, opened 2025-11-13
feat(l1): add `clear-data` target for tooling/sync/Makefile

fmoletta:
**Motivation**
When using the targets `start-%-metrics-docker` we often run into two annoying problems:
* The consensus client is stuck syncing from an old checkpoint as it was used previously and its data was not cleared
* The node has inconsistent data from a previous run
These two can be easily solved by removing the previous stale data on the clients (either removed manually or via purge-db/ removedb options). This PR proposes adding some makefile targets to make this easier (aka using the same preset paths according to the network that the other targets on the Makefile use)

**Description**
* Add targets `clear-lighthouse`, `clear-ethrex` and `clear-data` to remove stale data from previous runs of other targets on the Makefile


mpaulucci (approved):

ElFantasma (approved):

ElFantasma:
DATA_PATH does not longer exist. It's open since November and nobody needed this. Closing it.

Issue #5327, opened 2025-11-13
Simplify `FunctionProfilingLayer`


MegaRedHand:
Our current `FunctionProfilingLayer` implementation got quite complex, and has some problems: mainly that it requires a `String` allocation for the namespace (added in #5311), and that implementing functionality via `on_enter`/`on_exit` causes it to measure processing time instead of latency.

We should check if it's not easier to use a function wrapper (like `record_async_duration`), or if these issues can be fixed, and in either case, unify profiling metrics of RPC and block processing.

rodrigo-o:
I'm doing some test in separating the RPC handling (now accounting also for the result count) from the profiling.rs which was in the beggining just thought to work on block_execution timing. If this is clear and more readable i'll push create a new PR with it, which will left the old profiling as is and easier to refactor in the future if needed.

PR #5335, opened 2025-11-13
feat(l1): add rpc error rates to metrics and panels

### LINKED ISSUE #5379: Deagregate RPC/Engine error rate to actual methods
After #5335 we'll have just a global error rate, we want to be able to understand wich are the most comoon methods failing both for the engine as well as RPC. We might need to separate all of this panels in a new row to avoid overcomplicating the current RPC and Engine ones (given that we rely on vertical duplicate by instance and we are just expanding horizontally on each of them)

rodrigo-o:
**Motivation**

Add success/error rate panels for engine and rpc api

**Description**

This PR:
- Extracted the RPC metric logic in its own module to avoid overload the previous block_processing only profiling module
- Add the new error/success rates metrics and instrumentation along side the previous rpc instrumentation
- Move shared logic for registering default metrics outside of profiling
- Adds a whole new row for tracking both RPC and Engine errorr rates and deagregation by method and kind of error
- Additional Engine pie chart to show proportion of calls by method
- Dahsboard docs have been updated, see the changes [here](https://github.com/lambdaclass/ethrex/blob/rpc_error_rates/docs/developers/l1/dashboards.md#engine-api) and [here](https://github.com/lambdaclass/ethrex/blob/rpc_error_rates/docs/developers/l1/dashboards.md#engine-and-rpc-error-rates)

image

**Next Steps**

We still have some improvements that could be made to the block processing profiling:
- [ ] As tracked by #5327 we want to do some work apart from this refactor
- [x] A new issue has been created to rename the rest of the modules in metrics and remove the additional `metrics_` prefix #5378.
- [x] Apart from this we may want to have more information related to errors to be able to deagregate them as tracked by #5379

**NOTE**
Once this is merged and published in our shared grafana we'll need to update the servers with main for being able to see the rpc/engine latency panels given that the metric names changed in this PR due to the extraction from the previous block profiling module.

Closes #5379

### DIFF
--- a/crates/blockchain/metrics/rpc.rs
+++ b/crates/blockchain/metrics/rpc.rs
@@ -32,7 +32,7 @@ fn initialize_rpc_duration_histogram() -> HistogramVec {
 #[derive(Clone)]
 pub enum RpcOutcome {
     Success,
-    Error(String),
+    Error(&'static str),
 }
 
 impl RpcOutcome {
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -218,7 +218,7 @@ pub trait RpcHandler: Sized {
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr>;
 }
 
-fn get_error_kind(err: &RpcErr) -> String {
+fn get_error_kind(err: &RpcErr) -> &'static str {
     match err {
         RpcErr::MethodNotFound(_) => "MethodNotFound",
         RpcErr::WrongParam(_) => "WrongParam",
@@ -236,7 +236,6 @@ fn get_error_kind(err: &RpcErr) -> String {
         RpcErr::InvalidPayloadAttributes(_) => "InvalidPayloadAttributes",
         RpcErr::UnknownPayload(_) => "UnknownPayload",
     }
-    .to_string()
 }
 
 pub const FILTER_DURATION: Duration = {


crates/blockchain/metrics/rpc.rs
```diff
@@ -0,0 +1,81 @@
...
+        .with_label_values(&[namespace, function_name, outcome.as_label()])
+        .inc();
+}
+
+pub fn initialize_rpc_metrics() {
+    METRICS_RPC_REQUEST_OUTCOMES.reset();
+    METRICS_RPC_DURATION.reset();
+}
+
+/// Records the duration of an async operation in the function profiling histogram.
```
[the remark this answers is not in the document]
rodrigo-o: This was already taken care here: [simplify record_async_duration documentation](https://github.com/lambdaclass/ethrex/pull/5335/commits/b6092746b9b7165a7d62b7a42708e4d81f675311)

edg-l (approved):

crates/blockchain/metrics/rpc.rs
```diff
@@ -0,0 +1,85 @@
...
+        &["namespace", "method"],
+    )
+    .unwrap()
+}
+
+/// Represents the outcome of an RPC request when recording metrics.
+#[derive(Clone)]
+pub enum RpcOutcome {
+    Success,
+    Error(String),
```
MegaRedHand: This should enforce compile-time known strings: 
```suggestion
    Error(&'static str),
```

rodrigo-o: Great catch, i needed to change the error function return type also, done here: [use static str as error](https://github.com/lambdaclass/ethrex/pull/5335/commits/9611ee554e86406a11236e319605c433be7664e2)

PR #5340, opened 2025-11-13
chore: remove unused ThreadJoinError variant from TrieGenerationError

Fibonacci747:
Removed the ThreadJoinError variant from the TrieGenerationError enum in crates/common/trie/trie_sorted.rs. The current threading model uses a scoped thread pool that does not expose join handles or join results to the caller, and no code constructs or matches on this variant.


ElFantasma (approved):
**Staleness note:** This PR is from Nov 2025. The file has since gained new content but the `TrieGenerationError` enum is unchanged, so the diff should apply cleanly after a rebase.

ElFantasma:
We are in the process of cleaning the stalled and trivial PRs. 

I'm closing this one, but if you believe it is an error feel free to reopen with an explanation.

PR #5350, opened 2025-11-14
perf(l1): improve rlp encoding

edg-l:
**Motivation**

- In the current impl_encode there are 2 loops to get the position, fixed to a single pass.
- Added specialized branchless length methods to integers,
- Added specialized length to Vec<T>
- Added specialized length to tuples
- Added specialized length to most types in ethrex-common, only ones missing are in networking/p2p
- Removed unneeded arrayvec usage
- Many instances that deref to &[u8] didn't use its length specialized method, fixed to use it

Each missing length impl in the trait impl meant an extra allocation, this should remove them entirely when encoding

Helps with #4956

### DIFF
--- a/crates/common/rlp/encode.rs
+++ b/crates/common/rlp/encode.rs
@@ -699,6 +699,19 @@ mod tests {
     }
 
     #[test]
+    fn u64_lengths_match() {
+        for n in 0u64..=10_000 {
+            let mut encoded = Vec::new();
+            n.encode(&mut encoded);
+            assert_eq!(
+                encoded.len(),
+                n.length(),
+                "u64 length mismatch at value {n}"
+            );
+        }
+    }
+
+    #[test]
     fn can_encode_u8() {
         let mut encoded = Vec::new();
         0u8.encode(&mut encoded);


crates/common/rlp/encode.rs
```diff
@@ -13,6 +12,23 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
+    if payload_len < 56 {
+        // short prefix
+        1 + payload_len
+    } else {
+        let be = payload_len.to_be_bytes();
+        let mut i = 0;
+        while i < be.len() && be[i] == 0 {
+            i += 1;
+        }
+        let be_len = be.len() - i;
```
MegaRedHand: This can be rewritten as `let be_len = (payload_len.ilog2() / 8) + 1`, I think

crates/common/rlp/encode.rs
```diff
@@ -47,76 +63,120 @@ impl RLPEncode for bool {
...
 impl RLPEncode for u8 {
     fn encode(&self, buf: &mut dyn BufMut) {
         impl_encode(self.to_be_bytes(), buf);
     }
+
+    #[inline]
+    fn length(&self) -> usize {
+        1usize
+            + ((8 - self.leading_zeros() as usize).div_ceil(8)
+                * (((*self != 0) as usize) & ((*self > 0x7f) as usize)))
```
MegaRedHand: This is the same as the other integers and the while loops in `impl_encode` and `list_length`. We should move it to a `const` helper method.

crates/common/rlp/encode.rs
```diff
@@ -47,76 +89,120 @@ impl RLPEncode for bool {
...
 
 impl RLPEncode for u64 {
     fn encode(&self, buf: &mut dyn BufMut) {
         impl_encode(self.to_be_bytes(), buf);
     }
+
+    #[inline]
+    fn length(&self) -> usize {
+        1usize
+            + ((64 - self.leading_zeros() as usize).div_ceil(8)
```
MegaRedHand: We should use constants here:

```suggestion
            + ((u64::BITS - self.leading_zeros() as usize).div_ceil(8)
```

crates/common/rlp/encode.rs
```diff
@@ -47,76 +89,120 @@ impl RLPEncode for bool {
...
+                * (((*self != 0) as usize) & ((*self > 0x7f) as usize)))
```
MegaRedHand: This part looks confusing. Do we have tests for this?

crates/common/rlp/encode.rs
```diff
@@ -13,6 +12,49 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
+        1 + be_len + payload_len
+    }
+}
+
+/// Computes the length needed for a given bytes length and first byte
+#[inline]
+pub const fn bytes_length(bytes_len: usize, first_byte: u8) -> usize {
+    if bytes_len == 0 {
+        return 1;
+    }
```
MegaRedHand: This isn't needed, since we also receive the first byte, so we're assuming `bytes_len` isn't 0.

crates/common/types/block.rs
```diff
@@ -171,6 +190,58 @@ impl RLPEncode for BlockHeader {
...
+        payload_len += self.gas_limit.length();
+        payload_len += self.gas_used.length();
+        payload_len += self.timestamp.length();
+        payload_len += self.extra_data.length();
+        payload_len += self.prev_randao.length();
+        payload_len += self.nonce.to_be_bytes().length();
+
+        payload_len += match &self.base_fee_per_gas {
+            Some(v) => v.length(),
+            None => 1,
```
MegaRedHand: Why does this have a length of 1? As I understand, when `None`, these should be skipped when serializing.

crates/common/rlp/encode.rs
edg-l: added tests comparing the calculated length with the encoded length on the integer rlp tests

crates/common/types/block.rs
edg-l: you are right, fixed

crates/common/rlp/encode.rs
```diff
@@ -13,6 +12,45 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
+pub const fn list_length(payload_len: usize) -> usize {
+    if payload_len < 56 {
+        // short prefix
+        1 + payload_len
+    } else {
+        let be = payload_len.to_be_bytes();
+        let mut i = 0;
+        while i < be.len() && be[i] == 0 {
+            i += 1;
+        }
```
Arkenan: nit: You should explain these things in comments inside the code, they're not obvious. e.g. "the length is encoded as the shortest possible big-endian representation, so we don't consider the null bytes in the most significant bits".

crates/common/rlp/encode.rs
```diff
@@ -13,6 +12,45 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
+        let be_len = be.len() - i;
+        1 + be_len + payload_len
+    }
+}
+
+/// Computes the length needed for a given bytes length and first byte
```
Arkenan:
```suggestion
/// Computes the length needed for a given byte-string and first byte
```

crates/common/rlp/encode.rs
```diff
@@ -47,76 +85,120 @@ impl RLPEncode for bool {
...
 
 impl RLPEncode for u16 {
     fn encode(&self, buf: &mut dyn BufMut) {
         impl_encode(self.to_be_bytes(), buf);
     }
+    #[inline]
+    fn length(&self) -> usize {
+        1usize
+            + ((16 - self.leading_zeros() as usize).div_ceil(8)
+                * (((*self != 0) as usize) & ((*self > 0x7f) as usize)))
```
Arkenan: nit: we should evaluate making this a generic inline function in terms of the amount of bits and implement it only once, plus add a comment to explain how it works:

- Returns 1 if the integer is 0 or smaller than 0x7f
- Otherwise calculates the non-zero bits and uses that to calculate the minimum amount of bytes needed for the representation.

crates/common/rlp/encode.rs
```diff
@@ -47,76 +85,120 @@ impl RLPEncode for bool {
...
+    }
 }
 
 impl RLPEncode for () {
     fn encode(&self, buf: &mut dyn BufMut) {
         buf.put_u8(RLP_NULL);
     }
+    #[inline]
+    fn length(&self) -> usize {
+        0
```
Arkenan: If the encode puts `RLP_NULL` in the buffer, shouldn't the length return a 1?

crates/common/trie/nibbles.rs
```diff
@@ -207,6 +207,36 @@ impl Nibbles {
         compact
     }
 
+    /// Returns the length of this Nibbles as if you call encode_compact, without allocating.
+    #[inline]
+    pub const fn compact_length(&self) -> usize {
+        let d = self.data.len();
+        // subtract 1 if odd
+        let hex_len = d - (d & 1);
```
MegaRedHand: Won't division ignore the last bit anyways? I think this whole function can be simplified to:

```rust
(self.data.len() + 1) / 2
```

crates/common/trie/nibbles.rs
```diff
@@ -207,6 +207,36 @@ impl Nibbles {
...
+    /// Returns the first compacted byte, used to calculate the length without allocating.
+    #[inline]
+    pub fn compact_first_byte(&self) -> u8 {
+        let is_leaf = self.is_leaf();
+        let d = self.data.len();
+        let odd = d & 1 == 1;
+
+        let mut v = 0u8;
+
+        if odd {
```
MegaRedHand: For consistency:

```suggestion
        let is_odd = d & 1 == 1;

        let mut v = 0u8;

        if is_odd {
```

crates/common/types/transaction.rs
```diff
@@ -2891,6 +3122,12 @@ mod mempool {
                 .encode_field(&*self.inner)
                 .finish();
         }
+
+        #[inline]
+        fn length(&self) -> usize {
```
MegaRedHand: Are we using this encode implementation?

crates/common/types/tx_fields.rs
```diff
@@ -62,6 +62,24 @@ impl RLPEncode for AuthorizationTuple {
...
+            + self.s_signature.length();
+
+        list_length(payload_len)
+    }
+
+    fn encode_to_vec(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
+        self.encode(&mut buf);
+        buf
+    }
```
MegaRedHand: I think we can remove this

crates/common/rlp/encode.rs
```diff
@@ -13,6 +12,47 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
+        return 1 + bytes_len; // prefix (0x80 + len) + payload
+    }
+
+    // long (>=56 bytes)
+    let be = bytes_len.to_be_bytes();
+    let mut i = 0;
+    while i < be.len() && be[i] == 0 {
+        i += 1;
+    }
+    let be_len = be.len() - i;
```
MegaRedHand: This part is duplicated with `list_length`. Also, using `ilog2` might make this a bit more clear.

crates/common/rlp/encode.rs
```diff
@@ -47,76 +87,124 @@ impl RLPEncode for bool {
...
 impl RLPEncode for u128 {
     fn encode(&self, buf: &mut dyn BufMut) {
         impl_encode(self.to_be_bytes(), buf);
     }
+
+    #[inline]
+    fn length(&self) -> usize {
+        1usize
+            + ((128 - self.leading_zeros() as usize).div_ceil(8)
+                * (((*self != 0) as usize) & ((*self > 0x7f) as usize)))
```
MegaRedHand: We should move each part of this into variables and constants

crates/common/types/transaction.rs
```diff
@@ -44,6 +44,22 @@ pub enum Transaction {
...
+    #[inline]
+    fn tx_type_byte(&self) -> u8 {
+        match self {
+            Transaction::EIP2930Transaction(_) => 0x01,
+            Transaction::EIP1559Transaction(_) => 0x02,
+            Transaction::EIP4844Transaction(_) => 0x03,
+            Transaction::EIP7702Transaction(_) => 0x04,
+            Transaction::PrivilegedL2Transaction(_) => 0x05,
+            Transaction::FeeTokenTransaction(_) => 0x06,
+            Transaction::LegacyTransaction(_) => 0x0, // unused
```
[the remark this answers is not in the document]
mpaulucci: @edg-l is this so?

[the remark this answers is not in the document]
edg-l: yes, fixed

crates/common/rlp/encode.rs
```diff
@@ -47,76 +76,130 @@ impl RLPEncode for bool {
...
 impl RLPEncode for u32 {
     fn encode(&self, buf: &mut dyn BufMut) {
         impl_encode(self.to_be_bytes(), buf);
     }
+
+    #[inline]
+    fn length(&self) -> usize {
+        let bits = 32 - self.leading_zeros() as usize;
+        let needs_full_encoding = ((*self != 0) as usize) & ((*self > 0x7f) as usize);
+        let size = bits.div_ceil(8);
```
Oppen: Check the asm first, but if being branchless helps, avoiding a div probably helps too:
```rust
let size = (bits + 7) >> 3;
```

crates/common/rlp/encode.rs
```diff
@@ -141,47 +223,54 @@ impl RLPEncode for [u8] {
...
-            56..U8_MAX_PLUS_ONE => 1 + 1 + self.len(), // single byte prefix + payload len bytes
-            U8_MAX_PLUS_ONE..U16_MAX_PLUS_ONE => 1 + 2 + self.len(), // single byte prefix + payload len bytes
-            _ => {
-                // fallback if `self` is longer than 2^16 - 1 bytes
-                let payload_len_bytes =
-                    ((usize::BITS - self.len().leading_zeros()) as usize).div_ceil(8);
-                1 + payload_len_bytes + self.len()
-            }
+        if self.is_empty() {
+            return 1;
```
Oppen, line 253: I believe `bytes_length` covers this case already.

crates/common/rlp/encode.rs
```diff
@@ -190,6 +279,22 @@ impl RLPEncode for U256 {
...
+        let bits = 256 - self.leading_zeros() as usize;
+        let sig_len = bits.div_ceil(8);
+        let first_byte = if sig_len == 0 {
+            0
+        } else {
+            let shift = 8 * (sig_len - 1);
+            ((self >> shift) & U256::from(0xff)).as_u32() as u8
+        };
+
+        let is_nonzero_mask = (sig_len != 0) as usize;
```
Oppen: I'm not sure this mask is needed.

crates/common/rlp/encode.rs
```diff
@@ -141,47 +223,54 @@ impl RLPEncode for [u8] {
...
 }
 
 impl<const N: usize> RLPEncode for [u8; N] {
     fn encode(&self, buf: &mut dyn BufMut) {
         self.as_ref().encode(buf)
     }
+
+    fn length(&self) -> usize {
+        self.as_ref().length()
+    }
```
Oppen, line 283: This case can be optimized actually.
Encode is just prepending a length (that can be known at compile-time) in RLP encoding and then copying the bytes. This is specially true for `length`,  as it becomes a single constant number.

crates/common/types/account.rs
```diff
@@ -182,6 +187,14 @@ impl RLPEncode for AccountState {
             .encode_field(&self.code_hash)
             .finish();
     }
+
+    fn length(&self) -> usize {
+        let payload_len = self.nonce.length()
+            + self.balance.length()
+            + self.storage_root.length()
```
mpaulucci: what would happen if you added a field and forgot to add it here? Is there a unit test or something where this would break loudly?

crates/common/rlp/encode.rs
edg-l: good catch

crates/common/rlp/encode.rs
```diff
@@ -13,13 +12,79 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
+        return 1 + bytes_len; // prefix (0x80 + len) + payload
+    }
+
+    // long (>=56 bytes)
+    let be_len = bytes_len.ilog2() / 8 + 1;
+    1 + be_len as usize + bytes_len // prefix + len(len) + payload
+}
+
+#[derive(Debug, Clone, Copy, Default)]
+struct ByteCounter {
```
MegaRedHand, line 56: We should add some docs here:

```suggestion
/// Struct implementing `BufMut`, but only counting the number of bytes pushed into the buffer.
#[derive(Debug, Clone, Copy, Default)]
struct ByteCounter {
```

crates/common/rlp/encode.rs
```diff
@@ -13,13 +12,79 @@ pub fn encode<T: RLPEncode>(value: T) -> Vec<u8> {
...
 pub trait RLPEncode {
     fn encode(&self, buf: &mut dyn BufMut);
 
     fn length(&self) -> usize {
-        let mut buf = Vec::new();
-        self.encode(&mut buf);
-        buf.len()
+        let mut counter = ByteCounter::default();
+        self.encode(&mut counter);
+        counter.count
```
MegaRedHand, line 98: A comment here would be ideal:

```suggestion
        // Run the `encode` function, but only counting the bytes pushed.
        let mut counter = ByteCounter::default();
        self.encode(&mut counter);
        counter.count
```

crates/common/rlp/encode.rs
```diff
@@ -190,6 +327,21 @@ impl RLPEncode for U256 {
...
+        let first_byte = if sig_len == 0 {
+            0
+        } else {
+            let shift = 8 * (sig_len - 1);
+            ((self >> shift) & U256::from(0xff)).as_u32() as u8
+        };
+
+        let is_multibyte_mask = ((sig_len > 1) as usize) | ((first_byte > 0x7f) as usize);
+
+        1 + sig_len * is_multibyte_mask
```
Oppen: All of the integer types provide variants of the minimum number of bits needed to represent the value.
For `U256` it's `U256::bits(&self)`, for the primitive types it's `ilog2` (or its checked variant).
I think we can generalize the `length` implementation for them as follows:
```rust
fn impl_length_integers(bits: u32, lsb: u8) -> usize {
    let sig_len = (bits + 7) >> 3;
    let is_multibyte_mask = ((sig_len > 1) as usize) | ((lsb > 0x7f) as usize);
    1 + sig_len * is_multibyte_mask
}
```
Then the impl for `U256`:
```rust
fn length(&self) -> usize {
    impl_length_integers(self.bits(), (self.low_u32() & 0xff) as u8)
}
```
and for native types:
```rust
fn length(&self) -> usize {
    impl_length_integers(self.checked_ilog2().unwrap_or(0), (self & 0xff) as u8)
}
```

crates/common/rlp/encode.rs
edg-l: its needed to index self[0]

edg-l:
rlp benchmarks show an improvement in perf apparently

PR #5359, opened 2025-11-14
feat(l1,l2): make dev mode always use in memory db

fedacking:
**Motivation**

We want the dev mode to use the in memory db for development work.

**Description**

- Changed the init_l1 to use inmemory db if the dev option is on and the feature is enabled

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -279,7 +279,7 @@ impl Options {
     pub fn default_l2() -> Self {
         Self {
             network: Some(Network::LocalDevnetL2),
-            datadir: DB_ETHREX_DEV_L2.into(),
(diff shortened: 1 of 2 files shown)


cmd/ethrex/initializers.rs
```diff
@@ -382,7 +382,11 @@ pub async fn init_l1(
     opts: Options,
     log_filter_handler: Option<reload::Handle<EnvFilter, Registry>>,
 ) -> eyre::Result<(PathBuf, CancellationToken, PeerTable, NodeRecord)> {
-    let datadir = &opts.datadir;
+    let datadir: &PathBuf = if opts.dev && cfg!(feature = "dev") {
```
mpaulucci: instead of always, I would specify in memory by default. Also, I would set the default at the cli level, not here.

fedacking:
closed in favour of #5363

PR #5362, opened 2025-11-14
perf(levm): optimize set handling in the substate

iovoid:
**Motivation**

Currently `Substate::add_accessed_slot` takes about 3% of SLOAD. A couple of very similar structures also exist albeit with less impact.

**Description**

This switches from BTree to FxHash and always checks the current level first to prevent deep queries for hot values.


Oppen (approved):

iovoid:
No clear positive impact, the main improvement was likely from migrating to FxHash (done in #5824).

PR #5364, opened 2025-11-14
revert(l1, l2): revert building initially empty block

SDartayet:
**Description**

This reverts #5337. The time between payload building and fetching in devnet turned out to not be enough for some of the L2 test cases, so we'll have to figure out a new approach to building devnet blocks.


JereSalo (approved):
I suggest title `revert(l1, l2): revert revert revert`

PR #5366, opened 2025-11-14
perf(l1): cache account_codes blobs

MegaRedHand:
**Motivation**

Moving to BlobDB improved performance, but we suspect this also disabled caching for values.

**Description**

This PR configures an explicit blob cache of 128MB.

Flamegraph before:

Screenshot 2025-11-14 at 20 54 42

Flamegraph after:

Screenshot 2025-11-14 at 20 54 09

Notice `TableCache::Get` in the far left getting bigger than before. We also get a lot less decompression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-11-17
+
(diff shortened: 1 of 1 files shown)


jrchatruc:
Closed in favor of #5307

PR #5368, opened 2025-11-14
feat(l1): add target and script to run multiple benchmark runs

MegaRedHand:
This PR improves `tooling/import_benchmark` by adding a target that runs multiple repetitions of the same benchmark.


tooling/import_benchmark/Makefile
fedacking: Nitpick: I think that we should have the `rm -> cp -> cargo r -- import-bench` process defined in a single file, either the cargo or the make.

MegaRedHand: Done

PR #5373, opened 2025-11-17
fix(l2): return None when branch becomes empty in remove

sashass1315:
In BranchNode::remove() return None when the branch has zero children and no value after removal, instead of returning Mutated. Node::remove() relies on new_root.is_none() to mark the subtree as empty, and Trie::remove() uses that signal to reset the root to the empty trie (keccak(RLP_NULL)). LeafNode::remove() and ExtensionNode::remove() already adhere to this contract, so this change aligns BranchNode with the established semantics and prevents persisting a structurally empty branch at the root which would otherwise break the empty trie invariant.

### DIFF
--- a/crates/common/trie/node/branch.rs
+++ b/crates/common/trie/node/branch.rs
@@ -214,6 +214,9 @@ impl BranchNode {
             .enumerate()
             .filter(|(_, child)| child.is_valid())
             .collect::<Vec<_>>();
+        if children.is_empty() && self.value.is_empty() {
+            return Ok((None, value));
+        }
         let new_node = match (children.len(), !self.value.is_empty()) {
             // If this node still has a value but no longer has children, convert it into a leaf node
             (0, true) => NodeRemoveResult::New(


mpaulucci:
is this a bug? A perf optimization? How did you get into this? Why l2?

crates/common/trie/node/branch.rs
```diff
@@ -214,6 +214,9 @@ impl BranchNode {
             .enumerate()
             .filter(|(_, child)| child.is_valid())
             .collect::<Vec<_>>();
+        if children.is_empty() && self.value.is_empty() {
```
ElFantasma, line 217: Good placement — the early return sits between the children collection (line 210-215) and the restructuring match (line 219). One minor thought: you could alternatively add a `(0, false) => return Ok((None, value))` arm to the `match` below — the behavior is identical but it would keep all the restructuring cases in one place. Not a big deal either way.

ElFantasma (approved):

crates/common/trie/node/branch.rs
```diff
@@ -214,6 +214,9 @@ impl BranchNode {
...
+            return Ok((None, value));
+        }
```
iovoid, line 219: This can only happen if the branch has a single child and no value, which is already invalid. I think returning an error would be more appropiate.

iovoid (changes requested):

PR #5374, opened 2025-11-17
perf(l1): change encode_to_vec().len() -> length(), avoiding encoding to calc rlp length

edg-l:
**Motivation**

Followup on  #5350 since its already a big pr.

Changed:
- Replaced encode_to_vec().len() calls with length()
- Added more tests


Oppen (approved):

crates/networking/rpc/types/block.rs
```diff
@@ -72,9 +72,7 @@ impl RpcBlock {
         hash: H256,
         full_transactions: bool,
     ) -> Result<RpcBlock, RpcErr> {
-        let size = Block::new(header.clone(), body.clone())
-            .encode_to_vec()
-            .len();
+        let size = Block::new(header.clone(), body.clone()).length();
```
Oppen, line 75: I wonder if this can be done without cloning `header` and `body`. Not for this PR.

Maybe something like:
```rust
let tmp_block = Block::new(header, body);
let size = tmp_block.length();
let (header, body) = (tmp_block.header, tmp_block.body);
```
Ugly but I think *should*  work.

PR #5375, opened 2025-11-17
perf(levm): refactor `TrieLayerCache` structure

azteca1998:
**Motivation**

**Description**


crates/storage/trie_db/layering.rs
```diff
@@ -1,181 +1,88 @@
...
+                    if entry.previous.is_empty() {
+                        mem::take(&mut entry.value)
+                    } else {
+                        entry.previous.remove(0)
+                    },
+                ));
+
+                true
+            } else {
+                false
```
ElFantasma: I think these `true` and `false` should be the other way around: we want to retain the items that are not in last layer

PR #5384, opened 2025-11-18
[DNM] perf: add pprof instrumentation

Oppen:
Separate stages: initialization (includes `regenerate_head`) and
cruising.

**Motivation**

**Description**

Closes #issue_number


Oppen:
This was for temporary use during profiling.

PR #5388, opened 2025-11-19
perf(l1): validate blobs only once

Oppen:
**Motivation**

KZG signature verification takes a lot of time in flamegraphs.

**Description**

There are two changes here:
1. A fix to behavior, we should reject peers that send us unsolicited
   responses;
2. An optimization, we no longer revalidate a blob transaction that
   passed the verifications in `PooledTransactions::validate_requested`.

The first one might be a bit controversial. I want feedback on it and
will probably resend as a separate PR after. In my view, the current
behavior is an attack vector for DoS.


Arkenan:
I think we already have this. If not, we'll revive it.

PR #5389, opened 2025-11-19
feat(l1): handle parallel merkleization edge case

### LINKED ISSUE #5387: Support branch -> leaf/extension edge-case in parallel merkleization
In the current parallel merkleization implementation, we can fail on an edge case that can't happen in common networks: a state that starts with a branch as the root node and turns into an extension or leaf. We should add support for it, just for completeness.

MegaRedHand:
This PR adds handling for a specific edge case in the parallel merkleization implementation added in #5377

Closes #5387

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -1911,35 +1911,33 @@ fn collapse_root_node(
     };
     let path = Nibbles::from_hex(vec![choice as u8]);
     let child_bytes = match state_updates_map.get(&path) {
-        Some(v) => Some(v.clone()),
+        Some(v) => v.clone(),
         None => storage
             .state_trie(parent_header.hash())?
             .ok_or(StoreError::MissingStore)?
             .db()
-            .get(path)?,
+            .get(path)?
+            .ok_or_else(|| StoreError::Custom("Missing child node during root collapse".into()))?,
     };
-    let mut child = child_bytes.map(|b| Node::decode(&b)).transpose()?;
-    if let Some(child) = child.as_mut() {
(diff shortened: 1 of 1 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -432,20 +434,70 @@ impl Blockchain {
...
+                                extension_node.into()
+                            }
+                            Node::Leaf(leaf) => {
+                                let mut leaf = leaf.take();
+                                leaf.partial.prepend(choice as u8);
+                                leaf.into()
+                            }
+                        };
+                    }
+                    child
```
Oppen: I think we might want to extract this to a function, for two reasons:
1. It's a bit big and just a next-to-impossible edge-case, right in the middle of an already complex algorithm;
2. The compiler can't predict on its own that it's unlikely, but a function may be annotated as `#[cold]`, which helps the register allocator.

MegaRedHand: Done!

crates/blockchain/blockchain.rs
```diff
@@ -1872,5 +1888,59 @@ pub fn get_total_blob_gas(tx: &EIP4844Transaction) -> u32 {
...
+    };
+    let path = Nibbles::from_hex(vec![choice as u8]);
+    let child_bytes = match state_updates_map.get(&path) {
+        Some(v) => Some(v.clone()),
+        None => storage
+            .state_trie(parent_header.hash())?
+            .ok_or(StoreError::MissingStore)?
+            .db()
+            .get(path)?,
+    };
```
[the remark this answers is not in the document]
MegaRedHand: Done

PR #5392, opened 2025-11-19
feat(l2): add ZisK backend

xqft:
**Motivation**

To add ZisK as a new backend for ethrex-prover.

**Description**

- Implements ZisK guest program and code to build it
- Adds ZisK as a new backend to the ethrex-prover SDK
- Implements  `eip7702_recover_address()` using `k256` to make it compatible with ZisK (and this also optimizes the function for other zkVMs)

Closes #issue_number

### DIFF
--- a/.github/workflows/tag_release.yaml
+++ b/.github/workflows/tag_release.yaml
@@ -205,7 +205,7 @@ jobs:
 
       - name: Build ethrex elf - ${{ matrix.zkvm }}
         run: |
-          cargo build --release --package guest_program --features ${{ matrix.zkvm }}
+          cargo build --release --package guest_program --features "${{ matrix.zkvm }},ci"
           mkdir -p ${{ matrix.zkvm }}_verification_keys
           if [ "${{ matrix.zkvm }}" = "sp1" ]; then
             mv crates/l2/prover/src/guest_program/src/${{ matrix.zkvm }}/out/riscv32im-succinct-zkvm-elf ethrex-riscv32im-${{ matrix.zkvm }}-elf-${{ github.ref_name }}


crates/vm/levm/Cargo.toml
```diff
@@ -40,8 +40,9 @@ ark-bn254 = "0.5.0"
 ark-ec = "0.5.0"
 ark-ff = { version = "0.5.0", features = ["asm"] }
 strum = { version = "0.27.1", features = ["derive"] }
-k256 = "0.13.4"
+k256.workspace = true
```
ilitteri, line 43: Why isn't this optional?

crates/common/Cargo.toml
```diff
@@ -31,14 +31,21 @@ rayon.workspace = true
 url.workspace = true
 rkyv.workspace = true
 rustc-hash.workspace = true
+k256.workspace = true
```
tomip01, line 34: Maybe too nit but I think this should be used optionally with sp1, risc0 or zisk

crates/l2/common/Cargo.toml
```diff
@@ -18,13 +18,16 @@ bytes.workspace = true
...
 lambdaworks-crypto.workspace = true
-secp256k1.workspace = true
+sha3.workspace = true
 hex.workspace = true
 serde_with.workspace = true
 # We do not need to enable the "unaligned" feature here since we are just using
 # rkyv for ProverInputData, a data structure meant to be used in the host and not
 # inside a guest program
 rkyv.workspace = true
+k256.workspace = true
```
tomip01, line 28: same here

PR #5402, opened 2025-11-21
perf(l1): improve rlp decoding

edg-l:
**Motivation**

Basic integers like u64 used static_left_pad, that uses a temporary zeroed array, now they directly produce the integer. Did this by using decode_be_integer_u64, whose result can be casted to the relevant iX and decode_be_integer_u128, separate to avoid paying the cost of u128 always.

Modified decode_rlp_item so that it doesnt use `usize::from_be_bytes(static_left_pad(x))`, and instead uses a dedicated method `parse_be_length` which has less overhead.

Improved static_left_pad to not always zero init an array, and added the special case for data.len() == N to avoid a saturating sub.

Helps with https://github.com/lambdaclass/ethrex/issues/4956

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,3 +21,5 @@ hive/
 ethereum-package/
 tooling/ef_tests/blockchain/vectors
 tooling/ef_tests/state/vectors
+dev_ethrex_l1/
+dev_ethrex_l2/
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,20 @@
 # Lambdaclass core team
 * @lambdaclass/lambda-execution-reviewers
+
+## ethrex L2 code owners
+crates/l2 @ilitteri @manuelbilbao @avilagaston9
 crates/l2/contracts/src @jrchatruc @manuelbilbao
+cmd/ethrex/l2 @ilitteri @manuelbilbao @avilagaston9
+cmd/ethrex/build* @ilitteri @manuelbilbao @avilagaston9
+crates/blockchain/dev @ilitteri @manuelbilbao @avilagaston9
+crates/common/crypto @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/blobs_bundle.rs @ilitteri @manuelbilbao @avilagaston9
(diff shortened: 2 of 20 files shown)


crates/common/rlp/decode.rs
```diff
@@ -515,17 +518,22 @@ pub fn decode_bytes(data: &[u8]) -> Result<(&[u8], &[u8]), RLPDecodeError> {
 /// The size of the data must be less than or equal to the size of the output array.
 #[inline]
 pub fn static_left_pad<const N: usize>(data: &[u8]) -> Result<[u8; N], RLPDecodeError> {
```
Oppen: Is this still called after these changes? Maybe we can remove the function.

crates/common/rlp/decode.rs
```diff
@@ -534,6 +542,71 @@ pub fn static_left_pad<const N: usize>(data: &[u8]) -> Result<[u8; N], RLPDecode
     Ok(result)
 }
 
+/// Decodes a RLP integer limited by max_len into a u64.
+#[inline]
+fn decode_be_integer_u64(bytes: &[u8], max_len: usize) -> Result<u64, RLPDecodeError> {
+    debug_assert!(max_len <= 8);
```
Oppen: I suggest we either add an error case (in addition to the `debug_assert`) for when `max_len > 8` or make it a const generic and assert at compile time that it is `<= 8`.

crates/common/rlp/decode.rs
```diff
@@ -534,6 +542,71 @@ pub fn static_left_pad<const N: usize>(data: &[u8]) -> Result<[u8; N], RLPDecode
...
+        return Err(RLPDecodeError::MalformedData);
+    }
+
+    let mut value: usize = 0;
+    for byte in bytes {
+        value = value
+            .checked_mul(256)
+            .and_then(|v| v.checked_add(*byte as usize))
+            .ok_or(RLPDecodeError::InvalidLength)?;
+    }
```
Oppen: Probably just:
1. Check the length of the slice beforehand, it must not be greater than `size_of::<usize>()`;
2. Use shifts and binary ors.

In general, I wonder why this isn't just another call to `decode_be_integer_u64`.

PR #5404, opened 2025-11-21
chore(l2): use ZisK SDK instead of subprocesses

xqft:
**Motivation**

self explanatory.

Note: we require https://github.com/0xPolygonHermez/zisk/pull/674 for this to work. Don't merge until those changes are applied to ZisK.

### DIFF
--- a/crates/l2/prover/src/backend/zisk.rs
+++ b/crates/l2/prover/src/backend/zisk.rs
@@ -1,11 +1,6 @@
-use std::{
-    io::ErrorKind,
-    process::{Command, Stdio},
-    sync::OnceLock,
-    time::{Duration, Instant},
-};
+use std::{io::ErrorKind, sync::OnceLock};
 use zisk_common::io::ZiskStdin;
-use zisk_sdk::{Asm, Proof, ProverClient, ZiskProveResult, ZiskProver};
+use zisk_sdk::{Asm, Proof, ProverClient, ZiskProver};
 
 use ethrex_l2_common::prover::{BatchProof, ProofFormat};
 use guest_program::{ZKVM_ZISK_PROGRAM_ELF, input::ProgramInput, output::ProgramOutput};
@@ -13,8 +8,6 @@ use guest_program::{ZKVM_ZISK_PROGRAM_ELF, input::ProgramInput, output::ProgramO
 const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zisk_input.bin");
 const OUTPUT_DIR_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zisk_output");
 const ELF_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zkvm-zisk-program");
-
-pub struct ProveOutput(pub Vec<u8>);
 
 pub static PROVE_CLIENT: OnceLock<ZiskProver<Asm>> = OnceLock::new();
 pub static EXECUTE_CLIENT: OnceLock<ZiskProver<Asm>> = OnceLock::new();
@@ -70,6 +63,10 @@ pub fn prove(
     input: ProgramInput,
     format: ProofFormat,
 ) -> Result<Proof, Box<dyn std::error::Error>> {
+    if format != ProofFormat::Compressed {
+        unimplemented!("ZisK only supports the Compressed proof format");
+    }
+
     write_elf_file()?;
     let stdin_bytes = rkyv::to_bytes::<rkyv::rancor::Error>(&input)?.to_vec();
     let stdin = ZiskStdin::from_vec(stdin_bytes);
@@ -85,7 +82,7 @@ pub fn verify(_output: &ProgramOutput) -> Result<(), Box<dyn std::error::Error>>
 
 pub fn to_batch_proof(
     proof: Proof,
-    format: ProofFormat,
+    _format: ProofFormat,
 ) -> Result<BatchProof, Box<dyn std::error::Error>> {
     unimplemented!("to_batch_proof is not implemented for ZisK backend")
 }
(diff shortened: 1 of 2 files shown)


crates/l2/prover/src/backend/zisk.rs
```diff
@@ -10,87 +9,80 @@ const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zisk_input.bin");
...
 const ELF_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zkvm-zisk-program");
 
-pub struct ProveOutput(pub Vec<u8>);
+pub static PROVE_CLIENT: OnceLock<ZiskProver<Asm>> = OnceLock::new();
+pub static EXECUTE_CLIENT: OnceLock<ZiskProver<Asm>> = OnceLock::new();
 
-pub fn execute(input: ProgramInput) -> Result<(), Box<dyn std::error::Error>> {
-    write_elf_file()?;
+pub fn execute_client() -> &'static ZiskProver<Asm> {
+    if PROVE_CLIENT.get().is_some() {
```
ElFantasma, line 16: `panic!` here will crash the node if the prover is accidentally initialized for the wrong mode. Since this is a recoverable misconfiguration, return a `Result` (or `BackendError` after rebasing) instead.

```rust
pub fn execute_client() -> Result<&'static ZiskProver<Asm>, Box<dyn std::error::Error>> {
    if PROVE_CLIENT.get().is_some() {
        return Err("ZisK prover was previously initialized for proving...".into());
    }
    // ...
}
```

crates/l2/prover/src/backend/zisk.rs
```diff
@@ -10,87 +9,80 @@ const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zisk_input.bin");
...
+        );
+    }
+    PROVE_CLIENT.get_or_init(|| {
+        ProverClient::builder()
+            .asm()
+            .prove()
+            .aggregation(true)
+            .elf_path(ELF_PATH.into())
+            .unlock_mapped_memory(true)
+            .gpu(Default::default())
```
ElFantasma, line 45: `.gpu(Default::default())` is called unconditionally, meaning the prove client always attempts GPU initialization even when built without the `gpu` feature. This should be conditional:

```rust
let mut builder = ProverClient::builder()
    .asm()
    .prove()
    .aggregation(true)
    .elf_path(ELF_PATH.into())
    .unlock_mapped_memory(true);

#[cfg(feature = "gpu")]
{ builder = builder.gpu(Default::default()); }

builder.build()
```

crates/l2/prover/src/backend/zisk.rs
```diff
@@ -10,87 +9,80 @@ const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zisk_input.bin");
...
-        .stderr(Stdio::inherit())
-        .output()?;
-
-    if !output.status.success() {
-        return Err(format!(
-            "ZisK proof generation failed: {}",
-            String::from_utf8_lossy(&output.stderr)
-        )
-        .into());
+) -> Result<Proof, Box<dyn std::error::Error>> {
```
ElFantasma, line 65: `unimplemented!()` will panic at runtime. Use a proper error return instead:

```rust
if format != ProofFormat::Compressed {
    return Err("ZisK only supports the Compressed proof format".into());
}
```

This is especially important since `ProofFormat` is configurable — a user could set `Groth16` in config and get a panic instead of a clear error message.

crates/l2/prover/src/backend/zisk.rs
```diff
@@ -10,87 +9,80 @@ const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zisk_input.bin");
...
+        );
+    }
+    EXECUTE_CLIENT.get_or_init(|| {
+        ProverClient::builder()
+            .asm()
+            .verify_constraints()
+            .elf_path(ELF_PATH.into())
+            .unlock_mapped_memory(true)
+            .build()
+            .unwrap_or_else(|e| panic!("Failed to setup ZisK prover client: {e}"))
```
ElFantasma, line 28: Same `panic!` concern as `execute_client()` — the MPI mutual-exclusion guard should return an error, not crash the node.

crates/l2/prover/src/backend/mod.rs
```diff
@@ -15,15 +15,12 @@ pub mod sp1;
 #[cfg(feature = "zisk")]
 pub mod zisk;
 
-#[derive(Default, Debug, Deserialize, Serialize, Copy, Clone, ValueEnum)]
+#[derive(Default, Debug, Deserialize, Serialize, Copy, Clone, ValueEnum, PartialEq)]
```
ElFantasma, line 18: Removing `#[cfg(feature)]` from enum variants means users can *parse* `--backend zisk` even when ZisK isn't compiled, only to get a runtime error deep in the match arms. The current approach on `main` (feature-gating the variants) catches this at config/CLI time. Consider keeping the enum gated and just improving the error message from `FromStr`.

ElFantasma (changes requested):
This PR is heavily conflicting with `main` — the codebase was refactored to a trait-based `ProverBackend` system (with `BackendType`, `BackendError`, etc.) since this branch diverged. The SDK migration itself (replacing `ziskemu`/`cargo-zisk` subprocesses with `zisk_sdk::ZiskProver`) is the right direction, but the PR needs a rebase onto the current architecture before it can be reviewed meaningfully.

Beyond the conflicts, the main concerns are: (1) `panic!` in `execute_client()`/`prove_client()` for the MPI mutual-exclusion check — production node code should return errors, not crash; (2) `unimplemented!()` macro in `prove()` will also panic at runtime instead of returning an error; (3) `.gpu(Default::default())` is called unconditionally in `prove_client()`, but GPU support should be gated behind the `gpu` feature flag.

**`crates/l2/prover/src/backend/zisk.rs:12-13`** — `OUTPUT_DIR_PATH` and `INPUT_PATH` are now dead code since the SDK handles I/O internally. Only `ELF_PATH` is still referenced (by `write_elf_file`). Remove the unused constants.

PR #5405, opened 2025-11-21
chore(l2): fix prover timing

xqft:
**Motivation**

To measure execution and proving time correctly. This means removing the client initialization from the equation. In the case of ZisK, we use it's own timing exposed via the SDK.


xqft:
superseded by https://github.com/lambdaclass/ethrex/pull/5474

PR #5409, opened 2025-11-23
fix(l1,l2): make RLP integer impl_encode single-pass and zero-allocation

Forostovec:
Reworked impl_encode in crates/common/rlp/encode.rs to compute the first non-zero byte in a single pass and to encode using direct slices of the original big-endian array. This eliminates the previous double scan and avoids allocating/copying into ArrayVec, which was unnecessary for encoding the integer payload. As a direct consequence, this change also fixes a correctness bug where u128 encoding could panic at runtime due to relying on ArrayVec<[u8; 8]> and attempting to extend it with 16 bytes. The new logic preserves RLP semantics for zero (0x80), single-byte values in the 0x01..=0x7f range (encoded as-is), and multi-byte integers (prefixed with 0x80 + length followed by the trimmed big-endian bytes), aligning with how other parts of the module handle length-prefixed payloads without redundant work.

### DIFF
--- a/crates/common/rlp/encode.rs
+++ b/crates/common/rlp/encode.rs
@@ -48,32 +48,18 @@ impl RLPEncode for bool {
 // integer types impls
 
 fn impl_encode<const N: usize>(value_be: [u8; N], buf: &mut dyn BufMut) {
-    let mut is_multi_byte_case = false;
-
-    for n in value_be.iter().take(N - 1) {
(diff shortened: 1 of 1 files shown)


edg-l:
There is already a pr open which improves the encoding https://github.com/lambdaclass/ethrex/pull/5350

Forostovec:
> There is already a pr open which improves the encoding #5350

Ops, then i will close this

PR #5412, opened 2025-11-24
feat(l2): prover backend for OpenVM

xqft:
**Motivation**

Integrates OpenVM as a new prover backend. For now it only supports L1 proving, until we can implement blob verification using `openvm-kzg`.

Currently proving is slow as we didn't yet use the guest libraries/precompiles for crypto primitives.

L2 block proving is not supported for now. The issue blocking this is that openvm-kzg does not support blob KZG proof verification.

**Description**

- adds an OpenVM guest program for stateless execution
- adds a new prover backend for OpenVM
- added a new kzg implementation using `openvm-kzg` because of the guest incompatibility with `c-kzg` and `kzg-rs`.
- adds CI jobs for the new backend

### DIFF
--- a/.github/workflows/pr-main_l2_prover.yaml
+++ b/.github/workflows/pr-main_l2_prover.yaml
@@ -48,17 +48,12 @@ jobs:
         if: matrix.backend == 'zisk'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GH_RUNNER: 1
-        uses: nick-fields/retry@v3
-        with:
-          max_attempts: 3
-          retry_wait_seconds: 0
-          timeout_minutes: 15
-          command: |
-            sudo apt-get update
-            sudo apt-get install -y xz-utils jq curl build-essential qemu-system libomp-dev libgmp-dev nlohmann-json3-dev protobuf-compiler uuid-dev libgrpc++-dev libsecp256k1-dev libsodium-dev libpqxx-dev nasm libopenmpi-dev openmpi-bin openmpi-common libclang-dev clang gcc-riscv64-unknown-elf
-            curl https://raw.githubusercontent.com/0xPolygonHermez/zisk/main/ziskup/install.sh | bash
-            echo "$HOME/.zisk/bin" >> $GITHUB_PATH
+          SETUP_KEY: none
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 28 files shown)


crates/l2/prover/src/guest_program/build.rs
```diff
@@ -187,6 +190,39 @@ fn build_zisk_program() {
...
+        .current_dir("./src/openvm")
+        .status()
+        .expect("failed to execute cargo openvm build");
+
+    if !status.success() {
+        panic!("cargo openvm build failed with exit status: {}", status);
+    }
+
+    let elf_src =
+        Path::new("./src/openvm/target/riscv32im-risc0-zkvm-elf/release/zkvm-openvm-program");
```
ilitteri, line 212:
```suggestion
        Path::new("./src/openvm/target/riscv32im-openvm-zkvm-elf/release/zkvm-openvm-program");
```

crates/l2/prover/src/guest_program/Cargo.toml
```diff
@@ -28,6 +28,13 @@ risc0-build = { version = "=3.0.3", optional = true }
 risc0-zkvm = { version = "=3.0.3", optional = true, features = ["getrandom"] }
 sp1-build = { version = "=5.0.8", optional = true }
 sp1-sdk = { version = "=5.0.8", optional = true }
+openvm-build = { git = "https://github.com/openvm-org/openvm.git", package = "openvm-build", optional = true }
+openvm-sdk = { git = "https://github.com/openvm-org/openvm.git", package = "openvm-sdk", optional = true, features = [
+    "evm-prove",
+    "parallel",
+] }
+bincode = "1.3"
+serde.workspace = true
```
ilitteri: Why this if we are using the toolchain in the `build.rs`?

crates/l2/prover/src/guest_program/build.rs
xqft: it actually compiles into a folder with that name, probably because they leverage risc0's toolchain

crates/l2/prover/src/guest_program/Cargo.toml
xqft: no reason, removed!

crates/l2/prover/src/guest_program/build.rs
ilitteri: :O

Issue #5411, opened 2025-11-24
RLP Benchmarks


Arkenan:
We need benchmarks that measure how much time it takes to encode and decode each ethereum type and each basic type. 

## Basic types

- [ ] Integer
- [ ] String of size 5, 60, 500
- [ ] List of 10, 100, 1000 integers
- [ ] List of 10, 100, 1000 strings, each of size 10

### Common types

- [ ] AccountInfo
- [ ] AccountState
- [ ] BlobsBundle
- [ ] Block
- [ ] BlockHeader
- [ ] Withdrawal
- [ ] ForkId
- [ ] Receipt
- [ ] ReceiptWithBloom
- [ ] Log
- [ ] EncodedRequests

### Trie

- [ ] Nibble of size 65, 129, 130, 500
- [ ] NodeHash
- [ ] BranchNode
- [ ] ExtensionNode
- [ ] LeafNode

### Transactions

- [ ] P2PTransaction
- [ ] WrappedEIP4844Transaction
- [ ] TxKind
- [ ] LegacyTransaction
- [ ] EIP2930Transaction
- [ ] EIP1559Transaction
- [ ] EIP4844Transaction
- [ ] EIP7702Transaction
- [ ] PrivilegedL2Transaction
- [ ] FeeTokenTransaction
- [ ] MempoolTransaction

### Networking/p2p

- [ ] Endpoint
- [ ] NodeRecord
- [ ] Node
- [ ] PingMessage
- [ ] FindNodeMessage
- [ ] NeighborsMessage
- [ ] ENRRequestMessage
- [ ] ENRRequestMessage
- [ ] Capability
- [ ] AccountRangeUnit
- [ ] AccountStateSlim
- [ ] StorageSlot
- [ ] AuthMessage
- [ ] AckMessage
- [ ] HashOrNumber

edg-l:
we got a base pr here https://github.com/lambdaclass/ethrex/pull/5410 with a skeleton

we will now work to add the benches there

PR #5413, opened 2025-11-25
chore(l1): add rlp encode benches 

edg-l:
Adds the encoding benches for rlp

Part of https://github.com/lambdaclass/ethrex/issues/5411

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,3 +21,5 @@ hive/
 ethereum-package/
 tooling/ef_tests/blockchain/vectors
 tooling/ef_tests/state/vectors
+dev_ethrex_l1/
+dev_ethrex_l2/
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,20 @@
 # Lambdaclass core team
 * @lambdaclass/lambda-execution-reviewers
+
+## ethrex L2 code owners
+crates/l2 @ilitteri @manuelbilbao @avilagaston9
 crates/l2/contracts/src @jrchatruc @manuelbilbao
+cmd/ethrex/l2 @ilitteri @manuelbilbao @avilagaston9
+cmd/ethrex/build* @ilitteri @manuelbilbao @avilagaston9
+crates/blockchain/dev @ilitteri @manuelbilbao @avilagaston9
+crates/common/crypto @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/blobs_bundle.rs @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/l2 @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/l2.rs @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/transaction.rs @ilitteri @manuelbilbao @avilagaston9
+crates/common/rkyv_utils.rs @ilitteri @manuelbilbao @avilagaston9
+crates/networking/p2p/rlpx/l2 @ilitteri @manuelbilbao @avilagaston9
+crates/networking/rpc/types/transaction.rs @ilitteri @manuelbilbao @avilagaston9
+crates/networking/rpc/clients @ilitteri @manuelbilbao @avilagaston9
+crates/vm/levm/src/hooks/l2_hook.rs @ilitteri @manuelbilbao @avilagaston9
+crates/blockchain/dev @ilitteri @manuelbilbao @avilagaston9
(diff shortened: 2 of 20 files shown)


crates/common/rlp/benches/encode.rs
```diff
@@ -1,36 +1,1388 @@
...
+        let mut rng = StdRng::seed_from_u64(3);
+        (0..1_000_000)
+            .map(|_| {
+                let value: u32 = rng.r#gen();
+                value % 1_000_001
+            })
+            .collect::<Vec<u32>>()
+    };
+    group.bench_function("encode_u32_random_seeded", move |b| {
+        let mut buf = Vec::new();
```
Oppen: Use `with_capacity` for stability across iterations. Otherwise the first one will be allocating while the rest won't, creating an outlier in the set.

crates/common/rlp/benches/encode.rs
```diff
@@ -1,36 +1,1388 @@
...
+                value.encode(&mut buf);
+            }
+            black_box(&buf);
+        });
+    });
+
+    let u256_values = {
+        let mut rng = StdRng::seed_from_u64(6);
+        (0..1_000_000)
+            .map(|_| random_u256(&mut rng))
```
Oppen, line 214: Doesn't `U256` provide a `random` constructor like with `H256`?

crates/common/rlp/benches/encode.rs
Oppen, line 1: I think we should aim at keeping benches in `benches` subdir. I know a few of them are in other places (and I'm guilty of a few), but keeping crates free from this code would reduce noise in the day to day work and save us from overestimating our line count (assuming we only care about the code that implements functionality, rather than code that measures its performance or tests it).

Oppen: @mpaulucci @Arkenan wdyt?

crates/common/rlp/benches/encode.rs
edg-l: doesn't look like it

crates/common/rlp/benches/encode.rs
edg-l: you are right, added it for all benches

crates/common/rlp/benches/encode.rs
mpaulucci: I don't mind either way, but we should be consitent. If we go for the top level `benches` lets create a ticket to move all benches there

Arkenan: I agree with the benches dir for anything that the crate exposes. However, I wouldn't do that in this PR, let's add a different one moving everything there.

crates/common/rlp/benches/decode.rs
Arkenan: What is this file? Should we remove it?

.github/workflows/pr_perf_rlp.yaml
```diff
@@ -0,0 +1,70 @@
...
+        uses: ./.github/actions/setup-rust
+        with:
+          components: rustfmt, clippy
+
+      - name: Install critcmp
+        run: cargo install critcmp
+
+      - name: Run bench main
+        continue-on-error: true
+        run: make bench-rlp
```
ElFantasma, line 36: The benchmark comparison won't work as-is. Two problems:

1. `cargo bench` (via `make bench-rlp`) without `--save-baseline` writes to the default criterion baseline. The second run on the PR branch **overwrites** the first run's results, so there's nothing to compare.
2. `critcmp --list` (line 50) only lists the names of saved baselines — it doesn't produce a comparison table.

Fix: save each run to a named baseline and compare them:
```yaml
- name: Run bench main
  continue-on-error: true
  run: cd ./crates/common/rlp && cargo bench -- --save-baseline main

# ... checkout PR ...

- name: Run bench PR
  run: cd ./crates/common/rlp && cargo bench -- --save-baseline pr

- name: Save comparison
  run: |
    {
      printf '## RLP Bench Results\n\n```\n'
      critcmp main pr
      printf '\n```\n'
    } > result.md
```

crates/common/Cargo.toml
```diff
@@ -37,6 +37,7 @@ secp256k1 = { workspace = true, optional = true }
 
 [dev-dependencies]
 hex-literal.workspace = true
+ethrex-p2p.workspace = true
```
ElFantasma, line 40: The benchmarks live in `ethrex-rlp` (`crates/common/rlp/`), not in `ethrex-common`. This dev-dependency seems unnecessary here and creates a circular dev-dependency (`ethrex-common` ← `ethrex-p2p` → `ethrex-common`). While Cargo allows circular dev-deps, it adds compilation overhead without benefit. Consider removing this and keeping the `ethrex-p2p` dev-dep only in `crates/common/rlp/Cargo.toml` where the benches actually are.

crates/common/rlp/benches/encode.rs
```diff
@@ -0,0 +1,1440 @@
...
+        });
+    });
+
+    group.finish();
+}
+
+fn bench_encode_blobs_bundle(c: &mut Criterion) {
+    let mut group = c.benchmark_group("common_types");
+
+    let blobs_bundle = black_box(BlobsBundle {
```
ElFantasma, line 1195: nit: This `BlobsBundle` has 4 blobs but only 1 commitment and 1 proof. In practice these counts must always match (one commitment + one proof per blob). Might want to use matching counts so the benchmark encodes a realistic structure:
```rust
blobs: vec![[6u8; BYTES_PER_BLOB]; 4],
commitments: vec![[0x78u8; 48]; 4],
proofs: vec![[0x78u8; 48]; 4],
```

PR #5414, opened 2025-11-25
chore(l1): add rlp decode benches (part 1)

azteca1998:
**Motivation**

**Description**


crates/common/rlp/benches/decode.rs
```diff
@@ -0,0 +1,556 @@
...
+fn bench_decode_trie(c: &mut Criterion) {
+    let mut group = c.benchmark_group("decode_trie");
+
+    group.bench_function(BenchmarkId::new("NodeHash", "hashed/1000"), |b| {
+        b.iter_batched_ref(
+            || {
+                rand::rng()
+                    .random_iter::<[u8; 32]>()
+                    .map(H256)
+                    .map(NodeHash::Hashed)
```
ElFantasma, line 366: Bug: `random_iter::<[u8; 32]>()` produces an **infinite iterator** — there's no `.take(1000)` like the other benchmarks have. `.collect::<Vec<_>>()` on an infinite iterator will OOM/hang.

Should be:
```rust
rand::rng()
    .random_iter::<[u8; 32]>()
    .take(1000)
    .map(H256)
    .map(NodeHash::Hashed)
    .map(|x| x.encode_to_vec())
    .collect::<Vec<_>>()
```

crates/common/rlp/benches/decode.rs
```diff
@@ -0,0 +1,556 @@
...
+    // TODO: Benchmark NodeHash::Inline (len in {8, 16, 24}).
+
+    // TODO: Benchmark BranchNode (empty, full, random; hash, node).
+    // TODO: Benchmark ExtensionNode (short path, long path; hash, node).
+    // TODO: Benchmark LeafNode (short path, long path; short value, long value).
+
+    group.finish();
+}
+
+fn bench_decode_transactions(c: &mut Criterion) {
```
ElFantasma: nit: `bench_decode_transactions` is registered in `criterion_group!` but the function body is empty (just TODOs). Unlike an empty group (which at least calls `group.finish()`), this function does literally nothing — criterion will still set up and tear down a benchmark group for it. Consider removing it from the `criterion_group!` macro until it has actual benchmarks, to avoid confusing output.

crates/common/rlp/Cargo.toml
```diff
@@ -16,8 +16,22 @@ snap.workspace = true
 
 [dev-dependencies]
 hex-literal.workspace = true
+criterion = { version = "0.7.0", features = ["html_reports"] }
+ethrex-common.workspace = true
+ethrex-p2p.workspace = true
```
ElFantasma: `ethrex-p2p` as a dev-dependency of `ethrex-rlp` creates a circular dependency chain: rlp → p2p → blockchain → storage → common → rlp. While Cargo handles circular dev-deps at compile time, this pulls nearly the entire workspace into the benchmark build graph, significantly increasing compile times for what should be a leaf crate.

Consider either:
1. Moving the p2p-type benchmarks (Endpoint, Capability, StorageSlot) to a benchmark in `ethrex-p2p` itself, or
2. Creating the test data inline (manually construct the RLP bytes) instead of importing the types.

ElFantasma (changes requested):

ElFantasma (approved):
👍

crates/common/rlp/benches/decode.rs
```diff
@@ -0,0 +1,401 @@
...
+    let mut group = c.benchmark_group("decode_nibbles");
+
+    fn impl_bench<const N: usize, const L: usize, const IS_LEAF: bool>(b: &mut Bencher) {
+        b.iter_batched_ref(
+            || {
+                let mut data = Vec::with_capacity(N);
+                for _ in 0..N {
+                    data.push(
+                        Nibbles::from_raw(
+                            &rand::distr::Uniform::<u8>::new(0, 16)
```
iovoid: from_raw split bytes into nibbles, so all upper nibbles would be zero with this

azteca1998: Nice catch. This PR is from before the refactor that packed the nibbles together.

Issue #5043, opened 2025-10-24
Document that `Blockchain::execute_block` doesn't check header's `transactionsRoot` matches body


MegaRedHand:
Our implementation of `Blockchain::execute_block` doesn't check that the header's transactions root matches the transactions in the block's body. This assumes the caller validates it, which indeed happens when the block comes from P2P or from the engine API. We should, however, document this in the function, and check _all_ the callers indeed check this.

MegaRedHand:
Check in P2P (look for callers of the function):

https://github.com/lambdaclass/ethrex/blob/c7ec3b323b5d823857e8fe68c30bba95f82340da/crates/common/types/block.rs#L625

For the Engine API part, look at the spec

PR #5428, opened 2025-11-26
fix(l1): verify transactions root when validating blocks

mpaulucci:
**Motivation**
We were not validating the transactions root

**Description**
- Validates transaction root inside a block
- Adds tests


MegaRedHand:
Related to #5043, we _are_ validating the transactions root coming from P2P requests, and we don't receive one from consensus, just the transactions.

Issue #3754, opened 2025-07-21
L2: remove l2 logic from `eth_client`


avilagaston9:
Currently, the `eth_client` is full of L2 stuff:
- `RpcBatch` (duplicated in `crates/l2/networking/rpc/l2/batch.rs`)
- `L1MessageProof`
- `send_privileged_l2_transaction()`
- `get_last_commited/verified_batch()`
- etc.

We should move all L2 logic into `crates/l2/networking/rpc/clients.rs` and use the general `send_request` method to interact with the `ethClient`.

PR #5433, opened 2025-11-26
feat(l2): add helper functions for l2 eth client

tomip01:
**Motivation**

We want this new functions that will be useful for `rex` and future use.

**Description**

Add `get_batch_number()` function to call `ethrex_batchNumber`
Add `send_ethrex_transaction()` function to call `ethrex_sendTransaction`

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -7,7 +7,9 @@ use ethrex_crypto::keccak::keccak_hash;
 pub use mempool::MempoolTransaction;
 use rkyv::{Archive, Deserialize as RDeserialize, Serialize as RSerialize};
 use serde::{Serialize, ser::SerializeStruct};
-pub use serde_impl::{AccessListEntry, GenericTransaction, GenericTransactionError};
+pub use serde_impl::{
+    AccessListEntry, AuthorizationTupleEntry, GenericTransaction, GenericTransactionError,
+};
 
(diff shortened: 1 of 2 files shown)


crates/networking/rpc/clients/eth/errors.rs
```diff
@@ -327,6 +329,16 @@ pub enum GetBatchByNumberError {
...
+#[derive(Debug, thiserror::Error)]
+pub enum GetBatchNumberError {
+    #[error("{0}")]
+    SerdeJSONError(#[from] serde_json::Error),
+    #[error("{0}")]
+    RPCError(String),
+    #[error("{0}")]
+    ParseIntError(#[from] std::num::ParseIntError),
+}
+
```
JereSalo, line 343: I see the enum above this has a TODO that says `Move to L2`, do you know if there's an issue created for that? Would it be easy to achieve?

JereSalo (commented):
Looks good, idk if it was tested but it would be good if it was

crates/networking/rpc/clients/eth/errors.rs
tomip01: I didn't find any. The closes was [this](https://github.com/lambdaclass/ethrex/issues/3754)

tomip01:
> Looks good, idk if it was tested but it would be good if it was

It tested it manually

PR #5434, opened 2025-11-26
feat(l1): cache block hash lookup for BLOCKHASH opcode

### LINKED ISSUE #5344: gas benchmarks: blockhash
This looks like our worst performing benchmark, about 33x behind other nodes.

fmoletta:
**Motivation**
Addressing #5344 by using the existing `block_hash_cache` to store block hashes looked up during `BLOCKHASH` opcode execution to speed up subsequent lookups

**Description**
* Add interior mutability to block_hash_cache (HashMap -> Arc<Mutex<BTreeMap>>)
* Insert block hashes obtained during lookup to block_hash_cache (BLOCKHASH opcode)
* Use oldest cached block successor to shortcut ancestor lookup (BLOCKHASH opcode)

Benchmark Results: x33,6 improvement on BlockHash EEST gas benchmark: 1080.92 Mgas/s vs 32.18 Gas/s on main (7b4a0bab)

Closes  #5344
Closes #5443

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,20 @@
 # Lambdaclass core team
 * @lambdaclass/lambda-execution-reviewers
+
+## ethrex L2 code owners
+crates/l2 @ilitteri @manuelbilbao @avilagaston9
 crates/l2/contracts/src @jrchatruc @manuelbilbao
+cmd/ethrex/l2 @ilitteri @manuelbilbao @avilagaston9
+cmd/ethrex/build* @ilitteri @manuelbilbao @avilagaston9
+crates/blockchain/dev @ilitteri @manuelbilbao @avilagaston9
+crates/common/crypto @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/blobs_bundle.rs @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/l2 @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/l2.rs @ilitteri @manuelbilbao @avilagaston9
+crates/common/types/transaction.rs @ilitteri @manuelbilbao @avilagaston9
+crates/common/rkyv_utils.rs @ilitteri @manuelbilbao @avilagaston9
+crates/networking/p2p/rlpx/l2 @ilitteri @manuelbilbao @avilagaston9
+crates/networking/rpc/types/transaction.rs @ilitteri @manuelbilbao @avilagaston9
+crates/networking/rpc/clients @ilitteri @manuelbilbao @avilagaston9
+crates/vm/levm/src/hooks/l2_hook.rs @ilitteri @manuelbilbao @avilagaston9
+crates/blockchain/dev @ilitteri @manuelbilbao @avilagaston9
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-11-28
+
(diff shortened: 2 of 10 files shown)


crates/blockchain/vm.rs
```diff
@@ -90,12 +91,16 @@ impl VmDatabase for StoreVmDatabase {
                 .get_canonical_block_hash_sync(block_number)
                 .map_err(|err| EvmError::DB(err.to_string()))?
             {
+                block_hash_cache.insert(block_number, hash);
```
mpaulucci, line 101: this grows infinitely?

fmoletta: This is only used for the `BLOCKHASH` opcode implementation, which checks that the requested block number is less than 256 blocks away from the current block, so it should only grow up to 256

crates/blockchain/vm.rs
```diff
@@ -15,7 +19,7 @@ pub struct StoreVmDatabase {
     // Used to store known block hashes
     // We use this when executing blocks in batches, as we will only add the blocks at the end
     // And may need to access hashes of blocks previously executed in the batch
```
JereSalo: I know the PR is not open yet but we should update this comment, right?

fmoletta: nice catch!

Oppen:
I suggest we try preloading the 256 hashes in an array/vector first. For the most part they could live in memory (they take 8kiB for the canonical chain, with some extra for reorgs and forks) already, and even when they don't, the common case is blocks in the canonical chain, meaning they can be resolved with a single query to the canonical hashes table, and the rest require a bounded (65, I think, for forks starting at a non-finalized block that should be at most 64 blocks long, plus one for the intersection with the canonical chain).
That would avoid the internal management of block hashes, including the mutex, and I think would lead to overall simpler code. Iff it shows to perform poorly, I would consider something more complex than that.

fmoletta:
> I suggest we try preloading the 256 hashes in an array/vector first. For the most part they could live in memory (they take 8kiB for the canonical chain, with some extra for reorgs and forks) already, and even when they don't, the common case is blocks in the canonical chain, meaning they can be resolved with a single query to the canonical hashes table, and the rest require a bounded (65, I think, for forks starting at a non-finalized block that should be at most 64 blocks long, plus one for the intersection with the canonical chain). That would avoid the internal management of block hashes, including the mutex, and I think would lead to overall simpler code. Iff it shows to perform poorly, I would consider something more complex than that.

I implemented this on #5443 but the benchmark results are worse than this solution.
The benchmark only requests a single block hash multiple times so it makes sense that preloading all 256 won't be an improvement, but I also don't think it is very common for a contract to request so many different block hashes

Oppen:
> > I suggest we try preloading the 256 hashes in an array/vector first. For the most part they could live in memory (they take 8kiB for the canonical chain, with some extra for reorgs and forks) already, and even when they don't, the common case is blocks in the canonical chain, meaning they can be resolved with a single query to the canonical hashes table, and the rest require a bounded (65, I think, for forks starting at a non-finalized block that should be at most 64 blocks long, plus one for the intersection with the canonical chain). That would avoid the internal management of block hashes, including the mutex, and I think would lead to overall simpler code. Iff it shows to perform poorly, I would consider something more complex than that.
> 
> I implemented this on #5443 but the benchmark results are worse than this solution. The benchmark only requests a single block hash multiple times so it makes sense that preloading all 256 won't be an improvement, but I also don't think it is very common for a contract to request so many different block hashes

I'd argue it doesn't really implement what I said, as it's making a get to the DB for each block, which I'm saying we should avoid. It's also using a hashmap rather than a vector, and hashing is not free either.

PR #5438, opened 2025-11-27
refactor(l1): introduce get_block_headers and bulk fetching in gas_tip_estimator

figtracer:
this PR adds a new `get_block_headers(from, to)` method to the storage layer for bulk fetching block headers, mirroring the existing `get_block_bodies(from, to)` method.  the gas tip estimator now uses both bulk methods instead of fetching headers and bodies one-by-one.


figtracer:
considering adding a method that combines both `get_block_headers`and `get_block_bodies` but i think this is good

figtracer:
@MegaRedHand for sure! #5444

crates/networking/rpc/eth/gas_tip_estimator.rs
```diff
@@ -71,12 +71,19 @@ impl GasTipEstimator {
...
+            .get_block_headers(block_range_lower_bound, latest_block_number)
+            .await?;
+
+        for (idx, (block_body, header)) in bodies.into_iter().zip(headers.into_iter()).enumerate() {
+            let block_num = block_range_lower_bound + idx as u64;
+
+            let Some(block_body) = block_body else {
                 error!(
                     "Block body for block number {block_num} is missing but is below the latest known block!"
                 );
```
ElFantasma, line 89: The existing code on `main` does `.ok().flatten()` on `get_block_header`, silently swallowing any storage error. This bulk approach preserves that silent-error behavior (missing header → no base_fee → all tips counted as-is). Fine for now, but worth noting: if a header is genuinely missing while the body exists at the same block number, that's a storage inconsistency that arguably deserves a warning.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1102,6 +1102,55 @@ impl StoreEngine for Store {
...
+            })
+            .await?;
+
+        let mut i = 0;
+
+        // Fill in with None for missing headers
+        let headers = hashes
+            .into_iter()
+            .map(|opt| {
+                opt.and_then(|_| {
```
ElFantasma, line 1140: This two-phase bulk read (numbers → hashes via `flat_map`, then hashes → headers, then zip-back with a mutable index `i`) is fragile — it relies on `read_bulk_async` returning results in the same order as `hash_keys`. More importantly, **`read_bulk_async` no longer exists on `main`**. The storage layer was refactored to use `StorageBackend` / `StorageReadView`. See `get_block_bodies` at `store.rs:393` on `main` for the current pattern: simple loop over `begin_read().get()` inside `spawn_blocking`.

PR #5443, opened 2025-11-27
feat(l1): preload previous 256 block hashes when creating vm database

fmoletta:
**Motivation**
Addressing #5344 by using the existing `block_hash_cache` to preload the previous 256 block hashes so we can skip block hash lookup during execution
Alternative to #5434 as proposed by review comment https://github.com/lambdaclass/ethrex/pull/5434#issuecomment-3586227287

**Description**
* Preload previous 256 block hashes to the block hash cache upon VM database creation
* Rely on block hash cache when executing BLOCKHASH opcode instead of performing lookups

Benchmark Results: worse results than #5434 for BlockHash EEST gas benchmark: 
1080.92 Mgas/s PR #5434
802.74 Mgas/s This PR
32.18 Gas/s Main (7b4a0bab)

### DIFF
--- a/crates/vm/levm/bench/revm_comparison/src/levm_bench.rs
+++ b/crates/vm/levm/bench/revm_comparison/src/levm_bench.rs
@@ -57,7 +57,7 @@ fn init_db(bytecode: Bytes) -> GeneralizedDatabase {
         state_root: *EMPTY_TRIE_HASH,
         ..Default::default()
     };
-    let store: DynVmDatabase = Box::new(StoreVmDatabase::new(in_memory_db, header));
+    let store: DynVmDatabase = Box::new(StoreVmDatabase::new(in_memory_db, header).unwrap());
 
     let mut cache = FxHashMap::default();
     cache.insert(


crates/storage/store.rs
```diff
@@ -804,6 +804,14 @@ impl Store {
         self.engine.get_canonical_block_hash(block_number).await
     }
 
+    pub fn get_canonical_block_hashes(
+        &self,
+        newest: BlockNumber,
+        oldest: BlockNumber,
+    ) -> Result<Vec<BlockHash>, StoreError> {
+        self.engine.get_canonical_block_hashes(newest, oldest)
```
Oppen, line 812: This should be reading from memory most of the time. Only on startup it should need to go to the DB to populate its in-memory cache.

crates/storage/store_db/rocksdb.rs
```diff
@@ -1529,6 +1529,31 @@ impl StoreEngine for Store {
...
+            ))
+        })?;
+
+        let mut results = Vec::with_capacity(newest.saturating_sub(oldest) as usize);
+        for number in (oldest..newest).rev() {
+            match db.get_cf(&cf, number.to_le_bytes())? {
+                Some(res) => results.push(BlockHashRLP::from_bytes(res).to()?),
+                None => break,
+            }
+        }
```
Oppen, line 1552: This should fetch the whole range in batch:
```suggestion
        let keys = (oldest..newest).map(BlockNumber::to_le_bytes());
        let hashes = db.batched_multiget_cf(&cf, keys, true);
        hashes.into_iter().map(|h| h.ok().flatten().or_else(|| StoreError::Custom("missing value"))).collect()
```

PR #5445, opened 2025-11-27
fix(l1): avoid blocking during fullsync

iovoid:
**Motivation**

In fullsync, when in the final (head found) stage, we execute blocks individually. However, because we execute (sync) inside an async function, this blocks the runtime.

**Description**

Uses spawn_blocking to avoid blocking the runtime.

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -31,7 +31,6 @@ use ethrex_trie::trie_sorted::TrieGenerationError;
 use ethrex_trie::{Trie, TrieError};
 use rayon::iter::{ParallelBridge, ParallelIterator};
 use std::collections::{BTreeMap, BTreeSet, HashSet};
-use std::fmt::format;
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/sync.rs
```diff
@@ -31,6 +31,7 @@ use ethrex_trie::trie_sorted::TrieGenerationError;
 use ethrex_trie::{Trie, TrieError};
 use rayon::iter::{ParallelBridge, ParallelIterator};
 use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::fmt::format;
```
MegaRedHand:
```suggestion
```

iovoid: Fixed in [b3f3857](https://github.com/lambdaclass/ethrex/pull/5445/commits/b3f3857c8972fe437198ea132db66cf98a38785d)

Issue #5423, opened 2025-11-25
Use peer ENR info for RLPx connections


MegaRedHand:
See #5354

> This is a guess of mine, but we don't seem to support all IPv6 peers, since some of them might only set their connection address in their ENR. For example, some peer might use their IPv4 address in the ip field from the discv4 "Ping" message and set their TCP port to 0, but in their ENR they set their ip6 and tcp6 fields to a valid address and port. We're currently discarding these, which might be valid peers.
>
> Looking into the geth codebase, I found that their enode.Node type uses the data from the ENR instead of the data received in the Ping message from discv4:
> https://github.com/ethereum/go-ethereum/blob/5117f77af9339358699b7b9bf9a64cccdc9c6aed/p2p/discover/v4_udp.go#L381
> Called here:
> https://github.com/ethereum/go-ethereum/blob/2d3704c4d8332b0663ae3128e1ba472a99927be0/eth/backend.go#L543
>
> To support this, we need the RLPx protocol to use peer ENRs instead of discv4 node records when making connections.

lakshya-sky:
Hey @MegaRedHand, are you accepting external contributions? If so I would be happy to work on this ASAP.

lakshya-sky:
I did some digging and realized that we need to request ENR using `ENRRequest` before we make RLPx connection. This would also allow us to select peers whose forkid are valid for current network(e.g. we don't connect to other testnets or mainnet if we are on hoodi).

MegaRedHand:
Hey! Things might be changing on P2P discovery in the coming months, so I think it's best if we keep this issue for later.

MegaRedHand:
Done in #5500

PR #5446, opened 2025-11-27
feat(l1): periodically fetch node records from peers.

lakshya-sky:
**Motivation**

* This would also allow us to filter out peers on wrong network or fork.

**Description**

* uses similar flow as lookup
* store node records for future purposes such as filtering nodes on forkid and rlpx connections(#5423)

Towards #5423

### DIFF
--- a/crates/networking/p2p/discv4/server.rs
+++ b/crates/networking/p2p/discv4/server.rs
@@ -486,7 +486,11 @@ impl DiscoveryServer {
         }
 
         self.peer_table
-            .record_enr_response_received(&node_id, enr_response_message.request_hash ,enr_response_message.node_record)
+            .record_enr_response_received(
+                &node_id,
+                enr_response_message.request_hash,
+                enr_response_message.node_record,
+            )
             .await?;
 
         Ok(())


crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -67,6 +73,20 @@ impl Contact {
         self.validation_timestamp = Some(Instant::now());
         self.ping_hash = Some(ping_hash);
     }
+
+    pub fn record_enr_request_sent(&mut self, request_hash: H256) {
+        self.validation_timestamp = Some(Instant::now());
```
MegaRedHand: We do this for PING messages, but we shouldn't do it here. For ENR requests, we need to check the `enr-seq` field of PING/PONG messages and check if it differs from the one we have.

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -620,7 +715,7 @@ impl PeerTableServer {
         if sender_ip != contact.node.ip {
             return OutMessage::IpMismatch;
         }
-        OutMessage::Contact(contact.clone())
+        OutMessage::Contact(Box::new(contact.clone()))
```
MegaRedHand, line 710: Why is this boxed?

crates/networking/p2p/discv4/server.rs
```diff
@@ -263,6 +271,44 @@ impl DiscoveryServer {
         }
     }
 
+    async fn enr_lookup(&mut self) -> Result<(), DiscoveryServerError> {
```
MegaRedHand, line 269: I think it might be better to send an ENR request the moment we receive a PONG from a peer, instead of having an independent flow for these messages.

crates/networking/p2p/discv4/peer_table.rs
lakshya-sky: After adding node record in Contact I was getting `https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant` lint error, which makes sense. And the Contact is unboxed as soon as PeerTable receives it so outside api doesn't change.

crates/networking/p2p/discv4/server.rs
lakshya-sky: I was initially thinking the same but I was not sure. I will fix it.

crates/networking/p2p/discv4/peer_table.rs
lakshya-sky: How do you propose I store seq? should I create a node record with `enr-seq` on receiving pong and check against it?

MegaRedHand: No need to store it. We eventually revalidate the peer with PING-PONG, receiving the `enr-seq`. When that happens, we can compare it with the stored ENR's `seq` field, doing an ENRRequest if stale.
Should be easy to do with this change: https://github.com/lambdaclass/ethrex/pull/5446#discussion_r2572087486

crates/networking/p2p/discv4/server.rs
lakshya-sky: This would result in several tests failing.
image

ElFantasma: I believe that part of the setup of that tests it to do the Ping/Pong exchange before the actual test happens, and now we are pushing an ENRRequest message after the Pong, but before anything else...

ElFantasma: I think reverting to the version where it was sending decoupled ENRRequests in a periodic fashion will be better. I may have some comments there, tough

lakshya-sky: Hey @ElFantasma I'd love to hear them so we can address them now.

crates/networking/p2p/discv4/server.rs
```diff
@@ -41,6 +41,11 @@ const REVALIDATION_INTERVAL: Duration = Duration::from_secs(12 * 60 * 60); // 12
 /// contacts reaches [target_contacts](DiscoverySideCarState::target_contacts).
 pub const INITIAL_LOOKUP_INTERVAL: Duration = Duration::from_millis(100); // 10 per second
 pub const LOOKUP_INTERVAL: Duration = Duration::from_millis(600); // 100 per minute
+/// The initial interval between enr lookups, until the number of peers reaches
+/// [target_peers](DiscoverySideCarState::target_peers), or the number of
+/// contacts reaches [target_contacts](DiscoverySideCarState::target_contacts).
+pub const INITIAL_ENR_LOOKUP_INTERVAL: Duration = Duration::from_millis(100); // 10 per second
+pub const ENR_LOOKUP_INTERVAL: Duration = Duration::from_millis(600); // 100 per minute
```
ElFantasma: I think I prefer to re-use the same intervals as LOOKUP. No need to have a separated set of constants for this.

crates/networking/p2p/discv4/server.rs
```diff
@@ -263,6 +271,44 @@ impl DiscoveryServer {
...
+            }
+
+            self.peer_table
+                .record_enr_request_sent(&contact.node.node_id(), H256::from(enr_request_hash))
+                .await?;
+        }
+        Ok(())
+    }
+
+    async fn get_enr_lookup_interval(&mut self) -> Duration {
```
ElFantasma: I'd call the get_lookup_interval function in here, since we will want the same periodicity. We may want to change it in the future, but that will be a different PR

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -248,6 +270,51 @@ impl PeerTable {
...
+            .cast(CastMessage::RecordEnrRequestSent {
+                node_id: *node_id,
+                request_hash,
+            })
+            .await?;
+        Ok(())
+    }
+
+    /// Record a response received. Check previously saved hash and reset it if it matches
+    pub async fn record_enr_response_received(
```
ElFantasma, line 294: I think this function is not used (and so, neither the `RecordEnrResponseReceived` message) because when we receive a ENRResponse we should record it directly (`set_node_record`), right?

lakshya-sky: makes sense

lakshya-sky: I've instead removed set_node_record and record_enr_response_received would set it directly after matching the hash.

crates/networking/p2p/discv4/server.rs
```diff
@@ -263,6 +266,35 @@ impl DiscoveryServer {
...
+            let expiration: u64 = get_msg_expiration_from_seconds(EXPIRATION_SECONDS);
+            let enr_request = Message::ENRRequest(ENRRequestMessage { expiration });
+
+            let mut buf = Vec::new();
+            enr_request.encode_with_header(&mut buf, &self.signer);
+            let enr_request_hash: [u8; 32] = buf[..32]
+                .try_into()
+                .expect("first 32 bytes are the message hash");
+
+            if self.udp_socket.send_to(&buf, contact.node.udp_addr())
```
ElFantasma, line 280: Now it seems we have two kind of messages that save the hash. We may want to refactor it to have a single `send_and_get_hash` function (or some better naming 😬) to implement this logic only once.
It can be done on a different PR, though

MegaRedHand (approved):

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -45,7 +45,13 @@ pub struct Contact {
     /// None if no ping was sent yet or it was already acknowledged.
     pub ping_hash: Option<H256>,
 
+    /// The hash of the last unacknowledged ENRRequest sent to this contact, or
+    /// None if no request was sent yet or it was already acknowledged.
+    pub enr_request_hash: Option<H256>,
+
     pub n_find_node_sent: u64,
+    /// ENR associated with this contact, if it was provided by the peer.
+    pub record: Option<NodeRecord>,
```
MegaRedHand, line 54: We should set this to `None` if we receive a PING with a different `seq-num`. I opened issue #5488 for that

Issue #5348, opened 2025-11-14
gas benchmarks: modexp edge cases


Arkenan:
We're ok in modexp in general, but some edge cases seem to not be handled as such:

- mod vul nagydani: 5, 3 and 2 square and cube.
- mod 1024 exp 2.

MegaRedHand:
This might improve with v0.7.0 of malachite: https://github.com/mhogrefe/malachite/releases/tag/v0.7.0

MegaRedHand:
Bumping to v0.7.1 seems to improve performance up to ~10% in modexp benchmarks.

MegaRedHand:
Malachite v0.7.1 bumped the minimum Rust version to v1.90, which some prover toolchains don't support yet 😢

We could still bump it to v0.7.0, but v0.7.1 has some bugfixes in the changelog, so more testing is necessary for us to be sure this is OK. There's also the issue of bumping to more recent releases, like v0.8.0. We should look into this once we revisit this issue.

PR #5447, opened 2025-11-27
chore(levm): bump malachite dep 0.6.1 -> 0.7.1

MegaRedHand:
**Motivation**

The library we use for modular exponentation has a recent release which allegedly improves performance.

**Description**

This PR bumps the version of malachite to 0.7.1 from 0.6.1

Related to #5348

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3565,7 +3565,7 @@ dependencies = [
  "k256",
  "lambdaworks-math 0.13.0",
  "lazy_static",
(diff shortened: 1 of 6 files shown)


MegaRedHand:
Seems like malachite v0.7.1 bumped the minimum rust version to v1.90, which some prover toolchains don't support yet 😢 

We can bump to v0.7.0 to merge this now, but v0.7.1 has some bugfixes in the changelog, so more testing is necessary.

Issue #1766, opened 2025-01-21
Download old blocks & receipts when snap syncing


mpaulucci:
We want to download history (blocks and receipts) previous to the pivot block when doing snap sync.

From 1 May 2025, only the post-merge blocks and receipts will be downloaded.

This is a low priority ticket. We will eventually need it to be compliant with the spec. But I would implement it only if/when we needed. If it's after May 1, we should just download post-merge blocks.

Blocks and receipts are not necessary to follow the chain, but in general, nodes download them.

mpaulucci:
This was implemented in the new version of snap sync IIRC

PR #5452, opened 2025-11-27
refactor(l1,l2,levm): remove dead code

iovoid:
**Motivation**

Dead code in pub functions is not found by clippy, so they can easily accumulate.

**Description**

Using [a tool](https://github.com/cpg314/cargo-workspace-unused-pub), several functions that are never used are found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@
 ### 2025-10-17
 
 - Replaces incremental iteration with a one-time precompute method that scans the entire bytecode, building a `BitVec<u8, Msb0>` where bits mark valid `JUMPDEST` positions, skipping `PUSH1..PUSH32` data bytes.
+- Updates `is_blacklisted` to O(1) bit lookup.
 
 ### 2025-10-14
 


pablodeymo (approved):

crates/networking/rpc/debug/execution_witness.rs
```diff
@@ -65,85 +55,6 @@ impl TryFrom<ExecutionWitness> for RpcExecutionWitness {
     }
 }
 
-// TODO: Ideally this would be a try_from but crate dependencies complicate this matter
-pub fn execution_witness_from_rpc_chain_config(
```
MegaRedHand, line 69: `ethrex-replay` uses this. We can add a comment mentioning that

crates/l2/networking/rpc/clients.rs
JereSalo, line 1: see https://github.com/lambdaclass/ethrex/pull/5433/files

ilitteri (changes requested):
We need to test these changes in our tooling https://github.com/lambdaclass/ethrex-replay and https://github.com/lambdaclass/rex

crates/l2/sdk/src/sdk.rs
ilitteri, line 1: This is an SDK used by external users; this functionality might not be used in ethrex codebase, but it needs to be exposed to users.

crates/l2/networking/rpc/clients.rs
ilitteri, line 1: The whole RPC client module is used by external users; it is not right to just remove unused functions here.

pablodeymo (changes requested):
we need to review external dependencies

crates/networking/rpc/debug/execution_witness.rs
iovoid: Restored in 3c30aedb2

crates/l2/sdk/src/sdk.rs
iovoid: Restored in 3bcb75ef2.

crates/l2/networking/rpc/clients.rs
iovoid: Restored in 3bcb75ef2.

iovoid:
* rex seems to work after fixing unrelated incompatibilities (compile_contract and get_address_from_secret_key changed their interface), the changes are left in the branch test_remove_dead_pub if you want to reproduce

* ethrex-replay block execution tested

crates/networking/rpc/clients/eth/mod.rs
ilitteri, line 1: These are unused, but the idea of the `EthClient` is to be a wrapper of the RPC API, for users to be able to call any RPC method they want.

This module and the `auth` module should live in the `ethrex-sdk` module, but because of cyclic import reasons.

crates/networking/p2p/sync.rs
```diff
@@ -582,31 +582,6 @@ async fn store_block_bodies(
...
-            }
-            // Check if we need to ask for another batch
-            if block_hashes.is_empty() {
-                break;
-            }
-        }
-    }
-    Ok(())
-}
-
```
fmoletta, line 612: This would be used when we address #1766 , maybe we should leave it and link the issue?

crates/networking/rpc/clients/eth/mod.rs
iovoid: Restored in 588a3bfee.

cmd/ethrex/utils.rs
```diff
@@ -65,13 +64,6 @@ pub fn read_chain_file(chain_rlp_path: &str) -> Vec<Block> {
     decode::chain_file(chain_file).expect("Failed to decode chain rlp file")
 }
 
-pub fn read_block_file(block_file_path: &str) -> Block {
-    let encoded_block = std::fs::read(block_file_path)
-        .unwrap_or_else(|_| panic!("Failed to read block file with path {block_file_path}"));
-    Block::decode(&encoded_block)
-        .unwrap_or_else(|_| panic!("Failed to decode block file {block_file_path}"))
-}
```
Oppen, line 73: I though the import command used this.

crates/networking/p2p/sync.rs
ilitteri: Is #1766 going to be addressed soon? While that issue is not being addressed, I'd prefer to mention this thread in the issue and remove the code for now.

cmd/ethrex/utils.rs
iovoid: This is done by `read_chain_file`, which is almost identical.

CHANGELOG.md
```diff
@@ -91,7 +91,6 @@
 ### 2025-10-17
 
 - Replaces incremental iteration with a one-time precompute method that scans the entire bytecode, building a `BitVec<u8, Msb0>` where bits mark valid `JUMPDEST` positions, skipping `PUSH1..PUSH32` data bytes.
-- Updates `is_blacklisted` to O(1) bit lookup.
```
Oppen, line 94: The changelog shouldn't change. If anything, log that it was recently removed in a new entry.

crates/vm/levm/src/utils.rs
Oppen, line 1: Let's first send this one as a separate PR as this is surely unused.

CHANGELOG.md
iovoid: `is_blacklisted's` case is particular since it was already dead when that change happened, but I see why we might not want to pretend it never happened

Maybe adding a note or a strikethrough?

crates/vm/levm/src/utils.rs
iovoid: Why a separate PR?

CHANGELOG.md
pablodeymo: I agree that changelog shouldn't be removed, but only appends.

iovoid: Restored previous version in 454e76788

PR #5453, opened 2025-11-27
feat(l1): add "eth" entry to local ENR

### LINKED ISSUE #5066: Add "eth" entry to local ENR
The devp2p spec specifies an "eth" ENR entry with the fork ID. We don't currently set it, which could cause our peers to ignore us.

lakshya-sky:
**Motivation**

Geth uses forkId for making connection as outlined in below posts and failing to provide one could mean that newer Geth clients would reject connections.
https://ethereum-magicians.org/t/all-core-devs-testing-acdt-46-july-28-2025/24889/6 and https://github.com/ethereum/go-ethereum/pull/31944. 

**Description**

Closes #5066


crates/networking/p2p/discv4/server.rs
```diff
@@ -97,7 +99,8 @@ impl DiscoveryServer {
     ) -> Result<(), DiscoveryServerError> {
         info!("Starting Discovery Server");
 
-        let local_node_record = NodeRecord::from_node(&local_node, 1, &signer)
+        let fork_id = storage.get_fork_id().await.ok();
```
MegaRedHand: This won't work correctly for nodes that are syncing. At startup they'll use the fork-id of the genesis block (latest block is 0), and will never update this once synced.

I think we need to have some periodic message that updates this once per X mins or something like that.

crates/networking/p2p/types.rs
```diff
@@ -347,11 +347,26 @@ impl NodeRecord {
...
+    ) -> Result<Self, NodeError> {
         let mut record = NodeRecord {
             seq,
             ..Default::default()
         };
+        if let Some(fork_id) = fork_id {
+            // Without the Vec wrapper, RLP encoding fork_id directly would produce:
+            // [forkHash, forkNext]
+            // But the spec requires nested lists:
+            // [[forkHash, forkNext]]
```
MegaRedHand: Great comment

Wait. This means we were incorrectly encoding this entry?

crates/networking/p2p/types.rs
```diff
@@ -347,11 +347,26 @@ impl NodeRecord {
         Ok(result)
     }
 
-    pub fn from_node(node: &Node, seq: u64, signer: &SecretKey) -> Result<Self, NodeError> {
+    pub fn from_node(
+        node: &Node,
+        seq: u64,
+        signer: &SecretKey,
+        fork_id: Option<ForkId>,
+    ) -> Result<Self, NodeError> {
```
MegaRedHand: We should keep this function like before, and add another one for the case where `fork_id` is `Some`. That way changes are kept to this file, and API is clearer. We can reduce code duplication by moving common code to a helper in this module.

MegaRedHand (commented):
Thanks for your contribution! Also, great catch with the fork-id encoding!

crates/networking/p2p/discv4/server.rs
lakshya-sky: You're right! Geth uses https://github.com/ethereum/go-ethereum/blob/5e6f7374deaae104e243703f5aea25c68f4d6df6/eth/protocols/eth/discovery.go#L41 which would update ENR record updated with each new block head. Let me know your opinion.

lakshya-sky: I think we should tackle this in a different PR.

crates/networking/p2p/types.rs
lakshya-sky: pretty much since we never included `eth` to local node record so we never had a way to know if our encoding was right.

crates/networking/p2p/discv4/server.rs
MegaRedHand: Yeah. That's a good idea.

ElFantasma (approved):
LGTM

crates/networking/p2p/discv4/server.rs
MegaRedHand: Opened https://github.com/lambdaclass/ethrex/issues/5493 for this

PR #5457, opened 2025-11-28
perf(l1): avoid heap alloc in tx compute_hash

edg-l:
Added HashBuffer, this is a buffer that implements BufMut and feeds data to a keccak256 hasher, this way we avoid a temporary vec when computing hash of transactions.

A quick bench shows it goes from  4.70s to 4.28s in 10000000 iterations

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,3 +21,5 @@ hive/
 ethereum-package/
 tooling/ef_tests/blockchain/vectors
 tooling/ef_tests/state/vectors
+dev_ethrex_l1/
+dev_ethrex_l2/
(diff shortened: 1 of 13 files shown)


crates/common/utils.rs
```diff
@@ -79,6 +80,84 @@ pub fn truncate_array<const N: usize, const M: usize>(data: [u8; N]) -> [u8; M]
...
+            let take = cnt.min(CHUNK.len());
+            if val == 0 {
+                self.hasher.update(&CHUNK[..take]);
+            } else {
+                let mut tmp = CHUNK;
+                tmp[..take].fill(val);
+                self.hasher.update(&tmp[..take]);
+            }
+            cnt -= take;
+        }
```
Oppen:
```suggestion
        let chunk = [val; 64];
        while cnt >= 64 {
            self.hasher.update(&chunk);
            cnt -= 64;
        }
        self.hasher.update(&chunk[..cnt]);
```

PR #5460, opened 2025-11-28
feat(l1,l2): allow eip 7702 on generic transactions

tomip01:
**Motivation**

We want to be able to send EIP 7702 transactions through `GenericTransaction`

**Description**

- Add `authorization_list` to `Overrides`
- Implement `TryFrom` for `GenericTransaction` -> `EIP7702Transaction`

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2454,10 +2454,8 @@ mod serde_impl {
         InvalidTxType(TxType),
         #[error("Blob bundle error: {0}")]
         BlobBundleError(#[from] BlobsBundleError),
-        #[error("Missing fee token address")]
-        MissingFeeToken,
-        #[error("Missing To field")]
-        MissingToField,
+        #[error("Missing field: {0}")]
+        MissingField(String),
     }
 
     /// Unsigned Transaction struct generic to all types which may not contain all required transaction fields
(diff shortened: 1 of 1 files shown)


ilitteri (approved):

crates/common/types/transaction.rs
```diff
@@ -2456,6 +2456,8 @@ mod serde_impl {
         BlobBundleError(#[from] BlobsBundleError),
         #[error("Missing fee token address")]
         MissingFeeToken,
+        #[error("Missing to destination")]
```
pablodeymo:
```suggestion
        #[error("Missing To field")]
```

crates/common/types/transaction.rs
```diff
@@ -2456,6 +2456,8 @@ mod serde_impl {
         BlobBundleError(#[from] BlobsBundleError),
         #[error("Missing fee token address")]
         MissingFeeToken,
+        #[error("Missing To field")]
+        MissingToField,
```
ManuelBilbao: Would it be better to have a `MissingField(String)` error specifying wich field is missing in the string param?

tomip01: Agree! Done here [b9978b5](https://github.com/lambdaclass/ethrex/pull/5460/commits/b9978b5267f41ee6f86aecc7d288394a47692c30)

PR #5462, opened 2025-11-28
Stacked cache test

ElFantasma:
Replacing bloom filter with a "stacked layers" `HashMap` to improve `get` times from `TrieLayerCache`.
Only last layer is cached, In case a request for a previous layer is made, we transverse the layers. The idea is that 99% of the ´get´ requests for the cache are for current `state_root`, that is, for last layer.

Also, in case the block's parent is not last layer's `state_root`, we rebuild the stacked cache from the raw layers. This is costly, but has very little occurrence

Benchmarks did not show improvement in performance, but this may be relevant to:
- Remove the dependancy on bloom filter library
- Eventually this may become a bottleneck if we improve DB access enough.

### DIFF
--- a/.github/workflows/main_prover.yaml
+++ b/.github/workflows/main_prover.yaml
@@ -116,10 +116,12 @@ jobs:
         run: |
           cd crates/l2
           make init-prover-sp1 GPU=true &
-          docker logs --follow ethrex_l2 & DOCKER_LOGS_PID=$!
+          docker logs --follow ethrex_l2 & DOCKER_LOGS_L2_PID=$!
+          docker logs --follow ethrex_l1 & DOCKER_LOGS_L1_PID=$!
           PROPOSER_COINBASE_ADDRESS=0x0007a881CD95B1484fca47615B64803dad620C8d cargo test l2 --release -- --nocapture --test-threads=1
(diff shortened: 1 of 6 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -1217,6 +1217,24 @@ impl Blockchain {
                 "".to_string()
             };
             info!("{}{}", base_log, extra_log);
+            if as_gigas > 0.0 {
```
ElFantasma, line 1220: I added this log line to have an easy way to copy/paste metrics into a csv or excel sheet to calculate averages.

crates/storage/store_db/rocksdb.rs
```diff
@@ -782,37 +784,54 @@ impl Store {
         account_updates: Vec<(Nibbles, Vec<u8>)>,
         storage_updates: StorageUpdates,
     ) -> Result<(), StoreError> {
+        let start = Instant::now();
```
ElFantasma, line 787: Also added lot of timings for different stages of the trie updates process

ElFantasma:
Closing it as benchmarks didn't show an improvement. 

May return to this if eventually we find a bottleneck in layering

PR #5466, opened 2025-11-29
perf(l1,l2):  Optimize mulmod for zkVMs

JereSalo:
**Motivation**

- Try to see if we can optimize mulmod becuase we've seen mainnet blocks in which it's an expensive operation

**Description**

- Optimize mulmod for SP1. This is in draft and has experimental changes but using a SP1 patched function (not in a pretty manner) yielded 57x in Mulmod in Block 23890684, which is great.

Closes #issue_number


ilitteri:
This was a PoC and can be discarded

PR #5468, opened 2025-12-01
fix(l1,l2): forbid hash override on empty-path insert in branch/leaf

phrwlk:
Replace unresolved todo!() with explicit verification errors when inserting an external hash at the exact target path. In 
BranchNode::insert() (empty path) and the analogous case in LeafNode::insert() (attempt to set branch value from a hash), return TrieError::Verify instead of allowing override or panicking. This aligns with existing safeguards that allow hashes only in empty child slots and prevents replacing existing nodes/values at the target path.

### DIFF
--- a/crates/common/trie/node/branch.rs
+++ b/crates/common/trie/node/branch.rs
@@ -137,7 +137,9 @@ impl BranchNode {
             self.update(value);
         } else {
             // Value in branches don't happen in our use-case.
-            todo!("handle override case (error?)")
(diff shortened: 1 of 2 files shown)


MegaRedHand:
I believe handling the case for non-zero values here is a mistake, and we should instead hardcode at the type level that values are always empty.

I'll be closing this PR, but feel free to open a PR tackling the removal of values in branches (just opened #5772).

PR #5469, opened 2025-12-01
chore(l1): add trie hash bench

edg-l:
**Motivation**

We don't have a bench for how fast a trie is hashed

**Description**

- Adds a trie hashing bench, it uses hash_no_commit to not include db overhead.
- Improved the old benchmarks, made the rng deterministic
- Also updated some related deps


crates/common/trie/benches/trie_bench.rs
```diff
@@ -16,89 +19,163 @@ use ethrex_trie::Trie as EthrexTrie;
...
+            || {
+                let mut trie = EthrexTrie::new(Box::new(EthrexMemDB::new_empty()));
+                for i in 0..keys_1k.len() {
+                    trie.insert(keys_1k[i].clone(), values_1k[i].clone())
+                        .unwrap();
+                }
+                black_box(trie)
+            },
+            |trie| black_box(trie.hash_no_commit()),
+            criterion::BatchSize::LargeInput,
```
ElFantasma, line 115: `hash_no_commit()` memoizes all internal node hashes via `OnceLock` (`NodeRef::compute_hash_no_alloc` uses `hash.get_or_init(…)`). After the first call on a trie instance, the entire tree is hashed and cached — subsequent calls just return the root hash in O(1).

With `iter_batched_ref` + `LargeInput`, criterion runs `ceil(iters/1000)` iterations per batch sharing the same `&mut trie`. Only the first iteration per batch computes the real hash; the rest measure a cached `OnceLock::get()`. This biases the measurement downward.

Two options to fix:
1. Use `iter_batched` (takes ownership, drops after each iteration) instead of `iter_batched_ref` — this forces a fresh trie per iteration
2. Use `BatchSize::PerIteration` which runs setup before every single iteration

Option 1 is cleanest since it matches the semantics (each hash computation is independent). Same applies to the 10k and 1M benchmarks below.

ElFantasma (changes requested):

PR #5471, opened 2025-12-01
style(l1,l2,levm): remove unnecessarily mutable borrows

iovoid:
**Motivation**

Currently we have several unnecessary mutable references. Aside from misleading about what the function does, they could prevent spotting removable clones.

**Description**

Uses the `clippy::needless_pass_by_ref_mut` experimental lint to find these cases.


iovoid:
Do you think the lint should be left enabled?

PR #5473, opened 2025-12-01
refactor(l1): remove dependency of storage merkelization on the current account state

iovoid:
**Motivation**

In preparation for paralellization of storage merkelization, we want to remove the data dependency of storages on the current account state.

**Description**

Reads the root from the database when opening the trie.


Arkenan:
Now obsolete due to https://github.com/lambdaclass/ethrex/pull/5497

PR #5476, opened 2025-12-01
perf(l1,l2): add optional profiler

Oppen:
We currently use patches to run with pprof for good quality profiles. 
This structures the patches so it's easier to use and zero-cost if not using it,
so it becomes feasible to merge and enable as needed.


cmd/ethrex/initializers.rs
```diff
@@ -6,8 +6,8 @@ use crate::{
     },
 };
 use ethrex_blockchain::{Blockchain, BlockchainOptions, BlockchainType};
-use ethrex_common::fd_limit::raise_fd_limit;
 use ethrex_common::types::Genesis;
+use ethrex_common::{fd_limit::raise_fd_limit, utils::ProfilingGuard};
```
pablodeymo, line 10:
```suggestion
use ethrex_common::{fd_limit::raise_fd_limit, types::Genesis, utils::ProfilingGuard};
```

Oppen:
The current approach suffers from too few samples per block, which make the profile meaningless. It also seems like no sampling is done beyond 10kHz.

PR #5477, opened 2025-12-01
feat(l2): shared bridge forced inclusion

avilagaston9:
**Motivation**

In a shared bridge environment, we need a way to enforce L2s to include messages from other L2s.
We take the same approach as with deposits:
- After some time, a message is considered expired if it has not yet been included in a batch.
- When this happens, the sequencer is prevented from including non-privileged transactions until all expired messages are included.

**Description**

- Adds the forced inclusion mechanism to the documentation.
- Modifies `l1_committer` to commit the `l2_out_message_hashes` and the rolling hash of received processed messages in the batch.
- Modifies the `Batch` struct:
    - Adds the `l2_in_message_rolling_hashes` field.
    - Removes `l2_message_hashes`, since they are now included in the `BalanceDiff` struct.
- Removes all `l2_messages_merkle_root` related logic. Since we now send all message hashes, the Merkle root is no longer needed.
- Renames `privileged transactions | deposits` to `l1_in_messages` throughout the codebase. A privileged transaction represents a message from another chain (L1 or L2). `deposits` was also being used to refer to L1 inbound messages, but messages can be more than simple deposits.
- Updates the prover to output `L2RollingHashes`.
- Restores blobs due to changes in the L2 genesis.

### DIFF
--- a/crates/l2/networking/rpc/clients.rs
+++ b/crates/l2/networking/rpc/clients.rs
@@ -7,6 +7,7 @@ use ethrex_common::H256;
 use ethrex_common::U256;
 use ethrex_common::types::{AuthorizationList, AuthorizationTupleEntry};
 use ethrex_l2_common::messages::L1MessageProof;
+use ethrex_rpc::clients::eth::errors::GetBatchByBlockNumberError;
 use ethrex_rpc::clients::eth::errors::GetL1BlobBaseFeeRequestError;
 use ethrex_rpc::clients::eth::errors::GetL1FeeVaultAddressError;
 use ethrex_rpc::clients::eth::errors::GetOperatorFeeError;
@@ -51,15 +52,15 @@ pub async fn get_batch_by_block(
     client: &EthClient,
     block: BlockIdentifier,
 ) -> Result<Option<RpcBatch>, EthClientError> {
-    let params = Some(vec![block.into(), json!(true)]);
+    let params = Some(vec![block.into()]);
     let request = RpcRequest::new("ethrex_getBatchByBlock", params);
 
     match client.send_request(request).await? {
         RpcResponse::Success(result) => serde_json::from_value(result.result)
-            .map_err(GetBatchByNumberError::SerdeJSONError)
+            .map_err(GetBatchByBlockNumberError::SerdeJSONError)
             .map_err(EthClientError::from),
         RpcResponse::Error(error_response) => {
-            Err(GetBatchByNumberError::RPCError(error_response.error.message).into())
+            Err(GetBatchByBlockNumberError::RPCError(error_response.error.message).into())
         }
     }
 }
--- a/crates/networking/rpc/clients/eth/errors.rs
+++ b/crates/networking/rpc/clients/eth/errors.rs
@@ -87,6 +87,8 @@ pub enum EthClientError {
     GetL1BlobBaseFeeError(#[from] GetL1BlobBaseFeeRequestError),
     #[error("ethrex_sendTransaction request error: {0}")]
     SendEthrexTransactionError(#[from] SendEthrexTransactionError),
+    #[error("ethrex_getBatchByBlock request error: {0}")]
+    GetBatchByBlockNumberError(#[from] GetBatchByBlockNumberError),
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -397,4 +399,12 @@ pub enum SendEthrexTransactionError {
     RPCError(String),
     #[error("{0}")]
     ParseHashError(String),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum GetBatchByBlockNumberError {
+    #[error("{0}")]
+    SerdeJSONError(#[from] serde_json::Error),
+    #[error("{0}")]
+    RPCError(String),
 }


cmd/ethrex/cli.rs
```diff
@@ -112,6 +112,7 @@ pub struct Options {
         long = "log.level",
         default_value_t = Level::INFO,
         value_name = "LOG_LEVEL",
+        env = "ETHREX_LOG_LEVEL",
```
ManuelBilbao, line 120: `RUST_LOG` does the job too

crates/common/types/transaction.rs
```diff
@@ -1523,6 +1523,7 @@ impl PrivilegedL2Transaction {
 
         Some(crate::utils::keccak(
             [
+                U256::from(self.chain_id).to_big_endian().as_ref(),
```
ManuelBilbao, line 1526: The comment above the function should be updated with the new param

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -105,12 +105,14 @@ contract CommonBridge is
...
-    /// that the messages were published on L1, and that that batch was committed.
-    mapping(uint256 => bytes32) public l2MessagesMerkleRoots;
+    /// @notice Chain ID of the network
+    uint256 public CHAIN_ID;
+
+    /// @notice Mapping of chain ID to array of pending message hashes received from that chain.
+    mapping(uint256 => bytes32[]) public pendingMessagesHashesPerChain;
+
+    /// @notice Mapping of chain ID to index of the first unprocessed message hash in the array.
+    mapping(uint256 => uint256) public pendingMessagesIndexPerChain;
```
ManuelBilbao: For clarity
```suggestion
    mapping(uint256 chainId => bytes32[] hashes) public pendingMessagesHashesPerChain;

    /// @notice Mapping of chain ID to index of the first unprocessed message hash in the array.
    mapping(uint256 chainId => uint256 index) public pendingMessagesIndexPerChain;
```

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -164,6 +169,24 @@ contract CommonBridge is
...
+        uint256 pendingMessageIndex = pendingMessagesIndexPerChain[chainId];
+        bytes32[] memory pendingMessagesHashes = pendingMessagesHashesPerChain[
+            chainId
+        ];
+        bytes32[] memory buffer = new bytes32[](
+            pendingL2MessagesLength(chainId)
+        );
+        for (uint256 i = 0; i < pendingL2MessagesLength(chainId); i++) {
+            buffer[i] = pendingMessagesHashes[i + pendingMessageIndex];
+        }
```
ManuelBilbao, line 189: Not sure if the compiler optimize it, but i'd go from `i = pendingMessageIndex` until `i < pendingMessagesHashes.length` to avoid a sum on every iteration.

ManuelBilbao: Also, is it necessary to copy the hashes in memory (`pendingMessagesHashes`). Can't we just do something like:
```solidity
for (uint256 i = pendingMessagesIndex; i < pendingMessagesHashesPerChain[chainId].length; i++) {
    buffer[i] = pendingMessagesHashesPerChain[chainId][i];
}
```

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -304,6 +328,35 @@ contract CommonBridge is
...
+            "CommonBridge: number is greater than the length of pendingL2Messages (get)"
+        );
+
+        bytes memory hashes;
+        bytes32[] memory pendingMessagesHashes = pendingMessagesHashesPerChain[
+            chainId
+        ];
+        uint256 pendingMessageIndex = pendingMessagesIndexPerChain[chainId];
+
+        for (uint i = 0; i < number; i++) {
```
ManuelBilbao, line 352: ditto

crates/l2/contracts/src/l1/CommonBridge.sol
```diff
@@ -316,14 +369,58 @@ contract CommonBridge is
...
+            if (chainId == CHAIN_ID) {
+                continue;
+            }
+            uint256 pendingMessageIndex = pendingMessagesIndexPerChain[chainId];
+            bytes32[]
+                memory pendingMessagesHashes = pendingMessagesHashesPerChain[
+                    chainId
+                ];
+            if (pendingMessageIndex < pendingMessagesHashes.length) {
+                if (
```
ManuelBilbao: Nit. This can be a `&&`

crates/l2/contracts/src/l1/Router.sol
```diff
@@ -50,38 +53,32 @@ contract Router is
...
-        uint256 l2MessageBatchNumber,
-        bytes32 l2MessageLeaf,
-        bytes32[] calldata l2MessageProof
-    ) external view returns (bool) {
-        address bridge = bridges[chainId];
-        if (bridge == address(0)) {
-            revert ChainNotRegistered(chainId);
+        bytes32[] calldata message_hashes
+    ) public payable override {
+        uint256 senderChainId = registeredAddresses[msg.sender];
```
ManuelBilbao, line 69: Instead of having the `chainId <=> address` duplicated, maybe this function can receive the source chain ID and check `bridges[srcChainID] == msg.sender`

crates/l2/networking/rpc/l2/batch.rs
```diff
@@ -106,6 +107,65 @@ impl RpcHandler for GetBatchByBatchNumberRequest {
...
+pub struct GetBatchByBatchBlockNumberRequest {
+    pub block: BlockIdentifier,
+}
+
+impl RpcHandler for GetBatchByBatchBlockNumberRequest {
+    fn parse(params: &Option<Vec<Value>>) -> Result<GetBatchByBatchBlockNumberRequest, RpcErr> {
+        let params = params.as_ref().ok_or(ethrex_rpc::RpcErr::BadParams(
+            "No params provided".to_owned(),
+        ))?;
+        if params.len() != 2 {
```
ManuelBilbao: What's the second param? It's not parsed nor saved

crates/l2/sdk/contract_utils/src/compile.rs
```diff
@@ -39,6 +39,10 @@ pub fn compile_contract(
             .to_str()
             .ok_or(ContractCompilationError::FailedToGetStringFromPath)?,
     )
+    // Enable optimizer to keep bytecode size within limits.
+    .arg("--optimize")
+    .arg("--optimize-runs")
+    .arg("50")
```
ManuelBilbao: I think this should be max (999...)

crates/l2/sequencer/block_producer/payload_builder.rs
```diff
@@ -245,9 +243,8 @@ pub async fn fill_transactions(
             let id = head_tx.nonce();
             privileged_nonces.insert(privileged_tx.chain_id, Some(id));
 
-            if privileged_tx.chain_id == store.chain_config.chain_id {
-                privileged_tx_count += 1;
-            }
+            // Increment privileged tx count
```
ManuelBilbao: Think this comment isn't useful

crates/l2/sdk/contract_utils/src/compile.rs
ilitteri: This is not a problem, https://github.com/lambdaclass/ethrex/pull/5544 already bumps it to 999999

crates/l2/contracts/src/l1/CommonBridge.sol
ilitteri: If you do `i = pendingMessagesIndex `, then shouldn't you also do `buffer[i - pendingMessagesIndex]`?

crates/l2/contracts/src/l1/CommonBridge.sol
ilitteri: This one is ok.

crates/l2/common/src/messages.rs
JereSalo, line 1: Here when encoding an `L2Message` we are not encoding the `dest_chain_id`. Can this be a problem given that when getting the L2 Message hash we do `keccak(msg.encode())`? I mean, 2 equal messages to different L2s could possibly have the same hash. Is that a problem or is it intentional?

crates/l2/contracts/src/l1/Router.sol
ilitteri: I tend to believe that retrieving the src chain ID from the mapping via `msg.sender` is less error-prone than allowing it to be passed as an arg.

crates/l2/sequencer/block_producer/payload_builder.rs
ilitteri: [6bd103c](https://github.com/lambdaclass/ethrex/pull/5477/commits/6bd103c73b638a8ce80c8b9ea3a9494a085107b4)

crates/l2/networking/rpc/l2/batch.rs
ilitteri: [6bd103c](https://github.com/lambdaclass/ethrex/pull/5477/commits/6bd103c73b638a8ce80c8b9ea3a9494a085107b4)

crates/l2/contracts/src/l1/CommonBridge.sol
ilitteri: [6bd103c](https://github.com/lambdaclass/ethrex/pull/5477/commits/6bd103c73b638a8ce80c8b9ea3a9494a085107b4)

crates/l2/contracts/src/l1/CommonBridge.sol
ilitteri: [6bd103c](https://github.com/lambdaclass/ethrex/pull/5477/commits/6bd103c73b638a8ce80c8b9ea3a9494a085107b4)

crates/l2/contracts/src/l1/CommonBridge.sol
ilitteri: [6bd103c](https://github.com/lambdaclass/ethrex/pull/5477/commits/6bd103c73b638a8ce80c8b9ea3a9494a085107b4)

crates/common/types/transaction.rs
ilitteri: [6bd103c](https://github.com/lambdaclass/ethrex/pull/5477/commits/6bd103c73b638a8ce80c8b9ea3a9494a085107b4)

crates/l2/contracts/src/l1/Router.sol
```diff
@@ -21,6 +21,8 @@ contract Router is
 
     uint256[] public registeredChainIds;
 
+    mapping(address => uint256) public registeredAddresses;
```
ManuelBilbao: For clarity
```suggestion
    mapping(address bridge => uint256 chainId) public registeredAddresses;
```

ilitteri: [Refactor](https://github.com/lambdaclass/ethrex/pull/5477/commits/63798beb60d42c7dac8e5832bf709f7795949a80)

PR #5478, opened 2025-12-01
feat(l2): add timed prover API

ilitteri:
**Motivation**

To measure execution and proving time correctly. This means removing the client initialization from the equation. In the case of ZisK, we use it's own timing exposed via a JSON file generated after proving.

**Description**

Adds `_timed` versions for execution and proving for all zkVM backends.


crates/l2/prover/src/backend/zisk.rs
```diff
@@ -84,6 +112,30 @@ pub fn prove(
...
+    format: ProofFormat,
+) -> Result<(ProveOutput, std::time::Duration), Box<dyn std::error::Error>> {
+    let proof = prove(input, format)?;
+
+    #[derive(serde::Deserialize)]
+    struct ZisKResult {
+        #[serde(rename = "cycles")]
+        _cycles: u64,
+        #[serde(rename = "id")]
+        _id: String,
```
ManuelBilbao, line 126: Unknown fields are ignored by default. We can remove them if we are not going to use them. If we want them for completeness, it's ok then

PR #5481, opened 2025-12-01
fix(l1): sending ping to each neighbor found

### LINKED ISSUE #5480: Not initiating outbound connections except bootnodes.
Copied from Telegram:

> I might be missing something here but it seems like we never make PING except these scenarios
> * sending PING to bootnodes while starting
> * we try to initiate connection from out side when we receice PING from other peers.
> * when we revalidate peers every 12hours.
> and due to this we don't have large enough pool of peers to send FindNodes request even after few minutes. 
> 
> Geth when sending FindNode makes sure that we have bonded with peer using PING. 
> 
> I let the node run on hoodi without consensus client for 10minutes.
> 
> * only two nodes were validated out of almost 2000
> * 18 pings were sent out of those 16 were to bootnodes.
> * only 8 Pongs were received.
> complete log file https://gist.githubusercontent.com/lakshya-sky/bdeb1bc2658c4bbe844f03da246df0e2/raw/e8dc728c9f0749fffb7252e1be1825a965d38ba7/latest.log

ElFantasma:
**Motivation**
Our `FindNode` messages weren't responded as most of the remote nodes didn't know us. 

**Description**

In order to be a known peer for the remote nodes we need to send a `Ping` message first. So, when we receive Neighbors from peers we send a `Ping` message to each one.

Closes #5480

### DIFF
--- a/crates/l2/storage/src/store_db/sql.rs
+++ b/crates/l2/storage/src/store_db/sql.rs
@@ -29,23 +29,23 @@ impl Debug for SQLStore {
 }
 
 const DB_SCHEMA: [&str; 17] = [
-    "CREATE TABLE blocks (block_number INT PRIMARY KEY, batch INT)",
-    "CREATE TABLE messages (batch INT, idx INT, message_hash BLOB, PRIMARY KEY (batch, idx))",
-    "CREATE TABLE privileged_transactions (batch INT PRIMARY KEY, transactions_hash BLOB)",
-    "CREATE TABLE state_roots (batch INT PRIMARY KEY, state_root BLOB)",
-    "CREATE TABLE blob_bundles (batch INT, idx INT, blob_bundle BLOB, PRIMARY KEY (batch, idx))",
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/discv4/server.rs
```diff
@@ -395,7 +395,7 @@ impl DiscoveryServer {
             // Sending all in one packet would exceed bounds with the nodes only, weighing
             // up to 1424B.
             for chunk in neighbors.chunks(8) {
-                let _ = self.send_neighbors(chunk.to_vec(), &contact.node).await;
+                self.send_neighbors(chunk.to_vec(), &contact.node).await?;
```
MegaRedHand: Is it OK to stop on error here?

ElFantasma: probaby not 🤔

MegaRedHand (approved):

PR #5483, opened 2025-12-01
feat(l1): refactor chain config (#5233)

### LINKED ISSUE #4720: Improve the way we obtain the next and last forks

fmoletta:
**Motivation**
Restores #5233 by reverting #5464 + changes `get_fork` behaviour to default to Paris if the network is pre-merge (This is the change that broke Hive Daily Report tests). Also changes method `fork_activation_time_or_block` to `fork_activation_time`

**Description**
* Restore #5233
*  Change method `fork_activation_time_or_block` to `fork_activation_time` which returns `None` for pre-merge forks so we default to Paris on `get_fork` method if the network is pre-merge

Resolves #4720


crates/l2/sequencer/block_producer/payload_builder.rs
```diff
@@ -177,16 +176,6 @@ pub async fn fill_transactions(
...
 
-        // Check whether the tx is replay-protected
-        if head_tx.tx.protected() && !chain_config.is_eip155_activated(context.block_number()) {
-            // Ignore replay protected tx & all txs from the sender
-            // Pull transaction from the mempool
-            debug!("Ignoring replay-protected transaction: {}", tx_hash);
-            txs.pop();
-            blockchain.remove_transaction_from_pool(&tx_hash)?;
-            continue;
-        }
```
ilitteri, line 199: Why does this refactor have anything to do with removing this piece of code?

ilitteri: Taking into account the refactor, the correct thing to do here is to find another way to replace the `!chain_config.is_eip155_activated(context.block_number())` part of this code.

fmoletta: I think the reasoning is that we don't support running pre-merge blocks so this code should be unreachable

crates/common/types/genesis.rs
```diff
@@ -283,11 +284,11 @@ lazy_static::lazy_static! {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Default, Hash, Clone, Copy, Serialize, Deserialize)]
 pub enum Fork {
```
MegaRedHand, line 285: Maybe we should move the fork type and list to another file. wdyt?

crates/blockchain/payload.rs
```diff
@@ -11,7 +11,9 @@ use ethrex_common::{
     constants::{DEFAULT_OMMERS_HASH, DEFAULT_REQUESTS_HASH, GAS_PER_BLOB, MAX_RLP_BLOCK_SIZE},
     types::{
         AccountUpdate, BlobsBundle, Block, BlockBody, BlockHash, BlockHeader, BlockNumber,
-        ChainConfig, MempoolTransaction, Receipt, Transaction, TxType, Withdrawal, bloom_from_logs,
+        ChainConfig,
+        Fork::*,
```
MegaRedHand, line 15: Neat!

crates/common/types/genesis.rs
```diff
@@ -420,145 +467,42 @@ impl ChainConfig {
...
+    pub fn get_blob_schedule_for_time(&self, block_timestamp: u64) -> Option<ForkBlobSchedule> {
+        if let Some(fork_with_current_blob_schedule) = FORKS.into_iter().rfind(|fork| {
+            self.get_blob_schedule_for_fork(*fork).is_some()
+                && self.is_fork_activated(*fork, block_timestamp)
+        }) {
+            self.get_blob_schedule_for_fork(fork_with_current_blob_schedule)
         } else {
             None
         }
     }
```
JereSalo, line 488: Can this be simpler?
Like can we just do `let fork = self.get_fork(block_timestamp)` and then `self.get_blob_schedule_for_for(fork)` or there may be a problem with that? I'm assuming that every new fork will have a blob schedule set to it.

crates/common/types/genesis.rs
```diff
@@ -420,145 +467,42 @@ impl ChainConfig {
...
-        } else if self.is_osaka_activated(block_timestamp) && self.bpo1_time.is_some() {
-            Some(Fork::BPO1)
-        } else if self.is_prague_activated(block_timestamp) && self.osaka_time.is_some() {
-            Some(Fork::Osaka)
-        } else if self.is_cancun_activated(block_timestamp) && self.prague_time.is_some() {
-            Some(Fork::Prague)
-        } else if self.is_shanghai_activated(block_timestamp) && self.cancun_time.is_some() {
-            Some(Fork::Cancun)
+        let current_fork = self.get_fork(block_timestamp);
+        if (current_fork as usize) < FORKS.len() {
```
JereSalo: I believe this condition will always be true given that the first element starts with zero, so `current_fork` can be at most `FORKS - 1`

JereSalo (commented):
Defaulting to Paris if none of the forks is activated may work if we suppose that there's a previous check that doesn't let anybody initialize with a pre-Merge config but I still find it kinda weird. Doesn't need any change though, just saying. It may be the best thing to do anyway

crates/common/types/genesis.rs
fmoletta: Linking discussion from reverted PR as I made the same comment there: https://github.com/lambdaclass/ethrex/pull/5233#discussion_r2518764961

crates/common/types/genesis.rs
fmoletta: Updated!

crates/l2/sequencer/block_producer/payload_builder.rs
Oppen: Given we already broke main by changing behavior during a chain config refactor, wouldn't it be better to leave this part out and reintroduce it later separately?

fmoletta: This part didn't break main

crates/common/types/genesis.rs
```diff
@@ -283,11 +284,11 @@ lazy_static::lazy_static! {
...
     Frontier = 0,
-    FrontierThawing = 1,
-    Homestead = 2,
-    DaoFork = 3,
-    Tangerine = 4,
-    SpuriousDragon = 5,
+    Homestead = 1,
+    DaoFork = 2,
+    EIP150 = 3,
+    EIP155 = 4,
```
ElFantasma, line 290: The `EIP155` variant maps to `eip155_block` and `EIP158` maps to `eip158_block`, which are separate fields in `ChainConfig`. On Mainnet they always activate at the same block (2675000), but as separate enum variants with `PartialOrd` derived, `EIP155 = 4 < EIP158 = 5`. This means `fork >= EIP158` won't include `EIP155`, which is semantically wrong — EIP-155 (replay protection) was part of Spurious Dragon, same as EIP-158.

In the revm runner (`fork_to_spec_id`), both map to `SpecId::SPURIOUS_DRAGON`, confirming they're the same fork. Consider keeping a single `SpuriousDragon` variant (or at least documenting why they're split and ensuring no comparison relies on their relative ordering).

crates/blockchain/mempool.rs
```diff
@@ -445,7 +440,7 @@ pub fn transaction_intrinsic_gas(
             .checked_add(zero_count * TX_DATA_ZERO_GAS_COST)
             .ok_or(MempoolError::TxGasOverflowError)?;
 
-        if is_contract_creation && config.is_shanghai_activated(header.timestamp) {
+        if is_contract_creation && config.is_fork_activated(Shanghai, header.timestamp) {
```
ElFantasma, line 443: Good simplification — since this is a post-merge-only client, Istanbul is always active and `TX_DATA_NON_ZERO_GAS` (now 16) is the only valid gas cost. The removed pre-Istanbul branch was dead code.

crates/common/types/genesis.rs
```diff
@@ -420,145 +467,38 @@ impl ChainConfig {
...
-        } else if self.is_cancun_activated(block_timestamp) {
-            Some(self.blob_schedule.cancun)
+        FORKS
+            .into_iter()
+            .rfind(|fork| self.is_fork_activated(*fork, block_timestamp))
+            .unwrap_or(Paris)
+    }
+
+    pub fn get_blob_schedule_for_time(&self, block_timestamp: u64) -> Option<ForkBlobSchedule> {
+        if let Some(fork_with_current_blob_schedule) = FORKS.into_iter().rfind(|fork| {
```
ElFantasma, line 480: This `rfind` over `FORKS` is clean but has a subtle behavior difference from the old `get_fork_blob_schedule`: if a fork is activated but has no explicit blob schedule (e.g., BPO5 activated with `blob_schedule.bpo5 = None`), the old code returned `Some(Default::default())` (a zero-valued schedule), while this falls back to the next lower fork's schedule. The new behavior is arguably more correct (inherits the last configured schedule), but it silently changes semantics — worth a test and/or comment explaining the fallback intent.

crates/common/types/genesis.rs
```diff
@@ -420,145 +467,38 @@ impl ChainConfig {
         output
     }
 
+    /// Obtain the latest active fork given a block's timestamp
+    /// If no time-based (aka Post-Merge) fork is active, it will default to Paris
     pub fn get_fork(&self, block_timestamp: u64) -> Fork {
```
ElFantasma, line 472: nit: The `FORKS` array includes pre-merge forks (Frontier through GrayGlacier) which will never satisfy `is_fork_activated` (returns `false` since `fork_activation_time` returns `None` for them). The iteration still works correctly due to `unwrap_or(Paris)`, but scanning 15 extra variants that can never match is unnecessary. Consider either starting the iteration from `Shanghai`, or using a separate `POST_MERGE_FORKS` slice.

crates/blockchain/payload.rs
```diff
@@ -552,16 +554,6 @@ impl Blockchain {
             // TODO: maybe fetch hash too when filtering mempool so we don't have to compute it here (we can do this in the same refactor as adding timestamp)
             let tx_hash = head_tx.tx.hash();
 
-            // Check whether the tx is replay-protected
```
ElFantasma, line 555: Removing the EIP-155 replay protection check is correct for a post-merge client (EIP-155 is always active), but this is a semantic change that goes beyond pure refactoring. The commit message doesn't mention it — worth calling out in the PR description since it affects both L1 (`payload.rs`) and L2 (`payload_builder.rs`) payload building paths.

PR #5397, opened 2025-11-20
perf(l1,l2): avoid unnecessary code hashing

Oppen:
Implements the `Code::from_bytecode_unchecked` function to create the
code object from a trusted or bogus hash.

The trusted case is used for snap sync, where the downloader task
already checks the hash is correct, so it can pass it down.

The bogus hash case is for init code, where the hash is never used
(there is no account pointing to it, no entries in the DB addressed by
it and no opcode can access the current code's hash, only the hash for
an account associated to an address).

### Main
http://ethrex-performance-1:4000/reports/37
```
Title ↑ ↓	Max (MGas/s) ↑ ↓	p50 (MGas/s) ↑ ↓	p95 (MGas/s) ↑ ↓	p99 (MGas/s) ↑ ↓	Min (MGas/s) ↑ ↓	N	Description	Start Time	End Time	Duration (ms) ↑ ↓	FCU time (ms) ↑ ↓	NP time (ms) ↑ ↓
tests_benchmark_test_worst_bytecode.py__test_worst_initcode_jumpdest_analysis[fork_Prague-benchmark-blockchain_test_engine_x-00]-gas-value	303.92	303.92	303.92	303.92	303.92	1	Description not found on metadata file	2025-11-26 19:50:09.584877+00	2025-11-26 19:50:09.963341+00	378.46	0.42	329.04
tests_benchmark_test_worst_bytecode.py__test_worst_initcode_jumpdest_analysis[fork_Prague-benchmark-blockchain_test_engine_x-5b]-gas-value	269.02	269.02	269.02	269.02	269.02	1	Description not found on metadata file	2025-11-26 19:50:10.913177+00	2025-11-26 19:50:11.334488+00	421.31	0.38	371.72
tests_benchmark_test_worst_bytecode.py__test_worst_initcode_jumpdest_analysis[fork_Prague-benchmark-blockchain_test_engine_x-605b5b]-gas-value	259.06	259.06	259.06	259.06	259.06	1	Description not found on metadata file	2025-11-26 19:50:12.312645+00	2025-11-26 19:50:12.748253+00	435.61	0.70	386.01
tests_benchmark_test_worst_bytecode.py__test_worst_initcode_jumpdest_analysis[fork_Prague-benchmark-blockchain_test_engine_x-605b]-gas-value	241.64	241.64	241.64	241.64	241.64	1	Description not found on metadata file	2025-11-26 19:50:13.776154+00	2025-11-26 19:50:14.239418+00	463.27	0.46	413.83
[...]

PR #5485, opened 2025-12-02
fix(l1,l2,levm): avoid double hashing code in Account conversions

Snezhkko:
Introduced Code::from_hashed_bytecode and refactored Account conversions to reuse a single computed code hash, removing redundant keccak calls. This keeps info.code_hash and code.hash consistent while reducing work.

### DIFF
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -32,12 +32,20 @@ pub struct Code {
 }
 
 impl Code {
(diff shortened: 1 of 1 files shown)


MegaRedHand:
Hello @Snezhkko, and thanks for your contribution, but this is a duplicate of #5397

Snezhkko:
> Hello @Snezhkko, and thanks for your contribution, but this is a duplicate of #5397

Oh. see

PR #5487, opened 2025-12-02
chore(l1): add a explicit blockheader partialeq impl due to oncelock

### LINKED ISSUE #5125: Block header comparison is broken due to `OnceCell`
We are currently deriving the `PartialEq` trait for `BlockHeader`, but this is incorrect: we memoize its hash in a `OnceCell`, which starts uninitialized, and could make the comparison of two equal headers fail. This doesn't fail when tests are run using the in-memory engine because it doesn't serialize its contents, but when generalizing this for multiple backends in #4623, some tests started failing due to this. This was fixed indirectly in [d5f28b1](https://github.com/lambdaclass/ethrex/pull/4623/commits/d5f28b1f0f70a9b7949cbbffc8c62b7fe1e4eb71), but comparison should be fixed to avoid surprises.

As an additional measure, we should add a test that asserts hashes are being properly loaded from the DB.

edg-l:
**Motivation**

> We are currently deriving the PartialEq trait for BlockHeader, but this is incorrect: we memoize its hash in a OnceCell, which starts uninitialized, and could make the comparison of two equal headers fail. 

Closes #5125


crates/common/types/block.rs
```diff
@@ -146,6 +144,35 @@ pub struct BlockHeader {
...
+            && self.withdrawals_root == other.withdrawals_root
+            && self.difficulty == other.difficulty
+            && self.ommers_hash == other.ommers_hash
+            && self.requests_hash == other.requests_hash
+            && self.logs_bloom == other.logs_bloom
+            && self.extra_data == other.extra_data
+    }
+}
+
+impl Eq for BlockHeader {}
```
MegaRedHand: Isn't this the same as the derive?

crates/common/types/block.rs
```diff
@@ -146,6 +144,35 @@ pub struct BlockHeader {
     pub requests_hash: Option<H256>,
 }
 
+// Needs a explicit impl due to the hash OnceLock.
+impl PartialEq for BlockHeader {
+    fn eq(&self, other: &Self) -> bool {
+        self.parent_hash == other.parent_hash
```
MegaRedHand: We should use destructuring here to make any new field a compile error.

PR #5490, opened 2025-12-02
chore(l2): patch `ark` algebra for ZisK guest

ilitteri:


ilitteri:
`ark` libs are not used anymore in the guest programs

PR #5492, opened 2025-12-02
fix(l1): reduce spammy trace-level logs during snap-sync

MegaRedHand:
**Motivation**

Running ethrex with trace-level logs results in a lot of spam of "We didn't get a peer from the table" or similar, due to busy loops we have when we don't have many peers. This is especially important for the first few stages, since we start with a small number of peers.

**Description**

This PR reduces the spam by adding a sleep of 10ms, plus a counter to make the warning be printed only once every 1000 times (effectively once every 10s). Also does the same in other stages of snap-sync, adding logs in other busy loops to detect if/when we land in those cases instead of silently ignoring them.

### DIFF
--- a/crates/networking/p2p/sync/storage_healing.rs
+++ b/crates/networking/p2p/sync/storage_healing.rs
@@ -164,6 +164,8 @@ pub async fn heal_storage_trie(
     let (task_sender, mut task_receiver) =
         tokio::sync::mpsc::channel::<Result<TrieNodes, RequestStorageTrieNodes>>(1000);
 
+    let mut logged_no_free_peers_count = 0;
+
     loop {
         yield_now().await;
         if state.last_update.elapsed() >= SHOW_PROGRESS_INTERVAL_DURATION {
@@ -247,6 +249,7 @@ pub async fn heal_storage_trie(
             peers,
             state.state_root,
             &task_sender,
+            &mut logged_no_free_peers_count,
         )
         .await;
 
@@ -317,6 +320,7 @@ async fn ask_peers_for_nodes(
     peers: &mut PeerHandler,
     state_root: H256,
     task_sender: &Sender<Result<TrieNodes, RequestStorageTrieNodes>>,
+    logged_no_free_peers_count: &mut u32,
 ) {
     if (requests.len() as u32) < MAX_IN_FLIGHT_REQUESTS && !download_queue.is_empty() {
         let Some((peer_id, connection)) = peers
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/sync/storage_healing.rs
```diff
@@ -317,20 +320,24 @@ async fn ask_peers_for_nodes(
...
-            // If we have no peers we shrug our shoulders and wait until next free peer
-            trace!("We have no free peers for storage healing!");
+            // Log ~ once every 10 seconds
+            if *logged_no_free_peers_count == 0 {
+                trace!("We are missing peers in heal_storage_trie");
+                *logged_no_free_peers_count = 1000;
+            }
+            *logged_no_free_peers_count -= 1;
+            // Sleep for a bit to avoid busy polling
+            tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
```
Oppen, line 340: Should this be a `yield_now` instead?

crates/networking/p2p/peer_handler.rs
```diff
@@ -337,7 +339,14 @@ impl PeerHandler {
...
             else {
-                trace!("We didn't get a peer from the table");
+                // Log ~ once every 10 seconds
+                if logged_no_free_peers_count == 0 {
+                    trace!("We are missing peers in request_block_headers");
+                    logged_no_free_peers_count = 1000;
+                }
+                logged_no_free_peers_count -= 1;
+                // Sleep a bit to avoid busy polling
+                tokio::time::sleep(Duration::from_millis(10)).await;
```
Oppen, line 349: Ditto.

crates/networking/p2p/peer_handler.rs
```diff
@@ -736,7 +747,14 @@ impl PeerHandler {
...
             else {
-                trace!("We are missing peers in request_account_range_request");
+                // Log ~ once every 10 seconds
+                if logged_no_free_peers_count == 0 {
+                    trace!("We are missing peers in request_account_range");
+                    logged_no_free_peers_count = 1000;
+                }
+                logged_no_free_peers_count -= 1;
+                // Sleep a bit to avoid busy polling
+                tokio::time::sleep(Duration::from_millis(10)).await;
```
Oppen, line 757: Ditto.

crates/networking/p2p/peer_handler.rs
```diff
@@ -1012,6 +1032,14 @@ impl PeerHandler {
...
                 .await?
             else {
+                // Log ~ once every 10 seconds
+                if logged_no_free_peers_count == 0 {
+                    trace!("We are missing peers in request_bytecodes");
+                    logged_no_free_peers_count = 1000;
+                }
+                logged_no_free_peers_count -= 1;
+                // Sleep a bit to avoid busy polling
+                tokio::time::sleep(Duration::from_millis(10)).await;
```
Oppen, line 1042: Ditto.

crates/networking/p2p/peer_handler.rs
```diff
@@ -1526,6 +1556,14 @@ impl PeerHandler {
...
                 .await?
             else {
+                // Log ~ once every 10 seconds
+                if logged_no_free_peers_count == 0 {
+                    trace!("We are missing peers in request_storage_ranges");
+                    logged_no_free_peers_count = 1000;
+                }
+                logged_no_free_peers_count -= 1;
+                // Sleep a bit to avoid busy polling
+                tokio::time::sleep(Duration::from_millis(10)).await;
```
Oppen, line 1566: Ditto.

Oppen (approved):

crates/networking/p2p/sync/storage_healing.rs
MegaRedHand: The sleep is to make the log rate predictable, since this is effectively polling continuously the task completion channel and peer table.
If we reach this point, we haven't finished any requests recently (so no peers were recently freed), and no new peers were found. That probably won't change for the next 10ms, so sleeping for a bit will reduce resource usage without affecting performance.

PR #5496, opened 2025-12-02
fix(l1): preventing invalid contact errors from propagating on ENRRequest handling

ElFantasma:
**Motivation**

Contact validation shouldn't raise ERROR exception on logs.

**Description**

This PR keeps logging DEBUG lines when receiving invalid contacts or non-solicited ENRResponses from peers, but avoids returning an error that will be later logged as an ERROR log line

### DIFF
--- a/.github/workflows/pr-main_l2_prover.yaml
+++ b/.github/workflows/pr-main_l2_prover.yaml
@@ -19,7 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        backend: ["sp1", "risc0", "zisk", "openvm"]
+        # Note: openvm backend linting is disabled due to a problem in the installation https://github.com/lambdaclass/ethrex/issues/5509
+        backend: ["sp1", "risc0", "zisk"]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4


crates/networking/p2p/discv4/server.rs
```diff
@@ -488,13 +488,16 @@ impl DiscoveryServer {
     ) -> Result<(), DiscoveryServerError> {
         let node_id = node_id(&sender_public_key);
 
-        let contact = self
+        if let Ok(contact) = self
```
MegaRedHand: We can use a `let else` here

crates/networking/p2p/discv4/server.rs
```diff
@@ -541,6 +540,22 @@ impl DiscoveryServer {
...
 
+    async fn validate_enr_response(
+        &mut self,
+        sender_public_key: H512,
+        node_id: H256,
+        from: SocketAddr,
+    ) -> Result<Contact, DiscoveryServerError> {
+        let contact = self
+            .validate_contact(sender_public_key, node_id, from, "ENRRequest")
+            .await?;
```
Oppen, line 551: Shouldn't this trigger a disconnect or something?

ElFantasma: This is discovery, it's all UDP and there are no connections yet.

crates/networking/p2p/discv4/server.rs
```diff
@@ -488,13 +488,12 @@ impl DiscoveryServer {
...
-
-        if !contact.has_pending_enr_request() {
-            debug!(received = "ENRResponse", from = %format!("{sender_public_key:#x}"), "unsolicited message received, skipping");
-            return Err(DiscoveryServerError::InvalidContact);
+        if self
+            .validate_enr_response(sender_public_key, node_id, from)
+            .await
+            .is_err()
+        {
+            return Ok(());
```
pablodeymo, line 496: This function (`validate_enr_response`) returns a Contact element. Given that this call is not using the return value, I think we could have a version returning unit type (i.e. Result of nothing).

mpaulucci (approved):

PR #5497, opened 2025-12-02
perf(l1): parallelize storage trie merkelization

iovoid:
**Motivation**

Currently we split merkelization by account. However, some accounts have much higher activity than others, and this leads to the merkelization work being split unevenly.

**Description**

First we do all the storage merkelization work, split by storage slot and then after gathering the results we merkelize the accounts.

This only has pipelineing for the storage part, not the accounts.


crates/blockchain/blockchain.rs
```diff
@@ -471,190 +490,196 @@ impl Blockchain {
...
-    #[allow(clippy::too_many_arguments)]
-    fn process_incoming_update_message(
-        storage: &Store,
-        state_trie: &mut Trie,
-        updates: Vec<(H256, AccountUpdate)>,
-        storage_updates_map: &mut StoreUpdatesMap,
+    /// Collapses a root branch node into an extension or leaf node if it has only one valid child.
+    /// Returns None if there are no valid children.
+    ///
+    /// Precondition: the branch node's reference to their children are of type NodeRef::Node
```
MegaRedHand: I see we still try to fetch the node from the DB if it's `NodeRef::Hash`, does this precondition still apply?

crates/blockchain/blockchain.rs
```diff
@@ -340,128 +342,145 @@ impl Blockchain {
...
+                    let bucket = hashed_key.as_fixed_bytes()[0] >> 4;
+                    workers_tx[bucket as usize]
+                        .send(MerklizationRequest::MerklizeStorage {
+                            prefix: hashed_address,
+                            key: hashed_key,
+                            value: if value.is_zero() {
+                                vec![]
+                            } else {
+                                value.encode_to_vec()
+                            },
```
MegaRedHand: Can this `if` be moved to the worker thread?

crates/blockchain/blockchain.rs
```diff
@@ -340,128 +342,145 @@ impl Blockchain {
...
+
+                account_updates.push(update.clone());
+                if update.removed_storage | update.removed {
+                    for tx in &workers_tx {
+                        tx.send(MerklizationRequest::Delete(hashed_address))
+                            .map_err(|e| StoreError::Custom(format!("send error: {e}")))?;
+                    }
+                }
+                for (key, value) in update.added_storage {
+                    let hashed_key = keccak(key);
```
MegaRedHand, line 533: We should check if all this hashing doesn't hurt performance. One alternative is to use a cheaper hash here, and leave the keccak to the workers.

crates/blockchain/blockchain.rs
```diff
@@ -471,190 +490,196 @@ impl Blockchain {
...
-            );
-            if update.removed {
-                // Remove account from trie
-                state_trie.remove(&hashed_address)?;
-                account_states.remove(&hashed_address_h256);
-                continue;
+        prefix: Option<H256>,
+        mut root: BranchNode,
+    ) -> Result<Option<Node>, StoreError> {
+        root.choices.iter_mut().for_each(NodeRef::clear_hash);
```
MegaRedHand, line 680: We should comment on why this is needed

crates/blockchain/blockchain.rs
```diff
@@ -1892,56 +1917,46 @@ pub fn get_total_blob_gas(tx: &EIP4844Transaction) -> u32 {
...
-        // Replace root with an extension node leading to the child
-        Node::Branch(_) => {
-            ExtensionNode::new(Nibbles::from_hex(vec![choice as u8]), only_child).into()
-        }
-        // Replace root with the child extension node, updating its path in the process
-        Node::Extension(mut extension_node) => {
-            let mut extension_node = extension_node.take();
-            extension_node.prefix.prepend(choice as u8);
-            extension_node.into()
+fn branchify(node: Node) -> BranchNode {
```
MegaRedHand: It makes sense to keep the `BranchNode` in a box, I think:

```suggestion
/// Turns [`Node`] into a [`BranchNode`], if needed, and returns it
fn branchify(node: Node) -> Box<BranchNode> {
```

The only use of this function is in `collect_trie`, and after calling this function it puts the result in a `Box`

crates/blockchain/blockchain.rs
```diff
@@ -471,190 +490,196 @@ impl Blockchain {
...
+                    *EMPTY_TRIE_HASH,
+                )?;
+                let root = trie
+                    .db()
+                    .get(Nibbles::default())?
+                    .filter(|rlp| !rlp.is_empty())
+                    .map(keccak)
+                    .unwrap_or(*EMPTY_TRIE_HASH);
+                self.storage
+                    .open_storage_trie(account_hash, parent_header.state_root, root)?
```
MegaRedHand, line 665: Why are we opening the trie twice?

crates/blockchain/blockchain.rs
iovoid: Good catch, this was refactored into no longer needing that to be true. Fixed in 9f2690e67

crates/blockchain/blockchain.rs
iovoid: Moved in 4c13bb2ee

crates/blockchain/blockchain.rs
iovoid: keccak _is_ slow but it can't be moved, we need to know the first nibble (because the task is split based on that) to determine which worker should handle the task

crates/blockchain/blockchain.rs
iovoid: Once to obtain something readable, and the second time with the correct root. This could also be done by replacing trie.root with a reference to the correct root.

Also now that get_node() no longer uses checked gets, we might be able to avoid hashing the node too, since we only need to know if the tree is empty.

crates/blockchain/blockchain.rs
iovoid: Done in f0e49a931

crates/blockchain/blockchain.rs
iovoid: Added a comment in f0e49a931 detailing the immediate reason (it causes the children to be included in commit). However, the deeper reason isn't exactly known.

crates/blockchain/blockchain.rs
```diff
@@ -471,190 +486,195 @@ impl Blockchain {
...
-                account_state.balance = info.balance;
-                account_state.code_hash = info.code_hash;
-                // Store updated code in DB
-                if let Some(code) = &update.code {
-                    trace!("Updated code");
-                    code_updates.insert(info.code_hash, code.clone());
+                    if value.is_zero() {
+                        trie.remove(&key.as_bytes().to_vec())?;
+                    } else {
+                        trie.insert(key.as_bytes().to_vec(), value.encode_to_vec())?;
```
MegaRedHand, line 760: This can be replaced by `encode`, to reuse a single buffer across all calls (calling `clear` after each one).

crates/blockchain/blockchain.rs
```diff
@@ -471,190 +486,195 @@ impl Blockchain {
...
+                    if let Some(storage_root) = storage_root {
+                        old_state.storage_root = storage_root;
+                    }
+                    if let Some(info) = state.info {
+                        old_state.nonce = info.nonce;
+                        old_state.balance = info.balance;
+                        old_state.code_hash = info.code_hash;
+                    }
+                    if *old_state != AccountState::default() {
+                        state_trie.insert(pathrlp, old_state.encode_to_vec())?;
```
MegaRedHand: Same

MegaRedHand (approved):
LGTM

crates/blockchain/blockchain.rs
iovoid: Insert takes ownership of the buffer, you'd have to clone it anyway

crates/blockchain/blockchain.rs
```diff
@@ -340,128 +342,141 @@ impl Blockchain {
...
+                    let hashed_key = keccak(key);
+                    let bucket = hashed_key.as_fixed_bytes()[0] >> 4;
+                    workers_tx[bucket as usize]
+                        .send(MerklizationRequest::MerklizeStorage {
+                            prefix: hashed_address,
+                            key: hashed_key,
+                            value,
+                        })
+                        .map_err(|e| StoreError::Custom(format!("send error: {e}")))?;
+                }
```
Oppen, line 542: One message per storage slot touched sounds like a lot of overhead. Why not group the keys first?

iovoid: The idea was that since keccak is "slow", we want the worker to work in the meantime instead of waiting until every key is hashed.

We want that, by the time the update stream is closed, as much work is done as possible since everything after is overhead.

The only way to really know which one is better is to measure since each approach excels in a different scenario, but not sure how easy it would be to detect the difference.

ilitteri (approved):

Issue #4491, opened 2025-09-15
[Tracking] Implement discv5


ElFantasma:
From [#el-networking](https://discord.com/channels/595666850260713488/1413582163747803247)  channel in Eth R&D Discord server, [discv5](https://notes.ethereum.org/@cskiraly/el-discovery-v5-tracker), is planned to be supported by Jan 15, 2026 and mandatory by Jan 31, 2026.

See: https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md and related documents.

Issue #3660, opened 2025-07-16
Precompute our fork ID and store it inside the discv4 process


MegaRedHand:
We're going to the DB each time we receive an `ENRResponse` message. This is less than ideal, so we should instead precompute this and store it locally in the process, to make request/response handling as light as possible.

lakshya-sky:
Hey @MegaRedHand, #5539 checks and updates forkid in local node record so we can just read it form the local node record in the disc server. Furthermore, after #5546 it will be even faster to read forkid from the local node record.

PR #5500, opened 2025-12-03
feat(l1): enr fork id filtering

### LINKED ISSUE #5506: Check fork_id before attempting to initiate an RLPx connection
Now that we are [requesting ENRs](https://github.com/lambdaclass/ethrex/pull/5446) from peers, we can filter valid contacts to initiate RLPx connections on lookup. 

We should validate fork_id field when [receiving ENRResponses](https://github.com/lambdaclass/ethrex/blob/70463a1124b08d1a8b39d5a3059fb9a2f0bca0aa/crates/networking/p2p/discv4/peer_table.rs#L294) and use that information to filter on [get_contact_for_lookup](https://github.com/lambdaclass/ethrex/blob/70463a1124b08d1a8b39d5a3059fb9a2f0bca0aa/crates/networking/p2p/discv4/peer_table.rs#L662)

Geth seems to be doing the same: https://github.com/ethereum/go-ethereum/blob/d3679c2f2e9d7f6d354375bfad0165016e6d059c/eth/protocols/eth/discovery.go#L71-L80

Related: #5423

edg-l:
**Motivation**

Now that we are https://github.com/lambdaclass/ethrex/pull/5446 from peers, we can filter valid contacts to initiate RLPx connections on lookup.

Closes #5506

### DIFF
--- a/crates/common/trie/trie_sorted.rs
+++ b/crates/common/trie/trie_sorted.rs
@@ -6,7 +6,6 @@ use crossbeam::channel::{Receiver, Sender, bounded};
 use ethereum_types::H256;
 use ethrex_threadpool::ThreadPool;
 use std::{sync::Arc, thread::scope};
-use tracing::debug;
 
 /// The elements of the stack represent the branch node that is the parent of the current
 /// parent element. When the current parent is no longer valid (is not the parent of
@@ -139,16 +138,6 @@ fn add_current_to_parent_and_write_queue(
     };
     parent_element.element.choices[index as usize] =
         node.compute_hash_no_alloc(&mut nodehash_buffer).into();
-    debug!(
-        "branch {:x?}",
-        parent_element
-            .element
-            .choices
-            .iter()
-            .enumerate()
-            .filter_map(|(index, child)| child.is_valid().then_some(index))
-            .collect::<Vec<_>>()
-    );
     nodes_to_write.push((target_path, node));
     Ok(())
 }
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -281,6 +281,8 @@ impl PeerHandler {
 
         *METRICS.headers_download_start_time.lock().await = Some(SystemTime::now());
 
+        let mut logged_no_free_peers_count = 0;
+
         loop {
             if let Ok((headers, peer_id, _connection, startblock, previous_chunk_limit)) =
                 task_receiver.try_recv()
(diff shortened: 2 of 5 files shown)


crates/networking/p2p/discv4/server.rs
```diff
@@ -505,6 +511,49 @@ impl DiscoveryServer {
             )
             .await?;
 
+        let Some(remote_fork_id) = pairs.eth else {
```
ElFantasma, line 527: Is it possible to move the node_record parsing and validation logic to it's own function?

edg-l: moved it to its own function

ElFantasma (approved):
LGTM

I think there is room for further code refactoring, but there are some pending PRs that may conflict with this one. Let's merge all and take another look at this code afterwards.

Oppen:
> LGTM
> 
> I think there is room for further code refactoring, but there are some pending PRs that may conflict with this one. Let's merge all and take another look at this code afterwards.

On those lines, I would suggest:
1. Tracking validated peers separately from unvalidated ones, so we can validate once and forget about it afterwards;
2. Keeping track of invalid peers separately, in a more lightweight way.

Ideally, I think there would be a clear flow like this:
```
New peer -> validate +-> peer table
                     |
                     +-> bad peers list
```

Oppen (approved):

crates/networking/p2p/discv4/server.rs
```diff
@@ -501,10 +505,68 @@ impl DiscoveryServer {
...
+            .store
+            .get_block_header(latest_block_number)?
+            .ok_or(DiscoveryServerError::InvalidContact)?;
+
+        let local_fork_id = ForkId::new(
+            chain_config,
+            genesis_header.clone(),
+            latest_block_header.timestamp,
+            latest_block_number,
+        );
```
MegaRedHand, line 551: We should probably cache this: #3660 (in another PR)

ElFantasma:
> > LGTM
> > I think there is room for further code refactoring, but there are some pending PRs that may conflict with this one. Let's merge all and take another look at this code afterwards.
> 
> On those lines, I would suggest:
> 
>     1. Tracking validated peers separately from unvalidated ones, so we can validate once and forget about it afterwards;
> 
>     2. Keeping track of invalid peers separately, in a more lightweight way.
> 
> 
> Ideally, I think there would be a clear flow like this:
> 
> ```
> New peer -> validate +-> peer table
>                      |
>                      +-> bad peers list
> ```

Yeah, I definitely  like this approach. Probably several lists with different categories and different update intervals: non-validated-contacts, ping-requested-contacts, enr-requested-contacts, once-connected-contacts, etc... But I think this should be done after #4245, or #4491

PR #5507, opened 2025-12-03
feat(l1): detect and update stale enr on ping pong

### LINKED ISSUE #5488: Detect stale ENRs when PING/PONG is received
When a peer sends us a PING or PONG message, it includes a sequence number. If that sequence number is higher than the one in our local copy of the peer's ENR, then we should discard that ENR to later re-request the ENR.

lakshya-sky:
**Motivation**

**Description**

Closes #5488

### DIFF
--- a/crates/networking/p2p/discv4/server.rs
+++ b/crates/networking/p2p/discv4/server.rs
@@ -182,7 +182,8 @@ impl DiscoveryServer {
                     return Ok(());
                 }
 
-                self.handle_find_node(sender_public_key, from).await?;
+                self.handle_find_node(sender_public_key, find_node_message.target, from)
+                    .await?;
             }
             Message::Neighbors(neighbors_message) => {
                 trace!(received = "Neighbors", msg = ?neighbors_message, from = %format!("{sender_public_key:#x}"));
@@ -438,14 +439,18 @@ impl DiscoveryServer {
     async fn handle_find_node(
         &mut self,
         sender_public_key: H512,
+        target: H512,
         from: SocketAddr,
     ) -> Result<(), DiscoveryServerError> {
-        let node_id = node_id(&sender_public_key);
+        let sender_id = node_id(&sender_public_key);
         if let Ok(contact) = self
-            .validate_contact(sender_public_key, node_id, from, "FindNode")
+            .validate_contact(sender_public_key, sender_id, from, "FindNode")
             .await
         {
-            let neighbors = self.peer_table.get_closest_nodes(&node_id).await?;
+            // According to https://github.com/ethereum/devp2p/blob/master/discv4.md#findnode-packet-0x03
+            // reply closest 16 nodes to target
+            let target_id = node_id(&target);
+            let neighbors = self.peer_table.get_closest_nodes(&target_id).await?;
 
             // A single node encodes to at most 89B, so 8 of them are at most 712B plus
             // recursive length and expiration time, well within bound of 1280B per packet.


crates/networking/p2p/discv4/server.rs
```diff
@@ -384,6 +362,33 @@ impl DiscoveryServer {
         Ok(())
     }
 
+    async fn send_enr_request(&mut self, node: &Node) -> Result<(), DiscoveryServerError> {
```
ElFantasma, line 354: Let's factorize common code (message encoding, hash extraction, `udp_socket.send`, maybe `set_disposable`) with `send_ping`/`send_ping_internal` to an utility function to minimize code repetiton.

crates/networking/p2p/discv4/server.rs
```diff
@@ -398,11 +403,34 @@ impl DiscoveryServer {
...
+                .and_then(|c| c.record)
+                .map(|r| r.seq);
+
+            let received_enr_seq = ping_message.enr_seq;
+
+            if let (Some(received), Some(stored)) = (received_enr_seq, stored_enr_seq)
+                && received > stored
+            {
+                self.send_enr_request(&node).await?;
+            }
```
ElFantasma, line 407:
```suggestion
            if received_enr_seq.is_some_and(|r| stored_enr_seq.is_some_and(|s| r > s)) {
                self.send_enr_request(&node).await?;
            }
```
Not sure which one I prefer 🤔

Oppen: Maybe [transpose](https://doc.rust-lang.org/std/primitive.array.html#method.transpose-1):

```suggestion
            if [received_enr_seq, stored_enr_seq].transpose().is_some_and(|[received, stored]| received > stored) {
                self.send_enr_request(&node).await?;
            }
```

Oppen: Or:
```suggestion
            if let Some([received, stored]) = [received_enr_seq, stored_enr_seq].transpose() && received > stored {
                self.send_enr_request(&node).await?;
            }
```

crates/networking/p2p/discv4/server.rs
```diff
@@ -384,6 +362,33 @@ impl DiscoveryServer {
...
+        if self.udp_socket.send_to(&buf, node.udp_addr())
+                .await
+                .inspect_err( |e| error!(sending = "ENRRequest", addr = ?node.udp_addr(), to = %format!("{:#x}", node.public_key), err=?e, "Error sending message"),)
+                .is_err()
+            {
+                self.peer_table
+                    .set_disposable(&node.node_id())
+                    .await?;
+                METRICS.record_new_discarded_node().await;
+            }
```
Oppen: Either put this inside the `inspect_err` or convert everything to an `if let Err(e)`...
Having two ways to check the same thing in the same place makes the code more confusing than it needs to be.

crates/networking/p2p/discv4/server.rs
```diff
@@ -384,6 +362,33 @@ impl DiscoveryServer {
...
 
+    async fn send_enr_request(&mut self, node: &Node) -> Result<(), DiscoveryServerError> {
+        let expiration: u64 = get_msg_expiration_from_seconds(EXPIRATION_SECONDS);
+        let enr_request = Message::ENRRequest(ENRRequestMessage { expiration });
+
+        let mut buf = Vec::new();
+        enr_request.encode_with_header(&mut buf, &self.signer);
+        let enr_request_hash: [u8; 32] = buf[..32]
+            .try_into()
+            .expect("first 32 bytes are the message hash");
```
Oppen: It would probably be more convenient to just return the hash from `encode_with_header`.

Oppen: Nvm, this is just moved code.

crates/networking/p2p/discv4/server.rs
lakshya-sky: `transpose` is an unstable feature.

MegaRedHand: 😭

crates/networking/p2p/discv4/server.rs
lakshya-sky: Added send_else_dispose to remove code duplication.

crates/networking/p2p/discv4/server.rs
lakshya-sky: changed to `if let Err(e)..` for better error handling.

PR #5513, opened 2025-12-04
chore(l1): remove massive branch debug log spam

edg-l:
**Motivation**

This log provides no useful info and it spams so much my terminal couldnt keep up with the output, so much that i had to SIGINT ethrex to stop it, since ctrl-c didnt work

image


MegaRedHand:
Was already removed in #5492, as part of trace logs cleanup

PR #5518, opened 2025-12-04
feat(l1): improved lookup interval

### LINKED ISSUE #5149: Make Discovery and Connection attempt Lookup interval automatically adjustable.
Currently we have two modes: [Initial](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/discv4/server.rs#L42) and [final](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/discv4/server.rs#L43) lookup interval. We can improve this to make it dependent on the percentage of peers we have according to the [target](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/discv4/peer_table.rs#L35).

ElFantasma:
**Motivation**

Node lookup interval is used to send `FindNode` and `ENRRequest` messages for discovery and to try initiate new RLPx connections and currently it uses only two values.

**Description**
- This PR changes it for a variable interval according to the peer target completion.
- Also it provides a CLI argument to set up the initial interval, to be able to set the node more or less eager on startup.

Closes #5149 (also addresses #5148, already closed)

### DIFF
--- a/.github/actions/snapsync-run/action.yml
+++ b/.github/actions/snapsync-run/action.yml
@@ -15,6 +15,14 @@ inputs:
     description: Ethrex Docker image tag.
     required: false
     default: main
+  cl_type:
+    description: Consensus layer type (lighthouse, prysm, etc).
+    required: false
+    default: lighthouse
+  cl_image:
+    description: Consensus layer Docker image and tag.
+    required: false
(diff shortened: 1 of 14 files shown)


ilitteri (approved):

Oppen:
LGTM, I'll test it at home and approve if able to sync.

PR #5519, opened 2025-12-04
chore(l1): allow PR title to begin with UPPERCASE

JereSalo:
**Motivation**

**Description**

- Had to change `pull_request_target` to `pull_request` because this makes the workflow apply immediately. With pull_request_target the workflow that would've run would've been the `main` one, which still blocks uppercase characters.

Edit: Had to change the title to lowercase because the old workflow is still running as it was a `pull_request_target` 😞 

Example job run https://github.com/lambdaclass/ethrex/actions/runs/19934067966/job/57153864030?pr=5519


mpaulucci (changes requested):
It still should be a letter A to Z, whether underscore or upperscore

JereSalo:
> It still should be a letter A to Z, whether underscore or upperscore

Yeah, I agree. 
Note that the previous regex didn't cover this either though. Any title could've started with something like "123 Title" and it would've been accepted anyway because the only restriction it imposed was that the first symbol couldn't be an uppercase letter.

I changed it [here](https://github.com/lambdaclass/ethrex/pull/5519/commits/5e4263ccc60f8297a085e829ea3ad59e25afe30b), now it can only start with a letter. Thanks for pointing that out!

.github/workflows/pr_lint_pr_title.yml
```diff
@@ -1,7 +1,7 @@
 name: PR title
 
 on:
-  pull_request_target:
+  pull_request:
```
mpaulucci, line 4: why did you change this?

JereSalo: The reason is in the PR description:
With `pull_request_target` the workflow that will run will be the one in `main` and with `pull_request` it will run the workflow in the current branch. So if I make changes to the workflow now the version of it that would run would be the one in my branch and not the old one (in main).

PR #5527, opened 2025-12-04
feat(l2): add l2 block pipelined

gianbelinche:
**Motivation**
We want to add the performance improvement done in L1 with the block pipelining also for the L2.

**Description**
Adds the `add_l2_block_pipeline` function to the blockchain and uses it for regenerating the blockchain from checkpoints.

### DIFF
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -475,10 +475,7 @@ impl Command {
                         }
 
                         // Execute block
-                        blockchain.add_block_pipeline_with_fee_config(
-                            block.clone(),
-                            Some(fee_config.clone()),
-                        )?;
+                        blockchain.add_block_pipeline(block.clone())?;
 
                         // Add fee config to rollup store
                         rollup_store


ilitteri (approved):

cmd/ethrex/l2/command.rs
```diff
@@ -475,7 +475,7 @@ impl Command {
                         }
 
                         // Execute block
-                        blockchain.add_block(block.clone())?;
+                        blockchain.add_block_pipeline(block.clone())?;
```
avilagaston9, line 478: shouldn't we use `add_block_pipeline_with_fee_config`?

crates/blockchain/blockchain.rs
```diff
@@ -1091,8 +1084,26 @@ impl Blockchain {
     }
 
     pub fn add_block_pipeline(&self, block: Block) -> Result<(), ChainError> {
+        self.add_block_pipeline_with_fee_config(block, None)
+    }
+
+    pub fn add_block_pipeline_with_fee_config(
+        &self,
+        block: Block,
+        fee_config: Option<FeeConfig>,
```
avilagaston9: Why not directly use the `fee_config` inside `self.options.r#type`?

cmd/ethrex/l2/command.rs
ilitteri: Yes

crates/blockchain/blockchain.rs
ilitteri: Is it guaranteed to be updated?

avilagaston9: No. It’s the responsibility of the caller to keep it up to date. But until this PR, we were always assuming that it was up to date, especially our payload builder. See my comment below.

crates/blockchain/blockchain.rs
```diff
@@ -1664,13 +1683,25 @@ impl Blockchain {
 }
 
 pub fn new_evm(blockchain_type: &BlockchainType, vm_db: StoreVmDatabase) -> Result<Evm, EvmError> {
+    new_evm_with_fee_config(blockchain_type, vm_db, None)
+}
```
avilagaston9: Careful! Our payload builder calls `new_evm`, assuming that the inner `fee_config` is up to date. With this change, it will use `fee_config = None`.

crates/blockchain/blockchain.rs
```diff
@@ -1091,8 +1084,18 @@ impl Blockchain {
     }
 
     pub fn add_block_pipeline(&self, block: Block) -> Result<(), ChainError> {
+        // Validate if it can be the new head and find the parent
+        let Ok(parent_header) = find_parent_header(&block.header, &self.storage) else {
+            // If the parent is not present, we store it as pending.
+            self.storage.add_pending_block(block)?;
+            return Err(ChainError::ParentNotFound);
+        };
```
Oppen, line 1092: In L1 this is performed before sending the message in the first place. Why can't L2 do the same?

Oppen: Oh, I see we were doing it twice regardless. Maybe the parent header should be a parameter to avoid the redundant query.

PR #5531, opened 2025-12-05
refactor(l1): avoid extra allocations in RLPx handshake

Snezhkko:
Remove unnecessary allocations on the handshake path. 
receive_handshake_msg() now returns the owned buffer instead of cloning a slice, and receive_auth()/receive_ack() move msg_bytes into init_message instead of cloning. This keeps behavior unchanged while reducing copies and heap traffic during auth/ack processing.

### DIFF
--- a/crates/networking/p2p/rlpx/connection/handshake.rs
+++ b/crates/networking/p2p/rlpx/connection/handshake.rs
@@ -261,7 +261,6 @@ async fn receive_handshake_msg<S: AsyncRead + std::marker::Unpin>(
 
     // Read the rest of the message
     stream.read_exact(&mut buf[2..]).await?;
-    buf.truncate(msg_size + 2);
     Ok(buf)
 }
 


crates/networking/p2p/rlpx/connection/handshake.rs
```diff
@@ -258,15 +258,9 @@ async fn receive_handshake_msg<S: AsyncRead + std::marker::Unpin>(
...
-    if buf.len() < msg_size + 2 {
-        return Err(PeerConnectionError::CryptographyError(String::from(
-            "bad buf size",
-        )));
-    }
-    stream.read_exact(&mut buf[2..msg_size + 2]).await?;
-    let ack_bytes = &buf[..msg_size + 2];
-    Ok(ack_bytes.to_vec())
+    stream.read_exact(&mut buf[2..]).await?;
+    buf.truncate(msg_size + 2);
```
MegaRedHand: This truncate is unneeded

```suggestion
```

PR #5544, opened 2025-12-05
feat(l2): add optimize option to sdk compile contract

tomip01:
**Motivation**

We want to able to optimize compilation of contracts. With this we can reduce bytecode to deploy

**Description**

Add optional argument to do optimize runs


cmd/ethrex/build_l2.rs
ilitteri, line 1: Let's use 999999 (the maximum).

tomip01: We want to reduce the bytecode for the contract. Having a higher number will increase the bytecode but is more efficient to execute

JereSalo (commented):
I wanted to see how much optimization and optimization runs impacted on the contract size so I measured these in some contracts.

Size in bytes post compiling contracts:

| Contract              | No Optimizer | 1 Run | 50 Runs | 200 Runs | 999999 Runs |
|----------------------|--------------|-------|---------|----------|--------------|
| **Print.sol**        | 3816         | 2504  | 2507    | 2518     | 3107         |
| **OnChainProposer.sol** | 24772    | 10748 | 10763   | 10799    | 16570        |

The difference between 50 and 200 is negligible, yet when not optimizing it or going to the extreme it becomes very noticeable.
Maybe we can go for 999999 as @ilitteri suggested (which will make transactions to the contract cheaper and deployment more expensive) but if the contract size becomes too big at some point we should consider going for lower optimizer runs.

Edit: The good thing about this is that with just optimization enabled we have a large room for expanding `OnChainProposer` contract size, it literally went from 24772 to 10763, which is a relief. We can have more descriptive errors now if we want to haha

cmd/ethrex/build_l2.rs
tomip01: Done here [ce3c3b7](https://github.com/lambdaclass/ethrex/pull/5544/commits/ce3c3b71e4421bb49dcebbc5381d41e718bfc05d)

tomip01:
> I wanted to see how much optimization and optimization runs impacted on the contract size so I measured these in some contracts.
> 
> Size in bytes post compiling contracts:
> 
> Contract	No Optimizer	1 Run	50 Runs	200 Runs	999999 Runs
> **Print.sol**	3816	2504	2507	2518	3107
> **OnChainProposer.sol**	24772	10748	10763	10799	16570
> The difference between 50 and 200 is negligible, yet when not optimizing it or going to the extreme it becomes very noticeable. Maybe we can go for 999999 as @ilitteri suggested (which will make transactions to the contract cheaper and deployment more expensive) but if the contract size becomes too big at some point we should consider going for lower optimizer runs.
> 
> Edit: The good thing about this is that with just optimization enabled we have a large room for expanding `OnChainProposer` contract size, it literally went from 24772 to 10763, which is a relief. We can have more descriptive errors now if we want to haha

Nice! I agree. Changed that here [ce3c3b7](https://github.com/lambdaclass/ethrex/pull/5544/commits/ce3c3b71e4421bb49dcebbc5381d41e718bfc05d)

crates/l2/sdk/build.rs
```diff
@@ -44,6 +44,7 @@ fn main() {
         false,
         None,
         &[&contracts_path],
+        Some(50),
```
avilagaston9: We can use None or `999999` as well

tomip01: Done [4683b86](https://github.com/lambdaclass/ethrex/pull/5544/commits/4683b867728bd3c84baf9b98d9a3226f8e2a66f2)

JereSalo (approved):
◝(ᵔᗜᵔ)◜

PR #5546, opened 2025-12-06
chore(l1): use NodeRecordPairs to hold enr entries

### LINKED ISSUE #5068: Simplify ENR handling
Currently we have two separate structures to deal with Ethereum Node Records, a "generic" one, named `NodeRecord`, that allocates to keep the pairs, and a specific one with the ones we understand, named `NodeRecordPairs`. I think we should move to using only the latter. Our own `local_node_record` currently uses the former, cloning a vector for every message. We should always use `NodeRecordPairs`, ignoring fields we don't know, like Geth does, and rejecting peers with invalid values in the ones we do know.

lakshya-sky:
**Motivation**
Instead of storing vec of arbitary blob use NodeRecordPairs which allow updating ENRs less tedious. 
* Every time adding/removing a key-value would require sorting the vec which is not optimal.
* Checking a certain key-value requires decoding whole key-value list is also less optimal.

**Description**

Closes #5068

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -441,6 +441,8 @@ pub async fn init_l1(
     let network = get_network(&opts);
     let datadir = crate::cli::compute_effective_datadir(&opts.datadir, &network, opts.dev);
 
+    raise_fd_limit()?;
+
     migrate_datadir_if_needed(&opts.datadir, &datadir, &network, opts.no_migrate);
 
     if !is_memory_datadir(&datadir) {
@@ -449,8 +451,6 @@ pub async fn init_l1(
 
     let genesis = network.get_genesis()?;
     display_chain_initialization(&genesis);
-
-    raise_fd_limit()?;
     debug!("Preloading KZG trusted setup");
     ethrex_crypto::kzg::warm_up_trusted_setup();
 
--- a/crates/common/types/block_access_list.rs
+++ b/crates/common/types/block_access_list.rs
@@ -1159,13 +1159,16 @@ impl BlockAccessListRecorder {
                 }
             }
 
-            // Add storage reads (excluding slots that were promoted to writes)
+            // Add storage reads (excluding slots that were promoted to writes
+            // or that already exist in storage_writes from any transaction).
             // Sort for canonical BAL ordering (IndexSet preserves insertion order).
             if let Some(reads) = self.storage_reads.get(address) {
                 let promoted = self.reads_promoted_to_writes.get(address);
+                let writes = self.storage_writes.get(address);
                 let mut sorted_reads: Vec<_> = reads
                     .iter()
                     .filter(|slot| !promoted.is_some_and(|p| p.contains(slot)))
+                    .filter(|slot| !writes.is_some_and(|w| w.contains_key(slot)))
                     .copied()
                     .collect();
                 sorted_reads.sort();
@@ -1328,9 +1331,13 @@ impl BlockAccessListRecorder {
                     .and_then(|m| m.get(slot))
                     .copied()
                     .unwrap_or(0);
-                if changes.len() > checkpoint_len {
-                    // This slot had writes after checkpoint - ensure it's recorded as a read
-                    // (Reads that became writes are already in storage_reads since we don't remove them)
+                if changes.len() > checkpoint_len && checkpoint_len == 0 {
+                    // This slot was freshly written after checkpoint (no prior writes).
+                    // Convert to a read since the write is being reverted.
+                    // If checkpoint_len > 0, the slot already has writes from a prior
+                    // transaction and is already represented in storage_changes — adding
+                    // it to storage_reads would violate the invariant that a slot appears
+                    // in at most one of the two lists.
                     self.storage_reads.entry(*addr).or_default().insert(*slot);
                 }
             }
--- a/test/tests/levm/eip7928_tests.rs
+++ b/test/tests/levm/eip7928_tests.rs
@@ -954,3 +954,197 @@ fn test_bal_reverted_write_restores_read() {
     // And not in writes
     assert!(account.storage_changes.is_empty());
 }
+
+/// Regression test: a slot written in tx 1 and then written+reverted in tx 2
+/// must NOT appear in both storage_changes and storage_reads.
+/// See: restore() was adding reverted writes as reads even when prior writes existed.
+#[test]
+fn test_reverted_write_after_prior_tx_write_no_duplicate() {
+    let mut recorder = BlockAccessListRecorder::new();
+
+    // Tx 1: write slot 0x10
+    recorder.set_block_access_index(1);
+    recorder.capture_pre_storage(ALICE, U256::from(0x10), U256::from(0x00));
+    recorder.record_storage_write(ALICE, U256::from(0x10), U256::from(0x01));
+
+    // Transition to tx 2
+    recorder.set_block_access_index(2);
+
+    // Tx 2, inner call: read then write slot 0x10
+    let checkpoint = recorder.checkpoint();
+    recorder.record_storage_read(ALICE, U256::from(0x10));
+    recorder.capture_pre_storage(ALICE, U256::from(0x10), U256::from(0x01));
+    recorder.record_storage_write(ALICE, U256::from(0x10), U256::from(0x99));
+
+    // Inner call reverts
+    recorder.restore(checkpoint);
+
+    let bal = recorder.build();
+
+    // Slot 0x10 should be in storage_changes (from tx 1), NOT in storage_reads
+    let alice = bal.accounts().iter().find(|a| a.address == ALICE).unwrap();
+    assert_eq!(alice.storage_changes.len(), 1);
+    assert_eq!(alice.storage_changes[0].slot, U256::from(0x10));
+    assert!(
+        alice.storage_reads.is_empty(),
+        "slot should not be in both storage_changes and storage_reads"
+    );
+
+    // BAL should pass ordering validation
+    bal.validate_ordering().unwrap();
+}
+
+/// Nested reverts: slot written in inner call A (not reverted), then written
+/// in inner call B (reverted within same tx). Should only be in storage_changes.
+#[test]
+fn test_nested_revert_slot_only_in_changes() {
+    let mut recorder = BlockAccessListRecorder::new();
+    recorder.set_block_access_index(1);
+
+    // Inner call A: write slot 0x10 (succeeds)
+    recorder.capture_pre_storage(ALICE, U256::from(0x10), U256::from(0x00));
+    recorder.record_storage_write(ALICE, U256::from(0x10), U256::from(0x01));
+
+    // Inner call B: write same slot (reverts)
+    let checkpoint = recorder.checkpoint();
+    recorder.capture_pre_storage(ALICE, U256::from(0x10), U256::from(0x01));
+    recorder.record_storage_write(ALICE, U256::from(0x10), U256::from(0x99));
+    recorder.restore(checkpoint);
+
+    let bal = recorder.build();
+
+    let alice = bal.accounts().iter().find(|a| a.address == ALICE).unwrap();
+    assert_eq!(alice.storage_changes.len(), 1);
+    assert_eq!(alice.storage_changes[0].slot, U256::from(0x10));
+    assert!(alice.storage_reads.is_empty());
+    bal.validate_ordering().unwrap();
+}
+
+/// Net-zero write in tx 1 (filtered to read), then written+reverted in tx 2.
+/// Slot should stay in storage_reads only.
+#[test]
+fn test_net_zero_then_reverted_write_stays_read() {
+    let mut recorder = BlockAccessListRecorder::new();
+
+    // Tx 1: write slot 0x10 with net-zero (write 0x01 then back to 0x00)
+    recorder.set_block_access_index(1);
+    recorder.capture_pre_storage(ALICE, U256::from(0x10), U256::from(0x00));
+    recorder.record_storage_write(ALICE, U256::from(0x10), U256::from(0x01));
+    recorder.record_storage_write(ALICE, U256::from(0x10), U256::from(0x00));
+
+    // Transition to tx 2 (triggers net-zero filtering for tx 1)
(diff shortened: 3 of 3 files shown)


crates/networking/p2p/types.rs
```diff
@@ -284,38 +276,42 @@ pub struct NodeRecordPairs {
...
+    // Snap entry is being used by some tests such as `test_encode_enr_response`.
+    pub snap: Option<Vec<u32>>,
     // TODO implement ipv6 specific ports
 }
 
-impl NodeRecord {
-    pub fn decode_pairs(&self) -> NodeRecordPairs {
+impl NodeRecordPairs {
+    pub fn try_from_raw_pairs(
+        pairs: &Vec<(Bytes, Bytes)>,
```
Oppen: Why not:
```suggestion
        pairs: &[(Bytes, Bytes)],
```

crates/networking/p2p/types.rs
```diff
@@ -259,17 +259,9 @@ impl Display for Node {
...
-    pub seq: u64,
-    // holds optional values in (key, value) format
-    // value represents the rlp encoded bytes
-    // The key/value pairs must be sorted by key and must be unique
-    pub pairs: Vec<(Bytes, Bytes)>,
-}
-
-#[derive(Debug, Default, PartialEq)]
+/// Holds optional values in (key, value) format, value represents the rlp encoded bytes
+/// The key/value pairs must be sorted by key and must be unique
```
Oppen: Is this comment currently accurate? During the life of the struct, it seems like a regular struct with optional fields and no meaning to their order. I think the comment is more about the encoded form than the type.

lakshya-sky: You're right, should I move update and move it to encode method?

crates/networking/p2p/types.rs
lakshya-sky: no reason, just used the old decode_pairs which accepts vec. But your suggestion makes sense.

crates/networking/p2p/types.rs
```diff
@@ -284,38 +276,40 @@ pub struct NodeRecordPairs {
...
                     if bytes.len() < 33 {
-                        continue;
+                        return Err(RLPDecodeError::Custom(format!(
+                            "Invalid signature length {}",
+                            bytes.len()
+                        )));
                     }
                     decoded_pairs.secp256k1 = Some(H264::from_slice(&bytes))
                 }
+                "snap" => decoded_pairs.snap = Some(Vec::<u32>::decode(&value)?),
```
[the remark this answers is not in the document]
Oppen: Strictly speaking, the old version sounds like a bug. Fields are allowed to be absent, but they should not be allowed to be present and garbage.

crates/networking/p2p/types.rs
```diff
@@ -284,38 +276,42 @@ pub struct NodeRecordPairs {
...
-impl NodeRecord {
-    pub fn decode_pairs(&self) -> NodeRecordPairs {
+impl NodeRecordPairs {
+    pub fn try_from_raw_pairs(
+        pairs: &Vec<(Bytes, Bytes)>,
+    ) -> Result<NodeRecordPairs, RLPDecodeError> {
         let mut decoded_pairs = NodeRecordPairs::default();
-        for (key, value) in &self.pairs {
+        for (key, value) in pairs {
             let Ok(key) = String::from_utf8(key.to_vec()) else {
```
Oppen: Although the same probably can be said about this. If it's not valid UTF8 should we still take the packet?

Oppen: Actually, according to the spec:
> The key/value pairs must be sorted by key and must be unique, i.e. any key may be present only once. The keys can technically be any byte sequence, but ASCII text is preferred.

So, if we don't understand them we shall ignore them. That said, the UTF parsing is unnecessary, we can just use the slice and `b"..."` syntax for writing the match expression.

crates/networking/p2p/types.rs
```diff
@@ -284,38 +276,42 @@ pub struct NodeRecordPairs {
...
 }
 
-impl NodeRecord {
-    pub fn decode_pairs(&self) -> NodeRecordPairs {
+impl NodeRecordPairs {
+    pub fn try_from_raw_pairs(
+        pairs: &Vec<(Bytes, Bytes)>,
+    ) -> Result<NodeRecordPairs, RLPDecodeError> {
         let mut decoded_pairs = NodeRecordPairs::default();
-        for (key, value) in &self.pairs {
```
Oppen, line 297: Shouldn't we check order and uniqueness as well?

lakshya-sky:
Let me work on those suggestions.

crates/networking/p2p/types.rs
Oppen: Yes, I think it would make more sense there.

Oppen:
Looks good in general. I'll give it my 👍 as soon as I find the time to test it a bit, but if anyone else on the team beats me to it don't wait on me.

crates/networking/p2p/types.rs
```diff
@@ -284,52 +274,105 @@ pub struct NodeRecordPairs {
...
+                    let bytes = Bytes::decode(value)?;
                     if bytes.len() < 33 {
-                        continue;
+                        return Err(RLPDecodeError::Custom(format!(
+                            "Invalid secp256k1 public key length: expected at least 33 bytes, got {}",
+                            bytes.len()
+                        )));
                     }
                     decoded_pairs.secp256k1 = Some(H264::from_slice(&bytes))
                 }
```
Oppen: If you want, this can be simplified as follows:
```suggestion
                b"secp256k1" => decoded_pairs.secp256k1 = Some(H264([u8; 33]::decode(value)?)),
```

Oppen:
Did a partial sync. Let it run until storages were all downloaded without issue. Then I forgot and suspended the machine for the night, but I think it's enough of a smoke test for the code this touches.

crates/networking/p2p/types.rs
```diff
@@ -404,42 +431,19 @@ impl NodeRecord {
...
             .finish();
         keccak_hash(&rlp)
     }
+
+    pub fn pairs(&self) -> &NodeRecordPairs {
+        &self.pairs
+    }
 }
 
 impl From<NodeRecordPairs> for Vec<(Bytes, Bytes)> {
```
Oppen: This `From` implementation appears unused. If I delete it, the project still builds.
The reason I mention this is that this is the only non-codec function that uses `encode_pairs`, so maybe we can encode in place rather than allocating the vec of vecs?

lakshya-sky: Hey @Oppen, I wasn't sure how to exactly remove allocation but I tried something new which removes allocation altogether. Please have a look at [5973f67957cceaa192ea6715ae1ad0e5d6ba7efe](https://github.com/lambdaclass/ethrex/pull/5546/changes/5973f67957cceaa192ea6715ae1ad0e5d6ba7efe)

I forgot to try using `b"..."` string instead of `Bytes::from()`, once I have green-light I will change those as well.

lakshya-sky:
Hey @Oppen, any idea when this can get merged?

edg-l (approved):

crates/networking/p2p/types.rs
```diff
@@ -188,7 +188,7 @@ impl Node {
     }
 
     pub fn from_enr(record: &NodeRecord) -> Result<Self, NodeError> {
-        let pairs = record.decode_pairs();
+        let pairs = record.pairs();
```
ElFantasma, line 191: This `decode_pairs()` → `pairs()` migration is correct here, but the same change is missing in `discv5/`:

- `discv5/server.rs:307` — `record.decode_pairs()` → `record.pairs()`
- `discv5/messages.rs:1209` — `record.decode_pairs()` → `record.pairs()` (test)
- `discv5/messages.rs:1503` — `record.decode_pairs()` → `record.pairs()` (test)
- `discv5/peer_table.rs:889` — `record.decode_pairs().eth` → `record.get_fork_id().cloned()` (same pattern as the discv4 fix at line 549)

Also, two test sites in `discv5/messages.rs` construct `NodeRecord` via struct literal with `pairs: NodeRecordPairs{...}.into()` (lines ~1300 and ~1589), but `From<NodeRecordPairs> for Vec<(Bytes, Bytes)>` was removed and `pairs` is now private. These need `NodeRecord::new(signature, seq, pairs)` with `NodeRecordPairs` directly.

crates/networking/p2p/types.rs
```diff
@@ -408,15 +439,15 @@ impl NodeRecord {
         let mut rlp = vec![];
         structs::Encoder::new(&mut rlp)
             .encode_field(&self.seq)
-            .encode_key_value_list::<Bytes>(&self.pairs)
+            .encode_with(|buf| self.pairs.encode(buf))
```
ElFantasma: This re-encodes the record content from typed `NodeRecordPairs` fields, but `try_from_raw_pairs` (line 287) drops unknown ENR keys via `_ => {}`. This breaks the decode → encode roundtrip:

1. Peer sends ENR with non-standard keys (e.g., `eth2`, `attnets`, custom keys)
2. `NodeRecord::decode()` → `try_from_raw_pairs()` → unknown keys silently dropped
3. `verify_signature()` → `get_signature_digest()` → encodes from typed fields only → **different content than original** → signature verification fails

On `main` this works because `get_signature_digest()` encodes from raw `Vec<(Bytes, Bytes)>` which preserves all keys.

Fix: keep the raw pairs for encoding/signature verification. The typed accessors (`get_fork_id()`, `pairs()`) can decode on top. A `OnceLock<NodeRecordPairs>` cache would be natural but doesn't implement `Serialize`/`Deserialize` (which `NodeRecord` derives), so the simplest approach is to keep raw pairs as the source of truth and add typed accessor methods (like `get_fork_id()`) that decode on the fly — similar to `main`'s `decode_pairs()` but with the strict parsing this PR introduces.

crates/networking/p2p/types.rs
```diff
@@ -347,48 +385,41 @@ impl NodeRecord {
...
+        self.pairs.eth = Some(fork_id);
+        self.update(signer)
+    }
 
+    pub fn get_fork_id(&self) -> Option<&ForkId> {
+        self.pairs.eth.as_ref()
+    }
+
+    fn update(&mut self, signer: &SecretKey) -> Result<(), NodeError> {
+        self.seq += 1;
```
ElFantasma, line 427: `update()` increments `self.seq` — this is a behavioral change. Previously `set_fork_id` didn't touch `seq`, now it bumps it by 1. So `from_node(seq=1)` followed by `set_fork_id()` produces `seq=2` instead of `seq=1`.

This is arguably correct per [EIP-778](https://eips.ethereum.org/EIPS/eip-778) ("The sequence number SHOULD be increased whenever the record changes"), but has two consequences worth verifying:

1. **Test at line 670-672:** `set_fork_id` already signs via `update()`, so the explicit `record.sign_record(&signer)` on line 672 is redundant (signs the same content a second time). It's not wrong, just unnecessary — consider removing it to avoid confusion.

2. **Production callers:** Both `discv4/server.rs:113` and `discv5/server.rs:138` call `from_node(seq=1)` then `set_fork_id()`. After this PR, the initial ENR will have `seq=2` instead of `seq=1`. Make sure no peer comparisons or ENR freshness checks depend on the initial seq being exactly 1.

crates/networking/p2p/types.rs
```diff
@@ -288,52 +278,100 @@ pub struct NodeRecordPairs {
...
             }
         }
 
-        decoded_pairs
+        Ok(decoded_pairs)
+    }
+
+    /// Encodes to a list of (key, value) where keys are ascii bytes and values are rlp encoded bytes.
+    fn encode(&self, buf: &mut dyn BufMut) {
+        if let Some(eth) = self.eth.clone() {
```
ElFantasma: nit: `self.eth.clone()` clones the `Option<ForkId>` to destructure it. You can avoid the clone with a reference pattern:
```rust
if let Some(eth) = &self.eth {
    let eth = vec![eth.clone()];
    ...
}
```
Still needs one `clone()` for the `vec!`, but avoids the intermediate `Option` clone.

ElFantasma (changes requested):
Good refactor direction — typed `NodeRecordPairs` is much more ergonomic than raw `Vec<(Bytes, Bytes)>`, and the strict decoding via `try_from_raw_pairs` is correct per ENR spec.

Two blockers:

1. **discv5 won't compile.** During the time it took us to review this PR, we implemented ´discv5´. The PR removes `decode_pairs()` from `NodeRecord` (replaced by `pairs()`) and removes `From<NodeRecordPairs> for Vec<(Bytes, Bytes)>`, but 4 call sites in `discv5/` still use the old API — see inline comment on types.rs:191. 

2. **Lossy roundtrip breaks signature verification.** See inline comment on types.rs:442.

crates/networking/p2p/types.rs
lakshya-sky: Hey do you want to change it in this PR or I can create a different PR for discv5 specific change?

ElFantasma: We will have to update in this PR or discv5 won't compile when merging this.

crates/networking/p2p/types.rs
lakshya-sky: Fixed

crates/networking/p2p/types.rs
lakshya-sky: Fixed

crates/networking/p2p/types.rs
lakshya-sky: I have added an extra field `other` to hold all the other keys we don't care about. When signature verification is required it will be encoded as well to make sure it works. And a new unit test for this scenario https://github.com/lambdaclass/ethrex/pull/5546/commits/457579ff41833135795a151593d84710640bd5aa#diff-44fa7b327f9379733c50bc8196a0fd993a533e38b5ff57d7ac55de7980590957R735

crates/networking/p2p/types.rs
iovoid: Doesn't seem to be resolved

crates/networking/p2p/types.rs
iovoid: Agreed

PR #5548, opened 2025-12-07
fix(l1): skip fork-choice update if snap-syncing

### LINKED ISSUE #5547: Avoid applying fork choice during snap-sync
We're currently calling `apply_fork_choice` in our `forkChoiceUpdate` handler during a snap-sync. We should instead return a "SYNCING" early.

We already had this functionality, but it was commented out in #5196:

https://github.com/lambdaclass/ethrex/blob/1bebff3c049e767fc9974eff551d066ad0935761/crates/networking/rpc/engine/fork_choice.rs#L222-L237

MegaRedHand:
**Motivation**

We recently found some errors due to payloads being marked as invalid when FCU is triggered during a snap-sync.

**Description**

This PR uncomments an early return that handles these cases.

Closes #5547

### DIFF
--- a/crates/networking/rpc/engine/fork_choice.rs
+++ b/crates/networking/rpc/engine/fork_choice.rs
@@ -221,17 +221,8 @@ async fn handle_forkchoice(
     }
 
     if syncer.sync_mode() == SyncMode::Snap {
-        // Don't trigger a sync if the block is already canonical
-        if context
-            .storage
-            .is_canonical_sync(fork_choice_state.head_block_hash)?
-        {
-            // Disable snapsync mode so we can process incoming payloads
-            syncer.disable_snap();
-        } else {
-            syncer.sync_to_head(fork_choice_state.head_block_hash);
-            return Ok((None, PayloadStatus::syncing().into()));
-        }
+        syncer.sync_to_head(fork_choice_state.head_block_hash);
+        return Ok((None, PayloadStatus::syncing().into()));
     }
 
     match apply_fork_choice(


MegaRedHand:
This doesn't fully fix the problem, since an FCU could be received for a block before the last snap-sync pivot, after snap-sync was finished

MegaRedHand:
Reopening since we identified the issues are independent

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -219,22 +220,19 @@ async fn handle_forkchoice(
...
-    if context.syncer.sync_mode() == SyncMode::Snap {
+    if syncer.sync_mode() == SyncMode::Snap {
         // Don't trigger a sync if the block is already canonical
         if context
             .storage
             .is_canonical_sync(fork_choice_state.head_block_hash)?
         {
             // Disable snapsync mode so we can process incoming payloads
-            context.syncer.disable_snap();
+            syncer.disable_snap();
```
fedacking: I am worried about this scenario. If this is happening after snap sync, we shouldn't be seeing this message because we changed the mode. If this happening during snap sync, we should know why is the head_block_hash in the canonical list, and returning early.

MegaRedHand: I agree. Maybe we can disable the inner `if`.

MegaRedHand: Removed in [f21cada](https://github.com/lambdaclass/ethrex/pull/5548/commits/f21cadac4ddddd39f372cc70d93edf9ad115734f). Let's see if hive fails, which was the only reason for that `if` lol

fedacking (approved):

PR #5552, opened 2025-12-08
fix(l1): revert "reduce spammy trace-level logs during snap-sync (#5492)"

rodrigo-o:
This reverts commit 610df53674f307c7d9e2259e45cb52cc1c2bcbf9.

**Motivation**

The reverted PR appear to have produced a regression in sepolia snapsyncing.

**Description**

Probably due to the added sleep, this PR appears to be causing a regresion during snapsync in sepolia. We are right now testing this and will open this PR for review when confirmed.


rodrigo-o:
We validated that the issue we seen was not related to this PR, here is the corresponding issue: #5558

PR #5554, opened 2025-12-09
feat(l2): introduce EncodedTrie as a zkVM performant MPT

xqft:
## Performance Results

All benchmarks run on **mainnet block 24284399** (256 transactions, 25.5M gas).

### SP1

| Metric | main | EncodedTrie | Improvement |
|--------|------|-------------|-------------|
| **Total Cycles** | **601.4M** | **525.9M** | **-12.6%** |
| read_input | 106.7M | 17.0M | -84.1% |
| guest_program_state_init | 108.1M | 40.6M | -62.5% |
| execute_block | 313.7M | 349.9M | +11.5% |
| apply_account_updates | 35.1M | 46.6M | +32.9% |
| get_final_state_root | 20.2M | 30.9M | +52.8% |

### ZisK

| Metric | main | EncodedTrie | Improvement |
|--------|------|-------------|-------------|
| **Total Steps** | **531.4M** | **496.2M** | **-6.6%** |
| Total Cost | 61.0B | 57.4B | -5.9% |
| memcpy calls | 904K | 714K | -21.0% |
| memcpy steps | 135.2M | 114.0M | -15.7% |

## Motivation

zkVM environments have different performance characteristics than traditional VMs - cycles matter more than wall-clock time, and memory operations (particularly memcpy) are expensive. The current `Trie` implementation wasn't designed with these constraints in mind, leading to significant overhead from:
- Recursive node structures requiring serialization/deserialization
- Repeated RLP encoding during hash calculations
- Excessive cloning and memory copying in hot paths

## Description

This PR introduces `EncodedTrie`, a Merkle Patricia Trie implementation optimized for zkVM environments. The key optimizations are:

1. **Contiguous byte buffer storage**: All initial state nodes are stored RLP-encoded in a single contiguous `Vec<u8>`, referenced by ranges. This avoids repeated deserialization.

2. **Index-based node references**: Nodes are indexed by integers instead of recursive pointers, eliminating reference overhead.

3. **Separated structural information**: The `NodeType` enum stores structural info separately from encoded data, allowing efficient traversal without parsing.

4. **Lazy encoding**: New nodes and mutations are stored as overrides; RLP encoding only happens when computing the final root hash.

5. **Authentication vs Hashing distinction**: Separates initial state verification (authentication) from final state computation (hashing), avoiding redundant work.

### Analysis

The EncodedTrie trades slightly higher execution overhead for dramatically lower serialization costs:

- **SP1**: ~157M cycles saved in deserialization (`read_input` + `state_init`) vs ~58M cycles added during execution. Net: **75.5M cycles saved**.
- **ZisK**: 21M fewer memcpy steps from avoiding node cloning during trie traversal.

The serialization efficiency gains (pre-encoded nodes in a contiguous buffer, zero-copy reads) significantly outweigh the execution overhead from the new trie structure.

### Files Changed

- `crates/common/trie/encodedtrie.rs` - New `EncodedTrie` implementation
- `crates/common/trie/node.rs` - Node type modifications for pruned hash storage
- Integration with execution witness generation for L2 prover

## Checklist

### DIFF
--- a/crates/common/trie/encodedtrie.rs
+++ b/crates/common/trie/encodedtrie.rs
@@ -1329,12 +1329,14 @@ mod test {
         fn proptest_multi_round_updates((kv, _shuffle) in kv_pairs_strategy()) {
             // Test scenario: simulate multi-block execution with updates to same keys
             // This is like what happens with the beacon root contract across multiple blocks
-            if kv.len() < 4 {
+            if kv.len() < 8 {
                 return Ok(());
             }
 
             // Use first quarter for initial state
             let initial = &kv[..kv.len()/4];
+            // Reserve remaining keys for "new account" insertions (same length as initial keys)
+            let remaining = &kv[kv.len()/4..];
 
             // Create trie with initial data
             let mut trie = Trie::new_temp();
@@ -1362,12 +1364,13 @@ mod test {
                     flat_trie.insert(key.clone(), new_value).unwrap();
                 }
 
-                // Insert a new key (simulating new account)
-                if block < kv.len() / 4 {
-                    let new_key = format!("newkey_block{}", block).into_bytes();
+                // Insert a new key from remaining keys (same length, simulating new account)
+                // In Ethereum, all keys are 32 bytes (keccak hashes), so we use same-length keys
+                if block < remaining.len() {
+                    let (new_key, _) = &remaining[block];
                     let new_value = format!("newvalue_block{}", block).into_bytes();
                     trie.insert(new_key.clone(), new_value.clone()).unwrap();
-                    flat_trie.insert(new_key, new_value).unwrap();
+                    flat_trie.insert(new_key.clone(), new_value).unwrap();
                 }
 
                 // Compute hash after each "block"


crates/common/trie/flattrie.rs
```diff
@@ -0,0 +1,253 @@
...
+    serde::Deserialize,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+    rkyv::Archive,
+)]
+pub enum NodeChilds {
+    /// A leaf node doesn't have any childs
+    Leaf,
+    /// An extension node always has a branch as a child
+    Extension { child: Option<usize> },
```
Oppen: Why is the child an option, if it's guaranteed to exist and be a branch?

crates/common/trie/flattrie.rs
```diff
@@ -0,0 +1,253 @@
...
+    Clone,
+    Copy,
+    serde::Serialize,
+    serde::Deserialize,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+    rkyv::Archive,
+)]
+pub enum NodeChilds {
+    /// A leaf node doesn't have any childs
```
Oppen:
```suggestion
    /// A leaf node doesn't have any children
```

crates/common/trie/flattrie.rs
```diff
@@ -0,0 +1,253 @@
...
+                        }
+                    }
+                    NodeChilds::Branch { childs }
+                }
+                Node::Extension(node) => {
+                    let mut child = None;
+                    if let NodeRef::Node(child_node, _) = &node.child {
+                        recursive(child_node, trie);
+                        child = Some(trie.views.len() - 1);
+                    }
```
Oppen: How can we tell when a given child is unpopulated from it just being in hashed form?

crates/common/trie/flattrie.rs
```diff
@@ -0,0 +1,253 @@
...
+                    NodeChilds::Extension { child }
+                }
+                Node::Leaf(_) => NodeChilds::Leaf,
+            };
+
+            let offset = trie.data.len();
+            trie.data.extend(value.encode_to_vec());
```
Oppen: It's probably better to do:
```suggestion
            value.encode(&mut trie.data);
```
That should save a memcpy and a second realloc.

crates/common/trie/flattrie.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.data.is_empty()
+    }
+
+    /// Recursively traverses the trie, hashes each node's data and checks that their parents reference
+    /// those same hashes. This way the data gets authenticated.
+    ///
+    /// Initializes `Self::root_hash` if successful, and returns a boolean indicating success.
```
Oppen: Shouldn't auth failure be an error?

crates/common/types/block_execution_witness.rs
```diff
@@ -318,6 +335,14 @@ impl GuestProgramState {
         &mut self,
```
ElFantasma: This `SYSTEM_ADDRESS` constant is already defined in `ethrex_vm::system_contracts::SYSTEM_ADDRESS`. Duplicating it risks drift if the value ever changes (unlikely, but still). Can you import it instead?

Also, this entire early-return block for system address is new behavior — it silently returns `Ok(None)` instead of looking up the state trie. What happens if the system address legitimately has state (e.g., balance)? Is this a workaround for a missing key in the execution witness? If so, a comment explaining the rationale would be helpful.

crates/common/types/block_execution_witness.rs
```diff
@@ -171,6 +174,7 @@ impl TryFrom<ExecutionWitness> for GuestProgramState {
             chain_config: value.chain_config,
```
ElFantasma: The `state_trie` obtained from the witness is used directly without calling `authenticate()` here. Authentication happens later in `auth_state()` called from guest execution. Is there a risk that a corrupted/malicious witness trie gets used between construction and authentication? If execution reads from the trie before `auth_state()` is called, those reads are unauthenticated.

crates/common/trie/encodedtrie.rs
```diff
@@ -0,0 +1,1416 @@
...
+    pub fn remove(&mut self, path: &[u8]) -> Result<(), EncodedTrieError> {
+        self.ensure_hashes_initialized();
+        let path = Nibbles::from_bytes(path);
+        if let Some(root_index) = self.root_index {
+            self.root_index = self.remove_inner(root_index, path)?;
+        }
+        Ok(())
+    }
+
+    pub fn remove_inner(
```
ElFantasma, line 355: `remove_inner` is `pub` but it's a recursive implementation detail — callers should use `remove()`. Should this be `fn` (private) like `insert_inner`?

crates/common/trie/encodedtrie.rs
```diff
@@ -0,0 +1,1416 @@
...
+                // RLP encode 32-byte string directly: 0xa0 prefix + 32 bytes
+                // Avoids RLPEncode trait call overhead
+                buf.put_u8(0xa0);
+                buf.extend_from_slice(&hash.0);
+            }
+            NodeHash::Inline((_, 0)) => buf.put_u8(RLP_NULL),
+            NodeHash::Inline((encoded, len)) => buf.put_slice(&encoded[..*len as usize]),
+        }
+    }
+    buf.put_u8(RLP_NULL);
```
ElFantasma, line 1014: `encode_branch` always appends `RLP_NULL` as the 17th element (branch value). This is correct for the Ethereum state/storage tries where values only live at leaves, but it's a hidden invariant. A brief comment like `// Branch value is always empty for state/storage tries` would prevent confusion if someone tries to use EncodedTrie in a different context.

crates/common/types/block_execution_witness.rs
```diff
@@ -318,6 +335,14 @@ impl GuestProgramState {
         &mut self,
         address: Address,
     ) -> Result<Option<AccountState>, GuestProgramStateError> {
+        const SYSTEM_ADDRESS: H160 = H160([
```
ElFantasma, line 338: This SYSTEM_ADDRESS constant is already defined in ethrex_vm::system_contracts::SYSTEM_ADDRESS. Duplicating it risks drift. Since ethrex-common cannot import from ethrex-vm (circular dep), consider moving the constant to ethrex-common so both crates reference the same definition.

Also, a brief comment on why the system address early-returns Ok(None) would help — presumably because the system address (0xfff...ffe) is used for system calls and typically has no real account state in the witness trie.

crates/blockchain/blockchain.rs
```diff
@@ -1854,19 +1868,17 @@ impl Blockchain {
         transaction: EIP4844Transaction,
         blobs_bundle: BlobsBundle,
     ) -> Result<H256, MempoolError> {
+        // Validate blobs bundle
+
         let fork = self.current_fork().await?;
 
+        blobs_bundle.validate(&transaction, fork)?;
```
ElFantasma, line 1875: The blob validation reorder is a semantic change: previously, known transactions skipped expensive validation. Now validation runs unconditionally. If intentional, it should be called out in the PR description.

PR #5559, opened 2025-12-09
fix(l1): error on missing parent

iovoid:
**Motivation**

Currently if the state trie root does not match with the one being used for execution, we fail due to various errors that happen as a result of executing from incorrect state (invalid nonces, incorrect gas, state root mismatch) instead of alerting about the real error.

While this should only happen due to a bug, it has happened several times.

**Description**

Errors if the state root being used for execution isn't found.

**Checklist**

- [X] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync.


crates/blockchain/vm.rs
```diff
@@ -24,26 +24,38 @@ pub struct StoreVmDatabase {
 }
 
 impl StoreVmDatabase {
-    pub fn new(store: Store, block_header: BlockHeader) -> Self {
-        StoreVmDatabase {
+    pub fn new(store: Store, block_header: BlockHeader) -> Result<Self, EvmError> {
+        if !store
+            .has_state_root(block_header.state_root)
+            .map_err(|e| EvmError::DB(e.to_string()))?
```
MegaRedHand, line 34: We should add a comment here about why we check this.

crates/blockchain/vm.rs
```diff
@@ -24,26 +24,38 @@ pub struct StoreVmDatabase {
...
     pub fn new_with_block_hash_cache(
         store: Store,
         block_header: BlockHeader,
         block_hash_cache: BTreeMap<BlockNumber, BlockHash>,
-    ) -> Self {
-        StoreVmDatabase {
+    ) -> Result<Self, EvmError> {
+        if !store
+            .has_state_root(block_header.state_root)
+            .map_err(|e| EvmError::DB(e.to_string()))?
```
MegaRedHand, line 54: Same here

MegaRedHand (approved):
Great improvement!

crates/blockchain/vm.rs
iovoid: Added comments in 43889736d

crates/blockchain/vm.rs
iovoid: Added comments in 43889736d

Oppen (approved):

tooling/ef_tests/state/utils.rs
```diff
@@ -20,10 +20,8 @@ pub async fn load_initial_state_revm(test: &EFTest) -> (RevmState, H256, Store)
     let mut storage = Store::new("./temp", EngineType::InMemory).expect("Failed to create Store");
     storage.add_initial_state(genesis.clone()).await.unwrap();
 
-    let vm_db: DynVmDatabase = Box::new(StoreVmDatabase::new(
-        storage.clone(),
-        genesis.get_block().header,
-    ));
+    let vm_db: DynVmDatabase =
+        Box::new(StoreVmDatabase::new(storage.clone(), genesis.get_block().header).unwrap());
```
rodrigo-o, line 24: All unwraps are just in tooling/bench code right?

iovoid: Yes.

Note: `crates/vm/levm/runner/` is also tooling even if the path might imply otherwise.

iovoid: Created issue (#5610) to move them so we can avoid confusion.

PR #5560, opened 2025-12-09
docs(l1): snapsync docs

fedacking:
**Motivation**

We want to add documentation to Snap Sync and concerns that should be taken into account for a rewrite.

**Description**

- Adds two markdown files detailing the implementation and concerns with snap sync. 
- Deletes the old documentation files that don't apply anymore.


docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+To solve the second problem, we introduce an optimization, which is called the "Membatch". This allows us to maintain a new invariant:
+
+> If a node is present in the DB, then that node and all its children must be present.
+
+This removes the need to explore entire subtrees: when walking down the trie, if a node is in the DB, the whole subtree can be skipped.
+
+To maintain this invariant, we do the following:
+
+- When we get a new node, we don't immediately store it in the database. We keep track of the amount of every node's children that are not yet in the database. As long as it's not zero, we keep it in a separate in-memory structure "Membatch" instead of on the db.
```
jrchatruc, line 49: This is no longer how membatch is implemented. This was our implementation when the db was hash based, now that it's path based what we do is:

- Whenever we download a new node, we delete its parents from the database

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+[^1]: We update pivots based only on the timestamp, not on the peer response. According to the spec, stale roots return empty responses, but Byzantine peers may return empty responses at arbitrary times. Therefore, we rely on the rule that nodes must keep at least 128 blocks. Once `time > timestamp + 128 * 12` we mark the pivot as stale.
+
+Example of a possible state after stoping fast sync due to staleness.
+
+![Fast Sync Retaking Example - 1](snap_sync/Fast%20Sync%20Retaking%20Example%20-%201.png)
+
+In the example, even if we find that node { hash: 0x317f, path: 0 } is correct, we still need to check all its children in the DB (in this case none are present).
+
+To solve the second problem, we introduce an optimization, which is called the "Membatch". This allows us to maintain a new invariant:
```
jrchatruc: Explain here briefly why we call it membatch and reference geth (as it's an idea taken from their codebase); the name is very obscure and has nothing to do with what it actually means.

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+Example of a possible state after stoping fast sync due to staleness with membatch.
+
+![Fast Sync Membatch Example - 1](snap_sync/Fast%20Sync%20Membatch%20Example%20-%201.png)
+
+### Speeding up: Snap-Sync
+
+Fast-sync is slow (ethrex fast-sync of a fresh “hoodi” state takes ~45 minutes—4.5× slower than snap-sync). To accelerate it, we use a key property of fast-sync:
+> Fast-sync can “heal” any partial trie that obeys the invariant, even if that trie is not consistent with any real on-chain state.
+
+Snap-sync exploits this:
```
jrchatruc: > Snap-sync exploits this:

Snap sync exploits this by:

- downloading only the leaves (the accounts and storage slots) from any recent state
- assembling a trie from these leaves
- running fast-sync (“healing”) to repair it into a consistent trie.

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+- download only the leaves (the accounts and storage slots) from any recent state,
+- assemble a trie from these leaves,
+- run fast-sync (“healing”) to repair it into a consistent trie.
+
+For our code, we call the fast-sync the healing step.
```
jrchatruc: > For our code, we call the fast-sync the healing step.

On our code, we call the fast-sync step "healing".

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+Generalized Flowchart of snapsync
+
+![Flow - Snap Sync](snap_sync/Flow%20-%20Snap%20Sync.png)
+
+### Flags
+
+When testing snap sync there are flags to take into account:
+
+- If the `SKIP_START_SNAP_SYNC` environment variable is set and isn't empty, it will skip the step of downloading the leaves and will immediatly begin healing. This simulates the behaviour of fast-sync.
```
jrchatruc: > This simulates the behaviour of fast-sync

This simulates the behaviour of fast-sync and is used for testing.

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+- If debug assertions are on, the program will validate that the entire state and storage tries are valid by traversing the entire trie and recomputing the roots. If any is found to be wrong, it will print an error and exit the program.
```
jrchatruc: > - If debug assertions are on, the program will validate that the entire state and storage tries are valid by traversing the entire trie and recomputing the roots. If any is found to be wrong, it will print an error and exit the program.

I'd add here something like "this is also used for testing purposes; a validation error here means there is a 100% a bug in snap sync."

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+- We need to know which accounts have storage that needs to be downloaded
+- We need to know what storage root each account has to be able to verify it
+
+To solve these issues we take two action:
+- Before we download the storage slots we ensure that the state trie is in a consistent complete state. This is accomplished by doing the insertion of accounts step first and then healing the trie. If during the storage slot download the pivot becomes stale, we heal the trie again with the new pivot, to keep the trie up to date.
+- When inserting the accounts, we grab a list of all the accounts with their storage root. If the account is healed, we marked the storage root as `None`, to indicate we should check in the DB what is the state of the storage root.
+
+#### The time traveling problem
+
+During the development of snap sync we found a recurring problem, the time traveling problem. In hash based, when a node returned to a previous state, those accounts wouldn't be redownloaded, as it would just return to the previous hash which was present in the database.
```
jrchatruc: I would rephrase this paragraph to the following:

During snap sync development, we kept running into a problematic edge case: what we call "time traveling". This is when a certain account is in state `A` at a certain pivot, then changes to `B` in the next pivot, then **goes back to A** on a subsequent pivot change. This was a huge source of problems because it broke an invariant we assumed to be true, namely:

- If an account changes its state on pivot change, we will encounter it during state healing. This is NOT true if the trie is hash based.

The reason for this is the time traveling scenario. When an account goes back to a state we already have on our database, we do not heal it (because we already have it), even though its state has changed. This means the code won't realize the account has changed its storage root and won't update it.

This should not be a problem in a path based trie (which we currently have), but it's important to keep it in mind and make sure that whatever code we write keeps this time traveling edge case in mind and solves it.

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+#### Repeated Storage Roots
+
+A large amount of the accounts with storage have exactly the same storage as other accounts.[^5] As such, when we are creating tasks for download, it's important to group the tasks by storage root and not download them twice.
+
+[^5]: This may be for a variety of reasons, but the most likely is ERC20 tokens that were deployed and never used.
+
+#### Big Accounts
+
+The storage trie is very unneven distribution of the accounts sizes. Between accounts with a single or two storage slots are around 70% of all accounts with storage trie. And large accounts have more storage slots than accounts slots are present in the state accounts. As such they need to be downloaded with special consideration.
```
jrchatruc: I would reword it to this:

Storage trie sizes have a very uneven distribution. Around 70% of all ethereum mainnet contracts have only 1 or 2 storage slots. However, a few contracts have more storage slots than all account leaves in the entire state trie. As such, the code needs to take this pareto distribution into account to download storage tries fast.

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+
+Currently, if ethrex has been downloading storages for more than 2 pivots, the node will stop trying to download storage, and fallback to heal (fast sync) all the storage accounts that were still missing downloads. This stops ethrex hanging due to a problem but it indicates that we still have bugs in our storage slots download.
+
+### Downloading Bytecodes
+
+Whenever an account is download or healed we check if the code is not empty. If it isn't, we store it for future download. This is added to a list, and when the list grows beyond a certain size it is written to disk. After the healing is done and we have a complete state and storage tree, we start with the download of bytecodes, chunking them to avoid memory overflow.
+
+### Forkchoice update
+
+Once the entire files are downloaded, we change the snap sync mode to full, and we do an `apply_forkchoice` to mark that as the last pivot as the last block.
```
jrchatruc: > Once the entire files are downloaded, we change the snap sync mode to full...

Once the entire state trie, all storage tries and contract bytecodes are downloaded, we switch the sync mode from `snap` to `full`...

docs/l1/fundamentals/snap_sync_concerns.md
```diff
@@ -0,0 +1,39 @@
+# Snap Sync Concerns
+
+## Known Bugs
```
jrchatruc: These are not bugs, but rather code that needs to be improved; instead of "Known Bugs" let's call it something like "Code Improvement opportunities"

Let's also mention before this that all of these items should be taken into account when we rewrite our snap sync code

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,321 @@
...
+}
+```
+
+#### Big Accounts Flow
+
+![Big Account logic](snap_sync/Flow%20-%20Big%20Account%20Logic.png)
+
+#### Retry Limit
+
+Currently, if ethrex has been downloading storages for more than 2 pivots, the node will stop trying to download storage, and fallback to heal (fast sync) all the storage accounts that were still missing downloads. This stops ethrex hanging due to a problem but it indicates that we still have bugs in our storage slots download.
```
jrchatruc: > This stops ethrex hanging due to a problem but it indicates that we still have bugs in our storage slots download.

This prevents snap sync from hanging due to an edge case we do not currently handle if an account time travels. See the "snap sync concerns" document for details on what this edge case is.

docs/l1/fundamentals/snap_sync_concerns.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Storage downloads
+
+When downloading storages, there are scenarios when the storages are never finished downloading. The most likely explanation is that big accounts intervals aren't properly redownloaded.
```
jrchatruc: > When downloading storages, there are scenarios when the storages are never finished downloading. The most likely explanation is that big accounts intervals aren't properly redownloaded. 

When downloading storages, there's a possibility that storage leaves never finish downloading. This can happen if an account time travels (i.e. if it goes from state `A` to state `B` on a pivot change, then back to state `A` on a subsequent pivot). The reason for this is that we have an in-memory cache where we keep track of the storage root for every account that needs to be downloaded. On every state healing phase on a pivot change, we update the storage root of every account we encounter. However, if an account time travels, we will not encounter it during healing because we already had it, and thus our storage root for it will be wrong.

NOTE: this should no longer be a problem now that our trie is path based, because on every state healing phase we erase the account's previous state. However, it's important to keep this scenario in mind when developing snap sync; any change in how we handle our trie or our assumptions may run into a problem if it does not properly handle time traveling accounts.

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync_concerns.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Healing
+
+There are two healing challenges. We should have a single main algorithm for healing, not have the code duplicated across two files. On top of that, the Membatch structure should be replaced with "PendingNodes" as it's a far more descriptive name.
+
+## Memory Concerns
+
+### Sotrage accounts
+
+Currently, we use a struct `accounts_by_root_hash` that we don't check the memory size. When rewriting this algorithm we should check if we're not going over the memory limit.
```
jrchatruc: > that we don't check the memory size.

that has an unbounded size in memory. When rewriting this algorithm we should make sure that we never use more than a certain amount of memory for it; the rest should be on disk.

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync.md
fedacking: Clarified at the top that both of these flags are for testing only in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6)

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6).

docs/l1/fundamentals/snap_sync_concerns.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6).

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [dc0d14b](https://github.com/lambdaclass/ethrex/pull/5560/commits/dc0d14bd186e191e16a805d2ddb20e3c43678ad6).

docs/l1/fundamentals/snap_sync_concerns.md
fedacking: Done in [9439289](https://github.com/lambdaclass/ethrex/pull/5560/commits/94392890ba37f1983f1f490c7e92fa60cd320f33)

docs/l1/fundamentals/snap_sync_concerns.md
fedacking: Done in [9439289](https://github.com/lambdaclass/ethrex/pull/5560/commits/94392890ba37f1983f1f490c7e92fa60cd320f33)

docs/l1/fundamentals/snap_sync.md
fedacking: Done in [9439289](https://github.com/lambdaclass/ethrex/pull/5560/commits/94392890ba37f1983f1f490c7e92fa60cd320f33)

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,328 @@
...
+
+[^1]: We update pivots based only on the timestamp, not on the peer response. According to the spec, stale roots return empty responses, but Byzantine peers may return empty responses at arbitrary times. Therefore, we rely on the rule that nodes must keep at least 128 blocks. Once `time > timestamp + 128 * 12` we mark the pivot as stale.
+
+Example of a possible state after stopping fast sync due to staleness.
+
+![Fast Sync Retaking Example - 1](./snap_sync/FastSyncRetakingExample-1.png)
+
+In the example, even if we find that node { hash: 0x317f, path: 0 } is correct, we still need to check all its children in the DB (in this case none are present).
+
+To solve the second problem, we introduce an optimization, which is called the "Membatch"^2. This allows us to maintain a new invariant:
```
MegaRedHand: I think this is missing the brackets:

```suggestion
To solve the second problem, we introduce an optimization, which is called the "Membatch"[^2]. This allows us to maintain a new invariant:
```

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,328 @@
...
+
+### Flags
+
+When developing snap sync there are flags to take into account that are used only for testing:
+
+- If the `SKIP_START_SNAP_SYNC` environment variable is set and isn't empty, it will skip the step of downloading the leaves and will immediately begin healing. This simulates the behaviour of fast-sync.
+
+- If debug assertions are on, the program will validate that the entire state and storage tries are valid by traversing the entire trie and recomputing the roots. If any is found to be wrong, it will print an error and exit the program. This is used for debugging purposes; a validation error here means that there is 100% a bug in snap sync.
+
+- `--syncmode [full, default:snap]` which defines what kind of sync we use. Full is executing each block, and isn't possible for mainnet and sepolia.
```
MegaRedHand:
```suggestion
- `--syncmode [full, default:snap]` which defines what kind of sync we use. Full is executing each block, and isn't possible on a fresh sync for mainnet and sepolia.
```

docs/l1/fundamentals/snap_sync.md
```diff
@@ -0,0 +1,328 @@
+# Snap Sync
+
+API reference: https://github.com/ethereum/devp2p/blob/master/caps/snap.md
+
+Terminology: 
+ - Peers: Other Ethereum execution clients we are connected to, and which can respond to snap requests.
+ - Pivot: The block we have chosen to snap-sync to.
```
jrchatruc: > Pivot: The block we have chosen to snap-sync to.

Pivot: The block we have chosen to snap-sync to. The pivot block changes continually as it becomes too old, because nodes don't serve old data (i.e. more than 128 blocks in the past). Read below for more details on this.

fedacking: Done in [50fede1](https://github.com/lambdaclass/ethrex/pull/5560/commits/50fede1bf4b714600d260844eea40611534eaa8a).

PR #5562, opened 2025-12-10
chore(storage): remove redundant StoreEngine overrides in RocksDB backend

ANtutov:
The RocksDB StoreEngine implementation was re-defining get_transaction_by_hash, get_transaction_by_location and get_block_by_hash with logic that is byte-for-byte identical to the default implementations in the StoreEngine trait. This duplication brings no behavioral difference, but makes future changes harder to reason about and increases the risk of divergence between backends. This commit deletes the redundant overrides so that the RocksDB backend relies on the shared default implementations, matching the in-memory backend and centralizing the composite logic in api.rs.


MegaRedHand:
Closing this due to conflicts. Feel free to rebase and reopen if applicable

PR #5563, opened 2025-12-10
fix(l1): ignore FCUs if state is not reachable

### LINKED ISSUE #5558: Handle FCU calls with a head before known states
We currently update the snap-sync pivot optimistically, to "latest block number according to current time" - 128. This leads to errors if we receive an FCU for a block before the last snap-sync pivot.

We identified the issue because of a stale checkpoint-sync server, which was serving a state several days old. This caused the consensus client to start syncing from that state, continuously sending FCUs for old blocks to ethrex. Since the state was too old, syncing took ~1h30m, enough so that the old FCUs arrived between when we finished snap-sync and started full-sync. This, combined with FCUs also being processed during snap-sync (see #5547), caused an inconsistent forkchoice state in the DB, which overwrote the latest known block number to that of the FCU head, which didn't match our current world state. The mismatch produced errors during the resulting full-sync.

We should research how other clients handle this. A sensible way to handle this would be to wait for the consensus client to sync first, before we start snap-syncing. Apart from this, we should avoid going to a block before the snap-sync pivot (or before the last available state), and instead log an error/warning for those cases, asking for the operator to take a look.

MegaRedHand:
**Motivation**

We currently fail to sync if we receive an FCU for a block we can't reach from our DB.

**Description**

This PR ignores FCUs when the nearest link to the chain is a block with a state we pruned, or before our last snap-sync pivot. It also logs a warning, which recommends that the user re-sync their node if they consistently see it. Note that this is a temporary patch, which should be removed once we rewrite snap-sync.

Closes #5558

### DIFF
--- a/crates/blockchain/fork_choice.rs
+++ b/crates/blockchain/fork_choice.rs
@@ -111,7 +111,7 @@ pub async fn apply_fork_choice(
             link_block=%link_block_hash,
             link_number=%link_header.number,
             head_number=%head.number,
-            "FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, it can be fixed by removing the DB and re-syncing the execution client."
+            "FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, if consensus is synced and this is a consistent message it can be fixed by removing the DB and re-syncing the execution client."
         );
         return Err(InvalidForkChoice::StateNotReachable);
     }


crates/blockchain/fork_choice.rs
```diff
@@ -97,6 +98,24 @@ pub async fn apply_fork_choice(
...
+    };
+
+    // If the state can't be constructed from the DB, we ignore it and log a warning.
+    // TODO(#5564): handle arbitrary reorgs
+    if !store.has_state_root(link_header.state_root)? {
+        warn!(
+            link_block=%link_block_hash,
+            link_number=%link_header.number,
+            head_number=%head.number,
+            "FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, it can be fixed by removing the DB and re-syncing the execution client."
```
rodrigo-o:
```suggestion
            "FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, if consensus is synced and this is a consistent message it can be fixed by removing the DB and re-syncing the execution client."
```

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -220,6 +220,10 @@ async fn handle_forkchoice(
         ));
     }
 
+    // Ignore any FCU during snap-sync.
+    // Processing the FCU while snap-syncing can result in reading inconsistent data
+    // from the DB, and the later head update can overwrite changes made by the syncer
+    // process, corrupting the forkchoice state (see #5547)
```
rodrigo-o, line 226: 👌

rodrigo-o (approved):

crates/blockchain/fork_choice.rs
rodrigo-o: Just to be sure removing the db is not a fix when seeing this just once. I doubled the consensus sync part just in case, but could be removed

PR #5595, opened 2025-12-10
fix(l1): use correct target to get closest neighbors in discv4

lakshya-sky:
** Motivation **
Instead of using sender id to get the distance use `target` provided from the `FindNodesMessage` which is correct spec.


MegaRedHand (approved):
Great find!

pablodeymo (approved):

MegaRedHand:
Hive tests fail due to authentication issues. I'm testing a fix for that in #5596

PR #5597, opened 2025-12-10
feat(l1)!: fail opening DB without schema version

### LINKED ISSUE #5541: Check datadir schema matches before initializing `Store`
The current schema version check occurs after initializing the `Store`, which might drop existing tables if the schema doesn't match. This can't be easily solved in a backend-agnostic way, so I propose storing the schema version in a separate file (in plain-text, JSON, or similar), which can also be more easily read by users.

MegaRedHand:
**Motivation**

We're currently opening the Store backend before checking the DB schema version matches. This can lead to data loss in case the RocksDB schema is changed.

**Description**

This PR changes the metadata to be stored in JSON format in a backend-agnostic way. This allows for inspecting the DB version without starting ethrex. Example:

```text
$ cat "~/.local/share/ethrex/metadata.json"
{
  "schema_version": 1
}
```

It also allows for more easily extending the check with more validations, for example, for backend type.

Additionally, it also fails with a clear error when trying to run a new ethrex version on an old DB without this versioning scheme:

```
2025-12-11T14:33:47.045029Z  INFO Opening storage engine engine=RocksDB path="/Users/mega/Library/Application Support/ethrex"
Error: Incompatible DB Version: found 0, expected 1. Please run `ethrex removedb` and restart node

Location:
    /Users/mega/execution/ethrex/cmd/ethrex/initializers.rs:400:24
```

Closes #5541

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -2386,7 +2386,9 @@ impl Store {
     }
 
     pub fn create_checkpoint(&self, path: impl AsRef<Path>) -> Result<(), StoreError> {
-        self.backend.create_checkpoint(path.as_ref())
+        self.backend.create_checkpoint(path.as_ref())?;
+        init_metadata_file(path.as_ref())?;
+        Ok(())
     }
 
     pub fn get_store_directory(&self) -> Result<PathBuf, StoreError> {
@@ -2852,13 +2854,8 @@ fn validate_store_schema_version(path: &Path) -> Result<(), StoreError> {
             return Err(StoreError::NotFoundDBVersion {
                 expected: STORE_SCHEMA_VERSION,
             });
-        } else {
-            std::fs::create_dir_all(path)?;
         }
-        let metadata = StoreMetadata::new(STORE_SCHEMA_VERSION);
-        let serialized_metadata = serde_json::to_string_pretty(&metadata)?;
-        let mut new_file = std::fs::File::create_new(metadata_path)?;
-        new_file.write_all(serialized_metadata.as_bytes())?;
+        init_metadata_file(path)?;
         return Ok(());
     }
     if !metadata_path.is_file() {
@@ -2876,6 +2873,17 @@ fn validate_store_schema_version(path: &Path) -> Result<(), StoreError> {
             expected: STORE_SCHEMA_VERSION,
         });
     }
+    Ok(())
+}
+
+fn init_metadata_file(parent_path: &Path) -> Result<(), StoreError> {
+    std::fs::create_dir_all(parent_path)?;
+
+    let metadata_path = parent_path.join(STORE_METADATA_FILENAME);
+    let metadata = StoreMetadata::new(STORE_SCHEMA_VERSION);
+    let serialized_metadata = serde_json::to_string_pretty(&metadata)?;
+    let mut new_file = std::fs::File::create_new(metadata_path)?;
+    new_file.write_all(serialized_metadata.as_bytes())?;
     Ok(())
 }
 


cmd/ethrex/initializers.rs
```diff
@@ -396,9 +396,11 @@ pub async fn init_l1(
 
     let store = match init_store(datadir, genesis).await {
         Ok(store) => store,
-        Err(StoreError::IncompatibleDBVersion) => {
+        Err(err @ StoreError::IncompatibleDBVersion { .. })
+        | Err(err @ StoreError::NotFoundDBVersion { .. }) => {
+            // Error: Incompatible DB Version: not found, expected v1. Please erase your DB by running `ethrex removedb` and restart node to resync. Note that this will take a while.
```
pablodeymo: this comment is a bit suspicious, if you upgrade from no-version to a future v2 version, this is non-sense

MegaRedHand: Removed 🫡

cmd/ethrex/utils.rs
```diff
@@ -98,12 +98,8 @@ pub fn default_datadir() -> PathBuf {
...
 pub fn init_datadir(datadir: &Path) {
-    if datadir.exists() {
-        if !datadir.is_dir() {
-            panic!("Datadir {datadir:?} exists but is not a directory");
-        }
-    } else {
-        std::fs::create_dir_all(datadir).expect("Failed to create data directory");
+    if datadir.exists() && !datadir.is_dir() {
+        panic!("Datadir {datadir:?} exists but is not a directory");
     }
```
rodrigo-o, line 107: Now we don't create the datadir directoyr if it's not already created? what happens on a first run or removing the directory manually?

rodrigo-o: NVM, this now happens befor in the metadata check i assume!

MegaRedHand: We still create it. I just had to move it to the validation function

MegaRedHand: After [feeddd4](https://github.com/lambdaclass/ethrex/pull/5597/commits/feeddd494967a2403af8806e60881b09113b7f3b), this change was no longer needed, so I reverted it.

cmd/ethrex/initializers.rs
```diff
@@ -396,9 +396,10 @@ pub async fn init_l1(
 
     let store = match init_store(datadir, genesis).await {
         Ok(store) => store,
-        Err(StoreError::IncompatibleDBVersion) => {
+        Err(err @ StoreError::IncompatibleDBVersion { .. })
+        | Err(err @ StoreError::NotFoundDBVersion { .. }) => {
```
Oppen, line 400: We start at 1 so we can treat 0 and missing as equivalent, so I don't think the new error kind is necessary.

crates/storage/lib.rs
```diff
@@ -18,3 +18,6 @@ pub use trie_db::layering::apply_prefix;
 /// Store Schema Version, must be updated on any breaking change
 /// An upgrade to a newer schema version invalidates currently stored data, requiring a re-sync.
 pub const STORE_SCHEMA_VERSION: u64 = 1;
+
+/// Name of the file storing the metadata about the database
+pub const STORE_METADATA_FILENAME: &str = "metadata.json";
```
Oppen, line 22: I think the place for the schema versioning is the DB. Decoupling it is asking for trouble. E.g., I don't see the `removedb` command deleting this properly.
If we don't want it to be data, we may follow a scheme similar to what Postgres does, which is encoding versioning in file names (e.g., the internal directory might be `data-{schema_version}`, with failure indicating incompatibility).

cmd/ethrex/initializers.rs
MegaRedHand: It makes the distinction clear and also simplifies the formatting of the error here.

crates/storage/lib.rs
MegaRedHand: `removedb` removes the whole datadir, including the metadata file, so there should be no problem there.

The main reason to move the schema versioning to another file is to make it backend-agnostic. Keeping it inside the DB means we need to open a `Store` before checking if we support its version. Right now, this means we may have some data loss due to unknown columns being dropped by RocksDB. This can be solved, but it would require the backend itself to check the version.

It also has the benefit that we can include more information, like the engine type we're using, and we can easily inspect it because we use plaintext. Having it in filenames is a good alternative, but I don't see the benefits in moving to that.

PR #5598, opened 2025-12-10
ci(l1): improve snapsync smoketest

MegaRedHand:
**Motivation**

We found the smoketest lacking.

**Description**

This PR changes the snapsync smoketest job to run every 3 hours, with both Lighthouse and Prysm consensus clients.

### DIFF
--- a/.github/workflows/daily_snapsync.yaml
+++ b/.github/workflows/daily_snapsync.yaml
@@ -32,18 +32,18 @@ jobs:
         run: |
           event_name="${GITHUB_EVENT_NAME}"
           if [[ "$event_name" == "schedule" ]]; then
-            json='[{"network":"hoodi","timeout":"30m"},{"network":"sepolia","timeout":"2h"}]'
+            json='[{"network":"hoodi","timeout":"40m"},{"network":"sepolia","timeout":"2h30m"}]'
           elif [[ "$event_name" == "pull_request" ]]; then
-            json='[{"network":"hoodi","timeout":"30m"}]'
+            json='[{"network":"hoodi","timeout":"40m"}]'
           else
             network=$(jq -r '.inputs.network // empty' "$GITHUB_EVENT_PATH")
 
             case "$network" in
               hoodi)
-                json='[{"network":"hoodi","timeout":"30m"}]'
+                json='[{"network":"hoodi","timeout":"40m"}]'
                 ;;
               sepolia)
-                json='[{"network":"sepolia","timeout":"2h"}]'
+                json='[{"network":"sepolia","timeout":"2h30m"}]'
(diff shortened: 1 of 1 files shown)


.github/workflows/daily_snapsync.yaml
```diff
@@ -6,8 +6,8 @@ on:
     paths:
       - ".github/workflows/daily_snapsync.yaml"
   schedule:
-    # Every day at UTC 03:00
-    - cron: "0 3 * * *"
+    # Every 3 hours
+    - cron: "0 */3 * * *"
```
rodrigo-o: Will this work? as far as i Understand this 2 jobs will run sequentially (i might be wrong) taking nearly 2hs each. Also I'm not sure were the timeout is set i se the timeout true but missed the actual number.

.github/workflows/daily_snapsync.yaml
```diff
@@ -70,3 +70,25 @@ jobs:
         with:
           network: ${{ matrix.network }}
           timeout: ${{ matrix.timeout }}
+          cl_type: lighthouse
+          cl_image: "sigp/lighthouse:v8.0.1"
```
Oppen, line 74: If it's the exact same workflow except for this, consider setting these in the strategy matrix.

Oppen (approved):

.github/workflows/daily_snapsync.yaml
Oppen: It's set in a somewhat complicated way in the prepare step. I'm not sure why it was done like that when `strategy.matrix.include` would have done the trick AFAICT, but it predates this PR.

.github/workflows/daily_snapsync.yaml
MegaRedHand: I considered it, but it's being specified dynamically in another job. I feel this was easier and more clear, and we have plans to move this outside GHA anyways.

.github/workflows/daily_snapsync.yaml
MegaRedHand: I checked and running hoodi+sepolia took <1h30m, so 2x that should be right on time. We can bump this to 4h, but I feel it's better to try with 3h first and bump it if we see the need for it.

MegaRedHand: Changed to 4h. We might bump it further if we see start seeing false positives.

rodrigo-o (approved):
Looking good.

I'm just a bit hesitant about the 4hs runtime and the 30m + 2hs timeout in the jobs (which if all timeout gives us 5hs). I also seen some sepolia jobs taking more than 2hs, but we can start with this and refine.

PR #5599, opened 2025-12-11
refactor(l1): use threads::Genserver for synchronic threaded code

### LINKED ISSUE #5565: Use spawned sync (thread) version of GenServer for our synchronic theaded code.
Spawned has a [sync version](https://github.com/lambdaclass/spawned/blob/main/concurrency/src/threads/gen_server.rs) of `GenServer` that relies on threads instead of tasks. And we have some processes in ethrex that can get benefit from using them.

In Ethrex, block execution is sync code and it is a good place to start testing this feature. For example we are [spawning a thread](https://github.com/lambdaclass/ethrex/blob/main/crates/storage/store_db/rocksdb.rs#L471-L517) that applies tries updates. We can have a `GenServer` instead that does this job. We should identify if there are other places where we want to run threads with non-async code that could convert to a `GenServer`.

Also note that `threads::GenServer` was not maintained at the same pace of `tasks::GenServer` so it may be necessary to update it or add features

lakshya-sky:
** Motivation **
Block execution is sync code and can benefit from using sync version of `GenServer` that relies on threads. 

** Changes **
- use crossbeam channel instead of std::sync::mpsc which allows cloning.
- Introduces `TrieUpdateWorker` and moves thread spawning inside the GenServer::Init.

Closes #5565


lakshya-sky:
Waiting for upstream changes to be accepted https://github.com/lambdaclass/spawned/pull/64.

lakshya-sky:
Hey @ElFantasma, I've pinned the revision of the spawned for now. Might be better to wait for the create to be released.

crates/storage/store.rs
```diff
@@ -2471,6 +2442,22 @@ struct TrieUpdate {
 
 // NOTE: we don't receive `Store` here to avoid cyclic dependencies
 // with the other end of `fkv_ctl`
```
Oppen: Let's move this as a footnote to the bigger comment. With the changes it looks strange.

crates/storage/store.rs
```diff
@@ -3233,3 +3220,98 @@ mod tests {
...
+
+    fn init(
+        self,
+        handle: &spawned_concurrency::threads::GenServerHandle<Self>,
+    ) -> Result<Self, Self::Error> {
+        let mut handle = handle.clone();
+        let rx = self.trie_upd_rx.clone();
+        let _ = std::thread::spawn(move || {
+            let mut cancellation_token = handle.cancellation_token();
+            loop {
```
Oppen: I'm confused, with the new actor shouldn't this be handled by spawned? I thought the goal would be for the worker implementation to only need to worry about the loop body.

lakshya-sky: Async tasks Genserver uses `spawn_listener` but here the sync version doesn't have it implemented https://github.com/lambdaclass/spawned/blob/7af72c4ce40eab8b22d33b3302e9da9229319dc0/concurrency/src/threads/stream.rs#L8C1-L17C2 that's why. 
Might be a good idea to implement `spawn_listener` in the upstream. But the function signature of the `spawn_listener` relies on `Stream` and the close one I was able to think is `IntoInterator`. e.g. `fn spawn_listener(handle: .., iter: impl IntoIterator<Item = T::CastMsg>)` but I was not sure about that change.

Maybe you and @ElFantasma can share some suggestion to implement `spawn_listener`.

lakshya-sky: Raised a PR https://github.com/lambdaclass/spawned/pull/66 which should allow to write 

```rust    
fn init(
        self,
        handle: &spawned_concurrency::threads::GenServerHandle<Self>,
    ) -> Result<Self, Self::Error> {
        let handle = handle.clone();
        let rx = self
            .trie_upd_rx
            .clone()
            .into_iter()
            .map(|u| Self::CastMsg::TrieUpdates(u));

        spawn_listener(handle, rx);
        Ok(self)
    }
```

Cargo.toml
```diff
@@ -108,7 +108,7 @@ kzg-rs = "0.2.6"
 libsql = "0.9.10"
 futures = "0.3.31"
 aligned-sdk = { git = "https://github.com/yetanotherco/aligned_layer", rev = "c60d7eb147edbdf12bb7a7c6e92ec178d9f8da23" }
-spawned-concurrency = "0.4.2"
+spawned-concurrency = { git = "https://github.com/lakshya-sky/spawned", version = "0.4.5", rev = "c1251c1d6996b22c3d2facbcc5e3d35d056040b5" }
```
ElFantasma: **Supply chain concern:** This switches `spawned-concurrency` from the published crates.io version to the PR author's personal fork (`lakshya-sky/spawned`) pinned at a specific commit. The fork adds a simplified `spawn_listener` for the `threads` module (commit `c1251c1` — one commit after the published 0.4.5 release) with a different signature than the crates.io version.

This is the same kind of issue we'd flag for any dependency: pulling from a personal GitHub repo bypasses the crates.io publishing/audit pipeline. The proper path is to get the upstream `spawned` crate to publish a new release with this API change, then use the published version.

crates/storage/store.rs
```diff
@@ -1410,46 +1415,11 @@ impl Store {
...
-                        .inspect_err(|err| error!("apply_trie_updates failed: {err}"));
-                    }
-                    Err(err) => {
-                        debug!("Trie update sender disconnected: {err}");
-                        return;
-                    }
-                }
-            }
-        });
+        TrieUpdateWorker::new(backend, flatkeyvalue_control_tx, trie_cache, trie_upd_rx).start();
```
ElFantasma: **Bug — lost shutdown join:** The old code pushed the trie update thread's `JoinHandle` into `background_threads`, which is joined on `Drop` via `ThreadList`. Here, `.start()` returns a `GenServerHandle` that is immediately dropped. Inside `GenServerHandle::new`, the spawned thread's `JoinHandle` is stored as `_join_handle` (explicitly ignored). Additionally, `spawn_listener` in `init()` spawns yet another thread whose `JoinHandle` is also dropped.

This means the trie update worker thread is now fire-and-forget — on shutdown, `Store` will join the FKV generator thread (still in `background_threads`) but won't wait for the trie update worker. If the process exits while `apply_trie_updates` is mid-write (specifically during the disk-persist step), you could get a partially written diff layer.

Suggestion: store the returned `GenServerHandle` so its cancellation token can be used for clean shutdown, or at minimum store the thread `JoinHandle` so it's joined before exit.

crates/storage/store.rs
```diff
@@ -3233,3 +3220,83 @@ mod tests {
...
+            backend,
+            flatkeyvalue_control_tx,
+            trie_cache,
+            trie_upd_rx,
+        }
+    }
+
+    fn apply_trie_updates(&self, trie_update: TrieUpdate) {
+        // FIXME: what should we do on error?
+        tracing::info!("Apply Trie Updates");
```
ElFantasma: `info!` level is too noisy for a per-block operation — this will fire on every single block execution. Should be `trace!` or `debug!`, or removed entirely since the `FIXME` comment above already signals this path needs work.

crates/storage/store.rs
```diff
@@ -3233,3 +3220,83 @@ mod tests {
...
+        handle: &spawned_concurrency::threads::GenServerHandle<Self>,
+    ) -> Result<Self, Self::Error> {
+        let handle = handle.clone();
+        let rx = self
+            .trie_upd_rx
+            .clone()
+            .into_iter()
+            .map(Self::CastMsg::TrieUpdates);
+
+        spawn_listener(handle, rx);
```
ElFantasma: The `JoinHandle` returned by `spawn_listener` is dropped here, so this listener thread is fire-and-forget. If the GenServer's main loop stops (e.g. on error), this listener thread keeps running and calling `handle.cast()` which will fail — not catastrophic, but unclean. Consider storing the handle for coordinated shutdown.

ElFantasma (changes requested):
This will have to wait till new release of Spawned anyway, and it actually is going through a major rework.

ElFantasma (changes requested):
This will have to wait till new release of Spawned anyway, and it actually is going through a major rework.

ElFantasma:
Ethrex was migrated to Spawned v0.5.0 so we are ready to continue.

@lakshya-sky you should rebase to latest version if you wish to continue with this work.

crates/storage/store.rs
```diff
@@ -3458,3 +3454,62 @@ pub fn read_chain_id_from_db(path: &Path) -> Option<u64> {
...
+#[actor(protocol = TrieUpdateWorkerProtocol)]
+impl TrieUpdateWorker {
+    #[started]
+    fn started(&mut self, ctx: &Context<Self>) {
+        let rx = self
+            .trie_upd_rx
+            .clone()
+            .into_iter()
+            .map(|val| trie_update_worker_protocol::TrieUpdates { updates: val });
+        spawn_listener(ctx.clone(), rx);
```
ElFantasma, line 3496: **Shutdown leak:** When the Store is dropped, the crossbeam sender is dropped, so `into_iter()` ends and the `spawn_listener` thread exits — dropping its `Context` clone (which holds an `mpsc::Sender`). However, `run_actor` in spawned v0.5.0 holds a *second* `Context` on its stack for the entire loop. That keeps the `mpsc::Sender` alive, so `rx` never disconnects. The loop falls into a 100ms `recv_timeout` spin because nothing cancels the `CancellationToken` either.

The old `std::thread::spawn` loop exited cleanly via `recv() -> Err(RecvError)` when the Store's `SyncSender` was dropped. To fix: either cancel the actor's token when the crossbeam channel closes (e.g. have `spawn_listener`'s thread call `ctx.stop()` after the iterator ends), or skip `spawn_listener` and receive from the crossbeam channel directly in `handle_trie_updates` via a timer/poll pattern.

crates/storage/store.rs
```diff
@@ -3458,3 +3454,62 @@ pub fn read_chain_id_from_db(path: &Path) -> Option<u64> {
...
+    }
+
+    #[send_handler]
+    fn handle_trie_updates(
+        &mut self,
+        msg: trie_update_worker_protocol::TrieUpdates,
+        _ctx: &Context<Self>,
+    ) {
+        // FIXME: what should we do on error?
+        tracing::info!("Apply Trie Updates");
```
ElFantasma, line 3506: `tracing::info!` fires on every block execution during sync. Should be `debug!` or `trace!`.

crates/storage/store.rs
```diff
@@ -3458,3 +3454,62 @@ pub fn read_chain_id_from_db(path: &Path) -> Option<u64> {
...
+        }
+    }
+}
+
+#[actor(protocol = TrieUpdateWorkerProtocol)]
+impl TrieUpdateWorker {
+    #[started]
+    fn started(&mut self, ctx: &Context<Self>) {
+        let rx = self
+            .trie_upd_rx
```
ElFantasma, line 3492: nit: `self.trie_upd_rx.clone()` creates a second crossbeam `Receiver` that stays in `self` as dead weight — only the cloned one is consumed by `spawn_listener`. Consider using `Option<Receiver>` + `.take()` to make the ownership transfer explicit.

ElFantasma (changes requested):

PR #5604, opened 2025-12-11
fix(l1): retry head fetch during sync

iovoid:
**Motivation**

Currently [one of the hive tests](https://github.com/lambdaclass/ethrex/actions/runs/20121423289/attempts/1#summary-57742462357) is failing, with the log `Sync failed to find target block header, aborting`.

This is because currently we only retry 5 times and (for low-latency scenarios like a testnet) without delay between the attempts. If the header isn't immediately available, we abort.

**Description**

This solves the problem by introducing a retry-with-delay logic.

The currently existing logic in `request_block_headers_from_hash` is removed since it now becomes redundant.

**Checklist**

- [X] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync.


crates/networking/p2p/sync.rs
```diff
@@ -188,26 +191,34 @@ impl Syncer {
...
                 .await?
             else {
-                warn!("Sync failed to find target block header, aborting");
-                return Ok(());
+                if attempts > MAX_HEADER_FETCH_ATTEMPTS {
+                    warn!("Sync failed to find target block header, aborting");
+                    return Ok(());
+                }
+                attempts += 1;
+                tokio::time::sleep(Duration::from_millis(200)).await;
```
Oppen: This is a lot of time to sleep. Maybe you should try an exponential backoff?

crates/networking/p2p/sync.rs
```diff
@@ -323,16 +334,22 @@ impl Syncer {
...
             else {
-                warn!("Sync failed to find target block header, aborting");
-                debug!("Sync Log 8: Sync failed to find target block header, aborting");
-                return Ok(());
+                if attempts > MAX_HEADER_FETCH_ATTEMPTS {
+                    warn!("Sync failed to find target block header, aborting");
+                    return Ok(());
+                }
+                attempts += 1;
+                tokio::time::sleep(Duration::from_millis(200)).await;
```
Oppen: Ditto.

crates/networking/p2p/sync.rs
iovoid: Added in 2d7e667f4

crates/networking/p2p/sync.rs
iovoid: Added in 2d7e667f4

MegaRedHand (approved):

rodrigo-o:
> Checklist
> - [x] Updated STORE_SCHEMA_VERSION (crates/storage/lib.rs) if the PR includes breaking changes to the Store requiring a re-sync.

Is this outdated?

Issue #5601, opened 2025-12-11
Properly validate `withdrawals_root` on blocks


mpaulucci:
There are Hive tests that are failing here https://hive.ethpandaops.io/#/test/fusaka/1765117435-d508c29334cffe232f56986ae9487619

These test cases verify that blocks with an invalid withdrawals_root header field are rejected.

Acceptance criteria:
- All hive tests for withdrawals should pass

MegaRedHand:
It appears those are failing because we don't do some checks when importing blocks through `ethrex import`. In other cases, the checks are already there.

MegaRedHand:
Made #5606 fixing this issue and #5605 bumping to v5.4.0 of the ethereum-spec-tests fixtures, to run the version with the failing tests in our CI

PR #5608, opened 2025-12-11
fix(l2): validate block header and body match

MegaRedHand:
**Motivation**

We weren't checking the body contents match the block's header inside the guest program.

**Description**

This PR adds an additional check to the guest program that the header and body matches for each block.

Related to #5601

### DIFF
--- a/crates/common/common.rs
+++ b/crates/common/common.rs
@@ -17,6 +17,5 @@ pub mod utils;
 pub use errors::{EcdsaError, InvalidBlockError};
 pub use validation::{
     get_total_blob_gas, validate_block_access_list_hash, validate_block_pre_execution,
-    validate_gas_used,
-    validate_receipts_root, validate_requests_hash,
+    validate_gas_used, validate_receipts_root, validate_requests_hash,
 };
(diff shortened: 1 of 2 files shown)


xqft (approved):

crates/blockchain/blockchain.rs
```diff
@@ -1766,9 +1766,13 @@ pub fn find_parent_header(
...
 ///
-/// Note that this doesn't validate that the transactions or withdrawals root of the header matches the body
+/// # WARNING
+///
+/// This doesn't validate that the transactions or withdrawals root of the header matches the body
 /// contents, since we assume the caller already did it. And, in any case, that wouldn't invalidate the block header.
-pub fn validate_block(
+///
+/// To validate it, use [`ethrex_common::types::block::validate_block_body`]
+pub fn validate_block_pre_execution(
```
tomip01: If it validates the header, should it be called `validate_block_header_pre_execution`?

tomip01:
> We need to test this with GPU on the L2 and replay

I tried both, running the integration tests with GPU and replay with GPU a mainnet block, and succeeded

ElFantasma (approved):

Issue #5602, opened 2025-12-11
Fix `eth_config` showing bad info for next fork.


mpaulucci:
This can be reproduced by creating a kurtosis localnet with these parameters

```
    "blobSchedule": {
      "cancun": {
        "target": 3,
        "max": 6,
        "baseFeeUpdateFraction": 3338477
      },
      "prague": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      },
      "osaka": {
        "target": 6,
        "max": 9,
        "baseFeeUpdateFraction": 5007716
      },
      "bpo1": {
        "target": 10,
        "max": 15,
        "baseFeeUpdateFraction": 8346193
      }
    },
    "depositContractAddress": "0x00000000219ab540356cBB839Cbe05303d7705Fa",
    "pragueTime": 0,
    "osakaTime": 0,
    "bpo1Time": 0
````

When using Dora, you will also see an explanation mark next to ethrex, highlighting the fact that its not spec compliant.
<img width="3280" height="886" alt="Image" src="https://github.com/user-attachments/assets/6b7eaf0b-69b3-4d1c-aa2b-623397a2e8ee" />

This is the issue:
<img width="3114" height="984" alt="Image" src="https://github.com/user-attachments/assets/1f27374e-9e15-4db8-92d4-35a8b07e7c18" />

Dora uses `eth_config` so it's likely that there is some issue with that endpoint.

Acceptance criteria:
- Dora should show Ethrex as spec compliant

rodrigo-o:
This was an issue for v7.0.0, it's already fixed and will be part of our next release

PR #5609, opened 2025-12-11
fix(l1): fix bpo1 validation

rodrigo-o:
**Motivation**

We received a report that when setting bpo activation to 0 in kurtosis we weren't taking it into account

**Description**

Right now with our current setup, running `make localnet` starts a localnet with kurtosis and bpo1 activated in epoch 0. Unfortunately I'm not seeing the warn sign shown in #5602, instead everything appears to work correctly.

image

image

We need to work on reproduce the issue probably moving to ehereum-package main instead of the revision we are using by default and testing with latest from ethrex (7.0.0) instead of main.

### DIFF
--- a/fixtures/networks/default.yaml
+++ b/fixtures/networks/default.yaml
@@ -30,6 +30,13 @@ participants:
 
 ethereum_metrics_exporter_enabled: true
 
+network_params:
+  electra_fork_epoch: 0
+  bpo_1_epoch: 0
+  bpo_1_max_blobs: 15
+  bpo_1_target_blobs: 10
+  bpo_1_base_fee_update_fraction: 8346193
+
 additional_services:
   - dora
   - spamoor


barnabasbusa:
I'd recommend using the main ethereum package. As we make releases infrequently to it. We generally make one release per fork. Main is considered fairly stable, if there is a bug, its generally fixed under a few hours.

rodrigo-o:
> I'd recommend using the main ethereum package. As we make releases infrequently to it. We generally make one release per fork. Main is considered fairly stable, if there is a bug, its generally fixed under a few hours.

Thanks @barnabasbusa we've been discussing updating to main and we'll probably do it asap

rodrigo-o:
As discussed offline this was an issue for release 7.0.0 and is already fixed in main.

PR #5615, opened 2025-12-12
fix(l1): reject high-s transactions when recovering sender address

### LINKED ISSUE #4100: See if ethrex is accepting transactions with high-s and reject them if we are
As detailed in [EIP-2](https://eips.ethereum.org/EIPS/eip-2#:~:text=All%20transaction%20signatures%20whose%20s%2Dvalue%20is%20greater%20than%20secp256k1n/2%20are%20now%20considered%20invalid.%20The%20ECDSA%20recover%20precompiled%20contract%20remains%20unchanged%20and%20will%20keep%20accepting%20high%20s%2Dvalues%3B%20this%20is%20useful%20e.g.%20if%20a%20contract%20recovers%20old%20Bitcoin%20signatures.) we shouldn't accept signatures with high-s. By looking at the code it seems that we are indeed accepting these (if I'm not mistaken).
See if this is true and fix it. Otherwise close this issue.

Some ways to solve this:
- Use `k256` crate without normalizing the signature, and if it fails we know that it's invalid.
- Keep the `secp256k1` implementation but add the check for high-s.

JereSalo:
**Motivation**

- Ethrex was accepting transactions with high-s signatures. Ethereum doesn't allow these so I add a check in order not to accept them anymore.

**Description**

- Add high-s signature check. Read EIP-2 https://eips.ethereum.org/EIPS/eip-2

Closes #4100


crates/common/types/transaction.rs
```diff
@@ -1447,16 +1462,18 @@ pub fn recover_address(signature: Signature, payload: H256) -> Result<Address, k
...
-
-    let mut signature = k256::ecdsa::Signature::from_slice(&signature_bytes[..64])?;
-
-    let mut recovery_id_byte = signature_bytes[64];
-
-    if let Some(low_s) = signature.normalize_s() {
-        signature = low_s;
-        recovery_id_byte ^= 1;
+    // EIP-2: signatures must use "low-s" (s <= secp256k1n/2).
+    // Standard k256 rejects high-s signatures by default but it's best to leave this for 3 reasons:
```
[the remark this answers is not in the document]
JereSalo: I mean, if we didn't `normalize_s` then k256 would've rejected the signature anyway, that's why we normalize it in the first place

crates/common/Cargo.toml
```diff
@@ -26,6 +26,7 @@ libc = "0.2"
 crc32fast.workspace = true
 bytes.workspace = true
 hex.workspace = true
+hex-literal.workspace = true
```
MegaRedHand, line 29: Let's move this to dev-dependency

JereSalo: I'm using it in the constant though

MegaRedHand: My bad. Thought we were using it inside a test

PR #5625, opened 2025-12-12
ci(l1): bump hoodi timeout to 1h30m

MegaRedHand:
**Motivation**

We recently got a failure in the CI due to Hoodi reaching the timeout limit: https://github.com/lambdaclass/ethrex/actions/runs/20176312201/job/57925180164

**Description**

This PR bumps the Hoodi timeout to 1h30m


MegaRedHand:
We didn't have any timeout errors since Friday, so I think we can keep the timeout in 40m

PR #5628, opened 2025-12-13
test(l1): add FCU test for StateNotReachable case

### LINKED ISSUE #5603: Add test for newly added FCU cases
See https://github.com/lambdaclass/ethrex/pull/5563#discussion_r2606805639

We recently added an early return of SYNCING when we're snapsyncing, and another check that we have a state available for the nearest link in the chain for reorgs. We should add unit tests for each of those.

madisoncarter1234:
## Summary
- Add test for `StateNotReachable` error in `apply_fork_choice` covering the real-world pruning scenario
- Test builds a 130-block canonical chain, creates an early fork, simulates state pruning, then verifies reorg fails with `StateNotReachable`
- Added `clear_trie_cache_for_testing()` helper to Store for simulating pruning in tests (InMemory backend uses 10k threshold)

## Test plan
- `cargo test -p ethrex-blockchain test_state_not_reachable_on_reorg_beyond_pruning_range`

Closes #5603


crates/blockchain/smoke_test.rs
```diff
@@ -287,6 +287,48 @@ mod blockchain_integration_test {
...
+
+        // Now try to apply fork choice to this block.
+        // It should fail with StateNotReachable because the state_root doesn't exist.
+        let result = apply_fork_choice(&store, block_hash, genesis_hash, genesis_hash).await;
+
+        assert!(
+            matches!(result, Err(InvalidForkChoice::StateNotReachable)),
+            "Expected StateNotReachable, got {:?}",
+            result
+        );
```
MegaRedHand: This case isn't that useful. On normal operation, if we have the state for a block, and we receive a child, we'll only insert the header into the database if we executed the block (we have the state), so this can't happen.

The issue refers to cases where we have a reorg, and the nearest chain link is beyond our pruning range (128 blocks, currently).

madisoncarter1234: understood, rewrote the test to cover the actual reorg scenario you described. builds 130 blocks on canonical, forks early, then simulates the pruning by clearing the trie cache before attempting the reorg.

Had to add a test helper (clear_trie_cache_for_testing) to Store since the InMemory backend threshold is 10k and won't prune naturally in tests.

madisoncarter1234:
> is this being blocked by another task?

just waiting on approvals, ready for review

crates/storage/store.rs
```diff
@@ -2354,6 +2354,23 @@ impl Store {
...
+
+    /// Clears the in-memory trie layer cache.
+    ///
+    /// # Warning
+    /// This is a test helper that simulates state pruning. **Do not use in production.**
+    ///
+    /// In production, state older than 128 blocks is pruned from the diff layers and
+    /// may not be available on disk if using InMemory backend (which has a 10000 block
+    /// threshold for disk commits).
+    ///
```
ElFantasma, line 2366: `pub` visibility on a test-only method exposes it in the public API of `ethrex-storage`. Since this crate is `#[cfg(test)]`-incompatible (the test is in a different crate), consider either:

1. `#[doc(hidden)] pub fn clear_trie_cache_for_testing(...)` — hides from docs but still accessible
2. A `pub fn clear_trie_cache_for_testing(...)` behind a `#[cfg(feature = "test-utils")]` feature flag
3. Or simply accept the `pub` with a doc comment warning (which you already have) — this is test tooling and the method name makes intent clear

Option 3 is fine for now, but worth considering for the rebase.

crates/storage/store.rs
```diff
@@ -2354,6 +2354,23 @@ impl Store {
...
+    /// Use this to test scenarios where a reorg's common ancestor state is unavailable.
+    pub fn clear_trie_cache_for_testing(&self) {
+        use crate::layering::TrieLayerCache;
```
ElFantasma, line 2369: `TrieLayerCache::default()` sets `commit_threshold: 128`, but the store may have been initialized with a different value (e.g., `TrieLayerCache::new(10000)` for InMemory). This doesn't break the test since clearing the cache is the goal, but if the threshold matters for subsequent operations, the replacement cache will behave differently.

Consider `TrieLayerCache::new(cache.commit_threshold)` to preserve the original threshold (you'd need to add a getter or read it before replacing), or simply `*cache = Arc::new(TrieLayerCache::default());` is fine if the test doesn't execute more blocks after clearing.

crates/blockchain/smoke_test.rs
```diff
@@ -287,6 +287,85 @@ mod blockchain_integration_test {
...
+        let block_1 = new_block(&store, &genesis_header).await;
+        let hash_1 = block_1.hash();
+        let block_1_state_root = block_1.header.state_root;
+        blockchain.add_block(block_1.clone()).unwrap();
+        apply_fork_choice(&store, hash_1, genesis_hash, genesis_hash)
+            .await
+            .unwrap();
+
+        // Create a fork block at height 2, child of block 1 (will be non-canonical)
+        let fork_block = new_block(&store, &block_1.header).await;
```
ElFantasma, line 322: This loop builds 129 blocks, each calling `new_block` (which creates a `Blockchain`, builds a payload) + `add_block` + `apply_fork_choice`. This is likely slow — consider reducing `CHAIN_LENGTH` and using a smaller threshold, or documenting that this test is intentionally slow.

Alternatively, since `clear_trie_cache_for_testing` nukes all cached state regardless of how many blocks were built, the 128-block threshold is only conceptually relevant (it motivates *why* state would be pruned). You could use a much smaller chain (e.g., 5 blocks) and the test would still verify the same code path.

crates/storage/store.rs
```diff
@@ -2354,6 +2354,23 @@ impl Store {
...
+    ///
+    /// # Warning
+    /// This is a test helper that simulates state pruning. **Do not use in production.**
+    ///
+    /// In production, state older than 128 blocks is pruned from the diff layers and
+    /// may not be available on disk if using InMemory backend (which has a 10000 block
+    /// threshold for disk commits).
+    ///
+    /// Use this to test scenarios where a reorg's common ancestor state is unavailable.
+    pub fn clear_trie_cache_for_testing(&self) {
```
ElFantasma, line 2368: Silently ignoring lock poisoning: if another thread panicked while holding `trie_cache`, `lock()` returns `Err` and this does nothing. In a test helper this is probably fine, but an `expect("trie_cache lock poisoned")` would surface the issue rather than silently passing through.

ElFantasma (commented):
## Review Summary

The test scenario is well-designed — building a chain past the pruning window, creating a fork, clearing state, and verifying `StateNotReachable` is the correct approach. However, **this PR needs a rebase**: `smoke_test.rs` has been moved to `test/tests/blockchain/smoke_tests.rs` on `main`, and the old `mod blockchain_integration_test` module wrapper no longer exists.

Two design concerns: (1) `clear_trie_cache_for_testing` is a `pub` method on the production `Store` type — consider using `#[cfg(test)]` gating or a separate test helper trait to avoid exposing it in the public API; and (2) `TrieLayerCache::default()` resets `commit_threshold` to 128, which differs from the value the store was initialized with (see inline comment).

ElFantasma:
We are in the process of cleaning the stalled or trivial PRs. This one has conflicts and has unanswered feedback

I'm closing it, but feel free to reopen if you are planning to work on this.

Issue #5637, opened 2025-12-15
PeerTable error on snapsync


rodrigo-o:
During the weekend we've been running continues MAINNET and Sepolia Snapsync runs in loop, we've seen 3 runs across that failed and all of them with the same error:

 ```
2025-12-14T10:18:08.387651Z DEBUG State Healing status="in progress" snap_peers=95 inflight_tasks=8 longest_path_seen=6 leafs_healed=0 global_leafs_healed=0 downloads_rate=1.0 paths_to_go=11559 pending_nodes=32645 heals_per_cycle=71
 2025-12-14T10:18:10.387663Z DEBUG State Healing status="in progress" snap_peers=95 inflight_tasks=1 longest_path_seen=11 leafs_healed=17502 global_leafs_healed=17502 downloads_rate=1.0 paths_to_go=0 pending_nodes=329 heals_per_cycle=309
 2025-12-14T10:18:10.752217Z DEBUG Nothing more to heal found
 2025-12-14T10:18:10.891246Z DEBUG State Healing stopped, signaling storage healer
 2025-12-14T10:18:10.891688Z DEBUG Stopped state healing
 2025-12-14T10:18:10.891691Z  INFO Started request_storage_ranges with 1695855 accounts with storage root unchanged
 2025-12-14T10:18:12.873711Z  INFO Received new engine request: Requested Blobs V2
 2025-12-14T10:18:12.946136Z  INFO Received new payload block_hash=0x2567…f12c block_number=24010195
 2025-12-14T10:18:13.035906Z  INFO New fork choice update version=v3 head=0x256742477c976fffc98b931b430838c6e4f2d79e1a18b89737cd07f661b9f12c safe=0xda34fe2fe217592eddceeb12a775385b15ff762f1024c73d8d6eef302713939d finalized=v0xd31da016f10ca40e001b9c5497ca7142cc6455b6e70d658d46e84b1bfc31b220
 2025-12-14T10:18:21.768204Z ERROR Error Handling Discovery message err=PeerTable(InternalError(CallTimeout))
 2025-12-14T10:18:22.711519Z ERROR Sync cycle failed, exiting as the error is irrecoverable time_elapsed_s=2174 sync_head=0x0823…ff77 error=Peer handler error: Error in Peer Table: Internal error: Call to GenServer timed out
```

PR #5641, opened 2025-12-15
fix(l1): avoid doing collect in peer table get_contact functions

MegaRedHand:
**Motivation**

Inside the PeerTable we're potentially cloning an entire collection (after filtering).

**Description**

This PR removes a collect we had in `get_contact_for_lookup` and `get_contact_for_enr_lookup`.

Related to #5637 #5646

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -74,6 +74,7 @@
     - [L1MessageSender]()
   - [Based sequencing](./l2/fundamentals/based.md)
   - [Transaction fees](./l2/fundamentals/transaction_fees.md)
+  - [Shared Bridge](./l2/fundamentals/shared_bridge.md)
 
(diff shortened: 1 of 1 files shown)


edg-l (approved):

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -706,10 +706,8 @@ impl PeerTableServer {
                     && !c.disposable
                     && c.is_fork_id_valid != Some(false)
             })
-            .collect::<Vec<_>>()
             .choose(&mut rand::rngs::OsRng)
```
Oppen, line 709: Careful here, I broke this once. This samples differently for an iterator than for a vec, with vec always succeeding if any element matched the filter and that not being guaranteed for iterators.

PR #5647, opened 2025-12-15
perf(l1): avoid loading the whole state trie during validation

MegaRedHand:
**Motivation**

Running a node with debug assertions requires a lot of memory on mainnet.

**Description**

This PR changes validation to be done by checking trie consistency instead of recomputing the trie root. This way we avoid having the whole state trie in memory during validation.


MegaRedHand:
#5693 already includes a version of this

PR #5648, opened 2025-12-15
feat(l1): add `engine_getBlobsV3` rpc method

### LINKED ISSUE #5634: Implement `engine_getBlobsV3`
See https://github.com/ethereum/execution-apis/pull/719/changes for the spec.

https://ethresear.ch/t/gossipsubs-partial-messages-extension-and-cell-level-dissemination/23017/1 for more context

lakshya-sky:
**Motivation**

Adds new rpc method almost identical to `engine_getBlobsV2` expect the new method allows partial data as opposed to all or nothing.

**Description**

- create & share a new function for both `engine_getBlobsV{2,3} as they are identical.

Closes #5634


crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,83 @@ impl RpcHandler for BlobsV2Request {
...
+    }
+}
 
-        for blobs_bundle in context.blockchain.mempool.get_blobs_bundle_pool()? {
-            // Go over all blobs bundles from the blobs bundle pool.
-            let blobs_in_bundle = blobs_bundle.blobs;
-            let commitments_in_bundle = blobs_bundle.commitments;
-            let proofs_in_bundle = blobs_bundle.proofs;
+// Get blob data for a given list of blob versioned hashes.
+async fn get_blobs_and_proof_v2(
```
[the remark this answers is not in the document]
MegaRedHand: This. We should remove the `_v2` from the name

crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,83 @@ impl RpcHandler for BlobsV2Request {
...
+        serde_json::to_value(res).map_err(|error| RpcErr::Internal(error.to_string()))
+    }
+}
 
-        for blobs_bundle in context.blockchain.mempool.get_blobs_bundle_pool()? {
-            // Go over all blobs bundles from the blobs bundle pool.
-            let blobs_in_bundle = blobs_bundle.blobs;
-            let commitments_in_bundle = blobs_bundle.commitments;
-            let proofs_in_bundle = blobs_bundle.proofs;
+// Get blob data for a given list of blob versioned hashes.
```
MegaRedHand: It's better to use a rustdoc here (`///`)

```suggestion
/// Get blob data and proofs for a given list of blob versioned hashes.
```

crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,83 @@ impl RpcHandler for BlobsV2Request {
...
+            blob_versioned_hashes: serde_json::from_value(params[0].clone())?,
+        })
+    }
 
-        let mut res: Vec<Option<BlobAndProofV2>> = vec![None; self.blob_versioned_hashes.len()];
+    async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        info!("Received new engine request: Requested Blobs V3");
+        let res = get_blobs_and_proof_v2(&self.blob_versioned_hashes, context).await?;
+        serde_json::to_value(res).map_err(|error| RpcErr::Internal(error.to_string()))
+    }
```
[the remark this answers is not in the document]
MegaRedHand: This. Let's add a test. In this PR or another one.

crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,83 @@ impl RpcHandler for BlobsV2Request {
...
+        .storage
+        .get_block_header(context.storage.get_latest_block_number().await?)?
+        && !context
+            .storage
+            .get_chain_config()
+            .is_osaka_activated(current_block_header.timestamp)
+    {
+        // validation requested in https://github.com/ethereum/execution-apis/blob/a1d95fb555cd91efb3e0d6555e4ab556d9f5dd06/src/engine/osaka.md?plain=1#L130
+        return Err(RpcErr::UnsuportedFork(
+            "getBlobsV2 engine only supported for Osaka".to_string(),
```
MegaRedHand: Let's add a `version: u64` parameter and use it here for the error message.

MegaRedHand:
Thanks for your contribution! Left some comments, but should be good to merge once those are addressed

crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,185 @@ impl RpcHandler for BlobsV2Request {
...
+        let mut storage =
+            Store::new("test-blobs", EngineType::InMemory).expect("Failed to create test store");
+        storage
+            .set_chain_config(&chain_config(osaka_active))
+            .await
+            .expect("Failed to set chain config");
+        default_context_with_storage(storage).await
+    }
+
+    #[tokio::test]
```
lakshya-sky, line 279: Hey @MegaRedHand, I added few tests, do they seem useful?

MegaRedHand: They are useful!

MegaRedHand: Can we add one that's the same as `blobs_v3_returns_partial_results`, but for v2? (with a different result ofc)

MegaRedHand (approved):
Great work!! 😄

crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,205 @@ impl RpcHandler for BlobsV2Request {
...
+                // If the versioned hash is one of the requested we save its corresponding blob and proof in the returned vector. We store them in the same position as the versioned hash was received.
+                res[index] = Some(BlobAndProofV2 {
+                    blob: *blob,
+                    proofs: proofs.to_vec(),
+                });
             }
         }
-        if res.iter().any(|blob| blob.is_none()) {
-            return Ok(Value::Null);
+    }
```
Oppen, line 197: This looks like a lot of work to do for a single RPC call. We should probably, at the very least, hash once when we accept the blob tx to the mempool and keep that hash, rather than hash all of them each time someone calls this endpoint. Looks like a DoS vector to me.

NOTE: this is not necessarily the place to fix it, being already there, it just got it to my attention.
@MegaRedHand should we open an issue for this?

crates/networking/rpc/engine/blobs.rs
```diff
@@ -110,55 +115,205 @@ impl RpcHandler for BlobsV2Request {
...
+            if let Some(index) = blob_versioned_hashes
+                .iter()
+                .position(|&hash| hash == current_versioned_hash)
+            {
+                // If the versioned hash is one of the requested we save its corresponding blob and proof in the returned vector. We store them in the same position as the versioned hash was received.
+                res[index] = Some(BlobAndProofV2 {
+                    blob: *blob,
+                    proofs: proofs.to_vec(),
+                });
             }
```
Oppen, line 195: This search also looks rather worrisome in terms of performance. We iterate all blobs, hash them, then iterate all hashes in the request each time. I suggest we just build a hashmap mapping hash to desired position (pseudo-code):
```rust
let mut res = vec![None; blob_versioned_hashes.len()];
let versioned_hashes_to_pos = HashMap::from_iter(blob_versioned_hashes.iter().enumerate().map(|(i, h)| (*h, i)));
for blob in blobs_pool {
    let hash = blob.hopefully_precomputed_sha2();
    let Some(pos) = versioned_hashes_to_pos.get(&hash) else {
         continue;
    };
    res[pos] = build_the_thing(blob);
}
```
Extra points for keeping a mapping of versioned hash to blob as well.

Again, not to handle now, just pointing out the issues in the original code.

Oppen (approved):

crates/networking/rpc/engine/blobs.rs
MegaRedHand: Let's. I think we should also add a `get_blobs(versioned_hashes)` method for fetching the blobs with those versioned hashes only.

lakshya-sky: I will raise a PR with the a new method once the issue is up. Could you assign it to me when you create the issue @MegaRedHand. Thanks!

MegaRedHand: https://github.com/lambdaclass/ethrex/issues/5679

crates/networking/rpc/engine/blobs.rs
MegaRedHand: Should also be addressed by https://github.com/lambdaclass/ethrex/issues/5679

PR #5655, opened 2025-12-16
feat(l1): discv5

### LINKED ISSUE #4491: [Tracking] Implement discv5

ElFantasma:
**Motivation**

Ethereum nodes are expected to run `discv5` for early 2026. 

**Description**

On one hand, this is work in progress, but in the other we don't want to have a huge PR at the end of development. So we plan to start merging `discv5` work into main, but behind a feature flag so nothing is compiled or run, except for development. The flag is already present in this commit, so we should be clear to merge.

**Checklist**

Closes #4491

### DIFF
--- a/.github/workflows/daily_hive_report.yaml
+++ b/.github/workflows/daily_hive_report.yaml
@@ -115,6 +115,11 @@ jobs:
             FLAGS+=" --sim.buildarg branch=forks/osaka"
             FLAGS+=" --client.checktimelimit=180s"
           fi
+
+          if [[ "$SIMULATION" == "ethereum/rpc-compat" ]]; then
+            FLAGS+=" --sim.buildarg branch=d08382ae5c808680e976fce4b73f4ba91647199b"
+          fi
+
           if [[ -n "$SIM_LIMIT" ]]; then
             FLAGS+=" --sim.limit \"$SIM_LIMIT\""
           fi
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3580,7 +3580,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "anyhow",
  "bytes",
@@ -3631,7 +3631,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-benches"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bytes",
  "criterion",
@@ -3649,7 +3649,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-blockchain"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bytes",
  "ethrex-common",
@@ -3670,7 +3670,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-common"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bytes",
  "crc32fast",
@@ -3701,7 +3701,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-config"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "ethrex-common",
  "ethrex-p2p",
@@ -3712,7 +3712,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-crypto"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "c-kzg",
  "kzg-rs",
@@ -3723,7 +3723,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-dev"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bytes",
  "envy",
@@ -3742,7 +3742,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-l2"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "aligned-sdk",
  "anyhow",
@@ -3800,7 +3800,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-l2-common"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "bytes",
  "ethereum-types 0.15.1",
@@ -3823,7 +3823,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-l2-rpc"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "axum 0.8.7",
  "bytes",
@@ -3852,7 +3852,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-levm"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "ark-bn254",
  "ark-ec",
@@ -3889,7 +3889,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-metrics"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "axum 0.8.7",
  "ethrex-common",
@@ -3904,7 +3904,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-p2p"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "aes",
  "aes-gcm",
@@ -3950,7 +3950,7 @@ dependencies = [
 
 [[package]]
 name = "ethrex-prover"
-version = "7.0.0"
+version = "8.0.0"
 dependencies = [
  "anyhow",
  "bincode 1.3.3",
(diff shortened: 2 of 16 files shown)


crates/networking/p2p/discv5/messages.rs
```diff
@@ -0,0 +1,1212 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum PacketDecodeErr {
+    #[error("RLP decoding error")]
+    RLPDecodeError(#[from] RLPDecodeError),
+    #[error("Invalid packet size")]
+    InvalidSize,
+    #[error("Invalid protocol: {0}")]
+    InvalidProtocol(String),
+    #[error("Stream Cipher Error: {0}")]
+    ChipherError(String),
```
Oppen:
```suggestion
    CipherError(String),
```

crates/networking/p2p/discv5/messages.rs
```diff
@@ -0,0 +1,1212 @@
...
+    WhoAreYou(WhoAreYou),
+    Handshake(Handshake),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PacketHeader {
+    pub static_header: Vec<u8>,
+    pub flag: u8,
+    pub nonce: Vec<u8>,
+    pub authdata: Vec<u8>,
```
Oppen: Why not arrays if they are constant size?

crates/networking/p2p/discv5/messages.rs
```diff
@@ -0,0 +1,1212 @@
...
+    /// Offset in the encoded packet where authdata ends, i.e where the header ends.
+    pub header_end_offset: usize,
+}
+
+impl Packet {
+    pub fn decode(
+        dest_id: &H256,
+        decrypt_key: &[u8],
```
Oppen: Are the keys actually variable size or could they be arrays/references to arrays?

crates/networking/p2p/discv5/messages.rs
```diff
@@ -0,0 +1,1212 @@
...
+impl PingMessage {
+    pub fn new(req_id: Vec<u8>, enr_seq: u64) -> Self {
+        Self { req_id, enr_seq }
+    }
+}
+
+impl RLPEncode for PingMessage {
+    fn encode(&self, buf: &mut dyn BufMut) {
+        Encoder::new(buf)
+            .encode_field(&Bytes::from(self.req_id.clone()))
```
Oppen:
```suggestion
            .encode_field(self.req_id.as_slice())
```

crates/networking/p2p/discv5/messages.rs
```diff
@@ -0,0 +1,1212 @@
...
+            .encode_field(&self.enr_seq)
+            .finish();
+    }
+}
+
+impl RLPDecode for PingMessage {
+    fn decode_unfinished(rlp: &[u8]) -> Result<(Self, &[u8]), RLPDecodeError> {
+        let ((req_id, enr_seq), remaining): ((Bytes, u64), &[u8]) =
```
Oppen:
```suggestion
        let ((req_id, enr_seq), remaining): ((&[u8], u64), &[u8]) =
```

crates/networking/p2p/discv5/messages.rs
```diff
@@ -0,0 +1,1212 @@
...
+            }
+            _ => Err(RLPDecodeError::MalformedData),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PingMessage {
+    /// The request id of the sender.
+    pub req_id: Vec<u8>,
```
Oppen: Shouldn't this be `u64` like in other messages?

cmd/ethrex/Cargo.toml
```diff
@@ -101,6 +101,8 @@ jemalloc_profiling = [
   "ethrex-rpc/jemalloc_profiling",
 ]
 sync-test = ["ethrex-p2p/sync-test"]
+# discv5 is currently experimental and should only be enabled for development purposes
+discv5 = ["ethrex-p2p/discv5"]
```
Oppen: Why not call the feature itself `experimental-discv5`? A comment is not user visible and thus far less obvious than the name of what they're activating calling it out as such.

crates/networking/p2p/discv5/messages.rs
edg-l: authdata is variable sized

you are right about nonce though i think

crates/networking/p2p/discv5/session.rs
```diff
@@ -0,0 +1,196 @@
...
+pub struct SessionKeys {
+    pub initiator_key: [u8; 16],
+    pub recipient_key: [u8; 16],
+}
+
+/// A discv5 session
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Session {
+    pub keys: SessionKeys,
+    pub role: SessionRole,
```
MegaRedHand, line 26: Does this need to be kept? I think the role can be removed if we switch the keys in one of the two cases (maybe rename the keys to local and remote too).

crates/networking/p2p/discv5/session.rs
```diff
@@ -0,0 +1,196 @@
...
+///
+/// https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#identity-specific-cryptography-in-the-handshake
+fn compressed_shared_secret(dest_pubkey: &PublicKey, ephemeral_key: &SecretKey) -> [u8; 33] {
+    let xy_point = shared_secret_point(dest_pubkey, ephemeral_key);
+    let mut compressed = [0u8; 33];
+    let y = &xy_point[32..];
+    compressed[0] = if y[31] & 1 == 0 { 0x02 } else { 0x03 };
+    compressed[1..].copy_from_slice(&xy_point[..32]);
+    compressed
+}
```
MegaRedHand, line 117: I think we already had ECDH for RLPx

crates/networking/p2p/Cargo.toml
```diff
@@ -46,9 +46,12 @@ serde_json = "1.0.117"
 concat-kdf = "0.1.0"
 hmac = "0.12.1"
 aes = "0.8.4"
+aes-gcm = "0.10.3"
```
MegaRedHand: We should make these optional

```suggestion
aes-gcm = { version = "0.10.3", optional = true }
```

Also, we should group them up and add a comment mentioning they're optional due to discv5

crates/networking/p2p/Cargo.toml
```diff
@@ -46,9 +46,12 @@ serde_json = "1.0.117"
...
 ctr = "0.9.2"
 rand = "0.8.5"
 
+hkdf = "0.12.4"
```
MegaRedHand:
```suggestion
hkdf = { version = "0.12.4", optional = true }
```

crates/networking/p2p/Cargo.toml
```diff
@@ -65,6 +68,8 @@ sync-test = []
 l2 = ["dep:ethrex-storage-rollup"]
 test-utils = []
 metrics = ["dep:ethrex-metrics"]
+# discv5 is currently experimental and should only be enabled for development purposes
+experimental-discv5 = []
```
MegaRedHand:
```suggestion
experimental-discv5 = ["dep:aes-gcm", "dep:hkdf"]
```

cmd/ethrex/Cargo.toml
```diff
@@ -101,6 +101,8 @@ jemalloc_profiling = [
   "ethrex-rpc/jemalloc_profiling",
 ]
 sync-test = ["ethrex-p2p/sync-test"]
+# discv5 is currently experimental and should only be enabled for development purposes
+experimental-discv5 = ["ethrex-p2p/experimental-discv5"]
```
MegaRedHand, line 106: We might want to add a warning on startup when this is enabled

MegaRedHand (approved):

crates/networking/p2p/discv5/session.rs
edg-l: shared_secret_point is from the ecdh module, the rest of the code are details from the spec

crates/networking/p2p/discv5/messages.rs
ElFantasma: Spec says:
> request-id is an RLP byte array of length <= 8 bytes.

So probably we need to change the other messages. (or fix encode/decode for each one)

crates/networking/p2p/discv5/session.rs
ElFantasma: I will change it in a future PR

PR #5656, opened 2025-12-16
chore(l1): add a message related to a snapsync edge case

rodrigo-o:
**Motivation**

If we cant get storages after some retries we just mark everything for healing. This could take a really long time, we want to make it clear.

**Description**

Add a warn log related to this edge case

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -786,7 +786,7 @@ impl Syncer {
                     }
 
                     warn!("Storage could not be downloaded after multiple attempts. Marking for healing.
-                        This could severely impact snap sync time, please consider stopping the node and resyncing.");
+                        This could impact snap sync time (healing may take a while).");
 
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/sync.rs
```diff
@@ -785,6 +785,9 @@ impl Syncer {
                         );
                     }
 
+                    warn!("Storage could not be downloaded after multiple attempts. Marking for healing.
+                        This could severely impact snap sync time, please consider stopping the node and resyncing.");
```
MegaRedHand: I think this can be a bit confusing. How about:


```suggestion
                        This could impact snap sync time, consider restarting the node and resyncing if healing takes more than 30m.");
```

rodrigo-o: We already checked async and removed the whole resync/restart bit

MegaRedHand (approved):

PR #5661, opened 2025-12-16
fix(l1): wait for background threads before closing RocksDB

### LINKED ISSUE #5658: `test_rocksdb_store` is flaky
`test_rocksdb_store` is failing in CI without really failing, printing "pthread lock: Invalid argument" and exiting with a non-zero exit code. This problem can be reproduced locally by running:

```bash
#!/bin/bash
for i in $(seq 1 1000) ; do
    echo =========== TEST RUN $i ===========
    cargo test -p ethrex-storage -F rocksdb
    if [[ $? -ne 0 ]]; then
        echo
        echo "Failed in run $i"
        exit 1
    fi
done
```

The problem seems to be related with background threads trying to interact with the DB after it was dropped, either by us [or RocksDB's background workers](https://github.com/facebook/rocksdb/issues/11349#issuecomment-2587107605). Adding a sleep at the end of the test seems to solve this, but I think we should try to see if this can be solved by waiting for _our_ background workers instead.

See https://github.com/lambdaclass/ethrex/pull/5621 for a previous attempt to fix this

PivasDesant:
Fixes #5658

Adds graceful shutdown mechanism for Store background threads to prevent 
"pthread lock: Invalid argument" errors when RocksDB is closed while 
background threads are still accessing it.

Changes:
- Store now tracks background thread handles
- Added shutdown() method to wait for threads before closing DB
- Tests call shutdown() for RocksDB before cleanup

The test now passes consistently across multiple runs.

### DIFF
--- a/.github/workflows/pr-main_l1.yaml
+++ b/.github/workflows/pr-main_l1.yaml
@@ -106,7 +106,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          cargo test --workspace --exclude ethrex-l2
+          cargo test --workspace --exclude 'ethrex-l2*' --exclude ethrex-prover --exclude guest_program
 
       - name: Run Blockchain EF tests
         if: ${{ github.event_name != 'merge_group' }}
--- a/crates/l2/prover/src/guest_program/src/openvm/Cargo.lock
+++ b/crates/l2/prover/src/guest_program/src/openvm/Cargo.lock
@@ -917,6 +917,7 @@ dependencies = [
  "ethrex-rlp",
  "ethrex-trie",
  "hex",
+ "hex-literal 0.4.1",
  "k256 0.13.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "kzg-rs",
  "lazy_static",
@@ -965,6 +966,7 @@ dependencies = [
  "serde_with",
  "sha3",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/l2/prover/src/guest_program/src/risc0/Cargo.lock
+++ b/crates/l2/prover/src/guest_program/src/risc0/Cargo.lock
@@ -1166,6 +1166,7 @@ dependencies = [
  "ethrex-rlp",
  "ethrex-trie",
  "hex",
+ "hex-literal",
  "k256",
  "kzg-rs",
  "lazy_static",
@@ -1214,6 +1215,7 @@ dependencies = [
  "serde_with",
  "sha3",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/l2/prover/src/guest_program/src/sp1/Cargo.lock
+++ b/crates/l2/prover/src/guest_program/src/sp1/Cargo.lock
@@ -955,6 +955,7 @@ dependencies = [
  "ethrex-rlp",
  "ethrex-trie",
  "hex",
+ "hex-literal",
  "k256",
  "kzg-rs",
  "lazy_static",
@@ -1004,6 +1005,7 @@ dependencies = [
  "serde_with",
  "sha3",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
@@ -1393,6 +1395,12 @@ checksum = "5313b072ce3c597065a808dbf612c4c8e8590bdbf8b579508bf7a762c5eae6cd"
 dependencies = [
  "arrayvec",
 ]
+
+[[package]]
+name = "hex-literal"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6fe2267d4ed49bc07b63801559be28c718ea06c4738b7a03c94df7386d2cde46"
 
 [[package]]
 name = "hkdf"
--- a/crates/l2/prover/src/guest_program/src/zisk/Cargo.lock
+++ b/crates/l2/prover/src/guest_program/src/zisk/Cargo.lock
@@ -913,6 +913,7 @@ dependencies = [
  "ethrex-rlp",
  "ethrex-trie",
  "hex",
+ "hex-literal",
  "k256",
  "kzg-rs",
  "lazy_static",
@@ -960,6 +961,7 @@ dependencies = [
  "serde_with",
  "sha3",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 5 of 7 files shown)


crates/storage/store.rs
```diff
@@ -145,6 +145,42 @@ pub struct Store {
...
+            chain_config: self.chain_config.clone(),
+            trie_cache: self.trie_cache.clone(),
+            flatkeyvalue_control_tx: self.flatkeyvalue_control_tx.clone(),
+            trie_update_worker_tx: self.trie_update_worker_tx.clone(),
+            latest_block_header: self.latest_block_header.clone(),
+            last_computed_flatkeyvalue: self.last_computed_flatkeyvalue.clone(),
+            account_code_cache: self.account_code_cache.clone(),
+            background_threads: self.background_threads.clone(),
+        }
+    }
```
MegaRedHand, line 187: Why is this needed?

crates/storage/store.rs
```diff
@@ -2930,8 +3010,14 @@ mod tests {
...
         // Build a new store
         let store = Store::new(&path, engine_type).expect("Failed to create test db");
-        // Run the test
-        test_func(store).await;
+        // Run the test and get the store back
+        let store = test_func(store).await;
+        // For RocksDB, we need to shutdown the store to wait for background threads
+        // before dropping it, otherwise RocksDB's background threads may try to
+        // access the database after it's been dropped, causing "pthread lock: Invalid argument" errors.
+        if matches!(engine_type, EngineType::RocksDB) {
```
MegaRedHand: We should do this for all backends (including `InMemory`).

crates/storage/store.rs
```diff
@@ -1472,6 +1517,41 @@ impl Store {
         Ok(store)
     }
 
+    /// Shutdown the store and wait for background threads to finish.
+    /// This should be called before dropping the Store to ensure all background
+    /// threads have completed their work and RocksDB can be safely closed.
+    ///
+    /// This method consumes the Store. After calling shutdown, the Store should not be used.
+    pub fn shutdown(self) {
```
MegaRedHand, line 1424: Let's also call this when the node is shutting down.

crates/storage/store.rs
PivasDesant: Store gets cloned in a few places - mainly in async contexts and iterators like `AncestorIterator`. 

Since the backend and background threads are shared via Arc, cloning is safe and cheap. 

I added a doc comment to explain this.

MegaRedHand: Yes, but can we derive it instead of manually implementing it?

cmd/ethrex/l2/initializers.rs
```diff
@@ -344,6 +347,9 @@ pub async fn init_l2(
         store_node_config_file(node_config, node_config_path);
     }
     tokio::time::sleep(Duration::from_secs(1)).await;
+    info!("Shutting down stores...");
+    store_for_shutdown.shutdown();
+    rollup_store_for_shutdown.shutdown();
```
MegaRedHand: Let's shut down the rollup store first, since it might depend on the store.


```suggestion
    rollup_store_for_shutdown.shutdown();
    store_for_shutdown.shutdown();
```

crates/storage/store.rs
```diff
@@ -1347,6 +1389,13 @@ impl Store {
                 }
             }
         });
+
+        // Store the thread handles
+        *background_threads.lock().unwrap() = Some(BackgroundThreadHandles {
```
MegaRedHand: Let's explain why this `unwrap` won't panic:

```suggestion
        *background_threads.lock().expect("we just created this mutex") = Some(BackgroundThreadHandles {
```

crates/storage/store.rs
```diff
@@ -1365,6 +1414,41 @@ impl Store {
...
 
+    /// Shutdown the store and wait for background threads to finish.
+    /// This should be called before dropping the Store to ensure all background
+    /// threads have completed their work and the backend can be safely closed.
+    ///
+    /// This method consumes the Store. After calling shutdown, the Store should not be used.
+    pub fn shutdown(self) {
+        // Take the thread handles out of the Mutex before dropping self
+        let handles = {
+            let mut threads = self.background_threads.lock().unwrap();
```
MegaRedHand: We can avoid the `unwrap` here:

```suggestion
            let mut threads = self.background_threads.lock().unwrap_or_else(|e| e.into_inner());
```

MegaRedHand:
@PivasDesant There are some conflicts, please resolve.

Also, Clippy is complaining about a `clone`: https://github.com/lambdaclass/ethrex/actions/runs/20374909357/job/58550732084?pr=5661
To reproduce locally, run `make lint`

PivasDesant:
added shutdown method to StoreRollup

Oppen:
The git helper changes seem completely unrelated. Why are they here?

Why is this a `shutdown` method and not just a `Drop` implementation?

crates/storage/store.rs
```diff
@@ -145,6 +145,46 @@ pub struct Store {
     /// those changes already affect the code hash stored in the account, and only
     /// may result in this cache having useless data.
     account_code_cache: Arc<CodeCache>,
+
+    /// Handles to background threads for graceful shutdown
+    /// Wrapped in Arc<Mutex<>> to allow sharing across clones
+    background_threads: Arc<Mutex<Option<BackgroundThreadHandles>>>,
```
Oppen, line 151: This will only be accessed once, to join, so the `Mutex` is unnecessary.
The `Option` seems unnecessary as well.

crates/storage/store.rs
```diff
@@ -145,6 +145,46 @@ pub struct Store {
...
+            chain_config: self.chain_config,
+            trie_cache: self.trie_cache.clone(),
+            flatkeyvalue_control_tx: self.flatkeyvalue_control_tx.clone(),
+            trie_update_worker_tx: self.trie_update_worker_tx.clone(),
+            latest_block_header: self.latest_block_header.clone(),
+            last_computed_flatkeyvalue: self.last_computed_flatkeyvalue.clone(),
+            account_code_cache: self.account_code_cache.clone(),
+            background_threads: self.background_threads.clone(),
+        }
+    }
```
Oppen, line 187: Just `#[derive(Clone)]`.

crates/storage/store.rs
```diff
@@ -145,6 +145,46 @@ pub struct Store {
...
+    /// Handles to background threads for graceful shutdown
+    /// Wrapped in Arc<Mutex<>> to allow sharing across clones
+    background_threads: Arc<Mutex<Option<BackgroundThreadHandles>>>,
+}
+
+/// Handles for background threads that need to be joined on shutdown
+struct BackgroundThreadHandles {
+    flatkeyvalue_thread: std::thread::JoinHandle<()>,
+    trie_update_thread: std::thread::JoinHandle<()>,
+}
```
Oppen, line 158:
```suggestion
struct BackgroundThreadHandles {
    flatkeyvalue_thread: std::thread::JoinHandle<()>,
    trie_update_thread: std::thread::JoinHandle<()>,
}

impl Drop for BackgroundThreadHandles {
    fn drop(&mut self) {
        _ = self.flatkeyvalue_thread.join();
        _ = self.trie_update_thread.join();
    }
}
```

No need for `shutdown` AFAICT if you do this.

Arkenan:
Hey @PivasDesant thanks a lot for taking the time to analyze this issue and propose a solution. As we need this feature urgently because it blocks @iovoid 's high priority PR on parallelized merkelization, he will take over this issue in a new PR. Thanks for understanding!

PR #5672, opened 2025-12-17
chore(l1): move entire cache behind mutex and remove atomic 

### LINKED ISSUE #5612: Minimize critical section size in account code cache
See https://github.com/lambdaclass/ethrex/pull/4623#discussion_r2612047016

The current account code cache implementation can be improved to reduce contention.

lakshya-sky:
**Motivation**

Currently code cache uses both Mutex and AtomicU64 which is not necessary, and only mutex is enough. 

**Description**

move core logic in a new `InnerCodeCache` which is put behind a mutex.

Closes #5612

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -174,11 +174,6 @@ pub struct Store {
     /// those changes already affect the code hash stored in the account, and only
     /// may result in this cache having useless data.
     account_code_cache: Arc<Mutex<CodeCache>>,
-    /// LRU cache for contract bytecode, keyed by code hash.
-    ///
-    /// Entries are not removed on code changes since the code hash
-    /// in the account already reflects the change.
-    account_code_cache: Arc<CodeCache>,
 
     background_threads: Arc<ThreadList>,
 }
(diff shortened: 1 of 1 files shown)


crates/storage/store.rs
```diff
@@ -75,54 +74,73 @@ enum FKVGeneratorControlMessage {
 // 64mb
 const CODE_CACHE_MAX_SIZE: u64 = 64 * 1024 * 1024;
 
-// TODO: don't use atomic here, instead wrap in Mutex the whole cache
 #[derive(Debug)]
-struct CodeCache {
-    inner_cache: Mutex<LruCache<H256, Code, FxBuildHasher>>,
-    cache_size: AtomicU64,
+struct InnerCodeCache {
```
MegaRedHand: Let's keep this as `CodeCache`, and move the `Mutex` to the `Store`:

```rust
struct CodeCache {...}

struct Store {
    ...
    account_code_cache: Arc<Mutex<CodeCache>>
    ...
}
```

lakshya-sky: done.

iovoid (approved):

PR #5678, opened 2025-12-18
docs(l2): exit windows fundamentals

ilitteri:


docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,59 @@
+# Exit Window
+
+An exit window is a time window, or period, during which users can opt to exit the network before the execution of an upgrade or system modification. The purpose of exit windows in L2 rollups is to protect users from unwanted changes to the system, such as those mentioned above.
+
+The [Stages Framework](https://medium.com/l2beat/introducing-stages-a-framework-to-evaluate-rollups-maturity-d290bb22befe) defines 7-day exit windows for Stage 1 rollups and 30-day exit windows for Stage 2 rollups. This period may vary if there is a withdrawal delay, as it is subtracted from the total exit window.
```
JereSalo: We can give a little bit more details on the exit window thing because it's not just 7 days vs 30 days, there are some subtle differences.

Stage 1: If updates are initiated out of the Security Council the exit window should be at least 7 days. But the SC can upgrade instantly
Stage 2: Security Council can upgrade immediately only if there's a bug detected on-chain. Otherwise the exit window should be at least 30 days.

Also, maybe we can replace the medium link with the current source of truth, given that the former is outdated. 
https://forum.l2beat.com/t/the-stages-framework/291

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,59 @@
...
+
+The goal of the settlement window is to give enough time to the rollup operator to react in the event of a bug exploit in the L2 before the state is settled on the L1 and, thus, irreversible.
+
+As said before, the settlement window must be taken into account to calculate the real exit window.
+
+## Who owns the `Timelock`
+
+Whoever owns the `Timelock` decides its functioning. In our stack, the owner of the contract is established during its initialization, and then that owner can transfer the ownership to another account if desired.
+
+Our recommendation is that the owner of the contract be a security council. The [Stages Framework](https://medium.com/l2beat/introducing-stages-a-framework-to-evaluate-rollups-maturity-d290bb22befe) recommends that the security council be in the form of a multisig composed of at least 8 people with a consensus threshold of 50%, and at least 50% of the participants must be external to the organization that operates the rollup.
```
JereSalo: Consensus threshold recommended for SC is 75%, see [this](https://forum.l2beat.com/t/the-stages-framework/291)

JereSalo: I don't think what we call "owner" should be the security council. The SC can execute things immediately, and what we call owner just have roles of Proposer/Executor so it doesn't have permissions to bypass the minDelay. Maybe we should change names though, currently we have owner and security_council. Maybe the former shouldn't be called owner so that it's not misleading.

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,59 @@
...
+## How are `Timelock` upgrades protected
+
+Today, we only expose two types of modifications to our `Timelock` contract:
+
+1. Upgrade of the contract itself.
+2. Update of the delay time.
+
+An interesting question is how we protect users from instantaneous executions of these types of operations. One might think of a scenario where the `Timelock` owner updates the delay to 0 to execute a malicious operation, having removed the exit window for users; or perhaps upgrade the `Timelock` contract by removing the delay logic with the same objective.
+
+We solve this by making the `Timelock` itself the only one capable of invoking the corresponding functions for those operations. In this way, if the contract owner wants to push a malicious upgrade or modification to the “protector” contract, they must comply with the configured delay at the time of proposing the operation. For example, if a malicious or compromised owner wishes to set the delay to 0, they must wait the duration of the current exit window for the execution of their malicious operation, thus giving users sufficient time to exit the network.
```
JereSalo, line 61: Note that the Security Council can `emergencyExecute` and update the delay time to zero.
In case we don't want this behavior what we can do is to enable `emergencyExecute` only if the target address is different than the timelock contract itself. That means that the SC won't be able to upgrade the Timelock instantly in case of an emergency though, which may backfire.
For a rollup that is not mature it isn't a critical thing, but as the rollup becomes more mature we should consider a more sophisticated approach

JereSalo (commented):
I changed the name of `owner` in the Timelock contract to `governance` to make it more clear. I think it's more accurate given that they don't have permissions to bypass the delay unlike the Security Council.
Maybe we can change it again afterwards if we think there's a better name instead. Let me know what you think

docs/l2/fundamentals/exit_window.md
ilitteri: I agree.

ilitteri: LMK what you think https://github.com/lambdaclass/ethrex/pull/5678/commits/bf17a3bca46c9168c07c4dba6493ea996221941d

docs/l2/fundamentals/exit_window.md
ilitteri: [b732058](https://github.com/lambdaclass/ethrex/pull/5678/commits/b7320580181739da0613530fa7e358bf39972c21)

ilitteri: Yes, I'll review this part and clarify it.

docs/l2/fundamentals/exit_window.md
JereSalo: It's excellent

docs/l2/fundamentals/exit_window.md
ilitteri: [607914b](https://github.com/lambdaclass/ethrex/pull/5678/commits/607914b37279ce205eec7d1601650336c6341515)

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,63 @@
...
+
+The goal of the settlement window is to give enough time to the rollup operator to react in the event of a bug exploit in the L2 before the state is settled on the L1 and, thus, irreversible.
+
+As said before, the settlement window must be taken into account to calculate the real exit window.
+
+## Who owns the `Timelock`
+
+There's no such thing as a unique owner of the `Timelock` necessarily. `Timelock` is a `TimelockController`, which is also an `AccessControl`, so we can define different roles and assign them to different entities. By "owner" of the `Timelock`, we refer to the account that has the role to update the contract (i.e. the one that can modify the delay).
+
+This said, whoever owns the `Timelock` decides its functioning. In our stack, the owner of the contract is established during its initialization, and then that owner can transfer the ownership to another account if desired.
```
tomip01:
```suggestion
That said, whoever owns the `Timelock` decides its functioning. In our stack, the owner of the contract is established during its initialization, and then that owner can transfer the ownership to another account if desired.
```

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,63 @@
...
+
+Whoever owns the `Timelock` decides its functioning. In our stack, the owner of the contract is established during its initialization, and then that owner can transfer the ownership to another account if desired.
+
```
tomip01: This line is duplicated

```suggestion
```

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,63 @@
...
+It's worth noting that the designated security council can execute operations instantly in case of emergencies, so it's crucial that the members are trustworthy and committed to the network's security. The [Stages Framework](https://forum.l2beat.com/t/the-stages-framework/291) recommends that the security council be in the form of a multisig composed of at least 8 people with a consensus threshold of 75%, and the participants in the set need to be decentralized and diverse enough, possibly coming from different companies and jurisdictions.
+
+In the case of our `Timelock`, the owner is not the only one who can act on it. In fact, it is recommended that the security council only act in specific emergencies. The `Timelock` is also AccessControl, which means it has special functionality for managing accesses, in this case, in the form of roles.
```
tomip01: nit
```suggestion
In the case of our `Timelock`, the owner is not the only one who can act on it. In fact, it is recommended that the security council only act in specific emergencies. The `Timelock` is also `AccessControl`, which means it has special functionality for managing accesses, in this case, in the form of roles.
```

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,63 @@
...
+
+`TimelockController` defines two roles in its business logic: the “proposer” and the “executor.” The first is enabled to schedule operations and cancel them, while the second is enabled to execute them. These roles are assigned to a given set of addresses during the contract’s initialization.
```
tomip01, line 48: also a nit 😄 
```suggestion
`TimelockController` defines two roles in its business logic: the “proposer” and the “executor”. The first is enabled to schedule operations and cancel them, while the second is enabled to execute them. These roles are assigned to a given set of addresses during the contract’s initialization.
```

docs/l2/fundamentals/exit_window.md
```diff
@@ -0,0 +1,63 @@
...
+
+We define two additional roles besides the defaults: the “sequencer” and the “security council.” The first is confined to performing settlement operations (meaning it cannot operate as “proposer” or “executor”), while the second is enabled to revert batches, pause and unpause contracts, and execute emergency operations (i.e., without delay).
```
tomip01, line 50: ditto

```suggestion
We define two additional roles besides the defaults: the “sequencer” and the “security council”. The first is confined to performing settlement operations (meaning it cannot operate as “proposer” or “executor”), while the second is enabled to revert batches, pause and unpause contracts, and execute emergency operations (i.e., without delay).
```

docs/l2/fundamentals/exit_window.md
JereSalo: This is okay, depends on the style of writing

docs/l2/fundamentals/exit_window.md
JereSalo: ditto

PR #5681, opened 2025-12-18
fix(l1): add error handling to init_rpc_api()

### LINKED ISSUE #3821: L1: Error not logged when L1 ports are not available
When you want to launch a local L1 node and the port is not available, initialization does not stop and the error is not logged.
This is replicated by doing:
`make init-local-l1` in the `crates/l2` directory
and then running `RUST_LOG=debug cargo run --bin ethrex -- --network ./fixtures/genesis/kurtosis.json`

figtracer:
closes #3821

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -164,24 +164,39 @@ pub async fn init_rpc_api(
         None
     };
 
-    let rpc_api = ethrex_rpc::start_api(
-        get_http_socket_addr(opts),
-        ws_socket_opts,
-        get_authrpc_socket_addr(opts),
-        store,
-        blockchain,
-        read_jwtsecret_file(&opts.authrpc_jwtsecret),
-        local_p2p_node,
-        local_node_record,
-        syncer,
-        peer_handler,
-        get_client_version(),
-        log_filter_handler,
-        opts.gas_limit,
-        opts.extra_data.clone(),
-    );
+    let http_socket = get_http_socket_addr(opts);
+    let authrpc_socket = get_authrpc_socket_addr(opts);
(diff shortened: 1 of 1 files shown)


cmd/ethrex/initializers.rs
```diff
@@ -164,24 +164,39 @@ pub async fn init_rpc_api(
...
+            log_filter_handler,
+            gas_limit,
+            extra_data,
+        )
+        .await
+        {
+            Ok(()) => info!("RPC API server shut down gracefully"),
+            Err(e) => {
+                error!("Failed to start RPC API: {e}");
+                std::process::exit(1);
```
ElFantasma, line 196: `std::process::exit(1)` immediately terminates the process without running destructors — RocksDB handles won't be cleanly closed, the tracing `WorkerGuard` won't flush (so this very error message may not make it to the log file), and other tasks spawned on the `TaskTracker` are killed without cleanup.

Since `init_rpc_api` already receives a `cancel_token: CancellationToken`, a cleaner approach would be to cancel it here and let the existing shutdown machinery handle cleanup:

```rust
Err(e) => {
    error!("Failed to start RPC API: {e}");
    cancel_token.cancel();
}
```

Alternatively, you could change `init_rpc_api` to return `Result<()>` and propagate the bind error back to `init_l1`, which already uses `eyre::Result`. That would let the caller decide how to shut down — and the bind error would surface before any tasks are spawned, since the `TcpListener::bind` calls in `start_api` happen early.

The second approach would require splitting `start_api` or awaiting it briefly to catch the bind error synchronously before spawning the long-running server future. Either way, `process::exit` should be avoided in favor of cooperative shutdown.

PR #5682, opened 2025-12-18
feat(l1): display ascii art during startup

MegaRedHand:
**Motivation**

Why not?

**Description**

This PR adds some ASCII art to the node initialization.

image showing a terminal with the new ascii art


cmd/ethrex/utils.rs
```diff
@@ -171,6 +171,59 @@ pub fn get_minimal_client_version() -> String {
...
+                   ▒░▒▓▓▒░░░▒▓▓▒▓▓▒▒▓▓▒░░▓▓▓▒▒▓▓▓▓▒▒░░▒▒▒▒▓▒▒█████▓░▒░
+                   ▒░▒██▓░░░▓▓░▒▓▒▒▒▒▒▒▒▒▓▓▒▒▒▒▒▒▒▒▒░░▓▓▒▓▓▓▒██▓▒██░▓░
+                   ▒░▒███▒▒▒▓▓▒▓▓▓▓▓████▓███▓█▓▓▒▓▓▓▒░▓▓░▒██▓██▓▒█▓░▓░
+                   ▒░░▓████▒▓█████▓▓██▒██▓██▓██████▒░▒██▒▓██▓██░░▒░▒░
+                    ░▒░░▓██▒██░░██▒▓██▒░▓▒██▓██▓▓████▒████▒░░░░░▒▒▒░
+                      ░▒░▒▓▒▓█░░██▒▓██▒░░▒██████▒▓███▒██▓░░░▒▒▒░
+                        ░▒▒░░░░░░▒░▓██▒░░▒██▓██████▒░░░░░▒▒░
+                           ░▒▒▒▒░░░░░░░░▒▒░░░░░░░░░░▒▒▒▒░
+                                ░░░░░░░░ ░░░░░░░░░░░
+"#;
```
Oppen: Nit: this might be in a separate text file and imported with `include_str`. That way we could reuse it, e.g. in the tooling CLIs, like rex and the migration tools.

cmd/ethrex/utils.rs
```diff
@@ -171,6 +171,59 @@ pub fn get_minimal_client_version() -> String {
...
+    println!("{BANNER}");
```
ElFantasma: `println!` bypasses the tracing/logging framework, so this banner won't appear in log files and won't respect `--log-level` settings. The rest of this function uses `info!`. Was this intentional (always show the art even at `warn`/`error` level)? If so, a brief comment explaining the choice would help. If not, you could use `info!` for each line, or write to `stderr` via `eprintln!` to avoid interleaving with any stdout-based structured output.

MegaRedHand: Changed to `info!`

cmd/ethrex/utils.rs
MegaRedHand: Done

PR #5684, opened 2025-12-18
refactor(l1): reduce hashing in `engine_getBlobsV2`

### LINKED ISSUE #5679: Reduce hashing in `engine_getBlobsV2`
See https://github.com/lambdaclass/ethrex/pull/5648#discussion_r2631263372

We're currently hashing all the blob commitments in the mempool for each request. Instead, we should store the blobs keyed by versioned hash inside the mempool, to retrieve them inexpensively during `engine_getBlobsV2`.

lakshya-sky:
**Motivation**

We're currently hashing all the blob commitments in the mempool for each request. Instead, this  pr stores the blobs keyed by versioned hash inside the mempool, to retrieve them inexpensively during engine_getBlobsV2

**Description**

- Adds a new map to retrieve blob bundle by their commitment hashes.
- Adds a new method for blob bundle to retrieve data at given index.

Closes #5679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 2026-03-05
 
+- Doubly pipelined merkleization with self-coordinating shard workers [#6278](https://github.com/lambdaclass/ethrex/pull/6278)
 - Switch hot EVM and mempool HashMaps to FxHashMap for faster hashing [#6303](https://github.com/lambdaclass/ethrex/pull/6303)
 
 ### 2026-03-02
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3896,6 +3896,7 @@ name = "ethrex-blockchain"
 version = "9.0.0"
 dependencies = [
  "bytes",
+ "crossbeam 0.8.4",
  "ethrex-common",
  "ethrex-crypto",
  "ethrex-metrics",
--- a/crates/blockchain/Cargo.toml
+++ b/crates/blockchain/Cargo.toml
@@ -22,6 +22,7 @@ tracing.workspace = true
 bytes.workspace = true
 hex.workspace = true
 rustc-hash.workspace = true
+crossbeam.workspace = true
 tokio = { workspace = true, features = ["time", "rt", "sync", "macros"] }
 tokio-util.workspace = true
 
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -59,6 +59,7 @@ use error::MempoolError;
 use error::{ChainError, InvalidBlockError};
 use ethrex_common::constants::{EMPTY_TRIE_HASH, MIN_BASE_FEE_PER_BLOB_GAS};
 
+use crossbeam::channel::{self as cb, TryRecvError, select};
 // Re-export stateless validation functions for backwards compatibility
 #[cfg(feature = "c-kzg")]
 use ethrex_common::types::EIP4844Transaction;
@@ -93,9 +94,10 @@ use ethrex_vm::backends::levm::db::DatabaseLogger;
 use ethrex_vm::{BlockExecutionResult, DynVmDatabase, Evm, EvmError};
 use mempool::Mempool;
 use payload::PayloadOrTask;
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHashSet};
 use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap, HashSet};
+use std::sync::LazyLock;
 use std::sync::mpsc::Sender;
 use std::sync::{
     Arc, RwLock,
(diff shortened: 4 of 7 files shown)


crates/blockchain/mempool.rs
```diff
@@ -45,8 +49,13 @@ impl MempoolInner {
...
     fn remove_transaction_with_lock(&mut self, hash: &H256) -> Result<(), StoreError> {
         if let Some(tx) = self.transaction_pool.get(hash) {
-            if matches!(tx.tx_type(), TxType::EIP4844) {
-                self.blobs_bundle_pool.remove(hash);
+            if matches!(tx.tx_type(), TxType::EIP4844)
+                && let Some(h) = self.blobs_bundle_pool.remove(hash)
+            {
+                for commitment in &h.commitments {
+                    self.blobs_bundle_by_versioned_hash
+                        .remove(&kzg_commitment_to_versioned_hash(commitment));
```
MegaRedHand: Can two transactions share versioned hashes? If that's the case, we'll need to keep a reference counter here.

lakshya-sky: I guess you're right, looking at what Copilot flagged and how Geth implements there could be a possibility that same blob is part of two txns. Let me push a fix for it.

lakshya-sky: Hey @MegaRedHand, I've updated the cache to allow sharing versioned hash. Also I've added a test for insertion and deletion.

crates/blockchain/mempool.rs
```diff
@@ -24,6 +27,7 @@ struct MempoolInner {
     broadcast_pool: HashSet<H256>,
     transaction_pool: HashMap<H256, MempoolTransaction>,
     blobs_bundle_pool: HashMap<H256, BlobsBundle>,
+    blobs_bundle_by_versioned_hash: HashMap<H256, HashMap<H256, usize>>,
```
MegaRedHand: `blobs_bundle_pool` has the tx hash as the key, while `blobs_bundle_by_versioned_hash` has the blob versioned hash as the key, right?
Let's document what each key type represent here

crates/common/types/blobs_bundle.rs
```diff
@@ -14,12 +14,13 @@ use ethrex_rlp::{
...
 use serde::{Deserialize, Serialize};
 
-use super::{BYTES_PER_BLOB, SAFE_BYTES_PER_BLOB};
+use super::{BYTES_PER_BLOB, CELLS_PER_EXT_BLOB, SAFE_BYTES_PER_BLOB};
 
 pub type Bytes48 = [u8; 48];
 pub type Blob = [u8; BYTES_PER_BLOB];
 pub type Commitment = Bytes48;
 pub type Proof = Bytes48;
+pub type BlobTuple = (Box<Blob>, Commitment, Box<[Proof]>);
```
MegaRedHand: Do we need this to be `Box<[Proof]>`? Can't we use `Vec<Proof>` instead? We do an `into_vec` anyways inside `get_blobs_and_proof`.

lakshya-sky: Yes, but my thought was since we are using Box for Blob might be good to use it for proofs as well for consistency. But using a vec is good idea.

crates/blockchain/mempool.rs
lakshya-sky: sure let me fix that

lakshya-sky:
fixes pushed.

crates/blockchain/mempool.rs
```diff
@@ -44,19 +50,40 @@ impl MempoolInner {
...
-            self.broadcast_pool.remove(hash);
+        let Some(tx) = self.transaction_pool.remove(hash) else {
+            return Ok(());
         };
+        if matches!(tx.tx_type(), TxType::EIP4844) {
+            self.remove_blob_bundle(hash);
+        }
+
+        self.txs_by_sender_nonce.remove(&(tx.sender(), tx.nonce()));
+        self.transaction_pool.remove(hash);
```
ElFantasma: Bug: `self.transaction_pool.remove(hash)` on line 53 already removed the entry and returned the value via the `let Some(tx) = ...` binding. This second `.remove(hash)` is a no-op. Should be deleted.

crates/blockchain/mempool.rs
```diff
@@ -154,9 +181,16 @@ impl Mempool {
...
-        self.write()?
-            .blobs_bundle_pool
-            .insert(tx_hash, blobs_bundle);
+        let mut mempool = self.write()?;
+        for (i, c) in blobs_bundle.commitments.iter().enumerate() {
+            let versioned_hash = kzg_commitment_to_versioned_hash(c);
+            mempool
+                .blobs_bundle_by_versioned_hash
+                .entry(versioned_hash)
+                .or_insert(HashMap::new())
```
ElFantasma: nit: `.or_insert(HashMap::new())` → `.or_default()` — avoids the closure and is idiomatic.

lakshya-sky: @ElFantasma I fixed this nits and made changes to the test as well

ElFantasma: Nice job @lakshya-sky!
Do you think it's worth it to add the optimization to BlobsV1Request::handle that still uses the old get_blobs_bundle_pool()? It's a pre-Cancun era endpoint but we can update it at least for consistency, right?

PR #5686, opened 2025-12-19
perf(l1): validate tx blobs after checking if its already in the mempool

edg-l:
**Motivation**

Currently we always validate the tx blobs bundle in add_blob_transaction_to_pool, the validation is expensive and we can avoid it sometimes if we move it after the check of whether the tx is already added to the mempool

Whether this improves performance, theorically yes, its hard to measure but with `perf record` it seems to have gone from 72% to 61% when testing on hoodi

**Checklist**

### DIFF
--- a/.github/workflows/daily_hive_report.yaml
+++ b/.github/workflows/daily_hive_report.yaml
@@ -115,6 +115,11 @@ jobs:
             FLAGS+=" --sim.buildarg branch=forks/osaka"
             FLAGS+=" --client.checktimelimit=180s"
           fi
+
+          if [[ "$SIMULATION" == "ethereum/rpc-compat" ]]; then
+            FLAGS+=" --sim.buildarg branch=d08382ae5c808680e976fce4b73f4ba91647199b"
+          fi
+
           if [[ -n "$SIM_LIMIT" ]]; then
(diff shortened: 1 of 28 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2025-12-19
+
+- Validate tx blobs after checking if its already in the mempool [#5686](https://github.com/lambdaclass/ethrex/pull/5686)
```
MegaRedHand: Typo

```suggestion
- Validate tx blobs after checking if it's already in the mempool [#5686](https://github.com/lambdaclass/ethrex/pull/5686)
```

crates/blockchain/blockchain.rs
```diff
@@ -1400,17 +1400,19 @@ impl Blockchain {
...
 
-        blobs_bundle.validate(&transaction, fork)?;
-
         let transaction = Transaction::EIP4844Transaction(transaction);
         let hash = transaction.hash();
         if self.mempool.contains_tx(hash)? {
             return Ok(hash);
         }
+
+        // Validate blobs bundle after checking if its already added.
```
MegaRedHand:
```suggestion
        // Validate blobs bundle after checking if it's already added.
```

PR #5687, opened 2025-12-19
fix(l1): align mempool error messages with their actual usage

MozirDmitriy:
The human-readable messages for TxGasOverflowError and TxTipAboveFeeCapError were swapped. TxGasOverflowError is only used for arithmetic overflows while computing intrinsic gas in transaction_intrinsic_gas, and TxTipAboveFeeCapError is used when max_priority_fee_per_gas exceeds max_fee_per_gas in validate_transaction. This mismatch produced misleading logs and metrics but did not affect behavior. The change swaps the error annotations so that each variant's message matches its semantic meaning and call sites.


MegaRedHand (approved):
Good catch

ElFantasma (approved):

Arkenan (approved):
LGTM

PR #5693, opened 2025-12-19
feat(l1): expose fkv progress

iovoid:
**Motivation**

To support the geth2ethrex tool, we want to be able to tell the user when FKV generation is complete and how far we are into the process.

**Description**

This PR exposes that information.

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -21,11 +21,6 @@ use crate::{
 };
 
 use bytes::Bytes;
-
-/// Length of the FKV completion marker in bytes.
-/// Structure: account_path (64) + separator (2) + storage_path (64) + end (1) = 131
-pub const FKV_COMPLETION_MARKER_LEN: usize = 131;
-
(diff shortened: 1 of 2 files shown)


crates/storage/store.rs
```diff
@@ -21,6 +21,11 @@ use crate::{
 };
 
 use bytes::Bytes;
+
+/// Length of the FKV completion marker in bytes.
+/// Structure: account_path (64) + separator (2) + storage_path (64) + end (1) = 131
```
ElFantasma: nit: The doc comment breakdown is slightly off: it says `separator (2)` but the separator is a single nibble (value 17) added by `append_new(17)`. The actual breakdown is: account nibbles (64) + account leaf flag (1) + separator (1) + storage nibbles (64) + storage leaf flag (1) = 131. The total is correct, just the intermediate grouping is misleading.

iovoid:
Rolled back all these changes, the user tool can just adapt.

PR #5694, opened 2025-12-19
chore(l2): bump ZisK to 0.15.0 & add modexp syscall

xqft:
0.15.0 introduces a new modexp syscall and fixes for the existing patches, apart from other optimizations (see https://github.com/0xPolygonHermez/zisk/releases/tag/v0.15.0)

we found a regression some time ago in main, probably related to a ZisK patch, in which some blocks failed to execute with a gas mismatch error. We tested those blocks again and the regression was gone. So the new patches' versions contain fixes too.


crates/l2/prover/src/guest_program/src/zisk/Cargo.toml
```diff
@@ -22,13 +22,13 @@ ethrex-blockchain = { path = "../../../../../../blockchain", default-features =
...
-sha3 = { git = "https://github.com/0xPolygonHermez/zisk-patch-hashes.git", tag = "patch-sha3-0.10.8-zisk-0.14.0" }
-k256 = { git = "https://github.com/0xPolygonHermez/zisk-patch-elliptic-curves.git", tag = "patch-k256-0.13.4-zisk-0.14.0" }
-substrate-bn = { git = "https://github.com/0xPolygonHermez/zisk-patch-bn.git", tag = "patch-0.6.0-zisk-0.14.0" }
-sp1_bls12_381 = { git = "https://github.com/0xPolygonHermez/zisk-patch-bls12-381", tag = "patch-0.8.0-zisk-0.14.0" }
+sha2 = { git = "https://github.com/0xPolygonHermez/zisk-patch-hashes.git", tag = "patch-sha2-0.10.9-zisk-0.15.0" }
+sha3 = { git = "https://github.com/0xPolygonHermez/zisk-patch-hashes.git", tag = "patch-sha3-0.10.8-zisk-0.15.0" }
+k256 = { git = "https://github.com/0xPolygonHermez/zisk-patch-elliptic-curves.git", tag = "patch-k256-0.13.4-zisk-0.15.0" }
+substrate-bn = { git = "https://github.com/0xPolygonHermez/zisk-patch-bn.git", tag = "patch-0.6.0-zisk-0.15.0" }
+sp1_bls12_381 = { git = "https://github.com/0xPolygonHermez/zisk-patch-bls12-381", tag = "patch-0.8.0-zisk-0.15.0" }
 tiny-keccak = { git = "https://github.com/0xPolygonHermez/zisk-patch-tiny-keccak/", branch = "zisk" }
```
ilitteri: They pushed a tag for `tiny-keccak` in 0.15.0 (`patch-2.0.2-zisk-0.15.0`) https://github.com/0xPolygonHermez/zisk-patch-tiny-keccak/tree/patch-2.0.2-zisk-0.15.0

xqft: done!

crates/vm/levm/src/precompiles.rs
```diff
@@ -675,8 +675,22 @@ fn mod_exp(base: Natural, exponent: Natural, modulus: Natural) -> Natural {
...
+        {
+            use ziskos::zisklib::modexp_u64;
+            let (mut base, mut exponent, mut modulus) = (base, exponent, modulus);
+            let base_limbs = base.to_limbs_asc();
+            let exponent_limbs = exponent.to_limbs_asc();
+            let modulus_limbs = modulus.to_limbs_asc();
+            let result_limbs = modexp_u64(&base_limbs, &exponent_limbs, &modulus_limbs);
+            Natural::from_owned_limbs_asc(result_limbs)
+        }
     }
```
Oppen, line 706: Two things:
1. Try not to put these implementations ad-hoc, as it quickly grows with different provers;
2. I think we were using `target_vendor` rather than feature flags for this.

xqft: Yes, at some point we would need to create a module, maybe in `ethrex-crypto`, to refactor the syscalls and patched crates.

Initially I tried using
```
#[cfg(all(target_os = "zkvm", target_vendor = "zisk"))]
```
which I copied from a ZisK patched crate [here](https://github.com/aurora-is-near/aurora-engine/compare/develop...0xPolygonHermez:zisk-patch-modexp:zisk-patch/modexp#diff-97dba09355bcbaabd35f7c084b931f7a3b38f8c5854b3cb32a126af8d77400daR44), but when compiling the guest program it seems to not enable the ZisK path of the code because I had a snippet that didn't compile but the guest compiled anyways. We also lose linting (CI workflows use feature flags).

Considering that currently we are not using `target_vendor` anywhere in the codebase for zkVMs precompiles, we could leave for a future PR the task of:
1. Changing all precompile conditional compilation cases from feature flags to `target_vendor`
2. Have some kind of feedback to make sure we are compiling the right path for each case
3. Change the CI workflows to have linting in those cases

Because I think these are out of scope of this PR

ilitteri: I agree with this but I'd rather prefer to address it all in one single PR later.

PR #5696, opened 2025-12-19
fix(l1): create parent dir during jwtsecret generation

MegaRedHand:
**Motivation**

If the user supplies a directory that doesn't exist for the JWT secret, we just panic.

**Description**

This PR adds a step creating the parent directory for the JWT secret, so this doesn't fail.


rodrigo-o (approved):
LGTM. Copilot's comment, if true, would be a good change for consistency

PR #5697, opened 2025-12-22
perf(l1,l2): optimize pending messages lookups in watcher

Fibonacci747:
**Motivation**

Heavy redundancy/excessive work in filter_verified_messages:

```rust
for rpc_log in logs {
    ...
    let pending_l2_messages =
        get_pending_l2_messages(l1_client, bridge_address, l2_client.chain_id).await?;
    ...
}
```
For each log:
RPC get_pending_l2_messages is called on L1;
the same list of pending hashes for a given chain_id is returned.
This results in:
a linear number of RPC calls on L1 based on the number of logs (very heavy with a large flow);
unnecessary allocations in each call.

**Description**

L1Watcher was calling bridge view methods to fetch pending hashes once per log or transaction, which results in unnecessary RPC load and extra allocations under high throughput.

This change makes the watcher fetch pending L2 message hashes once per batch in filter_verified_messages and pending privileged L1 transaction hashes once per batch in process_privileged_transactions. The logic and ordering semantics are preserved: we still break on the first non-pending L2 message and still short-circuit if a privileged transaction is already known in the local store. Only the number of external calls is reduced, so behavior remains the same while performance improves.


ilitteri:
Hey! Thanks for the PR. The `performance` label wasn't assigned automatically. Now that it was labeled, please update the changelog file for the PR to be merged.

Fibonacci747:
> Hey! Thanks for the PR. The `performance` label wasn't assigned automatically. Now that it was labeled, please update the changelog file for the PR to be merged.

Thanks for review, added changelog

PR #5701, opened 2025-12-22
fix(l1): fix InvalidPayloadAttributes RPC error message

Galoretka:
**Motivation**

Invalid message for InvalidPayloadAttributes
In impl From<RpcErr> for RpcErrorMetadata branch RpcErr::InvalidPayloadAttributes(data) sets
``` rust
message: “Invalid forkchoice state”.to_string(),
```

which contradicts the variant itself (InvalidPayloadAttributes) and the adjacent InvalidForkChoiceState.

**Description**

The InvalidPayloadAttributes RpcErr variant was incorrectly mapped to the "Invalid forkchoice state" message while keeping the -38003 error code. This change aligns the message with the Engine API specification ("Invalid payload attributes") and avoids confusing it with InvalidForkChoiceState errors, without altering error codes or behavior in the surrounding validation logic.


ElFantasma:
Oops, I updated it and it seems it has been fixed by someone in the meantime. I'm closing it.

PR #5706, opened 2025-12-22
fix(l1): use current header instead of parent to get excess blob gas

lakshya-sky:
**Motivation**

* `./hive --sim ethereum/eels/execute-blobs --client ethrex` fails with `execution_testing.rpc.rpc_types.JSONRPCError: JSONRPCError(code=-32603, message=Internal Error: Parent block not found)`.
* GetBlobBaseFee is using parents header which causes to get blob base fee for genesis blocks. According to the specs `excess_blob_gas` is **Running total of blob gas consumed in excess of the target, prior to this block.**. That means we need to use latest block header only.

**Description**
- use latest block header and remove accessing parent header.


crates/networking/rpc/eth/block.rs
```diff
@@ -341,8 +335,6 @@ pub async fn get_all_block_rpc_receipts(
     if header.parent_hash.is_zero() {
         return Ok(receipts);
     }
```
Oppen, line 337: I think this check might also be wrong. Doesn't the genesis block also include its own transactions which might create receipts?

edg-l: i think genesis blocks don't have receipts

MegaRedHand: This. We don't have a way to accept a genesis block with transactions:

https://github.com/lambdaclass/ethrex/blob/1bebff3c049e767fc9974eff551d066ad0935761/crates/common/types/genesis.rs#L697-L703

It'd be good to add a comment here.

crates/networking/rpc/eth/block.rs
```diff
@@ -311,14 +310,9 @@ impl RpcHandler for GetBlobBaseFee {
...
         };
-        let parent_header = match find_parent_header(&header, &context.storage) {
-            Ok(option_header) => option_header,
-            Err(error) => return Err(RpcErr::Internal(error.to_string())),
-        };
-
         let config = context.storage.get_chain_config();
         let blob_base_fee = calculate_base_fee_per_blob_gas(
-            parent_header.excess_blob_gas.unwrap_or_default(),
+            header.excess_blob_gas.unwrap_or_default(),
```
MegaRedHand, line 315: For reference, the same code in geth is: https://github.com/ryanschneider/go-ethereum/blob/13bde9bb8572898abab200be7bd1616f5650257e/eth/gasprice/feehistory.go#L97-L103

MegaRedHand (approved):
Good catch!

PR #5707, opened 2025-12-22
fix(l1,l2): make GuestProgramState more strict when information is missing

JereSalo:
**Motivation**

- Main goal was to -> Fix problems with [ethrex-replay](https://github.com/lambdaclass/ethrex-replay) when asking for a witness to non-ethrex nodes.

Now, given that it turns out to be a vulnerability for the L2 we will change the approach, it's more important to guarantee safety in the Guest Program than to replay with non-ethrex nodes for now. We will try to achieve both though.

**Description**

Description of Replay with non-ethrex nodes problem:
> When using a non-ethrex node as RPC in ethrex-replay it will happen sometimes that we will request for a node that's not completely necessary for that block's execution, so in this case instead of failing (as we were doing before) we default to an empty value because the node wasn't relevant in the first place.
> In all the cases I've seen so far the storage slot didn't exist, but the non-ethrex witness didn't provide the nodes to check it by traversing nibble by nibble until we found 0x80 corresponding to that path. It just simply didn't give us the node.
> The fix for replay is to default to empty values if we don't have them but this is not correct for production, so we will try to find a good solution that satisfies both replay and the L2, prioritizing correctness in the L2 first.

Before this PR we were doing for accounts and storage "If you don't find them, default to empty values because they are not important". This works for replay but is a vulnerability in production (e.g. for an Ethrex L2) so we are going to be strict and see if we can be loose with replay only.


JereSalo:
Stale. If this hasn't been handled yet we need to think of an appropriate solution to what's described in the PR description, it's not necessarily this one though.

PR #5708, opened 2025-12-22
feat(l1): generate execution witnesses during payload execution

avilagaston9:
**Motivation**

To provide a quicker response to `debug_executionWitness` requests.

**Description**

This PR generates execution witnesses during payload execution and stores them for a period of time (128 blocks).

- Adds the node flag `generate-witness`.
- After successfully executing a payload received via the `newPayload` message, if the `generate-witness` flag is enabled, the node will generate and store the `ExecutionWitness` for that block.
- Upon receiving a `debug_executionWitness` request, the node first checks if it has the witness in storage and only generates it as a fallback.

**Measurements**

[execution_witness_measurements.md](https://github.com/lambdaclass/ethrex/blob/0343b46db8e4e18ccf64e8c919420441fd2ab1ca/docs/l2/fundamentals/execution_witness_measurements.md) contains a comparative analysis of `debug_executionWitness` latency for pre-generated vs on-demand execution witnesses.

**Checklist**

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -3,14 +3,21 @@ description: Builds the Ethrex docker image
 
 inputs:
   username:
-    description: "Username for docker registry login"
+    description: "Username for ghcr.io registry login"
     required: false
   password:
-    description: "Password or token for docker registry login"
+    description: "Password or token for ghcr.io registry login"
+    required: false
+  dockerhub_username:
+    description: "Username for DockerHub login (for rate limiting)"
+    required: false
+  dockerhub_password:
+    description: "Password or token for DockerHub login"
     required: false
   registry:
     description: "Docker registry to push the image to (if pushing)"
     required: false
+    default: ghcr.io
   tags:
     description: "Comma-separated list of tags to apply to the built image"
     required: false
@@ -31,6 +38,10 @@ inputs:
     description: "Comma-separated list of platforms to build for"
     required: false
     default: "linux/amd64"
+  variant:
+    description: "Build variant for cache separation (l1 or l2)"
+    required: false
+    default: "l1"
 
 outputs:
   artifact_path:
@@ -45,11 +56,43 @@ runs:
       run: |
         echo "artifact_path=/tmp/${{ inputs.artifact_path }}" >> $GITHUB_OUTPUT
 
-    - name: Login to Docker registry
+    - id: cache-config
+      shell: bash
+      run: |
+        # Sanitize platform for use in image tags (replace / with -)
+        PLATFORM_SAFE=$(echo "${{ inputs.platforms }}" | tr '/' '-')
+        echo "platform_safe=$PLATFORM_SAFE" >> $GITHUB_OUTPUT
+
+        # Determine cache scope based on GitHub context
+        # - PRs: use pr-<number> for isolation between PRs
+        # - Main branch: use 'main'
+        # - Tags/releases: use 'main' (read from main's cache)
+        if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+          CACHE_SCOPE="pr-${{ github.event.pull_request.number }}"
+        elif [[ "${{ github.ref }}" == "refs/heads/main" ]]; then
+          CACHE_SCOPE="main"
+        else
+          # For tags and other refs, use main's cache
+          CACHE_SCOPE="main"
+        fi
+        echo "cache_scope=$CACHE_SCOPE" >> $GITHUB_OUTPUT
+
+        # Build cache tag components
+        VARIANT="${{ inputs.variant }}"
+        echo "Cache scope: $CACHE_SCOPE, variant: $VARIANT, platform: $PLATFORM_SAFE"
+
+    - name: Login to DockerHub (for rate limiting)
+      if: inputs.dockerhub_username != '' && inputs.dockerhub_password != ''
+      uses: docker/login-action@v3
+      with:
+        username: ${{ inputs.dockerhub_username }}
+        password: ${{ inputs.dockerhub_password }}
+
+    - name: Login to ghcr.io (for cache)
       if: inputs.username != '' && inputs.password != ''
       uses: docker/login-action@v3
       with:
-        registry: ${{ inputs.registry }}
+        registry: ghcr.io
         username: ${{ inputs.username }}
         password: ${{ inputs.password }}
 
@@ -65,8 +108,10 @@ runs:
         push: ${{ inputs.push }}
         tags: ${{ inputs.tags }}
         outputs: ${{ inputs.push == 'false' && format('type=docker,dest={0}', steps.vars.outputs.artifact_path) || '' }}
-        cache-from: type=gha
-        cache-to: type=gha,mode=max
+        cache-from: |
+          type=registry,ref=${{ inputs.registry }}/${{ github.repository }}:cache-${{ steps.cache-config.outputs.cache_scope }}-${{ inputs.variant }}-${{ steps.cache-config.outputs.platform_safe }}
+          type=registry,ref=${{ inputs.registry }}/${{ github.repository }}:cache-main-${{ inputs.variant }}-${{ steps.cache-config.outputs.platform_safe }}
+        cache-to: type=registry,ref=${{ inputs.registry }}/${{ github.repository }}:cache-${{ steps.cache-config.outputs.cache_scope }}-${{ inputs.variant }}-${{ steps.cache-config.outputs.platform_safe }},mode=max
         build-args: ${{ inputs.build_args }}
         platforms: ${{ inputs.platforms }}
 
--- a/.github/actions/snapsync-run/action.yml
+++ b/.github/actions/snapsync-run/action.yml
@@ -15,6 +15,14 @@ inputs:
     description: Ethrex Docker image tag.
     required: false
     default: main
+  build_local:
+    description: Build ethrex image locally instead of using a pre-built image.
+    required: false
+    default: "false"
+  build_profile:
+    description: Cargo profile to use when building locally (e.g., release, release-with-debug-assertions).
+    required: false
+    default: release
   cl_type:
     description: Consensus layer type (lighthouse, prysm, etc).
     required: false
@@ -34,12 +42,25 @@ runs:
       shell: bash
       run: docker image rm -f ${{ inputs.ethrex_image }}:${{ inputs.ethrex_tag }} || true
 
+    - name: Build local ethrex image
+      if: inputs.build_local == 'true'
+      shell: bash
+      env:
+        BUILD_PROFILE: ${{ inputs.build_profile }}
+        IMAGE_TAG: ${{ inputs.ethrex_tag }}
+      run: |
+        echo "Building ethrex with profile: ${BUILD_PROFILE}"
+        docker build \
+          --build-arg PROFILE="${BUILD_PROFILE}" \
+          -t ethrex-local:${IMAGE_TAG} \
+          -f Dockerfile .
+
     - name: Generate Kurtosis args
       shell: bash
       env:
         NETWORK_NAME: ${{ inputs.network }}
         TIMEOUT: ${{ inputs.timeout }}
-        ETHREX_IMAGE: ${{ inputs.ethrex_image }}
+        ETHREX_IMAGE: ${{ inputs.build_local == 'true' && 'ethrex-local' || inputs.ethrex_image }}
         ETHREX_TAG: ${{ inputs.ethrex_tag }}
         CL_TYPE: ${{ inputs.cl_type }}
         CL_IMAGE: ${{ inputs.cl_image }}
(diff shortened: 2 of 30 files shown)


crates/networking/rpc/engine/payload.rs
```diff
@@ -617,7 +617,37 @@ async fn handle_new_payload_v1_v2(
...
     // All checks passed, execute payload
-    let payload_status = try_execute_payload(block, &context, latest_valid_hash).await?;
+    let payload_status = try_execute_payload(block.clone(), &context, latest_valid_hash).await?;
+
+    let block_hash = block.hash();
+    let block_number = block.header.number;
+
+    // Generate and store witness if required
+    if context.generate_witness {
+        let Ok(witness) = context
```
tomip01: a nit:

```suggestion
    // Generate and store witness if required
    if context.generate_witness {
        let block_hash = block.hash();
        let block_number = block.header.number;
 
        let Ok(witness) = context
```

crates/networking/rpc/engine/payload.rs
```diff
@@ -617,7 +617,37 @@ async fn handle_new_payload_v1_v2(
...
+            .await
+        else {
+            warn!(
+                %block_hash,
+                %block_number,
+                "Failed to generate witness for block. Skipping witness generation."
+            );
+            return Ok(payload_status);
+        };
+        _ = context
```
tomip01: I didn't know you could do `_` instead of `let _`

tomip01 (approved):

crates/networking/rpc/engine/payload.rs
avilagaston9: Done [cd2d029](https://github.com/lambdaclass/ethrex/pull/5708/commits/cd2d029259127bdf627300a12727f8641a925e3b)!

crates/networking/rpc/engine/payload.rs
avilagaston9: Now that you mentioned it, I prefer the `let _` version. [cd2d029](https://github.com/lambdaclass/ethrex/pull/5708/commits/cd2d029259127bdf627300a12727f8641a925e3b)

crates/networking/rpc/debug/execution_witness.rs
```diff
@@ -234,6 +234,22 @@ impl RpcHandler for ExecutionWitnessRequest {
             blocks.push(block);
         }
 
+        if blocks.len() == 1 {
```
iovoid, line 237: Why only if len=1?

crates/storage/store.rs
```diff
@@ -1701,6 +1705,100 @@ impl Store {
...
+
+    fn cleanup_old_witnesses(&self, latest_block_number: u64) -> Result<(), StoreError> {
+        // If we have less than 128 blocks, no cleanup needed
+        if latest_block_number <= MAX_WITNESSES {
+            return Ok(());
+        }
+
+        let threshold = latest_block_number - MAX_WITNESSES;
+
+        self.get_oldest_witness_number()?
```
iovoid: This could be an if-let

crates/blockchain/blockchain.rs
```diff
@@ -1104,7 +1370,28 @@ impl Blockchain {
...
 
-        let result = self.store_block(block, account_updates_list, res);
+        let result = if self.options.generate_witness && self.is_synced() {
+            let block_hash = block.hash();
+            let Some(logger) = logger else {
+                return Err(ChainError::Custom(
+                    "Logger not found for witness generation".to_string(),
+                ));
+            };
+            let account_updates = vm.get_state_transitions()?;
```
iovoid: Block execution should already consume the state transition, wouldn't this return empty?

crates/storage/store.rs
avilagaston9: done [263e0e5](https://github.com/lambdaclass/ethrex/pull/5708/commits/263e0e5e701815115e8031a85041e1b8052c69b7)!

crates/blockchain/blockchain.rs
avilagaston9: Yes, that explains the changes in `crates/vm/levm/src/db/gen_db.rs`

crates/networking/rpc/debug/execution_witness.rs
avilagaston9: I limit this PR to caching witnesses for only one block. In the future, we may want to merge all cached witnesses when a range is requested

crates/blockchain/blockchain.rs
```diff
@@ -1091,11 +1336,32 @@ impl Blockchain {
             return Err(ChainError::ParentNotFound);
         };
 
-        let vm_db = StoreVmDatabase::new(self.storage.clone(), parent_header.clone())?;
-        let vm = self.new_evm(vm_db)?;
+        let (mut vm, logger) = if self.options.generate_witness && self.is_synced() {
```
JereSalo: I'd add a quick comment here saying what the purpose of this code is. Like just to execute with the database logger in order to generate the witness in one execution instead of executing more than once for this goal.

crates/storage/api/tables.rs
```diff
@@ -87,7 +87,22 @@ pub const STORAGE_FLATKEYVALUE: &str = "storage_flatkeyvalue";
...
+///     // composite_key.extend_from_slice(&block_number.to_be_bytes());
+///     // composite_key.extend_from_slice(block_hash.as_bytes());
+///    ```
+/// - [`Vec<u8>`] = `serde_json::to_vec(&witness)`
+pub const EXECUTION_WITNESSES: &str = "execution_witnesses";
+
+/// Oldest witness block column family: [`Vec<u8>`] => [`Vec<u8>`]
+/// - [`Vec<u8>`] = `b"oldest_witness_block"`
+/// - [`Vec<u8>`] = `oldest_block_number.to_le_bytes()`
+pub const OLDEST_WITNESS_BLOCK: &str = "oldest_witness_block";
```
JereSalo: The name is okay but it'd be good to specify that it's a number, like `OLDEST_WITNESS_BLOCK_NUMBER`

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -27,6 +27,7 @@ pub struct GeneralizedDatabase {
     pub store: Arc<dyn Database>,
     pub current_accounts_state: CacheDB,
     pub initial_accounts_state: CacheDB,
+    pub previous_tx_accounts_state: CacheDB,
```
JereSalo: We should add comments to each because it may be confusing

crates/blockchain/blockchain.rs
```diff
@@ -1104,7 +1370,28 @@ impl Blockchain {
...
         );
 
-        let result = self.store_block(block, account_updates_list, res);
+        let result = if self.options.generate_witness && self.is_synced() {
+            let block_hash = block.hash();
+            let Some(logger) = logger else {
+                return Err(ChainError::Custom(
+                    "Logger not found for witness generation".to_string(),
+                ));
+            };
```
tomip01: I think we could do a let chain
```
let result =  if let Some(logger) = logger && self.options.generate_witness && self.is_synced() {
...
}
```

crates/blockchain/blockchain.rs
```diff
@@ -1104,7 +1370,28 @@ impl Blockchain {
...
+                &block,
+                parent_header,
+                &logger,
+                res,
+            )?;
+            self.storage
+                .store_witness(block_hash, block_number, witness)
+                .map_err(|e| e.into())
+        } else {
+            // generate_witness_from_account_updates already stores the block
```
tomip01: Maybe we could add `store_block` to the function name `generate_witness_from_account_updates`

crates/vm/levm/src/db/gen_db.rs
tomip01, line 71: We can change the comment here to be:
```
/// If it's the first time it's loaded store it in `initial_accounts_state` and in `previous_tx_accounts_state`. Also cache it in `current_accounts_state` for making changes to it
```

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -119,18 +124,23 @@ impl GeneralizedDatabase {
         key: H256,
     ) -> Result<U256, InternalError> {
```
tomip01, line 120: Also here we can update the comment

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -257,13 +266,15 @@ impl GeneralizedDatabase {
                 // Skip processing account that we know wasn't mutably accessed during execution
                 continue;
             }
-            // [LIE] In case the account is not in immutable_cache (rare) we search for it in the actual database.
+
             let initial_state_account =
-                self.initial_accounts_state.get(address).ok_or_else(|| {
```
tomip01, line 262: why is this change for?

crates/storage/api/tables.rs
avilagaston9: Done [14173ad](https://github.com/lambdaclass/ethrex/pull/5708/commits/14173adf8db430a5951896e4b4d169befbf94fde)!

crates/vm/levm/src/db/gen_db.rs
avilagaston9: During `execute_block_pipeline` this function is called after executing each tx, updating `initial_accounts_state` with `current_accounts_state` later in this function. This causes calls to `get_state_transitions` after block execution to return an empty array of `AccountUpdates`.

With this change, we modify only `previous_tx_accounts_state`, leaving `initial_accounts_state` unchanged (i.e., containing the state before the whole block execution).

I will add comments to the field definitions for clarity.

crates/blockchain/blockchain.rs
avilagaston9: Done [aa754b7](https://github.com/lambdaclass/ethrex/pull/5708/commits/aa754b78e6e5fa0dac2c558c44e29761f9e7c219)!

crates/vm/levm/src/db/gen_db.rs
avilagaston9: Done [aa754b7](https://github.com/lambdaclass/ethrex/pull/5708/commits/aa754b78e6e5fa0dac2c558c44e29761f9e7c219)!

crates/blockchain/blockchain.rs
avilagaston9: Done [b6f7f9b](https://github.com/lambdaclass/ethrex/pull/5708/commits/b6f7f9b7796ef8e7c0ddcf989d6d6a66245774f7)

crates/vm/levm/src/db/gen_db.rs
avilagaston9: Done [2ed6d66](https://github.com/lambdaclass/ethrex/pull/5708/commits/2ed6d66105fbfdc0b54bd9d86928cc83511ea663)!

crates/vm/levm/src/db/gen_db.rs
avilagaston9: Done [aa754b7](https://github.com/lambdaclass/ethrex/pull/5708/commits/aa754b78e6e5fa0dac2c558c44e29761f9e7c219)!

crates/blockchain/blockchain.rs
avilagaston9: Done [187f934](https://github.com/lambdaclass/ethrex/pull/5708/commits/187f934e1960718f23956d6b8cf4db77c07f7bbf)!

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -25,8 +25,11 @@ pub type CacheDB = FxHashMap<Address, LevmAccount>;
 #[derive(Clone)]
 pub struct GeneralizedDatabase {
     pub store: Arc<dyn Database>,
+    // Holds the current state of accounts being modified during execution
     pub current_accounts_state: CacheDB,
+    // Holds the initial state of accounts before the block execution started, used for calculating state transitions
     pub initial_accounts_state: CacheDB,
```
JereSalo: Maybe we should add that it could be either before a block or before a batch of blocks. This is because if you have a batch of 100 blocks the `initial_accounts_state` is going to store the initial state before those 100 blocks. Current explanation may be leaving that case out

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -25,8 +25,11 @@ pub type CacheDB = FxHashMap<Address, LevmAccount>;
...
+    // Holds the state of accounts before the current transaction started, used for calculating state transitions per transaction (e.g. pipeline)
```
JereSalo: I'd be more specific in the parenthesis. Saying something like `(used in pipeline execution)` would be best. Just for people that are new to this.

avilagaston9: I changed the PR. We now accumulate `AccountUpdates` inside `handle_merkleization()` only when `generate-witness` is enabled. If it is not enabled, this PR has no impact.

JereSalo:
I still dislike the `generate-witness` name for the CLI flag. 
Suggestions:
- `precompute-witnesses`
- `cache-witnesses`
- `please-precompute-and-cache-the-execution-witnesses-so-debug_executionwitness-is-super-fast`

I'd honestly go with one of the first two but it's up to you.

crates/vm/levm/src/db/gen_db.rs
JereSalo: That's good news. I'm glad you found another way

avilagaston9:
@JereSalo changed to `precompute-witnesses` [here](https://github.com/lambdaclass/ethrex/pull/5708/commits/127494f3d015422e2c36f3f28723e0f7a6751270)!

crates/storage/store.rs
iovoid: I think the change got undone

crates/storage/api/tables.rs
```diff
@@ -87,7 +87,22 @@ pub const STORAGE_FLATKEYVALUE: &str = "storage_flatkeyvalue";
...
+///     // composite_key.extend_from_slice(&block_number.to_be_bytes());
+///     // composite_key.extend_from_slice(block_hash.as_bytes());
+///    ```
+/// - [`Vec<u8>`] = `serde_json::to_vec(&witness)`
+pub const EXECUTION_WITNESSES: &str = "execution_witnesses";
+
+/// Oldest witness block number column family: [`Vec<u8>`] => [`Vec<u8>`]
+/// - [`Vec<u8>`] = `b"oldest_witness_block_number"`
+/// - [`Vec<u8>`] = `oldest_block_number.to_le_bytes()`
+pub const OLDEST_WITNESS_BLOCK_NUMBER: &str = "oldest_witness_block_number";
```
iovoid: We should reuse another cf (like MISC_VALUES) instead of having a cf that holds a single key.

crates/storage/store.rs
avilagaston9: Mmm, I think you are watching an outdated version

crates/storage/api/tables.rs
avilagaston9: Done [here](https://github.com/lambdaclass/ethrex/pull/5708/commits/77c5354925d4b74f02577054ed4bf72e5e22e8a6)!

PR #5713, opened 2025-12-23
chore(l1): fast validate

fedacking:
**Motivation**

We want to validate the tree quicker and consume less memory.

**Description**

- Made validate state and storage root not recompute the entire tree, instead just traverse the trie to found the missing nodes.

### DIFF
--- a/crates/common/trie/trie.rs
+++ b/crates/common/trie/trie.rs
@@ -529,6 +529,12 @@ impl Trie {
         trie
     }
 
+    /// Validates that the Trie isn't missing any nodes expected in the branches
+    ///
+    /// This is used internally with debug assertions to check the status of the trie
+    /// after syncing operations.
+    /// Note: this operation validates the hashes because the iterator uses
+    /// get_node_checked. We shouldn't downgrade that to the unchecked version
     pub fn validate(self) -> Result<(), TrieError> {
         let mut expected_count = if self.root.is_valid() { 1 } else { 0 };
         for (_, node) in self.into_iter() {


crates/common/trie/trie.rs
```diff
@@ -528,6 +528,32 @@ impl Trie {
...
+                        .iter()
+                        .filter(|child| child.is_valid())
+                        .count();
+                }
+                Node::Extension(_) => {
+                    expected_count += 1;
+                }
+                Node::Leaf(_) => {}
+            }
+        }
```
Oppen, line 555: Are we checking the hashes match at some point? Does the iterator still do that?

fedacking: The iterator uses `get_node_checked`, which will stop the iteration, which will make `expected_count` return a value indicating that the trie is not valid.

crates/common/trie/trie.rs
```diff
@@ -528,6 +528,32 @@ impl Trie {
         trie.root = root;
         trie
     }
+
+    pub fn validate(self) -> Result<(), TrieError> {
+        let mut expected_count = if self.root.is_valid() { 1 } else { 0 };
+        for (_, node) in self.into_iter() {
```
Oppen, line 540: Let's add a comment explaining why this works. Something like "`TrieIterator` uses `get_node_checked` which ensures the actual and expected node hash match" would suffice. A matching comment in `TrieIterator` would be appreciated as well, explaining that we shouldn't downgrade that to the unchecked version.

fedacking: Done in [f93276b](https://github.com/lambdaclass/ethrex/pull/5713/commits/f93276bf1246a7bba5b58f4defd178aa0e72ae80).

crates/common/trie/trie.rs
```diff
@@ -528,6 +528,38 @@ impl Trie {
...
         trie
     }
+
+    /// Validates that the Trie isn't missing any nodes expected in the branches
+    ///
+    /// This is used internally with debug assertions to check the status of the trie
+    /// after syncing operations.
+    /// Note: this operation validates the hashes because the iterator uses
+    /// get_node_checked. We shouldn't downgrade that to the unchecked version
+    pub fn validate(self) -> Result<(), TrieError> {
```
rodrigo-o: Nit: we need self instead of &self for something right? Was thinking about consuming vs not consuming it.

crates/common/trie/trie.rs
```diff
@@ -528,6 +528,38 @@ impl Trie {
...
+                Node::Extension(_) => {
+                    expected_count += 1;
+                }
+                Node::Leaf(_) => {}
+            }
+        }
+        if expected_count != 0 {
+            return Err(TrieError::Verify(format!(
+                "Node count mismatch, expected {expected_count} more"
+            )));
```
rodrigo-o, line 559: We lost context here, maybe for a follow-up but we might want better info for this error (like node/path) to understand where to look instead of just the difference? Not sure if right now that info is useful or not, i didn't debugged these cases before, if it is indeed useful we can create a small issue for later.

rodrigo-o (approved):

PR #5725, opened 2025-12-24
perf(l1,l2): dedup execution witness codes and trie nodes

Snezhkko:
This change deduplicates contract bytecodes when building the execution witness by tracking seen code hashes and only pushing each bytecode once. It also removes redundant intermediate Vec allocations when extending the list of trie nodes, extending directly from map values instead. Both adjustments preserve the semantics of the witness while reducing unnecessary allocations, storage lookups, and payload size.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -42,7 +42,7 @@ use mempool::Mempool;
 use payload::PayloadOrTask;
 use rustc_hash::FxHashMap;
 use std::collections::hash_map::Entry;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::sync::{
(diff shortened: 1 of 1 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -727,6 +727,7 @@ impl Blockchain {
 
         let mut blockhash_opcode_references = HashMap::new();
         let mut codes = Vec::new();
+        let mut seen_code_hashes = BTreeSet::new();
```
ElFantasma, line 730: nit: `BTreeSet` maintains sorted order, which isn't needed here — a `HashSet` (or `FxHashSet` since `FxHashMap` is already imported) would give O(1) insert/lookup instead of O(log n). Small difference for typical code counts, but since this is a perf PR it seems worth aligning the data structure choice.

ElFantasma (changes requested):

PR #5727, opened 2025-12-29
docs: fix LEVM FAQ U256 casting example and error naming

radik878:
**Motivation**

Tighten up the LEVM FAQ so it matches the actual levm code and doesn’t reference non‑existent errors or unresolved TODOs.

**Description**

fixed the usize/U256 example in docs/vm/levm/faq.md to use the real InternalError::TypeConversion variant instead of the obsolete ConversionError, corrects the typo in InternalError in the Errors section, and replaces the TODO about U256 addresses with a link to the existing State diffs doc that explains how U256 values are used.

### DIFF
--- a/docs/vm/levm/faq.md
+++ b/docs/vm/levm/faq.md
@@ -1,8 +1,7 @@
 # FAQ
 ## `usize` and `U256`
 In Rust, **accessing an index on a specific data structure requires a `usize` type variable.** This can be seen in methods like `get` and `get_mut`.
-
-<!-- TODO: Link in the documentation where the `U256` addresses are described -->
+For more details on how `U256` is used to represent storage keys and other 256-bit values, see the [State diffs](../../l2/fundamentals/state_diffs.md) documentation.
 On the other hand, the EVM specification requires all addresses to be in `U256`. Therefore, every opcode treats its arguments as `U256` values.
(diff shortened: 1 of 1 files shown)


docs/vm/levm/faq.md
```diff
@@ -1,8 +1,7 @@
 # FAQ
 ## `usize` and `U256`
 In Rust, **accessing an index on a specific data structure requires a `usize` type variable.** This can be seen in methods like `get` and `get_mut`.
-
-<!-- TODO: Link in the documentation where the `U256` addresses are described -->
+For more details on how `U256` is used to represent storage keys and other 256-bit values, see the [State diffs](../../l2/fundamentals/state_diffs.md) documentation.
```
ElFantasma, line 4: Two issues with this link:

1. **Broken relative path** — from `docs/vm/levm/faq.md`, `../../l2/fundamentals/state_diffs.md` resolves to `docs/vm/l2/fundamentals/state_diffs.md` which doesn't exist. You need three levels up: `../../../l2/fundamentals/state_diffs.md`.

2. **Questionable link target** — the state_diffs doc is about L2 data availability (and is even marked as deprecated). The original TODO was asking for a link to where `U256` addresses are described in the LEVM/EVM context. The state diffs doc doesn't really cover that. It might be better to just drop the TODO and leave the paragraph flowing naturally into the next sentence (which already explains that the EVM spec requires `U256`).

ElFantasma (changes requested):

PR #5736, opened 2026-01-04
chore(l2): remove unused ELASTICITY MULTIPLIER import

Olexandr88:
Removes an unused ELASTICITY MULTIPLIER import flagged by clippy

### DIFF
--- a/crates/networking/p2p/discv5/server.rs
+++ b/crates/networking/p2p/discv5/server.rs
@@ -418,7 +418,7 @@ impl DiscoveryServer {
             }
         }
         Message::FindNode(FindNodeMessage {
-            req_id: Bytes::from(rng.r#gen::<u64>().to_be_bytes().to_vec()),
+            req_id: generate_req_id(),
             distances,
         })
     }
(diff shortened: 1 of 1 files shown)


ElFantasma (commented):
**This PR has merge conflicts and is stale.** The target file `crates/l2/prover/src/guest_program/src/execution.rs` was moved to `crates/guest-program/` on `main`, so this diff no longer applies cleanly. The original change (removing the unused `#[cfg(not(feature = "l2"))]` import of `ELASTICITY_MULTIPLIER`) was correct in intent — I confirmed that `ELASTICITY_MULTIPLIER` is not used in the guest program's L2 execution path. However, this needs a rebase onto current `main` and the fix should target the new file location.

The import reorder (`ethrex_l2_common` ↔ `ethrex_rlp`) is cosmetic and fine either way.

Olexandr88:
After rebasing onto current main and checking the new file location, `ELASTICITY_MULTIPLIER` is now used in the guest program, so the original cleanup is no longer applicable

PR #5738, opened 2026-01-05
perf(l1): improve ecrecover by removing needless alloc and conversion

edg-l:
**Motivation**

Removed a vec alloc and a U256 conversion that could be avoided

Tried to run the benchmarks but unrelated things improve and regress, so i think its a bit flaky, this PR only changed the precompile function so its hard to imagine how it affects that

### DIFF
--- a/crates/vm/levm/src/precompiles.rs
+++ b/crates/vm/levm/src/precompiles.rs
@@ -376,6 +376,12 @@ pub(crate) fn fill_with_zeros(calldata: &Bytes, target_len: usize) -> Bytes {
 }
 
 #[expect(clippy::arithmetic_side_effects, clippy::indexing_slicing)]
+#[cfg(all(
+    not(feature = "sp1"),
+    not(feature = "risc0"),
+    not(feature = "zisk"),
+    feature = "secp256k1"
+))]
 #[inline(always)]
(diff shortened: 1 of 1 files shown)


crates/vm/levm/src/precompiles.rs
```diff
@@ -396,19 +396,35 @@ pub fn ecrecover(calldata: &Bytes, gas_remaining: &mut u64, _fork: Fork) -> Resu
...
+    // Total input size = 128
+    let mut raw_hash = [0u8; WORD];
+    let mut raw_v = [0u8; WORD];
+    let mut raw_sig = [0u8; SIG_LEN];
 
-    // len(raw_hash) == 32, len(raw_v) == 32, len(raw_sig) == 64
-    let (raw_hash, tail) = input.split_at(WORD);
-    let (raw_v, raw_sig) = tail.split_at(WORD);
+    #[expect(clippy::arithmetic_side_effects, clippy::indexing_slicing)]
+    let copy_segment = |dst: &mut [u8], start: usize| {
```
iovoid: This could be a regular function that takes &calldata as a parameter.

edg-l: done

azteca1998 (approved):

PR #5742, opened 2026-01-05
fix(l2): correctly accumulate operations counts in in-memory rollup store

Fibonacci747:
**Motivation**

The in-memory rollup store is used in L2 components that expose metrics about transactions, privileged transactions and messages. These metrics depend on the global counters exposed by get_operations_count. However, the in-memory implementation of update_operations_count was not updating the underlying state at all, so the counters always stayed at zero. This silently broke L2 metrics when the in-memory backend is used and also made its behavior diverge from the SQL backend.

**Description**

This change fixes the in-memory implementation of StoreEngineRollup::update_operations_count so that it mutates the operations_counts array inside StoreInner instead of working on a local copy. As a result, subsequent calls to get_operations_count now return the accumulated values, matching the semantics of the SQL backend. A small async unit test was added to in_memory.rs to verify that multiple updates correctly accumulate and are reflected in the result of get_operations_count.

### DIFF
--- a/crates/l2/storage/src/store_db/in_memory.rs
+++ b/crates/l2/storage/src/store_db/in_memory.rs
@@ -197,10 +197,10 @@ impl StoreEngineRollup for Store {
         privileged_tx_inc: u64,
         messages_inc: u64,
     ) -> Result<(), RollupStoreError> {
-        let mut values = self.inner()?.operations_counts;
-        values[0] += transaction_inc;
-        values[1] += privileged_tx_inc;
-        values[2] += messages_inc;
+        let mut inner = self.inner()?;
+        inner.operations_counts[0] += transaction_inc;
+        inner.operations_counts[1] += privileged_tx_inc;
+        inner.operations_counts[2] += messages_inc;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


crates/l2/storage/src/store_db/in_memory.rs
```diff
@@ -197,10 +197,10 @@ impl StoreEngineRollup for Store {
         privileged_tx_inc: u64,
         messages_inc: u64,
     ) -> Result<(), RollupStoreError> {
-        let mut values = self.inner()?.operations_counts;
-        values[0] += transaction_inc;
-        values[1] += privileged_tx_inc;
-        values[2] += messages_inc;
+        let mut inner = self.inner()?;
```
ElFantasma, line 200: Good catch. `self.inner()?.operations_counts` copies the `[u64; 3]` out of the `MutexGuard` (arrays are `Copy`), so the old code was a silent no-op. Holding the guard as `let mut inner = self.inner()?` and mutating through it is the correct fix and matches the rest of this file.

ElFantasma (approved):
Was submitted twice by another contributor on #6097. We should close one of these.

crates/l2/storage/src/store_db/in_memory.rs
```diff
@@ -481,3 +481,32 @@ impl Debug for Store {
         f.debug_struct("In Memory L2 Store").finish()
     }
 }
+
+#[cfg(test)]
```
ElFantasma, line 485: Test should be moved to tests directory. We changed recently our approach to have our tests separated from actual code

ElFantasma (changes requested):
We should move the test to its own file

PR #5746, opened 2026-01-06
perf(l1,l2): enable optional use of frame pointers

Oppen:
**Motivation**

Frame pointers are used by sampling profilers (notably the eBPF
profiler used by Pyroscope) to reconstruct the call stack of a sample.

**Description**

Pass the `-Cforce-frame-pointers=yes` flag via `RUSTFLAGS` when calling
`make` with `FRAME_POINTERS=1`.

For `tooling/sync`, the default is `1`, so you should use `FRAME_POINTERS=0`
if you don't want them. That's because it's mostly an aid to our own
development and not end-user facing.

While it is a typical compiler optimization to omit it and use RBP (or
the current architecture's equivalent) as an extra general purpose
register, the gains are often tiny (<5% on CPU bound processes) and
hinders our ability to profile and get more impactful optimizations out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### 2026-01-21
 
+- Use fastbloom as the bloom filter [#5968](https://github.com/lambdaclass/ethrex/pull/5968)
 - Improve snap sync logging with table format and visual progress bars [#5977](https://github.com/lambdaclass/ethrex/pull/5977)
 
 ### 2026-01-20
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4107,10 +4107,10 @@ dependencies = [
  "ethrex-crypto",
  "ethrex-rlp",
  "ethrex-trie",
+ "fastbloom",
  "hex",
  "hex-literal 0.4.1",
  "lru 0.16.2",
- "qfilter",
  "rayon",
  "rocksdb",
  "rustc-hash 2.1.1",
@@ -4159,6 +4159,7 @@ dependencies = [
  "ethrex-levm",
  "ethrex-p2p",
  "ethrex-rlp",
+ "ethrex-rpc",
  "ethrex-sdk",
  "ethrex-storage",
  "ethrex-storage-rollup",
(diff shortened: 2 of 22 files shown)


tooling/sync/Makefile
```diff
@@ -7,6 +7,18 @@ multisync-up multisync-down multisync-clean multisync-logs multisync-status \
...
 ETHREX_DIR ?= "../.."
+
+# Frame pointers for profiling (default on for sync/profiling targets)
+FRAME_POINTERS ?= 1
+
+ifeq ($(FRAME_POINTERS),1)
+# Read existing rustflags from .cargo/config.toml and append frame pointers
+# (setting RUSTFLAGS overrides config.toml, so we need to include its flags)
+HOST_TARGET := $(shell rustc -vV | sed -n 's/host: //p')
+CARGO_RUSTFLAGS := $(shell sed -n '/\[target.$(HOST_TARGET)\]/,/^\[/p' $(ETHREX_DIR)/.cargo/config.toml 2>/dev/null | grep '"-C' | sed 's/^[[:space:]]*"//; s/",\{0,1\}$$//' | tr '\n' ' ')
```
tomip01: Just a question, this only grabs "-C*" flags. Can be another types of flags? I see that in `.cargo/config.toml` we don't have any, but could it be added some in the future?

pablodeymo (approved):

tooling/sync/Makefile
Oppen: The regex does look overspecialized. I'll try to make it more general. Maybe there's a better way to build flags from the config file.

Oppen: OK, I think I found a better approach, but I'm not 100% sure it will work.

According to [cargo docs](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags) (added emphasis):
> There are four mutually exclusive sources of extra flags. They are checked in order, with the first one being used:
> 
> CARGO_ENCODED_RUSTFLAGS environment variable.
> RUSTFLAGS environment variable.
> **All matching target.<triple>.rustflags and target.<cfg>.rustflags config entries joined together.**
> build.rustflags config value.

That "all matching" seems to imply I can use an additional configuration file with [option --config](https://doc.rust-lang.org/cargo/commands/cargo-build.html#common-options):
> [--config KEY=VALUE or PATH](https://doc.rust-lang.org/cargo/commands/cargo-build.html#option-cargo-build---config)
> Overrides a Cargo configuration value. The argument should be in TOML syntax of KEY=VALUE, or provided as a path to an extra configuration file. This flag may be specified multiple times. See the [command-line overrides section](https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides) for more information.

At the same time, the doc mentions overriding for the flag, so it may just ignore the original one if I do that. I'll see how I can make sure that works.

Oppen: OK, tried that approach and cargo decides having two `target` sections matching causes conflicts, so that won't work. I'll review the regex later.

Oppen: I tried again and now it seems to work. I'm confused.

PR #5747, opened 2026-01-06
fix(l1): add context to trie validation count mismatch error

### LINKED ISSUE #5745: Add more context to count mismatch in trie validation function
> We lost context [here](https://github.com/lambdaclass/ethrex/pull/5713/changes/BASE..f93276bf1246a7bba5b58f4defd178aa0e72ae80#diff-8398e9197a9beb1ece5b777c9c6f9c5662438c649cd7377341724c64eb41a67cR556-R560), maybe for a follow-up but we might want better info for this error (like node/path) to understand where to look instead of just the difference? Not sure if right now that info is useful or not, i didn't debugged these cases before, if it is indeed useful we can create a small issue for later.
> 
> _Originally posted by @rodrigo-o in https://github.com/lambdaclass/ethrex/pull/5713#discussion_r2664632949_

RaveenaBhasin:
**Motivation** 
Trie validation errors currently only report the count mismatch without contextual information about where in the trie structure the issue occurred. This makes debugging sync failures difficult, as there is relevant starting point to investigate the inconsistency (ref: issue #5745, PR #5713 comment). 

**Description**
Improved the NodeCountMismatch error with diagnostic context including
- Path to the last successfully validated node
 - Node type (Branch/Extension/Leaf)
- Node hash for db cross-referencing
- Total nodes traversed before failure

Closes #5745


azteca1998 (approved):

crates/common/trie/trie.rs
```diff
@@ -533,29 +535,64 @@ impl Trie {
...
+
+        for (path, node) in self.into_iter() {
+            // Capture context before processing this node
+            last_valid_path = Some(path.clone());
+            last_node_type = Some(match &node {
+                Node::Branch(_) => NodeType::Branch,
+                Node::Extension(_) => NodeType::Extension,
+                Node::Leaf(_) => NodeType::Leaf,
+            });
+            last_node_hash = Some(node.compute_hash().finalize());
```
ElFantasma: `node.compute_hash().finalize()` does a full recursive hash computation (including `memoize_hashes` + RLP encoding) for **every** node during iteration. This is only needed for the diagnostic when validation fails — the happy path pays the cost for nothing.

Consider deferring the hash computation to the error case. You could store the last node itself (or its RLP) and only hash it when building the error:

```rust
let mut last_node: Option<Node> = None;
// ...
for (path, node) in self.into_iter() {
    last_valid_path = Some(path);
    last_node_type = Some(match &node { /* ... */ });
    last_node = Some(node.clone()); // or restructure to avoid clone
    // ...
}
// In error path:
let last_node_hash = last_node.map(|n| n.compute_hash().finalize());
```

Alternatively, just drop the hash from the diagnostic entirely — the path and node type are probably sufficient for debugging, and the hash can be looked up from the path.

crates/common/trie/trie.rs
```diff
@@ -533,29 +535,64 @@ impl Trie {
...
+
+        // Track context for error reporting
+        let mut last_valid_path: Option<Nibbles> = None;
+        let mut last_node_type: Option<NodeType> = None;
+        let mut last_node_hash: Option<H256> = None;
+        let mut nodes_traversed = 0;
+
+        for (path, node) in self.into_iter() {
+            // Capture context before processing this node
+            last_valid_path = Some(path.clone());
```
ElFantasma: nit: `path` is owned and not used after this line in the loop body — `last_valid_path = Some(path)` works without the `.clone()`.

crates/common/trie/error.rs
```diff
@@ -52,3 +54,57 @@ impl std::fmt::Display for ExtensionNodeErrorData {
...
+            NodeType::Extension => write!(f, "Extension"),
+            NodeType::Leaf => write!(f, "Leaf"),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct NodeCountMismatchData {
+    /// The count mismatch: positive = missing nodes, negative = extra nodes
+    pub expected_count: isize,
```
ElFantasma: nit: The field name `expected_count` is a bit misleading — it's actually the residual counter value (mismatch amount), not a count of expected nodes. The `Display` impl uses it correctly but a reader of the struct definition would expect it to hold the total expected count. Something like `count_difference` or `remaining_count` would be clearer.

ElFantasma (changes requested):

crates/common/trie/trie.rs
RaveenaBhasin: Fixed!

crates/common/trie/error.rs
RaveenaBhasin: Fixed in [commit](https://github.com/lambdaclass/ethrex/pull/5747/changes/13480dbd24b20bb457d6b4b6f768e41c1a0f7b61#diff-8c6b16a78b3f411d75cdc2b5a5b0d60c5ac48501537c6ae9681353864847fb8a)
 Renamed `expected_count` to `count_difference` for better clarity

crates/common/trie/trie.rs
RaveenaBhasin: Fixed in [commit](https://github.com/lambdaclass/ethrex/pull/5747/changes/13480dbd24b20bb457d6b4b6f768e41c1a0f7b61#diff-8c6b16a78b3f411d75cdc2b5a5b0d60c5ac48501537c6ae9681353864847fb8a)
Changed to store the `Node` itself and only compute the hash in the error path when validation fails.

ElFantasma (approved):
LGTM

PR #5761, opened 2026-01-07
perf(levm): inline common opcodes

### LINKED ISSUE #5752: Inlining the most frequently used opcodes
Test inlining hottest opcodes (PUSH, DUP, SWAP, ADD, SUB, MLOAD, MSTORE), instead of having them in the opcode table:

```rust

  loop {
      let opcode = self.current_call_frame.next_opcode();
      // Fast path for common opcodes
      match opcode {
          0x60..=0x7f => { /* inline PUSH1-PUSH32 */ }
          0x80..=0x8f => { /* inline DUP1-DUP16 */ }
          0x90..=0x9f => { /* inline SWAP1-SWAP16 */ }
          0x01 => { /* inline ADD */ }
          0x51 => { /* inline MLOAD */ }
          _ => {
              // Fallback to table for rare opcodes
              self.opcode_table[opcode as usize].call(self);
          }
      }
  }

```

edg-l:
**Motivation**

Perf experiment, inline common opcodes

Closes #5752

There is an increase in speed in benchmarks related to dup, push and swap by a 30%-40%

~some small regresion in jump and push0, inlining them may fix it~

inlining jump removed the regresion seemingly

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2026-01-15
+
+- Reduce state iterated when calculating partial state transitions [#5864](https://github.com/lambdaclass/ethrex/pull/5864)
+
 ### 2026-01-13
 
 - Inline common opcodes [#5761](https://github.com/lambdaclass/ethrex/pull/5761)
--- a/crates/common/crypto/Cargo.toml
+++ b/crates/common/crypto/Cargo.toml
@@ -11,7 +11,7 @@ path = "./lib.rs"
 [dependencies]
 # TODO(#1102): Move to Lambdaworks in the future
 c-kzg = { version = "2.1.1", default-features = false, optional = true }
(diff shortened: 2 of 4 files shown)


crates/vm/levm/src/opcode_handlers/push.rs
```diff
@@ -42,6 +43,7 @@ impl<'a> VM<'a> {
     }
 
     // PUSH0
+    #[inline]
```
iovoid: Why add an inline if it's not in the table?

edg-l: removed

rodrigo-o:
This is the last hour in mainnet:
image

1 is main, 2 is this branch, we see a ~20% improvement

Note: this was just 1 hour. We need to validate this is sustained and not an improvement for just a particular set of blocks

PR #5763, opened 2026-01-07
docs(levm): align LEVM database docs with implementation

Bashmunta:
**Description**

Updated the LEVM database documentation to match the current Database trait and CacheDB type. The text now references the actual method set (get_account_state, get_storage_value, get_block_hash, get_chain_config, get_account_code) and describes CacheDB as a map from Address to LevmAccount instead of a generic Account map.

### DIFF
--- a/docs/vm/levm/database.md
+++ b/docs/vm/levm/database.md
@@ -2,8 +2,8 @@
 
 ## Database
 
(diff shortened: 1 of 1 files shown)


MegaRedHand:
We don't accept PRs fixing typos

PR #5764, opened 2026-01-07
fix(l1): error on really long rlp payloads when decoding

edg-l:
**Motivation**

When decoding an RLP payload, error on big payloads (>1gb) since they are possibly malicious and make no sense in the current protocol

### DIFF
--- a/crates/storage/backend/rocksdb.rs
+++ b/crates/storage/backend/rocksdb.rs
@@ -203,6 +203,16 @@ impl RocksDBBackend {
     }
 }
 
+impl Drop for RocksDBBackend {
+    fn drop(&mut self) {
+        // When the last reference to the db is dropped, stop background threads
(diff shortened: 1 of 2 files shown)


crates/common/rlp/decode.rs
```diff
@@ -8,6 +8,9 @@ use ethereum_types::{
 };
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
+/// Max payload size accepted when decoding.
```
Oppen, line 11: I would add to the comment:
```suggestion
/// Max payload size accepted when decoding.
/// While technically any size is RLP spec-compliant, there are no well-formed messages
/// in our protocols that could carry such big payloads, so they are either bugs or malicious.
```
Just for clarity for the reader, someone might think we're misunderstanding the RLP spec and try to fix it otherwise.

PR #5766, opened 2026-01-07
ci(l1,l2): trigger CI but skip jobs on docs-only changes

ManuelBilbao:
**Motivation**

Our GH setup requires certain "Checks" (GHA jobs titles) to pass to merge a PR. Changes introduced in #5750 make docs-only PRs to not trigger some checks, blocking those PRs.

**Description**

Restore trigger on required checks and run the jobs conditionally on paths changed. According to [GH docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks):
> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success".

### DIFF
--- a/.github/workflows/pr-main_l2_tdx_build.yaml
+++ b/.github/workflows/pr-main_l2_tdx_build.yaml
@@ -40,7 +40,7 @@ jobs:
     # "Test" is a required check, don't change the name
     name: Test
     needs: detect-changes
-    if: ${{ needs.detect-changes.outputs.run_tests == 'true' }}
+    if: ${{ needs.detect-changes.outputs.run_tests == 'true' && always() }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources


.github/workflows/pr-main_l2_tdx_build.yaml
```diff
@@ -22,9 +18,29 @@ env:
...
+        id: finish
+        run: echo "run_tests=${{ steps.filter.outputs.run_tests }}" >> "$GITHUB_OUTPUT"
+      - name: Print result
+        run: echo "run_tests=${{ steps.finish.outputs.run_tests }}"
+
   build_tdx:
     # "Test" is a required check, don't change the name
     name: Test
+    needs: detect-changes
+    if: ${{ needs.detect-changes.outputs.run_tests == 'true' }}
```
tomip01: I think we should add `always()` here since it is a required check (this won't make it run the tests unconditionally)

```suggestion
    if: ${{ needs.detect-changes.outputs.run_tests == 'true' && always() }}
```

JereSalo (commented):
Q: Let's say I run one of these workflows in main using `workflow_dispatch`, won't `detect-changes` return false and therefore skip running the tests?

A: Maybe the fix is doing what we do in `pr-main_l2.yaml` in the rest of the files, which is: `elif [[ "${GITHUB_EVENT_NAME}" != "pull_request" ]]` then `run_tests=true`

PR #5773, opened 2026-01-07
feat(l1): discv5 server

ElFantasma:
**Motivation**
We need to support DiscV5

**Description**

This is a WIP that adds a `discv5::server` and `discv5::peer_table` to support the protocol.
It is an unfinished set of features, but I pretend to merge this big PR to be able to work in sub-features separately. All this functionality is still behind an `experimental-discv5` feature flag, so none of this code is run if the feature is not enabled. Discv4 still is the only supported protocol.

Currently this performs:
- Outgoing handshake (including secret sharing and encryption)
- `FindNode` requests
- handles `Nodes` responses
- Populates the PeerTable with these responses
With this functionality it is enough to start testing and even finding some useful peers, but this does not complete the spec:

I will start issuing in a set of tickets the missing functionality. But mostly:
- It does not handle incoming `FindNode` or handshake requests
- It does not filter peers using the ENR information
- It does not revalidate or clean up discarded peers
- Several other validations need to be done
- It lacks some encoding/decoding tests

### DIFF
--- a/crates/networking/p2p/discv5/messages.rs
+++ b/crates/networking/p2p/discv5/messages.rs
@@ -175,8 +175,8 @@ impl PacketHeader {
 }
 
 pub trait PacketTrait {
+    const TYPE_FLAG: u8;
     fn encode_authdata(&self, buf: &mut dyn BufMut) -> Result<(), PacketCodecError>;
-    fn packet_type_flag(&self) -> u8;
     fn get_encoded_message(&self) -> Vec<u8>;
 
     fn build_header(&self, nonce: &[u8; 12]) -> Result<PacketHeader, PacketCodecError> {
@@ -189,13 +189,13 @@ pub trait PacketTrait {
         let mut static_header: [u8; 23] = [0; 23];
         static_header[0..6].copy_from_slice(PROTOCOL_ID);
         static_header[6..8].copy_from_slice(&PROTOCOL_VERSION.to_be_bytes());
-        static_header[8] = self.packet_type_flag();
+        static_header[8] = Self::TYPE_FLAG;
         static_header[9..21].copy_from_slice(nonce);
         static_header[21..].copy_from_slice(&authdata_size.to_be_bytes());
         let header_end_offset = 16 + authdata.len() + static_header.len();
         Ok(PacketHeader {
             static_header,
-            flag: self.packet_type_flag(),
+            flag: Self::TYPE_FLAG,
             nonce: *nonce,
             authdata,
             header_end_offset,
@@ -239,13 +239,11 @@ pub struct Ordinary {
 }
 
 impl PacketTrait for Ordinary {
+    const TYPE_FLAG: u8 = 0x00;
+
     fn encode_authdata(&self, buf: &mut dyn BufMut) -> Result<(), PacketCodecError> {
         buf.put_slice(self.src_id.as_bytes());
         Ok(())
-    }
-
-    fn packet_type_flag(&self) -> u8 {
-        0x00
     }
 
     fn get_encoded_message(&self) -> Vec<u8> {
@@ -301,14 +299,12 @@ pub struct WhoAreYou {
 }
 
 impl PacketTrait for WhoAreYou {
+    const TYPE_FLAG: u8 = 0x01;
+
     fn encode_authdata(&self, buf: &mut dyn BufMut) -> Result<(), PacketCodecError> {
         buf.put_slice(&self.id_nonce.to_be_bytes());
         buf.put_slice(&self.enr_seq.to_be_bytes());
         Ok(())
-    }
-
-    fn packet_type_flag(&self) -> u8 {
-        0x01
     }
 
     fn get_encoded_message(&self) -> Vec<u8> {
@@ -353,6 +349,8 @@ pub struct Handshake {
 }
 
 impl PacketTrait for Handshake {
+    const TYPE_FLAG: u8 = 0x02;
+
     fn encode_authdata(&self, buf: &mut dyn BufMut) -> Result<(), PacketCodecError> {
         let sig_size: u8 = self
             .id_signature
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/discv5/codec.rs
```diff
@@ -74,17 +34,8 @@ impl Decoder for Discv5Codec {
...
-            (_, Some(session)) => session.outbound_key(),
-            (_, None) => return Err(PacketCodecError::SessionNotEstablished),
-        };
-
-        packet.encode(buf, masking_iv, &nonce, &self.dest_id, key)
+    fn encode(&mut self, _packet: Packet, _buf: &mut BytesMut) -> Result<(), Self::Error> {
+        // We are not going to use Discv5Coded to send messages, only to receive them
+        unimplemented!();
     }
 }
```
azteca1998: If the entire trait is not needed, maybe just not implementing it would be better. Unless it's needed due to a restriction in the type system elsewhere, of course.

crates/networking/p2p/p2p.rs
```diff
@@ -82,3 +82,12 @@ pub mod utils;
 
 pub use network::periodically_show_peer_stats;
 pub use network::start_network;
+
+#[cfg(not(feature = "experimental-discv5"))]
+pub use discv4::peer_table;
+#[cfg(not(feature = "experimental-discv5"))]
+pub use discv4::server as discovery_server;
```
azteca1998:
```suggestion
#[cfg(not(feature = "experimental-discv5"))]
pub use discv4::{peer_table, server as discovery_server};
```

This may be more readable.

edg-l (approved):

crates/networking/p2p/discv5/messages.rs
```diff
@@ -374,25 +339,24 @@ impl Handshake {
...
             u16::try_from(authdata.len()).map_err(|_| PacketCodecError::InvalidSize)?;
 
-        let mut static_header = Vec::new();
-        static_header.put_slice(PROTOCOL_ID);
-        static_header.put_slice(&PROTOCOL_VERSION.to_be_bytes());
-        static_header.put_u8(0x02);
-        static_header.put_slice(nonce);
-        static_header.put_slice(&authdata_size.to_be_bytes());
+        let mut static_header: [u8; 23] = [0; 23];
+        static_header[0..6].copy_from_slice(PROTOCOL_ID);
```
iovoid: This logic is duplicated (triplicated if only looking at the header code), maybe it could be simplified?

crates/networking/p2p/discv5/peer_table.rs
```diff
@@ -0,0 +1,1319 @@
...
+            let distance = distance(&node_id, contact_id);
+            if nodes.len() < MAX_NODES_IN_NEIGHBORS_PACKET {
+                nodes.push((contact.node.clone(), distance));
+            } else {
+                for (i, (_, dis)) in &mut nodes.iter().enumerate() {
+                    if distance < *dis {
+                        nodes[i] = (contact.node.clone(), distance);
+                        break;
+                    }
+                }
```
iovoid, line 818: Why not binary search or similar?

crates/networking/p2p/discv5/peer_table.rs
```diff
@@ -0,0 +1,1319 @@
...
+                    None
+                } else {
+                    Some(*node_id)
+                }
+            })
+            .collect::<Vec<_>>()
+            .len()
+    }
+
+    fn get_peer_connections(&self, capabilities: Vec<Capability>) -> Vec<(H256, PeerConnection)> {
```
iovoid, line 876: It should probably be documented that it gives a node with any of the requested capabilities, not all of them.

crates/networking/p2p/discv5/server.rs
```diff
@@ -0,0 +1,624 @@
...
+    async fn handle_packet(
+        &mut self,
+        Discv5Message { packet, from }: Discv5Message,
+    ) -> Result<(), DiscoveryServerError> {
+        // TODO retrieve session info
+        match packet.header.flag {
+            0x00 => self.handle_ordinary(packet, from).await,
+            0x01 => self.handle_who_are_you(packet, from).await,
+            0x02 => {
+                tracing::info!("NonWhoAreYou!");
```
iovoid: Unclear log

crates/networking/p2p/discv5/server.rs
```diff
@@ -0,0 +1,624 @@
...
+    };
+    Duration::from_micros(
+        // Use `progress` here instead of `ease_in_out_cubic` for a linear function.
+        (1000f64 * (ease_in_out_cubic * (upper_limit - lower_limit) + lower_limit)).round() as u64,
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    // use rand::{SeedableRng, rngs::StdRng};
```
iovoid: Why commented out?

crates/networking/p2p/discv5/peer_table.rs
ElFantasma: Binary search requires sorting. Each request uses a different node_id so we cannot pre-sort the contact list.

In any case, this is code copied from discv4 and will be replaced once we implement `FindNodes` handler and K-bucket table.

crates/networking/p2p/discv5/peer_table.rs
ElFantasma: This is an internal callback function, but I will document the external API function

crates/networking/p2p/discv5/server.rs
ElFantasma: Yes, it's work in progress. Will add a comment.

crates/networking/p2p/discv5/messages.rs
ElFantasma: You are right, I was expecting to solve this in a future PR, but this may be a good time to do it.

ElFantasma:
> I've had decoding errors, apparently related to nodes sending the signature in an uncompressed format. Maybe this should be fixed, or the error should be downgraded to trace/debug and the limitation documented.

Fixed 👍

iovoid (approved):

PR #5783, opened 2026-01-08
fix(l1): include sender in MempoolTransaction RLP encoding

Forostovec:
Align the RLP encoding of MempoolTransaction with its RLPDecode implementation by including the sender field between timestamp and the inner Transaction. Previously the encoder only serialized timestamp and the inner transaction, while the decoder expected timestamp, sender and inner, which would break round‑tripping if the type is ever serialized. This field is not currently used over the network or persisted storage, but making the format consistent now avoids subtle bugs if RLP serialization of MempoolTransaction is introduced in the future and keeps it in line with the rest of the transaction types.


azteca1998 (approved):

ElFantasma (approved):

Arkenan (approved):
LGTM, but weird this wasn't caught earlier

PR #5786, opened 2026-01-08
feat(l1, l2): introduce a trait for vm tracers

lakshya-sky:
**Motivation**
                                                                                                                                           
Introduce a VM tracer trait for L1 and wire tracing hooks through opcode handling and txn lifecycle, while cleaning up naming and disabled call sites.
                                                                                                                                           
**Description**
                                                                                                                                           
- Add a VM tracer trait and evolve it with env/db/txn hooks.
- Reorder and refine tracer methods, including tx start/end and error handling.
- Update opcode handlers to call tracer hooks and remove disabled/levm call sites.
- Normalize naming around tracing and simplify exit handling.

### DIFF
--- a/crates/common/trie/node/branch.rs
+++ b/crates/common/trie/node/branch.rs
@@ -200,8 +200,7 @@ impl BranchNode {
         } else {
             // Remove own value (if it has one) and return it
             if !self.value.is_empty() {
-                let value = mem::take(&mut self.value);
-                (!value.is_empty()).then_some(value)
+                Some(mem::take(&mut self.value))
             } else {
                 None
             }
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -439,48 +439,47 @@ impl PeerHandler {
         start: H256,
         order: BlockRequestOrder,
     ) -> Result<Option<Vec<BlockHeader>>, PeerHandlerError> {
-        for _ in 0..REQUEST_RETRY_ATTEMPTS {
-            let request_id = rand::random();
-            let request = RLPxMessage::GetBlockHeaders(GetBlockHeaders {
-                id: request_id,
-                startblock: start.into(),
-                limit: BLOCK_HEADER_LIMIT,
-                skip: 0,
-                reverse: matches!(order, BlockRequestOrder::NewToOld),
-            });
-            match self.get_random_peer(&SUPPORTED_ETH_CAPABILITIES).await? {
-                None => return Ok(None),
-                Some((peer_id, mut connection)) => {
-                    if let Ok(RLPxMessage::BlockHeaders(BlockHeaders {
-                        id: _,
-                        block_headers,
-                    })) = PeerHandler::make_request(
-                        &mut self.peer_table,
-                        peer_id,
-                        &mut connection,
-                        request,
-                        PEER_REPLY_TIMEOUT,
-                    )
-                    .await
+        let request_id = rand::random();
+        let request = RLPxMessage::GetBlockHeaders(GetBlockHeaders {
+            id: request_id,
+            startblock: start.into(),
+            limit: BLOCK_HEADER_LIMIT,
+            skip: 0,
+            reverse: matches!(order, BlockRequestOrder::NewToOld),
+        });
+        match self.get_random_peer(&SUPPORTED_ETH_CAPABILITIES).await? {
+            None => Ok(None),
+            Some((peer_id, mut connection)) => {
+                if let Ok(RLPxMessage::BlockHeaders(BlockHeaders {
+                    id: _,
+                    block_headers,
+                })) = PeerHandler::make_request(
+                    &mut self.peer_table,
+                    peer_id,
+                    &mut connection,
+                    request,
+                    PEER_REPLY_TIMEOUT,
+                )
+                .await
+                {
+                    if !block_headers.is_empty()
+                        && are_block_headers_chained(&block_headers, &order)
                     {
-                        if !block_headers.is_empty()
-                            && are_block_headers_chained(&block_headers, &order)
-                        {
-                            return Ok(Some(block_headers));
-                        } else {
-                            warn!(
-                                "[SYNCING] Received empty/invalid headers from peer, penalizing peer {peer_id}"
-                            );
-                        }
+                        return Ok(Some(block_headers));
+                    } else {
+                        warn!(
+                            "[SYNCING] Received empty/invalid headers from peer, penalizing peer {peer_id}"
+                        );
+                        return Ok(None);
                     }
-                    // Timeouted
-                    warn!(
-                        "[SYNCING] Didn't receive block headers from peer, penalizing peer {peer_id}..."
-                    );
                 }
+                // Timeouted
+                warn!(
+                    "[SYNCING] Didn't receive block headers from peer, penalizing peer {peer_id}..."
+                );
+                Ok(None)
             }
         }
-        Ok(None)
     }
 
     /// Given a peer id, a chunk start and a chunk limit, requests the block headers from the peer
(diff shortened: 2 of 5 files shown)


lakshya-sky:
Hey @iovoid, You're right that `on_opcode` is not being used so that overhead unnecessary. I wasn't sure how to track reads and writes by VM except using a Trait, but if you have some suggestions I'd be happy to work on them.

crates/vm/levm/src/tracing/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+        _db: &mut GeneralizedDatabase,
+    ) {
+    }
+
+    /// Called after txn execution ends
+    fn txn_end(&mut self, _gas_used: u64, err: Option<String>, _db: &mut GeneralizedDatabase) {}
+
+    /// Called before each opcode execution. Used by prestate tracer to capture account lookups.
+    /// Returns true if tracing should continue, false to interrupt execution.
+    fn on_opcode(
```
ElFantasma, line 60: `on_opcode` is defined with a return-based interruption mechanism (`-> bool`) and documented as "used by prestate tracer," but it's never called from the execution loop or any opcode handler.

The other hooks (`on_account_access`, `on_storage_access`, `on_create`, `on_selfdestruct`) are all wired into their respective opcode handlers — this one is the only dead method. To be useful, it would need to be called at the top of `run_execution`'s main loop before each opcode dispatch. But that's also where it becomes expensive: a `RefCell::borrow_mut()` + vtable call on *every single opcode* has non-trivial overhead in the hot loop (see comment on `vm.rs:320`).

Suggestion: either wire it in this PR to validate the signature design works (especially the `-> bool` interruption and the performance cost), or remove it until a follow-up PR actually needs it. Dead interface tends to drift from real requirements.

crates/vm/levm/src/tracing/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    ) -> bool {
+        true
+    }
+
+    /// Called when a storage slot is accessed (SLOAD/SSTORE).
+    fn on_storage_access(&mut self, _address: Address, _slot: H256, _db: &mut GeneralizedDatabase) {
+    }
+
+    /// Called when an account is accessed (BALANCE, EXTCODESIZE, EXTCODEHASH, EXTCODECOPY).
+    fn on_account_access(&mut self, _address: Address, _db: &mut GeneralizedDatabase) {}
```
ElFantasma, line 75: All the `on_*` hooks, `txn_start`, and `txn_end` take `_db: &mut GeneralizedDatabase`. Tracers are observation-only — they should read state, not modify it.

`&mut` is currently needed because `GeneralizedDatabase::get_account()` takes `&mut self` (for internal caching). But the signature as-is gives tracer implementations the power to call mutating methods (`cache_account`, write operations, etc.), which is a correctness hazard: a buggy tracer could silently corrupt EVM state during tracing.

Two possible fixes:
1. **Short-term:** Add a doc comment on the trait methods clarifying that `db` is `&mut` only because the DB cache requires it, and implementations must not modify state.
2. **Longer-term:** Introduce a `ReadableDatabase` wrapper (or trait) that exposes only `get_account`, `get_storage_value`, `get_account_code` — then pass `&ReadableDatabase` to tracer methods.

Option 1 is probably fine for now since all tracer implementations live in-tree.

crates/vm/levm/src/vm.rs
```diff
@@ -316,7 +317,7 @@ pub struct VM<'a> {
     /// Original storage values before the transaction. Used for gas calculations in SSTORE.
     pub storage_original_values: BTreeMap<(Address, H256), U256>,
     /// When enabled, it "logs" relevant information during execution
-    pub tracer: LevmCallTracer,
+    pub tracer: Rc<RefCell<dyn Tracer>>,
```
ElFantasma, line 320: `Rc<RefCell<dyn Tracer>>` means every tracer call pays for:
- `RefCell::borrow_mut()` — runtime borrow check
- vtable dispatch — no inlining through `dyn Tracer`

This runs on every `CALL*`, `CREATE*`, `SLOAD`, `SSTORE`, `BALANCE`, `EXTCODESIZE`, `EXTCODECOPY`, `EXTCODEHASH`, `LOG*`, and `SELFDESTRUCT`. For a block with thousands of transactions, each executing hundreds of these opcodes, the overhead of `NoOpTracer` being dynamically dispatched (instead of compiled away) adds up.

The existing `Hook` system uses the same `Rc<RefCell<dyn Hook>>` pattern, but hooks only fire on `prepare_execution`/`finalize_execution` (once per tx). Tracer hooks fire on every opcode that touches accounts or storage.

Consider making `VM` generic over the tracer: `VM<'a, T: Tracer = NoOpTracer>`. This lets the compiler monomorphize and inline all `NoOpTracer` calls to no-ops at zero runtime cost, while still supporting `VM<'a, LevmCallTracer>` for tracing. The `Rc<RefCell<dyn Tracer>>` approach would still be available as `VM<'a, Rc<RefCell<dyn Tracer>>>` if dynamic dispatch is needed.

crates/vm/levm/src/tracing/call_tracer.rs
```diff
@@ -167,6 +126,40 @@ impl LevmCallTracer {
...
+    fn enter(
+        &mut self,
+        call_type: CallType,
+        from: Address,
+        to: Address,
+        value: U256,
+        gas: u64,
+        input: &Bytes,
+    ) {
+        self.enter(call_type, from, to, value, gas, input);
```
ElFantasma, line 146: `self.enter(...)` inside `impl Tracer for LevmCallTracer` resolves to the inherent `LevmCallTracer::enter()` method (not the trait method) because Rust prioritizes inherent methods over trait methods on `self`. Same for `self.exit()` on line 157 and `self.log()` on line 161.

This works but is fragile: if someone renames or removes the inherent `enter` method, `self.enter()` silently becomes `<Self as Tracer>::enter()` — infinite recursion at runtime with no compiler warning.

Safer alternatives:
1. Rename the inherent methods (e.g., `enter_impl`, `exit_impl`) and call those explicitly
2. Remove the inherent methods entirely and implement the logic directly in the trait impl
3. Use fully-qualified syntax: `LevmCallTracer::enter(self, ...)` to make the delegation explicit

Option 2 is cleanest since only the trait interface is public.

crates/vm/levm/src/tracing/mod.rs
```diff
@@ -0,0 +1,86 @@
+#![allow(unused_variables)]
```
ElFantasma, line 1: nit: `#![allow(unused_variables)]` at module level is overly broad. The trait default methods already use underscore-prefixed parameters (`_env`, `_tx`, `_db`, etc.) which individually suppress unused-variable warnings. This module-level allow could mask legitimate unused variables in future code added to this module (e.g., in `call_tracer.rs` which is a submodule). Consider removing it.

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -543,6 +560,12 @@ impl<'a> VM<'a> {
             (target_address, to)
         };
 
+        // Notify tracer of account access and selfdestruct
+        self.tracer
+            .borrow_mut()
+            .on_account_access(beneficiary, self.db);
+        self.tracer.borrow_mut().on_selfdestruct(to, self.db);
```
ElFantasma, line 567: nit: Two consecutive `borrow_mut()` calls on the same `RefCell`. Each one acquires and releases the borrow. You can coalesce them:
```rust
{
    let mut tracer = self.tracer.borrow_mut();
    tracer.on_account_access(beneficiary, self.db);
    tracer.on_selfdestruct(to, self.db);
}
```
Same pattern appears in `generic_create` (lines 664-665) with `on_account_access` + `on_create`.

With `NoOpTracer` the optimizer likely eliminates both calls entirely, but with a real tracer this saves a borrow/unborrow cycle.

ElFantasma:
This PR got stalled and it has a lot of conflicts. 

We don't think this will be necessary, but if you consider otherwise feel free to reopen with a note and rebase.

Thanks anyway for the contribution.

PR #5788, opened 2026-01-08
test(l1): add fuzzing tests and security workflows

ManuelBilbao:
**Motivation**

Add more tests

**Description**

Add a fuzzing toolset that tests precompiles, trie, rlp, encoding/decoding.
Also added workflows to run the fuzzing tests and security checks

### DIFF
--- a/tooling/fuzz/fuzzers/precompile_bls12_381.rs
+++ b/tooling/fuzz/fuzzers/precompile_bls12_381.rs
@@ -17,19 +17,33 @@
 
 use arbitrary::Arbitrary;
 use bytes::Bytes;
-use ethrex_common::types::Fork;
 use ethrex_common::H160;
+use ethrex_common::types::Fork;
 use ethrex_levm::precompiles::execute_precompile;
 use libfuzzer_sys::fuzz_target;
 
 // BLS12-381 precompile addresses
-const BLS12_G1ADD: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0b]);
-const BLS12_G1MSM: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0c]);
-const BLS12_G2ADD: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0d]);
-const BLS12_G2MSM: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0e]);
-const BLS12_PAIRING: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0f]);
-const BLS12_MAP_FP_TO_G1: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x10]);
-const BLS12_MAP_FP2_TO_G2: H160 = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x11]);
+const BLS12_G1ADD: H160 = H160([
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0b,
+]);
+const BLS12_G1MSM: H160 = H160([
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0c,
+]);
+const BLS12_G2ADD: H160 = H160([
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x0d,
+]);
+const BLS12_G2MSM: H160 = H160([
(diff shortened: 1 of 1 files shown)


tooling/fuzz/fuzzers/transaction_decode.rs
```diff
@@ -0,0 +1,48 @@
...
+                0x03 => {
+                    let _ = EIP4844Transaction::decode(tx_data);
+                }
+                0x04 => {
+                    let _ = EIP7702Transaction::decode(tx_data);
+                }
+                _ => {}
+            }
+        }
+    }
```
iovoid, line 47: Why attempt to decode twice for each type? (L22-26, and here)

tooling/fuzz/fuzzers/precompile_all.rs
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    // Select precompile address
+    let precompile_idx = (input.precompile_index as usize) % PRECOMPILE_ADDRESSES.len();
+    let address = make_precompile_address(PRECOMPILE_ADDRESSES[precompile_idx]);
+
+    let calldata = Bytes::from(input.calldata);
+
+    // Use a generous gas limit to allow the precompile to run
+    let mut gas_remaining: u64 = 10_000_000;
```
iovoid, line 75: This should probably be set somewhere above the EIP-7825 cap.

tooling/fuzz/fuzzers/precompile_bls12_381.rs
```diff
@@ -0,0 +1,478 @@
...
+    RawG1Msm(Vec<u8>),
+    RawG2Add(Vec<u8>),
+    RawG2Msm(Vec<u8>),
+    RawPairing(Vec<u8>),
+    RawMapFpToG1(Vec<u8>),
+    RawMapFp2ToG2(Vec<u8>),
+}
+
+fuzz_target!(|input: Bls12Input| {
+    let mut gas_remaining: u64 = 100_000_000;
```
iovoid, line 198: We should probably use the same gas limit for all tests.

.github/workflows/fuzz.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+      - name: Run fuzzer
+        if: steps.check.outputs.run == 'true'
+        working-directory: fuzz
+        run: |
+          cargo fuzz run ${{ matrix.target }} -- \
+            -max_total_time=${{ steps.check.outputs.duration }} \
+            -max_len=4096
+
+      - name: Upload corpus
```
ElFantasma, line 119: `working-directory: fuzz` is wrong — the fuzz crate is at `tooling/fuzz`. This will fail with a "Cargo.toml not found" error.

Should be:
```yaml
working-directory: tooling/fuzz
```

Same issue on line 103 with `workspaces: fuzz -> target` — should be `tooling/fuzz -> target`. And the `fuzz-compile-check` job at the bottom has the same problem.

.github/workflows/security.yaml
```diff
@@ -0,0 +1,143 @@
+name: Security
+
+on:
+  push:
```
ElFantasma, line 4: This workflow triggers on every push to `main` and every PR. The `cargo-geiger` and `sarif-clippy` jobs are quite heavy (full workspace compile + analysis). Consider:
- Running `cargo-audit` and `cargo-deny` on PRs (they're fast)
- Restricting `cargo-geiger`, `clippy-security`, and `sarif-clippy` to `schedule` only

This prevents CI from being slowed down on every PR for checks better suited as periodic audits.

.github/workflows/security.yaml
```diff
@@ -0,0 +1,143 @@
...
+            -D clippy::unimplemented \
+            -D clippy::unreachable \
+            -W clippy::integer_arithmetic \
+            -W clippy::arithmetic_side_effects \
+            -W clippy::cast_possible_truncation \
+            -W clippy::cast_sign_loss \
+            -W clippy::cast_possible_wrap \
+            -W clippy::indexing_slicing
+        continue-on-error: true  # Report but don't fail initially
+
```
ElFantasma, line 85: `cargo-geiger` is known to be unreliable — it often fails to compile crates with proc macros or complex build scripts, and is very slow. The `continue-on-error: true` mitigates breakage, but this still consumes CI minutes on every PR for a tool that frequently produces incomplete results.

Consider either removing this job or restricting it to scheduled runs only.

crates/common/trie/node/leaf.rs
```diff
@@ -37,8 +37,9 @@ impl LeafNode {
     }
 
     /// Returns the stored value if the given path matches the stored path
+    /// Returns None if the value is empty (Ethereum trie semantics: empty value = no value)
     pub fn get(&self, path: Nibbles) -> Result<Option<ValueRLP>, TrieError> {
-        if self.partial == path {
+        if self.partial == path && !self.value.is_empty() {
```
ElFantasma, line 42: This makes `LeafNode::get()` consistent with `BranchNode::get()` (which already does `(!self.value.is_empty()).then_some(...)` at branch.rs:77). The semantics are correct for Ethereum tries: empty value = deleted.

However, this is a behavioral change in consensus-critical trie code. Was this found by the fuzzer, or added preemptively? If the fuzzer discovered a case where a leaf ends up with an empty value, that's worth documenting as a bug fix. If preemptive, the root cause (how a leaf gets an empty value) should be addressed instead — `insert()` and `remove()` don't seem to create this situation in normal operation.

Either way, this deserves its own focused PR with a targeted regression test rather than being bundled with the fuzzing infrastructure.

deny.toml
```diff
@@ -0,0 +1,68 @@
...
+name = "ring"
+expression = "MIT AND ISC AND OpenSSL"
+license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
+
+[bans]
+multiple-versions = "warn"
+wildcards = "allow"
+highlight = "all"
+
+# Deny specific crates known to have issues
```
ElFantasma, line 48: The `ring` license hash `0xbd0eed23` will break whenever `ring` updates its LICENSE file. This is a known `cargo-deny` pain point — just flagging that you'll need to update this hash periodically.

PR #5790, opened 2026-01-08
fix(l2): misc contract fixes

ManuelBilbao:
**Motivation**

An AI analysis threw some errors and vulnerabilities in the contracts. Some of them were not actually errors, but could lead to them in the future.

**Description**

- Changed `revertBatch` API to a more clearer one: now the `batchNumber` param indicates which batch to revert, among with its followings.
- Add shared bridge logic (balance transfers) in `verifyBatchesAligned`.
- Add validation of verifiers address in OCP initializer.
- Add reentrancy guard to `claimWithdraw`.
- Update `deposits` count in `receiveETHFromSharedBridge`.
- Don't allow crosschain transactions in Router when paused.
- Add `onlySelf` modifier to `mintETH` in CommonBridgeL2

### DIFF
--- a/crates/l2/contracts/src/l1/CommonBridge.sol
+++ b/crates/l2/contracts/src/l1/CommonBridge.sol
@@ -464,6 +464,7 @@ contract CommonBridge is
             IRouter(SHARED_BRIDGE_ROUTER).sendETHValue{
                 value: balanceDiffs[i].value
             }(balanceDiffs[i].chainId);
+            deposits[ETH_TOKEN][ETH_TOKEN] -= balanceDiffs[i].value;
 
             // Send ERC20 values if any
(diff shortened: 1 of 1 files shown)


crates/l2/contracts/src/l1/CommonBridge.sol
avilagaston9, line 1: Shouldn't we update `publishL2Messages` as well?
I think we should add something like this:
```
deposits[ETH_TOKEN][ETH_TOKEN] -= balanceDiffs[i].value;
```

ManuelBilbao: Nice catch! [2d427a5](https://github.com/lambdaclass/ethrex/pull/5790/commits/2d427a55527f02983d6f863d1ea3194e3e30fbac)

PR #5803, opened 2026-01-09
feat(l1): send ENR optionally on Handshake messages

### LINKED ISSUE #5777: Check if we have to send ENR according to enr-seq
When we handle `WhoAreYou` messages we need to decide if we have to send our ENR. Currently we are always sending it. We don't need to send it if the `enr-seq` is up to date. 
From the [spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#step-3-node-a-processes-the-challenge):

> node A compares the enr-seq element of the WHOAREYOU challenge against its own node record sequence number. If the sequence number in the challenge is lower, it includes its record into the handshake message packet.

See `TODO` comment [here](https://github.com/lambdaclass/ethrex/blob/dff4a4a2a2bdc932baa7274dac998563c1e6e704/crates/networking/p2p/discv5/server.rs#L179)

ElFantasma:
**Motivation**
From the [spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#step-3-node-a-processes-the-challenge):

> node A compares the enr-seq element of the WHOAREYOU challenge against its own node record sequence number. If the sequence number in the challenge is lower, it includes its record into the handshake message packet.

**Description**

Checks the sequence number and sends the ENR only if the sequences doesn't match.

depends on https://github.com/lambdaclass/ethrex/pull/5799

Closes #5777

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4128,6 +4128,7 @@ dependencies = [
 name = "ethrex-test"
 version = "9.0.0"
 dependencies = [
+ "anyhow",
  "bytes",
  "cita_trie",
  "ethereum-types 0.15.1",
@@ -4135,10 +4136,13 @@ dependencies = [
  "ethrex-blockchain",
  "ethrex-common",
  "ethrex-crypto",
+ "ethrex-l2-common",
  "ethrex-levm",
  "ethrex-p2p",
  "ethrex-rlp",
+ "ethrex-sdk",
  "ethrex-storage",
+ "ethrex-storage-rollup",
  "ethrex-trie",
  "hasher",
  "hex",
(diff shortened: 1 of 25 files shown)


crates/networking/p2p/discv5/server.rs
```diff
@@ -225,10 +223,13 @@ impl DiscoveryServer {
...
             .set_session_info(node.node_id(), session)
             .await?;
-        self.send_handshake(&message, signature, &ephemeral_pubkey, &node)
-            .await?;
 
-        Ok(())
+        // Check enr-seq to decide if we have to send the local ENR in the handshake.
+        let whoareyou = WhoAreYou::decode(&packet)?;
+        let record = (self.local_node_record.seq != whoareyou.enr_seq)
+            .then_some(self.local_node_record.clone());
```
Oppen: Note this is cloning the record unconditionally. You might want to use `.then(|| self.local_node_record.clone())` instead.

PR #5805, opened 2026-01-09
chore(l2): remove unwanted error log on L2

tomip01:
**Motivation**

This error is not useful in L2. Empty blocks with same state root is an expected behavior on L2.

On L1 due to calling the system contracts there is always a change in the state. This does not happen on L2, so having a block with no transactions leads to having the same state root.

### DIFF
--- a/crates/storage/layering.rs
+++ b/crates/storage/layering.rs
@@ -112,9 +112,9 @@ impl TrieLayerCache {
         if parent == state_root && key_values.is_empty() {
             return;
         } else if parent == state_root {
-            // L1 always changes the state root (system contracts run even on empty blocks), so this
-            // is inconsistent there. L2 can legitimately keep the same root on empty blocks because
-            // it has no system contract calls.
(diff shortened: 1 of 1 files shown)


crates/storage/layering.rs
```diff
@@ -112,7 +112,10 @@ impl TrieLayerCache {
         if parent == state_root && key_values.is_empty() {
             return;
         } else if parent == state_root {
-            tracing::error!("Inconsistent state: parent == state_root but key_values not empty");
+            // L1 always changes the state root (system contracts run even on empty blocks), so this
+            // is inconsistent there. L2 can legitimately keep the same root on empty blocks because
```
iovoid: Probably better to say "should not happen" or similar rather than inconsistent (which would imply a corrupt trie)

tomip01: Right, changed here [802a700](https://github.com/lambdaclass/ethrex/pull/5805/commits/802a700884390dac45c48a456a12739d5725e73d)!

iovoid (approved):

PR #5817, opened 2026-01-12
refactor(l2): trait-based prover backend API with associated types

ilitteri:
**Motivation**

The prover backend code had several issues:
  - Multiple backend implementations (SP1, RISC0, ZisK, OpenVM, Exec) with inconsistent APIs
  - Dynamic dispatch via Box<dyn ProverBackend> prevented type-safe proof handling
  - Error handling was scattered and inconsistent
  - Unnecessary indirection layers in lib.rs
  - super:: imports made module relationships unclear

**Description**

  This PR introduces a clean trait-based architecture for prover backends:

  - `ProverBackend` trait with associated type: Each backend defines its own type ProofOutput, enabling type-safe proof handling without dynamic dispatch
```
  pub trait ProverBackend {
      type ProofOutput;
      fn prove(&self, input: ProgramInput, format: ProofFormat) -> Result<Self::ProofOutput, BackendError>;
      // ...
  }
```
  - `BackendError` with `thiserror`: Centralized error type with clear variants for serialization, execution, proving, verification, and batch proof conversion
  - Static dispatch in `Prover<B>`: Generic prover struct eliminates runtime overhead
  `struct Prover<B: ProverBackend> { backend: B, ... }`
  - Removed indirection: Eliminated wrapper functions in `lib.rs` that just forwarded to trait methods
  - Consistent imports: Changed `super::` to `crate::backend::` throughout

**Checklist**

Closes #issue_number

### DIFF
--- a/crates/l2/prover/src/backend/exec.rs
+++ b/crates/l2/prover/src/backend/exec.rs
@@ -1,3 +1,5 @@
+use std::time::{Duration, Instant};
+
 use tracing::{info, warn};
 
 use ethrex_l2_common::{
@@ -20,7 +22,8 @@ impl ExecBackend {
         Self
     }
 
-    fn execution_program(input: ProgramInput) -> Result<ProgramOutput, BackendError> {
+    /// Core execution - runs the guest program directly.
+    fn execute_core(input: ProgramInput) -> Result<ProgramOutput, BackendError> {
         guest_program::execution::execution_program(input).map_err(BackendError::execution)
     }
 
@@ -45,11 +48,7 @@ impl ProverBackend for ExecBackend {
     }
 
     fn execute(&self, input: ProgramInput) -> Result<(), BackendError> {
-        let now = std::time::Instant::now();
-        Self::execution_program(input)?;
-        let elapsed = now.elapsed();
-
-        info!("Successfully executed program in {:.2?}", elapsed);
+        Self::execute_core(input)?;
         Ok(())
     }
 
@@ -59,7 +58,7 @@ impl ProverBackend for ExecBackend {
         _format: ProofFormat,
     ) -> Result<Self::ProofOutput, BackendError> {
         warn!("\"exec\" prover backend generates no proof, only executes");
-        Self::execution_program(input)
+        Self::execute_core(input)
     }
 
     fn verify(&self, _proof: &Self::ProofOutput) -> Result<(), BackendError> {
@@ -73,5 +72,13 @@ impl ProverBackend for ExecBackend {
         _format: ProofFormat,
     ) -> Result<BatchProof, BackendError> {
         Ok(BatchProof::ProofCalldata(Self::to_calldata()))
+    }
+
+    fn execute_timed(&self, input: ProgramInput) -> Result<Duration, BackendError> {
+        let start = Instant::now();
+        Self::execute_core(input)?;
+        let elapsed = start.elapsed();
+        info!("Successfully executed program in {:.2?}", elapsed);
+        Ok(elapsed)
     }
 }
--- a/crates/l2/prover/src/backend/openvm.rs
+++ b/crates/l2/prover/src/backend/openvm.rs
@@ -43,7 +43,9 @@ impl OpenVmBackend {
         let sdk = Sdk::standard();
         let proof = match format {
             ProofFormat::Compressed => {
-                let (proof, _) = sdk.prove(PROGRAM_ELF, stdin).map_err(BackendError::proving)?;
+                let (proof, _) = sdk
+                    .prove(PROGRAM_ELF, stdin)
+                    .map_err(BackendError::proving)?;
                 OpenVmProveOutput::Compressed(proof)
             }
             ProofFormat::Groth16 => {
--- a/crates/l2/prover/src/backend/sp1.rs
+++ b/crates/l2/prover/src/backend/sp1.rs
@@ -13,7 +13,11 @@ use sp1_sdk::{
     HashableKey, Prover, SP1ProofMode, SP1ProofWithPublicValues, SP1ProvingKey, SP1Stdin,
     SP1VerifyingKey,
 };
-use std::{fmt::Debug, sync::OnceLock, time::{Duration, Instant}};
+use std::{
+    fmt::Debug,
+    sync::OnceLock,
+    time::{Duration, Instant},
+};
 use url::Url;
 
 use crate::backend::{BackendError, ProverBackend};


crates/l2/prover/src/backend/openvm.rs
```diff
@@ -1,98 +1,89 @@
...
-    };
-    let duration = start.elapsed();
-
-    Ok((proof, duration))
+/// OpenVM prover backend.
+#[derive(Default)]
+pub struct OpenVmBackend;
+
+impl OpenVmBackend {
+    pub fn new() -> Self {
```
ManuelBilbao, line 25: Is it worth to have this function? The same could be achieved with `OpenVmBackend {}`

crates/l2/prover/src/lib.rs
```diff
@@ -1,109 +1,16 @@
...
-use ethrex_l2_common::prover::{BatchProof, ProofFormat};
-use guest_program::input::ProgramInput;
-use std::time::Duration;
 use tracing::warn;
 
-use crate::backend::{Backend, ProveOutput};
+pub use crate::backend::{BackendError, BackendType, ExecBackend, ProverBackend};
+
+#[cfg(feature = "sp1")]
+pub use crate::backend::Sp1Backend;
```
ManuelBilbao, line 11: Why only this one?

crates/l2/prover/src/backend/error.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    #[error("Proving error: {0}")]
+    Proving(String),
+
+    #[error("Verification error: {0}")]
+    Verification(String),
+
+    #[error("Verify not supported for this backend")]
+    VerifyNotSupported,
+
```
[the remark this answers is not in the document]
tomip01: agree. Maybe we could add a function in the impl block below to not make the enum bigger
```
fn verify_not_supported() -> Self {
    Self::NotImplemented("Verify not implemented for this backend".to_string())
}
```

crates/l2/prover/src/backend/mod.rs
```diff
@@ -33,33 +54,73 @@ pub enum Backend {
...
-    #[cfg(feature = "zisk")]
-    ZisK(zisk::ProveOutput),
-    #[cfg(feature = "openvm")]
-    OpenVM(openvm::ProveOutput),
+/// Trait defining the interface for prover backends.
+///
+/// All proving backends (SP1, RISC0, ZisK, OpenVM, Exec) implement this trait,
+/// providing a unified interface for execution, proving, verification, and
+/// batch proof conversion.
+pub trait ProverBackend {
```
tomip01, line 81: I think we should include `serialize_input` as part of the trait, all the backends implements this outside the trait

tomip01: Maybe in the future we want to add `to_calldata()` to the trait too. Since the backends that implement `to_batch_proof` also implement that

crates/l2/prover/src/lib.rs
ilitteri: Forgot to add the other ones. Adding them asap.

crates/l2/prover/src/backend/openvm.rs
ilitteri: Yes, it's Rust convention, it's always better to instantiate structs like: `MyStruct::new()` than `MyStruct {}`, even if they are a unit struct.

crates/l2/prover/src/backend/exec.rs
```diff
@@ -7,63 +6,64 @@ use ethrex_l2_common::{
...
-) -> Result<ProgramOutput, Box<dyn std::error::Error>> {
-    warn!("\"exec\" prover backend generates no proof, only executes");
-    let output = execution_program(input)?;
-    Ok(output)
+    fn to_calldata(proof: &ProgramOutput) -> ProofCalldata {
+        let public_inputs = proof.encode();
+        ProofCalldata {
+            prover_type: ProverType::Exec,
+            calldata: vec![Value::Bytes(public_inputs.into())],
+        }
```
avilagaston9: We don't need the public inputs anymore. We are simply returning `vec![Value::Bytes(vec![].into())]`.

crates/l2/prover/src/backend/exec.rs
```diff
@@ -7,63 +6,64 @@ use ethrex_l2_common::{
...
+/// This backend is useful for testing and debugging, as it runs the guest
+/// program directly without the overhead of proof generation.
+#[derive(Default)]
+pub struct ExecBackend;
 
-pub fn execute_timed(input: ProgramInput) -> Result<Duration, Box<dyn std::error::Error>> {
-    let now = Instant::now();
-    execution_program(input)?;
-    let duration = now.elapsed();
+impl ExecBackend {
```
avilagaston9, line 20: Shouldn't this implement the trait?

crates/l2/prover/src/backend/openvm.rs
avilagaston9, line 1: here we are using the default `prove_timed`, is that OK?

crates/l2/prover/src/backend/risc0.rs
```diff
@@ -6,158 +6,132 @@ use guest_program::{
...
+    }
+
+    fn to_calldata(receipt: &Receipt) -> Result<ProofCalldata, BackendError> {
+        let seal = Self::encode_seal(receipt)?;
+        let journal = receipt.journal.bytes.clone();
+
+        // bytes calldata seal,
+        // bytes32 imageId,
+        // bytes journal
+        let calldata = vec![Value::Bytes(seal.into()), Value::Bytes(journal.into())];
```
avilagaston9: Same here, the journal is no longer needed

crates/l2/prover/src/backend/openvm.rs
avilagaston9: Same with `execute_timed` and the rest of provers. Now we’re also measuring the time it takes to write the input to stdin.

crates/l2/prover/src/backend/sp1.rs
```diff
@@ -56,128 +55,127 @@ pub fn init_prover_setup(_endpoint: Option<Url>) -> ProverSetup {
...
+            ProofFormat::Groth16 => SP1ProofMode::Groth16,
+        }
+    }
 
-    let proof = setup.client.prove(&setup.pk, &stdin, format)?;
+    fn to_calldata(proof: &Sp1ProveOutput) -> ProofCalldata {
+        let calldata = vec![
+            Value::Bytes(proof.proof.public_values.to_vec().into()),
+            Value::Bytes(proof.proof.bytes().into()),
+        ];
```
avilagaston9:
```suggestion
            Value::Bytes(proof.proof.bytes().into()),
        ];
```

avilagaston9:
We must also ensure that the `L2 (SP1 Backend)` workflow passes before merging this PR

crates/l2/prover/src/backend/error.rs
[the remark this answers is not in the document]
xqft: agreed with tomi

crates/l2/prover/src/backend/error.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    pub fn batch_proof(e: impl std::fmt::Display) -> Self {
+        Self::BatchProofConversion(e.to_string())
+    }
+
+    pub fn not_implemented(msg: impl Into<String>) -> Self {
+        Self::NotImplemented(msg.into())
+    }
+}
```
xqft, line 51: we don't usually wrap errors in functions like this, to stay consistent we should return the enum directly, but it's a valid change anyways if we want to change the style.

crates/l2/prover/src/backend/openvm.rs
xqft: agree with Ivan here

crates/l2/prover/src/backend/exec.rs
```diff
@@ -7,63 +6,72 @@ use ethrex_l2_common::{
...
-    input: ProgramInput,
-    _format: ProofFormat,
-) -> Result<(ProgramOutput, Duration), Box<dyn std::error::Error>> {
-    warn!("\"exec\" prover backend generates no proof, only executes");
-    let now = Instant::now();
-    let output = execution_program(input)?;
-    let duration = now.elapsed();
+impl ProverBackend for ExecBackend {
+    type ProofOutput = ProgramOutput;
+    type SerializedInput = ();
```
avilagaston9, line 41: This method has its own timing + logging, but the default `execute_timed` also wraps `execute` with timing. If `execute_timed` is called, timing is measured twice.

Other backends use helper methods (`execute_with_stdin`, `execute_core`, etc.) to separate core logic from timing. Consider doing the same here, or override `execute_timed` like other backends.

crates/l2/prover/src/backend/zisk.rs
```diff
@@ -1,164 +1,192 @@
...
+    ) -> Result<Self::ProofOutput, BackendError> {
+        Self::write_elf_file()?;
+        self.serialize_input(&input)?;
+        self.prove_core(format)
     }
 
-    let zisk_result_bytes = std::fs::read(format!("{OUTPUT_DIR_PATH}/result.json"))?;
+    fn execute_timed(&self, input: ProgramInput) -> Result<Duration, BackendError> {
+        Self::write_elf_file()?;
+        self.serialize_input(&input)?;
```
avilagaston9, line 141: Timing inconsistency: `execute_timed` uses `Instant::now()` (wall-clock), but `prove_timed` reads from ZisK's `result.json`. Consider using consistent timing measurement across both methods.

ilitteri: I know this, but it's a limitation on the current development status of ZisK. This will be solved once we use ZisK's SDK.

PR #5822, opened 2026-01-13
perf(l1,l2): avoid unnecessary Arc::make_mut in trie iterator

Snezhkko:
**Motivation**

The code only reads the node and never mutates it, so Arc::make_mut could trigger deep clones when the node is shared.

**Description**

Replace Arc::make_mut in TrieIterator::advance’s first_ge helper with a match on &Node and drop the mutable binding.

### DIFF
--- a/crates/common/trie/trie_iter.rs
+++ b/crates/common/trie/trie_iter.rs
@@ -51,7 +51,7 @@ impl TrieIterator {
             };
             match next_node.as_ref() {
                 Node::Branch(branch_node) => {
-                    // Add all children to the stack (in reverse order so we process first child frist)
+                    // Add all children to the stack (in reverse order so we process first child first)
                     let Some(choice) = target_nibbles.next_choice() else {
                         return Ok(());
                     };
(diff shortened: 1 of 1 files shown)


azteca1998 (approved):

ElFantasma (approved):

crates/common/trie/trie_iter.rs
```diff
@@ -42,14 +42,14 @@ impl TrieIterator {
...
                 .get_node_checked(db, prefix_nibbles.clone())
                 .ok()
                 .flatten()
             else {
                 return Ok(());
             };
-            match Arc::make_mut(&mut next_node) {
+            match next_node.as_ref() {
                 Node::Branch(branch_node) => {
                     // Add all children to the stack (in reverse order so we process first child frist)
```
ElFantasma:
```suggestion
                    // Add all children to the stack (in reverse order so we process first child first)
```

ElFantasma (approved):

PR #5830, opened 2026-01-13
perf(l2): reduce allocations in blob reconstruction and blob utils

GarmashAlex:
**Motivation**

Expensive allocations in Reconstruct when working with blobs. First, `std::fs::read` returns `Vec<u8>`, then it is converted to `Bytes` inside `bytes_from_blob(blob.into())`, then back to `Vec<u8>` and then another copy is created via `Bytes::copy_from_slice(&blob)` for `blob_from_bytes`.

**Description**

Removed redundant blob re-encoding in the L2 Reconstruct command by reusing the raw blob bytes read from disk when sealing batches, while still decoding block and fee data via bytes_from_blob. Also rewrited blob_from_bytes and bytes_from_blob to operate in a single pass without intermediate Vec allocations, preserving the existing encoding format but significantly reducing heap usage across committer, prover and reconstruction paths that work with blobs.

### DIFF
--- a/crates/common/types/blobs_bundle.rs
+++ b/crates/common/types/blobs_bundle.rs
@@ -44,26 +44,35 @@ pub fn blob_from_bytes(bytes: Bytes) -> Result<Blob, BlobsBundleError> {
     }
 
     let mut buf = [0u8; BYTES_PER_BLOB];
-    buf[..(bytes.len() * 32).div_ceil(31)].copy_from_slice(
-        &bytes
-            .chunks(31)
-            .map(|x| [&[0x00], x].concat())
-            .collect::<Vec<_>>()
-            .concat(),
-    );
+    let mut dst = 0usize;
+
+    for chunk in bytes.chunks(31) {
+        let needed = 1 + chunk.len();
+        if dst + needed > BYTES_PER_BLOB {
+            return Err(BlobsBundleError::BlobDataInvalidBytesLength);
+        }
+
+        buf[dst] = 0x00;
+        dst += 1;
+
+        buf[dst..dst + chunk.len()].copy_from_slice(chunk);
+        dst += chunk.len();
+    }
 
     Ok(buf)
 }
 
 pub fn bytes_from_blob(blob: Bytes) -> [u8; SAFE_BYTES_PER_BLOB] {
     let mut buf = [0u8; SAFE_BYTES_PER_BLOB];
-    buf.copy_from_slice(
-        &blob
-            .chunks(32)
-            .map(|x| x[1..].to_vec())
-            .collect::<Vec<_>>()
-            .concat(),
-    );
+    let mut dst = 0usize;
+
+    for chunk in blob.chunks(32) {
+        let payload = &chunk[1..];
+        let len = payload.len();
+
+        buf[dst..dst + len].copy_from_slice(payload);
+        dst += len;
+    }
 
     buf
 }
--- a/crates/common/types/blobs_bundle.rs
+++ b/crates/common/types/blobs_bundle.rs
@@ -47,7 +47,8 @@ pub fn blob_from_bytes(bytes: Bytes) -> Result<Blob, BlobsBundleError> {
     let mut dst = 0usize;
 
     for chunk in bytes.chunks(31) {
-        let needed = 1 + chunk.len();
+        let len = chunk.len();
+        let needed = 1 + len;
         if dst + needed > BYTES_PER_BLOB {
             return Err(BlobsBundleError::BlobDataInvalidBytesLength);
         }
@@ -55,8 +56,8 @@ pub fn blob_from_bytes(bytes: Bytes) -> Result<Blob, BlobsBundleError> {
         buf[dst] = 0x00;
         dst += 1;
 
-        buf[dst..dst + chunk.len()].copy_from_slice(chunk);
-        dst += chunk.len();
+        buf[dst..dst + len].copy_from_slice(chunk);
+        dst += len;
     }
 
     Ok(buf)
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -8,14 +8,13 @@ use crate::{
     },
     utils::{self, default_datadir, init_datadir, parse_private_key},
 };
-use bytes::Bytes;
 use clap::{FromArgMatches, Parser, Subcommand};
 use ethrex_blockchain::{
     Blockchain, BlockchainOptions, BlockchainType, L2Config, fork_choice::apply_fork_choice,
 };
 use ethrex_common::{
     Address, U256,
-    types::{BYTES_PER_BLOB, Block, blobs_bundle, bytes_from_blob, fee_config::FeeConfig},
+    types::{BYTES_PER_BLOB, Block, bytes_from_blob, fee_config::FeeConfig},
 };
 use ethrex_common::{types::BlobsBundle, utils::keccak};
 use ethrex_config::networks::Network;
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -418,13 +418,13 @@ impl Command {
                     .enumerate()
                 {
                     let batch_number = file_number as u64 + 1;
-                    let blob = std::fs::read(file.path())?;
+                    let raw_blob = std::fs::read(file.path())?;
 
-                    if blob.len() != BYTES_PER_BLOB {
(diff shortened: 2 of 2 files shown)


crates/common/types/blobs_bundle.rs
```diff
@@ -44,26 +44,35 @@ pub fn blob_from_bytes(bytes: Bytes) -> Result<Blob, BlobsBundleError> {
...
+        let needed = 1 + chunk.len();
+        if dst + needed > BYTES_PER_BLOB {
+            return Err(BlobsBundleError::BlobDataInvalidBytesLength);
+        }
+
+        buf[dst] = 0x00;
+        dst += 1;
+
+        buf[dst..dst + chunk.len()].copy_from_slice(chunk);
+        dst += chunk.len();
```
Oppen: For consistency, why not do:
```rust
let len = chunk.len();
```
like you do in `bytes_from_blob`?

Oppen:
Looks good in principle and I would like this as a cleanup alone. Do we have numbers to back the perf tag?

Also, clippy is complaining about a few unused imports, could you delete those?

GarmashAlex:
image
image

GarmashAlex:
Bench code: 

```
use bytes::Bytes;
use ethrex_common::types::{BYTES_PER_BLOB, SAFE_BYTES_PER_BLOB, blob_from_bytes, bytes_from_blob};
use std::time::Instant;

fn format_duration(nanos: u128) -> String {
    if nanos < 1_000 {
        format!("{} ns", nanos)
    } else if nanos < 1_000_000 {
        format!("{:.2} µs", nanos as f64 / 1_000.0)
    } else if nanos < 1_000_000_000 {
        format!("{:.2} ms", nanos as f64 / 1_000_000.0)
    } else {
        format!("{:.2} s", nanos as f64 / 1_000_000_000.0)
    }
}

fn format_throughput(bytes: usize, nanos: u128) -> String {
    let bytes_per_sec = (bytes as f64 / nanos as f64) * 1_000_000_000.0;
    if bytes_per_sec > 1_000_000.0 {
        format!("{:.2} MB/s", bytes_per_sec / 1_000_000.0)
    } else if bytes_per_sec > 1_000.0 {
        format!("{:.2} KB/s", bytes_per_sec / 1_000.0)
    } else {
        format!("{:.2} B/s", bytes_per_sec)
    }
}

fn bench_blob_from_bytes(size: usize, iterations: usize) {
    let data = vec![0x42u8; size];
    let bytes = Bytes::from(data);

    let start = Instant::now();
    for _ in 0..iterations {
        let _ = blob_from_bytes(bytes.clone()).unwrap();
    }
    let elapsed = start.elapsed();

    let avg_nanos = elapsed.as_nanos() / iterations as u128;
    let throughput = format_throughput(size, avg_nanos);

    println!(
        "  blob_from_bytes({:>6}): {} per op, {}",
        format!("{}KB", size / 1024),
        format_duration(avg_nanos),
        throughput
    );
}

fn bench_bytes_from_blob(iterations: usize) {
    let blob_data = vec![0x42u8; BYTES_PER_BLOB];
    let bytes = Bytes::from(blob_data);

    let start = Instant::now();
    for _ in 0..iterations {
        let _ = bytes_from_blob(bytes.clone());
    }
    let elapsed = start.elapsed();

    let avg_nanos = elapsed.as_nanos() / iterations as u128;
    let throughput = format_throughput(SAFE_BYTES_PER_BLOB, avg_nanos);

    println!(
        "  bytes_from_blob(128KB): {} per op, {}",
        format_duration(avg_nanos),
        throughput
    );
}

fn bench_round_trip(size: usize, iterations: usize) {
    let data = vec![0x42u8; size];
    let bytes = Bytes::from(data);

    let start = Instant::now();
    for _ in 0..iterations {
        let blob = blob_from_bytes(bytes.clone()).unwrap();
        let blob_bytes = Bytes::from(blob.to_vec());
        let _ = bytes_from_blob(blob_bytes);
    }
    let elapsed = start.elapsed();

    let avg_nanos = elapsed.as_nanos() / iterations as u128;
    let total_bytes = size * 2; // encode + decode
    let throughput = format_throughput(total_bytes, avg_nanos);

    println!(
        "  round_trip({:>6}):     {} per op, {}",
        format!("{}KB", size / 1024),
        format_duration(avg_nanos),
        throughput
    );
}

fn main() {
    println!("\n=== Blob Utilities Benchmark ===\n");

    let iterations = 1000;

    println!("blob_from_bytes (encode):");
    bench_blob_from_bytes(1024, iterations);
    bench_blob_from_bytes(10 * 1024, iterations);
    bench_blob_from_bytes(50 * 1024, iterations);
    bench_blob_from_bytes(SAFE_BYTES_PER_BLOB, iterations);

    println!("\nbytes_from_blob (decode):");
    bench_bytes_from_blob(iterations);

    println!("\nRound trip (encode + decode):");
    bench_round_trip(10 * 1024, iterations);
    bench_round_trip(50 * 1024, iterations);
    bench_round_trip(SAFE_BYTES_PER_BLOB, iterations);

    println!("\n=== Benchmark Complete ===\n");
}
```

GarmashAlex:
> Looks good in principle and I would like this as a cleanup alone. Do we have numbers to back the perf tag?
> 
> Also, clippy is complaining about a few unused imports, could you delete those?

All is ready

azteca1998 (approved):

crates/common/types/blobs_bundle.rs
```diff
@@ -44,26 +44,36 @@ pub fn blob_from_bytes(bytes: Bytes) -> Result<Blob, BlobsBundleError> {
...
-    buf.copy_from_slice(
-        &blob
-            .chunks(32)
-            .map(|x| x[1..].to_vec())
-            .collect::<Vec<_>>()
-            .concat(),
-    );
+    let mut dst = 0usize;
+
+    for chunk in blob.chunks(32) {
```
ElFantasma, line 70: Subtle behavior change: the old code panicked (via `copy_from_slice` length mismatch) if `blob.len()` wasn't exactly `BYTES_PER_BLOB`, because the concatenated output of `chunks(32).map(|x| x[1..].to_vec())` wouldn't match `SAFE_BYTES_PER_BLOB`. The new loop silently returns a partially-filled zero buffer for short blobs, since `dst` just won't reach the end.

This is fine in practice (callers always pass `BYTES_PER_BLOB`-sized input), but if you want to preserve the old invariant, a debug assert at the end would make the contract explicit:
```rust
debug_assert_eq!(dst, SAFE_BYTES_PER_BLOB, "blob size mismatch");
```

crates/common/types/blobs_bundle.rs
```diff
@@ -44,26 +44,36 @@ pub fn blob_from_bytes(bytes: Bytes) -> Result<Blob, BlobsBundleError> {
...
-            .map(|x| [&[0x00], x].concat())
-            .collect::<Vec<_>>()
-            .concat(),
-    );
+    let mut dst = 0usize;
+
+    for chunk in bytes.chunks(31) {
+        let len = chunk.len();
+        let needed = 1 + len;
+        if dst + needed > BYTES_PER_BLOB {
```
ElFantasma, line 52: Good addition — this bounds check prevents writing past `buf` if the input exceeds `SAFE_BYTES_PER_BLOB`. The existing length check at the top (line 42-43) should already guarantee this, but belt-and-suspenders on unsafe-free buffer writes is reasonable.

ElFantasma (changes requested):
The `blob_from_bytes` / `bytes_from_blob` rewrites to single-pass loops are good — they eliminate the intermediate `Vec` allocations from `collect/concat`. However, the `raw_blob.clone()` on line 425 of `command.rs` still copies ~131KB, which offsets some of the savings in the reconstruct path. The `bytes_from_blob` rewrite also subtly changes panic-on-wrong-size behavior to silent zero-fill (see inline).

**cmd/ethrex/l2/command.rs:425** — `raw_blob.clone().into()` copies the entire ~131KB blob. The original code moved `blob` into `bytes_from_blob(blob.into())` (zero-copy move into `Bytes`), then later did `Bytes::copy_from_slice(&blob)` — one copy either way. To truly eliminate the extra allocation, `bytes_from_blob` could accept `&[u8]` instead of `Bytes`, since it only reads via `.chunks(32)` which is available on slices. That would let you write `bytes_from_blob(&raw_blob)` with no clone or conversion. This would be a small signature change but all call sites already have a contiguous byte slice or `Bytes` (which derefs to `&[u8]`).

PR #5835, opened 2026-01-14
perf(levm): improve sstore by adding a (address, key) → (current, original)  cache 

edg-l:
**Motivation**
Tested using https://github.com/lambdaclass/ethrex/pull/5821 seems to improve perf by up to 40%

```
sstore_clear                         442.63         267.89      -39.48%
sstore_noop                          444.24         258.66      -41.77%
sstore_set                           440.40         255.32      -42.03%
sstore_toggle                         85.07          58.24      -31.54%
```

Closes https://github.com/lambdaclass/ethrex/pull/5835


edg-l:
Ran on mainnet for some time, it looks like it has a regresion of about 6-15ms, depending on the block, some isolated blocks are faster but overall its slower

PR #5836, opened 2026-01-14
perf(l1): avoid unnecesary clones in engine_newPayloadV3/V4

edg-l:
- Avoid cloning the entire ExecutionPayload when building a Block
- Avoid cloning expected_blob_versioned_hashes


ilitteri (approved):

edg-l:
Somehow seems to perform worse on mainnet

PR #5839, opened 2026-01-14
perf(l1): reduce cloning in hooks

### LINKED ISSUE #5838: Hook cloning per opcode
Both `prepare_execution` and `finalize_execution`, both of which execute once per transaction, clone every hook just to borrow_mut:

```rust
for hook in self.hooks.clone() {
    hook.borrow_mut().prepare_execution(self)?;
}
```

we should avoid cloning and reference them instead.

iovoid:
**Motivation**

Currently in every transaction we clone the list of hooks.

**Description**

This PR reorders the type to avoid cloning the vector.

Closes #5838

### DIFF
--- a/crates/vm/levm/src/hooks/l2_hook.rs
+++ b/crates/vm/levm/src/hooks/l2_hook.rs
@@ -1,3 +1,5 @@
+use std::{cell::RefCell, rc::Rc};
+
 use crate::{
     constants::POST_OSAKA_GAS_LIMIT_CAP,
     db::gen_db::GeneralizedDatabase,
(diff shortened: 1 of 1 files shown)


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -665,7 +665,7 @@ fn simulate_common_bridge_call(
         LevmCallTracer::disabled(),
         VMType::L2(Default::default()),
     )?;
-    new_vm.hooks = vec![];
+    new_vm.hooks = Default::default();
```
ilitteri: Nit.
```suggestion
    new_vm.hooks = Vec::default();
```

iovoid: The hooks are no longer a Vec so it can't be done that way, but in 45764e4ff I made it more explicit `Rc::new(RefCell::new(Vec::new()))`

iovoid:
Had no measurable improvement.

PR #5844, opened 2026-01-14
chore(l1, l2): export block number and execution latency together

jrchatruc:
**Motivation**

**Description**

**Checklist**

### DIFF
--- a/crates/blockchain/metrics/rpc.rs
+++ b/crates/blockchain/metrics/rpc.rs
@@ -1,4 +1,7 @@
-use prometheus::{CounterVec, HistogramVec, register_counter_vec, register_histogram_vec};
+use prometheus::{
+    CounterVec, Gauge, HistogramVec, IntGauge, register_counter_vec, register_gauge,
+    register_histogram_vec, register_int_gauge,
+};
 use std::{future::Future, sync::LazyLock};
 
 pub static METRICS_RPC_REQUEST_OUTCOMES: LazyLock<CounterVec> =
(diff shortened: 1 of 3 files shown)


ElFantasma (commented):
Nice addition for correlating block numbers with latency! A few observations below.

crates/blockchain/metrics/rpc.rs
```diff
@@ -83,3 +108,10 @@ where
     timer.observe_duration();
     output
 }
+
+/// Records the block number and latency for a newPayloadV4 request.
+/// Both metrics are updated atomically to allow correlation in Grafana.
```
ElFantasma, line 113: Nit: "atomically" is a bit misleading here since these are two separate `.set()` calls. In practice it's unlikely a scrape happens between them, but the comment might confuse future readers expecting actual atomic semantics.

Maybe just say "updated together" or add a note that they're best-effort correlated?

crates/networking/rpc/engine/payload.rs
```diff
@@ -186,6 +189,32 @@ impl RpcHandler for NewPayloadV4Request {
         })
     }
 
+    /// Custom call implementation that tracks block number and latency metrics.
+    /// This overrides the default to capture the full request duration including parsing.
+    async fn call(req: &RpcRequest, context: RpcApiContext) -> Result<Value, RpcErr> {
+        let start_time = Instant::now();
```
ElFantasma, line 195: Minor: starting the timer before `Self::parse()` means parse time is included in the latency. This is fine if intentional, but differs from the default `call()` which only times `handle()`. Just flagging in case you want consistency with other endpoints.

ElFantasma: This may be confusing when comparing with other handlers

crates/networking/rpc/engine/payload.rs
```diff
@@ -186,6 +189,32 @@ impl RpcHandler for NewPayloadV4Request {
         })
     }
 
+    /// Custom call implementation that tracks block number and latency metrics.
+    /// This overrides the default to capture the full request duration including parsing.
+    async fn call(req: &RpcRequest, context: RpcApiContext) -> Result<Value, RpcErr> {
```
ElFantasma, line 194: This custom `call()` implementation doesn't call `record_async_duration()`, which the default implementation uses to populate the `rpc_duration_seconds` histogram. This means `engine_newPayloadV4` won't appear in the existing RPC duration metrics.

You could wrap the `handle()` call with `record_async_duration()` to preserve the histogram, or document that this endpoint intentionally uses only the new gauge-based metrics.

crates/blockchain/metrics/rpc.rs
ElFantasma: nit: "Both metrics are updated atomically" is misleading — two separate `.set()` calls on different Prometheus gauges are not atomic. A Prometheus scrape that lands between the two `.set()` calls will observe a mismatched (block_number, latency) pair. With a 15s scrape interval this is unlikely to matter in practice, but the comment overpromises. Suggest: "Both metrics are updated together" or just remove the atomicity claim.

PR #5852, opened 2026-01-14
chore(l1): change containerized `ethereum-metrics-exporter` port and avoid unneeded metrics setup

rodrigo-o:
**Motivation**

Avoid port collision with our current server setup

**Description**

This PR just set a different port for ethereum-metrics-exporter. Also removes the metrics setup by default in the server-sync make target


metrics/docker-compose-metrics-l1.overrides.yaml
```diff
@@ -9,7 +9,7 @@ services:
       - --consensus-url=http://host.docker.internal:5052
       - --metrics-port=9093
     ports:
-      - "9093:9093"
+      - "9094:9093"
```
rodrigo-o, line 12: No other changes are needed because other services reference the container directly with the internal port, not the external one, so they still hit the internal 9093.

samoht9277: Nice!

PR #5856, opened 2026-01-15
perf(levm): improve call and related opcodes

edg-l:
**Motivation**

- Skip empty account checks when CALL value is zero to avoid DB reads.
- Avoid copying calldata on early revert CALLs when tracing is off.
- Reduce precompile output clones when tracing is off.

Closes #issue_number

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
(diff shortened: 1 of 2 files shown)


edg-l:
Tested on mainnet, perf remains the same

PR #5867, opened 2026-01-15
feat(trie): implement Erigon-style grid-based Patricia trie

diegokingston:
Implement a grid-based trie pipelining algorithm inspired by Erigon's HexPatriciaHashed for more efficient state root computation.

Key changes:
- Add grid/ module with HexPatriciaGrid and ConcurrentPatriciaGrid
- Grid structure: 128 rows (depth) x 16 columns (nibbles)
- Iterative fold/unfold operations instead of recursive traversal
- 16-way parallelism support via rayon (ConcurrentPatriciaGrid)
- Comprehensive equivalence tests against recursive trie
- Criterion benchmarks for performance comparison

The grid trie processes keys in sorted order using:
- fold(): reduce depth, compute hashes, propagate upward
- unfold(): expand depth, load from DB or derive from parent

Currently, this is 2.5x slower than the default
Remaining Work

  - Integrate with blockchain.rs - use grid trie for large state updates
  - ConcurrentPatriciaGrid optimization - parallelize across 16 first-nibble shards
  - Performance tuning - optimize hot paths based on benchmark data

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -80,6 +80,9 @@ use ethrex_storage::{
 };
 use ethrex_trie::node::{BranchNode, ExtensionNode};
 use ethrex_trie::{Nibbles, Node, NodeRef, Trie};
+
+#[cfg(feature = "grid-trie")]
+use ethrex_trie::grid::ConcurrentPatriciaGrid;
 use ethrex_vm::backends::levm::db::DatabaseLogger;
 use ethrex_vm::{BlockExecutionResult, DynVmDatabase, Evm, EvmError};
 use mempool::Mempool;
@@ -672,6 +675,16 @@ impl Blockchain {
             .collect();
         let code_updates = code_updates.into_iter().collect();
 
+        // Validate state root using grid trie if enabled
+        #[cfg(feature = "grid-trie")]
+        {
+            let encoded_states: FxHashMap<H256, Vec<u8>> = account_states
+                .iter()
+                .map(|(k, v)| (*k, v.encode_to_vec()))
+                .collect();
+            Self::validate_state_root_with_grid(&encoded_states, state_trie_hash);
+        }
+
         let accumulated_updates = accumulator.map(|acc| acc.into_values().collect());
 
         Ok((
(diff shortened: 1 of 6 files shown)


crates/common/trie/grid/hex_patricia_grid.rs
```diff
@@ -0,0 +1,831 @@
...
+    ///               Empty value means deletion.
+    ///
+    /// # Returns
+    /// The computed state root hash.
+    pub fn apply_sorted_updates<I>(&mut self, updates: I) -> Result<H256, TrieError>
+    where
+        I: Iterator<Item = (H256, Vec<u8>)>,
+    {
+        let mut prev_key: Option<H256> = None;
+        let debug = std::env::var("GRID_DEBUG").is_ok();
```
ElFantasma, line 142: `std::env::var("GRID_DEBUG")` is called on **every key insertion** and allocates a `String` + locks the env mutex each time. In a loop processing thousands of keys, this is significant overhead even when the env var is unset.

Suggestion: read it once before the loop:
```rust
let debug = std::env::var("GRID_DEBUG").is_ok();
// ... move this line before the `for` loop, remove the one inside `fold()` and pass it as a parameter or store it in `self`.
```
Same issue in `fold()` at line 424.

crates/common/trie/grid/hex_patricia_grid.rs
```diff
@@ -0,0 +1,831 @@
...
+    /// * `updates` - Iterator of (hashed_key, value) pairs in sorted order.
+    ///               Empty value means deletion.
+    ///
+    /// # Returns
+    /// The computed state root hash.
+    pub fn apply_sorted_updates<I>(&mut self, updates: I) -> Result<H256, TrieError>
+    where
+        I: Iterator<Item = (H256, Vec<u8>)>,
+    {
+        let mut prev_key: Option<H256> = None;
```
ElFantasma, line 141: `prev_key` is maintained on every iteration but only consumed inside `#[cfg(debug_assertions)]` (line 146) and for `common_prefix_len` (line 160). The `common_prefix_len` usage is needed in release, so `prev_key` itself is fine — but consider gating the debug assertion under a feature or removing it entirely since sorted-order violations will produce wrong results (not UB) and the proptest suite covers this invariant.

crates/common/trie/grid/concurrent.rs
```diff
@@ -0,0 +1,396 @@
...
+        let results: Vec<Result<Option<H256>, TrieError>> = partitions
+            .into_par_iter()
+            .enumerate()
+            .map(|(nibble, partition)| {
+                if partition.is_empty() {
+                    return Ok(None);
+                }
+
+                // Create grid for this shard
+                let mut grid = HexPatriciaGrid::new(self.db.clone());
```
ElFantasma, line 72: Because `InMemoryTrieDB::clone()` is a shallow clone (shares the underlying `Arc<Mutex<BTreeMap>>`), all 16 shards share the **same** backing store. Currently the grid only reads from DB during `unfold_branch`, and since each shard starts with a fresh empty DB this is fine — but this is a subtle invariant. If a future change writes intermediate nodes to the DB during processing, shards would see each other's data.

Worth adding a comment here that the clone shares state, or using `InMemoryTrieDB::new_empty()` per shard if independence is intended.

crates/common/trie/grid/concurrent.rs
```diff
@@ -0,0 +1,396 @@
...
+
+        // Multiple children - create branch node
+        let branch = BranchNode::new(choices);
+        Ok(branch.compute_hash().finalize())
+    }
+}
+
+/// Partition updates by first nibble for parallel processing.
+#[allow(dead_code)]
+pub struct UpdatesByNibble {
```
ElFantasma, line 169: nit: `UpdatesByNibble` and its `impl` block are entirely `#[allow(dead_code)]`. If this isn't used anywhere, remove it to avoid carrying dead code. If it's planned for future use, consider leaving it out until needed.

crates/common/trie/trie.rs
```diff
@@ -11,6 +11,9 @@ mod test_utils;
 mod trie_iter;
 pub mod trie_sorted;
 mod verify_range;
+
+// Grid-based trie for efficient state root computation
+pub mod grid;
```
ElFantasma, line 16: The `grid` module is compiled unconditionally (not behind `#[cfg(feature = "grid-trie")]`), which means all builds — including production L1/L2 — compile `bitmap.rs`, `cell.rs`, `hex_patricia_grid.rs`, and the equivalence tests. The PR description says the feature is gated behind `grid-trie`, but only the `concurrent` sub-module is actually gated.

Consider wrapping this with `#[cfg(feature = "grid-trie")]` so the grid code only compiles when opted in, or document why it's intentionally always available.

crates/common/trie/grid/equivalence_tests.rs
```diff
@@ -0,0 +1,608 @@
...
+        ];
+
+        // Note: Empty value in recursive trie might behave differently
+        // This test documents the behavior
+        let recursive_root = compute_recursive_root(&data);
+        let grid_root = compute_grid_root(&data);
+
+        // For now, we just ensure no panic
+        // The roots may differ if empty value handling differs
+        let _ = (recursive_root, grid_root);
```
ElFantasma, line 258: This test discards the results without asserting anything (`let _ = ...`). It just checks for no panics, but doesn't verify that the grid and recursive roots match for empty values. Either assert equality or add a comment explaining why they're expected to differ and what the test is actually verifying.

crates/blockchain/blockchain.rs
```diff
@@ -672,6 +675,16 @@ impl Blockchain {
...
         let code_updates = code_updates.into_iter().collect();
 
+        // Validate state root using grid trie if enabled
+        #[cfg(feature = "grid-trie")]
+        {
+            let encoded_states: FxHashMap<H256, Vec<u8>> = account_states
+                .iter()
+                .map(|(k, v)| (*k, v.encode_to_vec()))
+                .collect();
+            Self::validate_state_root_with_grid(&encoded_states, state_trie_hash);
```
ElFantasma, line 685: This validation call is fire-and-forget: if the grid hash mismatches, it only logs at `debug!` level and silently continues. In a debug/validation feature, you might want this to be louder — e.g., `warn!` for mismatches so operators notice it in production logs, or `assert_eq!` in debug builds so CI catches regressions.

Also, note that this recomputes the **entire** state trie from scratch using a fresh `InMemoryTrieDB` on every block. For blocks with many accounts this could be very expensive, even behind a feature + env var gate.

PR #5875, opened 2026-01-16
fix(l1,l2): crypto default features false

gianbelinche:
**Motivation**
For our lambda vm, we need the guest program to not include kzg-rs, for that we need ethrex-crypto to have default features false by default.

**Description**
Makes ethrex-crypto use default features false by default.

**Checklist**


gianbelinche:
LambdaVM now supports c-kzg, so this pr is no longer necessary

PR #5879, opened 2026-01-16
docs: perf docs

Arkenan:
Create documentation to track the perf experiments


Arkenan:
Some wrong commit was included, closing.

PR #5882, opened 2026-01-16
fix(l2): balance diff parsing

avilagaston9:
**Motivation**

ERC20 cross-chain transfers were silently lost in `get_balance_diffs`, causing L1 deposits accounting to never update.

**Description**

When processing ERC20 messages, the code looked for matching token entries but never added new ones when the vector was empty. This caused `value_per_token` to always be empty, so L1 `publishL2Messages` never transferred tokens between bridges.

Fixed by pushing new entries when no match is found. Added test assertion to verify L1 deposits accounting.

**Checklist**


ilitteri:
Consider using `iter_mut().find()` to make the "find or insert" intent more explicit:

```rust
if let Some(value_per_token_decoded) = value_per_token_decoded {
    if let Some(existing) = entry.value_per_token.iter_mut().find(|v| {
        v.token_l1 == value_per_token_decoded.token_l1
            && v.token_src_l2 == value_per_token_decoded.token_src_l2
            && v.token_dst_l2 == value_per_token_decoded.token_dst_l2
    }) {
        existing.value += value_per_token_decoded.value;
    } else {
        entry.value_per_token.push(value_per_token_decoded);
    }
}
```

This eliminates the mutable `found` flag and manual `break`. Feel free to ignore if you prefer the current approach - both are correct.

avilagaston9:
> Consider using iter_mut().find() to make the "find or insert" intent more explicit:
> @ilitteri 
Done [here](https://github.com/lambdaclass/ethrex/pull/5882/commits/fa8e073cae92662338ead268eb44845a7faa55b0)!

PR #5887, opened 2026-01-18
refactor(l2): move monitor to tooling workspace

ilitteri:
## Summary
- Move the L2 monitor code from `crates/l2/monitor` to `tooling/monitor` to reduce main codebase line count
- Create `ethrex-monitor` crate with `SequencerStatusProvider` trait for decoupling from L2 internals
- L2 crate re-exports the monitor for backwards compatibility

## Line Count Impact
| Location | Before | After | Change |
|----------|--------|-------|--------|
| `crates/` | 84,388 | 82,170 | **-2,218** |
| `tooling/monitor/` | - | 2,278 | +2,278 |

## Test plan
- [x] `cargo check -p ethrex-monitor` passes
- [x] `cargo check -p ethrex-l2` passes


crates/l2/Cargo.toml
```diff
@@ -27,6 +27,7 @@ ethrex-vm.workspace = true
 ethrex-levm.workspace = true
 ethrex-l2-common.workspace = true
 ethrex-l2-rpc.workspace = true
+ethrex-monitor = { path = "../../tooling/monitor" }
```
tomip01: Can we add this in the workspace?

tooling/monitor/src/config.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// Osaka activation time
+    pub osaka_activation_time: Option<u64>,
+}
+
+/// Trait for providing sequencer status to the monitor.
+/// This allows the monitor to be decoupled from the sequencer state implementation.
+pub trait SequencerStatusProvider: Clone + Send + Sync + 'static {
+    /// Get the current status of the sequencer
+    fn status(&self) -> impl std::future::Future<Output = SequencerStatus> + Send;
+}
```
tomip01: Can we move `SequencerState` to `ethrex-l2-common` to avoid creating a trait that is only implemented in one place?

crates/l2/common/src/sequencer_state.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl SequencerState {
+    pub async fn status(&self) -> SequencerStatus {
+        *self.0.lock().await
+    }
+
+    pub async fn new_status(&self, status: SequencerStatus) {
+        *self.0.lock().await = status;
+    }
+}
```
tomip01, line 16: This is for a future PR but I had the idea while reading this, we could use a `RWLock` since we are more times reading the state than changing it

crates/l2/sequencer/sequencer_state.rs
```diff
@@ -1,52 +1,2 @@
...
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            SequencerStatus::Sequencing => write!(f, "Sequencing"),
-            SequencerStatus::Syncing => write!(f, "Syncing"),
-            SequencerStatus::Following => write!(f, "Following"),
-        }
-    }
-}
+// Re-export from ethrex-l2-common
+pub use ethrex_l2_common::sequencer_state::{SequencerState, SequencerStatus};
```
tomip01: I think we can delete it and re-export it in the `crates/l2/sequencer/mod.rs`

crates/l2/common/Cargo.toml
```diff
@@ -27,6 +27,7 @@ serde_with.workspace = true
 rkyv.workspace = true
 k256.workspace = true
 tracing.workspace = true
+tokio.workspace = true
```
ElFantasma: Adding `tokio` here makes it a dependency of every crate that depends on `ethrex-l2-common`, including the guest programs (SP1, RISC Zero, OpenVM, Zisk) — the four guest `Cargo.lock` updates in this PR confirm it. `SequencerState` uses `tokio::sync::Mutex`, which is why tokio is needed, but this type is a runtime-only concept that doesn't belong in a crate shared with zkVM guests.

Options:
1. Keep `SequencerState` in `crates/l2/sequencer/` and have `ethrex-monitor` depend on `ethrex-l2` directly (or accept `SequencerState` as a type parameter/trait)
2. Gate the `sequencer_state` module behind a feature flag (e.g., `tokio`) that guest programs don't enable
3. Use `std::sync::Mutex` instead of `tokio::sync::Mutex` — `SequencerStatus` is `Copy` so lock hold times are trivially short

crates/l2/sequencer/mod.rs
```diff
@@ -206,11 +206,30 @@ pub async fn start_l2(
     }
 
     if cfg.monitor.enabled {
+        let monitor_cfg = ExternalMonitorConfig {
```
ElFantasma: nit: This config construction block duplicates field extraction that could live in a `From<&SequencerConfig>` impl on `MonitorConfig`. Not blocking, but if `SequencerConfig` changes, this manual mapping is easy to get out of sync.

PR #5904, opened 2026-01-19
feat(l1): add --p2p.bind-addr to separate bind and advertised addresses

catwith1hat:
**Motivation**

ethrex uses the local-ip-address crate to find the IP address to advertise on the p2p layer. Inside Docker or in a NAT environment, none of the interfaces have the correct IP address, so using netlink calls won't find the correct IP.

While `--p2p.addr` can be used to set the IP address explicitly, setting that option also causes a `bind` call to that IP. This `bind` call panics because as said before there is no interface inside a Docker or NAT-ed environment with this IP address. 

**Description**

Introduced --p2p.bind-addr (defaulting to 0.0.0.0) to allow specifying the local interface to bind to, while keeping --p2p.addr for the advertised address in the Node Record (ENR). This enables running the node in Docker environments where the bind and advertised addresses may differ.

**Checklist**

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -218,10 +218,18 @@ pub struct Options {
     #[arg(
         long = "p2p.addr",
         value_name = "ADDRESS",
-        help = "Listening address for the P2P protocol.",
+        help = "Advertised address for the P2P protocol (Node Record). Defaults to local IP if not set.",
         help_heading = "P2P options"
     )]
     pub p2p_addr: Option<String>,
+    #[arg(
+        long = "p2p.bind-addr",
+        default_value = "0.0.0.0",
+        value_name = "ADDRESS",
+        help = "Bind address for the P2P protocol.",
+        help_heading = "P2P options"
+    )]
+    pub p2p_bind_addr: String,
     #[arg(
         long = "p2p.port",
         default_value = "30303",
@@ -308,6 +316,7 @@ impl Options {
             metrics_port: "9090".to_string(),
             authrpc_addr: "localhost".to_string(),
             authrpc_jwtsecret: "jwt.hex".to_string(),
+            p2p_bind_addr: "0.0.0.0".to_owned(),
             p2p_port: "30303".into(),
             discovery_port: "30303".into(),
             mempool_max_size: 10_000,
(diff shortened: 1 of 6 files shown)


MegaRedHand:
Commits have to be signed. Check https://github.com/lambdaclass/ethrex/blob/main/CONTRIBUTING.md#commit-signature-verification for more info

catwith1hat:
I fixed the tests as far as I could, but the 3 remaining failing ones seems to come from me not being a member of the right organization. I also signed my commit.

crates/networking/p2p/network.rs
```diff
@@ -128,7 +132,7 @@ pub async fn start_network(context: P2PContext, bootnodes: Vec<Node>) -> Result<
 }
 
 pub(crate) async fn serve_p2p_requests(context: P2PContext) {
-    let tcp_addr = context.local_node.tcp_addr();
+    let tcp_addr = SocketAddr::new(context.bind_ip, context.local_node.tcp_port);
```
ElFantasma, line 135: nit: After this change, the self-connection check on line 149 (`if tcp_addr == peer_addr`) compares the **bind** address (e.g. `0.0.0.0:30303`) against the incoming peer's address. Previously it compared the **advertised** address. Neither actually works reliably — `peer_addr` from `accept()` includes the remote's ephemeral source port, so it would never equal `local:30303` — but worth noting this is a pre-existing issue, not introduced by this PR.

cmd/ethrex/cli.rs
```diff
@@ -218,10 +218,18 @@ pub struct Options {
     #[arg(
         long = "p2p.addr",
         value_name = "ADDRESS",
-        help = "Listening address for the P2P protocol.",
+        help = "Advertised address for the P2P protocol (Node Record). Defaults to local IP if not set.",
         help_heading = "P2P options"
     )]
     pub p2p_addr: Option<String>,
+    #[arg(
```
ilitteri, line 225: The semantic change on `--p2p.addr` is silent — existing deployments using it to bind to a specific interface will now bind to `0.0.0.0` without any warning.

Safer default: when `--p2p.addr` is set but `--p2p.bind-addr` isn't explicitly passed, default `bind-addr` to the `--p2p.addr` value. That preserves current behavior and only splits them when the user explicitly asks for it.

Arkenan:
Hello ser, thanks a lot for the contribution. We saw this PR became inactive and we had to make progress with ipv6 so we started a PR for this issue ourselves. For this reason, I'll be closing this PR, but feel free to pick another "good first issue" and we'll make sure to review it!

catwith1hat:
How is my PR related to IPv6? What's the other PR? Thanks!

PR #5909, opened 2026-01-19
feat(l1): rate limit WHOAREYOU packets per source IP to prevent amplification attacks

### LINKED ISSUE #5849: Rate limit WHOAREYOU responses in discv5
Per the [discv5 spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#whoareyou):

  > Implementations should limit the number of WHOAREYOU packets they send per source IP to prevent amplification attacks.

  Currently, we respond to every unknown packet with WHOAREYOU without rate limiting. An attacker could:
  1. Spoof source IP of a victim
  2. Send many packets to our node
  3. We send WHOAREYOU (larger response) to victim's IP

  Should implement per-IP rate limiting (e.g., max 1 WHOAREYOU per IP per second).

ElFantasma:
**Motivation**

  Per the discv5 spec, implementations should limit `WHOAREYOU` packets per source IP to prevent
  amplification attacks. An attacker can spoof the source IP of a victim, send many small
  packets to our node, and we respond with larger `WHOAREYOU` packets to the victim's
  IP—amplifying the attacker's traffic.

  **Description**

  - Adds a `whoareyou_rate_limit` map to track last `WHOAREYOU` send time per source IP
  - Rate limits `WHOAREYOU` responses to 1 per second per IP address
  - Stale rate limit entries are automatically cleaned up during periodic maintenance
  - Adds unit test `test_whoareyou_rate_limiting` to verify the behavior

  Closes #5849


crates/networking/p2p/discv5/server.rs
```diff
@@ -602,6 +608,23 @@ impl DiscoveryServer {
...
+        let now = Instant::now();
+
+        if let Some(last_sent) = self.whoareyou_rate_limit.get(&ip)
+            && now.duration_since(*last_sent) < WHOAREYOU_RATE_LIMIT
+        {
+            trace!(
+                to_ip = %ip,
+                "Rate limiting WHOAREYOU packet (amplification attack prevention)"
+            );
+            return Ok(());
```
Oppen, line 636: Should we also reduce its score? So misbehaving peers are penalized.

crates/networking/p2p/discv5/server.rs
```diff
@@ -682,11 +705,17 @@ impl DiscoveryServer {
             });
         let removed_challenges = before_challenges - self.pending_challenges.len();
 
-        let total_removed = removed_messages + removed_challenges;
+        // Clean stale WHOAREYOU rate limit entries
+        let before_rate_limits = self.whoareyou_rate_limit.len();
+        self.whoareyou_rate_limit
+            .retain(|_ip, timestamp| now.duration_since(*timestamp) < WHOAREYOU_RATE_LIMIT);
```
Oppen, line 725: This might be a little expensive in bad cases. I think leaving a comment would be enough for now.
It may be a good idea to add how long it took to the trace message as well.

ElFantasma: Yes, we have several cases like this. I added #5972 to measure and deal with these potential performance issues.

crates/networking/p2p/discv5/server.rs
ElFantasma: Maybe... problem is that it's not necessary a misbehaving node that sends several packets each triggering WhoAreYou responses. The spec describes [some case](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#handshake-implementation-considerations)
So, we could add some scoring management here, but we should be able to distinguish between malicious behavior vs. standard behavior, which would complicate code for a not-so-important improvement.

ElFantasma:
> I see two potential problems that should be considered:
> 
>     * The whoareyou_rate_limit map is keyed by IpAddr and only cleaned every PRUNE_INTERVAL (5 seconds). During an amplification attack with many spoofed source IPs, this map can grow quickly.
> 
>     * A IPv6 host might have a /64 (or more) allocated, and an attacker would be able to both forward amplified traffic _and_ degrade the node by cycling the faked source ip.
> 
> 
> It might be a reasonable stopgag solution to stop accepting WHOAREYOU packets if the map is too large, considering this would only prevent new peers from connecting.

Good points! Both concerns are valid

The fix is to replace the `FxHashMap` with a bounded LRU cache, plus add a global rate limit on total WHOAREYOU packets/second to cap outgoing bandwidth.

Note: Even with these mitigations, during a sustained attack legitimate new peers may have trouble connecting - this is a fundamental limitation of connectionless UDP protocols where we can't distinguish legitimate peers from attackers before the handshake completes.

I think the complexity of the fix is enough to be tackled in a different PR: Created #6125 for this.

iovoid (approved):

iovoid (approved):

PR #5914, opened 2026-01-19
feat(l1): detect external IP via PONG recipient_addr voting

### LINKED ISSUE #5851: Detect external IP via PONG recipient_addr voting
Nodes behind NAT don't know their external IP. The `recipient_addr` field in PONG contains the IP:port as seen by the responder.

  To reliably detect external IP:
  1. Collect `recipient_addr` from multiple PONGs
  2. Use majority voting to determine real external IP
  3. Update local ENR when majority agrees

  Reference: [nim-eth implementation](https://github.com/status-im/nim-eth/blob/master/eth/p2p/discoveryv5/protocol.nim#L168)
  uses `ipMajorityInterval = 5.minutes`.

  **Note:** May overlap with #5493 "Periodically update local ENR".

ElFantasma:
**Motivation**

  Nodes behind NAT don't know their external IP address. When we send PINGs, peers respond with PONGs containing `recipient_addr` - the IP:port they see as our source. By collecting these from multiple peers and using majority voting, we can reliably detect our external IP and update our ENR.

  **Description**

  Implements external IP detection via PONG `recipient_addr` voting:

  - Uses voting rounds: first round ends after 3 votes (fast startup), subsequent rounds after
  5 minutes (to detect IP changes from NAT/network switches)
  - At round end, the IP with most votes wins if it has ≥3 agreeing votes
  - Each peer can only vote once per round (Sybil protection)
  - Private IPs (RFC 1918, loopback, link-local) are filtered to prevent local/external
  oscillation
  - Updates local ENR with new IP and incremented sequence number when external IP is detected

  Closes #5851

### DIFF
--- a/crates/networking/p2p/discv5/server.rs
+++ b/crates/networking/p2p/discv5/server.rs
@@ -360,6 +360,13 @@ impl DiscoveryServer {
             return Ok(());
         }
 
+        // Add the peer to the peer table
+        if let Some(record) = &authdata.record {
+            self.peer_table
+                .new_contact_records(vec![record.clone()], self.local_node.node_id())
+                .await?;
+        }
+
         // Derive session keys (we are the recipient, node B)
         let session = derive_session_keys(
             &self.signer,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2026-02-23
+
+- Add precompile result cache shared between warmer and executor threads [#6243](https://github.com/lambdaclass/ethrex/pull/6243)
+
 ### 2026-02-13
 
 - Optimize storage layer for block execution by reducing lock contention and allocations [#6207](https://github.com/lambdaclass/ethrex/pull/6207)
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -415,6 +415,8 @@ impl Blockchain {
                 let warm_handle = std::thread::Builder::new()
                     .name("block_executor_warmer".to_string())
                     .spawn_scoped(s, move || {
+                        // Warming uses the same caching store, sharing cached state with execution.
+                        // Precompile cache lives inside CachingDatabase, shared automatically.
                         let start = Instant::now();
                         if let Some(bal) = bal {
                             // Amsterdam+: BAL-based precise prefetching (no tx re-execution)
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -343,7 +343,6 @@ impl LEVM {
             |stack_pool, (sender, txs)| {
                 // Each sender group gets its own db instance for state propagation
                 let mut group_db = GeneralizedDatabase::new(store.clone());
-
                 // Execute transactions sequentially within sender group
                 // This ensures nonce and balance changes from tx[N] are visible to tx[N+1]
                 for tx in txs {
--- a/crates/vm/levm/Cargo.toml
+++ b/crates/vm/levm/Cargo.toml
@@ -42,13 +42,12 @@ ark-ec = "0.5.0"
 ark-ff = { version = "0.5.0", features = ["asm"] }
 strum = { version = "0.27.1", features = ["derive"] }
 k256.workspace = true
+rustc-hash.workspace = true
 substrate-bn = { version = "0.6.0", optional = true }
 secp256k1 = { workspace = true, optional = true }
 ziskos = { git = "https://github.com/0xPolygonHermez/zisk.git", tag = "v0.15.0", optional = true }
 
 bitvec = { version = "1.0.1", features = ["alloc"] }
-
-rustc-hash.workspace = true
 
 [dev-dependencies]
 hex.workspace = true
--- a/crates/vm/levm/src/db/mod.rs
+++ b/crates/vm/levm/src/db/mod.rs
@@ -1,4 +1,4 @@
-use crate::errors::DatabaseError;
+use crate::{errors::DatabaseError, precompiles::PrecompileCache};
 use ethrex_common::{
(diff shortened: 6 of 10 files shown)


Oppen:
What happens if an attacker with spoofed IPs and fake clients spams us with a different address? I'm not sure one vote per peer covers that, given from ethrex' point of view they look like many different ones.

EDIT: nevermind, we should reject unsolicited PONGs regardless, so they can't spam undetected (we need to send PINGs first).

crates/networking/p2p/discv5/server.rs
```diff
@@ -1075,4 +1208,299 @@ mod tests {
...
+        // Link-local should be ignored
+        let link_local: IpAddr = "169.254.1.1".parse().unwrap();
+        server.record_ip_vote(link_local, voter1);
+        assert!(server.ip_votes.is_empty());
+
+        // Public IP should be recorded
+        let public_ip: IpAddr = "203.0.113.50".parse().unwrap();
+        server.record_ip_vote(public_ip, voter1);
+        assert_eq!(server.ip_votes.get(&public_ip).map(|v| v.len()), Some(1));
+    }
```
Oppen, line 1586: This needs to be mover to the tests subtree now.

ElFantasma: Yes, created #5992 for these and other tests that are misplaced

crates/networking/p2p/discv5/server.rs
```diff
@@ -751,6 +777,99 @@ impl DiscoveryServer {
...
+
+    /// Updates local node IP and re-signs the ENR with incremented seq.
+    fn update_local_ip(&mut self, new_ip: IpAddr) {
+        // Update runtime node
+        self.local_node.ip = new_ip;
+
+        // Update ENR - increment seq and re-create with new IP
+        let new_seq = self.local_node_record.seq + 1;
+        if let Ok(mut new_record) = NodeRecord::from_node(&self.local_node, new_seq, &self.signer) {
```
iovoid: Why ignore the error?

crates/networking/p2p/discv5/server.rs
```diff
@@ -751,6 +777,99 @@ impl DiscoveryServer {
...
+        self.ip_votes.clear();
+        self.ip_vote_period_start = Some(Instant::now());
+        self.first_ip_vote_round_completed = true;
+    }
+
+    /// Returns true if the IP is private/local (not useful for external connectivity).
+    fn is_private_ip(ip: IpAddr) -> bool {
+        match ip {
+            IpAddr::V4(v4) => v4.is_private() || v4.is_loopback() || v4.is_link_local(),
+            IpAddr::V6(v6) => v6.is_loopback(),
```
iovoid: Does not cover local addresses like with IPv4. While [Ipv6Addr::is_global](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.is_global) is nightly, its documentation describes well the cases to handle.

crates/networking/p2p/discv5/server.rs
```diff
@@ -751,6 +777,99 @@ impl DiscoveryServer {
...
+        let new_seq = self.local_node_record.seq + 1;
+        if let Ok(mut new_record) = NodeRecord::from_node(&self.local_node, new_seq, &self.signer) {
+            // Preserve fork_id if present
+            if let Some(fork_id) = self.local_node_record.decode_pairs().eth {
+                let _ = new_record.set_fork_id(fork_id, &self.signer);
+            }
+            self.local_node_record = new_record;
+
+            // Clear IP votes after successful update
+            self.ip_votes.clear();
```
iovoid: The caller already clears ip_votes.

crates/networking/p2p/discv5/server.rs
```diff
@@ -742,6 +760,14 @@ impl DiscoveryServer {
...
+        if let Some(start) = self.ip_vote_period_start
+            && self.first_ip_vote_round_completed
+            && now.duration_since(start) >= IP_VOTE_WINDOW
+        {
+            self.finalize_ip_vote_round();
+        }
+
         let total_removed = removed_messages + removed_challenges + removed_rate_limits;
         if total_removed > 0 {
             trace!(
```
ilitteri, line 789: If fewer than 3 peers ever respond with PONGs, the first round never completes and `ip_votes` is never cleared — `cleanup_stale_entries` only checks timeout when `first_ip_vote_round_completed` is true.

```rust
if let Some(start) = self.ip_vote_period_start
    && now.duration_since(start) >= IP_VOTE_WINDOW
{
    self.finalize_ip_vote_round();
}
```

This finalizes on timeout regardless of which round we're in, so nodes with few peers don't accumulate stale votes indefinitely.

crates/networking/p2p/discv5/server.rs
ElFantasma: Fixed — if set_fork_id fails we now abort the IP update entirely, so we never publish an ENR without fork_id. Also moved the local_node.ip assignment after the ENR is fully built to avoid IP/ENR mismatch on failure.

crates/networking/p2p/discv5/server.rs
ElFantasma: Good catch. Added checks for unique local (fc00::/7), link-local (fe80::/10), and unspecified (::) IPv6 addresses. Also added IPv6 test cases.

crates/networking/p2p/discv5/server.rs
ElFantasma: Right, removed.

crates/networking/p2p/discv5/server.rs
ElFantasma: Good catch. Removed the first_ip_vote_round_completed guard so the timeout applies to all rounds — nodes with few peers will now finalize via timeout instead of accumulating stale votes.

PR #5917, opened 2026-01-19
feat(l1): reduce account range chunks from 800 to 16

ilitteri:
## Motivation

ethrex uses 800 chunks for account download, creating excessive task management overhead. Geth uses 16 concurrent tasks because:
- Larger chunks = fewer proof verifications per byte of data
- Larger chunks = fewer incomplete responses requiring re-requests
- Reduced coordination overhead

## Description

Change the account range chunk count from 800 to 16. Each chunk now covers 1/16th of the hash space instead of 1/800th, significantly reducing the overhead of task management while maintaining parallelism.

**Expected Impact:** 5-15% faster account download phase.

## How to Test

1. Run snap sync against Sepolia testnet
2. Monitor account download phase timing
3. Compare against baseline (800 chunks)

## Related Issues

Part of snap sync optimization effort.


JereSalo:
Has it been tested? Results?

PR #5919, opened 2026-01-19
feat(l1): add code existence cache to skip redundant bytecode downloads

ilitteri:
## Motivation

When resuming snap sync after a partial sync, bytecodes that were already downloaded and stored in the DB are collected again during healing and re-downloaded. This wastes bandwidth and time.

## Description

Add a HashSet cache that tracks which code hashes already exist in the database. Before queueing a bytecode for download:

1. Check if hash is in the `existing_code_cache` (fast cache lookup)
2. If not in cache, query the DB via `store.get_account_code(hash)`
3. If code exists, add to cache and skip download
4. If code doesn't exist, add to download queue

Also added logging to report how many bytecodes were skipped due to already existing.

**Expected Impact:** 3-8% faster bytecode download phase when resuming sync.

## How to Test

1. Start a snap sync and let it partially complete (download some bytecodes)
2. Stop the sync
3. Resume sync
4. Check logs for "Skipped N bytecodes that already existed in DB"
5. Verify bytecode download phase is faster than a fresh sync

## Related Issues

Part of snap sync optimization effort.

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -902,7 +902,10 @@ impl Syncer {
 
         let code_hashes_dir = get_code_hashes_snapshots_dir(&self.datadir);
         let mut seen_code_hashes = HashSet::new();
+        // Cache for code hashes that already exist in the DB to avoid redundant downloads
+        let mut existing_code_cache: HashSet<H256> = HashSet::new();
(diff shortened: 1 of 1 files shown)


iovoid:
This only brings an improvement when resuming a snap sync, which is currently not supported.

PR #5924, opened 2026-01-19
feat(l1): allow parallel DB write tasks during healing

ilitteri:
## Motivation

Increases throughput during state and storage healing by allowing multiple DB write tasks to run in parallel. The current implementation waits for each write to complete before starting the next batch's encoding, which wastes time that could be used for preparation.

## Description

Increases the maximum concurrent DB write tasks from 1 to 2 during healing phases. This allows the next batch to start encoding (CPU work) while the previous batch is being written to the DB (I/O work).

### Changes

- **`sync/state_healing.rs`**:
  - Added `MAX_PARALLEL_DB_WRITERS` constant (2)
  - Added `BATCH_WRITE_THRESHOLD` constant for clarity
  - Changed joinset check from `!is_empty()` to `len() >= MAX_PARALLEL_DB_WRITERS`
  - Pre-allocates replacement buffer (from PR 7)

- **`sync/storage_healing.rs`**:
  - Imports `MAX_PARALLEL_DB_WRITERS` from state_healing
  - Added `BATCH_WRITE_THRESHOLD` constant
  - Applied same parallel write pattern

### Expected Impact

3-8% faster healing due to overlapping encoding and I/O operations.

### Note

The parallelism is intentionally limited to 2 to:
- Avoid excessive memory usage from multiple pending batches
- Minimize risk of ordering issues
- Keep the change simple and testable

## Testing

- [x] `cargo check -p ethrex-p2p`
- [x] `cargo clippy -p ethrex-p2p`
- [x] `cargo test -p ethrex-p2p`

### DIFF
--- a/crates/networking/p2p/sync/state_healing.rs
+++ b/crates/networking/p2p/sync/state_healing.rs
@@ -35,6 +35,9 @@ pub const STORAGE_BATCH_SIZE: usize = 300;
 pub const NODE_BATCH_SIZE: usize = 500;
 /// Pace at which progress is shown via info tracing
 pub const SHOW_PROGRESS_INTERVAL_DURATION: Duration = Duration::from_secs(2);
+/// Maximum number of parallel DB write tasks
+/// Keep this small to avoid out-of-order writes causing issues
+pub const MAX_PARALLEL_DB_WRITERS: usize = 2;
 
 use super::SyncError;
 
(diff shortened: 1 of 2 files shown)


crates/networking/p2p/sync/state_healing.rs
```diff
@@ -35,6 +35,9 @@ pub const STORAGE_BATCH_SIZE: usize = 300;
 pub const NODE_BATCH_SIZE: usize = 500;
 /// Pace at which progress is shown via info tracing
 pub const SHOW_PROGRESS_INTERVAL_DURATION: Duration = Duration::from_secs(2);
+/// Maximum number of parallel DB write tasks
+/// Keep this small to avoid out-of-order writes causing issues
+pub const MAX_PARALLEL_DB_WRITERS: usize = 2;
```
azteca1998, line 40: If large numbers cause out-of-order writing issues, does that also happen with 2? Is it still ok to merge it if 2 parallel write tasks may cause that issue (even if the probability is small)?

crates/networking/p2p/sync/state_healing.rs
```diff
@@ -289,16 +292,23 @@ async fn heal_state_trie(
 
         let is_done = paths.is_empty() && nodes_to_heal.is_empty() && inflight_tasks == 0;
 
-        if nodes_to_write.len() > 100_000 || is_done || is_stale {
-            // PERF: reuse buffers?
-            let to_write = std::mem::take(&mut nodes_to_write);
+        // Pre-allocate batch write threshold
+        const BATCH_WRITE_THRESHOLD: usize = 100_000;
```
tomip01, line 296: I think it is better if we define it outside the function

crates/networking/p2p/sync/state_healing.rs
```diff
@@ -289,16 +292,23 @@ async fn heal_state_trie(
...
+
+        if nodes_to_write.len() > BATCH_WRITE_THRESHOLD || is_done || is_stale {
+            // Replace with a pre-allocated buffer to avoid repeated allocations
+            let to_write = std::mem::replace(
+                &mut nodes_to_write,
+                Vec::with_capacity(BATCH_WRITE_THRESHOLD),
+            );
             let store = store.clone();
-            // NOTE: we keep only a single task in the background to avoid out of order deletes
```
tomip01, line 296: Why is that that now er void Out of Order deletes?

crates/networking/p2p/sync/storage_healing.rs
```diff
@@ -206,11 +208,18 @@ pub async fn heal_storage_trie(
         let is_done = state.requests.is_empty() && state.download_queue.is_empty();
         let is_stale = current_unix_time() > state.staleness_timestamp;
 
-        if nodes_to_write.values().map(Vec::len).sum::<usize>() > 100_000 || is_done || is_stale {
+        // Threshold for batching DB writes
+        const BATCH_WRITE_THRESHOLD: usize = 100_000;
```
tomip01, line 212: same here

PR #5930, opened 2026-01-20
perf(levm): remove locking in LevmDatabase

Oppen:
**Motivation**

There was unnecessary contention caused by a mutex in a field that is only used
as immutable reference, with mutability already protected in lower layers.

**Description**

Just remove the indirection, store an `Arc` directly.

**Checklist**


pablodeymo (approved):

crates/vm/backends/levm/db.rs
```diff
@@ -17,11 +17,11 @@ pub struct DatabaseLogger {
     pub state_accessed: Arc<Mutex<HashMap<CoreAddress, Vec<CoreH256>>>>,
     pub code_accessed: Arc<Mutex<Vec<CoreH256>>>,
     // TODO: Refactor this
```
Oppen:
```suggestion
```

PR #5931, opened 2026-01-20
chore(l1): measure serialization and deserialization as part of RPC calls

rodrigo-o:
**Motivation**

Taking into account the whole RPC flow for measures

**Description**

We wanted to be sure we are correctly measuring RPC latency. Previously we just measure the handler time (which involved all the logic for processing requests) now we added the initial serialization, parameter parsing and the final response serialization.

### DIFF
--- a/.github/scripts/generate_block_time_report.sh
+++ b/.github/scripts/generate_block_time_report.sh
@@ -37,6 +37,53 @@ if [[ "$result_count" -eq 0 ]]; then
   exit 1
 fi
 
+# Query version information from eth_exe_web3_client_version metric
+VERSION_QUERY="eth_exe_web3_client_version"
+version_curl_args=(-sS -G "$QUERY_URL" --data-urlencode "query=$VERSION_QUERY")
+if [[ -n "${BLOCK_TIME_PROMETHEUS_BEARER_TOKEN:-}" ]]; then
+  version_curl_args+=(-H "Authorization: Bearer $BLOCK_TIME_PROMETHEUS_BEARER_TOKEN")
+fi
+if [[ -n "${BLOCK_TIME_PROMETHEUS_BASIC_AUTH:-}" ]]; then
+  version_curl_args+=(-u "$BLOCK_TIME_PROMETHEUS_BASIC_AUTH")
+fi
+
+version_response=$(curl "${version_curl_args[@]}")
+
+# Extract version for each client by instance pattern (bash 3.2 compatible)
+# Extracts only the version portion (e.g., "v1.2.3" from "Client/v1.2.3/platform")
+get_version_by_instance() {
+  local instance_pattern="$1"
+  jq -r --arg pattern "$instance_pattern" '
+    .data.result[] | select(.metric.instance | test($pattern)) | .metric.version // "unknown" | split("/")[1] // "unknown"
+  ' <<<"$version_response" 2>/dev/null | head -1
+}
+
+# Truncate version string if longer than max length
+truncate_version() {
+  local version="$1"
+  local max_len="${2:-24}"
+  if [[ ${#version} -gt $max_len ]]; then
+    echo "${version:0:$((max_len-3))}..."
+  else
+    echo "$version"
+  fi
+}
+
+version_ethrex=$(get_version_by_instance "^ethrex-mainnet-1:")
+version_reth=$(get_version_by_instance "^reth-mainnet-1:")
+version_geth=$(get_version_by_instance "^geth-mainnet-1:")
+version_nethermind=$(get_version_by_instance "^nethermind-mainnet-1:")
+: "${version_ethrex:=unknown}"
+: "${version_reth:=unknown}"
+: "${version_geth:=unknown}"
+: "${version_nethermind:=unknown}"
+
+# Truncated versions for Slack table display
+version_ethrex_short=$(truncate_version "$version_ethrex")
+version_reth_short=$(truncate_version "$version_reth")
+version_geth_short=$(truncate_version "$version_geth")
+version_nethermind_short=$(truncate_version "$version_nethermind")
+
 raw_series=()
 while IFS= read -r line; do
   raw_series+=("$line")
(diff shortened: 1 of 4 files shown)


rodrigo-o:
In a mac with hoodi, no difference could be seen across larger methods (newPayload) on quicker ones (exchange capabilities) the difference is seen. We still need to check on mainnet

| newPayload | exchangeCapabilities |
| ---- | ---- |
| image |  image |

ilitteri (commented):
## Review

The approach looks good - measuring full request duration including JSON parsing/serialization will give more accurate latency insights.

### Questions

1. **Scope**: Only `handle_authrpc_request` gets full duration tracking, but `handle_request` (standard RPC) doesn't. Is this intentional? If non-engine call latency is also relevant, it might be worth adding there too.

2. **Error path**: When JSON parsing fails (early return at line 576), no metric is recorded for the full duration. Could this be problematic if you want visibility into failed request latency, or is tracking only successful requests the intended behavior?

### Typo

In the PR description: "parameter **aprsing**" → "parameter **parsing**"

### CI

Lint is failing due to a formatting issue - needs `cargo fmt`.

ilitteri:
## Additional Analysis: E2E Measurement Gaps

After exploring the codebase more deeply, here are some additional gaps in the e2e measurements that might be worth considering:

### 1. HTTP Layer Overhead (not captured)

**Before the handler**: By the time `handle_authrpc_request` is called, Axum has already:
- Established TCP connection
- Parsed HTTP headers  
- Read the request body into a `String`

**After the handler**: When you return `Ok(Json(response))`, Axum still needs to:
- Serialize the `Value` to bytes (this is a **second serialization** - the first was in `rpc_response`)
- Write HTTP headers
- Send over the network

For large payloads like `engine_getPayloadBodiesByRangeV1`, this second serialization could be significant.

### 2. Double Serialization

Looking at `rpc_response`:
```rust
serde_json::to_value(RpcSuccessResponse { id, jsonrpc, result })
```
This creates a `Value`, then Axum's `Json` extractor serializes it again to bytes. Consider whether this double serialization should be:
- (a) Measured (by timing after `Json(response)` is fully sent)
- (b) Eliminated (by returning bytes directly)

### 3. Capturing True E2E

To capture true end-to-end latency including network I/O, you'd need a Tower middleware layer that wraps the entire request lifecycle. Something like:
```rust
// Pseudocode
.layer(LatencyTrackingMiddleware::new())
```

This would capture everything from connection accept to response sent.

---

These are just observations for future consideration - the current PR is a good incremental improvement for measuring the handler-level latency including JSON parsing. Whether capturing the full HTTP layer is necessary depends on your observability goals.

ilitteri:
## Suggestions

### 1. Capture HTTP Layer Overhead with Tower Middleware

To capture true e2e latency (including body reading and response sending), consider adding a Tower middleware layer:

```rust
use axum::{body::Body, http::Request, response::Response};
use std::time::Instant;
use tower::{Layer, Service};

#[derive(Clone)]
pub struct LatencyLayer;

impl<S> Layer<S> for LatencyLayer {
    type Service = LatencyMiddleware<S>;

    fn layer(&self, inner: S) -> Self::Service {
        LatencyMiddleware { inner }
    }
}

#[derive(Clone)]
pub struct LatencyMiddleware<S> {
    inner: S,
}

impl<S> Service<Request<Body>> for LatencyMiddleware<S>
where
    S: Service<Request<Body>, Response = Response> + Clone + Send + 'static,
    S::Future: Send,
{
    type Response = S::Response;
    type Error = S::Error;
    type Future = /* ... */;

    fn call(&mut self, req: Request<Body>) -> Self::Future {
        let start = Instant::now();
        let future = self.inner.call(req);
        
        async move {
            let response = future.await;
            // Record e2e duration here (from request received to response ready)
            record_http_e2e_duration(start);
            response
        }
    }
}
```

Then apply it:
```rust
let authrpc_router = Router::new()
    .route("/", post(authrpc_handler))
    .layer(LatencyLayer)  // <-- Add here
    .with_state(service_context.clone())
```

This would capture everything from when Axum starts processing the request to when the response is ready to send.

### 2. Avoid Double Serialization

Currently there's double serialization:
1. `rpc_response()` → `serde_json::to_value()` → `Value`
2. Axum's `Json<Value>` → serializes `Value` to bytes

Consider returning `Bytes` directly to serialize only once:

```rust
use axum::response::IntoResponse;
use bytes::Bytes;

pub async fn handle_authrpc_request(
    // ...
) -> impl IntoResponse {
    // ... existing logic ...
    
    let response = rpc_response(req.id, res)?;
    
    // Serialize directly to bytes (only once)
    let bytes = serde_json::to_vec(&response)?;
    
    record_full_rpc_duration(namespace, method, start);
    
    (
        [(axum::http::header::CONTENT_TYPE, "application/json")],
        Bytes::from(bytes),
    )
}
```

This eliminates the intermediate `Value` → bytes conversion that Axum does.

### 4. Warning: Missing Handlers

Note that `handle_http_request` (standard RPC for `eth_*`, `debug_*`, etc.) doesn't have full duration tracking. If you later wonder why some RPC methods don't show up in the `rpc_request_full_duration_seconds` metric, this is why. You might want to add the same timing there for consistency, or document that this metric is engine-only.

rodrigo-o:
@ilitteri double serialization is now fixed and I also implemented the missing handlers sharing a new `handle_http_request_inner`. Regarding Tower I'm not so sure, as far as I understand we won't be able to get the method name with tower (given that its part of the Json request). 

We could go with a follow-up PR for investigating different ways to use a middle-ware, but i guess we want to have this measuring serialization + deserialization merged first. Apart from that I'm gonna replace the old handler metric to avoid confusions or at least remove it from the panels.

rodrigo-o:
Here is a mainnet run, the difference between the handler and the full measurement is between 1~3ms (~5% in the worst case) for engine_newPayload which was the one we were most insterested about.

image

Arkenan:
CI not passing, we'll close and create issue

PR #5933, opened 2026-01-20
perf(l1,l2): use rwlock for trie_cache

Oppen:
**Motivation**

The trie cache is read heavy. Using an RwLock allows simultaneous readers to clone the inner Arc without blocking each other.

**Description**

Replace `std::sync::Mutex` with `parking_lot::RwLock`.

Initially tried `std::sync::RwLock`, which showed a small regression in mean throughput (0.537 → 0.539 Ggas/s):

image

Switched to `parking_lot::RwLock`, which shows a small improvement (0.523 → 0.526 Ggas/s):

image

`ethrex-mainnet-1` runs `main`, `ethrex-mainnet-2` runs this branch.

**Checklist**

### DIFF
--- a/crates/l2/tee/quote-gen/Cargo.lock
+++ b/crates/l2/tee/quote-gen/Cargo.lock
@@ -2475,6 +2475,7 @@ dependencies = [
  "fastbloom",
  "hex",
  "lru 0.16.2",
+ "parking_lot",
  "rayon",
  "rustc-hash",
  "serde 1.0.228",
--- a/crates/vm/levm/bench/revm_comparison/Cargo.lock
+++ b/crates/vm/levm/bench/revm_comparison/Cargo.lock
@@ -1213,6 +1213,7 @@ dependencies = [
  "fastbloom",
  "hex",
  "lru",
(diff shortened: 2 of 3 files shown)


ilitteri:
Nice improvement! One suggestion:

Consider using `parking_lot::RwLock` instead of `std::sync::RwLock`:

```rust
use parking_lot::RwLock;
```

Benefits:
- No lock poisoning (simpler API - no `.map_err(|_| StoreError::LockError)?` needed)
- Generally faster, especially for short critical sections like cloning an Arc
- `parking_lot` is already a dependency of the workspace (used in `ethrex-trie` and others)

The usage would simplify to:
```rust
// Read
self.trie_cache.read().clone()

// Write  
*self.trie_cache.write() = trie.clone();
```

Not blocking - the current implementation is correct. Just a potential follow-up improvement.

Oppen:
Implemented. std::sync::RwLock showed a small regression, but parking_lot::RwLock is a small win.

Arkenan:
Improvements not big enough, but let's check this again in the future (add to roadmap)

PR #5951, opened 2026-01-20
perf(l1): implement missing length functions

### LINKED ISSUE #5949: Implement RLP length() for missing types
`length()` is not implemented for some key types, which means that they will be encoded to calculate their size. These types are:

- Nibbles
- BranchNode
- ExtensionNode
- LeafNode
- Node

Arkenan:
(pending measurements)
Closes #5949


crates/networking/p2p/rlpx/snap.rs
```diff
@@ -376,6 +381,24 @@ impl RLPEncode for StorageSlot {
...
     }
+
+    fn length(&self) -> usize {
+        // self.data is encoded, then that encoded data is wrapped as bytes
+        let data_encoded_len = self.data.length();
+        // For bytes encoding: we need to know the first byte of the encoded data
+        // U256 encodes as bytes, first byte depends on value
+        // For simplicity, compute the bytes_length for the encoded data length
+        // The first byte of RLP-encoded U256 is >= 0x80 for multi-byte, or the value itself for single byte
+        // Since we don't know the first byte without encoding, use a conservative estimate
```
Oppen, line 392: If the encoded length is > 1, then we know there will be a prefix byte.
If it is == 1, then we only need to look at the least significant byte, or simply observe that it will be <= 0x80, with only 0x80 causing a +1 here, IIRC. The point is we don't really need to encode to know how many bytes it takes exactly, at least not in the obvious way.
We may want to call `length` to know which value to encode in a list of lists without encoding twice or memcpy, so I would refrain from using estimates.

crates/networking/p2p/rlpx/snap.rs
```diff
@@ -376,6 +381,24 @@ impl RLPEncode for StorageSlot {
...
+        let data_as_bytes_len = if data_encoded_len < 56 {
```
ElFantasma, line 393: Bug: this hand-rolled bytes-length calculation misses the RLP single-byte optimization. When `data_encoded_len == 1` and the first byte of the encoded U256 is ≤ 0x7f (i.e., values 1..127), `[u8]::encode` writes just that one byte (no length prefix), so the correct result is 1 — but this code returns `1 + 1 = 2`.

This can be fixed by using the existing `bytes_length` helper (already imported in this file), but you need the first byte of the encoded data. Since `U256::encode` for values 1..127 produces a single byte equal to the value itself, you could do:

```rust
fn length(&self) -> usize {
    let encoded = self.data.encode_to_vec();
    let data_as_bytes_len = if encoded.is_empty() {
        1 // RLP_NULL
    } else {
        bytes_length(encoded.len(), encoded[0])
    };
    let payload_len = self.hash.length() + data_as_bytes_len;
    list_length(payload_len)
}
```

Alternatively, if you want to avoid the allocation, you can compute the first byte from the U256 value directly (for values 0..127 it's the value itself, for 0 it's 0x80, for larger values it's `0x80 + byte_len`).

Issue #5958, opened 2026-01-21
EIP-7708: ETH transfers emit a log


Arkenan:
[EIP-7708](https://eips.ethereum.org/EIPS/eip-7708)

This is specially important for smart wallets that may transfer eth funds when calling or self destructing, which is not always processed correctly by indexers. Emitting logs would give eth the same ux/devx than erc20 transfers.

PR #5966, opened 2026-01-21
feat(l1): implement EIP-7708

edg-l:
**Motivation**

All ETH-transfers, including transactions, CALL and SELFDESTRUCT emit a log.

https://eips.ethereum.org/EIPS/eip-7708

A log, identical to a LOG3, is issued for:

- Any nonzero-value-transferring transaction, before any other logs created by EVM execution
- Any nonzero-value-transferring CALL, at the time that the value transfer executes
- Any nonzero-value-transferring SELFDESTRUCT to a different account, at the time that the value transfer executes

**Checklist**

Closes https://github.com/lambdaclass/ethrex/issues/5958

### DIFF
--- a/crates/vm/levm/src/opcode_handlers/system.rs
+++ b/crates/vm/levm/src/opcode_handlers/system.rs
@@ -836,8 +836,9 @@ impl<'a> VM<'a> {
             if should_transfer_value && ctx_result.is_success() {
                 self.transfer(msg_sender, to, value)?;
 
-                // EIP-7708: Emit transfer log for nonzero-value CALL/CALLCODE
-                if self.env.config.fork >= Fork::Amsterdam && !value.is_zero() {
+                // EIP-7708: Emit transfer log for nonzero-value CALL/CALLCODE to DIFFERENT accounts
+                // Self-transfers should NOT emit a log per the EIP spec
+                if self.env.config.fork >= Fork::Amsterdam && !value.is_zero() && msg_sender != to {
                     let log = create_eth_transfer_log(msg_sender, to, value);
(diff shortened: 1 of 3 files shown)


crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -560,18 +562,42 @@ impl<'a> VM<'a> {
...
+                self.substate.add_log(log);
+            }
+
             self.substate.add_selfdestruct(to);
+
+            // EIP-7708: Emit selfdestruct log when contract is destroyed
+            if self.env.config.fork >= Fork::Amsterdam {
+                let log = create_selfdestruct_log(to, balance);
+                self.substate.add_log(log);
+            }
```
iovoid: Shouldn't these be mutually exclusive?

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -560,18 +562,42 @@ impl<'a> VM<'a> {
...
                 // If target is the same as the contract calling, Ether will be burnt.
                 self.get_account_mut(to)?.info.balance = U256::zero();
 
                 self.substate.add_selfdestruct(to);
+
+                // EIP-7708: Emit selfdestruct log ONLY when destroying to self
+                if self.env.config.fork >= Fork::Amsterdam && to == beneficiary {
+                    let log = create_selfdestruct_log(to, balance);
+                    self.substate.add_log(log);
+                }
```
iovoid: Should also filter for balance being nonzero. I think it might be clearer (here and in the other selfdestruct case) to unify both log types under a same if.

PR #5968, opened 2026-01-21
perf(l1): use atomic fastbloom

iovoid:
**Motivation**

On the gigagas blocks benchmark and on bulk imports, we are running into issues with the bloom filter update taking too long.

We began observing this when the gigablock benchmarking tool changed how delay is interpreted, leading into blocks being sent without spacing and triggering this degradation.

**Description**

This PR replaces the library used with fastbloom that allows parallel insertion and setting a cheaper hash function.

The result is restoring the performance on the gigagas benchmarks (697mgas/s -> 975mgas/s). On mainnet, there seems to be no regression (3% worse in latency, 57.5ms->59.2ms).

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -57,11 +57,13 @@ use error::{ChainError, InvalidBlockError};
 use ethrex_common::constants::{EMPTY_TRIE_HASH, MIN_BASE_FEE_PER_BLOB_GAS};
 
 // Re-export stateless validation functions for backwards compatibility
+#[cfg(feature = "c-kzg")]
+use ethrex_common::types::EIP4844Transaction;
 use ethrex_common::types::block_execution_witness::ExecutionWitness;
 use ethrex_common::types::fee_config::FeeConfig;
 use ethrex_common::types::{
     AccountState, AccountUpdate, Block, BlockHash, BlockHeader, BlockNumber, ChainConfig, Code,
-    EIP4844Transaction, Receipt, Transaction, WrappedEIP4844Transaction,
+    Receipt, Transaction, WrappedEIP4844Transaction,
 };
 use ethrex_common::types::{ELASTICITY_MULTIPLIER, P2PTransaction};
 use ethrex_common::types::{Fork, MempoolTransaction};
@@ -94,7 +96,7 @@ use std::sync::{
     atomic::{AtomicBool, AtomicUsize, Ordering},
     mpsc::{Receiver, channel},
 };
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::sync::Mutex as TokioMutex;
 use tokio_util::sync::CancellationToken;
 
@@ -118,6 +120,7 @@ type BlockExecutionPipelineResult = (
     Option<Vec<AccountUpdate>>,
     usize,        // max queue length
     [Instant; 6], // timing instants
+    Duration,     // warmer duration
 );
 
 //TODO: Implement a struct Chain or BlockChain to encapsulate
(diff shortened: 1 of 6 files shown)


rodrigo-o:
Have we tested this change doesn't generate a regression in mainnet (outside of the enhancement that could be seen in the benchamark). Also please add the improvement seen in the benchmarks as part of this PR description.

crates/storage/layering.rs
```diff
@@ -11,31 +13,36 @@ struct TrieLayer {
...
+    }
 }
 
 impl Default for TrieLayerCache {
     fn default() -> Self {
-        // Try to create the bloom filter, if it fails use poison mode.
-        let bloom = Self::create_filter().ok();
         Self {
-            bloom,
+            bloom: Self::create_filter(1_000_000),
```
tomip01: We can make `1_000_000` a constant for better changes

crates/storage/layering.rs
```diff
@@ -45,27 +52,25 @@ impl Default for TrieLayerCache {
 
 impl TrieLayerCache {
     pub fn new(commit_threshold: usize) -> Self {
-        let bloom = Self::create_filter().ok();
         Self {
-            bloom,
+            bloom: Self::create_filter(1_000_000),
```
tomip01: ditto

crates/storage/layering.rs
```diff
@@ -45,27 +52,25 @@ impl Default for TrieLayerCache {
...
 
-    // TODO: tune this
-    fn create_filter() -> Result<qfilter::Filter, qfilter::Error> {
-        qfilter::Filter::new_resizeable(1_000_000, 100_000_000, 0.02)
-            .inspect_err(|e| tracing::warn!("could not create trie layering bloom filter {e}"))
+    fn create_filter(expected_items: usize) -> AtomicBloomFilter<FxBuildHasher> {
+        // 2% false positive rate, using FxHash for fast hashing
+        AtomicBloomFilter::with_false_pos(0.02)
+            .hasher(FxBuildHasher)
+            .expected_items(expected_items.max(1_000_000))
```
tomip01: ditto

crates/storage/layering.rs
```diff
@@ -150,27 +149,19 @@ impl TrieLayerCache {
...
-                "TrieLayerCache: rebuild_bloom could not create new filter. Poisoning bloom."
-            );
-            self.bloom = None;
-            return;
-        };
+        // Pre-compute total keys for optimal filter sizing
+        let total_keys: usize = self.layers.values().map(|layer| layer.nodes.len()).sum();
 
-        for layer in self.layers.values() {
+        let filter = Self::create_filter(total_keys.max(1_000_000));
```
tomip01: ditto

crates/storage/layering.rs
iovoid: Fixed in 2203e6d7bf

crates/storage/layering.rs
iovoid: Fixed in 2203e6d7bf

crates/storage/layering.rs
iovoid: Fixed in 2203e6d7bf

crates/storage/layering.rs
iovoid: Fixed in 2203e6d7bf

pablodeymo:
> Have we tested this change doesn't generate a regression in mainnet (outside of the enhancement that could be seen in the benchamark). Also please add the improvement seen in the benchmarks as part of this PR description.


I agree with this. We need to have this conclusion as part of PR description.

crates/storage/layering.rs
```diff
@@ -45,27 +54,25 @@ impl Default for TrieLayerCache {
...
         }
     }
 
-    // TODO: tune this
-    fn create_filter() -> Result<qfilter::Filter, qfilter::Error> {
-        qfilter::Filter::new_resizeable(1_000_000, 100_000_000, 0.02)
-            .inspect_err(|e| tracing::warn!("could not create trie layering bloom filter {e}"))
+    fn create_filter(expected_items: usize) -> AtomicBloomFilter<FxBuildHasher> {
+        // 2% false positive rate, using FxHash for fast hashing
+        AtomicBloomFilter::with_false_pos(0.02)
```
tomip01: I think we can make this a constant also

iovoid: Fixed in b3375abd82

iovoid:
Added the performance measurements (gigagas to show improvement, mainnet to show lack of regression).

PR #5980, opened 2026-01-21
refactor(l1): small cleanups to remove duplicate code

avilagaston9:
**Motivation**

Small cleanups to remove duplicate code and consolidate shared types.

**Description**

Three independent cleanups:
- Remove unused `parse_json_hex` function from L2 RPC utils (duplicated from ethrex-rpc)
- Re-export `H256Wrapper` from ethrex-common instead of duplicating it in rkyv_utils
- Consolidate `RpcResponse` enum into the shared utils module

**Checklist**


crates/common/rkyv_utils.rs
```diff
@@ -12,6 +12,9 @@ use std::{
     hash::{Hash, Hasher},
 };
 
+// Re-export H256Wrapper from ethrex-trie to avoid duplication
+pub use ethrex_trie::rkyv_utils::H256Wrapper;
+
```
tomip01, line 17: This is good, but can we keep the `H256Wrapper` in `ethrex-common` instead of the trie? I think it makes more sense to keep it here. Also the `trie/rkyv_utils.rs` only have this wrapper

Oppen: If `ethrex_trie` needs it, we need to fix the dependency tree first, as `ethrex_trie` depending on `ethrex_common` would cause a circular dependency.

PR #5995, opened 2026-01-22
refactor(l1,l2): extract guest program to dedicated ethrex-guest crate

ilitteri:
## Motivation

The guest program was deeply nested inside `crates/l2/prover/src/guest_program/`, making it harder to find and manage. This PR extracts it to a top-level `crates/guest-program/` directory as a standalone workspace member, simplifying the prover architecture.

## Description

- **Moved** `crates/l2/prover/src/guest_program/` → `crates/guest-program/`
- **Renamed** package from `guest_program` to `ethrex-guest-program`
- **Reorganized** zkVM binaries from `src/` subdirectories to `bin/` at crate root
- **Renamed** binary packages to `ethrex-guest-{risc0,sp1,zisk,openvm}`
- **Updated** all path references across:
  - Cargo.toml files (workspace, prover, l2, tee/quote-gen, ef_tests)
  - Rust imports (`guest_program` → `ethrex_guest_program`)
  - CI workflows (pr-main_l1, pr-main_l2, pr-main_l2_prover, main_prover, tag_release)
  - Docker compose volume mounts
- **Added** README.md documentation for the guest program crate

### New Structure

```
crates/guest-program/
├── Cargo.toml
├── build.rs
├── README.md
├── src/
│   ├── lib.rs
│   ├── common/
│   ├── l1/
│   ├── l2/
│   └── methods.rs
└── bin/
    ├── risc0/
    ├── sp1/
    ├── zisk/
    └── openvm/
```

## Test plan

- [x] Main workspace compiles (`cargo check`)
- [x] Tooling workspace compiles (`cd tooling && cargo check`)
- [x] `ethrex-guest-program` crate compiles (`cargo check -p ethrex-guest-program`)
- [x] Code formatting passes (`cargo fmt --check`)
- [ ] CI passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3741,6 +3741,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "ethrex-guest-program"
+version = "9.0.0"
+dependencies = [
+ "bincode 1.3.3",
+ "bytes",
+ "ethrex-blockchain",
+ "ethrex-common",
+ "ethrex-crypto",
+ "ethrex-l2-common",
+ "ethrex-rlp",
+ "ethrex-storage",
+ "ethrex-trie",
+ "ethrex-vm",
+ "hex",
+ "openvm-sdk",
+ "risc0-build",
+ "risc0-zkvm",
+ "rkyv",
+ "serde",
+ "serde_json",
+ "serde_with",
+ "sp1-build",
+ "sp1-sdk",
+ "thiserror 2.0.17",
+]
+
+[[package]]
 name = "ethrex-l2"
 version = "9.0.0"
 dependencies = [
@@ -3761,6 +3788,7 @@ dependencies = [
  "ethrex-common",
  "ethrex-config",
  "ethrex-dev",
+ "ethrex-guest-program",
  "ethrex-l2-common",
  "ethrex-l2-rpc",
  "ethrex-levm",
@@ -3774,7 +3802,6 @@ dependencies = [
  "ethrex-trie",
  "ethrex-vm",
  "futures",
- "guest_program",
  "hex",
  "jsonwebtoken 9.3.1",
  "lazy_static",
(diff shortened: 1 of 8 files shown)


.github/workflows/pr-main_l1.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          cargo test --workspace --exclude 'ethrex-l2*' --exclude ethrex-prover --exclude guest_program
+          cargo test --workspace --exclude 'ethrex-l2*' --exclude ethrex-prover --exclude ethrex-guest-program
```
tomip01, line 115: If we compile the guest program for l1, we could keep the testing for that

crates/guest-program/src/methods.rs
tomip01, line 1: I think a better name for this file could be `program_bindings.rs` or something that refers to the programs. I think `methods.rs` is a bit off

crates/l2/prover/src/backend/openvm.rs
```diff
@@ -1,15 +1,16 @@
...
-use guest_program::input::ProgramInput;
 use openvm_continuations::verifier::internal::types::VmStarkProof;
 use openvm_sdk::{Sdk, StdIn, types::EvmProof};
 use openvm_stark_sdk::config::baby_bear_poseidon2::BabyBearPoseidon2Config;
 use rkyv::rancor::Error;
 
 use crate::backend::{BackendError, ProverBackend};
 
-static PROGRAM_ELF: &[u8] = include_bytes!("../guest_program/src/openvm/out/riscv32im-openvm-elf");
+static PROGRAM_ELF: &[u8] =
```
tomip01, line 12: Just an idea, we could move all the program elfs to `crates/guest-program/src/methods.rs` and import them from there. I think that will make it more consistent. Also renaming it to `program_bindings.rs` or similar will make more sense. Let me know what you think

crates/guest-program/src/methods.rs
ilitteri: I think this is needed for RISC0 in particular. Check de documentation.

crates/l2/prover/src/backend/openvm.rs
ilitteri: As said in the other comment, this file is needed only for RISCO.

.github/workflows/pr-main_l1.yaml
ilitteri: I think we can, but right now we don't have any test in this crate.

Issue #5959, opened 2026-01-21
EIP-7778: Block Gas Accounting without Refunds


Arkenan:
[EIP-7778](https://eips.ethereum.org/EIPS/eip-7778)

Prevents refunds from making the total gas used per block exceed the gas limit.

PR #5996, opened 2026-01-22
feat(l1): implement EIP-7778

edg-l:
**Motivation**

Implements EIP-7778, which separates block-level gas accounting from user-level gas accounting. Block gas (cumulative_gas_used) now uses pre-refund gas, while user payment uses post-refund gas via a new gas_spent field in receipts. This prevents circumvention of block gas limits through refund exploitation (e.g., SSTORE clearing slots).

https://eips.ethereum.org/EIPS/eip-7778

Closes https://github.com/lambdaclass/ethrex/issues/5959

### DIFF
--- a/crates/vm/levm/src/execution_handlers.rs
+++ b/crates/vm/levm/src/execution_handlers.rs
@@ -2,11 +2,12 @@ use crate::{
     constants::*,
     errors::{ContextResult, ExceptionalHalt, InternalError, TxResult, VMError},
     gas_cost::CODE_DEPOSIT_COST,
+    utils::create_eth_transfer_log,
     vm::VM,
 };
 
 use bytes::Bytes;
-use ethrex_common::types::Code;
+use ethrex_common::types::{Code, Fork};
 
 impl<'a> VM<'a> {
     pub fn handle_precompile_result(
@@ -131,7 +132,15 @@ impl<'a> VM<'a> {
             }));
         }
 
-        self.increase_account_balance(new_contract_address, self.current_call_frame.msg_value)?;
+        let value = self.current_call_frame.msg_value;
+        self.increase_account_balance(new_contract_address, value)?;
+
+        // EIP-7708: Emit transfer log for nonzero-value contract creation transactions.
+        // Origin is sender, new_contract_address is the recipient.
+        if self.env.config.fork >= Fork::Amsterdam && !value.is_zero() {
+            let log = create_eth_transfer_log(self.env.origin, new_contract_address, value);
+            self.substate.add_log(log);
+        }
 
         self.increment_account_nonce(new_contract_address)?;
 


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -103,16 +103,24 @@ fn finalize_non_privileged_execution(
...
     let mut l1_gas = calculate_l1_fee_gas(vm, &fee_config.l1_fee_config)?;
 
     let mut total_gas = actual_gas_used
         .checked_add(l1_gas)
         .ok_or(InternalError::Overflow)?;
 
+    // EIP-7778: Track pre-refund gas including L1 gas
+    let mut total_gas_pre_refund = gas_used_pre_refund
+        .checked_add(l1_gas)
+        .ok_or(InternalError::Overflow)?;
```
avilagaston9: I know it was already like this before this PR, but I think we should add `l1_gas` to `gas_used_pre_refund` and use the total to compute the actual gas used. Something like this:
```Rust
// Save pre-refund gas for EIP-7778 block accounting
let gas_used_pre_refund = ctx_result.gas_used;
let mut l1_gas = calculate_l1_fee_gas(vm, &fee_config.l1_fee_config)?;
// EIP-7778: Track pre-refund gas including L1 gas
let mut total_gas_pre_refund = gas_used_pre_refund
        .checked_add(l1_gas)
let actual_gas_used = default_hook::compute_actual_gas_used(vm, gas_refunded, total_gas_pre_refund)?;
```

Then we can get rid of the `total_gas` variable, which has a terrible name.

crates/common/types/receipt.rs
```diff
@@ -17,29 +17,49 @@ pub type Index = u64;
 pub struct Receipt {
     pub tx_type: TxType,
     pub succeeded: bool,
+    /// Cumulative gas used by this and all previous transactions in the block.
+    /// Pre-EIP-7778: This is the post-refund gas.
+    /// Post-EIP-7778: This is the pre-refund gas (for block-level accounting).
```
ElFantasma: This comment seems misleading, as `cumulative_gas_used` is always POST-REFUND gas. The block-level accounting calculates it in `block_gas_used` that uses `receipt.gas_used` instead of `receipt.gas_spent`.

crates/common/types/receipt.rs
```diff
@@ -112,19 +141,34 @@ impl RLPDecode for Receipt {
 pub struct ReceiptWithBloom {
     pub tx_type: TxType,
     pub succeeded: bool,
+    /// Cumulative gas used by this and all previous transactions in the block.
+    /// Pre-EIP-7778: This is the post-refund gas.
+    /// Post-EIP-7778: This is the pre-refund gas (for block-level accounting).
```
ElFantasma: Same here: `cumulative_gas_used` is always POST-REFUND gas. The block-level accounting calculates it in `block_gas_used` that uses `receipt.gas_used` instead of `receipt.gas_spent`.

crates/vm/backends/levm/mod.rs
```diff
@@ -75,11 +82,24 @@ impl LEVM {
...
+            // - gas_used (PRE-REFUND for Amsterdam+) for block accounting
+            cumulative_gas_used += report.gas_spent;
+            block_gas_used += report.gas_used;
+
+            // EIP-7778: Set gas_spent for Amsterdam+ receipts
+            let gas_spent = if fork >= Fork::Amsterdam {
+                Some(report.gas_spent)
+            } else {
+                None
+            };
```
ElFantasma: This could be factored in a function, as it is [duplicated code](https://github.com/lambdaclass/ethrex/pull/5996/changes#diff-2509163e9dbbfbef498f470d6d0eba720e84ce2b81176464693181c7407726c9R185-R190)
(similar code also appears in tests and some other modules)

PR #5997, opened 2026-01-22
refactor(l1,l2): remove dependency of prover on the blockchain and storage modules

iovoid:
**Motivation**

Currently the guest programs depend on ethrex-blockchain, which transitively depends on ethrex-storage. This imposes several limitations on which crates ethrex-storage can use, despite the provers never actually using any of that code. 

**Description**

This PR moves the functions in ethrex-blockchain used by the prover (the block validation utilities) to ethrex-common/validations.

### DIFF
--- a/crates/common/validation.rs
+++ b/crates/common/validation.rs
@@ -4,45 +4,14 @@
 //! storage dependencies, making them suitable for use in zkVM guest programs.
 
 use crate::constants::{GAS_PER_BLOB, MAX_RLP_BLOCK_SIZE, POST_OSAKA_GAS_LIMIT_CAP};
+use crate::errors::InvalidBlockError;
 use crate::types::requests::{EncodedRequests, Requests, compute_requests_hash};
 use crate::types::{
-    Block, BlockHeader, ChainConfig, EIP4844Transaction, InvalidBlockBodyError,
-    InvalidBlockHeaderError, Receipt, compute_receipts_root, validate_block_header,
-    validate_cancun_header_fields, validate_prague_header_fields,
+    Block, BlockHeader, ChainConfig, EIP4844Transaction, Receipt, compute_receipts_root,
+    validate_block_header, validate_cancun_header_fields, validate_prague_header_fields,
     validate_pre_cancun_header_fields,
 };
 use ethrex_rlp::encode::RLPEncode;
-
-/// Errors that occur during block validation.
-///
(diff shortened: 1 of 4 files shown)


crates/common/validation.rs
```diff
@@ -0,0 +1,193 @@
...
+    ReceiptsRootMismatch,
+    #[error("Invalid Header, validation failed pre-execution: {0}")]
+    InvalidHeader(#[from] InvalidBlockHeaderError),
+    #[error("Invalid Body, validation failed pre-execution: {0}")]
+    InvalidBody(#[from] InvalidBlockBodyError),
+    #[error("Exceeded MAX_BLOB_GAS_PER_BLOCK")]
+    ExceededMaxBlobGasPerBlock,
+    #[error("Exceeded MAX_BLOB_NUMBER_PER_BLOCK")]
+    ExceededMaxBlobNumberPerBlock,
+    #[error("Gas used doesn't match value in header. Used: {0}, Expected: {1}")]
```
tomip01: I think we can move this to `crates/common/errors.rs`

crates/common/constants.rs
```diff
@@ -58,3 +58,7 @@ pub const RLP_BLOCK_SIZE_SAFETY_MARGIN: u64 = 2_097_152;
 pub const MAX_RLP_BLOCK_SIZE: u64 = MAX_BLOCK_SIZE - RLP_BLOCK_SIZE_SAFETY_MARGIN;
 // Blob base cost defined in EIP-7918
 pub const BLOB_BASE_COST: u64 = 8192;
+
+// === EIP-7825 constants ===
+// https://eips.ethereum.org/EIPS/eip-7825
+pub const POST_OSAKA_GAS_LIMIT_CAP: u64 = 16777216;
```
tomip01, line 64: It is already defined in `crates/vm/levm/src/constants.rs`, I would add a comment that is a duplicated constant

tomip01: Or maybe add in the levm crate this `pub use ethrex_common::constants::POST_OSAKA_GAS_LIMIT_CAP;
`

crates/common/validation.rs
iovoid: Done in 80b4a34a3d

crates/common/constants.rs
iovoid: Done in 80b4a34a3d

PR #6003, opened 2026-01-22
fix(l2): validate L1 address in withdrawERC20 to prevent invalid proofs

avilagaston9:
**Motivation**

`withdrawERC20` in `CommonBridgeL2` accepts both `tokenL1` and `tokenL2` as parameters but doesn't verify they match. While not currently exploitable (the L1 `deposits` mapping prevents claiming mismatched tokens), this could allow users to accidentally burn their L2 tokens and create withdrawal proofs that can never be claimed.

**Description**

- Added a `require(token.l1Address() == tokenL1)` check to `withdrawERC20` in `CommonBridgeL2.sol`
- Added integration test `test_erc20_withdraw_l1_address_mismatch` that:
  - Deploys ERC20 tokens on L1 and L2
  - Deposits tokens from L1 to L2
  - Attempts withdrawal with a wrong L1 address
  - Verifies the revert reason using `eth_call`
  - Confirms the transaction reverts and L2 balance is unchanged
- Fixed `EthClient::call` to include the nonce in `eth_call` requests when provided in Overrides. The nonce field was being ignored, causing the server to always receive nonce 0 regardless of what was passed by the caller. This is needed because the L2 VM validates nonces during `eth_call` execution.

**Checklist**

### DIFF
--- a/crates/l2/tests/tests.rs
+++ b/crates/l2/tests/tests.rs
@@ -919,15 +919,23 @@ async fn test_erc20_withdraw_l1_address_mismatch(
         "Expected revert reason to contain 'L1 address mismatch', got: {error_message}"
     );
 
-    let withdraw_receipt = test_send(
+    // Send the transaction with a fixed gas limit (gas estimation fails for reverting txs)
+    let withdraw_tx = build_generic_tx(
         &l2_client,
-        &rich_wallet_private_key,
+        TxType::EIP1559,
         COMMON_BRIDGE_L2_ADDRESS,
-        signature,
-        &data,
-        "test_erc20_withdraw_l1_address_mismatch",
+        rich_address,
+        calldata.clone(),
+        Overrides {
+            gas_limit: Some(500_000),
+            ..Default::default()
+        },
     )
     .await?;
+    let withdraw_tx_hash =
+        send_generic_transaction(&l2_client, withdraw_tx, &rich_wallet_signer).await?;
+    let withdraw_receipt =
+        ethrex_l2_sdk::wait_for_transaction_receipt(withdraw_tx_hash, &l2_client, 1000).await?;
 
     // The transaction should revert
     assert!(


crates/l2/tests/tests.rs
```diff
@@ -763,6 +769,211 @@ async fn test_erc20_roundtrip(
...
+    // Mint and approve tokens on L1
+    test_send(
+        &l1_client,
+        &rich_wallet_private_key,
+        token_l1,
+        "freeMint()",
+        &[],
+        "test_erc20_withdraw_l1_address_mismatch",
+    )
+    .await?;
```
tomip01: When deploying it mints token anyway, so this is not strictly necessary, just a comment

crates/l2/tests/tests.rs
```diff
@@ -763,6 +769,211 @@ async fn test_erc20_roundtrip(
...
+        ],
+        "test_erc20_withdraw_l1_address_mismatch",
+    )
+    .await?;
+
+    // Build a dummy tx to calculate its encoded size (needed for fee calculation)
+    let tx = Transaction::EIP1559Transaction(EIP1559Transaction {
+        data: Bytes::from(encode_calldata(
+            "approve(address,uint256)",
+            &[
```
tomip01: Can we make a helper function for this? In both places the only thing that changes is the calldata.

tomip01 (approved):
One small suggestion

crates/l2/tests/tests.rs
avilagaston9: Addressed [26a62be](https://github.com/lambdaclass/ethrex/pull/6003/commits/26a62be6c76066448252b6e909b16f7f052f77de)!

crates/l2/tests/tests.rs
avilagaston9: Addressed [26a62be](https://github.com/lambdaclass/ethrex/pull/6003/commits/26a62be6c76066448252b6e909b16f7f052f77de)!

ManuelBilbao (approved):

PR #6011, opened 2026-01-23
WIP: change default network fb path

Arkenan:


cmd/ethrex/cli.rs
```diff
@@ -63,10 +63,9 @@ pub struct Options {
     #[arg(
         long = "datadir",
         value_name = "DATABASE_DIRECTORY",
-        help = "If the datadir is the word `memory`, ethrex will use the InMemory Engine",
         default_value = default_datadir().into_os_string(),
-        help = "Receives the name of the directory where the Database is located.",
-        long_help = "If the datadir is the word `memory`, ethrex will use the `InMemory Engine`.",
+        help = "Base directory for the database. Network name is appended for public networks.",
+        long_help = "Base directory where the database is stored. For public networks (mainnet, hoodi, holesky, sepolia), the network name is automatically appended as a subdirectory (e.g., <datadir>/mainnet). For dev mode, a 'dev' subdirectory is used. If the value is 'memory', an in-memory engine is used instead.",
```
ElFantasma, line 68: Should we remove holesky as an option?

crates/common/config/networks.rs
```diff
@@ -109,6 +109,18 @@ impl Network {
...
 
+    /// Returns the directory suffix for this network, if any.
+    /// Used to separate database directories for different public networks.
+    pub fn datadir_suffix(&self) -> Option<&str> {
+        match self {
+            Network::PublicNetwork(PublicNetwork::Mainnet) => Some("mainnet"),
+            Network::PublicNetwork(PublicNetwork::Hoodi) => Some("hoodi"),
+            Network::PublicNetwork(PublicNetwork::Holesky) => Some("holesky"),
+            Network::PublicNetwork(PublicNetwork::Sepolia) => Some("sepolia"),
+            _ => None,
```
ElFantasma, line 120: When a new network appear, this "catch all" will make the compiler to miss it if the developer forgets to add it here

ilitteri:
## Review: Subcommands don't use network suffix

The network suffix logic is only applied in `init_l1()`, but the subcommands use `opts.datadir` directly without appending the network suffix. This will cause inconsistent behavior.

### Affected commands:

**1. `removedb`** (lines 475-477)
```rust
Subcommand::RemoveDB { datadir, force } => {
    remove_db(&datadir, force);  // Uses base dir, not network-specific
}
```
If a user runs `ethrex --network mainnet` (data in `<base>/mainnet`), then `ethrex removedb` will try to remove `<base>` instead of `<base>/mainnet`.

**2. `import`** (lines 480, 492)
```rust
remove_db(&opts.datadir.clone(), opts.force);  // Wrong path
import_blocks(&path, &opts.datadir, genesis, ...)  // Wrong path
```

**3. `import-bench`** (lines 504, 517) - same issue as import

**4. `export`** (line 528)
```rust
export_blocks(&path, &opts.datadir, first, last)  // Won't find the DB
```

### Suggestion

Extract the datadir computation logic into a helper function and use it consistently across `init_l1()` and all subcommands:

```rust
fn compute_effective_datadir(opts: &Options, network: &Network) -> PathBuf {
    if is_memory_datadir(&opts.datadir) {
        return opts.datadir.clone();
    }
    if opts.dev && cfg!(feature = "dev") {
        opts.datadir.join("dev")
    } else if let Some(suffix) = network.datadir_suffix() {
        opts.datadir.join(suffix)
    } else {
        opts.datadir.clone()
    }
}
```

Then update `Subcommand::run()` to compute the effective datadir before dispatching to individual commands.

### Additional note

The migration logic also only runs in `init_l1()`, so users running `import`/`export` commands won't be offered the migration prompt.

ElFantasma:
Superseded by #6247

PR #6020, opened 2026-01-26
feat(l1): bal types and engine_newPayloadV5 (continued)

edg-l:
**Motivation**

Continues work from https://github.com/lambdaclass/ethrex/pull/5726 by @lakshya-sky 

- Merged main, fixed conflicts
- Fixed prepending 0x when serializing hex

https://eips.ethereum.org/EIPS/eip-7928

### DIFF
--- a/.github/scripts/generate_block_time_report.sh
+++ b/.github/scripts/generate_block_time_report.sh
@@ -195,14 +195,12 @@ header_text="Daily block time report (24-hour average)"
         printf "• nethermind: %.3fms (mean)\n  %s\n" "$nether_value" "$version_nethermind"
         ;;
     esac
-  done < <(printf '%s\n' "${sort_entries[@]}" | sort -n)
+  done < <(printf '%s\n' "${sort_entries[@]}" | LC_ALL=C sort -n)
 } >"${OUTPUT_DIR}/block_time_report_github.txt"
 
-# Generate Slack message with code block table (truncated versions)
+# Generate Slack message (simple format, similar to Telegram)
 # Sorted by block time (ascending - fastest first)
-slack_table='```'$'\n'
-slack_table+='Client       Version                  Block Time'$'\n'
-slack_table+='-------------------------------------------------------'$'\n'
+slack_text=""
 
 # Build sortable entries: "value client"
 slack_sort_entries=()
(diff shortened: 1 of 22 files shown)


crates/networking/rpc/engine/payload.rs
```diff
@@ -541,6 +632,19 @@ fn validate_execution_payload_v3(payload: &ExecutionPayload) -> Result<(), RpcEr
...
 
+fn validate_execution_payload_v4(payload: &ExecutionPayload) -> Result<(), RpcErr> {
+    // This method follows the same specification as `engine_newPayloadV4` additionally
+    // rejects payload without block access list
+
+    validate_execution_payload_v3(payload)?;
+
+    if payload.block_access_list.is_none() {
+        return Err(RpcErr::WrongParam("block_access_list".to_string()));
+    }
```
Oppen, line 642: I think this should be first, as it is a much quicker check to reject early.

crates/networking/rpc/engine/payload.rs
```diff
@@ -644,6 +749,16 @@ async fn handle_new_payload_v3(
     handle_new_payload_v1_v2(payload, block, context).await
 }
 
+async fn handle_new_payload_v4(
+    payload: &ExecutionPayload,
+    context: RpcApiContext,
+    block: Block,
+    expected_blob_versioned_hashes: Vec<H256>,
+) -> Result<PayloadStatus, RpcErr> {
+    // TODO: V4 specific: validate block access list
```
ElFantasma, line 758: Is there a ticket for this? There should be a mention here.

crates/networking/rpc/types/payload.rs
```diff
@@ -162,6 +172,8 @@ impl ExecutionPayload {
             withdrawals: block.body.withdrawals,
             blob_gas_used: block.header.blob_gas_used,
             excess_blob_gas: block.header.excess_blob_gas,
+            // TODO: need to finish this after we are able to get BAL from blocks
```
ElFantasma, line 175: Let's link to a ticket (or fix it if we support BAL already)

crates/networking/rpc/engine/payload.rs
edg-l: its done in the next pr https://github.com/lambdaclass/ethrex/pull/6024

crates/networking/rpc/types/payload.rs
edg-l: its done in the next pr https://github.com/lambdaclass/ethrex/pull/6024

crates/common/types/block_access_list.rs
```diff
@@ -0,0 +1,498 @@
...
+        }
+
+        let buf = self.encode_to_vec();
+        keccak(buf)
+    }
+}
+
+impl RLPEncode for BlockAccessList {
+    fn encode(&self, buf: &mut dyn bytes::BufMut) {
+        let mut sorted = self.inner.clone();
```
iovoid: This should be doable without cloning, by sorting the pointers instead.

iovoid (approved):

PR #6024, opened 2026-01-26
feat(l1): eip-7928 part 2, BAL execution tracking, hash validation

edg-l:
Implements https://eips.ethereum.org/EIPS/eip-7928 for the Amsterdam fork. Block Access Lists (BAL) record all accounts and storage slots accessed during block execution, enabling ZK proof generation and optimized state reconstruction.

  Changes

  Core BAL Types (crates/common/types/block_access_list.rs)
  - Extended BlockAccessList with full EIP-7928 semantics
  - Added BlockAccessListRecorder for tracking state changes during execution
  - Implemented checkpoint/restore for proper revert handling
  - Block access indices use u16 per spec (0 = pre-execution, 1..n = transactions, n+1 = withdrawals)

  VM Integration (crates/vm/)
  - Added BAL recording hooks in LEVM for storage reads/writes, balance, nonce, and code changes
  - execute_block now optionally records and returns the BAL
  - Proper handling of reverts: storage reads persist, writes convert to reads
  - Coinbase and withdrawal recipients recorded as touched addresses

  Engine API (crates/networking/rpc/engine/)
  - Added engine_getPayloadV5 returning ExecutionPayloadV5 with BAL
  - Extended engine_newPayloadV5 to accept and validate BAL
  - BAL hash validation against block header

  Debug RPC (crates/networking/rpc/debug/)
  - Added debug_getBlockAccessList for retrieving BAL by block number/hash

  Blockchain (crates/blockchain/)
  - Block building now generates BAL for Amsterdam+ blocks
  - BAL validation integrated into block import

 Test Plan

  - Unit tests for BAL recording semantics (test/tests/levm/eip7928_tests.rs - 972 lines)
  - RLP encoding/decoding roundtrips
  - Revert behavior (reads persist, writes become reads)
  - Net-zero filtering within transactions
  - EF tests: Enabled fork_Amsterdam and ToAmsterdam test suites (~5971 tests passing)
  - EIP-7928-specific EF tests (94 failing - pending BAL hash validation alignment)
 

**Checklist**

Closes #issue_number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2026-02-06
+
+- Cache ECDSA sender recovery in transaction structs [#6153](https://github.com/lambdaclass/ethrex/pull/6153)
+
 ### 2026-01-27
 
 - Optimize prewarmer by grouping transactions by sender [#6047](https://github.com/lambdaclass/ethrex/pull/6047)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,7 +67,7 @@ version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
- "cpp_demangle",
+ "cpp_demangle 0.4.5",
  "fallible-iterator 0.3.0",
  "gimli 0.31.1",
  "memmap2",
@@ -165,6 +165,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "aligned-vec"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc890384c8602f339876ded803c97ad529f3842aba97f6392b3dba0dd171769b"
+dependencies = [
+ "equator",
 ]
 
 [[package]]
@@ -2249,7 +2258,7 @@ dependencies = [
  "num-bigint 0.4.6",
  "num-traits",
  "prost 0.13.5",
- "prost-build",
+ "prost-build 0.13.5",
  "rand 0.8.5",
  "ruint",
  "serde",
@@ -2556,6 +2565,15 @@ name = "cpp_demangle"
 version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2bb79cb74d735044c972aae58ed0aaa9a837e85b01106a54c39e42e97f62253"
+dependencies = [
+ "cfg-if 1.0.4",
+]
+
+[[package]]
+name = "cpp_demangle"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0667304c32ea56cb4cd6d2d7c0cfe9a2f8041229db8c033af7f8d69492429def"
 dependencies = [
  "cfg-if 1.0.4",
 ]
@@ -2870,7 +2888,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "73736a89c4aff73035ba2ed2e565061954da00d4970fc9ac25dcc85a2a20d790"
 dependencies = [
  "dispatch2",
- "nix",
+ "nix 0.30.1",
  "windows-sys 0.61.2",
 ]
 
@@ -3663,6 +3681,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "equator"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4711b213838dfee0117e3be6ac926007d7f433d7bbe33595975d4190cb07e6fc"
+dependencies = [
+ "equator-macro",
+]
+
+[[package]]
+name = "equator-macro"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "44f23cf4b44bfce11a86ace86f8a73ffdec849c9fd00a386a53d278bd9e81fb3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.114",
+]
+
+[[package]]
 name = "equivalent"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3798,6 +3836,7 @@ dependencies = [
  "itertools 0.14.0",
  "lazy_static",
  "local-ip-address",
+ "pprof",
  "rand 0.8.5",
  "rayon",
  "reqwest",
@@ -4562,6 +4601,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "findshlibs"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40b9e59cd0f7e0806cca4be089683ecb6434e602038df21fe6bf6711b2f07f64"
+dependencies = [
+ "cc",
+ "lazy_static",
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "fixed-hash"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4572,6 +4623,12 @@ dependencies = [
  "rustc-hex",
  "static_assertions",
 ]
+
+[[package]]
+name = "fixedbitset"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ce7134b9999ecaf8bcd65542e436736ef32ddca1b3e06094cb6ec5755203b80"
 
 [[package]]
 name = "fixedbitset"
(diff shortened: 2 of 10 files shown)


lakshya-sky:
Hey @edg-l, I did something similar in #5784 in which I took tracer based approach but this seems the right way to do it. However if this PR doesn't introduce `debug_getBlockAccessList` then I can create a seprate PR to add the rpc method that way we atleast test with Hive.

edg-l:
> Hey @edg-l, I did something similar in #5784 in which I took tracer based approach but this seems the right way to do it. However if this PR doesn't introduce `debug_getBlockAccessList` then I can create a seprate PR to add the rpc method that way we atleast test with Hive.

I'll add it in this PR, thanks for the heads up

crates/vm/levm/src/constants.rs
```diff
@@ -99,10 +99,8 @@ lazy_static::lazy_static! {
 
 // EIP-7708: ETH Transfers Emit a Log
 // System address for EIP-7708 logs (0xfffffffffffffffffffffffffffffffffffffffe)
-pub const EIP7708_SYSTEM_ADDRESS: H160 = H160([
-    0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
-    0xFF, 0xFF, 0xFF, 0xFE,
-]);
+// Re-exported from common constants for consistency
+pub use ethrex_common::constants::SYSTEM_ADDRESS as EIP7708_SYSTEM_ADDRESS;
```
iovoid: This change seems unrelated to the PR?

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -213,6 +230,7 @@ impl<'a> VM<'a> {
 
         self.substate.refunded_gas = gas_refunds;
 
+        // Main gas check - if this fails, the slot MUST NOT appear in BAL
```
iovoid: This contradicts the previous comment, that says if the main check fails (but not the stipend one) then the slot should appear as read.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -518,6 +616,9 @@ impl<'a> VM<'a> {
     ///
     /// Accessed storage slots are stored in the `accessed_storage_slots` set.
     /// Accessed storage slots take place in some gas cost computation.
+    ///
+    /// Per EIP-7928: BAL recording is deferred until after gas checks pass.
```
iovoid: Why not simply call access_storage_slot after the gas check passes? Which I think we already do.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -570,6 +684,23 @@ impl<'a> VM<'a> {
     ) -> Result<(), InternalError> {
         self.backup_storage_slot(address, key, current_value)?;
 
+        // Record storage write for BAL
+        // Per EIP-7928: actual writes (post != pre) are recorded as writes
+        // No-op writes (post == pre) are recorded as reads
+        if let Some(recorder) = self.db.bal_recorder.as_mut() {
+            let slot = U256::from_big_endian(key.as_bytes());
+            if new_value != current_value {
+                // Capture pre-storage value for net-zero filtering (first-write-wins)
```
iovoid: Could be clearer

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -528,9 +629,22 @@ impl<'a> VM<'a> {
...
+
         Ok((storage_slot, storage_slot_was_cold))
     }
 
+    /// Records a storage slot read to BAL after gas checks have passed.
+    /// Per EIP-7928: "If pre-state validation fails, the target is never accessed and must not appear in BAL."
+    /// This function should be called AFTER the gas check succeeds.
+    pub fn record_storage_slot_to_bal(&mut self, address: Address, key: H256) {
+        if let Some(recorder) = self.db.bal_recorder.as_mut() {
+            recorder.record_storage_read(address, U256::from_big_endian(key.as_bytes()));
```
iovoid: Every time this function is called we also have a U256 version of the key available, we could skip the conversion. Also taking H256 implies it takes keccak(key) when that's not the case

crates/vm/backends/levm/mod.rs
```diff
@@ -111,11 +166,17 @@ impl LEVM {
             VMType::L2(_) => Default::default(),
         };
 
-        Ok(BlockExecutionResult {
-            receipts,
-            requests,
-            block_gas_used,
-        })
+        // Extract BAL if recording was enabled
+        let bal = if record_bal { db.take_bal() } else { None };
```
iovoid: take_bal returns none if not enabled, the if is not needed

crates/vm/backends/levm/mod.rs
```diff
@@ -52,25 +53,52 @@ use std::sync::mpsc::Sender;
...
+    /// Execute a block and return the execution result.
+    ///
+    /// If `record_bal` is true, also records and returns the Block Access List (EIP-7928).
+    /// The BAL will be `None` if `record_bal` is false.
     pub fn execute_block(
         block: &Block,
         db: &mut GeneralizedDatabase,
         vm_type: VMType,
-    ) -> Result<BlockExecutionResult, EvmError> {
+        record_bal: bool,
```
iovoid: Shouldn't record_bal be derived from the current fork?

crates/common/types/block_access_list.rs
```diff
@@ -44,10 +45,31 @@ fn sorted_list_length<T: RLPEncode>(items: &[T]) -> usize {
...
+
+    /// Returns the block access index for this storage change.
+    pub fn block_access_index(&self) -> u16 {
+        self.block_access_index
+    }
+
+    /// Returns the post value for this storage change.
+    pub fn post_value(&self) -> U256 {
+        self.post_value
+    }
```
iovoid: Are the getters necessary?

crates/common/types/block_access_list.rs
```diff
@@ -100,10 +155,31 @@ impl RLPDecode for SlotChange {
...
+
+    /// Returns the block access index for this balance change.
+    pub fn block_access_index(&self) -> u16 {
+        self.block_access_index
+    }
+
+    /// Returns the post balance for this balance change.
+    pub fn post_balance(&self) -> U256 {
+        self.post_balance
+    }
```
iovoid: Are the getters necessary?

crates/common/types/block_access_list.rs
```diff
@@ -273,246 +447,653 @@ impl RLPDecode for BlockAccessList {
...
+    /// At build() time, we take only the last entry per transaction for each address.
+    pub fn record_balance_change(&mut self, address: Address, post_balance: U256) {
+        // SYSTEM_ADDRESS balance changes from system contract calls should not be recorded
+        // (system calls backup and restore SYSTEM_ADDRESS state)
+        if address == SYSTEM_ADDRESS {
+            return;
+        }
+
+        // Track initial balance for round-trip detection
+        self.initial_balances.entry(address).or_insert(post_balance);
```
iovoid: Can the balance ever change without reading it first? Maybe it would be better to take the initial balance as a parameter?

crates/networking/rpc/engine/payload.rs
```diff
@@ -577,6 +586,119 @@ fn build_payload_body_response(bodies: Vec<Option<BlockBody>>) -> Result<Value,
...
+                            .ok_or_else(|| {
+                                RpcErr::Internal(format!(
+                                    "Header not found for block {block_number}"
+                                ))
+                            })?;
+                    let block = Block { header, body };
+
+                    let bal = context
+                        .blockchain
+                        .generate_bal_for_block(&block)
```
iovoid, line 689: We shouldn't reexecute the block every time. This could be postponed since it won't have an impact until amsterdam is active, but an issue should be created.

crates/vm/backends/levm/mod.rs
```diff
@@ -429,17 +513,42 @@ impl LEVM {
...
+                    db.get_account(address)
+                        .map_err(|_| {
+                            EvmError::DB(format!("Withdrawal account {address} not found"))
+                        })?
+                        .info
+                        .balance,
+                )
+            } else {
+                None
+            };
```
iovoid: You can reuse the account lookup done below.

crates/networking/rpc/engine/payload.rs
edg-l: ill open an issue

crates/blockchain/payload.rs
```diff
@@ -575,6 +591,20 @@ impl Blockchain {
...
+            let tx_index = (context.payload.body.transactions.len() + 1) as u16;
+            context.vm.set_bal_index(tx_index);
+
+            // Record tx sender and recipient for BAL
+            if let Some(recorder) = context.vm.db.bal_recorder_mut() {
+                recorder.record_touched_address(head_tx.tx.sender());
+                if let TxKind::Call(to) = head_tx.to() {
+                    recorder.record_touched_address(to);
+                }
+            }
```
avilagaston9, line 606: We should do the same in `crates/l2/sequencer/block_producer/payload_builder.rs`

crates/vm/levm/src/hooks/default_hook.rs
```diff
@@ -246,7 +246,19 @@ pub fn pay_coinbase(vm: &mut VM<'_>, gas_to_pay: u64) -> Result<(), VMError> {
...
+    if !vm.env.gas_price.is_zero()
+        && let Some(recorder) = vm.db.bal_recorder.as_mut()
+    {
+        recorder.record_touched_address(vm.env.coinbase);
+    }
+
+    // Only pay coinbase if there's actually a fee to pay.
+    if !coinbase_fee.is_zero() {
+        vm.increase_account_balance(vm.env.coinbase, coinbase_fee)?;
+    }
```
avilagaston9, line 261: We should do the same for the `l2_hook.rs`

crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -263,10 +263,21 @@ fn pay_coinbase_l2(
...
         .ok_or(InternalError::Overflow)?;
 
-    if use_fee_token {
-        pay_fee_token(vm, vm.env.coinbase, coinbase_fee)?;
-    } else {
-        vm.increase_account_balance(vm.env.coinbase, coinbase_fee)?;
+    // Per EIP-7928: Coinbase must appear in BAL when there's a user transaction,
+    // even if the priority fee is zero. System contract calls have gas_price = 0,
+    // so we use this to distinguish them from user transactions.
+    if !vm.env.gas_price.is_zero()
```
avilagaston9: We don't have system contract calls in the L2

PR #6029, opened 2026-01-26
feat(metrics): add p50/p99 summary metrics

jrchatruc:
Add summary metrics with quantile support (p50, p90, p95, p99, p999) using the metrics-exporter-prometheus crate. This provides better observability for block execution performance and engine API latency.

Changes:
- Add metrics recorder with configurable sliding window (2 min default)
- Add block execution throughput metrics (gigagas/s)
- Add engine newPayload consensus layer latency metric
- Update comparison dashboard with correct queries for client comparison

The new metrics use a CKMS-based algorithm that provides accurate quantile estimation with bounded memory usage.

**Motivation**

**Description**

**Checklist**


crates/blockchain/metrics/rpc.rs
```diff
@@ -55,6 +56,14 @@ pub fn record_rpc_outcome(namespace: &str, method: &str, outcome: RpcOutcome) {
     METRICS_RPC_REQUEST_OUTCOMES
         .with_label_values(&[namespace, method, outcome.as_label(), outcome.error_kind()])
         .inc();
+    // Record to new metrics system
```
ElFantasma, line 59: The new `counter!` call drops the `error_kind` label that the old prometheus counter had (`["namespace", "method", "outcome", "error_kind"]`). This means the new metrics system loses the ability to distinguish between different error types (e.g., `BadParams` vs `Internal`). If the intention is to eventually replace the old metrics, you'll want to preserve that label:
```rust
counter!(
    "rpc_requests_total",
    "namespace" => namespace.to_string(),
    "method" => method.to_string(),
    "outcome" => outcome.as_label().to_string(),
    "error_kind" => outcome.error_kind().to_string()
)
.increment(1);
```

crates/networking/rpc/engine/payload.rs
```diff
@@ -618,6 +623,11 @@ async fn handle_new_payload_v1_v2(
 
     // All checks passed, execute payload
     let payload_status = try_execute_payload(block, &context, latest_valid_hash).await?;
+
+    // Record consensus layer newPayload latency
+    let duration_secs = start.elapsed().as_secs_f64();
```
ElFantasma, line 628: This histogram only records when `try_execute_payload` succeeds. If the function returns early (syncer not initialized at line 701, invalid block hash at line 708, invalid ancestors at line 713, snap sync at line 721), no duration is recorded. This biases the histogram toward successful execution times only.

If that's intentional ("only measure execution latency, not validation errors"), a comment would help. Otherwise, move the timing measurement to the outer `handle()` call site (e.g., in `NewPayloadV3Request::handle`) to capture the full CL→response latency regardless of outcome.

crates/blockchain/metrics/recorder.rs
```diff
@@ -0,0 +1,76 @@
...
+pub const DEFAULT_QUANTILES: &[f64] = &[0.0, 0.5, 0.9, 0.95, 0.99, 0.999, 1.0];
+
+/// Default bucket duration for the sliding time window
+/// Using shorter buckets for more responsive metrics
+pub const DEFAULT_BUCKET_DURATION: Duration = Duration::from_secs(30); // 30 seconds per bucket
+
+/// Default number of buckets in the sliding window
+/// Total window = bucket_count * bucket_duration = 4 * 30s = 2 minutes
+/// This provides more responsive quantiles that track recent block variations
+pub const DEFAULT_BUCKET_COUNT: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(4) };
```
ElFantasma, line 27: nit: `unsafe` can be avoided here. Since Rust 1.83, `Option::unwrap()` is const-stable, so you can use: `pub const DEFAULT_BUCKET_COUNT: NonZeroU32 = NonZeroU32::new(4).unwrap();`. Minor, but nice to avoid an unnecessary `unsafe` block.

crates/blockchain/metrics/blocks.rs
```diff
@@ -49,188 +40,153 @@ impl MetricsBlocks {
     pub fn new() -> Self {
         MetricsBlocks {
             gas_limit: Gauge::new(
-                "gas_limit",
-                "Keeps track of the percentage of gas limit used by the last processed block",
+                "old_gas_limit",
```
ElFantasma, line 43: Renaming `gas_limit` → `old_gas_limit` (and similarly for all other block metrics) is a breaking change for any existing Grafana dashboard, PromQL alert, or monitoring integration that references these metric names. Only `ethrex_comparison.json` is updated in this PR. Consider keeping the legacy names unchanged during the transition and giving the new `metrics` crate metrics a distinct prefix (e.g., `ethrex_gas_limit` or `block_gas_limit`). That way both systems emit concurrently without breaking existing consumers, and the old names can be removed in a follow-up PR once dashboards have been migrated.

PR #6031, opened 2026-01-26
feat(l1): add HTTP/2 support for RPC servers

ilitteri:
**Motivation**

Enable HTTP/2 support for RPC servers to allow clients to benefit from HTTP/2's multiplexing and header compression, which can improve performance for clients making many concurrent requests.

**Description**

Add HTTP/2 over cleartext (h2c) support for HTTP and Auth-RPC servers. The servers use hyper's auto-detection to support both HTTP/1.1 and HTTP/2 via the HTTP Upgrade mechanism or prior knowledge.

**Changes:**
- Add `serve_with_http2()` function using `hyper-util`'s `auto::Builder` with `serve_connection_with_upgrades()`
- HTTP and Auth-RPC servers now use h2c by default
- WebSocket server remains HTTP/1.1 only for client compatibility
- Graceful shutdown support via `tokio::select!`

**Testing:**
```bash
# Test HTTP/2
curl --http2 http://localhost:8545 -X POST -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}'

# Verify HTTP/2 negotiated
curl -v --http2 http://localhost:8545 2>&1 | grep "< HTTP/2"

# HTTP/1.1 still works
curl --http1.1 http://localhost:8545 -X POST -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}'
```

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. *(N/A - no storage changes)*


crates/networking/rpc/rpc.rs
```diff
@@ -353,6 +358,79 @@ pub fn start_block_executor(
...
+                        continue;
+                    }
+                };
+
+                let router = router.clone();
+
+                tokio::spawn(async move {
+                    let io = TokioIo::new(stream);
+                    let service = router.into_make_service();
+                    let mut make_service = service;
```
ElFantasma, line 463: As suggested by Copilot, the intermediate variable is not necessary:
```suggestion
                    let mut make_service = router.into_make_service();
```

PR #6033, opened 2026-01-27
perf(l1): replace `TransactionQueue::heads` from `Vec<_>` to `VecDeque<_>`

azteca1998:
**Motivation**

The field `heads` of `TransactionQueue` is being used as a FIFO.

**Description**

Use containers designed for FIFO-like operations (VecDeque) instead of simple arrays (Vec).

**Checklist**


azteca1998:
Analysis proved that `TransactionQueue::heads`'s `Vec::remove(0)` is not significant within the critical path and does not show an improvement greater than 0.5%-1% on block building benchmarks.

PR #6034, opened 2026-01-27
perf(l1): implement a cache for `EXTCODESIZE`

azteca1998:
**Motivation**

The opcode `EXTCODESIZE` currently reads the entire bytecode from memory just to return its length.

**Description**

Cache the bytecode length into the database so that only 8 bytes are read instead of the entire bytecode.

**Checklist**

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -26,7 +26,8 @@ use ethrex_common::{
     types::{
         AccountInfo, AccountState, AccountUpdate, Block, BlockBody, BlockHash, BlockHeader,
         BlockNumber, ChainConfig, Code, CodeMetadata, ForkId, Genesis, GenesisAccount, Index,
-        Receipt, Transaction, block_execution_witness::ExecutionWitness,
+        Receipt, Transaction,
+        block_execution_witness::{ExecutionWitness, RpcExecutionWitness},
     },
     utils::keccak,
 };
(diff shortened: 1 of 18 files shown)


crates/vm/levm/src/opcode_handlers/environment.rs
```diff
@@ -261,8 +261,8 @@ impl<'a> VM<'a> {
     pub fn op_extcodesize(&mut self) -> Result<OpcodeResult, VMError> {
         let address = word_to_address(self.current_call_frame.stack.pop1()?);
         let address_was_cold = !self.substate.add_accessed_address(address);
-        // FIXME: a bit wasteful to fetch the whole code just to get the length.
-        let account_code_length = self.db.get_account_code(address)?.bytecode.len().into();
+        // Optimized: only fetch code length, not full bytecode
```
Oppen: The comment is redundant.

crates/storage/store.rs
```diff
@@ -1295,7 +1389,9 @@ impl Store {
 
         for (code_hash, code) in update_batch.code_updates {
             let buf = encode_code(&code);
+            let metadata_buf = (code.bytecode.len() as u64).to_be_bytes().to_vec();
```
Oppen: No need for `to_vec` AFAICT. It will take a slice anyway.

crates/storage/store.rs
```diff
@@ -705,11 +709,93 @@ impl Store {
...
+
+        Ok(Some(metadata))
+    }
+
     /// Add account code
     pub async fn add_account_code(&self, code: Code) -> Result<(), StoreError> {
         let hash_key = code.hash.0.to_vec();
         let buf = encode_code(&code);
-        self.write_async(ACCOUNT_CODES, hash_key, buf).await
+        let metadata_buf = (code.bytecode.len() as u64).to_be_bytes().to_vec();
```
Oppen: Ditto.

crates/storage/store.rs
```diff
@@ -705,11 +709,93 @@ impl Store {
...
+            return Ok(Some(CodeMetadata { length: 0 }));
+        }
+
+        // Check cache first
+        if let Some(metadata) = self
+            .code_metadata_cache
+            .lock()
+            .map_err(|_| StoreError::LockError)?
+            .get(&code_hash)
+            .copied()
```
Oppen, line 731: Why have multiple levels of caching the same data? I'd suggest we either remove this one or the one in Levm. Preferably the one in Levm, as it will typically be colder.

crates/storage/api/tables.rs
```diff
@@ -97,9 +102,10 @@ pub const MISC_VALUES: &str = "misc_values";
 /// - [`Vec<u8>`] = `serde_json::to_vec(&witness)`
 pub const EXECUTION_WITNESSES: &str = "execution_witnesses";
 
-pub const TABLES: [&str; 18] = [
+pub const TABLES: [&str; 19] = [
     CHAIN_DATA,
     ACCOUNT_CODES,
+    ACCOUNT_CODE_METADATA,
```
Oppen, line 108: I think there must be a way to avoid using an extra table for this. The database necessarily stores the length of values. Haven't found a way to query for that yet, tho.

azteca1998: It's probably to avoid having to re-sync everything. I asked specifically about that and was told it was not necessary to resync anything. If it didn't have the data it'll be cached so that next time it's already there.

PR #6038, opened 2026-01-27
perf(l1): parallelize memoize_hashes on branch children in trie

edg-l:
**Motivation**

Paralelizes memoize_hashes on branch children 

On fast expb, seems to reduce p99 by 10-12 ms reliably, avg and median look the same with a little bit better in the pr

### DIFF
--- a/crates/common/trie/node.rs
+++ b/crates/common/trie/node.rs
@@ -8,6 +8,7 @@ pub use branch::BranchNode;
 use ethrex_rlp::{decode::RLPDecode, encode::RLPEncode};
 pub use extension::ExtensionNode;
 pub use leaf::LeafNode;
+use rayon::prelude::*;
 use rkyv::{
     de::Pooling,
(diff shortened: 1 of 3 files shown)


crates/common/trie/node.rs
```diff
@@ -178,12 +179,12 @@ impl NodeRef {
...
 
-    pub fn memoize_hashes(&self, buf: &mut Vec<u8>) {
+    pub fn memoize_hashes(&self) {
         if let NodeRef::Node(node, hash) = &self
             && hash.get().is_none()
         {
-            node.memoize_hashes(buf);
-            let _ = hash.set(node.compute_hash_no_alloc(buf));
+            node.memoize_hashes();
+            let _ = hash.set(node.compute_hash_no_alloc(&mut Vec::with_capacity(512)));
```
Oppen, line 187: Why does it allocate now? This may be detrimental for prover execution.

PR #6042, opened 2026-01-27
fix(levm): charge EIP-7702 delegation resolution gas for initial tx calls

ilitteri:
**Motivation**

When a transaction targets an EIP-7702 delegated account, resolving the delegation requires accessing the delegated address. This may incur a cold access cost (2600 gas) if the address hasn't been accessed yet. The gas cost returned from `eip7702_get_code()` was being dropped in `set_bytecode_and_code_address()`, causing transactions to be undercharged.

The CALL/CALLCODE/DELEGATECALL/STATICCALL opcodes correctly charge this gas via `get_call_gas_params()`, but the initial transaction call was missing this charge.

**Description**

The `eip7702_gas_consumed` returned from `eip7702_get_code()` is now charged to the call frame's consumed gas instead of being discarded:

```rust
// Before
let (_is_delegation, _eip7702_gas_consumed, code_address, bytecode) =
    eip7702_get_code(vm.db, &mut vm.substate, to)?;

// After
let (_is_delegation, eip7702_gas_consumed, code_address, bytecode) =
    eip7702_get_code(vm.db, &mut vm.substate, to)?;

vm.current_call_frame
    .increase_consumed_gas(eip7702_gas_consumed)
    .map_err(|_| TxValidationError::IntrinsicGasTooLow)?;
```

**Checklist**

### DIFF
--- a/crates/vm/levm/src/hooks/default_hook.rs
+++ b/crates/vm/levm/src/hooks/default_hook.rs
@@ -502,8 +502,13 @@ pub fn set_bytecode_and_code_address(vm: &mut VM<'_>) -> Result<(), VMError> {
     } else {
         // Here bytecode and code_address could be either from the account or from the delegated account.
         let to = vm.current_call_frame.to;
-        let (_is_delegation, _eip7702_gas_consumed, code_address, bytecode) =
+        let (_is_delegation, eip7702_gas_consumed, code_address, bytecode) =
             eip7702_get_code(vm.db, &mut vm.substate, to)?;
+
+        // Charge the gas cost for resolving EIP-7702 delegation (cold/warm access to delegated address)
+        vm.current_call_frame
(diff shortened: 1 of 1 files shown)


ilitteri:
Closing this PR - the original code is correct.

After further analysis and EF test regression, I realized the gas is intentionally dropped:

1. **EIP-7702 specifies delegation resolution for "opcodes which get code"** (CALL, CALLCODE, etc.) - the initial transaction is not an opcode
2. **EIP-2929 warm/cold access costs apply to opcodes**, not the transaction entry point
3. **The underscore prefix** (`_eip7702_gas_consumed`) indicates intentional unused value
4. **The function still serves its purpose**: adding the delegated address to `accessed_addresses` so subsequent CALLs are warm

The auditor's observation was technically correct (gas is dropped), but it's intentional and correct behavior per the spec.

PR #6043, opened 2026-01-27
test(levm): add EIP-7702 delegation gas behavior tests

ilitteri:
## Motivation

During audit review on a potential EIP-7702 gas accounting issue, it was determined that the current LEVM implementation is correct - `_eip7702_gas_consumed` is intentionally dropped at transaction setup because EIP-7702 specifies delegation resolution for "opcodes which get code" (CALL, CALLCODE, DELEGATECALL, STATICCALL, etc.), not for initial transactions.

These tests document and verify this behavior to prevent future regressions and serve as documentation of the expected semantics.

## Description

Add three tests to verify EIP-7702 delegation gas behavior:

1. **`test_initial_tx_to_delegated_account_no_cold_access_charge`**: Verifies that when a transaction is sent to a delegated account (with `0xef0100 || target_address` bytecode), the cold access cost (2600 gas) for the delegation target is NOT charged at transaction setup time.

2. **`test_delegated_address_is_warmed_after_resolution`**: Verifies that after delegation resolution, the target address IS added to `accessed_addresses` (warming it for subsequent calls), even though no gas was charged.

3. **`test_delegation_code_format`**: Verifies the delegation code format is correctly constructed (3 byte prefix + 20 byte address = 23 bytes).


test/tests/levm/eip7702_tests.rs
```diff
@@ -0,0 +1,259 @@
...
+    assert!(
+        result.is_success(),
+        "Transaction should succeed with delegated code execution"
+    );
+}
+
+/// Test: Verify delegated address is warmed after initial transaction
+///
+/// After the initial transaction to a delegated account, subsequent accesses
+/// to the delegated target should be warm (100 gas) not cold (2600 gas).
```
iovoid: The description implies we are testing the gas cost, but we aren't.

test/tests/levm/eip7702_tests.rs
```diff
@@ -0,0 +1,259 @@
...
+    vm::{VM, VMType},
+};
+use ethrex_storage::Store;
+use ethrex_vm::DynVmDatabase;
+use once_cell::sync::OnceCell;
+use rustc_hash::FxHashMap;
+use std::sync::Arc;
+
+// EIP-7702 delegation code prefix: 0xef0100
+const DELEGATION_PREFIX: [u8; 3] = [0xef, 0x01, 0x00];
```
ElFantasma: We could import the [constant](https://github.com/lambdaclass/ethrex/blob/1ac98d49455f02622ed0adabfb6aa3d8e5031eed/crates/vm/levm/src/constants.rs#L82) in levm

test/tests/levm/eip7702_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+/// - The delegation is resolved and B's code is executed
+/// - B is added to accessed_addresses (for warming)
+/// - But the cold access cost (2600 gas) for B is NOT charged
+///
+/// This test verifies the gas consumed matches expected (without cold access cost).
+#[test]
+fn test_initial_tx_to_delegated_account_no_cold_access_charge() {
+    let sender = Address::from_low_u64_be(1);
+    let delegated_account = Address::from_low_u64_be(2); // Account A
+    let target_account = Address::from_low_u64_be(3); // Account B
```
iovoid: Those are all in the precompile range, this has unintended side effects.

test/tests/levm/eip7702_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+        },
+    );
+
+    // Delegated account (A) - points to target (B)
+    let delegation_code = create_delegation_code(target_account);
+    accounts.insert(
+        delegated_account,
+        Account {
+            code: Code::from_bytecode(delegation_code),
+            ..Default::default()
```
iovoid, line 134: The info has an empty code hash

test/tests/levm/eip7702_tests.rs
```diff
@@ -0,0 +1,257 @@
...
+    let result = vm.execute().expect("Execution failed");
+
+    // The gas used should NOT include COLD_ADDRESS_ACCESS_COST for the target account.
+    // Base transaction cost is 21000, plus minimal execution cost.
+    // If cold access was incorrectly charged, we'd see an extra 2600 gas.
+    let gas_used = result.gas_used;
+
+    // Verify gas is reasonable (less than base + cold access cost buffer)
+    // If the bug was present, gas_used would be >= 21000 + 2600 = 23600
+    // With correct behavior, it should be around 21000 (just intrinsic gas)
```
iovoid: We should assert the lower bound too

iovoid (changes requested):

iovoid (approved):

PR #6045, opened 2026-01-27
fix(levm): reorder fee token validations to prevent storage rollback issue

ilitteri:
## Motivation

Addresses audit finding: fee token storage changes are not rolled back when validation fails after fee deduction.

In `prepare_execution_fee_token`, fee deduction via `deduct_caller_fee_token` was happening BEFORE several validation checks (e.g., `validate_sufficient_max_fee_per_gas`, `validate_priority_fee`). Since `transfer_fee_token` uses `vm.db.get_account_mut` directly (bypassing the backup mechanism), if any subsequent validation failed, the fee token storage changes could not be rolled back by `restore_cache_state`.

This could lead to fee tokens being locked in the fee token contract without a corresponding transfer to the sequencer.

## Description

Moves `deduct_caller_fee_token` to AFTER all validation checks that can fail, ensuring fee tokens are never deducted for transactions that will fail validation.

The key insight is that `vm.db.get_account_mut` (used by `transfer_fee_token`) does NOT backup storage changes, while `vm.get_account_mut` does. The code comment explicitly warns: "Use directly only if outside of the EVM, otherwise use `vm.get_account_mut` because it contemplates call frame backups."

## Checklist

- [x] Blockchain tests pass
- [x] Lint passes
- [x] Code review

### DIFF
--- a/crates/vm/levm/src/hooks/l2_hook.rs
+++ b/crates/vm/levm/src/hooks/l2_hook.rs
@@ -683,7 +683,6 @@ fn prepare_execution_fee_token(vm: &mut VM<'_>) -> Result<U256, crate::errors::V
     Ok(fee_token_ratio)
 }
 
-
 /// Helper function to encode the calldata for the fee token contract calls.
 /// <function>(address,uint256)
 fn encode_fee_token_call(selector: [u8; 4], address: Address, amount: U256) -> Bytes {


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -515,6 +518,11 @@ fn prepare_execution_fee_token(vm: &mut VM<'_>) -> Result<(), crate::errors::VME
     // Transaction is type 4 if authorization_list is Some
     // NOT CHECKED: fee token transactions are not type 4
 
+    // (3) INSUFFICIENT_ACCOUNT_FUNDS
```
Oppen: This number is inconsistent.

crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -462,9 +462,6 @@ fn prepare_execution_fee_token(vm: &mut VM<'_>) -> Result<(), crate::errors::VME
     // (2) INSUFFICIENT_MAX_FEE_PER_BLOB_GAS
     // NOT CHECKED: the blob price does not matter, fee token transactions do not support blobs
 
-    // (3) INSUFFICIENT_ACCOUNT_FUNDS
-    deduct_caller_fee_token(vm, gaslimit_price_product.saturating_mul(fee_token_ratio))?;
-
     // (4) INSUFFICIENT_MAX_FEE_PER_GAS
```
iovoid, line 616: Now there's a gap in the numbering.

avilagaston9:
This issue has already been addressed by #6330 (`fix(levm): make finalize mutations atomic with rollback`), which added `backup_storage_slot` calls to `transfer_fee_token`. Fee token storage changes are now properly backed up before being applied, and `restore_cache_state` (called by `execute()` when `prepare_execution` fails) correctly reverts them.

I'd suggest modifying this PR to only add a regression test that confirms the fix from #6330 stays in place.

PR #6050, opened 2026-01-28
docs(l1,l2): jemalloc memory profiling endpoint

Oppen:
## Motivation

We need better information for optimizations and we already have in place code that can be used for continuous memory profiling.

## Description

Document memory profiling with jemalloc pprof data.

- Add `docs/profiling/memory-profiling.md` covering build instructions, endpoint docs, Pyroscope integration, manual analysis, and troubleshooting.
- Remove broken `monitoring-stack` link (repo returns 404).
- Add link to `google/pprof` and flamegraph endpoint to manual analysis section.


docs/profiling/memory-profiling.md
```diff
@@ -0,0 +1,51 @@
+# Memory Profiling with Jemalloc
+
+## Overview
+
+Ethrex supports memory profiling using jemalloc's built-in profiling capabilities. This allows you to collect heap allocation profiles in pprof format for analysis.
+
+## Building with Profiling Enabled
+
+```bash
+cargo build --features jemalloc_profiling --release
```
ElFantasma, line 10: nit: This should be `cargo build --package ethrex --features jemalloc_profiling --release` since the feature is defined on the `ethrex` binary crate (`cmd/ethrex/Cargo.toml`), not the workspace. Running it without `--package` works but only because Cargo forwards the flag — it's more explicit and avoids confusion if someone tries from a different directory.

PR #6057, opened 2026-01-28
perf(l1): use Bytes for trie values to enable O(1) clones

pablodeymo:
## Summary

- Change `ValueRLP` type from `Vec<u8>` to `Bytes` (reference-counted) for O(1) clones in trie cache lookups
- Storage reads return `Bytes` (zero-copy from RocksDB)
- Storage writes use references (RocksDB passes directly to `put_cf`)
- In-memory backend copies on write (acceptable for testing, prioritizing RocksDB perf)

## Key Changes

**Trie layer:**
- `ValueRLP = Bytes` instead of `Vec<u8>`
- `LeafNode` and `BranchNode` store `Bytes` values
- Added `BytesWrapper` for rkyv serialization

**Storage API:**
- `get()` returns `Option<Bytes>` - zero-copy from RocksDB's `Vec<u8>`
- `put_batch(&[(&[u8], &[u8])])` - references passed directly to RocksDB

**Performance focus:**
- RocksDB (production): writes pass slices directly, reads are zero-copy
- In-memory (testing): stores `Bytes`, copies on write
- Trie cache lookups: O(1) `Bytes` clones instead of O(n) `Vec` copies

## Test plan

- [ ] Run existing tests
- [ ] Benchmark trie operations to measure improvement

### DIFF
--- a/tooling/sync/docker_monitor.py
+++ b/tooling/sync/docker_monitor.py
@@ -48,6 +48,18 @@ STATUS_EMOJI = {
     "block_processing": "📦", "success": "🎉", "failed": "❌"
 }
 
+# Phase completion patterns for parsing sync logs
+PHASE_COMPLETION_PATTERNS = {
+    "Block Headers": r"✓ BLOCK HEADERS complete: ([\d,]+) headers in (\d+:\d{2}:\d{2})",
+    "Account Ranges": r"✓ ACCOUNT RANGES complete: ([\d,]+) accounts in (\d+:\d{2}:\d{2})",
+    "Account Insertion": r"✓ ACCOUNT INSERTION complete: ([\d,]+) accounts inserted in (\d+:\d{2}:\d{2})",
+    "Storage Ranges": r"✓ STORAGE RANGES complete: ([\d,]+) storage slots in (\d+:\d{2}:\d{2})",
+    "Storage Insertion": r"✓ STORAGE INSERTION complete: ([\d,]+) storage slots inserted in (\d+:\d{2}:\d{2})",
+    "State Healing": r"✓ STATE HEALING complete: ([\d,]+) state paths healed in (\d+:\d{2}:\d{2})",
+    "Storage Healing": r"✓ STORAGE HEALING complete: ([\d,]+) storage accounts healed in (\d+:\d{2}:\d{2})",
+    "Bytecodes": r"✓ BYTECODES complete: ([\d,]+) bytecodes in (\d+:\d{2}:\d{2})",
+}
+
 
 @dataclass
 class Instance:
@@ -262,6 +274,30 @@ def rpc_call(url: str, method: str) -> Optional[Any]:
         return None
 
 
+def parse_phase_timings(run_id: str, container: str) -> list[tuple[str, str, str]]:
+    """Parse phase completion times from saved container logs.
+
+    Returns list of (phase_name, count, duration) tuples.
+    """
+    log_file = LOGS_DIR / f"run_{run_id}" / f"{container}.log"
+    if not log_file.exists():
+        return []
+
+    try:
+        logs = log_file.read_text()
+    except Exception:
+        return []
+
+    phases = []
+    for phase_name, pattern in PHASE_COMPLETION_PATTERNS.items():
+        match = re.search(pattern, logs)
+        if match:
+            count = match.group(1)
+            duration = match.group(2)
+            phases.append((phase_name, count, duration))
+    return phases
+
+
 def slack_notify(run_id: str, run_count: int, instances: list, hostname: str, branch: str, commit: str, build_profile: str = ""):
     """Send a single summary Slack message for the run."""
     all_success = all(i.status == "success" for i in instances)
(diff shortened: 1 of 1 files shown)


crates/common/trie/node/branch.rs
```diff
@@ -448,7 +452,7 @@ mod test {
             .unwrap();
 
         assert_eq!(new_node.choices, node.choices);
-        assert_eq!(new_node.value, value);
+        assert_eq!(new_node.value, Into::<ValueRLP>::into(value));
```
azteca1998:
```suggestion
        assert_eq!(new_node.value, ValueRLP::from(value));
```

azteca1998 (approved):

crates/common/trie/node/branch.rs
pablodeymo: Thanks! Applied.

pablodeymo:
Regarding the suggested comment for `Bytes::copy_from_slice`: skipping this because the suggested comment ("zero-copy conversion") would be misleading. `copy_from_slice` does copy the data - the zero-copy benefit comes later when cloning `Bytes` (O(1) ref-count increment). The code is self-explanatory in this context.

pablodeymo:
Phase Breakdown (from runs 103–113, perf/avoid-copying-trie-leaves)

```  ┌───────────────────┬──────────┬───────────┬─────────────────────┐
  │       Phase       │ Avg Time │ % of Sync │        Trend        │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ Block Headers     │ ~16 min  │ 5%        │ Stable              │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ Account Ranges    │ ~4 min   │ 1%        │ Stable              │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ Account Insertion │ ~22 min  │ 7%        │ Stable              │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ Storage Ranges    │ ~16 min  │ 5%        │ Stable              │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ Storage Insertion │ ~1h 2m   │ 21%       │ Stable              │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ State Healing     │ ~1 min   │ <1%       │ Negligible          │
  ├───────────────────┼──────────┼───────────┼─────────────────────┤
  │ Storage Healing   │ ~1h 41m  │ 34%       │ Dominant bottleneck │
  └───────────────────┴──────────┴───────────┴─────────────────────┘
```

  Key Observations

  1. Storage Healing is still the # 1 bottleneck — ~1h 41m average, consuming 34% of total sync time.
  This confirms our earlier analysis and the optimization plan in STORAGE_HEALING_OPTIMIZATION_PLAN.md.
  2. Storage Insertion is # 2 — ~1h 2m average (21%). Together with Storage Healing, storage accounts for
   55% of sync time.
  3. No measurable difference from Bytes optimization — perf/avoid-copying-trie-leaves (runs 103–113)
  shows roughly the same times as the baseline branches. The Bytes change only affected state.rs
  (healing path), and state healing is already <1 min, so this is expected.
  4. Bytecodes phase is missing from all breakdowns — the regex might not be matching, or bytecodes
  complete too fast to register.
  5. Run # 99 timeout (8h on background-header-download) was a one-off — subsequent runs on the same
  commit are fine.
  6. Total sync time variance: 4h 39m to 6h 26m — likely driven by network conditions and peer quality.

crates/storage/store.rs
```diff
@@ -1225,7 +1224,7 @@ impl Store {
 
         tokio::task::spawn_blocking(move || {
             let txn = backend.begin_read()?;
-            txn.get(table, &key)
+            txn.get(table, &key).map(|opt| opt.map(|b| b.to_vec()))
```
ElFantasma, line 1227: This `.to_vec()` introduces an extra copy for every non-trie read compared to the previous code. Before: RocksDB returned `Vec<u8>` directly. Now: RocksDB returns `Vec<u8>`, wraps in `Bytes` (zero-copy), then `.to_vec()` copies it back. Net effect is one additional allocation for all callers of `read_async`.

Same applies to `read()` below at line 1238.

This is fine as a transitional step if the plan is to eventually propagate `Bytes` to callers, but worth acknowledging as a minor regression on the non-trie read path.

crates/storage/backend/in_memory.rs
```diff
@@ -6,7 +7,7 @@ use std::collections::BTreeMap;
 use std::path::Path;
 use std::sync::{Arc, RwLock};
 
-type Table = BTreeMap<Vec<u8>, Vec<u8>>;
+type Table = BTreeMap<Bytes, Bytes>;
```
ElFantasma, line 10: nit: Using `Bytes` for BTreeMap *keys* adds atomic refcount overhead that's never useful (keys are never shared/cloned for zero-copy benefit). Consider keeping keys as `Vec<u8>` and only using `Bytes` for values: `BTreeMap<Vec<u8>, Bytes>`. Very minor since in-memory is testing-only.

crates/common/trie/trie.rs
```diff
@@ -224,6 +226,8 @@ impl Trie {
     /// the cached nodes.
     pub fn commit(&mut self) -> Result<(), TrieError> {
         let acc = self.commit_without_storing();
+        // Zero-copy conversion: Bytes::from(Vec<u8>) takes ownership
+        let acc = acc.into_iter().map(|(k, v)| (k, v.into())).collect();
```
ElFantasma, line 230: nit: This `.collect()` allocates a new `Vec` just to convert `Vec<u8>` → `Bytes`. Since `commit_without_storing()` already builds the vec internally, you could have it return `Vec<(Nibbles, Bytes)>` directly (doing the `.into()` inside), avoiding this intermediate collection. Alternatively, `put_batch` could accept an iterator.

crates/storage/trie.rs
```diff
@@ -97,17 +98,18 @@ impl TrieDB for BackendTrieDB {
             TrieError::DbError(anyhow::anyhow!("Failed to begin read transaction: {}", e))
         })?;
         tx.get(table, prefixed_key.as_ref())
+            .map(|opt| opt.map(|b| b.to_vec()))
```
ElFantasma, line 101: The `.to_vec()` here converts the `Bytes` from `StorageReadView::get` back to `Vec<u8>` because `TrieDB::get` returns `Option<Vec<u8>>`. This means the zero-copy RocksDB read benefit doesn't reach the trie — every trie node read still allocates. To realize the "zero-copy from RocksDB" benefit claimed in the PR description, `TrieDB::get` would need to return `Bytes` (or `Cow<[u8]>`), which is a bigger refactor but would be the logical next step.

pablodeymo:
close, due to lack of benefits

PR #6059, opened 2026-01-28
perf(l1): parallelize header download with state download during snap sync

pablodeymo:
## Summary

- Moves header downloading to a background task that runs in parallel with state download
- State download can now begin immediately instead of waiting for all headers to be downloaded
- Adds incremental header processing at strategic points during snap_sync()
- Fixes block_hashes ordering bug by replacing `Vec<H256>` with `BTreeMap<u64, H256>`

## Motivation

Previously, `sync_cycle_snap()` downloaded ALL block headers before starting state download. This caused unnecessary delays since:
1. State download is independent of header download
2. Pivot selection only needs recent headers, not all of them
3. For mainnet with millions of blocks, header download alone takes significant time

Additionally, the `block_hashes: Vec<H256>` field stored hashes without block numbers. The `numbers_and_hashes` construction for `forkchoice_update` inferred block numbers from position (`pivot_header.number - i`). When the background header download and `update_pivot` interleaved inserts, entries ended up out of order, causing `forkchoice_update` to write wrong canonical hashes.

## Implementation

- Add `header_receiver` and `download_complete` fields to `SnapBlockSyncState`
- Create `download_headers_background()` function that runs header download in a separate tokio task
- Modify `sync_cycle_snap()` to spawn the background task and start state download immediately
- Add `process_pending_headers()` helper to incrementally process headers at strategic points
- `snap_sync()` waits for initial headers only when needed for pivot selection
- Replace `block_hashes: Vec<H256>` with `BTreeMap<u64, H256>` keyed by block number — this naturally handles ordering and deduplicates overlapping ranges from `update_pivot` re-inserting the same block numbers
- Build `numbers_and_hashes` directly from BTreeMap keys instead of inferring from position

## Test plan

- [ ] Run snap sync on testnet and verify parallel execution via logs
- [ ] Verify state download starts before header download completes
- [ ] Check memory usage to ensure channel doesn't buffer too many headers
- [ ] Compare sync time before/after the change
- [ ] Verify forkchoice_update writes correct canonical hashes after snap sync


pablodeymo:
Regarding the comment on line 179-181 about "potential panic":

This is a false positive. The code is correct:

```rust
let current_head = store
    .get_block_header(current_head_number)?
    .map(|h| h.hash())
    .unwrap_or(first_block_hash);
```

1. The `?` operator propagates any `Result` error (database failure, etc.)
2. After `?`, we have `Option<BlockHeader>`
3. `.map(|h| h.hash())` transforms `Option<BlockHeader>` → `Option<H256>`. Calling `.map()` on `None` returns `None`, it does not panic.
4. `.unwrap_or(first_block_hash)` provides the fallback when the header is `None`

There is no panic path here - `Option::map` is safe on `None`.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -157,27 +173,34 @@ pub async fn sync_cycle_snap(
...
             None => continue,
         };
-        // TODO(#2126): This is just a temporary solution to avoid a bug where the sync would get stuck
-        // on a loop when the target head is not found, i.e. on a reorg with a side-chain.
+
+        // Handle reorg case where sync head is not reachable
+        let current_head = store
+            .get_block_header(current_head_number)?
+            .map(|h| h.hash())
+            .unwrap_or(first_block_hash);
```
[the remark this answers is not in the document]
pablodeymo: Regarding the comment on line 179-181 about "potential panic":

This is a false positive. The code is correct:

```rust
let current_head = store
    .get_block_header(current_head_number)?
    .map(|h| h.hash())
    .unwrap_or(first_block_hash);
```

    The ? operator propagates any Result error (database failure, etc.)
    After ?, we have Option<BlockHeader>
    .map(|h| h.hash()) transforms Option<BlockHeader> → Option<H256>. Calling .map() on None returns None, it does not panic.
    .unwrap_or(first_block_hash) provides the fallback when the header is None

There is no panic path here - Option::map is safe on None.

pablodeymo:
## Addressing Review Feedback: Early Abort When Switching to Full Sync

The review correctly identified that when the background task switches to full sync mode (line 229), the main task continues with `snap_sync()` unnecessarily, wasting resources.

### Changes Made

**Added early abort mechanism:**

1. New helper function `should_abort_snap_sync()` checks if `snap_enabled` has been set to `false` by the background task

2. Modified `snap_sync()` signature to accept `snap_enabled: &Arc<AtomicBool>`

3. Added abort checks at 4 strategic points in `snap_sync()`:
   - Inside the initial headers wait loop
   - After `request_account_range` completes
   - Inside the storage range download loop
   - Inside the healing loop

When an abort is detected, `snap_sync()` logs "Snap sync aborted: switching to full sync" and returns `Ok(())` early, allowing the background full sync to proceed without wasted work.

### Rationale

Previously, if the background task detected that full sync was needed (e.g., we are close to chain head), it would:
1. Set `snap_enabled = false`
2. Start full sync

But the main task would continue the entire `snap_sync()` (downloading accounts, storages, healing) before finally checking `snap_enabled` after `snap_sync()` returned. This could waste significant time and resources.

Now, `snap_sync()` periodically checks the flag and exits early when full sync is triggered, ensuring resources are not wasted on work that will be discarded.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -195,50 +218,144 @@ pub async fn sync_cycle_snap(
...
+        }
+        Err(e) => {
+            warn!("Background header download task panicked: {e}");
+        }
+    }
+
+    // Sync the snap_enabled state back from the Arc
+    if !snap_enabled_arc.load(Ordering::Relaxed) {
+        snap_enabled.store(false, Ordering::Relaxed);
+    }
```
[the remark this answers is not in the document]
pablodeymo: Addressing Review Feedback: Early Abort When Switching to Full Sync

The review correctly identified that when the background task switches to full sync mode (line 229), the main task continues with snap_sync() unnecessarily, wasting resources.
Changes Made

Added early abort mechanism:

    New helper function should_abort_snap_sync() checks if snap_enabled has been set to false by the background task

    Modified snap_sync() signature to accept snap_enabled: &Arc<AtomicBool>

    Added abort checks at 4 strategic points in snap_sync():
        Inside the initial headers wait loop
        After request_account_range completes
        Inside the storage range download loop
        Inside the healing loop

When an abort is detected, snap_sync() logs "Snap sync aborted: switching to full sync" and returns Ok(()) early, allowing the background full sync to proceed without wasted work.
Rationale

Previously, if the background task detected that full sync was needed (e.g., we are close to chain head), it would:

    Set snap_enabled = false
    Start full sync

But the main task would continue the entire snap_sync() (downloading accounts, storages, healing) before finally checking snap_enabled after snap_sync() returned. This could waste significant time and resources.

Now, snap_sync() periodically checks the flag and exits early when full sync is triggered, ensuring resources are not wasted on work that will be discarded.

pablodeymo:
## Addressing Review Feedback: `is_download_complete()` Semantics

The review noted that `is_download_complete()` returning `true` when `download_complete` is `None` could be misleading, since no download was actually started.

### Solution

Replaced the boolean method with a `DownloadStatus` enum that explicitly represents all three states:

```rust
pub enum DownloadStatus {
    NotStarted,  // No background download task was configured
    InProgress,  // Background task is still downloading headers
    Complete,    // Background task has finished
}
```

The enum includes an `is_terminal()` helper method that returns `true` for both `NotStarted` and `Complete` states.

### Result

The call site now reads:
```rust
// Before (ambiguous - what does true mean when None?)
if block_sync_state.is_download_complete() { ... }

// After (explicit about what we are checking)
if block_sync_state.download_status().is_terminal() { ... }
```

This makes the semantics explicit: "stop waiting if no more headers are expected (either because download finished or was never started)."

pablodeymo:
## Addressing Review Feedback: Busy-wait Loop

The review noted that the 100ms sleep loop for waiting on initial headers was inefficient.

### Solution

Added a new `recv_headers_timeout()` method that uses `tokio::time::timeout` to block until either:
- Headers arrive (wakes immediately)
- Timeout elapses (allows checking abort conditions)

### Before
```rust
process_pending_headers(block_sync_state).await?;  // try_recv()
if block_sync_state.block_hashes.is_empty() {
    tokio::time::sleep(Duration::from_millis(100)).await;  // always waits 100ms
}
```

### After
```rust
if let Some(headers) = block_sync_state
    .recv_headers_timeout(Duration::from_millis(100))
    .await
    && !headers.is_empty()
{
    block_sync_state.process_incoming_headers(headers.into_iter().skip(1)).await?;
}
```

**Benefit:** The loop now wakes immediately when headers arrive instead of waiting for the next poll cycle, while still checking abort conditions every 100ms if no headers are received.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -195,50 +218,144 @@ pub async fn sync_cycle_snap(
...
-        // Discard the first header as we already have it
-        if block_headers.len() > 1 {
-            let block_headers_iter = block_headers.into_iter().skip(1);
-
-            block_sync_state
-                .process_incoming_headers(block_headers_iter)
-                .await?;
+        // Send headers through channel (skip the first as we already have it)
+        if block_headers.len() > 1 && header_sender.send(block_headers).await.is_err() {
+            debug!("Background: Header receiver dropped, stopping download");
```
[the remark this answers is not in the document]
pablodeymo: I'll leave this as is, since I think it will be too spammy otherwise.

pablodeymo:
```
  Last 24h Runs (branch: feature/background-header-download, commit 441634d01)
  ┌─────┬──────────────┬─────────┬─────────┬─────────────┬────────────────────────┐
  │ Run │   Started    │ Network │ Result  │  Sync Time  │    Post-sync Block     │
  ├─────┼──────────────┼─────────┼─────────┼─────────────┼────────────────────────┤
  │ #94 │ Feb 4, 04:30 │ mainnet │ SUCCESS │ 4h 37m      │ 24382518 (+111 blocks) │
  ├─────┼──────────────┼─────────┼─────────┼─────────────┼────────────────────────┤
  │ #95 │ Feb 4, 09:36 │ mainnet │ SUCCESS │ 6h 31m      │ 24384610 (+106 blocks) │
  ├─────┼──────────────┼─────────┼─────────┼─────────────┼────────────────────────┤
  │ #96 │ Feb 4, 16:35 │ mainnet │ SUCCESS │ 4h 40m      │ 24386126 (+110 blocks) │
  ├─────┼──────────────┼─────────┼─────────┼─────────────┼────────────────────────┤
  │ #97 │ Feb 4, 21:43 │ mainnet │ SUCCESS │ 4h 49m      │ 24387704 (+109 blocks) │
  ├─────┼──────────────┼─────────┼─────────┼─────────────┼────────────────────────┤
  │ #98 │ Feb 5, 03:00 │ mainnet │ SUCCESS │ 4h 43m      │ 24389248 (+110 blocks) │
  ├─────┼──────────────┼─────────┼─────────┼─────────────┼────────────────────────┤
  │ #99 │ Feb 5, 08:12 │ mainnet │ FAILED  │ 8h+ timeout │ -                      │
  └─────┴──────────────┴─────────┴─────────┴─────────────┴────────────────────────┘
  ```

**Comment by pablodeymo** [comment]
⏺ Here's the timeline for Run #99:

  Run #99 Timeline (PR #6059 - feature/background-header-download)
  Time: 08:19
  Event: Sync started (mainnet from block 0)
  Elapsed: 0
  ────────────────────────────────────────
  Time: 11:54
  Event: Bytecodes download started
  Elapsed: 3h 35m
  ────────────────────────────────────────
  Time: 13:07
  Event: Bytecodes download finished (last batch: 42,217)
  Elapsed: 4h 48m
  ────────────────────────────────────────
  Time: 13:07
  Event: validate_bytecodes started
  Elapsed: 4h 48m
  ────────────────────────────────────────
  Time: 13:44
  Event: First sync cycle finished (19,490s = 5h 25m total snap sync)
  Elapsed: 5h 25m
  ────────────────────────────────────────
  Time: 13:44-16:18
  Event: Catch-up sync cycles (~4 min each, continuously)
  Elapsed: 5h 25m - 8h+
  ────────────────────────────────────────
  Time: ~16:18
  Event: 8h timeout hit — monitor killed it
  Elapsed: 8h 6m
  Root Cause

  The snap sync itself completed in 5h 25m (within normal range — previous runs were 4h
  26m-6h 31m). The problem is what happened after:

  - After sync completed, the node entered catch-up mode doing full sync cycles every ~4
  minutes
  - It kept doing these cycles for 2.5+ hours without the monitor detecting it as "done"
  - Meanwhile, the background header download feature was still downloading headers in
  parallel (log ends showing headers being fetched from block ~24.3M downward)
  - The node never transitioned to the "processing new blocks" state that the monitor expects
   within 22 minutes

  Comparison with Successful Runs

  In successful runs #92-#98, after snap sync the node caught up and started processing new
  blocks within ~22 minutes. In #99, it was stuck in catch-up sync cycles for 2.5h+. This
  could be:

  1. The background header download competing for resources/bandwidth with the catch-up sync,
   slowing it down
  2. A larger gap to catch up (the sync took 5h25m vs typical 4h35m, meaning the chain
  advanced further during sync)
  3. Network conditions — many WARN about invalid/empty headers from peers throughout the
  tail of the log

File: crates/networking/p2p/sync/snap_sync.rs [hunk 4]
```diff
@@ -198,61 +267,193 @@ pub async fn sync_cycle_snap(
...
 
         if sync_head_found {
+            debug!("Background: Sync head found, header download complete");
             break;
-        };
+        }
     }
 
-    snap_sync(peers, &store, &mut block_sync_state, datadir).await?;
+    download_complete.store(true, Ordering::Release);
```
ElFantasma: Bug: `download_complete` is only set to `true` in the happy path (here) and in two specific early-returns (max attempts, full-sync switch). But if the function returns via `?` error propagation (e.g., `peers.request_block_headers(...)` fails, `store.add_block_headers(...)` fails, etc.), `download_complete` is never set to `true`.

When this happens, the sender half of the channel is dropped (task exits), but `snap_sync()`'s initial-headers wait loop (line ~435) checks `download_status().is_terminal()` which returns `false` (flag still `false`), so it keeps calling `recv_headers_timeout()` → gets `None` (closed channel) → loops with 100ms sleeps forever.

Suggestion: use an RAII guard pattern to ensure `download_complete` is always set on exit:
```rust
struct DownloadCompleteGuard(Arc<AtomicBool>);
impl Drop for DownloadCompleteGuard {
    fn drop(&mut self) {
        self.0.store(true, Ordering::Release);
    }
}
// At the start of download_headers_background:
let _guard = DownloadCompleteGuard(download_complete.clone());
```
Then remove the manual `download_complete.store(true, ...)` calls.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -160,27 +222,34 @@ pub async fn sync_cycle_snap(
...
         {
-            // There is no path to the sync head this goes back until it find a common ancerstor
-            warn!("Sync failed to find target block header, going back to the previous parent");
-            current_head = first_block_parent_hash;
+            warn!(
+                "Background: Sync failed to find target block header, going back to the previous parent"
+            );
+            // We can't easily go back in the background task, so we just continue with the current head
+            // The update_pivot mechanism will handle this case
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
ElFantasma: The original code handled the reorg case by walking backward: `current_head = first_block_parent_hash`. The background task can't do this because `current_head` was removed in favor of `current_head_number`, and going back requires knowing the parent hash.

But the current code just sleeps 100ms and retries with the same `current_head_number`, which will get the same headers back and loop indefinitely. The comment says "The update_pivot mechanism will handle this case" but `update_pivot` runs in `snap_sync()` on the main task — it doesn't feed back into the background task's `current_head_number`.

Consider either:
1. Decrement `current_head_number` by 1 to walk backward (simpler, analogous to the old `parent_hash` approach)
2. Add a maximum retry count for this specific case
3. Use `_first_block_parent_hash` (currently prefixed with `_`) to look up the parent's block number

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -198,61 +267,193 @@ pub async fn sync_cycle_snap(
...
 
-        // Discard the first header as we already have it
-        if block_headers.len() > 1 {
-            let block_headers_iter = block_headers.into_iter().skip(1);
-
-            block_sync_state
-                .process_incoming_headers(block_headers_iter)
-                .await?;
+        // Send headers through channel (skip the first as we already have it)
+        if block_headers.len() > 1 && header_sender.send(block_headers).await.is_err() {
```
ElFantasma: The first header is skipped by the *receiver* in `process_pending_headers()` (via `.skip(1)`), but here the full batch including the first header is sent. This means every batch sent through the channel carries one header that will be discarded. Not a bug, but it's a subtle contract — the sender and receiver must agree on the "skip first" convention.

More importantly: the original code only skipped the first header of the *first* batch (to avoid the already-known current head). Here, `.skip(1)` is applied to *every* batch received from the channel. If the background task sends multiple batches, the first header of each batch (which is the continuation point from the previous batch's last header) gets silently dropped. This could cause gaps in `block_hashes` — missing one header per batch.

Double-check that the header fetching always returns overlapping ranges (i.e., the first header of batch N+1 == the last header of batch N). If so, the skip is correct. If not, this introduces gaps.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -198,61 +267,193 @@ pub async fn sync_cycle_snap(
...
+
+    // Create channel for header communication between background task and main snap_sync
+    let (header_sender, header_receiver) = tokio::sync::mpsc::channel(100);
+    let download_complete = Arc::new(AtomicBool::new(false));
+
+    // Setup block_sync_state with channel receiver
+    block_sync_state.set_header_channel(header_receiver, download_complete.clone());
+
+    // Create Arc wrapper for snap_enabled so we can share it with the background task
+    let snap_enabled_arc = Arc::new(AtomicBool::new(snap_enabled.load(Ordering::Relaxed)));
```
ElFantasma, line 400: nit: Creating a separate `Arc<AtomicBool>` copy of `snap_enabled` means the `Syncer`'s original `snap_enabled` (shared with `SyncManager`) doesn't see the full-sync switch until `snap_sync()` completes and the sync-back on line ~389 runs. If `SyncManager` or other code checks `snap_enabled` concurrently during the sync cycle, it'll still see `true` even though the background task decided to switch to full sync.

In practice this is probably fine since the sync cycle is the only writer, but it's worth a comment explaining why a copy is used instead of sharing the original Arc.

pablodeymo:
Attaching the investigation of the bug:
[BACKGROUND_HEADER_DOWNLOAD_INVESTIGATION.md](https://github.com/user-attachments/files/25218170/BACKGROUND_HEADER_DOWNLOAD_INVESTIGATION.md)

ElFantasma (changes requested):

ElFantasma:
## Review feedback addressed in 137bfd55b

Resolves conflicts with `main` and addresses the unresolved review comments. Summary by reviewer:

### From @ElFantasma
- ✅ **L315 — `download_complete` not set on `?` paths**: replaced manual flag-sets with an RAII `CompletionGuard` that flips the flag on Drop. Every return path (happy, `?`, panic unwind) now signals completion.
- ✅ **L247 — reorg detection broken**: `current_head` is now tracked as a hash locally in the background loop, mirroring main's original walk-back-via-`first_block_parent_hash` recovery. No longer depends on a store lookup that fails once we fetch past the local canonical head.
- ✅ **L304 — first-header skip contract mismatch**: the sender now strips the duplicate first header from each batch via `split_off(1)`; receivers no longer call `.skip(1)`. Single source of truth for the channel contract.
- ⚠️ **L355 — `Arc<AtomicBool>` copy of `snap_enabled` nit**: not addressed. Tying `Syncer::snap_enabled` and the background task to a single shared `Arc` would change `sync_cycle_snap`'s signature into the caller (`Syncer::sync_cycle`); happy to do this in a follow-up if you want.

### From @Copilot
- ✅ **L297 — full sync running concurrently in background task**: the background task no longer calls `sync_cycle_full`. It sets `snap_enabled = false` + signals completion and returns. `Syncer::sync_cycle` runs full sync on the next cycle. Eliminates Store/peer concurrency between snap and full.
- ✅ **L236 — reorg detection vs. unpersisted headers**: now persists each batch to the store (`add_block_headers` + `set_header_download_checkpoint`) before sending via channel. Local `current_head_hash` covers the case where the store read can't be trusted (e.g. canonical-head divergence).
- ✅ **L458 — mode-switch ambiguity**: with the previous fix, the background task no longer returns through a path that could be confused with normal completion. `snap_sync()` aborting on `snap_enabled=false` is the single signal, and `sync_cycle_snap` cleans up + returns; the caller routes to full sync on the next cycle. No event mode confusion.
- ✅ **`snap_server_tests.rs:294`**: grammar fix applied.

### From @greptile-apps
- ✅ **L348 — channel buffer 100 hardcoded**: extracted to `HEADER_CHANNEL_BUFFER` constant with documenting comment.
- ✅ **L93 — `is_download_complete()` semantics**: superseded by the `DownloadStatus` enum (NotStarted / InProgress / Complete) which makes the "no download started" case explicit. `is_terminal()` documents that both terminal states satisfy "no more headers expected."
- ✅ **L471 — busy-wait loop**: superseded by `recv_headers_timeout()` which uses `tokio::time::timeout(receiver.recv())` — proper async wait, not a 100ms-sleep loop.
- ℹ️ **L305 — silent send error**: left as-is per the prior reply (would be too spammy).

Code is `cargo check` / `cargo fmt` / `cargo clippy` clean.

PR #6060, opened 2026-01-28
fix(l1): auto-switch from snap to full sync when node has synced state

ilitteri:
## Motivation

The node starts snap sync regardless of existing state when using `--syncmode snap` (the default). The check for existing blocks only happened **inside** the snap sync cycle, causing the node to incorrectly report `SYNCING` to the CL before realizing it should use full sync.

## Description

Add state detection at startup in `SyncManager::new()` that automatically switches from snap to full sync when the node already has synced state.

A node is considered "synced" if:
1. It has blocks beyond the merge fork (post-merge state exists)
2. AND there's no ongoing snap sync (no `header_download_checkpoint`)

The `merge_netsplit_block` threshold is used to avoid false positives in hive tests that start with pre-merge blocks. For post-merge networks where `merge_netsplit_block` is 0, any block > 0 indicates prior sync.

### Network behavior

| Network | `merge_netsplit_block` | Behavior |
|---------|------------------------|----------|
| Mainnet | 15,537,394 | Pre-merge blocks allow snap sync, post-merge triggers full sync |
| Sepolia | ~1,450,409 | Same as mainnet |
| Holesky | 0 (post-merge genesis) | Any block > 0 triggers full sync |
| Devnets | Often 0 | Same as Holesky |

## How to Test

1. Run hive sync tests: `make run-hive SIMULATION=ethereum/sync`
2. Manual test scenarios:
   - Fresh node with `--syncmode snap` → should use snap sync
   - Node with existing blocks + `--syncmode snap` → should auto-switch to full sync
   - Node mid-snap-sync (has checkpoint) → should resume snap sync

### DIFF
--- a/crates/networking/p2p/sync_manager.rs
+++ b/crates/networking/p2p/sync_manager.rs
@@ -43,31 +43,32 @@ impl SyncManager {
     ) -> Self {
         let snap_enabled = Arc::new(AtomicBool::new(matches!(sync_mode, SyncMode::Snap)));
 
+        // Fetch checkpoint once to avoid duplicate DB reads
+        let has_checkpoint = store
+            .get_header_download_checkpoint()
+            .await
+            .unwrap_or_else(|e| {
+                warn!("Failed to read header download checkpoint: {e}");
+                None
+            })
+            .is_some();
+
         // Auto-switch from snap to full sync if node already has synced state.
-        // A node is considered "synced" if:
-        // 1. It has blocks beyond the merge fork (post-merge state exists)
-        // 2. AND there's no ongoing snap sync (no header_download_checkpoint)
-        //
-        // The merge_netsplit_block threshold is used to avoid false positives in
-        // hive tests that start with pre-merge blocks. For post-merge networks
+        // A node is considered "synced" if it has blocks beyond the merge fork AND
+        // there's no ongoing snap sync. The merge_netsplit_block threshold avoids
+        // false positives in hive tests with pre-merge blocks. For post-merge networks
         // where merge_netsplit_block is 0, any block > 0 indicates prior sync.
-        if snap_enabled.load(Ordering::Relaxed) {
-            let has_checkpoint = store
-                .get_header_download_checkpoint()
-                .await
-                .is_ok_and(|res| res.is_some());
-
-            if !has_checkpoint {
-                let latest_block = store.get_latest_block_number().await.unwrap_or(0);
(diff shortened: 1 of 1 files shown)


iovoid:
Our genesis files don't set mergeNetsplitBlock for mainnet, only for sepolia (to 1735371, not the value the table says) and hoodi (0). Do you think this could cause problems?

crates/networking/p2p/sync_manager.rs
```diff
@@ -42,6 +42,36 @@ impl SyncManager {
...
+            .await
+            .unwrap_or_else(|e| {
+                warn!("Failed to read header download checkpoint: {e}");
+                None
+            })
+            .is_some();
+
+        // Auto-switch from snap to full sync if node already has synced state.
+        // A node is considered "synced" if it has blocks beyond the merge fork AND
+        // there's no ongoing snap sync. The merge_netsplit_block threshold avoids
```
iovoid, line 58: Considering we can't have a pre-merge block as latest (since we can't execute it), maybe explain why a simpler (block>0) check wouldn't work?

ElFantasma (approved):

crates/networking/p2p/sync_manager.rs
```diff
@@ -42,6 +42,36 @@ impl SyncManager {
...
+        // false positives in hive tests with pre-merge blocks. For post-merge networks
+        // where merge_netsplit_block is 0, any block > 0 indicates prior sync.
+        // Skip auto-switch if merge_netsplit_block is None (custom genesis).
```
Arkenan, line 61: I wouldn't use the word "synced". If we previously snap synced a node, but then we had a day of downtime and start up the node again, we may be behind tens of thousands of blocks. Therefore, we are not synced. However, we DO have a valid state, so we don't need to do snap sync, we just do full sync with the missing blocks.

```suggestion
        // Auto-switch from snap to full sync if node already has valid state.
        // A node is considered to have valid state if it has executed blocks beyond 
        // the merge fork AND there's no ongoing snap sync. 
        //
        // The merge_netsplit_block threshold prevents false positives in hive tests 
        // with pre-merge blocks. For post-merge networks where merge_netsplit_block
        // is 0, any block > 0 indicates prior sync.
        // 
        // Skip auto-switch if merge_netsplit_block is None (custom genesis).
```

Arkenan (approved):
LGTM with a comment on a comment.

PR #6063, opened 2026-01-29
perf(l1): use VecDeque in TransactionQueue for O(1) front removal

VolodymyrBg:
**Motivation**

`TransactionQueue` used `Vec` with `remove(0)` operations in `pop()` and `shift()` methods. This caused `O(n)` complexity for each front removal because `Vec` shifts all remaining elements. With mempool sizes up to 10,000 transactions, this became a performance bottleneck during block building.

**Description**

- Replace `Vec<HeadTransaction>` with `VecDeque<HeadTransaction>` for the heads list
- Replace `HashMap<Address, Vec<MempoolTransaction>>` with `HashMap<Address, VecDeque<MempoolTransaction>>` for per-sender tx lists
- Update `pop()` and `shift()` to use `pop_front()` instead of `remove(0)`
- Update `peek()` to use `front()` instead of `first()`

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -1,6 +1,6 @@
 use std::{
     cmp::{Ordering, max},
-    collections::HashMap,
+    collections::{HashMap, VecDeque},
     ops::Div,
     sync::Arc,
     time::{Duration, Instant},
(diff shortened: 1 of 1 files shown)


VolodymyrBg:
Result:

Before change: 
image
After change: 
image

Arkenan:
Hi! Thanks for the PR, but considering that the goal of this is to have perf improvements and seeing that there's a perf regression, we'll be closing this.

PR #6064, opened 2026-01-29
fix(l1): prevent panic from legacy transaction v value overflow

SchnobiTobi:
# Description 
Replace `as_u64()` with `try_into()` in `sender()` and `derive_legacy_chain_id()`
to handle malicious transactions with v > u64::MAX gracefully instead of panicking.

### DIFF
--- a/crates/common/errors.rs
+++ b/crates/common/errors.rs
@@ -18,6 +18,8 @@ pub enum EcdsaError {
     ))]
     #[error("k256 error: {0}")]
     K256(#[from] k256::ecdsa::Error),
+    #[error("recovery id error: exceeds u64::MAX")]
+    RecoveryIdOverflow,
 }
 
 /// Errors that occur during block validation.
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -1045,9 +1045,12 @@ impl Transaction {
     pub fn sender(&self) -> Result<Address, EcdsaError> {
         match self {
             Transaction::LegacyTransaction(tx) => {
+                let v: u64 =
+                    tx.v.try_into()
+                        .map_err(|_| EcdsaError::RecoveryIdOverflow)?;
                 let signature_y_parity = match self.chain_id() {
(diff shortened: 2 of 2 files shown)


crates/common/types/transaction.rs
```diff
@@ -3651,4 +3652,16 @@ mod tests {
...
+    fn test_large_v_value_no_panic() {
+        let tx = LegacyTransaction {
+            v: U256::from(2).pow(U256::from(200)),
+            ..Default::default()
+        };
+        let transaction = Transaction::LegacyTransaction(tx);
+
+        assert!(transaction.chain_id().is_none());
+        assert!(transaction.sender().is_err());
+    }
```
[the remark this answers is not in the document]
SchnobiTobi: done

crates/common/errors.rs
```diff
@@ -18,6 +18,8 @@ pub enum EcdsaError {
     ))]
     #[error("k256 error: {0}")]
     K256(#[from] k256::ecdsa::Error),
+    #[error("invalid signature: v value exceeds u64::MAX")]
+    InvalidSignature,
```
iovoid: The name of the error implies it's generic but the description is specific to a single case.

SchnobiTobi: changed the name to something more specific, lmk if any changes needed 🙂

crates/common/types/transaction.rs
```diff
@@ -1045,9 +1045,12 @@ impl Transaction {
     pub fn sender(&self) -> Result<Address, EcdsaError> {
         match self {
             Transaction::LegacyTransaction(tx) => {
+                let v: u64 =
+                    tx.v.try_into()
```
ElFantasma, line 1049: nit: this `try_into()` makes the subsequent `self.chain_id()` call safe (since it early-returns on overflow), but `chain_id()` → `derive_legacy_chain_id()` also does its own `try_into().ok()?`. So the conversion happens twice for the happy path. Not a real issue — just noting the overlap.

ElFantasma (approved):

PR #6079, opened 2026-01-30
perf(l1): parallelize storage trie merkelization

iovoid:
**Motivation**

Currently we split merkelization by account. However, some accounts have much higher activity than others, and this leads to the merkelization work being split unevenly.

**Description**

First we do all the storage merkelization work, split by storage slot and then after gathering the results we merkelize the accounts.

This only has pipelineing for the storage part, not the accounts.

**Note: this is the fixed version of #5497**

Bug in #5497: load_trie used a "double-open" pattern that opened a storage trie with EMPTY_TRIE_HASH, then read the root node from the DB to discover the real root. Because the backend DB stores trie nodes from all committed blocks, this could return a stale root node from a prior block for accounts whose storage is currently empty. This caused incorrect storage roots to propagate into the state trie, producing a wrong state root hash.

Fix: Replace load_trie with a direct lookup of the account's storage_root from the state trie (which each worker already has), matching what the sequential code path does. This uses the known-correct value instead of guessing it from the DB.


ilitteri:
The CHANGELOG entry references `#5497` but this PR is `#6079`. Should be:

```markdown
- Parallelize storage merkelization [#6079](https://github.com/lambdaclass/ethrex/pull/6079)
```

iovoid:
Fixed PR number.

crates/blockchain/blockchain.rs
```diff
@@ -632,219 +652,202 @@ impl Blockchain {
         ))
     }
 
-    fn handle_merkleization_sequential(
+    fn load_trie(
```
ElFantasma, line 655: Potential issue: `load_trie` still uses double-open pattern

  The `load_trie` function still uses the "double-open" pattern that was identified as the bug in #5497:

  ```rust
  fn load_trie(
      &self,
      parent_header: &BlockHeader,
      prefix: Option<H256>,
  ) -> Result<Trie, StoreError> {
      Ok(match prefix {
          Some(account_hash) => {
              // 1. Opens with EMPTY_TRIE_HASH (wrong root)
              let trie = self.storage.open_storage_trie(
                  account_hash,
                  parent_header.state_root,
                  *EMPTY_TRIE_HASH,
              )?;
              // 2. Reads root node from DB and hashes it to discover actual root
              let root = trie
                  .db()
                  .get(Nibbles::default())?
                  .filter(|rlp| !rlp.is_empty())
                  .map(keccak)
                  .unwrap_or(*EMPTY_TRIE_HASH);
              // 3. Opens again with discovered root
              self.storage
                  .open_storage_trie(account_hash, parent_header.state_root, root)?
          }
          None => self.storage.open_state_trie(parent_header.state_root)?,
      })
  }
```
  Why this is problematic

  The PR description states:

  Bug in #5497: `load_trie` used a "double-open" pattern that opened a storage trie with `EMPTY_TRIE_HASH`, then read the root node from the DB to discover the real root. Because the backend DB stores trie nodes from all committed blocks, this could return a stale root node from a prior block for accounts whose storage is currently empty.

  This same pattern is still present in the new load_trie function.

  Context

  `load_trie` is called from `collapse_root_node` when we need to load an unchanged child node (a `NodeRef::Hash`) from the parent state. While this might work correctly in practice because we're loading nodes that weren't modified, using the same pattern that caused the original bug seems risky and inconsistent with the fix applied elsewhere in this PR.

iovoid: Agreed, let's avoid the risk. Changed in e810675ec

ElFantasma (approved):
LGTM, I guess greptile suggestions are probably unnecessary defensive programming checks. (worth double 
checking, anyway)

PR #6095, opened 2026-01-31
chore(l1, l2): enable `clippy::unused_async` lint

### LINKED ISSUE #5662: Enable `clippy::unused_async` lint
Related to #5012

We should enable the [clippy::unused_async lint](https://rust-lang.github.io/rust-clippy/master/index.html#unused_async), which flags functions marked `async` but without any internal `await` calls. Functions like that can cause problems due to them being perceived as functions safe to call from an `async` context, while they're really sync functions that can block the tokio runtime for long periods of time.

akronim26:
**Motivation**

**Description**

This PR enables the `clippy::unused_async` lint in CI to prevent functions from being marked async without awaiting, which can unintentionally block the Tokio runtime.

Resolves #5662

**Checklist**

Closes #5662

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ help: ## 📚 Show help for each of the Makefile recipes
 # Frame pointers for profiling (default off, set FRAME_POINTERS=1 to enable)
 FRAME_POINTERS ?= 0
 
+CLIPPY_DENY := -D warnings -D clippy::unused_async
+
 ifeq ($(FRAME_POINTERS),1)
 PROFILING_CFG := --config .cargo/profiling.toml
 endif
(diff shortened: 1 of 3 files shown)


akronim26:
There were a few files which failed the `unused_async` lint, should I also update them in this PR?

Makefile
```diff
@@ -17,17 +17,17 @@ build: ## 🔨 Build the client
 
 lint-l1:
 	cargo clippy --lib --bins -F debug,sync-test \
-		--release -- -D warnings
+		--release -- -D warnings -D clippy::unused_async
```
ElFantasma: nit: An alternative to repeating `-D clippy::unused_async` in each target is using a shared variable:

```makefile
CLIPPY_DENY := -D warnings -D clippy::unused_async
```

then referencing `$(CLIPPY_DENY)` in each target. This would make it easier to add more lints in the future without touching three lines. Not blocking — the PR is fine as-is for a single lint.

ElFantasma (approved):
CI also runs standalone cargo clippy calls outside the Makefile — pr-main_l1.yaml:84 runs cargo clippy -- -D warnings and pr-main_l2.yaml:105 runs cargo clippy --workspace --features l2,l2-sql -- -D warnings (before calling make lint). These standalone clippy invocations won't enforce unused_async. Might be worth adding the lint there too for full coverage, or alternatively consolidating CI to only use the Makefile targets.

ElFantasma (approved):

akronim26:
Hi @ElFantasma, just following up on this. Is there anything else needed from my side to get it merged?

MegaRedHand:
The CI is failing due to the new lint. Can you fix the clippy errors?

PR #6097, opened 2026-02-02
fix(l2): correctly update in-memory operations counts

GarmashAlex:
**Motivation**

The in-memory rollup store did not actually update the operations counters. The `update_operations_count` method modified a local copy of the operations_counts array instead of the field inside StoreInner, so `get_operations_count` always returned [0, 0, 0] regardless of how many operations were processed. This also made the in-memory backend diverge from the SQL implementation, which persists and increments these counters correctly.

**Description**

Fixed `update_operations_count` in the in-memory store to mutate `StoreInner::operations_counts` directly under the mutex guard. As a result, subsequent calls to get_operations_count now return the accumulated numbers of transactions, privileged transactions, and messages, matching the behavior of the SQL backend and making metrics based on these values accurate when using the in-memory engine.

### DIFF
--- a/crates/l2/storage/src/store_db/in_memory.rs
+++ b/crates/l2/storage/src/store_db/in_memory.rs
@@ -197,10 +197,10 @@ impl StoreEngineRollup for Store {
         privileged_tx_inc: u64,
         messages_inc: u64,
     ) -> Result<(), RollupStoreError> {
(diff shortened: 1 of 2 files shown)


ElFantasma (commented):
Was already submitted by another contributor: #5742.

This one got better location for the test, though

PR #5975, opened 2026-01-21
refactor(l1): snap sync code reorganization and error handling consolidation

pablodeymo:
## Summary

Major refactoring of the snap sync implementation (~6,500 lines across 7 files) to improve code organization, maintainability, and error handling. This PR completes all 5 phases of the snap sync refactoring plan.

## Motivation

The snap sync codebase had grown organically with:
- Large monolithic files (sync.rs at 1,648 lines, peer_handler.rs at 2,074 lines)
- Scattered error types across multiple modules
- Mixed concerns (server/client code, protocol messages, sync orchestration)
- Inconsistent naming conventions in healing modules

This refactoring addresses these issues through a structured, phased approach.

### Phase 1: Foundation ✅
- Created `snap/` module directory structure
- Moved server code from `snap.rs` to `snap/server.rs`
- Created `snap/constants.rs` with documented protocol constants
- Moved snap server tests to root test crate (`test/tests/p2p/snap_server_tests.rs`) to avoid `clippy::unwrap_used` lint denial inherited from the `ethrex-p2p` crate

### Phase 2: Protocol Layer ✅
- Split `rlpx/snap.rs` into `rlpx/snap/` directory:
  - `messages.rs` - Snap protocol message type definitions
  - `codec.rs` - RLPxMessage encoding/decoding implementations
  - `mod.rs` - Module re-exports

### Phase 3: Healing Unification ✅
- Created `sync/healing/` directory with unified structure:
  - `types.rs` - Shared healing types (HealingQueueEntry, StateHealingQueue)
  - `state.rs` - State trie healing (~420 lines)
  - `storage.rs` - Storage trie healing (~530 lines)
- Improved naming conventions:
  - `MembatchEntryValue` → `HealingQueueEntry`
  - `MembatchEntry` → `StorageHealingQueueEntry`
  - `Membatch` → `StorageHealingQueue`
  - `children_not_in_storage_count` → `missing_children_count`

### Phase 4: Sync Orchestration ✅
- Split `sync.rs` (1,648 lines) into focused modules:
  - `sync/full.rs` (~260 lines) - Full sync implementation
  - `sync/snap_sync.rs` (~1,100 lines) - Snap sync implementation
  - `sync/mod.rs` (~285 lines) - Syncer struct and orchestration
- Extracted snap client methods from `peer_handler.rs` to `snap/client.rs`:
  - `request_account_range`, `request_bytecodes`, `request_storage_ranges`
  - `request_state_trienodes`, `request_storage_trienodes`
[...]

PR #6099, opened 2026-02-02
refactor(l1): extract snapshot dumping helpers in snap client

pablodeymo:
## Summary

Extract repeated snapshot dumping logic into helper functions to improve code maintainability and reduce duplication in `snap/client.rs`.

## Motivation

During the snap sync refactoring (PR #5975), a TODO comment was left in the code:

```rust
// TODO: This is repeated code, consider refactoring
```

The same pattern for dumping snapshots to disk appeared **4 times** across two functions:
- `request_account_range`: 2 occurrences (in-loop + after-loop)
- `request_storage_ranges`: 2 occurrences (in-loop + after-loop)

Each occurrence repeated the same logic:
1. Take/collect data from accumulators
2. Check if directory exists, create if not
3. Get the snapshot file path
4. Dump data to file with error handling

This duplication made the code harder to maintain and increased the chance of inconsistencies.

### New Helper Functions

Added 3 helper functions in `snap/client.rs`:

```rust
/// Ensures a directory exists, creating it if necessary
fn ensure_dir_exists(dir: &Path) -> Result<(), SnapError>

/// Prepares and dumps account state snapshot to file
fn dump_account_state_snapshot(
    dir: &Path,
    chunk_index: u64,
    account_hashes: Vec<H256>,
    account_states: Vec<AccountState>,
) -> Result<(), DumpError>

/// Prepares and dumps storage snapshot to file
fn dump_storage_snapshot(
    dir: &Path,
    chunk_index: u64,
    storages: Vec<AccountsWithStorage>,
) -> Result<(), DumpError>
```

### Refactored Call Sites

| Location | Function | Context |
|----------|----------|---------|
| Line ~193 | `request_account_range` | In-loop (buffer full) |
| Line ~306 | `request_account_range` | After-loop (remaining) |
| Line ~620 | `request_storage_ranges` | In-loop (buffer full) |
| Line ~996 | `request_storage_ranges` | After-loop (remaining) |

### Additional Improvements

- Removed the TODO comment
- Added empty check before dumping remaining data (avoids unnecessary file operations)
- Simplified variable names (`current_account_hashes` → `hashes`)
- Consistent error handling using `SnapError::from`

## Code Diff Summary

```
1 file changed, 67 insertions(+), 71 deletions(-)
```

Net reduction of 4 lines, but more importantly:
- Single source of truth for snapshot dumping logic
- Easier to test helper functions in isolation
- Clearer intent from function names

## Testing

- [x] `cargo clippy -p ethrex-p2p` - no warnings
- [x] `cargo test -p ethrex-p2p` - all 12 tests pass

## Related

- Parent PR: #5975 (snap sync refactoring)
- Addresses TODO comment added in commit `68867ae13`

### DIFF
--- a/crates/networking/p2p/snap/client.rs
+++ b/crates/networking/p2p/snap/client.rs
@@ -81,6 +81,43 @@ struct StorageTask {
     end_hash: Option<H256>,
 }
 
+// =============================================================================
+// Snapshot Dumping Helpers
+// =============================================================================
+
+/// Ensures a directory exists, creating it if necessary
+fn ensure_dir_exists(dir: &Path) -> Result<(), SnapError> {
+    if !dir.exists() {
+        std::fs::create_dir_all(dir).map_err(|_| SnapError::dir_create_failed(dir.to_path_buf()))?;
+    }
+    Ok(())
+}
+
+/// Prepares and dumps account state snapshot to file
+fn dump_account_state_snapshot(
+    dir: &Path,
+    chunk_index: u64,
+    account_hashes: Vec<H256>,
+    account_states: Vec<AccountState>,
+) -> Result<(), DumpError> {
+    let chunk: Vec<(H256, AccountState)> = account_hashes
+        .into_iter()
+        .zip(account_states)
+        .collect();
+
+    let path = get_account_state_snapshot_file(dir, chunk_index);
+    dump_accounts_to_file(&path, chunk)
+}
+
+/// Prepares and dumps storage snapshot to file
+fn dump_storage_snapshot(
+    dir: &Path,
+    chunk_index: u64,
+    storages: Vec<AccountsWithStorage>,
+) -> Result<(), DumpError> {
+    let path = get_account_storages_snapshot_file(dir, chunk_index);
+    dump_storages_to_file(&path, storages)
+}
 
 /// Snap sync client methods for PeerHandler
 impl PeerHandler {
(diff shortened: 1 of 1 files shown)


azteca1998 (approved):

crates/networking/p2p/p2p.rs
```diff
@@ -74,7 +74,7 @@ pub(crate) mod metrics;
 pub mod network;
 pub mod peer_handler;
 pub mod rlpx;
-pub(crate) mod snap;
+pub mod snap;
```
ElFantasma, line 77: nit: This was `pub(crate)` before — why widen to `pub`? The snap module contains internal protocol implementation details (client request methods, server handlers, error types). Unless an external crate needs to import from `ethrex_p2p::snap`, `pub(crate)` is more appropriate and avoids leaking internal types into the public API.

crates/networking/p2p/snap/client.rs
```diff
@@ -0,0 +1,1392 @@
...
+    // end_hash is None if the task is for the first big storage request
+    end_hash: Option<H256>,
+}
+
+// =============================================================================
+// Snapshot Dumping Helpers
+// =============================================================================
+
+/// Ensures a directory exists, creating it if necessary
+fn ensure_dir_exists(dir: &Path) -> Result<(), SnapError> {
```
ElFantasma, line 89: nit: `dir.exists()` followed by `create_dir_all` is a TOCTOU race — the directory could be removed between the check and the creation. Since `create_dir_all` is a no-op when the directory already exists, you can simplify this to just:
```rust
fn ensure_dir_exists(dir: &Path) -> Result<(), SnapError> {
    std::fs::create_dir_all(dir).map_err(|_| SnapError::dir_create_failed(dir.to_path_buf()))
}
```
This is also slightly cheaper (one syscall instead of two in the happy path).

crates/networking/p2p/snap/client.rs
```diff
@@ -0,0 +1,1392 @@
...
+                                {
+                                    if !old_intervals.is_empty() {
+                                        acc_hash = *account;
+                                    }
+                                } else {
+                                    continue;
+                                }
+                            }
+                            if acc_hash.is_zero() {
+                                panic!("Should have found the account hash");
```
ElFantasma, line 719: Pre-existing, but worth noting: this `panic!` can crash the node during snap sync. The surrounding code already uses `SnapError::InternalError` for similar "shouldn't happen" conditions (e.g., lines 696, 701). Consider replacing with:
```rust
return Err(SnapError::InternalError("Should have found the account hash".to_owned()));
```
This way a bad peer response degrades gracefully instead of taking down the node.

plan_snap_sync.md
```diff
@@ -0,0 +1,284 @@
+# Snap Sync Refactoring Plan
```
ElFantasma, line 1: This development planning document probably shouldn't be committed to the repo root. Options: (a) move it to `docs/snapsync/refactoring-plan.md` if it's useful as long-term documentation, (b) remove it from the PR since its value is captured by the commit history and PR description, or (c) add it to `.gitignore` if it's a local working document.

PR #6103, opened 2026-02-03
test(l1): add storage trie branches reorg test

### LINKED ISSUE #5657: Add test reproducing reth state root mismatch
Reth had a [recent incident](https://laced-king-de5.notion.site/Incident-Post-Mortem-Reth-Mainnet-State-Root-Mismatch-26732f2c348480dea8b8c2a8753696dc#26732f2c3484809489d0c249a67b1acd) when following the chain on mainnet, where one of the storages wasn't reorged out correctly. We want to have a test for this case.

We already made some progress in #4921. The main idea is to have a tree with two branch nodes in a storage trie:

```
// These hash to the same 0x2b prefix, causing a branch in the storage trie
let slot0A = 15865164872778812924_u64;
let slot0B = 11566337052255443646_u64;

// These hash to the same 0x83 prefix, causing a branch in the storage trie
let slot1A = 5800641681667202032_u64;
let slot1B = 8343558256908705664_u64;
```

The reorg should revert this change.

Himess:
Closes #5657


tooling/reorgs/src/main.rs
```diff
@@ -411,3 +412,178 @@ async fn test_storage_slots_reorg(simulator: Arc<Mutex<Simulator>>) {
...
+    node1.send_call(&signer, contract_address, calldata).await;
+    base_chain = node1.build_payload(base_chain).await;
+    node1.notify_new_payload(&base_chain).await;
+    node1.update_forkchoice(&base_chain).await;
+
+    let calldata = [slot1a.to_big_endian(), value_a.to_big_endian()]
+        .concat()
+        .into();
+    node1.send_call(&signer, contract_address, calldata).await;
+    base_chain = node1.build_payload(base_chain).await;
```
ElFantasma, line 504: nit: Both `send_call` invocations on the base chain (node1) use the same `signer`, and `send_call` fetches the nonce from `get_nonce(..., Latest)`. Between these two calls, the first tx isn't mined yet (no `build_payload`/`notify_new_payload` between them) — the nonce only advances when the block is built. Since you do call `build_payload` + `notify_new_payload` + `update_forkchoice` after the *first* `send_call` before calling the *second*, the nonce will indeed be incremented. Just confirming this is intentional and correct — the pattern is fine.

tooling/reorgs/src/main.rs
```diff
@@ -411,3 +412,178 @@ async fn test_storage_slots_reorg(simulator: Arc<Mutex<Simulator>>) {
...
+        node1.get_storage_at(contract_address, slot1a).await,
+        value_a
+    );
+    assert_eq!(
+        node1.get_storage_at(contract_address, slot1b).await,
+        U256::zero()
+    );
+
+    // Make base chain longer so it becomes canonical
+    for _ in 0..5 {
```
ElFantasma, line 562: nit: The base chain ends up 7 blocks ahead of the fork point (2 tx blocks + 5 empty) while the side chain is only 2 blocks ahead. The asymmetry is fine for triggering the reorg, but you could add a brief comment like `// Make base chain longer (7 blocks) than side chain (2 blocks) so it becomes canonical` to make the reorg trigger explicit for readers.

ElFantasma (approved):
One optional improvement: consider adding a brief comment on the deploy initcode (line 427) explaining the runtime bytecode (SSTORE(CALLDATALOAD(0), CALLDATALOAD(32))) — the existing test_storage_slots_reorg has the same gap, but since this test is specifically about trie structure, grounding the contract behavior helps future readers.

tooling/reorgs/src/main.rs
Himess: Yes, intentional — each send_call is followed by build_payload + notify + update_forkchoice, so the nonce increments correctly before the next call.

tooling/reorgs/src/main.rs
Himess: Done, added the comment as suggested.

Himess:
Thanks for the review! Added the runtime bytecode explanation to both test functions and the block count detail to the reorg comment.

ElFantasma:
> Thanks for the review! Added the runtime bytecode explanation to both test functions and the block count detail to the reorg comment.

Thanks for your contribution. It needs two more approvals, and we have the pending issue with our CI to merge this, hopefully we will solve it in the near future.

PR #6109, opened 2026-02-03
feat(l1): store validated ENR from handshake in peer table

### LINKED ISSUE #6056: Store validated ENR from handshake in peer table
**Motivation**

When receiving a handshake from an unknown peer, we validate their ENR but don't store it in the peer table. Per the [discv5 spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md), after validating an ENR signature, we should store the record.

**Description**

Currently in `handle_handshake()`:
1. Unknown peer sends us a message
2. We send WhoAreYou
3. They respond with Handshake containing their ENR
4. We validate ENR signature ✅
5. We verify node_id matches ✅
6. We store the session ✅
7. We do NOT add them to peer table ❌

This means peers who contact us first work (session exists) but aren't discoverable for future lookups or included in NODES responses.

**Requirements**
- After validating ENR in `handle_handshake()`, add the peer to the peer table via `new_contact_records()`
- Consider also updating existing contacts if the received ENR has a higher seq (related TODO at `peer_table.rs:867`)

**Files**
- `crates/networking/p2p/discv5/server.rs` - `handle_handshake()`
- `crates/networking/p2p/discv5/peer_table.rs` - `new_contact_records()`

mikhailofff:
**Motivation**

When a handshake is received from an unknown peer, we verify their ENR but do not add it to the peer table. According to the discv5 specification, once an ENR signature has been validated, the record should be stored.

**Description**

1. Added the peer to the peer table using ```new_contact_records()```.
2. Updated existing contacts if the received ENR has a higher seq.

**Checklist**

Closes #6056.

### DIFF
--- a/.github/workflows/pr-main_mdbook.yml
+++ b/.github/workflows/pr-main_mdbook.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Check links
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --no-progress --exclude 'localhost' docs/
+          args: --no-progress --exclude 'localhost' --exclude 'medium.com' docs/
           fail: true
 
   deploy:
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -709,7 +709,7 @@ pub async fn import_blocks(
             } else {
                 // We need to have the state of the latest 128 blocks
                 blockchain
-                .add_block_pipeline(block)
+                .add_block_pipeline(block, None)
                 .inspect_err(|err| match err {
                     // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
                     ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
@@ -817,7 +817,7 @@ pub async fn import_blocks_bench(
                 .map_err(InvalidBlockError::InvalidBody)?;
 
             blockchain
-                .add_block_pipeline(block)
+                .add_block_pipeline(block, None)
                 .inspect_err(|err| match err {
                     // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
                     ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -620,7 +620,7 @@ pub async fn regenerate_head_state(
             .await?
             .ok_or_else(|| eyre::eyre!("Block {i} not found"))?;
 
-        blockchain.add_block_pipeline(block)?;
+        blockchain.add_block_pipeline(block, None)?;
     }
 
     info!("Finished regenerating state");
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -477,7 +477,7 @@ impl Command {
                         }
 
                         // Execute block
-                        blockchain.add_block_pipeline(block.clone())?;
+                        blockchain.add_block_pipeline(block.clone(), None)?;
 
                         // Add fee config to rollup store
                         rollup_store
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -278,6 +278,15 @@ struct PreMerkelizedAccountState {
     nodes: Vec<TrieNode>,
 }
 
+/// Work item for BAL state trie shard workers.
+struct BalStateWorkItem {
+    hashed_address: H256,
+    info: Option<AccountInfo>,
+    removed: bool,
+    /// Pre-computed storage root from Stage B, or None to keep existing.
+    storage_root: Option<H256>,
+}
+
 impl Blockchain {
     pub fn new(store: Store, blockchain_opts: BlockchainOptions) -> Self {
         Self {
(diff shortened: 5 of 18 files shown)


crates/networking/p2p/discv5/peer_table.rs
```diff
@@ -864,7 +864,18 @@ impl PeerTableServer {
                     vacant_entry.insert(contact);
                     METRICS.record_new_discovery().await;
                 }
-                // TODO Handle the case the contact is already present
+                // The contact is already present
+                else if let Some(existing_contact) = self.contacts.get_mut(&node_id) {
+                    if let Some(existing_record) = &existing_contact.record {
+                        if node_record.seq > existing_record.seq {
+                            existing_contact.record = Some(node_record);
```
ElFantasma: nit: `record_new_discovery` is semantically a "new peer found" metric. Firing it on ENR updates for existing contacts inflates the count. Consider either a separate `record_enr_update` metric or not recording anything here.

ElFantasma:
Thanks for your contribution, but a previous [PR](https://github.com/lambdaclass/ethrex/pull/6172) fixing this exists.

ElFantasma:
Sorry, there is no previous PR for this issue, I got mixed up. I'm reopening this

crates/networking/p2p/discv5/server.rs
```diff
@@ -303,6 +303,11 @@ impl DiscoveryServer {
                 .secp256k1
                 .and_then(|pk| PublicKey::from_slice(pk.as_bytes()).ok());
 
+            // Add the peer to the peer table
+            self.peer_table
+                .new_contact_records(vec![record.clone()], self.local_node.node_id())
+                .await?;
```
ElFantasma: **Peer added before full handshake authentication.** This call runs before the `derived_node_id != src_id` check (line 316) *and* before the id-signature verification (line 337). A malicious node claiming `src_id=X` but sending a valid ENR for node Y would get Y inserted into the peer table — the ENR signature is valid (it's Y's real ENR), so `new_contact_records` accepts it. The function returns early at line 317, but the peer table entry persists.

Move this call after the id-signature verification at line 346 — there's no reason to store anything until the handshake is fully authenticated.

crates/networking/p2p/discv5/peer_table.rs
ElFantasma: **`contact.node` not updated — stale IP/port.** Only `record` is updated, but `contact.node` keeps its old IP and port. If the new ENR has a different address, `send_ordinary` (which uses `contact.node.udp_addr()`) still sends to the old address. The `Node` and `NodeRecord` become inconsistent.

Need to also do `existing_contact.node = node;` (the `node` parsed from `Node::from_enr` at the top of the loop).

crates/networking/p2p/discv5/peer_table.rs
```diff
@@ -864,7 +864,18 @@ impl PeerTableServer {
                     vacant_entry.insert(contact);
                     METRICS.record_new_discovery().await;
                 }
-                // TODO Handle the case the contact is already present
+                // The contact is already present
+                else if let Some(existing_contact) = self.contacts.get_mut(&node_id) {
```
ElFantasma: **No validation reset on IP/port change.** When the new ENR changes the node's IP or port, `validation_timestamp` should be cleared and `ping_req_id` reset. Without this, `validate_contact` still considers the node validated at the *old* address, which undermines anti-amplification checks (see #6199 / PR #6200). The fix should look like:
```rust
if existing_contact.node.ip != node.ip || existing_contact.node.udp_port != node.udp_port {
    existing_contact.validation_timestamp = None;
    existing_contact.ping_req_id = None;
}
```

crates/networking/p2p/discv5/peer_table.rs
```diff
@@ -864,7 +864,18 @@ impl PeerTableServer {
...
+                    if let Some(existing_record) = &existing_contact.record {
+                        if node_record.seq > existing_record.seq {
+                            existing_contact.record = Some(node_record);
+                            METRICS.record_new_discovery().await;
```
ElFantasma: **Metric inflation.** `record_new_discovery()` fires on ENR updates to existing contacts, not just genuinely new discoveries. This inflates the discovery count. Should only emit the metric in the vacant-entry branch (which already does).

crates/networking/p2p/discv5/peer_table.rs
```diff
@@ -864,7 +864,18 @@ impl PeerTableServer {
                     vacant_entry.insert(contact);
                     METRICS.record_new_discovery().await;
                 }
-                // TODO Handle the case the contact is already present
+                // The contact is already present
```
ElFantasma: This was already implemented more thoroughly on a previous [PR](https://github.com/lambdaclass/ethrex/pull/6172)
This PR will conflict once that merges. Consider rebasing on top of it and focusing only on the `handle_handshake` insertion.

ElFantasma (dismissed):

crates/networking/p2p/discv5/server.rs
```diff
@@ -339,6 +339,11 @@ impl DiscoveryServer {
             return Ok(());
         }
 
+        // Add the peer to the peer table
+        self.peer_table
+            .new_contact_records(vec![record.clone()], self.local_node.node_id())
```
ElFantasma: Bug: `record` is out of scope here. It was bound inside the `else if let Some(record) = &authdata.record` arm (line 295), which only lives within that branch of the `let src_pubkey = ...` expression. By line 344, that binding is gone.

The first commit had this call *inside* the `else if let` arm (where `record` was in scope). The second commit moved it here, outside the expression. CI likely passed due to stale build cache.

Fix: wrap this in `if let Some(record) = &authdata.record { ... }` so the call only runs when an ENR was actually provided in the handshake. This also makes the logic correct — there's no record to store when the peer was already known (first branch).

PR #6115, opened 2026-02-03
feat(l1): gnosis network support

Arkenan:
Gnosis is fully compatible with the engine api, so the only modification we need is support for gnosis as a network such as mainnet, hoodi or sepolia.


iovoid:
On startup the node will print 'Chain ID: 100 (unknown)' instead of the network name.

iovoid:
Notes from trying to get ethrex to sync starting from this PR: [GNOSIS_DEBUGGING.md](https://github.com/user-attachments/files/25191027/GNOSIS_DEBUGGING.md)

PR #6120, opened 2026-02-04
feat(l1): add mainnet fork-id test cases (EIP-2124)

### LINKED ISSUE #5231: Test fork-id computation for mainnet chain configuration
We currently have unit tests for fork-id computation on all testnets (Holesky, Sepolia, Hoodi), but we lack unit tests for mainnet. We want to add something similar to this:

https://github.com/lambdaclass/ethrex/blob/353d5894c9e0e151a866d2c39dbdf693b2c1e46e/crates/common/types/fork_id.rs#L251

but for mainnet. The values need to be double-checked with other clients.

dik654:
**Motivation**

Add mainnet fork-id test cases to verify EIP-2124 implementation against go-ethereum, as requested in https://github.com/lambdaclass/ethrex/issues/5231.

**Description**

- Add comprehensive mainnet fork-id test cases covering all 16 forks from Genesis through Prague
- Test values verified against [go-ethereum's forkid_test.go](https://github.com/ethereum/go-ethereum/blob/master/core/forkid/forkid_test.go)
- Covers both PoW era (block-based forks) and PoS era (timestamp-based forks)
image

**Checklist**

Closes #5231


ElFantasma (approved):
Two minor observations:

1. **`assert_test_cases` only checks `is_valid()`, not equality** — since every test case has `is_valid: true`, the test verifies the computed fork_id passes validation against the expected one, but doesn't strictly assert `ForkId::new() == test_case.fork_id`. Adding `assert_eq!(fork_id, test_case.fork_id)` would make all fork-id tests (not just this one) strictly verify computed values. Pre-existing across all tests, but worth a follow-up.

2. **Minor divergence from geth reference values** — The PoS-era test cases use `head: 20_000_000` / `25_000_000` where geth uses `head: 30_000_000` / `50_000_000`. Functionally equivalent since only timestamps matter in PoS, but matching geth's exact values would make cross-referencing easier.

PR #6124, opened 2026-02-04
fix(p2p): store validated ENR from handshake in peer table

### LINKED ISSUE #6056: Store validated ENR from handshake in peer table

Himess:
Closes #6056

### DIFF
--- a/crates/networking/p2p/discv5/server.rs
+++ b/crates/networking/p2p/discv5/server.rs
@@ -354,6 +354,13 @@ impl DiscoveryServer {
             .set_session_info(src_id, session.clone())
             .await?;
 
+        // Store validated ENR in peer table
+        if let Some(record) = authdata.record.clone() {
+            self.peer_table
+                .new_contact_records(vec![record], self.local_node.node_id())
+                .await?;
+        }
+
         // Decrypt and handle the contained message
         let mut encrypted = packet.encrypted_message.clone();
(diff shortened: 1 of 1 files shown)


ElFantasma:
Thanks for your contribution, but a previous [PR](https://github.com/lambdaclass/ethrex/pull/6109) fixing this exists.

PR #6107, opened 2026-02-03
docs(l1): add UX/DevEx roadmap

iovoid:
**Motivation**

We want to improve the experience of both users and developers, and to do so we need to know what is in need of improvement and which areas are of greater priority.

**Description**

Adds a roadmap for UX/DevEx improvement.

PR #6126, opened 2026-02-04
refactor(l1): improve error types with actionable context

pablodeymo:
## Motivation

This PR implements Section 1.4 ("Improve Error Type Definitions") from the UX/DevEx improvement roadmap (PR #6107). Error enums currently have variants with no context, making debugging difficult for operators. When a decode error occurs, the message `InvalidLength` tells you nothing about what failed to decode or why.

**Current pain point example:**
```
ERROR ethrex::p2p: Failed to decode message: InvalidLength
```

Operator asks: "InvalidLength of what? The block? A transaction? The header?"

**After this PR:**
```
ERROR ethrex::p2p: Failed to decode message: Invalid RLP length decoding Transaction
```

Operator knows: "The Transaction decoder failed. Let me check if it's a known bad transaction format."

---

### Approach: Context at API Boundaries (Option B)

After analyzing the codebase, we chose an approach that balances debuggability with code simplicity:

1. **Keep internal errors simple** - Generic decoders in `decode.rs` don't know what type they're decoding, so they use constructors like `invalid_length()` that create errors without context.

2. **Add context at type boundaries** - Type-specific decoders (Transaction, Account, etc.) wrap errors with context using `.with_context("TypeName")`.

3. **Zero memory overhead** - Verified by measurement that adding `Option<&'static str>` fields doesn't increase enum size because existing `String` variants already dominate.

### Why Not Other Approaches?

| Approach | Rejected Because |
|----------|------------------|
| **Full context everywhere** | 70+ sites in `decode.rs` alone; generic decoders don't know the type |
| **`#[track_caller]`** | `thiserror` doesn't integrate well; location info helps devs, not operators |
| **Error wrapper type** | Changes return types at API boundaries; more invasive |

---

### 1. RLPDecodeError (`crates/common/rlp/error.rs`)

**Before:**
```rust
#[error("InvalidLength")]
InvalidLength,
#[error("MalformedBoolean")]
MalformedBoolean,
```

**After:**
```rust
#[error("Invalid RLP length{}", fmt_ctx(.0))]
InvalidLength(Option<&'static str>),
#[error("Malformed boolean: expected 0x80 or 0x01, got 0x{0:02x}")]
MalformedBoolean(u8),

// Helper constructors for internal use
impl RLPDecodeError {
    pub fn invalid_length() -> Self { Self::InvalidLength(None) }
    pub fn malformed_boolean(got: u8) -> Self { Self::MalformedBoolean(got) }
    
    pub fn with_context(self, ctx: &'static str) -> Self {
        match self {
            Self::InvalidLength(_) => Self::InvalidLength(Some(ctx)),
            // ...
        }
    }
}
```

### 2. PeerConnectionError (`crates/networking/p2p/rlpx/error.rs`)

**Before:**
```rust
#[error("No matching capabilities")]
NoMatchingCapabilities,
#[error("Invalid peer id")]
InvalidPeerId,
#[error("Invalid message length")]
InvalidMessageLength,
```

**After:**
```rust
#[error("No matching capabilities: {0}")]
NoMatchingCapabilities(String),
#[error("Invalid peer ID: {0}")]
InvalidPeerId(&'static str),
#[error("Invalid message length: {0}")]
InvalidMessageLength(&'static str),
```

### 3. Context at Type Boundaries

```rust
impl RLPDecode for Transaction {
        decode_transaction(rlp).map_err(|e| e.with_context("Transaction"))
    }
}
```

### 4. Removed Dead Code

- `StoreError::DecodeError` - defined but never used anywhere in the codebase
- TODO comments from error files (the TODOs are now addressed)

---

## Error Message Comparison

| Scenario | Before | After |
|----------|--------|-------|
| Transaction decode | `InvalidLength` | `Invalid RLP length decoding Transaction` |
| Boolean decode | `MalformedBoolean` | `Malformed boolean: expected 0x80 or 0x01, got 0x42` |
| Capability mismatch | `No matching capabilities` | `No matching capabilities: no common eth version` |
| Peer ID compression | `Invalid peer id` | `Invalid peer ID: failed to compress public key` |
| Handshake parse | `Invalid message length` | `Invalid message length: handshake message too short` |
| Frame size | `Invalid message length` | `Invalid message length: frame exceeds max size` |

---

## Memory Analysis

Verified that adding context fields has **zero memory overhead**:

```
Component sizes:
  String:               24 bytes
  Option<&'static str>: 16 bytes

Enum sizes:
  CurrentError:  40 bytes
  ProposedError: 40 bytes  ← No change!

Result sizes:
  Result<u8, Error>:     40 bytes  ← No change!
  Result<u64, Error>:    40 bytes  ← No change!
```

The enum size is determined by its largest variant (`String` at 24 bytes). Adding 16-byte `Option<&'static str>` fields to smaller variants doesn't increase the overall size.

---

## Files Changed

| Category | Files | Purpose |
|----------|-------|---------|
| Error definitions | `rlp/error.rs`, `rlpx/error.rs`, `storage/error.rs` | New error variants with context |
| Generic decoders | `rlp/decode.rs`, `rlp/structs.rs` | Use new constructors |
| Type decoders | `types/transaction.rs`, `types/account.rs`, `types/receipt.rs`, `trie/node_hash.rs` | Add context at boundaries |
| P2P code | `handshake.rs`, `server.rs`, `codec.rs`, `metrics.rs` | Update to new error variants |
| Network messages | `discv4/messages.rs`, `discv5/messages.rs`, `rlpx/message.rs`, etc. | Use new constructors |

---

## Breaking Changes

This is a **breaking change** for anyone matching on these error variants:

```rust
// Before
match err {
    RLPDecodeError::InvalidLength => { ... }
}

// After
match err {
    RLPDecodeError::InvalidLength(ctx) => {
        if let Some(c) = ctx {
            log::error!("Invalid length decoding {}", c);
        }
    }
}
```

All breakage is **compile-time only** - the compiler tells you exactly what to fix.

---

## Testing

- [x] `cargo check` passes
- [x] `cargo fmt` passes
- [x] `cargo clippy` passes
- [x] `cargo test -p ethrex-rlp` - 2 tests pass
- [x] `cargo test -p ethrex-common` - 62 tests pass
- [x] `cargo test -p ethrex-p2p` - 50 tests pass

---

## Related

- Parent roadmap: PR #6107 (UX/DevEx Improvement Plan)
- Addresses: Section 1.4 "Improve Error Type Definitions"
- Related sections: 1.3 (context-discarding `map_err`) can be tackled separately

### DIFF
--- a/benches/benches/build_block_benchmark.rs
+++ b/benches/benches/build_block_benchmark.rs
@@ -146,6 +146,7 @@ fn create_payload_block(genesis_block: &Block, store: &Store) -> (Block, u64) {
         random: genesis_block.header.prev_randao,
         withdrawals: None,
         beacon_root: genesis_block.header.parent_beacon_block_root,
+        slot_number: None,
         version: 3,
         elasticity_multiplier: 1,
         gas_ceil: DEFAULT_BUILDER_GAS_CEIL,
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -84,6 +84,7 @@ pub struct BuildPayloadArgs {
     pub random: H256,
     pub withdrawals: Option<Vec<Withdrawal>>,
     pub beacon_root: Option<H256>,
+    pub slot_number: Option<u64>,
     pub version: u8,
     pub elasticity_multiplier: u64,
     pub gas_ceil: u64,
@@ -170,6 +171,7 @@ pub fn create_payload(
         requests_hash: chain_config
             .is_prague_activated(args.timestamp)
             .then_some(*DEFAULT_REQUESTS_HASH),
+        slot_number: args.slot_number,
         ..Default::default()
     };
 
--- a/crates/l2/sequencer/block_producer.rs
+++ b/crates/l2/sequencer/block_producer.rs
@@ -183,6 +183,7 @@ impl BlockProducer {
             random: H256::zero(),
             withdrawals: Default::default(),
             beacon_root: Some(head_beacon_block_root),
+            slot_number: None,
             version,
             elasticity_multiplier: self.elasticity_multiplier,
             gas_ceil: self.block_gas_limit,
--- a/crates/networking/rpc/engine/fork_choice.rs
+++ b/crates/networking/rpc/engine/fork_choice.rs
@@ -11,7 +11,9 @@ use tracing::{debug, info, warn};
 use crate::{
     rpc::{RpcApiContext, RpcHandler},
     types::{
-        fork_choice::{ForkChoiceResponse, ForkChoiceState, PayloadAttributesV3},
+        fork_choice::{
+            ForkChoiceResponse, ForkChoiceState, PayloadAttributesV3, PayloadAttributesV4,
+        },
         payload::PayloadStatus,
     },
     utils::RpcErr,
@@ -122,6 +124,46 @@ impl RpcHandler for ForkChoiceUpdatedV3 {
         if let (Some(head_block), Some(attributes)) = (head_block_opt, &self.payload_attributes) {
             validate_attributes_v3(attributes, &head_block, &context)?;
             let payload_id = build_payload(attributes, context, &self.fork_choice_state, 3).await?;
+            response.set_id(payload_id);
+        }
+        serde_json::to_value(response).map_err(|error| RpcErr::Internal(error.to_string()))
+    }
+}
+
+#[derive(Debug)]
+pub struct ForkChoiceUpdatedV4 {
+    pub fork_choice_state: ForkChoiceState,
+    pub payload_attributes: Option<PayloadAttributesV4>,
+}
+
+impl From<ForkChoiceUpdatedV4> for RpcRequest {
+    fn from(val: ForkChoiceUpdatedV4) -> Self {
+        RpcRequest {
+            method: "engine_forkchoiceUpdatedV4".to_string(),
+            params: Some(vec![
+                serde_json::json!(val.fork_choice_state),
+                serde_json::json!(val.payload_attributes),
+            ]),
+            ..Default::default()
+        }
+    }
+}
+
+impl RpcHandler for ForkChoiceUpdatedV4 {
+    fn parse(params: &Option<Vec<Value>>) -> Result<Self, RpcErr> {
+        let (fork_choice_state, payload_attributes) = parse_v4(params)?;
+        Ok(ForkChoiceUpdatedV4 {
+            fork_choice_state,
+            payload_attributes,
+        })
+    }
+
+    async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        let (head_block_opt, mut response) =
+            handle_forkchoice(&self.fork_choice_state, context.clone(), 4).await?;
+        if let (Some(head_block), Some(attributes)) = (head_block_opt, &self.payload_attributes) {
+            validate_attributes_v4(attributes, &head_block, &context)?;
+            let payload_id = build_payload_v4(attributes, context, &self.fork_choice_state).await?;
             response.set_id(payload_id);
         }
         serde_json::to_value(response).map_err(|error| RpcErr::Internal(error.to_string()))
@@ -386,6 +428,7 @@ async fn build_payload(
         random: attributes.prev_randao,
         withdrawals: attributes.withdrawals.clone(),
         beacon_root: attributes.parent_beacon_block_root,
+        slot_number: None,
         version,
         elasticity_multiplier: ELASTICITY_MULTIPLIER,
         gas_ceil: context.gas_ceil,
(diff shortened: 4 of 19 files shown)


crates/networking/p2p/rlpx/connection/handshake.rs
```diff
@@ -255,7 +255,9 @@ async fn receive_handshake_msg<S: AsyncRead + std::marker::Unpin>(
     let ack_data = [buf[0], buf[1]];
     let msg_size = u16::from_be_bytes(ack_data) as usize;
     if msg_size > P2P_MAX_MESSAGE_SIZE {
-        return Err(PeerConnectionError::InvalidMessageLength);
+        return Err(PeerConnectionError::InvalidMessageLength(
+            "handshake message too short",
+        ));
```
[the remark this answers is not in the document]
pablodeymo: fixed!

crates/networking/p2p/rlpx/connection/handshake.rs
```diff
@@ -255,7 +255,9 @@ async fn receive_handshake_msg<S: AsyncRead + std::marker::Unpin>(
...
     }
```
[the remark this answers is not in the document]
pablodeymo: fixed!

pablodeymo:
@greptile-apps ?

crates/common/rlp/structs.rs
```diff
@@ -103,7 +103,7 @@ impl<'a> Decoder<'a> {
         if self.payload.is_empty() {
             Ok(self.remaining)
         } else {
-            Err(RLPDecodeError::MalformedData)
+            Err(RLPDecodeError::MalformedData(None))
```
iovoid: Why not use the helper?

pablodeymo:
Fixed! Made the helper constructors `const fn` so they can be used in `Decoder::finish()` which is also const. Thanks for catching that.

azteca1998 (approved):

crates/common/rlp/error.rs
```diff
@@ -1,30 +1,64 @@
...
-    #[error("IncompatibleProtocol: {0}")]
+    #[error("Incompatible protocol: {0}")]
     IncompatibleProtocol(String),
     #[error("{0}")]
     Custom(String),
 }
 
-// TODO: improve errors
+fn fmt_ctx(ctx: &Option<&'static str>) -> String {
+    ctx.map(|c| format!(" decoding {c}")).unwrap_or_default()
```
ElFantasma, line 24: nit: `fmt_ctx` allocates a `String` on every `Display` call via `format!()`. In the `None` case, `unwrap_or_default()` returns an empty `String` (heap-allocated). Since RLP decode errors can appear in P2P logging hot paths, you could avoid the allocation by inlining the formatting:

```rust
#[error("Invalid RLP length{}", .0.map(|c| format!(" decoding {c}")).unwrap_or_default())]
```

Or change the helper to return `&str`:
```rust
fn fmt_ctx(ctx: &Option<&'static str>) -> &'static str {
    // Unfortunately doesn't work with format — 
    // you'd need a different approach
}
```

Actually the simplest fix: implement `Display` manually for the variants that need context, writing directly to the formatter without intermediate `String`. But this is minor — the current approach works, it's just doing a small heap allocation per error display.

crates/common/rlp/error.rs
```diff
@@ -1,30 +1,64 @@
...
+        Self::UnexpectedList(None)
+    }
+
+    pub const fn unexpected_string() -> Self {
+        Self::UnexpectedString(None)
+    }
+
+    pub fn with_context(self, ctx: &'static str) -> Self {
+        match self {
+            Self::InvalidLength(_) => Self::InvalidLength(Some(ctx)),
```
ElFantasma, line 50: `with_context` unconditionally overwrites any existing context. Currently safe because nested type errors go through `field_decode_error` → `Custom(...)` which isn't touched here. But if someone later uses `with_context` on a type that's decoded inside another `with_context`-wrapped type *without* going through `Decoder::decode_field`, the inner context is silently lost.

Consider either:
1. A comment documenting "outermost caller wins" behavior
2. Or: `Self::InvalidLength(existing) => Self::InvalidLength(Some(existing.unwrap_or(ctx)))` to preserve inner context ("innermost wins")

crates/networking/p2p/rlpx/error.rs
```diff
@@ -18,15 +18,14 @@ pub enum CryptographyError {
...
-// TODO improve errors
 #[derive(Debug, Error)]
 pub enum PeerConnectionError {
     #[error("{0}")]
     HandshakeError(String),
     #[error("Invalid connection state: {0}")]
     StateError(String),
-    #[error("No matching capabilities")]
-    NoMatchingCapabilities,
+    #[error("No matching capabilities: {0}")]
```
ElFantasma, line 27: nit: Inconsistent context types — `NoMatchingCapabilities(String)` uses owned `String` while `InvalidPeerId(&'static str)` and `InvalidMessageLength(&'static str)` use borrowed `&'static str`. The `String` here is needed because the call site in `server.rs` builds the message dynamically (`"no common eth version".to_string()`), but that string is actually a literal — it could be `&'static str` too. Consider making this `&'static str` for consistency (unless you anticipate dynamic messages in the future).

PR #6143, opened 2026-02-05
feat(l1): add CPU profiling via pprof behind feature flag

ilitteri:
## Motivation

When investigating performance bottlenecks during sync or block execution, having a built-in CPU profiler that can be toggled with a feature flag is much more convenient than setting up external profiling tools. This follows the same pattern as the existing `jemalloc_profiling` feature.

## Description

Adds opt-in CPU profiling to the ethrex node using [pprof-rs](https://github.com/tikv/pprof-rs):

- **`cpu_profiling` feature flag** in `cmd/ethrex/Cargo.toml` — enables the `pprof` dependency with `frame-pointer` unwinding (not in `default`, zero impact on normal builds)
- **Profiler startup** — a `ProfilerGuard` starts at boot with 1000 Hz sampling, right after tracing is initialized
- **Profile output** — on shutdown (Ctrl+C / SIGTERM), a `write_cpu_profile` helper writes `profile.pb` to the current working directory in pprof protobuf format, with proper error handling via `?` (no unwrap panics during shutdown)
- **Frame-pointer unwinding** — required to avoid SIGTRAP crashes on macOS aarch64 (Apple Silicon) caused by DWARF-based stack unwinding in pprof-rs ([tikv/pprof-rs#187](https://github.com/tikv/pprof-rs/issues/187))
- **Documentation** — comprehensive profiling guide at `docs/developers/l1/profiling.md` covering CPU profiling with pprof, memory profiling with jemalloc, and samply usage

All profiling code is gated behind `#[cfg(feature = "cpu_profiling")]` — no new dependencies or code paths are compiled without the feature.

### Usage

```bash
cargo build -p ethrex --release --features cpu_profiling
./target/release/ethrex --authrpc.jwtsecret ./secrets/jwt.hex --network holesky
# Ctrl+C → profile.pb appears in CWD
go tool pprof -http=:8080 profile.pb
```

## How to Test

1. `cargo check -p ethrex` — confirms no regressions without the feature
2. `cargo build -p ethrex --features cpu_profiling` — compiles cleanly with the feature
3. Run the node for a few minutes, Ctrl+C, confirm `profile.pb` exists and the log line `CPU profile written to profile.pb` appears

Tested on macOS aarch64 (Apple Silicon) with Hoodi for 5 minutes — node runs stably, shuts down gracefully, produces a 730K profile.

Issue #5972, opened 2026-01-21
Detect and address performance degradation in discovery operations with large contact tables


ElFantasma:
## Description

When the number of contacts in the peer table grows large, we may experience performance degradation in key discovery operations. We need to detect and address any bottlenecks.

## Affected operations

- **Lookup** - Finding nodes closest to a target
- **Revalidate** - Checking liveness of existing contacts
- **Prune** - Removing stale/unresponsive contacts
- **Filtering contacts** - Selecting contacts by protocol (discv4/discv5)

## Tasks

- [ ] Add metrics to measure time elapsed on the affected operations
- [ ] Investigate and fix any performance degradation detected

## Related issues

- #4245 - Reintroduce a proper implementation for the Kademlia table (may help resolve this)
- #3716 - Cap number of contacts in Kademlia table

ElFantasma:
We may have similar issues dealing with internal server caches, we have to include them in the analysis:
- pending_by_nonce
- pending_challenges
- whoareyou_rate_limit

PR #6144, opened 2026-02-06
perf(l1): throttle discovery, deduplicate KZG validation, and add RocksDB block cache

ilitteri:
## Motivation

CPU profiling on Hoodi steady-state revealed three major bottlenecks consuming the majority of CPU time. See the profiling methodology in PR #6143.

### Optimization 1: Discovery server throttling

The discovery server was the **#1 CPU consumer** at 24% cumulative, driven by:
- Two independent lookup timers (`Lookup` + `EnrLookup`) each firing at 100ms intervals
- No backoff when peer target is already met
- O(n) Vec allocation on every contact selection

Changes:
- **Merge timers**: Single `Lookup` timer alternates between regular lookup and ENR lookup
- **Increase intervals**: `INITIAL_LOOKUP_INTERVAL_MS` 100→500, `LOOKUP_INTERVAL_MS` 600→5000
- **Add backoff**: 30s maintenance interval when `target_peers_completion() >= 1.0`
- **Reservoir sampling**: Replace `collect::<Vec<_>>().choose()` with O(1)-memory reservoir sampling in `get_contact_for_lookup` and `get_contact_for_enr_lookup`

### Optimization 2: KZG blob validation dedup

KZG blob validation was the **#2 CPU consumer** at 15% cumulative, with the same blobs validated twice:
1. In `PooledTransactions::validate_requested()` (P2P layer)
2. Again in `Blockchain::add_blob_transaction_to_pool()` (blockchain layer)

Changes:
- Add `blobs_already_validated: bool` parameter to `add_blob_transaction_to_pool`
- P2P caller passes `true` (already validated in `validate_requested`)
- RPC caller passes `false` (no prior validation)

### Optimization 3: RocksDB block cache + commit buffer pre-alloc

RocksDB was using the tiny 8MB default block cache per column family, causing `RandomAccessFileReader::Read` to consume 8.89% flat CPU.

Changes:
- **4GB shared LRU block cache** across all column families
- **`cache_index_and_filter_blocks(true)`** + **`pin_l0_filter_and_index_blocks_in_cache(true)`** to keep bloom filters and indexes in memory
- **`Vec::with_capacity(512)`** in `NodeRef::commit()` to avoid repeated small allocations during trie merkleization (matching the existing pattern in `hash_no_commit()`)

## Results

Measured with `cpu_profiling` feature (pprof-rs, 1000 Hz) on Hoodi steady-state:

### After Optimizations 1+2 (vs baseline)

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| **Total CPU samples** | 56.82s/600s (9.47%) | 30.42s/600s (5.07%) | **-46.5%** |
| Discovery (`handle_cast` cum) | 11.02s (19.39%) | 0.68s (2.24%) | **-94%** |
| `UdpSocket::send_to` flat | 9.09s (16.0%) | 0.51s (1.66%) | **-94%** |
| KZG `BlobsBundle::validate` abs | 8.52s | 5.74s | **-33%** |

### After Optimization 3 (vs post-1+2)

| Metric | After Opt 1+2 | After Opt 3 | Change |
|--------|---------------|-------------|--------|
| **Total CPU samples** | 30.42s/600s (5.07%) | 16.98s/720s (2.36%) | **-53%** |
| RocksDB `RandomAccessFileReader` flat | 2.71s (8.89%) | 0.68s (4.01%) | **-75%** |

### Cumulative (all 3 optimizations vs original baseline)

| Metric | Baseline | Final | Total Change |
|--------|----------|-------|--------------|
| **Steady-state CPU** | 9.47% | 2.36% | **-75%** |

## Test plan

- [x] `cargo check` passes for all affected crates
- [x] Release build succeeds
- [x] Profiling confirms CPU reduction at each stage
- [ ] Verify peer discovery still works correctly (connects to 4+ peers)
- [ ] Verify blob transactions are still validated correctly via RPC
- [ ] Run on Hoodi for extended period to confirm stability with 4GB cache

### DIFF
--- a/crates/networking/p2p/discv4/peer_table.rs
+++ b/crates/networking/p2p/discv4/peer_table.rs
@@ -702,32 +702,26 @@ impl PeerTableServer {
     }
 
     fn get_contact_for_lookup(&self) -> Option<Contact> {
-        // Reservoir sampling: O(n) scan with O(1) memory instead of collecting into a Vec.
-        let mut rng = rand::rngs::OsRng;
-        let mut result: Option<&Contact> = None;
-        let mut count = 0u64;
-        for c in self.contacts.values() {
-            if c.n_find_node_sent < MAX_FIND_NODE_PER_PEER && !c.disposable {
-                count += 1;
-                if rng.gen_range(0..count) == 0 {
-                    result = Some(c);
-                }
-            }
-        }
-        result.cloned()
+        self.random_contact(|c| c.n_find_node_sent < MAX_FIND_NODE_PER_PEER && !c.disposable)
     }
 
     fn get_contact_for_enr_lookup(&mut self) -> Option<Contact> {
-        // Reservoir sampling: O(n) scan with O(1) memory instead of collecting into a Vec.
+        self.random_contact(|c| {
+            c.was_validated()
+                && !c.has_pending_enr_request()
+                && c.record.is_none()
+                && !c.disposable
+        })
+    }
+
+    /// Pick a uniformly random contact matching `predicate` using reservoir sampling.
+    /// O(n) scan with O(1) extra memory (no intermediate Vec allocation).
+    fn random_contact(&self, predicate: impl Fn(&Contact) -> bool) -> Option<Contact> {
         let mut rng = rand::rngs::OsRng;
         let mut result: Option<&Contact> = None;
         let mut count = 0u64;
         for c in self.contacts.values() {
-            if c.was_validated()
-                && !c.has_pending_enr_request()
-                && c.record.is_none()
-                && !c.disposable
-            {
+            if predicate(c) {
                 count += 1;
                 if rng.gen_range(0..count) == 0 {
                     result = Some(c);
--- a/crates/networking/p2p/discv4/server.rs
+++ b/crates/networking/p2p/discv4/server.rs
@@ -41,8 +41,8 @@ const REVALIDATION_INTERVAL: Duration = Duration::from_secs(12 * 60 * 60); // 12
 /// The initial interval between peer lookups, until the number of peers reaches
 /// [target_peers](DiscoverySideCarState::target_peers), or the number of
 /// contacts reaches [target_contacts](DiscoverySideCarState::target_contacts).
-pub const INITIAL_LOOKUP_INTERVAL_MS: f64 = 500.0;
-pub const LOOKUP_INTERVAL_MS: f64 = 5000.0;
+pub const INITIAL_LOOKUP_INTERVAL_MS: f64 = 500.0; // 2 per second
+pub const LOOKUP_INTERVAL_MS: f64 = 5000.0; // 12 per minute
 /// Slow maintenance interval when target peers are met (30 seconds).
 const MAINTENANCE_LOOKUP_INTERVAL: Duration = Duration::from_secs(30);
 const CHANGE_FIND_NODE_MESSAGE_INTERVAL: Duration = Duration::from_secs(5);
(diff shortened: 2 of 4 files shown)


crates/networking/p2p/discv4/server.rs
```diff
@@ -92,6 +93,8 @@ pub struct DiscoveryServer {
     /// signatures being expensive.
     find_node_message: BytesMut,
     initial_lookup_interval: f64,
+    /// Alternates between regular lookup and ENR lookup on each tick.
+    next_lookup_is_enr: bool,
```
ElFantasma, line 97: Merging Lookup and EnrLookup into a single alternating timer changes the effective lookup rate: each type now fires at half the CLI-configured interval. If a user sets 600ms, they previously got a Lookup every 600ms and an EnrLookup every 600ms — now each fires every 1200ms. This is a silent behavioral change, and the `next_lookup_is_enr` alternation adds state tracking complexity that wasn't needed with two independent timers.

crates/networking/p2p/discv4/server.rs
```diff
@@ -734,25 +732,36 @@ impl GenServer for DiscoveryServer {
...
-                let _ = self
-                    .lookup()
+                let peer_completion = self
+                    .peer_table
+                    .target_peers_completion()
                     .await
-                    .inspect_err(|e| error!(err=?e, "Error performing Discovery lookup"));
+                    .unwrap_or_default();
+
+                let interval = if peer_completion >= 1.0 {
```
ElFantasma, line 741: This existing `lookup_interval_function` already handles the saturated case — when `peer_completion >= 1.0`, the ease-in-out cubic curve returns `LOOKUP_INTERVAL_MS` as the upper bound. This 30s hard override bypasses the designed progressive interval entirely.

crates/networking/p2p/discv4/server.rs
```diff
@@ -41,8 +41,10 @@ const REVALIDATION_INTERVAL: Duration = Duration::from_secs(12 * 60 * 60); // 12
 /// The initial interval between peer lookups, until the number of peers reaches
 /// [target_peers](DiscoverySideCarState::target_peers), or the number of
 /// contacts reaches [target_contacts](DiscoverySideCarState::target_contacts).
-pub const INITIAL_LOOKUP_INTERVAL_MS: f64 = 100.0; // 10 per second
-pub const LOOKUP_INTERVAL_MS: f64 = 600.0; // 100 per minute
+pub const INITIAL_LOOKUP_INTERVAL_MS: f64 = 500.0; // 2 per second
```
ElFantasma, line 44: These constants are already CLI-configurable via `initial_lookup_interval`. Changing the defaults in code is unnecessary — operators who need slower lookups can just set the flag.

crates/networking/p2p/discv4/peer_table.rs
```diff
@@ -702,28 +702,33 @@ impl PeerTableServer {
...
+            c.was_validated()
+                && !c.has_pending_enr_request()
+                && c.record.is_none()
+                && !c.disposable
+        })
+    }
+
+    /// Pick a uniformly random contact matching `predicate` using reservoir sampling.
+    /// O(n) scan with O(1) extra memory (no intermediate Vec allocation).
+    fn random_contact(&self, predicate: impl Fn(&Contact) -> bool) -> Option<Contact> {
```
ElFantasma, line 719: The reservoir sampling removes the Vec allocation but is still O(n) in predicate evaluation on every tick. Consider maintaining pre-filtered lists (e.g. contacts eligible for lookup, contacts eligible for ENR lookup) updated on mutation events (send FindNode, receive ENR, validate peer, etc.). That way random selection becomes O(1) via random index access. The trade-off is keeping filtered sets in sync with mutations, but those are discrete events at clear call sites and less frequent than lookup ticks.

crates/networking/p2p/discv4/server.rs
ElFantasma: And the same behavior can be obtained by using half speed lookup interval via command line argument

ElFantasma:
For the discovery Issues, there is a ticket, but it is planned to be tacked after we merge Discv4 and Discv5 in a single GenServer:
https://github.com/lambdaclass/ethrex/issues/5972

azteca1998 (approved):
Please check the current review comments before merging.

PR #6146, opened 2026-02-06
feat(l1): add on-demand CPU profiling HTTP endpoint

ilitteri:
## Motivation

Replace the boot-time profiler (start at launch, write profile.pb at shutdown) with an on-demand HTTP endpoint. This enables live profiling without restarting the node:

```bash
curl -o profile.pb "http://localhost:8545/debug/pprof/profile?seconds=30"
go tool pprof -top profile.pb
```

## Description

- **`GET /debug/pprof/profile`** — captures a CPU profile for a configurable duration and returns protobuf data
- **Query params**: `seconds` (default 30, max 300), `frequency` (default 1000 Hz)
- **Mutual exclusion** — `AtomicBool` with a `ProfilingGuard` drop guard ensures only one capture runs at a time (pprof-rs limitation). Concurrent requests get `409 Conflict`
- **Feature-gated** — `cpu_profiling` feature in `ethrex-rpc` (with `dep:pprof`), forwarded from `cmd/ethrex`. Without the feature, the endpoint returns `501 Not Implemented`
- **Follows jemalloc_profiling pattern** — pprof dependency lives in ethrex-rpc, not the binary crate

### Changes

1. **`crates/networking/rpc/Cargo.toml`** — add `pprof` optional dep + `cpu_profiling` feature
2. **`crates/networking/rpc/rpc.rs`** — add `cpu_profile` module (enabled + stub) and register route
3. **`cmd/ethrex/Cargo.toml`** — replace old `cpu_profiling = ["dep:pprof"]` with `cpu_profiling = ["ethrex-rpc/cpu_profiling"]`
4. **`cmd/ethrex/ethrex.rs`** — remove boot-time `ProfilerGuard` and `write_cpu_profile`
5. **`docs/developers/l1/profiling.md`** — update to reflect HTTP endpoint workflow

## How to Test

1. `cargo check -p ethrex-rpc` — no new deps without the feature
2. `cargo check -p ethrex-rpc --features cpu_profiling` — compiles cleanly
3. `cargo test -p ethrex-rpc --features cpu_profiling -- cpu_profiling` — unit test passes (basic capture, concurrent rejection, flag reset)
4. Start node with feature, `curl http://localhost:8545/debug/pprof/profile?seconds=5` — returns valid protobuf
5. Two concurrent requests — second gets `409 Conflict`
6. Without feature: endpoint returns `501 Not Implemented`

### DIFF
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -1296,9 +1296,9 @@ mod tests {
     #[cfg(feature = "cpu_profiling")]
     #[tokio::test]
     async fn cpu_profiling_endpoint() {
+        use crate::rpc::cpu_profile::{ProfileParams, handle_cpu_profile};
         use axum::extract::Query;
         use axum::http::StatusCode;
-        use crate::rpc::cpu_profile::{ProfileParams, handle_cpu_profile};
 
         let query = |seconds, frequency| {
             Query(ProfileParams {
@@ -1312,9 +1312,7 @@ mod tests {
         assert_eq!(resp.status(), StatusCode::OK);
 
         // 2. Concurrent request gets 409.
-        let long_capture = tokio::spawn(async move {
-            handle_cpu_profile(query(3, 100)).await
-        });
+        let long_capture = tokio::spawn(async move { handle_cpu_profile(query(3, 100)).await });
         tokio::time::sleep(std::time::Duration::from_millis(500)).await;
 
         let err = handle_cpu_profile(query(1, 100)).await.unwrap_err();


iovoid (approved):

crates/networking/rpc/rpc.rs
```diff
@@ -158,6 +158,97 @@ pub async fn handle_get_heap_flamegraph() -> Result<(), (StatusCode, String)> {
...
+        pub seconds: Option<u64>,
+        /// Sampling frequency in Hz (default: 1000).
+        pub frequency: Option<i32>,
+    }
+
+    pub async fn handle_cpu_profile(
+        params: Query<ProfileParams>,
+    ) -> Result<Response, (StatusCode, String)> {
+        let seconds = params.seconds.unwrap_or(30).min(300);
+        let frequency = params.frequency.unwrap_or(1000).max(1);
```
ElFantasma, line 200: nit: `seconds` is capped at 300 but `frequency` has no upper bound — a request with `frequency=1000000` (1 MHz) would attempt to fire a signal every microsecond, which could saturate the CPU with profiling overhead alone. Consider adding `.min(10000)` or similar to match pprof-rs's practical limits.

crates/networking/rpc/rpc.rs
```diff
@@ -158,6 +158,97 @@ pub async fn handle_get_heap_flamegraph() -> Result<(), (StatusCode, String)> {
...
+    fn internal_error(msg: &str, err: impl std::fmt::Display) -> (StatusCode, String) {
+        (StatusCode::INTERNAL_SERVER_ERROR, format!("{msg}: {err}"))
+    }
+
+    #[derive(Deserialize)]
+    pub struct ProfileParams {
+        /// Capture duration in seconds (default: 30, max: 300).
+        pub seconds: Option<u64>,
+        /// Sampling frequency in Hz (default: 1000).
+        pub frequency: Option<i32>,
```
ElFantasma, line 193: nit: `i32` matches pprof-rs's API, but `frequency` semantically can't be negative. Using `Option<u32>` (or `Option<u16>`) would reject negative values at deserialization time rather than silently clamping via `.max(1)`. Minor since the clamp is correct.

PR #6147, opened 2026-02-06
chore(l1): replace unjustified panics with proper error propagation

iovoid:
**Motivation**

We don't want panics in production code, since they can ungracefully crash the node.

**Description**

Removes some panics.

This makes progress towards goal 1.1 of the [UX/DevEx](https://github.com/lambdaclass/ethrex/pull/6107) roadmap


crates/networking/p2p/sync/healing/storage.rs
```diff
@@ -441,7 +441,19 @@ async fn zip_requeue_node_responses_score_peer(
...
+            received = nodes_size,
+            "Peer responded with more trie nodes than requested"
+        );
+        *failed_downloads += 1;
+        peer_handler
+            .peer_table
+            .record_failure(&request.peer_id)
+            .await?;
+        download_queue.extend(request.requests);
+        return Ok(None);
```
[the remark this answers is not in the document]
iovoid: The behavior goes against spec, a node exhibiting this behavior is malfunctioning.

The peer is in fact penalized.

crates/blockchain/dev/block_producer.rs
```diff
@@ -55,9 +55,17 @@ pub async fn start_block_producer(
...
+            Some(id) => id,
+            None => {
+                tracing::error!(
+                    "Failed to produce block: payload_id is None in ForkChoiceResponse"
+                );
+                sleep(Duration::from_millis(300)).await;
+                tries += 1;
+                continue;
+            }
+        };
```
ilitteri: `let Some = else` pattern here.

crates/networking/p2p/sync/state_healing.rs
```diff
@@ -386,16 +387,14 @@ fn commit_node(
...
-        return; // Case where we're saving the root
+        return Ok(()); // Case where we're saving the root
     }
 
-    let mut membatch_entry = membatch.remove(parent_path).unwrap_or_else(|| {
-        panic!("The parent should exist. Parent: {parent_path:?}, path: {path:?}")
-    });
+    let mut membatch_entry = membatch.remove(parent_path).ok_or(SyncError::CorruptPath)?;
 
     membatch_entry.children_not_in_storage_count -= 1;
```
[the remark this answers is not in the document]
iovoid: The invariant is correct.

crates/blockchain/dev/block_producer.rs
iovoid: Improved in 5511f8c5d4

azteca1998 (approved):

crates/networking/p2p/network.rs
```diff
@@ -102,12 +102,12 @@ pub enum NetworkError {
     DiscoveryServerError(#[from] DiscoveryServerError),
     #[error("Failed to start Tx Broadcaster: {0}")]
     TxBroadcasterError(#[from] TxBroadcasterError),
+    #[error("Failed to bind UDP socket: {0}")]
+    UdpSocketError(#[from] std::io::Error),
```
ElFantasma: nit: Adding a blanket `From<std::io::Error>` to `NetworkError` means any `io::Error` from any source in a function returning `Result<_, NetworkError>` will silently become `UdpSocketError`. Currently `start_network` only has the one UDP bind site, but this could be misleading if the function grows. A scoped `.map_err()` at the call site would be more precise:
```rust
let udp_socket = UdpSocket::bind(context.local_node.udp_addr())
    .await
    .map_err(|e| NetworkError::UdpSocketError(e))?;
```
and drop the `#[from]` on the variant.

crates/networking/p2p/sync/healing/state.rs
```diff
@@ -376,16 +377,16 @@ fn commit_node(
...
-        return; // Case where we're saving the root
+        return Ok(()); // Case where we're saving the root
     }
 
-    let mut healing_queue_entry = healing_queue.remove(parent_path).unwrap_or_else(|| {
-        panic!("The parent should exist. Parent: {parent_path:?}, path: {path:?}")
-    });
+    let mut healing_queue_entry = healing_queue
+        .remove(parent_path)
+        .ok_or(SyncError::CorruptPath)?;
```
ElFantasma: `SyncError::CorruptPath` is semantically wrong here — that variant was introduced for filesystem path failures (`create_dir_all`, `DirEntry`), not for a missing parent in the healing queue. More importantly, the old panic message included `parent_path` and `path` which are very useful for debugging sync issues; this replacement loses that context entirely.

Consider a dedicated variant like `SyncError::HealingQueueInconsistency(String)` or at least something that carries the parent/child path info, e.g.:
```rust
let mut healing_queue_entry = healing_queue.remove(parent_path).ok_or_else(|| {
    SyncError::Custom(format!("Parent not found in healing queue. Parent: {parent_path:?}, path: {path:?}"))
})?
```
(assuming a `Custom(String)` variant exists or is added).

crates/networking/p2p/network.rs
iovoid: Done in cbdbce5870

crates/networking/p2p/sync/healing/state.rs
iovoid: Done in cbdbce5870. Also replaced CorruptPath with something clearer.

PR #6151, opened 2026-02-06
test(l1): add restart stall reproduction test using eth-docker

pablodeymo:
## Motivation

A Discord user reported that ethrex sometimes stalls while downloading headers after a restart when running inside [eth-docker](https://github.com/ethstaker/eth-docker). This is intermittent and hard to catch manually, so we need an automated reproduction test.

## Description

Adds `tooling/sync/restart_stall_test.py` — a Python script that automates the full reproduction cycle using eth-docker with ethrex (execution) + Prysm (consensus).

### How it works

**Phase 1 — Fresh snap sync:**
1. Stops any existing eth-docker containers and removes volumes (`docker compose down -v`)
2. Starts eth-docker (`docker compose up -d`)
3. Polls `eth_syncing` RPC until sync completes (8h timeout)
4. Verifies block progress post-sync (22m of new blocks)
5. Sends Slack notification on completion

**Phase 2 — Continuous restart cycles (runs forever, Ctrl+C to stop):**

Each cycle:
1. Removes execution and consensus containers (`docker compose rm -f -s`)
2. Removes all data volumes: `ethrex-el-data`, `prysmconsensus-data`, `prysmvalidator-data` (preserves JWT)
3. Recreates containers with fresh volumes (`docker compose up -d execution consensus`)
4. Waits for node RPC to come up
5. Waits for full snap sync completion (8h timeout)
6. Verifies block progress post-sync (22m)
7. Reports PASS/FAIL with sync time
8. Sends Slack notification on failure
9. Saves execution + consensus logs
10. Repeats from step 1

On Ctrl+C, prints a summary of all completed cycles and saves it to `summary.txt`.

Use `--restart-count N` to limit to N cycles instead of running forever.

**Phase 2 with `--keep-data` — Quick restart (no wipe):**
1. Stops only the execution client (`docker compose stop execution`), keeping consensus running and volumes intact
2. Restarts the execution client (`docker compose start execution`)
3. Monitors for stall: watches `eth_blockNumber` progress, detects if the node stops advancing
4. Reports result: `ok` (caught up), `stall` (no progress), `timeout` (still progressing but didn't catch up in time), or `unresponsive` (node never came back)

This mode is useful for testing restart recovery without re-syncing.

### Configuration

The `--configure` flag auto-writes eth-docker's `.env` from `default.env` with:
- `COMPOSE_FILE=prysm.yml:ethrex.yml:el-shared.yml` (el-shared.yml publishes the RPC port to the host for monitoring)
- `NETWORK` (default: hoodi)
- `CHECKPOINT_SYNC_URL` (auto-set based on network: mainnet, hoodi, holesky, sepolia)
- `FEE_RECIPIENT` (optional, via `--fee-recipient`)
- `ETHREX_DOCKER_REPO=ghcr.io/lambdaclass/ethrex`
- Slack webhook URLs from the local `.env`

Uses `docker compose` directly instead of `./ethd` commands to avoid interactive confirmation prompts when running non-interactively (e.g. in tmux).

### Timeouts (all configurable via env vars, all in seconds)

| Variable | Default | Description |
|----------|---------|-------------|
| `SYNC_TIMEOUT` | 8 hours | Max time for initial sync |
| `BLOCK_PROCESSING_DURATION` | 22 min | Post-sync block progress verification |
| `BLOCK_STALL_TIMEOUT` | 10 min | Time without progress during initial sync to declare stall |
| `RESTART_STALL_TIMEOUT` | 15 min | Time without progress after restart to declare stall |
| `NODE_STARTUP_TIMEOUT` | 5 min | Time for node to respond after restart |
| `CHECK_INTERVAL` | 10 sec | RPC polling interval |

### Makefile targets

| Target | Description |
|--------|-------------|
| `make restart-stall-test` | Full test: configure, sync, then continuous restart cycles (Ctrl+C to stop) |
| `make restart-stall-test-skip-sync` | Skip phase 1 (node must already be synced) |
| `make restart-stall-test-keep-data` | Restart without wiping data (stop/start only) |

Variables: `ETH_DOCKER_DIR` (default: `~/eth-docker`), `RESTART_TEST_NETWORK` (default: `hoodi`), `RESTART_TEST_COUNT` (default: `0` = infinite), `RESTART_TEST_FEE_RECIPIENT` (optional).

### Files changed

| File | Change |
|------|--------|
| `tooling/sync/restart_stall_test.py` | Main script with Phase 1/2 logic, continuous cycling, data wipe, Slack notifications |
| `tooling/sync/Makefile` | Add `restart-stall-test`, `restart-stall-test-skip-sync`, and `restart-stall-test-keep-data` targets |

## How to Test

**Prerequisites:** eth-docker cloned at `~/eth-docker`, Docker installed.

```bash
cd tooling/sync

# Full test: configure, sync, then loop forever (Ctrl+C to stop)
make restart-stall-test \
  RESTART_TEST_NETWORK=mainnet \
  RESTART_TEST_FEE_RECIPIENT=0xYourAddress

# Limit to 5 cycles
make restart-stall-test \
  RESTART_TEST_NETWORK=mainnet \
  RESTART_TEST_COUNT=5

# Skip initial sync (node already synced)
make restart-stall-test-skip-sync \
  RESTART_TEST_NETWORK=mainnet

# Quick restart without wipe (stop/start only)
make restart-stall-test-keep-data \
  RESTART_TEST_NETWORK=mainnet

# Or directly with Python
PYTHONUNBUFFERED=1 python3 restart_stall_test.py \
  --eth-docker-dir ~/eth-docker \
  --configure \
  --network mainnet \
  --fee-recipient 0xYourAddress \
  --ethrex-dir ~/ethrex
```

Logs are saved to `tooling/sync/restart_stall_logs/run_YYYYMMDD_HHMMSS/`.

### DIFF
--- a/tooling/sync/restart_stall_test.py
+++ b/tooling/sync/restart_stall_test.py
@@ -539,8 +539,8 @@ def main():
                         help="Ethereum address for EL rewards (FEE_RECIPIENT in eth-docker)")
     parser.add_argument("--rpc-port", type=int, default=8545,
                         help="RPC port for ethrex (default: 8545)")
-    parser.add_argument("--restart-count", type=int, default=3,
-                        help="Number of restart cycles to test (default: 3)")
+    parser.add_argument("--restart-count", type=int, default=0,
+                        help="Number of restart cycles (0 = infinite, stop with Ctrl+C)")
     parser.add_argument("--no-slack", action="store_true",
                         help="Disable Slack notifications")
     parser.add_argument("--skip-phase1", action="store_true",
@@ -598,7 +598,7 @@ def main():
     print(f"  RPC:        {rpc_url}")
     print(f"  Branch:     {branch}")
     print(f"  Commit:     {commit}")
-    print(f"  Restarts:   {args.restart_count}")
+    print(f"  Restarts:   {'infinite (Ctrl+C to stop)' if args.restart_count == 0 else args.restart_count}")
     print(f"  Wipe data:  {not args.keep_data}")
     print(f"  Logs:       {run_dir}")
     print()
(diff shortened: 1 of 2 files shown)


tooling/sync/restart_stall_test.py
```diff
@@ -0,0 +1,717 @@
...
+    print(f"\n  Monitoring restart for stall (timeout: {fmt_time(timeout)}, stall: {fmt_time(stall_timeout)})...")
+    start = time.time()
+
+    # Wait for node to come back up
+    if not wait_for_node(rpc_url, NODE_STARTUP_TIMEOUT):
+        elapsed = time.time() - start
+        return "unresponsive", f"Node never responded after {fmt_time(elapsed)}"
+
+    # Monitor: is it syncing? Is it making progress?
+    last_block = rpc_block_number(rpc_url) or 0
```
ElFantasma, line 378: nit: `rpc_block_number(rpc_url) or 0` conflates two states: node at block 0 (valid) and RPC unreachable (returns `None`). With the block-0 fix in `rpc_block_number`, this should use explicit `None` check:
```python
last_block = rpc_block_number(rpc_url)
if last_block is None:
    last_block = 0
```

tooling/sync/restart_stall_test.py
```diff
@@ -0,0 +1,717 @@
...
+def wipe_data_volumes(eth_docker_dir: str):
+    """Remove and recreate containers and their data volumes for a fresh start.
+
+    Stops and removes the execution and consensus containers, then removes all
+    data volumes (EL, consensus, validator) while preserving the JWT secret.
+    """
+    docker_compose_in_ethd(eth_docker_dir, "rm", "-f", "-s", "execution", "consensus", check=True)
+
+    project = os.path.basename(eth_docker_dir)
+    volumes = [
```
ElFantasma, line 197: Docker Compose normalizes the project name: it lowercases it, strips non-alphanumeric characters (except `-` and `_`), and replaces path separators. `os.path.basename` returns the raw directory name, so if it contains uppercase or special chars (e.g. `Eth-Docker` → project `eth-docker`), the volume names won't match and removal silently fails.

Suggestion: use `docker compose config --format json` to get the actual project name, or normalize to match Compose's behavior:
```python
import re
project = re.sub(r'[^a-z0-9_-]', '', os.path.basename(eth_docker_dir).lower())
```
Or list/inspect the actual volumes:
```python
result = subprocess.run(["docker", "volume", "ls", "--filter", f"name={project}_", "--format", "{{.Name}}"], capture_output=True, text=True)
```

tooling/sync/restart_stall_test.py
```diff
@@ -0,0 +1,717 @@
...
+        payload = {"jsonrpc": "2.0", "method": method, "params": params or [], "id": 1}
+        resp = requests.post(url, json=payload, timeout=5)
+        return resp.json().get("result")
+    except Exception:
+        return None
+
+
+def rpc_block_number(url: str):
+    result = rpc_call(url, "eth_blockNumber")
+    if result:
```
ElFantasma, line 166: `int("0x0", 16)` is `0`, which is falsy — so when the node is at block 0 (e.g. right after a wipe+restart), this returns `None` as if the RPC call failed.

Suggestion:
```python
def rpc_block_number(url: str):
    result = rpc_call(url, "eth_blockNumber")
    if result is not None:
        return int(result, 16)
    return None
```

PR #6152, opened 2026-02-06
feat(l1): integrate GenServer block builder into ethrex --dev mode

ilitteri:
## Motivation

The `ethrex --dev` mode previously used an Engine API-based block producer that built blocks at fixed 1000ms intervals regardless of transaction arrival. This was slow for development and didn't match how a real dev node should behave. Meanwhile, the standalone `ethrex-dev` binary had a GenServer-based block builder that supports both on-demand and interval-based block production, but required running a separate binary.

## Description

This PR replaces the Engine API block producer with the GenServer block builder from `ethrex-block-builder`, making `ethrex --dev` the single dev node entry point.

**Key changes:**

- **Decoupled tx forwarding channel**: Added `dev_tx_sender: Option<UnboundedSender<(Box<Transaction>, Option<BlobsBundle>)>>` to `RpcApiContext`. This avoids coupling `ethrex-rpc` directly to `ethrex-block-builder` — a forwarding task bridges the channel to the GenServer.

- **Optional auth RPC**: Made `authrpc_addr` optional in `start_api()` since dev mode doesn't need a consensus client or auth RPC server.

- **Transaction forwarding**: `SendRawTransactionRequest::handle()` now forwards transactions to the block builder via the channel (fire-and-forget, after mempool insertion).

- **New CLI arguments**:
  - `--dev.block-time <ms>`: Interval-based block production (default: on-demand)
  - `--dev.coinbase <address>`: Set the block coinbase/fee recipient

- **New `init_l1_dev()` function**: Spawns the `BlockBuilder` GenServer, sets up the forwarding task, and starts only the HTTP RPC server (no P2P, no sync, no auth RPC).

- **Dependency swap**: Replaced `ethrex-dev` with `ethrex-block-builder` in the `dev` feature.

**Architecture:**

```
 cast send / curl                    ┌──────────────────┐
       │                             │  BlockBuilder     │
       ▼                             │  (GenServer)      │
 ┌───────────┐   tx channel    ┌───►│                    │
 │ HTTP RPC  │ ───────────────►│    │  on-demand or      │
 │ :8545     │                 │    │  interval mode      │
 └───────────┘          forwarding  └──────────────────┘
                        task
```

## How to Test

```bash
# Build with dev feature
cargo build --features dev

# On-demand mode (default) — blocks built immediately when txs arrive
cargo run --features dev -- --dev --datadir memory

# Interval mode — blocks built every 5 seconds
cargo run --features dev -- --dev --dev.block-time 5000 --datadir memory

# Send a transaction
cast send --private-key 0x385c546456b6a603a1cfcaa9ec9494ba4832da08dd6bcf4de9a71e4a01b74924 \
  0x0000000000000000000000000000000000000001 --value 1 --rpc-url http://localhost:8545
```

### DIFF
--- a/docs/roadmaps/ethrex-dev.md
+++ b/docs/roadmaps/ethrex-dev.md
@@ -1,0 +1,111 @@
+# ethrex dev node — Roadmap
+
+> The path to becoming the default development environment for Ethereum.
+
+## Vision
+
+ethrex dev is not a simulator. It's a **real Ethereum execution client** running in development mode — same EVM, same storage engine, same RPC stack that runs in production. This is the foundation everything else builds on: when your contracts work on ethrex dev, they work on Ethereum. No translation layer, no behavioral gaps, no surprises at deployment.
+
+The roadmap doubles down on this identity while adding the developer tooling that turns a correct execution client into an indispensable development environment.
+
+---
+
+## Phase 1 — Developer Primitives
+
+**Goal:** Every workflow a contract developer needs, built on a real execution client.
+
+| Feature | Description |
+|---------|-------------|
+| **Time manipulation** | `evm_increaseTime`, `evm_setNextBlockTimestamp`, `evm_mine` — control block timestamps for testing time-dependent logic (vesting, lockups, governance) |
+| **State snapshots** | `evm_snapshot` / `evm_revert` — checkpoint and rollback state during test runs without restarting the node |
+| **Account impersonation** | Send transactions as any address without its private key — essential for testing governance, multisig, and access-controlled contracts |
+| **State overrides** | Override balances, nonces, code, and storage in `eth_call` and `eth_estimateGas` without modifying actual state |
+| **Auto-mine controls** | Toggle between instant mining, interval mining, and manual mining (`evm_setAutomine`, `evm_setIntervalMining`) |
+| **Configurable base fee** | Set and manipulate base fee for gas price testing |
+
+**Why ethrex is different here:** These aren't bolted onto a mock — they manipulate a real client's state. The EVM behavior is identical to what validators run.
+
+---
+
+## Phase 2 — Live Chain Integration
+
+**Goal:** Develop against real mainnet/testnet state without leaving ethrex.
+
+| Feature | Description |
+|---------|-------------|
+| **Fork mode** | `--fork-url <RPC>` — fork any EVM chain at its latest block. Run your contracts against real mainnet state locally. |
+| **Fork pinning** | `--fork-block-number <N>` — pin to a specific block for reproducible tests |
+| **Lazy state loading** | Fetch storage slots on-demand from the fork source, cache locally. Minimal startup time even for mainnet forks. |
+| **State diff view** | Expose what your transactions changed relative to the forked state — useful for auditing and understanding side effects |
+
+**Why ethrex is different here:** Fork mode in a full execution client means the forked environment has the same consensus rules, the same precompile behavior, the same edge cases as the real chain. The execution path is identical.
+
+---
+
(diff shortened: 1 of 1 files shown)


cmd/ethrex/initializers.rs
```diff
@@ -262,35 +263,152 @@ pub async fn init_network(
...
-    let head_block_hash = {
-        let current_block_number = store.get_latest_block_number().await.unwrap();
-        store
-            .get_canonical_block_hash(current_block_number)
-            .await
-            .unwrap()
-            .unwrap()
+    // Suppress tracing logs — the banner replaces them in dev mode.
+    if let Some(ref handler) = log_filter_handler {
+        let _ = handler.reload(EnvFilter::new("off"));
```
ElFantasma, line 287: Setting the filter to `"off"` suppresses **all** tracing output — including `error!` and `warn!` from the block builder, RPC server, and storage layer. If a block build fails or the RPC server hits an error after startup, the user sees nothing.

Consider using `"warn"` or `"error"` instead so that important diagnostics are still visible:
```rust
let _ = handler.reload(EnvFilter::new("error"));
```
Or even `"ethrex=error"` to keep the quiet banner experience while still surfacing failures.

crates/networking/rpc/eth/transaction.rs
```diff
@@ -605,20 +605,31 @@ impl RpcHandler for SendRawTransactionRequest {
...
                 .blockchain
                 .add_transaction_to_pool(self.to_transaction())
                 .await
         }?;
+
+        // In dev mode, forward to the block builder (fire-and-forget).
+        // The transaction is already in the mempool, so we don't fail the
+        // RPC call if the channel is closed.
+        if let Some(ref sender) = context.dev_tx_sender {
+            let _ = sender.send((Box::new(self.to_transaction()), blobs_bundle));
```
ElFantasma, line 630: Two concerns here:

1. **`self.to_transaction()` is called twice** — once at line 619/622 for the mempool add, and again here for the channel. This clones the full transaction unnecessarily. For EIP-4844, `blobs_bundle` (up to ~4.7 MB for 6 blobs) is also cloned for the channel despite being unused.

2. **The block builder ignores this data** — `build_block_with_txs` only checks the vec for emptiness, then builds the block from mempool via `initiate_payload_build`. The channel is effectively just a "new tx arrived" signal.

Since the tx is already in the mempool before this send, the channel only needs to carry a lightweight signal (e.g., the tx hash or even `()` for on-demand mode). This avoids the extra clone and makes the data flow clearer.

crates/blockchain/dev/builder.rs
```diff
@@ -0,0 +1,329 @@
...
+            .unwrap_or_default()
+            .as_secs();
+
+        // Ensure timestamp is always increasing
+        let timestamp = std::cmp::max(now, self.last_timestamp + 1);
+        self.last_timestamp = timestamp;
+        timestamp
+    }
+
+    async fn build_block_with_txs(
```
ElFantasma, line 162: The `txs` parameter is misleading — `build_block_with_txs` never injects these transactions into the block. It only uses `txs.is_empty()` as a guard (line 166), then builds the payload via `initiate_payload_build` + `get_payload`, which reads from the **mempool**.

In interval mode, `pending_txs` accumulates full `(Transaction, Option<BlobsBundle>)` tuples that are `mem::take`'d and passed here but never consumed. A `pending_tx_count: usize` (or even just checking the mempool) would be more accurate and avoid holding transaction data in memory.

If the intent is to eventually pass transactions directly to the builder (bypassing the mempool), that would be a different architecture — worth a TODO comment clarifying the current design.

PR #6155, opened 2026-02-06
SP1 Hypercube baseline: prover instrumentation, load test fixes, and tooling

avilagaston9:
## Motivation

Preparation for benchmarking SP1 proving performance before upgrading to SP1 Hypercube. This branch includes the tooling changes needed to run reproducible proving benchmarks on the L2.

## Description

**Prover instrumentation:**
- Switch prover loop to use `prove_timed()` instead of `prove()`, logging batch number and elapsed time with structured fields
- Add `scripts/sp1_bench_metrics.sh` to parse prover logs and build a CSV + summary table

**Deployer fix:**
- Fix stale VK file paths in `cmd/ethrex/l2/deployer.rs` after the guest program was moved from `crates/l2/prover/src/ethrex_guest_program/` to `crates/guest-program/`

**Load test improvements:**
- Use `BlockTag::Pending` instead of `Latest` when fetching the starting nonce, so consecutive runs don't restart from nonce 0
- `wait_until_all_included` now tracks per-account target nonces instead of a flat `tx_amount`
- Add `--endless` flag for continuous load generation (infinite send+wait loop)

## Test plan

- [ ] Start L2 with `--mempool.maxsize 200000`, run load test, verify txs are included
- [ ] Run load test twice consecutively, verify second run picks up correct nonces
- [ ] Run with `--endless` flag, verify it loops automatically

### DIFF
--- a/crates/l2/sequencer/l1_committer.rs
+++ b/crates/l2/sequencer/l1_committer.rs
@@ -971,7 +971,7 @@ impl L1Committer {
             #[allow(clippy::as_conversions)]
             let blob_usage_percentage = blob_size as f64 * 100_f64 / ethrex_common::types::BYTES_PER_BLOB_F64;
             let batch_gas_used = batch_gas_used.try_into()?;
-            let batch_size = (last_added_block_number - first_block_of_batch).try_into()?;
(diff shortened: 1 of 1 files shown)


avilagaston9:
Superseded by #6157

PR #6156, opened 2026-02-08
refactor(l1,l2): remove duplicate bytes-to-nibbles conversion

viktorking7:
The `compact_to_hex` function in `nibbles.rs` contained a duplicate implementation of the bytes-to-nibbles conversion logic that already exists in `Nibbles::from_raw`. This duplication increases maintenance burden and risk of divergence.


ElFantasma (approved):

viktorking7:
@ElFantasma lint fixed

PR #6157, opened 2026-02-09
feat(l2): add prover benchmarking tooling and documentation

avilagaston9:
## Motivation

Add tooling to make prover benchmarking workflows reproducible and ergonomic: structured timing logs, a results collection script, load test improvements, and documentation. Also rewrite the proof coordinator's batch assignment logic to properly detect and surface version mismatches and prover type rejections.

## Description

**Proof coordinator rewrite:**
- Rewrite `handle_request` with a clear 5-step batch assignment protocol: (1) prover type check, (2) resolve next batch, (3) proof already exists, (4) batch existence + version check, (5) input version match. Each step is a sequential early return with comments explaining the logic.
- Rename `NoBatchForVersion` to `VersionMismatch` and remove the `commit_hash` field. The old name was misleading and the returned hash was confusing since sometimes the prover and coordinator versions match but the stored witness version differs.
- Add `ProverTypeNotNeeded` response so the coordinator can permanently reject provers whose backend isn't needed for the deployment. The prover logs an error and skips that coordinator, continuing with others.
- Add `InputRequest` enum (`Batch`, `RetryLater`, `ProverTypeNotNeeded`) so `request_new_input` returns typed outcomes instead of overloading `Option`/errors. This replaces the `process::exit(1)` that fired when the coordinator rejected the prover's type.
- Remove redundant `all_proofs_exist` loop and `contains_batch` check from `handle_request`.
- Document the batch assignment protocol in `docs/l2/architecture/prover.md` with step-by-step flow and prover-side handling table.

**Bug fixes:**
- Fix stale VK file paths in deployer after guest program was moved out of prover
- Fix off-by-one in `batch_size` metric (single-block batch reported as 0)
- Fix `--sp1` flag in Makefile to pass explicit `true` value (the arg uses `ArgAction::Set`, not `SetTrue`)

**Load test improvements:**
- Use pending nonce for consecutive runs without nonce conflicts
- Add `--endless` flag for continuous load generation
- Add env var support: `LOAD_TEST_RPC_URL`, `LOAD_TEST_TX_AMOUNT`, `LOAD_TEST_ENDLESS`

**Prover instrumentation:**
- Add proving time instrumentation (`prove_timed`) and log-parsing benchmark script (`scripts/bench_metrics.sh`)
- Enrich benchmark script with Prometheus batch metadata (gas, txs, blocks)
- Add `--timed` flag to prover CLI to control whether proving time is measured (default: disabled). Set via `PROVER_CLIENT_TIMED` env var.

**Proof coordinator (other):**
- Add `prover_type` field to `BatchRequest` so the coordinator knows which backend the client runs
- Add `prover_type()` method to the `ProverBackend` trait

**Documentation:**
- Add prover benchmarking guide (`docs/l2/prover-benchmarking.md`)
- Add agent workflow for prover benchmarking (`docs/workflows/prover_benchmarking.md`)

## Checklist

### DIFF
--- a/crates/l2/prover/src/prover.rs
+++ b/crates/l2/prover/src/prover.rs
@@ -171,10 +171,10 @@ impl<B: ProverBackend> Prover<B> {
                 input,
                 format,
             } => (batch_number, input, format),
-            ProofData::NoBatchForVersion { commit_hash } => {
+            ProofData::VersionMismatch => {
                 warn!(
                     "Version mismatch: the next batch to prove was built with a different code \
-                     version ({commit_hash}). This prover needs to be updated."
+                     version. This prover may need to be updated."
                 );
                 return Ok(InputRequest::RetryLater);
             }
--- a/crates/l2/sequencer/proof_coordinator.rs
+++ b/crates/l2/sequencer/proof_coordinator.rs
@@ -184,8 +184,7 @@ impl ProofCoordinator {
                      version, so this prover is stale.",
                     self.git_commit_hash
                 );
-                let response = ProofData::no_batch_for_version(self.git_commit_hash.clone());
-                send_response(stream, &response).await?;
+                send_response(stream, &ProofData::version_mismatch()).await?;
             } else {
                 debug!("Batch {batch_to_prove} not yet created, prover is ahead of the proposer");
                 send_response(stream, &ProofData::empty_batch_response()).await?;
@@ -205,8 +204,7 @@ impl ProofCoordinator {
                 "Batch {batch_to_prove} exists but has no input for prover version ({commit_hash}), \
                  version mismatch"
             );
-            let response = ProofData::no_batch_for_version(self.git_commit_hash.clone());
-            send_response(stream, &response).await?;
+            send_response(stream, &ProofData::version_mismatch()).await?;
             return Ok(());
         };
 
--- a/crates/l2/common/src/prover.rs
+++ b/crates/l2/common/src/prover.rs
@@ -187,10 +187,11 @@ pub enum ProofData {
     },
 
     /// 4.
-    /// The Server responds with a NoBatchForVersion if the code version is not the same as the one
-    /// generated in the batch.
-    /// The Client can only prove batches of its own version.
-    NoBatchForVersion { commit_hash: String },
+    /// The Server responds with VersionMismatch when the prover's code version
+    /// does not match the version needed to prove the next batch. This can happen
+    /// when the batch was stored with a different version, or when the prover is
+    /// stale and future batches will use a newer version.
+    VersionMismatch,
 
     /// 4b.
     /// The Server responds with ProverTypeNotNeeded when the connecting prover's
@@ -241,9 +242,9 @@ impl ProofData {
         }
     }
 
-    /// Builder function for creating a NoBatchForVersion
-    pub fn no_batch_for_version(commit_hash: String) -> Self {
-        ProofData::NoBatchForVersion { commit_hash }
+    /// Builder function for creating a VersionMismatch
+    pub fn version_mismatch() -> Self {
+        ProofData::VersionMismatch
     }
 
     /// Builder function for creating a BatchResponse
--- a/crates/l2/tee/quote-gen/src/sender.rs
+++ b/crates/l2/tee/quote-gen/src/sender.rs
@@ -42,12 +42,10 @@ pub async fn get_batch(commit_hash: String) -> Result<(u64, ProgramInput), Strin
             }
             _ => Err("No blocks to prove.".to_owned()),
         },
-        ProofData::NoBatchForVersion {
-            commit_hash: server_code_version,
-        } => Err(format!(
-            "Next batch does not match with the current version. Server code: {}, Prover code: {}",
-            server_code_version, commit_hash
-        )),
+        ProofData::VersionMismatch => Err(
+            "Version mismatch: the next batch to prove was built with a different code version"
+                .to_owned(),
+        ),
         _ => Err("Expecting ProofData::Response".to_owned()),
     }
 }
--- a/docs/l2/architecture/prover.md
+++ b/docs/l2/architecture/prover.md
@@ -27,7 +27,7 @@ sequenceDiagram
 
 When a prover sends a `BatchRequest`, the coordinator decides what to respond based on the following checks, evaluated in order:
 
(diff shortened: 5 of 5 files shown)


crates/l2/Makefile
```diff
@@ -100,15 +100,12 @@ deploy-l1: ## 📜 Deploys the L1 contracts
...
+deploy-l1-sp1: ## 📜 Deploys the L1 contracts with SP1 verifier
 	COMPILE_CONTRACTS=true \
 	cargo run --release --features l2,l2-sql --manifest-path ../../Cargo.toml -- l2 deploy \
 	--eth-rpc-url ${L1_RPC_URL} \
 	--private-key ${L1_PRIVATE_KEY} \
-	--risc0.verifier-address 0x00000000000000000000000000000000000000aa \
-	--sp1.deploy-verifier \
-	--tdx.verifier-address 0x00000000000000000000000000000000000000aa \
-	--aligned.aggregator-address 0x00000000000000000000000000000000000000aa \
+	--sp1 \
```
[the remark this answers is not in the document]
avilagaston9: Fixed in 07402411d — changed to `--sp1 true`.

crates/l2/prover/src/backend/openvm.rs
```diff
@@ -64,6 +64,10 @@ impl ProverBackend for OpenVmBackend {
     type ProofOutput = OpenVmProveOutput;
     type SerializedInput = StdIn;
 
+    fn prover_type(&self) -> ProverType {
```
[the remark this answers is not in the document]
avilagaston9: Intentional — OpenVM and ZisK are not yet wired as L2 backends. The `unimplemented!()` is a compile-time-gated guard: you can only select these backends if the corresponding feature flag is enabled, and nobody enables them for production. When they are ready, the correct `ProverType` variant will be returned here.

crates/l2/prover/src/backend/zisk.rs
```diff
@@ -113,6 +113,10 @@ impl ProverBackend for ZiskBackend {
     type ProofOutput = ZiskProveOutput;
     type SerializedInput = ();
 
+    fn prover_type(&self) -> ProverType {
```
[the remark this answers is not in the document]
avilagaston9: Same as OpenVM above — intentional. ZisK is not yet enabled as an L2 backend.

crates/l2/prover/src/prover.rs
```diff
@@ -136,6 +160,14 @@ impl<B: ProverBackend> Prover<B> {
                 );
                 return Ok(None);
             }
+            ProofData::ProverTypeNotNeeded { prover_type } => {
+                error!(
+                    "Proof coordinator does not need {prover_type} proofs. \
+                     This prover's backend is not in the required proof types for this deployment. \
+                     Shutting down."
+                );
+                std::process::exit(1);
```
ElFantasma: `std::process::exit(1)` is abrupt — it skips async runtime shutdown, pending I/O flushes, and all destructors. Since this runs inside a tokio context, consider returning an error from `request_new_input` (e.g. a dedicated `ProverTypeNotNeeded` error variant) and letting the caller in `start()` break out of the loop and shut down cleanly. Or at minimum, use the same pattern as the `_ =>` branch below: `return Err("Prover type not needed by coordinator".to_owned())`.

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -171,54 +184,44 @@ impl ProofCoordinator {
...
-                );
-                debug!("Sending BatchResponse for block_number: {batch_to_prove}");
-                ProofData::batch_response(batch_to_prove, input, format)
-            };
+        let Some(input) = self
+            .rollup_store
+            .get_prover_input_by_batch_and_version(batch_to_prove, &commit_hash)
+            .await?
+        else {
+            send_response(stream, &ProofData::empty_batch_response()).await?;
```
ElFantasma: This changes behavior: previously when `get_prover_input_by_batch_and_version` returned `None`, the coordinator sent `ProofData::no_batch_for_version(commit_hash)`, which caused the prover client to log a specific warning about version mismatch. Now it sends an empty batch response, which the prover interprets as "no batches to prove" — hiding the real reason (version mismatch). Was this intentional? If so, worth a comment explaining why the distinction no longer matters.

crates/l2/prover/src/backend/openvm.rs
```diff
@@ -64,6 +64,10 @@ impl ProverBackend for OpenVmBackend {
...
+        unimplemented!("OpenVM is not yet enabled as a backend for the L2")
```
ElFantasma, line 68: nit: `unimplemented!()` panics at runtime. Since `prover_type()` is called on every `BatchRequest` (via `self.backend.prover_type()` in `request_new_input`), anyone who accidentally starts a prover with `--backend openvm` will get a panic instead of a clean error. This is fine for now since OpenVM isn't wired up, but worth knowing — the trait signature (`-> ProverType` instead of `-> Result<ProverType, _>`) makes this the only option short of changing the trait.

ElFantasma (dismissed):

crates/l2/prover/src/prover.rs
avilagaston9: You were right about `process::exit(1)` being abrupt. I changed the behavior: instead of exiting, the prover now just logs an error and continues with the remaining proof coordinators — since different coordinators may have different configurations. So the shutdown problem is gone.

crates/l2/sequencer/proof_coordinator.rs
avilagaston9: You were right that hiding the version mismatch was wrong. I rewrote the coordinator's `handle_request` to detect it properly. The old `next_batch_to_prove_for_version` scanned past version-mismatched batches, which meant by the time we reached `get_prover_input_by_batch_and_version` the batch didn't exist at all — so the mismatch info was lost. Now the flow is sequential early returns without scanning: resolve the next batch (`1 + latest_sent`), check if the proof exists, check if the batch exists (version mismatch if versions differ), then check if the input matches the version (version mismatch if not). `NoBatchForVersion` is now sent in both version mismatch cases. Added documentation in `docs/l2/architecture/prover.md` with the full protocol.

crates/l2/prover/src/backend/openvm.rs
avilagaston9: Acknowledged — keeping `unimplemented\!()` intentionally since changing the trait signature to `Result` would add complexity to every backend for a case that can't happen in practice (these backends are feature-gated and not wired for L2 yet).

Issue #6098, opened 2026-02-02
L2: normalize error codes between OnChainProposer contracts


avilagaston9:
## Description

The non-based and based `OnChainProposer` contracts use different error formats for proof verification failures:

**Non-based OnChainProposer** (uses error codes):
- RISC0: `"00c"`
- SP1: `"00e"`
- ZisK: `"017"`
- TDX: `"00g"`

**Based OnChainProposer** (uses full strings, except ZisK):
- RISC0: `"OnChainProposer: Invalid RISC0 proof failed proof verification"`
- SP1: `"OnChainProposer: Invalid SP1 proof failed proof verification"`
- ZisK: `"017"` (already normalized)
- TDX: `"OnChainProposer: Invalid TDX proof failed proof verification"`

## Impact

The `l1_proof_sender` needs to detect these errors to delete invalid proofs from the database. Currently it checks for the non-based error codes, which means:
- Works correctly for non-based contract
- Works for ZisK in based contract (same code)
- **Does not work** for RISC0/SP1/TDX in based contract

## Proposed Solution

Update the based `OnChainProposer` to use the same error codes as the non-based version for consistency.

## Related

This issue was identified during PR #6014 (ZisK integration).

PR #6158, opened 2026-02-09
feat(l2): enable distributed proving and multi-batch verification

avilagaston9:
## Motivation

The proof coordinator currently assigns the same batch to every prover that requests work, meaning only one prover can be active at a time. This is a bottleneck when multiple provers are available. Additionally, the proof sender verifies one batch per L1 transaction even when multiple proofs are ready, wasting gas on separate transactions.

## Description

**Proof Coordinator — distributed batch assignment:**
- Track in-flight batch assignments with timestamps using `Arc<std::sync::Mutex<HashMap<(u64, ProverType), Instant>>>` (two-phase lock pattern: brief mutex for scan+assign, storage validation outside lock)
- When a prover requests work, assign the first unassigned or timed-out batch — different provers get different batches
- Clean up assignments when all proof types arrive for a batch or when batches are verified on-chain
- New CLI flag `--proof-coordinator.prover-timeout` (default 600s, env `ETHREX_PROOF_COORDINATOR_PROVER_TIMEOUT`) controls stale assignment timeout

**L1 Proof Sender — multi-batch verification:**
- Collect all consecutive proven batches from `last_verified_batch + 1` and send them in a single `verifyBatches()` transaction
- Always uses `verifyBatches()` (with a single-element array when only one batch is ready)
- On any multi-batch error, fall back to per-batch sending — this prevents the sequencer from getting stuck on gas limit or calldata size issues (see #6173 for adding a proper cap)
- On invalid proof revert during single-batch fallback, delete the offending proof from the store
- Invalid proof detection matches both full error messages (based contract) and error codes (standard contract); see #6098 for normalizing these across contracts

**OnChainProposer contracts (standard + based):**
- Extract shared verification logic into `_verifyBatchInternal()` to avoid code duplication
- Add `verifyBatches(uint256, bytes[], bytes[], bytes[])` that loops over `_verifyBatchInternal()`
- Use `calldata` instead of `memory` for proof array parameters in external functions, avoiding unnecessary calldata-to-memory copies (consistent with `verifyBatchesAligned`)
- Critical ordering preserved: `_getPublicInputsFromCommitment` called before `lastVerifiedBatch` update
- Based contract now enforces sequential verification (`batchNumber == lastVerifiedBatch + 1`), fixing a pre-existing gap
- Timelock and interface updates for `verifyBatches`

**Metrics & Grafana:**
- Add `tx_hash` label to `batch_verification_gas` metric so batches verified in the same multi-batch tx share the same gas value and tx hash
- New "Verification Gas by Batch" xychart panel (batch_number on X, gas on Y, tx_hash in tooltip)

image

**Aligned mode is unchanged** — it already supports multi-batch via `L1ProofVerifier`.

**Note:** The `prover_type()` method on `ProverBackend` and the `prover_type` field in `BatchRequest` overlap with #6157. OpenVM and ZisK use `unimplemented!()` since they are not yet enabled as L2 backends. Whichever PR lands first, the other will resolve on rebase.

## Checklist

### DIFF
--- a/.github/workflows/pr-main_mdbook.yml
+++ b/.github/workflows/pr-main_mdbook.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Check links
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --no-progress --exclude 'localhost' docs/
+          args: --no-progress --exclude 'localhost' --exclude 'medium.com' docs/
           fail: true
 
   deploy:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2026-02-23
+
+- Add precompile result cache shared between warmer and executor threads [#6243](https://github.com/lambdaclass/ethrex/pull/6243)
+
 ### 2026-02-13
 
 - Optimize storage layer for block execution by reducing lock contention and allocations [#6207](https://github.com/lambdaclass/ethrex/pull/6207)
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -709,7 +709,7 @@ pub async fn import_blocks(
             } else {
                 // We need to have the state of the latest 128 blocks
                 blockchain
-                .add_block_pipeline(block)
+                .add_block_pipeline(block, None)
                 .inspect_err(|err| match err {
                     // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
                     ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
@@ -817,7 +817,7 @@ pub async fn import_blocks_bench(
                 .map_err(InvalidBlockError::InvalidBody)?;
 
             blockchain
-                .add_block_pipeline(block)
+                .add_block_pipeline(block, None)
                 .inspect_err(|err| match err {
                     // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
                     ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -620,7 +620,7 @@ pub async fn regenerate_head_state(
             .await?
             .ok_or_else(|| eyre::eyre!("Block {i} not found"))?;
 
-        blockchain.add_block_pipeline(block)?;
+        blockchain.add_block_pipeline(block, None)?;
     }
 
     info!("Finished regenerating state");
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -477,7 +477,7 @@ impl Command {
                         }
 
                         // Execute block
-                        blockchain.add_block_pipeline(block.clone())?;
+                        blockchain.add_block_pipeline(block.clone(), None)?;
 
                         // Add fee config to rollup store
                         rollup_store
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -278,6 +278,15 @@ struct PreMerkelizedAccountState {
     nodes: Vec<TrieNode>,
 }
 
+/// Work item for BAL state trie shard workers.
+struct BalStateWorkItem {
+    hashed_address: H256,
+    info: Option<AccountInfo>,
+    removed: bool,
+    /// Pre-computed storage root from Stage B, or None to keep existing.
+    storage_root: Option<H256>,
+}
+
 impl Blockchain {
     pub fn new(store: Store, blockchain_opts: BlockchainOptions) -> Self {
         Self {
@@ -376,6 +385,7 @@ impl Blockchain {
         block: &Block,
         parent_header: &BlockHeader,
         vm: &mut Evm,
+        bal: Option<&BlockAccessList>,
     ) -> Result<BlockExecutionPipelineResult, ChainError> {
         let start_instant = Instant::now();
 
@@ -405,9 +415,16 @@ impl Blockchain {
                 let warm_handle = std::thread::Builder::new()
                     .name("block_executor_warmer".to_string())
                     .spawn_scoped(s, move || {
-                        // Warming uses the same caching store, sharing cached state with execution
+                        // Warming uses the same caching store, sharing cached state with execution.
+                        // Precompile cache lives inside CachingDatabase, shared automatically.
                         let start = Instant::now();
-                        let _ = LEVM::warm_block(block, caching_store, vm_type);
+                        if let Some(bal) = bal {
+                            // Amsterdam+: BAL-based precise prefetching (no tx re-execution)
+                            let _ = LEVM::warm_block_from_bal(bal, caching_store);
+                        } else {
+                            // Pre-Amsterdam / P2P sync: speculative tx re-execution
+                            let _ = LEVM::warm_block(block, caching_store, vm_type);
+                        }
                         start.elapsed()
                     })
                     .map_err(|e| {
@@ -448,14 +465,22 @@ impl Blockchain {
                 let merkleize_handle = std::thread::Builder::new()
                     .name("block_executor_merkleizer".to_string())
                     .spawn_scoped(s, move || -> Result<_, StoreError> {
-                        let (account_updates_list, accumulated_updates) = self
-                            .handle_merkleization(
+                        let (account_updates_list, accumulated_updates) = if bal.is_some() {
+                            self.handle_merkleization_bal(
+                                rx,
+                                parent_header_ref,
+                                queue_length_ref,
+                                max_queue_length_ref,
+                            )?
+                        } else {
+                            self.handle_merkleization(
                                 s,
                                 rx,
                                 parent_header_ref,
                                 queue_length_ref,
                                 max_queue_length_ref,
-                            )?;
+                            )?
+                        };
                         let merkle_end_instant = Instant::now();
                         Ok((
                             account_updates_list,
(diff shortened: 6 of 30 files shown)


crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -137,32 +140,81 @@ impl ProofCoordinator {
...
+            let Some(input) = self
+                .rollup_store
+                .get_prover_input_by_batch_and_version(candidate, commit_hash)
+                .await?
+            else {
+                if let Ok(mut assignments) = self.assignments.lock() {
+                    assignments.remove(&(candidate, prover_type));
+                }
+                return Ok(None);
+            };
```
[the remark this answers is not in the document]
avilagaston9: Prover inputs are generated sequentially by the sequencer for committed batches. If batch N has no input for the requested `commit_hash`, batch N+1 won't either. Returning `None` ("no work right now, try again later") is the correct behavior — the prover will request work again on its next tick.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+            let risc0_bytes = proofs
                 .get(&ProverType::RISC0)
                 .map(|proof| proof.calldata())
                 .unwrap_or(ProverType::RISC0.empty_calldata())
-                .as_slice(),
-            proofs
+                .into_iter()
+                .next()
+                .unwrap_or(Value::Bytes(vec![].into()));
+            risc0_array.push(risc0_bytes);
```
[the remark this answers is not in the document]
avilagaston9: Each proof type (RISC0, SP1, TDX) produces exactly one `Value::Bytes(...)` element in `calldata()`. The `.into_iter().next()` unwraps that single element. This matches the contract signature `verifyBatches(uint256, bytes[], bytes[], bytes[])` where each proof per batch is a single `bytes` element.

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -137,32 +140,81 @@ impl ProofCoordinator {
...
+                })?;
+
+                assignments.retain(|&(batch, _), _| batch >= base_batch);
+
+                let now = Instant::now();
+                let mut batch = base_batch;
+                // Upper bound: there can be at most assignments.len() consecutive
+                // assigned batches for this prover type.
+                let max_batch =
+                    base_batch.saturating_add(u64::try_from(assignments.len()).unwrap_or(u64::MAX));
```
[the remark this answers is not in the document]
avilagaston9: The assignment map is small in practice (number of active provers × batch range). This is not a real performance concern.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+
+        // On any error with multiple batches, fall back to single-batch sending
+        // so that a gas limit / calldata issue doesn't block the sequencer.
+        // For single-batch failures, try to delete the invalid proof if applicable.
+        if let Err(ref err) = send_verify_tx_result {
+            if batch_count > 1 {
+                warn!("Multi-batch verify failed ({err}), falling back to single-batch sending");
+                for (batch_number, proofs) in batches {
+                    self.send_single_batch_proof(*batch_number, proofs).await?;
+                }
```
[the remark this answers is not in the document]
avilagaston9: Since on-chain verification is sequential (`batchNumber == lastVerifiedBatch + 1`), you can't send batch N+1 if batch N failed. Stopping at the first failure is correct — remaining batches will be retried on the next tick.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+        }
+        Ok(())
+    }
+
+    /// Updates `latest_sent_batch_proof` in the store and removes the
+    /// checkpoint directory for the given batch.
+    async fn finalize_batch_proof(&self, batch_number: u64) -> Result<(), ProofSenderError> {
+        self.rollup_store
+            .set_latest_sent_batch_proof(batch_number)
+            .await?;
```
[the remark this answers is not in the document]
avilagaston9: The `exists()` check already prevents any harm. Batch 0 checkpoint simply doesn't exist, so nothing happens.

crates/l2/contracts/src/l1/OnChainProposer.sol
```diff
@@ -479,6 +468,33 @@ contract OnChainProposer is
...
+            "OnChainProposer: array length mismatch"
+        );
+        for (uint256 i = 0; i < batchCount; i++) {
+            _verifyBatchInternal(
+                firstBatchNumber + i,
+                risc0BlockProofs[i],
+                sp1ProofsBytes[i],
+                tdxSignatures[i]
+            );
+        }
```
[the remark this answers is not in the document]
avilagaston9: Already tracked in #6173.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+        let verify_tx_hash = send_verify_tx_result?;
+
+        metrics!(
+            let tx_hash_str = format!("{verify_tx_hash:?}");
+            let verify_tx_receipt = self
+                .eth_client
+                .get_transaction_receipt(verify_tx_hash)
+                .await?
+                .ok_or(ProofSenderError::UnexpectedError("no verify tx receipt".to_string()))?;
+            let tx_gas: i64 = verify_tx_receipt.tx_info.gas_used.try_into()?;
```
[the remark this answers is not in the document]
avilagaston9: This is intentional — mentioned in the PR description. The Grafana tooltip shows the `tx_hash` so it's clear when batches share a multi-batch tx.

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -247,26 +276,28 @@ impl ProofCoordinator {
...
             // If not, store it
             self.rollup_store
                 .store_proof_by_batch_and_type(batch_number, prover_type, batch_proof)
                 .await?;
         }
+
+        // Remove the assignment for this (batch, prover_type)
+        if let Ok(mut assignments) = self.assignments.lock() {
+            assignments.remove(&(batch_number, prover_type));
+        }
```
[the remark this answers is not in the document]
avilagaston9: Mutex poisoning requires a thread panic while holding the lock, which would indicate bigger problems. The stale assignment is harmless — it will time out and be reassigned.

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -137,32 +140,81 @@ impl ProofCoordinator {
...
+            // so the scan advances past it on next iteration)
+            if self
+                .rollup_store
+                .get_proof_by_batch_and_type(candidate, prover_type)
+                .await?
+                .is_some()
+            {
+                debug!("Proof for {prover_type} already exists for batch {candidate}, skipping");
+                continue;
+            }
```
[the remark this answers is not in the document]
avilagaston9: This is a known benign race — storing a duplicate proof is a no-op. The alternative (holding the mutex across an async store query) would block all other provers from getting assignments, which is worse. This tradeoff is intentional.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+        let single_batch = [(batch_number, proofs.clone())];
+        let result = self
+            .send_verify_batches_tx(batch_number, &single_batch)
+            .await;
+
+        if let Err(EthClientError::RpcRequestError(RpcRequestError::RPCError {
+            ref message, ..
+        })) = result
+        {
+            self.try_delete_invalid_proof(message, batch_number).await?;
```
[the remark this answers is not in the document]
avilagaston9: The behavior is correct: the bad proof is deleted, the error propagates up, and the next tick retries.

docs/l2/fundamentals/distributed_proving.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. Queries the on-chain `lastVerifiedBatch` and `lastCommittedBatch`.
+2. Collects all **consecutive** proven batches starting from `lastVerifiedBatch + 1`, checking that every required proof type is present for each batch.
+3. Sends them in a single `verifyBatches()` call to L1.
+
+For example, if batches 5, 6, 7 are fully proven but batch 8 is missing a proof, only batches 5–7 are sent. Batch 8 waits for its proof.
+
+### Fallback to single-batch sending
+
+On **any** multi-batch error (gas limit exceeded, calldata too large, invalid proof, etc.), the proof sender falls back to sending each batch individually. Since on-chain verification is sequential (`batchNumber == lastVerifiedBatch + 1`), the fallback stops at the first failing batch — remaining batches are retried on the next tick.
```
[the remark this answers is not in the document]
avilagaston9: Since on-chain verification is sequential (`batchNumber == lastVerifiedBatch + 1`), there is no point continuing with the following batches if one fails — they need to be verified in order. The function returns immediately and retries on the next tick.

crates/l2/contracts/src/l1/based/OnChainProposer.sol
```diff
@@ -437,6 +430,35 @@ contract OnChainProposer is
...
+            "OnChainProposer: array length mismatch"
+        );
+        for (uint256 i = 0; i < batchCount; i++) {
+            _verifyBatchInternal(
+                firstBatchNumber + i,
+                risc0BlockProofs[i],
+                sp1ProofsBytes[i],
+                tdxSignatures[i]
+            );
+        }
```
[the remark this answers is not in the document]
avilagaston9: Already tracked in #6173.

crates/l2/contracts/src/l1/based/interfaces/IOnChainProposer.sol
```diff
@@ -74,24 +74,16 @@ interface IOnChainProposer {
...
+    /// @notice Method used to verify one or more consecutive L2 batches in a single transaction.
+    /// @param firstBatchNumber The batch number of the first batch to verify. Must be `lastVerifiedBatch + 1`.
+    /// @param risc0BlockProofs An array of RISC0 proofs, one per batch.
+    /// @param sp1ProofsBytes An array of SP1 proofs, one per batch.
+    /// @param tdxSignatures An array of TDX signatures, one per batch.
+    function verifyBatches(
+        uint256 firstBatchNumber,
+        bytes[] memory risc0BlockProofs,
+        bytes[] memory sp1ProofsBytes,
+        bytes[] memory tdxSignatures
```
iovoid: Probably should be calldata to be consistent with the others.

crates/l2/contracts/src/l1/OnChainProposer.sol
```diff
@@ -479,6 +468,33 @@ contract OnChainProposer is
...
+    ) external override onlyOwner whenNotPaused {
+        require(
+            !ALIGNED_MODE,
+            "008" // Batch verification should be done via Aligned Layer. Call verifyBatchesAligned() instead.
+        );
+        uint256 batchCount = risc0BlockProofs.length;
+        require(batchCount > 0, "OnChainProposer: empty batch array");
+        require(
+            sp1ProofsBytes.length == batchCount && tdxSignatures.length == batchCount,
+            "OnChainProposer: array length mismatch"
```
iovoid, line 486: Shouldn't this use an error code?

crates/l2/contracts/src/l1/based/interfaces/IOnChainProposer.sol
```diff
@@ -74,24 +74,16 @@ interface IOnChainProposer {
...
-        //tdx
-        bytes memory tdxSignature
+    /// @notice Method used to verify one or more consecutive L2 batches in a single transaction.
+    /// @param firstBatchNumber The batch number of the first batch to verify. Must be `lastVerifiedBatch + 1`.
+    /// @param risc0BlockProofs An array of RISC0 proofs, one per batch.
+    /// @param sp1ProofsBytes An array of SP1 proofs, one per batch.
+    /// @param tdxSignatures An array of TDX signatures, one per batch.
+    function verifyBatches(
+        uint256 firstBatchNumber,
+        bytes[] memory risc0BlockProofs,
```
ElFantasma: The interface declares `bytes[] memory` but the implementation at `based/OnChainProposer.sol` uses `bytes[] calldata`. This is a signature mismatch — the function selector computed from the interface won't match the implementation, which will break callers that use the interface ABI.

Should be `calldata` here to match the implementation (and the standard interface which already uses `calldata`):
```solidity
bytes[] calldata risc0BlockProofs,
bytes[] calldata sp1ProofsBytes,
bytes[] calldata tdxSignatures
```

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -137,32 +140,81 @@ impl ProofCoordinator {
...
+                    ProofCoordinatorError::Custom("Assignment lock poisoned".to_string())
+                })?;
+
+                assignments.retain(|&(batch, _), _| batch >= base_batch);
+
+                let now = Instant::now();
+                let mut batch = base_batch;
+                // Upper bound: there can be at most assignments.len() consecutive
+                // assigned batches for this prover type.
+                let max_batch =
```
ElFantasma, line 161: This upper bound is incorrect. `assignments.len()` counts *all* assignments across *all* prover types, but the scan only checks `(batch, prover_type)` — i.e. a single prover type. Consider: `assignments` has 5 entries for `SP1` provers but none for `RISC0`. A RISC0 prover requests work. `max_batch = base_batch + 5`, but the loop should find `base_batch` immediately (no SP1 assignment blocks RISC0). The bound is technically safe here (it won't miss valid batches in this example), but the opposite case is problematic: if there are 0 assignments for ANY type but many batches to skip (because the proof already exists for that type), the `max_batch` is `base_batch + 0 = base_batch`, so the loop immediately exits after checking only batch `base_batch`.

Since the `continue` branch (proof already exists) doesn't go through the `while batch <= max_batch` loop (it's in the outer `loop`), this is actually fine for that path. But a simpler and correct bound would be `last_committed_batch` or just removing the bound entirely — the inner assignments scan is already bounded by the map size implicitly.

crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -137,32 +140,81 @@ impl ProofCoordinator {
...
+                        Some(&assigned_at)
+                            if now.duration_since(assigned_at) > self.prover_timeout =>
+                        {
+                            break;
+                        }
+                        Some(_) => batch += 1,
+                    }
+                }
+
+                assignments.insert(key(batch), now);
```
ElFantasma, line 177: This unconditionally inserts the assignment *before* checking whether the batch actually has prover input or whether the proof already exists. If either check fails below, the cleanup code correctly removes the assignment on the `None` path (line 189), but the "proof already exists" path (the `continue` at line 204) does *not* remove it — it deliberately keeps the assignment to advance past it. However, this means the map grows with stale entries for already-proved batches that will only be cleaned by the `retain` call on the next request.

This isn't a bug per se (the `retain` cleans them), but it means the `assignments.len()` used in `max_batch` above inflates with these phantom entries, making the bound even less accurate.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+
+        let send_verify_tx_result = self.send_verify_batches_tx(first_batch, batches).await;
+
+        // On any error with multiple batches, fall back to single-batch sending
+        // so that a gas limit / calldata issue doesn't block the sequencer.
+        // For single-batch failures, try to delete the invalid proof if applicable.
+        if let Err(ref err) = send_verify_tx_result {
+            if batch_count > 1 {
+                warn!("Multi-batch verify failed ({err}), falling back to single-batch sending");
+                for (batch_number, proofs) in batches {
```
ElFantasma, line 601: In the multi-batch fallback, if `send_single_batch_proof` fails for batch N (e.g. due to an invalid proof that gets deleted), the `?` propagates the error and skips batches N+1, N+2, etc. that might have valid proofs. This is fine for invalid proof errors (the on-chain state hasn't advanced past N), but for transient errors (RPC timeout, network issue), the remaining batches are lost for this tick.

Consider logging and continuing instead of `?` here, since on-chain verification is sequential anyway — batch N+1 can't verify until N succeeds, so bailing early is actually correct. But it's worth a comment explaining why `?` is intentional (on-chain sequential ordering means later batches would also fail).

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
-                    .await?;
-            } else if message.contains("Invalid SP1 proof") {
-                warn!("Deleting invalid SP1 proof");
-                self.rollup_store
-                    .delete_proof_by_batch_and_type(batch_number, ProverType::SP1)
-                    .await?;
-            }
+    /// Returns the prover type whose proof is invalid based on the error message,
+    /// or `None` if the message doesn't indicate an invalid proof.
+    fn invalid_proof_type(message: &str) -> Option<ProverType> {
```
ElFantasma, line 490: nit: The error code matching (`"00c"`, `"00e"`, `"00g"`) could produce false positives on unrelated RPC error messages that happen to contain these short strings. The standard contract's error codes are short enough that a message like "Error at position 00c in calldata" would trigger deletion. Consider matching with a word boundary or prefix, e.g. `message.contains("\"00c\"")` (the JSON-encoded error code) or `message.starts_with("00c")` — depends on what the actual RPC error message format looks like.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -398,85 +427,217 @@ impl L1ProofSender {
...
+
+    /// Sends a single batch proof via verifyBatches, deleting the invalid proof
+    /// from the store if the transaction reverts. On success, updates progress
+    /// and cleans up the checkpoint.
+    async fn send_single_batch_proof(
+        &self,
+        batch_number: u64,
+        proofs: &HashMap<ProverType, BatchProof>,
+    ) -> Result<(), ProofSenderError> {
+        let single_batch = [(batch_number, proofs.clone())];
```
ElFantasma: `proofs.clone()` clones the entire `HashMap<ProverType, BatchProof>` including proof bytes/calldata for each batch in the fallback loop. For proofs containing large Groth16 calldata or compressed STARKs, this is an unnecessary allocation — the original `batches` slice is still alive.

Consider changing `send_single_batch_proof` to take `&HashMap<ProverType, BatchProof>` and building the single-element array with a reference, or restructuring `send_verify_batches_tx` to accept `&[(&u64, &HashMap<...>)]` to avoid the clone entirely.

crates/l2/contracts/src/l1/based/interfaces/IOnChainProposer.sol
avilagaston9: Good catch, fixed — changed to `calldata` to match the implementation.

crates/l2/contracts/src/l1/OnChainProposer.sol
avilagaston9: You're right, it should. This will be addressed by #6206 which replaces all error code strings with proper custom errors.

crates/l2/contracts/src/l1/based/interfaces/IOnChainProposer.sol
avilagaston9: Fixed, changed to `calldata` to match the implementation.

crates/l2/sequencer/proof_coordinator.rs
avilagaston9: The bound is a conservative over-approximation — it works correctly in all cases:

- If there are 0 assignments, `max_batch = base_batch`. The inner while checks `(base_batch, prover_type)`, finds no assignment, breaks, and assigns `base_batch` — correct.
- If there are assignments for other prover types, the inner while still only checks `(batch, prover_type)` for the requesting prover, so those entries don't block anything — correct.
- The "proof already exists" path is in the outer `loop`, not the inner `while`, so it doesn't depend on `max_batch` at all.

The `retain` at the top of each call cleans up stale entries, keeping `assignments.len()` bounded in practice. Making the bound more precise (e.g. filtering by prover type or using `last_committed_batch`) would add complexity without changing behavior.

crates/l2/sequencer/proof_coordinator.rs
avilagaston9: Right, the `retain` on line 157 cleans these up on every request, and the wider `assignments.len()` only makes the bound more conservative (checks more batches), so it doesn't cause any incorrect behavior. Keeping the current structure avoids adding an extra lock acquisition on the "proof already exists" path.

crates/l2/sequencer/l1_proof_sender.rs
avilagaston9: Added a comment explaining why `?` is intentional (on-chain sequential ordering).

crates/l2/sequencer/l1_proof_sender.rs
avilagaston9: These only match against Solidity revert messages from the contract, so the false positive risk is very low in practice. In any case, #6206 replaces all these string error codes with proper custom errors, which will make the matching exact.

crates/l2/sequencer/l1_proof_sender.rs
avilagaston9: Good point — changed `send_verify_batches_tx` to accept `&[(u64, &HashMap<ProverType, BatchProof>)]` so the single-batch path no longer needs to clone.

docs/l2/fundamentals/ethrex_l2_aligned_integration.md
```diff
@@ -282,24 +282,19 @@ mapping(bytes32 commitHash => mapping(uint8 verifierId => bytes32 vk))
...
 ```solidity
-function verifyBatch(
-    uint256 batchNumber,
-    bytes memory risc0BlockProof,
-    bytes memory sp1ProofBytes,
-    bytes memory tdxSignature
+function verifyBatches(
+    uint256 firstBatchNumber,
+    bytes[] memory risc0BlockProofs,
+    bytes[] memory sp1ProofsBytes,
```
ElFantasma: Should be `calldata` to match the interface fix in `based/interfaces/IOnChainProposer.sol`.

avilagaston9: Addressed in 429c7e0ca — changed `bytes[] memory` to `bytes[] calldata` in the docs.

PR #6165, opened 2026-02-10
fix(l1): only rebuild payload when mempool receives new transactions

avilagaston9:
## Motivation

The block building loop (`build_payload_loop`) continuously rebuilds the block for 12 seconds (one slot), even when no new transactions have arrived. Each rebuild is CPU-intensive (EVM execution, trie operations) and contends with RPC handlers over the shared `trie_cache` mutex, starving the RPC server.

## Description

Add an `AtomicU64` generation counter to `Mempool` that increments on every transaction insertion. The payload build loop polls this counter with a 100ms sleep interval and only rebuilds when it detects a change, instead of looping as fast as possible.

- `Mempool::add_transaction()` increments the generation counter after inserting a tx
- `build_payload_loop` compares the current generation to the one from the last build; if unchanged, sleeps 100ms and rechecks
- The first `build_payload` call remains unconditional — we always build one payload immediately when FCU arrives
- No new tokio features required — uses only `std::sync::atomic` and the already-enabled `tokio::time`

## Checklist

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -81,6 +81,8 @@ impl MempoolInner {
 #[derive(Debug, Default)]
 pub struct Mempool {
     inner: RwLock<MempoolInner>,
+    /// Monotonically increasing counter, bumped on every transaction insertion.
+    /// The payload builder polls this to skip rebuilds when the mempool is unchanged.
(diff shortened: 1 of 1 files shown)


avilagaston9:
Superseded by #6169

PR #6169, opened 2026-02-10
fix(l1): only rebuild payload when mempool receives new transactions (Notify)

avilagaston9:
## Motivation

The block building loop (`build_payload_loop`) continuously rebuilds the block for 12 seconds (one slot), even when no new transactions have arrived. Each rebuild is CPU-intensive (EVM execution, trie operations) and contends with RPC handlers over the shared `trie_cache` mutex, starving the RPC server.

## Description

**Alternative approach to #6165** — uses `tokio::sync::Notify` instead of an `AtomicU64` generation counter.

Add a `tokio::sync::Notify` to `Mempool` that signals when new transactions or blobs bundles are inserted. The payload build loop waits on this notification (or cancellation, or the slot deadline) before rebuilding, instead of looping as fast as possible.

- `Mempool::add_transaction()` and `Mempool::add_blobs_bundle()` both call `notify_waiters()` after inserting data, so builders see complete blob tx data and all active payload builders are woken
- `build_payload_loop` uses `tokio::select!` to wait for a mempool notification, `cancel_token` cancellation, or the remaining slot deadline — whichever comes first
- The `notified()` future is pre-captured before `select!` to avoid missing permits between loop iterations
- The first `build_payload` call remains unconditional — we always build one payload immediately when FCU arrives
- `tx_added()` is `pub(crate)` to avoid exposing the notification primitive externally

**Tradeoffs vs #6165 (AtomicU64 polling):**
- Instant wake-up on new tx (no 100ms polling delay)
- Requires adding `sync` and `macros` tokio features to the blockchain crate

## Checklist

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -168,9 +168,6 @@ impl Mempool {
         self.write()?
             .blobs_bundle_pool
             .insert(tx_hash, blobs_bundle);
-        // Notify after the bundle is inserted so builders that woke up from
-        // add_transaction see the complete blob data.
-        self.tx_added.notify_waiters();
         Ok(())
     }
 
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -2057,10 +2057,11 @@ impl Blockchain {
             self.remove_transaction_from_pool(&tx_to_replace)?;
         }
 
-        // Add transaction and blobs bundle to storage
+        // Add blobs bundle before the transaction so that when add_transaction
+        // notifies payload builders the blob data is already available.
+        self.mempool.add_blobs_bundle(hash, blobs_bundle)?;
         self.mempool
             .add_transaction(hash, sender, MempoolTransaction::new(transaction, sender))?;
-        self.mempool.add_blobs_bundle(hash, blobs_bundle)?;
         Ok(hash)
     }
 


crates/blockchain/mempool.rs
```diff
@@ -123,6 +131,9 @@ impl Mempool {
             .insert((sender, transaction.nonce()), hash);
         inner.transaction_pool.insert(hash, transaction);
         inner.broadcast_pool.insert(hash);
+        // Drop the write lock before notifying to avoid holding it while waking waiters
+        drop(inner);
+        self.tx_added.notify_one();
```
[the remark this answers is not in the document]
avilagaston9: Switched to `notify_waiters()` in bf7a919 — all active payload builders are now woken on every insertion.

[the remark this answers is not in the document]
avilagaston9: Added `notify_waiters()` to `add_blobs_bundle()` as well in bf7a919. After `add_blob_transaction_to_pool` calls `add_transaction` (first notification) and then `add_blobs_bundle` (second notification), builders that woke early and failed on the missing bundle get a second wake-up with the complete data.

crates/blockchain/payload.rs
```diff
@@ -390,6 +390,14 @@ impl Blockchain {
...
         let mut res = self.build_payload(payload.clone())?;
         while start.elapsed() < SECONDS_PER_SLOT && !cancel_token.is_cancelled() {
+            // Wait for new transactions, cancellation, or slot deadline before rebuilding
+            let remaining = SECONDS_PER_SLOT.saturating_sub(start.elapsed());
+            let notified = self.mempool.tx_added().notified();
+            tokio::select! {
+                _ = notified => {}
+                _ = cancel_token.cancelled() => break,
+                _ = tokio::time::sleep(remaining) => break,
+            }
```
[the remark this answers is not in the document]
avilagaston9: Fixed in bf7a919 — switched to `notify_waiters()` which wakes all active payload build loops, not just one.

crates/blockchain/payload.rs
```diff
@@ -390,6 +390,14 @@ impl Blockchain {
         // TODO(#4997): start with an empty block
         let mut res = self.build_payload(payload.clone())?;
         while start.elapsed() < SECONDS_PER_SLOT && !cancel_token.is_cancelled() {
+            // Wait for new transactions, cancellation, or slot deadline before rebuilding
+            let remaining = SECONDS_PER_SLOT.saturating_sub(start.elapsed());
+            let notified = self.mempool.tx_added().notified();
+            tokio::select! {
+                _ = notified => {}
+                _ = cancel_token.cancelled() => break,
```
ElFantasma, line 398: nit: the `cancel_token.cancelled()` branch here is redundant with the `while !cancel_token.is_cancelled()` loop condition — if the token fires, the next iteration's condition will catch it anyway. Not harmful (it exits one iteration earlier), but it could be removed for clarity since the sleep branch already provides the timeout exit.

ElFantasma (approved):
Approved with a nit

crates/blockchain/mempool.rs
```diff
@@ -157,6 +168,9 @@ impl Mempool {
         self.write()?
             .blobs_bundle_pool
             .insert(tx_hash, blobs_bundle);
+        // Notify after the bundle is inserted so builders that woke up from
+        // add_transaction see the complete blob data.
```
iovoid: I think in add_blob_transaction_to_pool we should also reorder the calls to add_transaction and add_blobs_bundle, to ensure we don't have transactions without their blobs.

crates/blockchain/payload.rs
avilagaston9: Good point, but I think the branch is actually needed: without it, if `getPayload` fires the cancel token while we're blocked in the `select!`, we'd be stuck waiting for either a notification or the sleep timeout before exiting. The `while` condition only checks between iterations, so the `cancelled()` branch gives us immediate responsiveness to cancellation.

crates/blockchain/mempool.rs
avilagaston9: Good catch\! Reordered the calls so `add_blobs_bundle` happens before `add_transaction`. This way the blob data is already in the pool when `add_transaction` fires `notify_waiters()`, and we can also remove the extra `notify_waiters()` from `add_blobs_bundle` since it's no longer needed.

Issue #6170, opened 2026-02-10
Snap Sync Refactor: download storage slots plan


fedacking:
# Plan: Refactor Storage Download to Use `StorageTrieTracker` 

This is being done in branch `perf/refactor-storage-download-snap`

## Context

The current `AccountStorageRoots` struct in snap sync tracks storage downloads per-account, with complex index-based referencing into `accounts_by_root_hash`. This makes the download loop in `request_storage_ranges` hard to follow: tasks reference accounts by index, results carry index ranges, and big-account promotion involves mutating intervals in the tracking struct.

The new `StorageTrieTracker` (already defined in `sync.rs`) groups storage tries by root hash from the start, separating small (single-request) from big (multi-request) tries. The refactor simplifies data ownership by moving trie data into tasks and back in results, eliminating clones and the index-based referencing.

## Files to Modify

1. **`crates/networking/p2p/sync.rs`** — Add `healed_accounts` field and methods to `StorageTrieTracker`
2. **`crates/networking/p2p/snap/client.rs`** — New `StorageTask`/`StorageTaskResult` enums, refactor `request_storage_ranges` + worker
3. **`crates/networking/p2p/sync/snap_sync.rs`** — Replace `AccountStorageRoots` with `StorageTrieTracker` throughout, update `insert_accounts`
4. **`crates/networking/p2p/sync/healing/state.rs`** — Update parameter types, use `tracker.handle_healed_account()`
5. **`crates/networking/p2p/sync/healing/storage.rs`** — Update `get_initial_downloads()` to read from `tracker.healed_accounts`

---

## Step 1: Extend `StorageTrieTracker` (sync.rs)

Add two fields to `StorageTrieTracker`:
- `healed_accounts: HashSet<H256>` — used by storage healing (`get_initial_downloads`) to know which accounts need storage trie node repair.
- `account_to_root: HashMap<H256, H256>` — reverse lookup from account hash to its current storage root. Maintained by all mutating methods. Enables O(1) lookup of old root during healing.

Add methods:

- **`insert_account(&mut self, account_hash: H256, storage_root: H256)`** — Inserts into `small_tries`, grouping by root. If root already exists (in small or big), appends to its accounts vec. Also inserts into `account_to_root`.

- **`promote_to_big(&mut self, root: H256, first_slots: Vec<Slot>, intervals: Vec<Interval>)`** — Removes from `small_tries`, inserts into `big_tries` with the same accounts + provided slots/intervals.

[...]

PR #6171, opened 2026-02-10
perf(l1): refactor storage download to use StorageTrieTracker

fedacking:
## Summary

- Replace `AccountStorageRoots` with `StorageTrieTracker` throughout snap sync, eliminating index-based referencing and the `accounts_by_root_hash` intermediate structure
- Introduce `StorageTask` / `StorageTaskResult` enums that move trie data into tasks and back in results, removing clones and simplifying the download loop
- Use structured concurrency (`JoinSet` + `try_join_next`) instead of channels for worker communication in `request_storage_ranges`
- Extract `BigTrie::compute_intervals` helper from the inline chunking logic
- Fix peers with the lowest score (-50) being permanently unable to receive any requests. The previous strict inequality (`<`) in `can_try_more_requests` meant a score ratio of 0.0 resulted in `requests < 0.0`, which is always false — effectively blacklisting the peer. Changing to `<=` ensures every connected peer can always handle at least 1 concurrent request.

The full plan for this PR is documented in https://github.com/lambdaclass/ethrex/issues/6170

## Test plan

- [x] `cargo check -p ethrex-p2p` compiles cleanly (default + rocksdb features)
- [x] `cargo clippy -p ethrex-p2p` passes with no warnings (default + rocksdb features)
- [x] `cargo test -p ethrex-p2p` — all 38 tests pass
- [ ] Full snap sync test against a testnet peer (manual verification)

### DIFF
--- a/crates/networking/p2p/snap/client.rs
+++ b/crates/networking/p2p/snap/client.rs
@@ -811,7 +811,14 @@ pub async fn request_storage_ranges(
             chunk_index += 1;
         }
 
-        if let Some(result) = worker_joinset.try_join_next() {
+        // Drain a completed worker: block if we've hit the concurrency cap,
+        // otherwise poll without waiting.
+        let maybe_result = if worker_joinset.len() >= MAX_STORAGE_RANGE_WORKERS {
+            worker_joinset.join_next().await
+        } else {
+            worker_joinset.try_join_next()
+        };
+        if let Some(result) = maybe_result {
             let result = result.expect("Storage worker task panicked");
             let (peer_id, success) = process_storage_task_result(
                 result,
--- a/crates/networking/p2p/snap/constants.rs
+++ b/crates/networking/p2p/snap/constants.rs
@@ -74,6 +74,13 @@ pub const REQUEST_RETRY_ATTEMPTS: u32 = 5;
 /// Maximum number of concurrent in-flight requests during storage healing.
 pub const MAX_IN_FLIGHT_REQUESTS: u32 = 77;
 
+/// Maximum number of concurrent in-flight storage range workers.
+///
+/// Each worker holds up to ~2 MB of in-memory data (512 KB wire response
+/// expanded into Rust structs, plus temporary copies during validation).
+/// 1000 workers × ~2 MB ≈ 2 GB cap on in-flight storage download memory.
+pub const MAX_STORAGE_RANGE_WORKERS: usize = 1000;
+
 // =============================================================================
 // BLOCK SYNC CONFIGURATION
 // =============================================================================


fedacking:
One thing I think would be very useful is adding monitoring to request_storage_ranges so we can observe what's happening during the storage download phase. Right now it's hard to tell the state of progress and the mix of work being done.

  Concretely, I'd like to see periodic debug! logs (or at the very least before/after each request_storage_ranges call) that include:

  1. Small vs big tries count — tracker.small_tries.len() and tracker.big_tries.len() separately. We already log tracker.remaining_count() in snap_sync.rs but that collapses both into one
   number, which hides the distribution.
  2. Total intervals across big tries — the sum of intervals.len() for each entry in tracker.big_tries. This tells us how much sub-range work is still pending for the large tries.
  3. Number of small batch vs big interval requests sent — how many SmallBatch and BigInterval tasks were actually dispatched to workers during the call.

  This would give us good visibility into whether storage download is making progress, whether tries are getting promoted from small to big, and how the interval-based download is evolving over successive attempts. The existing metrics infrastructure (METRICS) seems like the right place to wire these into.

crates/networking/p2p/snap/client.rs
```diff
@@ -643,301 +788,17 @@ pub async fn request_storage_ranges(
...
-
-            METRICS
-                .storage_leaves_downloaded
-                .inc_by(effective_slots as u64);
-
-            debug!("Downloaded {n_storages} storages ({n_slots} slots) from peer {peer_id}");
-            debug!(
-                "Total tasks: {task_count}, completed tasks: {completed_tasks}, queued tasks: {}",
-                tasks_queue_not_started.len()
+        if let Some(Ok(result)) = worker_joinset.try_join_next() {
```
ElFantasma: `Some(Ok(result))` silently drops `Some(Err(JoinError))` — if a worker panics, its trie data is lost without any logging. Consider matching on `Some(Err(e))` to at least log a warning (and mark the peer as failed).

This also creates an inconsistency with `join_all` below (line 847), which *will* panic if any task panicked, crashing the entire sync.

crates/networking/p2p/snap/client.rs
```diff
@@ -965,42 +826,49 @@ pub async fn request_storage_ranges(
...
             peer_table,
             pivot_header.state_root,
-            chunk_account_hashes,
-            chunk_storage_roots,
-            tx,
         ));
     }
 
+    // Drain remaining in-flight workers so their trie data is not lost
+    for result in worker_joinset.join_all().await {
```
ElFantasma, line 878: `join_all` panics if any task returned a `JoinError` (e.g., a worker panicked). Since `try_join_next` above silently discards panicked tasks, this creates an asymmetry: the main loop tolerates panics but the drain loop crashes.

Suggestion: replace with a manual `while let Some(result) = worker_joinset.join_next().await` loop that handles `Err(join_error)` gracefully — log the error and continue draining.

crates/networking/p2p/sync.rs
```diff
@@ -161,16 +161,239 @@ impl Syncer {
...
+        slots_per_chunk: usize,
+    ) -> Vec<Interval> {
+        let start_hash_u256 = U256::from_big_endian(&last_downloaded_hash.0);
+        let missing_storage_range = U256::MAX - start_hash_u256;
+        let slot_count = slot_count.max(1);
+        let storage_density = start_hash_u256 / slot_count;
+        let chunk_size = storage_density
+            .checked_mul(U256::from(slots_per_chunk))
+            .unwrap_or(U256::MAX);
+        let chunk_count = (missing_storage_range / chunk_size).as_usize().max(1);
```
ElFantasma, line 411: Division by zero: when `start_hash_u256` is zero (first slot has hash near zero), `storage_density = 0 / slot_count = 0`, then `chunk_size = 0 * slots_per_chunk = 0`, and this line divides `missing_storage_range / 0` → panic.

This can happen when a trie's first key is very small (near `H256::zero()`). Suggest guarding: `if chunk_size.is_zero() { chunk_size = U256::MAX; }` (single interval covering the whole range).

crates/networking/p2p/sync.rs
```diff
@@ -161,16 +161,239 @@ impl Syncer {
...
+                .get_mut(&old_root)
+                .expect("big_tries should contain old_root");
+            if big.accounts.len() == 1 && big.accounts[0] == account_hash {
+                // Only account: re-key
+                let mut big = self
+                    .big_tries
+                    .remove(&old_root)
+                    .expect("big_tries should contain old_root");
+                big.accounts = vec![account_hash];
+                self.big_tries.insert(new_root, big);
```
ElFantasma: If `new_root` already exists in `big_tries` (another account was already tracked under that root), `insert` silently overwrites the existing entry, losing its accounts/slots/intervals.

Consider merging into the existing entry instead, similar to what `insert_account` does:
```rust
if let Some(existing) = self.big_tries.get_mut(&new_root) {
    existing.accounts.push(account_hash);
} else {
    self.big_tries.insert(new_root, big);
}
```

crates/networking/p2p/snap/client.rs
```diff
@@ -515,85 +540,205 @@ pub async fn request_bytecodes(
...
+            current_account_storages
+                .entry(big_root)
+                .or_insert_with(|| AccountsWithStorage {
+                    accounts: big_trie.accounts.clone(),
+                    storages: Vec::new(),
+                })
+                .storages
+                .extend(storages);
+
+            tracker.promote_to_big(big_root, big_trie.slots, intervals.clone());
```
ElFantasma: `promote_to_big` moves slots from `big_trie` into the tracker via `big_trie.slots`, but two lines above (line 636) those same slots were already iterated with `.iter()` to build `storages`. After this call, `big_trie.slots` is moved and the `SmallTrie` from the tracker's `small_tries` map may no longer have the original accounts (it was taken out by `take_small_batch` earlier). `promote_to_big` falls back to `unwrap_or_default()` which means the big trie may end up with zero accounts. 

This seems correct in the `SmallPromotedToBig` flow because `big_trie` was popped from the worker's `tries` vec (which was taken from the tracker), but it's worth a `debug_assert!(!accounts.is_empty())` after `promote_to_big` to catch this if assumptions change.

crates/networking/p2p/snap/client.rs
fedacking: Fixed in [81e0c9e](https://github.com/lambdaclass/ethrex/pull/6171/commits/81e0c9ebb8aff73e2c1634d314edd27b05b9104d)

crates/networking/p2p/sync.rs
fedacking: Fixed in [02e52b7](https://github.com/lambdaclass/ethrex/pull/6171/commits/02e52b73b71f60074777932078888dff4477776b)

crates/networking/p2p/snap/client.rs
fedacking: Fixed in [81e0c9e](https://github.com/lambdaclass/ethrex/pull/6171/commits/81e0c9ebb8aff73e2c1634d314edd27b05b9104d)

crates/networking/p2p/sync.rs
fedacking: Fixed in [1a051bb](https://github.com/lambdaclass/ethrex/pull/6171/commits/1a051bbde5a96d459de0d2b0d042ca56b4cd6351)

crates/networking/p2p/snap/client.rs
fedacking: Fixed in [19ebc14](https://github.com/lambdaclass/ethrex/pull/6171/commits/19ebc14f3f2a71e385acaaebccc23aefd78ae244)

crates/networking/p2p/snap/client.rs
```diff
@@ -515,85 +540,203 @@ pub async fn request_bytecodes(
...
+                })
+                .storages
+                .extend(storages);
+
+            tracker.promote_to_big(
+                big_root,
+                big_trie.accounts,
+                big_trie.slots,
+                intervals.clone(),
+            );
```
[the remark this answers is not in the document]
fedacking: This was done in [09fc842](https://github.com/lambdaclass/ethrex/pull/6171/commits/09fc842b4a305b693069f7877784421402e046d0)

PR #6178, opened 2026-02-11
perf(l1): disable WAL and improve concurrency in snap sync

ilitteri:
## Motivation

Snap sync performance profiling on Hoodi showed that RocksDB write-ahead log (WAL) overhead and single-peer healing dispatch were limiting throughput. Nethermind's snap sync (which achieves ~20 min on mainnet) disables WAL during sync since all data can be re-downloaded on crash, and uses multi-peer concurrent requests for healing.

## Description

Three changes, each in a separate commit:

**1. Storage layer: `commit_no_wal()` (1848388)**
- Add `commit_no_wal()` to `StorageWriteBatch` trait with default fallback to `commit()`
- RocksDB implementation uses `WriteOptions` with `disable_wal(true)` to skip the write-ahead log
- Wire through `BackendTrieDB` via `with_no_wal()` builder and add `_no_wal` variants to `Store`

**2. Multi-peer state healing (7d18550)**
- Extract `dispatch_state_healing_batches()` that sends up to `MAX_IN_FLIGHT_REQUESTS` (77) concurrent trie node requests instead of one at a time
- Replace `try_recv()` busy-polling with `tokio::select!` on channel recv + 1s timeout
- Process all pending healed batches with `while let` instead of `if let` to drain the queue each iteration

**3. Wire no-WAL through snap sync + fix storage healing polling (5730189)**
- Switch all snap sync DB writes to `_no_wal` variants (account trie, storage trie, bytecodes, storage roots)
- Replace `try_recv()` + `yield_now()` busy loop in storage healing with `tokio::select!` on recv + 1s timeout

### Benchmark results (Hoodi, ethrex-mainnet-4: 12 cores, 62 GB RAM)

| Phase | main (baseline) | This PR | Delta |
|-------|-----------------|---------|-------|
| 1. Block Headers | 1:30 | 1:20 | -10s (network) |
| 2. Account Ranges | 0:20 | 0:20 | — |
| 3. Account Insertion | 0:40 | 0:40 | — |
| 4. Storage Ranges | 2:50 | 2:20 | -30s (network) |
| 5. Storage Insertion | 4:10 | 4:10 | — |
| 6. State Healing | 0:10 | 0:10 | — |
| 7. Storage Healing | 0:40 | 0:50 | +10s (more accounts) |
| 8. Bytecodes | 3:11 | 1:13 | **-118s** |
| Finalize | ~17s | ~2s | **-15s** |
| **Total** | **837s (13:57)** | **664s (11:04)** | **-173s (21%)** |

The biggest win is the bytecodes phase (62% faster) where WAL-disabled writes dramatically reduce I/O. Phases 1, 2, 4 are network-bound so their deltas are peer variability. Storage healing healed more accounts (19,333 vs 16,555) due to slightly different chain state, explaining the +10s.

## How to Test

Run snap sync on Hoodi (or any testnet) and compare total sync time against main:

```bash
# Build and run
cargo build --release --bin ethrex
./target/release/ethrex --network hoodi --authrpc.jwtsecret <jwt> --p2p.lookup-interval 10
```

Look for `Sync cycle finished successfully time_elapsed_s=...` in the logs.

### DIFF
--- a/crates/networking/p2p/sync/healing/storage.rs
+++ b/crates/networking/p2p/sync/healing/storage.rs
@@ -27,13 +27,10 @@ use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use std::{
     collections::{HashMap, VecDeque},
     sync::atomic::Ordering,
-    time::Instant,
+    time::{Duration, Instant},
 };
-use tokio::{sync::mpsc::error::TryRecvError, task::JoinSet};
-use tokio::{
-    sync::mpsc::{Sender, error::TrySendError},
-    task::yield_now,
-};
+use tokio::task::JoinSet;
+use tokio::sync::mpsc::{Sender, error::TrySendError};
 use tracing::{debug, trace};
 
 /// This struct stores the metadata we need when we request a node
@@ -170,7 +167,6 @@ pub async fn heal_storage_trie(
     let mut logged_no_free_peers_count = 0;
 
     loop {
-        yield_now().await;
         if state.last_update.elapsed() >= SHOW_PROGRESS_INTERVAL_DURATION {
             METRICS
                 .global_storage_tries_leafs_healed
@@ -229,7 +225,7 @@ pub async fn heal_storage_trie(
                     encoded_to_write.push((hashed_account, account_nodes));
                 }
                 // PERF: use put_batch_no_alloc? (it needs to remove parent nodes too)
-                spawned_rt::tasks::block_on(store.write_storage_trie_nodes_batch(encoded_to_write))
+                spawned_rt::tasks::block_on(store.write_storage_trie_nodes_batch_no_wal(encoded_to_write))
                     .expect("db write failed");
             });
         }
@@ -258,14 +254,17 @@ pub async fn heal_storage_trie(
 
         let _ = requests_task_joinset.try_join_next();
 
-        let trie_nodes_result = match task_receiver.try_recv() {
-            Ok(trie_nodes) => trie_nodes,
-            Err(TryRecvError::Empty) => {
-                state.empty_count += 1;
-                continue;
+        let trie_nodes_result = tokio::select! {
+            result = task_receiver.recv() => {
+                match result {
+                    Some(trie_nodes) => trie_nodes,
+                    None => {
+                        state.disconnected_count += 1;
+                        continue;
+                    }
+                }
             }
-            Err(TryRecvError::Disconnected) => {
-                state.disconnected_count += 1;
+            _ = tokio::time::sleep(Duration::from_secs(1)) => {
                 continue;
             }
         };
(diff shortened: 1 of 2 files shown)


crates/networking/p2p/sync/healing/state.rs
```diff
@@ -335,6 +285,76 @@ async fn heal_state_trie(
...
+    peers: &mut PeerHandler,
+    state_root: H256,
+    task_sender: &tokio::sync::mpsc::Sender<(
+        H256,
+        Result<Vec<Node>, SnapError>,
+        Vec<RequestMetadata>,
+    )>,
+    logged_no_free_peers_count: &mut u32,
+) {
+    while (*inflight_tasks as u32) < MAX_IN_FLIGHT_REQUESTS && !paths.is_empty() {
```
ElFantasma, line 303: `(*inflight_tasks as u32) < MAX_IN_FLIGHT_REQUESTS` — the `as u32` truncates silently if `inflight_tasks` ever exceeds `u32::MAX`. Can't happen in practice (max 77), but the cast is unnecessary since `MAX_IN_FLIGHT_REQUESTS` is a constant: either change `inflight_tasks` to `u32` (it's only used as a counter here), or cast the constant up: `*inflight_tasks < MAX_IN_FLIGHT_REQUESTS as u64`.

crates/networking/p2p/sync/healing/state.rs
```diff
@@ -335,6 +285,76 @@ async fn heal_state_trie(
...
+            }
+            *logged_no_free_peers_count -= 1;
+            break;
+        };
+
+        let tx = task_sender.clone();
+        *inflight_tasks += 1;
+        let peer_table = peers.peer_table.clone();
+
+        tokio::spawn(async move {
```
ElFantasma, line 339: Fire-and-forget `tokio::spawn` — if this task panics (e.g., `request_state_trienodes` panics), `inflight_tasks` is never decremented and the outer `tokio::select!` will poll the channel forever (retrying every 1s on timeout). Pre-existing issue, but now amplified since up to 77 tasks can be in flight.

Consider collecting the `JoinHandle` (or using `JoinSet`) and checking for panics when draining responses, similar to how `db_joinset` is handled elsewhere in this file.

crates/networking/p2p/sync/healing/storage.rs
```diff
@@ -258,14 +254,17 @@ pub async fn heal_storage_trie(
...
+                    Some(trie_nodes) => trie_nodes,
+                    None => {
+                        state.disconnected_count += 1;
+                        continue;
+                    }
+                }
             }
-            Err(TryRecvError::Disconnected) => {
-                state.disconnected_count += 1;
+            _ = tokio::time::sleep(Duration::from_secs(1)) => {
```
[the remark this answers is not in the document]
iovoid: Agreed

PR #6181, opened 2026-02-11
perf(l1): adaptive request sizing, storage bisection, and parallel trie in snap sync

ilitteri:
## Motivation

Follow-up to #6178 (Tier 1: WAL disable, multi-peer healing, async select). Profiling Tier 1 results on Hoodi showed remaining bottlenecks in fixed-size network requests, sequential big-account storage downloads, and single-threaded account trie construction. Nethermind uses adaptive request sizing, range bisection, and parallel trie BulkSet to address these.

## Description

Two commits implementing three Tier 2 optimizations:

**1. Adaptive request sizing (5efae33)**
- Add `RequestSizerMap` — per-peer response_bytes tuning based on response latency
- Start at 128KB, scale up 1.5x when latency < 2s, scale down 1.5x when > 10s, clamped to [50KB, 2MB]
- Applied to all 5 snap protocol request types: account ranges, storage ranges, trie nodes (state + storage healing), and bytecodes
- Includes 7 unit tests

**2. Storage range bisection (5efae33)**
- When a big-account sub-range returns incomplete, bisect the remaining range into two sub-tasks instead of one
- Both halves are dispatched via the normal task queue to potentially different peers, doubling download bandwidth
- Minimum range threshold (size >= 2) prevents infinite subdivision
- Composes with existing density-based big-account splitting logic

**3. Parallel 16-way account trie construction (fb91a34)**
- Partition sorted account iterator into 16 buckets by first nibble of H256 key
- Build each subtrie in parallel using `std::thread::scope` (16 scoped threads)
- Merge 16 subtrie root hashes into a single root BranchNode
- Refactor `trie_from_sorted_accounts` into `_inner` + `finalize_root` layers for reuse
- Falls back to sequential when 0-1 nibble buckets are non-empty
- Includes 5 new parallel construction tests

### Benchmark results (Hoodi, ethrex-mainnet-4: 12 cores, 62 GB RAM)

| Phase | main (baseline) | This PR (Tier 1+2) | Delta |
|-------|-----------------|---------------------|-------|
| 1. Block Headers | 1:10 | 1:40 | +30s (network) |
| 2. Account Ranges | 0:10 | 0:20 | +10s (network) |
| 3. Account Insertion | 0:30 | **0:20** | **-10s (33%)** |
| 4. Storage Ranges | 2:40 | **2:00** | **-40s (25%)** |
| 5. Storage Insertion | 4:20 | 4:30 | +10s (noise) |
| 6. State Healing | 0:10 | 0:10 | — |
| 7. Storage Healing | 0:30 | **0:20** | **-10s** |
| 8. Bytecodes + finalize | 3:41 | **2:10** | **-91s (41%)** |
| **Total** | **791s (13:11)** | **691s (11:31)** | **-100s (13%)** |

Phases 1, 2 are network-bound (peer variability at startup). CPU/IO phases show consistent improvements:
- **Account Insertion**: 33% faster from 16-way parallel trie construction
- **Storage Ranges**: 25% faster from adaptive sizing + bisection improving network utilization
- **Bytecodes**: 41% faster from adaptive sizing scaling up response_bytes for fast peers

## How to Test

Run snap sync on Hoodi and compare total sync time against main:

```bash
cargo build --release --bin ethrex
./target/release/ethrex --network hoodi --authrpc.jwtsecret <jwt> --p2p.lookup-interval 10
```

Look for `Sync cycle finished successfully time_elapsed_s=...` in the logs. Run `cargo test -p ethrex-p2p -- request_sizer` and `cargo test -p ethrex-trie -- trie_sorted` for unit tests.

### DIFF
--- a/crates/common/trie/trie_sorted.rs
+++ b/crates/common/trie/trie_sorted.rs
@@ -156,39 +156,28 @@ fn flush_nodes_to_write(
     Ok(())
 }
 
-/// trie_from_sorted_accounts computes and stores into a db a trie from a sorted
-/// iterator of H256 paths and values. This function takes a ThreadPool Arc to send
-/// the writing task to be done concurrently.
-/// To limit the amount of memory this function can use, we use a crossbeam multiproducer
-/// multiconsumer queue, which gives the function a buffer to write nodes into before
-/// flushing to the db.
-pub fn trie_from_sorted_accounts<'scope, T>(
+/// Core trie building loop. Processes sorted (H256, Vec<u8>) pairs into a trie structure,
+/// writing intermediate nodes to DB via the ThreadPool. Returns the root-level BranchNode
+/// (with choices populated) and any unflushed nodes, or None if the iterator was empty.
+#[allow(clippy::type_complexity)]
+fn trie_from_sorted_accounts_inner<'scope, T>(
     db: &'scope dyn TrieDB,
     data_iter: &mut T,
     scope: Arc<ThreadPool<'scope>>,
     buffer_sender: Sender<Vec<(Nibbles, Node)>>,
     buffer_receiver: Receiver<Vec<(Nibbles, Node)>>,
-) -> Result<H256, TrieGenerationError>
+) -> Result<Option<(BranchNode, Vec<(Nibbles, Node)>)>, TrieGenerationError>
 where
     T: Iterator<Item = (H256, Vec<u8>)> + Send,
 {
     let Some(initial_value) = data_iter.next() else {
-        return Ok(*EMPTY_TRIE_HASH);
+        return Ok(None);
     };
     let mut nodes_to_write: Vec<(Nibbles, Node)> = buffer_receiver
         .recv()
         .expect("This channel shouldn't close");
-    // We have a stack of the parents of the current parent
-    let mut trie_stack: Vec<StackElement> = Vec::with_capacity(64); // Optimized for H256
-
-    // This is the current parent of the first element. We assume that the root node
-    // is always a parent, and we fix it afterwards if it's not true
-    // The root is a parent of all nodes
-    let mut nodehash_buffer = Vec::with_capacity(512);
+    let mut trie_stack: Vec<StackElement> = Vec::with_capacity(64);
     let mut current_parent = StackElement::default();
-
-    // The current node that is being used for computing. We compare it with the current
-    // parent and the next value to see where it should be written
     let mut current_node: CenterSide = CenterSide::from_value(initial_value);
     let mut next_value_opt: Option<(H256, Vec<u8>)> = data_iter.next();
 
@@ -198,7 +187,6 @@ where
             scope.execute_priority(Box::new(move || {
                 let _ = flush_nodes_to_write(nodes_to_write, db, buffer_sender);
             }));
-            // We wait to get a new buffer to avoid writing too much
             nodes_to_write = buffer_receiver
                 .recv()
                 .expect("This channel shouldn't close");
@@ -206,9 +194,6 @@ where
 
         let next_value_path = Nibbles::from_bytes(next_value.0.as_bytes());
 
-        // If the current parent isn't a parent of the next value, that means
-        // that the current value doesn't have a sibling to the right
-        // As such we write this node and change the current node to the current parent
         while !is_child(&next_value_path, &current_parent) {
             add_current_to_parent_and_write_queue(
                 &mut nodes_to_write,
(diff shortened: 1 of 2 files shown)


crates/common/trie/trie_sorted.rs
```diff
@@ -323,20 +301,141 @@ where
...
+            .collect();
+
+        // Collect subtrie roots and merge into root BranchNode
+        let mut root_choices = BranchNode::EMPTY_CHOICES;
+        for (nibble, handle) in handles {
+            let branch_opt = handle
+                .join()
+                .map_err(|_| TrieGenerationError::ThreadJoinError())??;
+            if let Some(branch) = branch_opt {
+                root_choices[nibble] = branch.choices[nibble].clone();
```
ElFantasma, line 420: This line relies on the invariant that `trie_from_sorted_accounts_inner`, when fed keys all sharing the same first nibble N, returns a `BranchNode` where only `choices[N]` is populated. That's true today because `_inner` processes full H256 paths (not stripped keys), but it's subtle and fragile — a future caller passing pre-stripped keys would silently produce a wrong trie.

Consider adding a debug assertion:
```rust
if let Some(branch) = branch_opt {
    debug_assert!(
        branch.choices.iter().enumerate()
            .filter(|(_, c)| c.is_valid())
            .all(|(i, _)| i == nibble),
        "subtrie for nibble {nibble} has unexpected populated choices"
    );
    root_choices[nibble] = branch.choices[nibble].clone();
}
```

crates/common/trie/trie_sorted.rs
```diff
@@ -323,20 +301,141 @@ where
...
+    }
+    // Single bucket: fall back to sequential (parallel overhead not worth it)
+    if non_empty_count == 1 {
+        let items: Vec<_> = buckets.into_iter().flatten().collect();
+        return trie_from_sorted_accounts_wrap(db, &mut items.into_iter());
+    }
+
+    // Build 16 subtries in parallel using scoped threads + shared ThreadPool for DB writes
+    scope(|s| {
+        let pool = Arc::new(ThreadPool::new(16, s));
```
ElFantasma, line 376: This creates a `ThreadPool` with 16 threads for DB writes, plus 16 scoped threads (one per nibble) are spawned at line 388. That's up to 32 threads competing for CPU. On the benchmark machine (12 cores) this is fine, but on smaller machines the over-subscription could hurt.

Consider sizing the pool based on available parallelism, e.g.:
```rust
let num_threads = std::thread::available_parallelism()
    .map(|n| n.get())
    .unwrap_or(4);
let pool = Arc::new(ThreadPool::new(num_threads, s));
```

crates/networking/p2p/snap/request_sizer.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+}
+
+/// Thread-safe map of per-peer adaptive request sizers.
+///
+/// Cheaply cloneable (Arc-backed). Pass clones to spawned tasks.
+#[derive(Debug, Clone, Default)]
+pub struct RequestSizerMap {
+    inner: Arc<Mutex<HashMap<H256, PeerSizer>>>,
```
ElFantasma, line 62: Entries for disconnected peers are never removed from this HashMap, so it grows monotonically over the lifetime of the sync. Each entry is small (~16 bytes), but during a long mainnet sync the node may see hundreds of unique peers.

Not critical, but worth noting — could add a simple cap (e.g., evict oldest entries when size exceeds 256) or tie cleanup to peer disconnection events.

crates/networking/p2p/snap/constants.rs
```diff
@@ -14,8 +14,19 @@ use std::time::Duration;
...
 pub const MAX_RESPONSE_BYTES: u64 = 512 * 1024;
 
+/// Initial response bytes budget for adaptive request sizing (128 KB).
+pub const INITIAL_RESPONSE_BYTES: u64 = 128 * 1024;
+
+/// Minimum response bytes for adaptive sizing (50 KB).
+pub const MIN_RESPONSE_BYTES_ADAPTIVE: u64 = 50 * 1024;
+
+/// Maximum response bytes for adaptive sizing (2 MB).
+pub const MAX_RESPONSE_BYTES_ADAPTIVE: u64 = 2 * 1024 * 1024;
```
ElFantasma, line 28: nit: `MAX_RESPONSE_BYTES_ADAPTIVE` (2 MB) exceeds the server-side cap `MAX_RESPONSE_BYTES` (512 KB). Against ethrex peers, requests above 512 KB will never return more data — the peer truncates at its own limit. This means the sizer will plateau at 512 KB effective throughput regardless of how high it scales.

Not a bug (peers just return less), but might be worth documenting the relationship, or capping at `MAX_RESPONSE_BYTES` unless explicitly targeting non-ethrex peers.

PR #6184, opened 2026-02-11
perf(l1): pipeline bytecode downloads and background storage healing in snap sync

ilitteri:
## Motivation

Snap sync phases are strictly sequential: bytecode downloads wait until all healing completes, and storage healing must reach 100% before finalization. This leaves significant time on the table where network I/O and CPU could overlap.

## Description

Two optimizations to the snap sync pipeline:

### 1. Concurrent bytecode downloads
Instead of downloading bytecodes sequentially after all healing completes (Phase 8), stream code hashes via an `mpsc` channel to a concurrent download task that runs alongside phases 3-7 (healing and storage). Bytecodes are content-addressed (hash = key), so this is safe regardless of pivot changes.

- `CodeHashCollector` gains an optional channel sender (`take_receiver()`)
- `download_bytecodes_concurrent()` task spawned before `insert_accounts`, consumes hashes as they arrive
- Channel closes when `code_hash_collector.finish()` drops the sender, signaling the task to drain and complete

### 2. Background storage healing with 99% threshold
Split healing into state (must complete fully) and storage (can proceed at 99% threshold):

- State healing runs first and must complete 100%
- Initial storage healing pass runs with a 5-minute time limit
- Continues without time limit until 99% of account roots are healed
- After finalization (FKV + block store + forkchoice_update), remaining <1% heals in a background task

`heal_storage_trie()` gains a `time_limit: Option<Duration>` parameter and `global_roots_healed` counter for progress tracking.

## Benchmark (Hoodi, ethrex-mainnet-4: 12 cores, 62 GB RAM)

| Run | Time | vs Baseline |
|-----|------|-------------|
| Baseline (Tier 2) | 564s | — |
| **Optimized (Tier 3)** | **489s** | **-13%** |

Bytecodes started downloading at ~6:49 (concurrent with healing) vs ~9:00+ in baseline (after healing).

### Cumulative improvement across tiers

| Tier | Baseline | Optimized | Improvement |
|------|----------|-----------|-------------|
| Tier 1 (PR #6178) | 837s | 664s | -21% |
| Tier 2 (PR #6181) | 791s | 691s | -13% |
| **Tier 3 (this PR)** | 564s | 489s | **-13%** |

## How to Test

1. Run snap sync on Hoodi: `cargo run --release --bin ethrex -- --network hoodi --authrpc.jwtsecret <jwt>`
2. Observe "Concurrent bytecode download complete" log appearing before "Finished healing"
3. Observe "Storage healing target: N/M accounts (99% threshold)" in logs
4. If <1% remains, observe "Spawning background storage healing" after finalization

### DIFF
--- a/crates/networking/p2p/sync/code_collector.rs
+++ b/crates/networking/p2p/sync/code_collector.rs
@@ -6,6 +6,7 @@ use ethrex_common::H256;
 use ethrex_rlp::encode::RLPEncode;
 use std::collections::HashSet;
 use std::path::PathBuf;
+use tokio::sync::mpsc;
 use tokio::task::JoinSet;
 use tracing::error;
 
@@ -19,6 +20,8 @@ pub struct CodeHashCollector {
     file_index: u64,
     // JoinSet to manage async disk writes
     disk_tasks: JoinSet<Result<(), DumpError>>,
+    // Optional channel to stream code hashes for concurrent bytecode downloading
+    code_hash_sender: Option<mpsc::UnboundedSender<Vec<H256>>>,
 }
 
 impl CodeHashCollector {
@@ -29,19 +32,38 @@ impl CodeHashCollector {
             snapshots_dir,
             file_index: 0,
             disk_tasks: JoinSet::new(),
+            code_hash_sender: None,
         }
     }
 
-    /// Adds a code hash to the buffer
+    /// Creates the channel and returns the receiver for concurrent bytecode downloading.
+    /// Must be called before any code hashes are added.
+    pub fn take_receiver(&mut self) -> mpsc::UnboundedReceiver<Vec<H256>> {
+        let (tx, rx) = mpsc::unbounded_channel();
+        self.code_hash_sender = Some(tx);
+        rx
+    }
+
+    /// Adds a code hash to the buffer and sends it through the channel if present
     pub fn add(&mut self, hash: H256) {
-        self.buffer.insert(hash);
+        if self.buffer.insert(hash) && let Some(sender) = &self.code_hash_sender {
+            // Ignore send errors — the receiver may have been dropped if
+            // the bytecode task finished early or errored out
+            let _ = sender.send(vec![hash]);
+        }
     }
 
     // The optimization for rocksdb database doesn't use this method
     #[cfg(not(feature = "rocksdb"))]
     /// Extends the buffer with a list of code hashes
     pub fn extend(&mut self, hashes: impl IntoIterator<Item = H256>) {
-        self.buffer.extend(hashes);
+        let new_hashes: Vec<H256> = hashes
+            .into_iter()
+            .filter(|h| self.buffer.insert(*h))
+            .collect();
+        if !new_hashes.is_empty() && let Some(sender) = &self.code_hash_sender {
+            let _ = sender.send(new_hashes);
+        }
     }
 
     /// Flushes the buffer to a file if the buffer is larger than [`CODE_HASH_WRITE_BUFFER_SIZE`]
@@ -55,13 +77,16 @@ impl CodeHashCollector {
         Ok(())
     }
 
-    /// Finishes the code collector and returns the final index of file
+    /// Finishes the code collector: flushes remaining hashes and closes the channel
     pub async fn finish(mut self) -> Result<(), SyncError> {
         // Final flush if needed
         if !self.buffer.is_empty() {
             let buffer = std::mem::take(&mut self.buffer);
             self.flush_buffer(buffer);
         }
+
+        // Drop the sender to close the channel, signaling the bytecode task to drain
+        self.code_hash_sender.take();
 
         // Wait for all pending writes
         self.disk_tasks
--- a/crates/networking/p2p/sync/healing/storage.rs
+++ b/crates/networking/p2p/sync/healing/storage.rs
@@ -116,6 +116,10 @@ pub struct NodeRequest {
 /// - When a node is downloaded:
 ///    - if it has no missing children, we store it in the db
 ///    - if the node has missing children, we store it in our healing_queue, which is preserved between calls
+///
+/// When `time_limit` is `Some(duration)`, healing will stop early after the duration elapses
+/// and return `Ok(false)`, allowing the caller to proceed with partial healing.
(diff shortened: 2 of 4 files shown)


crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -586,6 +615,112 @@ pub async fn snap_sync(
...
+        let bg_state_root = pivot_header.state_root;
+        // Only clone the healed_accounts set — accounts_with_storage_root is not needed
+        // by heal_storage_trie (get_initial_downloads only reads healed_accounts)
+        let bg_storage_accounts = AccountStorageRoots {
+            accounts_with_storage_root: std::collections::BTreeMap::new(),
+            healed_accounts: storage_accounts.healed_accounts.clone(),
+        };
+        let mut bg_peers = peers.clone();
+        let bg_store = store.clone();
+        tokio::spawn(async move {
```
ElFantasma, line 634: **Background healing concurrent with block execution**: after `forkchoice_update` completes (line 615), the node starts executing blocks, which writes new storage trie nodes to the same `Store`. This `tokio::spawn` runs `heal_storage_trie` which also writes storage trie nodes via `write_storage_trie_nodes_batch_no_wal` — potentially for the same accounts that block execution is updating.

Additionally, `staleness_timestamp: u64::MAX` means the background task will never self-abort on staleness. If the pivot state root becomes unavailable (pruned or overwritten by execution), `get_initial_downloads` will open a trie at a stale root.

Consider either:
- Adding a `CancellationToken` that's triggered when block execution starts, so the background task can abort cleanly
- Or at minimum, catching and logging errors from `open_locked_state_trie` in `get_initial_downloads` instead of `.expect()`

crates/networking/p2p/sync/code_collector.rs
```diff
@@ -29,19 +32,38 @@ impl CodeHashCollector {
...
             disk_tasks: JoinSet::new(),
+            code_hash_sender: None,
         }
     }
 
-    /// Adds a code hash to the buffer
+    /// Creates the channel and returns the receiver for concurrent bytecode downloading.
+    /// Must be called before any code hashes are added.
+    pub fn take_receiver(&mut self) -> mpsc::UnboundedReceiver<Vec<H256>> {
+        let (tx, rx) = mpsc::unbounded_channel();
```
ElFantasma, line 42: This creates an `mpsc::unbounded_channel`. If account insertion (which calls `add()`) runs significantly faster than the bytecode download task (which is network-bound), the channel can accumulate all code hashes in memory before any are consumed. With millions of accounts, that could be non-trivial memory.

Consider using a bounded channel with a reasonable capacity (e.g., 1000 batches) to provide natural backpressure. The `add()` method would need to become `async` or use `try_send` with a fallback — but since it already ignores send errors, a bounded channel with `try_send` would just drop overflow hashes, which the file-based backup already covers.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -463,101 +481,112 @@ pub async fn snap_sync(
...
+            pivot_header = update_pivot(
+                pivot_header.number,
+                pivot_header.timestamp,
+                peers,
+                block_sync_state,
+            )
+            .await?;
+        }
+        // Re-do state healing if pivot changed
+        let state_ok = heal_state_trie_wrap(
```
ElFantasma, line 529: After the initial timed pass, state healing is re-run on every loop iteration (`heal_state_trie_wrap`). If the pivot hasn't changed, this re-heals an already-healed state trie on each iteration — it should return quickly (no work to do), but it's still opening the trie and scanning. Consider skipping the state re-heal when the pivot hasn't changed since the last successful state heal.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -463,101 +481,112 @@ pub async fn snap_sync(
...
+        "Storage healing sufficient: {global_roots_healed}/{total_accounts_to_heal} roots healed (fully done: {storage_fully_done})"
+    );
 
     store.generate_flatkeyvalue()?;
 
     debug_assert!(validate_state_root(store.clone(), pivot_header.state_root).await);
-    debug_assert!(validate_storage_root(store.clone(), pivot_header.state_root).await);
+    // Only validate storage root if healing fully completed — background healing
+    // handles the remaining <1% of accounts after snap sync finalization
+    if storage_fully_done {
```
ElFantasma, line 568: nit: When `storage_fully_done` is true, `validate_storage_root` is skipped. But this means the <1% background case never validates storage root even after background healing completes (the spawned task at line 625 doesn't run validation). Consider logging a note or adding a `debug_assert!` in the background task's `Ok(_)` branch.

crates/networking/p2p/sync/snap_sync.rs
ElFantasma, line 634: The `JoinHandle` is dropped immediately, making this fully fire-and-forget. The `warn!` on error is good, but there's a harder problem: if the process exits (crash or SIGTERM) after `snap_sync` returns but before background healing finishes, those <1% of storage tries stay incomplete. On restart, since `clear_snap_state` is called right after `snap_sync` returns, snap sync won't re-run — so those accounts' storage remains broken. Consider storing the handle and joining it during graceful shutdown, or persisting a "background healing pending" flag in the store so it re-runs on restart.

crates/networking/p2p/sync/healing/storage.rs
```diff
@@ -230,13 +244,24 @@ pub async fn heal_storage_trie(
...
         if is_done {
+            *global_roots_healed += state.roots_healed as u64;
             db_joinset.join_all().await;
             return Ok(true);
         }
 
-        if is_stale {
+        if is_stale || time_limit_reached {
+            *global_roots_healed += state.roots_healed as u64;
             db_joinset.join_all().await;
```
ElFantasma, line 254: nit: the `if is_stale` branch that preserves `state.healing_queue` (vs clearing it) has no effect — `state` is dropped when the function returns `Ok(false)`, and callers always pass `HashMap::new()` on the next call anyway. This was dead code in the original too (the function takes `healing_queue` by value but the return type is `bool`, not `(bool, StorageHealingQueue)`). Worth either removing the conditional or adding a `// TODO: thread healing_queue through to avoid re-queuing` comment so the intent is clear.

PR #94, opened 2024-06-28
feat: implement `Ping` message decoding

ricomateo:
This PR depends on #92 

**Motivation**

**Description**

Enables `Ping` message decoding

Closes #79

Issue #87, opened 2024-06-28
Store peer data


MegaRedHand:
[More info in the spec](https://github.com/ethereum/devp2p/blob/master/discv4.md#kademlia-table)

PR #64, opened 2024-06-26
feat: store account storage in db

fmoletta:
Based on #63 
**Motivation**

 Enable storing account storage in DB
 
**Description**

* Create `AccountStorages` table mapping account addresses & storage keys to storage values

Closes #41

PR #63, opened 2024-06-26
feat: store account info in db

fmoletta:
**Motivation**

**Description**

Closes #issue_number

PR #6190, opened 2026-02-12
perf(trie): stack-allocated Nibbles for O(1) traversal

ilitteri:
## Motivation

CPU profiles show that `Nibbles::next()` and `Nibbles::skip_prefix()` are called on every trie node during traversal. The current implementation uses two `Vec<u8>` fields, making `next()` O(n) via `Vec::remove(0)` and `skip_prefix()` O(n) via `.to_vec()` allocation. This creates unnecessary heap allocation pressure during the hottest path in trie operations.

## Description

Replace the two heap-allocated `Vec<u8>` fields (`data` + `already_consumed`) with a single stack-allocated `[u8; 132]` buffer plus `start: u8` and `len: u8` cursors.

Buffer layout: `buf[0..start]` = consumed nibbles, `buf[start..start+len]` = remaining data.

**Complexity improvements:**
| Operation | Before | After |
|-----------|--------|-------|
| `next()` | O(n) — `Vec::remove(0)` | O(1) — `start += 1` |
| `skip_prefix()` | O(n) — `.to_vec()` | O(1) — `start += prefix_len` |
| `offset()` | O(n) — slice + concat | O(1) — adjust cursors |
| Construction | Heap alloc | Stack only |

The 132-byte buffer is sufficient for all Ethereum trie paths (max observed: 131 nibbles from `apply_prefix`). The struct is 134 bytes total on the stack vs 48+ bytes minimum plus heap allocations before.

## Benchmark Results

**Throughput (Mgas/s):**

| Scenario | Baseline | Rank 8 | Change |
|----------|----------|--------|--------|
| Fast (200 blocks) | 420.48 | 430.74 | **+2.4%** |
| Gigablocks (100 synthetic) | 909.63 | 933.92 | **+2.7%** |
| Slow (5000 blocks) | 811.04 | 856.98 | **+5.7%** |

**Latency:**

| Scenario | Metric | Baseline | Rank 8 |
|----------|--------|----------|--------|
| Fast | avg / med / p95 / p99 | 74.23 / 48.51 / 180.72 / 258.13 | 66.81 / 55.73 / 158.33 / 225.54 |
| Gigablocks | avg / med / p95 / p99 | 1340 / 1070 / 2840 / 3380 | 1320 / 1120 / 2870 / 3240 |
| Slow | avg / med / p95 / p99 | 28.84 / 21.56 / 73.51 / 131.37 | 27.06 / 20.44 / 69.28 / 122.4 |

**Profile comparison (gigablocks):**
- Nibbles functions (`next`, `skip_prefix`) no longer register as separate samples — the O(1) cursor operations are below the sampling threshold
- `get_node_mut`: 10.19% cum → 9.89% cum (improved, within noise)

Benchmark runs: #85 (fast), #94 (gigablocks), #87 (slow). Baseline: #63, #64, #65 on `main` @ `1a47fb18`.

## How to Test

All existing trie tests pass (55 unit tests + 12 nibbles tests + 3 trie DB tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 2026-02-11
 
-- Replace heap-allocated Nibbles with stack-allocated buffer for O(1) trie traversal [#PENDING](https://github.com/lambdaclass/ethrex/pull/PENDING)
+- Replace heap-allocated Nibbles with stack-allocated buffer for O(1) trie traversal [#6190](https://github.com/lambdaclass/ethrex/pull/6190)
 
 ### 2026-02-06
 


crates/common/trie/nibbles.rs
```diff
@@ -7,31 +7,39 @@ use ethrex_rlp::{
...
-    rkyv::Deserialize,
-    rkyv::Serialize,
-    rkyv::Archive,
-)]
+/// Stack-allocated nibble sequence for trie traversal.
+///
+/// Layout: `buf[0..start]` is the already-consumed prefix (for path tracking),
+/// and `buf[start..start+len]` is the remaining data. This makes `next()` and
+/// `skip_prefix()` O(1) by advancing `start` instead of shifting a Vec.
+#[derive(Debug, Clone, rkyv::Deserialize, rkyv::Serialize, rkyv::Archive)]
```
ElFantasma, line 19: The `rkyv::Serialize/Deserialize/Archive` derive now archives `buf: [u8; 132], start: u8, len: u8` instead of `data: Vec<u8>, already_consumed: Vec<u8>`. This changes the rkyv wire format.

This matters because `LeafNode` and `ExtensionNode` contain `Nibbles` and also derive rkyv — and these are embedded in `ExecutionWitness` which is rkyv-serialized for L2 prover input (`rkyv::to_bytes` in `prover/src/backend/*.rs`, `l2/storage/src/store_db/sql.rs`).

The custom serde impls below correctly preserve the old format. Consider doing the same for rkyv (custom `Archive`/`Serialize`/`Deserialize` impls that archive as two variable-length slices), or document that this is an intentional format break.

crates/common/trie/nibbles.rs
```diff
@@ -145,45 +178,65 @@ impl Nibbles {
...
     pub fn at(&self, i: usize) -> usize {
-        self.data[i] as usize
+        self.buf[self.start as usize + i] as usize
     }
 
     /// Inserts a nibble at the start
     pub fn prepend(&mut self, nibble: u8) {
-        self.data.insert(0, nibble);
+        let start = self.start as usize;
+        let end = start + self.len as usize;
```
ElFantasma, line 218: nit: `prepend` is O(n) due to `copy_within` — same as the old `Vec::insert(0, _)`. Not a regression, but the PR description claims all operations are O(1).

Also, when `self.start > 0` there's unused space before the data. You could make `prepend` truly O(1) by decrementing `start`:
```rust
pub fn prepend(&mut self, nibble: u8) {
    if self.start > 0 {
        self.start -= 1;
        self.buf[self.start as usize] = nibble;
    } else {
        // fallback: shift right
        let end = self.len as usize;
        self.buf.copy_within(0..end, 1);
        self.buf[0] = nibble;
    }
    self.len += 1;
}
```

crates/common/trie/nibbles.rs
```diff
@@ -7,31 +7,39 @@ use ethrex_rlp::{
...
+/// Layout: `buf[0..start]` is the already-consumed prefix (for path tracking),
+/// and `buf[start..start+len]` is the remaining data. This makes `next()` and
+/// `skip_prefix()` O(1) by advancing `start` instead of shifting a Vec.
+#[derive(Debug, Clone, rkyv::Deserialize, rkyv::Serialize, rkyv::Archive)]
 pub struct Nibbles {
-    data: Vec<u8>,
-    /// Parts of the path that have already been consumed (used for tracking
-    /// current position when visiting nodes). See `current()`.
-    already_consumed: Vec<u8>,
+    buf: [u8; MAX_NIBBLES],
```
ElFantasma, line 21: Each `Nibbles` is now 134 bytes on the stack (was ~48 bytes + heap). Worth keeping in mind for the sorted trie writer where `Vec<(Nibbles, Node)>` buffers hold up to 20,000 entries across 32 channel slots (`SIZE_TO_WRITE_DB` × `BUFFER_COUNT` in `trie_sorted.rs`). The extra ~86 bytes/entry adds ~55 MB of memory across all buffers.

The benchmarks show this is a net win, so probably fine — just noting the trade-off.

ElFantasma (changes requested):

PR #6191, opened 2026-02-12
perf(l1): speed up snap sync validation with parallelism and deduplication

ilitteri:
## Motivation

The three post-sync validation functions (`validate_state_root`, `validate_storage_root`, `validate_bytecodes`) run via `debug_assert!` in `release-with-debug-assertions` builds to verify trie integrity after snap sync. We use these to validate that optimization changes don't break sync correctness, but they take too long — slowing down the testing feedback loop.

Neither geth nor nethermind perform full trie traversal after snap sync (they trust healing completion), so these are ethrex-specific correctness checks we want to keep but make faster.

### `validate_state_root` — parallel subtree validation
- Add `Trie::validate_parallel()` that fetches the root branch node and validates each of its 16 subtrees independently via rayon
- Extracted `validate_subtree()` helper for iterative DFS with hash checking
- Expected ~4-8x speedup from 16-way parallelism

### `validate_storage_root` — skip empty + eliminate par_bridge overhead
- Filter accounts with `storage_root == EMPTY_TRIE_HASH` before processing (~90% of accounts are EOAs with empty storage)
- Replace `par_bridge()` with chunked `collect()` + `par_iter()` (4096 accounts per chunk) to eliminate the per-item mutex contention inherent to `par_bridge`'s sequential-to-parallel adapter

### `validate_bytecodes` — dedup + existence check + parallelize
- Collect unique code hashes into a `HashSet` first — on mainnet ~70M contracts share only ~2.5M unique bytecodes (28:1 ratio)
- Add `Store::code_exists()` that checks key existence without reading the full bytecode blob or RLP decoding (avoids BlobDB value reads)
- Parallelize with `par_iter()` over the deduplicated set

## How to Test

1. Build with debug assertions: `cargo build --release --bin ethrex --profile release-with-debug-assertions`
2. Run a snap sync on any network (hoodi, sepolia, mainnet)
3. After sync completes, observe validation timings in logs:
   - `Starting validate_state_root` / `Succesfully validated tree`
   - `Starting validate_storage_root` / `Finished validate_storage_root`
   - `Starting validate_bytecodes` / `Collected N unique code hashes for validation`
4. Compare timings against main branch


crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -710,21 +710,40 @@ pub async fn validate_state_root(store: Store, state_root: H256) -> bool {
...
+                .par_iter()
+                .try_for_each(|(hashed_address, account_state)| {
+                    store
+                        .open_locked_storage_trie(
+                            *hashed_address,
+                            state_root,
+                            account_state.storage_root,
+                        )
+                        .expect("couldn't open storage trie")
+                        .validate()
```
iovoid, line 744: Why not validate_parallel?

crates/common/trie/trie.rs
```diff
@@ -562,6 +562,88 @@ impl Trie {
...
+        use rayon::prelude::*;
+
+        if !self.root.is_valid() {
+            return Ok(());
+        }
+
+        let db = &*self.db;
+        let root_node = self
+            .root
+            .get_node_checked(db, Nibbles::default())?
```
iovoid, line 578: Note: get_node_checked won't detect a hash mismatch if the root node was already loaded.

This shouldn't matter right now given we usually call this on a freshly opened trie, but it might be a good idea to verify.

crates/common/trie/trie.rs
```diff
@@ -562,6 +562,88 @@ impl Trie {
...
+}
+
+/// Validate a subtree rooted at `start_ref`, checking that all referenced nodes exist
+/// and their hashes match.
+fn validate_subtree(
+    db: &dyn TrieDB,
+    start_path: Nibbles,
+    start_ref: NodeRef,
+) -> Result<(), TrieError> {
+    let mut expected_count: isize = 1;
```
iovoid, line 613: It shouldn't be necessary to keep this. `validate()` only had this because the iterator has no way of distinguishing "ended because of error" and "ended because all nodes were visited".

crates/storage/store.rs
```diff
@@ -710,6 +710,28 @@ impl Store {
         Ok(Some(code))
     }
 
+    /// Check if account code exists by its hash, without reading the full bytecode.
+    /// More efficient than get_account_code for existence checks since it skips
+    /// blob value reads and RLP decoding.
```
ElFantasma: nit: The docstring says this "skips blob value reads" but the underlying `StorageReadView::get()` still calls `rocksdb::get_cf()` which reads the full value into memory. The real savings vs `get_account_code` are skipping RLP decoding and `Code` struct construction (no `jump_targets` deserialization). Consider updating the comment to reflect that — or switching to `key_may_exist_cf` / a custom existence-check on the backend trait if you actually want to skip the value read.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -737,27 +756,37 @@ pub async fn validate_storage_root(store: Store, state_root: H256) -> bool {
...
+    info!(
+        "Collected {} unique code hashes for validation",
+        unique_hashes.len()
+    );
+
+    // Validate in parallel using existence-only check
+    use rayon::prelude::*;
+    let missing: Vec<_> = unique_hashes
+        .par_iter()
+        .filter(|code_hash| !store.code_exists(**code_hash).unwrap_or(false))
```
ElFantasma: `unwrap_or(false)` silently converts DB errors (lock poisoning, RocksDB I/O failure) into "code missing", which triggers `process::exit(1)` for the wrong reason. The original code had the same pattern (`is_ok_and(|code| code.is_some())`), but since this is new code it's a good time to propagate the error:
```rust
.filter(|code_hash| match store.code_exists(**code_hash) {
    Ok(exists) => !exists,
    Err(e) => { error!("DB error checking code hash {:x}: {e}", code_hash); true }
})
```
At least this way the logs distinguish "missing code" from "DB read failed".

PR #6193, opened 2026-02-12
perf(l1): split account range downloads into concurrent partitions

ilitteri:
## Motivation

Currently `request_account_range()` downloads accounts sequentially — one chunk at a time, one peer at a time. Nethermind splits the account address space into N partitions for concurrent download from different peers.

## Description

Split `[0, 2^256)` into 8 concurrent partitions, each downloading independently:

- `compute_partitions()` splits the address space into 8 non-overlapping ranges
- Each partition runs as a separate tokio task with its own peer via `JoinSet`
- Shared state (`pivot_header`, `block_sync_state`) wrapped in `Arc<RwLock/Mutex>` with double-checked locking for pivot updates
- Snapshot file naming uses `Arc<AtomicU64>` counter to avoid filename collisions between partitions
- Download count metrics use `fetch_add` with delta tracking so partitions don't clobber each other

SST files from different partitions have non-overlapping key ranges, so `ingest_external_file` handles them correctly without additional merge logic.

## How to Test

- Run snap sync on hoodi and verify account download uses multiple peers concurrently
- Check that all partitions complete and account trie is correctly built
- Compare total account download time vs single-stream baseline

### DIFF
--- a/crates/networking/p2p/snap/client.rs
+++ b/crates/networking/p2p/snap/client.rs
@@ -35,7 +35,10 @@ use ethrex_trie::{Node, verify_range};
 use std::{
     collections::{BTreeMap, HashMap, VecDeque},
     path::Path,
-    sync::atomic::Ordering,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
     time::{Duration, SystemTime},
 };
 use tracing::{debug, error, info, trace, warn};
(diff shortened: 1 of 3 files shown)


crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -249,6 +249,11 @@ pub async fn snap_sync(
     block_sync_state: &mut SnapBlockSyncState,
     datadir: &Path,
 ) -> Result<(), SyncError> {
+    // Enable sync mode: relax compaction triggers for higher write throughput
+    if let Err(e) = store.set_sync_mode() {
```
ElFantasma, line 253: `store.set_sync_mode()` does not exist on `Store` — this will fail to compile. This looks like it belongs to PR #6192 (TuneDbMode) which hasn't been merged yet. Either remove these calls or rebase onto that branch.

crates/networking/p2p/snap/client.rs
```diff
@@ -238,25 +331,32 @@ pub async fn request_account_range(
...
+        // Check pivot staleness with read lock first, then upgrade to write if needed
+        let state_root = {
+            let is_stale = block_is_stale(&*pivot_header.read().await);
+            if is_stale {
+                let mut ph = pivot_header.write().await;
+                // Double-check under write lock (another partition may have updated)
+                if block_is_stale(&ph) {
+                    info!("Partition {partition_id}: pivot is stale, updating");
+                    let mut bss = block_sync_state.lock().await;
+                    *ph = update_pivot(ph.number, ph.timestamp, &mut peers, &mut bss)
```
ElFantasma, line 351: Holding both `pivot_header.write()` (acquired at line 345) and `block_sync_state.lock()` (line 350) across the `update_pivot` await is problematic. `update_pivot` does peer selection, network requests, and retries with sleeps — this can block for 15+ seconds. During that time, **all other partitions** are blocked from even reading `pivot_header` (since the write guard excludes readers) or accessing `block_sync_state`.

Consider releasing the locks before the network call:
```rust
let (number, timestamp) = {
    let ph = pivot_header.read().await;
    if !block_is_stale(&ph) {
        return ph.state_root; // early return from block
    }
    (ph.number, ph.timestamp)
};
// Network call without holding any lock
let new_pivot = update_pivot(number, timestamp, &mut peers, &mut bss).await?;
let mut ph = pivot_header.write().await;
// Double-check: another partition may have updated while we were awaiting
if block_is_stale(&ph) {
    *ph = new_pivot;
}
```
Note: this also requires restructuring the `block_sync_state` access — `update_pivot` currently takes `&mut SnapBlockSyncState`, so you'd need to lock it only for the `process_incoming_headers` call inside, or change `update_pivot` to return the headers and let the caller commit them under the lock.

crates/networking/p2p/snap/client.rs
```diff
@@ -80,27 +83,115 @@ struct StorageTask {
...
+    let num_partitions = MAX_ACCOUNT_PARTITIONS;
+    let partitions = compute_partitions(num_partitions);
+
     METRICS
         .current_step
         .set(CurrentStepValue::RequestingAccountRanges);
+    *METRICS.account_tries_download_start_time.lock().await = Some(SystemTime::now());
+
+    let pivot = Arc::new(tokio::sync::RwLock::new(pivot_header.clone()));
+    let sync_state = Arc::new(tokio::sync::Mutex::new(block_sync_state.clone()));
```
ElFantasma, line 129: `SnapBlockSyncState` contains `block_hashes: Vec<H256>` which deep-clones potentially hundreds of thousands of hashes (~16 MB for 500K blocks). This is only needed here because `update_pivot` calls `process_incoming_headers` which appends to `block_hashes`.

Since `block_hashes` is not actually read by any partition (only `state_root` from the pivot header is used), consider wrapping `SnapBlockSyncState` directly in `Arc<Mutex<>>` *without* cloning — pass a reference from the caller. This avoids the upfront deep clone and ensures the state is always consistent.

crates/networking/p2p/snap/client.rs
```diff
@@ -238,25 +331,32 @@ pub async fn request_account_range(
...
+            let is_stale = block_is_stale(&*pivot_header.read().await);
+            if is_stale {
+                let mut ph = pivot_header.write().await;
+                // Double-check under write lock (another partition may have updated)
+                if block_is_stale(&ph) {
+                    info!("Partition {partition_id}: pivot is stale, updating");
+                    let mut bss = block_sync_state.lock().await;
+                    *ph = update_pivot(ph.number, ph.timestamp, &mut peers, &mut bss)
+                        .await
+                        .expect("Should be able to update pivot");
```
ElFantasma, line 353: `.expect()` inside a spawned task will panic and abort only that partition. The `JoinSet` error handler (line 164) catches this, but the error message loses the original `SyncError` context. Prefer `?` propagation:
```rust
*ph = update_pivot(ph.number, ph.timestamp, &mut peers, &mut bss).await
    .map_err(|e| SnapError::InternalError(format!("Partition {partition_id}: pivot update failed: {e}")))?;
```

PR #6194, opened 2026-02-12
perf(l1): overlap header downloads with state sync

ilitteri:
## Motivation

ethrex downloads ALL 24.4M headers sequentially (15-20 min on mainnet) before starting state sync. Nethermind runs header download concurrently with state sync, removing headers from the critical path entirely.

## Description

Restructure `sync_cycle_snap()` to overlap header downloads with state sync:

1. **Fetch pivot by hash**: `request_pivot_header()` fetches the sync_head header directly via `get_block_header_by_hash()` — no need to download the full chain first
2. **Background download**: `download_headers_background()` runs as a tokio task, downloading remaining headers concurrently with state sync
3. **Shared state**: `block_sync_state` wrapped in `Arc<tokio::sync::Mutex<>>` shared between background download and state sync (for `update_pivot` calls)
4. **Await before finalization**: `snap_sync()` awaits the header download handle before `store_block_bodies` and `forkchoice_update`, which need the full header chain

On mainnet, state sync takes 30+ minutes — headers (15-20 min) complete well before state sync finishes, making headers essentially "free".

### Files changed
- `snap_sync.rs`: Restructured `sync_cycle_snap`, new `request_pivot_header`, new `download_headers_background`, updated `snap_sync` and `update_pivot` signatures
- `peer_handler.rs`: Added `get_block_header_by_hash()` method
- `snap/client.rs`: Updated `request_account_range` to accept `Arc<Mutex<SnapBlockSyncState>>`

## How to Test

- Run snap sync on hoodi and verify headers download in background (logs show "Background header download started" then state sync progress immediately)
- Verify "Background header download complete" appears before finalization
- Compare total sync time vs sequential baseline

### DIFF
--- a/crates/networking/p2p/sync/snap_sync.rs
+++ b/crates/networking/p2p/sync/snap_sync.rs
@@ -224,6 +224,35 @@ async fn download_headers_background(
     peers: &mut PeerHandler,
     store: &Store,
     block_sync_state: &Arc<tokio::sync::Mutex<SnapBlockSyncState>>,
+    current_head_number: u64,
+    sync_head: H256,
+    pending_block_header: Option<BlockHeader>,
+) -> Result<(), SyncError> {
+    METRICS
+        .headers_downloading
+        .store(true, Ordering::Relaxed);
+    *METRICS.headers_download_start_time.lock().await = Some(SystemTime::now());
+
+    let result = download_headers_background_inner(
+        peers,
+        store,
+        block_sync_state,
+        current_head_number,
+        sync_head,
+        pending_block_header,
+    )
+    .await;
+
+    METRICS
+        .headers_downloading
+        .store(false, Ordering::Relaxed);
+    result
+}
+
+async fn download_headers_background_inner(
+    peers: &mut PeerHandler,
+    store: &Store,
+    block_sync_state: &Arc<tokio::sync::Mutex<SnapBlockSyncState>>,
     mut current_head_number: u64,
     sync_head: H256,
     pending_block_header: Option<BlockHeader>,
--- a/crates/networking/p2p/metrics.rs
+++ b/crates/networking/p2p/metrics.rs
@@ -3,7 +3,7 @@ use std::{
     collections::{BTreeMap, VecDeque},
     sync::{
         Arc, LazyLock,
-        atomic::{AtomicU8, AtomicU64, Ordering},
+        atomic::{AtomicBool, AtomicU8, AtomicU64, Ordering},
     },
     time::{Duration, SystemTime},
 };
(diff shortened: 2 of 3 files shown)


crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -106,52 +109,185 @@ pub async fn sync_cycle_snap(
...
+        )
+        .await;
+    }
+
+    // Store the pivot header and add it to block_sync_state
+    store
+        .add_block_headers(vec![pivot_header.clone()])
+        .await?;
+    let mut block_sync_state = block_sync_state;
+    block_sync_state.block_hashes.push(pivot_header.hash());
```
ElFantasma, line 151: The pivot hash is pushed here before spawning the background download. But `download_headers_background_inner` will also push it when it reaches `sync_head` — the `process_incoming_headers` call at line 333 extends `block_hashes` with all received headers including the sync_head.

This means the pivot hash appears twice in `block_hashes`. The `forkchoice_update` at line 684 computes `pivot_header.number - i` for each entry, so the duplicate shifts all preceding block numbers by one — every header before the pivot gets mapped to the wrong block number.

Fix: either skip pushing here and let the background task handle it, or add a guard in `process_incoming_headers` / the background loop to skip the sync_head hash if it's already present.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -106,52 +109,185 @@ pub async fn sync_cycle_snap(
...
+    store: &Store,
+    block_sync_state: &Arc<tokio::sync::Mutex<SnapBlockSyncState>>,
+    mut current_head_number: u64,
+    sync_head: H256,
+    pending_block_header: Option<BlockHeader>,
+) -> Result<(), SyncError> {
+    let mut current_head = {
+        let bss = block_sync_state.lock().await;
+        bss.store
+            .get_latest_canonical_block_hash()
```
ElFantasma, line 263: `current_head` (hash) is re-derived here from `get_latest_canonical_block_hash()` + checkpoint, but `current_head_number` was computed by the caller from a *different* `current_head` value (the one from `block_sync_state.get_current_head()`). Between those two reads, `sync_cycle_snap` stored the pivot header via `add_block_headers`, which could change what `get_latest_canonical_block_hash()` returns.

If the re-derived `current_head` hash doesn't match `current_head_number`, the first `request_block_headers(current_head_number, sync_head)` call starts from the wrong position.

Consider passing both `current_head` and `current_head_number` from the caller instead of re-deriving the hash.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -560,6 +665,12 @@ pub async fn snap_sync(
 
     debug_assert!(validate_bytecodes(store.clone(), pivot_header.state_root));
 
+    // Wait for background header download to complete before store_block_bodies
+    // and forkchoice_update, which need the full header chain.
+    info!("Waiting for background header download to complete...");
+    header_download_handle.await??;
+    info!("Background header download complete");
```
ElFantasma, line 672: nit: This duplicates the log at line 338 inside `download_headers_background_inner`. Since `header_download_handle.await??` only succeeds when the background task completed successfully (which already logged "complete"), this second message is redundant. Consider removing one of them or changing this to a debug-level log like `debug!("Header download handle joined successfully")`.

ElFantasma (changes requested):

PR #6200, opened 2026-02-12
fix(l1): add anti-amplification check to discv5 handle_find_node

### LINKED ISSUE #6199: fix(l1): add anti-amplification check to discv5 handle_find_node
## Context

The discv5 `handle_find_node` handler responds to `contact.node.udp_addr()` (the stored IP from the ENR) without any anti-amplification check. In contrast, discv4's FindNode handler uses `validate_contact` to verify `sender_ip == contact.node.ip` before responding.

This means a malicious peer with an active session could update its ENR to point to a victim's IP and then send FindNode requests, causing us to send large NODES responses to the victim.

Discovered during review of #6172.

## Proposed fix

1. Pass `sender_addr` to `handle_find_node` and respond to it instead of `contact.node.udp_addr()`
2. Add a `sender_addr.ip() == contact.node.ip` check before responding (like discv4's `validate_contact`)

Neither alone is sufficient:
- Only (1): attacker spoofs UDP source to victim, response goes to victim
- Only (2): still responding to stored IP, so the check alone doesn't prevent redirection

Together they require an attacker to both spoof UDP source AND have the stored contact IP match, which is much harder (BCP38/ingress filtering blocks UDP spoofing on many networks).

ElFantasma:
**Motivation**

Discv5's `handle_find_node` responds to `contact.node.udp_addr()` (stored IP from ENR) without any sender address validation. A malicious peer with a session could update its ENR to point to a victim IP and trigger large NODES responses sent to the victim. Discv4 already has this protection via `validate_contact`.

**Description**

Use `validate_contact` instead of `get_contact` in `handle_find_node` to verify the sender's IP matches the stored contact IP before sending NODES responses. This prevents amplification attacks and matches the existing protection already present in discv4.

Closes #6199

**Checklist**

- All existing tests pass (`cargo test -p ethrex-p2p --features experimental-discv5`)

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -42,6 +42,10 @@ inputs:
     description: "Build variant for cache separation (l1 or l2)"
     required: false
     default: "l1"
+  cache_write:
+    description: "Whether to write cache to GHCR registry (set to false for fork PRs)"
+    required: false
+    default: "true"
 
 outputs:
   artifact_path:
@@ -89,7 +93,7 @@ runs:
         password: ${{ inputs.dockerhub_password }}
 
     - name: Login to ghcr.io (for cache)
-      if: inputs.username != '' && inputs.password != ''
+      if: inputs.username != '' && inputs.password != '' && inputs.cache_write == 'true'
       uses: docker/login-action@v3
       with:
         registry: ghcr.io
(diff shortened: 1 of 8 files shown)


ElFantasma:
**Note on honest node rejection scenarios**

`validate_contact` can temporarily reject honest nodes in three cases:

1. **UnknownContact** — peer completed a handshake but isn't in our peer table (handshake doesn't add contacts today). This is a pre-existing limitation tracked in #6056. Once that's resolved, handshake peers with a valid ENR will be added to the table.

2. **InvalidContact** — peer is in our table but hasn't been PING/PONG validated yet. Self-heals once the normal revalidation cycle completes.

3. **IpMismatch** — peer changed IP but we still have the old ENR. Self-heals: their PONG carries a higher `enr_seq`, which triggers a FINDNODE(distance=0) to fetch the updated ENR.

All three cases are consistent with how `handle_ping` already behaves (unknown nodes are silently dropped there too). The tradeoff is correct: brief delay for honest nodes vs. preventing amplification abuse.

PR #6205, opened 2026-02-13
perf(l1): download bytecodes concurrently with storage during snap sync

ilitteri:
## Summary
- Download bytecodes concurrently with storage ranges/insertion instead of sequentially after healing, eliminating the bytecodes phase from the critical path
- Skip already-stored bytecodes, start download earlier (during account trie building), fix metrics conflict during concurrent run, and add progress logging

## Description

During snap sync, bytecodes are content-addressed (hash = key), so they're safe to download regardless of pivot staleness. Previously, bytecodes were downloaded sequentially after all healing completed (~6 min on mainnet). This PR spawns a background task that starts downloading bytecodes during account insertion and runs concurrently with storage download/insertion.

**Changes:**
- `snap_sync.rs`: Spawn background bytecode download task using `oneshot` channel to receive code hashes from `insert_accounts`. Extract `download_bytecodes()` and `read_code_hashes_from_dir()` helpers. Handle healing-discovered bytecodes separately.
- `code_collector.rs`: Add `flush_all()` (non-consuming flush for mid-insertion use) and `dir()` accessor.
- `client.rs`: Add `update_metrics` parameter to `request_bytecodes()` to prevent overwriting `current_step` during concurrent download.

**Benchmark (Ethereum mainnet, ethrex-mainnet-4):**

| Phase | With PR | Baseline | Delta |
|-------|---------|----------|-------|
| Headers | 15:30 | 15:50 | -0:20 |
| Account Insertion | 13:20 | 13:40 | -0:20 |
| Storage (DL + Insert) | 53:00 | 53:00 | 0:00 |
| Healing | 2:20 | 2:20 | 0:00 |
| Bytecodes (sequential) | **0:00** | **6:23** | **-6:23** |
| **Total** | **1:29:47** | **1:32:48** | **-3:01 (-3.3%)** |

Bytecodes (2.1M hashes, 8 min download) finished 46 minutes before storage insertion ended.

## How to Test
1. Run ethrex with snap sync on mainnet or Hoodi with a consensus client
2. Verify "Background bytecode download starting for N unique code hashes" appears during account insertion
3. Verify "Bytecode progress: X/Y downloaded" messages appear during storage phases
4. Verify "Background bytecode download complete" appears after healing
5. Verify no sequential bytecodes phase at the end

### DIFF
--- a/crates/networking/p2p/sync/snap_sync.rs
+++ b/crates/networking/p2p/sync/snap_sync.rs
@@ -307,11 +307,40 @@ pub async fn snap_sync(
         info!("Finish downloading account ranges from peers");
 
         *METRICS.account_tries_insert_start_time.lock().await = Some(SystemTime::now());
+        *METRICS.bytecode_download_start_time.lock().await = Some(SystemTime::now());
         METRICS
             .current_step
             .set(CurrentStepValue::InsertingAccountRanges);
+
+        // Set up early bytecode download: spawn a background task that receives
+        // code hashes via a channel after they're collected during the first pass
+        // of account insertion, then downloads while trie building (second pass) proceeds.
+        // Bytecodes are content-addressed (hash=key), so they're safe to download
+        // concurrently — pivot staleness doesn't affect them.
+        let (code_hashes_tx, code_hashes_rx) = tokio::sync::oneshot::channel::<Vec<H256>>();
+        bytecode_handle = Some({
+            let mut bytecode_peers = peers.clone();
+            let bytecode_store = store.clone();
+            tokio::spawn(async move {
+                // Wait for code hashes to be sent from insert_accounts (between passes)
+                let code_hashes = code_hashes_rx.await.map_err(|_| SyncError::BytecodesNotFound)?;
+                if code_hashes.is_empty() {
+                    return Ok(());
+                }
+                info!(
+                    "Background bytecode download starting for {} unique code hashes",
+                    code_hashes.len()
+                );
+                // update_metrics=false to avoid overwriting current_step during concurrent run
+                download_bytecodes(&mut bytecode_peers, &bytecode_store, code_hashes, false)
+                    .await
+            })
+        });
+
         // We read the account leafs from the files in account_state_snapshots_dir, write it into
-        // the trie to compute the nodes and stores the accounts with storages for later use
+        // the trie to compute the nodes and stores the accounts with storages for later use.
+        // Between the code hash collection pass and the trie building pass, the function
+        // flushes code hashes, reads them, and sends them to the background download task.
 
         // Variable `accounts_with_storage` unused if not in rocksdb
         #[allow(unused_variables)]
@@ -321,6 +350,7 @@ pub async fn snap_sync(
             &account_state_snapshots_dir,
             datadir,
             &mut code_hash_collector,
+            Some(code_hashes_tx),
         )
         .await?;
         info!(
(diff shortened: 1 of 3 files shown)


crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -966,6 +973,14 @@ async fn insert_accounts(
         }
     }
 
+    // Flush code hashes to disk, read them, and send to the background download
+    // task so bytecodes start downloading during the trie building pass (below).
+    if let Some(tx) = code_hashes_tx {
+        code_hash_collector.flush_all().await?;
+        let code_hashes = read_code_hashes_from_dir(code_hash_collector.dir())?;
+        let _ = tx.send(code_hashes);
```
ElFantasma, line 981: `let _ = tx.send(code_hashes)` silently drops a send failure. If the background task panicked or its receiver was dropped, the code hashes are lost — and since the snapshot dir is cleaned up and recreated immediately after `insert_accounts` returns (line 368), they can't be re-read. Bytecodes would silently never be downloaded for this sync attempt.

Suggestion: at minimum, log a warning if the send fails. Better yet, propagate the error:
```rust
tx.send(code_hashes).map_err(|_| SyncError::BytecodesNotFound)?;
```
(Same issue in the non-rocksdb path at line 860.)

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -1138,3 +1153,81 @@ async fn insert_storages(
...
+async fn download_bytecodes(
+    peers: &mut PeerHandler,
+    store: &Store,
+    code_hashes: Vec<H256>,
+    update_metrics: bool,
+) -> Result<(), SyncError> {
+    // Filter out already-stored bytecodes to avoid re-downloading on retry
+    let code_hashes: Vec<H256> = code_hashes
+        .into_iter()
+        .filter(|hash| !store.get_account_code(*hash).is_ok_and(|code| code.is_some()))
```
ElFantasma, line 1196: This filter calls `store.get_account_code(*hash)` individually for each of the ~2.1M code hashes. Each call acquires a `Mutex` lock on `account_code_cache` and potentially does a RocksDB read — that's a lot of synchronous blocking I/O in an async context.

For the common case (fresh sync, no codes stored yet), every single check will miss the cache and hit the DB, only to return `None`. Consider short-circuiting: skip the filter entirely when `is_fresh_sync` or similar flag is set, and only enable it for retry scenarios where some codes may already exist.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -305,11 +307,40 @@ pub async fn snap_sync(
...
+        // of account insertion, then downloads while trie building (second pass) proceeds.
+        // Bytecodes are content-addressed (hash=key), so they're safe to download
+        // concurrently — pivot staleness doesn't affect them.
+        let (code_hashes_tx, code_hashes_rx) = tokio::sync::oneshot::channel::<Vec<H256>>();
+        bytecode_handle = Some({
+            let mut bytecode_peers = peers.clone();
+            let bytecode_store = store.clone();
+            tokio::spawn(async move {
+                // Wait for code hashes to be sent from insert_accounts (between passes)
+                let code_hashes = code_hashes_rx.await.map_err(|_| SyncError::BytecodesNotFound)?;
```
ElFantasma, line 326: If the oneshot sender is dropped (e.g., `insert_accounts` returns early with an error before reaching the `tx.send`), this converts the `RecvError` to `SyncError::BytecodesNotFound`, which is misleading — the bytecodes weren't "not found", the code hash channel was closed. This could confuse debugging since the real error would be the one from `insert_accounts`, but the background task's error might surface first (or be logged) as a confusing "BytecodesNotFound".

nit: Consider a more descriptive error variant, e.g. `SyncError::BytecodeChannelClosed`.

PR #6206, opened 2026-02-13
refactor(l2): replace string errors with custom errors in OnChainProposer contracts

### LINKED ISSUE #4196: Use custom errors instead of `require(..., <string>)` or `revert <string>` in contracts
Currently we use `require()` or `revert <string>` to revert in case of errors. While this can be beneficial for manual debugging, ethrex components could leverage reverting with custom errors by easily parsing them and taking decisions accordingly. Also, `rex` can have this custom errors inlcuded as known errors and try to parse into them to show a better message to the user, decreasing the cons of using them instead of `require()`

### LINKED ISSUE #6098: L2: normalize error codes between OnChainProposer contracts

avilagaston9:
## Motivation

The two OnChainProposer contracts use inconsistent error formats — the non-based contract uses opaque hex codes (`"007"`, `"00c"`) while the based contract uses full strings (`"OnChainProposer: Invalid RISC0 proof failed proof verification"`). This causes a real bug: the Rust `l1_proof_sender` detects invalid proofs by string-matching on `message.contains("Invalid TDX proof")`, which works for the based contract but fails silently for the non-based contract. Invalid proofs are never deleted.

Custom errors are cheaper (no string storage in bytecode), self-documenting (the error name IS the documentation), and can be decoded programmatically from the ABI.

Closes #4196, closes #6098

## Description

Replace all `require(condition, "string")` / `revert("string")` in both OnChainProposer contracts with Solidity custom errors defined in the interfaces. Both contracts now use identical error names.

On the Rust side, add a `data: Option<String>` field to `RpcRequestError::RPCError` to capture the revert data from JSON-RPC error responses (custom errors encode their identity as a 4-byte selector in the `data` field, not in `message`). Update proof sender and verifier to match on these selectors instead of string-matching.

**Solidity changes:**
- Define custom errors in both `IOnChainProposer` interfaces
- Replace ~35 string errors in the non-based contract and ~25 in the based contract
- Eliminate the duplicate `"012"` code that was used for two different errors

**Rust changes:**
- Add `data: Option<String>` to `RpcRequestError::RPCError`
- Propagate `error_response.error.data` in all RPCError construction sites
- Define 4-byte selector constants for `InvalidRisc0Proof`, `InvalidSp1Proof`, `InvalidTdxProof`, and `AlignedProofVerificationFailed`
- Fix `l1_proof_sender` to match on selectors in `data` instead of strings in `message`
- Fix `l1_proof_verifier` to match on `AlignedProofVerificationFailed` selector instead of the stale `"00m"` code

## How to Test

Test that the proof sender correctly detects and deletes invalid proofs using the new custom error selectors.

### Steps

1. **Start the L1 and deploy contracts with SP1 verifier:**
   ```bash
   cd crates/l2
   make init-l1-docker
   make deploy-l1-sp1
   ```

2. **Start the L2 sequencer with debug logging and no monitor:**
   ```bash
   ETHREX_NO_MONITOR=true ETHREX_LOG_LEVEL=debug make init-l2 2>&1 | grep -E "INFO|WARN|ERROR"
   ```

3. **Inject an invalid proof and skip proof generation:** In `crates/l2/prover/src/prover.rs`, replace lines 121–149:
   ```rust
   let batch_proof = if self.timed {
       self.backend
           .prove_timed(prover_data.input, prover_data.format)
           // ... rest of proving logic ...
   };
   let Ok(batch_proof) = batch_proof.inspect_err(|e| error!("{e}")) else {
       continue;
   };
   ```
   with:
   ```rust
   let batch_proof = BatchProof::ProofCalldata(ethrex_l2_common::prover::ProofCalldata {
       prover_type: self.backend.prover_type(),
       calldata: vec![ethrex_l2_common::calldata::Value::Bytes(vec![0xde, 0xad].into())],
   });
   ```
   This skips proof generation entirely and submits garbage calldata, making the test nearly instant.

4. **Start the SP1 prover:**
   ```bash
   RUST_LOG=info,ethrex_prover_lib=debug make init-prover-sp1
   ```

5. **Observe the proof sender logs.** After the batch is committed and the proof sender tries to verify it on L1, you should see:
   ```
   WARN Deleting invalid SP1 proof for batch N
   ```
   This confirms the proof sender detected the `InvalidSp1Proof` custom error selector from the `data` field of the RPC error response and deleted the invalid proof from the store.

### What to verify
- The proof sender log shows the correct prover type being deleted
- The sequencer does not get stuck — it retries after deleting the bad proof
- No panics or unexpected errors from the missing `data` field on non-revert RPC errors

## Checklist

### DIFF
--- a/crates/networking/rpc/clients/eth/errors.rs
+++ b/crates/networking/rpc/clients/eth/errors.rs
@@ -9,7 +9,7 @@ pub enum RpcRequestError {
         method: String,
         source: serde_json::Error,
     },
-    #[error("{method}: {message}")]
+    #[error("{method}: {message} (data: {data:?})")]
     RPCError {
         method: String,
         message: String,


crates/networking/rpc/clients/eth/errors.rs
```diff
@@ -10,7 +10,11 @@ pub enum RpcRequestError {
         source: serde_json::Error,
     },
     #[error("{method}: {message}")]
```
ElFantasma: nit: The `#[error("{method}: {message}")]` format on the line above doesn't include the new `data` field. When an unknown revert error propagates (i.e. doesn't match any known selector in `l1_proof_sender`/`l1_proof_verifier`), operators won't see the revert selector in logs — makes debugging harder.

Consider:
```rust
#[error("{method}: {message} (data: {data:?})")]
```

avilagaston9: Addressed — changed the format to `#[error("{method}: {message} (data: {data:?})")]` in a8eeaa391.

PR #6207, opened 2026-02-13
perf(l1): optimize storage layer for block execution

ilitteri:
## Motivation

Profiling block execution on mainnet revealed that **lock contention, per-lookup allocations, and O(log n) data structure access** in the storage layer accounted for a significant portion of execution time. The storage hot path (`get_storage_at_root` → trie lookups) was hitting multiple bottlenecks on every trie node access.

## Description

Four incremental optimizations to the storage layer, each targeting a specific bottleneck identified via `perf` profiling on a 32-core AMD server:

### 1. BTreeMap → FxHashMap + Mutex → RwLock
- Replace `BTreeMap<Vec<u8>, Vec<u8>>` with `FxHashMap` for in-memory storage tables — O(1) lookups instead of O(log n)
- Replace `Mutex` with `RwLock` for `trie_cache` and `last_computed_flatkeyvalue` — allows concurrent readers (only the background worker writes, 2-3 times per block)

### 2. RCU pattern + held read view
- **RCU (Read-Copy-Update)** for `InMemoryBackend`: wrap the database as `Arc<RwLock<Arc<Database>>>` so readers clone the inner `Arc` (O(1)) and read entirely lock-free
- **Held read view** in `BackendTrieDB`: acquire the read view once at construction and reuse it for all trie node lookups, eliminating ~8000+ `Box` allocations per block

### 3. Shared read view across trie opens
- Change `begin_read()` to return `Arc<dyn StorageReadView>` instead of `Box<dyn StorageReadView>`, enabling cheap `Arc::clone()` sharing
- In `get_storage_at_root`, pre-acquire the read view, trie cache, and last_written value once, then share across both the state trie and storage trie opens — eliminates duplicate `RwLock` acquisitions
- Add `StorageReadView: Send + Sync` bound (required for `Arc` sharing)

### 4. Pre-compute prefix nibbles in TrieWrapper
- `apply_prefix` was called on every trie node lookup, creating 3 `Vec` allocations per call (`from_bytes` + `append_new` + `concat`)
- Pre-compute the prefix nibbles once in the `TrieWrapper` constructor, reducing to 1 allocation (just `concat`) per lookup
- Add `TrieWrapper::new()` constructor to encapsulate prefix pre-computation

## Benchmark Results (ethrex-replay)

30 runs each, back-to-back on same server (32-core AMD, ethrex-office-4):

| Block | Txs | Gas | Baseline | Optimized | Delta |
|-------|-----|-----|----------|-----------|-------|
| 24443168 | 442 | 37.5M | 57.19ms | 48.32ms | **-15.5%** |
| 24197547 | 556 | 53.0M | 66.31ms | 56.75ms | **-14.4%** |
| 24199455 | 651 | 37.9M | 45.26ms | 38.70ms | **-14.5%** |
| 24198607 | 116 | 12.0M | 22.62ms | 20.88ms | **-7.7%** |

14-15% improvement on heavy blocks, 8% on light blocks. Lock contention dropped from 8.1% to 1.4% of block executor time.

## How to Test

- `cargo test -p ethrex-storage`
- `cargo test -p ethrex-blockchain`
- `cargo clippy -p ethrex-storage -p ethrex-blockchain -- -D warnings`
- Run a full sync or replay a mainnet block to verify identical state roots

### DIFF
--- a/.github/workflows/daily_snapsync.yaml
+++ b/.github/workflows/daily_snapsync.yaml
@@ -86,6 +86,12 @@ jobs:
         run: |
           echo "timestamp=$EPOCHSECONDS" >> "$GITHUB_OUTPUT"
 
+      - name: Cleanup stale Docker and Kurtosis resources
+        run: |
+          kurtosis engine stop || true
+          docker network prune -f
+          docker image prune -f
+
       - name: Run Snapsync Test
         uses: ./.github/actions/snapsync-run
         id: snapsync
@@ -126,6 +132,12 @@ jobs:
         id: start
         run: |
           echo "timestamp=$EPOCHSECONDS" >> "$GITHUB_OUTPUT"
+
+      - name: Cleanup stale Docker and Kurtosis resources
+        run: |
+          kurtosis engine stop || true
+          docker network prune -f
+          docker image prune -f
 
       - name: Run Snapsync Test
         uses: ./.github/actions/snapsync-run
--- a/.github/workflows/manual_docker_performance_publish.yaml
+++ b/.github/workflows/manual_docker_performance_publish.yaml
@@ -2,6 +2,12 @@ name: Publish Performance tag Docker
 
 on:
   workflow_dispatch:
+  push:
+    branches: ["main"]
+
+concurrency:
+  group: performance-docker-publish
+  cancel-in-progress: false
 
 permissions:
   contents: read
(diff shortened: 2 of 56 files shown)


crates/storage/backend/in_memory.rs
```diff
@@ -2,22 +2,23 @@ use crate::api::{
...
-use std::collections::BTreeMap;
+use rustc_hash::FxHashMap;
+use std::collections::HashMap;
 use std::path::Path;
 use std::sync::{Arc, RwLock};
 
-type Table = BTreeMap<Vec<u8>, Vec<u8>>;
-type Database = BTreeMap<&'static str, Table>;
+type Table = FxHashMap<Vec<u8>, Vec<u8>>;
+type Database = HashMap<&'static str, Table>;
```
edg-l: why not fxhashmap for this one?

crates/storage/store.rs
```diff
@@ -2085,18 +2085,40 @@ impl Store {
...
         let storage_root = if self.flatkeyvalue_computed(account_hash)? {
             // We will use FKVs, we don't need the root
             *EMPTY_TRIE_HASH
         } else {
-            let state_trie = self.open_state_trie(state_root)?;
+            let state_trie = self.open_state_trie_shared(
+                state_root,
+                read_view.clone(),
+                cache.clone(),
+                last_written.clone(),
```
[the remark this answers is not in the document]
ilitteri: Thanks. Confirmed, this was intentional to avoid redundant lock/snapshot setup in this hot path.

crates/storage/backend/in_memory.rs
```diff
@@ -152,7 +150,8 @@ impl StorageWriteBatch for InMemoryWriteTx {
             .write()
             .map_err(|_| StoreError::Custom("Failed to acquire write lock".to_string()))?;
 
-        let table_ref = db.entry(table).or_insert_with(Table::new);
+        let db_mut = Arc::make_mut(&mut *db);
```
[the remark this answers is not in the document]
ilitteri: Good callout. This is an intentional RCU/COW tradeoff for the in-memory backend: readers get lock-free snapshot reads, and writes may clone if snapshots are still alive. I added explicit comments in `crates/storage/backend/in_memory.rs` documenting this behavior so the cost model is clear.

crates/storage/backend/in_memory.rs
ilitteri: Good catch. I switched the outer `Database` map to `FxHashMap` as well in `99c31db`.

crates/storage/store.rs
```diff
@@ -2645,15 +2710,14 @@ impl Store {
...
             .map_err(|_| StoreError::LockError)?;
         Ok(last_computed_flatkeyvalue.clone())
     }
 
-    fn flatkeyvalue_computed(&self, account: H256) -> Result<bool, StoreError> {
+    fn flatkeyvalue_computed_with_last_written(account: H256, last_written: &[u8]) -> bool {
         let account_nibbles = Nibbles::from_bytes(account.as_bytes());
-        let last_computed_flatkeyvalue = self.last_written()?;
-        Ok(&last_computed_flatkeyvalue[0..64] > account_nibbles.as_ref())
+        &last_written[0..64] > account_nibbles.as_ref()
```
ElFantasma, line 2720: nit: `&last_written[0..64]` will panic if the slice is shorter than 64 bytes. The caller (`self.last_written()`) always returns ≥64 bytes (initialized from `unwrap_or_else(|| vec![0u8; 64])`), but this static method's `&[u8]` signature doesn't encode that invariant. Consider:
```rust
last_written.get(0..64).is_some_and(|lw| lw > account_nibbles.as_ref())
```

crates/storage/store.rs
```diff
@@ -2695,13 +2759,13 @@ fn apply_trie_updates(
...
     let trie = trie_cache
-        .lock()
+        .read()
         .map_err(|_| StoreError::LockError)?
         .clone();
     let mut trie_mut = (*trie).clone();
     trie_mut.put_batch(parent_state_root, child_state_root, new_layer);
     let trie = Arc::new(trie_mut);
-    *trie_cache.lock().map_err(|_| StoreError::LockError)? = trie.clone();
+    *trie_cache.write().map_err(|_| StoreError::LockError)? = trie.clone();
```
ElFantasma, line 2768: nit: The `.read()` → clone → modify → `.write()` RCU pattern (here and again at line 2833) is safe because only the background worker thread calls `apply_trie_updates`. Worth adding a brief comment at the `.write()` sites to make this single-writer invariant explicit — e.g., `// Single writer: only called from trie update worker thread`.

crates/vm/levm/src/timings.rs
iovoid, line 1: I feel like the addition of the SLOAD counters might belong in another PR.

PR #6208, opened 2026-02-16
feat(l1): add offline snap sync profiling with dataset capture and replay

ilitteri:
## Motivation

We need to measure and optimize snap sync compute phases (InsertAccounts, InsertStorages) without requiring a live network. This enables reproducible benchmarking, profiling with external tools (perf, samply, heaptrack, jemalloc), and regression detection across commits.

## Description

Adds an offline profiling system for snap sync with two main components:

**Dataset capture** — During a live snap sync, setting `ETHREX_SNAP_PROFILE_CAPTURE_DIR` captures the state data as RLP chunks + a versioned manifest. Works with both in-memory and rocksdb builds (SST files are converted to RLP during capture).

**Offline replay** — `snap_profile_replay` example binary replays a captured dataset against an isolated store, measuring per-phase timings. Supports both in-memory and rocksdb backends.

Changes:
- `crates/networking/p2p/sync/profile.rs` — manifest types, `load_manifest()`, `run_once_with_opts()`, `ProfileBackend` enum, 7 unit tests
- `crates/networking/p2p/sync/snap_sync.rs` — `capture_snap_profile_dataset()`, SST-to-RLP conversion, streaming for large storage entries
- `crates/networking/p2p/examples/snap_profile_replay.rs` — standalone replay binary with `--backend`, `--db-dir`, `--keep-db`
- `docs/l1/fundamentals/snap_sync.md` — operator docs for offline profiling
- `tooling/snapsync_profile/` — 4 shell scripts for perf/samply/heaptrack/jemalloc profiling
- `Makefile` — 4 Make targets (`snapsync-prof-perf`, `snapsync-prof-samply`, `snapsync-prof-heap`, `snapsync-prof-jeprof`)

Validated end-to-end on Hoodi: 21.5M accounts, 280M storage slots, 14GB dataset, replay in 17 minutes with rocksdb backend (~5GB peak RSS).

## Test plan

- [ ] `cargo test -p ethrex-p2p` — 7 new unit tests for profile module
- [ ] `cargo build --release --example snap_profile_replay -p ethrex-p2p --features rocksdb,c-kzg` — builds successfully
- [ ] Normal snap sync unaffected when `ETHREX_SNAP_PROFILE_CAPTURE_DIR` is unset


tooling/snapsync_profile/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+        let account_states_snapshot: Vec<(H256, AccountState)> =
+            RLPDecode::decode(&snapshot_contents)
+                .map_err(|_| SyncError::SnapshotDecodeError(snapshot_path.clone()))?;
+
+        info!("Inserting accounts into the state trie");
+
+        let store_clone = store.clone();
+        let current_state_root: Result<H256, SyncError> =
+            tokio::task::spawn_blocking(move || -> Result<H256, SyncError> {
+                let mut trie = store_clone.open_direct_state_trie(computed_state_root)?;
```
ElFantasma, line 171: This uses `open_direct_state_trie` (with WAL enabled), but the production snap sync path calls `open_direct_state_trie_no_wal`. The WAL write overhead could skew profiling results, making replay non-representative of actual snap sync performance.

Same applies to storage tries if they follow a similar pattern — worth checking `open_direct_storage_trie` vs `open_direct_storage_trie_no_wal` in `compute_storage_roots`.

Suggestion: use `open_direct_state_trie_no_wal` here to match production behavior.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -926,9 +944,264 @@ async fn insert_storages(
...
+            chunk.clear();
+            *entries_in_chunk = 0;
+            *chunk_idx += 1;
+            Ok::<(), SyncError>(())
+        };
+
+    for item in db.full_iterator(rocksdb::IteratorMode::Start) {
+        let (key, value) =
+            item.map_err(|e| SyncError::FileSystem(format!("DB iter error: {e}")))?;
+        let account_hash = H256::from_slice(&key[0..32]);
```
ElFantasma, line 1149: `&key[0..32]` and `&key[32..64]` will panic if an SST entry has a key shorter than 64 bytes (e.g., from a corrupted SST file). Since this processes user-captured data, a bounds check would be more robust:

```rust
if key.len() != 64 {
    return Err(SyncError::FileSystem(format!(
        "Unexpected SST key length {}: expected 64", key.len()
    )));
}
```

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -926,9 +944,264 @@ async fn insert_storages(
...
+        std::fs::write(&path, chunk.encode_to_vec())
+            .map_err(|e| SyncError::FileSystem(format!("Failed to write {path:?}: {e}")))?;
+    }
+
+    drop(db);
+    let _ = std::fs::remove_dir_all(&tmp_db_dir);
+
+    info!(
+        "Converted account SST files to {} RLP chunks in {rlp_dir:?}",
+        chunk_idx + 1
```
iovoid, line 1092: The +1 should be inside the !is_empty branch to avoid off-by-one errors.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -994,8 +1267,7 @@ async fn insert_accounts(
 
     drop(db); // close db before removing directory
 
-    std::fs::remove_dir_all(account_state_snapshots_dir)
-        .map_err(|_| SyncError::AccountStateSnapshotsDirNotFound)?;
+    // Snapshot dir cleanup is handled by the caller (conditional on capture mode).
```
iovoid, line 1264: Note: this change will increase the required (peak) disk usage even when capture is disabled, since we keep the account states directory until the end.

crates/networking/p2p/sync/profile.rs
```diff
@@ -0,0 +1,230 @@
...
+
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SnapProfileManifest {
+    pub version: u32,
+    pub chain_id: u64,
+    pub rocksdb_enabled: bool,
```
iovoid, line 21: What is the purpose of this field, if it's always set to false?

Issue #6211, opened 2026-02-16
[BAL] Batched state reads


edg-l:
Improve state prefetching using EIP-7928 Block Access Lists (Amsterdam fork), building on the existing `warm_block()` prewarmer.

ethrex already has a per-sender parallel prewarmer (`LEVM::warm_block()` in `crates/vm/backends/levm/mod.rs`). It speculatively executes transactions grouped by sender to warm the shared `CachingDatabase`. However, this approach has limitations: it re-executes full transactions (costly), can miss cross-sender state, and doesn't know the exact read set upfront.

BALs provide the **exact** set of accounts and storage slots accessed in a block before execution begins. This enables more efficient prefetching:

**Opportunities:**
- **BAL-based prefetch**: Replace or complement speculative prewarming with direct cache loading from the BAL's account/storage list — no need to re-execute transactions.
- **RocksDB MultiGet**: Batch trie node fetches via RocksDB's `multi_get()` instead of individual trie traversals per key.
- **Parallel trie node fetching**: Use the BAL to identify independent subtrie branches and parallelize node I/O.

EIP-7928 explicitly targets reducing execution to "parallel IO + parallel EVM" — BAL-based prefetching is the "parallel IO" half and should be more efficient than speculative prewarming.

**Key challenges:** trie traversal dependency chains (account storage root needed before storage slots), integration with `StorageBackend`/`StorageReadView` traits, and deciding whether to replace or run alongside the existing prewarmer.

**References:**
- [EIP-7928: Block-Level Access Lists](https://eips.ethereum.org/EIPS/eip-7928)

Issue #6210, opened 2026-02-16
[BAL] Parallel state root calculation


edg-l:
Improve state root calculation parallelism leveraging EIP-7928 Block Access Lists (Amsterdam fork).

BALs include post-execution values for every modified account and storage slot, organized by address. This enables distributing trie updates across workers without re-executing transactions, and knowing the full set of dirty nodes upfront.

Currently (`crates/blockchain/blockchain.rs` lines 514-533), ethrex splits accounts into 16 shards by address hash prefix and hashes each shard in parallel. Within each shard, trie updates and hashing are sequential (`handle_merkleization_subtrie()`).

**Opportunities:**
- **BAL-guided work distribution**: Use the BAL's per-account structure to distribute trie updates more evenly across workers (not just by prefix).
- **Intra-shard parallelism**: Branch nodes have 16 independent children whose subtree hashes can be computed in parallel.
- **Storage trie parallelism**: Each account's storage trie root is independent; within each, branch children are also independent.

**Key challenges:** spawn overhead vs. benefit tradeoff, lock contention on shared parent nodes, and memory overhead from parallel RLP encoding buffers.

**References:**
- [EIP-7928: Block-Level Access Lists](https://eips.ethereum.org/EIPS/eip-7928)

Issue #6209, opened 2026-02-16
[BAL] Parallel block execution


edg-l:
Implement parallel transaction execution leveraging EIP-7928 Block Access Lists, introduced in the Amsterdam fork.

BALs provide advance knowledge of every account and storage slot accessed during block execution, including per-transaction `block_access_index` markers. This allows identifying tx dependencies upfront and executing non-conflicting transactions concurrently.

Currently, ethrex executes transactions sequentially in `execute_block()` (`crates/blockchain/blockchain.rs`), where each tx runs against the state committed by the previous one.

**Approach:**
- Use BAL data to build a dependency graph: txs that touch disjoint accounts/slots can run in parallel.
- For conflicting txs, apply Block-STM style speculative execution with conflict detection and retry.
- EIP-7928 states the goal is reducing execution time to "parallel IO + parallel EVM."

**Key challenges:** state conflicts between txs touching shared accounts/slots, preserving deterministic final state, and overhead from conflict detection on contended workloads (MEV bundles, DEX pools).

**References:**
- [EIP-7928: Block-Level Access Lists](https://eips.ethereum.org/EIPS/eip-7928)
- [Block-STM (PPoPP 2023)](https://arxiv.org/abs/2203.06871)
- [ParallelEVM (EuroSys 2025)](https://yajin.org/papers/EuroSys_2025_ParallelEVM.pdf)

edg-l:
EIP-7928 Block Access Lists give validators advance knowledge of every account/slot touched during block execution, including per-transaction `block_access_index` markers. This enables parallel execution of non-conflicting transactions. Research suggests 60–80% of transactions access disjoint state and can be fully parallelized.

ethrex already has:
- BAL recording, validation (hash + size cap), and RLP encoding
- BAL-based parallel prewarming (`warm_block_from_bal`)
- Pipelined merkleization with 16 storage + 16 state shard workers
- Sequential tx execution in `LEVM::execute_block()` / `execute_block_pipeline()`

The goal is to replace sequential tx execution with parallel execution while preserving deterministic final state.

---

## Proposed Approach: Phased Implementation

### Phase 1 — BAL-based dependency graph + independent group execution

Parse the BAL to build a tx dependency graph. Group independent transactions and execute groups on separate threads. This captures most parallelism (60–80%) with manageable complexity.

### Phase 2 — Block-STM for conflicting transaction chains

[...]

PR #6216, opened 2026-02-17
feat(l1): bal-devnet-3 

edg-l:
## Motivation

Tracking branch for Amsterdam bal-devnet-3 alignment.

### EIP-7708: Rename Selfdestruct event to Burn
- Updated topic hash to `0xcc16f5...` (`BURN_EVENT_TOPIC`)
- Renamed `create_selfdestruct_log` to `create_burn_log`

### EIP-7928: BAL size cap
- `validate_block_access_list_size` in validation.rs (3 call sites in blockchain.rs)
- Fixed formula to match EELS: `bal_items > gas_limit / GAS_BLOCK_ACCESS_LIST_ITEM` (no system allowance or per-tx base cost)

### EIP-8024: Branchless normalization + EXCHANGE extension
- `decode_single` uses `(x + 145) % 256` instead of branching
- `decode_pair` uses XOR `0x8F` instead of conditional subtraction
- EXCHANGE valid range extended from `0x00-0x4F` to `0x00-0x51` (two new swappable pairs)

### EIP-8037: State Creation Gas Cost Increase
Full implementation of the reservoir model:
- State gas reservoir derived from excess `gas_limit` beyond `TX_MAX_GAS_LIMIT`
- Two-dimensional block gas accounting: `block.gas_used = max(sum(regular), sum(state))` per EIP-7778
- Receipt `cumulative_gas_used` = post-refund total (what user pays)
- Reservoir subtraction from gas_used in finalize_execution
- EIP-7702 auth refund to reservoir for Amsterdam+
- Skip Osaka TX_MAX_GAS_LIMIT cap for Amsterdam (reservoir model replaces per-tx limit)
- CREATE state gas charged before early-failure checks (balance/depth/nonce) and before child gas reservation (63/64 rule)
- SSTORE state gas charged before regular gas per EELS ordering; refund via normal refund counter (subject to 1/5 cap per EIP-3529)
- `increase_state_gas` reordered: charge first, mutate counters only on success (matches EELS)
- State gas restoration on child revert: `new_reservoir = current_reservoir + child_state_gas_used` (matches EELS `incorporate_child_on_error`)
- Nested child revert fix: correctly restores reservoir when sub-child also reverted
- CREATE collision gas excluded from regular dimension (EELS escrow mechanism — `gas_used=0` on collision)
- Amsterdam intrinsic regular gas cap validation (`max(intrinsic_regular, calldata_floor) > TX_MAX_GAS_LIMIT`)

## BAL Parallel Execution Validation

Comprehensive BAL validation for the parallel block execution path:
- `validate_bal_withdrawal_index`: verify withdrawal/request-extraction index entries against actual post-withdrawal state
- `validate_tx_execution`: smarter PART B check comparing absent-from-BAL accounts with pre-execution state to distinguish warm-access artifacts from genuine missing entries
- `validate_ordering`: storage conflict check (slot in both storage_changes and storage_reads)
- `validate_header_bal_indices`: validate BAL indices don't exceed max valid index for the block
- `accessed_accounts` tracker in `GeneralizedDatabase` to detect extraneous BAL "pure-access" entries
- Edge case fixes: EXTCODECOPY with size=0 now loads target account; storage_read validation handles DestroyedModified status
- Parallel gas accounting: track regular and state gas separately, compute `block_gas_used = max(regular, state)` matching sequential path
- BAL hash mismatch fix in block building (#6317)

## CI Changes

- Added Amsterdam consume-engine hive tests (~1000 tests) to PR workflow using bal@v5.3.0 fixtures
- Bumped daily hive + ef_tests fixtures to bal@v5.3.0
- Amsterdam hive tests marked as optional (fork spec still evolving)

## Roadmap

- [x] EIP-7708: Rename `Selfdestruct` event to `Burn` with updated topic hash
- [x] EIP-7928: Implement BAL max items cap (corrected formula)
- [x] EIP-8024: Branchless normalization + EXCHANGE extension
- [x] EIP-8037: State creation gas cost increase — full reservoir model
- [x] Bump Amsterdam fixtures to bal@v5.3.0
- [x] [Batched state reads](https://github.com/lambdaclass/ethrex/issues/6211)
- [x] [Parallel block execution](https://github.com/lambdaclass/ethrex/issues/6209)
- [x] [Parallel state root calculation](https://github.com/lambdaclass/ethrex/issues/6210)
- [x] BAL parallel execution validation (extraneous entry detection)
- [x] Amsterdam consume-engine hive tests in PR CI

## Checklist


crates/vm/backends/levm/mod.rs
```diff
@@ -736,57 +934,110 @@ impl LEVM {
...
+        //    exceeding the gas limit produce GAS_USED_OVERFLOW instead of a BAL
+        //    mismatch error (the BAL is built assuming rejected txs, so the miner
+        //    balance in the BAL won't match execution that ran all txs).
+        let mut block_regular_gas_used = 0_u64;
+        let mut block_state_gas_used = 0_u64;
+        for (tx_idx, _, report, _, _, _) in &exec_results {
+            // Prospective check: same logic as sequential check_gas_limit.
+            // Rejects tx if running_gas + tx.gas_limit() > block_limit,
+            // even if actual usage would fit. This matches sequential behavior.
+            let running_block_gas = block_regular_gas_used.max(block_state_gas_used);
```
ElFantasma, line 968: This uses `max(regular, state)` for the pre-tx check, but the comment above says "same logic as sequential `check_gas_limit`" — it isn't. #6309 deliberately switched sequential to `cumulative_gas_used` (post-refund) to fix a bal-devnet-2 rejection bug, and #6342 extends that same pattern here.

Using `max(regular, state)` is stricter and can still reject valid blocks: if a tx's gas goes entirely to the *smaller* dimension, the max doesn't increase, so `max + tx.gas_limit` overshoots even when the block is valid.

Since the post-execution `validate_gas_used` is the authoritative check, the pre-tx guard should be as permissive as possible to avoid false rejections during validation.

crates/blockchain/payload.rs
```diff
@@ -776,15 +810,39 @@ pub fn apply_plain_transaction(
...
+            context.block_regular_gas_used,
+            context.block_state_gas_used,
+            context
+                .block_regular_gas_used
+                .max(context.block_state_gas_used),
+        );
+    }
+
+    // Update remaining_gas using gas_used (pre-refund) for block gas limit checks
+    context.remaining_gas = context.remaining_gas.saturating_sub(report.gas_used);
```
ElFantasma: `report.gas_used` is the sum of regular + state gas, but Amsterdam block capacity is `max(sum_regular, sum_state)`. Decrementing `remaining_gas` by the sum makes it hit zero faster than actual block capacity allows.

Example: gas_limit=30M, tx1 uses regular=8M + state=12M (gas_used=20M) → remaining_gas=10M. Next tx with gas_limit=12M is skipped, but actual block gas_used = max(8M, 12M) = 12M, leaving 18M of capacity.

For Amsterdam, `remaining_gas` could track `gas_limit - max(block_regular, block_state)` instead, or the tx selection loop could check against `gas_used()` directly.

ElFantasma (changes requested):

crates/vm/backends/levm/mod.rs
edg-l: Good catch — the parallel path was using `max(regular, state)` while sequential used `cumulative_gas_used` (post-refund sum). Both are now aligned to `max(regular, state)` per EIP-8037.

Also fixed the overflow-safe comparison (`tx_gas_limit > gas_limit.saturating_sub(used)` instead of `used + tx_gas_limit > gas_limit`).

Fixed in ff74f2d7f5.

crates/blockchain/payload.rs
edg-l: Valid point — the payload builder was subtracting `report.gas_used` (regular + state sum), but EIP-8037 block capacity is `max(sum_regular, sum_state)`. Now Amsterdam tracks `remaining_gas = gas_limit - max(block_regular, block_state)` so we don't skip transactions that would fit.

Fixed in ff74f2d7f5.

ElFantasma (approved):

crates/vm/levm/src/call_frame.rs
```diff
@@ -283,6 +283,10 @@ pub struct CallFrame {
     pub ret_size: usize,
     /// If true then transfer value from caller to callee
     pub should_transfer_value: bool,
+    /// EIP-8037: snapshot of VM.state_gas_reservoir at the start of this frame (for revert restoration)
+    pub reservoir_snapshot: u64,
```
iovoid: This doesn't seem to be actually used?

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -590,7 +620,7 @@ impl OpcodeHandler for OpSelfDestructHandler {
                 let log = if to != beneficiary {
                     create_eth_transfer_log(to, beneficiary, balance)
                 } else {
-                    create_selfdestruct_log(to, balance)
+                    create_burn_log(to, balance)
                 };
                 vm.substate.add_log(log);
             }
```
iovoid: This can't happen since Amsterdam>Cancun

crates/vm/backends/levm/mod.rs
```diff
@@ -1072,13 +1333,25 @@ impl LEVM {
...
+                    .get(addr)
+                    .map(|a| (a.info.balance, a.info.nonce))
+                    .or_else(|| {
+                        store
+                            .get_account_state(*addr)
+                            .ok()
+                            .map(|a| (a.balance, a.nonce))
+                    })
+                    .unwrap_or_default();
+                let post = (account.info.balance, account.info.nonce);
```
iovoid: Should we check code_hash? If not maybe leave a comment explaining why.

Makefile
iovoid: Why does it differ from the list on pr-main_l1.yaml (8024|7708|7778|7843|7928|7954|8037)

edg-l: its outdated ill fix it

iovoid (approved):

ilitteri (approved):

crates/vm/levm/src/utils.rs
```diff
@@ -333,12 +336,30 @@ impl<'a> VM<'a> {
...
-                let refunded_gas_if_exists = PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST;
-                refunded_gas = refunded_gas
-                    .checked_add(refunded_gas_if_exists)
-                    .ok_or(InternalError::Overflow)?;
+            // 7. Refund if authority exists in the trie.
+            // EIP-8037 (Amsterdam+): return STATE_BYTES_PER_NEW_ACCOUNT * COST_PER_STATE_BYTE
+            // to the state gas reservoir (the new-account portion of the auth state charge).
+            // Pre-Amsterdam: add REFUND_AUTH_PER_EXISTING_ACCOUNT (12500) to global refund counter.
+            // NOTE: Uses `exists` (account_exists in EELS / Exist in geth), NOT `!is_empty()`.
+            // An account can exist in the trie but be empty (e.g., has non-empty storage root).
```
JereSalo, line 344: Huh, I remember changing EIP-7702 so that we could use `is_empty()` instead of `exists` to make it simpler. It seems that in this new EIP it hasn't been taken into consideration. I wonder if we should propose the same change to that EIP as well given that it may not be necessary. I didn't analyze it in depth but I don't see a reason not to change it though. Anyway, it's just a minor change and perhaps not so relevant.

https://github.com/ethereum/EIPs/pull/9710

PR #6222, opened 2026-02-18
fix(l1): add tests to account creation and deletion

Arkenan:
**Motivation**

This is a relatively difficult edge case to identify and have run into it in different forms multiple times.

**Description**

We add an integration test for PR #6219

### DIFF
--- a/test/tests/levm/batch_execution_tests.rs
+++ b/test/tests/levm/batch_execution_tests.rs
@@ -21,7 +21,7 @@ use ethrex_levm::{
 use rustc_hash::FxHashMap;
 use std::sync::Arc;
 
-// ==================== Test Database ====================
+// ==================== Test Helpers ====================
 
 /// Minimal in-memory database that returns empty/default for everything.
 struct EmptyDatabase;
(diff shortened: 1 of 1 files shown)


test/tests/levm/batch_execution_tests.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/// Verify that removed_storage IS correctly set when the account actually
+/// had storage in the trie (i.e. existed before the batch started).
+///
+/// This tests that the fix doesn't break the legitimate case: an account
+/// that existed in the trie with storage, gets SELFDESTRUCT'd, and then
+/// gets modified (e.g., receives ETH). Its storage should be cleared.
+#[test]
+fn batch_removed_storage_set_when_account_had_trie_storage() {
```
ElFantasma, line 158: This test only covers `get_state_transitions()`. Since `get_state_transitions_tx()` has the same `removed_storage` logic (and received the same fix in #6219), it would be good to add a `_tx` variant — the same way tests 1 and 2 mirror each other for the spurious-update case.

test/tests/levm/batch_execution_tests.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/// Test that a DestroyedModified account with NO prior trie storage but with
+/// balance changes still gets an info update (but NOT removed_storage).
+///
+/// Scenario: Account didn't exist before batch, was created, destroyed, then
+/// received some ETH. It should appear in the trie with the new balance, but
+/// removed_storage should be false (nothing to remove from trie).
+#[test]
+fn batch_destroyed_modified_with_balance_but_no_prior_storage() {
```
ElFantasma, line 201: Same as above — this scenario only tests `get_state_transitions()`. A `_tx` variant would ensure parity with the first two tests and catch any future divergence between the two code paths.

PR #6224, opened 2026-02-18
fix(levm): backup account status

iovoid:
**Motivation**

Recently we had a bug in which an empty account was incorrectly inserted into the trie. The proximate cause was fixed in #6219, but the problem runs deeper.

The code used to send an update with removed_storage=true when an account was in the DestroyedModified state, but the problem is this state shouldn't have been reached since the account wasn't modified after being destroyed, otherwise it would have state.

The problem is that the state goes from Destroyed to DestroyedModified when the account is modified through get_account_mut, but isn't restored if the modification is reverted.

**Description**

This PR fixes the backup and adds an integration test that should fail unless one (this PR's and #6219's) of the fixes is present.


iovoid:
Superceded by #6225

PR #6225, opened 2026-02-18
test(l1): add regression test for batch selfdestruct spurious state root

azteca1998:
## Summary
- Adds integration tests exercising the `DestroyedModified` account state transition in batch execution, validating the fix from #6219.
- The main test deploys a self-destructing contract (Block 1), sends a 0-value tx to the destroyed address (Block 2), then re-executes both blocks via `add_blocks_in_batch` and asserts the state root matches.
- Without the fix, a spurious empty `AccountState` would be inserted into the state trie, causing a `StateRootMismatch`.

## Test plan
- [x] `batch_selfdestruct_created_account_no_spurious_state` passes with the fix
- [x] `batch_selfdestruct_created_account_no_spurious_state` fails (StateRootMismatch) when the fix is reverted
- [x] `batch_single_block_selfdestruct` passes (single-block batch sanity check)
- [x] Existing tests unaffected

### DIFF
--- a/test/tests/blockchain/batch_tests.rs
+++ b/test/tests/blockchain/batch_tests.rs
@@ -9,8 +9,8 @@ use ethrex_common::{
     Address, H160, H256, U256,
     evm::calculate_create_address,
     types::{
-        Block, BlockHeader, EIP1559Transaction, GenesisAccount, Transaction, TxKind,
-        DEFAULT_BUILDER_GAS_CEIL, ELASTICITY_MULTIPLIER,
+        Block, BlockHeader, DEFAULT_BUILDER_GAS_CEIL, EIP1559Transaction, ELASTICITY_MULTIPLIER,
+        GenesisAccount, Transaction, TxKind,
     },
 };
 use ethrex_l2_rpc::signer::{LocalSigner, Signable, Signer};
@@ -179,6 +179,11 @@ async fn batch_selfdestruct_created_account_no_spurious_state() {
 
     // 4. Build block 1 and validate it via single-block path.
     let block1 = build_block(&store_a, &blockchain_a, &genesis_header).await;
+    assert_eq!(
+        block1.body.transactions.len(),
+        1,
+        "block1 must include the deploy tx"
+    );
     blockchain_a
         .add_block(block1.clone())
         .expect("block1 should be valid (single-block)");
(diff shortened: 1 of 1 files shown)


test/tests/blockchain/batch_tests.rs
```diff
@@ -0,0 +1,281 @@
...
+        .expect("tx2 should enter pool");
+
+    // Remove Block 1's transactions from the pool so they don't interfere
+    // with building Block 2 (the mempool doesn't auto-prune on add_block).
+    blockchain_a
+        .remove_block_transactions_from_pool(&block1)
+        .expect("should remove block1 txs from pool");
+
+    // 6. Build block 2 and validate it via single-block path.
+    let block2 = build_block(&store_a, &blockchain_a, &block1.header).await;
```
ElFantasma, line 224: Same here — assert block2 includes tx2. This is the transaction that triggers the `Destroyed` -> `DestroyedModified` transition; if it's not included, the regression scenario is not exercised.

test/tests/blockchain/batch_tests.rs
```diff
@@ -0,0 +1,281 @@
...
+    let beneficiary = Address::from_low_u64_be(0xBEEF);
+
+    let tx = create_selfdestruct_deploy_tx(chain_id, 0, beneficiary, &signer).await;
+
+    blockchain_a
+        .add_transaction_to_pool(tx)
+        .await
+        .expect("tx should enter pool");
+
+    let block1 = build_block(&store_a, &blockchain_a, &genesis_header).await;
```
ElFantasma, line 273: Same suggestion for the single-block test: assert that block1 contains the deploy transaction.

ElFantasma (changes requested):
Clean, well-documented regression test that directly exercises the `DestroyedModified` batch bug from #6219. The test logic, EVM bytecode, and helper extraction all look correct.

The main concern is that neither test asserts the built blocks actually contain the expected transactions. If `build_block` produced empty blocks (e.g. due to mempool validation rejecting the tx), the batch would succeed trivially without exercising the selfdestruct path at all.

**test/tests/blockchain/batch_tests.rs:181** — Consider asserting that block1 actually contains tx1, e.g. `assert_eq!(block1.body.transactions.len(), 1, "block1 must include the deploy tx")`. Without this, if `build_block` fails to pick up tx1 from the mempool (e.g. nonce or fee validation rejects it), the block will be empty and `add_blocks_in_batch` will pass without ever hitting the selfdestruct/`DestroyedModified` path — the test would succeed vacuously. (See inline comments for block2 and the single-block test.)

iovoid (approved):

Arkenan (approved):
LGTM

PR #6226, opened 2026-02-18
fix(l1): fix snap-to-full sync auto-switch on mainnet

ilitteri:
## Motivation

After PR #6060, the auto-switch from snap to full sync at startup doesn't fire on mainnet. Every restart goes through snap sync (~15-20s wasted) before eventually switching to full sync mid-cycle. The snap progress display also continues printing after the switch.

## Description

Two bugs:

1. **Auto-switch never fires on mainnet** (`sync_manager.rs`): The check relied on `merge_netsplit_block`, which mainnet's genesis doesn't set (always `None`). Now uses `terminal_total_difficulty_passed` (which mainnet sets to `true`) — any stored block > 0 triggers the switch. Also removes the `!has_checkpoint` guard and clears stale snap state instead of skipping the check.

2. **Snap progress display continues after mode switch** (`snap_sync.rs`): When `sync_cycle_snap` switches to full sync internally, `METRICS.disable()` only runs after the outer function returns. Since full sync runs inside `sync_cycle_snap`, the snap progress bar kept printing. Now disables metrics before entering full sync.

## How to Test

1. Run ethrex on mainnet with `--syncmode snap` on a node that has previously synced state
2. On restart, logs should show `"Node has synced state (block N), switching to full sync"` immediately
3. No `"SNAP SYNC STARTED"` banner should appear

### DIFF
--- a/crates/networking/p2p/sync_manager.rs
+++ b/crates/networking/p2p/sync_manager.rs
@@ -68,13 +68,9 @@ impl SyncManager {
                 false
             };
             if is_synced {
-                info!(
-                    "Node has synced state (block {latest_block}), switching to full sync"
-                );
+                info!("Node has synced state (block {latest_block}), switching to full sync");
                 snap_enabled.store(false, Ordering::Relaxed);
-                if has_checkpoint
-                    && let Err(e) = store.clear_snap_state().await
-                {
+                if has_checkpoint && let Err(e) = store.clear_snap_state().await {
                     warn!("Failed to clear stale snap state: {e}");
                 }
             }


crates/networking/p2p/sync_manager.rs
```diff
@@ -54,21 +54,29 @@ impl SyncManager {
...
+            } else {
+                false
+            };
+            if is_synced {
                 info!(
-                    "Node has synced state (block {latest_block} > merge block {merge_block}), switching to full sync"
+                    "Node has synced state (block {latest_block}), switching to full sync"
                 );
                 snap_enabled.store(false, Ordering::Relaxed);
+                if has_checkpoint {
```
ElFantasma: nit: Since `is_synced` is true here, the node is already past the merge with stored blocks. Could a stale checkpoint realistically exist in this scenario? If so, might be worth a brief comment explaining when that happens (e.g., node crashed mid-snap-sync, restarted, and now has enough blocks from a previous successful sync).

ElFantasma (changes requested):
One issue not tied to a diff line: **sync_manager.rs:98** — After the auto-switch clears snap state via `clear_snap_state()` (line 76), the `has_checkpoint` boolean at line 98 still holds its original `true` value. This causes `start_sync()` to be called unnecessarily. The spawned task will loop on `"Resuming sync after node restart, waiting for next FCU"` every 5 seconds (since `last_fcu_head` is still `H256::zero()`), which is misleading and wasteful. Consider re-reading the checkpoint after the auto-switch block, or skipping `start_sync()` when `is_synced` was true:

```rust
if has_checkpoint && !is_synced_and_switched {
    sync_manager.start_sync();
}
```

Or simply move `has_checkpoint` to be read after the auto-switch block (it's only used in those two places).

ilitteri:
> One issue not tied to a diff line: **sync_manager.rs:98** — After the auto-switch clears snap state via `clear_snap_state()` (line 76), the `has_checkpoint` boolean at line 98 still holds its original `true` value. This causes `start_sync()` to be called unnecessarily. The spawned task will loop on `"Resuming sync after node restart, waiting for next FCU"` every 5 seconds (since `last_fcu_head` is still `H256::zero()`), which is misleading and wasteful. Consider re-reading the checkpoint after the auto-switch block, or skipping `start_sync()` when `is_synced` was true:
> 
> ```rust
> if has_checkpoint && !is_synced_and_switched {
>     sync_manager.start_sync();
> }
> ```
> 
> Or simply move `has_checkpoint` to be read after the auto-switch block (it's only used in those two places).

Fixed in [17014c0](https://github.com/lambdaclass/ethrex/pull/6226/commits/17014c01000c2d0cf721f07c1a1ad931487a5e94)

PR #6227, opened 2026-02-19
feat(l1): bal batched read and parallel state root calculation

edg-l:
**Motivation**

Two performance optimizations for the BAL (Block Access List) pipeline introduced in Amsterdam BAL (EIP-7928):

1. **BAL-based batched state reads for block prewarming** — instead of reading state entries individually before block execution, batch all reads from the BAL into a single RocksDB operation. This reduces overhead per entry and improves cache locality.
2. **Parallel state root calculation via BAL** — use the BAL's known write set to begin merkle trie updates concurrently with EVM execution, overlapping what was previously a sequential post-execution step.

**Description**

- `batched state reads`: on `fcU` with payload attributes, the warmer thread now issues a single batched read for all BAL entries rather than N individual reads.
- `parallel state root`: spawns a worker thread during execution that begins computing the state root over the BAL write set concurrently. After execution completes, only the remaining diff needs to be flushed.

**Benchmark**

The Kurtosis network setup and comparison script used to produce these results live in the [`bal-optimizations-kurtosis`](https://github.com/lambdaclass/ethrex/tree/bal-optimizations-kurtosis) branch. See `fixtures/networks/bal-benchmark.yaml` for the network config and `scripts/benchmark_compare.py` for the metrics collection script.

A local Kurtosis network was run with two ethrex nodes side by side — one built from `main` and one from this branch (`bal-opt`) — both connected to the same Lighthouse consensus clients on the `bal-devnet-2` devnet spec (Fulu at genesis, Gloas at epoch 1). The network used `spamoor` with `gasburnertx`, `evm-fuzz`, `uniswap-swaps`, and `eoatx` scenarios to generate sustained load (~60-115 txs/block).

**Important caveat:** both nodes ran on the same physical machine sharing CPU, memory, and disk I/O. This means results are not absolute performance numbers — they reflect relative behavior under identical resource contention. Any measured difference underestimates the real-world benefit since both nodes compete for the same resources.

Metrics were collected by tailing kurtosis service logs and parsing the `[METRIC] BLOCK` lines emitted by ethrex, comparing the same block numbers across both nodes.

**Results — Run 1** (260 blocks, Gloas-active)

| metric | `main` | `bal-opt` | diff | notes |
|---|---|---|---|---|
| gigagas/s | 1.078 | 1.097 | **+1.8%** | slight throughput gain |
| execution_ms | 19.3 ms | 18.8 ms | **−2.6%** | marginal exec speedup; warmer already finishes well ahead of exec so cache is warm either way |
| warmer_ms | 3.4 ms | 0.9 ms | **−73.5%** | core result: batched reads are ~4× faster to prewarm |
| warmer_early | +15.9 ms | +17.9 ms | **+12.6%** | bal-opt warmer finishes further ahead of exec start, leaving more slack |
| merkle_ms | ~0 ms | ~0 ms | flat | already zero-cost due to full concurrent overlap on both |
| merkle_overlap | ~58% | ~58% | flat | parallel state root equally effective on both branches |

**Results — Run 2** (440 loaded blocks, ~76 txs/block avg, Gloas-active)

| metric | `main` | `bal-opt` | diff | notes |
|---|---|---|---|---|
| gigagas/s | 1.719 | 1.727 | **+0.5%** | marginal throughput gain |
| execution_ms | 34.8 ms | 34.1 ms | **−2.0%** | ~0.7 ms faster execution |
| warmer_ms | 6.2 ms | 1.0 ms | **−83.9%** | batched reads ~6× faster to prewarm |
| warmer_early | +28.5 ms | +33.1 ms | **+16.1%** | warmer finishes 4.5 ms further ahead of exec |
| merkle_ms | ~0 ms | ~0.1 ms | flat | full concurrent overlap on both |
| merkle_overlap | 99.8% | 99.5% | flat | effectively 100% on both |

**Interpretation**

The warmer result is the clearest signal across both runs: batched reads consistently finish in ~1 ms vs ~3–6 ms on `main` — **74–84% faster prewarming**.

`warmer_early` measures how many milliseconds before block execution starts the prewarmer thread finishes. A positive value means the cache was fully warm by the time the EVM began; negative would mean the warmer was still running when execution started (cold-cache stalls). Both nodes finish comfortably ahead under these test loads, which explains why the execution time improvement is modest: with 16–33 ms of slack the EVM sees a fully warm cache either way. The practical significance of `warmer_early` improvement is in robustness under heavier load: as blocks get fuller or state working sets grow, the warmer races closer to execution start, and a 6× faster warmer means that margin erodes much more slowly before cache misses begin affecting execution latency.

The parallel state root (merkle_overlap ~100% on loaded blocks) is fully effective on both branches because `main` already had the parallel merkle pipeline from a prior PR. The `mrkl` time is effectively zero on loaded blocks for both nodes.

**Checklist**

Closes https://github.com/lambdaclass/ethrex/issues/6210
Closes https://github.com/lambdaclass/ethrex/issues/6211

### DIFF
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -174,8 +174,6 @@ pub enum RpcRequestWrapper {
     Multiple(Vec<RpcRequest>),
 }
 
-/// Shared context passed to all RPC request handlers.
-///
 /// Channel message type for the block executor worker thread.
 type BlockWorkerMessage = (
     oneshot::Sender<Result<(), ChainError>>,


edg-l:
### Fixed

**Stage C: comment explaining why all 16 threads must run unconditionally** (Claude, Codex)
Unlike Stage B which skips empty bins, Stage C must run all 16 shard workers to contribute their subtree to the merged root. Added a comment to prevent future incorrect \"optimization\".

**Re-insert rationale comment expanded** (Claude)
Added explanation that `collect_changes_since_last_hash` depends on this insert to include the node in its diff — needed for both updates and removals.

**EIP-161 empty-account deletion comment** (Claude)
Added comment connecting the `AccountState::default()` comparison to EIP-161 semantics.

---

### ❌ Dismissed

**\"Memory safety issue\" — `storage_roots[idx]` panics (Kimi)**
`storage_roots` is a `Vec<Option<H256>>`. Accessing `storage_roots[idx]` yields an `Option<H256>`, not an unwrapped value. There is no panic path here.

**Duplicate storage slot prefetching in `all_storage_slots` (Codex, Claude)**
`storage_reads` and `storage_changes` in `AccountChanges` are **disjoint by construction**. The BAL builder's `build()` method explicitly promotes reads-that-become-writes out of `storage_reads` into `storage_changes`. A slot cannot appear in both fields. `all_storage_slots()` does not produce duplicates.

**V1-V4 passing `None` for BAL (Kimi)**
Intentional by spec. BAL (`NewPayloadV5`) is an Amsterdam/Gloas feature. `NewPayloadV1-V4` predate the BAL spec and must not pass one. Passing `None` at those call sites is correct behavior.

**Files not in this PR (Kimi, Claude)**
The Makefile additions (`build-benchmark-images`, `localnet-benchmark`, `stop-localnet-benchmark`), `scripts/benchmark_compare.py`, and `fixtures/networks/bal-benchmark.yaml` are **not part of this PR**. They live on the [`bal-optimizations-kurtosis`](https://github.com/lambdaclass/ethrex/tree/bal-optimizations-kurtosis) branch. The bot reviewers appear to have read additional files outside the diff.

**`gloas` is a typo for `glados` (Kimi)**
\"Gloas\" is the devnet codename — not a typo. \"Amsterdam\" is the fork name.

---

### ⏳ Deferred / Tracked

**BAL size validation before selecting the BAL merkleization path (Codex, Claude, Kimi)**
Acknowledged. There is already a `TODO` comment at `payload.rs`. BAL size cap (EIP-7928) is being implemented in `bal-devnet-3` ([PR #6216](https://github.com/lambdaclass/ethrex/pull/6216)) which includes `validate_block_access_list_size`. Gating the BAL path behind that validation will be done there.

**BAL vs non-BAL state root equivalence regression test (Codex)**
Valid suggestion. Deferred — the ef-tests and hive suites provide coverage but a targeted unit test would be a good addition.

**`NUM_WORKERS = 16` CPU-adaptive for Stage B (Kimi, Claude)**
Stage C must be exactly 16 (hex nibble shards). Stage B could use `available_parallelism()`. Deferred as a follow-up.

**`Ordering::Acquire` → `Ordering::Relaxed` on queue counter (Claude)**
Noted, low priority. The semantics are stronger than necessary but not incorrect.

**`BlockWorkerMessage` newtype (Kimi, Claude), `BalStateWorkItem` doc comment, `EMPTY_KECCACK_HASH` typo (Claude)**
All pre-existing or cosmetic. Deferred.

crates/vm/backends/levm/mod.rs
```diff
@@ -375,6 +376,50 @@ impl LEVM {
...
+        // This warms the CachingDatabase account cache and the TrieLayerCache
+        // with state trie nodes, so Phase 2 storage reads benefit from cached lookups.
+        accounts.par_iter().for_each(|ac| {
+            let _ = store.get_account_state(ac.address);
+        });
+
+        // Phase 2: Prefetch storage slots and contract code in parallel.
+        // Each account's storage reads + writes need their pre-state loaded.
+        // Also prefetch code for all accounts with non-empty code_hash.
+        accounts.par_iter().for_each(|ac| {
```
iovoid, line 420: I suspect it might be better to iterate across storages instead of accounts, since the distribution of slots read per storage is not uniform.

edg-l: you are right, fixed

crates/common/types/block_access_list.rs
```diff
@@ -332,6 +332,15 @@ impl AccountChanges {
         self.code_changes.push(change);
     }
 
+    /// Returns an iterator over all unique storage slots that need prefetching
```
ElFantasma: The doc says "unique" but the iterator chains `storage_reads` and `storage_changes` without deduplication. If the same slot is both read and written (common for DEX pools — read balance, write new balance), it'll be prefetched twice in `warm_block_from_bal`.

Under typical load this is masked by the cache (second fetch hits CachingDatabase), but for hot accounts with many overlapping slots it adds unnecessary `H256::from_uint` conversions + `get_storage_value` calls.

Either deduplicate (e.g. `FxHashSet` or sort+dedup) or drop "unique" from the doc to match the actual behavior.

crates/networking/rpc/engine/payload.rs
```diff
@@ -326,11 +329,13 @@ impl RpcHandler for NewPayloadV5Request {
             )));
         }
 
+        let bal = self.payload.block_access_list.clone();
```
ElFantasma, line 332: nit: Since `self` is consumed by the handler, you could avoid cloning the BAL with `std::mem::take(&mut self.payload.block_access_list)` — the field becomes `None` before `&self.payload` is borrowed, which is fine since the rest of `handle_new_payload_v4` doesn't use the BAL from the payload struct. Same applies to `expected_blob_versioned_hashes` two lines below.

crates/common/types/block_access_list.rs
edg-l: Fixed, dropped "unique" from the doc.

crates/networking/rpc/engine/payload.rs
edg-l: The `handle` trait method signature is `async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr>` — it takes `&self`, not `self` or `&mut self`, so `std::mem::take` won't compile here. The `.clone()` is required.

crates/networking/rpc/rpc.rs
```diff
@@ -175,6 +176,13 @@ pub enum RpcRequestWrapper {
 
 /// Shared context passed to all RPC request handlers.
```
iovoid: This comment now documents the wrong type.

edg-l: fixed

PR #6231, opened 2026-02-19
fix(l1): limit memory consumed by fullsync

Arkenan:
**Motivation**

Fullsync uses batching, which behaves in unexpected ways with respect to trie layer caches: instead of a single layer representing a single block, it represents a single batch of 1024 blocks. That means that 128 layers are actually 128000 blocks, which may take more than 100GB of memory.

**Description**

In this PR we do two things:

1. Use only 4 layers for batched execution. That bounds the memory usage to about 4GB, which is reasonable.
2. Fix a small but important bug in trie layer. The previous implementation assumed that the parent was present (which is not the case for batches) and just advanced.

### DIFF
--- a/.github/actions/install-risc0/action.yml
+++ b/.github/actions/install-risc0/action.yml
@@ -7,9 +7,12 @@ runs:
     - name: Install RISC0
       shell: bash
       run: |
-        curl -L https://risczero.com/install | bash
-        ~/.risc0/bin/rzup install cargo-risczero 3.0.3
-        ~/.risc0/bin/rzup install risc0-groth16
-        ~/.risc0/bin/rzup install rust
-        ~/.risc0/bin/rzup install cpp
-        ~/.risc0/bin/rzup install r0vm 3.0.3
+        source "$GITHUB_WORKSPACE/.github/scripts/retry.sh"
+
+        retry curl -L https://risczero.com/install -o /tmp/risc0_install.sh
+        bash /tmp/risc0_install.sh
+        retry ~/.risc0/bin/rzup install cargo-risczero 3.0.3
+        retry ~/.risc0/bin/rzup install risc0-groth16
+        retry ~/.risc0/bin/rzup install rust
+        retry ~/.risc0/bin/rzup install cpp
+        retry ~/.risc0/bin/rzup install r0vm 3.0.3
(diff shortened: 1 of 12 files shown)


crates/storage/layering.rs
```diff
@@ -48,12 +69,16 @@ impl Default for TrieLayerCache {
             bloom: Self::create_filter(BLOOM_SIZE),
             last_id: 0,
             layers: Default::default(),
+            // TODO: this is coupled with DB_COMMIT_THRESHOLD in store.rs — unify them.
```
iovoid: Maybe open an issue for this? I don't think it even makes sense to carry this if we have a get_commitable_with_threshold method

crates/storage/layering.rs
```diff
@@ -97,19 +128,55 @@ impl TrieLayerCache {
...
+        self.get_commitable_with_threshold(state_root, self.commit_threshold)
+    }
+
+    /// Walks the layer chain starting from `state_root` toward older ancestors, counting
+    /// layers. When the count reaches `threshold`, returns `Some(state_root)` — the root
+    /// of the threshold-th layer, which is guaranteed to be a key in `self.layers`.
+    ///
+    /// The caller then passes the returned root to [`commit`](Self::commit), which removes
+    /// that layer **and all older ancestors below it**, flushing them to disk. This leaves
+    /// `threshold - 1` layers in the cache.
```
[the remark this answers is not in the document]
iovoid: This is true, but not the function's concern.

That said, I would remove the line about how many layers are left since it's false (reorgs could mean you have many more layers stored).

crates/storage/layering.rs
```diff
@@ -97,19 +128,55 @@ impl TrieLayerCache {
...
+    ///
+    /// Returns `None` if the chain has fewer than `threshold` layers (nothing to commit yet).
+    ///
+    /// # Why return the threshold-th layer (not the oldest)?
```
iovoid: Also you avoid problems from loops in the state (which can happen in an L2) where A->B->A.

Arkenan: Added this to the comment.

crates/storage/layering.rs
Arkenan: Opened an issue and linked it in the comment.

PR #6232, opened 2026-02-19
perf(l1): move witness generation off the critical path

ilitteri:
## Motivation

With `--precompute-witnesses` enabled, witness generation runs synchronously inside `add_block_pipeline`, adding 80-340ms to the "store" phase per block. On mainnet this makes store 87-90% of total block time, reducing throughput from ~1 Ggas/s to 0.06-0.16 Ggas/s.

The overhead comes from three phases that all run before `store_block` can proceed:
1. **Witness generation** (80-230ms): re-opens state/storage tries, re-accesses all accounts/keys via TrieLogger, re-applies account updates
2. **Serialization** (included in store): converts `ExecutionWitness` → `RpcExecutionWitness` (calls `encode_subtrie` per node) → JSON
3. **DB write**: writes serialized witness to RocksDB + cleans up old witnesses

## Description

**Commit 1 — `apply_account_updates_for_witness`** (store.rs)

The existing `apply_account_updates_from_trie_with_witness` calls `collect_changes_since_last_hash()` on every storage trie and the state trie, materializing a full `Vec<TrieNode>` of changed nodes into `AccountUpdatesList`. In the witness-only path (`generate_witness_from_account_updates`), this result is assigned to `_account_updates_list` and immediately discarded.

The new `apply_account_updates_for_witness` uses `hash_no_commit()` instead — it computes only the storage root hash without allocating the changed-nodes list. It also skips the final state trie hash computation entirely, returning only the `StorageTries` with their TrieLogger witnesses.

**Commit 2 — Background witness generation** (blockchain.rs)

Refactors `generate_witness_from_account_updates` into a static `generate_witness_impl` that takes `&Store` directly (instead of `&self`). In `add_block_pipeline`, witness generation + serialization + DB write now run on a named background thread (`witness_generator`). `store_block` proceeds immediately on the main thread.

Concurrent access is safe because:
- Witness gen reads from the parent state root; `store_block` writes new state — different trie layers
- `Store` is Clone (shares `Arc<InnerStore>`); the `TrieLayerCache` is behind `Arc<Mutex<>>` with short-lived locks
- RocksDB supports concurrent readers and writers

## Results (mainnet, ethrex-office-2)

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Store time | 169-479 ms (87-90%) | **1-3 ms (6-8%)** | ~100x |
| Throughput | 0.06-0.16 Ggas/s | **0.95-1.37 Ggas/s** | 6-20x |
| Block time | 193-552 ms | **13-46 ms** | 4-15x |
| Bottleneck | store | **exec (correct)** | — |

Witness generation still completes in the background (72-232ms) and writes are visible via the `[witness-bg]` log line.

## Test plan

- [x] Deployed on mainnet node (ethrex-office-2) with `--precompute-witnesses`
- [x] Verified store times dropped from 169-479ms to 1-3ms across 15+ consecutive blocks
- [x] Verified `[witness-bg]` logs confirm witness generation completes successfully in background
- [x] Verified no errors or warnings from background thread
- [ ] CI passes

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -1274,9 +1274,26 @@ impl Blockchain {
         parent_header: BlockHeader,
         logger: &DatabaseLogger,
     ) -> Result<ExecutionWitness, ChainError> {
+        Self::generate_witness_impl(
+            &self.storage,
+            account_updates,
+            block,
+            parent_header,
+            logger,
+        )
+    }
+
+    /// Static implementation of witness generation that takes `Store` directly.
+    /// This allows running witness gen on a background thread without needing `&self`.
+    fn generate_witness_impl(
+        storage: &Store,
+        account_updates: Vec<AccountUpdate>,
+        block: &Block,
+        parent_header: BlockHeader,
+        logger: &DatabaseLogger,
+    ) -> Result<ExecutionWitness, ChainError> {
         // Get state at previous block
-        let trie = self
-            .storage
+        let trie = storage
             .state_trie(parent_header.hash())
             .map_err(|_| ChainError::ParentStateNotFound)?
             .ok_or(ChainError::ParentStateNotFound)?;
@@ -1285,7 +1302,6 @@ impl Blockchain {
         let (trie_witness, trie) = TrieLogger::open_trie(trie);
 
         let mut touched_account_storage_slots = BTreeMap::new();
-        // This will become the state trie + storage trie
         let mut used_trie_nodes = Vec::new();
 
         // Store the root node in case the block is empty and the witness does not record any nodes
@@ -1341,7 +1357,7 @@ impl Blockchain {
             // Get storage trie at before updates
             if !acc_keys.is_empty()
                 && let Ok(Some(storage_trie)) =
-                    self.storage.storage_trie(parent_header.hash(), *account)
+                    storage.storage_trie(parent_header.hash(), *account)
             {
                 let (storage_trie_witness, storage_trie) = TrieLogger::open_trie(storage_trie);
                 // Access all the keys
@@ -1365,8 +1381,7 @@ impl Blockchain {
             })?
             .iter()
         {
-            let code = self
-                .storage
+            let code = storage
                 .get_account_code(*code_hash)
                 .map_err(|_e| {
                     ChainError::WitnessGeneration("Failed to get account code".to_string())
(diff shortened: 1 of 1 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -1625,18 +1636,42 @@ impl Blockchain {
             block.body.transactions.len(),
         );
 
+        // Spawn witness generation on a background thread so store_block proceeds immediately
         if let Some(logger) = logger
             && let Some(account_updates) = accumulated_updates
         {
             let block_hash = block.hash();
-            let witness = self.generate_witness_from_account_updates(
```
iovoid, line 1889: generate_witness_from_account_updates  is no longer called, it can be removed

crates/blockchain/blockchain.rs
```diff
@@ -1625,18 +1636,42 @@ impl Blockchain {
...
-                &block,
-                parent_header,
-                &logger,
-            )?;
-            self.storage
-                .store_witness(block_hash, block_number, witness)?;
+            let block_clone = block.clone();
+            let parent_header_clone = parent_header.clone();
+            let storage = self.storage.clone();
+            std::thread::Builder::new()
```
iovoid, line 1647: Maybe use a single background thread? On big full syncs this could spawn threads faster than they can finish.

crates/blockchain/blockchain.rs
```diff
@@ -1625,18 +1636,42 @@ impl Blockchain {
...
+                            }
+                            info!(
+                                "  [witness-bg] block {block_number}: gen={gen_ms} ms, store={} ms",
+                                store_start.elapsed().as_millis()
+                            );
+                        }
+                        Err(e) => warn!("Background witness generation failed for block {block_number}: {e}"),
+                    }
+                })
+                .ok();
```
ElFantasma, line 1674: `.ok()` silently discards both the `JoinHandle` and any spawn error. If `thread::Builder::spawn()` fails (OS thread limit), witness generation stops silently with no log — hard to diagnose in production.

The dropped `JoinHandle` (detached thread) is intentional and fine, but the error path should warn:
```rust
.spawn(move || { ... })
.map_err(|e| warn!("Failed to spawn witness_generator thread: {e}"))
.ok();
```

crates/storage/store.rs
```diff
@@ -1824,6 +1824,73 @@ impl Store {
         Ok((storage_tries, account_updates_list))
     }
 
+    /// Like `apply_account_updates_from_trie_with_witness`, but optimized for the witness-only path.
+    /// Uses `hash_no_commit()` instead of `collect_changes_since_last_hash()` to avoid
+    /// materializing the changed-nodes list (which the caller discards anyway).
+    pub fn apply_account_updates_for_witness(
```
ElFantasma, line 1830: nit: This function duplicates ~60 lines from `apply_account_updates_from_trie_with_witness` with two differences: `hash_no_commit()` instead of `collect_changes_since_last_hash()` for storage tries, and no final state trie hash. Any correctness fix in the shared logic (e.g. `removed_storage` handling, storage slot ordering) will need to land in both places.

Would be cleaner to extract a shared inner function or add a `witness_only: bool` flag — but if the maintenance burden seems acceptable for now, at minimum add a comment pointing to the twin function.

PR #6233, opened 2026-02-20
feat(levm): parallel block execution via BAL 

### LINKED ISSUE #6209: [BAL] Parallel block execution

edg-l:
Closes #6209

## Summary

Implements BAL-based parallel transaction execution for Amsterdam+ blocks using EIP-7928 Block Access Lists.

### Approach: BAL State Seeding

Each transaction runs independently on its own database, pre-seeded with BAL-derived intermediate state (same approach as geth). No conflict detection or grouping needed — the BAL provides the complete state dependency graph.

**Pipeline (3 concurrent threads):**
1. **Warmer** — prefetches all account states, storage slots, and contract codes listed in the BAL (parallel batch fetch)
2. **Executor** — runs all txs in parallel via rayon, each with its own `GeneralizedDatabase` seeded from BAL intermediate values
3. **Merkleizer** — computes state trie from `bal_to_account_updates` (BAL → AccountUpdates, no execution needed)

**Key functions (`crates/vm/backends/levm/mod.rs`):**
- `execute_block_parallel` — orchestrates the parallel path: sends BAL-derived AccountUpdates to merkleizer, then executes all txs in parallel
- `bal_to_account_updates(bal, store)` — converts BAL final values into `Vec<AccountUpdate>` for the merkleizer (last entry per field = post-block state)
- `seed_db_from_bal(db, bal, max_idx)` — pre-seeds a per-tx DB with cumulative BAL state through index `max_idx` (system calls + previous txs)
- `warm_block_from_bal(bal, store)` — 3-phase prefetch: accounts → storage slots → contract codes

**BAL indexing:** 0 = system calls, 1 = tx 0, 2 = tx 1, ..., N+1 = withdrawals. For tx at index `i`, `seed_db_from_bal` applies all changes with `block_access_index <= i`.

The parallel path is only triggered when `header_bal` is `Some` (Amsterdam+ blocks via `engine_newPayloadV4`). All other callers pass `None` and use the existing sequential loop unchanged.

## EF Test Verification (Two-Pass Parallel Check)

Amsterdam EF tests now exercise the parallel execution path as a correctness check. After the normal sequential run succeeds, a second two-pass run is performed:

1. **Pass 1 (sequential):** Re-executes all blocks on a fresh blockchain via `add_block_pipeline_returning_bal(block, None)`, collecting the produced BAL for each block.
2. **Pass 2 (parallel):** Re-executes all blocks on another fresh blockchain via `add_block_pipeline(block, Some(&bal))`, using the BAL from pass 1 to drive the parallel execution path.
3. **Post-state verification:** The final state from pass 2 is checked against the expected post-state from the test fixture.

This ensures that the parallel execution path produces identical results to the sequential path across the entire Amsterdam EF test suite. Non-Amsterdam tests are unaffected. All Amsterdam EF tests pass both sequential and parallel execution.

## Test plan

- [x] 6 unit tests for `bal_to_account_updates` — all pass (`cargo test -p ethrex-vm bal_tests`)
- [x] EF blockchain tests — sequential path (existing CI)
- [x] EF blockchain tests — parallel path via two-pass BAL verification (Amsterdam tests)
- [x] Hive regression on `devnets/bal/2` fixture suite with parallel path enabled
- [x] Kurtosis devnet benchmark (see results below)

## Notes

- No new crate dependencies (rayon and FxHashMap already used)
- **Per-tx BAL validation:** After each transaction executes in parallel, its actual state mutations are validated against the BAL claims. `GeneralizedDatabase::compute_tx_diff()` computes the diff between initial and final account state for the tx, then `BlockAccessList::validate_tx_diff(tx_idx, &diff)` checks that balances, nonces, storage values, and code changes match exactly what the BAL declares for that tx index. Blocks with mismatched state mutations are rejected. This matches geth's per-tx validation approach.
- BAL recording is disabled in the parallel path — the header-embedded BAL is trusted (validated by `validate_block_access_list_hash`) and used directly for both state root computation and per-tx state seeding. This mirrors geth's approach: `BALStateTransition.IntermediateRoot()` computes the state root entirely from BAL diffs (via `readAccountDiff` + `ModifiedAccounts`) without re-executing transactions, and `BALReader.getStateObject` / `initObjFromDiff` seeds per-tx state from BAL intermediate values. No BAL re-recording happens during validation in either implementation.
- Per-tx DBs skip `initial_accounts_state` tracking (state transitions come from BAL, not from diffing)

## Benchmark Results

Benchmarked on a Kurtosis devnet (`bal-devnet-2-ethrex.yaml`) comparing parallel execution (via BAL) vs sequential (`main`), both running on identical hardware. Nodes are in consensus (matching block hashes).

### Spamoor Setup (mainnet-like workload, ~700 tx/s target)

| Scenario | tx/s | Share | Mainnet Analog |
|---|---|---|---|
| `erc20tx` | 280 | 40% | USDT/USDC token transfers — hot storage (balances, allowances) |
| `uniswap-swaps` | 175 | 25% | DEX swaps — complex contract calls, hot pool state |
| `eoatx` | 105 | 15% | Simple ETH transfers |
| `storagespam` | 70 | 10% | Storage-heavy contract interactions |
| `erc721tx` | 35 | 5% | NFT mints/transfers |
| `blobs` | 35 | 5% | L2 sequencer-like blob transactions |

Blocks average ~231 txs and ~60 Mgas each.

### Results (43 non-empty blocks sampled)

| | Parallel (`bal-parallel-exec`) | Sequential (`main`) |
|---|---|---|
| **Avg Ggas/s** | **2.00** | 1.20 |
| **Avg exec time** | **31 ms** | 51 ms |
| **Peak Ggas/s** | **2.61** | 1.40 |

**~67% faster with BAL parallel execution** on realistic mainnet-like workloads.

### DIFF
--- a/crates/common/types/block_access_list.rs
+++ b/crates/common/types/block_access_list.rs
@@ -431,6 +431,89 @@ impl BlockAccessList {
         &self.inner
     }
 
+    /// Validates that the BAL has canonical ordering per EIP-7928.
+    /// - Accounts must be in strictly ascending order by address.
+    /// - Within each account: storage_changes by slot, storage_reads by slot value,
+    ///   slot_changes/balance_changes/nonce_changes/code_changes by block_access_index.
+    ///
+    /// Returns an error string describing the first violation found.
+    pub fn validate_ordering(&self) -> Result<(), String> {
+        let mut prev_addr = None;
+        for account in &self.inner {
+            if let Some(prev) = prev_addr
+                && prev >= account.address
+            {
+                return Err(format!(
+                    "Block access list accounts not in strictly ascending order: \
+                     {:#x} >= {:#x}",
+                    prev, account.address
+                ));
+            }
+            prev_addr = Some(account.address);
+
+            for window in account.storage_changes.windows(2) {
+                if window[0].slot >= window[1].slot {
+                    return Err(format!(
+                        "Block access list storage_changes not in strictly ascending order \
+                         for account {:#x}: {:#x} >= {:#x}",
+                        account.address, window[0].slot, window[1].slot
+                    ));
+                }
+            }
+            for slot_change in &account.storage_changes {
+                for window in slot_change.slot_changes.windows(2) {
+                    if window[0].block_access_index >= window[1].block_access_index {
+                        return Err(format!(
+                            "Block access list slot_changes not in strictly ascending order \
+                             for account {:#x} slot {:#x}: {} >= {}",
+                            account.address,
+                            slot_change.slot,
+                            window[0].block_access_index,
+                            window[1].block_access_index
+                        ));
+                    }
+                }
+            }
+            for window in account.storage_reads.windows(2) {
+                if window[0] >= window[1] {
+                    return Err(format!(
+                        "Block access list storage_reads not in strictly ascending order \
+                         for account {:#x}: {:#x} >= {:#x}",
+                        account.address, window[0], window[1]
+                    ));
+                }
+            }
+            for window in account.balance_changes.windows(2) {
+                if window[0].block_access_index >= window[1].block_access_index {
+                    return Err(format!(
+                        "Block access list balance_changes not in strictly ascending order \
+                         for account {:#x}: {} >= {}",
+                        account.address, window[0].block_access_index, window[1].block_access_index
+                    ));
+                }
+            }
+            for window in account.nonce_changes.windows(2) {
+                if window[0].block_access_index >= window[1].block_access_index {
+                    return Err(format!(
+                        "Block access list nonce_changes not in strictly ascending order \
+                         for account {:#x}: {} >= {}",
+                        account.address, window[0].block_access_index, window[1].block_access_index
+                    ));
+                }
+            }
+            for window in account.code_changes.windows(2) {
+                if window[0].block_access_index >= window[1].block_access_index {
+                    return Err(format!(
+                        "Block access list code_changes not in strictly ascending order \
+                         for account {:#x}: {} >= {}",
+                        account.address, window[0].block_access_index, window[1].block_access_index
+                    ));
+                }
+            }
+        }
+        Ok(())
+    }
+
     /// Computes the hash of the block access list (sorts accounts by address per EIP-7928).
     /// Use this when hashing a BAL constructed locally from execution.
     pub fn compute_hash(&self) -> H256 {
--- a/crates/l2/sdk/src/calldata.rs
+++ b/crates/l2/sdk/src/calldata.rs
@@ -1,6 +1,6 @@
 use ethrex_common::Bytes;
 use ethrex_common::H256;
-use ethrex_common::utils::keccak;
+use ethrex_common::utils::{decode_hex, keccak};
 use ethrex_common::{Address, H32, U256};
 use ethrex_l2_common::calldata::Value;
 use ethrex_rpc::clients::EthClientError;
@@ -521,7 +521,7 @@ fn copy_into(
 
 #[allow(clippy::indexing_slicing)]
 pub fn from_hex_string_to_h256_array(hex_string: &str) -> Result<Vec<H256>, EthClientError> {
-    let bytes = hex::decode(hex_string.strip_prefix("0x").unwrap_or(hex_string))
+    let bytes = decode_hex(hex_string)
         .map_err(|_| EthClientError::Custom("Invalid hex string".to_owned()))?;
 
     // The ABI encoding for dynamic arrays is:
--- a/crates/l2/sdk/src/sdk.rs
+++ b/crates/l2/sdk/src/sdk.rs
@@ -1331,15 +1331,8 @@ async fn _generic_call(
         .get(..4)
         .ok_or(EthClientError::Custom("Failed to get selector.".to_owned()))?
         .to_vec();
-
-    let mut calldata = Vec::new();
-    calldata.extend_from_slice(&selector);
-
-    let leading_zeros = 32 - ((calldata.len() - 4) % 32);
-    calldata.extend(vec![0; leading_zeros]);
-
     let hex_string = client
-        .call(contract_address, calldata.into(), Overrides::default())
+        .call(contract_address, selector.into(), Overrides::default())
         .await?;
 
     Ok(hex_string)
--- a/crates/networking/p2p/discv5/messages.rs
+++ b/crates/networking/p2p/discv5/messages.rs
@@ -144,15 +144,20 @@ impl PacketHeader {
         let nonce = static_header[9..21].try_into()?;
         let authdata_size = u16::from_be_bytes(static_header[21..23].try_into()?) as usize;
         let authdata_end = STATIC_HEADER_END + authdata_size;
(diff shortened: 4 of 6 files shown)


crates/vm/backends/levm/mod.rs
```diff
@@ -311,6 +367,605 @@ impl LEVM {
...
+                None
+            };
+
+            let update = AccountUpdate {
+                address: addr,
+                removed,
+                info,
+                code,
+                added_storage,
+                removed_storage: false, // EIP-6780: SELFDESTRUCT only same-tx
```
ElFantasma: Hardcoding `removed_storage: false` is correct for the common case, but misses the `DestroyedModified` scenario: if a contract at address X self-destructs during its creation tx (EIP-6780 allows this) and then gets modified in the same tx, the sequential path sets `removed_storage = true` so the merkleizer clears old storage slots from a prior contract at that address. The BAL path doesn't detect this because the BAL records post-state values, not the `Destroyed → Modified` status transition.

Does EIP-7928's BAL record individual storage slot deletions from SELFDESTRUCT (i.e., each old slot → 0)? If yes, `added_storage` would contain the zeroed slots and this is fine. If not, old slots from a prior contract at the same address would persist in the trie.

Extremely rare edge case (same-address re-creation + SELFDESTRUCT in creation tx), and the state root check would catch the discrepancy at block level — but worth confirming the BAL spec covers this.

crates/vm/backends/levm/mod.rs
```diff
@@ -311,6 +367,605 @@ impl LEVM {
...
+
+        // PART B: For each modified account in execution state,
+        //         verify no unexpected mutations (changes not claimed by BAL).
+        for (addr, account) in current_state {
+            if account.is_unmodified() {
+                continue;
+            }
+
+            let Some(&bal_acct_idx) = index.addr_to_idx.get(addr) else {
+                // Account not in BAL. Modified status can come from read-only
```
ElFantasma, line 877: nit: This skip makes sense (warm `get_account_mut` can set `Modified` without real mutations), but it means the per-tx validation has a blind spot: an execution bug that writes to an account the BAL doesn't mention would be silently ignored here. Consider at least logging a debug trace when skipping a `Modified` account not in BAL, so anomalies are visible during testing.

crates/vm/backends/levm/mod.rs
```diff
@@ -311,6 +367,605 @@ impl LEVM {
...
+            if !has_exact_change_code(&acct.code_changes, bal_idx) {
+                let seeded_pos = acct
+                    .code_changes
+                    .partition_point(|c| c.block_access_index <= seed_idx);
+                if seeded_pos > 0 {
+                    let seeded_code = &acct.code_changes[seeded_pos - 1].new_code;
+                    let seeded_hash = if seeded_code.is_empty() {
+                        *EMPTY_KECCACK_HASH
+                    } else {
+                        Code::from_bytecode(seeded_code.clone()).hash
```
ElFantasma: nit: `Code::from_bytecode(seeded_code.clone()).hash` recomputes keccak256 on each validation call. For frequently-deployed code (factory patterns), this could add up across many txs in a block. Consider caching the seeded code hash during `seed_db_from_bal` (which already computes it at line 551) and passing it through, or at minimum pre-hashing once per validation call instead of reconstructing the full `Code` object.

crates/vm/levm/Cargo.toml
```diff
@@ -43,6 +43,7 @@ ark-ff = { version = "0.5.0", features = ["asm"] }
 strum = { version = "0.27.1", features = ["derive"] }
 k256.workspace = true
 rustc-hash.workspace = true
+rayon.workspace = true
```
ElFantasma, line 46: `rayon` is added unconditionally to `ethrex-levm`, which is compiled into all zkVM guest programs (openvm, risc0, sp1, zisk — visible in their Cargo.lock diffs). While rayon falls back to single-threaded when parallelism isn't available, it still adds binary size overhead. For zkVM targets where binary size directly impacts proving time, consider gating this behind a feature flag (e.g., `parallel`) and only enabling it in the `ethrex-vm` backend crate.

crates/vm/backends/levm/mod.rs
edg-l: Good catch on the edge case. After checking the EIP-7928 spec:

> "Accounts destroyed within a transaction MUST be included in AccountChanges without nonce or code changes. However, if the account had a positive balance pre-transaction, the balance change to zero MUST be recorded."

And for storage, writes include "value changes and zeroing operations." So in the destroy+recreate scenario, the individual slot deletions (old slot → 0) appear in `storage_changes`, meaning `added_storage` already contains those zeroed entries and the trie update is correct without `removed_storage = true`.

Added an expanded comment explaining this reasoning. EF tests pass which confirms the BAL covers this case through explicit slot zeroing.

crates/vm/backends/levm/mod.rs
edg-l: Makes sense, added a `tracing::debug!` log when skipping a Modified account not present in BAL.

crates/vm/backends/levm/mod.rs
edg-l: Good point. Replaced `Code::from_bytecode(seeded_code.clone()).hash` with a direct `ethrex_common::utils::keccak(seeded_code)` — avoids constructing the full `Code` object (which also computes jump targets) just to get the hash.

crates/vm/levm/Cargo.toml
edg-l: Rayon falls back to single-threaded execution when parallelism isn't available, and the binary size overhead is minimal compared to the crypto dependencies already in the crate. Feature-gating would add conditional compilation complexity across the codebase for a marginal size saving. Not going to gate it for now — if zkVM proving time becomes a bottleneck we can revisit.

ElFantasma (approved):

PR #6235, opened 2026-02-20
feat(l1,l2): add environment variables to more CLI options

iovoid:
**Motivation**

Many CLI options lack env attributes. This makes container and deployment configuration harder, since environment variables are the standard way to configure services in Docker, Kubernetes, etc.

**Description**

This PR adds a corresponding environment variable to all ethrex options, except for the following:

  - --force, one-shot and not meant to remain set
  - tooling (repl, monitor, etc) options, since those are manually invoked by the user

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -728,7 +728,7 @@ pub async fn import_blocks(
             } else {
                 // We need to have the state of the latest 128 blocks
                 blockchain
-                .add_block_pipeline(block)
+                .add_block_pipeline(block, None)
                 .inspect_err(|err| match err {
                     // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
                     ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
@@ -836,7 +836,7 @@ pub async fn import_blocks_bench(
                 .map_err(InvalidBlockError::InvalidBody)?;
 
             blockchain
-                .add_block_pipeline(block)
+                .add_block_pipeline(block, None)
                 .inspect_err(|err| match err {
                     // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
                     ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
--- a/cmd/ethrex/l2/options.rs
+++ b/cmd/ethrex/l2/options.rs
@@ -216,6 +216,7 @@ impl TryFrom<SequencerOptions> for SequencerConfig {
                     .proof_coordinator_tdx_private_key,
                 qpl_tool_path: opts.proof_coordinator_opts.proof_coordinator_qpl_tool_path,
                 validium: opts.validium,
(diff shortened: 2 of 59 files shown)


cmd/ethrex/cli.rs
```diff
@@ -197,95 +203,107 @@ pub struct Options {
...
     )]
     pub authrpc_port: String,
     #[arg(
         long = "authrpc.jwtsecret",
         default_value = "jwt.hex",
         value_name = "JWTSECRET_PATH",
         help = "Receives the jwt secret used for authenticated rpc requests.",
-        help_heading = "RPC options"
+        help_heading = "RPC options",
+        env = "ETHREX_AUTHRPC_JWTSECRET"
```
ElFantasma: nit: `ETHREX_AUTHRPC_JWTSECRET` sounds like it should hold the secret hex value, but it actually takes a file path (as `value_name = "JWTSECRET_PATH"` hints). CLI users see the `value_name`, but operators configuring via env vars won't. Consider `ETHREX_AUTHRPC_JWTSECRET_PATH` to avoid a confusing failure when someone does `ETHREX_AUTHRPC_JWTSECRET=0x<hex>`.

cmd/ethrex/l2/options.rs
```diff
@@ -36,7 +36,8 @@ pub struct Options {
         long = "sponsorable-addresses",
         value_name = "SPONSORABLE_ADDRESSES_PATH",
         help = "Path to a file containing addresses of contracts to which ethrex_SendTransaction should sponsor txs",
-        help_heading = "L2 options"
+        help_heading = "L2 options",
+        env = "ETHREX_SPONSORABLE_ADDRESSES"
```
ElFantasma: nit: Same naming ambiguity as `ETHREX_AUTHRPC_JWTSECRET` — `ETHREX_SPONSORABLE_ADDRESSES` sounds like a list of addresses, but it's a file path. `ETHREX_SPONSORABLE_ADDRESSES_PATH` would match the `value_name = "SPONSORABLE_ADDRESSES_PATH"` already in the help text.

ElFantasma (approved):
Approved with two nits

cmd/ethrex/cli.rs
iovoid: I agree it's clearer that way, improved in 409ac47.

cmd/ethrex/l2/options.rs
iovoid: Done in 409ac47.

avilagaston9 (approved):
LGTM! Clean and well-scoped PR. Two minor optional suggestions:

1. **`ETHREX_DISCOVERY_PORT`** — All other P2P options use the `ETHREX_P2P_` prefix (`ETHREX_P2P_PORT`, `ETHREX_P2P_ADDR`, `ETHREX_P2P_DISABLED`, etc.). Consider `ETHREX_P2P_DISCOVERY_PORT` for consistency, since discovery is a P2P concern.

2. **`ETHREX_MEMPOOL_MAXSIZE`** — Other compound env vars use underscores between words (`ETHREX_P2P_TARGET_PEERS`, `ETHREX_BUILDER_GAS_LIMIT`). `ETHREX_MEMPOOL_MAX_SIZE` would be more consistent.

Both are very minor — feel free to take or leave.

PR #6237, opened 2026-02-20
feat(l1): add node replay tooling and checkpoint-safe replay execution

ilitteri:
**Motivation**

`ethrex` needed an agent-oriented replay workflow that can re-execute canonical blocks using real node execution logic without mutating the live node database. The previous tooling gap made repeatable benchmarking/debug runs hard to automate safely.

**Description**

This PR introduces a new `tooling/node_replay` workspace tool plus storage-side read-only safety improvements required by that tool.

Main changes:

- Add `tooling/node_replay` as a new workspace member.
- Implement an agent-friendly CLI with JSON envelopes and typed errors:
  - `checkpoint create|list`
  - `plan`
  - `run`
  - `status`
  - `resume`
  - `cancel`
  - `verify`
  - `report`
- Implement replay pipeline primitives:
  - checkpoint creation from live RocksDB datadir
  - canonical hash pinning for planned ranges
  - isolated per-run DB materialization under workspace
  - lock ownership + heartbeat for run concurrency control
  - persisted run state/events/summary artifacts
- Add read-only storage support used by sidecar replay operations:
  - `RocksDBBackend::open_read_only`
  - write guards for read-only backends (`begin_write`, `clear_table`)
  - bounded read-only RocksDB FD settings to reduce EMFILE risk
  - `Store::new_read_only(...)`
  - `Store::load_chain_config(...)`
  - read-only schema validation path that does not initialize metadata
- Add comprehensive `tooling/node_replay/README.md`:
  - architecture diagrams
  - operational flow and safety model
  - CLI contracts
  - examples for smoke runs, retries, multiple checkpoints, resume/cancel
- Add safety/integration-style tests in `tooling/node_replay/tests/safety.rs`.

User impact:

- Operators/agents can plan and replay block ranges deterministically from checkpoints.
- Live node datadir remains read-only from replay tool perspective.
- Checkpoints are preserved for re-replay while execution happens in isolated per-run copies.

**Validation**

- `cargo check -p node-replay`
- `cargo test -p node-replay`

**Checklist**

Reason unchecked: no store schema format change was introduced.

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -380,16 +380,14 @@ impl LEVM {
                 // Execute transactions sequentially within sender group
                 // This ensures nonce and balance changes from tx[N] are visible to tx[N+1]
                 for tx in txs {
-                    let Ok(mut env) = Self::setup_env(tx, sender, &block.header, &group_db, vm_type) else {
-                        continue;
-                    };
-                    env.skip_balance_check = true;
-                    let Ok(mut vm) = VM::new(env, &mut group_db, tx, LevmCallTracer::disabled(), vm_type) else {
-                        continue;
-                    };
-                    std::mem::swap(&mut vm.stack_pool, stack_pool);
-                    let _ = vm.execute();
-                    std::mem::swap(&mut vm.stack_pool, stack_pool);
+                    let _ = Self::execute_tx_in_block(
+                        tx,
+                        sender,
+                        &block.header,
+                        &mut group_db,
+                        vm_type,
+                        stack_pool,
+                    );
                 }
             },
         );
@@ -467,7 +465,6 @@ impl LEVM {
             difficulty: block_header.difficulty,
             is_privileged: matches!(tx, Transaction::PrivilegedL2Transaction(_)),
             fee_token: tx.fee_token(),
-            skip_balance_check: false,
         };
 
         Ok(env)
@@ -991,7 +988,6 @@ fn env_from_generic(
         difficulty: header.difficulty,
         is_privileged: false,
         fee_token: tx.fee_token,
-        skip_balance_check: false,
     })
 }
 
--- a/crates/vm/levm/src/environment.rs
+++ b/crates/vm/levm/src/environment.rs
@@ -41,8 +41,6 @@ pub struct Environment {
     pub block_gas_limit: u64,
     pub is_privileged: bool,
     pub fee_token: Option<Address>,
-    /// When true, skip sender balance validation (used during prewarming).
-    pub skip_balance_check: bool,
 }
 
 /// This struct holds special configuration variables specific to the
--- a/crates/vm/levm/src/hooks/default_hook.rs
+++ b/crates/vm/levm/src/hooks/default_hook.rs
@@ -488,9 +488,6 @@ pub fn validate_gas_allowance(vm: &mut VM<'_>) -> Result<(), TxValidationError> 
 }
 
 pub fn validate_sender_balance(vm: &mut VM<'_>, sender_balance: U256) -> Result<(), VMError> {
-    if vm.env.skip_balance_check {
-        return Ok(());
-    }
     // Up front cost is the maximum amount of wei that a user is willing to pay for. Gaslimit * gasprice + value + blob_gas_cost
     let value = vm.current_call_frame.msg_value;
 


tooling/node_replay/src/commands.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    // Planned/Paused, or Running with dead executor: write Canceled directly.
+    status.state = RunState::Canceled;
+    status.updated_at = Utc::now();
+    workspace.write_run_status(&status)?;
+
+    workspace.append_event(&ReplayEvent::new(
+        run_id.to_string(),
+        EventName::RunFailed,
```
ElFantasma, line 51: Both this path and `check_cancel_flag` in `runner.rs` emit `EventName::RunFailed` when writing a `Canceled` state. The state field in `status.json` is correct, but the event log will show `run_failed` for every cancellation — agents consuming `events.ndjson` can't distinguish a canceled run from a failed one by event name. `EventName` is missing a `RunCanceled` variant; add it and use it here and in `check_cancel_flag`.

tooling/node_replay/src/workspace.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+        let data = fs::read_to_string(path)?;
+        Ok(serde_json::from_str(&data)?)
+    }
+
+    // --- Run status read/write ---
+
+    pub fn write_run_status(&self, status: &RunStatus) -> Result<(), ReplayError> {
+        let json = serde_json::to_string_pretty(status)?;
+        fs::write(self.run_status_path(&status.run_id), json)?;
```
ElFantasma, line 185: `fs::write` is not atomic: on a crash between truncation and the full write, `status.json` ends up with partial JSON. On resume, `read_run_status` will return a JSON parse error and the run will be unrecoverable without manual intervention. The lock heartbeat code already uses temp-file + `fs::rename` for atomic updates — apply the same pattern here (and to `write_run_summary`). `fs::rename` on the same filesystem is atomic on POSIX.

crates/storage/store.rs
```diff
@@ -1428,6 +1439,31 @@ impl Store {
...
+            validate_store_schema_version_read_only(&db_path)?;
+        }
+
+        match engine_type {
+            #[cfg(feature = "rocksdb")]
+            EngineType::RocksDB => {
+                let backend = Arc::new(RocksDBBackend::open_read_only(path)?);
+                Self::from_backend(backend, db_path, DB_COMMIT_THRESHOLD)
+            }
+            EngineType::InMemory => {
```
ElFantasma, line 1460: `InMemoryBackend::open()` creates a **fresh, empty** in-memory store — not a read-only view of any existing data. A caller using `Store::new_read_only(path, EngineType::InMemory)` expecting to read state at `path` will silently get an empty store instead. This should either return `Err(StoreError::Custom("InMemory does not support read-only open".into()))` or at minimum have a `// SAFETY:` comment explaining that InMemory is only for testing and callers are expected to populate state manually.

tooling/node_replay/src/types.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub avg_block_ms: f64,
+    pub mismatch_count: u64,
+    pub final_head_number: Option<u64>,
+    /// Hex-encoded final head hash
+    pub final_head_hash: Option<String>,
+}
+
+impl RunSummary {
+    /// Creates a run summary from a completed run's status and manifest.
+    pub fn from_status(status: &RunStatus, manifest: &RunManifest, duration_ms: u64) -> Self {
```
ElFantasma, line 126: nit: `RunSummary::from_status` is never called — `runner.rs` constructs `RunSummary` inline. Either use this helper or remove it to avoid the divergence.

tooling/node_replay/src/types.rs
```diff
@@ -0,0 +1,247 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum EventName {
+    RunPlanned,
+    RunStarted,
+    BlockStarted,
+    BlockExecuted,
+    BlockVerified,
+    RunPaused,
+    RunResumed,
+    RunFailed,
```
iovoid, line 163: Why not differentiate with Canceled?

tooling/node_replay/src/commands.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// For Running state with a stale/missing lock: the executor is dead. Write
+/// Canceled directly and clean up the stale lock.
+///
+/// For non-Running states (Planned/Paused): writes Canceled directly.
+pub fn cancel_run(workspace: &Workspace, run_id: &str) -> Result<RunStatus, ReplayError> {
+    let mut status = workspace.read_run_status(run_id)?;
+
+    // Validate transition
+    status.state.transition_to(&RunState::Canceled)?;
```
iovoid, line 28: Throwing an error isn't very user-friendly. The user might trigger this by trying to cancel a failed run.

PR #6243, opened 2026-02-23
perf(l1): add precompile result cache shared between warmer and executor

ilitteri:
## Motivation

During block execution, the warmer thread pre-executes precompiles to warm their results, but those results were discarded — the executor thread would then recompute the same precompile calls from scratch. This is wasted work, especially for expensive precompiles like ecrecover and modexp.

## Description

Add a shared `PrecompileCache` (`Arc<RwLock<FxHashMap<(Address, Bytes), (Bytes, u64)>>>`) between the warmer and executor threads so precompile results computed during warming are reused during execution.

- The cache is keyed by `(precompile_address, calldata)` and stores `(output, gas_cost)`
- The identity precompile (0x04) is excluded from caching since it's cheaper to just copy the input
- The cache is created per-block in `add_block` and passed through to LEVM's `GeneralizedDatabase`
- On precompile execution, the cache is checked first; on miss, the result is computed and inserted

### Files changed
- `crates/blockchain/blockchain.rs` — create shared `Arc<PrecompileCache>` and pass to both warmer and executor
- `crates/vm/backends/levm/mod.rs` — accept `precompile_cache` in `warm_block()`
- `crates/vm/levm/src/db/gen_db.rs` — add `precompile_cache` field to `GeneralizedDatabase`
- `crates/vm/levm/src/precompiles.rs` — implement `PrecompileCache` struct with cache lookup/insert in `execute_precompile()`
- `crates/vm/levm/src/vm.rs` — pass cache to precompile execution
- `crates/vm/levm/src/opcode_handlers/system.rs` — pass cache to precompile execution in CALL handler
- `crates/vm/lib.rs` — re-export `PrecompileCache`

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -83,7 +83,6 @@ use ethrex_storage::{
 };
 use ethrex_trie::node::{BranchNode, ExtensionNode, LeafNode};
 use ethrex_trie::{Nibbles, Node, NodeRef, Trie, TrieError, TrieNode};
-use ethrex_vm::PrecompileCache;
 use ethrex_vm::backends::CachingDatabase;
 use ethrex_vm::backends::levm::LEVM;
 use ethrex_vm::backends::levm::db::DatabaseLogger;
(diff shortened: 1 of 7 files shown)


azteca1998 (approved):
Making the cache bounded may be important to avoid out of memory crashes on malicious txs.

crates/blockchain/blockchain.rs
```diff
@@ -409,9 +410,15 @@ impl Blockchain {
         // Replace the VM's store with the caching version
         vm.db.store = caching_store.clone();
 
+        // Create a shared precompile result cache so the executor can reuse
+        // results computed by the warmer (precompiles are pure functions).
+        let precompile_cache = Arc::new(PrecompileCache::new());
+        vm.db.precompile_cache = Some(precompile_cache.clone());
```
Arkenan: Do we need this to be external? If not, let's put it inside the CachingDatabase as we did with the prewarming cache. This is enough as we also pass that caching database to the warmers.

This way the blockchain module doesn't care about precompile caches and low level stuff, only the specific type of db we pass.

crates/vm/levm/src/precompiles.rs
```diff
@@ -336,9 +376,22 @@ pub fn execute_precompile(
...
+    let cache = if address == IDENTITY.address {
+        None
+    } else {
+        cache
+    };
+
+    if let Some((output, gas_cost)) = cache.and_then(|c| c.get(&address, calldata)) {
+        increase_precompile_consumed_gas(gas_cost, gas_remaining)?;
+        return Ok(output);
+    }
```
Arkenan: This is really convoluted. Should we not just do something like

```
if address != IDENTITY.address && let Some((output, gas_cost)) = cache.and_then(|c| c.get(&address, calldata)) {
        increase_precompile_consumed_gas(gas_cost, gas_remaining)?;
        return Ok(output);
}
```

This way we just check a single condition and early return instead of shadowing a cache variable.

crates/vm/levm/src/precompiles.rs
```diff
@@ -348,6 +401,11 @@ pub fn execute_precompile(
         timings.update(address, time);
     }
 
+    if let (Some(cache), Ok(output)) = (cache, &result) {
+        let gas_cost = gas_before.saturating_sub(*gas_remaining);
+        cache.insert(address, calldata.clone(), output.clone(), gas_cost);
+    }
```
Arkenan: Is this possible at all? Have we not already early returned if we had a cache?

PR #6244, opened 2026-02-23
perf(l1): check self before parent in Substate warm/cold lookups

ilitteri:
## Motivation

`add_accessed_slot` and `add_accessed_address` in `Substate` (EIP-2929 warm/cold tracking) check the parent chain **before** checking `self`. Since `Substate` is a linked list (`parent: Option<Box<Self>>`), every warm slot re-access triggers an O(depth) recursive walk even when the slot is already in the current level's set.

For MEV bot transactions with deep CALL stacks (10+ levels), this means every SLOAD/SSTORE pays the full parent chain traversal cost for already-warm slots — the common case.

Found via profiling block 24494268 (413 txs, 43M gas) with samply. `is_slot_accessed` showed 1.5% self-time but the recursive calls merge into parent frames, likely undercounting the true cost.

## Description

Add an early return in `add_accessed_slot` and `add_accessed_address` to check `self` first:

- **`add_accessed_slot`**: check `self.accessed_storage_slots.get(&address).contains(&key)` before walking parents
- **`add_accessed_address`**: check `self.accessed_addresses.contains(&address)` before walking parents

This short-circuits the common case (re-accessing a warm slot/address) in O(log n) / O(1) without touching any parent substates.

No semantic change — the function still returns `true` if the slot was already accessed (warm) and `false` if it's newly inserted (cold).

## How to Test

```bash
cargo test -p ethrex-levm
```

### DIFF
--- a/.github/workflows/pr-main_mdbook.yml
+++ b/.github/workflows/pr-main_mdbook.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Check links
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --no-progress --exclude 'localhost' docs/
+          args: --no-progress --exclude 'localhost' --exclude 'medium.com' docs/
           fail: true
 
   deploy:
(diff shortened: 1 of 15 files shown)


ElFantasma (approved):
nit: `add_selfdestruct` and `add_created_account` follow the same parent-first pattern — could get the same treatment in a follow-up if profiling shows they matter.

PR #6245, opened 2026-02-23
perf(l1): expand fast-path dispatch in LEVM interpreter loop

ilitteri:
## Motivation

The LEVM interpreter loop in `run_execution` has a hand-written `match` for ~70 common opcodes (PUSH1-32, DUP1-16, SWAP1-16, ADD, JUMP, JUMPI, JUMPDEST, MLOAD, CODECOPY) that enables inlining. All other opcodes fall through to an indirect function pointer call (`self.opcode_table[opcode as usize].call(self)`) which cannot be inlined and has higher dispatch overhead.

Profiling block 24494268 showed that ~25-35% of opcode executions go through the indirect path, including high-frequency opcodes like POP, EQ, ISZERO, AND, SUB, MSTORE, SLOAD, and PUSH0.

## Description

Expand the fast-path match with 18 additional opcodes and add `#[inline]` annotations to their handlers:

**Arithmetic**: SUB (0x03), MUL (0x02)
**Comparison/Bitwise**: LT, GT, EQ, ISZERO, AND, OR, SHL, SHR
**Stack/Memory/Storage**: POP, MSTORE, SLOAD, CALLDATALOAD
**System**: RETURN (0xf3), STOP (0x00)
**PUSH0** (0x5f, Shanghai-gated)

The `_ =>` fallback arm is preserved for remaining opcodes. Handler implementations are unchanged — only the dispatch path and `#[inline]` annotations are modified.

This brings the fast-path coverage from ~70 to ~88 opcodes, reducing the indirect dispatch share from ~30% to ~10-15% of executed opcodes.

## How to Test

```bash
cargo test -p ethrex-levm
```

### DIFF
--- a/.github/workflows/pr-main_mdbook.yml
+++ b/.github/workflows/pr-main_mdbook.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Check links
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --no-progress --exclude 'localhost' docs/
+          args: --no-progress --exclude 'localhost' --exclude 'medium.com' docs/
           fail: true
 
   deploy:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 
 ## Perf
 
-### 2026-02-23
+### 2026-02-24
 
 - Expand fast-path dispatch in LEVM interpreter loop [#6245](https://github.com/lambdaclass/ethrex/pull/6245)
+
+### 2026-02-23
+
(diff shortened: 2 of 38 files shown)


azteca1998 (approved):
With this, we have more than 55% of all defined opcodes in the fast path, is there even a need to have a fast path anymore?

ElFantasma (approved):
nit: The existing fast-path handlers `op_jump`, `op_jumpi`, `op_jumpdest`, `op_codecopy`, and `op_tstore` also lack `#[inline]` — might be worth adding for consistency in a follow-up.

PR #6247, opened 2026-02-23
feat(l1): network-specific datadir paths, remove holesky

iovoid:
**Motivation**

Currently running a node with multiple networks requires manually changing the datadir or removing the existing network data. This worsens user experience for developers who might want to have both hoodi and mainnet synced.

**Description**

- Append network-specific subdirectory to datadir for public networks (e.g. ~/.local/share/ethrex/mainnet)
- Add one-time interactive migration from old unsuffixed layout with dry-run + rollback
- Remove deprecated Holesky testnet support

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -594,6 +594,7 @@ pub fn migrate_datadir_if_needed(
     for suffix in Network::all_datadir_suffixes() {
         let subdir = base_datadir.join(suffix);
         if subdir.exists() && subdir.is_dir() {
+            info!("Found existing network subdirectory {subdir:?}, skipping migration.");
             return;
         }
     }


cmd/ethrex/initializers.rs
```diff
@@ -556,6 +554,128 @@ pub async fn init_l1(
...
+        warn!(
+            "Existing database at {base_datadir:?} has chain ID {db_chain_id}, \
+             expected {expected_chain_id} for {network}. Skipping migration."
+        );
+        return;
+    }
+
+    // All checks passed — prompt the user.
+    info!("Found existing database at {base_datadir:?} that can be moved to {network_datadir:?}.");
+    print!("Move existing database to the new network-specific directory? (y/n): ");
```
ElFantasma: This `print!`/`stdin().read_line()` runs during `init_l1()` on every startup. In headless deployments (systemd, Docker without `-it`, CI), stdin is `/dev/null` or a closed pipe, so `read_line` returns `Ok(0)` (EOF) → `input` is empty → migration silently skips at `info` level → node starts a fresh sync even though valid synced data exists one directory up.

Two suggestions:
1. **Detect TTY**: guard the prompt with `std::io::stdin().is_terminal()` (stable since Rust 1.70). When non-interactive, log at `warn!` level with an actionable message like `"Run ethrex interactively or use --migrate-datadir to move existing data"`.
2. **CLI flag**: add `--migrate-datadir` (or `--no-migrate-datadir`) so migration can be triggered/suppressed without interactive input, which is the typical pattern for server software that may also run interactively.

crates/storage/store.rs
```diff
@@ -3143,3 +3143,43 @@ fn dir_is_empty(path: &Path) -> Result<bool, StoreError> {
...
+/// store initialization. Returns `None` if the database doesn't exist or
+/// the chain config can't be read. Always returns `None` when compiled
+/// without the `rocksdb` feature.
+pub fn read_chain_id_from_db(path: &Path) -> Option<u64> {
+    if !has_valid_db(path) {
+        return None;
+    }
+    #[cfg(feature = "rocksdb")]
+    {
+        let backend = RocksDBBackend::open(path).ok()?;
```
ElFantasma, line 3196: nit: `RocksDBBackend::open()` allocates production-weight resources (1 GB write buffer, WAL, `create_if_missing(true)`) for a one-shot read of a single key. `rocksdb::DB::open_for_read_only()` would skip write-path setup entirely and guarantee no mutation of the DB.

iovoid: This is _only_ done when a migration is needed, the cost is small. I think it's better to avoid the extra code.

cmd/ethrex/initializers.rs
iovoid: Done in 7bc34b14f, choosing migration as the default solution. Otherwise the user would need to resync, when we could just do the migration. Since the migration is just a move, it's trivial to undo if it was unwanted.

Added a --no-migrate flag if for some reason the user wants to opt-out.

PR #6249, opened 2026-02-23
fix(l1): bound precompile cache memory with LRU eviction

ilitteri:
## Summary
This PR hardens the shared precompile cache by introducing a byte-bounded LRU eviction policy.

## Problem
`PrecompileCache` previously used an unbounded map keyed by `(Address, Bytes)` and cached successful precompile calls without any memory ceiling. Because calldata is part of the key, adversarial blocks can force unbounded growth in cache memory during warming and execution.

## Root Cause
The cache insert path had no upper bound (entries or bytes) and no eviction policy, so every distinct key could remain resident until block execution completed.

## Fix
- Replace the unbounded map with an `IndexMap`-backed LRU state.
- Add a hard cap of `128 MiB` for cached precompile entries.
- Track current cache footprint in bytes (metadata + calldata + output).
- On cache hit, refresh recency.
- On insert:
  - skip entries larger than the configured cap,
  - correctly account for replacement of existing keys,
  - evict least-recently-used entries until under the cap.
- Keep the public `PrecompileCache` API used by callers unchanged.

## Validation
- Ran `cargo test -p ethrex-levm precompile_cache`
- Added and passed focused tests:
  - evicts least-recently-used entries at capacity,
  - skips oversized entries,
  - disables caching with zero capacity.

### DIFF
--- a/crates/guest-program/bin/openvm/Cargo.lock
+++ b/crates/guest-program/bin/openvm/Cargo.lock
@@ -986,6 +986,7 @@ dependencies = [
  "ethrex-common",
  "ethrex-crypto",
  "ethrex-rlp",
+ "indexmap 2.13.0",
  "k256 0.13.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "lambdaworks-math",
  "lazy_static",
--- a/crates/guest-program/bin/risc0/Cargo.lock
+++ b/crates/guest-program/bin/risc0/Cargo.lock
@@ -1235,6 +1235,7 @@ dependencies = [
  "ethrex-common",
  "ethrex-crypto",
  "ethrex-rlp",
+ "indexmap 2.13.0",
  "k256",
  "lambdaworks-math",
  "lazy_static",
--- a/crates/guest-program/bin/sp1/Cargo.lock
+++ b/crates/guest-program/bin/sp1/Cargo.lock
@@ -1020,6 +1020,7 @@ dependencies = [
  "ethrex-common",
  "ethrex-crypto",
  "ethrex-rlp",
+ "indexmap 2.13.0",
  "k256",
  "lambdaworks-math",
  "lazy_static",
--- a/crates/guest-program/bin/zisk/Cargo.lock
+++ b/crates/guest-program/bin/zisk/Cargo.lock
@@ -978,6 +978,7 @@ dependencies = [
  "ethrex-common",
  "ethrex-crypto",
  "ethrex-rlp",
+ "indexmap 2.13.0",
  "k256",
  "lambdaworks-math",
  "lazy_static",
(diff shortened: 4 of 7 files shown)


crates/vm/levm/src/precompiles.rs
```diff
@@ -307,22 +326,63 @@ impl PrecompileCache {
...
-        self.cache
-            .read()
-            .unwrap_or_else(|poisoned| poisoned.into_inner())
-            .get(&(*address, calldata.clone()))
-            .cloned()
+        // holding it), recover the inner state and keep operating. The cache is
+        // a pure optimization, so this can never affect correctness.
+        let mut cache = self
+            .cache
+            .write()
```
ElFantasma, line 342: Switching `get()` from `.read()` to `.write()` serializes all cache access — the warmer's `insert()` calls and the executor's `get()` lookups now contend for the same exclusive lock, turning the `RwLock` into an effective `Mutex`.

For a per-block cache the LRU refresh on read is probably unnecessary: the warmer populates entries during warming, the executor consumes them during execution, and the cache is discarded at block end. If the working set fits in 128 MiB (very likely for a single block), no entries will be evicted anyway, so their relative order doesn't matter.

Consider keeping `.read()` here and dropping the `shift_remove`/re-insert dance — insertion-time eviction is sufficient. Alternatively, if `get()` will always need exclusive access, switch to `Mutex` to avoid the reader/writer tracking overhead of `RwLock`.

crates/vm/levm/src/precompiles.rs
```diff
@@ -289,16 +289,35 @@ pub fn is_precompile(address: &Address, fork: Fork, vm_type: VMType) -> bool {
...
+
+impl PrecompileCacheInner {
+    fn new() -> Self {
+        Self {
+            entries: IndexMap::new(),
+            used_bytes: 0,
+        }
+    }
+
+    fn entry_size(calldata_len: usize, output_len: usize) -> usize {
```
ElFantasma, line 307: nit: this only counts calldata + output bytes, but each entry also stores an `Address` (20 B), `u64` gas_cost (8 B), two `Bytes` handles (~24 B each), and `IndexMap` per-slot overhead (~50 B for hash + indices). For many small entries the tracked `used_bytes` could undercount actual resident memory by ~2×. Not blocking — with the 128 MiB default it's unlikely to matter in practice — but worth a `// NOTE:` so future readers know the cap is approximate.

ElFantasma (changes requested):

crates/vm/levm/src/precompiles.rs
```diff
@@ -389,22 +405,51 @@ impl PrecompileCache {
         Self::default()
     }
 
+    pub fn with_max_bytes(max_bytes: usize) -> Self {
```
MegaRedHand: This is only used for tests. I think we can remove it with the `max_bytes` field

MegaRedHand (approved):

PR #6253, opened 2026-02-24
fix(l1): fix broadcast_pool race and offload tx pool insertion to background task

edg-l:
## Summary

- **Fix broadcast_pool race condition**: `clear_broadcasted_txs()` wiped the entire `broadcast_pool`, including txs added between the read (`get_txs_for_broadcast()`) and the clear. With 2000 txs being inserted one-by-one and the broadcaster firing every 5ms, this reliably dropped txs that were never announced to peers. Replaced with targeted `remove_broadcasted_txs()` that only removes the txs that were actually fetched and broadcast.
- **Offload tx pool insertion to background task**: Spawn the incoming `Transactions` message pool-insertion loop as a background `tokio::spawn` task instead of running it synchronously in `handle_cast`. This prevents the ConnectionServer from being blocked during signature recovery + storage reads for large transaction batches.

## Motivation

The hive `LargeTxRequest` devp2p eth test intermittently fails in CI ([example run](https://github.com/lambdaclass/ethrex/actions/runs/22335807040)). Root cause is a race condition in the transaction broadcaster:

1. `get_txs_for_broadcast()` reads the `broadcast_pool` (releases lock)
2. Meanwhile, incoming txs continue to be added to `broadcast_pool`
3. `clear_broadcasted_txs()` clears **all** of `broadcast_pool`, including txs added in step 2

Any tx added between steps 1 and 3 is lost forever — never broadcast but cleared from the pool. With 2000 txs being inserted sequentially and the broadcaster firing every 5ms, this race reliably drops transactions, causing the test's 2-second timeout to expire before all tx hashes are announced.

## Test plan

- [x] `cargo check` (with and without `l2` feature)
- [x] Local hive devp2p eth tests pass 3/3 runs: `./hive --sim devp2p --sim.limit "eth" --client ethrex` (20/20 each run)
- [ ] CI hive daily run should show P2P Eth capability back to 20/20

### DIFF
--- a/crates/networking/p2p/rlpx/connection/server.rs
+++ b/crates/networking/p2p/rlpx/connection/server.rs
@@ -945,34 +945,41 @@ async fn handle_incoming_message(
         Message::Transactions(txs) if peer_supports_eth => {
             // https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transactions-0x02
             if state.blockchain.is_synced() {
+                let tx_hashes: Vec<_> = txs.transactions.iter().map(|tx| tx.hash()).collect();
+
+                // Offload pool insertion to a background task so we don't block
+                // the ConnectionServer (validation + signature recovery are expensive).
+                let blockchain = state.blockchain.clone();
+                let peer = state.node.to_string();
                 #[cfg(feature = "l2")]
                 let is_l2_mode = state.l2_state.is_supported();
-                for tx in &txs.transactions {
-                    // Reject blob transactions in L2 mode
-                    #[cfg(feature = "l2")]
-                    if (is_l2_mode && matches!(tx, Transaction::EIP4844Transaction(_)))
-                        || tx.is_privileged()
-                    {
-                        let tx_type = tx.tx_type();
-                        debug!(peer=%state.node, "Rejecting transaction in L2 mode - {tx_type} transactions are not broadcasted in L2");
-                        continue;
-                    }
+                tokio::spawn(async move {
+                    for tx in txs.transactions {
+                        #[cfg(feature = "l2")]
+                        if (is_l2_mode && matches!(tx, Transaction::EIP4844Transaction(_)))
+                            || tx.is_privileged()
+                        {
+                            let tx_type = tx.tx_type();
+                            debug!(peer=%peer, "Rejecting transaction in L2 mode - {tx_type} transactions are not broadcasted in L2");
+                            continue;
+                        }
 
-                    if let Err(e) = state.blockchain.add_transaction_to_pool(tx.clone()).await {
-                        debug!(
-                            peer=%state.node,
-                            error=%e,
-                            "Error adding transaction"
-                        );
-                        continue;
+                        if let Err(e) = blockchain.add_transaction_to_pool(tx).await {
+                            debug!(
+                                peer=%peer,
+                                error=%e,
+                                "Error adding transaction"
+                            );
+                        }
                     }
-                }
+                });
+
+                // Notify the broadcaster immediately — it only tracks hashes
+                // to avoid re-broadcasting to the sender. The actual broadcast
+                // happens on a periodic timer that queries the mempool directly.
                 state
                     .tx_broadcaster
-                    .cast(InMessage::AddTxs(
-                        txs.transactions.iter().map(|tx| tx.hash()).collect(),
-                        state.node.node_id(),
-                    ))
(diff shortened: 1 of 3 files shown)


edg-l:
## Addressing bot reviews

### Kimi findings

1. **Race condition (hashes collected before task runs):** By design. `AddTxs` only marks hashes as "known from this peer" in a `known_txs` map to avoid re-broadcasting back to the sender. It does NOT trigger any broadcast or require pool state. Actual broadcasting happens on a periodic timer that queries mempool directly via `get_txs_for_broadcast()`.

2. **Missing deduplication check before spawning:** `add_transaction_to_pool` already does `mempool.contains_tx(hash)` as its very first check (blockchain.rs:2421). Adding another check before spawning would be redundant.

3. **Unbounded task spawning (DoS):** One task per `Transactions` message, not per transaction. The original code was equally vulnerable to flooding — it just blocked the ConnectionServer instead of spawning. Bounded concurrency could be a follow-up but is out of scope here.

4. **Missing task cancellation on disconnect:** The spawned tasks just do pool inserts and exit. No long-running loops or waits. Adding cancellation logic would add complexity for negligible gain on a non-critical path.

5. **Inefficient `blockchain.clone()`:** `Blockchain` is `Arc`-wrapped. Clone is a ref count bump, not a deep copy.

6. **Debug logging inconsistency:** We can't capture `state.node` (a reference) into the spawned `async move` block. `peer` is `state.node.to_string()` — functionally identical output.

7. **Missing spawn failure handling:** `tokio::spawn` only fails if the runtime is shutting down, at which point everything is tearing down anyway.

8. **Transaction validation timing:** Validation still happens in the spawned task before pool insertion. The ordering is unchanged — validate then insert.

### Codex findings

1. **L2-rejected hashes sent to broadcaster:** Not a regression. The original code did the exact same thing — `txs.transactions.iter().map(|tx| tx.hash()).collect()` was called on ALL transactions without L2 filtering. The `AddTxs` handler only uses these hashes to suppress re-broadcast to the sender, so stale entries in `known_txs` are harmless (pruned after 10 minutes).

2. **Unbounded task spawning:** Same as Kimi #3 above.

### Claude review

Good observations — agree with all of them. The TOCTOU in `add_transaction_to_pool` and dead `known_txs` entries for rejected hashes are both pre-existing and not regressions. The `JoinHandle` drop is acceptable since insertion failures are already `debug!`-logged inside the task.

---

## Update: broadcast_pool race condition fix

Investigation revealed the **true root cause** of the flaky `LargeTxRequest` test was a race condition in the tx broadcaster, not the ConnectionServer blocking:

- `get_txs_for_broadcast()` reads the `broadcast_pool` (releases lock)
- Meanwhile, incoming txs continue to be added to `broadcast_pool` one-by-one
- `clear_broadcasted_txs()` clears **all** of `broadcast_pool`, including txs added since the read

Any tx added between the read and the clear is lost forever — never broadcast but removed from the pool. With 2000 txs being inserted sequentially and the broadcaster firing every 5ms, this reliably drops transactions.

**Fix:** Replaced `clear_broadcasted_txs()` with `remove_broadcasted_txs(hashes)` that only removes the specific txs that were actually fetched and sent. After the fix, hive devp2p eth tests pass 3/3 full runs (20/20 each).

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -945,34 +945,41 @@ async fn handle_incoming_message(
...
-                    // Reject blob transactions in L2 mode
-                    #[cfg(feature = "l2")]
-                    if (is_l2_mode && matches!(tx, Transaction::EIP4844Transaction(_)))
-                        || tx.is_privileged()
-                    {
-                        let tx_type = tx.tx_type();
-                        debug!(peer=%state.node, "Rejecting transaction in L2 mode - {tx_type} transactions are not broadcasted in L2");
-                        continue;
-                    }
+                tokio::spawn(async move {
```
ElFantasma, line 956: The `JoinHandle` is dropped here, making this fire-and-forget. If `add_transaction_to_pool` panics (e.g., poisoned `RwLock`), the panic is caught by the tokio runtime but won't surface clearly. More importantly, on node shutdown, in-flight insertion tasks may be silently dropped.

Not necessarily a blocker since the task is bounded (processes a fixed set of txs and the errors are already debug-logged), but worth considering whether to at least log on task failure, e.g.:
```rust
tokio::spawn(async move {
    // ... existing code ...
}.instrument(tracing::debug_span!("tx_pool_insert")));
```
or store the handle in a `JoinSet` for clean shutdown.

PR #6254, opened 2026-02-24
Use Arc<[u32]> for Code.jump_targets to make Code::clone() O(1)

ilitteri:
## Motivation

Every EVM CALL-family opcode clones the callee's bytecode (`Code`) when constructing a new call frame. On mainnet, contracts like routers and DEX aggregators trigger dozens of internal calls per transaction, each cloning potentially large bytecode vectors. The only expensive field to clone in `Code` is `jump_targets: Vec<u32>` — `Bytes` is already Arc-backed (O(1) clone) and `H256` is Copy.

## Description

Change `Code.jump_targets` from `Vec<u32>` to `Arc<[u32]>`, making `Code::clone()` three cheap operations:
- `H256`: Copy
- `Bytes`: refcount bump (already Arc-backed)
- `Arc<[u32]>`: refcount bump

This avoids the previous `Arc<Code>` approach (PR v1) which added pointer indirection on every opcode fetch in the hot interpreter loop. With `Arc<[u32]>`, `Code` stays inline in `CallFrame` — no extra cache-line fetch per opcode — while clone cost drops from O(n) to O(1).

## How to Test

Run on mainnet and compare throughput/block-time metrics against baseline. CI benchmarks should show no regression.

### DIFF
--- a/crates/vm/levm/src/call_frame.rs
+++ b/crates/vm/levm/src/call_frame.rs
@@ -10,7 +10,7 @@ use bytes::Bytes;
 use ethrex_common::types::block_access_list::BlockAccessListCheckpoint;
 use ethrex_common::{Address, U256};
 use ethrex_common::{H256, types::Code};
-use std::{collections::HashMap, fmt, hint::assert_unchecked, sync::Arc};
+use std::{collections::HashMap, fmt, hint::assert_unchecked};
 
 /// [`u64`]s that make up a [`U256`]
 const U64_PER_U256: usize = U256::MAX.0.len();
@@ -244,7 +244,7 @@ pub struct CallFrame {
     /// Bytecode to execute.
     /// Its hash field will be bogus for initcodes, as it is inaccessible to the VM
     /// unless associated to an account, which doesn't happen for its initcode.
-    pub bytecode: Arc<Code>,
+    pub bytecode: Code,
     /// Value sent along the transaction
     pub msg_value: U256,
     pub stack: Stack,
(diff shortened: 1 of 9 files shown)


crates/vm/levm/src/call_frame.rs
```diff
@@ -244,7 +244,7 @@ pub struct CallFrame {
     /// Bytecode to execute.
     /// Its hash field will be bogus for initcodes, as it is inaccessible to the VM
     /// unless associated to an account, which doesn't happen for its initcode.
-    pub bytecode: Code,
+    pub bytecode: Arc<Code>,
```
ElFantasma: `Arc<Code>` adds a pointer indirection every time `next_opcode()` reads `self.bytecode.bytecode[pc]`:

- **Before:** `CallFrame (inline) → Code.bytecode → Bytes buffer` — Code lives inside the struct, one fewer cache-line fetch.
- **After:** `CallFrame → Arc ptr → Code (heap) → Bytes buffer` — extra dereference on every opcode.

Since `Bytes` is already Arc-backed (O(1) clone) and `H256` is Copy, the only expensive field to clone is `jump_targets: Vec<u32>`. There's already a TODO in `crates/common/types/account.rs:32` noting this:

```rust
// TODO: Consider using Arc<[u32]> (needs to enable serde rc feature)
pub jump_targets: Vec<u32>,
```

If you change `jump_targets` to `Arc<[u32]>`, then `Code::clone()` becomes three cheap ops (Copy + refcount + refcount) and you can drop the `Arc<Code>` wrapper entirely — same clone savings, no extra indirection in the hot loop, likely no throughput regression.

PR #6256, opened 2026-02-24
fix(l1): fix p2p use consistent encoding for blob tx size in NewPooledTransactionHashes

### LINKED ISSUE #6255: bug(p2p): ethrex-to-ethrex peers stuck in connect/disconnect loop - Invalid pooled transaction size
## Description

When two ethrex nodes peer with each other, they enter a rapid connect/disconnect loop (~5-10 cycles/second). The handshake succeeds (eth/69 + snap/1 negotiated), but immediately after, the connection is dropped with:

```
WARN ethrex_p2p::rlpx::connection::server: disconnected from peer peer=ethrex #0x3c96…bcb8(172.16.0.11:30303) reason=Invalid pooled transaction size, differs from expected
```

This happens on `main` — not caused by any feature branch.

## Reproduction

1. Run the Kurtosis devnet with two ethrex nodes (no geth):
   ```
   fixtures/networks/bal-devnet-2-ethrex.yaml
   ```
   Both nodes use `ethrex:main`.

2. Check logs:
   ```
   kurtosis service logs -f lambdanet el-1-ethrex-lighthouse
   ```

3. Observe the connect/disconnect cycle repeating every ~100-200ms:
   ```
   DEBUG Negotatied eth version: eth/69
   DEBUG Negotatied snap version: snap/1
   WARN  disconnected from peer ... reason=Invalid pooled transaction size, differs from expected
   DEBUG Negotatied eth version: eth/69
   DEBUG Negotatied snap version: snap/1
   WARN  disconnected from peer ... reason=Invalid pooled transaction size, differs from expected
   (repeats indefinitely)
   ```

<details>
<summary>bal-devnet-2-ethrex.yaml</summary>

```yaml
participants:
  - cl_type: lighthouse
    cl_image: ethpandaops/lighthouse:bal-devnet-2
    el_type: ethrex
    el_image: ethrex:local
    el_extra_params:
      - "--syncmode=full"
    el_extra_env_vars:
      RUST_LOG: "debug,ethrex_p2p::tx_broadcaster=info,ethrex_blockchain::payload=info"
  - cl_type: lighthouse
    cl_image: ethpandaops/lighthouse:bal-devnet-2
    el_type: ethrex
    el_image: ethrex:local
    el_extra_params:
      - "--syncmode=full"
    el_extra_env_vars:
      RUST_LOG: "debug,ethrex_p2p::tx_broadcaster=info,ethrex_blockchain::payload=info"
global_log_level: 'debug'
network_params:
  genesis_delay: 20
  fulu_fork_epoch: 0
  gloas_fork_epoch: 1
  seconds_per_slot: 6
snooper_enabled: true
ethereum_genesis_generator_params:
  image: ethpandaops/ethereum-genesis-generator:master
dora_params:
  image: ethpandaops/dora:eip7928-support
additional_services:
  - dora
  - spamoor
port_publisher:
  additional_services:
    enabled: true
    public_port_start: 64400
spamoor_params:
  image: ethpandaops/spamoor:master
  spammers:
    # Mainnet-like workload mix (~700 tx/s target)
    # ~40% ERC20 transfers (hot storage: balances, allowances)
    - scenario: erc20tx
      config:
        throughput: 280
        max_wallets: 500
    # ~25% DEX swaps (complex contract calls, hot pool state)
    - scenario: uniswap-swaps
      config:
        throughput: 175
        max_wallets: 300
    # ~15% simple ETH transfers
    - scenario: eoatx
      config:
        throughput: 105
    # ~10% contract calls (storage-heavy)
    - scenario: storagespam
      config:
        throughput: 70
    # ~5% ERC721 mints/transfers (NFTs)
    - scenario: erc721tx
      config:
        throughput: 35
    # ~5% blob txs (L2 sequencer-like)
    - scenario: blobs
      config:
        throughput: 35
        max_pending: 16
```

</details>

## Impact

- **Missed slots**: With both nodes on `main`, ethrex-1 missed **8 out of 40 slots (20%)** in the first epoch alone. For comparison, with one node on the parallel-exec branch (which adds geth-unrelated optimizations only), misses were 5/120 (4%) — confirming this is a `main` p2p issue, not a feature branch regression.
- **CPU/log spam**: Continuous reconnection attempts waste resources and flood logs.
- **No ethrex-to-ethrex tx propagation**: Since connections never stabilize, transactions only reach ethrex nodes via the spamoor RPC, not via p2p gossip between peers.

## Root Cause

The error `Invalid pooled transaction size, differs from expected` comes from the `PooledTransactions` response handler. After the handshake succeeds, one node sends a `GetPooledTransactions` request, and the response fails validation — likely a size/count mismatch in the eth/69 protocol handling.

## Not Amsterdam-specific

The devnet config has `gloas_fork_epoch: 1` (Amsterdam at epoch 1). The disconnect loop starts immediately in epoch 0 (Fulu), before Amsterdam activates — confirming this is a general ethrex-to-ethrex p2p issue, not related to Amsterdam/BAL changes.

## Open Questions

- **ethrex-to-geth**: Does this also happen when ethrex peers with geth? Not yet tested with this specific config.

## Environment

- ethrex `main` branch
- eth/69 protocol
- bal-devnet-2 network (bug occurs during Fulu epoch, before Amsterdam activation)
- Kurtosis devnet with Lighthouse CL

edg-l:
## Summary

- Fix EIP-4844 blob transaction size mismatch between `NewPooledTransactionHashes` announcement and `PooledTransactions` validation, which caused ethrex-to-ethrex peers to enter a rapid connect/disconnect loop
- Remove now-dead `rlp_encode_as_pooled_tx` / `rlp_length_as_pooled_tx` methods from `EIP4844Transaction`

### Root cause

The advertised size used `rlp_encode_as_pooled_tx` which produced two concatenated RLP lists (`0x03 || rlp([tx_fields]) || rlp([blobs, commitments, proofs])`), while the validation used `P2PTransaction::encode_canonical_to_vec` which produces a single outer RLP list (`0x03 || rlp([rlp([tx_fields]), wrapper_version?, blobs, commitments, proofs])`). The structural difference (plus the missing `wrapper_version` field) meant sizes never matched for blob transactions, triggering an immediate disconnect.

### Fix

Construct the same `P2PTransaction::EIP4844TransactionWithBlobs` in the announcement and use `encode_canonical_to_vec().len()` — the exact same encoding path the validator uses.

## Test plan

- [x] `cargo clippy` / `cargo test` pass on `ethrex-p2p` and `ethrex-common`
- [x] Tested on Kurtosis `bal-devnet-2-ethrex` devnet: the connect/disconnect loop no longer occurs
- [x] Hive `engine-cancun` blob pooled transaction tests pass (all 3):
  ```
  make run-hive SIMULATION=ethereum/engine TEST_PATTERN="engine-cancun/Blob Pooled Transactions"
  ```

Closes #6255

### DIFF
--- a/crates/networking/p2p/rlpx/eth/transactions.rs
+++ b/crates/networking/p2p/rlpx/eth/transactions.rs
@@ -97,14 +97,13 @@ impl NewPooledTransactionHashes {
                         .mempool
                         .get_blobs_bundle(transaction_hash)?
                         .unwrap_or_default();
-                    let p2p_tx = P2PTransaction::EIP4844TransactionWithBlobs(
-                        WrappedEIP4844Transaction {
+                    let p2p_tx =
+                        P2PTransaction::EIP4844TransactionWithBlobs(WrappedEIP4844Transaction {
                             tx: eip4844_tx,
                             wrapper_version: (tx_blobs_bundle.version != 0)
                                 .then_some(tx_blobs_bundle.version),
                             blobs_bundle: tx_blobs_bundle,
-                        },
-                    );
+                        });
                     p2p_tx.encode_canonical_to_vec().len()
                 }
                 _ => transaction.encode_canonical_to_vec().len(),


crates/networking/p2p/rlpx/eth/transactions.rs
```diff
@@ -85,18 +85,27 @@ impl NewPooledTransactionHashes {
...
                     let tx_blobs_bundle = blockchain
                         .mempool
                         .get_blobs_bundle(transaction_hash)?
-                        .unwrap_or(BlobsBundle::empty());
-                    eip4844_tx.rlp_length_as_pooled_tx(&tx_blobs_bundle)
+                        .unwrap_or_default();
+                    let p2p_tx = P2PTransaction::EIP4844TransactionWithBlobs(
+                        WrappedEIP4844Transaction {
+                            tx: eip4844_tx,
+                            wrapper_version: (tx_blobs_bundle.version != 0)
```
ElFantasma, line 103: nit: This is correct — `version` defaults to `0` via `#[serde(skip, default)]` on `BlobsBundle`, and the decode side (`WrappedEIP4844Transaction::decode_unfinished`) reads `wrapper_version` as optional. So `None` here matches the no-version case and `Some(v)` matches the versioned case.

One thing to watch: if `get_blobs_bundle` returns `None`, `unwrap_or_default()` gives an empty `BlobsBundle` (version=0, no blobs). The encoded size would then reflect a blob tx with zero blobs, which won't match the size of the actual response that includes blobs. This is a pre-existing concern (the old code also fell back to `BlobsBundle::empty()`), but worth noting that `get_blobs_bundle` returning `None` for a mempool blob tx would cause a size mismatch on the receiving end.

PR #6259, opened 2026-02-24
perf(l1): disable balance check for prewarming to avoid early reverts

ilitteri:
## Motivation

During block execution, the prewarmer thread runs transactions speculatively to warm storage slots and account state into the cache. However, when a sender has insufficient balance, the prewarmer reverts early in `validate_sender_balance` / `deduct_caller`, missing the opportunity to warm the storage slots that the transaction would have touched deeper in execution.

## Description

- Add `disable_balance_check` flag to the LEVM `Environment` struct
- Skip `validate_sender_balance` and `deduct_caller` when the flag is set
- Pass `disable_balance_check: true` only from the prewarmer path, `false` for normal execution
- This lets the prewarmer reach deeper into the transaction's execution path, warming more storage slots and improving cache hit rates for the actual executor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Perf
 
+### 2026-02-25
+
+- Disable balance check for prewarming to avoid early reverts [#6259](https://github.com/lambdaclass/ethrex/pull/6259)
+
 ### 2026-02-24
 
 - Expand fast-path dispatch in LEVM interpreter loop [#6245](https://github.com/lambdaclass/ethrex/pull/6245)


crates/vm/backends/levm/mod.rs
```diff
@@ -353,6 +354,7 @@ impl LEVM {
                         &mut group_db,
                         vm_type,
                         stack_pool,
+                        true,
```
ElFantasma, line 357: nit: bare boolean at the call site (`true` / `false`) doesn't convey intent. Reads a bit better with a `/* disable_balance_check */` comment or a constant, e.g.
```rust
const WARMING_MODE: bool = true;
```
Same applies to the `false` at the normal execution call site (line 238).

ElFantasma (approved):

PR #6261, opened 2026-02-25
feat(tokamak-jit): add bytecode size limit with graceful interpreter fallback (D-2)

jason-h23:
## Summary

- Add negative cache (`oversized_hashes: FxHashSet<H256>`) to `JitState` for O(1) skip of known-oversized bytecodes (> 24576 bytes, EIP-170)
- Add early size gate in VM dispatch at compilation threshold — prevents compilation attempts before they reach the backend
- Add belt-and-suspenders size check in background compiler thread (defense-in-depth)
- Return interpreter-only benchmark results instead of silently dropping oversized scenarios

## Changes

| File | Change |
|------|--------|
| `crates/vm/levm/src/jit/dispatch.rs` | `oversized_hashes` field + `is_oversized()`/`mark_oversized()` + 4 unit tests |
| `crates/vm/levm/src/vm.rs` | Early size gate wrapping JIT compile+dispatch block |
| `crates/vm/tokamak-jit/src/lib.rs` | Size check in background compiler thread before compile |
| `crates/tokamak-bench/src/jit_bench.rs` | Return interpreter-only `JitBenchResult` on compile failure |
| `crates/vm/tokamak-jit/src/tests/oversized.rs` | **NEW** — 3 integration tests (`revmc-backend` gated) |
| `crates/vm/tokamak-jit/src/tests/mod.rs` | Register `oversized` module |

## Test plan

- [x] `cargo test -p ethrex-levm --features tokamak-jit,test-utils -- jit::dispatch::tests` — 4/4 pass
- [x] `cargo test -p tokamak-jit --no-default-features -- --test-threads=1` — 22/22 pass (existing unaffected)
- [x] `cargo test -p tokamak-bench` — 35/35 pass (existing unaffected)
- [ ] 3 integration tests (`revmc-backend` gated) — run in CI with LLVM 21
- [x] `cargo clippy -p ethrex-levm -p tokamak-jit -p tokamak-bench` — clean
- [x] `cargo fmt -- --check` — clean

### DIFF
--- a/crates/vm/levm/src/vm.rs
+++ b/crates/vm/levm/src/vm.rs
@@ -17,6 +17,8 @@ use crate::{
     tracing::LevmCallTracer,
 };
 use bytes::Bytes;
+#[cfg(feature = "tokamak-l2")]
+use ethrex_common::types::l2::tokamak_fee_config::TokamakFeeConfig;
 use ethrex_common::{
     Address, H160, H256, U256,
     tracing::CallType,
@@ -64,6 +66,9 @@ pub enum VMType {
     L1,
     /// L2 rollup execution with additional fee handling.
     L2(FeeConfig),
+    /// Tokamak L2 execution with proven execution metadata and JIT policy.
+    #[cfg(feature = "tokamak-l2")]
+    TokamakL2(TokamakFeeConfig),
 }
 
 /// Execution substate that tracks changes during transaction execution.
--- a/cmd/ethrex/Cargo.toml
+++ b/cmd/ethrex/Cargo.toml
@@ -130,7 +130,7 @@ risc0 = ["ethrex-prover/risc0", "ethrex-l2/risc0"]
 perf_opcode_timings = ["ethrex-vm/perf_opcode_timings"]
 tokamak-jit = ["ethrex-vm/tokamak-jit"]
 tokamak-debugger = ["ethrex-vm/tokamak-debugger"]
-tokamak-l2 = ["ethrex-vm/tokamak-l2"]
+tokamak-l2 = ["ethrex-vm/tokamak-l2", "ethrex-blockchain/tokamak-l2"]
 tokamak = ["ethrex-vm/tokamak"]
 cpu_profiling = ["dep:pprof"]
 
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -316,6 +316,16 @@ pub struct Options {
         env = "ETHREX_PRECOMPUTE_WITNESSES"
     )]
     pub precompute_witnesses: bool,
+    #[cfg(feature = "tokamak-l2")]
+    #[arg(
+        long = "tokamak-l2",
+        action = ArgAction::SetTrue,
+        default_value = "false",
+        help = "Run as Tokamak L2 node with proven execution and JIT-by-default policy",
+        help_heading = "Node options",
+        env = "ETHREX_TOKAMAK_L2"
+    )]
+    pub tokamak_l2: bool,
 }
 
 impl Options {
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -469,12 +469,27 @@ pub async fn init_l1(
     #[cfg(feature = "sync-test")]
     set_sync_block(&store).await;
 
+    #[cfg(feature = "tokamak-l2")]
+    let blockchain_type = if opts.tokamak_l2 {
+        use ethrex_blockchain::TokamakL2Config;
+        use ethrex_common::types::l2::tokamak_fee_config::JitPolicy;
+        BlockchainType::TokamakL2(TokamakL2Config {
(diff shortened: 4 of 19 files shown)


azteca1998:
Hello, I think you may have mistakenly opened a PR to our repository instead of your fork.
Can you confirm whether that's the case?

crates/vm/levm/src/vm.rs
```diff
@@ -30,6 +30,29 @@ use std::{
...
+#[cfg(feature = "tokamak-jit")]
+type ValidationSnapshot = (
+    GeneralizedDatabase,
+    CallFrame,
+    Substate,
+    FxHashMap<(Address, H256), U256>,
+);
+
+#[cfg(feature = "tokamak-jit")]
+lazy_static::lazy_static! {
```
ElFantasma, line 49: nit: The codebase uses `std::sync::LazyLock` for static initialization (see `timings.rs:OPCODE_TIMINGS`). Using `lazy_static!` here is inconsistent and pulls in an unnecessary dependency. Replace with:

```rust
pub static JIT_STATE: LazyLock<crate::jit::dispatch::JitState> =
    LazyLock::new(crate::jit::dispatch::JitState::new);
```

crates/vm/levm/src/vm.rs
```diff
@@ -572,7 +613,250 @@ impl<'a> VM<'a> {
...
+                                    &JIT_STATE.cache,
+                                ) {
+                                    Ok(()) => {
+                                        JIT_STATE
+                                            .metrics
+                                            .compilations
+                                            .fetch_add(1, Ordering::Relaxed);
+                                    }
+                                    Err(e) => {
+                                        eprintln!(
```
ElFantasma, line 671: Multiple `eprintln!` calls in the JIT dispatch path (here, line ~767 for validation failure, line ~823 for fallback) should use `tracing::warn!` or `tracing::error!` to match the rest of the codebase. `eprintln!` bypasses structured logging, log levels, and any configured log sinks.

crates/vm/levm/src/vm.rs
```diff
@@ -572,7 +613,250 @@ impl<'a> VM<'a> {
...
+                        // Only allocate when validation will actually run for this cache key.
+                        // Skip validation for bytecodes with CALL/CREATE — the state-swap
+                        // mechanism cannot correctly replay subcalls (see CRITICAL-1).
+                        let cache_key = (bytecode_hash, fork);
+                        let needs_validation = JIT_STATE.config.validation_mode
+                            && JIT_STATE.should_validate(&cache_key)
+                            && !compiled.has_external_calls;
+                        let pre_jit_snapshot = if needs_validation {
+                            Some((
+                                self.db.clone(),
```
ElFantasma, line 698: `self.db.clone()` deep-clones `GeneralizedDatabase` (including `current_accounts_state`, a full `HashMap` of all touched accounts). Even bounded to `max_validation_runs` (3) per bytecode, this is expensive for contracts that touch many accounts. The validation feature defaults to enabled (`validation_mode: true`), so this cost applies out-of-the-box. Consider making validation opt-in (default `false`) since this is a PoC correctness tool, not something that should run in production.

ElFantasma (commented):
This looks like AI slop or an error. I'm not closing it to let the rest of the team look at it and make a decision regarding anti-slop measures but recommend closing.

In any case, some summary:

The title says "add bytecode size limit with graceful interpreter fallback" but this PR is actually the entire Tokamak JIT subsystem — ~18,600 lines across 50+ files with 60+ commits. This needs to be split into reviewable chunks (infrastructure/types, JIT dispatch, compiler backend, CI workflows, docs, benchmarks, etc.). Reviewing 18K lines as a single unit makes thorough review impractical and increases the risk of subtle issues slipping through.

Additional concerns beyond the inline comments:

**`.claude/commands/` committed to repo** — 11 files (in Korean) under `.claude/commands/` are personal Claude Code developer commands. These are local development tooling and shouldn't be checked into the shared repository — similar to how `.claude/settings.local.json` stays local.

**`handle_jit_subcall` duplicates ~300 lines of call/create logic** — The CALL and CREATE handlers in `handle_jit_subcall` manually duplicate depth checking, balance checking, precompile detection, BAL checkpoints, EIP-7708 log emission, nonce management, collision checking, and EIP-3860 initcode limits from `generic_call`/`generic_create`. Any future changes to those paths must be manually mirrored here, which is a significant maintenance risk. Consider extracting shared logic.

**`docs/tokamak/context/volkov-reviews.md`** contains AI agent review transcripts — this is project scaffolding that shouldn't live in the repository.

**Dead code: `oversized_hashes`** — Since `max_bytecode_size` defaults to 24576 (EIP-170 limit) and all deployed bytecodes post–Spurious Dragon are ≤24576, the `is_bytecode_oversized()` check (`len > 24576`) never triggers. The entire oversized negative cache is a no-op for any bytecode the JIT would encounter in practice.

PR #6262, opened 2026-02-25
perf(l1): doubly pipelined merkelization

iovoid:
**Motivation**

Currently while merkelizing, the main thread has to hash every storage key to dispatch it, and do the storage root collection itself. By moving this job to the account workers we can save time.

This shows a decrease in latency 29.3ms->29.1ms while keeping throughput identical.

**Description**

This PR has:
- 16 account workers tasked with handling routing updates to storage workers
- 16 storage workers that handle updates to the corresponding storage shard, like we already had
- pre-collection of account/storage tries when the work queue is empty


crates/blockchain/blockchain.rs
```diff
@@ -536,35 +543,60 @@ impl Blockchain {
...
             let (tx, rx) = channel();
-            let handle = std::thread::Builder::new()
-                .name(format!("block_executor_merkleization_shard_worker_{i}"))
-                .spawn_scoped(scope, move || {
-                    self.handle_merkleization_subtrie(rx, parent_header, i)
+            let account_senders = account_workers_tx.clone();
+            let storage_clone = self.storage.clone();
+            std::thread::Builder::new()
+                .name(format!("block_executor_storage_shard_{i}"))
+                .spawn(move || {
```
ElFantasma, line 572: The old code used `spawn_scoped`, which auto-joins workers when the scope exits — propagating panics and letting `Result`s be collected. The new code uses `spawn` without storing `JoinHandle`s, so:

1. **Worker panics are invisible.** If a worker hits one of the `.expect()` calls below (or any other panic), the thread silently vanishes. The main thread sees only a channel `RecvError`.
2. **`StoreError`s are lost.** Workers return `Result<(), StoreError>`, but nobody joins the thread, so errors are silently dropped.
3. **Possible deadlock.** If a storage worker panics mid-drain (after `CollectStorages`), it may have sent `StorageShard` for some prefixes but not all. An account worker with `pending_storage_accounts > 0` defers `CollectState` and keeps waiting for shards that will never arrive. Since the other storage workers still hold senders for the account channel, it stays open — the main thread blocks on `gatherer_rx` forever.

Consider storing the `JoinHandle`s and joining them after the gatherer loop completes (or when a channel error is detected) to propagate the real error. Alternatively, if scoped threads are feasible (everything captured is owned/cloned), passing the scope through would restore the old safety guarantees.

crates/blockchain/blockchain.rs
```diff
@@ -2649,6 +2481,359 @@ impl Blockchain {
...
+            AccountRequest::ProcessAccount {
+                prefix,
+                info,
+                storage: account_storage,
+                removed,
+                removed_storage,
+            } => {
+                let senders = storage_workers_tx
+                    .as_ref()
+                    .expect("ProcessAccount after FinishRouting");
```
ElFantasma: nit: This and the other `.expect()` calls in the worker functions (`"just loaded"` at L2686/L2703, `"shard without state"` at L2764, `"loaded in ProcessAccount"` at L2781) would panic the detached thread on invariant violations — invisibly, due to the issue above. Converting to `.ok_or(StoreError::Custom(...))?` would at least propagate the failure via channel closure rather than a silent thread death.

crates/blockchain/blockchain.rs
iovoid: Having the threads detached instead of joined is intentional, to avoid immediately paying for the cost of dropping the resources inside.

What do you think about having a (detached) thread that joins the worker threads (using a joinset) and, if one errored, sends a message to the others instructing them to quit? This should prevent the deadlock scenario too, while keeping the main thread free from repetitive error handling logic.

---

Alternatively, another approach I've been investigating is having a global dropper thread that handles dropping heavy objects. This way I could go back to joining directly without huge costs.

crates/blockchain/blockchain.rs
iovoid: Done in c482b4c60

crates/blockchain/blockchain.rs
ElFantasma: Good question. I traced the channel lifecycle to evaluate each option concretely.

**Context on why detaching matters:** `JoinHandle::join()` blocks until the thread function returns, which includes running all local destructors. So with scoped/joined threads the main thread waits for the workers' `Trie` + `FxHashMap` drops before it can start the next block batch. With detached threads those drops happen concurrently. That's a real pipeline benefit.

### What happens today if a worker fails

Each account channel has **17 senders** (1 from main + 16 from storage workers holding cloned vecs). When a storage worker panics/errors during finalization:

1. Its 16 `Sender<AccountRequest>` drop (thread exits)
2. Remaining 15 storage workers finish and drop theirs
3. Main drops `account_workers_tx` after sending `CollectState`
4. All senders gone → account channels close → workers exit their `for msg in rx` loop
5. Account workers with `pending_storage_accounts > 0` exit **without** sending `CollectedStateMsg`
6. Main's `gatherer_rx` loop gets < 16 results → incomplete root → block rejected with a confusing "wrong state root" error

So it's not a permanent hang (channels eventually close), but the real failure cause is invisible. The only true deadlock scenario is a storage worker getting stuck on a trie/RocksDB operation that never returns — extremely unlikely.

### Evaluating the options

**Option A (monitor thread):** Can't cleanly break a deadlock — the monitor doesn't hold the channel senders, so it can't close them to unblock account workers. Would need `recv_timeout` + shared atomic flags in workers, which restructures the loops significantly. High complexity for a very unlikely scenario.

**Option B (dropper thread):** The most principled option. Workers `std::mem::take` heavy objects, ship them to a drop channel, then return quickly — so joining is fast. You get full error propagation via scoped/joined threads while keeping the pipeline overlap. Worth doing if drops are actually on the critical path, but I'd measure first to justify the complexity.

**My suggestion — Option D (minimal pragmatic fix):** Keep detached, but:
1. **Log errors in the spawn closure** so failures aren't completely invisible:
   ```rust
   .spawn(move || {
       if let Err(e) = handle_storage_subtrie(...) {
           tracing::error!("storage shard worker {i} failed: {e}");
       }
   })
   ```
2. Optionally, use `recv_timeout` in account workers' deferred-shard wait to turn the (near-impossible) deadlock into a clear error instead of a silent hang.

This gives you 90% of the safety benefit with minimal code changes, and the detached-thread pipeline overlap is preserved. Option B is the right long-term answer if you want full error propagation, but given how unlikely worker failures are after the `.expect()` cleanup, D seems like the right tradeoff for now.

iovoid:
Superceded by #6278

PR #6264, opened 2026-02-25
fix(l2): reserve L1 gas upfront in prepare_execution to prevent DA fee griefing

### LINKED ISSUE #6053: L2: Validate gas_limit covers L1 DA fee to prevent griefing attack
## Problem

When a user sends a transaction with `gas_limit` exactly equal to intrinsic gas (e.g., 21000 for a simple transfer), the L1 fee vault receives **zero payment**, allowing a griefing attack against the sequencer.

### Step 1: Calculate intended L1 gas (in `finalize_non_privileged_execution`)

```rust
// l1_fee (in wei) = (l1_fee_per_blob_gas * GAS_PER_BLOB / SAFE_BYTES_PER_BLOB) * tx_size
let l1_fee = calculate_l1_fee(fee_config, tx_size)?;

// l1_gas = l1_fee / gas_price (convert wei to gas units)
let l1_gas = l1_fee / vm.env.gas_price;
```

### Step 2: Check if total gas exceeds limit

```rust
let total_gas = actual_execution_gas + l1_gas;

if total_gas > gas_limit {
    // Transaction reverts, but is still included in block/blob
    // Recalculate l1_gas as the "remaining gas" after execution:
    l1_gas = gas_limit - actual_execution_gas;  // <-- THIS IS THE PROBLEM
    total_gas = gas_limit;
}
```

### Step 3: Pay L1 fee vault

```rust
// L1 fee vault receives: l1_gas * gas_price
let l1_fee_payment = l1_gas * gas_price;
vm.increase_account_balance(l1_fee_vault, l1_fee_payment)?;
```

## The Problem

When `gas_limit == intrinsic_gas` (e.g., 21000 for simple transfer):

1. `actual_execution_gas = intrinsic_gas = 21000`
2. `l1_gas` (intended) might be 5000 based on tx size
3. `total_gas = 21000 + 5000 = 26000 > gas_limit (21000)` → tx reverts
4. **Recalculated** `l1_gas = gas_limit - actual_execution_gas = 21000 - 21000 = 0`
5. L1 fee vault receives: `0 * gas_price = 0`

**But the transaction IS included in the blob**, so the sequencer pays L1 DA cost and gets nothing back.

## Griefing Attack

A malicious user can spam transactions with `gas_limit = intrinsic_gas`:
- Each tx passes validation (gas_limit ≥ intrinsic_gas ✓)
- Each tx gets included in blob
- Sequencer pays L1 DA for each tx
- L1 fee vault receives 0 for each tx
- **Sequencer loses money on every transaction**

## Proposed Solution

Add validation in `prepare_execution` **before** execution:

```rust
fn validate_gas_limit_covers_l1_fee(
    vm: &VM<'_>,
    l1_fee_config: &Option<L1FeeConfig>,
) -> Result<(), TxValidationError> {
    let l1_gas = calculate_l1_fee_gas(vm, l1_fee_config)?;
    if l1_gas == 0 {
        return Ok(()); // No L1 fee configured
    }
    
    let intrinsic_gas = vm.get_intrinsic_gas()?;
    let min_gas_limit = intrinsic_gas + l1_gas;
    
    if vm.current_call_frame.gas_limit < min_gas_limit {
        // Reject transaction - not enough gas to cover L1 DA fee
        return Err(TxValidationError::IntrinsicGasTooLow);
    }
    Ok(())
}
```

This ensures: **`gas_limit >= intrinsic_gas + l1_gas`** or transaction is rejected.

### Deposits (Privileged Transactions)

Deposits from L1 create privileged transactions. Need to ensure:
- Deposit fee on L1 accounts for L2's L1 DA cost
- Privileged tx `gas_limit` includes L1 DA gas
- Define: does depositor or sequencer pay L1 DA?

### L2-to-L2 Messages

Cross-L2 messages create privileged transactions. Same considerations:
- Message fee must cover L1 DA cost
- Privileged tx builder must set adequate `gas_limit`

### Regular Transactions
- [ ] `gas_limit < intrinsic_gas + l1_gas` → rejected in `prepare_execution`
- [ ] Test: tx with `gas_limit = intrinsic_gas` fails when L1 fee configured
- [ ] Test: tx with `gas_limit = intrinsic_gas + l1_gas` succeeds

### Deposits & L2-to-L2
- [ ] Define fee model for L1 DA cost on privileged txs
- [ ] Privileged tx builders set adequate `gas_limit`

## Related

Test that discovered this: `test_l1_gas_revert_with_minimal_gas_limit_l1_vault_gets_zero` in PR #6051

avilagaston9:
## Motivation

A user can send a transaction with `gas_limit = intrinsic_gas` (e.g., 21000). The tx passes validation, gets included in the blob, but at finalize time `actual_gas + l1_gas > gas_limit` triggers a revert. The recalculated `l1_gas = gas_limit - actual_gas = 0`, so the L1 fee vault receives nothing. The sequencer pays DA costs and gets nothing back — a griefing vector.

Closes #6053

## Description

Reserve `l1_gas` from `gas_remaining` during `prepare_execution` by calling `increase_consumed_gas(l1_gas)`, so execution physically cannot consume the L1 fee budget. This replaces the previous validate-only approach (`validate_gas_limit_covers_l1_fee`) which rejected transactions where `gas_limit < intrinsic_gas + l1_gas` but didn't prevent a contract from burning enough execution gas to squeeze the L1 fee at finalize time.

With reservation, the L1 fee vault always receives the full `l1_gas` payment:
- If `gas_limit < intrinsic_gas + l1_gas`, the tx is rejected upfront with `IntrinsicGasTooLow`
- The finalize revert path is removed (dead code — `actual_gas_used <= gas_limit` is guaranteed)
- The refund cap (EIP-3529) is applied to execution gas only since `l1_gas` is not refundable

### Design note: `calculate_l1_fee_gas` is called in both prepare and finalize

`calculate_l1_fee_gas` is computed twice: once in `reserve_l1_gas` (prepare) and once in `finalize_non_privileged_execution` (finalize). Both calls use `vm.tx` and `vm.env.gas_price` which are immutable during execution, so they always return the same value. An alternative would be storing the reserved `l1_gas` on the VM/environment to avoid recomputation, but we intentionally keep the current approach to avoid modifying the environment struct for now.

## How to Test

```bash
cargo test -p ethrex-test -- levm::l2_gas_reservation
cargo clippy -p ethrex-levm -- -D warnings
cargo fmt --all -- --check
```

## Checklist

### DIFF
--- a/crates/vm/levm/src/hooks/l2_hook.rs
+++ b/crates/vm/levm/src/hooks/l2_hook.rs
@@ -388,15 +388,11 @@ fn prepare_execution_privileged(vm: &mut VM<'_>) -> Result<(), crate::errors::VM
         let value = vm.current_call_frame.msg_value;
         if value > sender_balance {
             tx_should_fail = true;
-        } else {
-            // This should never fail, since we just checked the balance is enough.
-            vm.decrease_account_balance(sender_address, value)
-                .map_err(|_| {
-                    InternalError::Custom(
-                        "Insufficient funds in privileged transaction".to_string(),
-                    )
-                })?;
         }
+        // NOTE: Balance is NOT debited here — it is deferred until after all
+        // validation passes. This avoids permanent fund loss if a later check
+        // (e.g. intrinsic gas) triggers the failure path, which zeroes msg_value
+        // and makes undo_value_transfer a no-op.
     }
 
     // if fork > prague: default_hook::validate_min_gas_limit
@@ -452,6 +448,17 @@ fn prepare_execution_privileged(vm: &mut VM<'_>) -> Result<(), crate::errors::VM
             jump_targets: Vec::new(),
         })?;
         return Ok(());
+    }
+
+    // Debit sender balance now that all validation has passed.
+    // This must happen after the tx_should_fail checks above to avoid
+    // permanent fund loss when the failure path zeroes msg_value.
+    if sender_address != COMMON_BRIDGE_L2_ADDRESS {
+        let value = vm.current_call_frame.msg_value;
+        vm.decrease_account_balance(sender_address, value)
+            .map_err(|_| {
+                InternalError::Custom("Insufficient funds in privileged transaction".to_string())
+            })?;
     }
 
     default_hook::transfer_value(vm)?;
@@ -720,6 +727,7 @@ fn simulate_common_bridge_call(
         &tx,
         LevmCallTracer::disabled(),
         VMType::L2(Default::default()),
+        vm.crypto,
     )?;
     new_vm.hooks = vec![];
     default_hook::set_bytecode_and_code_address(&mut new_vm)?;
--- a/.github/scripts/generate_combined_report.sh
+++ b/.github/scripts/generate_combined_report.sh
@@ -42,6 +42,7 @@ if [[ -f "$LOC_JSON" ]]; then
 
   pct() {
     # Integer percentage: pct numerator denominator
+    [[ "$2" -eq 0 ]] && { printf "0"; return; }
     printf "%d" $(( ($1 * 100 + $2 / 2) / $2 ))
   }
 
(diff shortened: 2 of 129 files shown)


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -103,40 +106,29 @@ fn finalize_non_privileged_execution(
...
+    // ctx_result.gas_used includes l1_gas (reserved in prepare_execution).
+    // Separate execution gas for refund calculation — l1_gas is not refundable.
+    let execution_gas_pre_refund = ctx_result
+        .gas_used
+        .checked_sub(l1_gas)
+        .ok_or(InternalError::Underflow)?;
 
-    let gas_refunded: u64 = default_hook::compute_gas_refunded(vm, ctx_result)?;
+    // Refund cap based on execution gas only (EIP-3529)
+    let gas_refunded: u64 = vm
```
ElFantasma, line 119: Good that this doesn't use `compute_gas_refunded(vm, ctx_result)` — that helper caps refunds at `ctx_result.gas_used / 5`, which now includes `l1_gas` since it was consumed from `gas_remaining`. Using `execution_gas_pre_refund` here correctly excludes `l1_gas` from the denominator.

Worth noting for future maintainability: a "simplification" to call `compute_gas_refunded` would silently inflate the refund cap by `l1_gas / 5`, letting users reclaim gas that should cover the L1 fee.

ElFantasma (approved):

PR #6265, opened 2026-02-25
perf(l1,levm): use nested storage originals, fxhashmap callframe backup, and sstore helper

ilitteri:
**Motivation**

This change set focuses on concrete internal optimizations in LEVM and L1 storage access paths, especially around storage/account read bookkeeping and SSTORE original-value tracking.

**Description**

The PR bundles these optimizations:

- Added `Store::get_storage_at_root_with_known_storage_root(...)` in `crates/storage/store.rs`.
  - This allows direct storage-trie reads when account hash and storage root are already known, avoiding repeated state-trie account lookup and account RLP decode in that path.

- Optimized `StoreVmDatabase` read path in `crates/blockchain/vm.rs`.
  - Added memoized account-state entries (`state` + precomputed `hashed_address`) behind `RwLock<FxHashMap<...>>`.
  - Updated `get_account_state` to use this cache.
  - Updated `get_storage_slot` to use cached account data plus the new known-storage-root store helper.

- Optimized SSTORE original-value handling in LEVM (`crates/vm/levm/src/vm.rs`, `crates/vm/levm/src/db/gen_db.rs`, `crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs`).
  - Replaced flat `(Address, H256) -> U256` storage-original map with nested `Address -> (H256 -> U256)`.
  - Added `access_storage_slot_for_sstore(...)` to return `(current_value, original_value, was_cold)` in a single helper.
  - Updated `op_sstore` to use this dedicated helper.

- Switched call-frame backup maps to `FxHashMap` in `crates/vm/levm/src/call_frame.rs`.

- Tightened LEVM storage miss cache-fill logic in `get_storage_value` to update the current account storage map directly.

**Validation**

- Built and validated with `cargo check`/`cargo build` on the touched crates.
- Timed block-execution runs showed consistent improvements on the tested branch compared to `main` for the selected optimization set.

**Checklist**


crates/vm/levm/src/db/gen_db.rs
```diff
@@ -673,8 +696,12 @@ impl<'a> VM<'a> {
 
         let value = self.db.get_value_from_database(address, key)?;
 
-        // Update the account with the fetched value
-        let account = self.get_account_mut(address)?;
+        // Cache-fill only: this is a read-path miss, not a state mutation.
+        let account = self
```
ElFantasma: Good catch: the old path called `self.get_account_mut(address)?` which invoked `mark_modified()` and created a call frame backup entry. A read-path cache fill should have neither side effect — marking the account `Modified` could cause it to appear in state transitions even if nothing was actually written.

ilitteri:
4ba40540f7

ilitteri:
3b4d370f1a

PR #6266, opened 2026-02-25
feat(l1): add two-digit precision to block timing logs

iovoid:
**Motivation**

We often have blocks where each step takes single-digit milliseconds, where having more detail would be useful. There is a big difference between a step taking 0.5ms and 1.4ms.

**Description**

Adds two decimal places to every timing logged.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -2022,9 +2022,9 @@ impl Blockchain {
                 METRICS_BLOCKS.set_latest_gas_used(gas_used as f64);
                 METRICS_BLOCKS.set_latest_block_gas_limit(gas_limit as f64);
                 METRICS_BLOCKS.set_latest_gigagas(throughput);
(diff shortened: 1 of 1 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -2132,16 +2138,16 @@ impl Blockchain {
         ];
         let bottleneck = phases
             .iter()
-            .max_by_key(|(_, ms)| ms)
+            .max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
```
ElFantasma, line 2141: nit: `f64::total_cmp` gives a total ordering without needing `unwrap_or` and handles NaN deterministically (sorts above infinity). Slightly cleaner:
```rust
.max_by(|(_, a), (_, b)| a.total_cmp(b))
```
These values can't be NaN anyway, but `total_cmp` communicates that more directly.

ElFantasma (approved):

PR #6275, opened 2026-02-28
fix(l1): stop permanently invalidating blocks on transient EVM errors

### LINKED ISSUE #6274: Block 24548619 invalid
I am running grandine with ethrex on mainnet and this morning my instance of grandine went offline with the following error:

`WARN eth1_api::execution_service:209: engine_newPayload returned INVALID status (beacon_block_root: 0x5a7b46ce2d6e3e315e4976a440e28140627df503800aaade9202cab3816f563e, block_number: 24548619, block_hash: 0x1311add13604cd7870d137cce45a03190000c445df382668b67547db11c136d3, response: PayloadStatusV1 { status: Invalid, latest_valid_hash: Some(0x85ab277d35c34c4d9822f770a64faa842c7aee4b03e3842c4004a470870c6ae9), validation_error: Some("Header has been previously invalidated.") }) peers=[2/100]`

`WARN fork_choice_control::mutator:2384: block rejected (error: block has invalid execution payload, block root: 0x5a7b46ce2d6e3e315e4976a440e28140627df503800aaade9202cab3816f563e, origin: Requested(Some(PeerId("...")))) peers=[26/100]`

https://github.com/grandinetech/grandine/issues/617#issue-4001731321

ilitteri:
## Motivation

Closes #6274.

Block 24548619 is a valid, finalized mainnet block that ethrex permanently rejected. A user running Grandine + ethrex on mainnet had their node go offline because ethrex returned `INVALID` with `"Header has been previously invalidated."` for every subsequent `engine_newPayload`, with no way to recover short of wiping the database.

## Root Cause

`try_execute_payload()` in `payload.rs` treated `ChainError::EvmError` the same as `ChainError::InvalidBlock` — both permanently cached the block as invalid in the `INVALID_CHAINS` RocksDB column family via `set_latest_valid_ancestor()`.

However, `EvmError` includes **transient** infrastructure errors:
- `DB("state root missing")` — parent state not committed yet
- `DB("Code not found")` — code not fetched during sync
- `Custom("LockError")` — poisoned mutex
- Any `VMError::Internal` propagated as `EvmError::Custom`

The `From<EvmError> for ChainError` conversion (`error.rs:41-53`) already routes genuine consensus violations (`EvmError::Transaction`, `EvmError::InvalidDepositRequest`) to `ChainError::InvalidBlock`. Everything remaining in `ChainError::EvmError` is transient and should never trigger permanent invalidation.

Once cached, `validate_ancestors()` rejects the block (and cascades to all descendants) on every future `engine_newPayload` without attempting re-execution. There was no mechanism to clear the cache.

## Description

- **`payload.rs`**: The `ChainError::EvmError` arm in `try_execute_payload()` now returns `Err(RpcErr::Internal(...))` instead of caching in `INVALID_CHAINS`. The CL will retry the payload. Only `ChainError::InvalidBlock` triggers permanent invalidation.
- **`store.rs`**: Added `delete_latest_valid_ancestor()`, `clear_invalid_chains()`, and a `delete_async()` helper for managing the invalid chains cache.
- **`initializers.rs`**: `open_store()` now calls `clear_invalid_chains()` on startup, so existing nodes with false entries recover automatically after updating.
- **Tests**: 5 new tests covering `INVALID_CHAINS` set/get/delete operations and the core invariant that only `InvalidBlock` caches while `EvmError` does not.

## Recovery for affected users

**Just update and restart.** The `INVALID_CHAINS` table is cleared on every startup, so any falsely invalidated blocks will be re-executed normally. No database wipe or manual intervention needed.

## How to Test

```bash
cargo test -p ethrex-storage
```

Validates all 5 new INVALID_CHAINS tests including the caching invariant.

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -158,15 +158,19 @@ pub async fn load_store(datadir: &Path) -> Result<Store, StoreError> {
 
 /// Opens a pre-existing Store or creates a new one
 pub fn open_store(datadir: &Path) -> Result<Store, StoreError> {
-    if is_memory_datadir(datadir) {
-        Store::new(datadir, EngineType::InMemory)
+    let store = if is_memory_datadir(datadir) {
+        Store::new(datadir, EngineType::InMemory)?
     } else {
         #[cfg(feature = "rocksdb")]
         let engine_type = EngineType::RocksDB;
         #[cfg(feature = "metrics")]
         ethrex_metrics::process::set_datadir_path(datadir.to_path_buf());
-        Store::new(datadir, engine_type)
-    }
+        Store::new(datadir, engine_type)?
+    };
+    // Clear any false positives from a previous bug where transient EVM errors
+    // were cached as permanent block invalidity (see #6274).
+    store.clear_invalid_chains()?;
+    Ok(store)
 }
 
 pub fn init_blockchain(store: Store, blockchain_opts: BlockchainOptions) -> Arc<Blockchain> {
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -1353,6 +1353,13 @@ impl Store {
         self.backend.clear_table(FULLSYNC_HEADERS)
     }
 
+    /// Clears all entries from the invalid chains cache.
+    /// Called on startup to purge any false positives from a previous bug
+    /// where transient errors were cached as permanent block invalidity.
+    pub fn clear_invalid_chains(&self) -> Result<(), StoreError> {
+        self.backend.clear_table(INVALID_CHAINS)
+    }
+
     /// Delete a key from a table
     pub fn delete(&self, table: &'static str, key: Vec<u8>) -> Result<(), StoreError> {
         let mut txn = self.backend.begin_write()?;


crates/networking/rpc/engine/payload.rs
```diff
@@ -1076,15 +1076,8 @@ async fn try_execute_payload(
...
-            warn!("Error executing block: {error}");
-            context
-                .storage
-                .set_latest_valid_ancestor(block_hash, latest_valid_hash)
-                .await?;
-            Ok(PayloadStatus::invalid_with(
-                latest_valid_hash,
-                error.to_string(),
-            ))
+            error!("Internal error executing block {block_hash}: {error}");
```
ElFantasma, line 1079: Not all `EvmError` variants are transient. `EvmError::SystemContractCallFailed` is explicitly a consensus violation — per EIP-7002 and EIP-7251, a system contract revert *invalidates the entire block* (the code even says so: `"According to EIP-7002, the revert of this system call invalidates the block."`). `EvmError::Header` (missing `parent_beacon_block_root`) also indicates a genuinely invalid block.

With this change, those cases return `Err(RpcErr::Internal)` → the CL retries the block indefinitely instead of learning it's invalid.

Suggestion: update the `From<EvmError> for ChainError` conversion in `crates/blockchain/error.rs` to also route `SystemContractCallFailed` and `Header` to `ChainError::InvalidBlock`. Then only truly transient variants (`DB`, `Custom`, `InvalidEVM`, `Precompile`) remain as `ChainError::EvmError`, making this non-caching arm correct:

```rust
impl From<EvmError> for ChainError {
    fn from(value: EvmError) -> Self {
        match value {
            EvmError::Transaction(err) => {
                ChainError::InvalidBlock(InvalidBlockError::InvalidTransaction(err))
            }
            EvmError::InvalidDepositRequest => ChainError::InvalidBlock(
                InvalidBlockError::InvalidTransaction("Invalid deposit request layout".to_string()),
            ),
            EvmError::Header(err) => {
                ChainError::InvalidBlock(InvalidBlockError::InvalidTransaction(err))
            }
            EvmError::SystemContractCallFailed(err) => {
                ChainError::InvalidBlock(InvalidBlockError::InvalidTransaction(err))
            }
            other_errors => ChainError::EvmError(other_errors),
        }
    }
}
```

(You'd want a better `InvalidBlockError` variant than `InvalidTransaction` for these — maybe `InvalidBlockError::SystemContractFailed` and `InvalidBlockError::InvalidHeader` — but the routing is the important part.)

PR #6276, opened 2026-02-28
refactor(l2): abstract EVM crypto behind Crypto trait with dependency injection

ilitteri:
## Summary

- Introduces a `Crypto` trait (`crates/common/crypto/provider.rs`) covering all 20 EVM crypto operations (precompiles + tx-level ECDSA), with a `CryptoError` enum that doesn't leak library types.
- Implements `NativeCrypto` (`crates/common/crypto/native.rs`) using the existing native libraries (secp256k1, ark-bn254, c-kzg, lambdaworks, etc.).
- Wires `&dyn Crypto` through the VM layer: `Evm` holds `Arc<dyn Crypto>`, `VM` holds `&dyn Crypto`, precompiles receive it as a parameter.
- Rewrites all precompile functions to call `crypto.method()` instead of direct library imports — removes **all** `#[cfg(feature = "sp1/risc0/zisk/secp256k1")]` gates and duplicate implementations from levm.
- Migrates `Transaction::sender()` and EIP-7702 recovery to use `&dyn Crypto`, deleting the old `EcdsaError` type and dual cfg-gated `recover_address` functions.
- Creates zkVM-specific `Crypto` implementations in `guest-program` (Sp1Crypto, Risc0Crypto, ZiskCrypto, OpenVmCrypto) that restore the original k256/substrate-bn implementations per target, with NativeCrypto fallback for methods that had no zkVM-specific version.
- Migrates `get_address_from_secret_key` (`ethrex-l2-common`) from `secp256k1` (optional feature) to `k256` (unconditional dependency), fixing compilation for backends like TDX that depend on `ethrex-l2-common` with `default-features = false`.

**Result**: levm depends on **zero** crypto libraries — the binary brings the implementation. ~890 lines of duplicated cfg-gated code across 14 sites in 7 files replaced by a single trait dispatch.

### Behavioral changes

While this is primarily a refactor, it includes the following intentional behavioral changes:

- **Blake2f assembly fix (x86_64):** The old `pxor xmm3, [rcx]` (legacy SSE) zeroed the upper 128 bits of `ymm3`, corrupting `v[14..15]` of the Blake2b state. Fixed with `vmovdqu xmm12, [rcx]; vpxor ymm3, ymm3, ymm12` to preserve the full register. This corrects Blake2f results for inputs where `v[14..15]` are nonzero.
- **`substrate_bn_g1_mul` validation order (SP1/ZisK only):** The old SP1/ZisK code checked `scalar.is_zero()` before validating the point is on the curve, silently returning zeros for off-curve points with a zero scalar. The new code validates the point first, returning `InvalidPoint` for off-curve points regardless of the scalar. This is more correct per EIP-196.
- **`PRECOMPILES` array dedup:** The old array had `BLS12_MAP_FP_TO_G1` listed twice (19 elements). Fixed to 18 elements with correct ordering. No impact on precompile dispatch (which uses address-indexed lookup), but fixes iteration-based consumers.

### zkVM crypto provider mapping

| Method | SP1 | RISC0 | ZisK | OpenVM | Native |
|--------|-----|-------|------|--------|--------|
| ECDSA (ecrecover, recover_signer) | k256 | k256 | k256 | k256 | secp256k1 (C) |
| bn254_g1_add | NativeCrypto | NativeCrypto | substrate-bn | NativeCrypto | ark-bn254 |
| bn254_g1_mul | substrate-bn | NativeCrypto | substrate-bn | NativeCrypto | ark-bn254 |
| bn254_pairing | substrate-bn | substrate-bn | substrate-bn | NativeCrypto | ark-bn254 |
| Everything else | NativeCrypto | NativeCrypto | NativeCrypto | NativeCrypto | native libs |

## Test plan

- [x] `cargo test --workspace` — 316 tests pass, 0 failures
- [x] Zero `todo!()` in guest-program crypto stubs
- [x] Zero `#[cfg(feature = "sp1/risc0/zisk/secp256k1")]` gates in levm or common
- [x] Zero references to `EcdsaError` in codebase
- [ ] EF tests pass
- [ ] SP1/RISC0 guest binary compilation (blocked by pre-existing blst RISC-V issue, not a regression)

### DIFF
--- a/crates/common/crypto/provider.rs
+++ b/crates/common/crypto/provider.rs
@@ -31,6 +31,25 @@ pub enum CryptoError {
 ///
 /// Methods take `&self` to support `&dyn Crypto` (dynamic dispatch).
 /// Implementations are typically zero-sized structs.
+///
+/// # zkVM implementors
+///
+/// The following methods **must** be overridden for zkVM targets, as their
+/// default implementations use native C libraries (secp256k1, ark-bn254, etc.)
+/// that cannot run inside a zkVM guest:
+///
+/// - [`secp256k1_ecrecover`](Crypto::secp256k1_ecrecover) — uses `libsecp256k1` C library
+/// - [`recover_signer`](Crypto::recover_signer) — uses `libsecp256k1` C library
+/// - [`bn254_g1_add`](Crypto::bn254_g1_add), [`bn254_g1_mul`](Crypto::bn254_g1_mul),
+///   [`bn254_pairing_check`](Crypto::bn254_pairing_check) — use `ark-bn254`
+/// - [`bls12_381_g1_add`](Crypto::bls12_381_g1_add), [`bls12_381_g2_add`](Crypto::bls12_381_g2_add),
+///   [`bls12_381_g1_msm`](Crypto::bls12_381_g1_msm), [`bls12_381_g2_msm`](Crypto::bls12_381_g2_msm),
+///   [`bls12_381_pairing_check`](Crypto::bls12_381_pairing_check) — use `bls12_381` crate
+///   [`bls12_381_map_fp_to_g1`](Crypto::bls12_381_map_fp_to_g1),
+///   [`bls12_381_map_fp2_to_g2`](Crypto::bls12_381_map_fp2_to_g2) — use `bls12_381` crate
+///
+/// Non-overridden methods will silently use the native default, which will
+/// fail to compile or panic at runtime inside a zkVM guest.
 pub trait Crypto: Send + Sync + core::fmt::Debug {
     // ── ECDSA (secp256k1) ──────────────────────────────────────────────
 


crates/common/crypto/provider.rs
```diff
@@ -0,0 +1,762 @@
...
+    fn bn254_g1_add(&self, p1: &[u8], p2: &[u8]) -> Result<[u8; 64], CryptoError> {
+        use ark_bn254::Fq;
+        use ark_ec::CurveGroup;
+        use ark_ff::{BigInteger, PrimeField as _, Zero};
+
+        let parse_point = |bytes: &[u8]| -> Result<ark_bn254::G1Affine, CryptoError> {
+            if bytes.len() < 64 {
+                return Err(CryptoError::InvalidInput("G1 point must be 64 bytes"));
+            }
+            let x = Fq::from_be_bytes_mod_order(&bytes[..32]);
```
ElFantasma, line 164: `from_be_bytes_mod_order` silently reduces values >= the BN254 field modulus to their remainder. The substrate-bn equivalent in `shared.rs` (`Fq::from_slice`) returns an error for out-of-range inputs instead.

This means the native (ark) and zkVM (substrate-bn) implementations have asymmetric input validation: the same invalid coordinate would produce a (wrong) result on native but an error on zkVM.

Today this is harmless because the precompile layer validates coordinates `< ALT_BN128_PRIME` before calling into the trait (`validate_bn254_g1_coords` in `precompiles.rs`). But if the trait is called from a new context without that guard, the paths would silently diverge.

Consider using `Fq::from_be_bytes` (strict) + explicit error instead of `from_be_bytes_mod_order` here and in `bn254_g1_mul` (line 184-185), to make the trait self-validating. Same applies to `bn254_pairing_check`'s `from_le_bytes_mod_order` calls.

crates/common/crypto/provider.rs
```diff
@@ -0,0 +1,762 @@
...
+        let mut result = G2Projective::identity();
+
+        for (point_bytes, scalar_bytes) in pairs {
+            let point = parse_bls12_g2(*point_bytes)?;
+            if !bool::from(point.is_torsion_free()) {
+                return Err(CryptoError::InvalidPoint("G2 point not in subgroup"));
+            }
+            let scalar = parse_bls12_scalar(scalar_bytes);
+
+            if scalar != bls12_381::Scalar::zero() {
```
ElFantasma: nit: This uses `scalar != bls12_381::Scalar::zero()` while the equivalent check in `bls12_381_g1_msm` (line 516) uses `!bool::from(scalar.is_zero())`. Both work, but it'd be cleaner to pick one style — `!bool::from(scalar.is_zero())` is more idiomatic for `subtle::CtOption`-based types.

avilagaston9:
**@edg-l's review addressed:**

1. **`parse_bls12_scalar` Montgomery form**: Pre-existing pattern from `main`. Created #6346.
2. **`bn254_pairing_check` allocations**: Fixed in 5f3a01bba.
3. **`ecrecover` high-S**: No change needed — the old `normalize_s()` was a k256 workaround, not EVM validation. The secp256k1 C lib handles full S range natively. `recover_signer` still rejects high-S for tx validation.
4. **`crypto_error_to_precompile` wildcard**: Fixed in 55aede881.
5. **G1/G2 MSM zero-check**: Fixed in 7b049b006.
6. **Default impl on trait**: Added zkVM doc comment in cac21aa45.

crates/common/crypto/provider.rs
avilagaston9: Pre-existing pattern from main (precompiles.rs:816-819). Created #6347 to track.

crates/common/crypto/provider.rs
avilagaston9: Fixed in 7b049b006.

ElFantasma (approved):

PR #6278, opened 2026-03-02
perf(l1): doubly pipelined merkelization

iovoid:
## Motivation

  In the current merkelization pipeline on `main`, the dispatcher thread orchestrates three sequential
  phases:

  1. Send `LoadAccount` + `MerklizeStorage` to shard workers
  2. Send `CollectStorages`, wait for all storage roots, assemble account states
  3. Send `MerklizeAccount`, then `CollectState` for final results

  The dispatcher acts as a bottleneck between phases — all 16 workers idle while it collects storage roots
  and prepares account work. On heavy blocks this gap is measurable.

  ## Description

  Restructure merkelization so that shard workers self-coordinate, eliminating the dispatcher as intermediary
   between storage and account phases.

  ### Unified `ProcessAccount` message

  Replace the separate `LoadAccount` / `MerklizeStorage` / `MerklizeAccount` request types with a single
  `ProcessAccount` that carries account info + storage entries together. Workers handle storage trie updates
  internally.

  ### Worker-to-worker routing

  Each worker receives clones of all 16 senders. When a `ProcessAccount` contains storage slots that hash to
  a different shard, the worker routes them directly to the responsible peer — the dispatcher no longer fans
  out individual storage entries.

  ### `FinishRouting` / `RoutingDone` protocol

  Since workers now send to each other, they need to know when all peers are done routing. On `FinishRouting`
   from the dispatcher, each worker broadcasts `RoutingDone` to all 15 peers. Once a worker has received
  `RoutingDone` from all peers, it knows no more storage will arrive and can proceed to collect storage roots
   and merklize accounts autonomously.

  ### Early `MerklizeAccounts` for no-storage accounts

  Accounts with no storage changes are batched and sent to workers immediately after `FinishRouting`, so
  workers can begin account merklization without waiting for the `RoutingDone` protocol to complete.

  ### Deferred root deallocation

  The collapsed root node is sent to a background `DROP_SENDER` thread for deallocation, moving the dealloc
  cost off the critical path.

### Benchmarks

In our new servers we observed the merkelization overhead going down from 17% to 12%, resulting in a decrease of 0.4ms (1.3%) in total average block execution time.


crates/blockchain/blockchain.rs
```diff
@@ -536,35 +557,39 @@ impl Blockchain {
...
+        // Spawn 16 unified workers (each gets clone of all 16 senders)
+        for (i, rx) in workers_rx.into_iter().enumerate() {
+            let all_senders = workers_tx.clone();
+            let storage_clone = self.storage.clone();
+            std::thread::Builder::new()
+                .name(format!("block_executor_shard_{i}"))
+                .spawn(move || {
+                    handle_subtrie(storage_clone, rx, parent_state_root, i as u8, all_senders)
+                })
+                .map_err(|e| StoreError::Custom(format!("spawn failed: {e:?}")))?;
```
ElFantasma: Worker `JoinHandle`s are discarded here. The old code used `spawn_scoped`, which guaranteed panics in worker threads propagated to the caller. With `spawn`, the handle is dropped, detaching the thread — if a worker panics, the panic is silently lost and the dispatcher only sees a generic `RecvError` on the gatherer channel. This makes debugging significantly harder ("send error" instead of the actual panic message).

Suggest storing the 16 handles and joining them after `gatherer_rx` is consumed:

```rust
let mut handles = Vec::with_capacity(16);
for (i, rx) in workers_rx.into_iter().enumerate() {
    // ...
    let handle = std::thread::Builder::new()
        .name(...)
        .spawn(move || { ... })
        .map_err(|e| StoreError::Custom(format!("spawn failed: {e:?}")))?;
    handles.push(handle);
}
// ... dispatch + collect ...
for handle in handles {
    match handle.join() {
        Ok(result) => result?,
        Err(panic_payload) => {
            return Err(StoreError::Custom(
                format!("worker panicked: {panic_payload:?}")
            ));
        }
    }
}
```

This restores the panic propagation behavior of scoped threads while keeping the `move` ownership pattern.

crates/blockchain/blockchain.rs
```diff
@@ -536,35 +557,39 @@ impl Blockchain {
...
+            workers_rx.push(rx);
+        }
+
+        // Spawn 16 unified workers (each gets clone of all 16 senders)
+        for (i, rx) in workers_rx.into_iter().enumerate() {
+            let all_senders = workers_tx.clone();
+            let storage_clone = self.storage.clone();
+            std::thread::Builder::new()
+                .name(format!("block_executor_shard_{i}"))
+                .spawn(move || {
```
ElFantasma, line 616: These workers are now detached (`spawn` instead of `spawn_scoped`) — if one panics, its `handle_subtrie` silently disappears. The node stays correct because the gatherer loop at line ~695 will receive fewer than 16 `CollectedStateMsg` responses, producing a wrong state root that gets rejected during block validation. So this is a **diagnostic** concern, not a correctness one.

Still, a silent panic makes debugging painful. Two options:

**Option A — lightweight `catch_unwind` (minimal change):**
```suggestion
                .spawn(move || {
                    if let Err(panic) = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
                        handle_subtrie(storage_clone, rx, parent_state_root, i as u8, all_senders)
                    })) {
                        tracing::error!("merkelization worker {i} panicked: {panic:?}");
                    }
                })
```
This logs the panic payload without restructuring the collection flow. The worker still disappears, but you get the panic message in the logs.

**Option B — store handles and join (full propagation):**
```rust
let mut handles = Vec::with_capacity(16);
for (i, rx) in workers_rx.into_iter().enumerate() {
    let all_senders = workers_tx.clone();
    let storage_clone = self.storage.clone();
    let handle = std::thread::Builder::new()
        .name(format!("block_executor_shard_{i}"))
        .spawn(move || {
            handle_subtrie(storage_clone, rx, parent_state_root, i as u8, all_senders)
        })
        .map_err(|e| StoreError::Custom(format!("spawn failed: {e:?}")))?;
    handles.push(handle);
}
// ... after gatherer_rx collection ...
for handle in handles {
    match handle.join() {
        Ok(result) => result?,
        Err(panic) => {
            return Err(StoreError::Custom(format!("worker panicked: {panic:?}")));
        }
    }
}
```
This restores the panic propagation behavior of the old `spawn_scoped` approach, but requires moving the join loop after the gatherer collection and returning the error to the caller.

Option A is probably sufficient here — the state root mismatch already surfaces the failure at the block validation level.

ElFantasma (changes requested):

crates/blockchain/blockchain.rs
iovoid: I went with option B, which also prevents any kind of deadlock.

PR #6281, opened 2026-03-02
chore(l1): add rlp decode/encode benches (part 2)

azteca1998:
## Summary

- Expand decode benchmarks with common types (AccountInfo, AccountState, Withdrawal, Log, Receipt, BlockHeader), trie nodes (NodeHash::Inline, Leaf, Extension, Branch), and transactions (Legacy, EIP1559)
- Implement all encode benchmarks from scratch: scalars, bytes/strings, collections, tuples, IPs, common types, trie nodes, and transactions

> **Depends on #5414** (`rlp_benches_decode` — part 1)

## No part 3 needed

The remaining types were intentionally skipped as they don't exercise new RLP code paths:
- **EIP2930/EIP4844/EIP7702** — structurally near-identical to EIP1559, same encode/decode logic
- **BlobsBundle** — blobs are `[u8; 4096]`; benchmarks would measure bulk memcpy, not RLP
- **Block** — just `BlockHeader` + `BlockBody(Vec<Transaction>)`; marginal value over BlockHeader alone
- **ReceiptWithBloom** — dominated by the 256-byte Bloom field, not RLP-interesting
- **EncodedRequests** — not a clear public type

## Test plan
- [x] `cargo check -p ethrex-rlp --benches` compiles cleanly
- [x] `cargo bench -p ethrex-rlp --bench decode -- --test` — all benchmarks pass
- [x] `cargo bench -p ethrex-rlp --bench encode -- --test` — all benchmarks pass


crates/common/rlp/Cargo.toml
```diff
@@ -17,6 +17,8 @@ criterion = { version = "0.7.0", features = ["html_reports"] }
 ethrex-common.workspace = true
 ethrex-trie.workspace = true
 rand = "0.9.2"
+bytes.workspace = true
+once_cell = "1"
```
ElFantasma: Nit: `bytes` is already a regular `[dependencies]` of this crate (line 11), so bench code can already import it. This dev-dep entry is redundant.

crates/common/rlp/benches/decode.rs
```diff
@@ -1,14 +1,98 @@
...
+        r: U256(rng.random()),
+        s: U256(rng.random()),
+        inner_hash: OnceCell::new(),
+        sender_cache: OnceCell::new(),
+    }
+}
+
+fn random_eip1559_tx(rng: &mut impl Rng) -> EIP1559Transaction {
+    EIP1559Transaction {
+        chain_id: 1,
```
ElFantasma: Minor: `Nibbles::from_raw` treats each input byte as a full byte and splits it into two nibbles (high + low). Since you're generating values in `[0, 16)`, every high nibble is 0, so you get paths like `[0, N, 0, N, ...]`. If you want more realistic trie paths, you could use `Nibbles::from_hex(vec_of_nibbles)` which takes nibble values directly, or generate random bytes in the full `[0, 256)` range for `from_raw`. Not a correctness issue, just affects data representativeness.

crates/common/rlp/Cargo.toml
```diff
@@ -17,6 +17,8 @@ criterion = { version = "0.7.0", features = ["html_reports"] }
 ethrex-common.workspace = true
 ethrex-trie.workspace = true
 rand = "0.9.2"
+bytes.workspace = true
```
ElFantasma: nit: `bytes` is already a regular `[dependencies]` of this crate (line 11), so bench code can already import it. This `[dev-dependencies]` entry is redundant.

crates/common/rlp/benches/decode.rs
```diff
@@ -375,11 +576,159 @@ fn bench_decode_trie(c: &mut Criterion) {
...
+        bench_inline_hash::<24>,
+    );
+
+    group.bench_function(BenchmarkId::new("Node::Leaf", 1000), |b| {
+        b.iter_batched_ref(
+            || {
+                let mut rng = rand::rng();
+                (0..1000)
+                    .map(|_| {
+                        let nibbles = Nibbles::from_raw(
```
ElFantasma, line 620: nit: `Nibbles::from_raw` splits each input byte into two nibbles (high + low). Since you're generating values in `[0, 16)`, each byte only has a non-zero low nibble — the resulting paths are half-zeroed (`[0, x, 0, y, ...]`). Using `Nibbles::from_hex` or generating full-range `[0, 256)` bytes would produce more representative trie paths for benchmarking.

crates/common/rlp/benches/encode.rs
```diff
@@ -1,36 +1,722 @@
...
+fn random_block_header(rng: &mut impl Rng) -> BlockHeader {
+    BlockHeader {
+        hash: OnceCell::new(),
+        parent_hash: H256(rng.random()),
+        ommers_hash: H256(rng.random()),
+        coinbase: Address::from(rng.random::<[u8; 20]>()),
+        state_root: H256(rng.random()),
+        transactions_root: H256(rng.random()),
+        receipts_root: H256(rng.random()),
+        logs_bloom: Bloom::default(),
```
iovoid: Why empty? Considering this is the largest component of a block header it might be worth improving.

crates/common/rlp/benches/decode.rs
```diff
@@ -1,14 +1,98 @@
...
+        data: Bytes::from((0..32).map(|_| rng.random::<u8>()).collect::<Vec<_>>()),
+        access_list: vec![],
+        signature_y_parity: rng.random(),
+        signature_r: U256(rng.random()),
+        signature_s: U256(rng.random()),
+        inner_hash: OnceCell::new(),
+        sender_cache: OnceCell::new(),
+        cached_canonical: OnceCell::new(),
+    }
+}
```
avilagaston9, line 94: nit: `random_log`, `random_block_header`, `random_legacy_tx`, and `random_eip1559_tx` are copy-pasted identically in `encode.rs`. You could extract them into a `benches/helpers.rs` and `mod helpers;` from both files — keeps them in sync if a struct field changes later.

PR #6287, opened 2026-03-02
fix(l1): validate BLS12-381 coordinates against field modulus in parse_coordinate

qu0b:
## Summary

- Add field modulus validation to `parse_coordinate` in `precompiles.rs`, rejecting coordinates >= BLS12-381 field modulus `p` before passing them to the `bls12_381` crate
- The crate's `from_uncompressed` interprets the top 3 bits of the first coordinate byte as BLS serialization flags (compression, infinity, sort), masking them with `& 0x1F`. EIP-2537 uses a different encoding where all 48 bytes are pure coordinate data — no flag bits. Without this check, coordinates like `0x40...` trigger the "infinity" flag (bit 6), causing the precompile to return the identity point instead of failing
- Uses the same `UnsignedInteger::<6>` comparison pattern already present in `bls12_g1add` and `bls12_g2add`

## Impact

This caused a 905,242 gas deficit at block 154022 on bal-devnet-2, leading to an orphaned block at epoch 5402, slot 172886 (produced by `hc-lodestar-ethrex-super-1`). TX 62 called BLS12_G2MSM (precompile `0x0e`) with an invalid field element (`x_1 = 0x40000...0`, which is >= `p`). Geth correctly rejects it and consumes all 927,742 allocated gas; ethrex incorrectly succeeded using only 22,500 gas.

Affects all precompiles that use `parse_coordinate`: G1MSM, G2MSM, G1MUL, G2MUL, MAP_FP_TO_G1, MAP_FP2_TO_G2. The G1ADD, G2ADD, and PAIRING precompiles already have their own local parse functions with correct validation.

## Test plan

- [ ] Verify `cargo check -p ethrex-levm` passes
- [ ] Build Docker image and sync past block 154022 on bal-devnet-2
- [ ] Run EF BLS12-381 precompile test vectors (if available in CI)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/vm/levm/src/precompiles.rs
+++ b/crates/vm/levm/src/precompiles.rs
@@ -1957,8 +1957,8 @@ fn parse_coordinate(coordinate_raw_bytes: &[u8]) -> Result<[u8; 48], VMError> {
     // away. EIP-2537 uses a different encoding where all 48 bytes are pure coordinate
     // data. Rejecting values >= p here prevents the crate from misinterpreting
     // coordinate bits as flags.
-    let coord_value = UnsignedInteger::<6>::from_bytes_be(&coordinate_raw_bytes[16..64])
-        .unwrap_or_default();
+    let coord_value =
+        UnsignedInteger::<6>::from_bytes_be(&coordinate_raw_bytes[16..64]).unwrap_or_default();
     if coord_value >= BLS12381FieldModulus::MODULUS {
         return Err(PrecompileError::ParsingInputError.into());
     }


crates/vm/levm/src/precompiles.rs
```diff
@@ -1951,6 +1951,18 @@ fn parse_coordinate(coordinate_raw_bytes: &[u8]) -> Result<[u8; 48], VMError> {
         return Err(PrecompileError::ParsingInputError.into());
     }
 
+    // Validate that the coordinate is strictly less than the BLS12-381 field modulus.
+    // The bls12_381 crate's from_uncompressed interprets the top bits of the first
```
kevaundray, line 1955: I wonder if long term, we should be manually instantiating the points instead of using `from_uncompressed`, ie `G1Affine::new_unchecked(x,y)` where x and y are created with modulus checks.

Not a blocker, just a thought

ilitteri: Thanks for the suggestion! We're trying this in https://github.com/lambdaclass/ethrex/pull/6320.

PR #6289, opened 2026-03-03
perf(l1): cancel warmer thread early when block execution finishes

ilitteri:
## Motivation

The warmer thread (BAL-based prefetching or speculative re-execution) continues running even after the execution thread has finished processing the block. This wastes CPU cycles on prefetching that will never be used, and delays the overall pipeline because we join the warmer thread last.

## Description

- Add an `AtomicBool` cancellation signal shared between the execution and warmer threads.
- The execution thread sets the flag to `true` after `execute_block_pipeline` completes.
- The warmer checks the flag at key points (between rayon phases in `warm_block_from_bal`, before starting each sender group in `warm_block`) and returns early when cancelled.
- Fix the thread join order: join execution and merkleization first (their results are needed), then join the warmer (whose result is only a duration metric). Previously all three were joined in an arbitrary order inside a tuple constructor.

## How to Test

- Run the existing test suite — the cancellation is a cooperative check, so no behavior changes for tests.
- Benchmark with `import` on `l2-1k-erc20.rlp` to measure reduced CPU time from early warmer exit.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -432,9 +432,13 @@ impl Blockchain {
                             }
                         } else {
                             // Pre-Amsterdam / P2P sync: speculative tx re-execution
-                            if let Err(e) =
-                                LEVM::warm_block(block, caching_store, vm_type, &NativeCrypto, cancelled_ref)
-                            {
+                            if let Err(e) = LEVM::warm_block(
+                                block,
+                                caching_store,
+                                vm_type,
+                                &NativeCrypto,
+                                cancelled_ref,
+                            ) {
                                 debug!("Block warming failed (non-fatal): {e}");
                             }
                         }


crates/blockchain/blockchain.rs
```diff
@@ -432,12 +435,15 @@ impl Blockchain {
...
                 let max_queue_length_ref = &mut max_queue_length;
                 let (tx, rx) = channel();
+                let cancelled_exec = cancelled.clone();
                 let execution_handle = std::thread::Builder::new()
                     .name("block_executor_execution".to_string())
                     .spawn_scoped(s, move || -> Result<_, ChainError> {
                         let (execution_result, bal) =
                             vm.execute_block_pipeline(block, tx, queue_length_ref)?;
 
+                        cancelled_exec.store(true, Ordering::Relaxed);
```
ElFantasma: If `execute_block_pipeline` returns an `Err`, the `?` on line 443 exits the closure before reaching this line, so the cancellation flag is never set. Same if the thread panics. In both cases the warmer continues running to completion.

Not a correctness issue (the scope joins all threads regardless), but it defeats the purpose of the cancellation in failure scenarios. Consider setting the flag unconditionally after the call:

```rust
let result = vm.execute_block_pipeline(block, tx, queue_length_ref);
cancelled_exec.store(true, Ordering::Relaxed);
let (execution_result, bal) = result?;
```

crates/vm/backends/levm/mod.rs
```diff
@@ -325,6 +325,7 @@ impl LEVM {
         block: &Block,
         store: Arc<dyn Database>,
         vm_type: VMType,
+        cancelled: &std::sync::atomic::AtomicBool,
```
ElFantasma: nit: `AtomicUsize` and `Ordering` are already imported from `std::sync::atomic` (line 45). Could add `AtomicBool` to that import instead of using the fully qualified path here and in `warm_block_from_bal`.

crates/vm/backends/levm/mod.rs
iovoid, line 1241: Why not cancel here too?

ilitteri: It's a good idea, I'll test it in another branch and open a new PR if it improves.

crates/blockchain/blockchain.rs
ilitteri: 59f89e9b47

crates/blockchain/blockchain.rs
```diff
@@ -432,12 +435,15 @@ impl Blockchain {
...
+
```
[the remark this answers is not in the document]
ilitteri: 59f89e9b47

crates/blockchain/blockchain.rs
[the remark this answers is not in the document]
ilitteri: 59f89e9b47

crates/blockchain/blockchain.rs
```diff
@@ -491,22 +497,20 @@ impl Blockchain {
                     .map_err(|e| {
                         ChainError::Custom(format!("Failed to spawn merkleizer thread: {e}"))
                     })?;
+                let execution_result = execution_handle.join().unwrap_or_else(|_| {
+                    Err(ChainError::Custom("execution thread panicked".to_string()))
+                });
+                let merkleization_result = merkleize_handle.join().unwrap_or_else(|_| {
+                    Err(StoreError::Custom(
+                        "merklization thread panicked".to_string(),
```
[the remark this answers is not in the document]
ilitteri: f528455e52

crates/vm/backends/levm/mod.rs
```diff
@@ -342,6 +343,9 @@ impl LEVM {
         sender_groups.into_par_iter().for_each_with(
             Vec::with_capacity(STACK_LIMIT),
             |stack_pool, (sender, txs)| {
+                if cancelled.load(Ordering::Relaxed) {
+                    return;
+                }
```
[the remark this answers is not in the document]
ilitteri: fc80a7a885

crates/vm/backends/levm/mod.rs
ilitteri: fc80a7a885

crates/vm/backends/levm/mod.rs
ilitteri: 5aa9cb3086

ilitteri:
446a1c68fb

ilitteri:
Perf results didn't go well

PR #6292, opened 2026-03-03
perf(l1): reuse persistent thread pool for merkleization workers

### LINKED ISSUE #6282: Reuse 19 threads between block executions
The problem: `blockchain.rs` lines 367–487 spawn **19 fresh OS threads per block** via `std::thread::scope` and `Builder::new().spawn_scoped()`:

- 1 warmer thread
- 1 execution thread
- 1 merkleization coordinator
- 16 shard workers (`block_executor_merkleization_shard_worker_0..15`)

Each `pthread_create` costs ~50–100µs on Linux. 19 × 100µs = ~2ms per block purely in thread overhead. More importantly, it also causes: 8MB stack allocation per thread, TLB pressure from stack VA mappings, and thread initialization costs (TLS, stack guard pages). At 5–8 blocks/min the costs add up.

**Change:** Create persistent thread pools:
- A `MERKLE_POOL`: dedicated rayon `ThreadPool` with 16 threads, replacing the 16 shard workers
- Use `MERKLE_POOL.in_place_scope()` instead of `std::thread::scope` for the worker loop
- The warmer and execution threads can potentially also use pools, but they have more complex lifetimes

**Files:** `crates/blockchain/blockchain.rs` — the `std::thread::scope` block with the `for i in 0..16` merkleization worker loop.

Pitfalls:

- This is already done, but we need to make sure and document that this model does not support two blocks executing at the same time.
- We need to really make sure that the threads are clean between block executions.

azteca1998:
Closes #6282

Fixes the BAL (Amsterdam+) merkleization stages B and C, which were using a rayon `ThreadPool` (`in_place_scope`) for blocking channel-recv workers. This caused deadlocks when concurrent test threads each tried to hold all 16 rayon threads blocked on channel recv — the second caller's jobs would queue forever.

Replaces rayon `in_place_scope` with `std::thread::scope` in both Stage B (parallel storage root computation) and Stage C (state shard workers) of `handle_merkleization_bal`. The non-BAL `handle_merkleization` already used `std::thread::scope` and is unchanged.

### Changes

- Replace `MERKLE_POOL` / `merkle_pool()` rayon pool usage in `handle_merkleization_bal` Stage B and Stage C with `std::thread::scope` + `spawn_scoped`
- Remove `MERKLE_POOL`, `merkle_pool()`, `use rayon::ThreadPool`, and `rayon.workspace = true` from `Cargo.toml`
- Extract `run_merkleization_subtrie`, `load_trie`, and `collapse_root_node` as free functions taking `&Store` instead of `&self`, required for the scoped thread closures

### Coverage

- **Pre-Amsterdam (non-BAL path)**: `handle_merkleization` — unchanged, continues to use `std::thread::scope` with 16 scoped shard workers
- **Amsterdam+ (BAL path)**: `handle_merkleization_bal` — Stage B and Stage C now use `std::thread::scope` instead of rayon

## Test plan

- [x] `cargo build -p ethrex-blockchain` — clean build
- [x] `make lint` — no warnings or errors
- [x] All 201 Amsterdam ef-tests pass with `--test-threads=8` (no deadlock)

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -2649,9 +2649,8 @@ fn collapse_root_node(
 
 /// Core logic for a single shard worker during non-BAL merkleization.
 ///
-/// Extracted as a free function so it can be called from both the legacy
-/// `handle_merkleization_subtrie` method and the persistent `ShardWorkerPool`
-/// threads without borrowing `self`.
+/// Extracted as a free function so it can be called from scoped worker threads
+/// without borrowing `self`.
 fn run_merkleization_subtrie(
     rx: Receiver<MerklizationRequest>,
     storage: &Store,


azteca1998:
Superseded by #6344.

PR #6296, opened 2026-03-03
fix(l1): deduplicate transaction `input` hex decoding by reusing shared bytes deserializer

maradini77:
Removed a custom hex-to-`Bytes` decoding implementation for transaction `input` fields and reuses the existing shared `crate::serde_utils::bytes::deserialize` helper instead.


crates/common/types/transaction.rs
```diff
@@ -2195,26 +2195,23 @@ mod serde_impl {
...
-            ))?
+        .map_err(E::custom)?;
+
+        if !data_str.starts_with("0x") {
+            return Err(E::custom("'input' field must start with '0x'"));
         }
+
+        let deser = serde::de::value::StringDeserializer::<E>::new(data_str);
+        crate::serde_utils::bytes::deserialize(deser)
+            .map_err(|_| E::custom("Invalid hex format in 'input' field"))
```
ElFantasma: nit: `|_|` discards the error from `bytes::deserialize`, which already includes useful detail from `hex_simd` (e.g. the invalid character position). Since the `0x` prefix check is handled above, the only errors here are malformed hex — and the original message is more informative than the generic string.
```suggestion
            .map_err(|e| E::custom(e))
```
Or just drop the `.map_err` entirely — `bytes::deserialize` already returns `E::custom(...)` errors.

ElFantasma (changes requested):

ElFantasma (approved):
CI is failing because of the missing scope in the PR title — it needs to be renamed to something like `fix(l1): deduplicate transaction...`

PR #6297, opened 2026-03-03
perf(l1): replace RwLock/Mutex<HashMap> with DashMap

### LINKED ISSUE #6283: Dashmap in CachingDatabase
**The problem:** ethrex's `CachingDatabase` has three caches: `accounts: RwLock<FxHashMap<...>>`, `storage: RwLock<FxHashMap<...>>`, `code: RwLock<FxHashMap<...>>`. These are read by the warming workers in parallel and written during execution.

**Change options:**
- Replace `RwLock<FxHashMap>` with `DashMap` (per-bucket locking, drop-in replacement)
- Or replace with a lock-free fixed-size cache.

DashMap is the lower-risk option. `fixed-cache` has better theoretical properties for hot-key workloads but is less battle-tested.

**Files:** `crates/vm/levm/src/db/mod.rs` — the three `RwLock<FxHashMap>` fields in `CachingDatabase`.

**Estimated impact:** High. The warming phase has multiple rayon threads all reading from `CachingDatabase` simultaneously. Any lock contention there directly serializes warming. If the contention pattern resembles reth's, similar gains (5–10%) are plausible.

iovoid:
**Motivation**

We want to reduce lock contention for shared caches. This change also simplifies away the lock management logic.

**Description**

Replaces RwLock/Mutex containing a HashMap/FxHashMap with a DashMap which allows for lock-free concurrent usage.

Closes #6283

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4410,7 +4410,6 @@ dependencies = [
  "anyhow",
  "async-trait",
  "bytes",
- "dashmap",
  "ethereum-types 0.15.1",
  "ethrex-common",
  "ethrex-crypto",
@@ -4523,7 +4522,6 @@ version = "9.0.0"
 dependencies = [
  "bincode 1.3.3",
  "bytes",
- "dashmap",
  "derive_more 1.0.0",
  "dyn-clone",
  "ethereum-types 0.15.1",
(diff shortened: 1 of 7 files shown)


iovoid:
Re-measure shows a slight _decrease_ in performance

PR #6300, opened 2026-03-03
perf(l1): set high thread priority on block execution pipeline threads

ilitteri:
## Motivation

The block execution pipeline (warmer, executor, merkleizer) runs on scoped threads that compete with background work (networking, syncing, RocksDB compaction) for CPU time. Setting high thread priority ensures these latency-critical threads get scheduled preferentially by the OS.

## Description

- Add `thread-priority` crate dependency
- Add `increase_thread_priority()` helper that attempts max priority with a fallback to priority 62/100
- Call it at the start of all three pipeline threads: `block_executor_warmer`, `block_executor_execution`, and `block_executor_merkleizer`
- Failures are logged at debug level and silently ignored (non-critical optimization)

## How to Test

- `cargo check` passes
- Run block import and verify pipeline threads get elevated priority: `ps -eo pid,ni,comm | grep block_executor`
- Benchmark block import throughput (expected: small improvement on loaded systems)

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -228,12 +228,12 @@ pub struct BatchBlockProcessingFailure {
 fn increase_thread_priority() {
     use thread_priority::{ThreadPriority, ThreadPriorityValue};
     if let Err(err) = ThreadPriority::Max.set_for_current() {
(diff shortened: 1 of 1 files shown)


ilitteri:
Recreating with updated description.

PR #6301, opened 2026-03-03
perf(l1): switch hot EVM and mempool HashMaps to FxHashMap

ilitteri:
## Motivation

Standard `HashMap` uses SipHash which provides HashDoS resistance but is slower than necessary for keys that are already cryptographic hashes (H256, Address). Since these keys have high entropy by construction, we can safely use FxHash for faster lookups.

## Description

Switch `HashMap` → `FxHashMap` in hot paths:

**Mempool** (`mempool.rs`):
- `transaction_pool: FxHashMap<H256, MempoolTransaction>`
- `blobs_bundle_pool: FxHashMap<H256, BlobsBundle>`
- `filter_transactions` / `get_all_txs_by_sender` return types

**Payload building** (`payload.rs`):
- `TransactionQueue.txs: FxHashMap<Address, Vec<MempoolTransaction>>`

**EVM** (`call_frame.rs`, `environment.rs`, `utils.rs`, `vm.rs`):
- `CallFrameBackup` account info and storage slot maps
- `TransientStorage` type alias (EIP-1153)
- `Storage` type alias (slot → value mapping)

All keys are either `H256`, `Address`, `U256`, or tuples thereof — all cryptographic or high-entropy, safe for non-DoS-resistant hashing.

## How to Test

- `cargo check` passes
- Benchmark block import throughput

### DIFF
--- a/crates/vm/levm/src/vm.rs
+++ b/crates/vm/levm/src/vm.rs
@@ -95,7 +95,7 @@ impl Substate {
             accessed_storage_slots,
             created_accounts: FxHashSet::default(),
             refunded_gas: 0,
-            transient_storage: TransientStorage::new(),
+            transient_storage: TransientStorage::default(),
             logs: Vec::new(),
         }
     }


ilitteri:
Recreating with updated description.

PR #6303, opened 2026-03-03
perf(l1): switch hot EVM and mempool HashMaps to FxHashMap

ilitteri:
## Motivation

Standard `HashMap` uses SipHash which provides HashDoS resistance but is slower than necessary for keys that are already cryptographic hashes (H256, Address). Since these keys have high entropy by construction, we can safely use FxHash for faster lookups.

## Description

Switch `HashMap` → `FxHashMap` in hot paths:

**Mempool** (`mempool.rs`):
- `transaction_pool: FxHashMap<H256, MempoolTransaction>`
- `blobs_bundle_pool: FxHashMap<H256, BlobsBundle>`
- `filter_transactions` / `get_all_txs_by_sender` return types

**Payload building** (`payload.rs`):
- `TransactionQueue.txs: FxHashMap<Address, Vec<MempoolTransaction>>`

**EVM** (`call_frame.rs`, `environment.rs`, `utils.rs`, `vm.rs`):
- `CallFrameBackup` account info and storage slot maps
- `TransientStorage` type alias (EIP-1153)
- `Storage` type alias (slot → value mapping)

All keys are either `H256`, `Address`, `U256`, or tuples thereof — all cryptographic or high-entropy, safe for non-DoS-resistant hashing.

## How to Test

- `cargo check` passes
- Benchmark block import throughput


crates/blockchain/mempool.rs
```diff
@@ -250,9 +252,9 @@ impl Mempool {
     pub fn filter_transactions_with_filter_fn(
         &self,
         filter: &dyn Fn(&Transaction) -> bool,
-    ) -> Result<HashMap<Address, Vec<MempoolTransaction>>, StoreError> {
-        let mut txs_by_sender: HashMap<Address, Vec<MempoolTransaction>> =
-            HashMap::with_capacity(128);
+    ) -> Result<FxHashMap<Address, Vec<MempoolTransaction>>, StoreError> {
```
ElFantasma, line 259: `filter_transactions_with_filter_fn` now returns `FxHashMap`, but the test in `test/tests/blockchain/mempool_tests.rs:378` still compares against `HashMap::from([...])`. That's a type mismatch — this won't compile:

```rust
assert_eq!(txs, HashMap::from([(blob_tx.sender(), vec![blob_tx])]));
```

Needs to be updated to construct an `FxHashMap` instead (e.g. `FxHashMap::from_iter([...])` or collect into the right type).

crates/blockchain/mempool.rs
```diff
@@ -35,7 +37,7 @@ impl MempoolInner {
     fn new(max_mempool_size: usize) -> Self {
         MempoolInner {
             txs_order: VecDeque::with_capacity(max_mempool_size * 2),
-            transaction_pool: HashMap::with_capacity(max_mempool_size),
+            transaction_pool: FxHashMap::with_capacity_and_hasher(max_mempool_size, Default::default()),
```
ElFantasma: nit: `FxHashMap::with_capacity_and_hasher(max_mempool_size, Default::default())` is repeated in a few places (here, line 236, line 257). A tiny helper like:

```rust
fn fx_hashmap_with_capacity<K, V>(capacity: usize) -> FxHashMap<K, V> {
    FxHashMap::with_capacity_and_hasher(capacity, Default::default())
}
```

could reduce the verbosity — but totally optional.

crates/blockchain/mempool.rs
```diff
@@ -22,8 +24,8 @@ use tracing::warn;
 #[derive(Debug, Default)]
 struct MempoolInner {
     broadcast_pool: HashSet<H256>,
```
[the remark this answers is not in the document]
ilitteri: c92e061085

crates/blockchain/mempool.rs
ilitteri: Already addressed — the test uses `FxHashMap::from_iter([...])` now.

crates/blockchain/mempool.rs
ilitteri: Agreed it's a bit verbose, but only 3 call sites and the indirection isn't worth it for now. Leaving as-is.

crates/vm/levm/src/utils.rs
```diff
@@ -24,8 +24,8 @@ use ethrex_common::{
 };
 use ethrex_common::{types::TxKind, utils::u256_from_big_endian_const};
 use ethrex_rlp;
-use std::collections::HashMap;
-pub type Storage = HashMap<U256, H256>;
+use rustc_hash::FxHashMap;
+pub type Storage = FxHashMap<U256, H256>;
```
iovoid, line 28: Is this alias actually used?

crates/vm/levm/src/vm.rs
```diff
@@ -25,13 +25,13 @@ use ethrex_common::{
...
     cell::{OnceCell, RefCell},
-    collections::{BTreeMap, BTreeSet, HashMap},
+    collections::{BTreeMap, BTreeSet},
     mem,
     rc::Rc,
 };
 
 /// Storage mapping from slot key to value.
-pub type Storage = HashMap<U256, H256>;
+pub type Storage = FxHashMap<U256, H256>;
```
iovoid, line 35: Is this alias actually used?

PR #6305, opened 2026-03-04
perf(l1): move state diffing out of execution thread

### LINKED ISSUE #6284: Move state diffing out of execution thread
Problem: `get_state_transitions_tx()` call in `vm/backends/levm/mod.rs`. This function drains the `current_accounts_state` map, compares initial vs final values per account, and builds a `Vec<AccountUpdate>`. While this runs, the execution thread is blocked — no next transaction can start.

The merkleization workers are already parallel, but they can't start until the diff is built and sent.

**Change:** Instead of synchronously building `Vec<AccountUpdate>` in the execution thread, do a fast `mem::swap` of the dirty accounts map into a side buffer and send that buffer to a dedicated "diff thread" via a channel. The diff thread produces `Vec<AccountUpdate>` and forwards to the merkleization workers. The execution thread returns to processing transactions immediately after the swap.

**Files:** `crates/vm/backends/levm/mod.rs` (`send_state_transitions_tx()`, `get_state_transitions_tx()`), `crates/vm/levm/src/db/gen_db.rs` (the state drain logic).

**Estimated impact:** Depends on the relative cost of state diffing vs transaction execution. For ERC-20 heavy blocks (many small transfers touching 2 accounts each), diffing is cheap relative to execution. For complex DeFi blocks (50+ storage slot changes per tx), diffing cost is non-trivial and hiding it behind execution could be meaningful. Profile `get_state_transitions_tx` to quantify.

azteca1998:
## Summary

Closes #6284

- Introduces a dedicated **diff thread** in the block execution pipeline that computes `Vec<AccountUpdate>` from raw dirty state, unblocking the execution thread from state diffing work
- Adds `extract_dirty_state()` on `GeneralizedDatabase` that does a fast `mem::take` of the dirty accounts map and updates `initial_accounts_state` for subsequent transactions
- Adds `compute_state_transitions()` standalone function that performs the diff logic on the extracted payload (runs on the diff thread)
- Pipeline is now: **execution → diff thread → merkleizer** (previously: execution → merkleizer)
- BAL parallel path is unchanged — it sends pre-computed `Vec<AccountUpdate>` directly to the merkleizer, bypassing the diff thread

## Test plan

- [x] `cargo build` — full workspace compiles
- [x] `cargo clippy` — no warnings on changed crates
- [x] `cargo test` — all tests pass on `ethrex-levm`, `ethrex-vm`, `ethrex-blockchain`
- [ ] Run hive tests to validate block execution correctness end-to-end


azteca1998:
Did not improve.

PR #6306, opened 2026-03-04
feat(l1): implement eth/71 Block Access List exchange (EIP-8159)

edg-l:
## Summary
- Add `GetBlockAccessLists` / `BlockAccessLists` messages (0x12/0x13) per [EIP-8159](https://eips.ethereum.org/EIPS/eip-8159)
- Add `EthCapVersion::V71` with capability negotiation and offset constants
- Add `BLOCK_ACCESS_LISTS` storage table with store/get methods
- Store produced BALs after block execution in `add_block_pipeline_inner()`
- Serve BALs to peers via `GetBlockAccessLists` handler (response capped at `BLOCK_ACCESS_LIST_LIMIT` = 1024 to bound memory)
- Fetch BALs from eth/71 peers during full sync for Amsterdam+ blocks, enabling parallel execution with correct gas handling
- Fallback to sequential execution (`None` BAL) if peer doesn't support eth/71 or BAL is unavailable
- Reuse eth/69 receipts for eth/71 (no `Receipts71` message; `GetReceipts69` / `Receipts69` accepted under both `V69` and `V71`)

**No DB migration or resync required.** RocksDB automatically creates the new `BLOCK_ACCESS_LISTS` column family on startup. The table will be empty for existing blocks — BALs are only populated for newly executed blocks going forward. During sync, if a peer doesn't support eth/71 or BALs are unavailable, the node falls back to sequential execution as before.

## Implementation notes
- **BAL fetch is sequential** (one `GetBlockAccessLists` round-trip per batch, awaited in `add_blocks_in_batch` before execution). Moving the fetch into the background download task alongside block bodies is a possible follow-up.
- **BAL only used on the sequential pipeline path.** Main's full sync tries `Blockchain::add_blocks_in_batch` first (which does not consume BAL), then falls back to the per-block pipeline `run_blocks_pipeline` on post-execution errors and at the sync head. BAL feeds the pipeline path only; the batch path remains BAL-agnostic and unchanged.
- Request-id matching is validated on the response side of `PeerHandler::request_block_access_lists` to prevent unrelated responses from being substituted.
- Branch was rebased + squashed onto current `main`; previous commits' inline review comments are now marked outdated by GitHub but their fixes are baked in (request-id validation, RLP-framework refactor of `OptionalBal`, allocation cap on `Vec::with_capacity`, `V69 | V71` receipts guard, `lastest_block` typo fix).

## Test plan
- [x] `cargo build --workspace` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test -p ethrex-storage -p ethrex-blockchain -p ethrex-p2p` passes (incl. 7 unit tests in `block_access_lists::tests` covering RLP roundtrip, `OptionalBal` None/Some encoding, mixed responses, and `BLOCK_ACCESS_LIST_LIMIT` truncation)
- [ ] Manual devnet test: wipe DB on bal-devnet-2 node, restart with eth/71 support, verify full sync past Amsterdam fork via `GetBlockAccessLists` exchange (logs should show BAL fetches and the new column family populating)

> Note: there is currently no upstream Hive coverage for EIP-8159 (the existing EELS BAL suites — `run-hive-eels-amsterdam`, `run-hive-eels-bal-quick` — exercise EIP-7928 BAL execution semantics, not the eth/71 wire-protocol exchange).


crates/networking/p2p/sync/full.rs
```diff
@@ -204,10 +210,31 @@ async fn add_blocks_in_batch(
...
+    let blocks_hashes: Vec<H256> = blocks.iter().map(|b| b.hash()).collect();
+
+    // Fetch BALs for Amsterdam+ blocks (eth/71). If unavailable, fall back to None per block.
+    let chain_config = store.get_chain_config();
+    let bals: Vec<Option<BlockAccessList>> = {
+        // Only attempt BAL fetch if at least one block is Amsterdam+
+        let any_amsterdam = blocks
+            .iter()
+            .any(|b| chain_config.is_amsterdam_activated(b.header.timestamp));
+        if any_amsterdam {
```
ElFantasma, line 308: BALs are fetched here for every batch, but `add_blocks` only uses them when `sync_head_found` is true (the final batch). For non-final batches, the code falls into the `blockchain.add_blocks_in_batch()` path which doesn't accept BALs — so the network round-trip is wasted.

Consider gating on `final_batch`:
```rust
if any_amsterdam && final_batch {
```
Or, for full benefit, plumb BALs through to `blockchain.add_blocks_in_batch()` too so intermediate batches can also execute in parallel.

ElFantasma (changes requested):

azteca1998 (approved):

ElFantasma (approved):

ilitteri (approved):

PR #6309, opened 2026-03-04
fix(levm): use post-refund gas for check_gas_limit in sequential execution

edg-l:
## Summary
- Use post-refund gas (`report.gas_spent`) for the `check_gas_limit` accumulator in both sequential execution paths (`execute_block` and `execute_block_pipeline`)
- Keep pre-refund gas (`report.gas_used`) for `block_gas_used` in `BlockExecutionResult` (used for header validation via `validate_gas_used`)

Pre-Amsterdam, both values are identical so there is no behavior change.

## Post-mortem

**Observed error on bal-devnet-2:**
```
Error: Invalid Block: Invalid transaction: Gas allowance exceeded: Block gas used overflow: used 79340683 + tx limit 1000000 > block limit 80000000

Caused by:
    Invalid transaction: Gas allowance exceeded: Block gas used overflow: used 79340683 + tx limit 1000000 > block limit 80000000

Location:
    /ethrex/cmd/ethrex/initializers.rs:629:9
```

**Root cause:** When a node executes Amsterdam+ blocks sequentially (e.g. full sync, engine API catch-up, or any path without a BAL), `check_gas_limit` accumulates gas to verify each transaction fits within the block gas limit. EIP-7778 introduced separate gas tracking: `gas_used` (pre-refund, higher) for block-level accounting and `gas_spent` (post-refund, lower) for what the user actually pays. The `check_gas_limit` accumulator was using pre-refund gas, which inflates the running total.

**Why it failed:** The accumulated pre-refund gas (79.3M) + next tx's gas limit (1M) = 80.3M, exceeding the block gas limit (80M). With post-refund gas the accumulated value is lower (accounting for refunds from SSTORE clears, etc.), keeping the total within the 80M limit. The block is valid — it's at 99.2% capacity and the refund difference is just enough to push the pre-refund accumulator over.

**Fix:** Use post-refund gas (`report.gas_spent`) for the `check_gas_limit` accumulator, matching the pre-Amsterdam behavior where both values were identical. Pre-refund gas (`report.gas_used`) is still used for `block_gas_used` in the `BlockExecutionResult`, which is validated against `header.gas_used`.

**Why EF tests didn't catch this:** The existing EIP-7778 EF test (`test_multi_transaction_gas_accounting`) has `gasUsed == gasLimit` (fully packed block), but each transaction's pre-refund gas is bounded by its own `gas_limit`. The check is `accumulated_gas + next_tx.gas_limit <= block_gas_limit` — since a tx's actual gas can never exceed its own limit, the accumulated pre-refund gas from prior txs never pushes the check over. The bug only manifests with real-world blocks where large refunds (e.g. SSTORE clears) create a significant gap between pre-refund and post-refund gas AND the block is packed tightly enough that the inflated accumulator + next tx's gas limit overflows the block limit.

image

## Test plan
- [x] `cargo build` passes
- [x] `cargo test` passes
- [x] EIP-7778 EF tests pass (`test_multi_transaction_gas_accounting`)
- [ ] Manual devnet test on bal-devnet-2

### DIFF
--- a/.github/actions/install-risc0/action.yml
+++ b/.github/actions/install-risc0/action.yml
@@ -7,9 +7,12 @@ runs:
     - name: Install RISC0
       shell: bash
       run: |
-        curl -L https://risczero.com/install | bash
-        ~/.risc0/bin/rzup install cargo-risczero 3.0.3
-        ~/.risc0/bin/rzup install risc0-groth16
-        ~/.risc0/bin/rzup install rust
-        ~/.risc0/bin/rzup install cpp
-        ~/.risc0/bin/rzup install r0vm 3.0.3
+        source "$GITHUB_WORKSPACE/.github/scripts/retry.sh"
+
+        retry curl -L https://risczero.com/install -o /tmp/risc0_install.sh
+        bash /tmp/risc0_install.sh
+        retry ~/.risc0/bin/rzup install cargo-risczero 3.0.3
+        retry ~/.risc0/bin/rzup install risc0-groth16
+        retry ~/.risc0/bin/rzup install rust
+        retry ~/.risc0/bin/rzup install cpp
+        retry ~/.risc0/bin/rzup install r0vm 3.0.3
--- a/.github/actions/install-sp1/action.yml
+++ b/.github/actions/install-sp1/action.yml
@@ -1,0 +1,14 @@
+name: "Install SP1"
+description: "Install SP1 Toolchain"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install SP1
+      shell: bash
+      run: |
+        source "$GITHUB_WORKSPACE/.github/scripts/retry.sh"
+
+        retry curl -L https://sp1.succinct.xyz -o /tmp/sp1_install.sh
+        bash /tmp/sp1_install.sh
+        retry ~/.sp1/bin/sp1up --version 5.0.8
--- a/.github/actions/install-zisk/action.yml
+++ b/.github/actions/install-zisk/action.yml
@@ -1,0 +1,20 @@
+name: "Install ZisK"
+description: "Install ZisK Toolchain"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install ZisK dependencies
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y xz-utils jq curl build-essential qemu-system libomp-dev libgmp-dev nlohmann-json3-dev protobuf-compiler uuid-dev libgrpc++-dev libsecp256k1-dev libsodium-dev libpqxx-dev nasm libopenmpi-dev openmpi-bin openmpi-common libclang-dev clang gcc-riscv64-unknown-elf
+
+    - name: Install ZisK
+      shell: bash
+      run: |
+        source "$GITHUB_WORKSPACE/.github/scripts/retry.sh"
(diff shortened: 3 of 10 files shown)


crates/vm/backends/levm/mod.rs
```diff
@@ -103,13 +103,20 @@ impl LEVM {
         let mut cumulative_gas_used = 0_u64;
         // Block gas accounting (PRE-REFUND for Amsterdam+ per EIP-7778)
         let mut block_gas_used = 0_u64;
+        // Post-refund accumulator for check_gas_limit (pre-refund would reject valid
+        // tightly-packed Amsterdam blocks when executing sequentially without BAL)
+        let mut gas_used_for_limit_check = 0_u64;
```
iovoid: Always equal to cumulative_gas_used

crates/vm/backends/levm/mod.rs
```diff
@@ -263,12 +271,19 @@ impl LEVM {
...
```
iovoid: Always equal to cumulative_gas_used

crates/vm/backends/levm/mod.rs
iovoid: The name is somewhat confusing since it can be read as both:
- 'gas used', for limit check (incorrect, it's gas _spent_ that we track)
- 'gas', used for limit check (correct but more generic)

crates/vm/backends/levm/mod.rs
```diff
@@ -128,8 +132,8 @@ impl LEVM {
             let report = Self::execute_tx(tx, tx_sender, &block.header, db, vm_type)?;
 
             // EIP-7778: Separate gas tracking
-            // - gas_spent (POST-REFUND) for receipt cumulative_gas_used
-            // - gas_used (PRE-REFUND for Amsterdam+) for block accounting
+            // - gas_spent (POST-REFUND) for receipt cumulative_gas_used and limit check
+            // - gas_used (PRE-REFUND for Amsterdam+) for block accounting / header validation
             cumulative_gas_used += report.gas_spent;
```
iovoid, line 133: I think it might be a good idea to rename cumulative_gas_used to cumulative_gas_spent to make it even clearer what it tracks

crates/vm/backends/levm/mod.rs
```diff
@@ -268,7 +272,11 @@ impl LEVM {
         let mut tx_since_last_flush = 2;
 
         for (tx_idx, (tx, tx_sender)) in transactions_with_sender.into_iter().enumerate() {
-            check_gas_limit(block_gas_used, tx.gas_limit(), block.header.gas_limit)?;
+            check_gas_limit(
+                cumulative_gas_used,
```
ElFantasma: nit: `check_gas_limit`'s first parameter is still named `block_gas_used` and the error message says "Block gas used overflow", but it now receives the post-refund cumulative value. Consider renaming the parameter to `cumulative_gas` (or similar) and adjusting the error message so future debugging sessions aren't misleading.

ElFantasma (commented):
Clean fix — using post-refund `cumulative_gas_used` for `check_gas_limit` is correct per EIP-7778, and the post-mortem in the PR description is excellent. The change is symmetric across both `execute_block` and `execute_block_pipeline`, and pre-Amsterdam behavior is preserved since both values are identical.

PR #6313, opened 2026-03-04
fix(l2): defer aligned cursor until on-chain verification

iovoid:
## Motivation

In aligned-mode, `finalize_batch_proof()` advances the DB cursor (`set_latest_sent_batch_proof`) and deletes checkpoint data immediately after sending to the Aligned gateway, before on-chain verification. If Aligned accepts but fails to aggregate a proof, the sender never revisits that batch, causing a permanent L1 verification stall.

## Description

  - Add a separate `latest_sent_to_aligned` cursor so the sender tracks what's been submitted to the gateway, while the main cursor (`latest_sent_batch_proof`) only advances after the verifier confirms on-chain verification
  - Add a configurable `--aligned.resubmission-timeout` so lost proofs are automatically resent after a timeout
  - Move main cursor advancement and checkpoint cleanup from `L1ProofSender` into `L1ProofVerifier`, which runs after the on-chain verify tx succeeds

### Storage layer
  - New `get/set_latest_sent_to_aligned` methods on `StoreEngineRollup` trait, storing `(batch_number, timestamp_secs)`
  - Implemented for both in-memory and SQL backends (new `latest_sent_to_aligned` table)

### L1ProofSender
  - Aligned-mode block in `verify_and_send_proofs` now uses the separate aligned cursor instead of `latest_sent_batch_proof`
  - `verify_and_send_proofs_aligned` no longer calls `finalize_batch_proof` and only advances the aligned cursor
  - Time-based resubmission: if `now - sent_at > resubmission_timeout_secs`, resets to `last_verified_batch + 1`

### L1ProofVerifier
  - After successful on-chain verification tx: advances main cursor (`set_latest_sent_batch_proof`) and removes checkpoint directories
  - Accepts `checkpoints_dir` parameter threaded from sequencer startup


crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -190,7 +193,40 @@ impl L1ProofSender {
         let latest_sent_batch_db = self.rollup_store.get_latest_sent_batch_proof().await?;
```
ElFantasma: nit: `latest_sent_batch_db` is fetched here but no longer used in aligned mode (the aligned branch now reads `latest_sent_to_aligned` instead and returns early). This is an unnecessary DB round-trip on every sender tick when `aligned_mode` is true. Consider moving this read below the `if self.aligned_mode { ... }` block.

cmd/ethrex/l2/options.rs
```diff
@@ -880,6 +881,15 @@ pub struct AlignedOptions {
         help_heading = "Aligned options"
     )]
     pub from_block: Option<u64>,
+    #[arg(
+        long = "aligned.resubmission-timeout",
+        default_value = "300",
```
avilagaston9: Aligned usually aggregates proofs once per day, so this value is incorrect as a default. I would leave this field without any default to avoid skill issues in production deployments.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -190,7 +193,40 @@ impl L1ProofSender {
...
+                    .duration_since(std::time::UNIX_EPOCH)
+                    .map_err(|e| ProofSenderError::UnexpectedError(e.to_string()))?
+                    .as_secs();
+                if now.saturating_sub(sent_at) > self.resubmission_timeout_secs {
+                    warn!(
+                        latest_sent_to_aligned,
+                        last_verified_batch,
+                        elapsed_secs = now - sent_at,
+                        "Aligned verification timed out, resending proof"
+                    );
```
avilagaston9: I would log an `ERROR` here, because if the timeout is properly set, this should never happen unless Aligned is down or something unexpected occurred.

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -190,7 +193,40 @@ impl L1ProofSender {
         let latest_sent_batch_db = self.rollup_store.get_latest_sent_batch_proof().await?;
 
         if self.aligned_mode {
-            let batch_to_send = std::cmp::max(latest_sent_batch_db, last_verified_batch) + 1;
+            let (latest_sent_to_aligned, sent_at) =
+                self.rollup_store.get_latest_sent_to_aligned().await?;
```
avilagaston9: We are always checking `latest_sent_to_aligned` instead of the "first in the queue". This means that some batches may have timed out while waiting to be verified, but we have not noticed yet.

crates/l2/sequencer/l1_proof_verifier.rs
```diff
@@ -268,13 +279,29 @@ impl L1ProofVerifier {
...
             self.rollup_store
                 .store_verify_tx_by_batch(batch_number, verify_tx_hash)
                 .await?;
         }
 
+        // Advance verified cursor now that on-chain verification succeeded.
+        // Only the verifier writes this cursor; the sender reads it for timeout detection.
+        self.rollup_store
+            .set_latest_verified_batch_proof(last_batch_number, now)
+            .await?;
```
avilagaston9, line 313: The node may be stopped before reaching this step, leaving an outdated timestamp in the db. We can add a comparison (line 129) between the contract and the db cursor, and update it if `db_batch < contract_batch`.

iovoid: Addressed in f5371070

PR #6314, opened 2026-03-04
chore(l2): add selector hash comments to custom error declarations

avilagaston9:
**Motivation**

When debugging on-chain transactions, raw revert data only shows 4-byte selectors. Without a quick reference, identifying which custom error was triggered requires manually computing `keccak256` of each error signature. This slows down debugging, especially when triaging CI failures or investigating reverts from block explorers.

**Description**

Add inline comments with the 4-byte selector hash next to every custom error declaration across all L1 interface contracts:

- `IOnChainProposer.sol` (non-based) — 30 errors
- `IOnChainProposer.sol` (based) — 33 errors
- `IRiscZeroVerifier.sol` — 1 error
- `ITimelock.sol` — 2 errors
- `IRouter.sol` — 6 errors

Format: `error ErrorName(); // 0xabcd1234`

**Note on test coverage**

The existing `error_selectors.rs` test validates selector constants that are actually used in Rust code — that's its purpose. The inline comments added here are developer convenience for reading raw revert data, not constants consumed by the codebase, so extending the test to cover them is unnecessary.

**Checklist**

### DIFF
--- a/crates/l2/contracts/src/l1/based/interfaces/IOnChainProposer.sol
+++ b/crates/l2/contracts/src/l1/based/interfaces/IOnChainProposer.sol
@@ -7,52 +7,52 @@ pragma solidity =0.8.31;
 /// by the proposer to commit batches of l2 blocks and verify proofs.
 interface IOnChainProposer {
     // Initialization errors
-    error MissingRisc0Verifier();
-    error MissingSp1Verifier();
-    error MissingTdxVerifier();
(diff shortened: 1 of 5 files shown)


ElFantasma (approved):
Verified all 49 selector hashes against `keccak256` — they're all correct. The shared errors between based and non-based `IOnChainProposer` are consistent. Clean change, LGTM.

PR #6317, opened 2026-03-05
fix(l1): fix BAL hash mismatches in block building

edg-l:
## Summary

Fixes ethrex-built blocks being rejected by other clients (nimbus, geth) after Amsterdam fork activation with "Mismatched blockAccessListHash". Root cause was ordering differences between ethrex and geth in the block building and execution paths.

### 1. Withdrawal ordering in block building
Moved `apply_withdrawals` after `fill_transactions` so withdrawals are processed at BAL index `n+1` (post-tx phase) instead of before transactions. Previously, withdrawal balance changes were recorded at the wrong BAL index.

### 2. BAL snapshot/restore for rejected transactions
Clone the BAL recorder state before each transaction attempt during block building. If a transaction fails and is skipped, restore the snapshot so rejected transactions don't pollute the block access list with stale touched addresses or index changes.

### 3. Post-tx phase ordering: requests before withdrawals
Both building (`payload.rs`) and execution (`levm/mod.rs`) paths were processing withdrawals before requests in the post-tx phase. Geth does requests (system calls) first, then withdrawals via `Finalize()`. This ordering difference caused BAL hash mismatches since state changes from system calls and withdrawals were recorded in different order. Fixed in all 4 code paths: building, sequential execution, parallel execution, and pipeline sequential fallback.

### 4. Tolerate extraneous no-op BAL entries in parallel validation
During parallel block validation, handle no-op change entries (where post == pre) for accounts that are touched but not modified (e.g. coinbase with zero fee, reverted txs).

## Test plan

- [x] All EIP-7928 tests pass (45 tests)
- [x] All EF blockchain tests pass (6158 tests, 0 failures)
- [x] `cargo check` passes
- [x] Deployed to bal-devnet-2 — all 3 clients (nimbus, geth, ethrex) stay in sync post-Amsterdam with no orphaned blocks

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -960,11 +960,18 @@ impl LEVM {
                 if let Some(expected_code) = find_exact_change_code(&acct.code_changes, bal_idx) {
                     match actual {
                         Some(a) => {
-                            let actual_code = codes
-                                .get(&a.info.code_hash)
-                                .map(|c| &c.bytecode)
-                                .cloned()
-                                .unwrap_or_default();
+                            let actual_code = if let Some(c) = codes.get(&a.info.code_hash) {
+                                c.bytecode.clone()
+                            } else {
+                                store
+                                    .get_account_code(a.info.code_hash)
+                                    .map_err(|e| {
+                                        BalValidationError::Database(format!(
+                                            "DB error reading account code for {addr:?}: {e}"
+                                        ))
+                                    })?
+                                    .bytecode
+                            };
                             if actual_code != *expected_code {
                                 return Err(BalValidationError::Mismatch(format!(
                                     "account {addr:?} code mismatch at index {bal_idx}"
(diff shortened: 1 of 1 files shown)


crates/vm/backends/levm/mod.rs
```diff
@@ -758,6 +781,58 @@ impl LEVM {
...
+            acct.balance_changes[pos - 1].post_balance
+        } else {
+            system_seed
+                .get(&acct.address)
+                .map(|a| a.info.balance)
+                .unwrap_or_else(|| {
+                    store
+                        .get_account_state(acct.address)
+                        .map(|a| a.balance)
+                        .unwrap_or_default()
```
iovoid: We shouldn't be doing unwrap_or_default on an actual error.

crates/vm/backends/levm/mod.rs
```diff
@@ -758,6 +781,58 @@ impl LEVM {
...
+            acct.nonce_changes[pos - 1].post_nonce
+        } else {
+            system_seed
+                .get(&acct.address)
+                .map(|a| a.info.nonce)
+                .unwrap_or_else(|| {
+                    store
+                        .get_account_state(acct.address)
+                        .map(|a| a.nonce)
+                        .unwrap_or_default()
```
iovoid: We shouldn't be doing unwrap_or_default on an actual error.

crates/vm/backends/levm/mod.rs
```diff
@@ -855,6 +983,14 @@ impl LEVM {
                         let key = ethrex_common::utils::u256_to_h256(sc.slot);
                         let actual_value = actual.and_then(|a| a.storage.get(&key)).copied();
                         if actual_value != Some(expected_value) {
+                            // If account not in execution state, check pre-state
+                            if actual.is_none() || actual_value.is_none() {
+                                let pre_value =
+                                    store.get_storage_value(addr, key).unwrap_or_default();
```
iovoid: We shouldn't be doing unwrap_or_default on an actual error.

crates/vm/backends/levm/mod.rs
```diff
@@ -779,6 +856,8 @@ impl LEVM {
         codes: &FxHashMap<H256, Code>,
         bal: &BlockAccessList,
         index: &BalAddressIndex,
+        system_seed: &CacheDB,
+        store: &Arc<dyn Database>,
     ) -> Result<(), String> {
```
iovoid: Also why do we return a String?

crates/vm/backends/levm/mod.rs
```diff
@@ -834,15 +966,37 @@ impl LEVM {
...
-                                "account {addr:?} has BAL code change at {bal_idx} but not in execution state"
-                            ));
+                            // No-op check: compare against pre-state code.
+                            // Try system_seed + codes cache first, then fall
+                            // back to store (consistent with balance/nonce).
+                            let code_hash = system_seed
+                                .get(&addr)
+                                .map(|a| a.info.code_hash)
+                                .or_else(|| {
+                                    store.get_account_state(addr).ok().map(|a| a.code_hash)
```
iovoid: .ok() silently discards errors

crates/vm/backends/levm/mod.rs
```diff
@@ -834,15 +966,37 @@ impl LEVM {
...
+                                });
+                            let pre_code = match code_hash {
+                                Some(hash) => codes
+                                    .get(&hash)
+                                    .map(|c| c.bytecode.clone())
+                                    .or_else(|| {
+                                        store.get_account_code(hash).ok().map(|c| c.bytecode)
```
iovoid: .ok() silently discards errors

PR #6318, opened 2026-03-05
chore(l1): perf report improved v2

Arkenan:
Changes:

- Only published on mondays and thursdays.
- Improved telegram messages
- addressed comments on previous PR


.github/scripts/generate_combined_report.sh
```diff
@@ -42,6 +42,7 @@ if [[ -f "$LOC_JSON" ]]; then
 
   pct() {
     # Integer percentage: pct numerator denominator
+    [[ "$2" -eq 0 ]] && { printf "0"; return; }
```
ElFantasma, line 45: Good catch on the division-by-zero. If `loc_total` is ever 0, this would have blown up.

ilitteri (approved):

PR #6319, opened 2026-03-05
fix(l1): geth-compatible callTracer format for debug_trace* endpoints

avilagaston9:
## Motivation

Tools like Blockscout that consume `debug_traceTransaction` and `debug_traceBlockByNumber` expect geth's callTracer response format. Two incompatibilities prevented ethrex traces from being consumed correctly:

1. `debug_traceTransaction` returned an array of `CallTraceFrame` objects, but geth returns a single root frame (object, not array)
2. `debug_traceBlockByNumber` only accepted decimal block numbers, but clients like Blockscout send hex-encoded strings (e.g. `"0x13E"`)

Note: the Hive `ethereum/rpc-compat` suite does not cover `debug_trace*` endpoints, so these issues went undetected.

## Description

- Unwrap the first `CallTraceFrame` from the `Vec<CallTraceFrame>` trace result in both `debug_traceTransaction` and `debug_traceBlockByNumber` handlers, matching geth's single-object response format
- Parse block number parameter via `BlockIdentifier::parse()` which handles both hex strings (`"0x2a"`) and falls back to plain integer deserialization

## How to Test

```bash
# Start ethrex in dev mode
ethrex --dev

# Send a transaction, then trace it:
curl -X POST http://localhost:8545 -H 'Content-Type: application/json' -d '{
  "jsonrpc": "2.0", "id": 1,
  "method": "debug_traceTransaction",
  "params": ["<TX_HASH>", {"tracer": "callTracer"}]
}'
# Should return a single object {type, from, to, ...}, not an array

# Trace a block by hex number:
curl -X POST http://localhost:8545 -H 'Content-Type: application/json' -d '{
  "jsonrpc": "2.0", "id": 1,
  "method": "debug_traceBlockByNumber",
  "params": ["0x1", {"tracer": "callTracer"}]
}'
# Should accept hex block numbers and return [{txHash, result: {type, from, to, ...}}]
```

## Checklist

- [x] Lint passes
- [x] Format passes
- [x] Tests pass

### DIFF
--- a/.github/scripts/set-pr-status.js
+++ b/.github/scripts/set-pr-status.js
@@ -151,7 +151,12 @@ module.exports = async ({ github, context }) => {
     // Gets all issue numbers that would be closed if the PR is merged.
     function extractLinkedIssueNumbers(prBody) {
         const body = prBody || "";
-        const withoutComments = body.replace(/<!--[\s\S]*?-->/g, "");
+        let withoutComments = body;
+        let previous;
+        do {
+            previous = withoutComments;
+            withoutComments = withoutComments.replace(/<!--[\s\S]*?-->/g, "");
+        } while (withoutComments !== previous);
         const matches = [...withoutComments.matchAll(/(?:close[sd]?|fixe[sd]?|resolve[sd]?)\s+#(\d+)/gi)];
         return matches.map(match => parseInt(match[1], 10));
     }
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -1867,6 +1867,7 @@ impl Blockchain {
             receipts: vec![(block.hash(), execution_result.receipts)],
             blocks: vec![block],
             code_updates: account_updates_list.code_updates,
+            batch_mode: false,
         };
 
         self.storage
@@ -2377,6 +2378,7 @@ impl Blockchain {
             blocks,
             receipts: all_receipts,
             code_updates,
+            batch_mode: true,
         };
 
         self.storage
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -434,9 +434,25 @@ impl Blockchain {
         if let BlockchainType::L1 = self.options.r#type {
             self.apply_system_operations(&mut context)?;
         }
-        self.apply_withdrawals(&mut context)?;
         self.fill_transactions(&mut context)?;
+        // EIP-7928: Post-tx phase uses index n+1 for both requests and withdrawals.
+        // Order must match geth: requests (system calls) BEFORE withdrawals.
+        if context
+            .chain_config()
+            .is_amsterdam_activated(context.payload.header.timestamp)
+        {
+            #[allow(clippy::cast_possible_truncation)]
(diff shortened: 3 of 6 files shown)


crates/networking/rpc/tracing.rs
```diff
@@ -173,9 +186,15 @@ impl RpcHandler for TraceBlockByNumberRequest {
...
                     .map_err(|err| RpcErr::Internal(err.to_string()))?;
-                // We need to show transactions from newest to oldest
-                let block_trace: BlockTrace<CallTrace> =
-                    call_traces.into_iter().rev().map(Into::into).collect();
+                // Unwrap each CallTrace (Vec<CallTraceFrame>) to a single
+                // CallTraceFrame to match geth's callTracer response format.
+                let block_trace: BlockTrace<CallTraceFrame> = call_traces
+                    .into_iter()
+                    .rev()
+                    .filter_map(|(hash, trace)| {
```
ElFantasma: `filter_map` silently drops transactions whose `CallTrace` vec is empty. In contrast, `debug_traceTransaction` (line 124) returns an error for empty traces. This inconsistency could hide issues — a block trace response would simply omit a transaction instead of signaling a problem.

Consider using `.map()` + `ok_or(...)` to propagate an error, or at minimum returning a default/empty `CallTraceFrame` so every transaction is represented in the response (which is what geth does).

crates/networking/rpc/tracing.rs
```diff
@@ -173,9 +186,15 @@ impl RpcHandler for TraceBlockByNumberRequest {
...
                     .await
                     .map_err(|err| RpcErr::Internal(err.to_string()))?;
-                // We need to show transactions from newest to oldest
-                let block_trace: BlockTrace<CallTrace> =
-                    call_traces.into_iter().rev().map(Into::into).collect();
+                // Unwrap each CallTrace (Vec<CallTraceFrame>) to a single
+                // CallTraceFrame to match geth's callTracer response format.
+                let block_trace: BlockTrace<CallTraceFrame> = call_traces
+                    .into_iter()
+                    .rev()
```
ElFantasma: nit: This `.rev()` is inherited from the previous code, but geth's `debug_traceBlockByNumber` returns traces in transaction execution order (index 0 first). Since this PR targets geth compatibility, it'd be worth dropping the `.rev()` here.

ElFantasma (changes requested):

crates/networking/rpc/tracing.rs
avilagaston9: Fixed in 01d7ed0ab — replaced `filter_map` with `.map()` + `ok_or_else()` so empty traces now return `RpcErr::Internal("Empty call trace")`, consistent with `debug_traceTransaction`.

crates/networking/rpc/tracing.rs
avilagaston9: Fixed in 01d7ed0ab — removed `.rev()`. Verified against geth source (`eth/tracers/api.go`): both sequential and parallel paths store results at `results[i]` preserving transaction execution order.

ElFantasma (approved):

PR #119, opened 2024-07-04
feat: add datatest to consume ef tests.

mpaulucci:
**Motivation**
We want to be able to consume tests from the fixture files provided by the EF

**Description**
Adds datatest boilerplate with one working example.

PR #6322, opened 2026-03-05
ci(l1): loop HTML comment removal to prevent incomplete sanitization

klaus993:
## Summary

Fixes [CodeQL alert #119](https://github.com/lambdaclass/ethrex/security/code-scanning/119) (`js/incomplete-multi-character-sanitization`).

### Problem

In `.github/scripts/set-pr-status.js`, `extractLinkedIssueNumbers` strips HTML comments from PR bodies before extracting "closes #N" keywords. The regex used a single-pass replacement:

```js
const withoutComments = body.replace(/<!--[\s\S]*?-->/g, "");
```

A single-pass replacement of `<!--...-->` can leave behind a valid HTML comment if the input is crafted so that removing one match creates a new one. For example, input `<!-<!-- foo -->->` would produce `<!-- foo -->` after one pass.

This means a malicious PR author could hide closing keywords inside crafted comment nesting to manipulate project board status. The practical risk is low (CI context, trusted-ish input), but the fix is trivial.

### Fix

Replace the single-pass `.replace()` with a loop that repeats until stable:

```js
let withoutComments = body;
let previous;
do {
    previous = withoutComments;
    withoutComments = withoutComments.replace(/<!--[\s\S]*?-->/g, "");
} while (withoutComments !== previous);
```

This is the approach recommended by [CodeQL's own documentation](https://codeql.github.com/codeql-query-help/javascript/js-incomplete-multi-character-sanitization/) for this exact alert type.

## Test plan

- No behavioral change for normal PR bodies — the single pass already removes all comments; the loop just adds a no-op second iteration confirming stability
- Edge case: nested/overlapping comment markers like `<!-<!-- foo -->->` are now fully stripped across multiple iterations
- Resolves CodeQL alert `js/incomplete-multi-character-sanitization`

### DIFF
--- a/.github/scripts/set-pr-status.js
+++ b/.github/scripts/set-pr-status.js
@@ -151,7 +151,12 @@ module.exports = async ({ github, context }) => {
     // Gets all issue numbers that would be closed if the PR is merged.
     function extractLinkedIssueNumbers(prBody) {
         const body = prBody || "";
-        const withoutComments = body.replace(/<!--[\s\S]*?-->/g, "");
+        let withoutComments = body;
+        let previous;
+        do {
+            previous = withoutComments;
+            withoutComments = withoutComments.replace(/<!--[\s\S]*?-->/g, "");
+        } while (withoutComments !== previous);
         const matches = [...withoutComments.matchAll(/(?:close[sd]?|fixe[sd]?|resolve[sd]?)\s+#(\d+)/gi)];
         return matches.map(match => parseInt(match[1], 10));
     }


.github/scripts/set-pr-status.js
```diff
@@ -151,7 +151,12 @@ module.exports = async ({ github, context }) => {
     // Gets all issue numbers that would be closed if the PR is merged.
     function extractLinkedIssueNumbers(prBody) {
         const body = prBody || "";
-        const withoutComments = body.replace(/<!--[\s\S]*?-->/g, "");
+        let withoutComments = body;
+        let previous;
+        do {
+            previous = withoutComments;
+            withoutComments = withoutComments.replace(/<!--[\s\S]*?-->/g, "");
```
iovoid, line 158: Why `\s\S` instead of `.`?

klaus993:
> The do-while loop could potentially become an infinite loop if the regex replacement doesn't converge.

The loop always terminates: each iteration either removes at least one `<!--...-->` match (strictly shortening the string) or leaves the string unchanged (and the loop exits). Since the string length is non-increasing and bounded below by 0, convergence is guaranteed — there's no infinite loop risk.

> A single regex replacement with a more comprehensive pattern can handle nested comments in one pass.

The loop-until-stable approach is the one [recommended by CodeQL's own documentation](https://codeql.github.com/codeql-query-help/javascript/js-incomplete-multi-character-sanitization/) for this exact alert type. It's also easier to reason about correctness compared to a more complex single-pass regex.

.github/scripts/set-pr-status.js
klaus993: In JavaScript, `.` does not match newline characters (`\n`) by default. `[\s\S]` matches any character including newlines, which is needed since HTML comments in PR bodies typically span multiple lines. You could also use the `s` (dotAll) flag with `.`, but `[\s\S]` is the traditional JS idiom and what was already used here before this PR.

PR #6325, opened 2026-03-05
fix(levm): skip nonce validation in eth_call simulations

### LINKED ISSUE #6312: `eth_call` validates nonce and fails when nonce is not provided and nonce(sender) > 0
While deploying an Ethrex L2, I noticed that `eth_call` validates the transaction nonce against the sender account nonce. However, when the caller does not provide a nonce (which is the normal case for read-only calls), it defaults to `0`. As a result, any account with a nonce > 0 receives a `NonceMismatch` error when using `eth_call`.                                                   

`eth_estimateGas` already handles this correctly: when the nonce is not provided, it automatically fetches the account nonce from state ([transaction.rs:454-465](https://github.com/lambdaclass/ethrex/blob/68ac45d60c/crates/networking/rpc/eth/transaction.rs#L454-L465)). However, `eth_call` does not perform this step ([transaction.rs:101-121](https://github.com/lambdaclass/ethrex/blob/68ac45d60c/crates/networking/rpc/eth/transaction.rs#L101-L121)).

For reference, Geth skips nonce validation entirely for `eth_call`. In [core/state_transition.go](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go), the `Message` struct includes a `SkipNonceChecks` flag, which is used in the `preCheck()` method to bypass nonce validation.

A possible fix would be either:
  - auto-populate the nonce from the account state (similar to `eth_estimateGas`), or
  - skip nonce validation for `eth_call`, following Geth's approach.

  If this sounds reasonable, I can open a PR with a fix.

ilitteri:
## Motivation

`eth_call` validates the transaction nonce against the sender account nonce. When the caller does not provide a nonce (the normal case for read-only calls), it defaults to `0`. Any account with nonce > 0 gets a `NonceMismatch` error.

For reference, Geth skips nonce validation entirely for `eth_call` via a `SkipNonceChecks` flag in `core/state_transition.go`.

Closes #6312

## Description

Add a `disable_nonce_check` flag to `Environment` (following the existing `disable_balance_check` pattern) and set it to `true` in `simulate_tx_from_generic`, which is the code path used by `eth_call` and `eth_estimateGas`.

**Changes:**
- `environment.rs`: Added `disable_nonce_check: bool` field
- `backends/levm/mod.rs`: Set flag to `true` in `simulate_tx_from_generic`; set to `false` in explicit `Environment` constructors
- `hooks/default_hook.rs`, `hooks/l2_hook.rs`: Guard nonce mismatch check with `!vm.env.disable_nonce_check`

### DIFF
--- a/crates/vm/levm/src/environment.rs
+++ b/crates/vm/levm/src/environment.rs
@@ -44,6 +44,9 @@ pub struct Environment {
     /// When true, skip balance deduction in `deduct_caller`. Used by the prewarmer
     /// to avoid early reverts on insufficient balance so that warming touches more storage.
     pub disable_balance_check: bool,
+    /// When true, skip nonce validation. Used by `eth_call` simulations
+    /// where the caller typically doesn't provide a nonce.
+    pub disable_nonce_check: bool,
 }
 
 /// This struct holds special configuration variables specific to the
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -1162,6 +1162,7 @@ impl LEVM {
             is_privileged: matches!(tx, Transaction::PrivilegedL2Transaction(_)),
             fee_token: tx.fee_token(),
             disable_balance_check: false,
+            disable_nonce_check: false,
         };
 
         Ok(env)
@@ -1222,6 +1223,7 @@ impl LEVM {
         let mut env = env_from_generic(tx, block_header, db, vm_type)?;
 
         env.block_gas_limit = i64::MAX as u64; // disable block gas limit
+        env.disable_nonce_check = true;
 
         adjust_disabled_base_fee(&mut env);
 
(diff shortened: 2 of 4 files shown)


crates/vm/levm/src/environment.rs
```diff
@@ -44,6 +44,9 @@ pub struct Environment {
     /// When true, skip balance deduction in `deduct_caller`. Used by the prewarmer
     /// to avoid early reverts on insufficient balance so that warming touches more storage.
     pub disable_balance_check: bool,
+    /// When true, skip nonce validation. Used by `eth_call` simulations
+    /// where the caller typically doesn't provide a nonce.
+    pub disable_nonce_check: bool,
```
ElFantasma, line 49: Adding the field here is correct, but since `Environment` doesn't use `..Default::default()` at all construction sites, this will fail to compile. At minimum these files need `disable_nonce_check: false` added:
- `tooling/ef_tests/state_v2/src/modules/runner.rs` (~line 145)
- `tooling/ef_tests/state/runner/levm_runner.rs` (~line 225)
- `test/tests/levm/eip7708_tests.rs` (~line 203)

Issue #797, opened 2024-10-09
Ensure already executed transactions are removed from the tx pool


mpaulucci:
We need a way to ensure that transactions that are part of the canonical chain are removed from the transaction pool.

As part of this task, we should take a look at how Geth does it.

Some possible approaches:
- When a block is set as canonical, we should remove all the transactions of the block that are in the tx pool.
- When there is a reorg, should we re-add the transactions that were canonical and are not anymore?
- What should we do when building a block?

fedacking:
Geth updates the mempool and removes all invalid transactions when a forkchoiceupdate that moves the head is issued. Right now, for our method, each time we execute a transaction we remove the transactions. This is useful right now for L2, but for full L1 support that should be changed.

mpaulucci:
There was some work done with respect to this, we should look into Geth and see what is missing. @rodrigo-o observes that the tx pool is always increasing during syncing

PR #6326, opened 2026-03-06
feat(levm): implement EIP-8141 frame transactions

juanbono:
Implements EIP-8141 Frame Transaction (tx type `0x06`): account-abstraction transactions split into *frames* (VERIFY / SENDER / DEFAULT) with abstract validation and gas payment. Gated behind the Hegota fork; still a Draft EIP.

## What's in it

Execution (LEVM): frame tx type `0x06`, frame loop and modes, atomic batching, the `APPROVE` opcode and the introspection opcodes (`TXPARAM`, `FRAMEDATALOAD`, `FRAMEDATACOPY`, `FRAMEPARAM`, `SIGPARAM`), default code, expiry verifier predeploy at `0x8141`, secp256k1 + P256 signatures, blob-gas accounting, receipts and RPC types.

Mempool: static admission (fork gate, expiry, static constraints, signature auth, intrinsic gas) plus the full EIP §Mempool DoS policy: validation-prefix recognition with structural rules, a VM validation-trace simulation that enforces the banned opcodes and state-access rules during the prefix, the `MAX_VERIFY_GAS` budget, paymaster reservation accounting, one pending frame tx per sender, and revalidation on new canonical blocks.

## Known limitations

- The Draft EIP pins no canonical-paymaster bytecode, so all paymasters are treated as non-canonical: one pending sponsored frame tx per paymaster until the spec pins it.
- Reservations for txs re-queued by a reorg are not rebuilt (tracks the existing mempool reorg TODO).
- Blob-carrying frame txs are rejected at admission (no sidecar transport yet); consensus still accounts for them if included.
- `pending_withdrawal_amount` is treated as 0 (no paymaster withdrawal queue in ethrex).

The mempool policy is a local peer policy: conservative, never affects consensus or block building.


crates/common/types/genesis.rs
```diff
@@ -268,6 +268,19 @@ pub struct ChainConfig {
...
+    /// currently emits `bogotaTime` regardless of which name kurtosis args
+    /// use. Internal name stays `hegota_time` until upstream picks a final
+    /// name; this is purely a parsing alias.
+    #[serde(
+        alias = "hezeTime",
+        alias = "heze_time",
+        alias = "bogotaTime",
+        alias = "bogota_time"
+    )]
+    pub hegota_time: Option<u64>,
```
ElFantasma, line 283: The `hegota_time` field name and the `Fork::Hegota` variant are explicitly placeholders (per the doc comment at line 275: "Internal name stays `hegota_time` until upstream picks a final use"). Worth filing an in-repo tracking issue or TODO with the exact rename path so this doesn't bit-rot — once the EIP picks a fork, every reference (`is_hegota_activated`, `Fork::Hegota`, `hegotaTime` in genesis JSON, the RPC chain-config JSON at `rpc.rs:1528`, error variant `FrameTxPreFork`) needs a coordinated swap. Currently the rename surface is implicit knowledge; future readers landing in `genesis.rs:343` will see `Hegota = 26` without context. Non-blocking — just visibility on the rename touchpoints.

ElFantasma (approved):
Scope is large (+11646/-129 across 55 files, ~4944 LOC) but well-organized; I focused on the consensus-critical surface: frame-tx execution loop in `vm.rs::execute_frame_tx`, the new opcode handlers in `frame_tx.rs`, mempool admission + revalidation in `blockchain.rs`, the receipt RLP layout in `receipt.rs`, and the fork plumbing.

**Where the prior bot reviews landed — verified against current code:**

- **Codex 1st-pass C1 (committed state leak on `tx_invalid`)**: ADDRESSED. `tx_level_backup` absorbs each committed frame's backup (`vm.rs` around L1550, L1683); the `tx_invalid` exit path then calls `restore_cache_state(self.db, tx_level_backup)?` before returning `Err(InvalidFrameTransaction)`. The contract "`Err` leaves `db.current_accounts_state` unchanged" is explicit in the comment.
- **Codex 1st-pass C2 (cumulative log duplication)**: ADDRESSED. Per-frame logs are snapshotted via `self.substate.current_logs()` BEFORE `commit_backup()` (vm.rs:1519), with a comment explaining that calling `extract_logs()` after commit would pull in prior frames' logs. The snapshot feeds `frame_receipts[i].logs` directly.
- **Codex 1st-pass C4 (mempool: unauthenticated sender flood)**: ADDRESSED. `validate_transaction` now performs FULL signature authentication via `validate_frame_signatures` BEFORE any other mempool work, with an explicit comment about "the unauthenticated `sender` field lets fabricated senders flood the pool for free". Static constraints, prefix shape, and gas-budget pre-filter run before that, so unsigned-or-malformed admission is rejected cheaply.
- **Codex 2nd-pass #1 (revalidation doesn't re-check paymaster)**: ADDRESSED. `revalidate_frame_txs_after_block` (blockchain.rs:2830) re-simulates the validation prefix against the new head state AND re-checks `reserved_pending_cost(paymaster) <= balance` per paymaster — the comment walks through why the per-tx prefix simulation is insufficient on its own and why the aggregate invariant is also re-checked.
- **Codex 2nd-pass #2 (replacement race)**: ADDRESSED. The same-nonce predecessor is now kept until AFTER the locked paymaster re-check in `add_transaction` for frame txs (blockchain.rs:2741, comment references "review fix 2"); legacy non-frame replacement still removes-then-adds (unchanged).
- **Kimi #7 (receipt missing-field decode)**: NOT A BUG. The receipt RLP path dispatches on `tx_type_byte`: non-Frame uses the standard consensus decode unchanged, Frame uses the new `[cumulative_gas_used, payer, [frame_receipts]]` path. No graceful-missing-field handling is needed because the wire shape is type-discriminated.

Given that ~5 of the bot-flagged blockers have visible, intentional fixes with rationale comments, the author has clearly been iterating on review feedback — that's the strongest signal here.

**Things I want to flag from this pass:**

- **EIP-8141 spec is still moving** (this branch has explicit alignment commits for spec versions `0b197156`, `d56e73ad`, `235d5056`, `fe0940cae2`, and dropped the original Amsterdam gate in favor of a new Hegota gate mid-PR). Until the EIP exits draft and a fork is named, every spec churn is a re-review. Worth being explicit in the merge plan that this stays behind a not-yet-scheduled fork.
- **Reorg paymaster-reservation gap**: `revalidate_frame_txs_after_block` explicitly documents that txs re-queued from a reorged-out block don't have their reservations reconstructed (ties to existing TODO #797). For a future-fork feature that's a documented known-limitation; before mainnet activation it needs closure.
- **The atomic batch + frame-value-transfer interaction**: the commit `80369cf1e` added a regression test for frame-value unwind on atomic-batch revert specifically; worth surfacing as a path that previously had a bug, so reviewers spot-check `do_frame_value_transfer!` + revert behaviour when the spec settles.
- **Status code 0x3 (skipped) for atomic-batch failures**: pinned to spec commit `235d5056` per the most-recent alignment commit. If the spec moves again, this will need yet another touch.

**No blocking findings against the *current* code.** The shape of the implementation is sound; the open risks are spec-stability and operational/coordination, not code defects.

One small inline on the placeholder fork name.

ilitteri (approved):

crates/vm/levm/src/vm.rs
```diff
@@ -932,6 +1148,931 @@ impl<'a> VM<'a> {
         Ok(report)
     }
 
+    /// Execute a frame transaction (EIP-8141).
+    /// This bypasses the normal prepare/finalize hooks and orchestrates per-frame execution.
```
iovoid, line 1335: The delete_self_destruct_accounts task of the finalize hook should also be executed here.

docs/eip-8141.md
```diff
@@ -0,0 +1,649 @@
...
+
+**ethrex decision**: Uses `Address::from_low_u64_be(0xaa)` as-is (`vm.rs`).
+
+**Impact**: If a contract exists at address `0xaa` on the chain, DEFAULT/VERIFY frames would appear to originate from that contract. On the demo chain this is not an issue. On mainnet, this address should be either reserved or protected from deployment.
+
+### 9. Failed SENDER frame behavior
+
+**Spec gap**: The spec says "if frame reverts, discard state changes and continue." Two things are underspecified: (a) whether execution continues to subsequent frames, and (b) what the top-level receipt `status` should be.
+
+**ethrex decision**: A reverted SENDER frame discards its state changes but execution continues to the next frame. The frame is recorded as failed in `frame_results`. The top-level `succeeded` is set to `false` if any SENDER frame reverted or was skipped (`vm.rs`), matching the standard Ethereum model where `status: 0` means execution reverted (even though the transaction is still included and gas is paid).
```
iovoid: This contradicts with L451, which sets succeeded=false if _any_ frame (not just SENDER) reverts. Receipt encoding vs VM result also has this 'disagreement'. This should be unified or documented.

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -290,6 +290,16 @@ async fn handle_forkchoice(
                     context
                         .blockchain
                         .remove_block_transactions_from_pool(&block)?;
+                    // Re-simulate pending frame txs (EIP-8141) whose validity may
+                    // have changed because of this block, evicting any that no
+                    // longer pass. Best-effort housekeeping (local peer policy):
+                    // a failure must not fail an otherwise-successful FCU, so log
+                    // and continue. Mirrors the stale-blob sweep below.
+                    if let Err(err) = context.blockchain.revalidate_frame_txs_after_block(&block) {
```
iovoid: This is an expensive operation that can block the runtime, it should be inside a spawn_blocking. Even then it's probably better done outside the FCU handler, but that can be deferred.

crates/blockchain/blockchain.rs
```diff
@@ -2748,13 +2756,28 @@ impl Blockchain {
...
         let sender = transaction.sender(&NativeCrypto)?;
         // Validate transaction
-        if let Some(tx_to_replace) = self.validate_transaction(&transaction, sender).await? {
+        let (tx_to_replace, frame_reservation) =
+            self.validate_transaction(&transaction, sender).await?;
+        // For a frame tx, the same-nonce predecessor must NOT be removed here:
+        // `add_transaction` removes it only AFTER the locked paymaster re-check
+        // passes, so a rejected fee-bump leaves the original pending tx intact
+        // (atomic replacement, review fix 2). The hash `add_transaction` removes
+        // (via `check_frame_tx_sender_pending`) is the same same-nonce hash
```
iovoid: check_frame_tx_sender_pending only checks frame transactions elegible for replacement, non-frame transactions would be orphaned instead of properly replaced

iovoid (changes requested):

crates/vm/levm/src/vm.rs
```diff
@@ -932,6 +1148,953 @@ impl<'a> VM<'a> {
...
+                    Ok(ctx_result) => {
+                        let gas_used = ctx_result.gas_used;
+                        let success = ctx_result.is_success();
+
+                        if success {
+                            // Snapshot this frame's own logs before commit_backup merges
+                            // them into the parent substate (required for correct
+                            // frame_receipts[i].logs — walking extract_logs() after commit
+                            // would pull in prior frames' logs).
+                            let this_frame_logs = self.substate.current_logs();
```
iovoid: handle_state_backup inside run_execution already commits it, wouldn't it be duplicated?

crates/vm/levm/src/opcode_handlers/frame_tx.rs
```diff
@@ -0,0 +1,618 @@
...
+        // that FRAMEDATACOPY outside a frame tx halts exactly like INVALID
+        // regardless of operands (consensus parity with other clients).
+        let ctx = vm
+            .frame_tx_context
+            .as_ref()
+            .ok_or(ExceptionalHalt::InvalidOpcode)?;
+
+        if length == 0 {
+            return Ok(OpcodeResult::Continue);
+        }
```
iovoid: This check should be below the frame id validation

crates/common/types/transaction.rs
```diff
@@ -1628,6 +1682,749 @@ pub struct FeeTokenTransaction {
...
+        let mut sigs_buf = Vec::new();
+        self.signatures.encode(&mut sigs_buf);
+        for byte in &sigs_buf {
+            calldata_gas = calldata_gas.saturating_add(if *byte == 0 { 4 } else { 16 });
+        }
+        FRAME_TX_INTRINSIC_COST
+            .saturating_add((self.frames.len() as u64).saturating_mul(FRAME_TX_PER_FRAME_COST))
+            .saturating_add(calldata_gas)
+            .saturating_add(self.signature_verification_cost())
+            .saturating_add(self.frames.iter().map(|f| f.gas_limit).sum::<u64>())
```
iovoid: `.sum()` can overflow and, if overflow checks are enabled, panic

crates/vm/levm/src/vm.rs
iovoid: I think this still applies (no related commits).

It also might show up through address warmth and the selfdestruct set, so I suggest care when fixing.

iovoid:
This conversation https://github.com/lambdaclass/ethrex/pull/6326#discussion_r3459923251 still aplies

iovoid (approved):
Would be nice to have some more tests for the edge cases we found, but otherwise looks good.

PR #6327, opened 2026-03-06
feat(l1): implement EIP-7975 eth/70 partial receipt fetching

edg-l:
## Summary

Implements [EIP-7975](https://eips.ethereum.org/EIPS/eip-7975) — a new `eth/70` protocol version that extends the `GetReceipts` and `Receipts` messages with partial receipt list support.

As Ethereum scales with higher block gas limits, complete block receipt lists can exceed the 10 MiB devp2p message size limit, causing synchronization failures. eth/70 solves this by adding:

- **`GetReceipts` (eth/70)**: `[request-id, firstBlockReceiptIndex, [blockhash₁, ...]]` — allows the client to request receipts starting from a specific index in the first block (for pagination of large receipt lists).
- **`Receipts` (eth/70)**: `[request-id, lastBlockIncomplete, [[receipt₁, ...], ...]]` — includes a flag indicating whether the last block's receipt list is incomplete, signaling the client to make a follow-up request.

### Changes

**New files:**
- `crates/networking/p2p/rlpx/eth/eth70/` — `GetReceipts70`, `Receipts70`, and `StatusMessage70` message types with RLP encode/decode

**Modified files:**
- `message.rs` — `V70` variant in `EthCapVersion`, new `Message` enum variants, updated dispatch (decode/encode/code/request_id/Display)
- `p2p.rs` — `Capability::eth(70)` added to `SUPPORTED_ETH_CAPABILITIES`
- `server.rs` — V70 capability negotiation, Status70 handling, `GetReceipts70` handler with 10 MiB soft response limit and partial receipt pagination
- `store.rs` — `get_receipts_for_block_from_index()` for partial receipt retrieval

### Server-side partial receipt logic
- Accumulates receipts per block using `RLPEncode::length()` for zero-allocation size estimation
- Stops when the 10 MiB soft limit would be exceeded and sets `lastBlockIncomplete = true`
- Guarantees at least one receipt is always included, even if it alone exceeds the limit
- Avoids pushing empty trailing receipt lists that could mislead peers

## Test plan

- [x] 21 new unit tests for `GetReceipts70` and `Receipts70` roundtrip encoding
- [x] Wire format compatibility tests (eth/70 differs from eth/69, cross-decode rejected)
- [x] All 93 p2p crate tests pass
- [x] `cargo fmt`, `make lint` clean
- [ ] Hive eth/70 protocol tests (when available upstream)


crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1016,6 +1030,62 @@ async fn handle_incoming_message(
...
+        Message::GetReceipts70(GetReceipts70 {
+            id,
+            first_block_receipt_index,
+            block_hashes,
+        }) if peer_supports_eth => {
+            let mut all_receipts: Vec<Vec<Receipt>> = Vec::new();
+            let mut total_size: usize = 0;
+            let mut last_block_incomplete = false;
+
+            for (i, hash) in block_hashes.iter().enumerate() {
```
ElFantasma, line 1041: The `block_hashes` vector from the peer has no size bound — a malicious peer could send a request with thousands of hashes, each triggering a separate `get_receipts_for_block_from_index` call (and a separate DB read transaction). Consider capping this, e.g.:

```rust
let block_hashes = &block_hashes[..block_hashes.len().min(256)];
```

Other Ethereum clients (e.g. geth) impose similar limits on batched receipt requests. The existing eth68/69 `GetReceipts` handler has the same issue, but since you're writing a new handler this is a good opportunity to add the check.

crates/networking/p2p/rlpx/message.rs
```diff
@@ -81,8 +89,10 @@ pub enum Message {
     GetPooledTransactions(GetPooledTransactions),
     PooledTransactions(PooledTransactions),
     GetReceipts(GetReceipts),
+    GetReceipts70(GetReceipts70),
```
iovoid: Maybe rename GetReceipts to add the version number to make it explicit? We do this on the other messages.

edg-l: done

PR #6328, opened 2026-03-06
perf(l1): wrap ChainConfig in Arc to avoid 624-byte copy on Store::clone

### LINKED ISSUE #5127: [PERF] Avoid copying a `ChainConfig` on each `Store::clone`

aswinpq:
Fixes #5127

[Store](cci:2://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:153:0-189:1) is cloned extensively as a handle. Since #5071, each clone copies
ChainConfig (624 bytes). This wraps it in Arc so clones only copy an
8-byte pointer.

- Store.chain_config: ChainConfig → Arc<ChainConfig>
- set_chain_config() wraps in Arc::new()
- get_chain_config() returns clone via deref (no API change)

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -155,7 +155,8 @@ pub struct Store {
     /// Storage backend (InMemory or RocksDB).
     backend: Arc<dyn StorageBackend>,
     /// Chain configuration (fork schedule, chain ID, etc.).
-    chain_config: ChainConfig,
+    /// Wrapped in `Arc` to avoid copying 624 bytes on each `Store::clone`.
+    chain_config: Arc<ChainConfig>,
     /// Cache for trie nodes from recent blocks.
     trie_cache: Arc<RwLock<Arc<TrieLayerCache>>>,
     /// Channel for controlling the FlatKeyValue generator background task.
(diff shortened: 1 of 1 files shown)


crates/storage/store.rs
```diff
@@ -2152,7 +2153,7 @@ impl Store {
     }
 
     pub fn get_chain_config(&self) -> ChainConfig {
-        self.chain_config
+        (*self.chain_config).clone()
```
ElFantasma, line 2156: nit: `ChainConfig` is `Copy`, so you can simply write `*self.chain_config` instead of `(*self.chain_config).clone()`. The explicit `.clone()` is unnecessary for Copy types and slightly misleading — it suggests the type might be expensive to clone.

ElFantasma (approved):
Just one nit

PR #6329, opened 2026-03-06
perf(l1): lock-free ArcSwap for trie cache + O(1) canonical TX index

aswinpq:
**Motivation**

The storage layer has two performance bottlenecks on hot paths:

1. **Lock contention on reads**: `trie_cache` uses `Arc<RwLock<Arc<TrieLayerCache>>>` and [LatestBlockHeaderCache](cci:2://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:3109:0-3111:1) uses `Arc<Mutex<Arc<BlockHeader>>>`. Every trie lookup and RPC header request acquires a lock, even though reads vastly outnumber writes (RCU pattern).

2. **O(k) transaction lookups**: [get_transaction_location](cci:1://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:568:4-619:5) uses a `prefix_iterator` scan across all block hashes containing a transaction, then filters for canonical blocks — expensive for frequently-queried transactions.

**Description**

### ArcSwap Migration (Lock-Free Reads)
- `trie_cache`: `Arc<RwLock<Arc<TrieLayerCache>>>` → `Arc<ArcSwap<TrieLayerCache>>`
- [LatestBlockHeaderCache](cci:2://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:3109:0-3111:1): `Arc<Mutex<Arc<BlockHeader>>>` → `ArcSwap<BlockHeader>`
- All read sites use `.load_full()` (lock-free), writes use `.store()` (atomic swap)
- Eliminates `.expect("poisoned mutex")` panics in LatestBlockHeaderCache
- [apply_trie_updates()](cci:1://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:2773:0-2871:1) RCU pattern preserved, now fully lock-free

### Canonical TX Index (O(1) Lookups)
- New `CANONICAL_TX_INDEX` table: `tx_hash → (block_number, block_hash, tx_index)`
- Fast-path added in [get_transaction_location()](cci:1://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:568:4-619:5) with graceful fallback to existing prefix scan
- Index populated during [forkchoice_update_inner()](cci:1://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:1007:4-1071:5) for zero-downtime migration

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the [Store](cci:2://file:///c:/Users/Hi/Documents/zkp%20projects/ethrex/crates/storage/store.rs:153:0-189:1) requiring a re-sync.

> Note: The new `CANONICAL_TX_INDEX` table is auto-created from the `TABLES` array on DB initialization. Existing nodes will have an empty index and fall back to the prefix scan until new blocks populate it. A schema version bump may not be required — happy to add one if maintainers prefer.

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -6,9 +6,10 @@ use crate::{
         StorageBackend, StorageReadView,
         tables::{
             ACCOUNT_CODE_METADATA, ACCOUNT_CODES, ACCOUNT_FLATKEYVALUE, ACCOUNT_TRIE_NODES,
-            BLOCK_NUMBERS, BODIES, CANONICAL_BLOCK_HASHES, CHAIN_DATA, EXECUTION_WITNESSES,
-            FULLSYNC_HEADERS, HEADERS, INVALID_CHAINS, MISC_VALUES, PENDING_BLOCKS, RECEIPTS,
-            SNAP_STATE, STORAGE_FLATKEYVALUE, STORAGE_TRIE_NODES, TRANSACTION_LOCATIONS,
+            BLOCK_NUMBERS, BODIES, CANONICAL_BLOCK_HASHES, CANONICAL_TX_INDEX, CHAIN_DATA,
+            EXECUTION_WITNESSES, FULLSYNC_HEADERS, HEADERS, INVALID_CHAINS, MISC_VALUES,
+            PENDING_BLOCKS, RECEIPTS, SNAP_STATE, STORAGE_FLATKEYVALUE, STORAGE_TRIE_NODES,
+            TRANSACTION_LOCATIONS,
         },
     },
     apply_prefix,
@@ -41,6 +42,7 @@ use ethrex_trie::{Node, NodeRLP};
 use lru::LruCache;
 use rustc_hash::FxBuildHasher;
 use serde::{Deserialize, Serialize};
+use arc_swap::ArcSwap;
 use std::{
     collections::{BTreeMap, HashMap, hash_map::Entry},
     fmt::Debug,
@@ -157,7 +159,8 @@ pub struct Store {
     /// Chain configuration (fork schedule, chain ID, etc.).
     chain_config: ChainConfig,
     /// Cache for trie nodes from recent blocks.
-    trie_cache: Arc<RwLock<Arc<TrieLayerCache>>>,
+    /// Uses `ArcSwap` for lock-free reads (RCU pattern).
+    trie_cache: Arc<ArcSwap<TrieLayerCache>>,
     /// Channel for controlling the FlatKeyValue generator background task.
     flatkeyvalue_control_tx: std::sync::mpsc::SyncSender<FKVGeneratorControlMessage>,
     /// Channel for sending trie updates to the background worker.
@@ -573,7 +576,12 @@ impl Store {
             let tx_hash_bytes = transaction_hash.as_bytes();
             let tx = db.begin_read()?;
 
-            // Use prefix iterator to find all entries with this transaction hash
+            // Fast path: check canonical index first (O(1) lookup)
+            if let Some(value) = tx.get(CANONICAL_TX_INDEX, tx_hash_bytes)? {
+                return Ok(Some(<(BlockNumber, BlockHash, Index)>::decode(&value)?));
+            }
+
+            // Fallback: Use prefix iterator to find all entries with this transaction hash
             let mut iter = tx.prefix_iterator(TRANSACTION_LOCATIONS, tx_hash_bytes)?;
             let mut transaction_locations = Vec::new();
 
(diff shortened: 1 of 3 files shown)


crates/storage/store.rs
```diff
@@ -1014,17 +1022,73 @@ impl Store {
                 let head_key = block_number.to_le_bytes();
                 let head_value = block_hash.encode_to_vec();
                 txn.put(CANONICAL_BLOCK_HASHES, &head_key, &head_value)?;
+
+                // Populate canonical transaction index for O(1) lookups.
```
ElFantasma, line 1047: Bug: stale `CANONICAL_TX_INDEX` entries after reorgs.

The cleanup loop at line 1045 only purges TX index entries for blocks **above** `head_number`. But during a reorg, `new_canonical_blocks` contains blocks that **replace** old canonical blocks at the same numbers. Transactions unique to the old blocks (not in the new chain) keep their stale index entries, pointing to non-canonical blocks.

Example: old chain `1→2→3→4→5`, reorg to `1→2→3'→4'→5'`:
- `new_canonical_blocks` = `[(3, hash3'), (4, hash4')]`, head = `5'`, latest = `5`
- Cleanup loop range `6..=5` is empty — nothing gets purged
- Old blocks 3, 4, and 5's transaction entries remain in `CANONICAL_TX_INDEX`
- `get_transaction_location` fast path returns stale `(3, old_hash3, idx)` for txs only in old block 3

The old slow path explicitly checks `canonical_hash == Some(block_hash)`, so it's immune. But the fast path trusts the index blindly.

Fix: before indexing each entry in `new_canonical_blocks`, read the OLD canonical hash at that number, fetch the old body, and delete its TX entries. Same treatment needed for `head_number` before indexing the head block (see line 1075).

crates/storage/store.rs
```diff
@@ -1014,17 +1022,73 @@ impl Store {
...
+                // Read the block body to get transaction hashes.
+                let hash_key = block_hash.encode_to_vec();
+                if let Some(body_bytes) = {
+                    let read_txn = db.begin_read()?;
```
ElFantasma: nit: A new `db.begin_read()` is created per block inside both this loop and the cleanup loop below (lines 1049, 1055, 1078). Since the write transaction hasn't committed, all these read snapshots see the same pre-transaction state. A single `let read_txn = db.begin_read()?;` created once before both loops would avoid the repeated overhead.

ElFantasma (changes requested):

crates/storage/store.rs
```diff
@@ -1009,22 +1017,106 @@ impl Store {
         let db = self.backend.clone();
         tokio::task::spawn_blocking(move || {
             let mut txn = db.begin_write()?;
+            let read_txn = db.begin_read()?;
 
             for (block_number, block_hash) in new_canonical_blocks {
                 let head_key = block_number.to_le_bytes();
+
+                // Purge old entries from canonical TX index if this block height was already canonical.
```
ElFantasma, line 1025: nit: The purge pattern (read old block body → delete its TX entries from `CANONICAL_TX_INDEX`) is repeated 3 times (here, above-head cleanup, head block). Same for the index pattern (read body → insert TX entries) which appears twice. Consider extracting:

```rust
fn purge_block_tx_index(
    read: &dyn StorageReadView,
    txn: &mut dyn StorageWriteBatch,
    block_hash: &BlockHash,
) -> Result<(), StoreError> {
    let body_key = block_hash.encode_to_vec();
    if let Some(body_bytes) = read.get(BODIES, &body_key)? {
        let body: BlockBody = BlockBodyRLP::from_bytes(body_bytes)
            .to().map_err(StoreError::from)?;
        for tx in body.transactions.iter() {
            txn.delete(CANONICAL_TX_INDEX, tx.hash().as_bytes())?;
        }
    }
    Ok(())
}

fn index_block_txs(
    read: &dyn StorageReadView,
    txn: &mut dyn StorageWriteBatch,
    block_number: BlockNumber,
    block_hash: BlockHash,
) -> Result<(), StoreError> {
    let hash_key = block_hash.encode_to_vec();
    if let Some(body_bytes) = read.get(BODIES, &hash_key)? {
        let body: BlockBody = BlockBodyRLP::from_bytes(body_bytes)
            .to().map_err(StoreError::from)?;
        for (i, tx) in body.transactions.iter().enumerate() {
            let val = (block_number, block_hash, i as u64).encode_to_vec();
            txn.put(CANONICAL_TX_INDEX, tx.hash().as_bytes(), &val)?;
        }
    }
    Ok(())
}
```

The types (`dyn StorageReadView` / `dyn StorageWriteBatch`) are already trait objects so this works without generics.

ElFantasma (changes requested):

PR #6330, opened 2026-03-06
fix(levm): make finalize mutations atomic with rollback

iovoid:
**Motivation**

When we attempt to execute a transaction, and it fails during finalization (such as due to the fee contract token reverting, or more rarely an overflow) we leave the state inconsistent.

**Description**

- Rearrange finalize_non_privileged_execution into two phases: fallible computations first, then all state mutations grouped in apply_finalize_mutations with restore_cache_state() on error
- Prevents partial fee payments from leaking into DB when a mutation fails mid-way

Two regression tests are added, that should fail without the fix:
 - Fee token revert: fee token contract reverts on payFee during finalization, asserts that undo_value_transfer is also rolled back (contract retains transferred value)
 - Overflow: operator fee vault at U256::MAX causes pay_operator_fee to overflow, asserts that coinbase and vault balances unchanged after rollback

### DIFF
--- a/crates/vm/levm/src/hooks/l2_hook.rs
+++ b/crates/vm/levm/src/hooks/l2_hook.rs
@@ -145,6 +145,9 @@ fn finalize_non_privileged_execution(
     // === Phase 1: Fallible computations (no state mutations) ===
     // Perform contract calls and conversions that can fail BEFORE any
     // mutations, so an error here leaves the DB state unchanged.
+    // NOTE: get_fee_token_ratio now runs before delete_self_destruct_accounts
+    // (which moved into Phase 2). This is safe because the fee token ratio
+    // contract should never be in the selfdestruct set.
     let fee_token_ratio: u64 = if let Some(fee_token) = vm.env.fee_token {
         get_fee_token_ratio(vm, fee_token)?
             .try_into()
@@ -175,6 +178,11 @@ fn finalize_non_privileged_execution(
     );
 
     if let Err(e) = result {
+        // Rollback DB cache to undo partial Phase 2 mutations.
+        // Note: substate (logs, selfdestruct) is NOT rolled back here because
+        // the substate parent was already consumed by handle_state_backup()
+        // during run_execution(). This is safe because the Err propagates to
+        // the caller, which discards the entire VM context.
         vm.restore_cache_state()?;
         return Err(e);
     }
@@ -698,6 +706,12 @@ fn transfer_fee_token(vm: &mut VM<'_>, data: Bytes) -> Result<(), VMError> {
     for (key, new_value) in &new_storage {
         let old_value = current_storage.get(key).copied().unwrap_or_default();
         if old_value != *new_value {
+            vm.backup_storage_slot(fee_token, *key, old_value)?;
+        }
+    }
+    // Back up slots that will be removed by the bulk replacement
+    for (key, &old_value) in &current_storage {
+        if !new_storage.contains_key(key) {
             vm.backup_storage_slot(fee_token, *key, old_value)?;
         }
     }
--- a/test/tests/levm/l2_hook_tests.rs
+++ b/test/tests/levm/l2_hook_tests.rs
@@ -280,16 +280,16 @@ fn finalize_mutation_failure_reverts_all_changes() {
 }
 
 /// Fee token variant: the fee token contract reverts on `payFee` calls during
-/// finalization, triggering rollback of `undo_value_transfer` mutations.
+/// finalization, triggering rollback of Phase 2 mutations only.
 ///
 /// Flow:
 /// 1. prepare_execution_fee_token: lockFee succeeds, value transferred to contract
 /// 2. Execution: called contract REVERTs
-/// 3. Finalize: undo_value_transfer runs (backed up after prepare clears backup)
+/// 3. Finalize: undo_value_transfer runs, then backup is cleared (checkpoint after undo)
 /// 4. apply_finalize_mutations → refund_sender_fee_token → pay_fee_token → REVERT
-/// 5. restore_cache_state undoes undo_value_transfer
+/// 5. restore_cache_state undoes only Phase 2 mutations (undo_value_transfer preserved)
 ///
-/// Assertions verify the contract still holds the value (undo was rolled back).
+/// Assertions verify the contract has 0 balance (value returned to sender, not rolled back).
 #[test]
 fn fee_token_revert_during_finalize_triggers_rollback() {
     let sender = Address::from_low_u64_be(SENDER);


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -152,6 +153,48 @@ fn finalize_non_privileged_execution(
         1u64
     };
 
+    // === Phase 2: State mutations (with rollback on error) ===
+    // All mutations go through get_account_mut which records the original
```
ElFantasma: This comment is slightly misleading: `get_account_mut` records the original account **info** (balance, nonce, code_hash), but `transfer_fee_token` also does `vm.db.get_account_mut(fee_token)?.storage = fee_storage` which replaces the entire storage map — that replacement isn't tracked by the backup mechanism. So in the fee token path, storage mutations bypass the backup and won't be rolled back.

See the review body for a fuller explanation of the scenario.

crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -152,6 +153,48 @@ fn finalize_non_privileged_execution(
...
+    fee_config: &FeeConfig,
+    use_fee_token: bool,
+    fee_token_ratio: u64,
+    l1_gas: u64,
+    gas_refunded: u64,
+    execution_gas: u64,
+    actual_gas_used: u64,
+    total_gas_pre_refund: u64,
+) -> Result<(), crate::errors::VMError> {
+    default_hook::delete_self_destruct_accounts(vm)?;
```
ElFantasma, line 208: nit: On `main`, `delete_self_destruct_accounts` runs *before* `get_fee_token_ratio`. Now it runs after (inside Phase 2, while `get_fee_token_ratio` is in Phase 1). The reordering is likely safe since `get_fee_token_ratio` simulates on a cloned DB and system contracts at `FEE_TOKEN_RATIO_ADDRESS` shouldn't be in the self-destruct set — but worth being aware of the ordering change.

crates/vm/levm/src/hooks/l2_hook.rs
iovoid: Improved in 8a29006

crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -637,8 +680,25 @@ fn transfer_fee_token(vm: &mut VM<'_>, data: Bytes) -> Result<(), VMError> {
...
+    }
+
+    // Apply changed slots
+    let account = vm.db.get_account_mut(fee_token)?;
+    for (key, new_value) in new_storage {
+        let old_value = current_storage.get(&key).copied().unwrap_or_default();
+        if old_value != new_value {
+            account.storage.insert(key, new_value);
+        }
+    }
```
avilagaston9: nit: The slot-by-slot apply loop is unnecessary — the backup already recorded the old values for rollback, so the apply doesn't need to be selective. A direct assignment is simpler and faster (single pointer swap vs iterating every entry):

```suggestion
    // Apply new storage
    vm.db.get_account_mut(fee_token)?.storage = new_storage;
```

test/tests/levm/l2_hook_tests.rs
```diff
@@ -0,0 +1,396 @@
...
+
+    // Verify that undo_value_transfer was rolled back: the contract still
+    // holds the value from prepare_execution's transfer_value (the backup was
+    // cleared after prepare, so restore goes back to post-prepare state).
+    let contract_balance = db.get_account(contract_addr).unwrap().info.balance;
+    assert_eq!(
+        contract_balance, tx_value,
+        "Contract should still hold transferred value after rollback (undo_value_transfer reverted)"
+    );
+}
```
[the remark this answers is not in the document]
avilagaston9: Agree with this finding. The test is asserting the buggy behavior — `undo_value_transfer` should stick even when Phase 2 mutations fail, so the contract should have 0 balance, not `tx_value`.

The root cause is that the backup isn't cleared after the gas-overuse handling block (lines 122-139). Since `backup_account_info` uses `or_insert`, the backup still holds pre-execution entries from the first `undo_value_transfer`. When Phase 2 fails and `restore_cache_state()` runs, it restores to the post-prepare state, incorrectly undoing `undo_value_transfer`.

Fix: add `vm.current_call_frame.call_frame_backup.clear()` after the gas-overuse `if` block (after line 139), then update this assertion to `assert_eq!(contract_balance, U256::zero())`.

crates/vm/levm/src/hooks/l2_hook.rs
iovoid: Done in c0e9d20a6

crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -644,8 +691,19 @@ fn transfer_fee_token(vm: &mut VM<'_>, data: Bytes) -> Result<(), VMError> {
...
+    // Back up original values for changed slots so restore_cache_state can revert them
+    for (key, new_value) in &new_storage {
+        let old_value = current_storage.get(key).copied().unwrap_or_default();
+        if old_value != *new_value {
+            vm.backup_storage_slot(fee_token, *key, old_value)?;
+        }
+    }
+
+    // Apply new storage
+    vm.db.get_account_mut(fee_token)?.storage = new_storage;
```
avilagaston9, line 720: The backup loop only visits `new_storage` keys, so slots present in `current_storage` but absent from `new_storage` (removed by the simulation) aren't backed up and would be lost on rollback.

```suggestion
    // Back up original values for changed slots so restore_cache_state can revert them
    for (key, new_value) in &new_storage {
        let old_value = current_storage.get(key).copied().unwrap_or_default();
        if old_value != *new_value {
            vm.backup_storage_slot(fee_token, *key, old_value)?;
        }
    }
    // Back up slots that will be removed by the bulk replacement
    for (key, &old_value) in &current_storage {
        if !new_storage.contains_key(key) {
            vm.backup_storage_slot(fee_token, *key, old_value)?;
        }
    }

    // Apply new storage
    vm.db.get_account_mut(fee_token)?.storage = new_storage;
```

Low risk in practice (fee token simulations rarely remove slots), but closes the rollback gap.

test/tests/levm/l2_hook_tests.rs
```diff
@@ -94,10 +95,311 @@ const SENDER: u64 = 0x1000;
...
+/// finalization, triggering rollback of `undo_value_transfer` mutations.
+///
+/// Flow:
+/// 1. prepare_execution_fee_token: lockFee succeeds, value transferred to contract
+/// 2. Execution: called contract REVERTs
+/// 3. Finalize: undo_value_transfer runs (backed up after prepare clears backup)
+/// 4. apply_finalize_mutations → refund_sender_fee_token → pay_fee_token → REVERT
+/// 5. restore_cache_state undoes undo_value_transfer
+///
+/// Assertions verify the contract still holds the value (undo was rolled back).
```
avilagaston9: The docstring still describes the old (incorrect) behavior:
- Step 5 says "restore_cache_state undoes undo_value_transfer"
- Last line says "Assertions verify the contract still holds the value (undo was rolled back)"

But the test now correctly asserts `contract_balance == U256::zero()` (undo_value_transfer is preserved, not rolled back). The docstring should be updated to match.

PR #6333, opened 2026-03-06
fix(levm,l2): regression test for validation failure after fee token deduction

iovoid:
**Motivation**

Currently we deduct the fee token before performing other validations that could still fail. 

**Description**

The fix ended up as part of #6330, but we still want a regression test.

### DIFF
--- a/test/tests/levm/l2_fee_token_tests.rs
+++ b/test/tests/levm/l2_fee_token_tests.rs
@@ -1,0 +1,231 @@
+//! Regression tests for fee-token lock ordering (Audit Finding L).
+//!
+//! Issue: `deduct_caller_fee_token` mutates fee-token contract storage via
+//! `db.get_account_mut()` (bypassing call-frame backup) before later validation
+//! steps that can still fail. If validation fails, the fee-token lock persists.
+
+use bytes::Bytes;
+use ethrex_common::{
+    Address, H256, U256,
+    constants::EMPTY_TRIE_HASH,
+    types::{
+        Account, AccountState, ChainConfig, Code, CodeMetadata, EIP1559Transaction, Fork,
+        Transaction, TxKind,
+    },
+};
+use ethrex_crypto::NativeCrypto;
+use ethrex_levm::{
+    db::{Database, gen_db::GeneralizedDatabase},
+    environment::{EVMConfig, Environment},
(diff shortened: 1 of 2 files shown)


crates/vm/levm/src/hooks/l2_hook.rs
```diff
@@ -545,6 +542,13 @@ fn prepare_execution_fee_token(vm: &mut VM<'_>) -> Result<(), crate::errors::VME
     // Transaction is type 4 if authorization_list is Some
     // NOT CHECKED: fee token transactions are not type 4
 
+    // (10) INSUFFICIENT_ACCOUNT_FUNDS
+    // Deduct fee-token fees AFTER all fallible validations above. The fee-token lock
+    // mutates contract storage via db.get_account_mut() which bypasses the call-frame
+    // backup mechanism, so it cannot be reverted by restore_cache_state(). Moving it
+    // here ensures fees are only locked once the tx is certain to proceed.
```
ElFantasma: nit: "certain to proceed" is slightly strong — `transfer_value()` and `set_bytecode_and_code_address()` right below are still fallible (e.g., balance overflow in `increase_account_balance`). In practice they won't fail for a tx that passed all the validations above, but strictly speaking this isn't a guarantee. Maybe "highly unlikely to fail" or just drop that clause:

```suggestion
    // here ensures fees are only locked after all other validations have passed.
```

avilagaston9 (changes requested):
PR #6330 addresses a related issue (non-atomic finalization mutations) and fixes the underlying mechanism by making `transfer_fee_token` properly participate in the call-frame backup. That same fix would also solve this bug, since `restore_cache_state()` would correctly revert the fee-token lock when `prepare_execution` fails.

The rollback approach is preferable because it fixes the root cause — with reordering, any fallible step added after the fee-token deduction in the future would silently reintroduce the bug.

**Validation step numbering**

The numbered comments (`// (3) INSUFFICIENT_ACCOUNT_FUNDS`, `// (4) INSUFFICIENT_MAX_FEE_PER_GAS`, etc.) mirror the step numbering in `default_hook.rs`. This 1:1 correspondence makes it easy to verify that the fee-token path covers every validation the default path does. It'd be nice to preserve it.

PR #3866, opened 2025-07-29
feat(l2): discard invalid batch proofs

LeanSerra:
**Motivation**

In case a verify_batch transaction failed, the proof_sender would get stuck because the old proof would be retried instead of asking for a new one

**Description**

- Add a function to rollup_store to delete a proof
- Modify OnChainProposer to return an error message indicating which proof was determined to be invalid
- Modify l1_proof_sender to identify the the invalid proof and delete it

**How to test**

Build a working sp1 prover
```
make build-prover PROVER=sp1 G=true
```

copy the built binary to a different folder because we are going to overwrite it
```
cp target/release/ethrex_prover <destination-folder>
```

start a local testnet with sp1 enabled (set `--sp1.deploy-verifier` on l2 Makefile deploy-l1 target)
```
ETHREX_PROOF_COORDINATOR_DEV_MODE=false make restart
```

start the valid prover and check that the batches are correctly proved
```
cd <destination-folder> 
```

```
SP1_PROVER=cuda ./ethrex_prover --http.addr 127.0.0.1 --http.port 3900 --log.level debug
```

kill the prover with ctrl-c

modify the prover code to generate invalid proofs (I commented all the code from `execution_program` function and returned a `ProgramOutput` with everything set to 0)

```
make build-prover PROVER=sp1 G=true
```

and

```
SP1_PROVER=cuda make init-prover PROVER=sp1
```

you should see on the sequencer logs  `Deleting invalid SP1 proof` you can then kill this invalid prover with ctrl-c and restart the original correct prover. Now the valid prover should receives the batch that was previously sent to the invalid prover and batches should start verifying

PR #6334, opened 2026-03-06
fix(l2): add documentation about invalid proof deletion

iovoid:
**Motivation**

If we somehow generate an invalid proof or some attacker manages to connect to the proof collector, we want to delete the broken proof and generate a real one.

On aligned mode we don't know which proof caused the issue, so we remove all of them.

This behavior was implemented in #3866, but the assumptions weren't well-documented.

**Description**

Explains the assumption that the only errors a invalid proof could cause look like InvalidXYZProof.

### DIFF
--- a/crates/l2/sequencer/l1_proof_sender.rs
+++ b/crates/l2/sequencer/l1_proof_sender.rs
@@ -490,6 +490,12 @@ impl L1ProofSender {
 
     /// Returns the prover type whose proof is invalid based on the RPC error data
     /// (custom error selector), or `None` if the data doesn't indicate an invalid proof.
+    ///
+    /// These three selectors cover all proof-related errors:
+    /// `_verifyBatchInternal` wraps each verifier call in try/catch, so any
+    /// verifier failure becomes one of these. Other reverts (BatchNotSequential,
+    /// etc.) are ordering issues where deleting proofs would be wrong.
+    /// Aligned mode uses `AlignedProofVerificationFailed` instead (see l1_proof_verifier).
     fn invalid_proof_type_from_data(data: &str) -> Option<ProverType> {
         if data.starts_with(INVALID_TDX_PROOF_SELECTOR) {
             Some(ProverType::TDX)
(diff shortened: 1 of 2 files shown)


crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -514,6 +515,16 @@ impl L1ProofSender {
             self.rollup_store
                 .delete_proof_by_batch_and_type(batch_number, proof_type)
                 .await?;
+        } else {
+            warn!(
+                "L1 rejected batch {batch_number} with unrecognized error, \
```
ElFantasma: The warning doesn't include the actual error `data` that triggered this path. Including it would help operators understand why proofs were deleted and whether the error is a real on-chain rejection vs something unexpected.

```suggestion
            warn!(
                "L1 rejected batch {batch_number} with unrecognized error ({data}), \
                 deleting all proofs so provers can regenerate"
            );
```

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -514,6 +515,16 @@ impl L1ProofSender {
...
+                 deleting all proofs so provers can regenerate"
+            );
+            for proof_type in &self.needed_proof_types {
+                self.rollup_store
```
ElFantasma: `try_delete_invalid_proof` is called for every `RPCError`, which can include non-proof-related failures (gas estimation, nonce issues, etc.). With this `else` branch, any such transient error would delete all proofs — re-proving is expensive.

Worth considering: should this only run when the error message looks like an execution revert (e.g., contains `"execution reverted"` or a known revert-data prefix)? That way transient RPC failures don't unnecessarily trigger re-proving.

iovoid: Fixed in 93c9cc8

crates/l2/sequencer/l1_proof_sender.rs
iovoid: Done

crates/l2/sequencer/l1_proof_sender.rs
```diff
@@ -514,6 +516,17 @@ impl L1ProofSender {
             self.rollup_store
                 .delete_proof_by_batch_and_type(batch_number, proof_type)
                 .await?;
+        } else if data.starts_with(ALIGNED_PROOF_VERIFICATION_FAILED_SELECTOR) {
+            warn!(
```
avilagaston9: I would upgrade these new logs to `error!`. We may also want to combine this with the solution in #6313

iovoid: Done in 01be806b9

PR #6335, opened 2026-03-06
fix(l2): advance L1 watcher cursor only after processing

iovoid:
**Motivation**

Currently if there is an error while adding a privileged transaction, we permanently ignore it. This leaves a gap in processed privileged transactions, eventually leading us to exceed the inclusion deadline.

**Description**

We advance the processing cursor only after a success so it can be retried, and actually generate an error instead of silently discarding it.

### DIFF
--- a/crates/l2/sequencer/l1_watcher.rs
+++ b/crates/l2/sequencer/l1_watcher.rs
@@ -308,7 +308,11 @@ impl L1Watcher {
                 .await
                 .map_err(|e| {
                     L1WatcherError::Custom(format!(
-                        "Failed to add mint transaction to the mempool: {e:#?}"
+                        "Failed to add mint transaction to the mempool \
+                        (to: {:x}, value: {}, transactionId: {:#}): {e:#?}",
+                        privileged_transaction_data.to_address,
+                        privileged_transaction_data.value,
+                        privileged_transaction_data.transaction_id
                     ))
                 })?;
 


crates/l2/sequencer/l1_watcher.rs
```diff
@@ -165,15 +165,20 @@ impl L1Watcher {
...
-            let _ = self
-                .process_privileged_transactions(logs)
-                .await
-                .inspect_err(|err| error!("L1 Watcher Error: {}", err));
-        };
+        if !logs.is_empty()
+            && let Err(err) = self.process_privileged_transactions(logs).await
+        {
+            error!("L1 Watcher Error: {err}");
+            // Don't advance cursor — next tick will re-fetch and retry.
```
ElFantasma, line 172: Worth considering: if `add_transaction_to_pool` returns a *permanent* error for a specific tx (e.g., the tx is inherently invalid), this will block the cursor forever — every tick will re-fetch the same block range, re-encounter the same tx, and fail again.

The `privileged_transaction_already_processed` check only skips txs that were *successfully* added previously, so a permanently failing tx would loop infinitely.

Might be worth adding a bounded retry count or logging at a higher severity after N consecutive failures for the same tx, so operators can intervene. Not necessarily blocking for this PR, but worth tracking.

iovoid: Yeah the privileged transactions need to be ordered, so it's correct behavior.

As for logging, not sure what's the best way to bring the operator's attention. Exiting? The L2 will eventually fail due to the inclusion deadline.

crates/l2/sequencer/l1_watcher.rs
```diff
@@ -183,17 +188,15 @@ impl L1Watcher {
                     .await?
                     .into();
         }
```
avilagaston9, line 190: Nit: since this error now aborts the whole batch and triggers a retry, it might help operators to also log which transaction failed (e.g. `transaction_id`, `to_address`, `value`) in the error message, so they can quickly identify a persistently-failing privileged tx.

iovoid: Done in 7f96ad03

PR #6336, opened 2026-03-07
fix(l1): remove value from branch node to enforce empty invariant (#5772)

### LINKED ISSUE #5772: Set trie branch values to empty at the type level
We currently assume that branches don't have values, and our API keeps this assumption. However, we don't currently have this proven at the type-level.

aswinpq:
Fixes #5772

## Motivation
Previously, our API assumed that [BranchNode](cci:2://file:///crates/common/trie/node/branch.rs:22:0-24:1)s don't hold values, but the struct still contained a [value](cci:1://file:///zkp%20projects/ethrex/crates/storage/store.rs:2659:4-2663:5) field. We want to strictly enforce this invariant at the compiler level.

Also, regarding the concern about [withdrawals_root](cci:1://file:///crates/common/types/block.rs:351:0-358:1) computation: I verified that the root computations use sequential `idx.encode_to_vec()` RLP keys.

 These indices never produce prefix collisions that land exactly on a 16th terminator nibble, so they always securely terminate in [LeafNode](cci:2://file:///crates/common/trie/node/leaf.rs:27:0-30:1)s. The branch [value](cci:1://file:////crates/storage/store.rs:2659:4-2663:5) field is safely unused everywhere.

## Description
This PR makes [BranchNode](cci:2:///crates/common/trie/node/branch.rs:22:0-24:1) structural by completely removing its [value](cci:1://file:///zkp%20projects/ethrex/crates/storage/store.rs:2659:4-2663:5) field.

* **[branch.rs](cci:7://file:///crates/common/trie/node/branch.rs:0:0-0:0)**: Removed `value: ValueRLP`, `new_with_value()`, and [update()](cci:1://file:///crates/storage/store.rs:3104:4-3107:5).
* **[leaf.rs](cci:7://file:///crates/common/trie/node/leaf.rs:0:0-0:0) & [extension.rs](cci:7://file:///crates/common/trie/node/extension.rs:0:0-0:0)**: In the impossible edge-case where a key insertion tries to terminate exactly *on* an existing branch, it now explicitly returns `TrieError::Verify` instead of silently allocating a branch value.
* **[rlp.rs](cci:7://file:///crates/common/trie/rlp.rs:0:0-0:0)**: Branch payload serialization now hardcodes the 17th item as an empty byte slice (`&[]`) to maintain network encoding compatibility.
* **Tests**: Cleaned up [verify_range.rs](cci:7://file:///crates/common/trie/verify_range.rs:0:0-0:0), [trie_iter.rs](cci:7://file:///crates/common/trie/trie_iter.rs:0:0-0:0), and removed the `with_leaf` test macro since branch values can no longer be artificially constructed.

### DIFF
--- a/crates/common/trie/node/branch.rs
+++ b/crates/common/trie/node/branch.rs
@@ -1,5 +1,3 @@
-use std::mem;
-
 use ethrex_rlp::encode::RLPEncode;
 
 use crate::{
@@ -10,7 +8,7 @@ use crate::{
 use super::{ExtensionNode, LeafNode, Node, NodeRef, ValueOrHash};
 
 /// Branch Node of an an Ethereum Compatible Patricia Merkle Trie
-/// Contains the node's value and the hash of its children nodes
+/// Contains the hash of its children nodes
 #[derive(
     Debug,
     Clone,
@@ -24,7 +22,6 @@ use super::{ExtensionNode, LeafNode, Node, NodeRef, ValueOrHash};
 )]
 pub struct BranchNode {
     pub choices: [NodeRef; 16],
-    pub value: ValueRLP,
 }
 
 impl BranchNode {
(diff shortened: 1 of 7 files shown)


crates/common/trie/node/branch.rs
```diff
@@ -213,13 +187,11 @@ impl BranchNode {
...
-        let new_node = match (children.len(), !self.value.is_empty()) {
-            // If this node still has a value but no longer has children, convert it into a leaf node
-            (0, true) => NodeRemoveResult::New(
-                LeafNode::new(Nibbles::from_hex(vec![16]), mem::take(&mut self.value)).into(),
-            ),
-            // If this node doesn't have a value and has only one child, replace it with its child node
-            (1, false) => {
+        let new_node = match children.len() {
+            // If this node doesn't have children, the trie is empty
+            0 => NodeRemoveResult::Mutated,
```
ElFantasma, line 192: The comment says 0 remaining children is impossible, but if it ever happens, returning `Mutated` silently keeps an empty (childless, valueless) branch node in the trie — which would corrupt state. Consider adding a `debug_assert!(false, "branch with 0 children after removal")` or returning an error here so violations are caught early rather than silently propagated.

crates/common/trie/node/leaf.rs
```diff
@@ -70,32 +70,9 @@ impl LeafNode {
...
-                    choices,
-                    match value {
-                        ValueOrHash::Value(value) => value,
-                        // Value in branches don't happen in our use-case.
-                        ValueOrHash::Hash(_) => todo!("handle override case (error?)"),
-                    },
-                )
+            let branch_node = if self_choice_idx == 16 || new_leaf_choice_idx == 16 {
+                return Err(TrieError::Verify("Key paths may not be prefixes of each other since branch values are not supported".to_string()));
+
```
ElFantasma, line 75: nit: the blank line after the `return Err(...)` (before the closing `} else {`) could be removed — it's the only branch in this match with trailing whitespace.

ElFantasma (changes requested):

PR #6342, opened 2026-03-08
fix(levm): validate remaining block gas in parallel pipeline

### LINKED ISSUE #6341: Parallel LEVM pipeline skips remaining block gas validation
## Summary
The Amsterdam BAL-backed parallel execution path in `LEVM::execute_block_pipeline` does not enforce the same remaining block gas check that the sequential path does.

## Why this matters
The sequential path rejects a transaction when `cumulative_gas_used + tx.gas_limit() > block.header.gas_limit`, using post-refund cumulative gas per EIP-7778.

The parallel path currently executes every transaction, sorts the results, and builds receipts without any equivalent `check_gas_limit` step. That means a block can get through the parallel validation path even when a later transaction's `gas_limit` is larger than the remaining block gas, as long as that transaction's actual `gas_spent` stays low enough for the final receipts and `header.gas_used` to still line up.

## Concrete example
With `block_gas_limit = 100`:
- `tx0`: `gas_limit = 60`, `gas_spent = 60`
- `tx1`: `gas_limit = 50`, `gas_spent = 40`

Sequential execution rejects `tx1` because only `40` gas remains before it starts.
Parallel execution currently accepts the block because it only accumulates `gas_spent` after execution and never checks `tx1.gas_limit()` against the remaining block gas.

## Relevant code
- Sequential check: `crates/vm/backends/levm/mod.rs` in the sequential `execute_block` / `execute_block_pipeline` loops
- Missing check: `crates/vm/backends/levm/mod.rs` in `execute_block_parallel`, during the receipt-building phase after sorting results

## Suggested fix
Carry each transaction's `gas_limit()` into the parallel receipt finalization step and run the same `check_gas_limit` logic there before updating cumulative receipt gas.

I have a local patch and regression test for this and can open a PR if this direction looks good.

sueun-dev:
## Summary
- carry each transaction's `gas_limit()` into the BAL-backed parallel execution result set
- enforce the same remaining block gas check used by the sequential LEVM path when building parallel receipts
- add a regression test for a later transaction whose `gas_limit` exceeds the remaining block gas

Fixes #6341.

## Testing
- `cargo test -p ethrex-vm`


crates/vm/backends/levm/mod.rs
```diff
@@ -76,6 +76,32 @@ fn check_gas_limit(
...
+    mut results: Vec<(usize, TxType, u64, ExecutionReport)>,
+) -> Result<(Vec<Receipt>, u64), EvmError> {
+    results.sort_unstable_by_key(|(idx, _, _, _)| *idx);
+
+    let mut receipts = Vec::with_capacity(results.len());
+    let mut cumulative_gas_used = 0_u64;
+    let mut block_gas_used = 0_u64;
+
+    for (_, tx_type, tx_gas_limit, report) in results {
+        check_gas_limit(cumulative_gas_used, tx_gas_limit, block_gas_limit)?;
```
ElFantasma, line 91: Bug: this passes `cumulative_gas_used` (sum of `gas_spent`, post-refund) to `check_gas_limit`, but the sequential paths (lines 107 and 215) pass `block_gas_used` (sum of `gas_used`, pre-refund). Post-EIP-7778 these differ: `gas_spent <= gas_used`, so the parallel path becomes more lenient and could accept blocks that the sequential path rejects.

Should be:
```rust
check_gas_limit(block_gas_used, tx_gas_limit, block_gas_limit)?;
```

crates/vm/backends/levm/mod.rs
```diff
@@ -1950,4 +1959,44 @@ mod bal_tests {
         assert_eq!(u.info.as_ref().unwrap().code_hash, expected_hash);
         assert_eq!(u.code.as_ref().unwrap().bytecode, code);
     }
+
+    #[test]
+    fn test_build_parallel_receipts_rejects_tx_that_exceeds_remaining_block_gas() {
```
ElFantasma, line 2202: nit: the test uses `gas_used == gas_spent` for all entries, so it doesn't exercise the bug described above. Consider using different values (e.g., `gas_used: 70, gas_spent: 60`) to verify the check uses the pre-refund value.

crates/vm/backends/levm/mod.rs
sueun-dev: Thanks!

ElFantasma: Sorry, I got this wrong. The sequential paths (lines 122 and 285 on main) actually use `cumulative_gas_used` (post-refund), not `block_gas_used`. PR #6309 explicitly fixed this to use post-refund gas because pre-refund gas causes false "Gas allowance exceeded" rejections on tightly-packed Amsterdam blocks with large SSTORE refunds (see the post-mortem there).

Your original code using `cumulative_gas_used` was correct. Commit 6b7f80 switching to `block_gas_used` reintroduces the bug that #6309 fixed. Please revert to `cumulative_gas_used` for the parallel path to match the sequential paths.

crates/vm/backends/levm/mod.rs
ElFantasma: Disregard this too — since the check should use `cumulative_gas_used` (post-refund), the test values should match that. Though having `gas_used != gas_spent` in the test is still useful to verify the receipt uses `gas_spent` for `cumulative_gas_used` correctly.

ElFantasma (approved):
Approved

Sorry again for my mistake, and thanks for your contribution!

crates/vm/backends/levm/mod.rs
sueun-dev: Thanks so much! I have been following the project for a while, so I'm really happy I could contribute!!

ilitteri:
Hey @sueun-dev, the PR is good to go. At ethrex, we have a commit signature verification policy that states that all commits must be verified (read [here](https://github.com/lambdaclass/ethrex/blob/main/CONTRIBUTING.md#commit-signature-verification)).

Please, sign the commits of this PR for it to be able to be merged.

sueun-dev:
I pushed a verified signed commit and the PR now consists of a single signed commit (`6fcaf4c`). The remaining blocker appears to be maintainer approval for the fork workflows to run again.

PR #6344, opened 2026-03-09
perf(l1): replace per-block thread spawning with persistent thread pool for merkleization

### LINKED ISSUE #6282: Reuse 19 threads between block executions

azteca1998:
## Summary

Eliminates ~1.7ms overhead per block by replacing per-block spawning of 16 merkleization worker threads with a persistent rayon ThreadPool.

**Key changes:**
- Add `merkle_pool` field to `Blockchain` struct with 17-thread rayon pool
- Refactor `handle_merkleization` to use `pool.scope()` instead of `std::thread::scope`
- Use 17 threads to prevent deadlock: 16 for workers + 1 for coordination/overhead

**Performance impact:**
- Before: 16 workers spawned/joined per block (~100µs each × 2 = ~1.6ms overhead)
- After: Persistent pool reused across blocks (~0ms overhead for workers)

**Why 17 threads?**
The 17th thread is critical - with only 16 threads, all pool threads would be consumed by blocking workers, leaving no thread available for coordination inside `rayon::scope()`, causing deadlock. This was validated through systematic testing showing worker 15 never starting with 16 threads, but all workers starting successfully with 17 threads.

## Test plan

- [x] All 6,158 EF tests pass in 69.09 seconds
- [x] Verified thread pool eliminates per-block thread creation overhead
- [x] Validated 17-thread configuration prevents deadlock
- [x] Each Blockchain instance has its own pool for test isolation

Fixes #6282

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.90 AS chef
+FROM rust:1.91 AS chef
 
 RUN apt-get update && apt-get install -y \
     build-essential \


crates/blockchain/blockchain.rs
```diff
@@ -542,31 +571,26 @@ impl Blockchain {
...
-            workers_handles.push(handle);
-            workers_tx.push(tx);
-        }
+
+        self.merkle_pool.scope(|s| -> Result<_, StoreError> {
+            for i in 0..16 {
+                let (tx, worker_rx) = channel();
+
+                s.spawn(move |_| {
+                    let _ = self.handle_merkleization_subtrie(worker_rx, parent_header, i);
```
ElFantasma: Worker errors are silently discarded here. If `handle_merkleization_subtrie` returns a `StoreError`, the coordination code will continue with missing/incorrect storage roots and produce a wrong state root.

The old code had the same problem (`workers_handles` was populated but never joined for results), so this isn't a regression — but since you're refactoring this, it'd be a good opportunity to propagate errors. One approach: have each worker send its `Result` through a channel back to the coordinator, or store errors in a shared `Mutex<Option<StoreError>>` that the coordinator checks after `drop(workers_tx)`.

crates/blockchain/blockchain.rs
```diff
@@ -290,22 +302,40 @@ struct BalStateWorkItem {
...
             mempool: Mempool::new(blockchain_opts.max_mempool_size),
             is_synced: AtomicBool::new(false),
             payloads: Arc::new(TokioMutex::new(Vec::new())),
             options: blockchain_opts,
+            merkle_pool,
         }
     }
 
     pub fn default_with_store(store: Store) -> Self {
+        let merkle_pool = Arc::new(
```
ElFantasma: nit: This pool construction is copy-pasted from `new()`. Consider extracting a helper:

```rust
fn build_merkle_pool() -> Arc<rayon::ThreadPool> {
    Arc::new(
        rayon::ThreadPoolBuilder::new()
            .num_threads(17)
            .thread_name(|i| format!("merkle-worker-{i}"))
            .build()
            .expect("Failed to create merkle pool"),
    )
}
```

crates/blockchain/blockchain.rs
```diff
@@ -208,6 +208,9 @@ pub struct Blockchain {
     /// Maps payload IDs to either completed payloads or in-progress build tasks.
     /// Kept around in case consensus requests the same payload twice.
     pub payloads: Arc<TokioMutex<Vec<(u64, PayloadOrTask)>>>,
+    /// Persistent thread pool for merkleization workers.
+    /// 17 threads: 16 shard workers + 1 watcher/coordination.
+    merkle_pool: Arc<rayon::ThreadPool>,
```
iovoid: Is the arc needed?

azteca1998:
Fixed CI failures:

- ✅ Updated `rust-toolchain.toml` to 1.91.0 (required by alloy 1.8.3 dependencies)
- ✅ Updated `vergen-git2` to 9.1 (fixes vergen_lib version conflict)  
- ✅ Updated `pr_nostd.yaml` workflow to use Rust 1.91.0
- ✅ Merged latest main

All changes tested locally:
- ✅ `cargo test --workspace` passes
- ✅ `cargo build --features l2,l2-sql` succeeds
- ✅ no_std build for riscv64imac target succeeds

The L2 docker build and Test failures should resolve with these fixes.

PR #6364, opened 2026-03-13
ci(l2): add replay-witness-test job to L2 CI workflow

avilagaston9:
**Motivation**

We need CI coverage to verify that ethrex-replay works against the current ethrex code. This catches breaking changes to `ExecutionWitness`, the `debug_executionWitness` RPC endpoint, or any VM/blockchain APIs that ethrex-replay depends on. Follows up on #6356.

**Description**

Add a `replay-witness-test` job to the L2 CI workflow that:
1. Starts an L1 dev node using the existing docker image
2. Clones ethrex-replay at a **pinned commit** and patches its ethrex dependencies to the current PR's local checkout using Cargo `[patch]`
3. Sends a test transaction via `rex` and replays the block containing it with `--no-zkvm`
4. Asserts the replay succeeds (state root matches)

The `[patch]` approach means ethrex-replay's `Cargo.toml` points to a pinned ethrex rev, but CI overrides the dependencies with the local checkout. The replay commit is pinned to avoid unrelated replay regressions breaking ethrex CI.

**Breaking change procedure:**
When an ethrex PR introduces a breaking change to replay:
1. Create a companion PR in ethrex-replay that adapts to the new API
2. Update `REPLAY_REF` in the workflow to the companion branch/commit
3. After both merge, update `REPLAY_REF` to the new replay main commit

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` — N/A, CI-only change

### DIFF
--- a/.github/workflows/pr-main_l2.yaml
+++ b/.github/workflows/pr-main_l2.yaml
@@ -45,6 +45,8 @@ jobs:
               - "fixtures/contracts/**"
               - "crates/blockchain/dev/**"
               - "crates/vm/levm/**"
+              - "crates/common/types/block_execution_witness.rs"
+              - "crates/networking/rpc/debug/execution_witness.rs"
               - ".github/workflows/pr-main_l2.yaml"
               - "cmd/ethrex/l2/**"
             non_docs:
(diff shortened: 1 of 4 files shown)


avilagaston9:
Closed for inactivity.

Issue #5425, opened 2025-11-25
Decouple local interface address from externally announced address


MegaRedHand:
See #5354

We currently announce the same local IP address we listen on to our peers. However, this doesn't work when our node is behind a NAT. In those cases, we need to allow the user to set their externally reachable address (or automatically resolve our address by using an external service...).

Issue #5354, opened 2025-11-14
Add support for IPv6 addresses


MegaRedHand:
With #5297, we have a way for the user to set a specific address for the P2P layer. This works fine for IPv4; however, using an IPv6 address has the additional problem that most nodes use IPv4 addresses (specifically, bootnodes). We should find a way to translate between the two, or even have a dual stack of IPv4+IPv6.

MegaRedHand:
Did some research on this.

## IPv6 support for discv4

Implementing IPv6 support for discv4 is not worth it. discv4 requires both ends to be able to send messages to each other, which effectively makes IPv6 and IPv4 peers two different sets. This could be mitigated by keeping two discv4 servers, using IPv4 with IPv4 peers and upgrading to IPv6 with IPv6 peers, but this is made difficult by discv4 messages supporting a single local address (["Neighbors" packet has a single `ip` field](https://github.com/ethereum/devp2p/blob/master/discv4.md#neighbors-packet-0x04)). In the end, the node requires an IPv4 address anyway to connect to the bootnodes, since there aren't any IPv6 ones.

It'd be better to work on discv5 support (#4491), since that fixes the issue by moving to ENRs altogether: https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#nodes-response-0x04

Felix of geth on why IPv6 support for discv4 is a lost cause: https://github.com/ethereum/go-ethereum/issues/23263#issuecomment-984726573.

## IPv6 support for RLPx (listening address)

This _can_ be done, and could be useful for a user which doesn't need the discovery protocol, or which wants to use IPv4 for discv4 and IPv6 for RLPx. Geth supports this, but Nethermind doesn't seem to ([their `NodeRecordProvider` only sets the `ip` field](https://github.com/NethermindEth/nethermind/blob/e61dd7d48c77de62d3b944573f8ce2760f7926bd/src/Nethermind/Nethermind.Network.Discovery/NodeRecordProvider.cs#L27))
In order to support using a local IPv6 address for RLPx, we need to:

- [ ] Decouple the discv4 address from the RLPx address. We currently use a `Node` type to store our network configuration, but we should probably use an ENR for this: https://github.com/lambdaclass/ethrex/blob/fa3a298407bed400f3c8e63c0afd2eba9430dfc8/cmd/ethrex/initializers.rs#L319
[...]

PR #6375, opened 2026-03-17
feat(l1): decouple discv4 address from RLPx address

### LINKED ISSUE #5424: Support different addresses for discv4 and RLPx
See #5354

Some users might want to use an IPv6 socket for receiving inbound RLPx connections. For this, we need to support having different addresses for both protocols. We also need different flags to set both addresses independently.

azteca1998:
Closes #5424

## Summary

- Extends `NetworkConfig` with four address fields instead of two: `discovery_bind_addr`, `discovery_external_addr`, `rlpx_bind_addr`, `rlpx_external_addr`
- Adds `--discovery.addr` CLI flag (env: `ETHREX_P2P_DISCOVERY_ADDR`) to bind the UDP discovery socket independently of the RLPx TCP socket
- `discv4` and `discv5` servers now receive a `discovery_local_node` built from `discovery_external_addr`, so their Ping `from` endpoints advertise the correct address
- Implements `tcp6`/`udp6` ENR field parsing in `NodeRecordPairs` (removes the TODO)
- `NodeRecord::from_node()` now emits `ip6`/`tcp6`/`udp6` for IPv6 node addresses instead of incorrectly encoding them under the IPv4 `ip`/`tcp`/`udp` keys

## Address resolution logic

| `--discovery.addr` set? | Result |
|---|---|
| No | `discovery_bind_addr` = `rlpx_bind_addr`; `discovery_external_addr` = `rlpx_external_addr` |
| Yes, specific IP | `discovery_bind_addr` = given IP; `discovery_external_addr` = given IP |
| Yes, `0.0.0.0`/`::` | `discovery_bind_addr` = given wildcard; `discovery_external_addr` = `rlpx_external_addr` |

## Part of

This is PR 2 of 3 for IPv6 support (#5354):
- PR 1: decouple bind vs. external address (#5425) ← base branch
- **PR 2 (this):** decouple discv4 address from RLPx address (#5424)
- PR 3: ENR-based outbound connections + dual-stack listeners (#5354)

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -456,6 +456,10 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
     // Determine discovery bind address.
     // --discovery.addr sets the UDP bind addr independently of RLPx.
     // Defaults to rlpx_bind_addr so the two channels co-locate by default.
+    // NOTE: we currently enforce that discovery and RLPx use the same address
+    // family. Mixed-family (e.g. IPv4 discv4 + IPv6 RLPx) is structurally
+    // supported by NetworkConfig but not yet tested end-to-end. A future
+    // dual-stack PR (feat/p2p-dual-stack-ipv6) will relax this constraint.
     let discovery_bind_addr: IpAddr = opts
         .discovery_addr
         .as_deref()
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -446,16 +446,36 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
 
     let local_public_key = public_key_from_signing_key(signer);
 
-    let (bind_addr, external_addr) = resolve_p2p_endpoints(
+    let (rlpx_bind_addr, rlpx_external_addr) = resolve_p2p_endpoints(
         opts.p2p_addr.as_deref(),
         opts.nat_extip.as_deref(),
         local_ip().ok(),
         local_ipv6().ok(),
     );
 
-    let node = Node::new(external_addr, udp_port, tcp_port, local_public_key);
+    // Determine discovery bind address.
+    // --discovery.addr sets the UDP bind addr independently of RLPx.
+    // Defaults to rlpx_bind_addr so the two channels co-locate by default.
+    let discovery_bind_addr: IpAddr = opts
+        .discovery_addr
+        .as_deref()
+        .map(|a| a.parse().expect("Failed to parse --discovery.addr address"))
+        .unwrap_or(rlpx_bind_addr);
+
+    // Discovery external address: use the explicit discovery bind addr when it
+    // is a specific (non-wildcard) IP; otherwise fall back to rlpx_external_addr.
+    let discovery_external_addr = if !discovery_bind_addr.is_unspecified() {
+        discovery_bind_addr
+    } else {
+        rlpx_external_addr
+    };
+
+    let node = Node::new(rlpx_external_addr, udp_port, tcp_port, local_public_key);
     let network_config = NetworkConfig {
-        bind_addr,
+        discovery_bind_addr,
+        discovery_external_addr,
+        rlpx_bind_addr,
+        rlpx_external_addr,
         tcp_port,
         udp_port,
     };
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -459,7 +459,17 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
     let discovery_bind_addr: IpAddr = opts
         .discovery_addr
         .as_deref()
-        .map(|a| a.parse().expect("Failed to parse --discovery.addr address"))
+        .map(|a| {
+            let addr: IpAddr = a.parse().expect("Failed to parse --discovery.addr address");
+            if let Some(extip) = &opts.nat_extip {
+                let external: IpAddr = extip.parse().expect("Failed to parse --nat.extip address");
+                assert!(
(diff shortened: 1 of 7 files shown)


crates/networking/p2p/types.rs
```diff
@@ -310,13 +322,16 @@ pub struct NodeRecordPairs {
     // I think the confusion comes from the fact that geth decodes the bytes and then builds an IPV4/6 big-integer structure.
     pub tcp_port: Option<u16>,
     pub udp_port: Option<u16>,
+    /// TCP port for IPv6 RLPx connections (ENR key `tcp6`).
```
ElFantasma, line 325: These new fields are decoded/encoded correctly, but `NodeRecord::from_node` (line ~392) still unconditionally sets `tcp_port`/`udp_port` (the IPv4 ENR keys `tcp`/`udp`) even when the node IP is IPv6. The `match` below it dispatches `ip` vs `ip6` correctly, but ports don't follow suit.

The PR description says "NodeRecord::from_node() now emits ip6/tcp6/udp6 for IPv6 node addresses" but the code doesn't match. An IPv6-only ENR would have `ip6` + `tcp` + `udp` instead of `ip6` + `tcp6` + `udp6`, which other clients may not interpret correctly.

Suggested fix — move port assignment into the match:
```rust
match node.ip.to_canonical() {
    IpAddr::V4(ip) => {
        pairs.ip = Some(ip);
        pairs.tcp_port = Some(node.tcp_port);
        pairs.udp_port = Some(node.udp_port);
    }
    IpAddr::V6(ip) => {
        pairs.ip6 = Some(ip);
        pairs.tcp6_port = Some(node.tcp_port);
        pairs.udp6_port = Some(node.udp_port);
    }
}
```

ElFantasma (changes requested):

cmd/ethrex/initializers.rs
```diff
@@ -409,9 +412,39 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
...
+        })
+        .unwrap_or(rlpx_bind_addr);
+
+    // Discovery external address: use the explicit discovery bind addr when it
+    // is a specific (non-wildcard) IP; otherwise fall back to rlpx_external_addr.
+    let discovery_external_addr = if !discovery_bind_addr.is_unspecified() {
+        discovery_bind_addr
+    } else {
+        rlpx_external_addr
+    };
```
iovoid: What if nat.extip is set?

cmd/ethrex/initializers.rs
```diff
@@ -409,9 +412,39 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
...
+        .as_deref()
+        .map(|a| {
+            let addr: IpAddr = a.parse().expect("Failed to parse --discovery.addr address");
+            if let Some(extip) = &opts.nat_extip {
+                let external: IpAddr = extip.parse().expect("Failed to parse --nat.extip address");
+                assert!(
+                    addr.is_ipv4() == external.is_ipv4(),
+                    "--discovery.addr and --nat.extip must use the same address family (both IPv4 or both IPv6)"
+                );
+            }
```
iovoid: Why parse again?

cmd/ethrex/initializers.rs
```diff
@@ -409,9 +412,40 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
...
+    // Determine discovery bind address.
+    // --discovery.addr sets the UDP bind addr independently of RLPx.
+    // Defaults to rlpx_bind_addr so the two channels co-locate by default.
+    let discovery_bind_addr: IpAddr = opts
+        .discovery_addr
+        .as_deref()
+        .map(|a| {
+            let addr: IpAddr = a.parse().expect("Failed to parse --discovery.addr address");
+            assert!(
+                addr.is_ipv4() == rlpx_external_addr.is_ipv4(),
```
ElFantasma, line 424: The `NetworkConfig` doc (types.rs) advertises this flag as "enabling e.g. IPv4 discv4 with IPv6 RLPx" — a dual-stack scenario. But this assertion enforces `addr.is_ipv4() == rlpx_external_addr.is_ipv4()`, which blocks exactly that use case.

Example: `--p2p.addr [::] --discovery.addr 0.0.0.0` (IPv6 RLPx + IPv4 discovery) fails this assertion.

Either the doc is aspirational for a future PR — in which case the comment should be tightened to match the current single-family constraint — or the assertion should be relaxed to allow split families (the four addresses in `NetworkConfig` already support this structurally).

crates/networking/p2p/types.rs
```diff
@@ -430,13 +453,19 @@ impl NodeRecord {
...
-            IpAddr::V4(ip) => pairs.ip = Some(ip),
-            IpAddr::V6(ip) => pairs.ip6 = Some(ip),
+            IpAddr::V4(ip) => {
+                pairs.ip = Some(ip);
+                pairs.tcp_port = Some(node.tcp_port);
+                pairs.udp_port = Some(node.udp_port);
+            }
+            IpAddr::V6(ip) => {
+                pairs.ip6 = Some(ip);
+                pairs.tcp6_port = Some(node.tcp_port);
```
ElFantasma, line 466: The IPv4/IPv6 branch here is the key fix in this PR, but there's no test covering the IPv6 path. The existing `encode_ping_handshake_packet_with_enr` test only exercises IPv4 (updated here to add `tcp6_port: None, udp6_port: None`).

A round-trip test for an IPv6 node (construct a `Node` with IPv6 address → `NodeRecord::from_node` → `pairs()` → assert `ip6`, `tcp6_port`, `udp6_port` are set and `ip`, `tcp_port`, `udp_port` are not) would lock in this fix and catch regressions.

ElFantasma (approved):
Initial requests were addressed. 

Left a couple of comments for minor issues.

iovoid (approved):

PR #6386, opened 2026-03-19
fix(l1): fix BLOCKHASH cross-batch resolution during block import

edg-l:
Fixes #6385.

When importing blocks in batches, `BLOCKHASH` could fail to resolve hashes for blocks that landed in a prior batch -- those blocks aren't canonical yet during import, so they were absent from the cache and any lookup returned zero.

The fix walks the parent chain from the first block of the current batch, caching up to 256 prior block hashes into `block_hash_cache` before execution begins. The walk breaks early if storage doesn't have a header (e.g. genesis boundary), so there's no impact on single-batch or shallow imports.

Regression test in `test/tests/blockchain/batch_tests.rs`: deploys a contract that calls `BLOCKHASH(number - 1)` and invokes it from a second batch, confirming the hash is resolved correctly.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -2317,19 +2317,23 @@ impl Blockchain {
         // Walk the parent chain to cache the last 256 block hashes so that
         // BLOCKHASH can resolve references to blocks from previous batches
         // (they may not be canonical yet during import).
-        {
-            let mut hash = parent_header.hash();
-            let mut number = parent_header.number;
+        block_hash_cache
+            .entry(parent_header.number)
+            .or_insert_with(|| parent_header.hash());
+        let mut hash = parent_header.parent_hash;
+        let mut number = parent_header.number.saturating_sub(1);
+        let lookback = first_block_header.number.saturating_sub(256);
+        while number > lookback {
             block_hash_cache.entry(number).or_insert(hash);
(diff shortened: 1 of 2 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -2141,14 +2141,36 @@ impl Blockchain {
...
             .storage
             .get_block_header_by_hash(first_block_header.parent_hash)
             .map_err(|e| (ChainError::StoreError(e), None))?
             .ok_or((ChainError::ParentNotFound, None))?;
+
+        // Walk the parent chain to cache the last 256 block hashes so that
+        // BLOCKHASH can resolve references to blocks from previous batches
+        // (they may not be canonical yet during import).
+        {
+            let mut hash = parent_header.hash();
```
ElFantasma: nit: the first loop iteration fetches `get_block_header_by_hash(parent_header.hash())` which returns the same `parent_header` we already have. You could start the walk from `parent_header.parent_hash` directly:
```rust
let mut hash = parent_header.parent_hash;
let mut number = parent_header.number.saturating_sub(1);
```
and insert `parent_header` outside the loop (which you already do on line 2155). Saves one DB lookup per batch.

ElFantasma (approved):
LGTM, just pointed out the same nit that greptile found.

azteca1998 (approved):

iovoid (approved):

PR #6391, opened 2026-03-20
feat(l2): add --watcher.start-l1-block to skip old L1 blocks on restart

avilagaston9:
**Motivation**

When restarting a node that has been running for several months, the L1 watcher starts scanning from the block where the bridge contract was deployed. This means it re-scans all historical L1 blocks, which can take 3+ hours before the node catches up.

**Description**

Add a `--watcher.start-l1-block` CLI option (also configurable via `ETHREX_WATCHER_START_L1_BLOCK` env var) that lets operators specify an L1 block number to start scanning from, skipping older blocks. When set, the watcher uses this value instead of querying the bridge contract's `lastFetchedL1Block()`.

This is useful for node operators restarting nodes that have already processed all historical deposits — they can set the start block to a recent L1 block and avoid the multi-hour catch-up.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` — N/A, no storage changes


avilagaston9:
Closed for inactivity.

PR #6406, opened 2026-03-26
perf(l1): reuse BAL for witness generation

ArshLabs:
closes #6388

replaces DatabaseLogger wrapping in generate_witness_for_blocks_with_fee_configs with the BAL that execute_block already produces for Amsterdam+ blocks. the BAL contains exactly the same info that DatabaseLogger captures (touched accounts, storage slots, code changes), so we use it directly instead of intercepting every DB call through mutex-guarded fields.

changes:
- creates a plain VM instead of wrapping with DatabaseLogger
- uses bal.accounts() and all_storage_slots() to build state access map
- collects bytecodes from account states in storage (deduplicated)
- conservatively includes up to 256 ancestor block headers since BAL does not track BLOCKHASH opcode references
- returns error for pre-Amsterdam blocks (BAL is None), since this path targets post-Amsterdam


crates/blockchain/blockchain.rs
```diff
@@ -1309,25 +1309,31 @@ impl Blockchain {
...
-                    .storage
-                    .get_account_code(*code_hash)
-                    .map_err(|_e| {
-                        ChainError::WitnessGeneration("Failed to get account code".to_string())
-                    })?
-                    .ok_or(ChainError::WitnessGeneration(
-                        "Failed to get account code".to_string(),
-                    ))?;
-                codes.push(code.bytecode.to_vec());
+                let mut seen_code_hashes = HashSet::new();
```
ElFantasma, line 1330: This uses `parent_state_root` (block N-1's state) to look up account code hashes. Contracts deployed during block N's execution won't exist in the parent state — `get_account_state_by_root` returns `None` and their bytecode is silently missed from the witness.

The `account_updates` from `vm.get_state_transitions()` should have the `code_hash` for newly created accounts — you could collect those as a second pass.

ElFantasma (changes requested):

crates/blockchain/blockchain.rs
ArshLabs: > This uses `parent_state_root` (block N-1's state) to look up account code hashes. Contracts deployed during block N's execution won't exist in the parent state — `get_account_state_by_root` returns `None` and their bytecode is silently missed from the witness.
> 
> The `account_updates` from `vm.get_state_transitions()` should have the `code_hash` for newly created accounts — you could collect those as a second pass.

pl check now and lemme know if anything is required on my end

ElFantasma: Confirmed, thanks — the two-pass approach in :1338-1349 covers exactly the new-deployment case I flagged

ElFantasma (approved):

ElFantasma:
Thanks for your contribution!
In order to merge this, all commits must have verified signatures. (besides the required approvals)
See [here](https://github.com/lambdaclass/ethrex/blob/main/CONTRIBUTING.md#commit-signature-verification)

PR #6408, opened 2026-03-26
feat(l1): track unrecognized discovery packets

ArshLabs:
Closes #6400

Adds a metric counter for discovery packets that don't match discv4 or discv5. The multiplexer’s route_to_discv5 now returns whether decoding succeeded. When a packet fails both the discv4 hash check and discv5 decode, it gets counted via a new `unrecognized_discovery_packets` counter in both the internal Metrics struct and the Prometheus-exported MetricsP2P (exposed as `ethrex_p2p_unrecognized_discovery_packets`).

Note: this will likely need a rebase after #6398 merges, since both touch `route_packet`. The changes are complementary — #6398 adds positive filtering logic, this PR adds the tracking.

**Changes:**
- crates/networking/p2p/discovery/multiplexer.rs: `route_to_discv5` returns `bool`, `route_packet` detects and counts unrecognized packets
- crates/networking/p2p/metrics.rs: new `unrecognized_discovery_packets` IntCounter + `record_unrecognized_discovery_packet()` method
- crates/blockchain/metrics/p2p.rs: new `unrecognized_discovery_packets` IntCounter in MetricsP2P + Prometheus registration

### DIFF
--- a/crates/blockchain/metrics/p2p.rs
+++ b/crates/blockchain/metrics/p2p.rs
@@ -1,4 +1,6 @@
-use prometheus::{Encoder, IntCounterVec, IntGauge, IntGaugeVec, Opts, Registry, TextEncoder};
+use prometheus::{
+    Encoder, IntCounter, IntCounterVec, IntGauge, IntGaugeVec, Opts, Registry, TextEncoder,
+};
 use std::sync::LazyLock;
 
 use crate::MetricsError;
@@ -10,6 +12,7 @@ pub struct MetricsP2P {
     peer_count: IntGauge,
     peer_clients: IntGaugeVec,
     disconnections: IntCounterVec,
+    unrecognized_discovery_packets: IntCounter,
 }
 
 impl Default for MetricsP2P {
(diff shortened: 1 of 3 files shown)


ElFantasma (approved):
This will conflict with #6398 (also open, also touches `route_packet`). Both are complementary — #6398 adds `is_discv5_packet()` positive filtering, this adds the counter. Whichever merges second will need a trivial rebase. If #6398 goes first, the counter just moves into the `else` branch that #6398 introduces.

ArshLabs:
> Contributor



> This will conflict with #6398 (also open, also touches `route_packet`). Both are complementary — #6398 adds `is_discv5_packet()` positive filtering, this adds the counter. Whichever merges second will need a trivial rebase. If #6398 goes first, the counter just moves into the `else` branch that #6398 introduces.

yeah exactly and since this one is strictly additive (just a counter, no decode path changes), it's probably the lower type risk merge. happy to rebase on top of #6398 if it goes first though, should slot right into the else branch!

PR #6409, opened 2026-03-26
fix(l1): persist fullsync header progress across sync cycles

edg-l:
## Summary
- Full sync was losing all header download progress when a cycle aborted after peer failures. Each new cycle restarted from the chain head, re-downloading tens of thousands of headers before hitting the same failure point indefinitely.
- Now persists the lowest block reached and resume hash in `MISC_VALUES`, so the next cycle bridges the gap from the new sync head to stored headers and continues from the previous abort point.
- Fixes retry off-by-one (`>` to `>=`) in both full and snap sync so attempt count matches `MAX_HEADER_FETCH_ATTEMPTS`.
- Resets attempt counter after a successful fetch so it tracks consecutive failures, not cumulative.
- Replaces fixed 5s retry delay with gentle backoff (1s, 2s, 3s, 5s, 10s).

## Test plan
- [ ] Run full sync against mainnet/testnet and verify logs show "Connected to previously stored headers" on cycle 2+ after an abort
- [ ] Verify retry logs show correct "attempt N/5" counts and increasing delays
- [ ] Verify successful completion clears progress via `clear_fullsync_headers`

### DIFF
--- a/crates/networking/p2p/sync/full.rs
+++ b/crates/networking/p2p/sync/full.rs
@@ -218,7 +218,7 @@ pub async fn sync_cycle_full(
     // Request all block headers between the sync head and our local chain
     // We will begin from the sync head so that we download the latest state first, ensuring we follow the correct chain
     // This step is not parallelized
-    let mut start_block_number;
+    let mut start_block_number = 0;
     let mut end_block_number = 0;
     let mut headers = vec![];
     let mut single_batch = true;
@@ -230,6 +230,14 @@ pub async fn sync_cycle_full(
     // a synced node. Set on the first batch of headers we fetch.
     let mut started_behind = false;
     let mut sync_target_logged = false;
+
+    // Check for progress saved from a previous aborted cycle
+    let mut prior_progress = store.get_fullsync_progress()?;
+    if let Some((prev_lowest, _)) = &prior_progress {
+        info!(
+            "Found prior fullsync progress: headers stored down to block {prev_lowest}, will resume from there"
+        );
+    }
 
     // Request and store all block headers from the advertised sync head
     loop {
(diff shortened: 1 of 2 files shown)


ElFantasma (approved):

iovoid (changes requested):
Wouldn't a long-range attack leave the node permanently stuck?

edg-l:
@iovoid good catch, you're right. The number-only reconnection check made this exploitable: `FULLSYNC_HEADERS` is keyed by block number, so a stored header from a different fork (a reorg between cycles, or a long-range attack) sitting at the same number would falsely satisfy `connected`. The node would then jump to the stale/forged `prev_resume_hash`, fail to fetch it from honest peers, exhaust the retries, re-persist the same bad anchor on abort, and stay wedged every cycle.

Rebased onto main and fixed it:

- **Hash-verified reconnection** (`full.rs`): we only treat stored headers as a continuation when the hash chain is actually continuous. `sync_head` at that point is the parent hash our lowest header points to, so for `start_block_number == prev_lowest` we require `sync_head == prev_resume_hash`, and for `start_block_number > prev_lowest` we require the stored header at `start_block_number - 1` to hash-match `sync_head`.
- **Invalidation on mismatch**: if the stored progress doesn't link to the current consensus head, we `clear_fullsync_headers()` (drops the orphaned headers + anchor) and keep descending from the honest head, so a reorg/attack can't permanently wedge sync.

Also dropped the retry-logic changes from the original PR; main already landed the `attempts` reset and retry-delay tweaks independently, so the diff is now just the progress-persistence feature plus this hardening. PTAL.

crates/networking/p2p/sync/full.rs
```diff
@@ -406,6 +422,60 @@ pub async fn sync_cycle_full(
...
+                continue;
+            } else {
+                // Stored progress belongs to a different chain than the current
+                // consensus-provided sync head. Discard the orphaned headers and
+                // anchor so we keep descending from the honest head instead of
+                // chasing a stale/forged resume point on every cycle.
+                warn!(
+                    "Stored fullsync progress at block {prev_lowest} does not link to the current chain; discarding it and resyncing from the consensus head"
+                );
+                store.clear_fullsync_headers().await?;
```
iovoid, line 475: This deletes every header, not only the orphaned ones.

iovoid (changes requested):

crates/networking/p2p/sync/full.rs
```diff
@@ -218,7 +218,7 @@ pub async fn sync_cycle_full(
     // Request all block headers between the sync head and our local chain
     // We will begin from the sync head so that we download the latest state first, ensuring we follow the correct chain
     // This step is not parallelized
-    let mut start_block_number;
+    let mut start_block_number = 0;
```
iovoid, line 221: Never updated to not be zero.

iovoid (changes requested):

PR #6410, opened 2026-03-26
perf(l1): optimize trie building in snap sync insertion

ElFantasma:
## Motivation

Profiling snap sync's insertion phases on mainnet revealed that **trie building is 75-91% of insertion time**, and it's entirely CPU-bound (0% I/O wait). Account insertion took ~20 minutes and storage insertion ~43 minutes — together 61% of total snap sync time.

## Description

Three optimizations targeting the CPU-bound trie construction:

### 1. Eliminate redundant code hash iteration pass
The original code iterated all accounts **twice** — once to collect code hashes (async flush), once to build the trie. Merged both into a single pass by collecting code hashes in the trie-building iterator and flushing them incrementally to disk.

### 2. Reuse `nodehash_buffer` across calls
`add_current_to_parent_and_write_queue` allocated a new 512-byte `Vec` on every call (~700M calls on mainnet). Now receives the buffer as a parameter from the outer function. Also avoids a `Nibbles` path clone per call by computing the prefix index directly.

### 3. Parallel state trie building across 16 nibble ranges
Split the state trie build into 16 independent sub-tries (one per first nibble of account hash), built concurrently using streaming RocksDB iterators and a shared thread pool for DB flushes. Sub-trie results are assembled into the root BranchNode via `assemble_root_from_subtries`. Each sub-trie thread also collects its own code hashes (flushed incrementally to disk) and storage accounts, merged after completion.

### Code structure

- `trie_from_sorted_subtrie` — builds a single sub-trie, returns the child `NodeRef`
- `assemble_root_from_subtries` — assembles 16 sub-trie results into a root node (shared between `trie_from_sorted_parallel` and `insert_accounts`)
- `trie_from_sorted_parallel` — clean API used by tests; production uses the lower-level functions directly to collect side effects via `NibbleIterator`
- `NibbleIterator` — RocksDB iterator filtered to a single nibble range, collects code hashes + storage accounts as side effects

### Account Insertion

| Sub-phase | Before | After | Delta |
|-----------|--------|-------|-------|
| SST ingest | 104s | 112s | — |
| Code hash pass | **196s** | **0s** | **eliminated** |
| Trie building | **883s** | **700s** | **-21%** |
| **Total** | **1184s (19m 44s)** | **818s (13m 40s)** | **-31%** |

### Storage Insertion

| Sub-phase | Before | After | Delta |
|-----------|--------|-------|-------|
| Trie builds | 2357s | 2208s | -6% |
| **Total** | **2587s (43m 7s)** | **2433s (40m 30s)** | **-6%** |

## Multisync validation (ethrex-mainnet-10, release-with-debug-assertions)

Re-validated against current main on `ethrex-mainnet-10` (May 2026): **4 main runs vs 4 PR runs**, each running the full snap sync cycle on hoodi + sepolia + mainnet with `release-with-debug-assertions` (full state/storage/bytecode trie validation post-sync). **All 8 runs passed validation on all 3 networks.**

### Account Insertion (the optimization's main target)

| Network | Main avg (4 runs) | Range | PR avg (4 runs) | Range | Δ |
|---------|-------------------|-------|-----------------|-------|---|
| **Hoodi** | 1m 22s | 1:20–1:30 | **0m 57s** | 0:50–1:00 | **−30%** ✅ |
| **Sepolia** | 6m 52s | 5:30–8:20 | **4m 02s** | 3:30–4:30 | **−41%** ✅ |
| **Mainnet** | 17m 33s | 13:40–21:20 | **16m 48s** | 16:20–17:40 | **−4%** (within main's variance) |

On mainnet, single-run main numbers vary widely (13:40, 14:10, 21:00, 21:20). PR runs are much tighter (16:20–17:40, ±50s). The parallelization both improves the mean *and* stabilizes runtime.

### Storage Insertion

| Network | Main avg | PR avg | Δ |
|---------|----------|--------|---|
| Hoodi | 7m 20s | 10m 12s | +39% (~3min absolute) |
| Sepolia | 25m 55s | 29m 10s | +13% (~3min absolute) |
| Mainnet | 38m 43s | 40m 30s | +5% (~2min absolute) |

Small consistent regression on storage insertion (~2–3 min absolute on all networks). This phase wasn't the optimization's primary target. Worth investigating as follow-up but not large enough in absolute terms to block.

### Total wall (excluding outlier)

| Network | Main avg | PR avg | Δ |
|---------|----------|--------|---|
| Hoodi | 34m 45s | 41m 01s | +18% |
| Sepolia | 2h 21m 42s | 2h 23m 42s | +1% |
| Mainnet | 3h 10m 22s | 3h 13m 14s | +1.5% |

One main run hit a 2h 56m block-headers stall (peer/pivot issue, unrelated to this PR — see issue #6474), excluded from totals. Including all runs, mainnet main avg is 3h 46m vs PR's 3h 13m — but that's not a fair comparison; #6474 is a separate reliability issue.

### Storage Healing unchanged

| Network | Main avg | PR avg | Δ |
|---------|----------|--------|---|
| Hoodi | 11m 33s | 11m 53s | +3% |
| Sepolia | 44m 46s | 50m 14s | +12% |
| Mainnet | 1h 10m 40s | 1h 13m 07s | +3% |

Storage healing is dominated by `release-with-debug-assertions`'s trie validation, not the optimization itself.

### Run summary

| # | Branch | Hoodi | Sepolia | Mainnet | Total |
|---|--------|-------|---------|---------|-------|
| 1 | main | 29m 04s | 1h 59m 01s | 2h 59m 57s | 3h 29m 45s |
| 2 | perf | 39m 44s | 2h 09m 19s | 3h 13m 26s | 3h 44m 01s |
| 3 | perf | 45m 00s | 2h 13m 33s | 3h 18m 50s | 3h 49m 42s |
| 4 | perf | 40m 34s | 3h 03m 51s | 3h 04m 41s | 3h 36m 37s |
| 5 | perf | 38m 46s | 2h 08m 03s | 3h 16m 00s | 3h 48m 21s |
| 6 | main | 35m 37s | 2h 15m 13s | 3h 09m 29s | 3h 40m 16s |
| 7 | main | 30m 39s | 3h 03m 40s | 5h 33m 18s ⚠️ | 6h 06m 35s |
| 8 | main | 44m 10s | 2h 08m 53s | 3h 21m 41s | 3h 54m 20s |

⚠️ Run #7 mainnet hit a 2h 56m block-headers anomaly (likely #6474). Excluded from main avg.

All runs include full trie validation. **Validation passed on every run, both branches.**

## Checklist

- [x] Tested on mainnet snap sync (3 networks × 4 runs — all successful with validation)
- [x] Unit tests for parallel trie correctness (5 tests comparing parallel vs sequential hashes)
- [x] Profiling instrumentation removed
- [x] Root assembly logic deduplicated into `assemble_root_from_subtries`
- [x] Thread join panics handled gracefully (`ThreadJoinError` instead of `unwrap`)


crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -1046,39 +997,234 @@ async fn insert_accounts(
...
+            .compute_hash_no_alloc(&mut nodehash_buffer)
+            .finalize();
+        trie_db
+            .put_batch(vec![(Nibbles::default(), root_node.encode_to_vec())])
+            .map_err(SyncError::Trie)?;
+        hash
+    };
+
+    // Merge per-thread side effects
+    let collected_code_hashes: Vec<H256> = per_thread_code_hashes
```
fedacking: Memory risk: this may grow beyond our available memory

crates/common/trie/trie_sorted.rs
```diff
@@ -364,6 +498,126 @@ where
...
+/// `make_sub_iter` is called once per nibble (0..16) to produce a sorted iterator
+/// yielding only the accounts whose hash starts with that nibble.
+pub fn trie_from_sorted_parallel<F>(
+    db: &dyn TrieDB,
+    make_sub_iter: F,
+) -> Result<H256, TrieGenerationError>
+where
+    F: Fn(u8) -> Box<dyn Iterator<Item = (H256, Vec<u8>)> + Send> + Sync,
+{
+    let thread_count = std::thread::available_parallelism()
```
fedacking: I think this may break if the number is less than 16.

fedacking (dismissed):

crates/networking/p2p/sync/snap_sync.rs
ElFantasma: Fixed in 8dfa7f2 — code hashes are now flushed incrementally to disk from each thread via `NibbleIterator::flush_code_hashes()` (triggered every `CODE_HASH_WRITE_BUFFER_SIZE` entries). Only storage accounts remain in memory — those are just `(H256, H256)` pairs, much smaller.

crates/common/trie/trie_sorted.rs
ElFantasma: It works correctly with fewer than 16 entries. `trie_from_sorted_subtrie` returns `Ok(None)` for empty iterators, so nibble ranges with no accounts produce no sub-trie. The `assemble_root_from_subtries` function (extracted in the latest commit) handles 0, 1, or N valid sub-tries explicitly — the `valid_count == 1` branch collapses the single sub-trie into a proper root, and `valid_count == 0` returns the empty trie hash. The existing tests cover this — e.g. `test_parallel_1` has only 3 accounts touching 2 nibble ranges.

ElFantasma:
## Review feedback addressed

Thanks for the thorough reviews. Here's what was addressed and what wasn't (with reasoning).

### Addressed

| Finding | Raised by | Fix |
|---------|-----------|-----|
| Silent `dump_to_file` error in `NibbleIterator` | Greptile, Codex, Claude | Added `tracing::error!` logging instead of `let _` |
| `recv().expect()` panics on channel close in `trie_from_sorted_subtrie` | Greptile, Kimi | Converted to `.map_err(\|_\| TrieGenerationError::ThreadJoinError())?` |
| Final `flush_nodes_to_write` error silently discarded | Kimi, Claude | Changed `let _ =` to `?` to propagate |
| `.expect("Sub-trie wrote this node")` in `assemble_root_from_subtries` | Claude | Changed to `.ok_or(TrieGenerationError::FlushToDbError(TrieError::InvalidInput))?` |
| No `Drop` impl on `NibbleIterator` — code hashes lost on error paths | Claude | Added `Drop` impl that flushes code hashes and collects storage accounts |
| `handle.join().unwrap()` panics on thread failure | Fedacking (original review) | Converted to `.map_err(\|_\| TrieGenerationError::ThreadJoinError())?` |
| Mutex `unwrap()` rejected by clippy | CI | Changed to `.expect("storage accounts mutex poisoned")` |
| Parallel spawn duplication between `trie_from_sorted_parallel` and `insert_accounts` | Claude, Codex | Refactored `insert_accounts` to call `trie_from_sorted_parallel` directly. `NibbleIterator` now handles all side effects (code hash flushing + storage account collection) via `Drop`, so the generic API works for production. Tests now exercise the same code path as production. (-57 lines) |

### Not addressing (with reasoning)

1. **Loop duplication between `trie_from_sorted_accounts` and `trie_from_sorted_subtrie`** (Claude) — The inner trie-building loop is identical, but extracting a shared helper would require refactoring the pre-existing `trie_from_sorted_accounts` function, which is out of scope for this PR. The subtrie variant differs in finalization (returns `NodeRef` vs root hash).

2. **Code hash dedup regression** (Codex) — The old `CodeHashCollector` uses a `HashSet`; our `NibbleIterator` uses a `Vec`. Adding a per-thread `HashSet` would increase memory, which is what incremental flushing was designed to avoid. Dedup still happens at read time in `download_bytecodes` via `seen_code_hashes: HashSet`, so correctness is unaffected — we just write more files to disk.

3. **Thread panic payload lost in `ThreadJoinError`** (Claude) — The `Box<dyn Any>` payload is hard to extract usefully and the panic backtrace is already printed to stderr by the runtime.

4. **`Ordering::Relaxed` on `code_hash_file_index`** (Kimi) — Correct as-is. `fetch_add` with `Relaxed` guarantees atomic unique indices. No other data depends on ordering.

5. **`AccountState::decode(...).expect(...)` in `NibbleIterator`** (Kimi) — Pre-existing pattern. Data comes from our own temp RocksDB (not untrusted peers). A decode failure here would indicate internal corruption.

6. **`next_file_index` naming** (Claude) — Returns the next index to be used, which is the current counter value. Name is accurate from the caller's perspective.

azteca1998 (approved):

PR #6418, opened 2026-03-28
feat(l2): add EIP-8079 native rollups

avilagaston9:
**Motivation**

EIP-8079 (native rollups) and EIP-8025 (optional execution proofs) share the
same infrastructure: stateless block validation, execution witnesses, SSZ
encoding. EIP-8025 landed in main via #6427; this PR adds EIP-8079 on top. The
native-rollup code is always compiled and gated at runtime on `Fork::LStar` +
`VMType::L1` (no cargo feature).

**Description**

Reuses `verify_stateless_new_payload` for the `EXECUTE` precompile.

- `NativeRollup.sol` / `L2Bridge`: SSZ-based `advance()`; EIP-4788
  `BEACON_ROOTS` carries the L1 messages root; escrow-solvency accounting,
  failed-message recovery (`claimRecoverable`), configurable `FINALITY_DELAY`,
  and advancer/relayer rotation setters.
- L2 actors: `NativeBlockProducer`, `NativeL1Watcher`, `NativeL1Advancer`. The
  producer routes txs through `apply_tx_to_payload` so its EIP-7928 BAL matches
  L1 re-execution; the advancer regenerates the witness on demand and surfaces a
  reverted `advance()` via receipt status.
- `EXECUTE` precompile (`crates/vm/levm/src/execute_precompile.rs`): decodes SSZ
  input, validates L2 constraints, charges gas up front, delegates to a
  `StatelessValidator` trait. Invalid input and failed validation fail closed at
  the CALL level (tx included and charged).

**How to Test**

End-to-end demo (deposit, withdrawal, L1→L2 message):
[`docs/l2/deployment/native_rollups.md`](docs/l2/deployment/native_rollups.md).

```shell
# three terminals
make -C crates/l2 rm-db-l1 init-l1-native
make -C crates/l2 deploy-l1-native
make -C crates/l2 rm-db-l2 init-l2-native
cargo test -p ethrex-test --features l2 -- native_rollup_bridge_roundtrip --nocapture
```


crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,317 @@
...
+    // ===== Block Advancement =====
+
+    /// @notice Advance the L2 state by one block.
+    /// @param _l1MessagesCount Number of L1 messages consumed in this block.
+    /// @param _sszStatelessInput SSZ-encoded `StatelessInput` for the block.
+    /// @dev    Block fields are read from `_sszStatelessInput`; once the
+    ///         precompile reports `successful_validation == true` those bytes
+    ///         describe a correctly executed block, so the caller cannot
+    ///         substitute the new block_hash, state_root, etc.
+    function advance(
```
ElFantasma, line 260: **P1 SECURITY**: `_l1MessagesCount` is caller-controlled and unverified. The precompile validates the SSZ stateless input but never returns the count of L1 messages the L2 block actually consumed (`SszStatelessValidationResult` carries only `request_root`, `successful`, `chain_id`). So a malicious or buggy advancer can claim 5 L1 messages were consumed when the L2 block consumed 0, and this contract advances `l1MessageIndex` by 5, silently dropping queued L1→L2 messages.

**Fix shape**: the L2 block's L1-messages consumption needs to be cryptographically committed inside the SSZ payload (e.g., a count field bound to BEACON_ROOTS-derived root, or a dedicated SSZ `consumed_l1_message_count` field that the precompile validates against the block's actual deposits). Then the contract reads that count out of the SSZ payload (the same way it reads `provenBlockNumber`), not as a separate parameter. Without this, `onlyAdvancer` is the entire defense, and a compromised advancer can DoS L1→L2 messaging.

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,317 @@
...
+    uint256 constant RESULT_LEN = 41;
+    uint256 constant RESULT_SUCCESS_OFFSET = 32;
+    uint256 constant RESULT_CHAIN_ID_OFFSET = 33;
+
+    // Byte offsets inside the SSZ `ExecutionPayload` fixed prefix.
+    uint256 constant EP_PARENT_HASH_OFFSET = 0;
+    uint256 constant EP_STATE_ROOT_OFFSET = 52;
+    uint256 constant EP_BLOCK_NUMBER_OFFSET = 404;
+    uint256 constant EP_GAS_LIMIT_OFFSET = 412;
+    uint256 constant EP_BLOCK_HASH_OFFSET = 472;
```
ElFantasma, line 111: **Hardcoded SSZ field offsets** (`EP_PARENT_HASH_OFFSET = 0`, `EP_STATE_ROOT_OFFSET = 52`, `EP_BLOCK_NUMBER_OFFSET = 404`, ..., `EP_FIXED_PREFIX_LEN = 528`) are computed against the current `ExecutionPayload` SSZ schema. If the schema ever grows a new fixed-position field (or an existing variable offset shifts), the contract reads garbage and the `provenBlockHash`/`provenStateRoot` you commit to storage become attacker-controlled bytes. Since the contract is L1-deployed and upgrade-constrained, this is a significant brittleness. Mitigations:
- A schema version byte at the start of `_sszStatelessInput`, validated on entry.
- A unit test that compares these offsets against `SszEncode` output of a sample payload, gated on merge.

crates/vm/levm/src/precompiles.rs
```diff
@@ -255,6 +267,10 @@ pub fn precompiles_for_fork(fork: Fork) -> impl Iterator<Item = Precompile> {
 }
 
 pub fn is_precompile(address: &Address, fork: Fork, vm_type: VMType) -> bool {
```
ElFantasma, line 275: **`is_precompile()` for `VMType::L2`**: this unconditionally treats `0x0101` as a precompile when `experimental-devnet` is enabled, regardless of `vm_type`. For an L2 execution built with this feature, calls to `0x0101` route to the EXECUTE dispatcher and likely error since the validator isn't wired in the L2 path. Either gate the dispatch on `vm_type == VMType::L1` (the EXECUTE precompile is L1-only by design — it's the L1 contract calling L2 validation) or document why L2 mode also accepts the address. (Copilot flagged.)

crates/vm/levm/src/execute_precompile.rs
```diff
@@ -0,0 +1,107 @@
...
+    let validator = stateless_validator.ok_or_else(|| {
+        VMError::Internal(InternalError::Custom(
+            "EXECUTE precompile requires a StatelessValidator but none was provided".to_string(),
+        ))
+    })?;
+    run_execute(validator, calldata, gas_remaining)
+}
+
+/// Validate L2 constraints, charge gas, delegate.
+fn run_execute(
```
ElFantasma, line 57: **Double SSZ decode**: this decodes calldata into `SszStatelessInput` for L2 constraint validation, then `StatelessExecutor::verify` decodes the same bytes again. Two full SSZ decodes per precompile invocation. Pass the already-decoded `SszStatelessInput` through (or have `verify` accept `&SszStatelessInput` and re-serialize the result instead). (Greptile P2.)

tooling/l2/dev/blockscout_verify_native_rollup.py
```diff
@@ -0,0 +1,172 @@
...
+import http.client
+import json
+import os
+import sys
+import time
+import uuid
+
+COMPILER_VERSION = "v0.8.31+commit.fd3a2265"
+CONTRACT_NAME = "NativeRollup"
+CONTRACTS_DIR = os.path.join(
```
ElFantasma, line 28: Path is wrong (Copilot caught) — file lives under `crates/l2/contracts/src/nativeRollup/`, not `tooling/l2/crates/vm/levm/contracts/`.

tooling/l2/dev/blockscout_verify_native_rollup.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+    body += (
+        f"--{boundary}\r\n"
+        f'Content-Disposition: form-data; name="files[0]"; filename="input.json"\r\n'
+        f"Content-Type: application/json\r\n\r\n"
+        f"{standard_input}\r\n"
+    )
+    body += f"--{boundary}--\r\n"
+
+    conn = http.client.HTTPConnection(host, port)
```
ElFantasma: Uses `http.client.HTTPConnection` even for `https://` URLs (Copilot caught). Strip-scheme-then-pick-protocol pattern: parse with `urllib.parse.urlparse` and switch to `HTTPSConnection` when `scheme == "https"`. Will break verification against any TLS Blockscout instance.

ElFantasma (commented):
Substantial PR adding EIP-8079 native rollups. Reviewed with focus on security (this is the consensus boundary) and the StatelessValidator trait you flagged.

**Note on greptile/copilot P1 dismissals**: both reviewers flagged `advance()` as accepting caller-supplied `_newBlockHash` / `_newStateRoot`. Re-reading the current code, those values come from `_decodeProvenPayloadFields(_sszStatelessInput)` — i.e., from in-Solidity SSZ-decoded fields of the very payload the EXECUTE precompile validated, not from caller params. And `provenParentHash == blockHash` is checked. So the *block-hash and state-root binding* concerns are not present in this code; both bots seem to be reasoning about an earlier shape. Worth dismissing with a note in the response so the trail is clear. The real P1 is the `_l1MessagesCount` parameter — see inline.

**StatelessValidator trait (your feedback ask)**: the current shape (trait in `ethrex-vm`, impl in `ethrex-blockchain`, `Arc<dyn>` injected at `VM::new`) is the right pattern given the dep-graph constraint. The cycle `levm → blockchain` is real (blockchain → vm → levm); dep-inversion via a trait is the standard fix. Three small refinements:
1. Make the trait method generic over input type so the precompile can pass the already-decoded `SszStatelessInput` (avoids the double-decode noted inline).
2. Feature-gate the `Option<Arc<dyn …>>` slot on `Evm` with `#[cfg(feature = "experimental-devnet")]` so non-feature builds don't carry an unused slot.
3. Document the seam in `ethrex-vm` with a `# Why this is a trait` block so future contributors don't have to reason about the cycle from scratch.

**Cargo feature rename `eip-8025` → `experimental-devnet`**: makes sense as the umbrella for shared-infra EIPs. Worth noting in CHANGELOG, and grepping the workspace for stale `eip-8025` strings (CI configs, docs, READMEs).

**Tests gap**: integration test exists. Two coverage gaps to add before merge:
- `_l1MessagesCount` desync (caller claims N, payload consumes M, M < N) — should fail post-fix.
- Malformed-offset SSZ payload rejected by precompile rather than silently producing wrong proven values.

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,437 @@
...
+        // Configurable exit window (was hardcoded 0). A local demo passes 0 for
+        // instant finality; production should set a real reorg-safe delay.
+        FINALITY_DELAY = _finalityDelay;
+        advancer = _advancer;
+        lastFetchedL1Block = block.number;
+    }
+
+    /// @notice Rotate the authorized advancer. Callable only by the current
+    /// advancer (self-rotation), so a compromised/retired key can hand off.
+    function setAdvancer(address _newAdvancer) external onlyAdvancer {
```
ElFantasma, line 181: **Single-transaction total control loss.** `setAdvancer` immediately overwrites `advancer` with no confirmation from the target key: a typo permanently locks the contract (no upgrade path), and a compromised advancer key can transfer control to an attacker in one tx with no recovery window. For an L1-deployed immutable contract, the standard mitigation is a two-step handoff (OZ `Ownable2Step`-style): `setAdvancer` writes to a `pendingAdvancer`, and `acceptAdvancer` (only callable by that pending address) commits it. Optionally add a minimum delay between propose and accept to give observers time to react to a compromise.

Related: the L2Bridge relayer rotation likely wants the same treatment if it exists (I didn't check that side).

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,437 @@
...
+        _;
+    }
+
+    constructor(
+        bytes32 _initialStateRoot,
+        bytes32 _initialBlockHash,
+        uint256 _blockGasLimit,
+        uint256 _chainId,
+        address _advancer,
+        uint256 _finalityDelay
```
ElFantasma, line 142: **`FINALITY_DELAY = 0` accepted in production.** The comment below correctly says "production should set a real reorg-safe delay," but nothing enforces it — a deployment script bug (or a copy-paste from the demo) mints an immutable `FINALITY_DELAY = 0` on L1 with no recovery. Since the contract is not upgradable and the value is `immutable`, this is a one-shot misconfiguration risk.

Options: (a) a `MIN_FINALITY_DELAY` constant that the constructor floors against; (b) a distinct `deployProduction` factory that enforces a non-zero delay while keeping the demo path with `_finalityDelay = 0` via a separate `deployDemo`; (c) at minimum, require `_finalityDelay >= 1` and document that demos should explicitly pass 1 second.

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,437 @@
...
+    // Byte offsets inside the SSZ `ExecutionPayload` fixed prefix.
+    uint256 constant EP_PARENT_HASH_OFFSET = 0;
+    uint256 constant EP_STATE_ROOT_OFFSET = 52;
+    uint256 constant EP_BLOCK_NUMBER_OFFSET = 404;
+    uint256 constant EP_GAS_LIMIT_OFFSET = 412;
+    uint256 constant EP_BLOCK_HASH_OFFSET = 472;
+    // EIP-7843 slot_number is a trailing fixed u64 in the payload, right after the
+    // block_access_list offset slot (528..532). All reads above are unaffected.
+    uint256 constant EP_SLOT_NUMBER_OFFSET = 532;
+    uint256 constant EP_FIXED_PREFIX_LEN = 540; // 528 fixed fields + block_access_list offset(4) + slot_number(8)
```
ElFantasma, line 115: **Re-nudge: hardcoded SSZ field offsets** (flagged April 29). Concrete evidence the brittleness bites: `EP_FIXED_PREFIX_LEN` moved from 528 → 540 in this revision (+`EP_SLOT_NUMBER_OFFSET = 532` for EIP-7843), documented by hand. If any of the fixed-position fields shifts in a future `ExecutionPayload` schema (or a variable offset like `block_access_list` migrates position), the reads at lines 284–288 silently return attacker-controlled bytes as `provenBlockHash`/`provenStateRoot`.

The manual bump proves the maintenance burden — but not that it will be caught next time. Two mitigations worth combining:
1. **Solidity-side**: a schema-version byte at the start of `_sszStatelessInput`, checked in `_runExecutePrecompile`.
2. **Test-side**: a Rust unit test that constructs a canonical `SszStatelessInput`, calls the Rust `SszEncode` implementation, and asserts each hardcoded offset here matches the byte position. Gated on merge so drift breaks CI, not the deployed contract.

crates/vm/levm/src/execute_precompile.rs
```diff
@@ -0,0 +1,426 @@
...
+    validator: &dyn crate::StatelessValidator,
+    calldata: &Bytes,
+    gas_remaining: &mut u64,
+) -> Result<Bytes, VMError> {
+    use ethrex_common::types::stateless_ssz::SszStatelessInput;
+    use libssz::SszDecode;
+
+    // Attacker-controlled input: SSZ decode failure is a CALL-level failure, not an invariant.
+    let input = SszStatelessInput::from_ssz_bytes(calldata)
```
ElFantasma, line 66: **Re-nudge: double SSZ decode** (flagged April 29, greptile P2). Line 61 decodes `calldata` into `SszStatelessInput` for `validate_l2_constraints`, then line 81 passes raw `calldata` to `validator.verify(calldata)`, which decodes it again. Two full SSZ decodes per precompile invocation on adversary-controlled input. Fix shape: pass the already-decoded `SszStatelessInput` into `verify` (or have `verify` accept `&SszStatelessInput` and re-serialize its result payload separately).

tooling/l2/dev/blockscout_verify_native_rollup.py
ElFantasma: **Re-nudge: `HTTPConnection` for `https://` URLs** (flagged April 29, Copilot caught). Three call sites (lines 86, 105, 139) still unconditionally use `http.client.HTTPConnection(host, port)`. Any TLS Blockscout target breaks verification. Fix: parse `blockscout_url` with `urllib.parse.urlparse` up front, and branch to `http.client.HTTPSConnection` when `parsed.scheme == "https"` (or drop `http.client` for `requests` if the dev-tooling dependencies allow it).

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,437 @@
...
+        // Reject messages that can never fit in a single L2 block.
+        require(_gasLimit > 0, "gasLimit must be > 0");
+        require(_gasLimit <= l2GasLimit, "gasLimit exceeds L2 block gas limit");
+        _burnGas(_gasLimit);
+        // I7: account bridged-in ETH so claims can never exceed deposits.
+        totalDeposited += msg.value;
+        _recordL1Message(msg.sender, _to, msg.value, _gasLimit, _data);
+    }
+
+    receive() external payable {}
```
ElFantasma: nit: `receive()` accepts ETH without touching `totalDeposited`. Someone sending ETH directly (not via `sendL1Message`) increments the contract's balance but not the escrow accounting — the ETH is untracked and effectively stranded (no `L1MessageRecorded`, no L2 credit, no way to claim it back via `claimWithdrawal` since the escrow-solvency check gates on `totalDeposited`). Not a security hole (no path to over-withdraw), but the asymmetry is a footgun. Simplest fix: `revert()` in `receive()` so all deposits are forced through `sendL1Message`.

ElFantasma (commented):
Re-review of the revamped native rollup. 3 addressed since April (l1MessagesCount now bound via `provenL1MessagesRoot`, is_precompile L2-gates EXECUTE, blockscout script path fixed). Escrow solvency (`totalDeposited`/`totalClaimed`) and L2Bridge recoverable-balance are solid additions. 5 things worth addressing before this L1-deployed, immutable contract lands — 2 new security concerns on the hardening additions, 3 re-nudges on items still open from the April review.

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,496 @@
...
+    /// @param _data Calldata delivered to `_to` on L2.
+    function sendL1Message(address _to, uint256 _gasLimit, bytes calldata _data) external payable {
+        require(_gasLimit > 0, "NativeRollup: gasLimit must be > 0");
+        // Reject messages that can never fit in a single L2 block: the producer
+        // builds a relayer tx of `_gasLimit * 64/63 + RELAYER_GAS_BODY_ALLOWANCE`
+        // (block_producer.rs::build_relayer_transactions). Requiring that to be
+        // <= l2GasLimit means every accepted message is includable, so a valid
+        // deposit can never get permanently stuck.
+        require(
+            _gasLimit <= ((l2GasLimit - RELAYER_GAS_BODY_ALLOWANCE) * 63) / 64,
```
ElFantasma, line 216: **Missing constructor invariant on `l2GasLimit`.** The cap subtracts `RELAYER_GAS_BODY_ALLOWANCE` (300_000) from `l2GasLimit` before scaling. If a deployment ever bakes an immutable `l2GasLimit < 300_000`, Solidity 0.8+ reverts the subtraction on every `sendL1Message` call — permanently bricking the L1→L2 bridge with no way to reset (l2GasLimit is immutable, contract non-upgradeable). Realistic deployments use ≥15M so this is unlikely, but since the value is baked in for the contract's entire lifetime, a one-line constructor assert would be cheap defence in depth:

```solidity
require(_l2GasLimit > RELAYER_GAS_BODY_ALLOWANCE, "NativeRollup: l2GasLimit too small");
```

(A stricter form `_l2GasLimit >= RELAYER_GAS_BODY_ALLOWANCE + SOME_MIN_MESSAGE_GAS` would guarantee at least one non-trivial message is includable, but the weaker one already suffices to keep the bridge functional.) Non-blocking.

crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol
```diff
@@ -0,0 +1,496 @@
...
+    // Storage slot of `sentMessages` in L2Bridge.sol. Must match that layout —
+    // reordering L2Bridge storage silently breaks withdrawal proofs here.
+    uint256 constant SENT_MESSAGES_SLOT = 3;
+
+    // Fixed gas the L2 block producer reserves around each L1 message, on top of
+    // the 63/64 call-gas rule, when it builds the relayer tx (must match
+    // `build_relayer_transactions` in block_producer.rs). A message whose
+    // gasLimit leaves no room for this within one L2 block would be un-includable,
+    // so `sendL1Message` rejects it up front (see the cap there).
+    uint256 constant RELAYER_GAS_BODY_ALLOWANCE = 300_000;
```
ElFantasma, line 86: **Cross-file drift risk.** `RELAYER_GAS_BODY_ALLOWANCE = 300_000` is duplicated between here and `crates/l2/sequencer/native_rollup/block_producer.rs:53` (the `const RELAYER_GAS_BODY_ALLOWANCE: u64 = 300_000;` used by `relayer_tx_gas_limit`). The two must stay equal — a silent drift breaks the invariant that `sendL1Message`'s cap and the producer's actual relayer-tx sizing agree, either wedging the bridge again (rust > sol) or letting a message become un-includable (sol > rust). The new `native_rollup_sol_offsets.rs` drift-guard test already parses `.sol` constants against a runtime source of truth — extending it to also parse `RELAYER_GAS_BODY_ALLOWANCE` and compare with `block_producer::RELAYER_GAS_BODY_ALLOWANCE` would close this exactly the same way you closed the SSZ-offset drift. Nit / non-blocking.

cmd/ethrex/l2/deployer.rs
```diff
@@ -1697,3 +1777,244 @@ fn write_contract_addresses_to_env(
...
+        .unwrap_or_else(|| signer.address());
+    info!("Advancer address (authorized for advance()): {advancer:#x}");
+    // Exit window before a withdrawal can be claimed on L1, from
+    // --native-rollups.finality-delay (default 0 = instant finality for the local
+    // demo; production must pass a reorg-safe delay). Baked immutably into the
+    // contract at construction.
+    let finality_delay: u64 = opts.native_rollups_finality_delay;
+    if finality_delay == 0 {
+        warn!(
+            "NativeRollup FINALITY_DELAY is 0 (instant finality). This is only safe \
```
ElFantasma: **Re-nudge on the partial FINALITY_DELAY fix** (flagged 2026-07-13 as the second immutable-config trap). Adding the `--native-rollups.finality-delay` CLI flag is progress, but the default stays at 0 and the mitigation is only a `warn!` log — a CI or scripted deploy prints the warn and continues, and the immutable contract on L1 permanently commits FINALITY_DELAY=0. Options that would actually close the misconfig risk on this L1-deployed, non-upgradeable value:

- **No-default CLI**: require `--native-rollups.finality-delay=<seconds>` to be passed explicitly (no default), so a scripted deploy fails loudly if the operator forgot it. Preserves the demo path with an explicit `--native-rollups.finality-delay=0`.
- **Explicit demo opt-in**: keep `--native-rollups.finality-delay` defaulting to something safe (say 1 day), and gate zero behind `--native-rollups.allow-zero-finality-demo-only` — self-documenting in shell history / CI logs.

The demo path stays one flag away in either case; the difference is whether zero-in-production requires typing something the operator has to consciously read. Not blocking — flagging in case the team wants to close the immutable-config trap rather than accept it.

ElFantasma (commented):
Re-review after 385fc57 — 5/6 prior concerns cleanly addressed; three follow-ups inline (one partial FINALITY_DELAY re-nudge, two new non-blocking observations on the C1 hardening).

ElFantasma (approved):
LGTM

edg-l (commented):
**1. Default `--l2-gas-limit` bricks the contract**

Flag defaults to 30M:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/cmd/ethrex/l2/deployer.rs#L347-L354

`advance()` re-executes the L2 block in one L1 tx capped at `TX_MAX_GAS_LIMIT_AMSTERDAM` = 16.77M:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/common/constants.rs#L80

Constructor only checks the lower bound:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol#L140-L143

Docs deploy command doesn't pass `--l2-gas-limit`:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/docs/l2/deployment/native_rollups.md#L116-L122

Following the docs bakes 30M into the immutable `l2GasLimit` and every `advance()` runs out of gas. Only the Makefile's `15000000` avoids it. Add an upper-bound check in the deployer that fails loud, a matching `require` in the constructor, and fix the docs command.

**2. L1 messages dropped from the queue on any mid-block failure**

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/sequencer/native_rollup/block_producer.rs#L248-L270

Messages are popped off `pending_l1_messages` before the block is built. `produce_block` can bail on a relayer-tx or signer error before `store_block`, with no requeue:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/sequencer/native_rollup/block_producer.rs#L133-L221

Popped messages are lost, and `processL1Message` requires sequential nonces, so every later message reverts and the bridge is stuck until a node restart. Peek instead of pop, remove only after `store_block` succeeds.

**3. Watcher advances its cursor past parse-failed blocks**

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/sequencer/native_rollup/l1_watcher.rs#L136-L183

A log that fails parsing is dropped with a `warn!`, then `last_block_fetched` advances unconditionally, so a malformed event is lost on every rescan and its deposit is stranded. Don't advance the cursor past a range that had a parse failure.

**4. Deposited ETH can be locked with no way out**

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol#L189-L204

`sendL1Message` escrows value into `totalDeposited`. The only exit is a matching L2 withdrawal, and `advance()` accepts `_l1MessagesCount = 0`, so nothing forces consumption:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol#L241-L261

`receive()` reverts and there's no refund path, so a stalled advancer strands user funds permanently:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/l2/contracts/src/nativeRollup/l1/NativeRollup.sol#L209-L211

Add a per-message timeout refund that decrements `totalDeposited`, like OCP does for privileged txs.

**5. `base_fee_per_gas` truncation makes `hash_tree_root` malleable**

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/guest-program/src/l1/program.rs#L214-L220

Helper takes the low 8 bytes of the 32-byte SSZ field with no check that bytes `[8..32]` are zero. Inline copy does the same:

https://github.com/lambdaclass/ethrex/blob/adf6e266e57b7da56051f382eb1b96d8cce52941/crates/guest-program/src/l1/program.rs#L456-L462

Those bytes are in `NewPayloadRequest::hash_tree_root()` but don't affect validation, so ~2^192 distinct roots validate for the same block. Harmless for the current contract since it never uses the root, but it breaks the commitment invariant for any root-keyed caller or ZK variant. Reject non-zero upper bytes in the helper and reuse it at L458.

iovoid:
Fixed the issues. For 4, the existing OCP does not allow for _refunds_, it only stops the network if messages aren't being relayed. I think it might be out of scope to implement refunds now.

edg-l (approved):

ilitteri (approved):

PR #6419, opened 2026-03-29
fix(l1): guard full sync throughput when elapsed time is zero

ArshLabs:
**Motivation**

full sync batch throughput in `add_blocks_in_batch` used millisecond-truncated elapsed time. for very fast batches this can evaluate to `0.0`, which makes `blocks_per_second` log as `inf`.

**Description**

this PR updates the throughput calculation in `crates/networking/p2p/sync/full.rs` to:

- use `as_secs_f64()` for elapsed time
- guard the division when elapsed time is `0.0`
- keep existing sync logging format and behavior unchanged otherwise

no execution, storage, or forkchoice behavior is changed. this is an observability correctness fix for full sync logs.

no linked issue, self-discovered during code tracing.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. (not needed)
- [x] `cargo fmt --check`
- [ ] `cargo build` (blocked in this Windows env by upstream git dependency checkout path with trailing space in `agg_mode_sdk`)
- [ ] `cargo test` (same blocker as above)
- [ ] `cargo clippy -- -D warnings` (same blocker as above)


ElFantasma (approved):

crates/networking/p2p/sync/full.rs
```diff
@@ -247,8 +247,12 @@ async fn add_blocks_in_batch(
         )
         .await?;
 
-    let execution_time: f64 = execution_start.elapsed().as_millis() as f64 / 1000.0;
-    let blocks_per_second = blocks_len as f64 / execution_time;
+    let execution_time = execution_start.elapsed().as_secs_f64();
+    let blocks_per_second = if execution_time > 0.0 {
+        blocks_len as f64 / execution_time
+    } else {
+        0.0
```
iovoid, line 255: Not sure if zero is the best option here. That said using `as_secs_f64` intead of `as_millis` should already make this scenario very unlikely.

ArshLabs:
pushed an update here, branch is now rebased on latest `main` and the zero-fallback path has an inline comment clarifying intent.

behavior stays the same, throughput still returns `0.0` when elapsed time is zero so we avoid `inf`/`nan` in sync logs.

@iovoid does this look good to you now?

iovoid:
I was thinking some text, but considering `as_secs_f64` returning 0.0 shouldn't happen a comment is ok.

iovoid (approved):

ArshLabs:
quick triage on the remaining red check:
`L1/Hive - Consume Engine Amsterdam` is the only failing job, and the failed cases are in `tests/amsterdam/eip7778_block_gas_accounting_without_refunds/...`.

this does not touch those paths, and the same Hive Amsterdam job also failed in a recent `main` L1 run (`24261519308`).

looks like unrelated/flaky CI here, can someone with repo admin perms rerun failed jobs for this PR run?

PR #6429, opened 2026-03-31
feat(l1): add peer scoring by latency and bandwidth

azteca1998:
## Summary

- Introduces a pluggable `ScoringStrategy` enum (`Success`, `Latency`, `Bandwidth`) for P2P peer selection, with per-phase configuration
- Adds `PeerMetrics` struct on each `PeerData` tracking latency EMA (ms) and bandwidth EMA (bytes/s) using a rolling 20-sample window with exponential moving average (α=0.3)
- Instruments `make_request`, snap sync workers (account range, storage ranges, bytecodes), and header download workers to report `TransferStats` (elapsed time + response bytes)
- All three scoring dimensions are always collected regardless of active strategy

### Per-phase CLI configuration

During snap sync, bandwidth matters most (bulk data transfer). During live/full sync, latency matters most (block-by-block propagation). Two new CLI flags control this:

```
--p2p.snap-scoring <strategy>   # default: bandwidth
--p2p.live-scoring <strategy>   # default: latency
```

Where `<strategy>` is one of: `success`, `latency`, `bandwidth`.

The `Syncer` automatically switches strategy when entering snap vs full/live sync cycles.

## Test plan

- [x] `cargo check --workspace` passes with zero warnings
- [x] `cargo clippy -p ethrex-p2p -p ethrex` passes with zero warnings
- [ ] Verify peer selection behavior under each strategy with live peers
- [ ] Validate that latency/bandwidth EMA converges to expected values under controlled conditions

### DIFF
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -76,18 +76,15 @@ async fn ask_peer_head_number(
 
     debug!("(Retry {retries}) Requesting sync head {sync_head:?} to peer {peer_id}");
 
-    let req_start = Instant::now();
-    let result = connection
+    match connection
         .outgoing_request(request, PEER_REPLY_TIMEOUT)
-        .await;
-    let elapsed = req_start.elapsed();
-    match result {
+        .await
+    {
         Ok(RLPxMessage::BlockHeaders(BlockHeaders {
             id: _,
             block_headers,
         })) => {
             if !block_headers.is_empty() {
-                let _ = peer_table.record_response_latency(peer_id, elapsed);
                 let sync_head_number = block_headers
                     .last()
                     .ok_or(PeerHandlerError::BlockHeaders)?
@@ -684,18 +681,15 @@ impl PeerHandler {
             reverse: false,
         });
         debug!("get_block_header: requesting header with number {block_number}");
-        let req_start = Instant::now();
-        let result = connection
+        match connection
             .outgoing_request(request, PEER_REPLY_TIMEOUT)
-            .await;
-        let elapsed = req_start.elapsed();
-        match result {
+            .await
+        {
             Ok(RLPxMessage::BlockHeaders(BlockHeaders {
                 id: _,
                 block_headers,
             })) => {
                 if !block_headers.is_empty() {
-                    let _ = self.peer_table.record_response_latency(peer_id, elapsed);
                     return Ok(Some(
                         block_headers
                             .last()
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -22,6 +22,7 @@ use ethrex_common::{
     H256,
     types::{BlockBody, BlockHeader, block_access_list::BlockAccessList, validate_block_body},
 };
+use ethrex_rlp::encode::RLPEncode;
 use ethrex_crypto::NativeCrypto;
 use spawned_concurrency::{error::ActorError, tasks::ActorRef};
 use std::{
@@ -259,10 +260,13 @@ impl PeerHandler {
                 if !elapsed.is_zero() {
                     let _ = self.peer_table.record_response_latency(peer_id, elapsed);
                     if !headers.is_empty() {
-                        let approx_bytes = headers.len() as u64 * 500;
+                        let response_bytes: u64 = headers
+                            .iter()
+                            .map(|h| h.encode_to_vec().len() as u64)
+                            .sum();
                         let _ = self
                             .peer_table
-                            .record_bandwidth(peer_id, approx_bytes, elapsed);
+                            .record_bandwidth(peer_id, response_bytes, elapsed);
                     }
                 }
 
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -464,9 +464,7 @@ impl PeerHandler {
                         return Ok(None);
                     }
                     if are_block_headers_chained(&block_headers, &order) {
-                        let _ = self
-                            .peer_table
-                            .record_response_latency(peer_id, elapsed);
+                        let _ = self.peer_table.record_response_latency(peer_id, elapsed);
                         self.peer_table.record_success(peer_id)?;
                         return Ok(Some(block_headers));
                     }
@@ -556,9 +554,7 @@ impl PeerHandler {
                 {
                     // Check that the response is not empty and does not contain more bodies than the ones requested
                     if !block_bodies.is_empty() && block_bodies.len() <= block_hashes_len {
-                        let _ = self
-                            .peer_table
-                            .record_response_latency(peer_id, elapsed);
+                        let _ = self.peer_table.record_response_latency(peer_id, elapsed);
                         self.peer_table.record_success(peer_id)?;
                         return Ok(Some((block_bodies, peer_id)));
                     }
@@ -699,9 +695,7 @@ impl PeerHandler {
                 block_headers,
             })) => {
                 if !block_headers.is_empty() {
-                    let _ = self
-                        .peer_table
-                        .record_response_latency(peer_id, elapsed);
+                    let _ = self.peer_table.record_response_latency(peer_id, elapsed);
                     return Ok(Some(
                         block_headers
                             .last()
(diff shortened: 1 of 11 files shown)


crates/networking/p2p/peer_table.rs
```diff
@@ -810,36 +975,39 @@ impl PeerTableServer {
...
+    fn weight_peer(&self, peer_data: &PeerData) -> i64 {
+        let score = peer_data.effective_score(self.scoring_strategy);
+        score * SCORE_WEIGHT - peer_data.requests * REQUESTS_WEIGHT
     }
 
     // Returns if the peer has room for more connections given the current score
-    // and amount of inflight requests
-    fn can_try_more_requests(&self, score: &i64, requests: &i64) -> bool {
+    // and amount of inflight requests.
+    fn can_try_more_requests(&self, peer_data: &PeerData) -> bool {
```
ElFantasma, line 1011: `can_try_more_requests` now uses `effective_score(self.scoring_strategy)` which returns 0 for peers with no latency/bandwidth samples yet. With `score=0` and `MIN_SCORE=-50`, `score_ratio = 50/100 = 0.5`, so `max_requests = 50`. That's fine.

But `latency_score()` and `bandwidth_score()` both return 0 for `None` EMA. This means a brand-new peer (no samples) gets the same score as a peer at the 50ms/2MB boundary. Consider returning a neutral score (e.g. `(MAX_SCORE + MIN_SCORE) / 2`) for `None` to distinguish "unknown" from "measured-and-at-boundary".

crates/networking/p2p/peer_handler.rs
```diff
@@ -247,11 +259,21 @@ impl PeerHandler {
...
+            if let Ok((headers, peer_id, _connection, startblock, previous_chunk_limit, elapsed)) =
                 task_receiver.try_recv()
             {
                 // Release the reservation we made before spawning the task.
                 self.peer_table.dec_requests(peer_id)?;
+                // Feed transfer stats for latency/bandwidth scoring.
+                if !elapsed.is_zero() {
+                    let _ = self.peer_table.record_response_latency(peer_id, elapsed);
+                    if !headers.is_empty() {
+                        let approx_bytes = headers.len() as u64 * 500;
```
ElFantasma: nit: `headers.len() * 500` is a rough estimate. Headers are ~508 bytes RLP-encoded (post-merge, no ommers), so 500 is reasonable for the average case. But for bandwidth scoring accuracy, you could use the actual response size from the RLPx frame if available, similar to how the snap workers compute `response_bytes` from the actual data.

ElFantasma (changes requested):

crates/networking/p2p/peer_table.rs
```diff
@@ -204,6 +321,16 @@ impl PeerData {
...
             requests: Default::default(),
+            metrics: Default::default(),
+        }
+    }
+
+    /// Return the effective score for this peer under the given strategy.
+    pub fn effective_score(&self, strategy: ScoringStrategy) -> i64 {
+        match strategy {
+            ScoringStrategy::Success => self.score,
+            ScoringStrategy::Latency => self.metrics.latency_score(),
```
ElFantasma: **Critical: failure penalties are silently ignored under Latency/Bandwidth strategies.**

`record_failure()` and `record_critical_failure()` only mutate `peer_data.score`, but `effective_score()` here returns ONLY the metric-derived score, completely ignoring `self.score`. Since snap/full sync default to `Bandwidth`/`Latency` respectively, the bad-peer quarantine path is effectively disabled by default:

- Peer returns invalid block → `record_critical_failure()` → `score = MIN_SCORE_CRITICAL = -150`
- But under `Latency`: `effective_score` returns `latency_score()` (neutral midpoint = 0 if no samples, or whatever the EMA says)
- `can_try_more_requests`: `score_ratio = 0.5` → up to 50 concurrent requests allowed
- `weight_peer`: peer ranks neutrally instead of being quarantined

Minimum fix: hard-floor on critical failures regardless of strategy:

```rust
pub fn effective_score(&self, strategy: ScoringStrategy) -> i64 {
    // Always honor the critical-failure quarantine.
    if self.score <= MIN_SCORE {
        return self.score;
    }
    match strategy {
        ScoringStrategy::Success => self.score,
        ScoringStrategy::Latency => self.metrics.latency_score(),
        ScoringStrategy::Bandwidth => self.metrics.bandwidth_score(),
    }
}
```

Or better: blend `self.score` into all strategies so accumulated failures still degrade selection.

crates/networking/p2p/peer_table.rs
```diff
@@ -45,6 +46,120 @@ const SCORE_WEIGHT: i64 = 1;
...
+            snap: ScoringStrategy::Bandwidth,
+            live: ScoringStrategy::Latency,
+        }
+    }
+}
+
+/// Rolling performance metrics collected per peer.
+#[derive(Debug, Clone, Default)]
+pub struct PeerMetrics {
+    latency_samples: VecDeque<Duration>,
```
ElFantasma: nit: `latency_samples` and `bandwidth_samples` are written by `record_latency`/`record_bandwidth` but never read — only the EMA fields are consumed by `latency_score()`/`bandwidth_score()`. The `METRICS_WINDOW_SIZE` cap bounds them but they otherwise sit unused. Either remove them and the push/pop logic, or use them (e.g. for stddev / median scoring or EMA reset logic).

crates/networking/p2p/peer_handler.rs
```diff
@@ -114,8 +115,14 @@ impl PeerHandler {
         timeout: Duration,
     ) -> Result<RLPxMessage, PeerConnectionError> {
         peer_table.inc_requests(peer_id)?;
+        let start = Instant::now();
         let result = connection.outgoing_request(message, timeout).await;
+        let latency = start.elapsed();
         peer_table.dec_requests(peer_id)?;
+        // Record latency for successful responses (timeouts are already penalized via record_failure).
+        if result.is_ok() {
+            let _ = peer_table.record_response_latency(peer_id, latency);
```
ElFantasma: Telemetry is recorded based on `result.is_ok()` (RLPx layer) but before payload validation. A peer that responds quickly with malformed/incorrect data improves its latency score before the validation layer rejects the response. Combined with the failure-penalty issue above, a malicious peer could maintain a "fast" reputation while serving garbage. Consider deferring `record_response_latency` until after the caller validates the payload, or only crediting on validated responses.

ElFantasma (changes requested):
Thanks for addressing the previous round. The neutral midpoint and accurate header bandwidth fixes look good. However, there's a more serious issue worth addressing before merge that several AI reviewers also flagged: the failure-penalty path is silently bypassed under Latency/Bandwidth strategies. Two smaller items inline as well.

crates/networking/p2p/peer_table.rs
```diff
@@ -45,6 +46,120 @@ const SCORE_WEIGHT: i64 = 1;
...
+    latency_ema_ms: Option<f64>,
+    bandwidth_samples: VecDeque<f64>,
+    bandwidth_ema: Option<f64>,
```
iovoid: The samples are recorded but not used.

crates/networking/p2p/snap/client.rs
```diff
@@ -191,9 +203,22 @@ pub async fn request_account_range(
...
+        if let Ok((accounts, peer_id, chunk_start_end, stats)) = task_receiver.try_recv() {
             // Release the reservation we made before spawning the task.
             peers.peer_table.dec_requests(peer_id)?;
+            // Feed transfer stats for bandwidth/latency scoring.
+            if !stats.elapsed.is_zero() {
+                let _ = peers
+                    .peer_table
+                    .record_response_latency(peer_id, stats.elapsed);
+                if stats.response_bytes > 0 {
+                    let _ = peers.peer_table.record_bandwidth(
```
iovoid: This calculation mostly measures latency rather than throughtput.

Assuming all responses are 1MB for simplicity, compare the following scenarios:

- In 1s Node A sends 100 packets after 100ms of delay. Estimated bandwidth: 1MB/100ms=10MB/s, real bandwidth: 1MB*100/1s=100MB/s
- In 1s Node B sends 10 packets after 10ms of delay. Estimated bandwidth: 1MB/10ms=100MB/s, real bandwidth: 1MB*10/1s=10MB/s

Node A has 10x the bandwidth yet it will be scored lower

iovoid (changes requested):

ElFantasma (approved):

edg-l (approved):

PR #6433, opened 2026-04-01
feat(l1): injectable U256 backend via install_uint256_backend()

ilitteri:
## Motivation

zkVM vendors (ZisK, SP1, Airbender, etc.) need to swap or override the U256 arithmetic backend — either to use a different library (e.g., ruint instead of ethereum_types) or to delegate operations to hardware-accelerated instructions (e.g., custom RISC-V CSR instructions as in zksync-os/Airbender). Vendors should be able to install their backend with a single function call, matching the `install_crypto` pattern from revm and `install_default_provider` from alloy-consensus.

## Description

Replaces the compile-time feature flag approach (`uint-ethereum-types` / `uint-ruint`) with a runtime-injectable backend:

- **`Uint256Ops` trait** defines all U256/U512 operations. Every method has a default implementation using ethereum_types, so vendors only override what they need.
- **`install_uint256_backend()`** stores a custom backend in a global `OnceLock<Box<dyn Uint256Ops>>`.
- **`U256`/`U512`** are `#[repr(transparent)]` wrappers over `[u64; 4]`/`[u64; 8]` that delegate all operations through the installed backend.
- If no backend is installed, the default (ethereum_types) kicks in automatically.

### Vendor UX

```rust
// In the vendor's guest main():
use ethrex_common::install_uint256_backend;

struct MyU256Ops;
impl ethrex_common::Uint256Ops for MyU256Ops {
    fn overflowing_add(&self, a: [u64; 4], b: [u64; 4]) -> ([u64; 4], bool) {
        unsafe { hw_accelerated_add(&a, &b) }
    }
    // ... override only what you need, defaults handle the rest
}

fn main() {
    install_uint256_backend(MyU256Ops);
    // All U256 operations now use the custom backend
}
```

No ethrex source modifications needed. The ZisK guest includes a ruint backend as a usage example.

### Changes

- `crates/common/uint.rs`: Major rewrite — `Uint256Ops` trait, `OnceLock` global, `U256` over `[u64; 4]` with delegation, manual `Ord` (LE limb ordering)
- `crates/common/common.rs`: Export `Uint256Ops`, `DefaultUint256Ops`, `install_uint256_backend`
- `crates/common/Cargo.toml`: Remove `uint-ethereum-types`/`uint-ruint` features and `ruint` optional dep
- `Cargo.toml`, `crates/guest-program/Cargo.toml`, `crates/guest-program/bin/zisk/Cargo.toml`: Remove uint feature wiring
- `crates/guest-program/bin/zisk/`: Add `ruint_backend.rs` example + `install_uint256_backend` call in `main()`
- `test/tests/common/uint_tests.rs`: Backend compatibility tests

### Benchmark (ziskemu -X -S -D, mainnet block 24736913)

| Metric | ethereum_types (feature flag) | ruint (feature flag) | ethereum_types (injectable) | ruint (injectable) | injectable overhead (et) | injectable overhead (ruint) |
|--------|-------------------------------|----------------------|-----------------------------|---------------------|--------------------------|------------------------------|
| Status | Pass | Pass | Pass | Pass | | |
| STEPS | 377,407,002 | 361,368,662 | 396,994,519 | 379,365,290 | +5.2% | +5.0% |
| TOTAL COST | 43,986,607,573 | 42,410,545,492 | 45,789,509,184 | 44,056,877,515 | +4.1% | +3.9% |
| ELF size | 3,861,776 | 3,911,928 | 3,874,696 | 3,905,528 | +0.3% | -0.2% |

## How to Test

```bash
# Native execution (default backend, no setup needed)
cargo run --release -p ethrex-replay -- block 24736913 --rpc-url <RPC> --no-zkvm

# ZisK guest with ruint backend
make -C crates/guest-program zisk
ziskemu -X -S -D --elf crates/guest-program/bin/zisk/out/riscv64ima-zisk-elf --inputs <input.bin>
```


crates/common/uint.rs
```diff
@@ -0,0 +1,861 @@
...
+impl_binop!(BitXor, bitxor);
+
+// Cross-type arithmetic (U256 op primitive). ethereum_types supports these.
+macro_rules! impl_cross_binop {
+    ($prim:ty, $trait:ident, $method:ident) => {
+        impl core::ops::$trait<$prim> for U256 {
+            type Output = Self;
+            #[inline]
+            fn $method(self, rhs: $prim) -> Self {
+                core::ops::$trait::$method(self, U256::from(rhs as u64))
```
ElFantasma: Bug: `rhs as u64` for `i32` breaks sign extension. When `rhs` is negative (e.g. `-1i32`), `rhs as u64` yields `18446744073709551615u64`, then `U256::from(u64)` gives `U256::from_limbs([u64::MAX, 0, 0, 0])`. But the dedicated `From<i32> for U256` (line 477) correctly produces `U256::MAX` (all limbs `u64::MAX`).

For the `i32` variants, this should use `U256::from(rhs)` instead of `U256::from(rhs as u64)` to go through the proper sign-extending conversion. You could specialize the i32 impls separately or add a second macro parameter for the conversion.

The `usize` cast is also technically lossy on 32-bit targets (truncation in the other direction), though that's unlikely to matter in practice.

crates/common/uint.rs
```diff
@@ -0,0 +1,861 @@
...
+    /// Returns the low 64 bits.
+    #[inline]
+    pub fn as_u64(&self) -> u64 {
+        #[cfg(feature = "uint-ethereum-types")]
+        {
+            self.0.as_u64()
+        }
+        #[cfg(feature = "uint-ruint")]
+        {
+            self.0.as_limbs()[0]
```
ElFantasma: Behavioral difference: `ethereum_types::U256::as_u64()` panics if upper limbs are non-zero, while this ruint path silently truncates to the lowest limb. This could mask bugs when switching to the ruint backend. Consider adding an overflow check (or at least documenting the divergence) so both backends behave identically.

crates/common/uint.rs
```diff
@@ -0,0 +1,861 @@
...
+
+    /// Convert U256 to H256 (big-endian representation).
+    #[inline]
+    pub fn to_h256(self) -> ethereum_types::H256 {
+        ethereum_types::H256::from(self.to_big_endian())
+    }
+}
+
+/// Implement the BigEndianHash trait from ethereum_types for our wrapper.
+impl ethereum_types::BigEndianHash for U256 {
```
ElFantasma, line 1030: Nit: this `BigEndianHash` impl is an identity operation (`from_uint` and `into_uint` both return `*self`), which is semantically wrong — `BigEndianHash` is meant to convert between hash types and uint types (`H256 <-> U256`), not to be a no-op on a single type. Since all callers have been migrated to `to_h256()`/`from_h256()` it's harmless, but it could confuse future contributors who import `BigEndianHash` expecting actual conversion behavior. Consider either removing it or adding a comment explaining why it exists.

ElFantasma (changes requested):

azteca1998 (approved):

crates/common/uint.rs
```diff
@@ -0,0 +1,966 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// U512
+// ---------------------------------------------------------------------------
+
+/// 512-bit unsigned integer. Used for intermediate ADDMOD arithmetic.
+#[repr(transparent)]
+#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct U512([u64; 8]);
```
iovoid, line 1069: Not documented if this is LE or BE. If it's LE, the Ord implementation is wrong. If it's BE, low_u256 is wrong.

iovoid (changes requested):

crates/common/uint.rs
ilitteri: f07fc4e45

crates/common/uint.rs
```diff
@@ -0,0 +1,848 @@
...
+impl_cross_binop!(u64, Mul, mul);
+impl_cross_binop!(u64, Div, div);
+impl_cross_binop!(usize, Add, add);
+impl_cross_binop!(usize, Sub, sub);
+impl_cross_binop!(usize, Mul, mul);
+impl_cross_binop!(usize, Div, div);
+impl_cross_binop!(i32, Add, add);
+impl_cross_binop!(i32, Sub, sub);
+impl_cross_binop!(i32, Mul, mul);
+impl_cross_binop!(i32, Div, div);
```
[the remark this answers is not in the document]
ilitteri: f07fc4e45

crates/common/uint.rs
ilitteri: 86ebb4331

crates/common/uint.rs
ilitteri: d9eb942c1

ElFantasma (approved):

crates/common/uint.rs
```diff
@@ -0,0 +1,1147 @@
...
+#[derive(Debug)]
+pub struct ParseU256Error(pub String);
+
+impl core::fmt::Display for ParseU256Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl std::error::Error for ParseU256Error {}
```
iovoid, line 249: Why not use thiserror like the other modules?

iovoid (approved):

PR #6449, opened 2026-04-07
refactor(l1): unify discovery servers into single DiscoveryServer actor

### LINKED ISSUE #5990: Unify discovery protocol GenServers into a single DiscoveryServer
## Description

Refactor the discovery protocol architecture to use a single GenServer instead of three separate ones (DiscoveryMultiplexer + Discv4Server + Discv5Server). Since PeerTable is already shared and the multiplexer already discriminates packets, merging these GenServers simplifies the architecture and eliminates inter-GenServer communication overhead.

## Current Architecture (4 GenServers)

```
┌─────────────────────┐
│ DiscoveryMultiplexer│ ← routes packets
└──────────┬──────────┘
           │
     ┌─────┴─────┐
     ▼           ▼
┌─────────┐ ┌─────────┐
│ Discv4  │ │ Discv5  │ ← separate GenServers
│ Server  │ │ Server  │
└────┬────┘ └────┬────┘
     │           │
     └─────┬─────┘
           ▼
    ┌───────────┐
    │ PeerTable │ ← shared state
    └───────────┘
```

## Proposed Architecture (2 GenServers)

```
┌──────────────────────────────────┐
│        DiscoveryServer           │
│  ┌─────────────┬──────────────┐  │
│  │ Discv4State │ Discv5State  │  │ ← protocol-specific state
│  └─────────────┴──────────────┘  │
│  ┌────────────────────────────┐  │
│  │   Shared: socket, signer,  │  │ ← common fields
│  │   local_node, bootnodes    │  │
│  └────────────────────────────┘  │
│  • Single message loop           │
│  • Protocol dispatch via enum    │
│  • Unified timers                │
└───────────────┬──────────────────┘
                ▼
         ┌───────────┐
         │ PeerTable │
         └───────────┘
```

## Benefits

- Eliminates inter-GenServer communication overhead (no more casting between multiplexer and protocol servers)
- Single lifecycle & shutdown coordination
- Unified logging with protocol tags
- Simpler architecture to reason about
- Single place for revalidate/lookup/prune logic with protocol parameter
- Reduced code duplication (~30-40% of current code is duplicated between servers)

## Implementation Approach

1. **Combine state structs**: Create `Discv4State` and `Discv5State` structs for protocol-specific fields
2. **Unified InMessage enum**: Single enum with protocol-tagged variants
   ```rust
   enum InMessage {
       Discv4(Discv4Message),
       Discv5(Discv5Message),
       Revalidate(DiscoveryProtocol),
       Lookup(DiscoveryProtocol),
       Prune,
       Shutdown,
   }
   ```
3. **Protocol dispatch**: Route to `handle_discv4_message()` or `handle_discv5_packet()` based on variant
4. **Unified timers**: Single timer management with protocol parameter where intervals differ
5. **Extract shared logic**: Common implementations for `revalidate()`, `lookup()`, `prune()` with protocol parameter

## Protocol-Specific State

**Discv4:**
- `find_node_message: BytesMut` (cached, re-signed every 5 seconds)
- Revalidation interval: 12 hours

**Discv5:**
- `counter: u32` (nonce generation)
- `pending_by_nonce: FxHashMap` (messages awaiting WhoAreYou)
- `pending_challenges: FxHashMap` (WhoAreYou challenges)
- `whoareyou_rate_limit: FxHashMap` (per-IP rate limiting)
- `ip_votes` + voting state (external IP detection)
- Revalidation interval: 30 seconds

## Dependencies

- Should be done after #5962 (discovery multiplexer) is merged

## Related Issues

- #5971 - Remove "experimental-discv5" feature flag
- #5972 - Detect and address performance degradation in discovery operations

ElFantasma:
**Motivation**

Resolves #5990. The discovery protocol uses four GenServers (DiscoveryMultiplexer + Discv4Server + Discv5Server + PeerTable) with significant inter-GenServer communication overhead and ~30-40% code duplication between v4 and v5 servers. Since PeerTable is already shared and the multiplexer already discriminates packets, merging the three discovery GenServers into a single DiscoveryServer simplifies the architecture.

**Description**

Merges `DiscoveryMultiplexer`, `discv4::DiscoveryServer`, and `discv5::DiscoveryServer` into a single unified `discovery::DiscoveryServer` actor.

**Architecture: 4 actors → 2 actors** (DiscoveryServer + PeerTable)

```
Before:                              After:
┌─────────────────────┐              ┌──────────────────────────────────┐
│ DiscoveryMultiplexer│              │        DiscoveryServer           │
└──────────┬──────────┘              │  ┌─────────────┬──────────────┐  │
           │                         │  │ Discv4State │ Discv5State  │  │
     ┌─────┴─────┐                   │  └─────────────┴──────────────┘  │
     ▼           ▼                   │  • Single UDP receive loop       │
┌─────────┐ ┌─────────┐             │  • Inline packet discrimination  │
│ Discv4  │ │ Discv5  │             │  • Unified timers                │
│ Server  │ │ Server  │             └───────────────┬──────────────────┘
└────┬────┘ └────┬────┘                             ▼
     └─────┬─────┘                           ┌───────────┐
           ▼                                 │ PeerTable │
    ┌───────────┐                            └───────────┘
    │ PeerTable │
    └───────────┘
```

**Key changes:**

- **New:** `discovery/server.rs` — unified actor struct, protocol, startup, periodic task handlers, packet discrimination
- **New:** `discovery/discv4_handlers.rs` — all discv4 handler logic as `impl DiscoveryServer` methods
- **New:** `discovery/discv5_handlers.rs` — all discv5 handler logic as `impl DiscoveryServer` methods
- **Modified:** `discv4/server.rs` — replaced old actor with `Discv4State` data struct + `Discv4Message` type
- **Modified:** `discv5/server.rs` — replaced old actor with `Discv5State` data struct + `Discv5Message` type
- **Modified:** `network.rs` — spawns single `DiscoveryServer` instead of 3 actors
- **Removed:** `discovery/multiplexer.rs` — no longer needed
- Moved `INITIAL_LOOKUP_INTERVAL_MS`, `LOOKUP_INTERVAL_MS`, `lookup_interval_function` to `discovery/mod.rs` (shared by discv4, discv5, and RLPx initiator)
- `DiscoveryConfig` now includes `initial_lookup_interval` field

No behavioral changes — same protocols, same packet formats, same peer discovery semantics. Net -696 lines.

Closes #5990

**Checklist**

All items addressed — no Store schema changes.


crates/networking/p2p/discovery/discv4_handlers.rs
```diff
@@ -0,0 +1,541 @@
...
+/// Discv4 revalidation interval.
+const REVALIDATION_INTERVAL: Duration = Duration::from_secs(12 * 60 * 60); // 12 hours
+
+impl DiscoveryServer {
+    pub(crate) fn discv4_process_message(&mut self, msg: Discv4Message) {
+        let _ = tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(self.discv4_process_message_inner(msg))
+        })
+        .inspect_err(|e| error!(protocol = "discv4", err=?e, "Error Handling Discovery message"));
+    }
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 506ec7f — made `route_packet`, `route_to_discv4`, and `route_to_discv5` async, removing the `block_in_place` + `block_on` bridge entirely. The `#[send_handler]` on `handle_raw_packet` is already async, so the handlers now just `.await` directly.

crates/networking/p2p/discv5/server.rs
```diff
@@ -1077,117 +85,92 @@ impl DiscoveryServer {
...
+                updater.local_node_record,
+                updater.signer,
+                winning_ip,
+            );
         }
 
-        // Reset for next round
         self.ip_votes.clear();
         self.ip_vote_period_start = Some(Instant::now());
         self.first_ip_vote_round_completed = true;
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 506ec7f — removed `IpUpdater` entirely. `record_ip_vote` and `cleanup_stale_entries` now return `Option<IpAddr>` (the winning IP to apply). The caller in `discv5_handle_pong` and the `prune()` handler apply the update directly where they have access to `self.local_node` / `self.local_node_record` / `self.signer`.

crates/networking/p2p/discovery/discv5_handlers.rs
```diff
@@ -0,0 +1,800 @@
...
+        tracing::trace!(protocol = "discv5", received = "WhoAreYou", from = %node.node_id(), %addr);
+
+        let challenge_data = build_challenge_data(
+            &packet.masking_iv,
+            &packet.header.static_header,
+            &packet.header.authdata,
+        );
+
+        let ephemeral_key = SecretKey::new(&mut rand::thread_rng());
+        let ephemeral_pubkey = ephemeral_key.public_key(secp256k1::SECP256K1).serialize();
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 506ec7f — replaced with `OsRng`.

crates/networking/p2p/discovery/mod.rs
```diff
@@ -12,18 +9,38 @@
...
+pub use server::{DiscoveryServer, DiscoveryServerError, is_discv4_packet};
 
 use std::time::Duration;
 
+/// Configuration for which discovery protocols to enable.
+#[derive(Debug, Clone)]
+pub struct DiscoveryConfig {
+    pub discv4_enabled: bool,
+    pub discv5_enabled: bool,
+    pub initial_lookup_interval: f64,
```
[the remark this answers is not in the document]
ElFantasma: Acknowledged — the default in `DiscoveryConfig::default()` is mainly useful for tests and standalone usage. In production, `start_network` always overrides it from `P2PContext::initial_lookup_interval`. Changed the override pattern to use struct update syntax (`DiscoveryConfig { initial_lookup_interval: context.initial_lookup_interval, ..config }`) to make the intent clearer.

ElFantasma:
Addressed all review feedback in 506ec7f:

**Fixed:**
1. **IP voting regression** — `record_ip_vote` and `cleanup_stale_entries` now return `Option<IpAddr>` when a round ends with a winning IP. The caller (`discv5_handle_pong` and `prune()`) applies the update directly where it has access to `local_node`/`local_node_record`/`signer`. Removed the unused `IpUpdater` struct.
2. **`block_in_place` removed** — `route_packet`, `route_to_discv4`, `route_to_discv5` are now `async fn` and awaited directly from the `#[send_handler]`. No more sync bridge.
3. **`rand::thread_rng()` → `OsRng`** — consistent with the rest of the file.
4. **`get_lookup_interval` takes `&self`** instead of `&mut self`.
5. **`initial_lookup_interval` override** — uses struct update syntax for clarity.

**Acknowledged but not changing:**
- **`pub` fields on `DiscoveryServer`** — needed for external test crate access (`test/tests/p2p/discovery/`). The old discv5 server had the same fields as `pub`.
- **`expect("discv5 state must exist")` panics** — the invariant is maintained by the routing guard (`if self.discv5.is_none() { return; }`). Making every handler return early with `if let` would add noise for a situation that can't happen in normal operation. The `new_for_test` constructor always sets `discv5 = Some(...)`.

azteca1998 (approved):

crates/networking/p2p/discovery/server.rs
```diff
@@ -0,0 +1,473 @@
...
+                server.discv4_send_ping(bootnode).await?;
+            }
+        }
+
+        server.start();
+
+        Ok(())
+    }
+
+    pub fn new_for_test(
```
iovoid: This probably shouldn't be here.

crates/networking/p2p/discovery/server.rs
```diff
@@ -0,0 +1,473 @@
...
+            local_node_record,
+            signer,
+            udp_socket,
+            store: Store::new("", ethrex_storage::EngineType::InMemory)
+                .expect("Failed to create store"),
+            peer_table,
+            initial_lookup_interval: 1000.0,
+            config: DiscoveryConfig::default(),
+            discv4: None,
+            discv5: Some(Discv5State::default()),
```
iovoid: Only has discv5 but the name or the (nonexisting) documentation doesn't mention it.

crates/networking/p2p/discovery/server.rs
```diff
@@ -0,0 +1,473 @@
...
+
+pub struct DiscoveryServer {
+    pub local_node: Node,
+    pub local_node_record: NodeRecord,
+    pub(crate) signer: SecretKey,
+    pub(crate) udp_socket: Arc<UdpSocket>,
+    pub(crate) store: Store,
+    pub peer_table: PeerTable,
+    pub(crate) initial_lookup_interval: f64,
+    pub(crate) config: DiscoveryConfig,
```
iovoid: Isn't initial_lookup_interval already in the DiscoveryConfig?

ElFantasma: Fixed in 672fa03 — removed the standalone `initial_lookup_interval` field and now reads it from `self.config.initial_lookup_interval` directly.

crates/networking/p2p/discovery/server.rs
ElFantasma: Fixed in 672fa03 — moved the test helper behind `#[cfg(any(test, feature = "test-utils"))]`. The `ethrex-test` crate now enables the `test-utils` feature on its `ethrex-p2p` dependency. Not in the production build path anymore.

crates/networking/p2p/discovery/server.rs
ElFantasma: Fixed in 672fa03 — renamed to `new_for_discv5_test` and added a docstring making it explicit that only discv5 state is initialized. The config now also explicitly sets `discv4_enabled: false, discv5_enabled: true`.

iovoid (approved):

edg-l (approved):
LGTM — solid refactor, faithful preservation of packet discrimination and timer semantics. Two non-blocking items worth a follow-up:

1. **Test gap on IP voting guard.** `test_ip_voting_no_update_if_same_ip` in `test/tests/p2p/discovery/discv5_server_tests.rs:301` now calls `discv5(&mut server).record_ip_vote(...)`, which only returns an `Option<IpAddr>` and never mutates `server.local_node`. The assertion at line 313 trivially passes, so the server-level guard at `discovery/discv5_handlers.rs:379` (`winning_ip != self.local_node.ip`) is no longer exercised. Suggest adding a test that drives `server.discv5_handle_pong(...)` with a winning IP equal to the current local IP and asserts `server.local_node_record.seq` is unchanged — that's the actual guard we want to lock in.

2. **Observability regression on stale-entry cleanup.** The old `discv5` `cleanup_stale_entries` emitted a `trace!("Cleaned up {} stale entries", ...)`; the new `Discv5State::cleanup_stale_entries` at `crates/networking/p2p/discv5/server.rs:87` no longer does. Not a correctness issue but worth restoring for debuggability.

Issue #6339, opened 2026-03-07
Investigate making guest program `no_std`


kevaundray:
## Problem

Currently compilation of guest programs depend on forks of the Rust compiler maintained by zkVM teams. This is not ideal as it:

- Introduces supply chain risk
- Requires zkVM maintainers to regularly update their fork to the latest upstream Rust

## Suggested Solution

- We can use the upstream targets that are maintained by Rust such as `rv64im-unknown-none-elf` and add a CI job to ensure its always compatible. 

The main blocker for this is the guest program being `no_std` compatible. The forked Rust compilers had support for std, so we didn't have this issue.

Upon further investigation, the affected crates look like:

 - [x] ethrex-rlp (crates/common/rlp)                                                                             
 - [x] ethrex-crypto (crates/common/crypto)                                                                       
 - [ ] ethrex-common (crates/common)                                 
 - [ ] ethrex-vm (crates/vm) — which transitively pulls in:
    - [ ] ethrex-levm (crates/vm/levm)
    - [ ] ethrex-trie (crates/common/trie)

Asked AI to order them in terms of hardest to change:

1. ethrex-rlp — Almost no std dependencies. Mostly just gate thiserror/std. Minimal changes needed.          
  
2. ethrex-crypto — Small crate. Gate tracing, kzg-rs, and thiserror/std. The k256 and sha3 already support   
  no_std.
                                                                                                      
3. ethrex-trie — Gate tracing, rayon, crossbeam, serde_json, lazy_static, anyhow behind std. Replace         
  rustc_hash::FxHashMap/FxHashSet with hashbrown equivalents. Handle rkyv/bytecheck feature gating.
  
4. ethrex-common — Larger surface area. Gate serde_json, kzg-rs, rayon, tracing, url, hex-simd, once_cell,
  libc, lazy_static behind std. Replace FxHashMap with hashbrown. Gate serde derives on types containing Bytes.
   Handle k256::ecdsa::Error no_std differences.
  
5. ethrex-vm — Mostly a thin wrapper. Gate tracing, rayon, rkyv, bincode, ethereum-types behind std. Replace
  FxHashMap with hashbrown. Add hashbrown/rayon to std feature.
  
6. ethrex-levm — Hardest by far. Many precompile crypto deps (p256, ripemd, malachite, lambdaworks-math,
  bls12_381, ark-bn254, ark-ec, ark-ff) all pull std transitively and must be made optional behind std. Need a
[...]

PR #6455, opened 2026-04-07
feat(l1): make common crate no_std compatible

avilagaston9:
## Motivation

Part of #6339. Continuation of #6424 (crypto crate).

## Description

Makes the `ethrex-common` crate `no_std` compatible via a feature-gated `std` feature (default-on). The approach prioritizes shared code over feature-flag branching — most changes work in both modes without any `#[cfg]`.

**What's shared (no cfg needed):**
- All core types (Block, Transaction, Receipt, AccountState, etc.)
- RLP encoding/decoding, serde derives, constants
- `std::cmp/fmt/ops/hash/time` → `core::*` replacements
- `std::collections::BTreeMap/BTreeSet` → `alloc::collections::*`
- All 7 `LazyLock<H256>` constants → `const H256` using `hex_literal::hex!` (eliminates runtime init, strictly better perf)
- `tracing` macros kept unconditional (no-ops without subscriber in no_std)

**What's feature-gated behind `std`:**
- `fd_limit`, `genesis_utils` modules (file I/O)
- `block_execution_witness` module (tightly coupled to trie, which is not yet no_std)
- `ethrex-trie` dependency and functions using Trie (`compute_*_root`, `validate_block_body`)
- `GLOBAL_SIGNER_CACHE` (LazyLock + Mutex + LRU) — sender recovery works in both modes, just no cache in no_std
- `rayon::par_iter` (sequential fallback in no_std)
- HashMap-using code (serde_utils::hashmap, rkyv_utils::OptionStorageWrapper)
- FxHashMap-using types (Account, AccountUpdate, BlockAccessListRecorder)

**OnceCell:** Conditional type alias — `once_cell::sync::OnceCell` in std, `core::cell::OnceCell` in no_std (single-threaded, suitable for guest programs).

**Dependencies:** Pinned directly with `default-features = false` (same pattern as ethrex-rlp, since Cargo doesn't allow overriding workspace default-features). `std` feature propagates std to all deps.

**Downstream:** blockchain, vm, levm Cargo.tomls updated to explicitly request `features = ["std"]`. CI no_std check now includes ethrex-common.

## Checklist


avilagaston9:
Closing for inactivity. Tracked by #6339.

PR #6463, opened 2026-04-10
fix(l1): add reverse check in BAL withdrawal validation

edg-l:
## Summary

- `validate_bal_withdrawal_index` only checked the BAL->DB direction (every BAL claim matches actual state). It did not verify the reverse: that every account modified during the withdrawal/request phase has a corresponding BAL entry.
- A malicious block builder could omit a withdrawal recipient from the BAL. Since `bal_to_account_updates` derives the state root entirely from BAL entries, the omitted account's withdrawal balance would be excluded from the state trie. The BAL-derived state root would match the tampered header, passing all validation.
- This adds Part B (DB->BAL) to the withdrawal validation, mirroring the bidirectional check already present in `validate_tx_execution`.

## Perf notes

- Withdrawal/request phase typically touches <16 accounts, so the reverse iteration is negligible.
- Pre-state lookups first check the last BAL entry (in-memory), only falling back to `db.store` when no BAL entry exists for that field.

## Test plan

- [x] All 1963 Amsterdam EF tests pass
- [ ] EEST test for missing withdrawal account in BAL (PR pending in execution-spec-tests)


azteca1998 (approved):

crates/vm/backends/levm/mod.rs
```diff
@@ -1573,6 +1583,164 @@ impl LEVM {
...
+            }
+
+            // Storage: for each slot in the withdrawal/request-phase state,
+            // verify the BAL has a corresponding entry or the value is unchanged.
+            for (key_h256, &value) in &account.storage {
+                let slot_u256 = u256_from_big_endian_const(key_h256.0);
+                let pos = acct
+                    .storage_changes
+                    .partition_point(|sc| sc.slot < slot_u256);
+                if pos < acct.storage_changes.len() && acct.storage_changes[pos].slot == slot_u256 {
```
iovoid, line 1752: What if storage_changes is incomplete? Wouldn't this skip checking?

iovoid (changes requested):

crates/vm/backends/levm/mod.rs
```diff
@@ -1573,6 +1583,164 @@ impl LEVM {
...
+            }
+
+            // Code
+            if !has_exact_change_code(&acct.code_changes, withdrawal_idx) {
+                let seeded_hash = acct
+                    .code_changes
+                    .last()
+                    .map(|c| {
+                        if c.new_code.is_empty() {
+                            *EMPTY_KECCACK_HASH
```
ElFantasma, line 1723: When the account IS in the BAL but a specific slot is absent from `storage_changes`, the code silently skips it. This leaves a gap: a malicious builder could include the account in the BAL (passing the `addr_to_idx` check) but omit a storage slot that was actually written during the withdrawal/request phase.

The "absent from BAL entirely" path (above) correctly compares every slot against `db.store`; this path should do the same. Since the withdrawal phase touches <16 accounts (as noted in the PR description), a fallback store lookup here is cheap:

```rust
} else {
    // Slot not in BAL storage_changes — verify it wasn't written.
    let pre_value = db
        .store
        .get_storage_value(*addr, *key_h256)
        .unwrap_or_default();
    if value != pre_value {
        return Err(EvmError::Custom(format!(
            "BAL validation failed for withdrawal: account {addr:?} storage \
             slot {slot_u256} changed during withdrawal/request phase but \
             has no BAL storage_changes entry"
        )));
    }
}
```

(Note: `validate_tx_execution` has the same gap but justifies it with "can't verify cheaply" since it runs per-tx. That justification doesn't apply here.)

ElFantasma (changes requested):

crates/vm/backends/levm/mod.rs
```diff
@@ -1580,6 +1590,179 @@ impl LEVM {
...
+            }
+
+            let Some(&bal_acct_idx) = index.addr_to_idx.get(addr) else {
+                // Account modified during withdrawal/request phase but absent
+                // from BAL entirely. Compare with pre-state (store) to
+                // distinguish genuine mutations from warm-access artifacts.
+                let pre = db
+                    .store
+                    .get_account_state(*addr)
+                    .ok()
```
iovoid, line 1636: Silently ignores db errors

iovoid (changes requested):

ElFantasma (approved):

iovoid (approved):

PR #6464, opened 2026-04-10
fix(l1): adding block by block exec as a fallback

Arkenan:
We detected that we were consistently failing fullsync at hoodi block 443,055 due to batching reasons. This PR adds pipelined block by block execution as a fallback when a single batch fails. After this stability fix, hoodi full sync went past that block without issues. This should improve Full sync stability in general.


ilitteri (approved):

edg-l (approved):

azteca1998 (approved):

avilagaston9 (approved):

crates/networking/p2p/sync/full.rs
```diff
@@ -279,26 +284,81 @@ async fn add_blocks(
...
+            | InvalidBlockError::BlobGasUsedMismatch
+    )
+}
+
+async fn run_blocks_pipeline(
+    blockchain: Arc<Blockchain>,
+    blocks: Vec<Block>,
+) -> Result<(), (ChainError, Option<BatchBlockProcessingFailure>)> {
+    tokio::task::spawn_blocking(move || {
+        let mut last_valid_hash = H256::default();
```
iovoid, line 346: If the first block fails, the 'last good' value is zero which is unintuitive.

crates/networking/p2p/sync/full.rs
```diff
@@ -279,26 +284,81 @@ async fn add_blocks(
...
+            | InvalidBlockError::RequestsHashMismatch
+            | InvalidBlockError::BlockAccessListHashMismatch
+            | InvalidBlockError::BlobGasUsedMismatch
+    )
+}
+
+async fn run_blocks_pipeline(
+    blockchain: Arc<Blockchain>,
+    blocks: Vec<Block>,
+) -> Result<(), (ChainError, Option<BatchBlockProcessingFailure>)> {
```
iovoid, line 344: Should also use the cancel_token

crates/networking/p2p/sync/full.rs
```diff
@@ -279,26 +284,81 @@ async fn add_blocks(
...
+fn is_post_execution_error(err: &InvalidBlockError) -> bool {
+    matches!(
+        err,
+        InvalidBlockError::GasUsedMismatch(_, _)
+            | InvalidBlockError::StateRootMismatch
+            | InvalidBlockError::ReceiptsRootMismatch
+            | InvalidBlockError::RequestsHashMismatch
+            | InvalidBlockError::BlockAccessListHashMismatch
+            | InvalidBlockError::BlobGasUsedMismatch
+    )
```
iovoid, line 338: If the worry is invalid intermediate state, InvalidTransaction (which includes 'insufficient funds') could also be caused by bad state.

PR #6465, opened 2026-04-10
refactor(l1): improve error handling in set_sync_block with eyre::Result

emirongrr:
**Motivation**

The current `set_sync_block()` function uses identical error messages for two different failure 
scenarios (store query failure vs. hash not found), making it impossible to distinguish which 
layer failed. Using `eyre::Result` with distinct error contexts for each failure point provides 
better diagnostics and aligns with the codebase's error handling standard.

**Description**

Refactored `set_sync_block()` function in `cmd/ethrex/initializers.rs` to use `eyre::Result<()>` 
error handling instead of `.expect()` calls:

- Changed function from `async fn()` to `async fn() -> eyre::Result<()>`
- Replaced `.expect()` with explicit error handling using `wrap_err()` and `ok_or_else()`
- Added distinct error messages for each failure point:
- Updated `init_l1()` to propagate errors with `?` operator

## Note

If this refactoring is considered excessive or unnecessary, feel free to reject or close this PR. 
A simpler alternative would be to just fix the duplicate error messages:

```rust
.expect("Failed to query block hash from store")
.expect("Block hash not found for block number {block_number}");
```
**Checklist**


cmd/ethrex/initializers.rs
```diff
@@ -461,21 +461,28 @@ pub fn get_ws_socket_addr(opts: &Options) -> SocketAddr {
...
 #[cfg(feature = "sync-test")]
-async fn set_sync_block(store: &Store) {
-    if let Ok(block_number) = env::var("SYNC_BLOCK_NUM") {
-        let block_number = block_number
+async fn set_sync_block(store: &Store) -> eyre::Result<()> {
+    if let Ok(block_number_str) = env::var("SYNC_BLOCK_NUM") {
+        let block_number: u64 = block_number_str
             .parse()
-            .expect("Block number provided by environment is not numeric");
+            .map_err(|_| eyre::eyre!("Block number provided by environment is not numeric"))?;
```
ElFantasma: nit: Since `StoreError` implements `std::error::Error`, you can use eyre's `WrapErr` trait here (and on the `forkchoice_update` call below) instead of `map_err`. This preserves the full error chain and is more idiomatic with eyre:

```rust
use eyre::WrapErr;

let block_hash = store
    .get_canonical_block_hash(block_number)
    .await
    .wrap_err("Failed to query block hash from store")?
    .ok_or_else(|| {
        eyre::eyre!("Block hash not found for block number {block_number}")
    })?;

store
    .forkchoice_update(vec![], block_number, block_hash, None, None)
    .await
    .wrap_err("Could not set sync block")?;
```

`ok_or_else` still needs `eyre::eyre!()` since `Option` has no error to wrap — that part is fine as-is.

ElFantasma (approved):

cmd/ethrex/initializers.rs
emirongrr: Updated to use `wrap_err` so we preserve the full error chain as suggested. Thanks!

PR #6466, opened 2026-04-11
feat(l1): fullsync metrics

Arkenan:
Add metrics for full sync so that they can be seen in a grafana dashboard.


crates/blockchain/blockchain.rs
```diff
@@ -2188,7 +2190,8 @@ impl Blockchain {
         let mut total_gas_used = 0;
         let mut transactions_count = 0;
 
-        let interval = Instant::now();
+        let interval = std::time::Instant::now();
```
ElFantasma, line 2193: These four timing `Instant`s and the `[FULLSYNC TIMING]` log run regardless of the `metrics` feature. Cost is small but inconsistent with the rest of the metrics infra — wrap with `#[cfg(feature = "metrics")]` to match. The `info!` line below should also be feature-gated.

crates/blockchain/blockchain.rs
```diff
@@ -2270,6 +2280,17 @@ impl Blockchain {
         self.storage
             .store_block_updates(update_batch)
             .map_err(|e| (e.into(), None))?;
+        let store_ms = store_start.elapsed().as_millis();
+
+        info!(
+            "[FULLSYNC TIMING] Batch {}: exec={}ms, transitions={}ms, merkle={}ms, store={}ms, total={}ms",
+            blocks_len,
+            exec_ms,
+            transitions_ms,
```
ElFantasma, line 2289: `transitions_ms` is logged here but never exposed as a Prometheus gauge — Grafana can't graph it. Add `set_batch_transitions_ms` on `MetricsFullSync` (mirroring `batch_execution_ms` etc.) and call it in the metrics block below. (Greptile flagged.)

crates/blockchain/metrics/fullsync.rs
```diff
@@ -0,0 +1,320 @@
...
+    batch_total_ms: Gauge,
+    batch_size: IntGauge,
+
+    // Timestamps (Unix epoch seconds, for Grafana elapsed time calculation)
+    header_stage_start_timestamp: Gauge,
+    execution_stage_start_timestamp: Gauge,
+
+    // Reliability
+    header_failures: IntCounter,
+    body_failures: IntCounter,
```
ElFantasma, line 38: **P1**: `inc_body_failures()` is exposed and surfaced on the dashboard but never *called* anywhere in this PR — any panel reading this metric will permanently show zero, masking real body-fetch failures. Either wire it at the body-download error site in `crates/networking/p2p/sync/full.rs` (alongside `inc_header_failures()`), or drop the metric until both increment and dashboard panel land together. (Greptile flagged.)

crates/blockchain/metrics/fullsync.rs
```diff
@@ -0,0 +1,320 @@
...
+            .expect("Failed to create fullsync_cycles_completed metric"),
+        }
+    }
+
+    // Phase & progress setters
+    pub fn set_stage(&self, stage: i64) {
+        self.stage.set(stage);
+    }
+    pub fn set_target_block(&self, block: u64) {
+        self.target_block.set(block.cast_signed());
```
ElFantasma, line 178: `block.cast_signed()` is `u64 as i64` — fine for any realistic block number, but if `u64::MAX` is ever passed as a sentinel "unknown target", this silently reports `-1` and Grafana queries break. `i64::try_from(block).unwrap_or(0)` would be more defensive at zero perf cost. Same pattern in the next three setters.

crates/blockchain/metrics/fullsync.rs
```diff
@@ -0,0 +1,320 @@
...
+    pub fn set_execution_stage_start_now(&self) {
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs_f64();
+        self.execution_stage_start_timestamp.set(now);
+    }
+
+    /// Reset gauges at the start of a new sync cycle so stale data doesn't persist
+    pub fn reset_cycle(&self) {
```
ElFantasma, line 244: Worth a doc comment that this resets gauges only — the `IntCounter`s (`headers_downloaded`, `bodies_downloaded`, `*_failures`, `cycles_*`) intentionally stay monotonic for `rate()` queries. Right now the asymmetry looks like an oversight.

ElFantasma (changes requested):
Historically we approached metrics modules in a manual way, but we have done some work to fix this. Now there's a split in how Prometheus is wired:

- **`register_*` macro / default-registry pattern (2 modules)**: `rpc.rs` (introduced in #5335 by @rodrigo-o, 2025-11-24) and `sync.rs` (#6470, 2026-04-20). Each metric is created with `register_int_gauge!` / `register_counter_vec!` / etc., which registers it into Prometheus's default registry at construction time. `api.rs` surfaces them automatically via `gather_default_metrics()`.

- **Manual `Registry::new()` pattern (6 modules)**: `blocks.rs`, `node.rs`, `p2p.rs`, `process.rs`, `transactions.rs`, `l2/metrics.rs`. Each module owns a singleton struct of bare `IntGauge` / `IntCounter` collectors, exposes a `gather_metrics()` method that builds a fresh `Registry::new()` per call, and `api.rs` has to know about each module and concatenate their output.

We should start moving all the metrics to the first method, but that's tech debt to be addressed in some future effort. 

As this PR is adding a new module from scratch, it's a free moment to land it in the better shape rather than join the manual-Registry side and grow the divergence. Concrete deltas vs the current `fullsync.rs`:

1. **Use the `register_*` macros** (`register_int_gauge!`, `register_int_counter!`, `register_int_counter_vec!`) instead of `IntGauge::new()` / `IntCounter::new()`. The macros register into the default registry at construction time. New metrics auto-surface via `gather_default_metrics()` with zero `api.rs` changes — the current PR builds a fresh `Registry::new()` on every `gather_metrics()` call and forces `api.rs` to know about this module, both of which the macro pattern avoids.

2. **Collapse parallel counters into label vectors** (`IntCounterVec` with `&["component"]` or `&["outcome"]`). #6470 does this for `pivot_updates`, `storage_requests`, `header_resolution`. Here, `header_failures` + `body_failures` are a natural pair that should be one `fullsync_failures_total{component="header|body"}` — Grafana queries become `sum(rate(...)[5m]) by (component)` instead of two separate series.

3. **Naming convention.** Every other metric in this codebase is `ethrex_*` (`ethrex_blocks_*`, `ethrex_p2p_*`, `ethrex_sync_*`). `fullsync_*` will look orphaned in Grafana's metric picker — rename to `ethrex_fullsync_*`.

4. **Drop `reset_cycle()`.** It exists because some of the "counters" here (`cycles_started`, `cycles_completed`) are gauges-pretending-to-be-counters. With proper `IntCounter` registration via the macros, monotonicity is enforced and reset becomes meaningless — and counter monotonicity is what `rate()` queries assume.

5. **Drop `api.rs`'s explicit `gather_metrics()` call.** Once metrics register into the default registry via macros, `api.rs` doesn't need to know this module exists. See the comment in `api.rs:57-58` from #6470: *"METRICS_SYNC registers into the default Prometheus registry at init, so its metrics are already included in `gather_default_metrics()` above."*

---

The 5 inline comments above stand independently of this — they're tactical fixes that apply either way.

PR #6468, opened 2026-04-13
feat(l1): snap sync metrics and dashboard

Arkenan:

PR #6470, opened 2026-04-13
feat(l1): add snap sync observability endpoints and tooling

ElFantasma:
**Motivation**

Recent mainnet-9 multisync runs surfaced intermittent snap sync failures that are hard to diagnose from standard logs — in particular, pivot-update failures fired during the healing transition where peer-selection bottlenecks lead to `process::exit(2)`. To investigate and eventually prevent these, we need tools to (a) inspect live sync state from outside the node, (b) capture detailed peer diagnostics when something degrades, and (c) post-mortem a failure with full peer-table context.

**Description**

End-to-end observability for snap sync, across five layers.

*Node — sync metrics (new `crates/blockchain/metrics/sync.rs`):*
- Per-phase progress gauges: headers, accounts, storage, healing, bytecodes (downloaded/inserted/total)
- Peer health: eligible peers, snap peers, inflight requests
- Pivot tracking: block number, timestamp, age
- Phase timing: per-phase start timestamps for elapsed calculation
- Outcome counters: pivot updates, storage requests, header resolution (by outcome label)
- Registered into the default Prometheus registry at init — no per-request overhead

*Node — RPC endpoints:*
- `admin_syncStatus`: current phase, pivot block, staleness info, phase progress. Resets to `idle` on both success and error paths.
- `admin_peerScores`: full peer table with scores, capabilities, supported block ranges, and per-capability eligibility. Computed live per query (not read from a possibly-stale snapshot)
- Additional header-download-phase diagnostics in `snap_sync.rs`

*Grafana dashboard (`metrics/provisioning/grafana/dashboards/common_dashboards/snapsync_dashboard.json`):*

image
- [Live dashboard on ethrex-grafana](https://grafana.ethrex.xyz/d/ethrex-snapsync/ethrex-snap-sync)
- 7 sections: Sync Overview, Peer Health, Headers, Accounts, Storage, Healing, Bytecodes
- Each phase section: progress gauge with downloaded/total counts, rate gauge, ETA, elapsed time, full-width rate-over-time timeseries
- Peer Health row: eligible peers, pivot age (live from timestamp), inflight requests, pivot update count
- All rates computed in Grafana via `rate(metric[5m])` — no in-app rate computation
- ETAs computed as `remaining / rate` with division-by-zero guards
- Multi-instance support via `$instance` variable (mainnet:3701, hoodi:3702, sepolia:3703)
- Docker compose updated to expose metrics ports and enable `--metrics` on all containers

*Monitor (`tooling/sync/docker_monitor.py`):*
- Rolling snapshot buffer per instance, dumped to run directory on demand
- Degradation detection: low eligible peers, high staleness ratio
- Configurable watched phases (`--watched-phases`): sync phases that trigger TRACE logging and fast polling (5s). Empty by default — opt-in, not opt-out. Use `MULTISYNC_WATCHED_PHASES=healing` in the Makefile or `--watched-phases "healing"` directly.
- On degradation: poll drops to 5s, log level bumps to TRACE via `admin_setLogLevel`, snapshots dumped to disk; restored on recovery
- Force-dump on any failure for post-mortem (with final RPC poll to capture peer state at failure time)

*Live TUI (`tooling/sync/peer_top.py`):*

image
- Top-style live view of sync status + peer table sorted by score
- Responsive layout — adapts to terminal width, trims client names when narrow
- Score delta arrows (green ↑ / red ↓) to spot scoring changes at a glance
- Polls `admin_syncStatus` and `admin_peerScores`

*REPL (`tooling/repl/src/formatter.rs` + admin commands):*
- Table rendering for arrays of objects (needed to read the peer table returned by `admin_peerScores`)
- Admin commands wired to the new endpoints

**Note:** This PR incorporates the snap sync progress metrics and Grafana dashboard from #6468 (Tomi/Esteve), with adaptations: metrics use the default Prometheus registry, throughput rates are computed in Grafana via rate() instead of in-app, and a new peer-health row was added. I closed #6468, as it was superseded.

**Checklist**

- [x] Tested with `make multisync-loop-auto` on ethrex-mainnet-9 (mainnet, sepolia, hoodi in parallel)
- [x] AI agent review feedback addressed (see comment)
- [x] Grafana dashboard tested with live data from 3 networks


ElFantasma:
## Review feedback addressed

Thanks for the thorough reviews. Here's how each finding was handled.

### Fixed

| Finding | Raised by | Fix |
|---------|-----------|-----|
| Clippy: redundant closure in `formatter.rs:187` | CI lint | `.map(\|v\| inline_value(v))` → `.map(inline_value)` |
| Clippy: unnecessary `u64` casts in `sync_manager.rs` | CI lint | Removed 3 redundant `as u64` casts |
| `peer_top.sh` broken on Python < 3.12 (nested f-string quotes) | Claude, Codex | Removed `peer_top.sh` entirely — `peer_top.py` is the canonical tool |
| `MetricsSync` creates new `Registry` every `gather_metrics()` call | Kimi | Migrated to default Prometheus registry: metrics register once at init via `register_int_gauge!` / `register_int_counter_vec!`, exported via `gather_default_metrics()`. Removed `gather_metrics()` entirely. The same per-call pattern exists in 6 other modules — migrating those is out of scope but this sets the example. |
| `admin_syncStatus` reports stale phase after recoverable sync error | Codex | Reset `current_phase = "idle"` on the error path in `sync.rs` |
| `on_failure()` never executes the "final poll" | Codex | Added `force: bool` param to `poll()`. `on_failure()` now calls `poll(inst, force=True)` to bypass the status check |
| Dead `DEGRADATION_STALL_TIMEOUT` constant | Claude | Removed constant and stale comment reference |
| Healing phase unconditionally triggers degradation (TRACE + fast polling for hours) | Claude | Made configurable: `WATCHED_PHASES` set (empty by default — opt-in, not opt-out). Override via `--watched-phases "healing"` CLI flag or `MULTISYNC_WATCHED_PHASES=healing` Makefile variable. No TRACE logging unless explicitly requested. |
| Progress stall detection is dead code | Claude | Removed the dead constant and updated comment to `# stall detection not yet implemented` |

### Not addressing (with reasoning)

1. **`read_peer_diagnostics()` in `update_pivot` loop adds actor roundtrip overhead** (Greptile) — This is in a debug logging path that only fires during pivot updates (rare, seconds-scale operations). The overhead of one actor call is negligible relative to the 15s peer timeout per attempt.

2. **Duplicate `phase_progress` keys (`accounts_downloaded` vs `account_ranges_downloaded`)** (Greptile) — These are intentionally distinct: `account_ranges_downloaded` counts accounts received from peers (network phase), while `accounts_downloaded` counts accounts committed to temp storage (insertion phase). They track different stages of the pipeline.

3. **`peer_top.py` uncaught `KeyError` on malformed response** (Greptile) — This is a dev/debugging tool, not a production service. Crashing on malformed data is the correct behavior — it surfaces issues rather than hiding them.

4. **Lock poisoning / integer casting / feature flag alignment** (Kimi) — `tokio::sync::RwLock` doesn't poison. `usize` → `i64` for peer counts won't overflow. Feature flag follows the existing pattern in `mod.rs`.

5. **Python type hints use `dict[str, ...]` / `list[dict]` (Python 3.9+)** (Kimi) — Our Docker images and development environments use Python 3.10+. Python 3.8 is EOL. Acceptable minimum.

crates/networking/p2p/sync.rs
```diff
@@ -106,7 +160,24 @@ impl Syncer {
...
+                    recoverable = recoverable,
+                    action = if recoverable { "retry" } else { "exit" },
+                    "Sync cycle error classification"
+                );
+                self.diagnostics.write().await.push_error(SyncErrorEvent {
+                    timestamp: std::time::SystemTime::now()
+                        .duration_since(std::time::UNIX_EPOCH)
+                        .unwrap_or_default()
+                        .as_secs(),
+                    error_type: format!("{:?}", std::mem::discriminant(&error)),
```
iovoid, line 176: Not sure if something like 'Discriminant(7)' is that useful

iovoid (approved):

tooling/repl/src/formatter.rs
```diff
@@ -136,6 +156,84 @@ fn flatten_object(map: &serde_json::Map<String, Value>, prefix: &str) -> Vec<(St
...
+    out.push_str(&format!("\n      {}", sep_parts.join("──")));
+
+    // Rows
+    for row in &col_values {
+        let parts: Vec<String> = row
+            .iter()
+            .zip(&col_widths)
+            .map(|(val, w)| {
+                let truncated = if val.len() > *w {
+                    format!("{}…", &val[..*w - 1])
```
avilagaston9, line 223: `val[..*w - 1]` slices by byte index, not character index. If `val` contains multi-byte UTF-8 (peer client versions sometimes include emoji or non-ASCII), this panics with `byte index is not a char boundary`.

Suggestion:
```rust
let truncated = if val.len() > *w {
    let end = val.char_indices()
        .take(*w - 1)
        .last()
        .map(|(i, c)| i + c.len_utf8())
        .unwrap_or(0);
    format!("{}…", &val[..end])
```
Or simpler: `val.chars().take(w - 1).collect::<String>()` (allocates, but this isn't a hot path).

crates/networking/p2p/sync.rs
```diff
@@ -106,7 +160,24 @@ impl Syncer {
 
             // If the error is irrecoverable, we exit ethrex
             Err(error) => {
-                match error.is_recoverable() {
+                let recoverable = error.is_recoverable();
+                self.diagnostics.write().await.current_phase = "idle".to_string();
```
avilagaston9, line 164: Two separate write-lock acquisitions here and on the `.push_error()` call below. If diagnostics are read between the two locks (e.g. by the RPC handler), the phase shows `idle` but the error event is missing. Consider combining into a single lock:
```rust
{
    let mut diag = self.diagnostics.write().await;
    diag.current_phase = "idle".to_string();
    diag.push_error(SyncErrorEvent { ... });
}
```
The `debug!` macro between them doesn't need the lock.

avilagaston9 (commented):
Solid observability PR — well-structured across all 5 layers. Left a few inline comments on things I noticed.

ilitteri (approved):

avilagaston9 (approved):

PR #6471, opened 2026-04-14
refactor(l1): shared trie abstraction

edg-l:
## Summary

Introduces a trie-agnostic abstraction layer so ethrex can support both MPT and a future binary trie (EIP-7864) without backend-specific types leaking across crate boundaries. MPT is the only backend in this PR; `BackendKind` / `StateBackend` / `Merkleizer` / `NodeUpdates` are single-arm enums ready for a `Binary` arm in PR 2.

## Docs

- [Spec](https://github.com/lambdaclass/ethrex/blob/shared-trie/docs/shared-trie/spec.md)
- [Adding a new backend guide](https://github.com/lambdaclass/ethrex/blob/shared-trie/docs/shared-trie/adding-a-backend.md)

## Crate layout after the refactor

| Crate | Responsibility |
|---|---|
| `ethrex-state-backend` (new) | `StateReader` / `StateCommitter` traits, `BackendKind`, `CodeReader`, shared types (`AccountMut`, `MerkleOutput`, `NodeUpdates`, `StateError`) |
| `ethrex-trie` | MPT is the implementation: `MptBackend`, `MptMerkleizer`, `TrieProvider` trait, `Trie`, node types, `compute_*_root`, `validate_block_body`, `genesis_block` / `genesis_root` |
| `ethrex-storage` | `StateBackend` enum, `Merkleizer` enum, `Store`, layer cache, FKV generator, MPT wiring (`mpt_wiring.rs`) |
| `ethrex-common` | Data types only; no trie dependency. Hosts `ExecutionWitness` + `RpcExecutionWitness` + conversions |
| `ethrex-vm` | `GuestProgramState` (zkVM execution state), `Evm`, `VmDatabase` |

## Key architectural changes

- **Dep-cycle broken**: `ethrex-common` no longer depends on `ethrex-trie`. `compute_*_root`, `validate_block_body`, `genesis_block` moved to `ethrex-trie`.
- **Enum dispatch, not `dyn`**: `StateBackend` and `Merkleizer` are enums in `ethrex-storage`. No vtable on the production path.
- **Merkleization extracted**: 16-shard `handle_merkleization` + `handle_subtrie` moved from `blockchain.rs` into `MptMerkleizer` in `ethrex-trie/merkleizer.rs`. `blockchain.rs` has zero MPT imports.
- **Pipelined path** uses `Merkleizer::new_mpt` + `feed_updates` / `finalize`; non-pipelined path uses `StateCommitter::commit`.
- **Read path** goes through `StateReader` trait via `Store::new_state_reader`; `vm.rs` has zero `StateBackend::Mpt` downcasts.
- **`TrieProvider` trait** is the dep-inversion seam: `ethrex-trie` cannot name `Store`, so `ethrex-storage` provides `StoreTrieProvider` implementing the trait.
- **`ExecutionWitness` is backend-agnostic**: `state_proof: Vec<Vec<u8>>` replaces `state_trie_root: Option<Node>` + `storage_trie_roots: BTreeMap<H256, Node>`. Trie reconstruction lives in `MptBackend::from_witness_bytes`.
- **`GuestProgramState` moved to `ethrex-vm`**: it is a zkVM execution concept, not a trie concept. Holds an `MptBackend` rather than raw `Trie` fields.
- **`VmDatabase` / LEVM `Database`** return `AccountStateInfo` (no `storage_root` leak) instead of `AccountState`.
- **`AccountProof`** uses `AccountInfo` + `storage_root` instead of `AccountState`.
- **SLOAD cache** internalized in `MptBackend` via `Mutex<FxHashMap>`; lock poisoning is surfaced as `StateError::Storage`.
- **Format marker**: `STATE_BACKEND_FORMAT_KEY` in `MISC_VALUES` prevents opening a DB with a backend that doesn't match its on-disk format.
- **Genesis entry point**: `StateBackend::compute_genesis_root(kind, &genesis)` / `compute_genesis_block(kind, &genesis)` for CLI / deployer / banner callers with no `Store`.

## Adding a new backend requires

1. Implement `StateReader` / `StateCommitter` for the new backend type.
2. Add enum arms to `StateBackend`, `Merkleizer`, `NodeUpdates`, and `BackendKind`.
3. Add a parallel `*_wiring.rs` in `ethrex-storage` and a `TrieProvider`-equivalent.
4. Zero changes required in `blockchain.rs`, `vm.rs`, `payload.rs`, `store.rs`.

## Testing

- `make -C tooling/ef_tests/blockchain test`: **8675/8675** stateful, **8675/8675** stateless.
- `cargo test --workspace` green on both main and tooling workspaces.
- `cargo clippy -p ethrex-trie -p ethrex-storage -p ethrex-state-backend --lib` clean.

## Checklist

- [x] Break `ethrex-common` -> `ethrex-trie` dependency
- [x] Create `ethrex-state-backend` crate with `StateReader` / `StateCommitter` / `BackendKind`
- [x] Implement `MptBackend` + `MptMerkleizer`
- [x] Create `StateBackend` / `Merkleizer` enums
- [x] Migrate pipelined merkleization path
- [x] Route read path through `StateBackend`
- [x] Purge all MPT code from `store.rs`
- [x] Wire genesis through `StateCommitter`
- [x] All EF blockchain tests pass (stateful + stateless)
- [x] Remove MPT downcasts from `vm.rs`
- [x] `VmDatabase` / LEVM `Database` traits return `AccountStateInfo`
- [x] Extract witness generation from `blockchain.rs` into `StateBackend`
- [x] Make `ExecutionWitness` backend-agnostic (state_proof bytes)
- [x] Move `GuestProgramState` to `ethrex-vm`
- [x] Keep `ExecutionWitness` + conversions in `ethrex-common`
- [x] Format marker guard (`STATE_BACKEND_FORMAT_KEY`)
- [x] `TrieProvider` trait unifying state + storage trie opening
- [x] Unified `Merkleizer::new_mpt` / `new_bal_mpt` API
- [x] Backend guide (`adding-a-backend.md`)
- [ ] Benchmark regression <= 2% (measure before merge)

## Follow-up (PR 2+)

- Binary trie backend (`BinaryBackend`, `BinaryMerkleizer`, `binary_wiring.rs`).
- Backend-specific FKV generators (each backend provides its own; documented).
- Backend-specific snap sync (MPT snap sync stays in `mpt_wiring.rs`; binary sync added alongside).
- zkVM compilation verification (sp1, risc0 targets) with the moved `GuestProgramState`.


crates/common/types/block_execution_witness.rs
```diff
@@ -1,93 +1,38 @@
...
     pub chain_config: ChainConfig,
-    /// Root node embedded with the rest of the trie's nodes
-    pub state_trie_root: Option<Node>,
-    /// Root nodes per account storage embedded with the rest of the trie's nodes,
-    /// keyed by the keccak256 hash of the account address.
-    #[rkyv(with = MapKV<H256Wrapper, Identity>)]
-    pub storage_trie_roots: BTreeMap<H256, Node>,
+    /// Serialized trie proof data (RLP-encoded nodes for MPT, backend-specific for others).
+    #[rkyv(with = VecVecWrapper)]
+    pub state_proof: Vec<Vec<u8>>,
```
ElFantasma, line 29: **Breaking-change surface worth calling out explicitly in the description**:

1. **`ExecutionWitness` rkyv layout flips entirely**: `Option<Node>` + `BTreeMap<H256, Node>` → `Vec<Vec<u8>>`. Combined with the move to `serde_json::to_vec(&rpc_witness)` for `EXECUTION_WITNESSES` storage, any rkyv-cached witness from an older binary becomes unreadable.

2. **`STORE_SCHEMA_VERSION` not bumped**: the convention to skip the bump is for PRs that *don't* change the Store. This PR materially changes the on-disk format of `EXECUTION_WITNESSES` and adds a new `MISC_VALUES` key (`STATE_BACKEND_FORMAT_KEY`). Without the bump, an existing v1 DB without a marker gets stamped with `Mpt` regardless of what's actually in its trie tables. Bump and migrate (or drop) `EXECUTION_WITNESSES` rows on first open.

crates/common/state-backend/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+pub enum BackendKind {
+    Mpt,
+}
+
+/// Closure type for reading code by hash from the database.
+///
+/// Lives here (not in `ethrex-trie`) so it can be used across crates without
+/// the `ethrex-trie` dependency. The error type is `StateError` to match the
+/// backend trait surface.
+pub type CodeReader = Arc<dyn Fn(H256) -> Result<Option<Vec<u8>>, StateError> + Send + Sync>;
```
ElFantasma, line 39: `CodeReader = Arc<dyn Fn(...)>` reintroduces virtual dispatch on the code-read path despite the design goal of enum dispatch on prod paths. Code reads are likely cold enough not to matter, but the inconsistency deserves a comment explaining why `Fn` was chosen here over an enum (probably: the closure captures backend-specific state).

crates/common/state-backend/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct AccountMut {
+    pub account: Option<AccountInfo>,
+    pub code: Option<CodeMut>,
+    /// Current total code size. MPT ignores it; other backends may encode it
+    /// into their leaf layout. Populated by callers so the interface is stable
+    /// across backends.
+    pub code_size: usize,
```
ElFantasma: "MPT ignores it; other backends may encode it into their leaf layout." Means MPT-only callers must set `code_size` correctly or risk silent data loss when the binary backend lands. For MPT today there's no enforcement. Worth either a `#[must_use]` or an explicit constructor that takes `code_size` as a non-defaulted parameter, so binary-backend integration time isn't a quiet bug-fix tour.

crates/common/trie/backend.rs
```diff
@@ -0,0 +1,1033 @@
...
+    /// Code cache keyed by code_hash. Checked before `code_reader`.
+    pub(crate) codes: BTreeMap<H256, Code>,
+    pub(crate) crypto: Arc<dyn Crypto>,
+    /// Opens storage tries for accounts not yet in `storage_tries`.
+    storage_opener: Arc<dyn TrieProvider>,
+    /// Reads code by hash when not in the `codes` cache.
+    code_reader: CodeReader,
+    /// Cached storage roots to avoid re-reading accounts on repeated SLOAD.
+    /// Maps hashed_address -> storage_root.
+    storage_root_cache: Mutex<FxHashMap<H256, H256>>,
```
ElFantasma: **`Mutex` on the read path** — fine for sequential per-block execution, but if `StateReader::storage` is ever called in parallel (BAL execution, parallel tx pre-validation) this becomes a contention point. PR description mentions BAL — confirm the BAL path doesn't hit this Mutex from multiple worker threads. Worth either a comment documenting the assumed access pattern, or a switch to `RwLock` / per-thread caches when parallel reads land.

crates/common/trie/merkleizer.rs
```diff
@@ -0,0 +1,1233 @@
...
+    /// Open the storage trie for `account_hash` rooted at `storage_root`.
+    /// Called lazily by workers when they hit a storage-modifying account.
+    fn open_storage_trie(&self, account_hash: H256, storage_root: H256) -> Result<Trie, TrieError>;
+}
+
+// ---------------------------------------------------------------------------
+// Background drop thread (avoids recursive deallocation on hot path)
+// ---------------------------------------------------------------------------
+
+static DROP_SENDER: LazyLock<Sender<Box<dyn Send>>> = LazyLock::new(|| {
```
ElFantasma, line 86: Process-lifetime thread that consumes boxed values to dodge recursive deallocation on the hot path. Clever, but it never shuts down and isn't observable in any debug output. Worth a comment explaining why this trade is worth it (presumably profiling showed deep-trie drops were measurable) — otherwise this looks like a leak to anyone reviewing it cold.

ElFantasma (commented):
Architectural review of a 12.5K-line refactor across 128 files — focused on design soundness, dep-graph health, and breaking-change surface rather than line-by-line.

**Architectural shape: this is the right design.** `ethrex-state-backend` hosts the abstract traits, `ethrex-trie` hosts the MPT impl, `ethrex-storage` hosts the dispatching enum. `ethrex-common` no longer pulls `ethrex-trie` — the dep cycle that motivated this PR is genuinely broken. Single-arm-enum + enum dispatch keeps the prod path vtable-free; the `TrieProvider` dep-inversion seam is exactly the abstraction needed.

**Process suggestion**: candidate split points if it's still feasible to break this up — (a) `ethrex-common` → `ethrex-trie` dep break + new `ethrex-state-backend` crate, (b) merkleizer extraction, (c) witness format change, (d) `GuestProgramState` move. Each independently verifiable. If the team chooses to land it as one bundle, the BAL Mutex review and benchmark gate below are the safety net.

**Tests + benchmark gate**: 8675/8675 stateful + stateless EF tests passing is the right behavior-preservation evidence. The checklist still lists `Benchmark regression <= 2%` as `[ ]` — don't merge without that measurement. A 12.5K-line refactor to the merkleization hot path has plenty of room to silently regress block-build wall-clock by 5–10%; `make multisync-loop-auto` is the right comparison point.

**Negative format-marker test**: tests cover first-open-writes-marker and reopen-success. What's missing is opening with `Mpt`, closing, attempting to open with a hypothetical `Binary`, expecting `StoreError::Custom("state backend format mismatch...")`. Currently un-testable since `Binary` doesn't exist; worth adding the moment PR 2 lands.

**Workspace grep** for direct uses of removed `state_trie_root` / `storage_trie_roots` fields — public API change, easy to miss external callers (test harnesses, scripts).

PR #6472, opened 2026-04-14
chore(l1): move all benchmarks to benches/ folder

azteca1998:
## Summary

Consolidates all benchmarks into the `benches/` folder for easier maintenance and cleaner crate structure.

## Changes

- **Moved benchmark files:**
  - `crates/common/rlp/benches/decode.rs` → `benches/benches/rlp_decode.rs`
  - `crates/common/rlp/benches/encode.rs` → `benches/benches/rlp_encode.rs`
  - `crates/networking/p2p/benches/decode.rs` → `benches/benches/p2p_decode.rs`

- **Updated `benches/Cargo.toml`:**
  - Added dependencies: `ethrex-rlp`, `ethrex-trie`, `ethrex-p2p`, `ethereum-types`, `rand = "0.9.2"`, `once_cell = "1"`
  - Updated `criterion` to version `0.7.0` (to match RLP benchmarks)
  - Added benchmark targets for `rlp_decode`, `rlp_encode`, and `p2p_decode`

- **Cleaned up crate Cargo.tomls:**
  - Removed benchmark-related dev-dependencies from `crates/common/rlp/Cargo.toml`
  - Removed benchmark entry from `crates/networking/p2p/Cargo.toml`
  - Deleted old benchmark directories

## Benefits

- All benchmarks in one location for easier discovery and maintenance
- Reduces clutter in individual crate directories
- Consistent with the existing `benches/` structure

## Test Plan

- [x] `cargo check -p ethrex-benches --benches` compiles cleanly
- [ ] CI passes

## Notes

The `p2p_decode.rs` file has deprecation warnings due to using old rand 0.8 API. This is expected and doesn't affect functionality. We can update it later to use the newer rand 0.9 API for consistency.


iovoid (approved):

benches/Cargo.toml
```diff
@@ -12,13 +12,19 @@ ethrex-blockchain.workspace = true
...
 ethrex-storage.workspace = true
+ethrex-rlp.workspace = true
+ethrex-trie.workspace = true
+ethrex-p2p.workspace = true
 
 bytes.workspace = true
 tokio.workspace = true
 serde_json.workspace = true
+ethereum-types.workspace = true
+rand = "0.9.2"
```
ElFantasma, line 26: (Greptile P2) `rand` and `once_cell` are bench-only and conventionally go in `[dev-dependencies]`. The existing file is already inconsistent (`bytes`/`tokio`/`serde_json` in `[dependencies]`, `criterion` in `[dev-dependencies]`) so this PR perpetuates rather than introduces the issue, but worth reconciling — if there's no library/bin output, everything except criterion belongs in dev-deps.

benches/Cargo.toml
```diff
@@ -12,13 +12,19 @@ ethrex-blockchain.workspace = true
...
+once_cell = "1"
 
 [dev-dependencies]
-criterion = { version = "0.5.1", features = [
+criterion = { version = "0.7.0", features = [
```
ElFantasma, line 28: Two-major bump (0.5 → 0.7). `cargo check --benches` (the test plan) catches compile breakage but not runtime panics from criterion API drift (`BenchmarkGroup::sample_size`, `Criterion::measurement_time`, async helpers). Worth running `cargo bench --bench rlp_decode --bench rlp_encode --bench p2p_decode` once before merge.

benches/Cargo.toml
```diff
@@ -31,5 +37,17 @@ ethrex-l2-rpc.workspace = true
...
 
+[[bench]]
+name = "rlp_decode"
+harness = false
+
+[[bench]]
+name = "rlp_encode"
+harness = false
+
+[[bench]]
```
ElFantasma, line 46: Make sure these benches actually run in some CI workflow — easy for a `[[bench]]` entry to compile but never get invoked, then silently break the next time someone touches the underlying code. If no workflow targets them today, follow-up to wire one in.

ElFantasma (approved):
Left some nits

PR #6486, opened 2026-04-15
fix(l1): defer Amsterdam block gas overflow check to post-execution

azteca1998:
## Summary
- The Hive consume-engine Amsterdam tests for EIP-7778 and EIP-8037 were failing because ethrex's per-tx gas limit checks were incompatible with Amsterdam's new gas accounting rules.
- **EIP-7778** uses pre-refund gas for block accounting, so cumulative pre-refund gas can exceed the block gas limit even when a block builder correctly included all transactions.
- **EIP-8037** introduces 2D gas accounting (`block_gas = max(regular, state)`), meaning cumulative total gas (regular + state) can legally exceed the block gas limit.
- The fix skips the per-tx cumulative gas check for Amsterdam and adds a **post-execution** block-level overflow check using `max(sum_regular, sum_state)` in all three execution paths (sequential, pipeline, parallel).

## Local test results
- **200/201** EIP-7778 + EIP-8037 Hive consume-engine tests pass
- **105/105** EIP-7778 + EIP-8037 EF blockchain tests pass (4 + 101)
- The single remaining Hive failure (`test_block_regular_gas_limit[exceed=True]`) expects `TransactionException.GAS_ALLOWANCE_EXCEEDED` but we return `BlockException.GAS_USED_OVERFLOW` — the block is correctly rejected, just with a different error classification.

## Test plan
- [x] All EIP-7778 EF blockchain tests pass locally
- [x] All EIP-8037 EF blockchain tests pass locally
- [x] 200/201 Hive consume-engine Amsterdam tests pass locally
- [ ] Full CI Amsterdam Hive suite passes


edg-l (commented):
**Bug: payload building path missing state gas overflow check**

The validation path (this PR) now correctly checks `max(block_regular, block_state) <= gas_limit` post-execution. But `crates/blockchain/payload.rs::apply_plain_transaction` still only tracks `remaining_gas` against regular gas:

```rust
// payload.rs:858-863
if context.is_amsterdam {
    context.remaining_gas = context
        .payload
        .header
        .gas_limit
        .saturating_sub(context.block_regular_gas_used);  // <-- only regular, ignores state
}
```

If state gas becomes the bottleneck (i.e. `block_state_gas_used > block_regular_gas_used > gas_limit`), the payload builder will happily include txs that push `max(regular, state)` past the block gas limit. The block then fails validation and the slot is missed.

**Suggested fix** (two parts):

1. Post-execution check in `apply_plain_transaction` — compute new totals *before* committing them, reject the tx if `max(new_regular, new_state) > gas_limit`:
```rust
let new_regular = context.block_regular_gas_used.saturating_add(tx_regular_gas);
let new_state = context.block_state_gas_used.saturating_add(tx_state_gas);
if context.is_amsterdam && new_regular.max(new_state) > context.payload.header.gas_limit {
    return Err(EvmError::Custom("block gas limit exceeded (state gas overflow)").into());
}
context.block_regular_gas_used = new_regular;
context.block_state_gas_used = new_state;
```

2. Update `remaining_gas` to reflect both dimensions so the pre-execution heuristic also rejects:
```rust
if context.is_amsterdam {
    context.remaining_gas = context.payload.header.gas_limit
        .saturating_sub(new_regular.max(new_state));
}
```

edg-l (changes requested):

azteca1998:
Fixed the payload building path to match the validation logic for Amsterdam EIP-8037 2D gas accounting.

**Changes in commit 43ea8e74f:**

1. **Added post-execution overflow check** - Before committing gas totals, check if `max(new_regular, new_state) > gas_limit` and reject the transaction if true. This prevents the builder from creating invalid blocks.

2. **Updated `remaining_gas` calculation** - Changed from tracking only `block_regular_gas_used` to tracking `max(block_regular_gas_used, block_state_gas_used)`. This ensures the pre-tx heuristic check accounts for both gas dimensions.

**Testing:**
- ✅ All EIP-7778 tests pass (3/3)
- ✅ All LEVM integration tests pass (149/149)
- ✅ Code compiles cleanly

The fix ensures block building and validation now use consistent gas accounting, preventing the scenario where a builder creates a block that would fail validation due to state gas overflow.

@edg-l This addresses your review feedback - the payload building path now has both the post-execution check and the updated `remaining_gas` calculation as you suggested.

edg-l (approved):
LGTM. Verified against EIP-7778 and EIP-8037 specs.

- Validation path: correctly defers block gas check to post-execution using `max(sum_regular, sum_state) <= gas_limit`, pre-refund values for block accounting, post-refund for receipts.
- Payload building (43ea8e74fd): correctly adds post-execution overflow guard and updates `remaining_gas` to reflect both gas dimensions.
- Pre-Amsterdam paths unchanged and correct.

crates/vm/backends/levm/mod.rs
```diff
@@ -126,18 +125,13 @@ impl LEVM {
...
-                    tx.gas_limit().min(TX_MAX_GAS_LIMIT_AMSTERDAM),
-                    block.header.gas_limit,
-                )?;
-            } else {
+            // Pre-tx gas limit guard:
+            // Pre-Amsterdam: reject tx if cumulative post-refund gas + tx.gas > block limit.
+            // Amsterdam+: skip — EIP-8037's 2D gas model means cumulative gas (regular +
+            // state) can legally exceed the block gas limit as long as
+            // max(sum_regular, sum_state) stays within it. Block-level overflow is
+            // detected post-execution.
```
iovoid, line 133: I think we should still have some sort of check to prevent DoS by blocks using more gas than allowed. In principle in a 10MB block with 100B transactions each with a 16M limit we could end up executing 1.6Tgas of code before realizing the block is actually invalid.

iovoid (changes requested):

avilagaston9 (approved):

ilitteri (approved):

iovoid (approved):

PR #6487, opened 2026-04-15
refactor(l1): migrate tooling workspace to ethrex-tooling repo

ilitteri:
## Motivation

The ethrex repo mixes core node code with development tooling (EF tests, load tests, monitor TUI, REPL, loc counter, etc.). This makes the repo harder to navigate and couples tooling release cadence to node releases. This PR moves all tooling to the new [`lambdaclass/ethrex-tooling`](https://github.com/lambdaclass/ethrex-tooling) repository.

## Description

- **Remove `tooling/` directory** — all contents migrated to `ethrex-tooling`
- **Convert `ethrex-monitor` and `ethrex-repl`** from local path deps to git deps pointing to `ethrex-tooling` (`branch = "main"`)
- **Add `[patch."https://github.com/lambdaclass/ethrex"]` section** to root `Cargo.toml` so `ethrex-tooling`'s transitive deps resolve to local workspace crates (prevents duplicate crate versions)
- **Update 7 CI workflows** to checkout `ethrex-tooling` into the `tooling/` path before use — all existing `cd tooling/...` commands remain unchanged
- **Update Dockerfile** to remove `COPY tooling` lines (monitor/repl are now fetched as git deps by cargo)
- **Add `.cargo/config.toml.example`** documenting local path patch overrides for cross-repo development

### Cross-repo dev workflow

For changes that touch both repos:
1. Use `.cargo/config.toml` patches for local development (see `.cargo/config.toml.example`)
2. Merge `ethrex-tooling` PR first
3. Run `cargo update` in ethrex to pick up new tooling revision
4. Merge ethrex PR

## How to Test

- [ ] `cargo check --workspace` passes
- [ ] CI workflows that use `tooling/` paths pass (EF tests, LOC, hive report, upgradeability, reorgs, license check, cargo lock check)
- [ ] Docker build succeeds
- [ ] `ethrex-tooling` repo builds standalone: `cargo check --workspace`


iovoid (changes requested):
The references in the Makefile and docs should be updated, and maybe the repo's existence should be mentioned in the README.

MegaRedHand (approved):
LFG

Cargo.toml
```diff
@@ -144,5 +143,29 @@ libssz-types = { git = "https://github.com/lambdaclass/libssz", rev = "7262a4f"
...
+ethrex-levm = { path = "./crates/vm/levm" }
+ethrex-p2p = { path = "./crates/networking/p2p" }
+ethrex-prover = { path = "./crates/prover" }
+ethrex-rlp = { path = "./crates/common/rlp" }
+ethrex-rpc = { path = "./crates/networking/rpc" }
+ethrex-sdk = { path = "./crates/l2/sdk" }
+ethrex-storage = { path = "./crates/storage" }
+ethrex-storage-rollup = { path = "./crates/l2/storage" }
+ethrex-trie = { path = "./crates/common/trie" }
+ethrex-vm = { path = "./crates/vm" }
```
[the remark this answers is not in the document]
MegaRedHand: I ran `make lint` and it's working

avilagaston9 (commented):
The `on.pull_request.paths` trigger in `.github/workflows/pr-main_levm.yaml` still references `tooling/ef_tests/state/**` and `tooling/ef_tests/state` (lines 11-16), but these paths no longer exist in the repo since tooling was moved to `ethrex-tooling`. PRs that only change EF test code won't trigger this workflow anymore. These stale path entries should be removed.

ilitteri:
b00af965e — Added ethrex-tooling reference to README

ilitteri:
7b4a7965f — Removed stale tooling/ef_tests/state paths from LEVM workflow trigger

docs/developers/l1/testing/ef-tests.md
```diff
@@ -10,7 +15,7 @@ See [docs](https://eest.ethereum.org/v4.1.0/consuming_tests/state_test/).
 To run the test first:
 
 ```sh
-cd tooling/ef_tests/state
+cd ethrex-tooling/ef_tests/state
```
iovoid, line 18: Inconsistent with scripts assuming the folder is named 'tooling'

iovoid (changes requested):
Some places assume the folder is cloned to tooling, and others mention ethrex-tooling.

Also the folder should be gitignored to avoid pushing it.

Dockerfile
```diff
@@ -42,7 +41,13 @@ ARG PROFILE="release"
...
 
 COPY --from=planner /ethrex/recipe.json recipe.json
-RUN cargo chef cook --release --recipe-path recipe.json $BUILD_FLAGS
+# cargo-chef cook may fail on git dependencies (ethrex-monitor/ethrex-repl)
+# because the [patch] section references local paths that don't have real
+# source code yet (only stubs). This is fine — cook is a caching optimization,
+# and the real build below has all sources available.
+COPY Cargo.* ./
+COPY .cargo/ ./.cargo
+RUN cargo chef cook --release --recipe-path recipe.json $BUILD_FLAGS || true
```
ElFantasma, line 50: **`|| true` is too broad**: the comment explains why this can fail (cook can't resolve the git deps when local-patch source isn't in scope yet), but it swallows *every* cook failure — network outages, dep-resolution bugs, lockfile drift, cargo upgrades. The real build below succeeds because it has all sources, but you've lost any signal that cook genuinely broke. Two safer shapes:
1. Filter on the known error pattern (`grep -v 'failed to load source for dependency'`), return non-zero only on unexpected lines.
2. Skip cook entirely for git-patched workspaces and document the cache trade-off.

Otherwise the current shape will hide regressions until production CI starts behaving oddly months later.

Cargo.toml
```diff
@@ -77,8 +76,8 @@ ethrex-guest-program = { path = "./crates/guest-program", default-features = fal
 ethrex-storage-rollup = { path = "./crates/l2/storage" }
 ethrex = { path = "./cmd/ethrex" }
 ethrex-l2-rpc = { path = "./crates/l2/networking/rpc" }
-ethrex-monitor = { path = "./tooling/monitor" }
-ethrex-repl = { path = "./tooling/repl" }
+ethrex-monitor = { git = "https://github.com/lambdaclass/ethrex-tooling", branch = "main" }
```
ElFantasma, line 79: (Copilot) `branch = "main"` means the resolved revision changes silently when `Cargo.lock` is regenerated. Three consequences:
- Reproducing CI failures locally requires checking out the same lockfile.
- A force-push or amend on `ethrex-tooling/main` invalidates every cached lockfile entry.
- The cross-repo dev workflow assumes `cargo update` after merging the tooling PR; manual ceremony that's easy to forget.

Use `rev = "<sha>"` and bump it explicitly per cross-repo change. Cost: one extra commit on the ethrex side; benefit: reproducible builds.

Cargo.toml
```diff
@@ -144,5 +143,29 @@ libssz-types = { git = "https://github.com/lambdaclass/libssz", rev = "7262a4f"
 libssz-merkle = { git = "https://github.com/lambdaclass/libssz", rev = "7262a4f" }
 libssz-derive = { git = "https://github.com/lambdaclass/libssz", rev = "7262a4f" }
 
+# Ensure ethrex-tooling's transitive deps resolve to local workspace crates
+# instead of fetching a second copy from git (avoids "two versions of the same crate" errors).
+[patch."https://github.com/lambdaclass/ethrex"]
```
ElFantasma, line 148: **Discoverability concern**: this section enumerates every workspace crate. Adding a new crate later requires remembering to extend this list, otherwise tooling's transitive dep resolves to a *fetched* git copy of the new crate while the rest go local — silent duplicate-crate-version situation. Two safer mechanisms:
1. CI lint that diffs the workspace `members` against the patch entries and fails when they drift.
2. A README block right next to this `[patch]` section documenting "if you add a new workspace crate, also add it here." Won't catch oversights but at least the next contributor has a fighting chance.

.github/workflows/pr-main_levm.yaml
```diff
@@ -6,11 +6,9 @@ on:
   pull_request:
     branches: ["**"]
     paths:
-      - "tooling/ef_tests/state/**"
```
ElFantasma, line 9: (per avilagaston9) `tooling/ef_tests/state/**` no longer exists in this repo. PRs that only change EF test code (now in `ethrex-tooling`) won't trigger LEVM CI here. Either drop these path triggers and rely on `ethrex-tooling`'s own CI, or move to `repository_dispatch` that `ethrex-tooling` fires when its EF tests change. First is simpler; second is correct. Worth a follow-up issue.

ElFantasma (changes requested):

PR #6496, opened 2026-04-16
feat(l1): add eth_subscribe/eth_unsubscribe WebSocket subscription support

avilagaston9:
**Motivation**

Implement `eth_subscribe("newHeads")` WebSocket subscription support. This is standard Ethereum JSON-RPC functionality needed by external tools (e.g., the Credible Layer sidecar) that track new blocks in real time.

**Description**

Add a `SubscriptionManager` GenServer actor that owns all subscription state. When a new block becomes canonical (via fork choice update on L1, or block production on L2), the head `BlockHeader` is sent to the actor, which serializes it and fans out notifications to all subscribed WebSocket clients through bounded per-connection channels.

- `SubscriptionManager` actor (`crates/networking/rpc/subscription_manager.rs`): handles `Subscribe`, `Unsubscribe`, and `NewHead` messages. Dead subscribers are cleaned up automatically on the next `NewHead` fan-out.
- `WebSocketConfig` struct bundles the WS socket address and the actor handle.
- `handle_websocket` is public so L2 reuses it directly — zero duplicate WS code in L2.
- Subscriptions are created only on explicit `eth_subscribe("newHeads")`, not on connection open.
- Bounded per-subscriber channel (capacity 512): slow clients lose notifications via `try_send` instead of accumulating unbounded memory. The actor is never blocked.
- L1 broadcasts `newHeads` from `engine/fork_choice.rs` after successful canonical head advancement.
- L2 broadcasts from the block producer after sealing each block.
- L2 uses L1's existing `--ws.enabled` / `--ws.addr` / `--ws.port` CLI flags — no L2-specific WS flags.

**How to Test**

```bash
# Start ethrex with WS enabled
ethrex --ws.enabled --ws.port 8546 ...

# In another terminal, subscribe via websocat or python:
python3 -c "
import asyncio, json, websockets
async def test():
    async with websockets.connect('ws://localhost:8546') as ws:
        await ws.send(json.dumps({'jsonrpc':'2.0','method':'eth_subscribe','params':['newHeads'],'id':1}))
        print('Subscribe:', json.loads(await ws.recv()))
        print('Notification:', json.loads(await asyncio.wait_for(ws.recv(), timeout=15)))
asyncio.run(test())
"
```

**Checklist**

- [x] Code compiles (`cargo check -p ethrex --features l2`)
- [x] Zero warnings
- [x] L2 reuses L1 implementation (no duplicate WS code)
- [x] Bounded channels prevent memory issues with slow clients
- [ ] CI passing (pending)


crates/networking/rpc/rpc.rs
iovoid, line 458: No longer true (we don't take just an address, there are now ws-specific endpoints)

iovoid, line 467: Should be updated

crates/networking/rpc/rpc.rs
```diff
@@ -670,29 +690,199 @@ pub async fn handle_authrpc_request(
...
+    out_tx: &tokio::sync::mpsc::Sender<String>,
+    subscription_ids: &mut Vec<String>,
+    route_request: &F,
+) -> Option<String>
+where
+    F: Fn(RpcRequest) -> Fut,
+    Fut: std::future::Future<Output = Result<Value, E>>,
+    E: Into<RpcErrorMetadata>,
+{
+    let req: RpcRequest = match serde_json::from_str(body) {
```
iovoid, line 766: This doesn't use `RpcRequestWrapper` like `handle_http_request`, meaning this breaks support for batch requests

iovoid (changes requested):

crates/networking/rpc/rpc.rs
```diff
@@ -670,29 +690,199 @@ pub async fn handle_authrpc_request(
...
+            let resp = serde_json::json!({
+                "jsonrpc": "2.0",
+                "id": null,
+                "error": {
+                    "code": -32700,
+                    "message": "Parse error"
+                }
+            });
+            return Some(resp.to_string());
+        }
```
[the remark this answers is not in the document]
avilagaston9: Addressed in 80647af8. `handle_ws_request` now parses the body into `serde_json::Value` first (→ -32700 for malformed JSON), then into `RpcRequestWrapper` (→ -32600 for structurally-invalid requests). Batch support is restored: each entry is routed through `process_ws_request`, which still intercepts `eth_subscribe` / `eth_unsubscribe` per-connection. Empty batches respond with -32600 per the JSON-RPC 2.0 spec.

crates/networking/rpc/rpc.rs
```diff
@@ -475,7 +488,7 @@ pub fn start_block_executor(blockchain: Arc<Blockchain>) -> UnboundedSender<Bloc
 #[allow(clippy::too_many_arguments)]
 pub async fn start_api(
     http_addr: SocketAddr,
-    ws_addr: Option<SocketAddr>,
+    ws: Option<WebSocketConfig>,
     authrpc_addr: SocketAddr,
     storage: Store,
```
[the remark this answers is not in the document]
avilagaston9: Addressed in 3f9195e0 — updated the doc comment to describe `ws: Option<WebSocketConfig>` and the subscription methods it serves.

crates/networking/rpc/subscription_manager.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// Build the standard Ethereum subscription notification envelope.
+///
+/// Takes a pre-serialized `result_json` string to avoid re-serializing the
+/// header for every subscriber during fan-out.
+fn build_subscription_notification(sub_id: &str, result_json: &str) -> String {
+    format!(
+        r#"{{"jsonrpc":"2.0","method":"eth_subscription","params":{{"subscription":"{sub_id}","result":{result_json}}}}}"#
+    )
+}
```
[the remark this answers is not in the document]
avilagaston9: Addressed in c4d662ea. `build_subscription_notification` now uses `serde_json::json!` and takes the header as `&serde_json::Value` directly, so escaping is delegated to serde_json and the envelope stays correct regardless of what `sub_id` or the result contain.

crates/networking/rpc/rpc.rs
```diff
@@ -578,21 +592,27 @@ pub async fn start_api(
...
+    if let Some(ref ws_config) = ws {
+        let ws_handler = |ws: WebSocketUpgrade, State(ctx): State<RpcApiContext>| async {
+            ws.on_upgrade(|mut socket| async move {
+                handle_websocket(&mut socket, &ctx, |req| {
+                    let c = ctx.clone();
+                    async move { map_http_requests(&req, c).await }
+                })
+                .await;
+            })
         };
```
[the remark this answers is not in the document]
avilagaston9: Addressed in 6cdaa3a3 — L1 handler now uses `async move` to match L2.

crates/networking/rpc/rpc.rs
avilagaston9: Addressed in 3f9195e0.

avilagaston9: Addressed in 3f9195e0.

crates/networking/rpc/rpc.rs
avilagaston9: Addressed in 80647af8 — WS now parses via `RpcRequestWrapper`, so single and batched requests both work, matching the HTTP handler.

iovoid (approved):

crates/networking/rpc/subscription_manager.rs
```diff
@@ -0,0 +1,168 @@
...
+
+/// Maximum number of buffered notifications per subscriber.
+/// If a subscriber's channel is full (slow WebSocket client), the notification
+/// is dropped rather than blocking the actor. Matches Geth's approach of
+/// dropping slow clients (Geth uses 20,000; we use a smaller buffer since
+/// each notification is already serialized JSON).
+pub const SUBSCRIBER_CHANNEL_CAPACITY: usize = 512;
+
+/// Maximum number of active subscriptions allowed per WebSocket connection.
+pub const MAX_SUBSCRIPTIONS_PER_CONNECTION: usize = 128;
```
ElFantasma, line 34: **No global subscription cap (DoS surface)**: `MAX_SUBSCRIPTIONS_PER_CONNECTION = 128` is enforced per-connection, but no cap on *total* subscribers. A malicious client opening 1000 connections × 128 subscriptions = 128k entries in `self.subscribers`, each holding a 512-cap `mpsc::Sender<String>`. Memory cost grows linearly with no upper bound. For a public RPC endpoint, that's a DoS vector. Two mitigations:
- Add a `MAX_TOTAL_SUBSCRIPTIONS` cap inside the actor (return `Err` from `subscribe`).
- Cap concurrent WS connections at the accept layer (probably belongs upstream, but coordinate with this cap).

Not blocking unless this is exposed to the public internet at deploy time, but worth pinning down.

crates/networking/rpc/subscription_manager.rs
```diff
@@ -0,0 +1,168 @@
...
+        let mut dead_ids: Vec<String> = Vec::new();
+
+        for (sub_id, sender) in &self.subscribers {
+            let notification = build_subscription_notification(sub_id, &header_value);
+            match sender.try_send(notification) {
+                Ok(()) => {}
+                Err(tokio::sync::mpsc::error::TrySendError::Closed(_)) => {
+                    dead_ids.push(sub_id.clone());
+                }
+                Err(tokio::sync::mpsc::error::TrySendError::Full(_)) => {
```
ElFantasma, line 122: **Silent drops**: `try_send` on Full leaves the subscriber's view of the chain non-contiguous without any indication. Some Credible-Layer-shaped consumers will quietly miss a block and not realize. Two protocol-level options:
- On `Full`, close the connection (forces resubscribe — at least *visible*).
- Emit a synthetic `subscription_gap` notification (also via `try_send` so it stays bounded) — preserves the connection but tells the client "you missed something."

Geth chose silent-drop, so matching that is defensible — just want to make sure the choice is deliberate.

crates/networking/rpc/subscription_manager.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    #[request_handler]
+    async fn handle_subscribe(
+        &mut self,
+        msg: subscription_manager_protocol::Subscribe,
+        _ctx: &Context<Self>,
+    ) -> String {
+        let id = generate_subscription_id();
+        self.subscribers.insert(id.clone(), msg.sender);
```
ElFantasma, line 148: (Copilot collision concern) Mathematically negligible at any realistic scale (~2^64 IDs for ~50% chance), but the fix is trivially cheap:
```rust
loop {
    let id = generate_subscription_id();
    if let std::collections::hash_map::Entry::Vacant(e) = self.subscribers.entry(id.clone()) {
        e.insert(msg.sender);
        return id;
    }
}
```
Not worth blocking; easy slot-in if you touch this line.

crates/networking/rpc/subscription_manager.rs
```diff
@@ -0,0 +1,168 @@
...
+            map.insert(
+                "hash".to_string(),
+                Value::String(format!("{block_hash:#x}")),
+            );
+        }
+
+        let mut dead_ids: Vec<String> = Vec::new();
+
+        for (sub_id, sender) in &self.subscribers {
+            let notification = build_subscription_notification(sub_id, &header_value);
```
ElFantasma, line 116: Per-subscriber clone: this calls `serde_json::json!({...result: result, ...}).to_string()` once per subscriber, cloning the `Value`. For 100 subscribers × ~2 KB header, ~200 KB allocation per block — cheap, but if subscriber count climbs, a faster shape would be to serialize the result *once* into a string, then `format!` only the `"subscription":"<id>"` field per subscriber. Not blocking; mark for later if you ever profile a hot newHeads cycle.

azteca1998 (approved):

ElFantasma (approved):
This LGTM, but I added some comments that may be ignored, fixed or trigger new issues. None seem blocking

Also:

**Tests** (per Copilot): 676 lines added, zero new tests in the diff (other than `subscription_manager_tests.rs` which wasn't in the file list I saw — apologies if it actually exists). The actor + WS integration is exactly where protocol-level bugs slip in: what does the WS write loop do when its mpsc receiver returns `None`? does the connection close cleanly? does `unsubscribe` of a non-existent ID return the right error? Worth at minimum:
- An actor unit test (`subscribe` → `NewHead` → assert notification shape).
- An integration test driving `handle_websocket` with a mock socket through subscribe → NewHead → unsubscribe.
- Regression test for the `RpcRequestWrapper` batch handling iovoid flagged (now addressed in 80647af8).

**`rand::thread_rng()` in the actor (greptile)** is correct — `thread_rng()` is per-OS-thread and re-seeds lazily, so even if the actor moves between threads the IDs stay random. Acceptable as-is.

**`WebSocketConfig` L2-only deployment**: PR says "L2 uses L1's existing flags" — worth confirming an L2-only node with `--ws.enabled` works. If something silently binds to L1-default address, that's a foot-gun. Add to test plan: L2-only mode + WS enabled + verify subscribe works.

ilitteri (approved):

ilitteri (approved):

PR #6498, opened 2026-04-16
feat(l1): bump zkEVM fixtures to v0.3.3 and tighten witness validation

ilitteri:
## Summary
- Bump `.fixtures_url_zkevm` from `zkevm@v0.3.0` to `zkevm@v0.3.3`. v0.3.3 introduces new `witness_validation_*` tests that exercise invalid execution witnesses (missing code, reordered headers, extra unused trie nodes, etc.).
- 16 of the new tests initially failed, exposing real gaps in ethrex's stateless validation. Fixed all of them:
  - `get_account_code` / `get_code_metadata` now error on missing bytecode instead of silently falling back to empty code. A missing code hash means the witness is invalid; the guest program must not conclude anything about the state.
  - `RpcExecutionWitness::into_execution_witness` tolerates extra unused state entries that don't decode as trie nodes (matches other clients, passes `validation_state_extra_unused_trie_node`).
  - `GuestProgramState::from_witness` rejects non-contiguous or reordered `block_headers_bytes`. Headers must form a contiguous ascending chain.
  - `execute_blocks` now propagates errors from `get_first_invalid_block_hash` instead of swallowing `NoncontiguousBlockHeaders`.
- Test runner: `run_stateless_from_fixture` reads the fixture's `statelessOutputBytes` valid flag (byte 32) and expects execution to fail when the fixture marks the block invalid (`valid=0x00`).

Final: **93/93 zkevm tests pass** under `--features stateless`.

## Test plan
- [ ] `cd tooling/ef_tests/blockchain && make test-stateless-zkevm` passes
- [ ] Full EF blockchain suite (`make test`) is no worse than main

### DIFF
--- a/crates/common/types/block_execution_witness.rs
+++ b/crates/common/types/block_execution_witness.rs
@@ -151,15 +151,26 @@ impl RpcExecutionWitness {
             ));
         }
 
-        let mut initial_state_root = None;
-
-        for h in &self.headers {
-            let header = BlockHeader::decode(h)?;
-            if header.number == first_block_number - 1 {
-                initial_state_root = Some(header.state_root);
-                break;
+        // Decode headers up front so we can validate the ordering invariant:
+        // an RPC witness must provide headers in ascending, contiguous block-number
+        // order. Reordered or gapped inputs would otherwise be silently normalized
+        // by the BTreeMap in `GuestProgramState::from_witness` and yield an
+        // incorrect conclusion about the block's validity.
+        let decoded_headers: Vec<BlockHeader> = self
+            .headers
+            .iter()
+            .map(|h| BlockHeader::decode(h))
+            .collect::<Result<_, _>>()?;
+        for window in decoded_headers.windows(2) {
+            if window[1].number != window[0].number + 1 {
+                return Err(GuestProgramStateError::NoncontiguousBlockHeaders);
             }
         }
+
+        let initial_state_root = decoded_headers
+            .iter()
+            .find(|h| h.number == first_block_number - 1)
+            .map(|h| h.state_root);
 
         let initial_state_root = initial_state_root.ok_or_else(|| {
             GuestProgramStateError::Custom(format!(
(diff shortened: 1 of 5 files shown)


crates/common/types/block_execution_witness.rs
```diff
@@ -151,16 +151,27 @@ impl RpcExecutionWitness {
...
+        // order. Reordered or gapped inputs would otherwise be silently normalized
+        // by the BTreeMap in `GuestProgramState::from_witness` and yield an
+        // incorrect conclusion about the block's validity.
+        let decoded_headers: Vec<BlockHeader> = self
+            .headers
+            .iter()
+            .map(|h| BlockHeader::decode(h))
+            .collect::<Result<_, _>>()?;
+        for window in decoded_headers.windows(2) {
+            if window[1].number != window[0].number + 1 {
```
ElFantasma, line 165: **Security observation**: this only validates *numbering*, not *chain linkage*. An attacker could provide headers numbered correctly but with arbitrary `parent_hash` values, fabricating any `state_root` for the block being verified. The check is a useful guard against accidental reordering / gaps from honest peers, but a malicious witness can trivially bypass it.

#6541 (open) adds the load-bearing check: `window[1].parent_hash == window[0].hash()`. That's the one that closes the door on fabricated state roots. This PR's numbering check is a useful preliminary but not sufficient on its own.

Worth either (a) rebasing on #6541 if the order works out, or (b) adding the parent_hash check here so this PR is self-sufficient even if #6541 lingers.

crates/common/types/block_execution_witness.rs
```diff
@@ -177,10 +188,13 @@ impl RpcExecutionWitness {
                     // which would fail to decode in ours
                     return None;
                 }
+                // Tolerate extra unused witness entries that don't decode as trie nodes.
```
ElFantasma, line 191: The tolerance is correct per EIP-8025 §Tolerance — extra unused witness entries should not be a hard error. But the change from `.collect::<Result<_, RLPDecodeError>>()?` to `Node::decode(&b).ok()` silently swallows *all* decode errors, including ones that aren't "extra junk" — e.g., a corrupted node that the witness genuinely needs. Those errors would now manifest later as "node not found" lookup failures during execution, which is a less direct error message.

Not blocking — the spec is permissive here — but a `tracing::debug!` or counter on undecodable entries would help diagnose corrupt-witness cases when they happen.

tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -571,12 +591,39 @@ async fn run_stateless_from_fixture(
...
+/// Layout: `new_payload_request_root (32 bytes) || valid (1 byte) || padding`.
+/// If absent or unparseable, default to expecting success (true) to preserve
+/// the prior behavior for fixtures without this field.
+#[cfg(feature = "stateless")]
+fn expected_stateless_valid(hex_str: Option<&str>) -> bool {
+    let Some(s) = hex_str else { return true };
+    let s = s.strip_prefix("0x").unwrap_or(s);
+    // valid byte is at hex offset 64..66 (bytes[32])
+    match s.get(64..66).and_then(|b| u8::from_str_radix(b, 16).ok()) {
+        Some(0) => false,
```
ElFantasma, line 626: nit: the `_ => true` default-to-true preserves prior behavior for fixtures without `statelessOutputBytes`, but it also defaults to true for hex strings that are *malformed* (not 64+ chars, non-hex digits at offset 64..66, etc.). For a new test fixture with a typo'd `statelessOutputBytes`, the test would silently expect success and pass when it shouldn't. Three distinct cases collapse into the same default:
- Field absent → expect success (correct fallback)
- Field present but malformed → ??? (currently expect success)
- Field present with valid byte 0x01 → expect success

Returning `Result<bool, ParseError>` (or `Option<bool>` with the absent vs malformed distinction) would let the caller treat malformed fixtures as a hard error rather than a silent default. (Same concern as the parse helper in #6527 — worth keeping consistent.)

azteca1998 (approved):

PR #6500, opened 2026-04-17
fix(l1): prune Docker build cache and handle disk-full errors in multisync monitor

avilagaston9:
**Motivation**

The multisync monitor on srv9 crashed with `OSError: [Errno 28] No space left on device` after completing Run #9 successfully. The Docker build cache was growing unbounded across runs (750 GB accumulated), eventually filling the 1.8 TB disk entirely. When the monitor tried to write run logs, ENOSPC killed the process, leaving containers running unattended with no Slack notifications for 42+ hours.

**Description**

Two changes to `docker_monitor.py`:

1. **Prune Docker build cache after each image build** — runs `docker builder prune -f` after every successful `docker build`, preventing cache from accumulating across multisync cycles. The prune is non-fatal: if it fails, a warning is printed and the monitor continues.

2. **Handle disk-full errors gracefully in log writing** — `save_all_logs` and `log_run_result` now catch `OSError` so a full disk degrades to a warning instead of crashing the monitor loop. On write failure, the run result is printed to stdout so it's still captured in tmux history.

**Checklist**

- [x] No code compilation needed (Python script)
- [x] Already deployed and running on srv9

### DIFF
--- a/tooling/sync/docker_monitor.py
+++ b/tooling/sync/docker_monitor.py
@@ -398,6 +398,19 @@ def build_docker_image(profile: str, image_tag: str, ethrex_dir: str) -> bool:
             check=True
         )
         print("✅ Docker image built successfully")
+        # Prune build cache to prevent unbounded disk growth across runs
+        try:
+            result = subprocess.run(
+                ["docker", "builder", "prune", "-f"],
+                capture_output=True, text=True, timeout=120
+            )
+            if result.returncode == 0:
+                reclaimed = result.stdout.strip().split("\n")[-1] if result.stdout.strip() else ""
+                print(f"🧹 Build cache pruned. {reclaimed}")
+            else:
+                print(f"⚠️ Build cache prune failed: {result.stderr.strip()}")
+        except Exception as e:
+            print(f"⚠️ Build cache prune error: {e}")
         return True
     except subprocess.CalledProcessError as e:
         print(f"❌ Failed to build Docker image: {e}")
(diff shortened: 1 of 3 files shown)


ilitteri (approved):

tooling/sync/docker_monitor.py
```diff
@@ -662,15 +675,18 @@ def save_container_logs(container: str, run_id: str, suffix: str = ""):
...
+        for inst in instances:
+            # Save ethrex logs
+            save_container_logs(inst.container, run_id)
+            # Save consensus logs (convention: consensus-{network})
+            consensus_container = inst.container.replace("ethrex-", "consensus-")
+            save_container_logs(consensus_container, run_id)
+
+        print(f"📁 Logs saved to {LOGS_DIR}/run_{run_id}/\n")
+    except OSError as e:
+        print(f"⚠️ Failed to save some logs (disk full?): {e}", flush=True)
```
iovoid: `save_container_logs` already catches exceptions

iovoid (changes requested):

tooling/sync/docker_monitor.py
avilagaston9: Addressed in [da22a35](https://github.com/lambdaclass/ethrex/pull/6500/commits/da22a354ee49b29364d6a942dfb49f86808b504d).

iovoid (approved):

ElFantasma (approved):

PR #6501, opened 2026-04-17
feat(l1): implement prestateTracer for debug_traceTransaction and debug_traceBlockByNumber

avilagaston9:
> **Depends on:** #6496 (eth_subscribe/eth_unsubscribe WS support) — must be merged first.

**Motivation**

The `prestateTracer` is a standard Ethereum debug tracer that captures pre-execution account state (balance, nonce, code, storage) and optionally computes a post-execution diff. It's needed by external tools like the Credible Layer sidecar, which uses `debug_traceBlockByNumber` with `prestateTracer` in diff mode to build its local state database.

**Description**

Implement the `prestateTracer` (with optional `diffMode`) across the tracing stack:

- Prestate types in `ethrex-common` with serde formatting to match Geth's output format.
- `trace_tx_prestate` in `ethrex-vm` snapshots the `GeneralizedDatabase` cache before executing a transaction, then compares against the post-execution cache to identify touched accounts. For already-cached accounts whose new storage slots were loaded during the tx, it merges values from `initial_accounts_state` to capture original slot values.
- `trace_transaction_prestate` / `trace_block_prestate` in `ethrex-blockchain` rebuild parent state and orchestrate per-tx tracing with timeouts.
- `PrestateTracer` variant wired into the RPC `debug_traceTransaction` and `debug_traceBlockByNumber` handlers with `PrestateTracerConfig { diffMode }`.

Also extracts the duplicated `TestDatabase` into a shared `test/tests/levm/test_db.rs` module reused by both `l2_hook_tests` and the new `prestate_tracer_tests`.

**How to Test**

```bash
# Trace a transaction with prestateTracer
curl -X POST http://localhost:8545 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"debug_traceTransaction","params":["<TX_HASH>",{"tracer":"prestateTracer"}],"id":1}'

# Trace with diff mode
curl -X POST http://localhost:8545 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"debug_traceTransaction","params":["<TX_HASH>",{"tracer":"prestateTracer","tracerConfig":{"diffMode":true}}],"id":1}'

# Trace a block
curl -X POST http://localhost:8545 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"debug_traceBlockByNumber","params":["latest",{"tracer":"prestateTracer","tracerConfig":{"diffMode":true}}],"id":1}'
```


crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +121,142 @@ impl LEVM {
...
+    pre_snapshot: &'a CacheDB,
+    post_cache: &'a CacheDB,
+    db: &'a GeneralizedDatabase,
+) -> Vec<(Address, &'a LevmAccount, &'a LevmAccount)> {
+    let mut touched = Vec::new();
+
+    for (addr, post_account) in post_cache {
+        let pre_account = match pre_snapshot.get(addr) {
+            Some(pre) => {
+                if pre.info == post_account.info && pre.storage == post_account.storage {
```
ElFantasma: **Correctness gap vs Geth**: this `pre == post` early-exit is correct for *diff* mode (no diff to report), but Geth's prestateTracer in **non-diff mode** is supposed to capture every account *accessed* by the tx — read-only loads included (e.g., a contract that does only `SLOAD` and `BALANCE`). Right now those vanish from the trace. Either split `find_touched_accounts` or gate this `continue` on `diff_mode`. (Greptile flagged this.)

Also: this loop iterates `post_cache` only, so self-destructed accounts removed from `current_accounts_state` will silently disappear from the trace. Worth a SELFDESTRUCT test to lock down behavior either way.

crates/common/tracing.rs
```diff
@@ -65,3 +66,46 @@ pub struct CallLog {
     pub data: Bytes,
     pub position: u64,
 }
+
+/// Account state as captured by the prestateTracer.
+/// Matches Geth's prestateTracer output format.
+/// https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#prestate-tracer
+#[derive(Debug, Serialize, Default, Clone)]
+pub struct PrestateAccountState {
+    pub balance: U256,
```
ElFantasma: **Output format gap vs Geth**: no `skip_serializing_if` on `balance`, so accounts with zero balance emit `"balance":"0x0"`. Geth omits the field entirely. Add `skip_serializing_if = "is_zero_balance"` to match the existing `is_zero_nonce` pattern. (Greptile flagged.)

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +121,142 @@ impl LEVM {
...
+        }
+
+        result.insert(addr, state);
+    }
+
+    result
+}
+
+/// Build the post-tx state map for all accounts touched by a transaction.
+fn build_post_state_map(
```
ElFantasma, line 340: **Diff-mode post-state for newly-loaded accounts**: this skips its `pre_cached.storage` retain block when the account was first accessed this tx (no entry in `pre_snapshot`). Every slot loaded during execution lands in `post`, even slots that were only `SLOAD`-read with unchanged values. Geth filters post to slots actually mutated. Fix: also retain only slots whose `post != initial` for the new-account branch.

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +121,142 @@ impl LEVM {
...
+}
+
+/// Build the pre-tx state map for all accounts touched by a transaction.
+///
+/// For already-cached accounts, the pre_snapshot only contains storage slots
+/// loaded by *previous* transactions. Any slot first accessed during *this*
+/// transaction has its original value in `initial_accounts_state`. We merge
+/// both sources so the output includes every accessed slot, then filter to
+/// only slots actually touched by this tx (newly loaded or value changed).
+fn build_pre_state_map(
```
ElFantasma, line 269: Storage merge for already-cached accounts is correct (start from `pre_snapshot.storage`, fill in slots from `initial_accounts_state.storage`, retain only slots touched by this tx). One asymmetry: for an account first loaded this tx (no `pre_snapshot` entry), `initial_accounts_state` is used directly without the retain step. Today this is fine because `initial_accounts_state` only contains slots accessed during this tx, but if a future change ever caches more eagerly the assumption breaks. Defensive `retain` keyed on `post_account.storage` would future-proof it.

crates/blockchain/tracing.rs
```diff
@@ -82,6 +86,73 @@ impl Blockchain {
...
+        // Trace each transaction sequentially — state accumulates between calls
+        // We need to do this in order to pass ownership of block & evm to a blocking process without cloning
+        let vm = Arc::new(Mutex::new(vm));
+        let block = Arc::new(block);
+        let mut traces = vec![];
+        for index in 0..block.body.transactions.len() {
+            let block = block.clone();
+            let vm = vm.clone();
+            let tx_hash = block.as_ref().body.transactions[index].hash();
+            let result = timeout_trace_operation(timeout, move || {
```
ElFantasma, line 149: Cancellation-safety: each tx trace runs under `timeout_trace_operation` and mutates the shared `Mutex<Evm>`. If the timeout fires mid-tx, the spawn drops — lock guard releases, but `vm` may be in a partially-applied state (some opcodes ran, no rollback). Subsequent txs run on top of corrupted state and emit nonsense traces. Same pattern exists in `call_trace_block_with_logs`, so this isn't novel, but worth either documenting or making `Evm::trace_tx_prestate` cancellation-safe. Right now `vm.execute()` mutates `current_accounts_state` directly — not safe.

azteca1998 (approved):

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +121,142 @@ impl LEVM {
...
+    }
+
+    touched
+}
+
+/// Build the account state output for one account.
+fn build_account_output(account: &LevmAccount, db: &GeneralizedDatabase) -> PrestateAccountState {
+    let code = if account.info.code_hash != *EMPTY_KECCACK_HASH {
+        db.codes
+            .get(&account.info.code_hash)
```
iovoid: If the account has code but it was never used (balance-only changes) we emit empty code. If that's expected behavior it should be documented.

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +121,142 @@ impl LEVM {
...
+/// `initial_accounts_state` if the account was first loaded during this tx.
+fn find_touched_accounts<'a>(
+    pre_snapshot: &'a CacheDB,
+    post_cache: &'a CacheDB,
+    db: &'a GeneralizedDatabase,
+) -> Vec<(Address, &'a LevmAccount, &'a LevmAccount)> {
+    let mut touched = Vec::new();
+
+    for (addr, post_account) in post_cache {
+        let pre_account = match pre_snapshot.get(addr) {
```
iovoid, line 142: Should also check if the account was destroyed

iovoid (changes requested):

crates/vm/backends/levm/tracing.rs
avilagaston9: Fixed in 41895b88d — `find_touched_accounts` includes destroyed accounts, omitted from diff post.

crates/vm/backends/levm/tracing.rs
avilagaston9: Fixed in 41895b88d — `preload_touched_codes` fetches code for every touched contract before output.

iovoid (approved):

ElFantasma (changes requested):

crates/vm/backends/levm/tracing.rs
avilagaston9: - Read-only: `pre == post` early-exit removed in 41895b88d. Covered by `prestate_trace_includes_read_only_account`.
- Self-destruct: LEVM only marks `Destroyed` for same-tx-created contracts post-Cancun, so pre-existing ones keep `Modified` and survive. No integration test yet — follow-up.

crates/common/tracing.rs
avilagaston9: Refined in ecda774db — `Option<U256>` so balance→0 isn't suppressed in diff post. Covered by `prestate_diff_post_emits_zero_balance_when_changed`.

crates/vm/backends/levm/tracing.rs
avilagaston9: Fixed in ecda774db — `pre_storage_value` falls back to `initial_accounts_state.storage` for newly-accessed slots. Covered by `prestate_diff_post_excludes_unchanged_storage_for_newly_accessed_account`.

crates/vm/backends/levm/tracing.rs
avilagaston9: Done in 02fdd6fdd — `build_pre_state_map` retains slots only if present in `post_account.storage`. Covered by `prestate_pre_storage_excludes_slots_not_present_in_post`.

crates/blockchain/tracing.rs
avilagaston9: I think it's safe: `await?` propagates the timeout out of the loop, so no later tx runs on corrupted state. The blocking task finishes on its own and drops its `vm` Arc — the result is never observed. Same shape as `trace_block_calls`. WDYT?

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +132,273 @@ impl LEVM {
...
+/// Reads code from `db.codes`; caller must `preload_touched_codes` first.
+/// Storage values are passed through as-is (including zero); per-field filtering
+/// for diff-mode post is applied by `build_post_output`.
+fn build_account_output(account: &LevmAccount, db: &GeneralizedDatabase) -> PrestateAccountState {
+    let has_code = account.info.code_hash != *EMPTY_KECCACK_HASH;
+    let code = if has_code {
+        db.codes
+            .get(&account.info.code_hash)
+            .map(|c| c.bytecode.clone())
+            .expect("code preloaded by preload_touched_codes")
```
ElFantasma: **Panic on missing preloaded code**: `.expect("code preloaded by preload_touched_codes")` will crash the tracer if `preload_touched_codes` ever fails to load a code hash that subsequently appears here. The function does `db.get_code(hash)?` for every touched contract's code_hash, but if `get_code` ever returns `Ok(None)` for a hash that was supposed to be there (e.g., a backend race or a code-hash mismatch from a corrupted DB), this `expect` panics inside an RPC handler — caller sees a 500 with no useful message.

Safer: return a `EvmError::Custom("missing preloaded code for {hash}")` instead. The invariant the expect captures is real, but a panic is a poor failure mode for a tracer that runs on user-supplied tx hashes. Same expect appears at line 225 — fix in both.

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +132,273 @@ impl LEVM {
...
+                .unwrap_or_default();
+            *v != H256::from_uint(&post_val)
+        });
+    }
+}
+
+/// Returns the set of addresses whose state changed in this tx (i.e. would appear
+/// in diff `post`). Used to prune diff `pre` to the same set, plus destroyed accounts
+/// which appear only in `pre`.
+fn modified_or_destroyed_addresses(
```
ElFantasma: nit: `modified_or_destroyed_addresses` calls `build_post_output` per-account, and then `build_post_state_map` calls `build_post_output` *again* for every address that `modified_or_destroyed_addresses` returned (since `kept.contains(addr)` was checked already). For a heavy block this is 2× the per-account work just for the diff-mode path.

Refactor: have `build_post_state_map` return both the post map AND the set of "actually modified" addresses in one pass, then drop `modified_or_destroyed_addresses`. Or memoize `build_post_output` results by address. Not blocking — tracer cost is dominated by tx execution, not output building — but if you ever profile a heavy diffMode trace, this is the first thing to fix.

crates/blockchain/tracing.rs
```diff
@@ -82,6 +86,77 @@ impl Blockchain {
...
```
ElFantasma, line 149: **Cancellation safety — sharper framing than my original comment**: I previously said state could be left in a partially-applied state on timeout. That was wrong — the `Mutex<vm>` lock is held by the closure for its full duration, so state IS consistent. The actual concern is different:

`tokio::time::timeout` wraps `tokio::task::spawn_blocking`, but **dropping a `JoinHandle` does not abort the task**. So when timeout fires:
1. The `await` returns `Err(Tracing Timeout)`.
2. The blocking task *keeps running on the blocking thread pool*, holding its `Arc<Mutex<vm>>` clone.
3. `trace_block_prestate` returns early via `?`, dropping its `vm` Arc.
4. The orphaned blocking task runs to completion (could be hours for a pathological tx), occupying a worker, then finally drops its Arc and frees the vm.

Upshot: the timeout is **advisory**, not enforced. It doesn't free resources for retry. A pathological tx — or a malicious one crafted to be slow — can occupy blocking pool workers for arbitrary duration after the user's "timeout" returned an error. Multiple such timeouts × multiple concurrent traces could exhaust the pool.

Fixes:
- Document that the timeout is advisory and not enforced (cheapest).
- Use a cancellation token threaded through `vm.execute()` so opcode dispatch can check and bail early. Heavyweight but correct.
- Cap the gas budget passed to the trace VM at something well below the block gas limit, so even a pathological tx can't run for too many opcodes. Indirect but effective.

Not blocking; flagging because the original wording was technically wrong and this re-review is the right place to fix that record.

ElFantasma (commented):
Re-review against commits 41895b88d, ecda774db, and 02fdd6fdd (the three follow-ups since my CHANGES_REQUESTED on 2026-04-30T14:55Z).

**Original concerns — status:**

| # | Concern | Status |
|---|---------|--------|
| 1 | Read-only accounts dropped (`find_touched_accounts` `pre == post` filter) | ✅ Addressed in `41895b88d` — filter removed; equality skip moved to `build_post_state_map` only |
| 2 | `balance` always serialized | ✅ Addressed in `ecda774db` — switched to `Option<U256>` with `skip_serializing_if = "Option::is_none"`. Geth-compat: `Some(0)` for explicit drain, `None` for unmodified |
| 3 | Diff-mode post-state for newly-loaded accounts | ✅ Addressed in `ecda774db` — `build_post_output` now returns `Option<PrestateAccountState>` and only sets fields where `pre != post` |
| 4 | Storage merge asymmetry in `build_pre_state_map` | ✅ Addressed in `02fdd6fdd` — added `if !post_account.storage.contains_key(k) { return false }` to the retain block, defending against future caching changes |
| 5 | Cancellation safety on block-tracing mutex+timeout | ❌ Still open — see inline; concern is sharper than I originally framed it |

New concerns inline below — all minor relative to the original 5. The `expect()` panic risk on missing preloaded code is the one I'd lean on fixing; the others are polish.

crates/vm/backends/levm/tracing.rs
avilagaston9: Done in 104a0e89a — added a `get_preloaded_code` helper that returns `EvmError::Custom("missing preloaded code for {hash}")`; both call sites updated.

crates/vm/backends/levm/tracing.rs
avilagaston9: Done in 104a0e89a — `build_post_state_map` now returns `(PrestateTrace, HashSet<Address>)` in one pass; `modified_or_destroyed_addresses` removed.

crates/blockchain/tracing.rs
avilagaston9: Tracked in #6577 — out of scope here since the same pattern predates this PR.

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +132,273 @@ impl LEVM {
...
+    for (key, post_val) in &post_account.storage {
+        if post_val.is_zero() {
+            continue;
+        }
+        let pre_val = pre_storage_value(addr, key, pre_snapshot, db).unwrap_or_default();
+        if pre_val == *post_val {
+            continue;
+        }
+        state.storage.insert(*key, H256::from_uint(post_val));
+        modified = true;
```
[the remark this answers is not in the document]
avilagaston9: Fixed in 2eceb224f — `is_zero` check now gates only the `state.storage.insert` call, so clears flip `modified=true` and the cleared slot is encoded by absence in `post.storage`. Test: `prestate_diff_post_omits_cleared_storage_slot`.

crates/vm/backends/levm/tracing.rs
```diff
@@ -79,3 +132,273 @@ impl LEVM {
...
+            None => {
+                let Some(initial) = db.initial_accounts_state.get(addr) else {
+                    continue;
+                };
+                initial
+            }
+        };
+
+        touched.push((*addr, pre_account, post_account));
+    }
```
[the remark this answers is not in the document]
avilagaston9: Tracked in #6579 — the cache-delta approach inherent to this implementation cannot distinguish "accessed this tx" from "loaded earlier, untouched" without a per-tx access tracker in LEVM. Out of scope for this PR.

crates/common/tracing.rs
```diff
@@ -65,3 +66,65 @@ pub struct CallLog {
...
+            && self.code.is_empty()
+            && self.code_hash.is_zero()
+            && self.storage.is_empty()
+    }
+}
+
+/// Per-transaction prestate trace (non-diff mode).
+/// Maps account address to its state before the transaction.
+pub type PrestateTrace = HashMap<Address, PrestateAccountState>;
+
```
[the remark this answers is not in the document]
avilagaston9: Fixed in 2eceb224f — `PrestateTrace`, `PrePostState.{pre,post}`, and `PrestateAccountState.storage` now use `BTreeMap` for deterministic, sorted JSON output.

crates/common/tracing.rs
```diff
@@ -65,3 +66,65 @@ pub struct CallLog {
...
+    /// no code hash, no storage). Used to drop entries that would serialize to `{}`.
+    /// A `Some(0)` balance is treated as no balance for emptiness purposes — it carries
+    /// no diff information beyond what the absence of the field would imply.
+    pub fn is_empty(&self) -> bool {
+        self.balance.unwrap_or_default().is_zero()
+            && self.nonce == 0
+            && self.code.is_empty()
+            && self.code_hash.is_zero()
+            && self.storage.is_empty()
+    }
```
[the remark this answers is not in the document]
avilagaston9: Doc tightened in 2eceb224f — now reads "True when no field conveys information; `Some(0)` balance counts as empty."

ElFantasma (approved):

PR #6503, opened 2026-04-20
perf(p2p): randomize contact selection for RLPx connection initiation

azteca1998:
## Summary

- Randomize contact selection in `get_contact_to_initiate` to avoid sequential bucket traversal bias

Builds on #6497 (peer pruning fix).

## Problem

After #6458 introduced the Kademlia k-bucket routing table, snap sync times regressed (~50-75% slower across hoodi, sepolia, mainnet). Part of the slowdown comes from `do_get_contact_to_initiate` iterating buckets sequentially from 0→255. Due to XOR distance distribution, ~87% of contacts cluster in buckets 253-255, while buckets 0-200 are nearly empty. Sequential traversal wastes connection attempts on sparse low-numbered buckets before reaching the dense high-numbered ones where most peers live.

## Changes

Replace the sequential `for bucket in &self.buckets` loop with random selection from all eligible contacts across all buckets (main + replacements). This follows the same pattern already used by `get_contact_for_lookup` and `get_contact_for_enr_lookup`.

## Test plan

- [x] All 33 p2p unit tests pass
- [ ] Run daily snapsync test and compare sync times against baseline

### DIFF
--- a/crates/networking/p2p/peer_table.rs
+++ b/crates/networking/p2p/peer_table.rs
@@ -1055,24 +1055,31 @@ impl PeerTableServer {
     }
 
     fn do_get_contact_to_initiate(&mut self) -> Option<Contact> {
-        // Check both main contacts and replacements in each bucket.
-        // Replacements may contain fresher peers that haven't been tried yet.
-        for bucket in &self.buckets {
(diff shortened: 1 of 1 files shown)


azteca1998:
Superseded by #6511 (Kademlia v2) — randomized selection replaced by O(1) random index probe.

PR #6507, opened 2026-04-20
feat(l1): add meta entries + muiltiget for faster receipt retrieval

Arkenan:
**Motivation**

We saw that the method to retrieve a single receipt was taking too much time (seconds). We traced this to the sequential get of all the receipts of a block to calculate the gas used and the log index.

**Description**

This PR adds two mechanisms to solve this:
- Cache the values (non cumulative gas used and log index) in a separate receipt metadata CF. This is written on block execution.
- If receipts for older transactions are requested the receipts are pulled in multiget instead of one by one, and then the meta is calculated and written so future requests will benefit from this as well, specially if requesting from the same block, which some clients might do.


crates/storage/api/tables.rs
```diff
@@ -35,6 +35,15 @@ pub const ACCOUNT_CODE_METADATA: &str = "account_code_metadata";
 /// - [`Vec<u8>`] = `receipt.encode_to_vec()`
 pub const RECEIPTS: &str = "receipts";
 
+/// Per-block receipt metadata column family: [`Vec<u8>`] => [`Vec<u8>`]
+/// - Key: `block_hash.as_bytes()` (32 bytes)
+/// - Value: packed `[u32 gas_used_le | u32 log_index_offset_le]` * N_transactions
```
ElFantasma, line 40: Doc/impl mismatch: this says `Value: packed [u32 gas_used_le | u32 log_index_offset_le]`, but the implementation uses `u64::to_le_bytes()` for both fields and `BLOCK_RECEIPT_META_ENTRY_LEN = 16` (`store.rs:148`). Should read `[u64 gas_used_le | u64 log_index_offset_le]`. (Greptile flagged this.)

crates/networking/rpc/eth/block.rs
```diff
@@ -352,17 +361,18 @@ pub async fn get_all_block_rpc_receipts(
...
+        .iter()
+        .zip(raw_receipts.into_iter())
+        .enumerate()
+    {
         let index = index as u64;
-        let receipt = match storage.get_receipt(block_number, index).await? {
-            Some(receipt) => receipt,
-            _ => return Err(RpcErr::Internal("Could not get receipt".to_owned())),
-        };
         let gas_used = receipt.cumulative_gas_used - last_cumulative_gas_used;
```
ElFantasma, line 376: **`saturating_sub` vs plain `-` inconsistency**: `Receipt::build_block_meta` uses `saturating_sub` for `gas_used`; this path uses plain `-`. Same inputs flow through both paths post-backfill, so a hypothetical pathological cumulative-gas sequence saturates in one and panics in the other. Pre-existing rather than introduced here, but worth aligning while you're touching this code.

crates/networking/rpc/eth/transaction.rs
```diff
@@ -303,15 +304,99 @@ impl RpcHandler for GetTransactionReceiptRequest {
...
+            None => {
+                // One batched MultiGet pulls the whole block's receipts in
+                // roughly the cost of a single SST block read.
+                let all = storage
+                    .get_receipts_by_block_hash(block_hash, tx_count)
+                    .await?;
+                let meta = Receipt::build_block_meta(&all);
+                // Backfill the CF so the next request takes the fast path.
+                // Errors here are non-fatal.
+                let _ = storage.put_block_receipt_meta(block_hash, &meta).await;
```
ElFantasma, line 337: Silent error drop: if RocksDB returns a transient error here, we lose observability *and* the next request re-derives. `tracing::warn!` on `Err` is cheap and non-fatal.

crates/networking/rpc/eth/transaction.rs
```diff
@@ -303,15 +304,99 @@ impl RpcHandler for GetTransactionReceiptRequest {
...
+                };
+                (entry, receipt)
+            }
+            None => {
+                // One batched MultiGet pulls the whole block's receipts in
+                // roughly the cost of a single SST block read.
+                let all = storage
+                    .get_receipts_by_block_hash(block_hash, tx_count)
+                    .await?;
+                let meta = Receipt::build_block_meta(&all);
```
ElFantasma, line 334: Concurrent-request race on lazy backfill: two concurrent `GetTransactionReceiptRequest`s on a pre-meta block both reach this `None` arm, both compute the same vec, both `put_block_receipt_meta`. Idempotent so no correctness issue, but flagging for potential thundering-herd patterns on archival queries.

crates/storage/backend/rocksdb.rs
```diff
@@ -302,6 +302,39 @@ impl StorageReadView for RocksDBReadTx {
...
+            return Ok(Vec::new());
+        }
+        let cf = self
+            .db
+            .cf_handle(table)
+            .ok_or_else(|| StoreError::Custom(format!("Table {} not found", table)))?;
+
+        let results = self
+            .db
+            .batched_multi_get_cf(&cf, keys.iter().map(|k| k.as_slice()), false);
```
ElFantasma, line 320: `sorted=false` is correct for general callers, but in `get_receipts_by_block_hash` keys are constructed from `(block_hash, 0..count)` — already sorted by RLP encoding. Passing `true` lets RocksDB skip the internal sort. Tiny win; not blocking.

crates/storage/store.rs
```diff
@@ -677,6 +678,61 @@ impl Store {
             .map_err(StoreError::from)
     }
 
+    /// Fetch every receipt for a block in a single batched backend call.
+    ///
+    /// Keys share a `block_hash` prefix so RocksDB's MultiGet coalesces them
+    /// into ~one SST-block read rather than the `n` independent reads a
+    /// `join_all` over `get_receipt_by_block_hash` would do.
+    pub async fn get_receipts_by_block_hash(
```
ElFantasma, line 686: Returns `Err` if any receipt is missing — a partially-pruned block fails outright in `GetTransactionReceiptRequest`'s `None` arm and never gets meta backfilled. Probably fine since partial-prune is a separate problem, but worth a thought if pruning lands on the L1 path.

ElFantasma (changes requested):
**Follow-up worth filing**: `eth_getBlockReceipts` (`get_all_block_rpc_receipts` / `get_all_block_receipts`) doesn't use the meta CF — both still iterate from scratch. Could short-circuit via the meta CF for blocks that have it.

PR #6508, opened 2026-04-20
fix(l1,l2): serialize forkchoice_update

iovoid:
**Motivation**

Currently if forkchoice_update is called by two sites simultaneously, we could end up with a corrupted canonical chain.

**Description**

There are two underlying problems:
- forkchoice_update reads the latest block outside the atomic write segment, allowing two threads to read the same 'current latest'
- latest_block_header could end up in an inconsistent state

This PR fixes them by serializing writes with a mutex.

It also adds a regression test.

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -187,6 +187,10 @@ pub struct Store {
     /// Uses FxHashMap for efficient lookups, much smaller than code cache.
     code_metadata_cache: Arc<Mutex<rustc_hash::FxHashMap<H256, CodeMetadata>>>,
 
+    /// Serializes concurrent `forkchoice_update` callers so that the cache
+    /// update and the DB write transaction remain mutually ordered.
+    fcu_lock: Arc<tokio::sync::Mutex<()>>,
+
     background_threads: Arc<ThreadList>,
 }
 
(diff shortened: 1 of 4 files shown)


ilitteri (approved):

crates/storage/store.rs
```diff
@@ -1024,6 +1033,12 @@ impl Store {
                 txn.put(CANONICAL_BLOCK_HASHES, &head_key, &head_value)?;
             }
 
+            // Delete canonical entries above the new head by enumerating each key.
+            // `delete_range` is not safe here: keys are `u64::to_le_bytes()`, and
```
ElFantasma, line 1037: This comment is gold — exactly the kind of trap that bites people six months later.

ElFantasma (commented):
Solid TOCTOU fix. The mutex serializes the cache update + DB write so they can't interleave. The cache rollback on DB-write failure is the right complement — without it, a failed FCU would leave the cache pointing at a non-canonical head.

The regression test is excellent: the BASE=250, EXT=10 choice deliberately spans block 256 to catch any future refactor that tries to replace the per-block delete loop with a `delete_range`, since LE-encoded keys aren't lexicographically monotone. The threshold calibration (4999/5000 hits pre-fix) gives confidence the test would actually catch a regression.

The inline comment about why `delete_range` isn't safe is gold — exactly the kind of trap that bites people six months later.

ElFantasma (approved):

azteca1998 (approved):

PR #6510, opened 2026-04-21
fix(l2): preserve storage backup in undo_last_transaction

avilagaston9:
**Motivation**

`undo_last_transaction` silently fails to restore storage slot changes on L2, corrupting state when transactions are speculatively executed and then rolled back (e.g., Credible Layer integration).

The root cause: `L2Hook::finalize_non_privileged_execution` clears `call_frame_backup` at line 151 to scope its Phase 2 rollback, but this also destroys the SSTORE backup before `BackupHook::finalize_execution` can capture it into `tx_backup`. The result is that `undo_last_transaction` restores account info (balances, nonces) but silently loses all storage slot changes.

**Description**

Instead of clearing the backup, save it with `std::mem::take` before Phase 2 begins, then merge it back after Phase 2 succeeds. This preserves the Phase 2 rollback semantics (on error, only Phase 2 mutations are reverted) while ensuring `BackupHook` sees the complete execution + finalize backup.

**How to Test**

```bash
cargo test -p ethrex-test --test ethrex_tests 'levm::l2_hook'
```

New regression test `undo_last_transaction_restores_storage_slots`: creates a contract with `slot[0] = 42`, writes `0xDEAD`, calls `undo_last_transaction()`, asserts the slot is restored. Without the fix, it remains `0xDEAD`.

**Checklist**

- [x] All existing `l2_hook_tests` pass (5/5)
- [x] New regression test added and verified (fails without fix, passes with fix)
- [x] `cargo fmt` clean
- [x] `cargo clippy` clean (1 pre-existing warning unrelated to this change)


crates/vm/levm/src/call_frame.rs
```diff
@@ -326,8 +326,14 @@ impl CallFrameBackup {
     }
 
     pub fn extend(&mut self, other: CallFrameBackup) {
```
ElFantasma, line 332: The overwrite-on-duplicate behavior here is correct (callers always pass the older/more-original backup as `other`, so its values win for duplicate slots → truly-original is preserved), but that invariant is non-obvious. Worth documenting:

```rust
/// Merges `other` into `self`, per-address. For slots present in both,
/// `other`'s values win. Callers MUST pass the older/more-original backup
/// as `other` so the truly-original value is preserved (matches the
/// `or_insert` semantic in `backup_storage_slot`).
```
Both current call sites (`l2_hook::finalize_non_privileged_execution` passing `execution_backup` and `backup_hook::finalize_execution` passing `pre_execution_backup`) follow this — but a future caller could easily get it wrong without the comment.

ElFantasma (approved):
Just one suggestion. It LGTM

crates/vm/levm/src/call_frame.rs
avilagaston9: Addressed in https://github.com/lambdaclass/ethrex/pull/6510/commits/0c3c45c94cf7d5d34be8aa1cb5c4ee1894b6fa4b

test/tests/levm/l2_hook_tests.rs
```diff
@@ -202,6 +202,37 @@ fn fee_token_lock_only_contract() -> Account {
...
+        coinbase,
+        timestamp: 1000,
+        prev_randao: Some(H256::zero()),
+        chain_id: U256::from(1),
+        base_fee_per_gas: U256::from(base_fee_per_gas),
+        base_blob_fee_per_gas: U256::from(1),
+        gas_price: U256::from(gas_price),
+        tx_max_priority_fee_per_gas: Some(U256::from(gas_price - base_fee_per_gas)),
+        tx_max_fee_per_gas: Some(U256::from(gas_price)),
+        block_gas_limit: gas_limit * 2,
```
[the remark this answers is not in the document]
avilagaston9: Addressed in https://github.com/lambdaclass/ethrex/pull/6510/commits/0c3c45c94cf7d5d34be8aa1cb5c4ee1894b6fa4b

azteca1998 (approved):

ilitteri (approved):

PR #6511, opened 2026-04-21
feat(l1): kademlia k-bucket routing table v2

azteca1998:
## Summary

Replaces the flat peer table with a proper Kademlia k-bucket routing table and iterative convergence lookup (geth-style), while maintaining comparable traffic and sync performance to main.

### Key changes

- **Kademlia k-bucket routing table**: 256 buckets × 16 entries with replacement lists, XOR distance-based routing
- **Flat connection pool**: Decoupled from k-buckets (up to 10K entries) so RLPx connection initiation has a larger candidate pool
- **Iterative convergence lookup**: Seeds from connection pool, queries alpha=3 closest unqueried nodes per timer tick, feeds Neighbors/Nodes responses back into the result set
- **Timer-driven lookup advance**: Each timer tick advances by alpha=3 queries. Responses feed results into the lookup but don't trigger new queries, naturally throttling traffic
- **Scaled lookup interval**: 500ms initial → 10s steady-state (vs main's 100ms→600ms) to account for iterative lookups generating ~16 FindNodes each vs main's 1

### Traffic comparison (Prometheus metrics, 30 min capture)

| Metric | main (srv10, /min) | kademlia (srv3, /min) | Ratio |
|--------|-------------------|-----------------------|-------|
| v4 FindNode (steady) | 131 | 144 | 1.1x |
| v5 FindNode (steady) | 132 | 148 | 1.1x |
| v4 Ping | 61 | 34 | 0.6x |
| v5 Ping | 61 | 60 | 1.0x |
| **Total FindNode** | **263** | **292** | **1.1x** |
| **Total Ping** | **122** | **94** | **0.8x** |

Bootstrap burst (first 5 min): ~630 FindNode/min total, dropping to ~290/min by minute 29.

### Multisync comparison (same hardware, same Docker setup)

Three-network snapsync (mainnet + sepolia + hoodi) on srv3:

| Network | main | kademlia | Delta |
|---------|------|----------|-------|
| hoodi | baseline | −1 min | faster |
| sepolia | baseline | −1 min | faster |
| mainnet | baseline | +2 min | slower |

**Result: within noise.** No regression on sync time despite the architectural change.

Note: Docker multisync uses bridge networking (outbound-only, no inbound P2P), which limits peer count to ~50-60 and disproportionately affects kademlia (pool diversity matters more when 100% of peers come from the pool). Bare-metal performance should be slightly better since inbound connections add peer diversity independent of discovery strategy.

### Bare-metal peer discovery (clean restart, no peer cache)

| Time | Kademlia peers | Main peers |
|------|---------------|------------|
| 1m   | 3             | 8          |
| 5m   | 17            | 34         |
| 10m  | 55            | 61         |
| 20m  | 97            | 88         |

Slower first 5 min, catches up by 10 min, beats main at 20 min.

### Why Kademlia matters

The current flat peer table works well for filling RLPx slots but doesn't provide structured routing. Kademlia gives us:
- **O(log n) routing** for future DHT lookups (e.g., EIP-7594 PeerDAS)
- **Resilience**: k-buckets with replacement lists handle churn better
- **Scalability**: structured routing scales to large networks without random probing
- **Standards compliance**: aligns with devp2p/discv5 spec expectations

### What was explored and removed

Several approaches were tried and removed for generating excessive traffic:
- **Pre-bonding** (pinging discovered nodes before querying): 7x traffic burst
- **Concurrent lookups**: multiplied traffic without proportional peer gains
- **Response-driven advance**: discv5 lookups converged in <200ms, churning 60+ lookups/min
- **Back-to-back lookup chaining**: kept tick permanently active

The final design uses timer-driven advance with a scaled easing curve, naturally throttling traffic while benefiting from structured routing.


crates/networking/p2p/discovery/discv4_handlers.rs
```diff
@@ -343,11 +430,22 @@ impl DiscoveryServer {
...
+
+        // Feed results into ALL active lookups (but don't advance — the timer
+        // drives lookup progress so that traffic stays controlled).
+        if let Some(discv4) = &mut self.discv4 {
+            let entries: Vec<(H256, Node)> =
+                nodes.iter().map(|n| (n.node_id(), n.clone())).collect();
+            for (lookup, _) in &mut discv4.active_lookups {
+                lookup.feed_results(entries.clone());
+            }
+            if let Some((lookup, _)) = discv4.active_lookups.first_mut() {
```
ElFantasma, line 444: **`record_response` attribution mismatch.** `feed_results` is called on EVERY active lookup just above (line 437-441), but `record_response` is called only on `active_lookups.first_mut()`. So if multiple lookups are in flight, only the first one's `queries_in_flight` counter is decremented for any Neighbors response, regardless of which lookup the FindNode actually came from.

crates/networking/p2p/peer_table.rs
```diff
@@ -963,13 +1054,104 @@ impl PeerTableServer {
...
+    }
+
+    /// Insert a node into the flat connection pool for RLPx initiation.
+    /// Evicts the oldest entry when the pool is at capacity.
+    fn insert_to_connection_pool(&mut self, node_id: H256, node: Node) {
+        if self.connection_pool.contains_key(&node_id) {
+            return;
+        }
+        if self.connection_pool.len() >= MAX_CONNECTION_POOL_SIZE {
+            self.connection_pool.shift_remove_index(0);
```
ElFantasma, line 1112: **`shift_remove_index(0)` is O(N) per overflow.** Once the pool reaches the 10K cap, every new insert pays a 10K-element shift to drop the front entry. Under churn (e.g., bootstrap burst, peer rotation), that's a real cost on the hot path of every newly-discovered contact.

`IndexMap` has `swap_remove_index(0)` which is O(1) but breaks insertion order — and the LRU-like eviction here depends on insertion order to drop the OLDEST entry. So `swap_remove_index` would evict the wrong thing.

Two cleaner shapes:
1. **`VecDeque<(H256, Node)>` + separate `FxHashMap<H256, usize>` index**: `pop_front()` is O(1), `push_back()` is O(1) amortized. Indices in the map need updating on pop, but only ~once per overflow, not per access.
2. **Ring-buffer with monotonic head/tail counters**: pure O(1), no shifting, no map index updates.

Not urgent — under steady state the pool fills slowly. But under bootstrap or recovery the cost is real (the PR description's "~630 FindNode/min during bootstrap" implies a steady stream of inserts the moment the pool hits cap).

ElFantasma (approved):
There is no CHANGELOG.md entry for this change in the diff. For a routing-table architecture migration this size, an entry under a Perf or Refactor section (with a one-liner pointing at the structural shift) would help anyone bisecting a future P2P regression understand the cutover. Easy to add as a follow-up.

edg-l (approved):

ilitteri (approved):

PR #6516, opened 2026-04-21
feat(l1): fix EIP-8025 compliance

jsign:
## Summary

This branch fixes the EIP-8025 guest-program execution flow and aligns the SSZ-side `NewPayloadRequest` model with the Electra/CL spec.

At a high level, it adds a byte-oriented EIP-8025 execution entrypoint, moves execution requests into a typed SSZ container, and uses that typed representation to reconstruct the EL `requests_hash` correctly when converting a `NewPayloadRequest` into a block.

## What changed

- Added `execution_program_eip8025_bytes`, which decodes and executes the EIP-8025 wire format directly:
  - `[ssz_len: u32 LE][ssz_bytes][rkyv_bytes]`
- Refactored the EIP-8025 validation flow so both entrypoints share the same logic for:
  - `NewPayloadRequest -> Block` conversion
  - block hash validation
  - blob versioned hash validation
  - stateless execution via `execute_blocks`
- Updated the EIP-8025 SSZ types in `ethrex-common` to better match the Electra consensus spec:
  - `ExecutionPayload` no longer carries deposit/withdrawal/consolidation request lists
  - `NewPayloadRequest.execution_requests` is now a typed `ExecutionRequests` container
  - spec constants were renamed/normalized for clarity
  - the consolidation request limit was corrected to the Electra value
- Added `ExecutionRequests::to_encoded_requests()` to convert the SSZ typed request container into the EIP-7685 encoded request list expected by `compute_requests_hash`
- Re-exported the new EIP-8025 byte entrypoint from `ethrex-guest-program`

## Behavior notes

- Malformed EIP-8025 wire input still returns an error
- For well-formed input, the byte-entrypoint now always computes the `new_payload_request_root`
- If validation/execution fails after decoding, the byte-entrypoint returns `ProgramOutput { valid: false, ... }` instead of failing outright

## Why this matters

This makes the EIP-8025 guest path closer to the consensus-side structure we actually need to prove against, while also giving us a direct raw-bytes execution entrypoint for the EIP-8025 wire format.

It also removes the mismatch between the SSZ request representation and the EL request-hash computation path by deriving `requests_hash` from a properly typed `ExecutionRequests` container.

## External usage

I'm already using this PR branch into an ere-guests PR here: https://github.com/eth-act/ere-guests/pull/39
This allowed `ere-guests` to basically remove a lot of wrapper code before calling the guest program regarding execution-payload->Block, output calculation, etc -- making now Ethrex own all that logic as expected.

### DIFF
--- a/crates/common/types/eip8025_ssz.rs
+++ b/crates/common/types/eip8025_ssz.rs
@@ -4,33 +4,38 @@
 //! `NewPayloadRequest` and producing the `PublicInput` committed to by
 //! execution proofs.
 
+use bytes::Bytes;
 use libssz::{SszDecode, SszEncode};
 use libssz_derive::{HashTreeRoot, SszDecode, SszEncode};
 use libssz_merkle::{HashTreeRoot, Sha256Hasher};
 use libssz_types::{SszList, SszVector};
 
+use super::requests::EncodedRequests;
+
 // ── Spec limits (Electra) ──────────────────────────────────────────
 
-/// MAX_TRANSACTIONS_PER_PAYLOAD (Electra).
-const MAX_TRANSACTIONS: usize = 1_048_576;
-/// MAX_WITHDRAWALS_PER_PAYLOAD (Electra).
-const MAX_WITHDRAWALS: usize = 16;
-/// MAX_BYTES_PER_TRANSACTION.
+/// `MAX_TRANSACTIONS_PER_PAYLOAD` (Electra).
+const MAX_TRANSACTIONS_PER_PAYLOAD: usize = 1_048_576;
+/// `MAX_WITHDRAWALS_PER_PAYLOAD` (Electra).
+const MAX_WITHDRAWALS_PER_PAYLOAD: usize = 16;
+/// `MAX_BYTES_PER_TRANSACTION`.
 const MAX_BYTES_PER_TRANSACTION: usize = 1_073_741_824;
-/// MAX_EXTRA_DATA_BYTES.
+/// `MAX_EXTRA_DATA_BYTES`.
 const MAX_EXTRA_DATA_BYTES: usize = 32;
-/// MAX_DEPOSIT_REQUESTS_PER_PAYLOAD (Electra).
-const MAX_DEPOSIT_REQUESTS: usize = 8192;
-/// MAX_WITHDRAWAL_REQUESTS_PER_PAYLOAD (Electra).
-const MAX_WITHDRAWAL_REQUESTS: usize = 16;
-/// MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD (Electra).
-const MAX_CONSOLIDATION_REQUESTS: usize = 1;
-/// MAX_BLOB_COMMITMENTS_PER_BLOCK (Electra).
-const MAX_BLOB_COMMITMENTS: usize = 4096;
-/// MAX_EXECUTION_REQUESTS (EIP-7685).
-const MAX_EXECUTION_REQUESTS: usize = 16;
-/// MAX_EXECUTION_REQUEST_BYTES.
-const MAX_EXECUTION_REQUEST_BYTES: usize = 1_073_741_824;
+/// `MAX_DEPOSIT_REQUESTS_PER_PAYLOAD` (Electra).
+const MAX_DEPOSIT_REQUESTS_PER_PAYLOAD: usize = 8192;
+/// `MAX_WITHDRAWAL_REQUESTS_PER_PAYLOAD` (Electra).
+const MAX_WITHDRAWAL_REQUESTS_PER_PAYLOAD: usize = 16;
+/// `MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD` (Electra).
+const MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD: usize = 2;
+/// `MAX_BLOB_COMMITMENTS_PER_BLOCK` (Electra).
+const MAX_BLOB_COMMITMENTS_PER_BLOCK: usize = 4096;
+
+// ── EIP-7685 request type prefixes ─────────────────────────────────
+
+const DEPOSIT_REQUEST_TYPE: u8 = 0x00;
+const WITHDRAWAL_REQUEST_TYPE: u8 = 0x01;
+const CONSOLIDATION_REQUEST_TYPE: u8 = 0x02;
 
 // ── Bytes20 wrapper (address) ──────────────────────────────────────
 //
@@ -95,7 +100,7 @@ impl From<Bytes20> for [u8; 20] {
 // `logs_bloom` is `ByteVector[BYTES_PER_LOGS_BLOOM]` in the CL spec —
 // a fixed-length SSZ vector of 256 bytes.
 
-/// BYTES_PER_LOGS_BLOOM from the CL spec.
+/// `BYTES_PER_LOGS_BLOOM` from the CL spec.
 pub const BYTES_PER_LOGS_BLOOM: usize = 256;
 
 /// `ByteVector[256]` — the logs bloom as a fixed-size SSZ vector.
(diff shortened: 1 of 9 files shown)


crates/common/types/eip8025_ssz.rs
```diff
@@ -4,33 +4,38 @@
...
 use libssz::{SszDecode, SszEncode};
 use libssz_derive::{HashTreeRoot, SszDecode, SszEncode};
 use libssz_merkle::{HashTreeRoot, Sha256Hasher};
 use libssz_types::{SszList, SszVector};
 
+use super::requests::EncodedRequests;
+
 // ── Spec limits (Electra) ──────────────────────────────────────────
 
-/// MAX_TRANSACTIONS_PER_PAYLOAD (Electra).
```
jsign, line 14: Some renamings to conform to the [consensus-spec names](https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md), but also some fixes that I'll comment in particular.

crates/common/types/eip8025_ssz.rs
```diff
@@ -4,33 +4,38 @@
...
 const MAX_BYTES_PER_TRANSACTION: usize = 1_073_741_824;
-/// MAX_EXTRA_DATA_BYTES.
+/// `MAX_EXTRA_DATA_BYTES`.
 const MAX_EXTRA_DATA_BYTES: usize = 32;
-/// MAX_DEPOSIT_REQUESTS_PER_PAYLOAD (Electra).
-const MAX_DEPOSIT_REQUESTS: usize = 8192;
-/// MAX_WITHDRAWAL_REQUESTS_PER_PAYLOAD (Electra).
-const MAX_WITHDRAWAL_REQUESTS: usize = 16;
-/// MAX_CONSOLIDATION_REQUESTS_PER_PAYLOAD (Electra).
-const MAX_CONSOLIDATION_REQUESTS: usize = 1;
```
jsign, line 27: Note that the previous `1` value wasn't correct, this was fixed below L30 with value `2` as [defined in the specs](https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#execution).

crates/common/types/eip8025_ssz.rs
```diff
@@ -158,41 +163,52 @@ pub struct ExecutionPayload {
...
     pub block_hash: [u8; 32],
-    pub transactions: SszList<SszList<u8, MAX_BYTES_PER_TRANSACTION>, MAX_TRANSACTIONS>,
-    pub withdrawals: SszList<Withdrawal, MAX_WITHDRAWALS>,
+    pub transactions: SszList<SszList<u8, MAX_BYTES_PER_TRANSACTION>, MAX_TRANSACTIONS_PER_PAYLOAD>,
+    pub withdrawals: SszList<Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD>,
     pub blob_gas_used: u64,
     pub excess_blob_gas: u64,
-    pub deposit_requests: SszList<DepositRequest, MAX_DEPOSIT_REQUESTS>,
-    pub withdrawal_requests: SszList<WithdrawalRequest, MAX_WITHDRAWAL_REQUESTS>,
-    pub consolidation_requests: SszList<ConsolidationRequest, MAX_CONSOLIDATION_REQUESTS>,
```
jsign, line 167: In the specs, `ExecutionPayload` does not have these three fields. They are actually part of the `NewPayloadRequest`.

crates/common/types/eip8025_ssz.rs
```diff
@@ -158,41 +163,52 @@ pub struct ExecutionPayload {
...
 
-// ── ExecutionPayloadHeader ─────────────────────────────────────────
+// ── ExecutionRequests ──────────────────────────────────────────────
 
-/// Headerized version of `ExecutionPayload`: variable-length lists are
-/// replaced by their `hash_tree_root`.
+/// SSZ `ExecutionRequests` container (Electra) — the typed EIP-7685 bundle
+/// that the CL commits to alongside `ExecutionPayload`.
 #[derive(Debug, Clone, PartialEq, Eq, SszEncode, SszDecode, HashTreeRoot)]
-pub struct ExecutionPayloadHeader {
```
jsign, line 175: This isn't strictly required so we can remove.

crates/common/types/eip8025_ssz.rs
```diff
@@ -158,41 +163,52 @@ pub struct ExecutionPayload {
...
-    pub base_fee_per_gas: [u8; 32],
-    pub block_hash: [u8; 32],
-    pub transactions_root: [u8; 32],
-    pub withdrawals_root: [u8; 32],
-    pub blob_gas_used: u64,
-    pub excess_blob_gas: u64,
-    pub deposit_requests_root: [u8; 32],
-    pub withdrawal_requests_root: [u8; 32],
-    pub consolidation_requests_root: [u8; 32],
+pub struct ExecutionRequests {
```
jsign, line 177: See my next comment to understand this.

crates/common/types/eip8025_ssz.rs
```diff
@@ -202,20 +218,9 @@ pub struct ExecutionPayloadHeader {
...
-/// Headerized version of `NewPayloadRequest`.
-#[derive(Debug, Clone, PartialEq, Eq, SszEncode, SszDecode, HashTreeRoot)]
-pub struct NewPayloadRequestHeader {
-    pub execution_payload_header: ExecutionPayloadHeader,
-    pub versioned_hashes: SszList<[u8; 32], MAX_BLOB_COMMITMENTS>,
+    pub versioned_hashes: SszList<[u8; 32], MAX_BLOB_COMMITMENTS_PER_BLOCK>,
     pub parent_beacon_block_root: [u8; 32],
-    pub execution_requests:
-        SszList<SszList<u8, MAX_EXECUTION_REQUEST_BYTES>, MAX_EXECUTION_REQUESTS>,
+    pub execution_requests: ExecutionRequests,
```
jsign, line 223: The consensus-specs [don't define `execution_requests` as a list of bytes, but an `ExecutionRequests` deserialized struct](https://github.com/ethereum/consensus-specs/blob/master/specs/electra/beacon-chain.md#modified-newpayloadrequest).

Fixing it here.

crates/guest-program/src/l1/program.rs
```diff
@@ -72,36 +72,36 @@ pub fn execution_program(
...
+/// Decode and execute the L1 stateless validation program from EIP-8025 wire
+/// bytes.
+///
+/// The wire format is `[ssz_len: u32 LE][ssz_bytes][rkyv_bytes]`, matching
+/// [`decode_eip8025`](super::decode_eip8025).
+#[cfg(feature = "eip-8025")]
+pub fn execution_program_eip8025_bytes(
+    bytes: &[u8],
+    crypto: Arc<dyn Crypto>,
+) -> Result<ProgramOutput, ExecutionError> {
```
jsign, line 66: This is a new entrypoint which is aligned [with the execution-specs](https://github.com/ethereum/execution-specs/blob/713d1eed0d3b730b7d3c72c961fc4d7ed0b8d656/src/ethereum/forks/amsterdam/stateless_guest.py#L33).

In `ere-guests` now we use it to send raw bytes directly, removing a lot of "wrapper logic" we had in ere-guest guest program before calling Ethrex entry point.

crates/guest-program/src/l1/program.rs
```diff
@@ -72,36 +72,36 @@ pub fn execution_program(
...
 
-    // Validate blob versioned hashes
-    validate_versioned_hashes(&block, &new_payload_request)?;
+/// Decode and execute the L1 stateless validation program from EIP-8025 wire
+/// bytes.
+///
+/// The wire format is `[ssz_len: u32 LE][ssz_bytes][rkyv_bytes]`, matching
+/// [`decode_eip8025`](super::decode_eip8025).
+#[cfg(feature = "eip-8025")]
+pub fn execution_program_eip8025_bytes(
```
avilagaston9: Thanks for the PR! Looks good to me.

One thing: since `execution_program` is already under the `eip-8025` feature flag, I'd keep only the bytes version and drop the typed one. I think there's no reason to maintain the typed entrypoint.

WDYT @ilitteri?

jsign: Yep, I was thinking about the same when making the PR but prefered to be conservative in removing existing APIs. 

Leaving that decision to you guys!

ilitteri: I agree, let's keep the bytes version only.

jsign: Sounds good. I'll adjust and reping here when ready.

jsign: @avilagaston9 @ilitteri, all done see https://github.com/lambdaclass/ethrex/pull/6516/commits/0058d278302e4b44616cecaf2560a52bd5f3b5af

avilagaston9 (approved):

ilitteri (approved):

jsign:
@ilitteri, is this mergeable? :)

ilitteri:
> @ilitteri, is this mergeable? :)

We need one more approval. I'd get it asap.

edg-l (approved):

azteca1998 (approved):

PR #6517, opened 2026-04-22
fix(l1): prevent leak of peer request slots

iovoid:
**Motivation**

Currently we are doing `inc_requests` manually, but on staleness exit or failure the task that decrements the counter isn't executed.

**Description**

Before exiting, reap tasks.


iovoid:
Superceded by #6523

PR #6519, opened 2026-04-23
feat(l1): add automatic DB migration framework

azteca1998:
## Summary

- Adds a migration framework to `ethrex-storage` so schema version bumps run upgrade functions automatically instead of forcing a full resync
- `MIGRATIONS` array in `migrations.rs` maps each version step (index 0 = v1→v2, index 1 = v2→v3, etc.) with a compile-time assert to keep it in sync with `STORE_SCHEMA_VERSION`
- `Store::new` detects the on-disk version and runs pending migrations before proceeding, writing metadata with fsync after each step for crash safety
- `has_valid_db` now accepts any migratable version (`1..=STORE_SCHEMA_VERSION`)

### Files changed

| File | Change |
|------|--------|
| `crates/storage/migrations.rs` | **New** — `MigrationFn` type, `MIGRATIONS` array, `run_pending_migrations()`, unit tests |
| `crates/storage/lib.rs` | Add `pub mod migrations`, update docstring |
| `crates/storage/store.rs` | Replace `validate_store_schema_version` → `read_store_schema_version`, restructure `Store::new` with migration support, update `has_valid_db` |
| `crates/storage/error.rs` | Add `MigrationFailed` variant |
| `cmd/ethrex/initializers.rs` | Catch `MigrationFailed` at startup |

### How to add a future migration

1. Write `fn migrate_1_to_2(backend: &dyn StorageBackend) -> Result<(), StoreError>` in `migrations.rs`
2. Append to `MIGRATIONS` array (index 0)
3. Bump `STORE_SCHEMA_VERSION` to `2` in `lib.rs`
4. Compile-time assert ensures array length matches

## Test plan

- [x] `cargo check -p ethrex-storage` — clean
- [x] `cargo check -p ethrex` — clean
- [x] `cargo test -p ethrex-storage` — 3 unit tests pass (array length, no-op migration, metadata correctness)


crates/storage/migrations.rs
```diff
@@ -0,0 +1,129 @@
...
+        tracing::info!("Running migration v{version} → v{target}");
+
+        MIGRATIONS[idx](backend).map_err(|e| StoreError::MigrationFailed {
+            from: version,
+            to: target,
+            reason: e.to_string(),
+        })?;
+
+        // Persist the new version to metadata.json after each migration step
+        write_metadata_version(db_path, target)?;
```
iovoid: We should also wrap inside MigrationFailed

iovoid (approved):
It probably doesn't matter, but update of the db and metadata file aren't atomic.

edg-l (approved):

crates/storage/store.rs
```diff
@@ -1451,12 +1451,49 @@ impl Store {
...
+                    });
+                }
+                None => {
+                    // Fresh / empty directory — write initial metadata
+                    init_metadata_file(&db_path)?;
+                }
+                Some(v) if v < 1 => {
+                    return Err(StoreError::IncompatibleDBVersion {
+                        found: v,
+                        expected: STORE_SCHEMA_VERSION,
```
Arkenan: Shouldn't this be a specific migration error? ie, something like 

```rust
MigrationFailed { from: v, to: STORE_SCHEMA_VERSION, reason: "DB version {from} is previous to the existence of migrations." }
```

crates/storage/store.rs
```diff
@@ -1451,12 +1451,49 @@ impl Store {
...
+                    });
+                }
+                Some(v) if v > STORE_SCHEMA_VERSION => {
+                    return Err(StoreError::IncompatibleDBVersion {
+                        found: v,
+                        expected: STORE_SCHEMA_VERSION,
+                    });
+                }
```
Arkenan, line 1485: Similar, the message could be something like

```rust
MigrationFailed { from: v, to: STORE_SCHEMA_VERSION, reason: "DB version is more recent than the one the client expects. Migrating would involve rolling back, which is not supported." }
```

crates/storage/store.rs
```diff
@@ -1451,12 +1451,49 @@ impl Store {
...
-            // Check that the last used DB version matches the current version
-            validate_store_schema_version(&db_path)?;
+            let version = read_store_schema_version(&db_path)?;
+
+            match version {
+                None if db_path.exists() && !dir_is_empty(&db_path)? => {
+                    // Pre-metadata DB — cannot migrate safely
+                    return Err(StoreError::NotFoundDBVersion {
+                        expected: STORE_SCHEMA_VERSION,
+                    });
```
Arkenan: Is it possible to have this more explicit somehow? It's not that we expect `STORE_SCHEMA_VERSION` specifically, we can't migrate this at all because there's no version data in the db at all. We should output something like

```
Cannot migrate the database. It's version is unavailable, which means that it predates versioning and migrations.
```

we may just remove the "expected" part of the message entirely, as we are not really expecting that specific version now, we're expecting any migratable version.

crates/storage/migrations.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Returns `Ok(())` if `current_version == STORE_SCHEMA_VERSION` (no-op).
+pub fn run_pending_migrations(
+    backend: &dyn StorageBackend,
+    db_path: &Path,
+    current_version: u64,
+) -> Result<(), StoreError> {
+    assert!(
+        MIGRATIONS.len() == (STORE_SCHEMA_VERSION - 1) as usize,
+        "MIGRATIONS length must equal STORE_SCHEMA_VERSION - 1"
+    );
```
Arkenan: Isn't this done before as a compile time check just above? I don't think we need to do it at runtime again.

crates/storage/migrations.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    for version in current_version..STORE_SCHEMA_VERSION {
+        let idx = (version - 1) as usize; // v1→v2 is index 0
```
Arkenan: nit: let's have a `migration_for_version(i)` function that returns the correct migration for that version. That way we don't think about the indexing tricks in here..

crates/storage/migrations.rs
```diff
@@ -0,0 +1,133 @@
...
+        tracing::info!("Running migration v{version} → v{target}");
+
+        MIGRATIONS[idx](backend).map_err(|e| StoreError::MigrationFailed {
+            from: version,
+            to: target,
+            reason: e.to_string(),
+        })?;
+
+        // Persist the new version to metadata.json after each migration step
+        write_metadata_version(db_path, target).map_err(|e| StoreError::MigrationFailed {
```
Arkenan, line 65: Shouldn't this be part of a storage abstraction instead? Having to pass the path_db around is a bit weird.

Arkenan (approved):
LGTM with some nit considerations. We can merge as is but if we do so let's start a PR fixing these issues.

PR #6518, opened 2026-04-23
feat(l1): support bal-devnet-4 (bal@v5.7.0)

edg-l:
Brings ethrex to bal@v5.7.0 (bal-devnet-4) spec compliance. Three commits on top of main covering 9 EIPs, builder/validator parity fixes, and regression tests.

### EIP-7928 (BAL)

- Widen `BlockAccessIndex` `u16 -> u32`.
- Shadow BAL recorder on per-tx `tx_db` in parallel validator with missing-entry detection across touched_addresses, storage_reads, storage_changes, code_changes.
- Stop whitelisting `SYSTEM_ADDRESS` from `unaccessed_pure_accounts`.
- PART B pre-state fallback for missing `code_changes`.

### EIP-8037 (state gas 2D)

- Dynamic `cost_per_state_byte(block_gas_limit)`; `cpsb(120_000_000) == 1174`.
- Two-counter reservoir (`state_gas_spill_outstanding` + `state_gas_credit_against_drain`) for correct revert math across nested sub-calls (aligns with PR #2733).
- Per-tx 2D inclusion check in `fill_transactions` (PR #2703).
- Immutable `intrinsic_state_gas` (PR #2711).
- CREATE failure state-gas refund; same-tx SELFDESTRUCT refund clamped against execution-only state gas (PR #2707).
- Top-level reservoir reset on tx failure (PR #2689); revert-path X-Z formula.
- Zero `gas_remaining` on precompile exceptional halt.

### EIP-7976 / EIP-7981

- Calldata floor 10->16 gas-per-token under Amsterdam.
- Access-list data bytes (20/addr, 32/key) fold into floor-token count.

### EIP-7708

- Lex-ordered burn logs, no coinbase priority-fee log, SELFDESTRUCT-dest coalescing.

### Builder/validator parity (commit `2613d549f`)

Three miss-slot risks fixed:

- Mempool `transaction_intrinsic_gas` was using `TX_CREATE_GAS_COST = 53000` unconditionally; routed through `intrinsic_gas_dimensions` (re-exported from `ethrex-vm`) for Amsterdam+ so mempool admission matches VM charge.
- `fill_transactions` had no EIP-8037 2D inclusion check; `check_2d_gas_allowance` is now `pub` and called before any BAL touches so rejected txs contribute nothing.
- L2 payload builder missing BAL checkpoint/restore around `undo_last_tx`; added `bal_checkpoint` + `tx_restore` on both rejection paths.

### Tests (commit `13422b493`)

[...]

PR #6520, opened 2026-04-23
test(l1): bal-devnet-4 follow-up regression guards + doc polish

edg-l:
Follow-up to #6518. Addresses the test-gap list from the session-3 review. Every item in `TODO.md` is now covered except the upstream zkevm@v0.4.x fixture re-enable, which is tracked externally.

Stacked on #6518 — this PR's base is \`bal-devnet-4-pr\`, so the diff shown is only the 1-commit delta. Land after #6518.

## Summary

- **10 new regression tests** covering cpsb quantization boundaries, BAL u32 RLP wire format at high indices, mempool Amsterdam CREATE admission, `intrinsic_gas_dimensions`-vs-VM parity across forks, and CALL-to-empty state-gas retention on parent success.
- **Code polish**: \`debug_assert\` guards on \`BlockAccessIndex\` tx-count invariant at each block-exec entry; clarifying comment on the \`frame_outstanding_delta\` subtraction in \`credit_state_gas_refund\`.
- **Docs**: \`forks-roadmap.md\` updated — EIP-7976 / EIP-7981 flipped 🔴→✅, EIP-8037 status expanded, priority note refreshed for bal-devnet-4.

## New tests

| Area | Tests |
|---|---|
| cpsb formula boundaries | \`test_cpsb_clamp_to_one_for_tiny_gas_limit\`, \`test_cpsb_30m_bin_boundary\` |
| EIP-7928 u32 RLP | \`test_change_variants_rlp_roundtrip_index_above_u16_max\` |
| Mempool Amsterdam CREATE | \`amsterdam_create_intrinsic_matches_vm_dimensions\` |
| Standalone ↔ VM intrinsic parity | \`test_intrinsic_parity_plain_transfer\`, \`test_intrinsic_parity_create_tx\`, \`test_intrinsic_parity_with_calldata_and_access_list\`, \`test_intrinsic_parity_eip7702_auth_list\` |
| CALL-to-empty + parent success | \`test_call_to_empty_account_with_value_retains_parent_state_gas\` |

Every test is deterministic, isolates one spec rule, and fails loudly if the corresponding invariant regresses.

## Why each matters

- **cpsb boundaries**: the quantization formula has two failure modes — off-by-one at the \`quantized > CPSB_OFFSET\` cutoff (returns 1 incorrectly for low gas limits) and bin-boundary surprises as \`raw\` increments by 1. Tests fix both pre-conditions.
- **u32 RLP**: devnet-4 widened \`BlockAccessIndex\` from \`u16\` to \`u32\` (PR #2730). An accidental revert would silently truncate indices above 65535, scrambling the BAL for busy blocks.
- **Mempool CREATE**: the original squashed rollup had the legacy \`TX_CREATE_GAS_COST = 53000\` leaking into Amsterdam admission (fixed in #6518 via \`intrinsic_gas_dimensions\`). Test pins the new behavior so a future refactor can't regress.
- **Standalone ↔ VM parity**: \`intrinsic_gas_dimensions\` runs in two places (mempool, payload builder) with no VM instance. If it ever drifts from \`VM::get_intrinsic_gas\`, txs would pass one gate but fail the other.
- **CALL-to-empty**: EIP-8037 charges new-account state gas in the PARENT frame before entering the child. Pairs with the existing revert-direction test for symmetric coverage.

## Test plan

- [x] \`cargo test -p ethrex-test --test ethrex_tests\` — 478 pass / 0 fail.
- [x] \`cargo fmt --all\` clean.
- [x] \`cargo clippy --workspace --lib\` clean (no new warnings introduced).
- [ ] Hive re-run optional — no runtime behavior change.

No behavior changes. This PR is purely test coverage, one invariant comment, one debug-only assertion pair, and a doc update.

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -158,6 +158,15 @@ impl LEVM {
         let chain_config = db.store.get_chain_config()?;
         let is_amsterdam = chain_config.is_amsterdam_activated(block.header.timestamp);
 
+        // EIP-7928 BlockAccessIndex is uint32. Block validity forbids >= 2^32 txs
+        // long before we'd reach this point, but guard the invariant explicitly
+        // so any upstream bug that inflates tx counts panics in debug instead of
+        // silently producing a `u32::MAX` index.
+        debug_assert!(
+            block.body.transactions.len() < u32::MAX as usize,
+            "tx count overflows u32 BlockAccessIndex"
+        );
+
         // Enable BAL recording for Amsterdam+ forks
         if is_amsterdam {
             db.enable_bal_recording();
@@ -331,6 +340,12 @@ impl LEVM {
     ) -> Result<(BlockExecutionResult, Option<BlockAccessList>), EvmError> {
         let chain_config = db.store.get_chain_config()?;
         let is_amsterdam = chain_config.is_amsterdam_activated(block.header.timestamp);
+
+        // EIP-7928 BlockAccessIndex invariant — see `execute_block` for rationale.
+        debug_assert!(
+            block.body.transactions.len() < u32::MAX as usize,
+            "tx count overflows u32 BlockAccessIndex"
+        );
 
         let transactions_with_sender =
             block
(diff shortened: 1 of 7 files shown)


edg-l:
Folding these commits into PR #6518 instead of a stacked PR. Sorry for the noise.

PR #2711, opened 2025-05-08
chore(l1): fix precompile tests from EIP-7702

DiegoCivi:
**Motivation**

On #2586  new tests were added and some of them regarding the [EIP-7702](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md) and precompiles were failing in LEVM.

**Description**

- Tests from file `eip7702_set_code_tx/set_code_txs/set_code_to_precompile.json` were failing because we were not transferring WEI when the target address of a delegation was one of the precompiles. As a result, balances were not updated, causing an incorrect state root.
- Tests from file `eip7702_set_code_tx/set_code_txs/set_code_to_precompile_not_enough_gas_for_precompile_execution.json` were failing because when a precompile was the target of a delegation, and its execution failed a revert was returned. This caused an incorrect receipts root because if a precompile fails in this context, a success should be returned.

Closes #2691

PR #2703, opened 2025-05-08
docs(levm): added Rust docs and modified handle_return function to make it more d…

SDartayet:
**Motivation**

Making the code easier to understand.

**Description**

Some comments were added to the main VM code. Some Rust docs were added to various data structures too to make them easier to understand.

PR #6521, opened 2026-04-23
feat(l1): support bal-devnet-4 (bal@v5.7.0)

edg-l:
Brings ethrex up to `bal@v5.7.0` (bal-devnet-4). Replaces closed PR #6518 — that one was accidentally opened on the long-lived `bal-devnet-4` branch that ethpanda tracks; this one uses the dedicated `bal-devnet-4-pr` branch so the PR lifecycle doesn't disturb the rolling progress branch.

## What bal-devnet-4 requires

The spec update rolls up five EIPs that interlock:

- **EIP-7928** — Block Access List, `BlockAccessIndex` widened `u16 → u32`.
- **EIP-8037** — two-dimensional gas accounting for state creation, with dynamic `cost_per_state_byte`, per-frame reservoir, clamp-and-spill refunds, per-tx 2D inclusion check, immutable intrinsic state gas, same-tx SELFDESTRUCT refund, and top-level reservoir reset on tx failure.
- **EIP-7976** — calldata floor raised from 10 to 16 gas per token (effective 64 gas/byte).
- **EIP-7981** — access-list data bytes (20 per address, 32 per storage key) fold into the floor-token count.
- **EIP-7708** — transfer + burn logs at SYSTEM_ADDRESS with lexicographic ordering.

Against bal@v5.6.1 the notable deltas are the u32 index widening, dynamic cpsb (previously a static 1174), immutable intrinsic state gas (PR #2711), and the per-tx 2D inclusion check (PR #2703). Fixtures were regenerated accordingly.

## Results

- Hive `eels/consume-engine` amsterdam: **1342 pass / 0 fail** (from 10 failing at the start).
- ef_tests/blockchain: **8781 pass** (6 EIP-8025 `witness_codes_*` zkevm@v0.3.x fixtures skipped — filled against bal@v5.6.1 pre-#2711, explicit skip in `tooling/ef_tests/blockchain/tests/all.rs`, re-enable when zkevm@v0.4.x ships).
- Full LEVM + blockchain unit test suite: **478 pass**.

## Builder/validator parity

The bigger-than-expected part of this PR is making sure ethrex-as-builder and ethrex-as-validator agree on every Amsterdam output. An audit turned up three miss-slot risks on the builder side (Amsterdam CREATE intrinsic gas in the mempool, missing 2D inclusion check in `fill_transactions`, missing BAL checkpoint/restore in the L2 builder). All three are fixed.

To keep this from regressing silently, there's a new test file, `test/tests/blockchain/builder_validator_parity_tests.rs`, that builds a block via the payload builder and then runs the built block+BAL through the validator pipeline. The module doc explains the rationale and what a failure means (P0: would cause a missed slot on devnet). Nine positive scenarios exercise the shared path; five negative scenarios corrupt the BAL deterministically and assert the validator rejects — each mirrors one of the Hive `test_bal_invalid_*` cases we fixed in this PR.

## Where to focus review

- `crates/vm/backends/levm/mod.rs` — biggest diff. Parallel validator with the shadow BAL recorder, PART B validation interactions, SYSTEM_ADDRESS handling.
- `crates/vm/levm/src/vm.rs` — new state-gas counters + CallFrame snapshot fields for the EIP-8037 reservoir model.
- `crates/vm/levm/src/gas_cost.rs::cost_per_state_byte` — formula implementation, matched against EELS. Sanity: `cpsb(120_000_000) == 1174`.
- `crates/vm/levm/src/utils.rs` — `get_intrinsic_gas` (VM) and `intrinsic_gas_dimensions` (standalone, used by mempool + builder) must stay in sync. Parity is checked by `test_intrinsic_parity_*`.
- `crates/blockchain/payload.rs::fill_transactions` — 2D inclusion check + BAL checkpoint semantics.
- `test/tests/blockchain/builder_validator_parity_tests.rs` — the primary regression guard for everything else.

## Relationship to PR #6463

#6463 (withdrawal-phase BAL reverse check) merged into main before this branch was rebased. This PR builds on its `validate_bal_withdrawal_index` Part B; the u32 widening + #6463's new `&BalAddressIndex` parameter are merged into a combined signature. Not re-introducing #6463's changes.

## Test plan

- [x] `make -C tooling/ef_tests/blockchain test`
- [x] `make run-hive-eels-amsterdam`
- [x] `cargo test -p ethrex-test --test ethrex_tests`
- [x] `cargo fmt --all` + `cargo check --workspace --lib`

## Pins

- Fixtures: [bal@v5.7.0](https://github.com/ethereum/execution-spec-tests/releases/tag/bal%40v5.7.0)
- EELS: `524b44617e410ab21b5122f0be5113b62a0e76ee` on branch `devnets/bal/4`


crates/vm/levm/src/utils.rs
```diff
@@ -590,6 +636,163 @@ impl<'a> VM<'a> {
...
 }
 
+/// Compute `(regular, state)` intrinsic gas for a transaction without needing
+/// a full VM instance. Mirrors `VM::get_intrinsic_gas` but operates on the raw
+/// transaction, fork, and block gas limit (for cpsb derivation). Pre-Amsterdam
+/// returns `(regular, 0)`.
+///
+/// Used by the block executor to perform the EIP-8037 (PR #2703) per-tx 2D
+/// inclusion check before the tx runs.
+pub fn intrinsic_gas_dimensions(
```
ElFantasma, line 646: (Greptile P2) `intrinsic_gas_dimensions` mirrors `VM::get_intrinsic_gas` line-for-line. `test_intrinsic_parity_*` is the right immediate guard, but the duplication will eventually drift — a future Amsterdam-tweak adds a charge to one and forgets the other. The good outcome: parity test catches it. The bad outcome: someone disables/relaxes the parity test under deadline pressure. Worth a follow-up that extracts shared logic into `compute_intrinsic_gas_internal(tx, fork, block_gas_limit, mode: { Vm | Standalone })` and calls it from both. Not blocking this PR.

crates/vm/levm/src/gas_cost.rs
```diff
@@ -162,14 +162,27 @@ pub const CODE_DEPOSIT_COST: u64 = 200;
...
+pub const CPSB_OFFSET: u64 = 9578;
+
+/// Compute cost_per_state_byte from the block gas limit (EIP-8037, execution-specs#2687).
+///
+/// TEMPORARY for bal-devnet-4: returns the fixed value 1174 used by bal-devnet-3
+/// regardless of `block_gas_limit`. The dynamic formula (BLOCKS_PER_YEAR /
+/// TARGET_STATE_GROWTH_PER_YEAR / CPSB_SIGNIFICANT_BITS / CPSB_OFFSET) is preserved
+/// in the consts above so this commit can be reverted with a single `git revert` to
+/// restore the formula body. See execution-specs#2687.
+pub fn cost_per_state_byte(_block_gas_limit: u64) -> u64 {
```
ElFantasma, line 182: **Description / implementation mismatch**: PR description says "dynamic cpsb (previously a static 1174)" as a delta against bal@v5.6.1, but this is still pinned to 1174 with the same `TEMPORARY for bal-devnet-4` doc that #6546 has been working to clean up. Either (a) the dynamic formula isn't actually being used yet on this branch and the description's delta refers to the spec change, or (b) the dynamic body is in a follow-up. Worth clarifying the description, since reading it suggests dynamic CPSB lands here. Also intersects with #6546 — the EIP-11573 sync that nukes the dynamic formula entirely. Whichever lands second should be aware of the ordering.

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -18,6 +18,18 @@ const SKIPPED_BASE: &[&str] = &[
     "ValueOverflowParis",
     // Skip because it's a "Create" Blob Transaction, which doesn't actually exist. It never reaches the EVM because we can't even parse it as an actual Transaction.
     "createBlobhashTx",
+    // EIP-8025 optional-proofs fixtures filled against bal@v5.6.1 (devnets/bal/3),
+    // which predates EELS PR #2711 "immutable intrinsic_state_gas for EIP-7702".
+    // Expected gas assumes the auth refund still deducts from block-accounted state
+    // gas; our devnet-4 (bal@v5.7.0) impl correctly keeps intrinsic_state_gas
+    // immutable and routes the refund to the reservoir only. Re-enable once the
+    // zkevm@v0.4.x release ships fixtures regenerated against devnet-4.
+    "witness_codes_redelegation_old_marker_included_new_marker_excluded",
```
ElFantasma, line 27: **Action items for these skipped fixtures**:
- Add a TODO comment naming the upstream version that lifts the skip (zkevm@v0.4.x), so a future updater knows when to remove it without re-deriving the reasoning.
- File a tracking issue (or link an existing one) so the skip doesn't outlive its trigger silently.

PR #6522, opened 2026-04-23
refactor(levm): skip redundant bytecode bounds check in JUMP validation

MegaRedHand:
## Motivation

In `jump()`, target validity was checked twice:

1. `bytecode.get(target)` to fetch the byte and verify in-bounds.
2. Compare that byte against `Opcode::JUMPDEST as u8`.
3. Binary search the position in `jump_targets`.

But `Code::compute_jump_targets` (in `crates/common/types/account.rs`) already scans the bytecode and only records positions whose byte is exactly `JUMPDEST (0x5B)` and that are not inside a PUSH literal. So a successful binary search already guarantees both:

- the index is within the bytecode (in-bounds), and
- the byte at that index is `JUMPDEST`.

The explicit `get(target)` bounds check and the byte comparison are therefore redundant.

## Description

- Add `Code::is_valid_jump_target(target: usize) -> bool` on `crates/common/types/account.rs`, which encapsulates the validity check as a single binary search.
- Replace the open-coded check in `jump()` (`crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs`) with a call to the new method, and drop the now-unused `opcodes::Opcode` import.
- Use `u32::try_from(target)` instead of `target as u32`. The `as` cast would silently truncate a `usize` like `2^32 + k` down to a valid-looking `k`, so a target with nonzero upper bits could alias onto a real JUMPDEST. `try_from` rejects those inputs explicitly. (Not exploitable today since EVM bytecode is bounded by `u32::MAX`, but the helper now stands on its own without that assumption.)
- Document the invariants of `jump_targets` that make the single-check correct.

Origin: cherry-picked optimization by @jrchatruc (commit `dbf40f4`), refactored into a method and hardened against the truncating cast.

## TODO

- [ ] Benchmark `jump()` performance vs. `main` to confirm the simplified check is at least neutral.
- [ ] Benchmark with `Code::is_valid_jump_target` annotated `#[inline]` (or `#[inline(always)]`) vs. the default to see whether the extra method call boundary affects hot-path codegen.


crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -417,21 +416,13 @@ fn jump(vm: &mut VM<'_>, target: usize) -> Result<(), VMError> {
...
-            value == Opcode::JUMPDEST as u8
-                && vm
-                    .current_call_frame
-                    .bytecode
-                    .jump_targets
-                    .binary_search(&(target as u32))
-                    .is_ok()
-        })
+        .jump_targets
+        .binary_search(&(target as u32))
```
iovoid: This would allow a jump to addresses whose lowest 32 bits correspond to a valid target.

iovoid (changes requested):

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
MegaRedHand: Right. I'll add a check here that `target < bytecode.len()`

MegaRedHand: I changed the `as` cast to `u32::try_from`

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -417,21 +416,13 @@ fn jump(vm: &mut VM<'_>, target: usize) -> Result<(), VMError> {
...
+        .is_ok()
```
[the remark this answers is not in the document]
MegaRedHand: Fixed

iovoid (approved):

ElFantasma (approved):

azteca1998 (approved):

Issue #6502, opened 2026-04-20
High deserialize input cost for ziskVM


benbencik:
Blog [ZisK v0.16.1 - AIR Cost Reth and Ethrex](https://hackmd.io/@jsign/eth-ziskv0_16_1-air-cost-reth-ethrex) indicates a large gap in AIR-costs for deserialization input in ZisK

- Reth: 274.3M (0.9% relative cost)
- Ethrex: 4.5B (13.9% relative cost)

Ethrex decodes `ExecutionWitness` using `rkyv`. While `rkyv` deserializes input efficiently by encoding data as it is stored in memory, the overhead suggests that deserialization is not achieving true zero-copy. There is an additional cost likely in the [bytecheck validation](https://rkyv.org/validation.html).

The proposal is to use SSZ encoding as in the [spec](https://github.com/ethereum/execution-specs/blob/projects/zkevm/src/ethereum/forks/amsterdam/stateless_ssz.py#L100). The goal is to move the work to the host and provide efficient serialization such that the zkVM needs to perform minimal work.

PR #6524, opened 2026-04-24
feat(l1): add ssz encoding for witness in eip-8025

benbencik:
**Update:** Let's wait until the changes are finalized in #6550 

**Motivation**

Replace `rkyv` encoding of `ExecutionWitness` with SSZ. The structure of `ExecutionWitness` might change further, so this PR is a good middle step towards making EIP-8025 more [spec-compliant](https://github.com/ethereum/execution-specs/blob/projects/zkevm/src/ethereum/forks/amsterdam/stateless.py).

Related to issue https://github.com/lambdaclass/ethrex/issues/6502, but this PR does not yet make decoding faster.

**Description**

Changes add SSZ encoding by using `libssz`

- Added `SszExecutionWitness` as the SSZ representation for `ExecutionWitness`.
- Implemented conversion between SSZ bytes and `ExecutionWitness`.
- Integrated related changes from PR6516 https://github.com/lambdaclass/ethrex/pull/6516
- Added byte encoding for `ChainConfig`

Can be (almost) directly plugged into [ere-guests PR39](https://github.com/eth-act/ere-guests/pull/39)

### DIFF
--- a/crates/common/types/block_execution_witness.rs
+++ b/crates/common/types/block_execution_witness.rs
@@ -62,7 +62,7 @@ pub struct GuestProgramState {
 /// It is essentially an `RpcExecutionWitness` but it also contains `ChainConfig`,
 /// and `first_block_number`.
 #[derive(
-    Default, Serialize, Deserialize, rkyv::Serialize, rkyv::Deserialize, rkyv::Archive, Clone,
+    Default, Serialize, Deserialize, rkyv::Serialize, rkyv::Deserialize, rkyv::Archive, Clone, Debug,
 )]
 pub struct ExecutionWitness {
     // Contract bytecodes needed for stateless execution.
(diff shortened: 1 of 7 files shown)


crates/guest-program/src/l1/program.rs
```diff
@@ -72,36 +72,37 @@ pub fn execution_program(
...
-        |db, _| Ok(Evm::new_for_l1(db.clone(), crypto.clone())),
-        crypto.clone(),
-    )?;
+    let (new_payload_request, execution_witness) = super::decode_eip8025(bytes, crypto.as_ref())
+        .map_err(|err| {
+            ExecutionError::Internal(format!("failed to decode EIP-8025 input: {err}"))
+        })?;
+
+    let request_root = new_payload_request.hash_tree_root(&Sha2Hasher);
+    let valid = validate_eip8025_execution(&new_payload_request, execution_witness, crypto).is_ok();
```
ElFantasma, line 101: **P1**: this collapses every error path — payload conversion, blob hash mismatch, witness conversion bugs, EVM internal errors, OOM-like infra failures — into a normal `valid: false` outcome. Wrong shape for a stateless verifier: it makes infra bugs look like consensus-relevant rejections. The earlier `execution_program` propagates `Err(...)` for those cases, which is correct. Either match it (only `valid: false` for the categories EIP-8025 §Tolerance permits) or document explicitly which errors map to `valid: false` and assert with a typed enum split. (Greptile flagged.)

crates/common/types/block_execution_witness.rs
```diff
@@ -83,6 +83,329 @@ pub struct ExecutionWitness {
...
+                                NodeRef::Hash(hash) => SszNodeRef::Hash(
+                                    SszList::try_from(hash.as_ref().to_vec()).map_err(|e| {
+                                        ExecutionWitnessSszError::InvalidSszType(e.to_string())
+                                    })?,
+                                ),
+                            };
+                            choices.push(ssz_child);
+                        }
+                        SszNode::Branch(SszBranchNode {
+                            choices: SszList::try_from(choices).unwrap(),
```
ElFantasma, line 213: **P1**: `SszList::try_from(...).unwrap()` here (and lines 141, 154, 159, 160, 178, 193, 194) panics on overflow. Bounds:
- `SszBranchNode.value: 128` — branch values for state trie can hit ~110 bytes (RLP-encoded `AccountState`), so 128 is on the edge.
- `SszLeafNode.value: 128` — same.
- `SszExtensionNode.prefix: 64` / `SszLeafNode.partial: 64` — `encode_compact()` outputs of nibble paths up to 64 nibbles, worst case 33 bytes. Comfortable but un-asserted.

Replace each `unwrap()` with `?`-propagating `try_from` and let the existing `ExecutionWitnessSszError::InvalidSszType` carry the error up. Encoder side, where input is locally trusted, but still a robustness-vs-spec-evolution bug. (Greptile flagged.)

crates/common/types/block_execution_witness.rs
```diff
@@ -83,6 +83,329 @@ pub struct ExecutionWitness {
...
+                    }
+                    SszNodeRef::Hash(hash_bytes) => {
+                        let node_hash = decode_node_hash(hash_bytes.as_ref())?;
+                        Ok(NodeRef::Hash(node_hash))
+                    }
+                    SszNodeRef::None => Ok(NodeRef::Hash(ethrex_trie::NodeHash::default())),
+                }
+            }
+
+            fn rebuild_node_by_index(
```
ElFantasma, line 316: **Decoder DoS via cyclic references**: this recurses into `SszNodeRef::Index(i)` without checking that `i < current_index` (or any forward-only invariant). The encoder is post-order so indices are forward-only by construction, but a malicious witness can craft `nodes[0]` referencing `nodes[1]` referencing `nodes[0]` → infinite recursion → stack overflow. For a verifier accepting bytes from untrusted sources (CL relaying execution proofs), that's a DoS vector. Add an explicit `i as usize > current_node_index` check, or a visited-set, or convert to iterative reconstruction with topological-order validation.

crates/common/types/block_execution_witness.rs
```diff
@@ -83,6 +83,329 @@ pub struct ExecutionWitness {
...
+    use libssz_derive::{SszDecode as DeriveSszDecode, SszEncode as DeriveSszEncode};
+    use libssz_types::SszList;
+    use std::sync::Arc;
+
+    // Constants set based on the Ethereum spec
+    const MAX_WITNESS_NODES: usize = 1 << 20;
+    const MAX_WITNESS_CODES: usize = 1 << 16;
+    const MAX_WITNESS_HEADERS: usize = 256;
+    const MAX_BYTES_PER_WITNESS_NODE: usize = 1 << 20;
+    const MAX_BYTES_PER_CODE: usize = 1 << 24;
```
ElFantasma, line 99: `MAX_BYTES_PER_CODE = 1<<24` (16 MB) is *much* larger than EIP-170's 24576-byte cap and EIP-3860's 49152 init-code cap. Tighter bound = tighter spec contract = fewer witness shapes a verifier must accept. Tighten to `49_152` (covers init code) or document why the loose bound is intentional.

crates/common/types/block_execution_witness.rs
```diff
@@ -83,6 +83,329 @@ pub struct ExecutionWitness {
...
+    const MAX_BYTES_PER_HEADER: usize = 1 << 10;
+
+    // TODO: think about these constants
+    const MAX_CHAIN_CONFIG_BYTES: usize = 1 << 10;
```
ElFantasma, line 103: TODO: think about these constants — worth resolving in this PR rather than carrying it forward. 1024 is fine for the current 30-ish-field config, but the `TODO` invites bit-rot.

crates/common/types/block_execution_witness.rs
```diff
@@ -83,6 +83,329 @@ pub struct ExecutionWitness {
...
+    use libssz::{SszDecode, SszEncode};
+    use libssz_derive::{SszDecode as DeriveSszDecode, SszEncode as DeriveSszEncode};
+    use libssz_types::SszList;
+    use std::sync::Arc;
+
+    // Constants set based on the Ethereum spec
+    const MAX_WITNESS_NODES: usize = 1 << 20;
+    const MAX_WITNESS_CODES: usize = 1 << 16;
+    const MAX_WITNESS_HEADERS: usize = 256;
+    const MAX_BYTES_PER_WITNESS_NODE: usize = 1 << 20;
```
ElFantasma, line 98: (Greptile P2) `MAX_BYTES_PER_WITNESS_NODE` declared but never referenced. Current encoding stores `SszNode` structs directly in `SszList<SszNode, MAX_WITNESS_NODES>`, so per-node *byte* bounds aren't enforceable at the SSZ-list level. Either remove the constant, or restructure to encode each node as `SszList<u8, MAX_BYTES_PER_WITNESS_NODE>` and parse on decode (latter buys a uniform bound but adds an indirection).

crates/guest-program/src/l1/program.rs
```diff
@@ -72,36 +72,37 @@ pub fn execution_program(
...
+        valid: true,
+    })
+}
 
-    // Validate blob versioned hashes
-    validate_versioned_hashes(&block, &new_payload_request)?;
+/// Decode and execute the L1 stateless validation program from EIP-8025 wire
+/// bytes.
+///
+/// The wire format is `[ssz_len: u32 LE][ssz_bytes][rkyv_bytes]`, matching
```
ElFantasma, line 86: (Greptile P2) Stale doc comment: still says `[rkyv_bytes]` even though the encoding flipped to SSZ in this PR. Wire format is now SSZ end-to-end.

crates/common/types/block_execution_witness.rs
```diff
@@ -83,6 +83,329 @@ pub struct ExecutionWitness {
...
+                initial_state_root_index,
+                storage_trie_root_indices: SszList::try_from(storage_root_indices).unwrap(),
+                state_nodes: SszList::try_from(state_nodes).unwrap(),
+            };
+            Ok(ssz.to_ssz())
+        }
+
+        pub fn from_ssz_bytes(
+            bytes: &[u8],
+            _crypto: &dyn Crypto,
```
ElFantasma, line 274: `_crypto: &dyn Crypto` is unused. Either wire it through future signature recovery / hash verification or drop the parameter — keeping unused parameters in a feature-gated path tends to bit-rot.

ElFantasma (changes requested):

PR #6527, opened 2026-04-24
feat(l1): bump zkevm EF-tests fixtures to v0.3.3 and extend stateless witness coverage to all for_amsterdam tests

avilagaston9:
**Motivation**

`zkevm@v0.3.3` ships `executionWitness` on every `for_amsterdam` fixture, but the Makefile
only extracts `eip8025_optional_proofs` — leaving ~2670 witnessed fixtures unused by the
stateless path.

**Description**

- Bump `.fixtures_url_zkevm` to `zkevm@v0.3.3` (same `bal@v5.6.1` base).
- Widen `zkevm-vectors` to extract all of `for_amsterdam`, overlaying the non-witness
  `bal@v5.6.1` JSONs so every fixture runs through `run_stateless_from_fixture`.
- Parse `statelessOutputBytes` to honor the `valid` byte at offset 32: both
  `(valid=1, Ok)` and `(valid=0, Err)` now count as pass.
- Move stateless-specific skips into a feature-gated `EXTRA_SKIPS`. 9 `validation_*`
  fixtures skipped pending follow-up — see below.

**Follow-up**

9 skipped `validation_*` tests will be re-enabled in a separate PR: drop malformed entries
lazily in witness conversion, and error on missing bytecode / discontinuous header chains
during stateless execution.

**How to test**

```
make -C tooling/ef_tests/blockchain download-test-vectors
make -C tooling/ef_tests/blockchain test-stateless
```

Expected: 8720 passed, ~160s. `test-levm` unchanged.

**Checklist**


tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -551,32 +551,60 @@ async fn run_stateless_from_fixture(
...
+            let program_input = ProgramInput::new(vec![block.clone()], execution_witness);
+            let res = match backend_type {
+                BackendType::Exec => ExecBackend::new().execute(program_input),
+                #[cfg(feature = "sp1")]
+                BackendType::SP1 => Sp1Backend::new().execute(program_input),
+            };
+            res.map(|_| ()).map_err(|e| format!("execution: {e}"))
+        })();
+
+        match (expected_valid, stateless_outcome) {
```
ElFantasma: **Block on this**: the `(false, Err(_)) => {}` arm is too permissive — it treats *any* host-side failure as a pass, including `executionWitness` JSON parse errors, `into_execution_witness` conversion errors, and program-input setup errors. None exercise the stateless rejection path the fixture is asserting. A regression in `RpcExecutionWitness::deserialize` would silently pass every `valid=false` fixture without ever invoking the guest. Split the closure so JSON/conversion errors propagate as real failures regardless of `expected_valid`. (Copilot flagged.)

tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -551,32 +551,60 @@ async fn run_stateless_from_fixture(
...
 
     Ok(())
 }
+
+/// Extract the `valid` byte from a zkevm-fixture `statelessOutputBytes` hex string.
+///
+/// The output encoding is `new_payload_request_root (32 bytes) ++ valid (1 byte) ++ padding`,
+/// so byte index 32 carries the validity marker.
+#[cfg(feature = "stateless")]
+fn parse_expected_valid_flag(hex: &str) -> Option<bool> {
```
ElFantasma: `unwrap_or(true)` (at the call site above) collapses three distinct cases — field absent, field present but malformed, field present with valid byte 0x00 — into the same default. Right fix: `parse_expected_valid_flag` returns `Result` (or at minimum an `Option` with the absent case kept distinct from the malformed case), so a bad fixture can't silently flip into the success bucket. With ~2670 new fixtures landing, having one silently misinterpreted is a real risk. (Greptile + Copilot both flagged.)

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -28,7 +28,28 @@ const EXTRA_SKIPS: &[&str] = &[
     "Return50000",
     "static_Call1MB1024Calldepth",
 ];
-#[cfg(not(feature = "sp1"))]
+#[cfg(feature = "stateless")]
```
ElFantasma, line 31: `cfg(feature = "sp1")` and `cfg(feature = "stateless")` aren't mutually exclusive at the cfg level — if both are enabled together, this file fails to compile (two definitions of `EXTRA_SKIPS`). Either gate the third arm with `cfg(all(not(feature="sp1"), not(feature="stateless")))` (which it already does via `not(any(...))`) AND add a `compile_error!` in `cfg(all(feature="sp1", feature="stateless"))` to make the exclusivity explicit. Currently those features may be exclusive in practice, but the constraint isn't documented anywhere.

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -28,7 +28,28 @@ const EXTRA_SKIPS: &[&str] = &[
...
+    // completeness checks land (missing delegation/external-code bytecodes, non-contiguous
+    // header chain detection).
+    "validation_codes_missing_delegated_code_on_insufficient_balance_call",
+    "validation_codes_missing_external_code_read_target",
+    "validation_codes_missing_redelegation_old_marker",
+    "validation_codes_missing_sender_delegation_marker",
+    "validation_headers_non_contiguous_chain",
+];
+#[cfg(not(any(feature = "sp1", feature = "stateless")))]
 const EXTRA_SKIPS: &[&str] = &[];
```
iovoid, line 60: if sp1 is set but stateless isn't this won't compile, and if sp1 implies stateless then why specify both?

tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -551,32 +551,60 @@ async fn run_stateless_from_fixture(
...
+    let trimmed = hex.strip_prefix("0x").unwrap_or(hex);
+    let byte_hex = trimmed.get(64..66)?;
+    u8::from_str_radix(byte_hex, 16).ok().map(|b| b != 0)
+}
```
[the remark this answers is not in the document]
avilagaston9: Done in 5a597e6 — `parse_expected_valid_flag` now returns `Result<bool, String>`. Truncated input, non-hex bytes, and validity bytes other than `0x00`/`0x01` all become hard errors at the call site.

tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -551,32 +551,60 @@ async fn run_stateless_from_fixture(
...
+        // When the fixture signals `valid = false` the witness is deliberately incomplete
+        // and the stateless path must reject it; absent bytes means "expected to succeed".
+        let expected_valid = block_data
+            .stateless_output_bytes
+            .as_deref()
+            .and_then(parse_expected_valid_flag)
+            .unwrap_or(true);
+
         let block: CoreBlock = block_data.clone().into();
         let block_number = block.header.number;
```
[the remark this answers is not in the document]
avilagaston9: Done in 5a597e6 — took your suggested shape: `match` on `Some/None`, and a present-but-malformed field returns an explicit error rather than defaulting to `true`.

tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -551,32 +551,60 @@ async fn run_stateless_from_fixture(
...
+                BackendType::Exec => ExecBackend::new().execute(program_input),
+                #[cfg(feature = "sp1")]
+                BackendType::SP1 => Sp1Backend::new().execute(program_input),
+            };
+            res.map(|_| ()).map_err(|e| format!("execution: {e}"))
+        })();
+
+        match (expected_valid, stateless_outcome) {
+            (true, Ok(())) | (false, Err(_)) => {}
+            (true, Err(e)) => {
```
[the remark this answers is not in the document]
avilagaston9: Done in 5a597e6 — split the IIFE. `serde_json::from_value` and `into_execution_witness` now propagate via `?` unconditionally; only backend execution errors are matched against `expected_valid`.

tooling/ef_tests/blockchain/test_runner.rs
avilagaston9: Done in 5a597e6 — IIFE removed. JSON parse and witness conversion always fail the test; only `Backend::execute` errors satisfy `(false, Err(_))`.

tooling/ef_tests/blockchain/test_runner.rs
avilagaston9: Done in 5a597e6 — `parse_expected_valid_flag` returns `Result<bool, String>`, the call site `?`s the error, and only `0x00`/`0x01` are accepted. Absent / malformed / `0x00` / `0x01` are now all distinguishable.

tooling/ef_tests/blockchain/tests/all.rs
avilagaston9: `compile_error!` is at lines 6–7. In ad031d5 I gated the `sp1` arms with `not(feature = "stateless")` so the both-on combo now fails with only that diagnostic (previously two duplicate-definition errors rode along with it).

tooling/ef_tests/blockchain/tests/all.rs
avilagaston9: `sp1` alone compiles (verified). The `sp1 + stateless` combo is rejected by the `compile_error!` at lines 6–7. They are not redundant: stateless uses the in-process Exec backend; sp1 runs the guest ELF inside the SP1 zkVM executor.

ElFantasma (approved):

azteca1998 (approved):

ilitteri (approved):

PR #6529, opened 2026-04-25
perf(l1): add binaryandhash data block index to all rocksdb CFs for faster point lookups

### LINKED ISSUE #5941: Optimize for Point Lookups
```rust
block_opts.set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash);
block_opts.set_data_block_hash_ratio(0.75);
```

Adds hash index inside data blocks for faster point lookups.

dicethedev:
**Motivation**
Point lookups into RocksDB data blocks currently use binary search. For trie nodes and flat key-value stores, which are heavily read during block execution, this adds unnecessary CPU overhead on each block scan.

**Description**
Enables `DataBlockIndexType::BinaryAndHash` with a hash ratio of `0.75` on all column families. This adds a hash index inside each SST data block so point lookups hit O(1) hash probe instead of O(log n) binary
search, at the cost of ~25% extra space within each data block.

Applied to all CFs: `HEADERS`, `BODIES`, `CANONICAL_BLOCK_HASHES`, `BLOCK_NUMBERS`, `ACCOUNT_TRIE_NODES`, `STORAGE_TRIE_NODES`,  `ACCOUNT_FLATKEYVALUE`, `STORAGE_FLATKEYVALUE`, `ACCOUNT_CODES`, `RECEIPTS`, and the default arm.

**Checklist**
- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR
  includes breaking changes to the `Store` requiring a re-sync.
  
 Closes #5941

### DIFF
--- a/crates/storage/backend/rocksdb.rs
+++ b/crates/storage/backend/rocksdb.rs
@@ -110,8 +110,6 @@ impl RocksDBBackend {
 
                     let mut block_opts = BlockBasedOptions::default();
                     block_opts.set_block_size(32 * 1024); // 32KB blocks
-                    block_opts.set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash); // Hash index for faster lookups
-                    block_opts.set_data_block_hash_ratio(0.75); // Hash index covers 75% of entries for good performance
                     block_opts.set_block_cache(&block_cache);
                     cf_opts.set_block_based_table_factory(&block_opts);
                 }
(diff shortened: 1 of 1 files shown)


crates/storage/backend/rocksdb.rs
```diff
@@ -121,6 +121,8 @@ impl RocksDBBackend {
                     let mut block_opts = BlockBasedOptions::default();
                     block_opts.set_block_size(16 * 1024); // 16KB
                     block_opts.set_bloom_filter(10.0, false);
+                    block_opts.set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash);
```
ElFantasma, line 124: Direct conflict point with [#6530](https://github.com/lambdaclass/ethrex/pull/6530) — that PR adds the identical `set_data_block_index_type(BinaryAndHash)` + `set_data_block_hash_ratio(0.75)` calls to the same six BlockBasedOptions blocks. The textual merge resolves cleanly to one copy.

The overlap suggests this PR's content is a *subset* of #6530, which also adds LZ4 compression and broadens `point_lookup_cfs`. If the team wants the binary-hash index landed independently of the broader perf changes (smaller surface area, easier to revert if it ever needs to go), this PR is the cleaner unit. If the team is happy to land both together, #6530 is the strictly-larger superset and this PR can close.

crates/storage/backend/rocksdb.rs
```diff
@@ -121,6 +121,8 @@ impl RocksDBBackend {
                     let mut block_opts = BlockBasedOptions::default();
                     block_opts.set_block_size(16 * 1024); // 16KB
                     block_opts.set_bloom_filter(10.0, false);
+                    block_opts.set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash); 
+                    block_opts.set_data_block_hash_ratio(0.75);
```
ElFantasma, line 125: nit: trailing whitespace on this line (and lines 139-140, 154-155, 171, 183-184, 196). `cargo fmt` would clean these up — worth running before merge.

PR #6530, opened 2026-04-25
perf(l1): enable lz4 compression for trie and flat key-value tables

### LINKED ISSUE #5939: Consider LZ4 for State Tables (RocksDB)
```rust
cf_opts.set_compression_type(rocksdb::DBCompressionType::Lz4);
```

If we're IO-bound, reading less bytes from disk can be faster than reading uncompressed. LZ4 decompression is ~4GB/s.

dicethedev:
**Motivation**
State tables (`ACCOUNT_TRIE_NODES`, `STORAGE_TRIE_NODES`, `ACCOUNT_FLATKEYVALUE`, `STORAGE_FLATKEYVALUE`) were previously uncompressed. On IO-bound nodes, reading fewer bytes from disk outweighs the decompression cost, LZ4 decompresses at ~4 GB/s, making it effectively free relative to disk latency.

**Description**
Adds the four state CFs to `compressible_tables` so they receive `DBCompressionType::Lz4` alongside the existing block/receipt tables. No other logic changes.

**Checklist**

Closes #5939

### DIFF
--- a/crates/storage/backend/rocksdb.rs
+++ b/crates/storage/backend/rocksdb.rs
@@ -125,8 +125,9 @@ impl RocksDBBackend {
                     let mut block_opts = BlockBasedOptions::default();
                     block_opts.set_block_size(16 * 1024); // 16KB
                     block_opts.set_bloom_filter(10.0, false);
-                    block_opts.set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash); 
-                    block_opts.set_data_block_hash_ratio(0.75); 
+                    block_opts
+                        .set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash);
+                    block_opts.set_data_block_hash_ratio(0.75);
                     block_opts.set_block_cache(&block_cache);
                     cf_opts.set_block_based_table_factory(&block_opts);
                 }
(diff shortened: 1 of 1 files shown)


crates/storage/backend/rocksdb.rs
```diff
@@ -71,6 +71,10 @@ impl RocksDBBackend {
             RECEIPTS,
             TRANSACTION_LOCATIONS,
             FULLSYNC_HEADERS,
+            ACCOUNT_TRIE_NODES,
```
ElFantasma, line 74: **Scope difference vs #6529 worth surfacing**: this PR expands `point_lookup_cfs` to include `ACCOUNT_TRIE_NODES`, `STORAGE_TRIE_NODES`, `ACCOUNT_FLATKEYVALUE`, `STORAGE_FLATKEYVALUE`. #6529 doesn't touch this array. The `point_lookup_cfs` set drives other optimizations (the `optimize_for_point_lookup` path elsewhere in this file) — adding trie/FKV CFs here changes their broader RocksDB tuning, not just the binary-hash index.

Worth a brief justification in the PR description for why these four CFs belong in the point-lookup set. Trie nodes and flat KV are both keyed lookups (32-byte keys), so the choice is sound, but it's a non-trivial config change that's separate from the binary-hash index addition. Could be split into its own PR if the team wants surgical perf work.

crates/storage/backend/rocksdb.rs
```diff
@@ -121,6 +125,9 @@ impl RocksDBBackend {
                     let mut block_opts = BlockBasedOptions::default();
                     block_opts.set_block_size(16 * 1024); // 16KB
                     block_opts.set_bloom_filter(10.0, false);
+                    block_opts
+                        .set_data_block_index_type(rocksdb::DataBlockIndexType::BinaryAndHash);
```
ElFantasma, line 129: Direct conflict point with [#6529](https://github.com/lambdaclass/ethrex/pull/6529) — both add `.set_data_block_index_type(BinaryAndHash)` + `.set_data_block_hash_ratio(0.75)` here (and at five other identical blocks below: lines 145, 161, 179, 192, 206). Code is identical; the merge will be a textual conflict that resolves to one copy.

No functional disagreement to resolve — just merge order. The author teams (you and @dicethedev) might want to pick one PR as the canonical landing for this index change and close the other, instead of letting the merge race decide which becomes the no-op rebase.

PR #6536, opened 2026-04-27
fix(l1): pre-check sync head height to skip snap on low-block networks

edg-l:
## Summary
- When joining a fresh network whose head is only a few blocks deep, snap sync stalls in the header download loop before ever reaching the existing `head_close_to_0` fallback in `sync_cycle_snap` (peers are barely synced themselves and `request_block_headers` returns no usable batch).
- Adds an upfront probe in `Syncer::sync_cycle`: fetch the sync head's header by hash; if `header.number < MIN_FULL_BLOCKS`, flip `snap_enabled` to false and dispatch directly to `sync_cycle_full`.
- If the probe can't reach a peer after 3 attempts (2s apart), falls through to the existing snap path unchanged — no behavior regression on healthy mainnet/testnet.

## Test plan
- [ ] Bring up a fresh local devnet of multiple ethrex nodes; verify a newly started node skips snap and full-syncs from genesis without stalling.
- [ ] Confirm normal mainnet/testnet snap sync is unaffected (probe returns `>= MIN_FULL_BLOCKS`, snap path runs as before).
- [ ] Confirm probe-failure case (no peers): snap path runs as before, no immediate fallback to full.

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -10,8 +10,8 @@ mod healing;
 mod snap_sync;
 
 use crate::metrics::METRICS;
-use crate::peer_handler::{PeerHandler, PeerHandlerError};
-use crate::snap::constants::EXECUTE_BATCH_SIZE_DEFAULT;
+use crate::peer_handler::{BlockRequestOrder, PeerHandler, PeerHandlerError};
+use crate::snap::constants::{EXECUTE_BATCH_SIZE_DEFAULT, MIN_FULL_BLOCKS};
 use crate::utils::delete_leaves_folder;
 use ethrex_blockchain::{Blockchain, error::ChainError};
 use ethrex_common::H256;
@@ -29,7 +29,7 @@ use std::sync::{
 use tokio::sync::mpsc::error::SendError;
 use tokio::time::Instant;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 // Re-export types used by submodules
 pub use snap_sync::{
@@ -204,6 +204,28 @@ impl Syncer {
     async fn sync_cycle(&mut self, sync_head: H256, store: Store) -> Result<(), SyncError> {
         // Take picture of the current sync mode, we will update the original value when we need to
         if self.snap_enabled.load(Ordering::Relaxed) {
+            // Probe the sync head's block number before committing to snap sync.
+            // On a fresh devnet the chain head may be only a few blocks deep; the
+            // existing in-loop `head_close_to_0` guard in `sync_cycle_snap` is
+            // only reached after a successful header batch, which can stall when
+            // peers are barely synced themselves. Pre-checking avoids that.
+            if let Some(sync_head_number) = probe_sync_head_number(&mut self.peers, sync_head).await
+                && sync_head_number < MIN_FULL_BLOCKS
+            {
+                info!(
+                    sync_head_number,
+                    "Sync head below MIN_FULL_BLOCKS ({MIN_FULL_BLOCKS}), using full sync"
+                );
+                self.snap_enabled.store(false, Ordering::Relaxed);
+                return full::sync_cycle_full(
+                    &mut self.peers,
+                    self.blockchain.clone(),
+                    self.cancel_token.clone(),
+                    sync_head,
+                    store,
+                )
+                .await;
+            }
             METRICS.enable().await;
             // We validate that we have the folders that are being used empty, as we currently assume
             // they are. If they are not empty we empty the folder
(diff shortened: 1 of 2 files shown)


iovoid (approved):

crates/networking/p2p/sync.rs
```diff
@@ -204,6 +204,28 @@ impl Syncer {
     async fn sync_cycle(&mut self, sync_head: H256, store: Store) -> Result<(), SyncError> {
         // Take picture of the current sync mode, we will update the original value when we need to
         if self.snap_enabled.load(Ordering::Relaxed) {
+            // Probe the sync head's block number before committing to snap sync.
+            // On a fresh devnet the chain head may be only a few blocks deep; the
+            // existing in-loop `head_close_to_0` guard in `sync_cycle_snap` is
+            // only reached after a successful header batch, which can stall when
+            // peers are barely synced themselves. Pre-checking avoids that.
+            if let Some(sync_head_number) = probe_sync_head_number(&mut self.peers, sync_head).await
```
ElFantasma, line 215: **Probe data discard**: when `probe_sync_head_number` returns `Some(n)` with `n >= MIN_FULL_BLOCKS`, we proceed to snap sync but throw away the probe's response headers — the snap loop re-fetches them. That's fine for data-flow simplicity (no need to thread a header through `sync_cycle_snap`'s entry), but worth a one-liner comment confirming the choice is intentional, otherwise a future contributor might "fix" it by caching the probe result and complicate the data path.

crates/networking/p2p/sync.rs
```diff
@@ -233,6 +255,44 @@ impl Syncer {
     }
 }
 
+/// Number of attempts to fetch the sync head's header for the snap-vs-full pre-check.
+const PROBE_SYNC_HEAD_ATTEMPTS: u32 = 3;
```
ElFantasma, line 269: 6s max probe latency (3 × 2s) is reasonable, but worth stating *why* this isn't user-tunable. On a peer-starved devnet (the scenario this PR targets), 6s is an upper bound before falling through to the regular snap path; on a healthy mainnet the first attempt usually succeeds in < 1s. If you ever need to tune it (e.g., a network with persistently slow peers but enough blocks), surfacing these as CLI flags is the next step. Not blocking.

crates/networking/p2p/sync.rs
```diff
@@ -233,6 +255,44 @@ impl Syncer {
...
+/// `PROBE_SYNC_HEAD_ATTEMPTS`. Callers should treat that as "couldn't decide"
+/// and fall through to the regular sync path.
+async fn probe_sync_head_number(peers: &mut PeerHandler, sync_head: H256) -> Option<u64> {
+    for attempt in 1..=PROBE_SYNC_HEAD_ATTEMPTS {
+        match peers
+            .request_block_headers_from_hash(sync_head, BlockRequestOrder::NewToOld)
+            .await
+        {
+            Ok(Some(headers)) => {
+                if let Some(header) = headers.iter().find(|h| h.hash() == sync_head) {
```
ElFantasma, line 291: Defensive `headers.iter().find(|h| h.hash() == sync_head)` is the right call — eth/68 `BlockHeaders` responses can return multiple headers in one batch (the request asks for *up to* N starting from the requested hash), and we only trust the one whose hash actually matches. Without this, a peer could return any header at the requested position and we'd happily trust its `.number`. Good.

Minor: when the response *doesn't* contain the requested header (current `debug!`), it's worth recording this as a peer-quality signal — not blocking, but for the broader peer-scoring story.

crates/networking/p2p/sync.rs
```diff
@@ -233,6 +255,44 @@ impl Syncer {
...
+const PROBE_SYNC_HEAD_ATTEMPTS: u32 = 3;
+/// Delay between probe attempts.
+const PROBE_SYNC_HEAD_RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(2);
+
+/// Tries to fetch the block header for `sync_head` and return its number.
+///
+/// Returns `None` if peers don't respond with the requested header within
+/// `PROBE_SYNC_HEAD_ATTEMPTS`. Callers should treat that as "couldn't decide"
+/// and fall through to the regular sync path.
+async fn probe_sync_head_number(peers: &mut PeerHandler, sync_head: H256) -> Option<u64> {
```
ElFantasma, line 284: Threshold duplication: `MIN_FULL_BLOCKS` is checked here and *also* in `sync_cycle_snap` (snap_sync.rs:220, the `head_close_to_0` guard). Both compare `<` and gate the same fallback to full sync. With the pre-check, the in-loop guard is now mostly redundant — but should stay as a safety net (in case the pre-check is bypassed e.g. via someone calling `sync_cycle_snap` directly). Worth a comment in *both* places noting the relationship so neither gets removed without considering the other.

azteca1998 (approved):

ElFantasma (approved):

PR #6537, opened 2026-04-27
test(l2): pin operator-fee priority-fee underflow on validation gap

avilagaston9:
**Motivation**

`validate_sufficient_max_fee_per_gas_l2` only enforces
`max_fee_per_gas >= base_fee_per_gas + operator_fee_per_gas`. It does
not check `max_priority_fee_per_gas >= operator_fee_per_gas`. A user
can submit an EIP-1559 transaction that passes that validator with
`max_priority_fee_per_gas` set below the operator fee. The effective
`gas_price = min(max_priority + base_fee, max_fee)` then sits below
`base_fee + operator_fee`, and `compute_priority_fee_per_gas` computes
`(gas_price - base_fee) - operator_fee` which underflows. The error
surfaces as `VMError::Internal(InternalError::Underflow)` rather than
a `TxValidationError`. Internal errors are reserved for invariant
violations; a user-controlled input must never hit them.

**Description**

Adds a regression test `priority_fee_below_operator_fee_underflows_in_finalize`
in `test/tests/levm/l2_hook_tests.rs` that constructs the missing-case
inputs (`base_fee=10`, `operator_fee=100`, `max_fee=200`,
`max_priority=50`) and pins the current (buggy) outcome — the VM
returns `VMError::Internal(InternalError::Underflow)`. The assertion is
written so that any change in error handling (for example, tightening
`validate_sufficient_max_fee_per_gas_l2` to also reject
`max_priority_fee_per_gas < operator_fee_per_gas`, which would make
this case bounce out as a `TxValidationError`) breaks this test,
prompting an explicit update of the expected error.

The fix itself is intentionally not in this PR — surfacing the gap and
its reproducer first lets the right validation error variant be picked
deliberately. Two reasonable shapes:

1. Extend `validate_sufficient_max_fee_per_gas_l2` to also require
   `tx_max_priority_fee_per_gas >= operator_fee_per_gas` and return
   the existing `TxValidationError::InsufficientMaxFeePerGas` (or a
   more specific new variant).
2. Saturate at zero in `compute_priority_fee_per_gas` so the
   coinbase simply receives no priority fee instead of crashing —
   weaker, but matches how some L2s "round down to zero" rather
   than reject.

**Reproduction**

```
  priority_fee_below_operator_fee_underflows_in_finalize
```

The test passes on `main` (the bug is real and reachable).

**Impact**

A user-submitted L2 transaction with operator fee active and
`max_priority_fee_per_gas < operator_fee_per_gas` (with
`max_fee_per_gas >= base_fee + operator_fee`) reaches the
internal-error path during finalize. Whether the consequence is
"tx silently dropped from the block" or "block production aborts"
depends on the caller's handling of `VMError::Internal`, but in
either case the validator should have rejected the tx upfront —
internal errors are not the right escape hatch for malformed user
input.

**Checklist**

### DIFF
--- a/test/tests/levm/l2_hook_tests.rs
+++ b/test/tests/levm/l2_hook_tests.rs
@@ -677,3 +677,135 @@ fn privileged_tx_intrinsic_gas_failure_preserves_sender_balance() {
         "Recipient should not receive funds from a reverted privileged tx"
     );
 }
+
+/// Reproducer for an L2 fee-validation gap.
+///
+/// `validate_sufficient_max_fee_per_gas_l2` only requires
+/// `max_fee_per_gas >= base_fee_per_gas + operator_fee_per_gas`, so a
+/// transaction can satisfy that cap while still setting
+/// `max_priority_fee_per_gas < operator_fee_per_gas`. In that case the
+/// effective `gas_price = min(max_priority + base_fee, max_fee)` becomes
+/// `max_priority + base_fee`, the priority component
+/// (`gas_price - base_fee = max_priority`) is below `operator_fee_per_gas`,
+/// and `compute_priority_fee_per_gas` underflows when subtracting the
+/// operator fee — surfacing as `InternalError::Underflow` rather than a
+/// regular `TxValidationError`.
+///
+/// This test pins the current (buggy) behaviour: the transaction should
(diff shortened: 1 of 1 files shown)


avilagaston9:
Closed for inactivity.

PR #6541, opened 2026-04-27
fix(l1): close EIP-8025 stateless witness validation gaps (re-enable 9 zkevm tests)

avilagaston9:
**Motivation**

PR #6527 widened stateless coverage to all `for_amsterdam` fixtures and skipped 9 `validation_*` tests that exposed gaps in ethrex's witness handling. This PR closes those gaps so the 9 tests pass.

**Description**

Four spec-aligned changes (refs to EELS `execution-specs@projects/zkevm` and geth `core/stateless/database.go` are inline in the code):

- **RPC-side tolerance** — `into_execution_witness` drops trie-node entries that fail RLP-decode instead of failing the whole conversion. Bad/extra entries can't be looked up by hash; if execution requires one, the trie walk errors there.
- **Guest-side strictness** — `from_witness` rejects malformed headers and chains where `headers[i].parent_hash != keccak(headers[i-1])`. Mirrors EELS `validate_headers`.
- **Codes completeness** — `get_account_code` / `get_code_metadata` error on missing bytecode instead of silently returning empty code.
- **Generator alignment** — `generate_witness_for_blocks` now emits ancestors in ascending order (was descending). Required so ethrex-generated witnesses satisfy the contiguity check above.

_Cached pre-fix witnesses are descending; they self-heal within the ~120-block cache window, so no read-side sort is added._

**How to test**

```
make -C tooling/ef_tests/blockchain test-stateless
```

Expected: 8720 passed / 0 failed / 0 ignored, ~155 s. `test-levm` unaffected.

**Checklist**


crates/common/types/block_execution_witness.rs
```diff
@@ -325,17 +320,28 @@ impl GuestProgramState {
...
+        // EIP-8025: headers must form a contiguous chain in list order (each
+        // `parent_hash` matches keccak of the previous header bytes). Reordered
+        // or fragmented chains are invalid even if by-number lookup would resolve.
+        // Ref: https://github.com/ethereum/execution-specs/blob/projects/zkevm/src/ethereum/forks/amsterdam/stateless.py#L171-L191
+        let mut block_headers: BTreeMap<u64, BlockHeader> = BTreeMap::new();
+        let mut prev_hash: Option<H256> = None;
+        for bytes in &value.block_headers_bytes {
+            let Ok(header) = BlockHeader::decode(bytes.as_ref()) else {
+                // Malformed entry is a chain break; the next parent_hash check fails.
+                prev_hash = None;
```
ElFantasma: **Malformed-header recovery skips contiguity check on next header** (the inline I posted before — re-anchoring with `line+side`):

After a malformed header, `prev_hash = None` skips the contiguity check on the *next* successful header — `if let Some(expected_parent) = prev_hash` is `None`, so the check is skipped, and we proceed with `prev_hash` reset to the current header's keccak.

**Consequence**: an attacker can insert a malformed entry at position N, then place a fake header at position N+1 with arbitrary `parent_hash`. The for-loop accepts it without verification. Since `block_headers` is then indexed by number, the fake header's `state_root` (or anything else) is silently used downstream.

The doc comment says "Malformed entry is a chain break; the next parent_hash check fails" — but the actual behavior is that the check is *skipped*, not failed.

Two possible fixes:
1. **Hard-fail on any malformed header**: `return Err(GuestProgramStateError::Custom("malformed header in witness"))` instead of `prev_hash = None; continue`. Matches the strict-validation theme of this PR.
2. **Accept segments but require each segment to start at a known-good boundary** (e.g., genesis or the parent of the first executed block).

Option 1 is simpler and more defensible. If EELS does intend lenient acceptance, the comment should at least be corrected to reflect that the rest of the chain after a break is unvalidated.

crates/blockchain/blockchain.rs
```diff
@@ -1436,6 +1436,9 @@ impl Blockchain {
 
             block_headers_bytes.push(current_header.encode_to_vec());
         }
+        // EELS expects ascending order; we walked the chain backward, so reverse.
+        // Ref: https://github.com/ethereum/execution-specs/blob/projects/zkevm/src/ethereum/forks/amsterdam/stateless.py#L171-L191
+        block_headers_bytes.reverse();
```
ElFantasma: Worth confirming: this code walks the chain backward (parent → ... → child) and pushes to `block_headers_bytes`, then reverses to ascending. The reverse is correct *if* the loop pushes in strictly child-to-parent order. The hunk shows the same pattern at line 1678 — both call sites do the same walk and reverse. If they ever diverge (say, one of them changes the iteration direction), the reverse becomes wrong silently. Worth extracting the build-headers logic into a single helper to avoid the parallel-implementation drift risk.

Also worth a unit test asserting `block_headers_bytes` is in ascending order after this loop — would catch any future change to the loop direction.

crates/common/types/block_execution_witness.rs
```diff
@@ -151,36 +151,31 @@ impl RpcExecutionWitness {
...
-            GuestProgramStateError::Custom(format!(
-                "header for block {} not found",
-                first_block_number - 1
-            ))
-        })?;
+        // Skip headers that fail to decode, then pick parent by number.
+        let initial_state_root = self
+            .headers
+            .iter()
+            .filter_map(|h| BlockHeader::decode(h).ok())
```
ElFantasma, line 159: Lenient `.ok()` on RPC side is intentional per EIP-8025 — the witness can carry extra entries that don't decode as headers. That's a different choice from the guest-side strict parsing at line 327+ (which now contiguity-checks each header).

This split is correct in shape (RPC = lenient input, guest = strict invariant). Worth a short comment explaining the asymmetry — currently a future maintainer reading just one of these blocks would wonder why the other is different.

crates/common/types/block_execution_witness.rs
```diff
@@ -588,28 +594,22 @@ impl GuestProgramState {
...
-                println!(
-                    "Missing bytecode for hash {} in witness. Defaulting to empty code.", // If there's a state root mismatch and this prints we have to see if it's the cause or not.
-                    hex::encode(code_hash)
-                );
-                Ok(Code::default())
-            }
-        }
+        self.codes_hashed.get(&code_hash).cloned().ok_or_else(|| {
+            GuestProgramStateError::Database(format!(
+                "missing bytecode for hash {} in witness",
```
ElFantasma: Good change — the old `println!("Missing bytecode... defaulting to empty")` silently masked real bugs (missing-witness bytecode would lead to a state root mismatch later, but the println would have already scrolled past in a noisy log). Hard-erroring here surfaces the witness-corruption scenario where it should be surfaced.

One thing worth confirming: are there any callers that legitimately tolerated a missing-bytecode return in the old code? The comment said "In client implementations there are differences and it's natural for some clients to access more/less information in some edge cases" — that's a hint the lenient default was masking a known issue. If there's a real EELS test fixture where bytecode is intentionally absent (e.g., for self-destructed accounts), this strict error would now flag it. Worth confirming against the v0.3.3 fixture suite that bumped in #6498.

ElFantasma (changes requested):

crates/common/types/block_execution_witness.rs
avilagaston9: Done in b531dab5f — hard-fail + typed `NoncontiguousBlockHeaders`. Also propagated the previously-discarded `Err` from `get_first_invalid_block_hash`.

crates/blockchain/blockchain.rs
avilagaston9: Done in 29c80c2f5 — extracted `build_ascending_ancestor_headers_bytes`, both sites share it. Test in `test/tests/blockchain/witness_tests.rs`.

crates/common/types/block_execution_witness.rs
avilagaston9: Done in b531dab5f.

crates/common/types/block_execution_witness.rs
avilagaston9: Verified — `test-stateless` passes 8720/0 with the strict-on-missing-bytecode change in place.

crates/blockchain/blockchain.rs
```diff
@@ -3151,6 +3104,46 @@ fn branchify(node: Node) -> Box<BranchNode> {
...
+                .then(|| blockhash_opcode_references.get(n))?
+                .copied()
+        })
+        .unwrap_or(first_block_parent_hash);
+
+    let mut block_headers_bytes = Vec::new();
+    let mut current_header = walk_start_header.clone();
+    while current_header.hash() != first_needed_block_hash {
+        let parent_hash = current_header.parent_hash;
+        let current_number = current_header.number - 1;
```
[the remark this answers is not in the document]
avilagaston9: Done in 0f78f21e2 — `checked_sub(1)` returning `WitnessGeneration` if walked past genesis.

test/tests/blockchain/witness_tests.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let numbers: Vec<u64> = witness
+        .block_headers_bytes
+        .iter()
+        .map(|b| BlockHeader::decode(b).unwrap().number)
+        .collect();
+
+    assert!(
+        numbers.windows(2).all(|w| w[0] < w[1]),
+        "ancestor headers must be ascending, got {numbers:?}"
```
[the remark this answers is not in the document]
avilagaston9: Done in f9f0fdc21 — pinned exact ancestor numbers via `assert_eq!`.

crates/common/types/block_execution_witness.rs
```diff
@@ -325,17 +321,20 @@ impl GuestProgramState {
...
+        // Headers must decode and form a contiguous chain in list order.
+        // Ref: https://github.com/ethereum/execution-specs/blob/projects/zkevm/src/ethereum/forks/amsterdam/stateless.py#L171-L191
+        let mut block_headers: BTreeMap<u64, BlockHeader> = BTreeMap::new();
+        let mut prev_hash: Option<H256> = None;
+        for bytes in &value.block_headers_bytes {
+            let header = BlockHeader::decode(bytes.as_ref())?;
+            if let Some(expected_parent) = prev_hash
+                && header.parent_hash != expected_parent
+            {
+                return Err(GuestProgramStateError::NoncontiguousBlockHeaders);
```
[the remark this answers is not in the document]
avilagaston9: Stale — `from_witness` now hard-errors on malformed headers (b531dab5f). PR description updated.

ElFantasma (approved):

ilitteri (approved):

PR #6542, opened 2026-04-28
fix(l1): accept EIP-1898 object form in BlockIdentifierOrHash

edg-l:
## Summary
- `BlockIdentifierOrHash::parse` only accepted string params (`\"0x10\"`, `\"latest\"`, `\"0xhash\"`). The EIP-1898 object form (`{\"blockHash\": \"0x..\"}`, `{\"blockNumber\": \"0x..\"}`), which geth/erigon-compatible clients use, returned `Invalid params: invalid type: map, expected a string`.
- Caught while running spamoor against an ethrex kurtosis devnet: spamoor's receipt-tracking loop calls `eth_getBlockReceipts` with the object form, ethrex rejected it, spamoor stopped sending txs and the mempool stayed empty.
- Fix accepts the object form alongside the existing string forms. `requireCanonical` is accepted but not enforced (matches geth's permissive behavior).
- Six unit tests cover the new and existing parse paths.

## Test plan
- [x] `cargo test -p ethrex-rpc --lib types::block_identifier` (6 new tests pass)
- [x] `cargo clippy -p ethrex-rpc --no-deps --tests` clean
- [x] `cargo fmt`
- [ ] Spin up a kurtosis devnet with spamoor and confirm txs flow (pending fixture rebuild on `bal-devnet-4-bench`)

### DIFF
--- a/crates/networking/rpc/types/block_identifier.rs
+++ b/crates/networking/rpc/types/block_identifier.rs
@@ -114,6 +114,12 @@ impl BlockIdentifierOrHash {
         // EIP-1898 object form: {"blockHash": "0x.."} or {"blockNumber": "0x.." | tag}.
         // `requireCanonical` is accepted but not enforced (matches geth's permissive behavior).
         if let Value::Object(map) = &serde_value {
+            if map.contains_key("blockHash") && map.contains_key("blockNumber") {
+                return Err(RpcErr::BadParams(
+                    "EIP-1898 block identifier cannot specify both `blockHash` and `blockNumber`"
+                        .to_string(),
+                ));
+            }
             if let Some(hash_value) = map.get("blockHash") {
                 let hex_str = serde_json::from_value::<String>(hash_value.clone())
                     .map_err(|e| RpcErr::BadParams(e.to_string()))?;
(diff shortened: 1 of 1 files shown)


ilitteri (approved):

crates/networking/rpc/types/block_identifier.rs
```diff
@@ -111,6 +111,25 @@ impl BlockIdentifierOrHash {
...
+                    BlockHash::from_str(&hex_str).map_err(|_| RpcErr::BadHexFormat(arg_index))?;
+                return Ok(BlockIdentifierOrHash::Hash(block_hash));
+            }
+            if let Some(number_value) = map.get("blockNumber") {
+                return BlockIdentifier::parse(number_value.clone(), arg_index)
+                    .map(BlockIdentifierOrHash::Identifier);
+            }
+            return Err(RpcErr::BadParams(
+                "EIP-1898 block identifier requires `blockHash` or `blockNumber`".to_string(),
+            ));
```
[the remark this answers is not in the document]
Arkenan: Already fixed.

crates/networking/rpc/types/block_identifier.rs
```diff
@@ -111,6 +111,31 @@ impl BlockIdentifierOrHash {
...
+            }
+            if let Some(hash_value) = map.get("blockHash") {
+                let hex_str = serde_json::from_value::<String>(hash_value.clone())
+                    .map_err(|e| RpcErr::BadParams(e.to_string()))?;
+                let block_hash =
+                    BlockHash::from_str(&hex_str).map_err(|_| RpcErr::BadHexFormat(arg_index))?;
+                return Ok(BlockIdentifierOrHash::Hash(block_hash));
+            }
+            if let Some(number_value) = map.get("blockNumber") {
+                return BlockIdentifier::parse(number_value.clone(), arg_index)
```
Arkenan, line 131: Is there a `serde_json::from_value<Number>` or something like that we could use?

Arkenan (approved):
Looks great in general, one small comment.

crates/networking/rpc/types/block_identifier.rs
edg-l: `serde_json::from_value::<serde_json::Number>` would not help here since JSON-RPC sends hex strings (`"0x10"`), not JSON numbers, so we would still need the manual hex parse. I checked geth: its `BlockNumberOrHash.UnmarshalJSON` defers to `BlockNumber.UnmarshalJSON`, which accepts tags (`latest`, `earliest`, etc.) inside the object form too. EIP-1898 technically says QUANTITY only, but matching geth seems safer for client compatibility, so I left it delegating to `BlockIdentifier::parse`.

pablodeymo (approved):

PR #6543, opened 2026-04-28
perf(l1): reduce BAL parallel-path overhead

edg-l:
## Summary

Bundle of independent improvements to the BAL parallel-execution path, validated against a 149-block stress fixture (100M gas, 200–500 tx/block, ~25M-gas median blocks).

| Metric (median) | Sequential | Parallel (no bundle) | **Parallel + bundle** | vs sequential | vs parallel-no-bundle |
|---|---|---|---|---|---|
| **Ggas/s** | 1.78 | 2.88 | **3.64** | **+104.3%** | **+26.4%** |
| total (ms) | 23.86 | 14.43 | **11.44** | −52.1% | −20.7% |
| **exec (ms)** | 21.97 | 12.94 | **6.67** | **−69.6%** | **−48.5%** |
| warmer (ms) | 7.41 | 5.39 | 3.93 | −47.0% | −27.1% |
| store (ms) | 1.60 | 1.19 | 1.25 | −21.9% | +5.0% |

The bundle **doubles the speedup margin** the parallel path was already providing over sequential.

## What's in the bundle

Each change is independently shippable; combined here for atomic review since they touch overlapping code in \`execute_block_parallel\`.

- **A. handle_merkleization_bal overlap fix** (\`blockchain.rs\`) — replace channel drain-loop with single \`recv()\`. Stage B (parallel storage roots) now overlaps with exec instead of serializing after it.
- **B. Adaptive threshold** \`BAL_PARALLEL_TX_THRESHOLD = 5\` — below threshold falls through to sequential exec (which produces a BAL during exec; \`blockchain.rs\` hash-compares against header). Mirrors reth's \`SMALL_BLOCK_TX_THRESHOLD\`.
- **C. import-bench inter-block sleep 500ms → 100ms** (bench tooling change, no production effect) — cuts bench wall-clock 80%.
- **Q1. Skip prestate read in \`bal_to_account_updates\`** when BAL covers all info fields. Two fast paths: storage-only updates and full-info-coverage with non-empty post.
- **Q2. Per-tx \`GeneralizedDatabase\` capacity cap @32** (was sized to full BAL account count, often 100s; p50 tx touches <10).
- **Q3. Memoize \`code_from_bal\`** results — pre-compute Code objects (hash + jump_targets) once per BAL code change before the par_iter; pass cache via optional param to \`seed_db_from_bal\`.
- **Q8. Move per-tx BAL validation into the rayon par_iter closure** — eliminates a serial post-exec validation pass; drops \`current_state\`/\`codes\` inside the closure (no longer cross rayon boundary).
- **DashMap swap in \`CachingDatabase\`** — perf record showed 11% of CPU in \`RwLock::read_contended\` with 16 rayon workers hammering the single account RwLock. Replaced with sharded \`DashMap<_, _, FxBuildHasher>\`. Sequential paths unaffected (only 2 threads, weren't contended).

## Effect on non-BAL paths

- **Block production / pre-Amsterdam / sequential fallback**: DashMap is neutral (low contention); threshold-fallback adds a protective branch; other changes only fire on the BAL parallel-validation path.
- **No regressions in non-parallel paths.**

## Tried-and-rejected (documented for context)

- **Drop \`accessed_accounts\` tracker**: not actually redundant — superset/subset of shadow recorder, distinct correctness roles.
- **\`rayon::join\` warmer Phase 2 + Phase 3**: nested rayon on shared pool starved exec workers (−12%), warmer didn't speed up (already I/O-bound saturating internal par_iter).
- **Validation-only BAL recorder**: exec saved 5%, but those savings shifted to "after exec" merkle drain — net per-block flat. Once exec < merkle wall-clock, exec-side savings have diminishing returns on per-block time.

## Test plan
- [x] \`cargo check -p ethrex-blockchain -p ethrex-levm -p ethrex-vm\` (clean)
- [x] Stress fixture (149 blocks, 100M gas, mainnet-shape): per-block medians match the table above
- [ ] Hive Amsterdam consume-engine
- [ ] EF blockchain tests (BAL fixtures \`bal@v6.0.0\`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4145,6 +4145,7 @@ name = "ethrex-levm"
 version = "9.0.0"
 dependencies = [
  "bytes",
+ "dashmap",
  "derive_more 1.0.0",
  "ethrex-common",
  "ethrex-crypto",
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -941,13 +941,12 @@ pub async fn import_blocks_bench(
                     _ => warn!("Failed to add block {number} with hash {hash:#x}"),
                 })?;
 
-            // TODO: replace this
-            // This sleep is because we have a background process writing to disk the last layer
-            // And until it's done we can't execute the new block
-            // Because this wants to compare against running a real node in terms of reported performance
-            // It takes less than 500ms, so this is good enough, but we should report the performance
-            // without taking into account that wait.
-            tokio::time::sleep(Duration::from_millis(500)).await;
+            // TODO: replace this with an explicit "wait for persistence idle" call.
+            // The previous block's Phase 2 (disk write of bottom-most diff layer) runs in
+            // a background thread; we sleep so its completion doesn't bleed into the next
+            // block's per-block timer. 100ms is comfortably above Phase 2 cost on SSD for
+            // small blocks; raise if metrics get noisy.
+            tokio::time::sleep(Duration::from_millis(100)).await;
         }
 
         // Make head canonical and label all special blocks correctly.
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -837,36 +837,40 @@ impl Blockchain {
         const NUM_WORKERS: usize = 16;
         let parent_state_root = parent_header.state_root;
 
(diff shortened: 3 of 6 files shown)


edg-l:
Superseded by #6639, which rebases this onto `bal-devnet-6-pr`.

PR #1440, opened 2024-12-09
fix(levm): insert storage slot to account when accessing to it

maximopalopoli:
**Motivation**

**Description**

Closes #issue_number

PR #6544, opened 2026-04-28
feat(l1):  add EIP-8189 snap/2 (BAL-based state healing)

edg-l:
## Motivation

Implements [EIP-8189](https://eips.ethereum.org/EIPS/eip-8189): snap/2, which replaces iterative trie-node fetching during state healing with Block Access List (EIP-7928) replay. This is cleaner, avoids the need for partial-trie reconstruction, and lets the local node serve the same BALs to peers over snap/2 once sync is complete.

Depends on EIP-7928 (BAL header commitment) already present in this repo.

## Description

**Wire protocol**

- Advertises `snap(2)` alongside `snap(1)`; per-connection negotiation picks the highest common version.
- New messages: `GetBlockAccessLists` (0x08) and `BlockAccessLists` (0x09).
- snap/2 connections reject `GetTrieNodes`/`TrieNodes`; snap/1 connections serve them unchanged.
- `BlockAccessLists` response: `Vec<Option<BlockAccessList>>` with 0x80 RLP sentinel for absent entries, preserving position correspondence with the request.
- Compile-time assertions in `rlpx/message.rs` guard the new message range against `BASED_CAPABILITY_OFFSET`.

**BAL persistence**

- New storage table `BLOCK_ACCESS_LISTS` with `store_block_access_list`, `get_block_access_list`, `iter_block_access_lists_by_hashes`.
- `execute_block` returns `Option<BlockAccessList>`; `store_block` takes `Option<&BlockAccessList>`. Pre-Amsterdam blocks pass `None`.
- All block-import paths thread the BAL: `add_block`, `add_block_pipeline_inner`, `add_blocks_in_batch`.
- BAL hash validated once per block in `execute_block` / `execute_block_from_state`.

**Server / client / replay engine**

- Server `process_block_access_lists_request` respects a 2 MiB soft cap (`BAL_RESPONSE_SOFT_CAP_BYTES`); first slot always included; remaining over-cap slots filled with `None`.
- Client `request_block_access_lists` returns `(Vec<Option<BlockAccessList>>, peer_id)` to keep failure attribution coherent.
- `apply_bal` (`sync/bal_healing/apply.rs`) applies BAL as state diffs: balance/nonce/code/storage post-values; implicit-empty entry triggers account deletion; missing local slots treated as zero pre-value; storage writes go directly to backend.
- `advance_state_via_bals` fetches in batches, validates ordering/hash before each apply, applies in strict block order using locally-tracked `current_root`, persists each BAL. On peer exhaustion falls back to snap/1 `heal_state_trie_wrap`.

**Sync state machine**

Staleness loop in `snap_sync.rs` branches on snap/2 peer reachability: V2 path runs `advance_state_via_bals` then validates final `state_root`; V1 path is unchanged.

**Tests**

- 6 codec round-trip tests for new wire types.
- 5 server tests for `process_block_access_lists_request` (empty, all-known, mixed, 2 MiB cap, snap/1 rejection).
- 7 `apply_bal` unit tests: account creation/destruction, storage deletion, code deployment, EIP-7702 delegation clear, fresh storage slot, bad-state-root detection.
- 1 batch-import BAL persistence regression test in `test/tests/blockchain/batch_tests.rs`.
- 3 storage tests for the new BAL API.
- `cargo test -p ethrex-p2p`: 77 passing. `cargo test -p ethrex-test`: 987 passing.

## Cross-client sync on a live devnet

The client side has now synced a node end to end over snap/2, against real peers on a kurtosis Amsterdam devnet (mainnet preset, `ethpandaops/geth:glamsterdam-devnet-7` with `--snap.v2`, `ethpandaops/lighthouse` as CL). The harness lives in `scripts/snap2-devnet/join.py`: it joins a fresh ethrex node to a running enclave, drives it with `engine_forkchoiceUpdatedV4`, and asserts on which path the sync actually took.

A run from a fresh datadir:

```
Starting snap/2 state sync (BAL-based, no trie healing) pivot=510
snap/2 state download complete, pivot frozen block=510
snap/2 reconstruction matches the pivot state root block=510 root=0x957f…6f07

Headers: 511   Accounts: 1,674   Storage: 1,619   Bytecodes: 9
Healed: 0 state paths + 0 storage accounts
```

Verified afterwards against the geth seeder over RPC, at the pivot and at the joiner's head: identical block hashes, identical state roots, identical balances, contract code and contract storage. The node then full-syncs on top of that state, which re-derives the root independently. Zero heals, so nothing fell back to snap/1.

Path selection is asserted on a latched `snap2_reconstructed_block` in `admin_syncStatus` rather than on a sampled phase string; a snap/1 fallback also ends with a correctly synced node, so "the node synced" is not by itself evidence of anything.

What this run does not cover: access-list catch-up did not execute, because the devnet's state is small enough to download inside a single pivot, so the pivot never went stale. Catch-up was exercised on an earlier, slower run (48 blocks over 4 rounds, 0 validation failures), and the harness now reports explicitly when a run leaves it untested rather than passing silently.

## Notes

- Hive-side end-to-end coverage (8.5/8.6 from the EIP) is still open. `test/tests/p2p/` has no multi-node RLPx harness and the existing hive devp2p snap simulator calls the external Go `devp2p` CLI, so a dedicated hive snap/2 simulator is still needed; the devnet harness above covers the same ground manually in the meantime.
- Reorg past the pivot is unimplemented; the sync takes the discard-and-restart escape hatch that `caps/snap.md` allows.
- Reconstruction is single-threaded where snap/1's storage insert fans out. Not implicated in correctness, but likely the largest gap at mainnet scale.
- Updated `STORE_SCHEMA_VERSION` in `crates/storage/lib.rs` for the new `BLOCK_ACCESS_LISTS` table.

**Checklist**


crates/networking/p2p/sync/bal_healing/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+                            continue;
+                        };
+
+                        // Strict in-batch ordering: defer apply until every prior
+                        // slot has been filled. Without this, an out-of-order
+                        // response could apply BAL[2] against a state that hasn't
+                        // yet had BAL[1] applied — producing the wrong root.
+                        let all_prior_filled = (0..batch_idx).all(|k| batch_filled[k]);
+                        if !all_prior_filled {
+                            continue;
```
ElFantasma: **Strict-ordering `continue` silently drops a usable BAL.** If the peer returns `[BAL[0]=None, BAL[1]=Some, BAL[2]=Some]`, we increment `retry_counts[0]` and then hit this guard for `batch_idx=1` and `batch_idx=2` — both `Some` BALs we just paid bandwidth for are discarded, and the next request re-asks for indices [0, 1, 2]. With BAL_REQUEST_BATCH_SIZE=64 and any small per-block hit rate at slot 0, you wind up re-fetching almost the entire batch on every miss.

Low-effort fix: buffer the BAL into a `pending: Vec<Option<&BlockAccessList>>` parallel to `batch_filled`, and when slot K finally fills, drain forward through K+1, K+2, ... applying any already-buffered BALs in order. Pure perf, not correctness — the spec doesn't require this — but the savings are real on a flaky peer set.

Non-blocking.

crates/networking/p2p/sync/bal_healing/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+    if start_number > end_number {
+        return Ok(vec![]);
+    }
+
+    let mut headers = Vec::with_capacity((end_number - start_number + 1) as usize);
+    for number in start_number..=end_number {
+        let hash = store
+            .get_canonical_block_hash(number)
+            .await?
+            .ok_or(SyncError::MissingHeaderForBal(H256::zero()))?;
```
ElFantasma: `SyncError::MissingHeaderForBal(H256::zero())` here loses the only piece of information that would let an operator debug this — the missing block number. If `get_canonical_block_hash(number)` returns `None`, we know the canonical chain is missing at `number` (a real DB inconsistency given we just successfully walked from `start_number` to `end_number`), but the error reports a zero hash instead of the number.

Two options:
- Add a `MissingCanonicalAtNumber(u64)` variant for this branch and use it (clearest).
- Or at minimum keep the existing variant but include the number in the error string by formatting it into the hash, or extending the variant payload.

Non-blocking, but cheap to fix and saves a 30-minute debug session later.

crates/networking/p2p/sync/bal_healing/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ethrex_common::H256;
+    use ethrex_common::types::BlockHeader;
+    use ethrex_storage::{EngineType, Store};
+
+    /// `PeerHandler` requires an `RLPxInitiator` actor to construct; that makes
+    /// it impractical to directly unit-test `advance_state_via_bals` here. The
+    /// orchestration is covered by the deferred E2E test (M4 — Phase 3). What
```
ElFantasma: The honesty about the coverage gap is appreciated, but the orchestration in `advance_state_via_bals` has the densest bug surface of the PR — out-of-order batches, per-block retry accounting, batch-boundary parent linkage, peer-failure recording, the early-exit conditions. "Deferred to M4 — Phase 3" leaves the most error-prone code uncovered until the integration phase.

A cheap intermediate: factor out the inner peer-response → fill-batch step into a pure function that takes a `Vec<Option<BlockAccessList>>` (the response) + the current `batch_filled` / `retry_counts` / `current_root` state, and returns the next state. That function is unit-testable without a `PeerHandler`, and the cases that benefit are exactly the ones I'm worried about (out-of-order responses, single-slot exhaustion, mid-batch reorg detection).

Non-blocking — the comment is fair — but worth pulling forward if M4 slips.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1572,6 +1615,72 @@ async fn handle_incoming_message(
...
+        Vec::with_capacity(hashes.len());
+    let mut bytes_used: u64 = 0;
+
+    for (hash, raw_bal) in hashes.iter().zip(raw_bals.into_iter()) {
+        // Keep at least one entry then stop once cap is exceeded.
+        if !bals.is_empty() && bytes_used >= cap {
+            break;
+        }
+
+        let header = storage
```
ElFantasma: **Per-hash header lookup runs even when we already have a BAL.** When `raw_bals[i]` is `Some(bal)`, we *already know* the block is post-Amsterdam (we wouldn't have stored a BAL for a pre-Amsterdam block). The only branch that needs `header.block_access_list_hash` is the §100 case — i.e. when raw_bal is `None` and we need to distinguish "pre-Amsterdam" from "unknown/pruned".

Cheap rewrite:
```rust
let slot = match raw_bal {
    Some(bal) => Some(bal),
    None => match storage.get_block_header_by_hash(*hash)? {
        Some(h) if h.block_access_list_hash.is_none() => None, // §100
        _ => None, // unknown/pruned
    },
};
```
Saves N storage reads on the happy path where the peer has all the BALs. On a 1024-hash request that's ~1024 fewer disk hits per response, which is the kind of work this defense-against-DoS limit was designed to bound in the first place.

Non-blocking perf.

crates/networking/p2p/rlpx/snap/codec.rs
```diff
@@ -306,6 +313,102 @@ impl RLPxMessage for TrieNodes {
...
+        match &self.0 {
+            None => 1,
+            Some(bal) => bal.length(),
+        }
+    }
+}
+
+impl RLPDecode for Snap2OptionalBal {
+    fn decode_unfinished(rlp: &[u8]) -> Result<(Self, &[u8]), RLPDecodeError> {
+        if rlp.first() == Some(&0x80) {
```
ElFantasma, line 380: The byte-peek decoder is correct, but the invariant it relies on isn't written down here. RLP can encode a single-byte short string as `0x80..0xb7` and a list as `0xc0..0xff`. `0x80` is unambiguous for `None` *only because* a `BlockAccessList` is always encoded as a list (first byte `>= 0xc0`). If anyone ever refactors `BlockAccessList` into a wrapper type whose RLP encoding could collapse into a short string, this decoder silently mis-decodes the first BAL of the batch as `None`.

Worth a one-line invariant comment right above the `if`, e.g.:
```rust
// INVARIANT: BlockAccessList encodes as an RLP list (first byte >= 0xc0),
// so 0x80 is unambiguously the None sentinel.
```
or a stricter check (`rest @ [0x80]` -> None; otherwise decode_unfinished). Either keeps the next refactor from biting silently.

Non-blocking.

crates/networking/p2p/rlpx/snap/codec.rs
```diff
@@ -306,6 +313,102 @@ impl RLPxMessage for TrieNodes {
...
+        Ok((Snap2OptionalBal(Some(bal)), rest))
+    }
+}
+
+impl RLPxMessage for Snap2BlockAccessLists {
+    const CODE: u8 = codes::SNAP2_BLOCK_ACCESS_LISTS;
+
+    fn encode(&self, buf: &mut dyn BufMut) -> Result<(), RLPEncodeError> {
+        let mut encoded_data = vec![];
+        let bals: Vec<Snap2OptionalBal> = self.bals.iter().cloned().map(Snap2OptionalBal).collect();
```
ElFantasma: `.iter().cloned()` deep-clones every `BlockAccessList` in the response just to wrap it in `Snap2OptionalBal` for the encoder. BALs are not small (kilobytes per block, hundreds in a 64-block batch), so this is a real allocation hit on the server side, every response.

The wrapper exists only to give `RLPEncode` somewhere to dispatch the `None`-as-0x80 case. A borrowing variant — e.g. `struct Snap2OptionalBalRef<'a>(Option<&'a BlockAccessList>)` with its own `RLPEncode` impl — gets the same wire format with zero copies. Then:
```rust
let bals: Vec<Snap2OptionalBalRef<'_>> =
    self.bals.iter().map(|b| Snap2OptionalBalRef(b.as_ref())).collect();
```
Decode side can keep `Snap2OptionalBal` (owning) since the wire bytes are owned anyway.

Non-blocking perf.

ElFantasma (approved):
It LGTM. Left several comments but none is too important. Feel free to decide if they are actionable in some way (this PR, new PR or disregard)

edg-l:
Pushed fixes for the review:

- snap/2: call heal_storage_trie after a successful BAL replay. The head state-root check only commits storage_root hashes, so storage tries of accounts not touched by any replayed BAL could stay incomplete on disk and the check still passes. No-op when the download left nothing incomplete.
- apply_bal: empty BAL now checks parent_state_root == header.state_root instead of returning blindly.
- codec: drop the per-response deep clone of every BAL on encode (borrowing wrapper); added the 0x80 None-sentinel invariant comment.
- load_headers_range: error reports the missing block number instead of a zero hash.

greptile P2 about the fallback CodeHashCollector temp dir no longer applies, that path was already reworked to reuse the main collector. Left the out-of-order BAL buffering and the header-lookup skip as follow-ups.

crates/networking/p2p/sync/bal_healing/mod.rs
```diff
@@ -0,0 +1,406 @@
...
+                            }
+                            continue;
+                        };
+
+                        // Strict in-batch ordering: defer apply until every prior
+                        // slot has been filled. Without this, an out-of-order
+                        // response could apply BAL[2] against a state that hasn't
+                        // yet had BAL[1] applied — producing the wrong root.
+                        let all_prior_filled = (0..batch_idx).all(|k| batch_filled[k]);
+                        if !all_prior_filled {
```
ElFantasma: Re-nudge on the strict-order `continue` (original inline flagged 2026-05-28). To clarify the concern: when the peer returns `[BAL[0]=None, BAL[1]=Some, BAL[2]=Some]`, slots 1 and 2 hit `!all_prior_filled` here, so `batch_filled[1]` and `batch_filled[2]` stay `false`. Correctness is fine — the outer retry loop re-derives `pending_hashes` from `!batch_filled` and re-requests those slots, so nothing is *lost*. The concern is **bandwidth waste**: we paid for two usable BALs and threw them away because slot 0 was missing, and now we have to fetch them again (possibly repeatedly if slot 0 keeps missing).

One shape that fixes it: keep a `deferred: FxHashMap<usize, BlockAccessList>` and, after applying slot 0, drain deferred entries in order. Similar to the classic "holding-buffer + in-order apply" pattern for out-of-order chunk assembly. Optional — the current strict-order path is correct, just inefficient under partial-response scenarios. Non-blocking.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1746,6 +1789,72 @@ async fn handle_incoming_message(
...
+            .get_block_header_by_hash(*hash)
```
ElFantasma: Re-nudge on per-hash header lookup (original inline flagged 2026-05-28). The lookup at `get_block_header_by_hash(*hash)` runs unconditionally per request-hash to check the §100 `block_access_list_hash.is_none()` gate. But when `raw_bal.is_some()` we already know the block is post-Amsterdam — a stored BAL implies the block passed the Amsterdam-fork admission that gates BAL storage in the first place. Short-circuit:

```rust
let slot = match raw_bal {
    Some(bal) => Some(bal), // stored implies post-Amsterdam per §100
    None => match storage.get_block_header_by_hash(*hash)? {
        Some(h) if h.block_access_list_hash.is_none() => None, // pre-Amsterdam per §100
        Some(_) | None => None, // known post-Amsterdam w/o stored BAL, or unknown block
    },
};
```

Skips the header lookup for the common case (peer requesting BALs for blocks that have them). Non-blocking perf; matters at scale under snap/2 serving load.

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -525,29 +526,152 @@ pub async fn snap_sync(
...
+                        // complete. But storage tries of accounts not touched by any replayed BAL
+                        // keep whatever (possibly incomplete) storage they had from the snap download
+                        // phase: the state root only commits each account's storage_root *hash*, so
+                        // it matches even when storage leaves are still missing on disk. Heal those
+                        // exactly as the snap/1 path does before declaring healing complete; otherwise
+                        // the node passes its state-root check yet cannot serve those storage slots.
+                        // When the download left no incomplete storage (`storage_accounts` empty),
+                        // this is a no-op: the work queue is empty and it returns immediately.
+                        healing_done = heal_storage_trie(
+                            pivot_header.state_root,
```
ElFantasma: New observation on the BAL-replay-success heal path. The check just above matches `new_root == final_header.state_root` (where `final_header` is the block at `latest_head_hash`), but the heal call passes `pivot_header.state_root`. If BAL replay advanced state past the pivot — which is the entire reason we set up the pivot-to-head replay — those two roots differ, and `heal_storage_trie` is being asked to verify storage against a root the node has already moved past.

A `storage_accounts` entry's storage_root was captured at pivot time. If BAL replay changed that account's storage, the storage_root under `pivot_header.state_root` may no longer be what disk actually holds. Two possibilities:

1. **Intentional**: the storage layer commits storage-root hashes at pivot, and heal-against-pivot is the right way to complete the pivot snapshot before letting the state root move on. Then a comment explaining why we don't heal against the current head would clarify.
2. **Bug**: heal should be against `final_header.state_root` (the state we've actually reached), since that's what the node will serve going forward.

My read leans (1) — the pivot snapshot is the incomplete-storage payload, so healing against pivot is what completes it. But since the site is repeated four more times below with `pivot_header.state_root` too, worth confirming the semantic once so a future refactor doesn't switch to `final_header.state_root` on the assumption that's more "current" and silently break the invariant.

Non-blocking clarification.

ElFantasma (approved):
Re-review — 3/6 addressed cleanly + 1 acknowledged; three non-blocking follow-ups inline.

edg-l:
Ran the hive devp2p `snap2` suite against this branch locally: **4/4 tests, 10/10 sub-cases passing.**

```
ok 1 Status
ok 2 GetBlockAccessLists
ok 3 TrieNodesRemoved
```

Setup is ethereum/hive#1440 (adds the suite) plus geth master, which has `devp2p rlpx snap2-test` since ethereum/go-ethereum#34626 merged. Added `make run-hive-snap2`; note `make setup-hive` checks out hive master and will discard the #1440 patch until it merges.

The sub-cases cover empty requests, single and multi-hash, unknown hashes, duplicates, the 16-hash canonical window, and `responseBytes` of 1 and 0, so this exercises the wire format, the `0x80` unavailable sentinel, tail truncation, request ordering, and `GetTrieNodes` rejection on snap/2.

One fix was needed to get there. Every test initially failed at the handshake:

```
could not negotiate snap protocol (remote caps: [eth/68 eth/69 eth/70 eth/71 snap/1], local snap version: 2)
```

snap/2 advertisement was gated on `is_synced`, which by its own definition "does not reflect whether there is an ongoing sync process". A node that obtained its chain by any means other than finishing a sync, which is how the hive suite runs one, never sets it and so could never negotiate snap/2. The condition that actually has to hold is narrower: reconciling the state trie needs snap/1's `GetTrieNodes`, so only a node *running a snap sync* has to keep negotiating snap/1. Now gated on that instead.

Worth noting the suite also settles the response encoding empirically. It hashes the wire bytes directly:

```go
have := crypto.Keccak256Hash(raw)
want := *block.Header().BlockAccessListHash
```

so entries must be the BAL's own RLP spliced in verbatim, matching geth's `AccessLists rlp.RawList[rlp.RawValue]` + `AppendRaw(bal)`. That is what this branch does. The `bals: [bal1: B, ...]` annotation in devp2p `caps/snap.md` reads as a spec bug: a byte-string header would be inside that hash, and the section's own `keccak256(bal_i)` rule only holds for the raw form. Worth a devp2p issue.

Still open, and not covered by this suite since it only tests the serving side: the client-side BAL replay path, and reorg-past-pivot recovery.

ilitteri (approved):

PR #6545, opened 2026-04-28
feat(l1): bound memory usage of healing

iovoid:
**Motivation**

For large networks storage healing might end up loading hundreds of millions of nodes into the pending queue.

**Description**

We use a priority queue to ensure deep nodes are resolved first instead of opening new tries. We also set a soft cap after which download goes slower to ensure parallelism doesn't cause the limit to be substantially exceeded. The limit is not hard as we prefer a risk of OOM rather than a stall.

### DIFF
--- a/crates/networking/p2p/sync/healing/state.rs
+++ b/crates/networking/p2p/sync/healing/state.rs
@@ -75,6 +75,13 @@ pub async fn heal_state_trie_wrap(
 /// Returns true if healing was fully completed or false if we need to resume healing on the next sync cycle
 /// This method also stores modified storage roots in the db for heal_storage_trie
 /// Note: downloaders only gets updated when heal_state_trie, once per snap cycle
+// TODO: `paths` and `healing_queue` are both unbounded here. Storage healing
+// caps its pending-parents map via `HEALING_QUEUE_SOFT_LIMIT` and uses a
+// depth-ordered download queue to keep memory in check; state healing has
+// no equivalent guard. The state trie is much smaller than aggregate storage
+// so this has not been a problem in practice, but if state-healing OOMs ever
+// appear the same pattern (depth-priority heap + soft-limit backpressure)
+// applies.
 #[allow(clippy::too_many_arguments)]
 async fn heal_state_trie(
     state_root: H256,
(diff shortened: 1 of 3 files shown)


crates/networking/p2p/sync/healing/state.rs
```diff
@@ -38,6 +38,36 @@ use crate::{
...
+///
+/// Used inside a `BinaryHeap` (max-heap) so the dispatcher pops the deepest
+/// pending node available. Depth-first draining is what shrinks `healing_queue`
+/// fastest: committing a leaf cascades up through its ancestors via
+/// `commit_node`, freeing pending parents. Shallow-first would instead keep
+/// expanding the frontier and grow the queue without bound.
+#[derive(Debug, Clone)]
+struct DepthOrderedMetadata(RequestMetadata);
+
+impl PartialEq for DepthOrderedMetadata {
```
ElFantasma, line 51: **`PartialEq` surprise**: equality is purely on path length, so two structurally different requests at the same depth compare as equal. Fine for `BinaryHeap` (which only uses `Ord`), but if anyone later adds dedup or a `HashSet<DepthOrderedMetadata>` they'll get silent collisions. Either (a) doc-comment the equality as ordering-only, or (b) derive structural `PartialEq`/`Eq` and put the depth-only comparison in `Ord` alone — `BinaryHeap` doesn't require `PartialEq` to match `Ord` semantics. Same applies to `DepthOrderedRequest` in `storage.rs`.

crates/networking/p2p/sync/healing/state.rs
```diff
@@ -194,66 +239,83 @@ async fn heal_state_trie(
...
+                        .peer_table
+                        .get_best_peer(SUPPORTED_SNAP_CAPABILITIES.to_vec())
+                        .await
+                        .inspect_err(|err| {
+                            debug!(err=?err, "Error requesting a peer to perform state healing")
+                        })
+                        .unwrap_or(None)
+                    else {
+                        // If there are no peers available, re-add the batch to the paths vector, and continue
+                        paths.extend(batch.into_iter().map(DepthOrderedMetadata));
```
ElFantasma, line 289: Re-pushes a freshly-popped batch back into the heap on the no-peers path, costing O(BATCH_SIZE · log n). Acceptable for a slow path, but a peer-starved cluster eats more CPU than the old `VecDeque` would. Not blocking.

ElFantasma (approved):
Two minor comments

edg-l (approved):

azteca1998 (approved):

PR #6546, opened 2026-04-28
chore(l1): sync EIP-8037 with EIPs PR 11573

edg-l:
## Summary

ethereum/EIPs#11573 ("Update EIP-8037: fixed CPSB + frame accounting") rewrote large parts of the spec we already implement on this branch. The bulk of the rewrite (frame-end state-gas accounting, deferred SELFDESTRUCT, revert/halt cancellation) was already in place; this PR closes the remaining drift between our code and the new EIP text.

## Gaps closed

1. **Fixed CPSB doc + dead-const cleanup** — PR 11573 collapses the dynamic, gas-limit-dependent `cost_per_state_byte` formula (with quantization) into a single permanent constant of 1174. We were already returning 1174, but the function carried a "TEMPORARY for bal-devnet-4" doc and the formula constants (`BLOCKS_PER_YEAR`, `TARGET_STATE_GROWTH_PER_YEAR`, `CPSB_SIGNIFICANT_BITS`, `CPSB_OFFSET`) were dead code. Removed the consts, rewrote the doc, replaced the obsolete dynamic-formula unit tests in `test/tests/levm/eip8037_tests.rs` with a single `test_cpsb_is_fixed_at_1174` that asserts 1174 across a spread of inputs.

2. **Auth-bytes constant rename** — Spec now defines `STATE_BYTES_PER_AUTH_BASE = 23` and computes the per-auth state cost additively as `(STATE_BYTES_PER_NEW_ACCOUNT + STATE_BYTES_PER_AUTH_BASE) x CPSB`. We had precomputed `STATE_BYTES_PER_AUTH_TOTAL = 135` and `STATE_BYTES_PER_AUTH_ONLY = 23`. Renamed `_AUTH_ONLY` to `_AUTH_BASE` and dropped `_AUTH_TOTAL`; consumers in `crates/vm/levm/src/state_diff.rs:38` (`bytes()`), `crates/vm/levm/src/utils.rs:570` (intrinsic gas in `eip7702_set_access_code`), and `crates/vm/levm/src/utils.rs:690` (`intrinsic_gas_dimensions`) now compute the additive form inline. Bytes accounting is unchanged (135 = 112 + 23).

3. **System-call gas-limit bump** — PR 11573 introduces `SYSTEM_CALL_GAS_LIMIT = 30_000_000 + STATE_BYTES_PER_STORAGE_SET x CPSB x SYSTEM_MAX_SSTORES_PER_CALL` (= 30,601,088 with `SYSTEM_MAX_SSTORES_PER_CALL = 16`, matching `MAX_WITHDRAWAL_REQUESTS_PER_BLOCK` from EIP-7002). The extra 601,088 is placed in `state_gas_reservoir`; `gas_left` keeps the legacy 30M execution budget. System calls remain not subject to the EIP-7825 `TX_MAX_GAS_LIMIT` cap and don't contribute to `block_regular_gas_used`/`block_state_gas_used`. Wired this in `crates/vm/levm/src/constants.rs` (new constants `SYS_CALL_GAS_LIMIT_AMSTERDAM`, `SYS_CALL_STATE_GAS_RESERVOIR`, `SYSTEM_MAX_SSTORES_PER_CALL`), `crates/vm/backends/levm/mod.rs:2533` (system-call env uses the bumped budget on Amsterdam+), and `crates/vm/levm/src/utils.rs:451` (reservoir setup in `prepare_execution` special-cases `is_system_call` and bypasses `TX_MAX_GAS_LIMIT_AMSTERDAM`).

4. **Test parity** — Replaced the now-obsolete dynamic-CPSB unit tests with one that pins CPSB = 1174 across a spread of `block_gas_limit` inputs (1, 5M, 30M, 96M, 120M, 500M, u64::MAX).

## Verification

- `cargo fmt --all`, `cargo check --workspace --tests`, `cargo clippy -p ethrex-levm -p ethrex-vm -p ethrex-blockchain --tests` all clean.
- `levm::eip8037_*` unit tests: 28/28 pass.
- `blockchain::builder_validator_parity_tests`: 14/14 pass.
- `blockchain::mempool_tests`: 16/16 pass.
- `make -C tooling/ef_tests/blockchain test` against `bal@v5.7.0` Amsterdam fixtures: identical pass/fail set vs. baseline (131 pass / 55 fail in `eip8037_state_creation_gas_cost_increase`, all 55 pre-existing fixture-vs-spec drift). The `bal@v5.7.0` set predates PR 11573 so it does not exercise the new system-call gas-limit; that gap will be covered when EEST publishes a fixtures tag matching PR 11573.

## Test plan

- [ ] CI green (lint, build, unit, ef-tests-blockchain, ef-tests-state)
- [ ] Spot-check on bal-devnet-5 kurtosis fixture that system calls (EIP-2935, 4788, 7002, 7251) execute to completion without OOG under heavier state-creation load

### DIFF
--- a/crates/vm/levm/src/constants.rs
+++ b/crates/vm/levm/src/constants.rs
@@ -20,17 +20,26 @@ pub const MEMORY_EXPANSION_QUOTIENT: u64 = 512;
 // Dedicated gas limit for system calls according to EIPs 2935, 4788, 7002 and 7251
 pub const SYS_CALL_GAS_LIMIT: u64 = 30000000;
 
-// EIP-8037 (ethereum/EIPs#11573): per-fork system-call gas limit. From Amsterdam, system
-// calls receive an extra `STATE_BYTES_PER_STORAGE_SET * CPSB * SYSTEM_MAX_SSTORES_PER_CALL`
-// budget on top of the legacy 30M, placed in `state_gas_reservoir`. This preserves the
+// EIP-8037 (ethereum/EIPs#11573): per-system-call upper bound on new storage slots
+// written. Matches MAX_WITHDRAWAL_REQUESTS_PER_BLOCK (EIP-7002), the largest per-block
+// bound across the existing system contracts.
+pub const SYSTEM_MAX_SSTORES_PER_CALL: u64 = 16;
+
+// EIP-8037: extra state-gas budget reserved for system contracts that perform up to
+// SYSTEM_MAX_SSTORES_PER_CALL new SSTOREs per invocation. Derived symbolically from
+// the EIP-8037 byte constant so the bound stays in lockstep with
+// SYSTEM_MAX_SSTORES_PER_CALL. The CPSB factor (1174) is inlined here because
+// `cost_per_state_byte` is a runtime fn; if CPSB ever changes, this constant and
+// `cost_per_state_byte` must move together.
+pub const SYS_CALL_STATE_GAS_RESERVOIR: u64 =
+    crate::gas_cost::STATE_BYTES_PER_STORAGE_SET * 1174 * SYSTEM_MAX_SSTORES_PER_CALL;
+
+// EIP-8037: From Amsterdam, system calls receive the legacy 30M execution budget plus
+// SYS_CALL_STATE_GAS_RESERVOIR placed in `state_gas_reservoir`. This preserves the
 // existing 30M execution margin under the higher state-creation costs.
 //
 // 30_000_000 + 32 * 1174 * 16 = 30_601_088
 pub const SYS_CALL_GAS_LIMIT_AMSTERDAM: u64 = SYS_CALL_GAS_LIMIT + SYS_CALL_STATE_GAS_RESERVOIR;
-// Extra state-gas budget reserved for system contracts that perform up to
-// SYSTEM_MAX_SSTORES_PER_CALL (=16) new SSTOREs per invocation (matches
-// MAX_WITHDRAWAL_REQUESTS_PER_BLOCK from EIP-7002).
-pub const SYS_CALL_STATE_GAS_RESERVOIR: u64 = 601_088; // 32 * 1174 * 16
 
 // Transaction costs in gas
 pub const TX_BASE_COST: u64 = 21000;
--- a/crates/vm/levm/src/constants.rs
+++ b/crates/vm/levm/src/constants.rs
@@ -20,6 +20,18 @@ pub const MEMORY_EXPANSION_QUOTIENT: u64 = 512;
 // Dedicated gas limit for system calls according to EIPs 2935, 4788, 7002 and 7251
 pub const SYS_CALL_GAS_LIMIT: u64 = 30000000;
 
+// EIP-8037 (ethereum/EIPs#11573): per-fork system-call gas limit. From Amsterdam, system
+// calls receive an extra `STATE_BYTES_PER_STORAGE_SET * CPSB * SYSTEM_MAX_SSTORES_PER_CALL`
+// budget on top of the legacy 30M, placed in `state_gas_reservoir`. This preserves the
+// existing 30M execution margin under the higher state-creation costs.
+//
+// 30_000_000 + 32 * 1174 * 16 = 30_601_088
+pub const SYS_CALL_GAS_LIMIT_AMSTERDAM: u64 = SYS_CALL_GAS_LIMIT + SYS_CALL_STATE_GAS_RESERVOIR;
+// Extra state-gas budget reserved for system contracts that perform up to
+// SYSTEM_MAX_SSTORES_PER_CALL (=16) new SSTOREs per invocation (matches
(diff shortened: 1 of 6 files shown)


ilitteri (approved):

crates/vm/levm/src/gas_cost.rs
```diff
@@ -164,22 +164,15 @@ pub const CREATE_BASE_COST: u64 = 32000;
...
-/// regardless of `block_gas_limit`. The dynamic formula (BLOCKS_PER_YEAR /
-/// TARGET_STATE_GROWTH_PER_YEAR / CPSB_SIGNIFICANT_BITS / CPSB_OFFSET) is preserved
-/// in the consts above so this commit can be reverted with a single `git revert` to
-/// restore the formula body. See execution-specs#2687.
+/// Pinned at 1174, the value derived from a 100 GiB/year state-growth target at a
+/// 96M-gas reference block limit. The original draft computed this dynamically from
+/// the block gas limit; PR 11573 collapses that to a single fixed constant, so
+/// `block_gas_limit` is no longer an input. The argument is retained to keep call
+/// sites stable; if a future EIP re-derives CPSB, change the body.
 pub fn cost_per_state_byte(_block_gas_limit: u64) -> u64 {
```
iovoid, line 176: Do we still need this function if the value will be left fixed?

crates/vm/levm/src/constants.rs
```diff
@@ -20,6 +20,27 @@ pub const MEMORY_EXPANSION_QUOTIENT: u64 = 512;
...
+// `cost_per_state_byte` must move together.
+pub const SYS_CALL_STATE_GAS_RESERVOIR: u64 =
+    crate::gas_cost::STATE_BYTES_PER_STORAGE_SET * 1174 * SYSTEM_MAX_SSTORES_PER_CALL;
+
+// EIP-8037: From Amsterdam, system calls receive the legacy 30M execution budget plus
+// SYS_CALL_STATE_GAS_RESERVOIR placed in `state_gas_reservoir`. This preserves the
+// existing 30M execution margin under the higher state-creation costs.
+//
+// 30_000_000 + 32 * 1174 * 16 = 30_601_088
+pub const SYS_CALL_GAS_LIMIT_AMSTERDAM: u64 = SYS_CALL_GAS_LIMIT + SYS_CALL_STATE_GAS_RESERVOIR;
```
ElFantasma, line 42: `SYS_CALL_GAS_LIMIT_AMSTERDAM = 30_000_000 + 32 * 1174 * 16 = 30_601_088` checks out against the spec. `SYSTEM_MAX_SSTORES_PER_CALL = 16` is correctly anchored to `MAX_WITHDRAWAL_REQUESTS_PER_BLOCK` (EIP-7002, the loosest bound). The hardcoded `1174` here (rather than calling `cost_per_state_byte()`) is justified by the doc — `cost_per_state_byte` isn't `const`. Worth a `const_assert!` (or similar) somewhere that pins `cost_per_state_byte(0) == 1174` so the coupling can't drift silently if a future PR parameterizes CPSB.

crates/vm/levm/src/utils.rs
```diff
@@ -451,14 +452,36 @@ impl<'a> VM<'a> {
...
         if self.env.config.fork >= Fork::Amsterdam {
-            let gas_limit = self.tx.gas_limit();
+            // System calls have `tx.gas_limit() == 0` (the tx is built from
+            // `EIP1559Transaction::default()` in `generic_system_contract_levm`); their
+            // budget lives in `env.gas_limit` instead. For ordinary user txs the two
+            // agree (set together at the env-builder), but for `eth_call`/simulation
+            // paths `env.gas_limit` may default to the block max while the wrapping
+            // `tx` has `gas_limit = 0`. So we read whichever source carries the budget
+            // for the path we're on.
+            let gas_limit = if self.env.is_system_call {
```
ElFantasma, line 470: Doc/impl mismatch: the surrounding comment says *"For ordinary user txs the two agree (set together at the env-builder), but for `eth_call`/simulation paths `env.gas_limit` may default to the block max while the wrapping `tx` has `gas_limit = 0`. So we read whichever source carries the budget for the path we're on"* — but the branch below only switches on `is_system_call`, not on `is_simulation`/eth_call. So for an Amsterdam+ `eth_call` (non-system-call) we hit the `tx.gas_limit()` arm, which is `0`, making `execution_gas = 0` and the reservoir/gas_left logic effectively no-op. In practice that's probably fine because `eth_call` doesn't depend on the `GAS` opcode return value the way real txs do, but the source comment is misleading as written.

Two possible fixes — pick whichever matches your intent:
- **Tighten the comment** to system-call vs user-tx only and drop the eth_call digression, since the code doesn't actually treat eth_call specially.
- **Add an eth_call/simulation-aware branch** (e.g., key on `self.env.is_simulation` if that flag exists) that also reads `env.gas_limit` instead of `tx.gas_limit()`, if you do want simulations to get the same reservoir handling as system calls.

ElFantasma (approved):
Just a couple of nit suggestions

PR #6548, opened 2026-04-29
perf(l1): batch receipt retrieval in single read txn

azteca1998:
## Summary
- Batch all receipt point lookups into a single `spawn_blocking` task with one shared read transaction, eliminating per-receipt `spawn_blocking` + transaction setup overhead
- Refactor `get_all_block_rpc_receipts` and `get_all_block_receipts` to fetch all receipts in one pass instead of one-by-one
- Add cursor-based benchmark alongside point lookups to compare retrieval strategies

## Context
Targets `feat/automatic-db-migrations` (#6519) so we can add a DB migration if the final approach requires schema changes (e.g. new meta entries or prefix-based storage layout).

## Test plan
- [ ] Verify RPC receipt endpoints return correct results (`eth_getTransactionReceipt`, `eth_getBlockReceipts`, `eth_getRawReceipts`)
- [ ] Check benchmark logs on a synced node to compare point lookup vs cursor performance
- [ ] Verify eth/70 partial receipt fetching still works

### DIFF
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -307,9 +307,6 @@ impl RpcHandler for GetTransactionReceiptRequest {
             block::get_all_block_rpc_receipts(block.header, block.body, storage, Some(index))
                 .await?;
 
-        if receipts.len() != (index as usize + 1) {
-            return Err(RpcErr::Internal("Could not get receipt".to_owned()));
-        }
         serde_json::to_value(receipts.last())
             .map_err(|error| RpcErr::Internal(error.to_string()))
     }
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -307,7 +307,9 @@ impl RpcHandler for GetTransactionReceiptRequest {
             block::get_all_block_rpc_receipts(block.header, block.body, storage, Some(index))
                 .await?;
 
-        // The last receipt in the list is the one at `index` (we fetched 0..=index)
+        if receipts.len() != (index as usize + 1) {
+            return Err(RpcErr::Internal("Could not get receipt".to_owned()));
+        }
         serde_json::to_value(receipts.last())
             .map_err(|error| RpcErr::Internal(error.to_string()))
     }
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -307,8 +307,7 @@ impl RpcHandler for GetTransactionReceiptRequest {
             block::get_all_block_rpc_receipts(block.header, block.body, storage, Some(index))
                 .await?;
 
-        serde_json::to_value(receipts.last())
-            .map_err(|error| RpcErr::Internal(error.to_string()))
+        serde_json::to_value(receipts.last()).map_err(|error| RpcErr::Internal(error.to_string()))
     }
 }
 
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -292,7 +292,7 @@ impl RpcHandler for GetTransactionReceiptRequest {
             "Requested receipt for transaction {:#x}",
             self.transaction_hash,
         );
-        let (block_number, block_hash, index) = match storage
+        let (_block_number, block_hash, index) = match storage
             .get_transaction_location(self.transaction_hash)
             .await?
         {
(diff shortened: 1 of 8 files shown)


crates/storage/store.rs
```diff
@@ -1059,33 +1061,58 @@ impl Store {
...
-                Some(receipt_bytes) => {
-                    let receipt = Receipt::decode(receipt_bytes.as_slice())?;
-                    receipts.push(receipt);
-                    index += 1;
+        tokio::task::spawn_blocking(move || {
+            let txn = backend.begin_read()?;
+            let prefix = block_hash.as_bytes().to_vec();
+            let iter = txn.prefix_iterator(RECEIPTS, &prefix)?;
+            let mut receipts = Vec::new();
+            for result in iter {
```
iovoid, line 1090: If the receipts are missing for the block, we return [] instead of an error.

iovoid (changes requested):

iovoid (approved):

crates/networking/rpc/eth/transaction.rs
```diff
@@ -304,10 +304,13 @@ impl RpcHandler for GetTransactionReceiptRequest {
             None => return Ok(Value::Null),
         };
         let receipts =
-            block::get_all_block_rpc_receipts(block_number, block.header, block.body, storage)
+            block::get_all_block_rpc_receipts(block.header, block.body, storage, Some(index))
                 .await?;
 
-        serde_json::to_value(receipts.get(index as usize))
+        if receipts.len() != (index as usize + 1) {
```
ElFantasma: This length check is now belt-and-suspenders — `get_all_block_rpc_receipts` already returns `RpcErr::Internal` when `all_receipts.len() != fetch_count` (block.rs:66), so the inner zip never silently truncates. Can be dropped. (The greptile P1 about `receipts.last()` being unsafe is mistaken for the same reason — the length guard is upstream.)

crates/storage/migrations.rs
```diff
@@ -93,6 +95,79 @@ fn write_metadata_version(db_path: &Path, version: u64) -> Result<(), StoreError
...
 
+/// Migrates the RECEIPTS table key format from RLP-encoded `(BlockHash, u64)`
+/// to raw `block_hash (32B) || index (8B big-endian u64)`.
+///
+/// This enables efficient cursor-based prefix iteration by block hash.
+///
+/// Crash safety: if interrupted mid-migration, metadata still says v1,
+/// so the migration restarts from scratch on next boot. Keys that fail
+/// RLP decode are assumed to be already migrated and are skipped.
+fn migrate_1_to_2(backend: &dyn StorageBackend) -> Result<(), StoreError> {
```
ElFantasma, line 112: **Snapshot semantics worth pinning down**: this holds one `begin_read()` iterator while issuing multiple `begin_write()` batches that delete old keys and insert new ones into the *same* `RECEIPTS` CF. RocksDB transactions backed by snapshots will hold the iterator's view stable, but the contract isn't documented at the `StorageBackend` trait level. If a backend implementation ever returns a non-snapshot iterator, the migration could either re-read freshly-inserted 40-byte keys (idempotent — they hit the `len() == 40` skip below) or skip old keys that were just deleted (data loss). Worth either documenting the snapshot requirement on the trait or restructuring to materialize the old-key list before any writes.

crates/storage/migrations.rs
```diff
@@ -93,6 +95,79 @@ fn write_metadata_version(db_path: &Path, version: u64) -> Result<(), StoreError
...
+    const BATCH_SIZE: usize = 10_000;
+
+    let txn = backend.begin_read()?;
+    let iter = txn.prefix_iterator(RECEIPTS, &[])?;
```
ElFantasma, line 116: **No test exercises the migration body itself.** `migrations_length_matches_schema_version` and `run_pending_migrations_noop_when_current` cover the framework, not v1→v2. A unit test that seeds old-format keys via `(BlockHash, u64).encode_to_vec()`, runs `migrate_1_to_2`, and verifies the new keys round-trip via `get_receipt`/`get_receipts_for_block` would catch any future regression in the key encoding (e.g., if `to_be_bytes` ever drifts from the migration's reconstruction).

crates/storage/store.rs
```diff
@@ -1059,33 +1061,61 @@ impl Store {
...
+                let (k, v) = result?;
+                if !k.starts_with(&prefix) {
+                    break;
+                }
+                if k.len() != 40 {
+                    continue;
+                }
+                // Skip entries before start_index (for eth/70 partial requests)
+                if start_index > 0 {
```
ElFantasma, line 1099: For `start_index > 0` (eth/70 partial path), the prefix iterator starts at the block-hash boundary and skips entries below `start_index` linearly. A `seek` to `block_hash || start_index.to_be_bytes()` would jump straight to the target — minor on small batches but matters for partial requests deep into a block. Follow-up if `prefix_iterator` doesn't expose a seek-to-key today.

ElFantasma (approved):
LGTM, but left some suggestions

avilagaston9 (approved):

pablodeymo (approved):

ilitteri (changes requested):
We are testing a potential memory usage issue during the migration. Blocking until we know more.

pablodeymo (changes requested):
I change my approve to match my concern about the migration

PR #6549, opened 2026-04-29
fix(l1): use hash precompiles in eip8025 hash tree root

benbencik:
**Motivation**
Earlier work #6365 added `&dyn Crypto`, so zkVM can route hashing to precompiles. However, it is not applied everywhere in the eip-8025 execution path.

**Description**
The [execution_program](https://github.com/lambdaclass/ethrex/blob/main/crates/guest-program/src/l1/program.rs#L73) is calling `hash_tree_root` with `libssz_merkle::Sha2Hasher`. This fix implements `libssz_merkle::Sha256Hasher` trait for `&dyn Crypto` such that hash precompiles can be used in computing the hash of the tree root.

This fix **significantly reduces the AIR-costs** inside ZisK v0.16.1, and it is the main reason for the previously observed inefficiency in #6524. Profiling on mainnet blocks: 24949787 - 24949836:
- Average cost (main `1e85b1c36`): 57.6B
- Average cost (this change): 45.2B


iovoid (approved):

crates/guest-program/src/l1/program.rs
```diff
@@ -54,6 +57,18 @@ pub fn execution_program(
     })
 }
 
+/// Wrapper to pass Crypto into libssz_merkle::hash_tree_root,
+/// so hashing is computed by precompiles
```
ElFantasma: nit: the doc comment is on the right track but doesn't explain *why* a wrapper is needed. The actual reason is the orphan rule — `Sha256Hasher` lives in `libssz_merkle` and `Arc<dyn Crypto>` lives in `ethrex_crypto`, so we can't `impl Sha256Hasher for Arc<dyn Crypto>` directly. A future contributor wondering "why not just impl on `Arc<dyn Crypto>`?" would benefit from one extra line:

```rust
/// Wrapper to bridge `ethrex_crypto::Crypto` to `libssz_merkle::Sha256Hasher`,
/// so `hash_tree_root` is computed via crypto precompiles in the zkVM.
/// Required because the orphan rule prevents a direct impl on `Arc<dyn Crypto>`.
```

crates/guest-program/src/l1/program.rs
```diff
@@ -64,13 +79,13 @@ pub fn execution_program(
...
 ) -> Result<ProgramOutput, ExecutionError> {
-    use libssz_merkle::{HashTreeRoot, Sha2Hasher};
+    use libssz_merkle::HashTreeRoot;
 
     let (new_payload_request, execution_witness) = super::decode_eip8025(bytes).map_err(|err| {
         ExecutionError::Internal(format!("failed to decode EIP-8025 input: {err}"))
     })?;
 
-    let request_root = new_payload_request.hash_tree_root(&Sha2Hasher);
+    let request_root = new_payload_request.hash_tree_root(&CryptoWrapper(crypto.clone()));
```
ElFantasma, line 89: nit: a unit test asserting the wrapped path produces a root identical to `&Sha2Hasher` for some sample `NewPayloadRequest` would lock down the equivalence — cheap to write since `CryptoWrapper(Arc::new(NativeCrypto))` and `Sha2Hasher` should be byte-for-byte equal. Not blocking, but it'd catch any future drift if `Crypto::sha256`'s default impl ever changes.

ElFantasma (approved):
LGTM, just a couple of nits.

ilitteri (approved):

crates/guest-program/src/l1/program.rs
benbencik: Thanks for the comment. I did not find a clean way to get a sample `NewPayloadRequest` for the unit test. What do you think about adding `debug_assert_eq` to check the equivalence?

ElFantasma: You're right, drop the nit — `Crypto::sha256` default impl (`provider.rs:181`) is literally `sha2::Sha256::digest`, same as `Sha2Hasher::hash`.

PR #6550, opened 2026-04-29
feat(l1): support canonical SSZ input bytes

jsign:
Adds support for canonical EIP-8025 stateless input bytes in the L1 guest program.

The goal of this PR is to take the "last step" regarding API support towards the [execution-specs definition for the guest program](https://github.com/ethereum/execution-specs/blob/b1d870f8f06d9eba3d74179c39660cedb98f1f38/src/ethereum/forks/amsterdam/stateless_guest.py#L33-L41). For context and reference, these bytes are SSZ of [StatelessInput](https://github.com/ethereum/execution-specs/blob/b1d870f8f06d9eba3d74179c39660cedb98f1f38/src/ethereum/forks/amsterdam/stateless.py#L92-L118).

Ethrex guest program still supports both the current SSZ bytes (what I now call `Legacy`) and the new canonical SSZ bytes (which I now call `Canonical`). The SSZ byte decoder now expects a single byte at the front, which indicates which bytes were sent, so it can decode accordingly.

The reality is that this two-style support is somewhat temporary, since medium/long term we should aim the canonical being the only one. But prefer to walk this middle step since we need all guest program users to always support generating canonical inputs.

The main goal here is to have a way for Ethrex to receive the [`sszInputBytes` from the zkEVM EEST fixtures](https://github.com/ethereum/execution-spec-tests/releases?q=zkevm&expanded=true) directly without any middle layer manipulation, which always adds complexity and more surface for bugs in transformations.

Two main notes to have in mind:
1. The new `Canonical` bytes have format `0x02 + size(SSZ_bytes_EEST) + SSZ_bytes_EEST + size(Ethrex_ChainConfig_rkyv) + Ethrex_ChainConfig_rkyv`. The main reason for this is that the full `ChainConfig` isn't standarized in EEST. [Started a discussion](https://hackmd.io/@jsign/eth-zkevm-chainconfig-statelessinput) to see if this can be more formalized in the specs.
2. The way the canonical `ExecutionWitness` is transformed into usable tries leverages the existing method that was useful for the `Legacy` style, but which was used at the host level. I suspect doing this in the guest will be slow, but for now I'm only interested in Ethrex supporting spec input bytes. I think further PRs/optimizations can always find more optimal ways to do this. 

Summary of changes:
- Introduces versioned EIP-8025 decoding:
  - `0x00` for the existing legacy payload format
  - `0x01` for canonical stateless inputs
- Adds canonical SSZ input types for Amsterdam payloads, execution witnesses, chain config, and public keys.
- Reconstructs Amsterdam `NewPayloadRequest` values into L1 blocks, including block access list hash, slot number, request hash, block hash, and versioned hash validation.
- Converts canonical witnesses into the existing execution witness path so canonical inputs reuse the current stateless execution flow.
- Validates block access list hashes during guest block execution.

Leaving this PR as a draft for a bit since I want to run some extra tests in ere-guests using EEST fixtures and double-check the whole pipeline of running official EEST fixture releases `statelessInputBytes` works as expected.


crates/common/types/eip8025_ssz.rs
```diff
@@ -17,7 +17,11 @@ use super::requests::EncodedRequests;
 /// `MAX_TRANSACTIONS_PER_PAYLOAD` (Electra).
 const MAX_TRANSACTIONS_PER_PAYLOAD: usize = 1_048_576;
 /// `MAX_WITHDRAWALS_PER_PAYLOAD` (Electra).
-const MAX_WITHDRAWALS_PER_PAYLOAD: usize = 16;
+/// TODO: the specs have a non-compliant value compared to consensus
+/// specs. Whenever the specs can resolve an underlying issue,
+/// this value should be updated.
+/// See https://github.com/ethereum/execution-specs/blob/ec23140720d6a9257a907c470ba1874623bd7b50/src/ethereum/forks/amsterdam/stateless_ssz.py#L40-L43
+const MAX_WITHDRAWALS_PER_PAYLOAD: usize = 65536;
```
jsign, line 24: There is a current issue in the testing framework that forced us to temporarily use a non-consensus valid value. For now, sticking to the correct max since, if not, the hash tree root won't match the fixture expectations.

This will be eventually fixed, but leaving a comment here since somebody else might find the value surprising.

crates/common/types/eip8025_ssz.rs
```diff
@@ -169,6 +175,31 @@ pub struct ExecutionPayload {
     pub excess_blob_gas: u64,
 }
 
+/// SSZ `ExecutionPayload` execution payload V4.
```
jsign, line 178: The current [EEST zkVM releases](https://github.com/ethereum/execution-spec-tests/releases/tag/zkevm%40v0.3.4) are based on the latest BAL devnet.

The current `NewPayloadRequest` is based on the current active fork in mainnet. I'm adding support now for new execution witness and new payload request versions for Amsterdam.

I think at some point it might be worth moving these structs to another module for better organisation or similar, but it might be good as is for a while.

crates/guest-program/src/common/execution.rs
```diff
@@ -166,6 +166,18 @@ where
...
+            report_cycles("validate_block_access_list_hash", || {
+                validate_block_access_list_hash(
+                    &block.header,
+                    &chain_config,
+                    bal,
+                    block.body.transactions.len(),
+                )
+                .map_err(ExecutionError::BlockValidation)
+            })?;
+        }
```
jsign, line 179: Since prev code had `_bal`, this validation was missing.

crates/guest-program/src/l1/output.rs
```diff
@@ -36,24 +36,28 @@ impl ProgramOutput {
...
+/// `chain_id` (8 bytes).
 #[cfg(feature = "eip-8025")]
 #[derive(Serialize, Deserialize)]
 pub struct ProgramOutput {
     /// The `hash_tree_root` of the `NewPayloadRequest`.
     pub new_payload_request_root: [u8; 32],
     /// Whether execution was valid.
     pub valid: bool,
+    /// Chain ID from the stateless validation chain configuration.
+    pub chain_id: u64,
```
jsign, line 50: Technically is needed for the spec [StatelessValidationResult](https://github.com/ethereum/execution-specs/blob/ec23140720d6a9257a907c470ba1874623bd7b50/src/ethereum/forks/amsterdam/stateless_ssz.py#L171).

This part of the API is a bit unstable, so not sure yet if this will be the final form. Opened a convo to explore this a bit more with [this doc](https://hackmd.io/@jsign/eth-zkevm-chainconfig-statelessinput).

For now sticking to the current specs -- we can easily adjust whenever that convo comes to some conclusion and we change the specs.

crates/guest-program/src/l1/program.rs
```diff
@@ -57,26 +60,114 @@ pub fn execution_program(
...
+            })
+        }
+    }
+}
 
-    Ok(ProgramOutput {
-        new_payload_request_root: request_root,
-        valid,
-    })
+#[cfg(feature = "eip-8025")]
```
jsign, line 129: Next four methods are simply extracted existing logic so can be reused between the legacy and canonical block processing.

crates/guest-program/src/l1/program.rs
```diff
@@ -200,6 +340,53 @@ fn validate_versioned_hashes(
     Ok(())
 }
 
+#[cfg(feature = "eip-8025")]
```
jsign, line 359: As mentioned in the PR description, in my PR I'm focusing on Ethrex working with canonical SSZ bytes and not necessarily being efficient.

I've the sensation the canonical `ExecutionWitness` validation might have some other more efficient form in the future, but I think that can be done in later work.

Or maybe this approach is totally fine -- requires some benchmarks to be sure.

crates/guest-program/src/l1/program.rs
```diff
@@ -200,6 +340,53 @@ fn validate_versioned_hashes(
...
+            stateless_input.chain_config.chain_id, chain_config.chain_id
+        )));
+    }
+
+    let block_number = stateless_input
+        .new_payload_request
+        .execution_payload
+        .block_number;
+    let rpc_witness = canonical_execution_witness_to_rpc(stateless_input.witness);
+    let execution_witness = rpc_witness.into_execution_witness(chain_config, block_number)?;
```
jsign, line 402: To highlight reg my previous comment: these are the two lines that might require to be evaluated if are optimal, since the RPC->Execution witness logic today I think was mainly thought to happen in the host.

crates/guest-program/src/l1/input.rs
```diff
@@ -46,20 +54,125 @@ pub fn encode_eip8025(
...
+
+/// Mirrors `SszStatelessInput` from the Amsterdam stateless-validation spec.
+#[cfg(feature = "eip-8025")]
+#[derive(Debug, Clone, PartialEq, Eq, libssz_derive::SszEncode, libssz_derive::SszDecode)]
+pub struct CanonicalStatelessInput {
+    pub new_payload_request: ethrex_common::types::eip8025_ssz::NewPayloadRequestAmsterdam,
+    pub witness: CanonicalExecutionWitness,
+    pub chain_config: CanonicalChainConfig,
+    pub public_keys:
+        libssz_types::SszList<libssz_types::SszList<u8, MAX_BYTES_PER_PUBLIC_KEY>, MAX_PUBLIC_KEYS>,
```
ElFantasma, line 115: `public_keys` is decoded as part of `CanonicalStatelessInput` (mirroring `SszStatelessInput` from the spec) but I don't see it referenced anywhere downstream — `validate_eip8025_canonical_execution` only consumes `new_payload_request`, `witness`, and `chain_config`.

With `MAX_PUBLIC_KEYS = 1 << 20` and `MAX_BYTES_PER_PUBLIC_KEY = 65`, this field's allocation budget alone is ~67 MB per input. Two questions:

1. Is this intended to be used for ecrecover-skipping (so transactions don't need on-chain key recovery in the guest)? If so, that's a follow-up worth marking with a TODO at the use site (`validate_eip8025_canonical_execution`) so it isn't silently dropped.
2. If it's purely there to match the spec layout for now, a one-line comment here (`// Reserved for sender-recovery acceleration; not consumed yet`) would help future readers.

crates/guest-program/src/l1/program.rs
```diff
@@ -200,6 +340,53 @@ fn validate_versioned_hashes(
...
+
+    fn copy_ssz_bytes<const MAX_BYTES: usize>(
+        bytes: &libssz_types::SszList<u8, MAX_BYTES>,
+    ) -> Bytes {
+        Bytes::from(bytes.iter().copied().collect::<Vec<u8>>())
+    }
+
+    ethrex_common::types::block_execution_witness::RpcExecutionWitness {
+        state: witness.state.iter().map(copy_ssz_bytes).collect(),
+        keys: Vec::new(),
```
ElFantasma, line 378: `keys: Vec::new()` looks like a magic empty value that will trip up the next reader. The legacy/RPC paths populate `keys` from the witness; here we drop it on the floor.

If `RpcExecutionWitness::into_execution_witness` reconstructs keys from the state nodes when `keys` is empty, that's fine — but please add a one-line comment saying so. Otherwise, this is silently building a witness that may be missing data the executor expects.

crates/guest-program/src/l1/program.rs
```diff
@@ -57,26 +60,114 @@ pub fn execution_program(
...
+            address: Address::from_slice(&w.address.0),
+            amount: w.amount,
+        })
+        .collect()
+}
+
+#[cfg(feature = "eip-8025")]
+fn base_fee_per_gas_from_le_bytes(bytes: &[u8; 32]) -> Result<u64, String> {
+    Ok(u64::from_le_bytes(
+        bytes[..8]
```
ElFantasma, line 166: Pre-existing behavior (this PR just extracted it into a helper), so feel free to defer — but worth noting now that there's a single extraction site:

`base_fee_per_gas` in the SSZ `ExecutionPayload` is `[u8; 32]` LE (256-bit). This helper takes only `bytes[..8]` and silently ignores the upper 24 bytes. A malformed input with a non-zero high half would yield a wrong `base_fee_per_gas` rather than an explicit decode error; the `block_hash` mismatch downstream would still catch it, but with a confusing error.

Suggestion: add `if bytes[8..].iter().any(|b| *b != 0) { return Err("base_fee_per_gas overflows u64".into()); }` before the conversion. Cheap and turns a silent miscompute into a clear error.

crates/guest-program/src/l1/input.rs
jsign: Yes, great point about this, it was my miss to add a comment about this.
In the specs today we have the `public_keys` field but the spec still have this TODO to actually fill the values.

The goal is that the public keys will be provided as part of `StatelessInput` and the guest program can only check that the txs signatures validate against those provided public keys. 

So, things are still validated and not trusting inputs; but we have measured last year that verifying a signature against a public key is more efficient than doing the full pubkey recovery with ecdsa. So that is the main intention.

I'll will be implementing this proper `public_keys` filling in the specs so it can be properly used quite soon, so it will be part of the next spec release.

In the meantime, I'll add a comment as suggested to avoid confusion.

crates/guest-program/src/l1/program.rs
jsign: I actually imagine this block code to be optimized eventually by your tema, since as mentioned in my other comments, I don't think this logic was thought to be executed inside the guest program.

But yep, I'll add a comment to make things a bit more clear in the meanwhile.

(Sidenote: in the near future we might discover that we might add `keys` into `StatlessInput` -- but I think we should justify that with some benchmarks proving this makes things more efficient for all guest programs instead of including it just because was part of some legacy RPC)

jsign:
@ElFantasma, thanks for the review. I pushed a commit addressing your feedback.

docs/eip-8025.md
```diff
@@ -401,8 +401,8 @@ Output: ProgramOutput { initial_state_hash, final_state_hash, last_block_hash, c
 **EIP-8025 mode** (with `eip-8025` feature):
 ```
 Input:  (NewPayloadRequest [SSZ], ExecutionWitness [rkyv])
-Output: ProgramOutput { new_payload_request_root: [u8; 32], valid: bool }
-        → 33 bytes: 32-byte root + 1-byte boolean
+Output: ProgramOutput { new_payload_request_root: [u8; 32], valid: bool, chain_id: u64 }
+        → 41 bytes: 32-byte root + 1-byte boolean + 8-byte chain_id
```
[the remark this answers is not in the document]
iovoid: Agreed

iovoid (approved):

ilitteri (approved):

edg-l (approved):

PR #6553, opened 2026-04-30
fix(l1): fix storage range fetching for big accounts

iovoid:
**Motivation**

Storage range download would get stuck on large networks with small pivot times.

**Description**

We always queued bulk tasks from [0, MAX] regardless of an account's existing intervals, relying on the response handler's big account split path to re-fire each call. That path only fires when a peer's response is truncated. When a peer fully covered a known-big account in one response, no branch ran, and the account's intervals stayed in accounts_with_storage_root forever.


crates/networking/p2p/snap/client.rs
```diff
@@ -557,19 +557,65 @@ pub async fn request_storage_ranges(
...
+    // fully without hitting their response limit on it: the split path doesn't
+    // fire, no per-interval tasks get queued, intervals never drain, the
+    // account is stuck pending forever even after its data is on disk.
     let mut tasks_queue_not_started = VecDeque::<StorageTask>::new();
-    for i in 0..chunk_count {
-        let chunk_start = chunk_size * i;
-        let chunk_end = (chunk_start + chunk_size).min(accounts_by_root_hash.len());
+    let mut bulk_chunk_start: Option<usize> = None;
+    for (i, (_, accounts)) in accounts_by_root_hash.iter().enumerate() {
+        let first_account = *accounts.first().ok_or_else(|| {
```
ElFantasma: Implicit invariant: this uses `first_account` to look up intervals for the entire group (multiple accounts sharing the same `storage_root`), assuming all accounts in the group share the same interval state. The pre-existing handler at line 836 has the same assumption, so this isn't introduced here — but it's worth a doc comment near `accounts_by_root_hash`'s declaration (~line 549) noting that all accounts in a group are expected to share intervals because they share storage. If that invariant ever breaks (e.g., one account in a group finalizes while a sibling is still pending), this scheduling code silently uses the first account's state for all of them.

crates/networking/p2p/snap/client.rs
```diff
@@ -869,6 +915,46 @@ pub async fn request_storage_ranges(
...
+                        .get(account)
+                        && !old_intervals.is_empty()
+                    {
+                        acc_hash = *account;
+                    }
+                }
+                // acc_hash stays zero when a sibling per-interval task for the
+                // same account already drained the last interval and finalized
+                // it earlier in this call's loop — there's nothing left to do.
+                if !acc_hash.is_zero() {
```
ElFantasma: Worth pinning the contrast with the existing similar block at lines 766-768: that one `panic!`s when `acc_hash.is_zero()`, while this one silently skips. The reason is given in the comment above (sibling already drained), but the differing-invariant-with-sibling-block is subtle. A one-liner like `// (Compare to lines 766-768 which panic on zero — different scenario: that path runs only on remaining_start < remaining_end, so a missing acc_hash there genuinely means corruption.)` would help future readers tell the two paths apart.

crates/networking/p2p/snap/client.rs
```diff
@@ -869,6 +915,46 @@ pub async fn request_storage_ranges(
...
+                    let (_, old_intervals) = account_storage_roots
```
ElFantasma: nit: this block (940-956) and the existing block at lines 769-787 do the same three-step dance — find acc_hash, remove the matching interval by `end == hash_end`, and on empty-intervals mark the account `done` + healed. Worth extracting a helper like `clear_completed_interval(account_storage_roots, accounts_by_root_hash, accounts_done, start_index, hash_end) -> Result<(), SnapError>` and calling it from both. Reduces drift risk if the bookkeeping shape changes later.

crates/networking/p2p/snap/client.rs
```diff
@@ -557,19 +557,65 @@ pub async fn request_storage_ranges(
...
+            .get(&first_account)
+            .ok_or_else(|| {
+                SnapError::InternalError(
+                    "Could not find intervals for account while scheduling".to_owned(),
+                )
+            })?
+            .1;
+        if intervals.is_empty() {
+            let chunk_start = *bulk_chunk_start.get_or_insert(i);
+            if i + 1 - chunk_start >= chunk_size {
```
ElFantasma: Off-by-one check: `i + 1 - chunk_start >= chunk_size` flushes when the accumulated batch reaches `chunk_size` accounts (since `i+1` is the exclusive end and `chunk_start` is the inclusive start). That matches `STORAGE_BATCH_SIZE` semantics elsewhere in the file. ✓ Just confirming I read this right.

azteca1998 (approved):

ilitteri (approved):

edg-l (approved):

ElFantasma (approved):

PR #6560, opened 2026-05-01
perf(l1): remove rayon from eip-8025 execution path

benbencik:
**Motivation**

Observed that execution of the Ethrex guest program in ziskVM makes calls to `rayon`. However, the execution inside the zkVM is strictly sequential, so `rayon` provides no benefit and only adds runtime overhead.

**Description**

This change uses the `eip-8025` feature to control parallel execution paths. A few methods are removed from the `eip-8025` build, and some have a new sequential fallback that replaces the parallel iterator with standard iterator.

_Note:_ The `eip-8025` continues to use `std::sync` which is also redundant. Removing it would require a larger refactor (perhaps future work).

**Results**
Profiling in ZisK v0.16.1 on 25 mainnet blocks 24949787 - 24949836:
- Average cost (main `1c5e518f8`): 55.69B
- Average cost (this change): 53.19B

The change decreases costs by 4-5% and reduces the complexity of the code running inside zkVM.


crates/common/Cargo.toml
```diff
@@ -44,7 +44,7 @@ libssz-derive = { workspace = true, optional = true }
 [features]
 default = ["secp256k1"]
 c-kzg = ["ethrex-crypto/c-kzg"]
-secp256k1 = ["dep:secp256k1", "ethrex-crypto/secp256k1"]
+secp256k1 = ["dep:secp256k1", "ethrex-crypto/secp256k1", "dep:rayon"]
```
iovoid: dep:rayon should be on eip-8025 which is the feature that decides whether it's used

iovoid (changes requested):
The commits should be signed.

crates/common/Cargo.toml
benbencik: Sorry about the unsigned commits, I just fixed that. You are right that this is an odd placement for the feature. However, the `eip-8025` feature is supposed to do the exact opposite. So, rayon should not be used when `eip-8025` is enabled.

I changed it such that the `rayon` feature is part of the default feature. That should maintain the current behavior. Before this PR, rayon was a required dependency. Now it is optional, but included in the default.

crates/vm/backends/levm/mod.rs
```diff
@@ -271,7 +288,11 @@ impl LEVM {
                     EvmError::Transaction(format!("Couldn't recover addresses with error: {error}"))
                 })?;
 
-        // When BAL is provided (Amsterdam+ validation path): use parallel execution
+        // eip-8025 (zkVM guest) is intentionally sequential.
```
iovoid: We still want to validate the BAL is correct, right?

iovoid (changes requested):
Check the CI, you need to update the lockfiles with `make update-cargo-lock`.

crates/vm/backends/levm/mod.rs
benbencik: From my undetstanding, BAL is not needed in the guest-program as can be seen [here](https://github.com/lambdaclass/ethrex/blob/main/crates/guest-program/src/common/execution.rs#L127). 

The function `execute_block_pipeline` is actually not in the execution path of `eip-8025`. I updated the comment to make this clear. Then perhaps the more rigorous way is to do conditional compile on the whole `execute_block_pipeline`, but that results in problems in other parts of the codebase. 

In the current approach, I keep the whole function and gate just the parallel code. This seemed simpler and less invasive. What do you think?

crates/blockchain/blockchain.rs
iovoid, line 1: Currently execute_block is meant to be the sequential option (used for ZK) and execute_block_pipeline the parallel one, which call identically named LEVM functions.

Just to clarify the intent, the reason to change execute_block_pipeline is to fully remove rayon as a dependency?

iovoid (commented):
CI failed due to fmt again.

crates/blockchain/blockchain.rs
benbencik: Yes, exactly. If you think that is not a good idea, I can rewrite it such that the dependency on rayon stays, but is not used in `eip-8025`.

benbencik: There is a similar problem with `once_cell::sync` atomics that cause overhead in the `eip-8025` execution. I started working on that as well.

iovoid: I think it's ok to leave as-is for now.

The sequential/parallel path split should be refactored to reduce the number of cfg-gates needed, but that's out of scope for this PR.

iovoid (approved):

ilitteri (approved):

edg-l (approved):

PR #6562, opened 2026-05-01
feat(l1): add --no-precompile-cache flag for benchmarking

ilitteri:
**Motivation**

The per-block precompile result cache (`PrecompileCache` in `crates/vm/levm/src/precompiles.rs`) is always on during block execution. To benchmark how much the cache contributes to block-execution time, we need an opt-in switch to disable it without otherwise altering the execution pipeline.

**Description**

Adds a `--no-precompile-cache` CLI flag (env `ETHREX_NO_PRECOMPILE_CACHE`, default off) that disables the per-block precompile result cache for the full-node entry point and the `import-bench` subcommand. Default behavior is preserved everywhere else.

- `BlockchainOptions` gains a new `precompile_cache_enabled: bool` field (defaults to `true`), which `Blockchain::execute_block_pipeline` passes into `CachingDatabase::new`.
- `CachingDatabase::precompile_cache` becomes `Option<PrecompileCache>` so that, when disabled, no cache is allocated. `Database::precompile_cache()` returns `self.precompile_cache.as_ref()`. `execute_precompile` already short-circuits cleanly on `None`.
- Wired only at two `BlockchainOptions` construction sites:
  - `cmd/ethrex/initializers.rs` — full-node startup
  - `cmd/ethrex/cli.rs` — `Subcommand::ImportBench`
- Other entry points (plain `import`, L2, ef-tests, migrations, state_v2, benches) keep their existing cached behavior via `Default`.

### Files changed

| File | Change |
|------|--------|
| `crates/vm/levm/src/db/mod.rs` | `CachingDatabase::precompile_cache` is now `Option<PrecompileCache>`; `new` takes `precompile_cache_enabled: bool`; trait impl returns `as_ref()` |
| `crates/blockchain/blockchain.rs` | New `precompile_cache_enabled: bool` on `BlockchainOptions` (default `true`); passed to `CachingDatabase::new` in `execute_block_pipeline` |
| `cmd/ethrex/cli.rs` | New `--no-precompile-cache` clap arg; `Subcommand::ImportBench` reads it |
| `cmd/ethrex/initializers.rs` | Full-node `BlockchainOptions` reads `!opts.no_precompile_cache` |
| `cmd/ethrex/l2/initializers.rs` | Adds `precompile_cache_enabled: true` (struct literal spelled exhaustively; behavior unchanged — L2 keeps cache on) |

### Usage

```bash
# baseline (cache on, default)
ethrex import-bench /path/to/blocks.rlp

# disabled (for A/B benchmarking)
ethrex --no-precompile-cache import-bench /path/to/blocks.rlp

# or via env
ETHREX_NO_PRECOMPILE_CACHE=true ethrex …
```

The flag sits on the parent `Options` struct, so it goes before the subcommand (mirroring `--no-migrate`).

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A, no `Store` changes.


pablodeymo (approved):
The code is righr t.

One nit, the new CLI arg should be documented in: `docs/CLI.md`

iovoid (approved):
Only issue is the CLI docs not being updated, but the CI already flagged that.

azteca1998 (approved):

edg-l (approved):

avilagaston9 (approved):

PR #6570, opened 2026-05-05
perf(l1): codeless EOA fast path

iovoid:
**Motivation**

For transfers to codeless EOAs we perform some setup that isn't needed since they can't revert.

**Description**

If the account has no code, skip vm.run_execution() which does nothing since the target account has no code.


crates/vm/levm/src/vm.rs
```diff
@@ -580,6 +580,23 @@ impl<'a> VM<'a> {
...
+                .gas_limit
+                .checked_sub(self.current_call_frame.gas_remaining as u64)
+                .ok_or(InternalError::Underflow)?;
+            let context_result = ContextResult {
+                result: TxResult::Success,
+                gas_used,
+                gas_spent: gas_used,
+                output: Bytes::new(),
+            };
+            return self.finalize_execution(context_result);
```
ElFantasma, line 794: **BAL checkpoint skip rationale**: the slow path below this `return` takes a `bal_checkpoint` snapshot post-prepare_execution so that a top-level call failure can roll back inner-call BAL entries. The fast path skips that because:
- The tx can't fail past this point — there's no bytecode to execute, no revert possible.
- No inner calls happen — nothing to roll back.
- The value transfer + nonce increment from `prepare_execution` are already "committed" to BAL via the recorder hooks (which run inside `prepare_execution`'s setup, not at the checkpoint).

Reasoning is sound but worth pinning as a comment here so a future contributor doesn't "add the missing checkpoint" thinking it's an oversight. Suggested wording: *"Fast path skips the BAL checkpoint below because a codeless transfer cannot fail past this point — there's nothing to roll back."*

crates/vm/levm/src/vm.rs
```diff
@@ -603,6 +620,23 @@ impl<'a> VM<'a> {
         Ok(report)
     }
 
+    /// Must run after `prepare_execution` so EIP-7702 delegation is already resolved into
+    /// `bytecode`. Inspired by Nethermind's `IsSimpleTransferFastPathCandidate`.
+    #[inline(always)]
+    fn is_simple_transfer_fast_path(&self) -> bool {
```
ElFantasma, line 823: Conditions look right — would feel safer with a Nethermind-cross-reference comment naming the exact `IsSimpleTransferFastPathCandidate` revision we copied, since it's the upstream invariant we're inheriting. If their list ever grows a condition (e.g., they add a check we haven't replicated), the comment makes it easier to spot the divergence on next sync.

One case to think about: empty bytecode + non-empty `access_list` (EIP-2930). The access-list pre-warming has already happened by `prepare_execution`, and there's no bytecode to actually consume those entries — but BAL recording for access-list addresses might rely on `run_execution` walking the list. Does anything need the access-list entries to be recorded in BAL during the no-code path? If not, we're good; if yes, that's a 6th condition (`access_list().is_empty()`).

ElFantasma (commented):
Maybe it is worth a new test? 
For a fast path bypassing `run_execution` entirely, any divergence between fast and slow would manifest as silent state-root divergence on simple-transfer-heavy blocks. The EF blockchain suite likely covers this incidentally, but a focused unit test that runs the same simple-transfer through both paths and asserts identical `ExecutionReport` (gas_used, gas_spent, output, logs, state diff) would lock down the parity. Cheap to write — construct a tx with empty calldata, recipient with no code, run it once with the fast-path conditions met and once forced into the slow path (e.g., temporarily flip a condition), compare reports.

crates/vm/levm/src/vm.rs
edg-l: agree on adding the comment

edg-l (approved):

ilitteri (approved):

azteca1998 (approved):
note that there are some interesting comments to check before merging

crates/vm/levm/src/vm.rs
iovoid: Dropped the reference to these internals in 8c6f562, the reason for requiring the invariant could also diverge.

crates/vm/levm/src/vm.rs
iovoid: Added in 8c6f562

ElFantasma (approved):

Issue #2689, opened 2025-05-07
Bring nodes to memory on cache miss


Arkenan:
Currently, if we get a node and don't find it in the cache, we get it from the db. However, we don't put it in the cache, so if we get the same node multiple times, we won't retrieve it.

Note: this is a cache policy, it shouldn't leak outside of the cache and should be easy to change if the access pattern proves this policy to not be useful.

Arkenan:
If we have snapshots, as per #1997 we won't access the trie in execution. If we have in-place updates as per #2989 we will already be doing this. Closing this issue as it's redundant by those two.

PR #2827, opened 2025-05-16
ci(l1,l2): add 'build block' benchmark to PR checks

fkrause98:
**Motivation**

Make the "build block" benchmark run in the CI.

PR #2816, opened 2025-05-15
feat(l2): add flag to delay the watcher until a trusted block

LeanSerra:
**Motivation**

L1 reorgs can left the L2 in a bad state if a reorged block deposits' are processed.

**Description**

- Add the flag `watcher_block_delay` with default value 0 that represents the amount of blocks of delay the l1 watcher has. 
   - If the latest block in l1 is 100 and we set this delay to 10 l1 watcher will look for deposits until block 90
- Add logs and return empty from the function if 
   - We are too close to genesis (current_block - block_delay < 0)
   - We changed the block delay and now the last block verified by the contract is no longer a trusted block

Closes #2187

PR #6574, opened 2026-05-06
feat(l1): bal-devnet-6 support (EIP-7928 BAL + EIP-8037 2D gas)

edg-l:
**Motivation**

Bring ethrex up to [bal-devnet-6 spec](https://notes.ethereum.org/@ethpandaops/bal-devnet-6) so we can participate in the devnet. Rolls up the work that landed across `bal-devnet-3` → `bal-devnet-4` → `bal-devnet-5` → `bal-devnet-6`. The original `bal-devnet-6` branch is kept for ongoing devnet work; this `*-pr` branch is the one intended for merge.

**Description**

- **EIP-7928 (BAL)** — Block Access List support. Fixtures bumped to snøbal-devnet-6 (`v1.1.0`), `BlockAccessIndex` widened from `u16` to `u32`, shadow recorder for per-tx access validation in parallel execution, `SYSTEM_ADDRESS` exclusion.
- **EIP-8037 (2D gas)** — full EELS PR #2689 semantics:
  - `cost_per_state_byte` pinned to 1174 for bal-devnet-4..6.
  - Top-level / sub-frame `ExceptionalHalt` reclassifies state gas → regular dim (both CALL and CREATE return paths).
  - REVERT cascade keeps `state_gas_credit_against_drain` elevated so credit burns propagate.
  - CREATE-TX top-halt formula: `gross_spill − credit_against_drain − already_reclassified` reclassified to regular dim.
  - EIP-7702 `set_delegation` refund: subtracts from `state_gas_used` / `intrinsic_state_gas_charged` (bal-devnet-7-prep accounting, SELFDESTRUCT-style). See note below.
  - `credit_against_drain` capped by `regular_gas_reclassified` so phantom-drain credits don't cancel real spill.
  - `check_2d_gas_allowance` runs per-tx before BAL recording (rejected txs don't pollute the BAL).
- **EIP-7976 / EIP-7981** — calldata + access-list floor cost adjustments wired into intrinsic gas + sender refund.
- **Engine / Execution-API** — [execution-apis PR 786](https://github.com/ethereum/execution-apis/pull/786) FCU semantics: no-reorg restricted to finalized prefix (point 2), `-38006 TooDeepReorg` returned when the requested reorg depth exceeds ethrex's state-history retention (point 6).
- **ef-tests** — runner continues processing blocks after expected-exception (fork-transition tests); 74 bal-devnet-6 Amsterdam fixtures intentionally skipped (see `docs/known_issues.md`).

**bal-devnet-7 direction (intentional)**

The bal-devnet-6 spec acknowledges as a known-bug that `set_delegation` for an existing authority should subtract the auth refund from `tx_state_gas` (mirroring SELFDESTRUCT) but currently does not. Upstream has a regression test (`9b3961a65 test_snobal_block_gas_used_inflated_by_7702_auth_refund`) that locks in the un-subtracted behavior for bal-devnet-6 verification. This PR carries the bal-devnet-7-prep subtraction so we are already aligned with where EELS is moving; the cost is that 74 snobal-devnet-6 fixtures expect the old un-subtracted accounting and are skipped via `tooling/ef_tests/blockchain/tests/all.rs::SKIPPED_BASE` until fixtures bump to snobal-devnet-7.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. *(N/A — no Store schema changes)*


crates/vm/levm/src/utils.rs
```diff
@@ -337,23 +337,24 @@ impl<'a> VM<'a> {
...
                         .state_gas_reservoir
-                        .checked_add(state_refund)
-                        .ok_or(InternalError::Overflow)?;
-                    // Track as intrinsic state gas adjustment (matches EELS intrinsic_state_gas -= refund).
-                    // Do NOT reduce state_gas_used here — that would inflate regular_gas in block accounting.
-                    self.intrinsic_state_gas_refund = self
-                        .intrinsic_state_gas_refund
-                        .checked_add(state_refund)
+                        .checked_add(refund)
                         .ok_or(InternalError::Overflow)?;
```
iovoid, line 358: Shouldn't only state_gas_reservoir change acording to [the spec](https://github.com/ethereum/execution-specs/blob/e87390b69ca5113f8f99db81dae7e4d6a8420342/src/ethereum/forks/amsterdam/vm/eoa_delegation.py#L199-L204)?

edg-l: Yes, fixed in `aa8cb9bac` on this branch to match the bal-devnet-6 spec exactly (only `state_gas_reservoir += refund`, no mutation of `intrinsic_state_gas` / `state_gas_used`). bal-devnet-7 (`90324443d`) layers an additional `state_refund += refund` channel on top per EELS PR #2816 for block-level accounting, but that is out of scope for bal-6 where the spec keeps block-level `state_gas_used` "inflated" and credits the sender via the reservoir only. snobal-devnet-6 EIP-7702 fixtures pass with the current code.

crates/blockchain/fork_choice.rs
```diff
@@ -11,6 +11,22 @@ use crate::{
...
+///
+/// The spec (execution-apis PR 786, "engine: Restrict no-reorg to the prefix of known
+/// finalized") only forbids reorging past the finalized prefix. The finalized check is
+/// applied first; this cap is a secondary guard for the implementation limit.
+///
+/// Reference values across ELs (devnet branches, 2026-04-30):
+/// - besu (main): 90_000 — effectively unlimited
+/// - erigon (glamsterdam-devnet-0): 96, env-configurable via `MAX_REORG_DEPTH`
+/// - geth / nethermind / reth: no engine-API rejection; trust the CL's fork choice
+pub const REORG_DEPTH_LIMIT: u64 = 128;
```
ElFantasma, line 28: `REORG_DEPTH_LIMIT: u64 = 128` is hardcoded as a `const`. The doc comment above (lines 24-27) already cites the cross-EL variance (Besu unlimited, Erigon env-configurable at 96, geth/nethermind/reth no rejection). Since this number is load-bearing for `engine_forkchoiceUpdated -> -38006`, two suggestions worth considering:

1. Make it env-configurable (Erigon-style — `MAX_REORG_DEPTH`). At 128 blocks * 12s ≈ 25.6 min on mainnet, a network event that causes a longer reorg would be unrecoverable through CL-driven means. Operators may want to set this higher temporarily.
2. Couple it explicitly to the state-history retention setting (which the comment already says it's a proxy for). Right now the two could drift if retention is tuned without updating this constant.

Not blocking for the devnet-6 PR; the `128` value is reasonable as a default.

crates/vm/levm/src/hooks/default_hook.rs
```diff
@@ -148,11 +150,11 @@ impl Hook for DefaultHook {
...
-            // Block accounting: gas_used = intrinsic_regular + intrinsic_state
-            // state_gas_used already = intrinsic_state (no execution state gas)
-            let state_gas = vm
-                .state_gas_used
-                .saturating_sub(vm.intrinsic_state_gas_refund);
+            // Block accounting: gas_used = intrinsic_regular + intrinsic_state.
+            // state_gas_used already = intrinsic_state (no execution state gas).
+            // Per EELS, `tx_env.intrinsic_state_gas` is immutable — any auth refund
+            // goes to the reservoir, not to block-accounted state_gas.
+            let state_gas = vm.state_gas_used;
```
ElFantasma, line 157: Confirms the bal-devnet-7-prep walkback in `aa8cb9bac`: no longer subtracting `intrinsic_state_gas_refund` (the field is gone; replaced by `intrinsic_state_gas_charged` which serves a different purpose).

One thing worth double-checking against EELS `devnets/bal/6/fork.py`: on a CREATE-tx collision that ALSO had a set_delegation auth refund, `state_gas_used` here reflects the gross intrinsic state cost (auth refund went to reservoir instead). The line below (`gas_used = effective_regular + state_gas`) therefore charges the user for the un-refunded auth bytes via state_gas. EELS docstring "intrinsic_state_gas is immutable" supports this — just noting that collision + delegation interaction is a path worth confirming against the snobal-devnet-6 EIP-7702 fixtures referenced in `aa8cb9bac`.

crates/vm/levm/src/hooks/default_hook.rs
```diff
@@ -238,18 +242,35 @@ pub fn refund_sender(
...
+        // PR #2689: state-gas charges that were halted (top-level or sub-frame) get
+        // reclassified to regular_gas_used via `regular_gas_reclassified`. The base
+        // formula subtracts every spill (treats them all as state-gas); the
+        // reclassification term adds back the halted portion so it counts toward the
+        // regular dimension.
+        let regular_gas = raw_consumed
+            .saturating_sub(vm.intrinsic_state_gas_charged)
+            .saturating_sub(vm.state_gas_reservoir_initial)
+            .saturating_sub(vm.state_gas_spill)
+            .saturating_add(vm.regular_gas_reclassified);
```
ElFantasma, line 273: The reclassification formula is `raw_consumed - intrinsic_state - reservoir_initial - state_gas_spill + regular_gas_reclassified`. Three saturating-subs in a row mean any invariant breach (e.g. `intrinsic_state_gas_charged > raw_consumed`, which should be impossible but is unenforced) silently produces `0` and then re-adds `regular_gas_reclassified`. The user-visible symptom would be "sender refund larger than it should be" — a value bug, not a panic.

For a path this hot, two cheap improvements: (a) `debug_assert!(raw_consumed >= vm.intrinsic_state_gas_charged + vm.state_gas_reservoir_initial + vm.state_gas_spill)` once at the top of this block to fail loud in tests; (b) consider `checked_sub` and propagate an `InternalError` on underflow (matches the `checked_add` at line 276). Same suggestion would apply to the collision branch.

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -1133,32 +1165,94 @@ impl<'a> VM<'a> {
...
-                    .checked_add(child_state_gas_used)
-                    .ok_or(InternalError::Overflow)?;
+            TxResult::Revert(err) => {
+                let outstanding_delta = self
+                    .state_gas_spill_outstanding
+                    .saturating_sub(state_gas_spill_outstanding_snapshot);
+                let credit_against_drain_delta = self
+                    .state_gas_credit_against_drain
+                    .saturating_sub(state_gas_credit_against_drain_snapshot);
+                debug_assert!(
```
ElFantasma, line 1184: Same observation as on `default_hook.rs:273` — `debug_assert!` only catches in debug. The two `saturating_sub`s right above (`outstanding_delta`, `credit_against_drain_delta`) silently produce `0` in release if the invariant is violated, and the downstream `regular_gas_reclassified` math compounds the masked value (since `local_excess = outstanding_delta - credit_against_drain_delta` would be incorrect).

The invariant doc says it's structurally guaranteed ("reservoir revert invariant violated"), so this is a defense-in-depth concern, not a known bug. But given how subtle the 2D-gas reclassification math is, a release-mode panic (`expect(...)` + `checked_sub`) feels worth the few ns to avoid silently incorrect gas accounting. Not blocking.

ElFantasma (commented):
Four findings inline — all minor. None blocking.

**Body finding (not inline)**: the 74-fixture skip list in `tools/ef_tests/blockchain/tests/all.rs` + `docs/known_issues.md` is the right contract, but worth wiring a tracking issue so the skip list shrinks as snobal-devnet-7 fixtures land — easy to forget once devnet attention moves on.

edg-l:
> Four findings inline — all minor. None blocking.
> 
> **Body finding (not inline)**: the 74-fixture skip list in `tools/ef_tests/blockchain/tests/all.rs` + `docs/known_issues.md` is the right contract, but worth wiring a tracking issue so the skip list shrinks as snobal-devnet-7 fixtures land — easy to forget once devnet attention moves on.

The skip list is alredy gone in the bal-devnet-7 branch

crates/vm/levm/src/hooks/default_hook.rs
edg-l: This reclassification chain is removed on bal-devnet-7, via commits `c82b3bd44` (Policy A: halt refunds spilled state-gas via the reservoir in `finalize_execution`) and `8d8cc1d8f` (specs #2827/#2828), which collapse the formula to `raw_consumed - intrinsic_state - reservoir_initial - state_gas_spill`, dropping both `regular_gas_reclassified` and one of the saturating subs. Not worth hardening on the bal-6 branch.

crates/vm/levm/src/opcode_handlers/system.rs
edg-l: The `outstanding_delta` / `credit_against_drain_delta` chain and this `debug_assert!` are removed in bal-devnet-7 via `f4e47e163` ("cross-frame revert leaks inline credits"). The revert path is rewritten per EELS PR #2823 to track `child_net_used = used_delta - absorbed_delta`, and the snapshot fields the invariant was guarding are gone.

crates/blockchain/fork_choice.rs
edg-l: Agreed. bal-devnet-7 does not touch this either (`git diff bal-devnet-6-pr..origin/bal-devnet-7 -- crates/blockchain/fork_choice.rs` is empty), so making `REORG_DEPTH_LIMIT` env-configurable and coupling it to state-history retention is a clean follow-up PR.

crates/vm/levm/src/hooks/default_hook.rs
edg-l: Confirmed: `aa8cb9bac` keeps bal-6 spec-exact, so any set_delegation auth refund goes only to `state_gas_reservoir`. Block-level `state_gas_used` stays inflated by the un-refunded auth bytes and is charged via `gas_used = effective_regular + state_gas`. The snobal-devnet-6 EIP-7702 fixtures pass with that accounting. The block-level subtraction lands in bal-devnet-7 via the dedicated `state_refund` channel (`90324443d`, EELS PR #2816).

ElFantasma (approved):

iovoid (approved):

azteca1998 (approved):

ilitteri (approved):

Issue #1615, opened 2025-01-03
L1: get pooled transactions in batch


Arkenan:
Our current implementation of GetPooledTransactions iterates over the transaction hash list and gets transactions one by one. There may be a better more batch-oriented way to get them.

Arkenan:
This was relevant when the mempool was in the db. Both the transactions and the blob bundles are now in memory hashmaps and batching is not a thing.

PR #6576, opened 2026-05-06
feat(l1): add --mempool.private flag for non-propagating local txs

ilitteri:
**Motivation**

Mirror reth's [`--txpool.no-local-transactions-propagation`](https://github.com/paradigmxyz/reth/blob/12ec7c57281c611355a02a2032ea7330ed9d3c68/crates/node/core/src/args/txpool.rs#L375): a flag to keep RPC-submitted transactions private to this node so they're only included in blocks built locally — no peer gossip. Use cases:
- Operators running a private validator wanting MEV-protected ordering
- Devnet A/B tests where one node should never propagate
- Searcher / sequencer setups separating local vs public mempools

**Description**

New flag `--mempool.private` (env `ETHREX_MEMPOOL_PRIVATE`, off by default), placed under `Node options` next to the existing `--mempool.maxsize`. When enabled, transactions submitted via this node's `eth_sendRawTransaction` enter the mempool and may be included in locally-built blocks but are NOT inserted into the broadcast pool. P2P-received transactions are unaffected.

Plumbing follows the no-feature-gated-args pattern (separate methods + shared private helper):

- `BlockchainOptions::private_mempool: bool` (default `false`).
- `Mempool` gets a sibling `add_transaction_no_broadcast` next to `add_transaction`; both share a private `add_transaction_inner(broadcast: bool)` so the broadcast-pool insertion is the only behavioral difference.
- `Blockchain::add_local_transaction_to_pool` and `add_local_blob_transaction_to_pool` are the **RPC entry points** — they consult `options.private_mempool` and route to the no-broadcast variant when set. The existing `add_transaction_to_pool` (used by the P2P / sync paths) keeps current behavior and always broadcasts.
- `eth_sendRawTransaction` handler now calls the local variants.

**P2P-gossip race + perf hardening**

The new-peer pooled-hash dump (`send_all_pooled_tx_hashes`) and the local-submission early-return both have to handle the case where a tx is already in the public pool via gossip:

- `Mempool::get_txs_for_new_peer_dump` returns the broadcast-eligible snapshot under a single read lock — replaces the per-tx `is_private` lock the previous loop took, which under a hot pool would serialize peer handshakes against any mempool write.
- `add_transaction_to_pool_inner` (+ EIP-4844 path) logs a `warn!` when `contains_tx` short-circuits with `broadcast=false`. If the tx has already been seen via gossip, the local `--mempool.private` request cannot retroactively un-broadcast it; the log surfaces this to the operator. State is intentionally unchanged — mutating the existing entry would itself violate the original submitter's intent.

**Checklist**


crates/blockchain/blockchain.rs
```diff
@@ -2302,12 +2310,44 @@ impl Blockchain {
...
     pub async fn add_blob_transaction_to_pool(
         &self,
         transaction: EIP4844Transaction,
         blobs_bundle: BlobsBundle,
+    ) -> Result<H256, MempoolError> {
+        self.add_blob_transaction_to_pool_inner(transaction, blobs_bundle, true)
+            .await
+    }
+
+    /// EIP-equivalent of `add_blob_transaction_to_pool` for transactions
```
iovoid: Equivalent to which EIP?

iovoid (changes requested):
We leak the private transactions through send_all_pooled_tx_hashes, which is used to send a list of pooled transactions to new peers. We also send them if asked by txhash.

ilitteri:
Pushed `2ca83911a` addressing both review comments:

**1. P2P leak paths (@iovoid CHANGES_REQUESTED)** — fixed both leaks you identified:

- New `Mempool::private_pool` (`FxHashSet<H256>`) populated by `add_transaction_no_broadcast`, cleared on tx removal alongside `broadcast_pool`.
- New `Mempool::is_private(hash)` for P2P call sites to consult.
- `send_all_pooled_tx_hashes` (in `p2p/rlpx/connection/server.rs`) now filters out private hashes before sending pooled-hash dumps to a newly-connected peer.
- `Blockchain::get_p2p_transaction_by_hash` (called from `GetPooledTransactions::handle`) returns the same `not found`-shaped `StoreError` for private hashes — the eth/68 spec explicitly allows skipping unavailable txs, so peers handle this cleanly.

New test `add_transaction_no_broadcast_marks_tx_as_private_for_p2p_filters` exercises the flag round-trip (add → `is_private` → remove → `!is_private`) and the unknown-hash path.

**2. "Equivalent to which EIP?" inline (`crates/blockchain/blockchain.rs:2329`)** — fixed: renamed "EIP-equivalent of …" → "Local-RPC counterpart of …". My bad on the wording.

Ready for re-review.

iovoid (approved):

crates/blockchain/blockchain.rs
ilitteri: 63e702bc5

crates/blockchain/blockchain.rs
```diff
@@ -2332,22 +2374,56 @@ impl Blockchain {
...
+        broadcast: bool,
     ) -> Result<H256, MempoolError> {
         // Blob transactions should be submitted via add_blob_transaction along with the corresponding blobs bundle
         if matches!(transaction, Transaction::EIP4844Transaction(_)) {
             return Err(MempoolError::BlobTxNoBlobsBundle);
         }
         let hash = transaction.hash();
         if self.mempool.contains_tx(hash)? {
+            if !broadcast {
+                warn!(%hash, "tx already public; --mempool.private cannot retroactively un-broadcast");
```
ElFantasma, line 3147: The `warn!` correctly surfaces "this private submission lost the race to a public gossip" — but a log line is easy to miss for operators running with INFO+ and impossible to alert on. Consider adding a counter (`METRICS_MEMPOOL.inc_private_collisions()` or similar) so the rate of these races is visible in Grafana. Two reasons it matters:

- An operator using `--mempool.private` for MEV-protected ordering needs to know how often their privacy was already breached by peer gossip before their RPC submission arrived. The warn gives one-time anecdotal data, a metric gives the rate.
- The same warn-path exists in the blob path at `:2357`, so two log lines for the same condition — a single counter would collapse them.

Not blocking — happy to defer to a follow-up if you'd rather keep the PR scoped to the flag itself.

crates/blockchain/blockchain.rs
```diff
@@ -2531,6 +2612,15 @@ impl Blockchain {
     }
 
     pub fn get_p2p_transaction_by_hash(&self, hash: &H256) -> Result<P2PTransaction, StoreError> {
+        // --mempool.private: never serve private txs over P2P, even if a peer
+        // somehow learned the hash. The spec for `GetPooledTransactions`
+        // explicitly allows skipping unavailable transactions, so we mirror
+        // the "not found" path the caller already handles.
+        if self.mempool.is_private(*hash)? {
```
ElFantasma, line 3942: Consistency observation with the `get_txs_for_new_peer_dump` optimization called out in the PR description ("replaces the per-tx `is_private` lock the previous loop took"): the same concern still applies to `GetPooledTransactions::handle` (`networking/p2p/rlpx/eth/transactions.rs:228-237`), which calls `get_p2p_transaction_by_hash` in a loop. Each call takes two read locks here (one for `is_private`, one for `get_transaction_by_hash`), and a `GetPooledTransactions` request can carry up to 256 hashes per the eth spec.

A batched `get_p2p_transactions_by_hashes(&[H256]) -> Vec<P2PTransaction>` that takes a single read lock and applies the `is_private` filter inline would mirror the dump-path fix and remove a similar peer-handshake-vs-mempool-write contention. Same direction as the existing TODO at `transactions.rs:222` ("TODO(#1615): get transactions in batch instead of iterating over them") — this would dovetail with that.

Not blocking for this PR — the flag itself is correctly implemented and tested. Flagging because the description's lock-contention point applies broader than the one path.

ElFantasma (approved):
Two small observability concerns. Both follow-ups, not blocking.

azteca1998 (approved):

crates/blockchain/blockchain.rs
ElFantasma: Merged `main` in 39c9b719b; leaving this one as follow-up work rather than doing it here, with the reasoning written down so it's actionable.

There is no mempool metrics module today — `ethrex-metrics` exposes `blocks::METRICS_BLOCKS` and `transactions::METRICS_TX`, but nothing for the pool. Adding the counter therefore means a new module in `ethrex-metrics`, registering the metric, and threading it through the `metrics` feature gate in `ethrex-blockchain` — a fair amount of surface in a crate this PR doesn't otherwise touch, for a +298 PR whose job is the flag itself.

The point stands though: a `warn!` is not alertable, and an operator running `--mempool.private` for MEV-protected ordering genuinely needs the *rate* at which their submissions lose the race to peer gossip, not one-off log lines. Concretely what's wanted is a counter incremented at both `blockchain.rs:2902` and `:3017` (the two sites that log the same condition today), so the two collapse into one series.

Happy to file that as a tracked follow-up if you'd like it captured rather than left in a resolved thread.

crates/blockchain/blockchain.rs
ElFantasma: Also deferring this one — it belongs in its own PR.

Batching `get_p2p_transaction_by_hash` into `get_p2p_transactions_by_hashes(&[H256])` touches `crates/networking/p2p/rlpx/eth/transactions.rs` rather than the mempool flag, changes the shape of the `GetPooledTransactions` response path, and overlaps directly with the existing `TODO(#1615)` ("get transactions in batch instead of iterating over them"). Doing it here would mean this PR carries a p2p refactor with its own review surface, and it would collide with whoever picks up #1615.

Worth noting the correctness side is already covered: `get_p2p_transaction_by_hash` does consult `is_private`, so private txs are not served over `GetPooledTransactions` — this is purely about the per-tx double read-lock under a request that can carry up to 256 hashes, not a leak.

crates/blockchain/blockchain.rs
ElFantasma: None — this isn't an EIP behavior, it's node-local policy. The flag only decides whether RPC-submitted transactions are gossiped; it doesn't change validity, execution, or anything on the wire protocol. Other clients have the same kind of knob outside the spec (geth's `--txpool.nolocals` family).

The help text was reworded to say so explicitly: *"Node-level config (not a protocol/EIP behavior): keep RPC-submitted transactions private..."*

ElFantasma:
### Merged `main` (39c9b719b)

Branch was 213 commits behind; 13 conflicts across 6 files. Threads answered inline.

**The substantive merge.** This branch had split `Mempool::add_transaction` into `add_transaction` / `add_transaction_no_broadcast` / `add_transaction_inner(broadcast)`, while `main` changed `add_transaction`'s signature to take `frame_reservation` + `sender_admission` and return `MempoolError` (the atomic per-sender gating from #6938). Threaded `main`'s parameters through all three, so the private path gets **identical admission gating** and differs only in propagation, then updated both `blockchain.rs` call sites keeping `main`'s blob-bundle rollback on insert failure.

One merge decision worth calling out: `main`'s `send_all_pooled_tx_hashes` uses `get_all_txs_by_sender`, which has **no private-tx filter**. Keeping this branch's `get_txs_for_new_peer_dump` is therefore load-bearing for correctness, not just the lock optimisation the description mentions — taking `main`'s version would leak `--mempool.private` transactions to every newly-connected peer through the pooled-hash dump and defeat the flag entirely.

Also updated this branch's two tests for API drift over those 213 commits: `Transaction::hash()` now takes a crypto argument, and the add-transaction entry points take five.

**Verified:** `cargo fmt --check` clean, `clippy --workspace --all-targets -D warnings` clean, 70/70 mempool tests including both private-mempool tests.

**Deferred (replied inline, not fixed):** the private-collision metric and the batched `get_p2p_transactions_by_hashes`. Both are follow-up work — the first needs a new mempool module in `ethrex-metrics` that doesn't exist yet, the second is a p2p refactor overlapping `TODO(#1615)`. Neither is a correctness gap; `get_p2p_transaction_by_hash` already consults `is_private`, so private txs are not served over `GetPooledTransactions`.

ElFantasma:
### Correction: the Cancun reorg failure was not flaky — it was an ordering regression in this PR (d230bad1c)

I previously wrote this failure off as flaky after it passed a local run and a CI re-run. That was wrong, and the reasoning was bad: I never found a mechanism, and I treated "no mechanism found" as "no mechanism exists". A third failure forced a closer look and the cause was real.

**What broke.** This PR replaced the new-peer dump's `get_all_txs_by_sender` with `get_txs_for_new_peer_dump`. The old function ended with:

```rust
txs_by_sender.iter_mut().for_each(|(_, txs)| txs.sort());
```

`MempoolTransaction`'s `Ord` sorts by nonce, so each sender's transactions were announced in nonce order. The replacement iterates `transaction_pool` — a hash map — and did not sort, so a sender's transactions were announced in **arbitrary order**.

**Why that breaks syncing peers.** A peer fetching those announcements out of nonce order sees gaps. Gapped transactions are future/queued on the receiving side, so `GapAdmissionDeniedUnderPressure` (#6609) and the per-sender queued cap (#6603) can turn them away — they never enter the peer's pool. The `Invalid Missing Ancestor Syncing ReOrg` tests spin up a second syncing client, which is exactly the path that hits this.

**Why it looked like flakiness.** Hash-map iteration order varies per process, so a *different* member of that test family failed on each run, and sometimes none did. Deterministic given an ordering, random-looking across runs.

**Fix.** Restore sender-grouped, nonce-ascending order, with a comment recording that the sort is load-bearing rather than cosmetic. Added `new_peer_dump_is_grouped_by_sender_and_nonce_ordered`, verified by A/B that it fails with the sort removed and passes with it restored — a disappearing `sort()` is invisible to both the compiler and every existing test.

**Verification.** Local hive, whole affected family rather than one variant: `engine-cancun/Invalid Missing Ancestor Syncing ReOrg` → **25 tests, failed=0**, including the two that failed in CI ("Transaction Nonce" and "Incomplete Transactions"). Plus 901/901 unit tests, clippy and fmt clean.

edg-l (approved):

PR #6581, opened 2026-05-06
perf(l1): cross-block caches

iovoid:
**Motivation**

Currently we cache values within a single block: within each evm, and between the main executor and the warmer. This means that frequently read accounts will be fetched every time.

**Description**

Adds an optimistic cache, that in case of a reorg simply invalidates the cache.


crates/vm/levm/src/db/mod.rs
```diff
@@ -1,18 +1,27 @@
...
 
-// Type aliases for cache storage maps
-type AccountCache = FxHashMap<Address, AccountState>;
-type StorageCache = FxHashMap<(Address, H256), U256>;
-type CodeCache = FxHashMap<H256, Code>;
+// Bounded so the cache survives cross-block reuse without unbounded growth.
+#[allow(clippy::unwrap_used)]
+const ACCOUNT_CACHE_CAP: NonZeroUsize = NonZeroUsize::new(1024 * 1024).unwrap();
+#[allow(clippy::unwrap_used)]
+const STORAGE_CACHE_CAP: NonZeroUsize = NonZeroUsize::new(4 * 1024 * 1024).unwrap();
```
ElFantasma: Quick sanity-check on memory at full saturation:

- `STORAGE_CACHE_CAP = 4M` × `((Address, H256), U256)` ≈ 84 B value + ~30 B `LruCache` overhead ≈ ~460 MB.
- `ACCOUNT_CACHE_CAP = 1M` × ~150 B (state + LRU node) ≈ ~150 MB.
- `CODE_CACHE_CAP = 64K` × variable (up to 24 KB per contract) ≈ up to ~1.5 GB if filled with max-sized bytecode (unlikely in practice but worth bounding by bytes, not entries).

Not necessarily wrong, but the constants currently read like "large round numbers." Was this sized against a measured working-set (RPC + sync + execution), or chosen as an upper bound? A one-line comment with the rationale would help future readers retune. Bonus: bounding `code` by total bytes rather than entry count would prevent a pathological worst case.

crates/vm/levm/src/db/mod.rs
```diff
@@ -41,175 +50,309 @@ pub trait Database: Send + Sync {
...
+            .write()
+            .unwrap_or_else(|p| p.into_inner());
+        let mut code = self.cache.code.write().unwrap_or_else(|p| p.into_inner());
+
+        // Wipe path is cold on Cancun+ (EIP-6780 prevents cross-tx storage
+        // wipes); kept for pre-Cancun replay and L2s on older forks.
+        let mut storage_wiped: FxHashSet<Address> = FxHashSet::default();
+
+        for update in account_updates {
+            accounts.pop(&update.address);
```
ElFantasma, line 374: `accounts.pop(&update.address)` evicts every touched account each block. Concretely, this means:

- Frequently-**written** accounts (system contracts, popular routers, validator deposit, beacon root storer, etc.) get evicted every block they're touched — i.e. the accounts where cross-block caching would matter most are exactly the ones that don't benefit.
- The cross-block cache primarily helps **read-only** accounts (queried by `eth_call`, RPC, builders) — those aren't in `account_updates`, so they survive.

This matches the PR description ("frequently read accounts will be fetched every time"), and the comment block at 298-302 acknowledges the trade-off — but it's worth confirming that's the intended optimization target, not a regression vs. a more aggressive cache that keeps hot writes around. One alternative: have the merkleizer pass back the post-block `storage_root` per touched address, so we can `put` an updated `AccountState` here instead of evicting. More wiring, but recovers the hot-write case if measurements show it matters.

crates/blockchain/blockchain.rs
```diff
@@ -211,6 +212,14 @@ pub struct Blockchain {
...
     /// 17 threads: 16 shard workers + 1 watcher/coordination.
     merkle_pool: rayon::ThreadPool,
+    /// Cross-block state cache. Invalidated on parent mismatch.
+    cross_block_cache: CrossBlockCache,
+    /// Serializes `add_block_pipeline_inner` so the cross-block cache lifecycle
+    /// (`set_inner` → execute → `promote_block`) is atomic per block. Required
+    /// for callers that share `Arc<Blockchain>` across tasks (L2 peer actors).
+    /// `StdMutex` (not `TokioMutex`) because the pipeline is sync and must not
+    /// be held across `.await`.
+    pipeline_lock: StdMutex<()>,
```
ElFantasma: Behavioral change worth flagging: `pipeline_lock` serializes ALL `add_block_pipeline_inner` calls across an `Arc<Blockchain>`. The doc comment notes L2 peer actors as the motivating case, but anywhere `Arc<Blockchain>` is shared and concurrent block adds were possible (e.g. multiple sync streams), this is now a global bottleneck.

For L1 this is fine — engine_newPayload is serialized by consensus anyway. For L2 with N peer actors pulling blocks concurrently, this changes throughput from "min(N, cores) parallel" to "strictly serial". Two questions worth answering before merge:

1. Is that the actual L2 calling pattern today (was concurrent block-add already a no-op in practice because of upstream serialization)?
2. Could the lock be scoped tighter — only around `is_valid_for_parent → invalidate → set_inner` and `promote_block` — instead of the entire `add_block_pipeline_inner`? Execution + merkleization could then run unlocked, with only the cache lifecycle serialized.

ilitteri (approved):

crates/vm/levm/src/db/mod.rs
iovoid: I went ahead and measured with real-world data (replay of mainnet blocks 25087279-25087684) and updated the caps.

crates/vm/levm/src/db/mod.rs
iovoid: It's a known limitation. I agree it would be better to keep them if possible.

azteca1998 (approved):

iovoid:
Outdated

PR #6588, opened 2026-05-08
perf(l1): remove sync from eip-8025 execution path

benbencik:
**Motivation**

Observed that execution of the Ethrex guest program in ziskVM uses `sync`. However, the execution inside the zkVM is strictly sequential, so `sync` provides no benefit and only adds runtime overhead.

**Description**

This change is gated behind the `eip-8025` feature. We add replacements for `once_cell::sync::OnceCell` and `std::sync::OnceLock`. These new structs copy implementation from `once_cell::unsync::OnceCell` and use `unsafe` to get around the `Sync` requirement. The intent here is to avoid making huge rewrites in other parts of the project or adding sequential fallbacks. These changes are only sound because the execution inside zkVM is guaranteed to be single-threaded.

_Note:_ this work extends PR #6560 

**Results**
Profiling in ZisK v0.16.1 on 50 mainnet blocks 24949787 - 24949836:
- Average cost (main `accc87dfe`): 57.5869B
- Average cost (this change): 57.3293B

The change (~0.5%) may seem small, but it was profiled on the execution of zkVM. These costs compound when we measure the computation required for proving.


ilitteri (approved):

crates/common/types/eip8025_cell.rs
```diff
@@ -0,0 +1,109 @@
+/// `OnceCell` replacement for zkVM guest gated on `eip-8025` feature.
+///
+/// `once_cell::sync::OnceCell` atomics are pure overhead in zkVM guest.
+/// This struct copies the methods from `once_cell::unsync::OnceCell` and uses unsafe
+/// to get around the Sync requirement.
+///
+/// This code is only sound because the guest is guaranteed to be single-threaded.
+pub struct OnceCell<T>(core::cell::UnsafeCell<Option<T>>);
+
+unsafe impl<T: Send> Sync for OnceCell<T> {}
```
azteca1998: Is this intentional? Usually I'd expect to see the following:
```suggestion
unsafe impl<T: Send> Send for OnceCell<T> {}
unsafe impl<T: Sync> Sync for OnceCell<T> {}
```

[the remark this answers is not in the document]
iovoid: Maybe there should be some kind of guard to prevent eip-8025 mode from running in a multi-threaded environment?

I don't think it's _currently_ an issue, and it depends on how the prover is built, but wouldn't hurt to be sure.

crates/common/trie/node.rs
```diff
@@ -2,7 +2,133 @@ mod branch;
...
+        let val = f()?;
+        let slot = unsafe { &mut *self.0.get() };
+        debug_assert!(slot.is_none());
+        Ok(slot.insert(val))
+    }
+
+    #[inline]
+    fn take(&mut self) -> Option<T> {
+        self.0.get_mut().take()
+    }
```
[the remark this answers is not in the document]
iovoid: OnceLock and OnceCell should probably be in the same file, away from the details of trie internals.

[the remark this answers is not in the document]
benbencik: That is exactly what I inteded to do. However there is dependancy problem. The crate[`common` depends on `trie`](https://github.com/lambdaclass/ethrex/blob/main/crates/common/Cargo.toml#L13), so moving the implementation to `common` and using it in `trie` would create a cycle. It is also possible to make a new shared crate but that seems like too much. Do you have any clearner approach in mind?

crates/common/types/eip8025_cell.rs
benbencik: From my understanding, it should not matter in this case. But fair, I like your suggestion better. I added rewrite but did not include `impl<T: Send> Send for OnceCell<T>` since [`UnsafeCell` already implements `Send`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html#impl-Send-for-UnsafeCell%3CT%3E) trait.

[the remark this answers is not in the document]
benbencik: I might try to add a multithreaded assert to `eip-8025`. If we just care about the issue above, we can make a stronger compile gate by also including the target architecture running on zkVM. So we would have something like:
`#[cfg(all(feature = "eip-8025", target_arch="riscv64"))]`

iovoid (approved):

crates/common/types/eip8025_cell.rs
```diff
@@ -0,0 +1,109 @@
+/// `OnceCell` replacement for zkVM guest gated on `eip-8025` feature.
+///
+/// `once_cell::sync::OnceCell` atomics are pure overhead in zkVM guest.
+/// This struct copies the methods from `once_cell::unsync::OnceCell` and uses unsafe
+/// to get around the Sync requirement.
+///
+/// This code is only sound because the guest is guaranteed to be single-threaded.
+pub struct OnceCell<T>(core::cell::UnsafeCell<Option<T>>);
```
ElFantasma, line 8: **The type name shadows `std`'s** `OnceCell` (and `OnceLock` in `trie/node.rs`). At every use site the import line is `use super::eip8025_cell::OnceCell;` — identical in shape to `use once_cell::sync::OnceCell;`, and a future contributor adding a new field of type `OnceCell<T>` will reach for the import that's already in scope without realizing this version is single-threaded-only and triggers UB if ever touched from another thread.

Rename to something that carries the unsafety into every use site, e.g. `UnsafeSingleThreadedCell` / `Eip8025OnceCell` / `GuestOnceCell`. The cost is a one-time s/OnceCell/.../ across the cfg-gated paths; the win is that every future reader sees `pub hash: GuestOnceCell<BlockHash>` and immediately knows the surrounding code is in single-threaded territory.

crates/common/types/eip8025_cell.rs
```diff
@@ -0,0 +1,109 @@
...
+
+unsafe impl<T: Sync> Sync for OnceCell<T> {}
```
ElFantasma, line 10: The soundness contract ("only sound because the guest is guaranteed to be single-threaded") lives only in the doc comment. There's nothing that fails compilation or panics at startup if someone enables `eip-8025` in a multi-threaded host build — e.g. a debug build that stacks features, or a future host-side tool that picks up the feature transitively. The result would be a clean compile and silent UB.

Light defensive options:
- A `compile_error!` in this module under a config that's known multi-threaded (e.g. when `not(target_os = "zkvm")` or whatever zkVM marker is available), so the build fails loud instead of compiling unsoundly.
- At minimum, a `// SAFETY:` block above the `unsafe impl` listing the invariant explicitly ("caller asserts that no two threads ever hold an `&OnceCell` concurrently; only enable `eip-8025` in single-threaded guest builds"), so the contract is right next to the impl, not just in the module doc.

Non-blocking; the doc comment isn't *nothing*. But once the type compiles in a multi-threaded build, the only thing preventing UB is operator discipline.

crates/common/types/eip8025_cell.rs
```diff
@@ -0,0 +1,109 @@
...
+            None => d.field(&format_args!("<uninit>")),
+        };
+        d.finish()
+    }
+}
+
+impl<T> From<T> for OnceCell<T> {
+    #[inline]
+    fn from(value: T) -> Self {
+        OnceCell {
```
ElFantasma: nit: `OnceCell { 0: core::cell::UnsafeCell::new(Some(value)) }` reads as struct-literal syntax for a tuple struct, which is unusual. Prefer the tuple-struct constructor: `OnceCell(core::cell::UnsafeCell::new(Some(value)))`. Equivalent generated code; one fewer thing for clippy to flag (`init_numbered_fields`).

crates/common/types/eip8025_cell.rs
benbencik: Thanks, good point. However, I do not think that is possible with the current implementation. Right now, we shadow the struct as follows:
```rust
#[cfg(feature = "eip-8025")]
use super::eip8025_cell::OnceCell;
#[cfg(not(feature = "eip-8025"))]
use once_cell::sync::OnceCell as OnceCell;
```
The reason for the current design is that I wanted to make minimal changes to the current codebase. 

I added the mentioned compile error for non-zkvm builds, which should prevent incorrect usage. But if you prefer, I can rewrite this properly and change the naming of the struct.

azteca1998 (approved):

ElFantasma (approved):
Compile_error gate handles the safety concern, and your minimal-changes argument is the right call — a rename would have churned every use site for no behavioral gain once the gate is in place. Approving.

Thanks for the careful back-and-forth on this one.

benbencik:
Just wanted to note that the CI fails due to the `compile_error` we added in the changes. Should I think about ways to change the compile error to pass CI?

ElFantasma:
Sorry about this — the compile_error gate came out of my earlier review on this PR, and I didn't think through the `--features stateless` case before suggesting it. That's on me.

The fix is small: tighten the cfg from `feature = "eip-8025"` to `all(feature = "eip-8025", target_arch = "riscv64")` on the local unsafe `OnceLock` / `OnceCell` variants (and the matching `use` imports), then drop the compile_error itself. That way the guest (riscv64) still gets the single-threaded perf win, while host builds — including `--features stateless` on aarch64 CI — fall back to `std::sync::OnceLock` / `once_cell::sync::OnceCell` as before. I verified locally that `cargo check -p ethrex-common` and `cargo check -p ef_tests-blockchain --features stateless` both pass with this change.

If it'd save you time, I can push the fix as a PR against your fork's branch — `maintainer_can_modify` is on, so it would land as one commit on `remove-sync-on-eip-8025` for you to review and merge. Happy to do that, or to just hand you the diff if you'd rather apply it yourself. Whichever you prefer.

benbencik:
Great, thanks for the suggestion, I should have checked that first. I added the changes, but there seems to be a problem with downloading the tests. Could you try to run the tests again?

PR #6590, opened 2026-05-08
fix(levm): replace misleading [LIE] comment in get_state_transitions_tx

braveocheretovych:
Remove the stale [LIE] comment that incorrectly claimed a database fallback exists for missing accounts in get_state_transitions_tx. Since load_account always inserts accounts into initial_accounts_state when skip_initial_tracking is false, the account is guaranteed to be present. Also update the error message to accurately reflect the actual source (initial accounts state instead of immutable cache)


ElFantasma (approved):

braveocheretovych:
Hey @ElFantasma could you approve and merge this

PR #6592, opened 2026-05-08
fix(levm): revert doesn't unmark the account as existing

iovoid:
**Motivation**

When an account creation reverts, we incorrectly mark it as existing for the purposes of EIP7702, leading to charging gas incorrectly.

**Description**

Currently we track accounts that exist in the trie for purposes of charging EIP7702 gas (the price is different if the authority already exists in the trie). This is saved in the `exists` field of the account, which is currently not saved in the callframe backup.

This means when an account creation is reverted, we don't unmark it as existing. This causes an incorrect amount of gas to be charged.

The PR contains a regression test, that should fail on the first commit and pass on the second.

### DIFF
--- a/crates/vm/levm/src/db/gen_db.rs
+++ b/crates/vm/levm/src/db/gen_db.rs
@@ -534,12 +534,13 @@ impl<'a> VM<'a> {
 
     */
     pub fn get_account_mut(&mut self, address: Address) -> Result<&mut LevmAccount, InternalError> {
-        let account = self.db.get_account_mut(address)?;
-
+        // Backup must be taken before mark_modified flips `exists` to true.
+        let account = self.db.get_account(address)?;
         self.current_call_frame
             .call_frame_backup
             .backup_account_info(address, account)?;
 
+        let account = self.db.get_account_mut(address)?;
         Ok(account)
     }
 
--- a/crates/vm/levm/src/hooks/default_hook.rs
+++ b/crates/vm/levm/src/hooks/default_hook.rs
@@ -362,11 +362,13 @@ pub fn delete_self_destruct_accounts(vm: &mut VM<'_>) -> Result<(), VMError> {
 
     // Delete the accounts
     for address in vm.substate.iter_selfdestruct() {
-        let account_to_remove = vm.db.get_account_mut(*address)?;
+        // Backup must be taken before mark_modified flips `exists` to true.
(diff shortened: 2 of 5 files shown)


ilitteri (approved):

edg-l:
Reviewed the patch and ran a merge-tree test against a long-lived branch (no conflicts). Fix is clean and the test exercises the exact path. Two follow-up observations:

**1. One more raw `db.get_account_mut` callsite that has the same antipattern**

Grepping the levm tree turns up a third caller that bypasses `VM::get_account_mut`:

```rust
// crates/vm/levm/src/hooks/l2_hook.rs:780
vm.db.get_account_mut(fee_token)?.storage = new_storage;
```

`transfer_fee_token` backs up storage slots via `vm.backup_storage_slot(...)` (lines 773-779) but never calls `backup_account_info(fee_token, ...)`. So `mark_modified` flips `exists=true` for the fee-token account with no pre-flip backup, and on revert `restore_cache_state` (now with the new line) would restore an `exists` that was already `true` — practically a no-op since fee-token contracts are deployed at genesis and always pre-exist, but it's the same code smell. Worth either:
- applying the same backup-before-flip pattern there, or
- adding a comment justifying why the fee-token path doesn't need it.

Either way it keeps the antipattern from proliferating to a future site where the account isn't guaranteed to pre-exist.

**2. Worth auditing the storage-slot side separately**

The `exists` bug exists because `mark_modified` is called inside `db.get_account_mut` before any backup. Storage slots have an analogous risk surface: if anyone reads/writes `db.get_account_mut(addr)?.storage` without calling `backup_storage_slot` first, a revert could leak the storage write. Out of scope for this PR (which is just `exists`), but a quick grep for that idiom in a follow-up would be cheap insurance.

Otherwise the fix looks correct:
- backup pre-flip + restore on revert is the right two-sided fix
- `LevmAccount` now has all 5 fields (info, storage, has_storage, status, exists) consistently captured and restored
- the regression test correctly fails without commit 2 and passes with it (verified by tracing the path through `restore_cache_state`)
- one extra `db.get_account` lookup per `VM::get_account_mut` call is O(1) on FxHashMap, no perf concern

azteca1998 (approved):

edg-l (approved):

avilagaston9 (approved):

PR #6593, opened 2026-05-09
refactor(l1): unify the declaration of crate dependencies using workspace-style, for easier management.

Pana:
**Motivation**
Unify the declaration of crate dependencies using workspace-style, for easier management. Avoid using relative file paths.

**Description**

**Checklist**

Closes #issue_number


Cargo.toml
```diff
@@ -132,6 +133,13 @@ tower-http = { version = "0.6.2", features = ["cors"] }
 indexmap = { version = "2.11.4" }
 k256 = "0.13.4"
 anyhow = "1.0.86"
+directories = "5.0.1"
+itertools = "0.14.0"
```
ElFantasma, line 137: `itertools = "0.14.0"` here, but two other crates still use it with a different inline version:

- `tooling/ef_tests/state/Cargo.toml:28` → `itertools = "0.13.0"`
- `crates/l2/sdk/Cargo.toml:20` → `itertools = "0.13.0"`

Not dangerous (Cargo will resolve both versions), but it defeats the "single source of truth" point of moving to workspace deps — and you'll still pay the compile-time cost of building both versions. Worth either migrating those two to workspace + `0.14.0` (the bigger refactor, but consistent), or pinning workspace to `0.13.0` to match what's already in use (smaller change). Same direction either way.

Cargo.toml
```diff
@@ -132,6 +133,13 @@ tower-http = { version = "0.6.2", features = ["cors"] }
...
+tracing-appender = "0.2"
+local-ip-address = "0.6"
+tikv-jemallocator = "0.6.0"
+pprof = "0.15"
+vergen-git2 = "9.1"
```
ElFantasma, line 142: `vergen-git2 = "9.1"` here — but two other crates still have it as inline at a *major* version behind:

- `crates/l2/Cargo.toml:60` → `vergen-git2 = { version = "1.0.7" }`
- `crates/l2/tee/quote-gen/Cargo.toml:8` → `vergen-git2 = { version = "1.0.7" }`

This is a 1.x → 9.x gap, not a patch difference. If the L2 crates intentionally stay on `1.0.7` (e.g. because newer versions broke something), the workspace dep declaration is misleading — a reader would think the workspace standardizes on `9.1`. If the older pin is just stale, this is the natural moment to upgrade both.

Either way: a one-line comment in the workspace entry explaining the split, or a follow-up that unifies them, would resolve the confusion.

Pana:
The migration will be completed in multiple PRs, done crate by crate.

Issue #6575, opened 2026-05-06
[Fuzzing] Add `statetest` CLI subcommand for goevmlab differential fuzzing


bshastry:
## TL;DR

Add a small CLI surface — `<bin> statetest --json <path>` and `<bin> statetest --json-outcome <path>` that runs a single EF state test through LEVM and writes EIP-3155 JSONL traces (or just a stateroot line) to stderr. This is the missing piece needed to plug ethrex into [goevmlab](https://github.com/protocol-security/goevmlab) alongside geth, nethermind, besu, erigon, nimbus, evmone, eels and revm. The goevmlab adapter side is already scaffolded and waiting on this. The goevmlab repo is currently a private fork, but this issue specs out the ethrex side of things independently.

## Problem

[goevmlab](https://github.com/protocol-security/goevmlab) is the standard differential-fuzzing harness for EL clients (geth, nethermind, besu, erigon, nimbus, evmone, eels, revm/reth). It runs the same EF state test against every configured EVM and diffs their EIP-3155 traces and post-state roots.

To plug ethrex into that fleet, goevmlab needs to invoke an `ethrex` (or `ethrex-statetest`) binary with a small, well-defined CLI surface. Today no ethrex binary exposes this:

- `tooling/ef_tests/state_v2` is a *bulk* runner over a fixtures directory; it writes `success_report.txt` / `failure_report.txt` and has no single-file mode and no per-op trace.
- `tooling/ef_tests/state` is the older bulk runner, same shape.
- `crates/vm/levm/runner` takes a custom `RunnerInput` JSON, not the EF state-test format, and has no per-op trace output.
- `ethrex_levm::tracing::LevmCallTracer` exists (LEVM, `crates/vm/levm`, is ethrex's primary EVM impl) but is not wired to an EIP-3155 JSONL writer behind a CLI.

The goevmlab side is already scaffolded on a fork (adapter `evms/ethrex.go`, `--ethrex` CLI flag) and is waiting on this CLI landing upstream.

## Required CLI

| Mode | Invocation | Output |
|---|---|---|
| Full trace | `<bin> statetest --json <path>` | EIP-3155 JSONL on **stderr**, one op per line, terminated by a single `{"stateRoot":"0x<64 hex>"}` line |
| Speed / outcome | `<bin> statetest --json-outcome <path>` | One line on stderr containing `"stateRoot":"0x<64 hex>"` |

[...]

PR #6595, opened 2026-05-11
feat(l1): opcodeTracer (EIP-3155 per-opcode tracer)

edg-l:
## Summary

Adds a per-opcode tracer (`opcodeTracer`) wired through `debug_traceTransaction` and `debug_traceBlockByNumber`. The wrapper and per-step shape match the conventional cross-client `structLogger` format (`{failed, gas, returnValue, structLogs}` with `op` as the mnemonic string, numeric `gas`/`gasCost`/`refund`) while keeping the EIP-3155 step content (`pc`, `op`, `gas`, `gasCost`, `stack`, `depth`, `memory`, `storage`, `refund`, `memSize`, `returnData`).

Default tracer remains `CallTracer` (Blockscout compat); `opcodeTracer` is opt-in per request.

## Usage

Select the tracer with `"tracer": "opcodeTracer"`:

```bash
curl -s http://localhost:8545 -H 'content-type: application/json' -d '{
  "jsonrpc":"2.0","id":1,
  "method":"debug_traceTransaction",
  "params":[
    "0x<txhash>",
    { "tracer": "opcodeTracer" }
  ]
}' | jq
```

Block-level:

```bash
curl -s http://localhost:8545 -H 'content-type: application/json' -d '{
  "jsonrpc":"2.0","id":1,
  "method":"debug_traceBlockByNumber",
  "params":[
    "latest",
    { "tracer": "opcodeTracer" }
  ]
}' | jq
```

`tracerConfig` flags (all optional):

| Flag | Default | Effect |
|---|---|---|
| `disableStack` | `false` | When true, `stack` is serialized as `null` |
| `enableMemory` | `false` | Include `memory` (array of 32-byte hex strings) per entry |
| `disableStorage` | `false` | Omit `storage` map on SLOAD/SSTORE |
| `enableReturnData` | `false` | Populate `returnData` with the buffer (otherwise emitted as `"0x"`) |
| `limit` | `0` (unlimited) | Stop collecting after N entries |

## Output shape

Per-step entry:

```json
{
  "pc": 0,
  "op": "PUSH1",
  "gas": 100000,
  "gasCost": 3,
  "depth": 1,
  "stack": [],
  "memSize": 0,
  "returnData": "0x",
  "refund": 0
}
```

Top-level wrapper:

```json
{
  "failed": false,
  "gas": 21000,
  "returnValue": "0x",
  "structLogs": [ /* ... */ ]
}
```

Matches what every existing `debug_traceTransaction` consumer expects (geth/besu/reth all converge on this shape). `stateRoot` is omitted — computing it mid-trace would be expensive and consumers rarely use it.

## What's covered

- `crates/common/tracing.rs`: `OpcodeStep`, `MemoryChunk`, `OpcodeTraceResult` with manual `Serialize` impls.
- `crates/vm/levm/src/opcode_tracer.rs`: `LevmOpcodeTracer` with `active: bool` gate, hoisted once per opcode in the dispatch loop. Storage capture is **per-step single entry** (no accumulation across SLOAD/SSTOREs).
- `crates/vm/levm/src/vm.rs`: dispatch-loop hook (pc captured pre-advance, stack reversed top->bottom for wire format), end-of-tx capture using post-refund `gas_spent`.
- `crates/vm/levm/src/opcode_handlers/system.rs`: explicit `gasCost = intrinsic + callGasTemp` for CALL/CALLCODE/DELEGATECALL/STATICCALL/CREATE/CREATE2.
- `crates/networking/rpc/tracing.rs`: `TracerType::OpcodeTracer` variant.
- Plumbing: `crates/blockchain/tracing.rs` (`trace_transaction_opcodes` / `trace_block_opcodes`), `crates/vm/tracing.rs`, `crates/vm/backends/levm/tracing.rs`.
- `crates/vm/levm/benches/opcode_tracer_disabled.rs`: Criterion microbench for the disabled hot path.

## Test plan

- [x] 26 wire-format unit tests pinning each EIP-3155 field rule (`ethrex-common`)
- [x] 2 LEVM unit tests: dispatch hook + single-entry SSTORE storage capture
- [x] 1 end-to-end smoke test exercising the full `trace_tx_opcodes` pipeline
- [x] Disabled-path Criterion microbench
- [x] `cargo fmt --all`, `cargo check --workspace --all-targets`, `make lint-l1`
- [ ] `make test`
- [ ] `make -C tooling/ef_tests/blockchain test`

## Known follow-ups (from internal perf review, not blocking)

- Enabled-path: stack `Vec<U256>` reallocated per opcode (reuse buffer).
- Enabled-path: memory bytes copied twice when `enableMemory` is set.
- Enabled-path: per-stack-entry hex encoding allocates twice (`hex::encode` + `format!`); could write into a `[u8; 66]` stack buffer.
- Bench: add an enabled-path variant and a realistic mixed-opcode workload.


crates/vm/levm/src/vm.rs
```diff
@@ -762,6 +850,102 @@ impl<'a> VM<'a> {
...
+
+        let stack_values = &self.current_call_frame.stack.values;
+        let key_u256 = *stack_values.get(offset)?;
+        let key = u256_to_h256(key_u256);
+
+        if opcode == SLOAD {
+            let value = match self.get_storage_value(addr, key) {
+                Ok(v) => H256::from(v.to_big_endian()),
+                // Account not yet cached — graceful fallback per R16.
+                Err(_) => H256::zero(),
```
ElFantasma: Silent fallback to `H256::zero()` on `AccountNotFound` is a debugging foot-gun: if a consumer is using this tracer to investigate a state-transition issue, an account that's not yet been hot-loaded would appear in the trace as reading `0x00...00` from any slot — which is also a legitimate value for a never-written slot.

Geth's tracer doesn't have this case because by the time SLOAD runs, the account is cached. In LEVM the comment cites "graceful fallback per R16" — is there a code path where SLOAD/SSTORE genuinely runs before `load_account`?

If the answer is "no, this can't happen in practice but we're being defensive": consider returning `None` and skipping the storage field entirely for that step, so the consumer sees the gap rather than a misleading zero. If it can happen: worth a more specific comment about which scenario triggers it.

crates/common/tracing.rs
```diff
@@ -125,3 +126,328 @@ pub struct PrePostState {
...
+        0xF1 => Cow::Borrowed("CALL"),
+        0xF2 => Cow::Borrowed("CALLCODE"),
+        0xF3 => Cow::Borrowed("RETURN"),
+        0xF4 => Cow::Borrowed("DELEGATECALL"),
+        0xF5 => Cow::Borrowed("CREATE2"),
+        0xFA => Cow::Borrowed("STATICCALL"),
+        0xFD => Cow::Borrowed("REVERT"),
+        0xFE => Cow::Borrowed("INVALID"),
+        0xFF => Cow::Borrowed("SELFDESTRUCT"),
+        b => Cow::Owned(format!("opcode 0x{:02x}", b)),
```
ElFantasma: EIP-3155 (https://eips.ethereum.org/EIPS/eip-3155) says of `opName`: *"Required, but **if the operation is unknown, this field MAY be missing/null**"*. Returning `"opcode 0x5f"` for an unknown byte is more permissive than the spec allows for tools that strictly parse — they'd see this as a real opcode name.

Geth's choice is to leave `opName: null` for unassigned bytes. Two reasonable paths: (a) match geth's null behavior and serialize this branch as `None` in `OpcodeStep::serialize`; (b) keep the current string and add a brief comment saying we intentionally diverge from geth's `null` here for debuggability. Either way, worth documenting which client we're matching.

crates/vm/levm/src/vm.rs
```diff
@@ -762,6 +850,102 @@ impl<'a> VM<'a> {
...
+            };
+            Some((addr, key, value))
```
ElFantasma: The returned tuple is `(addr, key, value)` but `addr` is always `self.current_call_frame.code_address` (line 913) — i.e. the current frame. Downstream (`opcode_tracer.rs:148`) destructures it as `_addr` and discards it. The storage map in the serialized output is keyed by slot only.

So `Option<(H256, H256)>` would carry the same information and avoid the dead field. Not blocking, just cleanup.

edg-l:
## Wrapper realignment + JUMPDEST fix

Tested live against geth and besu on a kurtosis localnet. The PR's previous wrapper was strict-EIP-3155 (`pass`/`gasUsed`/`output`/`steps`, hex-string `gas`/`gasCost`/`refund`, numeric `op` byte, separate `opName`), which no existing tooling can read — every other major client emits the de-facto `structLogger` shape.

### Changes

- Wrapper renamed to match the cross-client convention:
  - `pass` → `failed` (inverted)
  - `gasUsed` → `gas` (number)
  - `output` → `returnValue`
  - `steps` → `structLogs`
- Per-step:
  - `op` now emits the mnemonic string (e.g. `"PUSH1"`); unassigned bytes fall back to `"opcode 0xNN not defined"`.
  - `opName` dropped (redundant with the new `op`).
  - `gas`, `gasCost`, `refund` emitted as JSON numbers (not hex strings).
- Per-step content remains EIP-3155 (`pc`, `op`, `gas`, `gasCost`, `stack`, `depth`, `memory`, `storage`, `refund`, `memSize`, `returnData`). `memSize`, `returnData`, and `refund` are always emitted — additive beyond the minimal cross-client step shape, harmless to consumers that ignore extras.

### Bug found while testing

`jump()` in `stack_memory_storage_flow.rs` fuses the destination JUMPDEST into the JUMP/JUMPI handler as an optimization (advances PC past JUMPDEST and charges its 1 gas inline). With the previous tracer, this caused:

1. **Missing JUMPDEST trace entries** — the JUMPDEST handler was never dispatched, so the tracer never emitted a step for it. Geth/besu emit a JUMPDEST step.
2. **Inflated parent `gasCost`** — JUMPI's `gasCost` came out as 11 instead of 10 (the absorbed JUMPDEST charge).

For an ERC-20 transfer with 24 JUMPDESTs hit, step count was 329 vs geth's 353.

**Fix:** keep the fusion on the disabled-tracer hot path (zero perf change). When the tracer is active, override the parent JUM(I)'s gasCost with the correct value via `last_opcode_gas_cost`, and push a synthetic JUMPDEST entry via a new `synthesize_step` helper. Required swapping `last_step_captured: bool` for `last_step_index: Option<usize>` so the synthetic push doesn't shadow the parent's `finalize_step` target.

### Verification

Same tx (full ERC-20 transfer, 353 steps), kurtosis localnet:

| Comparison | Result |
|---|---|
| Wrapper keys match besu/geth | ✓ identical |
| Step count vs geth (353 vs 353) | ✓ exact match |
| Step count vs besu (353 vs 353) | ✓ exact match |
| Field-level diff `{pc,op,depth,gas,gasCost,stack}` vs geth | ✓ 0 differing lines / 353 |
| Field-level diff `{pc,op,depth,gas,gasCost,stack}` vs besu | ✓ 0 differing lines / 353 |
| `enableMemory` memory contents vs geth | ✓ identical chunks |
| `debug_traceBlockByNumber` block 0x6 (750 txs) vs geth | ✓ same tx order, hashes, gas, step counts |

Regression test added: `opcode_tracer_jumpdest_synthesized_after_jump` pins JUMPDEST emission with `gasCost = 1` and the parent JUMP at `gasCost = 8` (not 9).

azteca1998 (approved):

crates/common/tracing.rs
```diff
@@ -125,3 +125,347 @@ pub struct PrePostState {
...
+}
+
+// ─── Helpers ──────────────────────────────────────────────────────────────
+
+/// Returns the opcode mnemonic for `byte`.
+///
+/// Known opcodes → their uppercase name (`"PUSH1"`, `"ADD"`, `"INVALID"` for
+/// 0xFE). Unassigned bytes → `None`; callers wanting the conventional unknown
+/// string should fall back to `format!("opcode 0x{:02x} not defined", byte)`.
+pub fn opcode_name(byte: u8) -> Option<&'static str> {
```
iovoid, line 238: Should the table be fork-dependent?

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
```diff
@@ -433,14 +441,97 @@ fn jump(vm: &mut VM<'_>, target: usize) -> Result<(), VMError> {
...
+                            dst.copy_from_slice(c);
+                        }
+                        MemoryChunk(arr)
+                    })
+                    .collect(),
+            )
+        }
+    } else {
+        None
+    };
```
iovoid: Duplicated code

iovoid (commented):
The "output shape" in the description has opName while the code doesn't.

crates/common/tracing.rs
edg-l: Verified: the static table matches geth's `opCodeToString` (also flat 256-entry). Fork validity is enforced at *dispatch* via LEVM's per-fork opcode table — pre-Shanghai bytecode that hits 0x5F halts with InvalidOpcode before the tracer emits any step, so the name lookup never fires for invalid-for-this-fork bytes in practice. Added a doc comment near `opcode_name` spelling this out in 47dd368.

crates/vm/levm/src/opcode_handlers/stack_memory_storage_flow.rs
edg-l: Real duplication — `build_jumpdest_step` re-implemented the cfg-conditional stack/memory/return-data logic already in `LevmOpcodeTracer::pre_step_capture`. Extracted to a shared `opcode_tracer::build_step` free function in 47dd368; both call sites now hand off the raw inputs to the same constructor.

ElFantasma (approved):

crates/common/tracing.rs
```diff
@@ -125,3 +125,354 @@ pub struct PrePostState {
...
+        0x3B => Some("EXTCODESIZE"),
+        0x3C => Some("EXTCODECOPY"),
+        0x3D => Some("RETURNDATASIZE"),
+        0x3E => Some("RETURNDATACOPY"),
+        0x3F => Some("EXTCODEHASH"),
+        0x40 => Some("BLOCKHASH"),
+        0x41 => Some("COINBASE"),
+        0x42 => Some("TIMESTAMP"),
+        0x43 => Some("NUMBER"),
+        0x44 => Some("PREVRANDAO"),
```
iovoid, line 288: geth still calls this opcode 'DIFFICULTY' even in new forks, should we copy geth or keep PREVRANDAO?

https://github.com/ethereum/go-ethereum/blob/21c5a287f91640f383bc09ffe97d922999e171c7/core/vm/opcodes.go#L317

ElFantasma:
Re-reviewed the per-step wire shape against EIP-3155 and geth's actual `debug_traceTransaction` response. Pushed `dc11a20e1` with what I believe is a small course-correction; flagging the reasoning here in case I've missed something.

**Where the wrapper realignment landed** — `pass→failed`, `gasUsed→gas`, `output→returnValue`, `steps→structLogs` — that's correct and matches geth's structLogger output. No changes there.

**Where the per-step fields seem to have drifted from the stated target** — geth's actual structLogger MarshalJSON (from [`eth/tracers/logger/gen_structlog.go`](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/logger/gen_structlog.go)):

```go
type StructLog struct {
    Pc      uint64               `json:"pc"`
    Op      vm.OpCode            `json:"op"`         // type OpCode byte → JSON number
    Gas     math.HexOrDecimal64  `json:"gas"`        // → "0xN" hex string
    GasCost math.HexOrDecimal64  `json:"gasCost"`    // → "0xN" hex string
    ...
    OpName  string               `json:"opName"`     // mnemonic via s.OpName()
    ...
}
```

So geth's per-step shape is `op:<number>, opName:<string>, gas:"0xN", gasCost:"0xN"` — which is *also* exactly what EIP-3155 §"Required Fields" specifies. Strict-EIP-3155 and geth-structLogger-per-step are the same shape for these fields; the divergence between them is only at the wrapper level (which your wrapper realignment already handled).

The current code emits `op:"PUSH1"` (string mnemonic), no `opName`, `gas:100000` (decimal number). That's a third shape, different from both EIP-3155 and from what geth's RPC actually returns on the wire.

**About the "0 differing lines vs geth" verification** — I think the diff tool used was tolerant to type-shape differences (parsing into a permissive Go struct, then comparing semantic content). That would normalize away `op:"PUSH1"` vs `op:96` for clients that emit either, while reading wire-format-strict consumers (goevmlab's `gen_oplog.go` opLog UnmarshalJSON parses `op` as `*math.HexOrDecimal64` → cast to `vm.OpCode(uint64)`; string mnemonics don't survive that). If you have the exact tooling/script you used for the comparison, would be good to re-run with `dc11a20e1` to confirm I haven't broken something I don't understand.

**What `dc11a20e1` changes** (in `crates/common/tracing.rs` `OpcodeStep` Serialize impl):

1. `op` emitted as numeric byte (EIP-3155 §Required: "Number"; geth: `vm.OpCode` default-marshals as number).
2. `opName` added as always-emitted string companion (EIP-3155 §Optional; geth `gen_structlog.go` line `enc.OpName = s.OpName()` always sets it).
3. `gas`, `gasCost`, `refund` emitted as `"0xN"` hex strings (EIP-3155 §Required: "Hex-Number"; geth: `math.HexOrDecimal64` marshals as `"0xN"`).
4. `stack` emits `[]` when disabled instead of `null` (EIP-3155: "All array attributes MUST be initialized to empty arrays NOT to null").
5. Field order matches the spec's listed order (spec §"Clients SHOULD output the fields in the same order as listed in this EIP").

The 6 LEVM wire-format tests in `test/tests/levm/opcode_tracer_tests.rs` are updated to assert the new shape (all still passing). PR `feat/6575-statetest-cli` is stacked on this branch and will inherit the change for its goevmlab CLI surface, which was the original motivation.

If I've misread your verification setup or there's a reason structLogger-RPC consumers wouldn't accept the spec shape (Blockscout, Foundry, …), happy to revert and discuss — but my reading of geth source is that they all expect it.

PREVRANDAO vs DIFFICULTY: leaving as `PREVRANDAO` since EIP-4399 calls that the canonical post-Merge name. EIP-3155 doesn't constrain mnemonic strings, so this is correctness-neutral for spec consumers.

crates/vm/levm/src/opcode_tracer.rs
```diff
@@ -0,0 +1,265 @@
...
+    gas_cost: u64,
+    depth: u32,
+    refund: u64,
+    stack_view: &[U256],
+    memory_view: &[u8],
+    mem_size: u64,
+    return_data: &Bytes,
+    storage_kv: Option<(H256, H256)>,
+) -> OpcodeStep {
+    // Stack: Some(vec) when capture enabled; None when disabled (emits JSON null).
```
iovoid, line 241: Not true

ElFantasma:
**Retracting [my earlier comment](https://github.com/lambdaclass/ethrex/pull/6595#issuecomment-4463812701)** — I was wrong on the per-step claim. Verified live against geth and besu on a kurtosis localnet (using the new `tooling/trace_compare/compare.sh` helper): both clients emit exactly the structLogger shape your tracer originally produced — string `op` mnemonic, no `opName`, decimal `gas`/`gasCost`/`refund`. The `gen_structlog.go` source I quoted earlier doesn't tell the whole story (geth's `vm.OpCode` and `math.HexOrDecimal64` MarshalJSON impls do their own thing on the wire; the type annotations alone misled me). Your "0 differing lines" verification was empirically correct; my source-reading was the wrong analysis.

`dc11a20e1` (the EIP-3155 conformance commit on this branch) was therefore broken for the RPC use case. It's been replaced with `bfcb482ae`, which addresses the underlying design issue more cleanly:

**`OpcodeStep` and `OpcodeTraceResult` no longer have `Serialize` impls.** They're pure data. Wire formats live in dedicated newtypes:

```rust
// crates/common/tracing.rs

pub struct StructLoggerStep<'a>(pub &'a OpcodeStep);
pub struct StructLoggerResult<'a>(pub &'a OpcodeTraceResult);
pub struct Eip3155Step<'a>(pub &'a OpcodeStep);
```

`StructLoggerStep`/`StructLoggerResult` produce the geth-RPC shape (exactly what was in `dc11a20e1`'s parent). `Eip3155Step` produces the strict spec shape (used by the `statetest` CLI subcommand for goevmlab in #6575). Both serializers consume the same captured `OpcodeStep` data — one tracer, two output formats, no second EVM pass.

The RPC handler in `crates/networking/rpc/tracing.rs` now does `serde_json::to_value(StructLoggerResult(&result))`, so the wire output is byte-identical to what your original tracer emitted. The 6 wire-format tests in `test/tests/levm/opcode_tracer_tests.rs` are reverted to assert your original geth-RPC shape (string `op`, decimal gas, `stack:null` when disabled) and all pass.

PREVRANDAO vs DIFFICULTY: left as `PREVRANDAO` per [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399). Geth's lingering `DIFFICULTY` string is a known TODO in their tree and isn't load-bearing for RPC consumers.

Sorry for the noise. The architecture cleanup ended up worth doing even though it started from a wrong premise — adding a third format now (Parity-style flat call, anything else) is a new newtype with its own `Serialize`, no churn to the data layer.

ElFantasma:
## Summary of follow-up work

I picked up this PR mid-review and pushed a series of commits that pin down byte-level compatibility with geth's `debug_traceTransaction` output, plus add tooling for verifying it. End result: **the trace is byte-identical to geth's** on the test transactions we exercised against a local kurtosis multi-client network.

### What changed

**Architecture refactor** — `bfcb482ae`
- `OpcodeStep` / `OpcodeTraceResult` are now pure data types with no `Serialize` impl.
- Two wire-format newtypes wrap them:
  - `StructLoggerStep` / `StructLoggerResult` — geth-RPC `debug_traceTransaction` shape (string `op`, no `opName`, decimal `gas`/`gasCost`/`refund`, `stack: null` when disabled).
  - `Eip3155Step` — strict EIP-3155 spec shape (numeric `op` + `opName`, hex `gas`/`gasCost`/`refund`, `stack: []` when disabled).
- Module-level docs explain why both formats coexist and which consumers expect which.
- RPC handler wraps with `StructLoggerResult`; #6575's statetest CLI will wrap with `Eip3155Step` for goevmlab.

**Field-emission gating** — `2e86843ff`, `e1f38c26a`
- `StructLoggerEmit` opt-in flags for `memSize` / `returnData` / `refund` so the wire output matches geth's empirical behavior:
  - `memSize` / `returnData` follow the corresponding capture-config flags (off by default, on when `enableMemory` / `enableReturnData` is set).
  - `refund` is emitted whenever non-zero (matches geth's `omitempty`); the `emit.refund` flag now only force-emits zero values.

**Storage accumulation** — `3498f5853`
- `LevmOpcodeTracer` carries a cumulative `BTreeMap<H256, H256>` of every slot touched by SLOAD/SSTORE in the trace so far.
- Each SLOAD/SSTORE step embeds a snapshot, matching geth's structLogger behavior. (Per your message, the non-cumulative version was unintentional.)
- Regression test pinning that two consecutive SSTOREs produce step-2 `{slot1}`, step-5 `{slot1, slot2}`.

**Refund timing** — `29e8a1571` — closes #6672
- `finalize_step` now patches the buffered log's `refund` with the post-opcode counter value (`substate.refunded_gas` after the handler returns), not just `gas_cost`.
- This matches geth's hook timing: refund credited by SSTORE / pre-London SELFDESTRUCT appears on **that** opcode's step, not the next one.
- For opcodes that don't touch refund (every opcode except those two), post == pre so it's a no-op.
- Regression test prefills a slot with `0x42`, clears it, and asserts the SSTORE step itself carries the resulting refund credit.

**Tooling** — `c157f7a1d`, `7576dec67`, `0197fc7d2`, `92c00ff47`
- `tooling/trace_compare/compare.sh` — discovers EL services in a kurtosis enclave, picks a tx from `latest`, calls `debug_traceTransaction` on each, and prints suggested `diff` commands. Asks ethrex for `tracer: "opcodeTracer"` explicitly since our default is `callTracer` rather than structLogger.
- Bash-3.2-compatible (no `mapfile` / `declare -A`); reads service names from column 2 of `kurtosis enclave inspect`.

### Verification

Live comparison on `make localnet` (besu / geth / ethrex / lighthouse), tracing a single mainnet-shape tx (353 structLogs) across all three EL clients:

```
                ORIGINAL    BYTE-COMPAT    
                            (after these commits)
geth vs ethrex   4968 lines      0 lines ✓
besu vs geth     n/a          1560 lines  (pre-existing cross-client drift)
besu vs ethrex   n/a          1560 lines  (= besu vs geth)
```

Confirmed at every level:
- Same set of fields per step (`pc, op, gas, gasCost, depth, stack`; `refund` / `storage` when applicable).
- Same JSON types (number / string / array / object).
- Same field values.
- Same encoding (decimal numbers for gas/gasCost/refund; compact `"0xN"` hex for stack; full 64-char hex for storage keys/values; string mnemonics for `op`).
- Only difference is in-source key ordering (geth: insertion order, us: alphabetical via `BTreeMap`), which is JSON-spec irrelevant — every parser treats them as equivalent.

### Tests

`test/tests/levm/opcode_tracer_tests.rs`: 8 tests passing.
- Pre-existing: basic, sstore single-entry, memory capture, return-data capture, stack-disabled (now asserts JSON null per structLogger), JUMPDEST synthesis.
- New: `opcode_tracer_sstore_storage_accumulates`, `opcode_tracer_sstore_refund_on_clearing_step`.

`cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

### Not addressed

- **Key ordering on the wire.** Production parsers don't care; only raw text diffs would. One-line `Cargo.toml` change (`serde_json` `preserve_order` feature) if any tool ever needs it.
- **PREVRANDAO vs DIFFICULTY**: kept `PREVRANDAO` (EIP-4399-correct). Lucas's original concern was based on observing geth still emit `DIFFICULTY`, but that's geth's legacy naming, not a spec requirement. EIP-3155 doesn't constrain the mnemonic string.

crates/common/tracing.rs
```diff
@@ -125,3 +162,566 @@ pub struct PrePostState {
...
+    pub return_data: bytes::Bytes,
+    /// Gas refund counter.
+    pub refund: u64,
+    /// `Some(vec)` when stack capture is enabled (bottom-first); `None` when disabled.
+    /// Each format serializer decides how to render `None`: structLogger emits JSON null,
+    /// EIP-3155 emits `[]` (per spec's "MUST initialize to empty array" rule).
+    pub stack: Option<Vec<U256>>,
+    /// `Some(chunks)` when memory capture is enabled; `None` when disabled (field omitted).
+    pub memory: Option<Vec<MemoryChunk>>,
+    /// `Some(map)` at SLOAD/SSTORE steps when storage capture is enabled (single entry); `None` otherwise.
```
iovoid: "single entry"?

iovoid (approved):

crates/vm/levm/src/opcode_tracer.rs
ElFantasma: Confirmed — that statement was not true. The non-accumulation here diverged from geth's structLogger, which carries a cumulative map of every slot touched by SLOAD/SSTORE so far in the trace.

Fixed in [`3498f5853`](https://github.com/lambdaclass/ethrex/pull/6595/commits/3498f5853): `LevmOpcodeTracer` now owns a `cumulative_storage: BTreeMap<H256, H256>` updated in `pre_step_capture` on every SLOAD/SSTORE, and each captured step's `storage` field is overwritten with a clone of that map. Synthetic steps (e.g. fused JUMPDEST) pass `storage_kv: None` and are unaffected.

The misleading comment is gone too — the build_step helper now reads:

```rust
// Storage: presence/absence of `storage_kv` is what signals "this step
// touches storage". Callers from `pre_step_capture` overwrite this with a
// snapshot of the tracer's cumulative storage map; callers from synthetic-
// step paths (e.g. fused JUMPDEST) pass `None` and get `None` here.
```

A regression test (`opcode_tracer_sstore_storage_accumulates`) pins that two consecutive SSTOREs produce step-2 with `{slot1}` and step-5 with `{slot1, slot2}`. Cross-validated live against geth on a kurtosis multi-client enclave — `debug_traceTransaction` byte-equivalent on the storage field after this commit (see [verification comment](https://github.com/lambdaclass/ethrex/pull/6595#issuecomment-4490641822)).

crates/common/tracing.rs
ElFantasma: Good catch — that doc comment was stale from before the storage-accumulation fix in `3498f5853`. The field really does carry a cumulative map now; `(single entry)` was a leftover from the old non-cumulative behavior. Updated in `a4caab16d`:

```rust
/// `Some(map)` at SLOAD/SSTORE steps when storage capture is enabled; `None`
/// otherwise. The map is a cumulative snapshot of every slot touched by an
/// SLOAD/SSTORE so far in the transaction — matching geth's structLogger.
/// The tracer maintains this in `LevmOpcodeTracer::cumulative_storage`.
pub storage: Option<BTreeMap<H256, H256>>,
```

ilitteri (approved):

PR #6596, opened 2026-05-11
fix(l1): final mempool re-drain in build_payload_loop

edg-l:
## Summary

Closes a race in the payload builder that causes `engine_getPayload` to occasionally return a stale, often empty, payload when a transaction arrives in the narrow window between the loop's most recent rebuild and the cancellation triggered by `engine_getPayload`.

Fix: one additional synchronous `build_payload` call at the end of `build_payload_loop`, just before returning. It re-reads the mempool one last time so any transaction that landed during the race window is included.

## Aim: kill the Hive Paris reorg-flake family

There is a family of four Hive Paris Engine tests that have been flaking on this repo for weeks. Three of them are already on a known-flaky ignore list in `.github/scripts/check-hive-results.sh:62-66`:

```
Invalid Missing Ancestor Syncing ReOrg, Timestamp,           EmptyTxs=False, CanonicalReOrg=False, Invalid P8
Invalid Missing Ancestor Syncing ReOrg, Timestamp,           EmptyTxs=False, CanonicalReOrg=True,  Invalid P8
Invalid Missing Ancestor Syncing ReOrg, Transaction Value,   EmptyTxs=False, CanonicalReOrg=False, Invalid P9
Invalid Missing Ancestor Syncing ReOrg, Transaction Nonce,   EmptyTxs=False, CanonicalReOrg=False, Invalid P9   ← new this week
```

The argument that this fix addresses all four follows.

### Shared precondition

The four test names share parameter `EmptyTxs=False`. That parameter asserts "this test expects a payload containing at least one transaction." It is set at test-instance generation time, not a behavioral choice the runtime makes.

### Shared script structure

The `Invalid Missing Ancestor Syncing ReOrg` family runs this exact sequence:

1. `eth_sendRawTransaction` — seed the mempool
2. `engine_forkchoiceUpdatedV1` with attrs — start building
3. `engine_getPayloadV1` — pull the built payload
4. Mutate **one** field of the payload (needs at least one tx in step 3)
5. `engine_newPayloadV1` with mutated payload — submit, expect INVALID
6. Assert client behaviour on the reorg

The mutated field changes per variant (`Timestamp`, `Transaction Value`, `Transaction Nonce`), but step 4 only succeeds when step 3 produced a non-empty payload. So all four share the precondition **"step 3 returns a payload with the seeded tx in it."** The mutated field is irrelevant to the precondition.

### The race produces exactly this precondition failure

Captured Hive log on the new failing variant shows:

```
FAIL: Unable to customize payload: no transactions available for modification
```

i.e. step 3 returned an empty payload. The race is in `build_payload_loop`:

```
T0   FCU-with-attrs           initial build pass, res = empty
T1   eth_sendRawTransaction   mempool.insert + tx_added.notify_waiters()
T2   loop wakes on notified   spawns pass 1 (would include the tx)
T3   engine_getPayload        cancel_token.cancel()
T4   run_until_cancelled      returns None, pass 1 result dropped
T5   loop exits               returns res = empty (from T0)
```

The mempool has the tx since T1 (we verified `add_transaction` holds the write lock until insert is complete, so the tx is durably visible to readers). The bug is solely on the build side: pass 1's result is dropped on cancellation.

### Why this family is the canary

Other engine-api tests either:

- Don't seed transactions (`EmptyTxs=True` variants) — empty payload is fine.
- Use the CL-mock's own payload producer instead of ethrex (`Modified Geth Module` in the log) — bypasses ethrex's builder.
- Pad the FCU-to-getPayload window with deliberate sleeps, hiding the race.

The reorg family doesn't pad — reorg semantics require tight timing between FCU and getPayload, so this family exposes the race more than any other.

### Why the fix targets all four

After the loop exits — regardless of how it exited — the fix runs one more synchronous `build_payload(payload)`. That call:

- Reads the mempool fresh via `fetch_mempool_transactions`
- Re-executes block building with whatever's in the mempool now
- Assigns the result to `res` before returning

The fix doesn't care which field the test will mutate. It only guarantees that the returned payload reflects the current mempool state. If the mempool had the tx at any point before getPayload returned, the payload will include it.

### Honest limits

- **Inference, not yet verified for all four.** The argument is structural: same precondition, same race, same fix. The Hive run on this branch should confirm.
- **Possible alternative cause per variant.** It is conceivable that one or more variants fail for a different reason (e.g. the timestamp variants might also have a sub-race on block timing). If so, the fix removes the empty-payload symptom but a different test failure could surface. The Hive log on this branch will reveal that.
- **The existing comment in `check-hive-results.sh:61` is wrong.** It says "These are hive framework issues, not ethrex bugs." Framework behavior is correct; the empty payload is the bug. Worth re-examining as a follow-up.

## Fix

`crates/blockchain/payload.rs` — add one final `build_payload` call after the rebuild loop exits. Errors are downgraded to a warning that keeps the previous `res`; we never make the payload worse than the loop already produced.

## Cost

One extra full payload build per `engine_getPayload`. For typical mempool depths this is a few milliseconds; geth and reth pay the same cost. The latency lands on the block-proposal critical path but is bounded by the same code path that runs in the rebuild loop today.

## Affected paths

`engine_getPayloadV1/V2/V3/V4` only. Non-builder full nodes never call `get_payload` and do not hit this code.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo check --workspace --all-targets`
- [x] `make lint-l1`
- [x] Local Hive run: `make run-hive SIMULATION=ethereum/engine TEST_PATTERN='engine-api/Invalid Missing Ancestor Syncing ReOrg, (Timestamp|Transaction (Value|Nonce))'`
- [x] CI: `Hive - Paris Engine tests` passes without the entries in `KNOWN_FLAKY_TESTS`
- [ ] No regression in `Hive - Cancun Engine tests` / `Engine withdrawal tests` / `Reorg Tests`

## Follow-up (not in this PR)

After this PR lands and Hive shows clean on the four variants for a few runs:

1. Remove the three entries from `KNOWN_FLAKY_TESTS` in `.github/scripts/check-hive-results.sh`.
2. Revisit the architectural alternative: drop the periodic-rebuild loop entirely and lazy-build at `getPayload` time (snapshot-then-build model). That removes the timing dependency by construction.

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -413,6 +413,9 @@ impl Blockchain {
         const SECONDS_PER_SLOT: Duration = Duration::from_secs(12);
         // Attempt to rebuild the payload as many times within the given timeframe to maximize fee revenue
         // TODO(#4997): start with an empty block
+        // Snapshot the mempool sequence *before* the build so any tx that lands
+        // during the build is seen as newer than the current `res`.
+        let mut last_built_seq = self.mempool.tx_seq();
         let mut res = self.build_payload(payload.clone())?;
         while start.elapsed() < SECONDS_PER_SLOT && !cancel_token.is_cancelled() {
             // Wait for new transactions, cancellation, or slot deadline before rebuilding
@@ -425,6 +428,7 @@ impl Blockchain {
             }
             let payload = payload.clone();
             let self_clone = self.clone();
+            let seq_before = self.mempool.tx_seq();
             let building_task =
                 tokio::task::spawn_blocking(move || self_clone.build_payload(payload));
             // Cancel the current build process and return the previous payload if it is requested earlier
@@ -433,6 +437,7 @@ impl Blockchain {
             match cancel_token.run_until_cancelled(building_task).await {
                 Some(Ok(current_res)) => {
                     res = current_res?;
+                    last_built_seq = seq_before;
                 }
                 Some(Err(err)) => {
                     warn!(%err, "Payload-building task panicked");
@@ -440,6 +445,23 @@ impl Blockchain {
                 None => {}
             }
         }
+
+        // If a tx landed after the snapshot that produced `res`, do one final
+        // build before returning. Covers both races: (a) cancellation dropping
+        // an in-progress rebuild via `run_until_cancelled`, and (b) the slot-
+        // timeout `select!` arm winning over a simultaneous `tx_added`
+        // notification near the slot boundary.
+        if self.mempool.tx_seq() > last_built_seq {
+            let blockchain = self.clone();
+            match tokio::task::spawn_blocking(move || blockchain.build_payload(payload)).await {
+                Ok(Ok(final_res)) => res = final_res,
+                Ok(Err(err)) => {
+                    warn!(%err, "Final payload rebuild failed; returning previous result")
+                }
+                Err(err) => warn!(%err, "Final payload rebuild task panicked"),
+            }
+        }
+
         Ok(res)
     }
 
--- a/.github/scripts/check-hive-results.sh
+++ b/.github/scripts/check-hive-results.sh
@@ -58,11 +58,8 @@ rm -rf "${failed_logs_root}"
 mkdir -p "${failed_logs_root}"
 
 # Tests excluded from the failure count (substring match against test case
-# name). Reserved for genuinely flaky hive-framework tests, not ethrex bugs.
+# name).
 KNOWN_EXCLUDED_TESTS=(
-  "Invalid Missing Ancestor Syncing ReOrg, Timestamp, EmptyTxs=False, CanonicalReOrg=False, Invalid P8"
-  "Invalid Missing Ancestor Syncing ReOrg, Timestamp, EmptyTxs=False, CanonicalReOrg=True, Invalid P8"
-  "Invalid Missing Ancestor Syncing ReOrg, Transaction Value, EmptyTxs=False, CanonicalReOrg=False, Invalid P9"
 )
 
 # Build a jq filter that excludes the known-excluded tests.
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{BTreeMap, VecDeque, hash_map::Entry},
     sync::RwLock,
+    sync::atomic::{AtomicU64, Ordering},
 };
 
 use rustc_hash::{FxHashMap, FxHashSet};
@@ -116,6 +117,11 @@ pub struct Mempool {
     /// Signaled on transaction and blobs bundle insertions so payload
     /// builders can await new work instead of busy-looping.
     tx_added: tokio::sync::Notify,
+    /// Monotonic counter incremented on every transaction insertion. Used by
+    /// the payload builder to detect whether new txs landed since it last
+    /// snapshotted the mempool, so it can decide whether a stale build is safe
+    /// to return.
+    tx_seq: AtomicU64,
 }
 
 impl Mempool {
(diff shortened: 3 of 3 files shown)


crates/blockchain/payload.rs
```diff
@@ -440,6 +440,24 @@ impl Blockchain {
...
             }
         }
+
+        // On the cancellation path only, re-read the mempool one more time
+        // before returning. `run_until_cancelled` may have dropped an
+        // in-progress rebuild, leaving `res` stale (possibly empty) even
+        // though a tx is already visible in the mempool. The slot-timeout
+        // path skips this — by then the loop has been rebuilding continuously
+        // for `SECONDS_PER_SLOT` and the latest result is already fresh.
+        if cancel_token.is_cancelled() {
```
ElFantasma: Gating on `is_cancelled()` misses a smaller-but-real version of the same race in the slot-timeout exit. Sequence:

1. T0: loop is `.await`ing the `select!` on `notified` / `cancelled` / `sleep(remaining)`.
2. T0+: tx arrives in mempool, fires `tx_added` notify — `notified` future becomes ready.
3. Simultaneously: `start.elapsed() >= SECONDS_PER_SLOT` triggers `sleep(remaining)` to also be ready.
4. Tokio picks `sleep` arm (race-fair), `break` exits the loop.
5. `is_cancelled()` is false → final build skipped → return `res` from before the just-missed tx.

The "loop has been rebuilding continuously" argument is true on the macro scale but doesn't cover the very last hundred-ms window of the slot, which is exactly when block-proposers are most likely to see late-arriving tip-bumping txs. Two options: drop the `is_cancelled()` guard (always do the final build), or guard on a weaker condition like "current mempool's `tx_added` count > snapshot taken before the last completed build." Not blocking — the current change captures the dominant failure mode from the Hive log.

crates/blockchain/payload.rs
```diff
@@ -440,6 +440,24 @@ impl Blockchain {
...
+            let blockchain = self.clone();
+            match tokio::task::spawn_blocking(move || blockchain.build_payload(payload)).await {
```
ElFantasma, line 456: Synchronous `build_payload` on the engine_getPayload critical path. For typical mempool depths this is the few-ms the description mentions, but on a near-full mempool with many high-gas calls (DEX day, mint storms) `build_payload` runs the full executor and can take 100-400 ms in practice — and this fires every single time on the cancellation path, even when the loop's `res` is already fresh.

Optional micro-optimization that keeps the same guarantee: skip the final build if no `tx_added` notification has fired since the last completed loop iteration. Concretely, snapshot a counter (or `Instant`) when `res` is assigned at line 435 and compare here. Saves the rebuild on "clean cancellation" (the common case — `engine_getPayload` arrives right at slot boundary with no new tx) while still covering the race window. Not blocking — happy to take this as a follow-up if you want symptom fix in now.

ilitteri (approved):

crates/blockchain/payload.rs
edg-l: Good catch on the slot-timeout race — addressed in the latest commit. Replaced the `is_cancelled()` guard with a `mempool.tx_seq() > last_built_seq` check (atomic counter bumped on every `add_transaction`). It fires on both the cancellation race and your slot-timeout `select!` edge case (any tx that landed after the snapshot driving `res` triggers the final rebuild), regardless of which loop-exit arm won.

crates/blockchain/payload.rs
edg-l: Took the micro-optimization you suggested. The new commit gates the final rebuild on `mempool.tx_seq() > last_built_seq` — `tx_seq` snapshotted right before each `build_payload` call and committed to `last_built_seq` only when that build's result becomes `res`. Clean cancellation with no new tx since the last successful build skips the rebuild entirely, so the 100-400 ms hit on a busy mempool no longer fires on the happy path.

ElFantasma (approved):

azteca1998 (approved):

PR #6598, opened 2026-05-11
perf(l1): two-CF receipts migration

azteca1998:
## Summary

- Replace the in-place cursor-held migration (re-key + per-key deletes on `receipts`) with a two-CF approach: copy entries from `receipts` to `receipts_v2` with fixed-width keys, then drop the old CF — all in a single startup
- No per-key delete tombstones → less compaction overhead
- Old CF dropped atomically by RocksDB (drops SST files directly, no compaction needed)
- One read pass + one write pass, no mixed reads/writes on same CF

### Migration lifecycle (single startup)

1. `RocksDBBackend::open()` opens all existing CFs (including legacy `receipts`) plus `TABLES` CFs
2. `run_pending_migrations()` copies entries from `receipts` → `receipts_v2` with new key format
3. `drop_obsolete_cfs()` drops `receipts` (not in `TABLES`) — same startup, right after migration
4. On subsequent startups, `receipts` no longer exists on disk and is never recreated

### Benchmark (150M synthetic entries)

| Approach | Wall-clock time | Notes |
|----------|----------------|-------|
| **Two-CF (this PR)** | **88s** | Copy to new CF, old CF dropped same startup |
| Cursor-held (prev) | 487s | In-place re-key + per-key deletes |
| Temp-file dump | 221s | Dump keys to file, re-key + delete |

### Files changed

- `tables.rs` — Add `RECEIPTS_V2` constant, remove `RECEIPTS` from `TABLES` (kept as const for migration reads only)
- `rocksdb.rs` — Switch CF options to `RECEIPTS_V2`; extract cleanup into `drop_obsolete_cfs()` so it runs after migrations
- `migrations.rs` — Rewrite `migrate_1_to_2` to copy to new CF (no deletes)
- `store.rs` — All receipt read/write ops use `RECEIPTS_V2`; call `drop_obsolete_cfs()` after migrations
- `Cargo.toml` — Gate `libc`/`tracing-subscriber` behind `migration-tools` feature
- Added `bench_migration.rs` and `seed_migration_test.rs` binaries for benchmarking (require `--features migration-tools`)

## Test plan

- [x] `cargo test -p ethrex-storage --features rocksdb -- migrations` passes
- [x] `cargo clippy -p ethrex-storage --features rocksdb` clean
- [x] `cargo check` (full workspace) passes
- [ ] Benchmark on production-sized data (srv1 has 16 GB receipts CF)


crates/storage/api/tables.rs
```diff
@@ -113,6 +118,7 @@ pub const TABLES: [&str; 19] = [
     PENDING_BLOCKS,
     TRANSACTION_LOCATIONS,
     RECEIPTS,
+    RECEIPTS_V2,
```
ElFantasma, line 127: **The old `RECEIPTS` CF won't auto-drop.** The migration doc (`migrations.rs:104-106`) and the PR description both say:

> *"the old `receipts` CF is **not** deleted here — it will be dropped automatically by the auto-cleanup in `RocksDBBackend::open()` on the next startup (since `RECEIPTS` is no longer listed in `TABLES`)."*

But `RECEIPTS` IS still in `TABLES` (line 120 above this addition). The auto-cleanup at `backend/rocksdb.rs` only drops CFs whose name is NOT in `TABLES`:

```rust
for cf_name in &existing_cfs {
    if cf_name != "default" && !TABLES.contains(&cf_name.as_str()) {
        warn!("Dropping obsolete column family: {}", cf_name);
        let _ = db.drop_cf(cf_name) ...
    }
}
```

So with both `RECEIPTS` and `RECEIPTS_V2` in `TABLES`, the old CF survives forever. After migration both CFs hold the same data — on srv1's 16 GB receipts CF, that's 16 GB of duplicate state that never goes away.

The test at `migrations.rs:250-254` confirms the intent ("Old keys should still be in RECEIPTS (dropped at startup)") but doesn't actually verify the drop happens.

Fix: remove `RECEIPTS,` from the `TABLES` array (line 120). The migration code still references it via `tables::RECEIPTS` for read-only reading, which is fine — `prefix_iterator` against a CF name that's been opened (via the auto-create path) works regardless of whether the constant is in `TABLES`. The auto-cleanup loop will then see `"receipts"` is unlisted and call `drop_cf` on next startup.

Worth a quick test of this scenario before merge — e.g., a unit test that runs migrate + reopens the backend + asserts `receipts` CF is gone.

edg-l (dismissed):
(superseded; see review below)

crates/storage/api/tables.rs
```diff
@@ -118,6 +123,7 @@ pub const TABLES: [&str; 20] = [
     PENDING_BLOCKS,
     TRANSACTION_LOCATIONS,
     RECEIPTS,
```
edg-l: Blocker: leaving `RECEIPTS` here defeats the auto-drop in `RocksDBBackend::open()` (it only drops CFs not in `TABLES`). After migration the old `receipts` CF stays on disk forever, which contradicts the PR description and the comment on `migrate_1_to_2`.

Either:
1. Drop `RECEIPTS` from this list and re-order startup so migrations run before the obsolete-CF sweep.
2. Expose `drop_cf` on `StorageBackend` and drop it from inside `migrate_1_to_2` after the copy. Cleaner and more atomic.

crates/storage/migrations.rs
```diff
@@ -93,6 +95,69 @@ fn write_metadata_version(db_path: &Path, version: u64) -> Result<(), StoreError
...
 }
 
+/// Migrates the RECEIPTS table from RLP-encoded `(BlockHash, u64)` keys
+/// to raw `block_hash (32B) || index (8B big-endian u64)` keys in a new
+/// `receipts_v2` column family.
+///
+/// This two-CF approach copies entries from the old `receipts` CF to
+/// `receipts_v2` with the new key format. The old `receipts` CF is **not**
+/// deleted here — it will be dropped automatically by the auto-cleanup in
+/// `RocksDBBackend::open()` on the next startup (since `RECEIPTS` is no
```
edg-l: Comment is wrong as the code stands. `RECEIPTS` is still in `TABLES` (see `api/tables.rs`), so the auto-cleanup will not drop it. Either fix the implementation to drop the CF, or fix the comment.

crates/storage/migrations.rs
```diff
@@ -93,6 +95,69 @@ fn write_metadata_version(db_path: &Path, version: u64) -> Result<(), StoreError
...
+    }
+
+    // Flush remaining entries.
+    if !batch.is_empty() {
+        let count = batch.len() as u64;
+        let mut tx = backend.begin_write()?;
+        tx.put_batch(RECEIPTS_V2, batch)?;
+        tx.commit()?;
+        migrated += count;
+        tracing::info!("Migration v1→v2: migrated {migrated} RECEIPTS entries so far");
```
edg-l: nit: the line below already logs the total (`Migration v1→v2 complete: migrated N entries total`), so the `"so far"` message on the final flush is redundant. Keep only the summary.

crates/storage/Cargo.toml
```diff
@@ -26,6 +26,8 @@ tokio = { workspace = true, features = ["rt", "sync"] }
 fastbloom = "0.14"
 rayon.workspace = true
 lru.workspace = true
+libc = "0.2"
+tracing-subscriber = { workspace = true }
```
edg-l: `libc` and `tracing-subscriber` are only used by the bench/seed binaries (`getrusage`, subscriber init), but they're in `[dependencies]` so they ship in every prod build. Move them to `[dev-dependencies]` and the binaries to `crates/storage/examples/` or a separate `tooling/storage-bench` crate.

edg-l (commented):
Nice speedup, two-CF is the right shape. One blocker and a few cleanups.

### Blocker

The old `receipts` CF never gets dropped. `TABLES` still contains both `RECEIPTS` and `RECEIPTS_V2`, so the auto-cleanup in `RocksDBBackend::open()` (which only drops CFs not in `TABLES`) skips it. The doc comment on `migrate_1_to_2` claims the opposite. Net effect: after migration users keep the full old CF (~16 GB on srv1) forever, which undoes the main benefit of the PR.

Two fixes:
1. Drop the CF inside `migrate_1_to_2` after the copy finishes. Atomic, deterministic. Needs `drop_cf` on `StorageBackend`. Recommended.
2. Remove `RECEIPTS` from `TABLES` and re-order startup so migrations run before the obsolete-CF sweep. More invasive since the sweep is inside `open()`.

### Should fix

- Bench/seed binaries are committed to `crates/storage/` with `libc` and `tracing-subscriber` in `[dependencies]`, not `[dev-dependencies]`. They end up in every prod build. Move them to `crates/storage/examples/` or a separate `tooling/storage-bench` crate. `seed_migration_test.rs` also duplicates the CF setup from `rocksdb.rs`, which will drift.
- CHANGELOG entry: date is `2026-04-29` (today is `2026-05-21`), PR link is `#6548` (this is `#6598`), and the bullet describes the prior PR's work, not this one. Rewrite or drop.
- `fetch_count` mismatch returns `Internal` instead of `Null`. Old code surfaced corruption as either a 500 per-tx or `Null` via `receipts.get(index)`. Pick one consistent behavior and document it.

### Nits

- `migrations.rs` tail flush logs `"migrated N entries so far"`; redundant since the next line logs the total.
- `block.rs:386`: `.take(fetch_count)` is redundant; `zip` short-circuits and the length check above already asserted both sides are `fetch_count`.
- `block.rs:273`: `GetRawReceipts` no longer needs `body`. Drop the `get_block_body` call.
- `store.rs:1102-1120`: could `seek` to `block_hash || start_index.to_be_bytes()` instead of decoding every key. Negligible for typical block sizes.
- `Cargo.toml`: stray double blank line before `[lib]`.

crates/storage/Cargo.toml
```diff
@@ -35,9 +37,20 @@ rocksdb = ["dep:rocksdb"]
 tempfile.workspace = true
 tokio = { workspace = true, features = ["full"] }
 
+
```
edg-l: nit: stray double blank line.

tooling/migrations/src/bin/seed_migration_test.rs
```diff
@@ -0,0 +1,286 @@
+/// Standalone binary to seed ~150M old-format (RLP-encoded) RECEIPTS keys
```
edg-l, line 1: Duplicates the full RocksDB CF setup from `backend/rocksdb.rs` (`ALL_CFS`, per-CF compression rules, block opts). Will drift the moment someone touches `rocksdb.rs`. If we keep the binary, reuse the CF descriptor logic from the storage crate.

crates/networking/rpc/eth/block.rs
```diff
@@ -353,16 +360,32 @@ pub async fn get_all_block_rpc_receipts(
...
     let block_info = RpcReceiptBlockInfo::from_block_header(header);
-    // Fetch receipt for each tx in the block and add block and tx info
+    // Fetch receipts: only up to target_index+1 when set, otherwise all
+    let fetch_count = target_index
+        .map(|ti| (ti + 1) as usize)
+        .unwrap_or(body.transactions.len());
+    let all_receipts = storage
+        .get_receipts_for_block_from_index(&block_hash, 0, Some(fetch_count))
+        .await?;
+    if all_receipts.len() != fetch_count {
```
edg-l, line 372: Behavior change: the old per-tx loop returned `RpcErr::Internal("Could not get receipt")` on the first missing receipt, surfaced as a 500. The RPC could also return `Null` via `receipts.get(index)`. Now any length mismatch errors out before the loop. Pick one consistent behavior for corruption (500 or null) and document it.

crates/networking/rpc/eth/block.rs
```diff
@@ -353,16 +360,32 @@ pub async fn get_all_block_rpc_receipts(
...
+    }
     let mut last_cumulative_gas_used = 0;
     let mut current_log_index = 0;
-    for (index, tx) in body.transactions.iter().enumerate() {
+    for (index, (tx, receipt)) in body
+        .transactions
+        .iter()
+        .zip(all_receipts.iter())
+        .enumerate()
+        .take(fetch_count)
```
edg-l: `.take(fetch_count)` is redundant. `zip` short-circuits to the shorter side, and the length check above just asserted both sides are exactly `fetch_count`.

crates/networking/rpc/eth/block.rs
```diff
@@ -270,11 +270,11 @@ impl RpcHandler for GetRawReceipts {
         };
         let header = storage.get_block_header(block_number)?;
         let body = storage.get_block_body(block_number).await?;
-        let (header, body) = match (header, body) {
+        let (header, _body) = match (header, body) {
```
edg-l: `body` is no longer used (now `_body`). Drop the destructuring and skip the `get_block_body` call; saves a DB read on every `eth_getRawReceipts`.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 - Skip `vm.run_execution()` for transfers to codeless EOAs [#6570](https://github.com/lambdaclass/ethrex/pull/6570)
 
+### 2026-04-29
```
edg-l: Three issues:
- date is `2026-04-29`, today is `2026-05-21`;
- PR link is `#6548`, this PR is `#6598`;
- the bullet describes the previous PR's work (fixed-width keys plus cursor iteration), not this one's (two-CF migration).

Rewrite to describe the two-CF migration, or drop if `#6548` already shipped its own.

edg-l (approved):
LGTM, all prior review items addressed. A few non-blocking nits for a follow-up:

- Migration progress log fires every batch (BATCH_SIZE=10k), so ~15k INFO lines on a 150M-entry migration. Drop to `debug!` or log every Nth batch.
- `seed_migration_test.rs` duplicates the CF/options setup from `rocksdb.rs`; it will silently drift if those constants change. Acceptable while gated behind `migration-tools`.
- `drop_obsolete_cfs` swallows `list_cf` errors via `unwrap_or_default()`. Best-effort is fine here, worth a one-line comment.

Crash safety verified: an interrupted migration leaves metadata at v1, and `drop_obsolete_cfs` only runs after `run_pending_migrations` returns Ok, so the legacy `receipts` CF survives until the copy completes. Re-runs are idempotent.

ElFantasma (approved):

tooling/migrations/src/bin/bench_migration.rs
iovoid, line 1: Shouldn't this be on the tooling folder?

tooling/migrations/src/bin/seed_migration_test.rs
iovoid, line 1: Shouldn't this be in test/ ?

ilitteri (approved):

PR #6599, opened 2026-05-11
feat(l1): enforce per-transaction wire-size cap in mempool admission

ilitteri:
**Motivation**

Every major Ethereum execution client (geth `txMaxSize`, reth `DEFAULT_MAX_TX_INPUT_BYTES`, nethermind `MaxTxSize` / `MaxBlobTxSize`, erigon size enforcement) ships a per-transaction wire-size cap at admission. Without one a single oversized transaction can chew bandwidth and pool capacity at near-zero attacker cost.

**Description**

- New constants `MAX_TX_SIZE = 128 KiB` and `MAX_BLOB_TX_SIZE = 1 MiB` in `crates/common/types/constants.rs`.
- `Blockchain::validate_transaction` enforces `MAX_TX_SIZE` on the canonical RLP encoding of non-blob transactions.
- `Blockchain::add_blob_transaction_to_pool` enforces `MAX_BLOB_TX_SIZE` on the **wire wrapper** — `Transaction::encode_canonical_to_vec().len() + BlobsBundle::encode_to_vec().len()` — to match geth (`tx.Size()` includes the attached sidecar), nethermind (`MaxBlobTxSize` is checked on the wrapper form), and erigon (`ValidateSerializedTxn` works on the raw wire bytes which for blob txs is the wrapper-with-sidecar). The two encoded sizes are summed because ethrex stores the core tx and the bundle in separate structs; the ±few bytes of outer list framing are rounding error at the 1 MiB scale.
- New error `MempoolError::TxSizeExceeded { actual, limit }`.
- Drops the redundant `MAX_TRANSACTION_DATA_SIZE` calldata-only check from `validate_transaction` — the new encoded-size cap is strictly tighter (a 128 KiB calldata payload always encodes to > 128 KiB).
- Unit test in `mempool_tests.rs` covering the non-blob path. Integration test for the wrapper check deferred (the `c-kzg`-gated `add_blob_transaction_to_pool` isn't currently exercised by `mempool_tests`, same pattern as PRs #6603/#6576).

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -2321,6 +2321,21 @@ impl Blockchain {
             return Ok(hash);
         }
 
+        // Wire-wrapper size cap for blob txs. Matches geth `txMaxSize = 1 MiB`
+        // (blobpool) and nethermind `MaxBlobTxSize`, which both bound the
+        // wire-wrapper form including the sidecar. ethrex stores the core tx
+        // and the bundle in separate structs, so sum the two encoded sizes
+        // (the ±few bytes of outer list framing are rounding error at this
+        // scale).
+        let wrapper_len =
+            transaction.encode_canonical_to_vec().len() + blobs_bundle.encode_to_vec().len();
+        if wrapper_len > MAX_BLOB_TX_SIZE {
+            return Err(MempoolError::TxSizeExceeded {
+                actual: wrapper_len,
+                limit: MAX_BLOB_TX_SIZE,
+            });
+        }
+
         // Validate blobs bundle after checking if it's already added.
         if let Transaction::EIP4844Transaction(transaction) = &transaction {
             blobs_bundle.validate(transaction, fork)?;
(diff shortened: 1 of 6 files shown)


avilagaston9 (approved):

crates/blockchain/blockchain.rs
```diff
@@ -2432,7 +2433,23 @@ impl Blockchain {
             .ok_or(MempoolError::NoBlockHeaderError)?;
         let config = self.storage.get_chain_config();
 
-        // NOTE: We could add a tx size limit here, but it's not in the actual spec
+        // Wire size cap: peer-policy default, not consensus. Matches geth
+        // `txMaxSize` (legacypool / blobpool), reth `DEFAULT_MAX_TX_INPUT_BYTES`,
+        // nethermind `MaxTxSize` / `MaxBlobTxSize`. For EIP-4844 the canonical
+        // encoding of `Transaction` excludes the sidecar (which lives in the
+        // adjacent `BlobsBundle`), so this caps the core tx only.
+        let encoded_len = tx.encode_canonical_to_vec().len();
```
MegaRedHand: We might want to find another way to compute this.

crates/common/types/transaction.rs
```diff
@@ -3764,4 +3764,31 @@ mod tests {
...
+        let payload = vec![0u8; MAX_TX_SIZE + 1];
+        let tx = Transaction::EIP1559Transaction(EIP1559Transaction {
+            data: Bytes::from(payload),
+            ..Default::default()
+        });
+
+        assert!(
+            tx.encode_canonical_to_vec().len() > MAX_TX_SIZE,
+            "tx with > 128 KiB calldata must encode larger than MAX_TX_SIZE"
+        );
```
MegaRedHand: I'm not sure what this test is testing

MegaRedHand (approved):
LGTM. I think we can remove the unit tests

crates/blockchain/blockchain.rs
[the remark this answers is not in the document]
ilitteri: a73179014

ilitteri: a73179014

ilitteri:
@MegaRedHand 3ee3f72b5 — removed the two encode-size sanity tests per your review.

ilitteri:
9d21d06d5 — moved the non-blob size check before sender recovery per the @codex + @claude review feedback. The blob wrapper-size check is already first in add_blob_transaction_to_pool_inner.

edg-l (approved):

PR #6601, opened 2026-05-11
feat(l1): require percentage fee bump for mempool RBF (10%/100%)

ilitteri:
**Motivation**

The previous strict-`>` replacement rule lets a replacement bump fees by a single wei, enabling near-zero-cost replacement-spam that forces nodes to re-validate and re-gossip on every iteration (the Nethermind-#3094 attack class). Every major Ethereum execution client requires a percentage bump on each applicable fee field: 10% for non-blob, 100% for blob.

**Description**

- `BlockchainOptions` gains `price_bump_percent` (default 10) and `blob_price_bump_percent` (default 100), exposed through new constants `DEFAULT_PRICE_BUMP_PERCENT` and `DEFAULT_BLOB_PRICE_BUMP_PERCENT`.
- `Mempool::find_tx_to_replace` now takes the bump percents and applies the appropriate one per tx type:
  - blob bump for EIP-4844 (covers all three fee fields including `max_fee_per_blob_gas`)
  - standard bump for the 1559 fee pair on non-blob typed txs
  - standard bump for `gas_price` on legacy
  - replaces the prior OR-of-everything logic with a per-type match
- Type-change rejection: replacing a tx with a different `Transaction` variant (e.g. legacy → 1559, or blob → non-blob) is rejected via `std::mem::discriminant`. Cross-type replacement skews accounting (blob vs non-blob slot reservation) and matches geth's `tx.Type() == old.Type()` guard.
- New `is_bumped_u64` / `is_bumped_u256` helpers using `checked_mul` u128 arithmetic so a huge `existing` value rejects under the floor instead of silently saturating to `u64::MAX` / `U256::MAX`.
- New CLI flags `--mempool.price-bump` (env `ETHREX_MEMPOOL_PRICE_BUMP`) and `--mempool.blob-price-bump` (env `ETHREX_MEMPOOL_BLOB_PRICE_BUMP`), plumbed through both L1 and L2 initializers.
- Unit tests in `mempool.rs` covering helper edge cases plus end-to-end scenarios: 1-wei bump rejected, full 10% bump accepted, asymmetric bump rejected, blob 50% rejected, blob 100% on three axes accepted, zero-fee case, overflow rejected, blob↔non-blob and legacy-bump cross-type rejection.


crates/blockchain/mempool.rs
```diff
@@ -456,38 +456,75 @@ impl Mempool {
...
-            let eip1559_higher_fees = new_tx_max_fee_per_gas > old_tx_max_fee_per_gas
-                && new_tx_max_priority_fee_per_gas > old_tx_max_priority_fee_per_gas;
-            let legacy_higher_fees = new_tx_gas_price > old_tx_gas_price;
+        // Reject type-change replacements. Peer clients keep blob and
+        // non-blob transactions in separate sub-pools precisely so a
+        // cheap-to-replicate non-blob tx can't displace a blob tx (with
+        // its expensive sidecar) or vice versa. ethrex has a single pool,
+        // so the same guarantee has to be enforced here.
+        if std::mem::discriminant(tx) != std::mem::discriminant(tx_in_pool.transaction()) {
+            return Err(MempoolError::UnderpricedReplacement);
```
MegaRedHand: We should return a new error variant here

MegaRedHand (approved):
LGTM. Left a comment

crates/blockchain/mempool.rs
ilitteri: 7283c274b

crates/blockchain/mempool.rs
```diff
@@ -457,38 +457,75 @@ impl Mempool {
...
 
-            let eip1559_higher_fees = new_tx_max_fee_per_gas > old_tx_max_fee_per_gas
-                && new_tx_max_priority_fee_per_gas > old_tx_max_priority_fee_per_gas;
-            let legacy_higher_fees = new_tx_gas_price > old_tx_gas_price;
+        // Reject type-change replacements. Peer clients keep blob and
+        // non-blob transactions in separate sub-pools precisely so a
+        // cheap-to-replicate non-blob tx can't displace a blob tx (with
+        // its expensive sidecar) or vice versa. ethrex has a single pool,
+        // so the same guarantee has to be enforced here.
+        if std::mem::discriminant(tx) != std::mem::discriminant(tx_in_pool.transaction()) {
```
ElFantasma: **This rejects *any* type change, but the documented intent (and the error message at `error.rs:123`) is only blob ↔ non-blob.** That's a real behavioural regression vs. the old code, which permitted type changes as long as the fee bump rule held.

Concrete user-visible failure: someone submits a `LegacyTransaction` at `(sender, n)`, then tries to replace it with an `EIP1559Transaction` at the same `(sender, n)` bumping `gas_price` 10%+. Pre-PR this worked. Post-PR they hit `ReplacementTypeMismatch` and are stuck — no path forward except waiting for the tx to time out. EIP-7702 → EIP-1559 → EIP-2930 are all in the regular sub-pool in geth/reth/nethermind, and each accepts type changes inside it.

The intended check is just blob-vs-non-blob:
```rust
let new_is_blob = matches!(tx, Transaction::EIP4844Transaction(_));
let old_is_blob = matches!(
    tx_in_pool.transaction(),
    Transaction::EIP4844Transaction(_)
);
if new_is_blob != old_is_blob {
    return Err(MempoolError::ReplacementTypeMismatch);
}
```
The rest of the function already dispatches per tx-type correctly, so once the gate matches intent, no other change is needed.

I'd hold the PR on this one — it's a small fix but it stranding users on `LegacyTransaction → EIP1559Transaction` replacements is worse than the over-rejection cost. Add a unit test for the now-allowed 1559↔legacy bumped replacement to lock the behaviour in.

crates/blockchain/mempool.rs
```diff
@@ -457,38 +457,75 @@ impl Mempool {
...
+                    bump,
+                );
+                let bumped_blob = is_bumped_u256(
+                    tx_in_pool.max_fee_per_blob_gas().unwrap_or_default(),
+                    tx.max_fee_per_blob_gas().unwrap_or_default(),
+                    bump,
+                );
+                bumped_fee && bumped_tip && bumped_blob
+            }
+            // EIP-2930 / EIP-1559 / EIP-7702 / FeeToken / Privileged: 1559-style
```
ElFantasma: Catch-all `_` for the non-blob/non-legacy arm bundles EIP-2930, EIP-1559, EIP-7702, FeeToken, and Privileged into one branch via a comment. That means the next tx-type variant added to `Transaction` silently inherits this 1559-style fee comparison — which might be wrong for whatever the new variant's fee model is.

Cheap fix: enumerate the variants explicitly so the compiler forces a deliberate decision on the next addition. Either:
```rust
Transaction::EIP2930Transaction(_)
| Transaction::EIP1559Transaction(_)
| Transaction::EIP7702Transaction(_)
| Transaction::FeeTokenTransaction(_)
| Transaction::PrivilegedL2Transaction(_) => { ... }
```
or mark with `#[deny(non_exhaustive_omitted_patterns)]` if that's the style preference. Non-blocking, but it removes a foot-gun that's easy to miss in a future tx-type PR.

crates/blockchain/error.rs
```diff
@@ -119,6 +119,10 @@ pub enum MempoolError {
     InvalidTxSender(#[from] ethrex_crypto::CryptoError),
     #[error("Attempted to replace a pooled transaction with an underpriced transaction")]
     UnderpricedReplacement,
+    #[error(
+        "Attempted to replace a pooled transaction with one of a different type (e.g. blob vs. non-blob)"
```
ElFantasma: After the discriminant gate (`mempool.rs:472`) is narrowed to blob ↔ non-blob, this message is accurate. As-is today it's narrower than reality — "e.g. blob vs. non-blob" suggests this fires only for that case, but the code also rejects `LegacyTransaction → EIP1559Transaction` etc. (see the linked inline). Tied to the same fix; no separate action needed once the gate is narrowed.

crates/blockchain/mempool.rs
```diff
@@ -499,6 +536,33 @@ impl Mempool {
     }
 }
 
+/// Returns true iff `new >= floor(existing * (100 + bump_percent) / 100)`.
+/// Uses `u128` intermediates with checked arithmetic so an overflow on the
+/// threshold computation is treated as "reject" rather than silently
+/// admitting an under-priced replacement. A `bump_percent` of 0 collapses to
+/// `new >= existing`.
+fn is_bumped_u64(existing: u64, new: u64, bump_percent: u64) -> bool {
```
ElFantasma: nit: the `is_bumped_u64` / `is_bumped_u256` helpers compute `(existing * (100 + bump)) / 100` via integer division, which rounds **down** the threshold. Combined with the `>=` comparison, this admits replacements at *exactly* the bumped floor. That's geth/reth-compatible (verified) but worth one inline doc line on the function to pin the invariant against future refactors:

```rust
/// Threshold rounds down (toward zero), so `new == floor(existing * (100+bump) / 100)`
/// is admitted. Matches geth's `core/types/transaction_signing.go` / reth's
/// `txpool/pool/replacements.rs` behaviour.
```
Non-blocking — the tests already lock in the boundary.

ElFantasma (dismissed):
One inline I'd treat as blocking: the discriminant check over-rejects in a way that strands users; the rest are non-blocking.

ilitteri:
Branch updated to latest main and review feedback addressed in 00de88056:

**@ElFantasma (cross-type gate)** — Narrowed to blob ↔ non-blob only. Type changes within the regular pool (legacy ↔ 2930 ↔ 1559 ↔ 7702) are allowed again, matching geth/reth/nethermind and pre-PR behavior. Added `legacy_to_1559_replacement_allowed_when_bumped` and `legacy_to_legacy_and_1559_to_legacy_bump_rules` to lock it in.

**@Copilot (0-default field comparison)** — Fee cap and tip are now compared uniformly via `gas_fee_cap()`/`gas_tip_cap()`, so a legacy pooled tx's gas price acts as both dimensions (geth's `GasFeeCap`/`GasTipCap` translation). An underpriced typed replacement against a legacy pooled tx is now rejected; regression test added (`typed_replacement_against_legacy_pool_tx_uses_translated_dimensions`).

**@Copilot / @greptile (saturating_mul)** — Already resolved on the branch: both helpers use checked arithmetic with u128 intermediates (u64) / checked_mul (U256); overflow → reject. The monotonicity bug is gone, and `is_bumped_u64_huge_existing_rejects_under_floor` covers the edge.

**@greptile (blob bump-rate selection)** — Moot under the narrowed category gate: both sides must share blob-ness before the bump is selected, so the 100% rate can't be bypassed cross-type.

**@ElFantasma (catch-all arm)** — The per-variant match is gone entirely (uniform `gas_fee_cap`/`gas_tip_cap` + blob-fee comparison), so future tx variants can't silently inherit a wrong comparison.

**@ElFantasma (rounding nit)** — Doc line on the helpers pins the floor-rounding semantics (`new >= floor(existing * (100+bump) / 100)`), matching geth/reth.

`cargo test -p ethrex-blockchain --lib` — 49/49 green.

crates/blockchain/mempool.rs
```diff
@@ -1278,40 +1278,60 @@ impl Mempool {
...
-                && new_tx_max_priority_fee_per_gas > old_tx_max_priority_fee_per_gas;
-            let legacy_higher_fees = new_tx_gas_price > old_tx_gas_price;
+        // Reject blob ↔ non-blob category changes. Peer clients keep blob
+        // and non-blob transactions in separate sub-pools precisely so a
+        // cheap-to-replicate non-blob tx can't displace a blob tx (with its
+        // expensive sidecar) or vice versa. ethrex has a single pool, so the
+        // same guarantee has to be enforced here. Type changes WITHIN the
+        // regular pool (legacy ↔ 2930 ↔ 1559 ↔ 7702) are allowed, matching
+        // geth/reth/nethermind.
+        let new_is_blob = matches!(tx, Transaction::EIP4844Transaction(_));
```
ElFantasma: The narrowed gate is right, but this predicate under-counts what "blob" means in this tree. `FrameTransaction` already carries `max_fee_per_blob_gas` and `blob_versioned_hashes` (`transaction.rs:1982,1984`), so a blob-bearing frame tx classifies as non-blob here.

The 4844↔frame direction is incidentally safe (mismatch → rejected, though `ReplacementTypeMismatch`'s "blob vs. non-blob" wording will read as wrong to whoever hits it). The frame↔frame direction is not:

- `bump` at L1309 resolves to `price_bump_percent` (10%) instead of `blob_price_bump_percent` (100%)
- `bumped_blob` at L1325 short-circuits to `true`, so the blob-fee dimension is never compared at all

That's the cheap-sidecar-re-propagation case the 100% bump exists to prevent — a blob-carrying frame tx could be displaced on a 10% bump with no blob-fee check. How reachable it is today depends on whether admission currently accepts blob-carrying frame txs; #7038 is enabling exactly that and is approved-but-unmerged, so this goes live when it lands.

A shared `is_blob_carrying()` helper on `Transaction` (4844 OR frame-with-blob-hashes) would fix it in one place. `mempool.rs:915` uses the same `matches!` pattern for `is_blob_tx`, so it likely needs the same treatment.

crates/blockchain/mempool.rs
```diff
@@ -1383,6 +1403,39 @@ impl Mempool {
...
+        return false;
+    };
+    (new as u128) >= threshold
+}
+
+/// U256 variant of [`is_bumped_u64`]. Used for `gas_price` (legacy) and
+/// `max_fee_per_blob_gas` (EIP-4844). Same overflow → reject semantic via
+/// `checked_mul`.
+fn is_bumped_u256(existing: U256, new: U256, bump_percent: u64) -> bool {
+    let multiplier = U256::from(100u64 + bump_percent);
```
ElFantasma: greptile's multiplier point looks still open: `100u64 + bump_percent` adds in `u64` and wraps for `bump_percent > u64::MAX - 100`, producing a nonsensically small multiplier and admitting undersized replacements.

The reply to that thread ("both helpers use checked arithmetic with u128 intermediates (u64) / checked_mul (U256)") covers overflow of the *multiplication* — `checked_mul` does handle that — but not the wrapping of the multiplier itself. `is_bumped_u64` already has the correct widening (`100u128 + bump_percent as u128`), so the weaker arithmetic is in the production helper and the correct pattern is in the one production never calls.

```suggestion
    let multiplier = U256::from(100u64) + U256::from(bump_percent);
```

Practically unreachable, but it's a one-liner and makes the two helpers consistent.

crates/blockchain/mempool.rs
```diff
@@ -1383,6 +1403,39 @@ impl Mempool {
...
+/// Uses `u128` intermediates with checked arithmetic so an overflow on the
+/// threshold computation is treated as "reject" rather than silently
+/// admitting an under-priced replacement. A `bump_percent` of 0 collapses to
+/// `new >= existing`.
+///
+/// Threshold rounds down (toward zero), so `new == floor(existing * (100 +
+/// bump) / 100)` is admitted — matches geth's and reth's replacement rules.
+/// Currently only exercised by unit tests: production compares fees via the
+/// U256 accessors (`gas_fee_cap`/`gas_tip_cap`).
+#[cfg(test)]
```
ElFantasma: Worth flagging what this `#[cfg(test)]` implies for the overflow discussion earlier in the thread: `is_bumped_u64` is compiled only under test, and its own doc says so. Production compares fees exclusively through `is_bumped_u256` (three call sites: fee cap, tip, blob fee).

So `is_bumped_u64_huge_existing_rejects_under_floor` — cited in the reply as covering the monotonicity/overflow edge — is exercising a function that never runs in production. The behaviour it pins is real, but it isn't the code path that matters.

Two options, either is fine: drop the `#[cfg(test)]` and use the u64 helper where the inputs really are u64, or keep it test-only and move the edge-case coverage onto `is_bumped_u256` (see the other comment on the test module).

nit while you're here: the doc on `is_bumped_u256` at L1425 opens "U256 variant of [`is_bumped_u64`]" — an intra-doc link to a `#[cfg(test)]` item, so a normal `cargo doc` build has nothing to resolve it against.

crates/blockchain/mempool.rs
```diff
@@ -1876,4 +1929,323 @@ mod tests {
...
+
+    // --- is_bumped_u256 ------------------------------------------------
+
+    #[test]
+    fn is_bumped_u256_blob_100_percent_accepted() {
+        assert!(is_bumped_u256(U256::from(50u64), U256::from(100u64), 100));
+    }
+
+    #[test]
+    fn is_bumped_u256_blob_99_percent_rejected() {
```
ElFantasma: These two are the only tests for `is_bumped_u256`, and both are ordinary bump cases. Its `checked_mul(...) else { return false }` branch — the overflow → reject semantic the review thread was about, and the one production actually reaches — has no coverage.

The u64 mirror of this case is well covered (`is_bumped_u64_huge_existing_rejects_under_floor`), so it's a direct port: pick an `existing` near `U256::MAX` so `existing * (100 + bump)` overflows, and assert the call returns `false` rather than admitting.

ElFantasma (commented):
All four earlier comments are addressed; one new blob-classification gap inline is worth closing before this lands.

crates/blockchain/mempool.rs
ElFantasma: Fixed in 3abe5f5f9. Added a shared predicate rather than patching the two call sites:

```rust
/// True for EIP-4844 and for EIP-8141 frame txs with a non-empty
/// `blob_versioned_hashes` — i.e. exactly when `max_fee_per_blob_gas()` yields a fee.
pub fn is_blob_carrying(&self) -> bool {
    self.max_fee_per_blob_gas().is_some()
}
```

`find_tx_to_replace` now uses it on **both** sides of the category gate, so a blob-carrying frame tx takes the 100% bump and gets its blob-fee dimension compared. Confirmed the underlying asymmetry: `FrameTransaction` carries `max_fee_per_blob_gas` + `blob_versioned_hashes` (`transaction.rs:1982,1984`) and `max_fee_per_blob_gas()` already returns `Some` for it when the hash list is non-empty — so the accessor was the natural definition and no new match arm was needed.

Reachability is as described: admission still rejects blob-carrying frame txs (`FrameTxBlobsUnsupported`), so this isn't hittable today and goes live with #7038. Covered by a new `is_blob_carrying_covers_frame_txs_with_blob_hashes` unit test.

`ReplacementTypeMismatch`'s "blob vs. non-blob" wording is accurate once the predicate means *carries a sidecar*, so I left it as is.

Deliberately **not** changed: the `is_blob_tx` filter at `mempool.rs:915`. It has the same narrow `matches!`, but it's a pre-existing payload-builder query filter that this PR doesn't touch, and widening it changes block-building selection. Better as its own PR alongside #7038 — I'll raise it there.

crates/blockchain/mempool.rs
ElFantasma: Applied in 3abe5f5f9:

```rust
let multiplier = U256::from(100u64) + U256::from(bump_percent);
```

Both the addition and the multiplication now happen in `U256`, so neither can wrap into a smaller threshold. Pinned by a new assertion in `is_bumped_u256_overflowing_threshold_rejects` using `bump_percent = u64::MAX`.

crates/blockchain/mempool.rs
ElFantasma: Took the stronger of the two options in 3abe5f5f9: **deleted `is_bumped_u64`** and ported all five of its cases onto `is_bumped_u256`.

Production only ever compares through the U256 accessors (`gas_fee_cap` / `gas_tip_cap` / `max_fee_per_blob_gas`), so the u64 mirror was carrying test coverage for a function that never ran — including the overflow case cited as covering the monotonicity edge. Removing it puts the coverage on the code path that actually executes and resolves the intra-doc link too: the doc on `is_bumped_u256` no longer references a `#[cfg(test)]` item, so it resolves in a normal `cargo doc` build.

(For context, `cargo doc -p ethrex-blockchain` still reports broken links on this branch, but all of them are pre-existing on `main` — `compute_reorg_ceiling`, `check_admission`, `add_block_pipeline*`, etc. This change removes one rather than adding any.)

crates/blockchain/mempool.rs
ElFantasma: Added in 3abe5f5f9 — `is_bumped_u256_overflowing_threshold_rejects` is the direct port:

```rust
assert!(!is_bumped_u256(U256::MAX, U256::MAX, 100));        // checked_mul → None → reject
assert!(!is_bumped_u256(U256::MAX, U256::MAX, u64::MAX));   // multiplier can't wrap small
let _ = is_bumped_u256(U256::MAX, U256::zero(), u64::MAX);  // no panic on extremes
```

The u64 helper it mirrored is gone (see the other thread), so `is_bumped_u256` now carries the whole set: exact-bump, just-below, zero-bump, zero-existing, round-down boundary, and this overflow branch.

crates/blockchain/mempool.rs
ElFantasma: Done in 3abe5f5f9 — the invariant now lives on `is_bumped_u256` (the production helper) and is pinned by a test rather than just prose:

```rust
// floor(101 * 110 / 100) = 111 → exactly 111 admitted, 110 not.
assert!(is_bumped_u256(U256::from(101u64), U256::from(111u64), 10));
assert!(!is_bumped_u256(U256::from(101u64), U256::from(110u64), 10));
```

crates/blockchain/mempool.rs
ElFantasma: This was addressed — the blob ↔ non-blob category change now returns a dedicated `MempoolError::ReplacementTypeMismatch` rather than reusing `UnderpricedReplacement`:

```
Attempted to replace a pooled transaction with one of a different category (blob vs. non-blob)
```

Confirmed it's new in this PR (the variant doesn't exist on `main`). Leaving the thread open for you to close.

ElFantasma:
### Review sweep — took ownership, all open threads addressed (3abe5f5f9)

Merged `main` (was 4 behind, clean) and worked through the 14 open threads.

**Real fix — blob category under-counted frame txs.** `find_tx_to_replace` decided blob-ness with `matches!(tx, EIP4844Transaction(_))`, but `FrameTransaction` carries `max_fee_per_blob_gas` + `blob_versioned_hashes` of its own. A blob-carrying frame tx therefore classified as *non-blob*, so a frame↔frame replacement took the 10% bump **and** short-circuited `bumped_blob` to `true` — the blob-fee dimension was never compared at all. That's exactly the cheap sidecar re-propagation the 100% bump exists to prevent.

Fixed with a shared predicate on `Transaction`:

```rust
pub fn is_blob_carrying(&self) -> bool { self.max_fee_per_blob_gas().is_some() }
```

used on both sides of the category gate. Not reachable today — admission rejects blob-carrying frame txs (`FrameTxBlobsUnsupported`) — but it goes live with #7038.

**Threshold arithmetic.** `U256::from(100u64 + bump_percent)` did the addition in `u64` and could wrap for a large `bump_percent`, yielding a nonsensically small multiplier. Now `U256::from(100u64) + U256::from(bump_percent)`, so neither the addition nor the product can wrap into a smaller threshold.

**Test-only mirror removed.** `is_bumped_u64` was `#[cfg(test)]` while production compares exclusively through `is_bumped_u256` — so five tests, including the cited overflow/monotonicity case, were exercising a function that never runs. Deleted it and ported the full set onto `is_bumped_u256`, which previously had only two ordinary bump cases and **no** coverage of its `checked_mul → reject` branch. Side benefit: the doc on `is_bumped_u256` no longer intra-doc-links a `cfg(test)` item, so it resolves in a normal `cargo doc` build.

`is_bumped_u256` now covers: exact bump, just-below, zero-bump, zero-existing, round-down boundary, overflow→reject, and multiplier-wrap.

**Verified:** `cargo fmt --check` clean, `clippy --workspace --all-targets -D warnings` clean, 39 mempool + 1 new `ethrex-common` test passing.

**Scoped out on purpose:** the `is_blob_tx` filter at `mempool.rs:915` has the same narrow `matches!`, but it's a pre-existing payload-builder query filter this PR doesn't touch, and widening it changes block-building selection. Belongs with #7038.

Resolved the bot threads and my own; left @MegaRedHand's open to close.

ElFantasma:
### Fixed a devp2p conformance regression in this PR (70470ae58)

`Hive - Devp2p tests` has been failing on this branch (not from the latest commits — it also failed on 00de88056). `main` passes the same test on the same day, so it was ours. Reproduced locally with Docker + hive, root-caused, fixed, and re-verified.

**Failing test:** `eth/InvalidTxs`, with:

```
failed to send invalid txs: received bad tx: 0x86381df746823d3b6236fa02fa694cd2b8c67c7347dc315e47f00e639bb7325c
```

i.e. the node **propagated a transaction it should have rejected** — the opposite direction from what a "stricter admission" change looks like at first glance.

**Root cause.** This PR replaced the old strict `>` fee comparison with a pure threshold test:

```
new >= floor(existing * (100 + bump) / 100)
```

That threshold is computed with integer division, so it collapses back onto `existing` whenever `existing * bump / 100 < 1` — with the default 10% bump, that is **every fee value below 10 wei**. For a 1 wei tip, `floor(1 * 110 / 100) == 1`, so a transaction with *identical* fees satisfies `new >= threshold` and is admitted as a valid "bump".

`TestInvalidTxs` walks straight into this: it submits several invalid transactions at an already-pooled `(sender, nonce)` with `GasTipCap: 1`, and fails if any of them propagate. One was accepted as a replacement, evicted the pooled tx, and got re-gossiped.

This isn't only a conformance detail — the admitted replacement evicts and re-broadcasts, so a peer could cycle a node's pooled transaction indefinitely at zero fee increase, which is exactly the churn the price bump exists to prevent.

**Fix.** geth applies the rule in two stages; we had only implemented the second. From `legacypool`'s `list.Add`:

```go
if old.GasFeeCapCmp(tx) >= 0 || old.GasTipCapCmp(tx) >= 0 {
    return false, nil          // strict increase required FIRST
}
// ... only then the percentage threshold
```

Added the missing strict-increase guard before the threshold check. It is load-bearing rather than redundant, and the comment in the code says so, so it doesn't get "simplified" away later.

**Verification** — hive `devp2p --sim.limit eth`, same machine, before vs after:

| | result |
|---|---|
| before (`3abe5f5f9`) | `tests=21 failed=2` — `InvalidTxs` `pass=false` |
| after (`70470ae58`) | `tests=21 failed=0` — `InvalidTxs` `pass=true` |

Also added `identical_fees_rejected_even_when_threshold_rounds_down`, covering both the 10%-rounds-down case and a zero-bump configuration. It uses same-fee/different-payload transactions, since byte-identical ones share a hash and get deduplicated before the replacement path.

ElFantasma (approved):

pablodeymo:
Reviewed at `70470ae58`. CI is green including `Hive - Devp2p tests`, so the conformance regression is confirmed fixed in CI and not just locally — nice root-cause writeup on that one; "the node propagated a tx it should have rejected, from a change that only makes admission stricter" is a genuinely counter-intuitive failure and the integer-division explanation is exactly right.

Things I checked that hold up:

- **EIP-2930.** The earlier finding is properly closed by routing everything through `gas_fee_cap()`/`gas_tip_cap()`, which fall back to `gas_price()` when the per-field maxes are `None`. That's geth's `GasFeeCap`/`GasTipCap` translation, and it means 2930 (whose `max_fee_per_gas()` is `None`) is now compared on its real gas price instead of against a 0 threshold. `typed_replacement_against_legacy_pool_tx_uses_translated_dimensions` pins the direction that was actually exploitable.
- **`is_blob_carrying`.** Keys on `max_fee_per_blob_gas()`, which for `FrameTransaction` returns `None` when `blob_versioned_hashes` is empty (`transaction.rs:1557-1563`) — so it really is "has a sidecar", not "has a blob-fee field", and a plain frame tx isn't accidentally pushed into the 100% category. The test asserts both sides.
- **Strict guard on blob txs.** Applying it to `gas_fee_cap`/`gas_tip_cap` but not to the blob fee is fine: at a 100% bump the blob threshold is `2 × existing`, which already implies a strict increase for any nonzero value, and a zero `max_fee_per_blob_gas` can't be admitted in the first place.

## The price bump is the one per-sender gate not re-checked under the insertion lock

`SenderAdmission` / `check_admission` exist precisely because the unlocked checks in `validate_transaction` are a pre-filter — they re-run the queued cap, the gap gate and the cumulative-balance check under the same write lock as the insert, and `check_admission` even re-derives `replaced` live rather than trusting the precomputed flag (`mempool.rs:388-393`, issue #6938). The fee-bump comparison didn't come along: `find_tx_to_replace` runs under a short-lived read lock in `validate_transaction`, and `add_transaction` then removes **whatever tx currently occupies `(sender, nonce)`** (`mempool.rs:665-672` for the non-frame path, `:745-747` for frame) rather than the hash the comparison was made against.

`Mempool::inner` is a `std::sync::RwLock` taken and released per call, so on the multi-threaded runtime two submissions for the same `(sender, nonce)` interleave with no await point needed:

1. Pooled: **A** at tip 100.
2. Attacker submits **B** (tip 110) and **C** (tip 110) concurrently — two RPC connections, or gossip.
3. Both pass `find_tx_to_replace` against **A**; each is a legitimate 10% bump.
4. **B** wins the write lock: removes A, inserts B.
5. **C** takes the write lock: `replaced` is now **B**, so it's exempt from the queued cap and the gap gate, the balance sum excludes B, and the non-frame removal drops B and inserts C.

**C** never out-bid **B**. The survivor only has to beat whatever was pooled when it started, so N racing submissions buy N evictions + re-announcements for a single bump step — the churn multiplier the percentage bump is there to price. It doesn't break the pool's invariants (the slot-based removal keeps exactly one tx per `(sender, nonce)`, which is why this is a weakening rather than a corruption), and each racer still has to pay a valid bump against A, so it's not free. But it's retriable at will and it lands on the same TOCTOU class #6938 closed for the neighbouring gates.

The fix looks mechanical given the existing structure: carry `price_bump_percent` / `blob_price_bump_percent` (or the incoming tx's fee vector) on `SenderAdmission` and re-run the category gate + the two/three comparisons against the live `replaced` inside `check_admission`, which already has it in hand. `find_tx_to_replace` stays as the pre-filter, same split as everything else in that function.

Worth noting the window is much wider for frame txs, where the EVM validation simulation and its storage reads sit between the unlocked check and the insert — and that path deliberately defers to "the locked re-check in `add_transaction` is the authoritative guard" for the paymaster gates, while the fee bump has no locked re-check to defer to.

## Smaller things

- **`bump_percent = 0` doc.** `is_bumped_u256`'s doc says a zero bump "collapses to `new >= existing`", which is true of the helper but no longer of the rule — the strict guard above it means the effective floor is `new > existing`. The CLI help for `--mempool.price-bump` has the same gap: setting it to `0` doesn't disable replacement protection, it falls back to strict-increase. `identical_fees_rejected_even_when_threshold_rounds_down` already locks the behaviour; it's just the wording.
- **L2 fee-token category (pre-existing, genuine question).** The new gate splits blob from non-blob because the two aren't interchangeable. On L2 the same argument seems to apply to `FeeTokenTransaction`: its `max_fee_per_gas` is denominated in `fee_token` and converted by a ratio at execution (`l2_hook.rs:156`), so comparing it numerically against a native-fee tx — or against a *different* `fee_token` — isn't an economic comparison, and a token with a low ratio could out-bid on paper while paying less. This PR plumbs both flags through `init_l2`, so the question is in scope even if the answer is "separate issue".
- `legacy_to_legacy_and_1559_to_legacy_bump_rules` only covers 1559 → legacy; legacy → legacy is covered by `legacy_replacement_requires_10_percent_bump`. Just the name.

edg-l (approved):

PR #6600, opened 2026-05-11
feat(l1): reject contract senders at mempool admission (EIP-3607)

ilitteri:
**Motivation**

Every major Ethereum execution client enforces EIP-3607: transactions whose recovered sender has deployed code are rejected, except where the code is an EIP-7702 delegation designation (the account is still an EOA in spirit, just pointing at delegate code). Without this check, contract accounts can submit invalid transactions that waste pool space and gossip bandwidth.

**Description**

- New helper `is_eip7702_delegation(code: &[u8]) -> bool` plus `EIP7702_DELEGATION_PREFIX = 0xef0100` and `EIP7702_DELEGATED_CODE_LEN = 23` constants in `crates/common/types/account.rs`.
- `Blockchain::validate_transaction` now inspects the sender's `code_hash`; if non-empty, it fetches the code (a single extra store hit only when the sender actually has code) and routes through `is_eip7702_delegation`. Non-delegation contract senders are rejected with the new `MempoolError::SenderIsContract`.
- Six unit tests for the delegation predicate covering: empty, valid 23-byte designation, too-short, too-long, wrong-prefix, and arbitrary contract code.

PR #6603, opened 2026-05-11
feat(l1): cap mempool queued (future-nonce) txs per sender (geth AccountQueue-style)

ilitteri:
**Motivation**

A single sender can currently occupy 100% of the 10,000-tx pool by parking many *future* (nonce-gapped) transactions from one address, trivially bypassing the size cap (#6599), EIP-3607 check (#6600) and percentage-RBF rule (#6601). This adds a per-account cap on the **queued (future-nonce) subpool** — geth's `AccountQueue` model — to bound that blast radius without penalizing legitimate high-throughput senders.

**Description**

- `BlockchainOptions` gains `max_queued_txs_per_account: usize` (default `DEFAULT_MAX_QUEUED_TXS_PER_ACCOUNT = 64`, matching geth's `AccountQueue`).
- Only **queued / future** txs count against the cap: a tx is *future* when its nonce is beyond the sender's contiguous executable run from the on-chain nonce (`is_future` / `queued_count_for_sender`, ranging the existing `txs_by_sender_nonce` index). **Executable (contiguous-nonce) txs are never capped**, so a legitimate high-throughput single sender is unaffected — only nonce-gap parking spam is bounded.
- The cap is enforced **atomically inside `Mempool::add_transaction`, under the same write lock as the insertion**, so two concurrent future-nonce submissions from one sender can't both pass a stale count and race past the limit. `Blockchain::validate_transaction` threads the sender's on-chain nonce out (via `QueuedCap`) so the locked check can distinguish future from executable without a storage read under the lock.
- **Replacements are exempt:** the admission path checks `find_tx_to_replace` first and passes `queued_cap = None` when it matches — a same-`(sender, nonce)` replace reuses the slot and never grows the queue, so it routes through the normal RBF logic even when the sender is at the cap.
- Over-cap submissions are rejected with `MempoolError::MaxQueuedTxsPerAccountExceeded { sender, count, limit }`.
- New CLI flag `--mempool.max-queued-txs-per-account` (env `ETHREX_MEMPOOL_MAX_QUEUED_TXS_PER_ACCOUNT`), plumbed through both L1 and L2 initializers.

**Behavioral change**

Admission becomes stricter only for *future-nonce* txs: a sender with 64 queued (nonce-gapped) txs submitting a 65th future tx is rejected at ingress. Executable txs and replacements remain unaffected.

**Tests**

Unit tests in `mempool.rs` cover the future/queued classification and per-sender isolation (`gapped_tx_is_future_and_counts_only_queued`, `executable_flood_is_never_capped`, `queued_cap_isolates_senders`, and the `queued_count_if_future` cases). The atomic enforcement + replacement-exemption wiring in the admission path is exercised by the existing mempool integration tests.

**Checklist**

- [x] Cap re-scoped to the queued (future-nonce) subpool, enforced atomically in `add_transaction`; replacements exempt.


crates/blockchain/mempool.rs
```diff
@@ -451,6 +451,17 @@ impl Mempool {
...
     }
 
+    /// Returns the number of pending transactions currently held in the
+    /// mempool for `sender`. Used by the per-sender slot cap at admission.
+    pub fn count_for_sender(&self, sender: Address) -> Result<usize, MempoolError> {
+        let inner = self.read()?;
+        let count = inner
+            .txs_by_sender_nonce
+            .range((sender, 0)..=(sender, u64::MAX))
+            .count();
```
MegaRedHand: This is bounded, so I think it's OK for now.

MegaRedHand (approved):
LGTM

crates/blockchain/error.rs
```diff
@@ -83,6 +83,8 @@ pub enum MempoolError {
     TxMaxInitCodeSizeError,
     #[error("Transaction max data size exceeded")]
     TxMaxDataSizeError,
+    #[error("Sender has {count} pending transactions, exceeds the per-account cap of {limit}")]
```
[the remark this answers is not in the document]
ilitteri: 149b829e8

crates/blockchain/error.rs
```diff
@@ -83,6 +83,8 @@ pub enum MempoolError {
...
+    MaxPendingTxsPerAccountExceeded { count: usize, limit: usize },
```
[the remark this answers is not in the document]
ilitteri: 149b829e8

crates/blockchain/blockchain.rs
```diff
@@ -2507,6 +2515,19 @@ impl Blockchain {
...
+        // sender's pool footprint.
+        if tx_to_replace_hash.is_none() {
+            let count = self.mempool.count_for_sender(sender)?;
+            if count >= self.options.max_pending_txs_per_account {
+                return Err(MempoolError::MaxPendingTxsPerAccountExceeded {
+                    count,
+                    limit: self.options.max_pending_txs_per_account,
+                });
+            }
+        }
```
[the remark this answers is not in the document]
ilitteri: 3df6bc84f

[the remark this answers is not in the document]
ilitteri: 3df6bc84f

cmd/ethrex/cli.rs
```diff
@@ -183,6 +183,15 @@ pub struct Options {
...
     pub mempool_max_size: usize,
+    #[arg(
+        help = "Maximum number of pending transactions a single sender may hold in the mempool. Replacements at an existing (sender, nonce) bypass this cap.",
+        long = "mempool.max-pending-txs-per-account",
+        default_value_t = 16,
+        value_name = "MAX_PENDING_TXS_PER_ACCOUNT",
+        help_heading = "Node options",
+        env = "ETHREX_MEMPOOL_MAX_PENDING_TXS_PER_ACCOUNT"
+    )]
+    pub mempool_max_pending_txs_per_account: usize,
```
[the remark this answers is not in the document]
ilitteri: f6de3cf71

cmd/ethrex/cli.rs
```diff
@@ -450,6 +459,7 @@ impl Default for Options {
             dev: Default::default(),
             force: false,
             mempool_max_size: Default::default(),
+            mempool_max_pending_txs_per_account: 16,
             tx_broadcasting_time_interval: Default::default(),
```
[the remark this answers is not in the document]
ilitteri: f6de3cf71

crates/blockchain/mempool.rs
```diff
@@ -142,14 +142,33 @@ impl Mempool {
...
         hash: H256,
         sender: Address,
         transaction: MempoolTransaction,
-    ) -> Result<(), StoreError> {
+        max_pending_txs_per_account: usize,
+    ) -> Result<(), MempoolError> {
         let mut inner = self.write()?;
+        let count = inner
+            .txs_by_sender_nonce
+            .range((sender, 0)..=(sender, u64::MAX))
```
ElFantasma: **Same `range(...)` logic now lives in two places** — here inside `add_transaction` (lines 162–166) and in the new `count_for_sender` at line 476. The latter is `pub` but only referenced from the 5 new test cases below; the production cap check inlines the same `range(...).count()` instead of going through `count_for_sender`.

Two reasonable cleanups:
- **Use it here**: replace lines 162–166 with `let count = inner.count_for_sender_locked(sender);` (or similar private helper that takes `&MempoolInner`) so the lock-free counting logic isn't duplicated.
- **Or drop it**: if there's no consumer outside tests, just inline the `range(...).count()` in the tests too and remove the public method — fewer API-surface points to maintain.

Non-blocking. The current shape works; it's just two copies of the same query.

crates/blockchain/mempool.rs
```diff
@@ -599,3 +629,70 @@ pub fn transaction_intrinsic_gas(
...
+        let b = Address::from_low_u64_be(2);
+        add_tx(&pool, a, 0);
+        add_tx(&pool, a, 1);
+        add_tx(&pool, b, 0);
+        assert_eq!(pool.count_for_sender(a).unwrap(), 2);
+        assert_eq!(pool.count_for_sender(b).unwrap(), 1);
+    }
+
+    #[test]
+    fn count_for_sender_unknown_returns_zero() {
```
ElFantasma: **5 tests cover `count_for_sender` thoroughly, but 0 tests cover the cap's actual *enforcement* in `add_transaction`.** The feature's main behaviour is "return `MaxPendingTxsPerAccountExceeded` when the cap is hit" — that exact path isn't asserted anywhere. All `add_transaction` calls in the test module pass `usize::MAX`, which deliberately disables the cap.

At minimum, two cases worth adding:
```rust
#[test]
fn add_transaction_returns_error_when_cap_reached() {
    let pool = Mempool::new(64);
    let sender = Address::from_low_u64_be(1);
    // Fill exactly to cap with cap=3, then assert 4th rejected.
    for n in 0..3 { /* add */ }
    let err = pool.add_transaction(.., 3).unwrap_err();
    assert!(matches!(err, MempoolError::MaxPendingTxsPerAccountExceeded { count: 3, limit: 3 }));
}

#[test]
fn add_transaction_replacement_bypasses_cap() {
    // Fill to cap. Remove one (simulating the replacement path). Assert
    // the next add at a fresh nonce now succeeds.
}
```

The replacement-bypass test in particular guards the invariant that the comment block in `blockchain.rs:2567` is asserting — that the bypass works because the caller removes first.

Non-blocking but the main feature should not ship without direct test coverage.

crates/blockchain/error.rs
```diff
@@ -83,6 +83,10 @@ pub enum MempoolError {
     TxMaxInitCodeSizeError,
     #[error("Transaction encoded size ({actual} bytes) exceeds the {limit}-byte limit")]
     TxSizeExceeded { actual: usize, limit: usize },
+    #[error(
+        "Sender has {count} pending transactions; adding a new one would exceed the per-account cap of {limit}"
+    )]
+    MaxPendingTxsPerAccountExceeded { count: usize, limit: usize },
```
ElFantasma: nit: when this error fires it's almost always because a specific sender is misbehaving (spam / stuck nonce sequence). Including the sender address in the error makes the operator log directly actionable instead of needing to correlate with submission logs:

```rust
#[error(
    "Sender {sender:?} has {count} pending transactions; adding a new one would exceed the per-account cap of {limit}"
)]
MaxPendingTxsPerAccountExceeded { sender: Address, count: usize, limit: usize },
```

Cheap to thread the address through from the call site at `mempool.rs:167`. Non-blocking.

cmd/ethrex/l2/initializers.rs
```diff
@@ -228,6 +228,7 @@ pub async fn init_l2(
         max_blobs_per_block: None, // L2 doesn't support blob transactions
         precompute_witnesses: opts.node_opts.precompute_witnesses,
         precompile_cache_enabled: true,
+        max_pending_txs_per_account: opts.node_opts.mempool_max_pending_txs_per_account,
```
ElFantasma: L2 inherits the cap straight from `node_opts.mempool_max_pending_txs_per_account`, so L1 and L2 share the same default (16) and the same CLI flag. That's a fine starting point, but worth checking whether L2 actually wants the same cap as L1 — a high-throughput L2 sequencer might want a higher cap (or none), and a sponsored-tx L2 might want a much lower one as anti-abuse.

Not for this PR; just flagging for a future `--l2.mempool.max-pending-txs-per-account` (or equivalent) if the L2 team has a different number in mind. Non-blocking.

ElFantasma (approved):
Four non-blocking inlines — the missing cap-enforcement test is the most worth picking up before merge.

azteca1998 (approved):

ElFantasma:
Rebased onto current `main` (merge) and addressed the review feedback:

- **Missing cap-enforcement tests** (our review) — added `add_transaction_rejects_at_per_sender_cap` (fill to cap → 4th distinct nonce rejected with `MaxPendingTxsPerAccountExceeded`) and `add_transaction_cap_frees_a_slot_after_removal` (the replacement-path bypass: remove one, then admission succeeds again).
- **Sender in the error** (our review) + **`count == limit` wording** (Copilot / greptile) — `MaxPendingTxsPerAccountExceeded` now carries `sender` and the message reads `Sender {sender:#x} has {count} pending transactions (per-account cap {limit}); rejecting new transaction`, which is accurate regardless of `count` vs `limit` and gives operators an actionable address.
- **Duplicated `range(...).count()`** (our review) — extracted a lock-free `MempoolInner::count_for_sender`; both the public `Mempool::count_for_sender` and the in-write-lock cap check now go through it (the cap check can't call the lock-taking public method without deadlocking, hence the inner helper).
- **CLI default `16` should use the shared constant** (Copilot) — already does: `cli.rs` uses `DEFAULT_MAX_PENDING_TXS_PER_ACCOUNT` for both the clap default and `Options::default()`. No change needed.
- **TOCTOU: cap not atomic with insertion** (Copilot / greptile P1) — real, but shared with the other per-sender admission gates (#6606, #6609, EIP-3607). Tracked for a single fix at the validate→insert boundary in #6938 rather than per-PR.
- **L2 cap reuse** (our review) — left as-is; flagged for a future dedicated L2 knob, out of scope here.

Inline `mempool::tests` (incl. the new cap tests) + the `ethrex-test` mempool suite pass; clippy/fmt clean.

edg-l (approved):

iovoid (commented):
To consider: transactions that were obsoleted (account nonce increased) but not pruned count against the limit.

crates/blockchain/blockchain.rs
```diff
@@ -2828,8 +2836,12 @@ impl Blockchain {
         // Add blobs bundle before the transaction so that when add_transaction
         // notifies payload builders the blob data is already available.
         self.mempool.add_blobs_bundle(hash, blobs_bundle)?;
```
iovoid, line 2858: If the transaction add gets rejected, the blob is leaked.

iovoid (changes requested):

ElFantasma:
## ⚠️ Pausing for a design discussion — hard per-account cap diverges from geth and breaks devp2p conformance

While bringing this current with `main`, CI surfaced a deterministic failure in the **devp2p `LargeTxRequest`** Hive test (3/3 runs). I isolated the cause with a CI experiment: temporarily defaulting the cap to `usize::MAX` made `LargeTxRequest` **pass**, and restoring the cap made it fail again — so the per-account cap is definitively the cause (not flaky, not the earlier compile issue, which is fixed).

**Root cause (semantics mismatch with geth):**
- `LargeTxRequest` sends ~2000 transactions concentrated on ~one sender, then retrieves them via `GetPooledTransactions`.
- This PR enforces a **hard reject at 16 pending/account** at admission (for *all* origins, including P2P-gossiped txs), so the node keeps only 16 and the test fails.
- geth also has `AccountSlots = 16`, but it's a **soft fairness-floor**, not a hard cap: a single account may hold many more *executable* txs when there's global pool room. geth therefore retains all ~2000 and passes the same test.

So a hard per-account admission cap:
1. breaks a conformance test that geth passes, and
2. would also reject legitimate high-throughput single-sender activity.

**Proposal:** reconsider the design before merging — options discussed internally are (a) a high coarse anti-DoS ceiling, (b) reworking to geth-style fairness-floor + global-limit eviction, or (c) rethinking whether a hard admission cap is the right model at all. Marking this as draft pending that discussion.

For reference, the merge-to-`main` and the earlier review feedback (cap-enforcement tests, sender-in-error, `count_for_sender` dedup) are already done on the branch; only the cap *semantics* are the open question.

ElFantasma:
# geth per-account tx limits — how they actually work (and why our hard cap breaks conformance)

_Research compiled 2026-07-01, cited against go-ethereum master (`core/txpool/legacypool/`). Context: ethrex #6603 added a hard per-account cap that rejects at admission when a sender has ≥16 pending txs; this deterministically breaks the devp2p `LargeTxRequest` test, which geth passes despite also defaulting `AccountSlots=16`._

## TL;DR
geth's `AccountSlots=16` is **not** a hard admission cap. It's a **soft fairness *floor*** — "the minimum executable slots *guaranteed* per account" — enforced **only during global eviction**, never at admission. A single sender can hold **thousands** of pending txs as long as the pool has global room. That's exactly why geth accepts ~2000 same-sender txs (`LargeTxRequest`) and we don't.

## The config (geth `DefaultConfig`, master)
| Knob | Default | Role |
|---|---|---|
| `PriceLimit` | 1 | min gas price to admit |
| `PriceBump` | 10 | % bump required to replace |
| `AccountSlots` | **16** | **soft floor** — executable slots *guaranteed* per account |
| `GlobalSlots` | **5120** (4096 urgent + 1024 floating, 4:1) | global pending cap |
| `AccountQueue` | **64** | **hard cap on the QUEUED (future/gapped) subpool, per account** |
| `GlobalQueue` | 1024 | global queued cap |

(v1.9.23 had `GlobalSlots=4096`; the +1024 floating capacity + `ErrTxPoolOverflow` came in the ~v1.11 legacypool refactor.)

## Two sub-pools
- **pending** = executable: contiguous nonces from the account's current nonce, applicable to current state.
- **queued** = future: nonce-gapped / not-yet-executable.
Txs migrate pending↔queued as state advances (`promoteExecutables` / demotion).

## The key mechanic — `AccountSlots` is a floor, not a ceiling
- `AccountSlots` is referenced **only in `truncatePending`** — it appears **nowhere** in `add()`, `validateTx()`, `enqueueTx()`, or `promoteExecutables()`.
- `validateTx` builds `UsedAndLeftSlots(addr) → (have, math.MaxInt)`, so the shared `ErrAccountLimitExceeded` (`left <= 0`) **never fires** in legacypool. **There is no per-account executable admission reject.**
- `promoteExecutables` promotes **all** executable txs for an account — no per-account count check. → one account can hold far more than 16 pending.

## When txs are actually dropped
1. **Admission (`add`)** rejects only on:
   - `ErrTxPoolOverflow` ("txpool is full") — the **global** condition `all.Slots()+numSlots(tx) > GlobalSlots+GlobalQueue` (=6144), and **only for remote txs** (locals exempt).
   - price/dup: `ErrUnderpriced`, `ErrReplaceUnderpriced`, `ErrAlreadyKnown`.
   - **No error is ever triggered by "sender already has ≥16 pending."**
2. **Post-admission eviction** (in `runReorg`):
   - `truncatePending`: early-returns `if pending <= GlobalSlots`. Only when total pending > 5120 does it build a **"spammers"** queue of **non-local** accounts whose pending length > `AccountSlots`, and iteratively trims the **biggest offenders, lowest-priced first**, toward a shared threshold (gradual, multi-offender — not draining one account). **Locals exempt.**
   - `truncateQueue`: drops oldest queued txs over `GlobalQueue`. `AccountQueue` (64) hard-caps queued per account via `list.Cap` in `enqueueTx`.

## Why geth passes `LargeTxRequest`
~2000 txs from one sender is **well under** `GlobalSlots=5120`, so nothing evicts and no per-account admission check exists → geth retains all ~2000 and serves them via `GetPooledTransactions`. Our hard reject at the 17th drops ~1984 → the test fails.

## Cross-client: reth does the same
Four subpools (Pending/Queued/BaseFee/Blob), promotion/demotion, capacity managed by **discarding worst txs during maintenance** (truncate non-locals first, spread ~evenly per sender). `max_account_slots` default `TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER = 16`, documented "guaranteed per account, **compatible with geth/erigon**" — a truncation target/floor, **not** an admission cap.

## Prior art: hard per-sender caps are a known anti-pattern
CCS'21 (Perez & Livshits, tx-pool DoS) shows the clients with **hard per-sender caps — Parity (81), Nethermind (17)** — are the ones that force attackers into *account-splitting*, and it explicitly lists geth as **not** having such a cap. A hard per-account admission cap is the outlier design, not the norm.

## Recommendations for ethrex
1. **Don't hard-reject on the pending/executable side.** Model `AccountSlots` as a **fairness floor** + a **global cap** (`GlobalSlots`+`GlobalQueue`) enforced by **price/age eviction**, matching geth/reth.
2. **If we want any hard per-account cap, put it on the QUEUED (nonce-gapped/future) subpool** — geth's only hard per-account bound is `AccountQueue=64` on queued, never pending. And any such bound must sit **well above** the ~2000-tx conformance workload.
3. **Origin matters.** geth exempts **local** txs from spammer eviction and overflow/underpriced discarding; **remote/gossiped** txs are the ones subject to eviction. So gating **remote** txs at admission (what #6603 does) is exactly the wrong lever. (`NoLocals` flips local treatment.) This is where #6608's `TxOrigin` (Local/External) becomes the right primitive.

## Implication for the umbrella
This affects any **hard-admission-reject** gate, not just #6603. **#6604** (min-priority-fee floor) and **#6609** (gapped-nonce reject under pressure) should be re-checked against conformance / geth semantics before merge — hard admission rejects on gossiped txs are the risky pattern.

## Sources
- geth `core/txpool/legacypool/legacypool.go` (master + `830f3c764c`), `list.go`, `core/txpool/validation.go`, `subpool.go`
- geth CLI docs (`--txpool.accountslots`), pkg.go.dev godoc
- reth `reth_transaction_pool` docs, `crates/node/core/src/args/txpool.rs`
- CCS'21 "Revisiting Transaction Mempool DoS" (tristartom.github.io/docs/ccs21.pdf); arXiv 2312.02642

ElFantasma:
# Is geth's soft/eviction mempool model safe against DoS? — cited assessment

_Research compiled 2026-07-02. Sources: CCS'21 (DETER), USENIX Sec'24 (MemPurge), arXiv 2312.02642 (MPFUZZ/ADAMS), arXiv 2407.03543 (saferAd), SIGMETRICS'25 amplification, geth source/releases. Context: ethrex is deciding whether to replace its hard per-account admission cap with geth's soft/eviction model._

## Direct answer
**No — geth's model is not *provably* safe, and mempool DoS is an accepted OPEN research problem.** geth's soft/eviction design (AccountSlots=16 floor, GlobalSlots=5120, AccountQueue=64 hard on *queued*, PriceBump=10%, spammer-first eviction) is the pragmatic standard, but it has a documented, still-active attack surface. Reactive patching of specific patterns has **not** closed the underlying eviction/locking surface.

**But the key finding for us:** a hard per-account cap does **not** make a client *more* DoS-safe — it's bypassable via account-splitting, harms legit users/builders, *and* breaks conformance. Neither model is a clean win; the real DoS levers are **validation rules + pricing**, not slot caps.

## History: DETER → patch → still broken
- **DETER (CCS'21, Li/Wang/Chen):** evicted **100% of geth's mempool at near-zero cost**. DETER-X: high-fee *future* (nonce-gap) txs that evict honest pending txs and are never charged (invalid). DETER-Z: chains of latent-overdraft txs at high fee, exploiting isolated per-tx validation — **as low as 0.021 ETH/block** (3 orders of magnitude below the ~12.5 ETH/block spam baseline).
- **geth patched in v1.11.4 (Mar 2023):** **policy rule 4** = *never evict pending txs for future txs*; **policy rule 2** = *validate a sender's cumulative worst-case cost against balance*. Fixed across geth/Nethermind/Erigon/Besu by Fall 2023.
- **Patched geth still falls:**
  - **MPFUZZ / ADAMS (arXiv 2312.02642):** 24 new asymmetric-DoS bugs across 6 clients (15 confirmed); v1.11.4 still vulnerable (e.g. **XT6**). Includes **mempool *locking*** (occupy the pool to *decline* new txs without evicting) — adversary fees ~**100× lower** than victim fees, **84.6–99.6% success**.
  - **MemPurge (Yaish et al., USENIX Sec'24):** crafted to evade the v1.11.4 fixes — sends the overdraft chain's **lowest-nonce (funding) tx LAST**, so earlier txs are treated as unvalidated "future" txs and only partial checks run → the whole chain is admitted as pending. One address evicts many, paying for ~one slot.
- **Formal gap:** saferAd-CP (best-known defense) raises attacker eviction cost **>10⁴×** (bound ~6.05 ETH vs geth's ~**0.00063 ETH**) at ≤7.3% latency — quantifying how cheap geth's default is to attack.

## Both cap models lose (the crux for ethrex)
- **Hard per-account cap doesn't stop eviction DoS — it forces cheap account-splitting.** DETER+ **ED4** defeats OpenEthereum's per-sender limit by spreading ~8192 high-fee txs across **≥102 accounts**, still evicting 100%.
- **The MemPurge authors explicitly evaluated and *rejected* the hard per-account-slot cap** (§"Limit per-account mempool slots"): (1) harms builders (fewer txs to pack), (2) breaks users needing multiple in-flight txs → forces multiple accounts or costly fee-bumping, (3) only partial (doesn't stop proposer/censorship variants).
- So ethrex's hard cap: **breaks conformance (LargeTxRequest), still bypassable (account-splitting), and hurts legit users** — worst of both worlds.

## What actually provides DoS resistance (load-bearing mitigations)
1. **Rule 4** — never evict *pending* for *future* txs.
2. **Rule 2** — validate a sender's **cumulative worst-case cost vs balance**. (Note: ethrex **#6606's cumulative-balance check is essentially this** — a genuine DoS mitigation, keep it.)
3. **Spammer-first eviction ordering** (`truncatePending` builds a prque of accounts over AccountSlots, trims biggest/lowest-priced first, non-locals first).
4. **PriceBump (10%)** replacement rule — blocks cheap replacement churn.
5. **AccountQueue=64 hard cap on the *queued* (future/gapped) subpool** — the one place a hard per-account cap is safe.
6. **Local-tx exemption / NoLocals** — origin-aware (remote txs are the eviction/DoS targets).

## Separate surface: gossip / CPU-IO DoS (independent of pool caps)
- Per-tx admission cost is **dominated by state-dependent validation (~86%; nonce checks ~82%** via MPT disk traversal) — **not** signature recovery (ecrecover ~12%, ~0.08ms). So processing gossiped txs is a real **state-lookup/IO DoS** regardless of pool limits.
- A malicious/modified node that skips filtering can drive a **3,600× traffic amplification** (SIGMETRICS'25; 2,591 real-world instances / 345 addresses). The gossip layer is weaponizable independent of any node's pool limits → bound per-peer tx flooding (fetcher/announcement limits).

## Community stance
Open problem. Even the academics withdrew an over-claiming defense paper (arXiv:2309.11721) admitting its security model "lacks a proof and verification of the security bounds on a real Ethereum client." Formalization is ongoing.

## Recommendations for ethrex
1. **Adopt geth's soft model** for conformance/compatibility — but treat it as *necessary-not-sufficient*; you inherit geth's residual DoS exposure.
2. **Copy the load-bearing mitigations:** rules 4 & 2, spammer-first + price-based eviction, PriceBump, AccountQueue-on-queued, local/remote origin awareness.
3. **Do NOT rely on a hard per-account admission cap for DoS** — it breaks conformance, is bypassed by account-splitting, and harms builders/users.
4. **Mind the gossip/state-lookup DoS** (nonce-check IO): bound per-peer tx intake; don't do heavy state work on unbounded gossiped floods.
5. If we ever want to *exceed* geth's posture, the frontier is **formal eviction-cost bounds** (saferAd-CP style).

## Sources
- CCS'21 DETER: tristartom.github.io/docs/ccs21.pdf ; DETER+: tristartom.github.io/docs/deter+-ef.pdf
- MemPurge (USENIX Sec'24): usenix.org/system/files/sec24summer-prepub-32-yaish.pdf
- MPFUZZ/ADAMS: arxiv.org/abs/2312.02642 ; saferAd: arxiv.org/abs/2407.03543 ; withdrawn predecessor: arxiv.org/pdf/2309.11721
- Amplification (SIGMETRICS'25): arxiv.org/html/2408.01508v1 ; dl.acm.org/doi/10.1145/3711697
- geth source/defaults: core/txpool/legacypool/legacypool.go ; release v1.11.4

ElFantasma:
## Re-scoped: per-account cap now targets the *queued* (future-nonce) subpool

Following the design discussion above (geth's `AccountSlots` is a soft fairness-floor, not a hard admission cap; a hard per-account *pending* cap breaks devp2p `LargeTxRequest` and diverges from geth/reth), this PR is re-scoped to match geth's **`AccountQueue`** model:

- **Only *queued* (future / nonce-gapped) txs are capped.** A tx is "queued" when its nonce is beyond the sender's contiguous executable run from the on-chain nonce. **Executable (contiguous-nonce) txs are never capped**, so a legitimate high-throughput single sender — and `LargeTxRequest` (~2000 sequential txs from one sender) — is unaffected. Only nonce-gap parking spam is bounded.
- Enforced in `validate_transaction` (which has the sender's on-chain nonce) via a single-lock `Mempool::queued_count_if_future`; gap-aware counting lives in `MempoolInner` (`next_executable_nonce` / `is_future` / `queued_count_for_sender`).
- Renamed for accuracy: `MaxQueuedTxsPerAccountExceeded`, `max_queued_txs_per_account`, `--mempool.max-queued-txs-per-account`, default **64** (geth's `AccountQueue` default).

**Result:** the devp2p **`LargeTxRequest` conformance test now passes** (it deterministically failed with the old hard pending-cap). Workspace clippy clean; inline gap-logic unit tests + the mempool integration suite green.

Note: the TOCTOU between this check and insertion is tracked separately for a unified fix across all per-sender gates in #6938 (not addressed here).

crates/blockchain/blockchain.rs
ElFantasma: Addressed. Two parts:

- The concrete trigger — the per-account cap rejecting the tx *after* the bundle was stored — is gone: the cap moved into `validate_transaction` (which runs before `add_blobs_bundle`), so a cap rejection now returns before any bundle is inserted.
- The residual (`add_transaction` itself erroring — lock poisoning / eviction — after the bundle is in) is handled in b5bca41f9: on `add_transaction` `Err` we roll back the orphaned bundle via a new `remove_blobs_bundle` (clears both `blobs_bundle_pool` and the `blobs_bundle_by_versioned_hash` index) and return the original error. The bundle-before-tx ordering (needed so payload builders see the blob data on notify) is preserved.

ElFantasma:
@iovoid thanks for the review — both points are addressed on the current head:

1. **Blob leak (inline thread):** the cap re-scope moved the check upstream of `add_blobs_bundle`, and b5bca41f9 rolls back the orphaned bundle if `add_transaction` errors afterward. Details in the thread.
2. **Obsoleted txs counting against the limit:** resolved by the re-scope to the queued (future-nonce) subpool. `queued_count_for_sender` now counts only txs at/above `next_executable_nonce(sender, account_nonce)`, so obsoleted txs (nonce below the on-chain account nonce) fall outside the range and no longer count toward the per-account cap.

Ready for another look whenever you have a moment.

crates/blockchain/blockchain.rs
```diff
@@ -3112,6 +3153,14 @@ impl Blockchain {
             return Err(MempoolError::InvalidChainId(config.chain_id));
         }
 
+        // The per-account pending-tx cap is enforced atomically inside
```
iovoid: We release the lock after checking, so it's not truly atomic.

crates/blockchain/blockchain.rs
```diff
@@ -3050,6 +3072,25 @@ impl Blockchain {
...
+            // on-chain nonce + its contiguous pooled run) count against the cap.
+            // Executable/contiguous txs are never capped, so a legitimate
+            // high-throughput single sender (and the devp2p `LargeTxRequest`
+            // conformance case) is unaffected; only nonce-gap parking spam is
+            // bounded. Replacements (same `(sender, nonce)`) are handled below.
+            if let Some(queued) =
+                self.mempool
+                    .queued_count_if_future(sender, sender_acc_info.nonce, nonce)?
+                && queued >= self.options.max_queued_txs_per_account
+            {
```
iovoid: This prevents RBF of queued transactions, we should probably check find_tx_to_replace first.

iovoid (commented):
The PR description mentions the old design.

crates/blockchain/blockchain.rs
ElFantasma: Fixed in 9d936b5f5. The queued-cap check moved out of `validate_transaction` (where it read under a released read lock) and **into `Mempool::add_transaction`, under the same write lock as the insert**, so the check and insert are atomic — two concurrent future-nonce txs from one sender can no longer both pass a stale count. `validate_transaction` now just threads the sender's on-chain nonce out via a `QueuedCap` so the locked check can tell future from executable without an async storage read under the lock.

crates/blockchain/blockchain.rs
ElFantasma: Fixed in 9d936b5f5. The admission callers now check `find_tx_to_replace` first and pass `queued_cap = None` when it matched, so a replacement — which reuses the same `(sender, nonce)` slot and never grows the queue — is exempt from the cap on all paths (including frame txs, whose predecessor is removed inside `add_transaction` after the check).

ElFantasma:
@iovoid thanks — all three addressed:

- **Not atomic:** the queued-cap check moved into `Mempool::add_transaction` under the insertion write lock (9d936b5f5), so check+insert are atomic; `validate_transaction` only threads the sender's on-chain nonce out via `QueuedCap` (no storage read under the lock). [thread](https://github.com/lambdaclass/ethrex/pull/6603#discussion_r3520692326)
- **RBF of queued txs:** the admission path checks `find_tx_to_replace` first and passes `queued_cap = None` for replacements, so a same-`(sender, nonce)` fee-bump is exempt on all paths. [thread](https://github.com/lambdaclass/ethrex/pull/6603#discussion_r3520705019)
- **Stale description:** rewritten to the current queued-subpool design (per-account *queued/future-nonce* cap, default 64, atomic enforcement, replacement-exempt).

Also merged latest `main` (frame-tx work integrated). 96 integration + 6 mempool unit tests pass, clippy clean.

crates/blockchain/mempool.rs
```diff
@@ -438,17 +484,43 @@ impl Mempool {
     /// Add transaction to the pool without doing validity checks, except for the
     /// one-pending-frame-tx-per-sender policy which must run under this lock to
     /// avoid a TOCTOU race (EIP-8141, review fix 1.6).
+    ///
+    /// The per-account **queued** (future/nonce-gapped) cap is enforced upstream
+    /// in `Blockchain::validate_transaction`, which has the sender's on-chain
```
iovoid: False, it's enforced here.

crates/blockchain/mempool.rs
```diff
@@ -1146,3 +1244,106 @@ pub fn transaction_intrinsic_gas(
     };
     Ok(intrinsic.max(calldata_floor))
 }
+
+#[cfg(test)]
+mod tests {
```
iovoid, line 1251: There are no tests for the rejection case.

iovoid (changes requested):

crates/blockchain/mempool.rs
ElFantasma: You're right — that doc was stale from before the cap moved into `add_transaction`. Fixed in fbc476f52: the doc now says the queued cap is enforced *here* (atomically under the write lock, via the passed `QueuedCap`), not upstream in `validate_transaction`.

crates/blockchain/mempool.rs
ElFantasma: Added in fbc476f52: `add_transaction_rejects_future_tx_over_queued_cap` (sender at cap → `MaxQueuedTxsPerAccountExceeded` with count/limit asserted), plus `add_transaction_accepts_future_tx_below_queued_cap` and `add_transaction_never_caps_executable_txs` to bracket the boundary and confirm executable txs are never capped.

iovoid (approved):

PR #6605, opened 2026-05-12
docs(l1): add mempool architecture documentation

ilitteri:
**Motivation**

There is no centralized, up-to-date description of ethrex's mempool in the L1 architecture docs. The mempool currently appears only as a one-paragraph mention in `overview.md`, with no description of its data structures, admission validation, RBF semantics, eviction policy, or P2P propagation paths. With six mempool-hardening PRs in flight (#6576, #6599, #6600, #6601, #6603, #6604), the surface area changing inside the pool warrants its own document — one that can be updated in place as each PR merges.

**Description**

New `docs/l1/architecture/mempool.md` covering:

- Role of the mempool in the node (RPC ingress, P2P ingress, block-building source, block-removal source) with an ASCII data-flow diagram.
- `Mempool` / `MempoolInner` fields with file:line citations: `transaction_pool`, `broadcast_pool`, `private_pool`, `txs_by_sender_nonce`, `txs_order`, `blobs_bundle_pool`, `in_flight_txs`, plus the supporting indices.
- Full admission-validation chain in `Blockchain::validate_transaction`, every check ordered, with the corresponding `MempoolError` variant. The six in-flight PRs are cited inline (`via #6599`, `via #6603`, etc.) so the citation can stay or be replaced once each PR merges.
- RBF semantics: type-change rejection, percentage bump (10% non-blob, 100% blob, per fee field), the configurable thresholds.
- Insertion path under the write lock (per-sender cap atomic with insertion, FIFO order queue management, broadcast vs private routing).
- FIFO eviction and the tombstoned `txs_order` queue.
- P2P propagation paths: periodic `tx_broadcaster`, new-peer pooled-hash dump (`get_txs_for_new_peer_dump`), `GetPooledTransactions` handler (`get_p2p_transaction_by_hash`).
- Private-mempool flag (`--mempool.private`): what is and isn't gated.
- Operator-facing CLI table for every `--mempool.*` flag with defaults and env vars.
- `MempoolError` taxonomy with one-line descriptions, in-flight-PR additions marked.
- Known limitations / open work as a short list, framed as gaps not future-tier commitments.

Also adds an entry for the new page in `docs/l1/architecture/README.md`.

**Deliberately omitted**

- No description of future tier work (sub-pool partitioning, sweep tasks, cumulative-balance check, tip-aware eviction, local-origin first-class marker beyond what #6576 introduces) beyond a short "open work" list. The doc is anchored to what is on `main` plus what is sitting in the six in-flight PRs — nothing further is documented as decided.

**Maintenance**

When any of the six in-flight PRs merges, the `(via #NNNN)` citations in the corresponding section can be left as-is for one release cycle and then dropped, or replaced with `as of vN.N.N`. The doc is structured so each PR's behavior lives in one section, minimizing the merge-time edits.


ElFantasma (approved):
Looks good as-is. A few staleness notes, all expected for a baseline doc that openly says "It will be updated as those PRs merge" (line 3):

Doc will need updates as the mempool stack lands. The doc tracks 6 in-flight PRs (https://github.com/lambdaclass/ethrex/pull/6576, https://github.com/lambdaclass/ethrex/pull/6599, https://github.com/lambdaclass/ethrex/pull/6600, https://github.com/lambdaclass/ethrex/pull/6601, https://github.com/lambdaclass/ethrex/pull/6603, https://github.com/lambdaclass/ethrex/pull/6604), but the stack has grown since: https://github.com/lambdaclass/ethrex/pull/6606 (cumulative balance), https://github.com/lambdaclass/ethrex/pull/6609 (gapped-nonce gate), https://github.com/lambdaclass/ethrex/pull/6610 (sweep task), https://github.com/lambdaclass/ethrex/pull/6624 (blob-shrink reject), https://github.com/lambdaclass/ethrex/pull/6625 (punish-spammer), https://github.com/lambdaclass/ethrex/pull/6630 (7702 cap), https://github.com/lambdaclass/ethrex/pull/6631 (reorg re-injection) are all in flight or merged. The "Known Limitations" entries that explicitly call out gaps (cumulative balance, sweep tasks, per-sender accounting) will be the first to need updating as those PRs merge.

Cumulative-balance section (line 120) flips polarity once https://github.com/lambdaclass/ethrex/pull/6606 lands. It currently reads "Not present on main" — that'll become a positive statement of how it works.

"Six in-flight" count in the lede (line 3) will quickly go stale. Could rephrase as something like "the in-flight mempool-hardening PRs" without a number, or scope it explicitly to the merge cohort represented here.

None of these are blocking — they're inherent to documenting a moving target, and the doc explicitly invites itself to be updated.

azteca1998 (approved):

ElFantasma:
Merged latest `main` and rewrote the doc to describe **current `main` only**, rather than `main` + the in-flight PRs. That was the root of nearly all the review feedback — the page documented code that hadn't merged (and, for #6603, code that changed shape before merging). Every comment below is addressed by the rewrite:

**Staleness / documents-unmerged-code (Copilot, Codex, our own approval notes):**
- `private_pool` field, `--mempool.private`, `add_transaction_no_broadcast`, `is_private`, non-broadcast RPC ingress (#6576) — removed. `--mempool.price-bump`/`--mempool.blob-price-bump` + `ReplacementTypeMismatch` (#6601), `--mempool.min-tip` + `TipBelowMinimum` (#6604) — removed. The CLI table now lists only the three flags that exist on `main` (`maxsize`, `max-queued-txs-per-account`, `gap-admit-occupancy-threshold`).
- Per-sender cap: rewritten from the never-merged 16-slot *pending* cap to the actual **queued (future-nonce) cap, default 64** (`MaxQueuedTxsPerAccountExceeded`) — the #6603 re-scope.
- "Cumulative-balance not present on main" → now documented as live (#6606); gap-admission (#6609) and L2-only rejection added; the EIP-8141 frame/paymaster machinery documented.
- `MempoolTransaction` path corrected to `crates/common/types/transaction.rs`.

**Correctness (Greptile, Claude):**
- `blobs_bundle_by_versioned_hash` now described as the nested `versioned_hash → { tx_hash → position }` (a blob may be referenced by multiple txs).
- The `TxMaxDataSizeError` / `MAX_TRANSACTION_DATA_SIZE` check Greptile flagged as missing from the flowchart is now **gone from `main`** (superseded by the `TxSizeExceeded` wire-size cap), so it's removed from the doc entirely rather than added.

**Clarity (Kimi):** the EIP-7702 delegation designation is now described as the 3-byte `0xef0100` prefix + 20-byte delegate address = 23 bytes.

The documentation for the still-in-flight features (#6576/#6601/#6604) isn't lost — it's set aside to move into each of those PRs when they land, so each PR carries its own doc delta (docs-with-code) instead of this page tracking unmerged work. I also dropped most `file:line` pins from prose, since those rot on any refactor; the code remains the source of truth for exact ordering/constants.

docs/l1/architecture/mempool.md
```diff
@@ -0,0 +1,273 @@
...
+**Gapped-nonce rejection under pressure.** When the pool is heavily occupied, a new *non-replacement* transaction whose nonce is not the sender's on-chain nonce is rejected with `GapAdmissionDeniedUnderPressure { occupancy_pct, nonce_gap }`. The gate fires only when `occupancy_pct ≥ gap_admit_occupancy_threshold` (default 90; a threshold of 100 disables it), so nonce-gap parking spam is bounded without affecting normal load.
+
+**Per-sender queued cap.** Enforced inside `add_transaction` (see [Insertion path](#insertion-path)), not here, so it re-checks the live pool atomically with the insert.
+
+## Replacement by Fee (RBF)
+
+`Mempool::find_tx_to_replace` decides whether a new transaction at an existing `(sender, nonce)` is accepted as a replacement. A replacement is accepted only when it strictly increases every applicable fee field versus the in-pool transaction:
+
+- Legacy: `gas_price` strictly greater.
+- EIP-1559 / EIP-2930 / EIP-7702 / fee-token: both `max_fee_per_gas` and `max_priority_fee_per_gas` strictly greater.
```
MegaRedHand: This groups EIP-2930 with EIP-1559/7702/fee-token ("both max_fee_per_gas and max_priority_fee_per_gas strictly greater"). But EIP-2930's max_fee_per_gas() / max_priority_fee() return None (transaction.rs), so in find_tx_to_replace (crates/blockchain/mempool.rs) the eip1559 arm is always false and the replacement is actually governed by the legacy gas_price comparison. EIP-2930 belongs in the "Legacy: gas_price strictly greater" bucket (line 143).

docs/l1/architecture/mempool.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Frame transactions (EIP-8141).** Frame txs go through an extended set of gates — fork activation (`FrameTxPreFork`), expiry (`FrameTxExpired`), static structural validity (`InvalidFrameTransaction`), a no-blobs rule (`FrameTxBlobsUnsupported`), a signature-verification gas budget (`FrameTxVerifyGasExceeded` / `FrameTxVerifyGasBudgetExceeded`), signature authenticity and low-`s` malleability (`InvalidFrameSignature` / `FrameTxMalleableSignature`), and validation-prefix shape (`FrameTxUnrecognizedPrefix` / `FrameTxInvalidPrefixStructure`). The validation-prefix simulation and paymaster accounting run last (see [Frame transactions](#frame-transactions-eip-8141)).
+
+**Per-tx wire-size cap.** Non-blob transactions are bounded by `MAX_TX_SIZE = 128 KiB` against `Transaction::encode_canonical_len()`. Blob transactions are bounded by `MAX_BLOB_TX_SIZE` in `add_blob_transaction_to_pool` against the wire wrapper, since ethrex stores the core transaction and the sidecar in separate structs.
+
+**Init code size.** Active from Shanghai. Limit is `MAX_INITCODE_SIZE = 48 KiB` (`2 × MAX_CODE_SIZE`); from Amsterdam onward (EIP-7954) it becomes `AMSTERDAM_MAX_INITCODE_SIZE = 64 KiB`.
```
MegaRedHand: > [EIP-7954: Increase Maximum Contract Size](https://eips.ethereum.org/EIPS/eip-7954)
> 
> Raise the maximum contract code size from 24KiB to 64KiB and initcode size from 48KiB to 128KiB.

```suggestion
**Init code size.** Active from Shanghai. Limit is `MAX_INITCODE_SIZE = 48 KiB` (`2 × MAX_CODE_SIZE`); from Amsterdam onward (EIP-7954) it becomes `AMSTERDAM_MAX_INITCODE_SIZE = 128 KiB`.
```

MegaRedHand (approved):
LGTM. Left some minor comments

docs/l1/architecture/mempool.md
ElFantasma: Good catch — fixed in d6d11ddaa.

Confirmed in the code: `max_fee_per_gas()` and `max_priority_fee()` both return `None` for `EIP2930Transaction`, so with `unwrap_or_default()` the `eip1559_higher_fees` arm can never be true and replacement is governed entirely by the `gas_price` comparison (`gas_price()` returns the `gas_price` field for EIP-2930).

While fixing it I found the section was inaccurate in a second way: the actual condition is

```
blob_fee_ok && (both_1559_fields_higher || gas_price_higher)
```

The two fee arms are OR-ed, and `gas_price()` returns `max_fee_per_gas` for every typed-fee tx — so raising *only* `max_fee_per_gas` on an EIP-1559 tx is already sufficient, meaning "both `max_fee_per_gas` and `max_priority_fee_per_gas` strictly greater" was wrong too. Rewrote the section around the real condition rather than just moving EIP-2930 to the legacy bullet.

docs/l1/architecture/mempool.md
ElFantasma: Applied in d6d11ddaa. The previous text conflated code size with initcode size — spelled out the derivation so it can't drift again:

`AMSTERDAM_MAX_CODE_SIZE = 0x10000` (64 KiB) → `AMSTERDAM_MAX_INITCODE_SIZE = 2 × AMSTERDAM_MAX_CODE_SIZE` = **128 KiB** (`crates/blockchain/constants.rs:29,36`).

ElFantasma:
### Review-comment sweep

Took ownership of this PR and worked through the open threads.

**@MegaRedHand's two comments — both valid, fixed in d6d11ddaa** (replies inline):
- EIP-2930 was grouped with the EIP-1559 replacement rule; it's governed by the `gas_price` comparison. While fixing it, the "both fields strictly greater" framing turned out to be wrong too — the two fee arms are OR-ed — so the section was rewritten around the actual condition.
- Amsterdam initcode limit was 64 KiB; it's **128 KiB** (`2 × AMSTERDAM_MAX_CODE_SIZE`). The text had conflated code size with initcode size.

**The nine bot threads (May 12) — resolved as stale.** Each was re-verified against the code rather than assumed; all are obsolete either because the doc was rewritten since or because the PR they referenced has landed:

| Comment | Why it's stale |
|---|---|
| `blobs_bundle_by_versioned_hash` 1-to-1 wording | Already reworded to `versioned_hash → { tx_hash → position }` |
| Flowchart omits `TxMaxDataSizeError` | Check removed from `main` by #6599 |
| `private_pool` field doesn't exist | No longer in the table |
| `encode_canonical_len()` doesn't exist | Now exists (`transaction.rs:2741`); `BlobsBundle.length()` no longer referenced |
| P2P `get_txs_for_new_peer_dump` / `is_private` | Doc now cites `get_all_txs_by_sender`, which exists |
| CLI flags don't exist on `main` | All four now exist (#6603, #6609 merged) |
| "Defaults pinned to named constants" | Sentence removed |
| Per-sender cap not enforced in `add_transaction` | True as of #6603 |
| `MempoolTransaction` in `types/mempool.rs` | Doc now points at `transaction.rs` (`mod mempool`) |

Also spot-checked the remaining cited constants (`MAX_TX_SIZE` 128 KiB, `EIP7702_DELEGATED_CODE_LEN` 23, defaults 64 / 90, `MIN_BASE_FEE_PER_BLOB_GAS` 1, prune threshold `max + max/2`) and confirmed the six PRs listed under *Known Limitations* (#6576, #6601, #6604, #6607, #6608, #6610) are all still open.

ElFantasma:
### This should merge *first*, not last

Flagging a change to the plan: the umbrella previously had this PR as step 9, "docs last so it reflects the merged state". We're switching to the opposite model — **each PR carries the documentation for the functionality it introduces**, in the same PR, rather than batching doc updates into a trailing docs PR.

That inverts the ordering. The page has to exist before a feature PR can amend it, so this one goes first and the remaining mempool PRs each add their own section afterwards. The rule is already stated in the page's own header ("...the matching section of this page, **in the same PR**"), which only works once the page has landed.

Checked that this PR is consistent with that model — it documents `main` and nothing else:

- The CLI table lists only the four flags that exist on `main` today (`--mempool.maxsize`, `--mempool.max-queued-txs-per-account`, `--mempool.gap-admit-occupancy-threshold`, `--p2p.tx-broadcasting-interval`).
- `--mempool.private` (#6576), the min-tip floor (#6604), `TxOrigin` (#6608) and the RBF percentage bump (#6601) appear **only** under *Known Limitations / Open Work*, flagged with their PR numbers — not in the CLI table, admission flowchart or error taxonomy.
- The admission flowchart and error taxonomy cover merged behavior only (#6603, #6606, #6609, #6935 have all landed).

Umbrella #6623 updated with the new order and the policy written out.

pablodeymo (approved):

PR #6627, opened 2026-05-12
feat(l1,l2): harden HTTP JSON-RPC defaults

edg-l:
## Summary

- Bind `--http.addr` to `127.0.0.1` by default instead of `0.0.0.0`.
- Add `--http.api` flag (default `eth,net,web3`) that gates HTTP/WS JSON-RPC dispatch on a namespace allowlist. Disabled namespaces return `MethodNotFound`.
- Add `--http.api.ethrex` (L2 only, default `true`) for the L2-specific `ethrex_*` namespace.
- Preserve current devnet behavior by passing `--http.addr=0.0.0.0 --http.api=eth,net,web3,debug,admin,txpool` to the ethrex participant in `fixtures/networks/default.yaml`.
- Update in-tree tooling (sync compose/Makefile, L2 dev compose) to opt into the namespaces they actually use.
- Update in-tree docs and CLI reference to reflect the new defaults.

The two new defaults are independent. `--http.addr` controls who can reach the RPC port (loopback vs. remote). `--http.api` controls which methods are served on the port. To call `debug_*` from the same machine, pass `--http.api ...,debug`; you do not need `--http.addr 0.0.0.0` for that.

> **BREAKING:** the HTTP JSON-RPC now binds to `127.0.0.1` and only serves `eth,net,web3` by default. Operators that want remote callers must pass `--http.addr 0.0.0.0`. Operators that want `admin_*`, `debug_*`, or `txpool_*` (locally or remotely) must pass `--http.api` to enable those namespaces.

## Why

Before this change, a fresh `ethrex` install:

1. Bound the HTTP JSON-RPC to `0.0.0.0`, so any host that could route to the machine could reach port 8545.
2. Served every JSON-RPC namespace on that endpoint, including `admin_*`, `debug_*`, and `txpool_*`, with **no authentication**. `handle_http_request` (`crates/networking/rpc/rpc.rs:647`) takes no auth header and never calls `authenticate()`; only the engine-RPC handler at `:673` does JWT.

So before this PR: anyone who could reach port 8545 could call:

- `admin_addPeer(enode)` — inject a peer of their choice into your peer set.
- `admin_removePeer(enode)` — kick a peer you depend on.
- `admin_setLogLevel("trace")` — flip your node to trace logging, fill the disk, slow it down.
- `admin_nodeInfo` / `admin_peers` / `admin_peerScores` — full enumeration of your topology.
- `debug_traceTransaction` / `debug_traceBlockByNumber` — re-execute arbitrary blocks under tracing; trivial CPU-pinning DoS.
- `txpool_content` / `txpool_inspect` — read every pending tx in your local mempool in real time.

No JWT, no API key, no source check, nothing. Just `POST /` and you're in. And with the old `--http.addr` defaulting to `0.0.0.0`, a fresh install on any cloud VM without a host firewall exposed all of that to the public internet on port 8545.

The other major EL clients (geth, reth, nethermind, besu) all bind the HTTP RPC to `127.0.0.1` by default and only expose `eth,net,web3` (or nothing at all); ethrex was an outlier.

After this change:

- The HTTP RPC is loopback-only out of the box, so a fresh install behind no firewall is not exposed.
- Even when an operator opts in to `--http.addr 0.0.0.0` (reverse proxy, docker, kurtosis), only `eth,net,web3` are reachable unless they also pass `--http.api` to enable additional namespaces. They have to make a deliberate decision to expose `admin`/`debug`/`txpool`.
- The two new defaults close two different doors: loopback prevents remote reach even if the operator enables admin/debug/txpool, and the allowlist prevents unsafe namespaces from being served even if the operator binds `0.0.0.0`.
- The `engine` namespace is doubly protected: the CLI parser rejects `--http.api engine`, and `map_http_requests` rejects it even if the set is constructed in code. It is only served on the authenticated RPC port (which already defaulted to `127.0.0.1`).

Requiring JWT (or some other auth) for `admin_*` even when it's exposed is the proper long-term fix — that's how geth's `--authrpc.api admin,debug` pattern works. Out of scope here; worth a follow-up.

## Implementation notes

- `RpcApiContext` now carries an `Arc<HashSet<RpcNamespace>>` allowlist. `map_http_requests` returns `MethodNotFound` for namespaces not in the set, matching the behavior operators get from other EL clients.
- L2's `RpcApiContext` reuses the L1 allowlist (so `debug`/`admin`/`txpool` gating is shared) and adds a separate `ethrex_namespace_allowed: bool` for L2-specific methods.
- The WebSocket handler routes through the same `map_http_requests`, so WS is gated transparently.

## Tests

- `crates/networking/rpc/rpc.rs::tests::http_api_allowlist_blocks_debug_namespace_by_default` — `debug_traceTransaction` with the default allowlist returns `MethodNotFound` and never reaches the handler.
- `crates/networking/rpc/rpc.rs::tests::http_api_allowlist_default_routes_standard_namespaces` — default allowlist still routes `eth_*`, `net_*`, `web3_*`.
- `crates/networking/rpc/rpc.rs::tests::engine_namespace_rejected_on_http` — even if `Engine` is somehow placed in the set, HTTP dispatch refuses it.
- `crates/l2/networking/rpc/rpc.rs::tests::ethrex_namespace_blocked_when_disabled` — `ethrex_*` returns `MethodNotFound` when `--http.api.ethrex` is false.
- `cmd/ethrex/cli.rs::tests` — 5 tests covering default `http_addr` is `127.0.0.1`, default `http_api` is `[Eth, Net, Web3]`, parsing comma-separated values, and that `engine` and unknown namespaces are rejected at parse time.

`cargo fmt`, `cargo clippy --workspace --all-targets`, `make lint-l1`, `make lint-l2` all clean.

## Hive will break without a paired upstream PR

The hive client launcher (`ethereum/hive` `clients/ethrex/ethrex.sh`) lives in the upstream hive repo, not in this tree. Once this PR merges, the launcher will keep working for the engine/eth namespaces but `rpc-compat` suites that hit `debug_*` / `admin_*` / `txpool_*` will start returning `MethodNotFound`, and the corresponding hive tests will turn red in CI.

Plan: open a PR against `ethereum/hive` that feature-detects the flag so the same launcher works against both old and new ethrex binaries:

```bash
# Configure RPC. Probe the binary for --http.api support so this script
# works against both old (no allowlist) and new (default allowlist of
# eth,net,web3) ethrex builds.
HTTP_API_FLAG=""
if "$ethrex" --help 2>&1 | grep -q -- '--http.api'; then
    HTTP_API_FLAG="--http.api=eth,net,web3,debug,admin,txpool"
fi
FLAGS="$FLAGS --http.addr=0.0.0.0 --authrpc.addr=0.0.0.0 $HTTP_API_FLAG"
```

The grep-on-help approach avoids hard-coding which ethrex versions support the flag and works regardless of build order across the two repos. The upstream PR can ship before, with, or after this one; whichever lands second won't break hive.

## Test plan

- [ ] CI: `cargo test`, `make lint-l1`, `make lint-l2`
- [ ] Devnet smoke: `kurtosis run --enclave ethrex github.com/lambdaclass/ethrex-package --args-file fixtures/networks/default.yaml` and verify ethrex still answers `debug_traceTransaction` from the kurtosis network
- [ ] Sync tooling: bring up `tooling/sync/docker-compose.yml` and run `peer_top.py` to confirm `admin_*` still works after the new `--http.api` flag
- [ ] Hive: once the paired upstream PR lands, rerun the engine/rpc-compat suites

### DIFF
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -74,7 +74,7 @@ use serde::Deserialize;
 use serde_json::Value;
 use spawned_concurrency::tasks::ActorRef;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     future::IntoFuture,
     net::SocketAddr,
     sync::{Arc, Mutex},
@@ -213,6 +213,12 @@ pub struct RpcApiContext {
     pub block_worker_channel: UnboundedSender<BlockWorkerMessage>,
     /// WebSocket configuration. `None` when the WS server is disabled.
     pub ws: Option<WebSocketConfig>,
+    /// Set of RPC namespaces that are allowed over the public HTTP/WS endpoints.
+    ///
+    /// Methods belonging to namespaces not in this set return `MethodNotFound`.
+    /// The `engine` namespace is always served via the authenticated RPC port
+    /// and is not gated here.
+    pub allowed_namespaces: Arc<HashSet<RpcNamespace>>,
 }
 
 /// Configuration for the WebSocket RPC server.
@@ -504,6 +510,7 @@ pub async fn start_api(
     log_filter_handler: Option<reload::Handle<EnvFilter, Registry>>,
     gas_ceil: u64,
     extra_data: String,
+    allowed_namespaces: HashSet<RpcNamespace>,
 ) -> Result<(), RpcErr> {
     // TODO: Refactor how filters are handled,
     // filters are used by the filters endpoints (eth_newFilter, eth_getFilterChanges, ...etc)
@@ -527,6 +534,7 @@ pub async fn start_api(
         gas_ceil,
         block_worker_channel,
         ws: ws.clone(),
+        allowed_namespaces: Arc::new(allowed_namespaces),
     };
 
     // Periodically clean up the active filters for the filters endpoints.
@@ -939,17 +947,24 @@ pub async fn handle_eth_unsubscribe(
 
 /// Handle requests that can come from either clients or other users
 pub async fn map_http_requests(req: &RpcRequest, context: RpcApiContext) -> Result<Value, RpcErr> {
-    match req.namespace() {
-        Ok(RpcNamespace::Eth) => map_eth_requests(req, context).await,
-        Ok(RpcNamespace::Admin) => map_admin_requests(req, context).await,
-        Ok(RpcNamespace::Debug) => map_debug_requests(req, context).await,
-        Ok(RpcNamespace::Web3) => map_web3_requests(req, context),
-        Ok(RpcNamespace::Net) => map_net_requests(req, context).await,
-        Ok(RpcNamespace::Mempool) => map_mempool_requests(req, context),
-        Ok(RpcNamespace::Engine) => Err(RpcErr::Internal(
-            "Engine namespace not allowed in map_http_requests".to_owned(),
-        )),
-        Err(rpc_err) => Err(rpc_err),
+    let namespace = match req.namespace() {
+        Ok(ns) => ns,
+        Err(rpc_err) => return Err(rpc_err),
+    };
+    // Engine is served on the authenticated port only; treat as not found here.
+    if matches!(namespace, RpcNamespace::Engine) || !context.allowed_namespaces.contains(&namespace)
+    {
+        return Err(RpcErr::MethodNotFound(req.method.clone()));
+    }
+    match namespace {
+        RpcNamespace::Eth => map_eth_requests(req, context).await,
+        RpcNamespace::Admin => map_admin_requests(req, context).await,
+        RpcNamespace::Debug => map_debug_requests(req, context).await,
+        RpcNamespace::Web3 => map_web3_requests(req, context),
+        RpcNamespace::Net => map_net_requests(req, context).await,
+        RpcNamespace::Mempool => map_mempool_requests(req, context),
+        // Engine is rejected by the guard above.
+        RpcNamespace::Engine => unreachable!(),
     }
 }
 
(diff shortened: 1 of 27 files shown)


azteca1998 (approved):

ilitteri (approved):

crates/networking/rpc/rpc.rs
```diff
@@ -821,12 +829,20 @@ where
...
 {
     match req.method.as_str() {
-        "eth_subscribe" => {
-            let result = handle_eth_subscribe(&req, context, out_tx, subscription_ids).await;
-            rpc_response(req.id, result).ok()
-        }
-        "eth_unsubscribe" => {
-            let result = handle_eth_unsubscribe(&req, context, subscription_ids).await;
+        "eth_subscribe" | "eth_unsubscribe" => {
+            // Subscriptions are part of the `eth` namespace and must obey the
```
iovoid, line 833: Do we need to gate eth_unsubscribe?

crates/networking/rpc/utils.rs
```diff
@@ -235,6 +236,35 @@ pub enum RpcNamespace {
     Mempool,
 }
 
+impl RpcNamespace {
+    /// Returns the lowercase prefix used in method names (e.g., "eth", "txpool").
+    pub fn as_prefix(&self) -> &'static str {
```
iovoid: Never used

crates/networking/rpc/utils.rs
```diff
@@ -235,6 +236,35 @@ pub enum RpcNamespace {
...
+            RpcNamespace::Admin => "admin",
+            RpcNamespace::Debug => "debug",
+            RpcNamespace::Web3 => "web3",
+            RpcNamespace::Net => "net",
+            RpcNamespace::Mempool => "txpool",
+        }
+    }
+
+    /// Parses a namespace name from its CLI/method-prefix form.
+    pub fn from_prefix(s: &str) -> Option<Self> {
```
iovoid, line 241: Isn't this equivalent to resolve_namespace?

crates/networking/rpc/rpc.rs
```diff
@@ -939,17 +955,24 @@ pub async fn handle_eth_unsubscribe(
...
-        Ok(RpcNamespace::Engine) => Err(RpcErr::Internal(
-            "Engine namespace not allowed in map_http_requests".to_owned(),
-        )),
-        Err(rpc_err) => Err(rpc_err),
+    let namespace = match req.namespace() {
+        Ok(ns) => ns,
+        Err(rpc_err) => return Err(rpc_err),
+    };
+    // Engine is served on the authenticated port only; treat as not found here.
+    if matches!(namespace, RpcNamespace::Engine) || !context.allowed_namespaces.contains(&namespace)
```
iovoid: Why is this handled separately?

iovoid (changes requested):

iovoid:
docs/workflows/debug_execution_witness_benchmarking.md should be updated to enable the debug endpoint

iovoid (approved):

PR #6631, opened 2026-05-12
feat(l1): reorg re-injection for mempool transactions and blob sidecars

ilitteri:
## Motivation

When the engine API's `forkchoiceUpdated` moves the canonical head to a different chain branch, transactions in the orphaned blocks (but not in the new canonical chain) currently disappear from the mempool. They're still valid on the new chain but the node loses them. Every other major EL client re-injects these. This is the biggest remaining correctness gap in ethrex's mempool.

## Description

**Reorg detection** lives in the engine API FCU handler (`crates/networking/rpc/engine/fork_choice.rs`). The previous canonical head + finalized hash are snapshotted before `apply_fork_choice`. If the new head's parent is NOT the previous head, the FCU is a genuine reorg.

- **Fast-path**: when `new_head.parent_hash == previous_head_hash`, the FCU is a normal slot advance, not a reorg. Skip the re-injection spawn (and its ~3 header-by-hash DB reads) entirely. Genuine reorgs fall through.
- **Spawned, not awaited**: re-injection is `tokio::spawn`-ed off the FCU response path so the engine API ack returns immediately. A near-cap 64-block reorg with ~10K orphaned txs would otherwise add ~1+ second of ECDSA recovery + KZG re-verification to the FCU response (under the 8s spec timeout but tight enough to destabilize CL timing budgets).

**Re-injection** via `Blockchain::reinject_orphaned_transactions`: for each orphaned tx not present in the new-canonical set, call the regular admission path. Failures are debug-logged and best-effort. Honors a `--mempool.reorg-depth` cap (default 64); deeper reorgs are skipped with a warning. Honors mempool occupancy: a deep reorg won't evict freshly-arrived txs to make room for orphaned ones.

**Blob sidecar limbo**:

- On block inclusion, sidecars are **moved** (not cloned) from `blobs_bundle_pool` into `blobs_bundle_limbo`. The versioned-hash index cleanup runs separately so the bundle never leaves the write lock without a stable index state.
- On reorg, orphaned blob txs pull their sidecars from limbo. If `add_blob_transaction_to_pool` fails for a transient reason (pool full, etc.), the sidecar is **re-inserted into limbo** via `Mempool::insert_blob_limbo_entry` rather than dropped permanently — a later re-injection attempt can still recover it.
- On FCU `finalized_block_hash` advancement, limbo entries are purged by **canonical-block-number range walk** (no `reorg_depth` cap). The cap was a stale defense against malicious inputs, but the input here is a CL-supplied finalized hash, already canonical. The previous cap leaked sidecars whenever a finalization gap exceeded it (fresh-node first finalization, CL downtime catch-up).
- CLI: `--mempool.reorg-depth` (default 64).

## Behavioral change

- Txs orphaned by reorgs of depth ≤ `reorg_depth` are re-injected with full admission re-validation. Deeper reorgs skip re-injection (logged warning).
- Blob sidecars are retained between inclusion and finalization so blob-tx re-injection works through reorgs.
- The FCU response time is unaffected by re-injection cost (spawned).

## Deferred

End-to-end test driving `engine_forkchoiceUpdated` through a real multi-branch reorg. The load-bearing logic (common-ancestor, orphan subtraction, depth cap, blob-limbo lifecycle, capacity gate, best-effort failure handling, idempotence) is covered by unit tests.


crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -304,6 +324,46 @@ async fn handle_forkchoice(
...
+                        debug!(
+                            reinjected = count,
+                            "Re-injected orphaned transactions into mempool"
+                        );
+                    }
+                    Err(error) => {
+                        debug!(%error, "Failed to re-inject orphaned transactions");
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
ilitteri: 2a3dc0389

crates/blockchain/mempool.rs
```diff
@@ -227,6 +237,49 @@ impl Mempool {
...
+    /// Returns whether the transaction was present in the mempool.
+    pub fn remove_included_transaction(&self, hash: &H256) -> Result<bool, StoreError> {
+        let mut inner = self.write()?;
+        let was_present = inner.transaction_pool.contains_key(hash);
+        // First retain the bundle in limbo (only if the tx is a blob tx whose
+        // sidecar we already have). We do this before removing the transaction
+        // because `remove_transaction_with_lock` also drops the bundle.
+        if let Some(bundle) = inner.blobs_bundle_pool.get(hash).cloned() {
+            inner.blobs_bundle_limbo.insert(*hash, bundle);
+        }
```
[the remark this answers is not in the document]
ilitteri: 9604ab027

crates/blockchain/blockchain.rs
```diff
@@ -2374,10 +2385,175 @@ impl Blockchain {
...
+                    match self.add_blob_transaction_to_pool(inner, blobs_bundle).await {
+                        Ok(_) => reinjected += 1,
+                        Err(error) => {
+                            debug!(
+                                tx_hash = %tx_hash,
+                                %error,
+                                "Failed to re-inject orphaned EIP-4844 transaction",
+                            );
+                        }
+                    }
```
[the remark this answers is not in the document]
ilitteri: 9604ab027

crates/blockchain/blockchain.rs
```diff
@@ -2374,10 +2385,202 @@ impl Blockchain {
...
+            "Reorg detected; re-injecting orphaned transactions into mempool",
+        );
+
+        let mut reinjected = 0usize;
+        let mut skipped_full = 0usize;
+        for tx in txs {
+            // Don't evict freshly-arrived (post-reorg) txs to make room for
+            // orphaned ones. If the pool is already at capacity, stop
+            // re-injecting and log how many were skipped.
+            if self.mempool.is_full()? {
```
ElFantasma, line 3277: **TOCTOU between `is_full()` gate and inner admission.** Once this check passes, `add_blob_transaction_to_pool` / `add_transaction_to_pool` eventually call `Mempool::add_transaction`, which itself triggers `remove_oldest_transaction()` when the pool hits capacity (mempool.rs:187-188). So if a fresh tx arrives between this `is_full()` and the inner insert (or if multiple orphaned txs admit concurrently with a fresh submission), a freshly-arrived tx CAN still get evicted to make room for the orphaned one — exactly the scenario the comment at line 2464-2466 says we're preventing.

Not a behavior regression vs. main (this is just the normal mempool race), but the stated intent ("won't evict freshly-arrived txs to make room") is slightly stronger than what the code guarantees under contention. Either drop the gate (since the inner path handles capacity anyway) or document that the gate is best-effort and a tight race can still evict.

crates/blockchain/blockchain.rs
```diff
@@ -2374,10 +2385,202 @@ impl Blockchain {
...
+                    match self.add_blob_transaction_to_pool(inner, blobs_bundle).await {
+                        Ok(_) => reinjected += 1,
+                        Err(error) => {
+                            debug!(
+                                tx_hash = %tx_hash,
+                                %error,
+                                "Failed to re-inject orphaned EIP-4844 transaction; restoring sidecar to limbo",
+                            );
+                            self.mempool
+                                .insert_blob_limbo_entry(tx_hash, bundle_backup)?;
```
ElFantasma, line 3307: **Duplicate-insert risk if `add_blob_transaction_to_pool` partially succeeds.** Looking at the admission flow (blockchain.rs:2329-2354), `add_blobs_bundle` is called at line 2351 BEFORE `add_transaction` at line 2352. If `add_transaction` fails (e.g. `StoreError` from a poisoned lock during `remove_oldest_transaction`), the bundle is already in `blobs_bundle_pool` and the versioned-hash index. The catch arm here then re-inserts the same bundle into `blobs_bundle_limbo` — leaving the bundle in BOTH pool and limbo.

In practice the failure modes that reach this arm are mostly pre-pool-insert validation errors (`BlobsBundleInvalid`, `validate_transaction` returning `Err`), so the duplicate state is rare. But a cheap fix would be calling `take_blob_limbo_entry(&tx_hash)` first (or `get_blobs_bundle(hash)` to check) to make the re-insertion safe under partial failure. Alternatively, a one-line comment noting the assumption ("only the pre-insert validation failures reach this arm") would help future readers.

crates/blockchain/blockchain.rs
```diff
@@ -2374,10 +2385,202 @@ impl Blockchain {
...
+        } else {
+            match self
+                .storage
+                .get_block_header_by_hash(previous_finalized_hash)?
+            {
+                Some(header) => header.number.saturating_add(1),
+                // Previous finalized hash isn't known to this store (sync gap?).
+                // Fall back to a full walk from genesis so we still drain stale
+                // limbo entries.
+                None => 0,
```
ElFantasma: **Cold-start fallback walks from genesis.** When `previous_finalized_hash` isn't found in the store, `start_number` drops to `0` and the subsequent loop at line 2572 walks every canonical block from genesis (`get_canonical_block_hash` + `get_block_body_by_hash`) building `hashes_to_purge`. On a synced node restarting mid-finalization or after a sync gap on a deep chain (say 1M+ blocks), that's 2M+ DB reads holding the mempool write lock at the end via `purge_blob_limbo_entries`. The vector also grows to all tx hashes in the chain.

The comment at line 2541-2543 acknowledges removing the `reorg_depth` cap to fix the leak case, but doesn't address this fresh-node / sync-gap walk. A reasonable bound would be: when `previous_finalized_hash` isn't in the store, fall back to `new_header.number.saturating_sub(MAX_FINALIZATION_WALK)` (where MAX is e.g. some multiple of the reorg cap) instead of 0 — the limbo can't realistically contain entries older than the deepest possible reorg anyway, since sidecars only enter limbo via `remove_included_transaction` from FCU processing. The trade-off is leaking sidecars from blocks older than that bound on the cold path, vs. potentially scanning the entire chain.

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -304,6 +324,59 @@ async fn handle_forkchoice(
...
+            // orphaned txs would otherwise add ~1+ second of ECDSA recovery + KZG
+            // re-verification to the engine API response (well under the 8s spec
+            // timeout but tight enough to destabilize CL timing budgets).
+            //
+            // Fast-path: in the normal slot-advance case the new head's parent
+            // IS the previous head, i.e. no reorg. Skip the spawn (and its 3
+            // header-by-hash DB reads) to keep typical FCUs lean. Genuine
+            // reorgs — where the new head's parent differs from the previous
+            // head — still fall through to the spawned re-injection.
+            if !previous_head_hash.is_zero()
```
ElFantasma, line 376: nit: the comment block above (line 334-338) describes only the "new head's parent IS the previous head" fast-path, but the actual guard has three conjuncts. A first-time FCU (`previous_head_hash.is_zero()`) and a same-head FCU (`previous_head_hash == head.hash()`) also skip — both are correct, but a one-liner noting them in the comment would help future readers reason about the condition without re-deriving it from the code.

azteca1998 (approved):

crates/blockchain/mempool.rs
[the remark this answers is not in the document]
ElFantasma: Confirmed fixed in 9604ab027 — the bundle is **moved**, not cloned:

```rust
if let Some(bundle) = inner.blobs_bundle_pool.remove(hash) {
    inner.blobs_bundle_limbo.insert(*hash, bundle);
}
```

Verified against the current code after merging `main`.

crates/blockchain/blockchain.rs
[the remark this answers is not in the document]
ElFantasma: Confirmed fixed in 9604ab027 — the sidecar is cloned before admission and restored to limbo if it fails, so a transient rejection no longer loses it permanently:

```rust
let bundle_backup = blobs_bundle.clone();
match self.add_blob_transaction_to_pool(inner, blobs_bundle).await {
    Ok(_) => reinjected += 1,
    Err(error) => {
        debug!(... "restoring sidecar to limbo");
        self.mempool.insert_blob_limbo_entry(tx_hash, bundle_backup)?;
    }
}
```

crates/blockchain/blockchain.rs
ElFantasma: Fixed in 0b27187ea, though not the way this comment implied — bounding the walk by `reorg_depth` trades the hang for a leak (that's exactly what the other review thread on this function objected to), so the two suggestions pull against each other.

Resolved by removing the walk from those paths entirely. When the finalization delta can't be determined — `previous_finalized_hash` is zero, or it isn't in this store after a sync gap — limbo is now cleared wholesale via a new `Mempool::clear_blob_limbo()` and the function returns:

```rust
let start_number = if previous_finalized_hash.is_zero() {
    let cleared = self.mempool.clear_blob_limbo()?;
    ...
    return Ok(());
} else {
    match self.storage.get_block_header_by_hash(previous_finalized_hash)? {
        Some(header) => header.number.saturating_add(1),
        None => { /* same: clear + return */ }
    }
};
```

That's sound because limbo is in-memory and best-effort. On a genuine cold start it is already empty, so this is a no-op — the case that actually matters is a sync gap, where the cost is at most a few orphaned blob txs not being re-injected, which the re-injection path already handles by logging and skipping. No unbounded walk, and no leak either.

Covered by `clear_blob_limbo_drops_everything`, which checks it empties a populated limbo and is a no-op on an empty one.

crates/blockchain/blockchain.rs
ElFantasma: Not an issue after merging `main` — `add_blob_transaction_to_pool` now rolls the bundle back itself when the insert fails:

```rust
if let Err(e) = inserted {
    let _ = self.mempool.remove_blobs_bundle(&hash);
    return Err(e);
}
```

So by the time the re-injection error arm runs, the bundle is no longer in `blobs_bundle_pool`, and restoring it to limbo can't leave it held in both places. Verified against the merged code.

crates/blockchain/blockchain.rs
ElFantasma: Leaving this one as-is deliberately, and documenting rather than closing it.

The window is real: `is_full()` takes the read lock, releases it, and the subsequent admission takes the write lock, where `add_transaction` will evict if the pool is at capacity. So a transaction arriving in between can be displaced by an orphan — the exact thing the gate exists to prevent.

What it costs is bounded, though: at most one fresh transaction per orphan that races, and only while the pool is exactly at capacity. Closing it properly needs an atomic "insert only if not full" entry point on `Mempool`, which is real API surface added for a best-effort path whose failure mode is "one transaction is evicted slightly earlier than it would have been". Re-injection already tolerates admission failures by design.

Happy to add that API if you'd rather have the guarantee — flagging the trade-off rather than silently leaving a known gap.

crates/networking/rpc/engine/fork_choice.rs
ElFantasma: Done in 33d782cc3 — the comment now names all three conjuncts:

```rust
// The guard has three conjuncts, and the other two are also
// no-reorg cases worth naming: a zero `previous_head_hash` is the
// node's first FCU (nothing to orphan), and `previous_head_hash ==
// head.hash()` is a repeated FCU for the head we already have.
```

PR #6633, opened 2026-05-12
test(l1): cover discv5 IP-voting guard and restore stale-entry cleanup trace

ElFantasma:
**Motivation**

Follow-up to [#6449](https://github.com/lambdaclass/ethrex/pull/6449) (unified DiscoveryServer). Addresses two non-blocking items from edg-l's review:

1. **Test gap on IP voting guard** — `test_ip_voting_no_update_if_same_ip` calls `Discv5State::record_ip_vote` directly, which never mutates `server.local_node`, so the assertion at `local_node.ip == original_ip` trivially passes. The actual guard at `discv5_handlers.rs:379` (`winning_ip != self.local_node.ip`, gating `update_local_ip`) was unexercised.

2. **Observability regression on stale-entry cleanup** — the old `discv5::cleanup_stale_entries` emitted `trace!("Cleaned up N stale entries (M messages, K challenges)")`; the new `Discv5State::cleanup_stale_entries` no longer did.

**Description**

- Added `test_handle_pong_same_ip_does_not_bump_enr_seq` in `test/tests/p2p/discovery/discv5_server_tests.rs`. Drives `server.discv5_handle_pong` three times with `recipient_addr.ip()` equal to `server.local_node.ip` and distinct sender IDs. After the threshold is reached the new test asserts that `server.local_node_record.seq` is unchanged — directly locking in the guard at `discv5_handlers.rs:379`.
- Restored the cleanup trace in `Discv5State::cleanup_stale_entries`, only emitted when something was actually removed (avoids per-tick noise).

**Checklist**

- [x] No `Store` schema change.


edg-l (approved):

ElFantasma:
Addressed feedback from bot reviews in 87de359 and 4cddfb0:

- **From Claude (moderate)** — Added `assert!(first_ip_vote_round_completed)` to `test_handle_pong_same_ip_does_not_bump_enr_seq` to make the test falsifiable: without it, a regression that prevents finalization would make both `local_node.ip` and `local_node_record.seq` assertions trivially pass without ever exercising the guard.
- **From Claude (minor)** — Added a comment explaining why an empty `req_id` is fine (`record_pong_received` silently ignores unmatched req_ids).

Not addressed:
- **From Kimi** — `saturating_sub` for the cleanup arithmetic. `retain()` can only shrink the maps, so the subtraction cannot underflow; defensive arithmetic for an impossible scenario isn't worth adding.
- **From Claude (minor)** — Threshold coupling (`for i in 1..=3u64`). The new round-completion assertion already catches this: if `IP_VOTE_THRESHOLD` changes and the loop runs short, the round won't complete and the assertion fails. Not worth exposing the constant as `pub` just for this.

ilitteri (approved):

azteca1998 (approved):

PR #6639, opened 2026-05-13
perf(l1): reduce BAL parallel-path overhead

edg-l:
## Summary

Rebased onto current `main` (post bal-devnet-7-pr squash merge). **Scope reduced** from the original bundle: dropped the `CachingDatabase` `RwLock<HashMap>` → `DashMap` swap and the 500ms → 100ms `import-bench` sleep tweak. Both require coordinated zkVM-feature-gating work that isn't done here.

Remaining changes target the BAL parallel-execution path (`execute_block_parallel` + `handle_merkleization_bal` + `seed_db_from_bal`). All changes are within already rayon-gated code paths -- zkVM builds (sp1/risc0/zisk/eip-8025) are unaffected.

### What is in this PR

- **A. `handle_merkleization_bal` overlap fix** (`crates/blockchain/blockchain.rs`): `for updates in rx { ... }` blocked until channel close (= exec end). `execute_block_parallel` sends exactly one batch up front from `bal_to_account_updates`, so draining nothing useful serialized Stage B (parallel storage roots) after exec instead of overlapping with it. Replaced with a single `rx.recv()` and dropped the `FxHashMap` merge step (BAL guarantees one entry per address).
- **Q1. Skip prestate read in `bal_to_account_updates` when BAL covers all info fields** (`crates/vm/backends/levm/mod.rs`): two fast paths added -- storage-only updates (info: None, removed: false by construction); full info coverage with non-empty post (removal impossible, info from BAL alone). Slow path keeps existing behavior for partial coverage.
- **Q2. Per-tx `GeneralizedDatabase` capacity cap at 32** (`execute_block_parallel`): previously sized to `bal.accounts().len()` (often 100s on stress blocks); p50 tx touches <10 accounts. Reduced allocator pressure across rayon workers.
- **Q3. Memoize `code_from_bal` results across `seed_db_from_bal` calls**: pre-compute `Code` objects (hash + jump_targets) once per BAL code change before the par_iter; pass cache via optional param to `seed_db_from_bal`. Saves N-1 keccak+jump-target scans per code change per block (N = tx count).
- **Q8. Move per-tx BAL validation into the rayon par_iter closure** (`execute_block_parallel`): eliminates a serial post-exec validation pass (~3 ms median across 200 txs). Drops `current_state` and `codes` inside the closure after validation runs -- they no longer cross the rayon boundary, reducing per-tx allocator pressure. Closure returns deferred `Option<EvmError>` so gas-limit check still takes priority over BAL mismatch errors.
- **Coinbase exemption fix**: gated the post-exec `unaccessed_pure_accounts` coinbase removal on `!exec_results.is_empty()` to fix 0-tx-block regression (Q8 had hoisted the removal outside the per-tx loop, silently exempting coinbase on empty/withdrawal-only blocks). Found via EELS `test_bal_invalid_extraneous_coinbase[empty_block|withdrawal_only]`.
- **Pre-size `backup_storage_slot` inner map** (separate commit): avoids rehash chain growth in the slot backup tracker.

### What was dropped from the original bundle (deferred to follow-ups)

- **DashMap. `CachingDatabase` RwLock<HashMap> → DashMap**: the `Cargo.toml` rayon gating from #6662 makes the levm crate's rayon dep optional (zkVM builds skip rayon entirely). The DashMap swap rewrites struct fields and methods unconditionally; to land cleanly it needs paired zkVM-fallback variants matching the existing `#[cfg(all(feature = "rayon", not(feature = "eip-8025")))]` pattern. Split out to its own PR.
- **C. `import-bench` 500ms → 100ms sleep**: superseded by #6666 which replaces the sleep with `Store::wait_for_persistence_idle()` entirely.

### Headline numbers (from the original bundle, before reduced scope)

460-block mainnet-mix fixture from bal-devnet-7 kurtosis localnet (~670 tx/s spamoor mix, 65 Mgas median, ~200 tx/block, 452/460 blocks carry BAL):

| Metric                  | Value          |
|-------------------------|----------------|
| Wall time               | 5.06 s         |
| Per-block median        | ~8 ms          |
| Per-block Ggas/s median | 7.67           |
| Per-block Ggas/s p90    | 8.32           |
| Per-block Ggas/s max    | 13.49          |
| Warmer/exec overlap     | 99 %           |

Numbers above include the DashMap change; expect a regression of a few % once that's removed (perf record showed 11% CPU on the RwLock contention -- DashMap removed that; without it some of that comes back, exact figure pending re-bench).

## Test plan

- [x] `cargo check -p ethrex` (default features) -- clean
- [x] `cargo check -p ethrex-levm --no-default-features --features sp1` (zkVM mode, no rayon) -- clean
- [x] `cargo fmt --check` -- clean
- [ ] CI: `make test`, EF tests, Hive Amsterdam
- [ ] Re-bench on 460-block fixture without DashMap to quantify regression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 2026-05-14
 
 - Skip `vm.run_execution()` for transfers to codeless EOAs [#6570](https://github.com/lambdaclass/ethrex/pull/6570)
+- Reduce BAL parallel-path overhead: overlap merkleization with execution, memoize per-BAL code derivation, and move per-tx BAL validation inside the exec closure [#6639](https://github.com/lambdaclass/ethrex/pull/6639)
 
 ### 2026-04-27
 
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -865,36 +865,40 @@ impl Blockchain {
         const NUM_WORKERS: usize = 16;
         let parent_state_root = parent_header.state_root;
 
-        // === Stage A: Drain + accumulate all AccountUpdates ===
-        // BAL guarantees completeness, so we block until execution finishes.
-        let mut all_updates: FxHashMap<Address, AccountUpdate> = FxHashMap::default();
-        for updates in rx {
-            let current_length = queue_length.fetch_sub(1, Ordering::Acquire);
-            *max_queue_length = current_length.max(*max_queue_length);
-            for update in updates {
-                match all_updates.entry(update.address) {
-                    Entry::Vacant(e) => {
-                        e.insert(update);
-                    }
-                    Entry::Occupied(mut e) => {
-                        e.get_mut().merge(update);
-                    }
-                }
+        // === Stage A: receive the single BAL-derived batch ===
+        // execute_block_parallel calls bal_to_account_updates BEFORE the rayon tx
+        // loop and sends exactly one Vec<AccountUpdate>. Receiving once (instead of
+        // draining until channel close = exec end) lets Stage B's parallel storage
+        // roots overlap with parallel exec instead of serializing after it.
+        //
+        // BAL accounts are unique by address (one entry per touched address), so
+        // no merge step is needed — skip the FxHashMap detour entirely.
+        let updates: Vec<AccountUpdate> = match rx.recv() {
+            Ok(updates) => {
+                let current_length = queue_length.fetch_sub(1, Ordering::Acquire);
+                *max_queue_length = current_length.max(*max_queue_length);
+                updates
             }
-        }
+            Err(_) => {
+                // Channel closed without a message — execution failed before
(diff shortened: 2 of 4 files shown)


edg-l:
Closing. Of the original bundle:
- A (Stage A rx.recv fix) is obsoleted by #6655 (validation path deletes Stage A entirely).
- Q1 fast paths' validation-side benefit is obsoleted by #6655 (skips bal_to_account_updates on validation path); builder-path benefit remains and may land as a separate PR if benched.
- Q2 cap-32 per-tx DB is already in #6669.
- Q3 code_from_bal cache is partially obsolete given #6669; remaining utility lives only on the system-call / outer-seed paths and may be relanded separately.
- Coinbase fix already in main.
- Q8 (move per-tx BAL validation into par_iter closure) and the backup_storage_slot pre-size are still uniquely useful; re-landing as small focused PRs.

PR #6655, opened 2026-05-14
perf(l1): bal optimistic merkleization (or post-state root optimization)

### LINKED ISSUE #6584: BAL Optimistic merkleization
## Summary

Decouple merkleization from execution timing on the validation path. Instead of waiting for `AccountUpdate`s to stream out of the EVM, synthesize the equivalent updates directly from the input BAL the moment its hash matches the block header, and run all four merkleization stages fully in parallel with execution.

## Background

`handle_merkleization_bal` (`crates/blockchain/blockchain.rs:830`) today:

- **Stage A** drains `rx` until execution finishes pushing every `AccountUpdate` batch.
- **Stages B–D** (parallel storage roots → sharded state-trie update → root assembly) start only after Stage A finishes.

PR #6543 makes Stage A's drain a single `recv()` so Stage B starts at the *tail* of execution. That still gates merkleization on the channel close.

EIP-7928 BALs carry **post-execution values** for every storage write (one per `(slot, tx_index)`) and post-block account info. Once `keccak256(rlp(bal)) == header.block_access_list_hash` is verified, every input merkleization needs is already in hand — no `rx` required.

## Proposal

Add a new entry path that takes `&BlockAccessList` directly (no channel) and runs all four stages from t=0 of `engine_newPayload` validation:

1. Verify `header.block_access_list_hash` against the input BAL — already required for block validity.
2. Synthesize `FxHashMap<Address, AccountUpdate>` from the BAL: for each account, take the last (highest `tx_index`) post-value per slot; pull `info` (balance, nonce, code_hash) from the BAL; mark `removed`/`removed_storage` from BAL flags.
3. Spawn the new merkleization handle from the same `spawn_scoped` block at `blockchain.rs:529`, alongside `execution_handle` and `warm_handle` — not after.
4. Keep the existing `validate_block_access_list_hash(produced_bal, ...)` post-execution check as the correctness gate. If it fails the block is invalid and both execution and optimistic merkle results are discarded.

The optimistic merkle root is correct iff the block is valid: there is no scenario where a valid block produces a wrong root via this path.

## Scope

- **Validation path only** (`engine_newPayload`). Block production has no input BAL — keep streaming on the builder path.
- **Amsterdam+ only** (no BAL pre-fork).

## Implementation notes

- Reuse the Stage B/C/D structure unchanged; just feed it the synthesized map instead of the post-Stage-A map.
- Witness accumulator: build once from the synthesized map (matches what PR #6406 does for the witness path).
- Code bytes: trie work only needs `code_hash`; bytes can stay on the warmer's path.
- The warmer and optimistic-merkle Stage B both pull the same keys; second consumer just hits warm RocksDB block cache. Likely fine but worth measuring.

edg-l:
## Summary

Decouples BAL-merkleization from EVM execution on the `engine_newPayload` validation path. Synthesizes per-field state deltas from the input BAL pre-`thread::scope` and runs merkle stages B/C/D in parallel with execution + warming. BAL-driven state prefetch lives on the warmer thread (Phase 1 of `warm_block_from_bal`), not duplicated on the merkleizer.

Closes #6584. Validation path only (builder still streams). Amsterdam+ only.

## Design

- `BalSynthesisItem` carries per-field optionals (no `Option<AccountInfo>` blob), so a balance-only recipient doesn't fabricate zero-nonce / `EMPTY_KECCACK_HASH` deltas.
- `handle_merkleization_bal` → `handle_merkleization_bal_from_updates`; Stage A (channel drain) deleted on the BAL path.
- EVM side: `execute_block_pipeline` / `execute_block_parallel` take `Option<Sender>`. BAL caller passes `None` and the EVM skips `bal_to_account_updates` + `merkleizer.send`. Outer `if let Some(bal) = header_bal` is also gated on `is_amsterdam`.
- **State prefetch**: `warm_block_from_bal`'s Phase 1 (`prefetch_accounts`) opens the parent state trie and walks each touched account's path; the merkleizer reuses those warmed pages and does not duplicate the walk.
- Hashed-key sort on `added_storage` before per-slot inserts; trie walks node-arena paths in order.

## Correctness

- `removed` / `removed_storage` not inferred from BAL; Stage B `value.is_zero()` + Stage C EIP-161 collapse handle selfdestruct identically to streaming.
- The Amsterdam path doesn't rebuild a BAL to hash-compare; the header BAL drives execution and is validated in `execute_block_pipeline` via `validate_header_bal_indices`, `validate_bal_withdrawal_index`, the `unread_storage_reads` / `unaccessed_pure_accounts` post-exec checks, and `validate_block_access_list_size`. Downstream `state_root` vs header still applies. (Pre-Amsterdam / streaming paths still return a produced BAL and the existing `validate_block_access_list_hash` check runs over it.)

## Benchmark

Fixture: `bal-devnet-7-mainnet-mix-460` (460 blocks, ~30 Ggas, transfer/EVM mix). `release-with-debug`, `import-bench --with-bal`. Baseline = `feat/import-bench-bal-tooling` (bench tooling only, no perf change).

| metric             | baseline | this PR | delta              |
|--------------------|---------:|--------:|--------------------|
| wall time          | 8.575 s  | 8.437 s | -138 ms (-1.6%)    |
| agg Ggas/s         | 3.901    | 3.991   | +2.3%              |
| avg ms / block     | 16.884   | 16.505  | -0.38 ms (-2.2%)   |
| **merkle drain**   | **0.479 ms** | **0.047 ms** | **-90%**     |
| merkle concurrent  | 15.571 ms| 1.587 ms| merkle finishes alongside exec |
| merkle overlap     | 96.7%    | 97.2%   | +0.5 pp            |
| exec               | 15.574 ms| 15.650 ms | flat             |
| store              | 0.667 ms | 0.648 ms | flat             |
| warmer             | 1.365 ms | 1.877 ms | +0.5 ms (state prefetch on warmer thread) |

The fixture is exec-bound, so wall savings are bounded by `max(0, merkle_total - exec_window)`. On mainnet-shaped blocks merkleization is a much larger share of the budget; the structural win scales there.

Compare dashboard: <https://edgl.dev/share/compare-feat-import-bench-bal-tooling@02a5663c-vs-perf-bal-optimistic-merkleization@ebd60c81.html>

## Test plan

- [x] `cargo test -p ethrex-common synthesize_tests` (13/13)
- [x] `cargo fmt --check` / `make lint`
- [x] EF blockchain tests (two-pass parallel runner exercises this path on every Amsterdam fixture)
- [ ] Hive `bal` group

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### 2026-05-14
 
 - Skip `vm.run_execution()` for transfers to codeless EOAs [#6570](https://github.com/lambdaclass/ethrex/pull/6570)
+- BAL optimistic merkleization: synthesize state deltas from the input Block Access List pre-execution and merkleize in parallel with the EVM on the `engine_newPayload` validation path. Includes a parallel state-trie pre-warm and per-account hashed-key-sorted storage inserts to keep the trie node arena hot for Stage B/C [#6655](https://github.com/lambdaclass/ethrex/pull/6655)
 
 ### 2026-05-21
 
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -64,8 +64,9 @@ use ethrex_common::types::block_access_list::BlockAccessList;
 use ethrex_common::types::block_execution_witness::ExecutionWitness;
 use ethrex_common::types::fee_config::FeeConfig;
 use ethrex_common::types::{
-    AccountInfo, AccountState, AccountUpdate, Block, BlockHash, BlockHeader, BlockNumber,
-    ChainConfig, Code, Receipt, Transaction, WrappedEIP4844Transaction, validate_block_body,
+    AccountInfo, AccountState, AccountUpdate, BalSynthesisItem, Block, BlockHash, BlockHeader,
+    BlockNumber, ChainConfig, Code, Receipt, Transaction, WrappedEIP4844Transaction,
+    synthesize_bal_updates, validate_block_body,
 };
 use ethrex_common::types::{ELASTICITY_MULTIPLIER, P2PTransaction};
 use ethrex_common::types::{Fork, MempoolTransaction};
@@ -93,6 +94,7 @@ use ethrex_vm::backends::levm::db::DatabaseLogger;
 use ethrex_vm::{BlockExecutionResult, DynVmDatabase, Evm, EvmError};
 use mempool::Mempool;
 use payload::PayloadOrTask;
+use rayon::prelude::*;
 use rustc_hash::{FxHashMap, FxHashSet};
 use std::collections::hash_map::Entry;
 use std::collections::{BTreeMap, HashMap, HashSet};
@@ -137,7 +139,7 @@ type BlockExecutionPipelineResult = (
     Option<Vec<AccountUpdate>>,
     Option<BlockAccessList>, // produced BAL (Some on Amsterdam+ blocks)
     usize,                   // max queue length
-    [Instant; 6],            // timing instants
+    [Instant; 7],            // timing instants
     Duration,                // warmer duration
 );
 
@@ -317,7 +319,9 @@ struct PreMerkelizedAccountState {
 /// Work item for BAL state trie shard workers.
 struct BalStateWorkItem {
     hashed_address: H256,
-    info: Option<AccountInfo>,
+    nonce: Option<u64>,
+    balance: Option<U256>,
+    code_hash: Option<H256>,
     removed: bool,
     /// Pre-computed storage root from Stage B, or None to keep existing.
     storage_root: Option<H256>,
@@ -468,6 +472,24 @@ impl Blockchain {
 
         let cancelled = AtomicBool::new(false);
 
+        // Synthesize BAL updates pre-scope so the merkleizer thread can start
+        // trie work immediately, in parallel with execution.
+        let optimistic_updates: Option<FxHashMap<Address, BalSynthesisItem>> =
+            bal.map(synthesize_bal_updates);
+        let optimistic_witness: Option<Vec<AccountUpdate>> = if self.options.precompute_witnesses {
+            optimistic_updates.as_ref().map(|m| {
+                m.iter()
+                    .map(|(addr, item)| AccountUpdate {
+                        address: *addr,
+                        added_storage: item.added_storage.clone(),
+                        ..Default::default()
+                    })
+                    .collect()
+            })
+        } else {
+            None
+        };
+
         let (execution_result, merkleization_result, warmer_duration) =
             std::thread::scope(|s| -> Result<_, ChainError> {
                 #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
@@ -505,7 +527,16 @@ impl Blockchain {
                         ChainError::Custom(format!("Failed to spawn warmer thread: {e}"))
                     })?;
                 let max_queue_length_ref = &mut max_queue_length;
-                let (tx, rx) = channel();
+                // Channel only exists on the streaming (non-BAL) path. On the BAL path the
+                // EVM merkleizes nothing on its own (the synthesized map drives merkleization),
+                // so no Sender / drain thread are needed.
+                let (tx, rx_for_merkle) = if bal.is_some() {
+                    (None, None)
+                } else {
+                    let (tx, rx) = channel();
+                    (Some(tx), Some(rx))
+                };
+
                 let execution_handle = std::thread::Builder::new()
                     .name("block_executor_execution".to_string())
                     .spawn_scoped(s, move || -> Result<_, ChainError> {
@@ -551,28 +582,40 @@ impl Blockchain {
                         ChainError::Custom(format!("Failed to spawn execution thread: {e}"))
                     })?;
                 let parent_header_ref = &parent_header; // Avoid moving to thread
+                // Merkleizer returns (list, streaming witness or None on BAL path, merkle_start, merkle_end).
+                type MerkleResult = Result<
+                    (
+                        AccountUpdatesList,
+                        Option<Vec<AccountUpdate>>,
+                        Instant,
+                        Instant,
+                    ),
+                    StoreError,
+                >;
                 let merkleize_handle = std::thread::Builder::new()
                     .name("block_executor_merkleizer".to_string())
-                    .spawn_scoped(s, move || -> Result<_, StoreError> {
-                        let (account_updates_list, accumulated_updates) = if bal.is_some() {
-                            self.handle_merkleization_bal(
-                                rx,
-                                parent_header_ref,
-                                queue_length_ref,
-                                max_queue_length_ref,
-                            )?
-                        } else {
-                            self.handle_merkleization(
-                                rx,
-                                parent_header_ref,
-                                queue_length_ref,
-                                max_queue_length_ref,
-                            )?
-                        };
+                    .spawn_scoped(s, move || -> MerkleResult {
+                        let merkle_start_instant = Instant::now();
+                        let (account_updates_list, streaming_witness) =
+                            if let Some(prepared) = optimistic_updates {
+                                let list = self.handle_merkleization_bal_from_updates(
+                                    prepared,
+                                    parent_header_ref,
+                                )?;
+                                (list, None)
+                            } else {
+                                self.handle_merkleization(
+                                    rx_for_merkle.expect("rx is Some on non-BAL path"),
+                                    parent_header_ref,
+                                    queue_length_ref,
+                                    max_queue_length_ref,
+                                )?
+                            };
                         let merkle_end_instant = Instant::now();
                         Ok((
                             account_updates_list,
-                            accumulated_updates,
+                            streaming_witness,
+                            merkle_start_instant,
                             merkle_end_instant,
                         ))
                     })
@@ -597,8 +640,12 @@ impl Blockchain {
                 let warmer_duration = Duration::ZERO;
                 Ok((execution_result, merkleization_result, warmer_duration))
             })?;
-        let (account_updates_list, accumulated_updates, merkle_end_instant) = merkleization_result?;
+        let (account_updates_list, streaming_witness, merkle_start_instant, merkle_end_instant) =
+            merkleization_result?;
         let (execution_result, produced_bal, exec_end_instant) = execution_result?;
+
+        // Synthesized witness wins when BAL is present; streaming witness wins otherwise.
+        let accumulated_updates = optimistic_witness.or(streaming_witness);
 
         let exec_merkle_end_instant = Instant::now();
 
@@ -612,6 +659,7 @@ impl Blockchain {
                 start_instant,
                 block_validated_instant,
                 exec_merkle_start,
+                merkle_start_instant,
                 exec_end_instant,
                 merkle_end_instant,
                 exec_merkle_end_instant,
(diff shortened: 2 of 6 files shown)


crates/blockchain/blockchain.rs
```diff
@@ -843,85 +891,68 @@ impl Blockchain {
...
             }
-            accounts.push((hashed, update));
+            accounts.push((hashed, item));
         }
 
+        // Warm parent state-trie pages for all touched accounts in parallel before
+        // Stage B / Stage C race for them. This replaces the prefetch that the old
+        // streaming path got for free via `bal_to_account_updates`.
+        accounts
+            .par_iter()
```
ElFantasma: **Pre-warm opens one state trie per account.** Stage B and Stage C below use the canonical one-trie-per-worker pattern (line ~997 and ~1093: `let state_trie = self.storage.open_state_trie(parent_state_root)?;` inside each worker's body, then loop over the bin/shard). This pre-warm pass does the opposite — `accounts.par_iter()` opens a fresh state trie for *every account*, then issues one `.get()`.

If `open_state_trie` is cheap (just a cursor against a shared RocksDB handle), the cost is dominated by the `.get()` and this is fine. If it's not (e.g. it builds a per-call cache, allocates a node arena, etc.), the warming pass becomes N opens vs Stage B's `max(num_bins, 16)` opens — strictly more work to warm than to execute. Either way, mirroring Stage B's pattern (rayon chunk the addresses, one trie open per chunk, batch the gets) keeps the three stages consistent and removes the question.

Non-blocking — measurable either way, but worth at least a comment noting the deliberate choice.

crates/common/types/block_access_list.rs
```diff
@@ -1591,6 +1592,81 @@ impl BlockAccessListRecorder {
...
+        let nonce = account.nonce_changes.last().map(|c| c.post_nonce);
+        let code = account.code_changes.last().map(|c| {
+            let hash = keccak(&c.new_code);
+            Code::from_bytecode_unchecked(c.new_code.clone(), hash)
+        });
+        let code_hash = code.as_ref().map(|c| c.hash);
+
+        let mut added_storage: FxHashMap<H256, U256> = FxHashMap::default();
+        for sc in &account.storage_changes {
+            debug_assert!(
```
ElFantasma: nit: `debug_assert!(!sc.slot_changes.is_empty(), …)` followed by `if sc.slot_changes.is_empty() { continue; }` two lines down express the same intent — one panics in debug, one defensively skips in release. If the invariant is real (canonical BAL ordering forbids empty `slot_changes`), the `continue` is unreachable and the `.expect()` at line 1649 should be the only fail-safe. If the invariant is defensible (a malformed BAL on the wire can produce empty `slot_changes`), the debug_assert is misplaced — that's a wire-format violation that should return a real `MempoolError`/decode error upstream, not panic in dev and silently skip in prod.

Pick one. The test at line 1936 (`synthesize_skips_when_slot_changes_empty`) is currently `#[cfg_attr(debug_assertions, should_panic(…))]` which encodes the inconsistency.

crates/vm/backends/levm/mod.rs
```diff
@@ -548,7 +551,10 @@ impl LEVM {
             ));
         }
 
-        // Sequential path (existing code, for block production and non-Amsterdam)
+        // Sequential path (existing code, for block production and non-Amsterdam).
+        // The non-BAL caller always provides a Sender; the BAL path returned above.
+        let merkleizer = merkleizer
+            .expect("sequential execution path requires a merkleizer Sender (non-BAL caller)");
```
ElFantasma: `.expect("sequential execution path requires a merkleizer Sender (non-BAL caller)")` is a structural panic that depends on the early-return for the BAL branch a few hundred lines up staying exactly where it is. If a future refactor reshapes the BAL branch (e.g. extracts it into a helper, moves the early return, or introduces a third path), the contract here can silently break and the next non-BAL block on mainnet panics the EVM thread.

Cheap belt-and-suspenders: `let Some(merkleizer) = merkleizer else { return Err(EvmError::Custom("non-BAL execution path called without merkleizer Sender".into())); };` — same control flow, returns instead of panics, surfaces the broken contract through the normal error path instead of bringing down the executor thread.

ElFantasma (commented):
Three inline findings + two body observations, all minor. Not blocking.

- **`validate_block_access_list_hash` is the only correctness net** for the optimistic merkle output, and it's gated on `produced_bal.is_some()` at `blockchain.rs:569-576`. On the BAL path this should always fire (Amsterdam+ produces a BAL), but the correctness story is now load-bearing on a runtime invariant rather than a static one. Worth a `debug_assert!(produced_bal.is_some(), "BAL validation path must produce a BAL")` on the optimistic branch to surface a future regression early.

- **`execute_block_pipeline`'s `if let Some(bal) = header_bal` does NOT check `is_amsterdam`** (`levm/mod.rs:391`, pre-existing). The `debug_assert!(is_amsterdam, …)` inside `execute_block_parallel` is a no-op in release. This PR adds another caller (the new optimistic merkle path) that depends on parallel execution running only on Amsterdam+. If `add_block_pipeline` is ever called with `Some(bal)` on a pre-Amsterdam block in release, the parallel path runs incorrectly AND the synthesized merkle is consumed unchecked. Recommend gating the outer `if let Some(bal)` on `is_amsterdam && header_bal.is_some()`, or upgrading the inner debug_assert to a real `EvmError`. Practical risk is low (header BAL is feature-gated input to engine_newPayload), but the cost-to-fix is tiny.

edg-l:
Pushed fixes for review comments in e7aaec8c3:

- dropped unused `BalSynthesisItem.address`
- dropped dead `BalStateWorkItem.removed` branch
- pre-warm now chunks accounts across workers (one trie open per worker, matching Stage B/C) instead of one open per account
- removed the conflicting `debug_assert!` + `.expect()` on empty `slot_changes`; defensive `continue` stays, test updated
- replaced the sequential-path `merkleizer.expect(...)` with a typed `EvmError`
- gated `execute_block_pipeline`'s `header_bal` branch on `is_amsterdam`
- when the BAL validation path produces no BAL, return a `ChainError` instead of skipping the hash check

azteca1998 (approved):

crates/blockchain/blockchain.rs
```diff
@@ -843,64 +896,40 @@ impl Blockchain {
...
-    ///
-    /// When the Block Access List is available (Amsterdam+), all dirty accounts
-    /// and storage slots are known upfront. This enables computing storage roots
-    /// in parallel across accounts before feeding final results into state trie
-    /// shards.
+    /// Validation path synthesizes `BalSynthesisItem`s from the input BAL pre-execution and
+    /// merkleizes optimistically in parallel with EVM execution. Two gates guard the result:
+    /// (1) `validate_block_access_list_hash` against the produced BAL post-execution, and
+    /// (2) the downstream `state_root` comparison against the block header. A missing
+    /// `produced_bal` on this path is treated as a hard error so gate (1) is never silently
```
ElFantasma: **Docstring is stale after the `79cf5327c` revert.** This block was added by `98c512adc` (09:25) under the assumption that a missing `produced_bal` would be a hard error. `79cf5327c` (10:44) reverted that hard error because the Amsterdam BAL validation path intentionally returns `produced_bal = None` (header BAL is authoritative; no rebuild for hash-comparison). The execution thread now has a fresh comment at `blockchain.rs:565-571` correctly explaining this — but this docstring still says *"A missing `produced_bal` on this path is treated as a hard error so gate (1) is never silently skipped"*, and the `if let Some(bal) = &produced_bal { validate_block_access_list_hash(...) }` has no else.

So on the Amsterdam validation path, gate (1) IS silently skipped (correctly so). The real "two gates" are:
1. The header-BAL invariants inside `execute_block_pipeline` (`unread_storage_reads` empty, `unaccessed_pure_accounts` empty, structural index/size validation)
2. The downstream `state_root` comparison against the block header (done by the caller in `add_block_pipeline`)

`validate_block_access_list_hash` is gate (1) only on the streaming path where `produced_bal` is `Some(...)`. Worth rewriting the docstring to reflect what `79cf5327c` actually settled on.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### 2026-05-14
 
 - Skip `vm.run_execution()` for transfers to codeless EOAs [#6570](https://github.com/lambdaclass/ethrex/pull/6570)
+- BAL optimistic merkleization: synthesize state deltas from the input Block Access List pre-execution and merkleize in parallel with the EVM on the `engine_newPayload` validation path. Includes a parallel state-trie pre-warm and per-account hashed-key-sorted storage inserts to keep the trie node arena hot for Stage B/C [#6655](https://github.com/lambdaclass/ethrex/pull/6655)
```
ElFantasma: **Stale claim: "parallel state-trie pre-warm" was removed in `5052c5804`.** The merkleizer-side `accounts.par_iter()` pre-warm was dropped after profiling showed `warm_block_from_bal` Phase 1 (`prefetch_accounts`) already covers it on the warmer thread. The CHANGELOG entry still advertises the parallel pre-warm as a feature of this PR. Drop that fragment so users don't go looking for code that isn't there.

Related trade-off note (not blocking, just calling it out): the old in-merkleizer pre-warm was a deterministic synchronous step that guaranteed warm OS pages before Stage B's workers spawned. The new design relies on the warmer thread winning the race against Stage B's startup. On a fresh process where OS page cache is empty, Stage B's first reads can hit cold DB if the warmer's parallel `prefetch_accounts` hasn't ramped up yet. The author measured -0.2 ms / -11% on the bal fixture so the average case is fine, but cold-start scenarios (container scale-up, process restart) have weaker timing guarantees than before.

crates/blockchain/blockchain.rs
```diff
@@ -468,6 +470,24 @@ impl Blockchain {
 
         let cancelled = AtomicBool::new(false);
 
+        // Synthesize BAL updates pre-scope so the merkleizer thread can start
+        // trie work immediately, in parallel with execution.
+        let optimistic_updates: Option<FxHashMap<Address, BalSynthesisItem>> =
+            bal.map(synthesize_bal_updates);
+        let optimistic_witness: Option<Vec<AccountUpdate>> = if self.options.precompute_witnesses {
```
ElFantasma, line 477: **Witness completeness on the optimistic path depends on `state_accessed` catching reads not in the mutation set.** `optimistic_witness` here is built from synthesis's `added_storage` (mutation set only) — no balance / nonce / code_hash, and crucially no storage_reads. The PR description correctly notes that read-only slots come from `DatabaseLogger.state_accessed` populated during execution.

What I want to confirm: `state_accessed` fires on **actual** reads during tx execution. If a BAL declares a `storage_reads` slot that execution doesn't actually exercise (branch not taken, opcode short-circuit, etc.), the slot is in the header BAL but **never** in the witness — neither here (`added_storage` doesn't include it) nor in `state_accessed` (no real read). The safety net is `unread_storage_reads` empty-check at `levm/mod.rs` flagging the unused declaration and rejecting the block.

Worth one of the following:
- Add an inline comment here explicitly stating "reads come from `DatabaseLogger.state_accessed`; a declared-but-unread storage_reads slot is caught by `unread_storage_reads` empty-check at execution time" so future readers don't have to derive the safety argument from scratch.
- Or confirm via an ef-tests blockchain fixture that exercises a header BAL with a declared-but-not-actually-read storage_reads slot, just to make the check load-bearing in CI.

Not blocking — design seems sound — but the implicit invariant is doing a lot of work.

ElFantasma (approved):
Follow-up after the fix series (e7aaec8c → aa5f7ad0). All 3 inline findings from the previous review landed cleanly, and the produced_bal-gate body suggestion was correctly identified as wrong and reverted in `79cf5327c` (the Amsterdam BAL validation path intentionally returns `None`). 

On a fresh-eyes pass over the post-fix code, three small drift items remain — all minor, none blocking, mostly doc-vs-code cleanup that's the natural consequence of an iterative fix series.

ilitteri (approved):

Issue #6572, opened 2026-05-05
[Tracking] Implement missing debug_* RPC endpoints


avilagaston9:
ethrex currently implements a subset of the `debug_*` namespace ([rpc.rs:1037-1048](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/rpc/rpc.rs#L1037-L1048)). Only 5 methods are formally specified ([execution-apis OpenRPC](https://github.com/ethereum/execution-apis/blob/main/src/debug/getters.yaml)); the rest are Geth extensions that downstream tooling (debuggers, simulators) relies on.

Note: hive `rpc-compat` only exercises `debug_getRaw{Block,Header,Transaction,Receipts}`. `debug_getBadBlocks` is in-spec but has no hive fixture.

### Status

**Merged:**
- #6501 — `prestateTracer` for `debug_traceTransaction` / `debug_traceBlockByNumber` (follow-ups: #6577, #6579, #6580)

**In flight:**
- #6595 — `opcodeTracer` / structLogger (default tracer)

**Already in `main` (historical context):**
- #2971 — initial `debug_traceBlockByNumber` with `callTracer`
- #3036 — wired LEVM tracer to RPC endpoint
- #3026 — `debug_executionWitness`
- #6319 — Geth-compatible `callTracer` output format

### Spec-defined (OpenRPC)
- [ ] `debug_getBadBlocks` — [spec](https://github.com/ethereum/execution-apis/blob/main/src/debug/getters.yaml) · [geth](https://github.com/ethereum/go-ethereum/blob/master/eth/api_debug.go)

### Tracing (Geth extensions)
- [ ] `debug_traceCall` — [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-tracecall) · [geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go)
- [ ] `debug_traceBlockByHash` — [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-traceblockbyhash) · [geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go)
- [ ] `debug_traceBlock` (raw RLP) — [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-traceblock) · [geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go)
[...]

PR #6657, opened 2026-05-14
feat(l1): add noopTracer for debug_traceTransaction and debug_traceBlockByNumber

### LINKED ISSUE #6645: Implement noopTracer for debug_trace*
Implement `noopTracer` for `debug_traceTransaction` / `debug_traceBlock*`. Does nothing — useful for benchmarking the trace harness without per-step overhead.

[geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/native/noop.go)

Implementation notes:
- New `TracerType::NoopTracer` variant.
- All hook methods are no-ops.
- Returns `{}` as the result.

Parent: #6572

ElFantasma:
**Motivation**

The `noopTracer` is a standard Geth debug tracer that runs a transaction through the trace harness without recording any per-step data. It is mostly used for benchmarking trace-pipeline overhead and as a smoke test for the tracer dispatch machinery, and several downstream tools expect it to be available.

**Description**

Adds `noopTracer` as a third `TracerType` alongside `callTracer` and `prestateTracer`, wired through `debug_traceTransaction` and `debug_traceBlockByNumber`:

- `LEVM::trace_tx_noop` in `crates/vm/backends/levm/tracing.rs` — executes the tx with `LevmCallTracer::disabled()` (same disabled tracer already used by `trace_tx_prestate`) and discards the result.
- `Evm::trace_tx_noop` in `crates/vm/tracing.rs` — thin façade.
- `Blockchain::trace_transaction_noop` / `Blockchain::trace_block_noop` in `crates/blockchain/tracing.rs` — rebuild parent state, replay earlier txs, run target tx(s) under the noop tracer with the usual timeout wrapping.
- `TracerType::NoopTracer` variant in `crates/networking/rpc/tracing.rs` plus arms in both handlers. Single-tx returns `{}`; block returns `[{ txHash, result: {} }, …]` using the existing `BlockTraceComponent` wrapper.

No new types, no new files, no storage changes.

Closes #6645

**How to Test**

```bash
curl -X POST http://localhost:8545 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"debug_traceTransaction","params":["<TX_HASH>",{"tracer":"noopTracer"}],"id":1}'
# Expected: {"jsonrpc":"2.0","id":1,"result":{}}

curl -X POST http://localhost:8545 -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"debug_traceBlockByNumber","params":["latest",{"tracer":"noopTracer"}],"id":1}'
# Expected: [{"txHash":"0x…","result":{}}, …]
```

**Checklist**

- [x] Tracker issue #6572 updated (sub-issue #6645).


ElFantasma:
Follow-up commit `705073046` folds in two of the points raised:

- Added the `// Run system calls but stop before tx 0` comment in `trace_block_noop` to match the other block tracers.
- Changed `trace_block_noop` to return `Vec<(H256, ())>` for symmetry with `trace_block_calls` / `trace_block_prestate`; RPC handler now uses the existing `BlockTraceComponent` `From` impl.

Skipped intentionally:

- **`block.as_ref()` redundancy** (Claude review): kept verbatim from the call/prestate tracers since `Arc<Block>` already `Deref`s. Cleaning it up cleanly would mean editing all three block tracers, which is out of scope for this PR — happy to do it as a separate cleanup.
- **RPC-level regression tests** (Codex review): no precedent in this crate — neither `callTracer` nor `prestateTracer` have RPC-handler tests, only LEVM-backend tests of their substantive behavior. There isn't substantive behavior to test at the LEVM level for `noopTracer` (it's `vm.execute()` with `LevmCallTracer::disabled()`, which is already exercised by `trace_tx_prestate`). Adding RPC-level harness scaffolding here would set a new precedent for one tracer; better as a separate "add tracer test harness" PR if the team wants it.

Also: kept `#[allow(clippy::enum_variant_names)]` instead of renaming variants to `Call`/`Prestate`/`Noop` — that route would require three per-variant `#[serde(rename = "...")]` attributes to keep the wire format (`"callTracer"`, etc.) intact, which trades one allow for three renames.

ilitteri (approved):
Approving. Verified the implementation against geth's `eth/tracers/native/noop.go` and tested it live on a `--dev` node:

- `debug_traceTransaction` + `noopTracer` → `{"result":{}}`
- `debug_traceBlockByNumber` + `noopTracer` → `[{"txHash":"0x…","result":{}}]`
- empty block → `[]`
- `tracerConfig` supplied alongside `noopTracer` → still `{}` (correctly ignored)
- the tx is genuinely executed (a `callTracer` run of the same tx returns a full frame), and reverts/OOG return `{}` rather than erroring, matching geth's `GetResult` always returning `{}`.

The wire format matches geth's `txTraceResult` (`txHash`/`result`) byte-for-byte.

### Suggestion: add a regression test for the tracer

There's currently no test for this tracer (we have `prestate_tracer_tests.rs` and `opcode_tracer_tests.rs`, but nothing for noop). A small test at the LEVM boundary pins the key geth-alignment contract — `GetResult` is always `{}` regardless of tx outcome, so a **reverting** tx must trace as `Ok` and not surface as an error (i.e. `vm.execute()?` must not propagate the tx outcome). This is cheap insurance against a future refactor turning a revert into an RPC error.

New file `test/tests/levm/noop_tracer_tests.rs`:

```rust
//! End-to-end tests for the noop tracer used by `debug_traceTransaction` /
//! `debug_traceBlockByNumber` with `"tracer": "noopTracer"`.
//!
//! geth's `noopTracer` runs the transaction through the trace harness but records
//! nothing, and `GetResult` always returns the literal JSON `{}` regardless of how
//! the transaction ended (success, revert, out-of-gas). These tests pin that
//! contract at the LEVM boundary: [`LEVM::trace_tx_noop`] must execute the tx and
//! return `Ok(())` (the RPC layer turns that into `{}`), and a reverting tx must
//! NOT surface as an error.

use super::test_db::TestDatabase;
use bytes::Bytes;
use ethrex_common::types::{Account, BlockHeader, Code, EIP1559Transaction, Transaction, TxKind};
use ethrex_common::{Address, U256};
use ethrex_crypto::NativeCrypto;
use ethrex_levm::db::gen_db::GeneralizedDatabase;
use ethrex_levm::vm::VMType;
use ethrex_vm::backends::levm::LEVM;
use once_cell::sync::OnceCell;
use rustc_hash::FxHashMap;
use std::sync::Arc;

// ── Helpers ──────────────────────────────────────────────────────────────

fn default_header() -> BlockHeader {
    BlockHeader {
        coinbase: Address::from_low_u64_be(0xCCC),
        base_fee_per_gas: Some(1),
        gas_limit: 30_000_000,
        ..Default::default()
    }
}

fn make_tx(contract: Address, sender: Address) -> Transaction {
    Transaction::EIP1559Transaction(EIP1559Transaction {
        chain_id: 1,
        nonce: 0,
        max_priority_fee_per_gas: 1,
        max_fee_per_gas: 10,
        gas_limit: 100_000,
        to: TxKind::Call(contract),
        value: U256::zero(),
        data: Bytes::new(),
        access_list: vec![],
        signature_y_parity: false,
        signature_r: U256::one(),
        signature_s: U256::one(),
        inner_hash: OnceCell::new(),
        sender_cache: {
            let cell = OnceCell::new();
            let _ = cell.set(sender);
            cell
        },
        cached_canonical: OnceCell::new(),
    })
}

/// Runs `bytecode` under a contract account through `LEVM::trace_tx_noop`,
/// returning the result so callers can assert on success/failure.
fn run_noop(bytecode: Vec<u8>) -> Result<(), ethrex_vm::EvmError> {
    let contract_addr = Address::from_low_u64_be(0xC000);
    let sender_addr = Address::from_low_u64_be(0x1000);

    let mut accounts = FxHashMap::default();
    accounts.insert(
        contract_addr,
        Account::new(
            U256::zero(),
            Code::from_bytecode(Bytes::from(bytecode), &NativeCrypto),
            1,
            FxHashMap::default(),
        ),
    );
    accounts.insert(
        sender_addr,
        Account::new(
            U256::from(10u64) * U256::from(10u64).pow(U256::from(18)),
            Code::default(),
            0,
            FxHashMap::default(),
        ),
    );

    let mut db = GeneralizedDatabase::new(Arc::new(TestDatabase { accounts }));
    let header = default_header();
    let tx = make_tx(contract_addr, sender_addr);

    LEVM::trace_tx_noop(&mut db, &header, &tx, VMType::L1, &NativeCrypto)
}

// ── Tests ──────────────────────────────────────────────────────────────────

/// A normally-completing tx (`PUSH1 0x01 PUSH1 0x02 ADD STOP`) traces with no error.
/// The RPC layer maps this `Ok(())` to `{}`.
#[test]
fn noop_tracer_successful_tx_returns_ok() {
    let bytecode = vec![0x60, 0x01, 0x60, 0x02, 0x01, 0x00];
    run_noop(bytecode).expect("noop trace of a successful tx should return Ok");
}

/// A reverting tx (`PUSH1 0x00 PUSH1 0x00 REVERT`) must still trace successfully:
/// geth's noopTracer returns `{}` regardless of tx outcome, so a revert must not
/// surface as an RPC error. Guards against `trace_tx_noop` propagating the tx
/// outcome through `vm.execute()?` (reverts are reported in the execution report,
/// not as `Err`).
#[test]
fn noop_tracer_reverting_tx_returns_ok() {
    let bytecode = vec![0x60, 0x00, 0x60, 0x00, 0xfd];
    run_noop(bytecode).expect("noop trace of a reverting tx should still return Ok");
}
```

And register it in `test/tests/levm/mod.rs`:

```rust
mod noop_tracer_tests;
```

Both tests pass locally (`cargo test -p ethrex-test --test ethrex_tests -- levm::noop`), and `cargo clippy` is clean.

### Optional, non-blocking

`trace_tx_noop` uses `LevmCallTracer::disabled()`, so — unlike geth's noop, which attaches empty hooks and still dispatches into them per opcode — it doesn't exercise per-opcode hook dispatch. The result is `{}` either way, so this is purely about what it measures: it won't reflect trace-harness overhead the way geth's noop does. The doc comment "the trace harness overhead is exercised" slightly overstates this. Fine to leave as-is given ethrex's tracing is opt-in per tracer.

ElFantasma:
Pushed `f0d4542` which brings in the strengths of @azteca1998's now-closed #6694 (the parallel noopTracer effort), so we don't lose that work:

- **Tests** — ported Esteve's full coverage from #6694: the integration suite (`test/tests/rpc/debug_trace_tests.rs`: `trace_tx_noop_tracer`, `…_unknown_hash_errors`, `trace_block_noop_tracer`) plus the RPC-layer parse/deserialize unit tests in `crates/networking/rpc/tracing.rs`. This PR previously shipped with no tests.
- **Simpler whole-block path** — adopted his single-pass `rerun_block(None)` for `trace_block_noop` (returning `Vec<H256>`), replacing the per-tx loop + `Arc<Mutex>` + per-tx VM reconstruction. Cleaner and a more faithful benchmark for the noop path, since it doesn't pay per-tx `setup_env`/`VM::new` overhead.
- **Doc** — clarified on the `NoopTracer` variant that the output matches geth's `noopTracer` (`{}` per tx) but the mechanism differs: geth runs an empty-hook tracer (per-opcode dispatch), whereas we skip tracer hooks entirely, isolating raw execution cost.

Verified locally: `cargo fmt` / `cargo check` / `cargo clippy --all-targets` clean; 9 unit + 3 integration tests pass (incl. the block path against the new single-pass impl).

Credit to @azteca1998 for the tests and the whole-block design — #6694 was the more complete of the two; this folds its good parts back in.

azteca1998 (approved):

ElFantasma:
Thanks @ilitteri — both points addressed:

**Regression test** (`33e61fead`): added `trace_tx_noop_tracer_reverting_tx` in `test/tests/rpc/debug_trace_tests.rs`. It seeds a contract with `PUSH1 0x00 PUSH1 0x00 REVERT` runtime code, calls it, and asserts `debug_traceTransaction` + `noopTracer` returns `{}` rather than erroring. This pins exactly the contract you called out — `vm.execute()?` must not propagate the reverting tx outcome as an `Err`. I put it at the RPC boundary alongside the existing noop tests (rather than a new `test/tests/levm/noop_tracer_tests.rs`) so it also exercises the full `trace_transaction_noop` → LEVM path; happy to move it if you'd prefer the LEVM-boundary placement.

**Doc comment** (`9649364b4`): you're right that `LevmCallTracer::disabled()` skips per-opcode hook dispatch, so "trace harness overhead is exercised" overstated it. Reworded to say the tx still runs (errors surface) but no per-opcode hooks are dispatched — unlike geth's noop, which attaches empty hooks and still calls into them.

PR #6660, opened 2026-05-14
feat(l1,l2): add State Override Set and Block Override Set to RPC simulation methods

ilitteri:
**Motivation**

ethrex rejects geth's State Override Set and Block Override Set on every RPC simulation method: `eth_call`, `eth_estimateGas` and `eth_createAccessList` cap out at two params and return `BadParams` for a third, and `debug_traceCall` captures `stateOverrides`/`blockOverrides` only to reject them explicitly. Tooling that integrators rely on — Foundry's `forge script --rpc-url`, Tenderly-style simulators, MEV simulators, wallet "what-if" balance previews, Multicall3 aggregators — assumes these parameters work. Closing the gap is a parity decision: tools that work against a geth RPC must work against an ethrex RPC.

Spec: https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#state-override-set

**Description**

Implements both override sets across all four simulation methods, on L1 and L2.

**State Override Set** — `eth_call`, `eth_estimateGas`, `eth_createAccessList` (3rd param), `debug_traceCall` (nested, see below). Per-address `balance`, `nonce`, `code`, `state` (replace storage), `stateDiff` (overlay storage), `movePrecompileToAddress`.

**Block Override Set** — `eth_call`, `eth_estimateGas` (4th param), `debug_traceCall` (nested). Fields `number`, `time`, `gasLimit`, `feeRecipient`, `prevRandao`, `baseFeePerGas`, `blobBaseFee` (back-computed by inverting `fake_exponential`, rounding down for a fee that is not representable), `difficulty`.

Names follow geth's `override.BlockOverrides`. That struct carries no `json` tags, so Go marshals it under its field names and unmarshals case-insensitively, which puts two spellings of every field in circulation; both are accepted, along with every other spelling in use, so a request shaped for any client works: geth's `feeRecipient`/`prevRandao` and the PascalCase `FeeRecipient`/`PrevRandao` Go emits from geth's own type, **and** the older `coinbase`/`random` that alloy still treats as canonical; geth's `baseFeePerGas` and alloy's `baseFee`; erigon's `blockNumber`/`timestamp`. The State Override Set's account fields do carry `json` tags upstream, and its parser matches keys case-insensitively as Go does. `beaconRoot`, `withdrawals` and reth/alloy's `blockHash` extension are declared but **refused with a reason** — see below.

**L1 + L2 parity** — L2 inherits all four methods through the existing `_other_eth_method` fall-through in `crates/l2/networking/rpc/rpc.rs`; the L2 sponsor flow's `EstimateGasRequest` literals gain `..Default::default()` for the new fields.

### Implementation shape

- **`OverlaidVmDatabase<Inner>`** (`crates/blockchain/vm.rs`) decorates any `VmDatabase` with per-address overrides — a six-method intercept, so nothing below the RPC layer needs to know overrides exist.
- **`Blockchain::new_overlaid_evm`** is the only way to build that overlay. An override set has two effects that live in different places — the account overlay in the `VmDatabase`, `movePrecompileToAddress` in EVM *dispatch* — and a two-step API let the second be forgotten at a call site. One constructor installs both, so the omission isn't representable.
- **`movePrecompileToAddress` reaches dispatch.** `PrecompileMoves` rides on `GeneralizedDatabase` and is consulted at all three dispatch sites (top-level in `vm.rs`, the CALL family in `opcode_handlers/system.rs`). It is `None` on every consensus path, where precompile dispatch stays bit-identical to the static table.
- **`StateOverrideSet::into_overrides` validates the moves**, for the same reason `new_overlaid_evm` owns the overlay: it is the one function every override-carrying endpoint must call to obtain a usable map, so the checks cannot be skipped at a call site — and it is where `BadParams` is reachable, rather than surfacing as a VM (-32015) or internal error. Deciding whether an address is a precompile needs the fork and the `VMType`, so `Blockchain::vm_type()` exposes the latter without building an `Evm`, and callers pass the fork of the *effective* header, since a `time` override can cross a fork boundary and change the precompile set.
- **`Blockchain::new_evm` is generalized** from concrete `StoreVmDatabase` to `D: VmDatabase + 'static` so the overlay composes through it. Public-API change touching every caller — worth a look.
- **`TraceCallOverrides`** (`crates/blockchain/tracing.rs`) carries both sets into `Blockchain::trace_call_*`, which builds the tracing EVM itself.
- **Perf**: `eth_estimateGas` converts the override set once per request instead of once per binary-search step — `into_overrides` hashes every override code blob, and the search runs up to ~64 times.

### `debug_traceCall` takes the overrides in geth's shape

Geth carries them as *fields of the 3rd config object* (`TraceCallConfig` embeds `TraceConfig`), not as separate positional params. Both are accepted, nested winning when both are present; positional params 4 and 5 remain as a backwards-compatible ethrex extension.

`TraceConfig` deliberately keeps accepting unknown fields, because geth embeds its struct-logger config at that level and a legitimate client may send `enableMemory`/`disableStack` there. The fix for silently-dropped overrides is to *accept* the fields, not to reject unknown ones. `BlockOverrideSet`, by contrast, now sets `deny_unknown_fields` to match `StateOverrideSet`, so a typo or geth's unmodelled `withdrawals` is an error rather than a silent drop.

### How the overrides compose with `txIndex`

A State Override Set applies on both of `debug_traceCall`'s state paths. Tracing on top of a block whose post-state is stored reads it straight out of the trie, as `eth_call` does. With `txIndex`, or on a block that needs re-execution, the blockchain layer rebuilds the parent state and re-runs the block's own transactions — and those transactions must not see the override, because they really happened and executed against the state they really saw.

So the replay finishes in its own `Evm`, is materialised as a **`ReplayedVmDatabase`** (a `VmDatabase` view of the finished replay's `AccountUpdate`s), and the overlay goes on top of *that*, for a second `Evm` that runs only the traced call. That is geth's ordering — `StateAtTransaction` first, `StateOverride.Apply` second — and reth's, `replay_block_until` then `prepare_call_env`. `GethDebugTracingCallOptions` carries `txIndex`, `stateOverrides` and `blockOverrides` in one struct with no mutual exclusion, and neither does this. A trace that carries no overrides keeps running on the replay's own `Evm`, since there is nothing to layer over it and the projection reconstructs `storage_root` rather than carrying it.

`test/tests/blockchain/trace_call_replay_tests.rs` is where the composition is proven at the state level: the same state reached two ways (`txIndex == tx_count` against the stored post-state) must trace identically, with overrides and without, and a separate test pins that the override is invisible to the replayed transactions. An earlier revision of this branch refused the combination instead, and the RPC-level tests now pin that the parser accepts it.

`beaconRoot` and `withdrawals` are **refused**, which is what geth's own `BlockOverrides.Apply` does with them. Either would need the block's system contracts to run, and no simulation path runs them: setting `parentBeaconBlockRoot` on the header alone leaves the EIP-4788 ring buffer untouched, so a contract reading `BEACON_ROOTS_ADDRESS` would not observe it. To simulate that, override the beacon-roots contract's storage through the State Override Set. `blockHash` is refused by name too, so the error says it is a reth/alloy extension rather than just "unknown field".

### Geth semantics

| Rule | Behavior |
|---|---|
| `state` + `stateDiff` together for one address | `BadParams` at parse time |
| Unknown field in a State or Block Override Set | `BadParams` naming the field. A deliberate deviation: geth ignores it, and a silently dropped override returns a plausible-looking answer computed against inputs the caller never asked for |
| `code` override on a 7702-delegated EOA | Override wins |
| `movePrecompileToAddress` | Destination dispatches; source becomes an ordinary account |
| Any overridden address that is a precompile | Stops dispatching as one, matching geth's `delete(precompiles, addr)` — including for a `balance`-only override |
| `movePrecompileToAddress` naming a non-precompile | `BadParams` |
| `movePrecompileToAddress` onto an address that is itself overridden | `BadParams` |
| `BLOCKHASH(n)` at or above the base block's number, under a synthetic future block | Zero, matching geth's `GetHashFn`, which is built from the real header *before* the overrides are applied (`ref.Number.Uint64() <= n`). For *either* override set, since it is a property of the synthetic block. LEVM's own 256-block window is measured against the synthetic number, as geth's `opBlockhash` is |
| `blobBaseFee` on a fork with no blob schedule | `BadParams` — there is no update fraction to invert |
| `blobBaseFee` between two representable fees | Rounds **down**, to the largest fee at or below the request, so a `maxFeePerBlobGas` set to the same value is never undercut. geth assigns the field directly and needs no inversion. Zero is the one request that cannot be honoured downwards: EIP-4844 floors the blob base fee at 1 |
| `beaconRoot` / `withdrawals` / `blockHash` | `BadParams` explaining why, matching geth's `BlockOverrides.Apply` |
| `stateOverrides` + `txIndex` on `debug_traceCall` | Both apply, and the override is invisible to the block's own transactions |
| `eth_createAccessList` with a 4th param | `BadParams`. The 3rd param is itself an ethrex extension: geth's `CreateAccessList` takes two params and no override sets at all |
| Mixed-case override key | Tolerated |
| A quantity without the `0x` prefix | Read as hex, as for every other quantity ethrex's RPC layer parses (`serde_utils` trims the prefix optionally), so `"30000000"` is taken for `0x30000000`. geth's `hexutil` requires the prefix; tightening it belongs in `serde_utils`, for the whole RPC surface at once |
| Empty override object `{}` | No-op |

### Test coverage

Unit coverage for the two override sets' parse semantics, `OverlaidVmDatabase`, `ReplayedVmDatabase` and per-endpoint param parsing, plus 26 integration tests over the full JSON-RPC HTTP dispatch, so an override that parses but never reaches the EVM fails the test:

- `test/tests/rpc/state_override_tests.rs` — `eth_call` with `code`, `balance` (read back via SELFBALANCE) and `state` (via SLOAD) overrides, a no-override control, and the `eth_estimateGas` paths.
- `test/tests/rpc/precompile_move_tests.rs` — the identity precompile at `0x04` echoes its calldata, so a relocation either dispatches or it doesn't with no ambiguous middle ground. Covers all three dispatch sites: top-level via `eth_call`, top-level via `debug_traceCall`, and the CALL family via a `STATICCALL` from overridden bytecode — plus a `balance`-only override silencing a precompile, with a control proving it still dispatches otherwise.
- `test/tests/rpc/trace_call_tests.rs` — the nested geth config shape for both sets, callTracer, prestateTracer (which must report the *overridden* pre-state), opcodeTracer, and the `txIndex` refusal including its error *code*, so a request-shape error cannot regress into `-32603 Internal`.
- `crates/networking/rpc/types/block_override.rs` — geth's current field names, alloy's and erigon's aliases, and each refused field naming itself in the error.

Every fix in the last review round is mutation-verified: reverting it fails exactly its own test. Where a test passed on first write it was mutation-checked rather than trusted — the `estimateGas` short-circuit guard, the precompile-move install sites, and the two patched dispatch expressions in `system.rs` are each individually guarded, and the `system.rs` pair discriminate (reverting the address resolution fails only the destination case, not the vacated one).

### Notes for reviewers

- **A prior review round found five defects, all fixed** (`af66a890be`), each reproduced with a failing test first. Two are worth knowing as a class: `eth_estimateGas` had a re-simulation path that carried no overrides, and the `BLOCKHASH` clamp was reachable through only one of the two override sets. The override sets must reach *every* site that reads state, and that set of sites is not stable — the first of those two arrived from `main` while this branch was adding overrides, so neither side conflicted and the merge produced a clean but wrong program. If this area grows further, a `SimulationContext` owning both sets and required by the simulate/estimate entry points would turn that into a compile error.
- **The merge with `main` dropped work from this branch.** `main` had implemented `debug_traceCall` independently and more completely (`trace_call_calls`/`trace_call_prestate`/`trace_call_opcodes`, whole-block variants, `txIndex`, nonce filling), including a deliberate choice to skip the gas-allowance check rather than raise `block_gas_limit`, on the grounds that a trace must report the GASLIMIT the traced block executed with. This branch's duplicate `*_from_generic` entry points are gone in favour of those, and the override support is built on that architecture. `main`'s two tests asserting the override sets are rejected are replaced by their opposite.
- **`movePrecompileToAddress` now matches geth's three semantics**, confirmed against `internal/ethapi/override/override.go`. Worth calling out because one of them changes what an innocuous-looking request does: geth's `delete(precompiles, addr)` runs for **any** overridden address, so `{"0x04": {"balance": "0x1"}}` takes the identity precompile out of the active set and leaves an ordinary account. That reads like a no-op and isn't. The other two are validation — `"account %s is not a precompile"` and `"account %s is already overridden"` — and both replaced silent no-ops. Rejecting the second means dispatch never has to arbitrate between a relocated precompile and a code override at the same address.

- **No CHANGELOG entry.** `CHANGELOG.md` has only `## Observability` and `## Perf` sections and every entry in its history comes from a `perf(...)` or perf-adjacent `fix(...)` PR; no feature PR adds one. Adding an entry would mean inventing a `## Features` section. Happy to add one if that's wanted.

### Second review round — what changed

Six commits, each with the geth reference it was checked against. The first four change behaviour and every one is mutation-verified: reverting the fix fails exactly its own test and nothing else.

1. **`blobBaseFee` rounds down.** The module docs said down, the search and its own doc said up. Down is what a caller needs, since the usual reason to set the field is to pair it with a `maxFeePerBlobGas` of the same value, and rounding up rejects that call for undercutting a fee it asked for. The old round-trip test passes either way and now says so; the new one asks for a fee one wei under a representable step.
2. **geth's own field spellings are accepted.** `BlockOverrides` has no `json` tags, so a client built on geth's type sends `Number`/`FeeRecipient`/`BaseFeePerGas`, and `deny_unknown_fields` rejected all of them. The State Override Set's visitor now matches keys case-insensitively, as Go does.
3. **`BLOCKHASH` cuts off at the base block, not the chain tip.** geth builds its hash function from the real header before applying the overrides. The two rules agree on `latest`, which is why the tests could not tell them apart; they diverge on a historical block with `number` moved forward, where the old rule served real canonical hashes for the blocks in between. Taking the number from the header each caller already has removes the tip plumbing and a `get_latest_block_number` per request.
4. **An override-free `txIndex` trace stays on the replay's own `Evm`.** The `ReplayedVmDatabase` projection reconstructs `storage_root` from an `AccountUpdate`, so it is now confined to the requests that actually ask for an override. Not distinguishable by a test — the two routes are meant to agree, and the differential test asserts they do.
5. **The overlay forwards `code_cache_budget_bytes`** instead of falling through to the trait default, which is documented as the floor for a backend holding no cache.
6. **Three comments corrected**, including the one review question left open: a `state`-only override on an absent account leaves it absent while its slots still read from the overlay, and geth is not observably different, because the object its `SetStorage` creates is empty and both clients gate what the EVM sees on emptiness rather than presence.

Two things deliberately not changed, both recorded in comments: the `0x` prefix stays optional, because tightening it in these two files alone would leave a block override's `gasLimit` refusing what the transaction object's own `gas` accepts, and `.gitignore` keeps the unrelated `docs/superpowers/` entry from the merge that brought it in.

### Third review round — what changed

One correctness fix and five cleanups, against the review at `a46ab68`.

1. **`storage_root` now follows the storage override.** It was passed through from the chain, and no slot read consults it — but `LevmAccount::from` turns it into `has_storage`, and the EIP-7610 collision check reads that. So `{"code":"0x","nonce":"0x0","state":{}}` left an account the EVM could see was empty in every way it can inspect, while a simulated CREATE2 onto it still aborted with `AddressAlreadyOccupied`. The root is now derived from the storage mode: `state` is a closed world and reports only what its own map holds, `stateDiff` keeps whatever the chain had underneath. The reviewer flagged the geth half as unverified; it holds. `SetStorage` installs an object from `NewEmptyStateAccount`, whose `Root` is `EmptyRootHash`, so geth's own collision check sees no storage and does not abort. Covered by the reviewer's repro end to end, plus a control proving the fixture collides without the override.
2. **The doc block above `recap_with_account_balances` is reattached.** `convert_state_overrides` had been inserted between the two, which left the balance-capping function undocumented and gave the new one a doc describing a `fee_cap` it does not take.
3. **The Block Override Set is applied once per request.** The caller already builds the synthetic header, so it is threaded down instead of rebuilt inside `simulate_tx_with_overrides`. That also removes the build that was computed and discarded for a call carrying only `blockOverrides`. Adjacent, as suggested: the converted override map is now shared by `Arc` rather than deep-copied on every step of the gas search.
4. **`ReplayedVmDatabase::new`'s dedup doc said the wrong direction.** The map overwrites, so the last update for an address wins, not the first.
5. **`needs_overlay` no longer fires for a Block Override Set that cannot need it.** Only a forward `number` move can reach the overlay's `BLOCKHASH` clamp, since LEVM measures its own window against the synthetic number. A `time`-only override with `txIndex` no longer takes the trace through the replay projection.

Minor: the wrappers now say why their batch methods keep the trait defaults, and what to know before wiring them up. Two suggestions were considered and left alone. The redundant zero-denominator guard stays, as cheap defensiveness in a function whose caller is the only thing standing between it and a division it cannot do. The code-hash lookup map was not added, since the scan is per simulation rather than per read and the reviewer measured it as negligible; a second source of truth for the same data costs more than it saves here.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A, no storage changes.


crates/networking/rpc/types/block_override.rs
```diff
@@ -0,0 +1,200 @@
...
+                .unwrap_or(0);
+            header.excess_blob_gas = Some(invert_blob_base_fee(desired, denom));
+        }
+        // Force hash recomputation by replacing the OnceCell.
+        header.hash = Default::default();
+        header
+    }
+}
+
+/// Binary-search the smallest `excess_blob_gas` whose `fake_exponential`-derived
```
ElFantasma, line 95: Rounding-direction doc/code inconsistency. The file-level comment (lines 11-13) says

> for values that fall between representable steps the result rounds down to the closest fee ≤ requested

but this function's doc says it returns the smallest `excess_blob_gas` whose derived fee is **≥ desired**, and the binary search below implements exactly that (loop invariant: `compute(hi) >= desired`, `compute(lo-1) < desired`). So the simulated `BLOBBASEFEE` ends up ≥ the user's requested value, not ≤ — the file-level doc looks wrong.

Semantically ≥ is probably not what a simulator caller wants ("simulate my tx as if blob fee were X" is the usual framing — Foundry, MEV sims), and geth just uses the header field directly with no rounding, so any deviation is our own. Would you prefer to (a) update the file-level doc to say the returned fee is ≥ requested, or (b) flip direction so the derived fee stays ≤ requested? Either is defensible; the current mix is the confusing part.

crates/blockchain/vm.rs
```diff
@@ -14,6 +15,194 @@ use std::{
...
+        }
+        // Synthesize an account if the address is unknown on chain but the override
+        // gives it state. Other overrides (e.g. movePrecompileToAddress only) keep
+        // the account absent.
+        let mut state = match base {
+            Some(s) => s,
+            None if ov.balance.is_some() || ov.nonce.is_some() || ov.code.is_some() => {
+                AccountState::default()
+            }
+            None => return Ok(None),
```
ElFantasma, line 119: Edge case, non-blocking: if an override sets only `storage_mode` (Replace/Diff) on an address with no on-chain account and no balance/nonce/code override, `is_noop()` returns false so we hit this match, and this `None => return Ok(None)` branch fires — but `get_storage_slot` still returns the synthetic storage values from the overlay. The account appears absent to the EVM (`EXTCODESIZE == 0`, `SELFDESTRUCT`/`EXISTS` semantics), but its storage is populated. Probably not a real caller shape, but worth confirming that's the intended geth behavior (or promoting a storage-only override to also synthesize a default account by including `storage_mode.is_some()` in the branch above).

MegaRedHand (commented):
Reviewed at `a46ab68`. One substantive correctness issue, a few cleanups, and two findings I raised and then withdrew after checking them (listed at the end, so they don't get re-raised).

---

## 1. `OverlaidVmDatabase` passes `storage_root` through, which leaks into the EIP-7610 create check

`crates/blockchain/vm.rs:142`

```rust
// storage_root is left untouched: the wrapper intercepts get_storage_slot
// directly, so the EVM never observes a storage_root that has to match.
Ok(Some(state))
```

The comment is true for slot reads, but `storage_root` has a second consumer that never goes through `get_storage_slot`:

```rust
// crates/vm/levm/src/account.rs:78, :83
has_storage: state.storage_root != *EMPTY_TRIE_HASH,
exists: !is_default,   // is_default compares the whole AccountState, root included
```

and `create_would_collide()` (`account.rs:113`) is `has_code() || has_nonce() || has_storage`, reached from `execution_handlers.rs:151` and `opcode_handlers/system.rs:995`. The path is live for all four override-carrying RPCs: `gen_db.rs:454` does `LevmAccount::from(self.store.get_account_state(address)?)`, and `self.store` is the overlay.

So a `state:` override that empties storage is invisible to the collision check. The reachable repro combines fields, which makes `has_code`/`has_nonce` false and leaves the stale real root as the only thing voting:

```json
{"0xA": {"code": "0x", "nonce": "0x0", "state": {}}}
```

`synthetic_code(Bytes::new())` gives `EMPTY_KECCAK_HASH` (`vm.rs:421`), every slot reads 0 (`vm.rs:154`), so the EVM sees an account that is empty in every way it can inspect, and a simulated CREATE2 onto it still aborts with `AddressAlreadyOccupied`.

Suggested fix: derive `storage_root` from `ov.storage_mode` in `get_account_state` instead of passing it through. `Replace(empty)` gives `EMPTY_TRIE_HASH`; `Replace(non-empty)` and `Diff(non-empty)` give a non-empty sentinel. Note this also makes the wrapper agree with `ReplayedVmDatabase`, whose own doc already says `has_storage`/`exists` read the root; right now the two wrappers in this PR document opposite positions on whether it is observable.

One caveat: I verified the ethrex side only. Whether geth's `SetStorage` resets `data.Root` so `GetStorageRoot` returns `EmptyRootHash` I did not check against geth source, so please treat the "diverges from geth" half as unconfirmed.

The mirror case (`stateDiff` on an absent account, where the account stays absent but slots still answer) I did check, and it is by design: `vm.rs:110-125` documents it and argues it matches geth. No action there.

---

## 2. Doc block detached from the function it describes

`crates/networking/rpc/eth/transaction.rs:789-805`

`convert_state_overrides` was inserted between the `recap_with_account_balances` doc block and `recap_with_account_balances`. There is no blank line at 797/798, so lines 789-804 now read as one rustdoc on `convert_state_overrides`, ending with "...which is the same split this function's `fee_cap` argument exists to close". `convert_state_overrides` has no `fee_cap` and performs no division; `recap_with_account_balances` (818), which has both, is now undocumented. On `main` the same text sits at `transaction.rs:675-683` directly above it.

---

## 3. The Block Override Set is applied twice per request

`transaction.rs:164-167` builds `effective_header`, and `simulate_tx_with_overrides` rebuilds the identical header at `:900-903` from the raw set passed down at `:180`. Same in `eth_estimateGas` (`:590-593`).

The sharpest case: when `stateOverrides` is absent or empty, `convert_state_overrides` returns at `:811` *before* reading `effective_header`, so for an `eth_call` carrying only `blockOverrides` the entire `:164-167` build, `invert_blob_base_fee` included, is computed and thrown away.

Threading the already-built header into `simulate_tx_with_overrides` removes both. Two corrections to my first pass, for the record: the cost is smaller than I implied (the estimateGas loop is bounded at ~18 iterations for realistic block gas limits, not ~64, so ~21 `apply_to` calls and ~630 `fake_exponential` evaluations, on the order of tens of microseconds against ~20 full EVM simulations), and the duplicated `BadParams` is not user-visible since `:165` propagates with `?` before the second call. Redundancy worth removing for clarity, not a latency problem.

Adjacent: `:916` does `state_overrides.cloned()`, deep-copying the map and its per-address storage maps once per search step. An `Arc` would avoid it, though that needs `new_overlaid_evm` (`blockchain.rs:4219`) to stop taking the map by value.

---

## 4. `ReplayedVmDatabase::new` documents the dedup direction backwards

`crates/blockchain/vm.rs:260-267`. The doc says "A second update for the same address is silently discarded rather than merged", but `updates.into_iter().map(|u| (u.address, u)).collect()` into an `FxHashMap` overwrites: the *first* is discarded and the last wins. Harmless under the stated precondition, but it points a reader the wrong way if `get_state_transitions` ever stops guaranteeing one entry per address.

---

## 5. `needs_overlay()` is broader than it needs to be

`crates/blockchain/tracing.rs:43`. It returns true for any Block Override Set, so `debug_traceCall` with `txIndex` plus a `time`-only override goes through `ReplayedVmDatabase`'s reconstructed root, which the comment at `:363-366` argues the override-free `txIndex` trace should avoid for exactly the same reason.

Only a *forward* `number` move actually needs the overlay: `get_block_hash` (`vm.rs:166`) clamps at `base_block_number`, and `opcode_handlers/block.rs:44` already gates on the synthetic number, so with the number unchanged or moved down the clamp never fires. A `time`/`coinbase`/`difficulty` override reaches the env and the fork, never the database layer.

Low severity: the sentinel's wrong answers fail closed, and the loss only applies on the re-execution path. Mentioning it because the code right above already makes the argument.

---

## Minor

- **`vm.rs:190`** Neither wrapper forwards the batch `VmDatabase` methods, so `get_account_codes_batch` falls back to the default, which turns `StoreVmDatabase`'s deliberate `Ok(None)` (`vm.rs:804`) into `Err` (`vm.rs:769`). Currently unreachable twice over: the override paths never build a `CachingDatabase`, and `prefetch_codes`, the sole consumer, has no call sites anywhere in the tree. Latent trap rather than a bug; a comment or a forwarding impl would keep it that way.
- **`block_override.rs:228`** `if denominator == 0 { return 0; }` sits under a `debug_assert` for the same condition, and `apply_to` (`:193-204`) already rejects a zero fraction with `BadParams`. Also removing it would not change release behavior: `fake_exponential` errors on a zero denominator, `.unwrap_or(U256::MAX)` at `:236` makes every probe read as too expensive, and the search returns 0 anyway. Redundant, not load-bearing.
- **`vm.rs:184-206`** The linear override scan in `get_account_code`/`get_code_metadata` is `O(overrides × distinct code hashes)` per simulation, not per read (`gen_db.rs:481-516` memoizes by hash), and there is no cardinality limit on `StateOverrideSet` beyond the 256 MB body limit. Negligible in practice; an `FxHashMap<H256, &Code>` built in `new` would be cheap hardening. Correcting my own earlier "every bytecode read" wording here.
- **`transaction.rs:629`** `has_overrides` is recomputed from the raw fields while `state_overrides.is_some()` already encodes the state half. The two are equivalent today, so this is a maintainability nit only. My earlier framing was wrong: `state_overrides` carries nothing about `block_overrides`, so the second disjunct is load-bearing and cannot be dropped.
- **`rpc/tracing.rs:473-487`** duplicates the fork + `into_overrides` sequence from the module-private `convert_state_overrides`. I checked and the copies do not disagree today. The shared part is a two-line body, so centralizing it is optional.

---

## Withdrawn after checking

- **Hex deserializers in `block_override.rs`.** I flagged `trim_start_matches("0x")` as drift from `serde_utils::{u64,u256}::hex_str_opt`. It is not: the shared helpers use the identical call (`serde_utils.rs:129`, `:264`), so switching would fix neither `"0x0x10"` nor unprefixed input, and it would loosen acceptance (the shared `u64` helper also takes a JSON number). `state_override.rs:189-195` already documents this as deliberate and says tightening belongs in `serde_utils` for the whole RPC surface at once. Agreed.
- **The import under the re-export comment in `blockchain.rs:68`.** The comment was already detached on `main` (`main:67-69`), so this PR adds a line to a pre-existing mismatch rather than creating one.

MegaRedHand (approved):

PR #6663, opened 2026-05-15
feat(l1): add statetest subcommand to ef_tests-statev2 for goevmlab fuzzing

### LINKED ISSUE #6575: [Fuzzing] Add `statetest` CLI subcommand for goevmlab differential fuzzing

ElFantasma:
**Motivation**

goevmlab is the standard differential-fuzzing harness for EL clients (geth, nethermind, besu, erigon, nimbus, evmone, eels, revm). It runs the same EF state test through every configured EVM and diffs their per-opcode (EIP-3155) traces and post-state roots. To plug ethrex into that fleet, goevmlab needs a tiny CLI surface on ethrex that emits EIP-3155 trace JSONL plus a `stateRoot` line on stderr. This PR provides that.

**Description**

Adds a `statetest` subcommand to the existing `ef_tests-statev2` binary. Two modes:

- `statetest --json <fixture>`: emits EIP-3155 per-opcode JSONL on stderr, followed by one `{"stateRoot":"0x..."}` line per `(fork, post-index)` case in the fixture.
- `statetest --json-outcome <fixture>`: same but without the per-opcode trace — only the `stateRoot` line(s).

Output goes exclusively to stderr (stdout reserved for crash diagnostics, matching geth/revm convention). Exit code `0` means every case's computed post-state root matched the fixture's expected root; `1` means at least one mismatch (tolerated by goevmlab); any other non-zero / panic indicates an actual crash.

The subcommand reuses the existing fixture parser, env/tx builders, and post-state-root computation from `state_v2`. The per-opcode trace is generated by `LevmOpcodeTracer` (added in #6595) — for each test case we activate the tracer, execute the tx, and serialize each `OpcodeStep` as one JSONL line. The geth-`structLogger` wrapper used by the RPC adapter is sidestepped: goevmlab expects raw JSONL.

Also rounds out the `state_v2` fixture deserializer with the fork mappings for `Osaka`, `Amsterdam`, and `BPO1`–`BPO5`. These variants are already in the `Fork` enum but were missing from the deserializer's match arm, so any fixture's `post` map containing them aborted parsing (the unmodified strict-error behavior is preserved for truly unknown fork names — adding a new variant remains a build break that has to be addressed).

Backwards-compatible: running the binary without a subcommand still triggers the existing bulk runner. Branch stacked on top of #6595 (`feat/eip-3155-tracer`) since the implementation depends on its `LevmOpcodeTracer`.

Closes #6575

**How to Test**

```bash
cd tooling

# full per-opcode trace + stateRoot per case
cargo run --release -p ef_tests-statev2 -- statetest --json \
  ef_tests/state/vectors/state_tests/cancun/eip5656_mcopy/test_mcopy_on_empty_memory.json \
  2> trace.jsonl
echo "exit=$?"
head -3 trace.jsonl     # per-opcode JSONL: {"pc":..., "op":..., ...}
grep stateRoot trace.jsonl   # final stateRoot line(s)

# outcome only
cargo run --release -p ef_tests-statev2 -- statetest --json-outcome \
  ef_tests/state/vectors/state_tests/cancun/eip5656_mcopy/test_mcopy_on_empty_memory.json \
  2>&1
```

**Checklist**

- [x] Sub-issue of the `debug_*` / fuzzing tracker (#6572).


ElFantasma:
Follow-up commit `fa34b00bc` addresses the three concrete bugs flagged across the reviews:

**Fixed**

- **Exit-code collision between mismatches and internal errors** (Greptile P1, Codex High, Claude Medium). `main` now returns `ExitCode` directly. Any `RunnerError` propagated from the `statetest` path maps to `ExitCode::from(2)`; deliberate state-root mismatches stay at `1`; success stays at `0`. goevmlab can now distinguish a tolerated differential from an actual crash.
- **Silent false-green on Osaka/BPO fixtures** (Codex Medium). `types.rs::DEFAULT_FORKS` only allows `Merge`/`Shanghai`/`Cancun`/`Prague`/`Amsterdam`, so adding Osaka/BPO1–5 to the deserializer made those entries *parse* but `Tests::from` silently dropped them, leaving `test_cases` empty and producing exit 0 with no `stateRoot` emitted. `statetest::run` now errors (→ exit 2) when every `test_cases` list is empty.
- **`--json` / `--json-outcome` required-one-of constraint** (Claude Minor, Greptile P2). `StatetestOptions` now uses `#[group(required = true, multiple = false)]` so clap rejects the invocation at parse time with a clear usage message, instead of the previous runtime `RunnerError::Custom` whose message was also misleading for the `(None, None)` case.

**Not addressed (with rationale)**

- **`stateRoot` zero-padding** (Claude Medium): non-bug. `H256`'s `LowerHex` impl writes every byte as `{:02x}`, always 64 hex chars regardless of leading zeros. Confirmed in the smoke test outputs — every line emitted is exactly `0x` + 64 chars.
- **`vm.db.clone()` before `get_state_transitions`** (Kimi, Claude, Codex): pre-existing pattern, mirroring `result_check::check_root`. Greptile correctly noted "no change needed now". A real refactor target but unrelated to this PR's scope.
- **Compound `|=` assignment nit** (Claude): adopted in the new commit alongside the other changes (`any_mismatch |= run_case(...)?;`).
- **`expectException` not checked** (Codex High): valid concern for direct EF-fixture use, but out of scope for the goevmlab CLI. goevmlab compares state roots across implementations and never feeds us `expectException`-bearing fixtures (its generator produces simple txs). The bulk runner still enforces this check via `result_check::check_exception`.
- **txbytes vs re-signed tx** (Codex Open Question): re-using the existing `get_tx_from_test_case` helper that re-signs matches what `state_v2`'s bulk runner does and what the issue's spec requested. Re-using raw `txbytes` is a separate ticket about test-runner accuracy, independent of the goevmlab interface.
- **Flattened `RunnerOptions` conflict** (Claude Risk): `RunnerOptions` has no required fields, both subcommand and no-subcommand invocations work in smoke tests. Speculative; not reproducible.

ElFantasma:
## Verification + parser hardening update

Rebased on the latest [`feat/eip-3155-tracer`](https://github.com/lambdaclass/ethrex/pull/6595) tip and verified end-to-end that the wire output matches what goevmlab's `gen_oplog.go` unmarshaler expects.

### Commits now on the branch

1. `feat(l1): add statetest subcommand to ef_tests-statev2 for goevmlab fuzzing` — the original implementation.
2. `fix(l1): distinguish statetest internal errors from root mismatches, enforce --json/--json-outcome at clap level` — exit-code semantics + clap `ArgGroup`.
3. `feat(l1): emit EIP-3155 wire shape from statetest sink via Eip3155Step wrapper` — switches the per-step serializer to the new `Eip3155Step` newtype landed in #6595. Strict EIP-3155 shape: numeric `op` byte + separate `opName`, `"0xN"` hex `gas`/`gasCost`/`refund`, `stack: []` (never null) when disabled.
4. `fix(l1): propagate fixture parse errors instead of panicking` — `parser::parse_file` / `parse_dir` previously `.unwrap()`-ed on file-not-found, malformed JSON, etc., producing an exit-101 panic that goevmlab would read as a real client crash. Now returns `RunnerError::ParseFixture { path, source }` which the main handler maps to exit code 2 with a clear message.

### Verification matrix

Run against `cancun/eip5656_mcopy/test_mcopy_on_empty_memory.json` (16 cases, 288 trace steps, 16 stateRoot lines):

| Check | Result |
|---|---|
| `--json` emits per-op JSONL + stateRoot to stderr | exit 0, 304 lines |
| `--json-outcome` emits only stateRoot to stderr | exit 0, 16/16 stateRoot lines |
| stdout silent in both modes | ✓ |
| Field types per step: `pc int, op int (0-255), opName str, gas "0xN", gasCost "0xN", refund "0xN", memSize int, depth int, stack [], returnData "0x..."` | 10/10 fields correct |
| goevmlab `gen_oplog.go`-style UnmarshalJSON parse (Python mimic of `HexOrDecimal64` and `vm.OpCode`) | every line round-trips |
| stateRoot lines match the literal byte pattern goevmlab scans for (`"stateRoot":"0x<64 hex>"`) | ✓ |
| Storage accumulation across SSTOREs (inherited from #6595) | SSTORE #0: 1 slot · #1: 2 slots · #2: 3 slots · … |
| Missing fixture file | exit 2 (was exit 101) |
| Malformed JSON | exit 2 (was exit 101) |
| No-flag invocation | clap rejects with exit 2 + helpful message |

Exit-1 (post-state root mismatch) path is straightforward (`any_mismatch |= run_case(...)`) and exercised by its dual (root-match → exit 0) above. Provoking a real mismatch needs an intentionally wrong fixture which we don't have lying around; happy to add a synthetic regression test if reviewers want it.

### Example output

```bash
$ cargo run --release -p ef_tests-statev2 -- statetest --json fixture.json 2> trace.jsonl
$ head -2 trace.jsonl
{"pc":0,"op":96,"gas":"0xeee94","gasCost":"0x3","memSize":0,"stack":[],"depth":1,"returnData":"0x","refund":"0x0","opName":"PUSH1"}
{"pc":2,"op":53,"gas":"0xeee91","gasCost":"0x3","memSize":0,"stack":["0x40"],"depth":1,"returnData":"0x","refund":"0x0","opName":"CALLDATALOAD"}
$ grep -c stateRoot trace.jsonl
16
```

`cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

iovoid (approved):

ilitteri (approved):

azteca1998 (approved):

PR #6664, opened 2026-05-16
fix(l1): resolve memory limit crash on zisk

benbencik:
**Motivation**
In the [zkEVM Benchmark Results](https://eth-act.github.io/zkevm-benchmark-runs/benchmarks/), we have seen some crashing EEST tests. A common problem is hitting the memory limit on zkVM. This change resolves previously crashing tests:
```
test_mod_arithmetic[fork_Osaka-blockchain_test-opcode_MULMOD-mod_bits_127-benchmark-gas-value_60M]
test_mod_arithmetic[fork_Osaka-blockchain_test-opcode_MULMOD-mod_bits_191-benchmark-gas-value_60M]
test_mod_arithmetic[fork_Osaka-blockchain_test-opcode_MULMOD-mod_bits_255-benchmark-gas-value_60M]
test_ecrecover[fork_Osaka-blockchain_test-ecrecover-benchmark-gas-value_60M]
test_msize[fork_Osaka-blockchain_test-mem_size_100000-benchmark-gas-value_60M]
test_return_revert[fork_Osaka-blockchain_test-1KiB of non-zero data-opcode_RETURN-benchmark-gas-value_60M]
test_return_revert[fork_Osaka-blockchain_test-1KiB of non-zero data-opcode_REVERT-benchmark-gas-value_60M]
test_return_revert[fork_Osaka-blockchain_test-1KiB of zero data-opcode_RETURN-benchmark-gas-value_30M]
test_return_revert[fork_Osaka-blockchain_test-1KiB of zero data-opcode_RETURN-benchmark-gas-value_60M]
test_return_revert[fork_Osaka-blockchain_test-1KiB of zero data-opcode_REVERT-benchmark-gas-value_30M]
test_return_revert[fork_Osaka-blockchain_test-1KiB of zero data-opcode_REVERT-benchmark-gas-value_60M]
```
**Description**
The ZisK zkVM uses bump allocator by default. This allocator never frees, for simplicity and efficiency. In Ethrex, when a callframe finishes, the function `clean_from_base` zeros the respective region. However, the memory capacity under the bump allocator grows monotonically. The changes propose adding `truncate_to_base` that shrinks the length of the vector such that the capacity is reused. 

I do not have a complete understanding of how the default allocator handles this in the "standard execution", but the choice to not truncate seems to be intentional. Therefore, this change is gated behind the `eip-8025` feature.


ilitteri (changes requested):
The fix looks ok, but if this is a zkVM constraint, I'd prefer to use `#[cfg(target_os = "zkvm")]` or `#[cfg(target_arch = "riscv64")]`.

benbencik:
> The fix looks ok, but if this is a zkVM constraint, I'd prefer to use `#[cfg(target_os = "zkvm")]` or `#[cfg(target_arch = "riscv64")]`.

Yes, I agree. It seems that the proposed RISCV target triple [specifies OS as `none`](https://github.com/eth-act/zkvm-standards/blob/main/standards/riscv-target/target.md). I changed the code to use `#[cfg(target_arch = "riscv64")]`.

ilitteri (approved):

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -1133,8 +1133,12 @@ impl<'a> VM<'a> {
             ..
         } = executed_call_frame;
 
+        #[cfg(not(target_arch = "riscv64"))]
         old_callframe_memory.clean_from_base();
 
+        #[cfg(target_arch = "riscv64")]
+        old_callframe_memory.truncate_to_base();
```
ElFantasma, line 1140: **Top-level frame memory isn't truncated at tx end.** The two call sites here (line 1136-1140 + 1289-1293) cover the CALL-return and CREATE-return paths — sub-frames. But at top-level tx finalization (the outermost call frame), neither `clean_from_base` nor `truncate_to_base` is invoked in this diff. So for a single long tx that grows memory to 1 MiB and never makes a sub-call, the bump allocator still holds 1 MiB at tx end and into the next tx.

Is the per-tx Memory replaced rather than reused? Looking at `Memory::default()`/`new()` would clarify. If a fresh Memory is allocated per tx (drops the old buffer), the bump allocator only needs to be poked on sub-frame returns — current code is correct. If Memory is reused across txs, top-level cleanup is missing.

Worth a one-line comment confirming which model, so a future reader doesn't add a redundant truncate at top-level (or realize the gap and add the missing one). Non-blocking — the 11 fixed tests presumably exercise sub-call cleanup which is what's covered.

crates/vm/levm/src/memory.rs
```diff
@@ -54,6 +54,15 @@ impl Memory {
         }
     }
 
+    /// Truncates the memory back to base. This is crucial for constrained
+    /// memory in zkVMs. The memory is not freed, but rather shrunk in `len`,
+    /// so that the already allocated `capacity` is reused.
+    #[cfg(target_arch = "riscv64")]
```
ElFantasma, line 60: nit: no unit test for `truncate_to_base` in this PR. The function is small (`self.buffer.borrow_mut().truncate(self.current_base)`) and behavior-critical (the whole point of the PR), but it lives inside `#[cfg(target_arch = "riscv64")]` which most developer machines can't compile.

A `#[cfg(test)]` test that exercises the SAME truncate behavior via a non-cfg-gated test helper (or by inverting the gate just for the test) would lock the contract: "after truncate, `len()` equals `current_base`, `capacity()` is unchanged, and reading from `[current_base..old_len]` returns whatever was there before (since we didn't zero)."

The 11 failing→passing benchmark tests confirm the fix in aggregate, but they're upstream EEST fixtures, not local. Lock the function-level contract locally so a future refactor of `clean_from_base` / Memory layout can't silently break it.

ElFantasma (approved):
PR description says "this change is gated behind the `eip-8025` feature" but the actual `#[cfg(...)]` is `target_arch = "riscv64"` (line 60 of memory.rs, lines 1136-1140 + 1289-1293 of system.rs). These aren't equivalent: `target_arch = "riscv64"` fires on **any** RISC-V backend (zisk, risc0, openvm), while `feature = "eip-8025"` would only fire when the EIP-8025 feature is enabled. The current gate is broader than the body suggests — probably the right call (any RISC-V bump-allocator benefits), but worth either updating the body to match the code, or tightening the gate if zisk-only was the actual intent.

Also two small inline observations.

azteca1998 (approved):

PR #6665, opened 2026-05-18
feat(l1): add in-process engine API ef_tests runner

edg-l:
Add `tooling/ef_tests/engine`, an in-process drop-in for hive's `consume-engine` simulator. Instead of running ethrex in a Docker container behind HTTP+JWT, it drives `ethrex_rpc::rpc::map_engine_requests` / `map_http_requests` directly via `serde_json` deserialization. The serde+dispatch round-trip hive exercises is preserved; HTTP framing and JWT auth are out of scope.

Runs **5609 EEST JSON fixture files / 82,818 inner fixtures, all green in ~50s on an 8-core machine**.

Key implementation notes:

- Shared `SyncManager` via `tokio::sync::OnceCell` in `engine_only_context` keeps OS-thread count bounded to one per process.
- `merkle_pool` is `thread_local!` per tokio worker, not process-wide. A process-wide rayon pool deadlocks: the merkle protocol uses `in_place_scope` with 16 concurrent jobs that cross-broadcast via channels, starving all but one caller.
- Added `Blockchain::default_with_store_and_pool(store, Arc<ThreadPool>)` constructor; existing `new` / `default_with_store` are unchanged, so production is unaffected.
- `src/exception_mapper.rs` ports the Python `EthrexExceptionMapper` from `execution-specs/.../clis/ethrex.py` (~42 entries). No automated drift check yet; README warns.
- `ETHREX_ENGINE_LIMIT=<regex>` mirrors hive's `--sim.limit` for fixture-name filtering.
- `#[ctor::dtor]` prints a final colored summary (`fixture result: ok. N passed; ...`) after libtest's per-file output so the inner 82k count is visible.
- Makefile mirrors `tooling/ef_tests/blockchain` conventions (`.fixtures_url`, `.fixtures_url_amsterdam`, extracts `blockchain_tests_engine/`).

## TODOs (not blockers)
- Parity spot-check against real hive on ~25 fixtures.
- Nightly RocksDB-backed CI job.
- Document recommended `RUST_TEST_THREADS` for CI runners.
- CI drift check comparing `src/exception_mapper.rs` against the upstream Python source.
- Evaluate replacing 3 of 11 daily hive matrix entries (Consume Engine Rest/Prague/Amsterdam in `.github/workflows/daily_hive_report.yaml`) with this crate.

## Test plan
- `make -C tooling/ef_tests/engine test` runs all 5609 fixture files locally.
- `ETHREX_ENGINE_LIMIT=<pattern>` for targeted runs.
- Sample fixtures in `tooling/ef_tests/engine/fixtures_sample/` for smoke testing without downloading the full suite.


azteca1998 (approved):

tooling/ef_tests/engine/src/engine_ctx.rs
```diff
@@ -0,0 +1,95 @@
...
+    let shared_syncer = SHARED_SYNCER
+        .get_or_init(|| async { Arc::new(dummy_sync_manager().await) })
+        .await
+        .clone();
+    let blockchain = Arc::new(Blockchain::default_with_store_and_pool(
+        storage.clone(),
+        thread_local_merkle_pool(),
+    ));
+    let local_node_record = example_local_node_record();
+    let block_worker_channel = start_block_executor(blockchain.clone());
```
ElFantasma, line 67: **`start_block_executor(blockchain.clone())` spawns an OS thread per fixture.** This function (rpc.rs `start_block_executor`) creates a new `std::thread::Builder::new()` per call. Combined with `engine_only_context` being called per-fixture via `EngineApiHarness::from_genesis`, you get one OS thread spawn + termination per fixture × 82,818 fixtures.

Linux thread spawn is ~10-100µs, so that's ~1-8 seconds out of the ~50s run that's pure spawn/teardown overhead. The thread terminates when the `block_worker_channel`'s sender (`UnboundedSender` field on `RpcApiContext`) drops, which happens when `RpcApiContext` drops at end of each `run_fixture`. Correct lifecycle, but expensive.

Two cheap improvements:
1. **Pool block executors** via a `LazyLock` similar to `SHARED_SYNCER`: one thread per process, fed by a multi-sender channel where each context registers a per-fixture reply path. Per-fixture thread spawn collapses to zero.
2. **`thread_local!` block_executor** matching the `THREAD_LOCAL_MERKLE_POOL` pattern: one per tokio worker thread (`num_cpus` total). Cheaper than pooling, doesn't need MPSC routing.

Low priority — current shape works. But if RocksDB nightly + larger fixture sets ever push this past a minute, this is the first place to look.

tooling/ef_tests/engine/src/engine_ctx.rs
```diff
@@ -0,0 +1,95 @@
...
+    },
+};
+use ethrex_storage::Store;
+use tokio::sync::{Mutex as TokioMutex, OnceCell};
+
+/// Shared SyncManager for `engine_only_context`. Allocated once per process so the
+/// RLPxInitiator OS thread (spawned by `dummy_actor::spawn_on_thread`) is created
+/// exactly once regardless of how many harnesses are built.
+///
+/// Verified unused in the engine and eth/block handler paths:
```
ElFantasma, line 28: **Invariant verification is rg-only, not enforceable.** The comment says "Verified unused in the engine and eth/block handler paths: `rg context.peer_handler` -> empty". That's accurate today, but it's an extra-code check that's easy to forget across refactors. If a future engine handler grows a `ctx.peer_handler.as_ref().unwrap()` access (e.g., for peer-aware payload builds, hive simulator parity, …), the shared `None` peer_handler panics — and the failure surfaces only at fixture-test time, not at the offending handler's compile time.

Two cheap hardenings:
1. **`#[deny]` lint or grep check in CI**: add a CI step that runs `! rg "ctx\.peer_handler|context\.peer_handler" crates/networking/rpc/engine/ crates/networking/rpc/eth/block.rs` and fails the build if it produces output. Two lines in `.github/workflows/`.
2. **Compile-time guard via a marker trait**: a sealed `EngineOnlyContext` trait that prevents peer_handler access at type-check time. More invasive but durable.

Option 1 is the cheap fix; the comment turns into a CI-enforced invariant.

tooling/ef_tests/engine/tests/all.rs
```diff
@@ -0,0 +1,143 @@
...
+
+// Aggregate fixture counters across all `engine_runner` calls. libtest reports
+// one test per JSON file, but each file holds many fixtures; these atomics
+// surface the inner totals in a final line printed at process exit.
+static F_PASSED: AtomicUsize = AtomicUsize::new(0);
+static F_FAILED: AtomicUsize = AtomicUsize::new(0);
+static F_SKIPPED: AtomicUsize = AtomicUsize::new(0);
+static F_FILTERED: AtomicUsize = AtomicUsize::new(0);
+
+#[ctor::dtor]
```
ElFantasma, line 22: nit: `#[ctor::dtor]` on `print_fixture_summary` is cosmetic-only (prints aggregate fixture counts at process exit). The `ctor` crate's documented caveats:
- destructor ordering across `dtor` functions is undefined
- accessing other statics in `dtor` is undefined behavior territory if they've been dropped

This function only touches `'static` atomics (`F_PASSED`, `F_FAILED`, etc.) plus stdlib I/O (`eprintln!`, `is_terminal()`). Both safe at exit on Linux. But the function is `pub` (default visibility) and called automatically — any future addition to this file that loads more statics during `print_fixture_summary` could hit dragons.

Worth a `#[ctor::dtor]` doc-comment saying "DO NOT touch non-`'static` state from here; ordering is undefined." Or wrap the body in a `std::panic::catch_unwind` so a misbehaving future addition doesn't kill the test process exit code.

Low priority; current code is safe.

ElFantasma (approved):
Only some minor comments.

The TODO list has good shape. At least two of these — "Parity spot-check against real hive on ~25 fixtures" and "Evaluate replacing 3 of 11 daily hive matrix entries" — are the validation steps that prove this harness is honest.

Worth landing them in a follow-up issue (linked from the PR description) so they're tracked, not just buried in the body.

ilitteri (approved):

PR #6666, opened 2026-05-18
feat(l1): improve rlp import

edg-l:
## Summary

- Adds `--export-bal <FILE>` and `--with-bal <FILE>` flags to `import-bench` (`cmd/ethrex/cli.rs`). `--export-bal` runs sequential execution and writes one RLP-encoded `BlockAccessList` per Amsterdam+ block into a single file; `--with-bal` decodes that file and feeds BALs into the parallel execution path.
- Replaces the magic-number `tokio::time::sleep(500ms)` between blocks in `import_blocks_bench` with an explicit `Store::wait_for_persistence_idle()` call. Uses a new `TrieMessage::Ping` over the `sync_channel(0)` rendezvous channel: a successful send proves the trie-update worker finished its Phase 2 disk write and returned to `recv()`.
- No production/consensus code paths are affected; changes are scoped to bench tooling and one new `Store` method.

## Import speed improvement

The previous inter-block `sleep(500ms)` is replaced by a ping-based wait that resolves in ~3-10ms (typical Phase 2/3 drain time).

For a 200-block import, total inter-block waiting:
- Old: 200 x 500ms ~= **100s**
- New: 200 x ~5ms ~= **<1s**

So ~99 seconds saved in pure waiting per 200 blocks. The per-block timer is also now accurate -- it no longer absorbs the previous block's Phase 2/3 background persistence cost.

## Notes

Cherry-picked from two feature branches:
- `bal-benchmark-setup` -- `--export-bal` / `--with-bal` plumbing
- `perf/bal-parallel-overhead-rebased` -- persistence-idle wait

Isolated here so they can ship as a focused tooling PR without dragging in unrelated work from either parent branch.

## Test plan

- [ ] \`cargo fmt --all\`
- [ ] \`cargo check -p ethrex\`
- [ ] Manual round-trip: run \`import-bench --export-bal /tmp/bals.rlp\`, then \`import-bench --with-bal /tmp/bals.rlp\` and verify parallel execution path is exercised
- [ ] Confirm no sleep regression: bench should not stall between blocks without the old sleep

### DIFF
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -518,13 +518,13 @@ pub enum Subcommand {
         #[arg(
             long = "export-bal",
             value_name = "FILE",
-            help = "Export BALs produced during sequential execution to a single RLP file (concatenated, one per block)"
+            help = "Export BALs produced during sequential execution to a single RLP file (concatenated, one per block). All BALs are buffered in memory before writing; suitable for benchmark-sized runs, not 100k+ block exports."
         )]
         export_bal: Option<String>,
         #[arg(
             long = "with-bal",
             value_name = "FILE",
-            help = "Load BALs from a single RLP file and use the parallel execution path",
+            help = "Load BALs from a single RLP file and use the parallel execution path. The entire file is decoded into memory upfront; suitable for benchmark-sized runs, not 100k+ blocks.",
             conflicts_with = "export_bal"
         )]
         with_bal: Option<String>,
@@ -926,31 +926,12 @@ pub async fn import_blocks_bench(
     let store = init_store(datadir, genesis).await?;
     let blockchain = init_blockchain(store.clone(), blockchain_opts);
     regenerate_head_state(&store, &blockchain).await.unwrap();
-    let path_metadata = metadata(path).expect("Failed to read path");
+    let path_metadata =
+        metadata(path).unwrap_or_else(|e| panic!("failed to stat path {path:?}: {e}"));
 
     if let Some(bal_path) = export_bal_path {
         info!(path = %bal_path, "Will export BALs to file");
     }
-
-    // Pre-load all BALs into memory upfront to avoid per-block I/O during benchmark
-    let preloaded_bals = if let Some(bal_path) = with_bal_path {
-        info!(path = %bal_path, "Loading BALs from file (parallel path)");
-        use ethrex_common::types::block_access_list::BlockAccessList;
-        use ethrex_rlp::decode::RLPDecode as _;
-        let data = std::fs::read(bal_path).expect("Failed to read BAL file");
-        let mut remaining = data.as_slice();
-        let mut bals = Vec::new();
-        while !remaining.is_empty() {
-            let (bal, rest) =
-                BlockAccessList::decode_unfinished(remaining).expect("Failed to decode BAL");
-            bals.push(bal);
-            remaining = rest;
-        }
-        info!(count = bals.len(), "Loaded BALs into memory");
-        Some(bals)
-    } else {
-        None
-    };
 
     // If it's an .rlp file it will be just one chain, but if it's a directory there can be multiple chains.
     let chains: Vec<Vec<Block>> = if path_metadata.is_dir() {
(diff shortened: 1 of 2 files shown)


azteca1998 (approved):

ilitteri (approved):

cmd/ethrex/cli.rs
```diff
@@ -965,21 +1017,44 @@ pub async fn import_blocks_bench(
...
-                .inspect_err(|err| match err {
-                    // Block number 1's parent not found, the chain must not belong to the same network as the genesis file
-                    ChainError::ParentNotFound if number == 1 => warn!("The chain file is not compatible with the genesis file. Are you sure you selected the correct network?"),
-                    _ => warn!("Failed to add block {number} with hash {hash:#x}"),
-                })?;
+            // Look up preloaded BAL for this block (if --with-bal was provided).
+            // BALs are only produced for Amsterdam+ blocks, so use a separate counter
+            // that only advances for blocks that have a BAL hash in the header.
+            let bal = if block.header.block_access_list_hash.is_some() {
+                let b = preloaded_bals.as_ref().and_then(|bals| bals.get(bal_index));
```
ElFantasma, line 1041: **Silent BAL truncation when `--with-bal` file has fewer entries than Amsterdam blocks.** `preloaded_bals.get(bal_index)` returns `None` once `bal_index` exhausts the vec — then `bal = None` and the block falls through to the sequential path silently. A user who passed a mismatched (truncated) BAL file would see a mixed sequential/parallel run instead of a clear error, and the perf numbers would be misleading.

Three cheap options:
1. **Upfront count check** — before the loop, count Amsterdam blocks across all chains (`block.header.block_access_list_hash.is_some()`) and assert it equals `preloaded_bals.len()`. Surfaces the mismatch before any wasted work.
2. **Per-block warn** — when `bal_index >= preloaded_bals.len()` on an Amsterdam block, `warn!(block = number, "--with-bal exhausted at block N, falling through to sequential")`.
3. **Hard error** — `assert!(bal_index < preloaded_bals.len(), …)` and bail. Probably too strict if the user knowingly truncates.

Option 1 is the cleanest (fail-fast, clear message). Option 2 is the cheap belt-and-suspenders.

crates/storage/store.rs
```diff
@@ -270,6 +270,23 @@ pub struct AccountUpdatesList {
...
+    /// Block until the trie-update background worker has drained every prior
+    /// message and is waiting for new work — i.e. Phase 2 (disk write of the
+    /// bottom-most diff layer) and Phase 3 (in-memory layer removal) for all
+    /// previously-applied updates have completed.
+    ///
+    /// Implementation: the worker channel is `sync_channel(0)`, so a send only
+    /// returns once the worker calls `recv()` on the next loop iteration.
+    /// `TrieMessage::Ping` carries no work, so the send completing is itself
+    /// the idle signal.
+    pub async fn wait_for_persistence_idle(&self) -> Result<(), StoreError> {
```
ElFantasma, line 288: **Document the single-producer expectation.** The doc-comment correctly explains the rendezvous-channel proof of "worker drained prior message and is back at `recv()`". But under concurrent senders, the guarantee is weaker than the docstring implies: if thread A sends `Update` and thread B (our caller) sends `Ping` concurrently, `sync_channel(0)` serializes them — but the order isn't guaranteed. If Ping wins the race, the worker no-ops while a new `Update` from thread A is still in-flight after `wait_for_persistence_idle` returns.

For the bench-tooling use case there's only one producer (the main import loop), so this is moot. But the method is `pub` and a future caller (e.g. on the engine-API path) could land in a multi-producer context.

Worth adding one line:
```rust
/// Caller's responsibility: hold off other senders to `trie_update_worker_tx`
/// while this is in flight. Under concurrent producers the rendezvous
/// guarantee degrades to "the prior message has been drained", not
/// "persistence is idle going forward."
```

Not blocking — current usage is single-producer and safe.

cmd/ethrex/cli.rs
```diff
@@ -905,6 +928,30 @@ pub async fn import_blocks_bench(
...
+    if let Some(bal_path) = export_bal_path {
+        info!(path = %bal_path, "Will export BALs to file");
+    }
+
+    // Pre-load all BALs into memory upfront to avoid per-block I/O during benchmark
+    let preloaded_bals = if let Some(bal_path) = with_bal_path {
+        info!(path = %bal_path, "Loading BALs from file (parallel path)");
+        use ethrex_common::types::block_access_list::BlockAccessList;
+        use ethrex_rlp::decode::RLPDecode as _;
+        let data = std::fs::read(bal_path).expect("Failed to read BAL file");
```
ElFantasma: nit: `std::fs::read(bal_path).expect("Failed to read BAL file")` — when this panics (e.g. wrong path, perm denied), the error message doesn't include the path, so the user gets a generic "Failed to read BAL file: <io error>" without knowing which path was attempted (especially if `--with-bal` was set via env var or shell substitution).

Cheap fix: `.unwrap_or_else(|e| panic!("failed to read BAL file at {bal_path:?}: {e}"))`, or convert to a proper `Result` and propagate. Same applies to the `std::fs::write(bal_path, …).expect("Failed to write BAL file")` at line ~1081 and the two `.expect("Failed to decode BAL")` / `.expect("Failed to read path")` calls.

For benchmark tooling the panic is acceptable; just the error message is the suggestion.

ElFantasma (approved):
Both modes batch-load everything into RAM. `--export-bal` accumulates `exported_bals: Vec<BlockAccessList>` across all blocks before writing; `--with-bal` does `preloaded_bals: Vec<BlockAccessList>` upfront. For a 200-block bench this is fine. For 100k+ blocks on a long chain, total memory could reach multiple GB. Worth a comment in the help text noting the in-memory bound, or a follow-up to stream both sides via length-prefixed framing.

PR #6667, opened 2026-05-18
perf(l1): make BAL recorder checkpoint O(1) via journal snapshots

edg-l:
## Summary

`BlockAccessListRecorder::checkpoint()` previously walked the entire `storage_writes` / `balance_changes` / `nonce_changes` / `code_changes` / `reads_promoted_to_writes` maps and built fresh `BTreeMap` clones each call. With one checkpoint per CALL/STATICCALL/CREATE frame, cost was **O(call_frames × total_writes_in_block_so_far)** on the builder path. As the block accumulates state, every later checkpoint pays for everything before it ; the worst case scales with block state.

Replaced with append-only journals:

- Each push to a change map also appends `(addr [, slot], prev_len)` to a parallel `Vec`.
- `checkpoint()` = **O(1)** snapshot of 5 journal lengths.
- `restore()` walks `journal[checkpoint_len..]` in reverse, truncating each affected per-address container back to its `prev_len`. **O(reverted_entries)**, independent of the block's accumulated state.

Storage-write slots whose container becomes empty after restore (= fresh writes since the checkpoint, `prev_len == 0`) are promoted to `storage_reads` per EIP-7928. `tx_restore` skips the promotion since rejected txs must leave no trace.

Follow-up commit also swaps `reads_promoted_to_writes` from `BTreeMap<Address, Vec<U256>>` to `BTreeMap<Address, IndexSet<U256>>` so the dedup check in `record_storage_write` is O(1) instead of O(promoted_slots_per_address).

## Why this matters

The worst case in `O(call_frames × accumulated_writes)` no longer scales with block state. On a realistic-mix workload (`bal-devnet-7-mainnet-mix-460`, ~30 Mgas/block, shallow call stacks) the steady-state win is modest (~3% on the builder path) because the block-accumulated map stays small. On stress workloads ; deep nested DeFi calls with many storage writes per frame, or EIP-7928 `test_bal_max_storage_write`-style fixtures specifically designed to maximize checkpoint count × write count ; the previous implementation balloons quadratically. With this change it stays flat.

## Design notes

**Journal staleness**. `filter_net_zero_storage` / `filter_net_zero_code` mutate the change maps in `set_block_access_index`, which would invalidate outstanding journal entries' `prev_len` values. We clear all 5 journals after the filter pass when the index advances. This is safe because no checkpoint can outlive a tx boundary in the current call flow:

- Inner-call `checkpoint`s are consumed within the tx ; success commits them, revert restores them.
- `tx_checkpoint` is consumed before the next `set_block_access_index` ; success commits the tx, failure calls `tx_restore`.

**Call-order invariant**. `tx_checkpoint` must be taken **after** `set_block_access_index`, not before ; `set_block_access_index` clears journals, so checkpointing first would record a journal length the post-clear journal can never grow back to. This matches the existing `payload.rs` and `execute_block_pipeline` call order. A `debug_assert!` in both `restore` and `tx_restore` catches the violation (journal shrank between checkpoint and restore).

**`track_selfdestruct`** mutates the change maps but only runs in `finalize_execution` (end of successful tx, no checkpoint outstanding); the next `set_block_access_index` clears journals before any new checkpoint could observe stale entries.

## Tests

13 unit tests in `checkpoint_restore_tests`:

- No-op restore (checkpoint then restore with nothing in between)
- Balance / nonce / code change rollback
- Fresh-write → read promotion (EIP-7928 semantics)
- Prior-write preservation (slot must not appear in both writes and reads)
- Multiple writes to same slot after checkpoint
- Post-checkpoint write to an already-promoted slot exercises the IndexSet dedup path
- Nested checkpoints (inner restore then outer restore)
- Read-promoted-to-write rollback
- `tx_restore` does NOT promote emptied writes to reads
- `set_block_access_index` clears journals on advance
- Journal walk preserves correct per-address ordering with interleaved pushes

## Parity verification

The recorder is used on two paths:

- **Builder** (`payload.rs::build_payload`): full per-block recorder that produces the BAL written to the block header.
- **Parallel validation** (`execute_block_parallel`): per-tx shadow recorder; checks recorded accesses match the header BAL.

Both call `checkpoint`/`restore` on inner-call frames, so any divergence in checkpoint semantics would surface as `BAL validation failed for tx N` errors. The bench round-trip exercises both:

```
ethrex import-bench --export-bal …   # builder path with record_* + checkpoint/restore + build()
ethrex import-bench --with-bal  …    # parallel-path shadow recorder vs the BAL from above
```

Ran the round-trip on `bal-devnet-7-mainnet-mix-460` (460 blocks). Zero validation errors.

## Test plan

- [x] `cargo test -p ethrex-common checkpoint_restore_tests` (13/13)
- [x] `make -C tooling/ef_tests/blockchain test` (8819 / 8819)
- [x] `cargo fmt --check`
- [x] Builder ↔ validator round-trip on `bal-devnet-7-mainnet-mix-460`: no BAL mismatches
- [ ] Hive `bal` group


azteca1998 (approved):

crates/common/types/block_access_list.rs
```diff
@@ -1349,89 +1389,138 @@ impl BlockAccessListRecorder {
...
-                changes.truncate(len);
-                len > 0
+    /// Like the storage_writes walk in `restore`, but does NOT promote
+    /// emptied slots to `storage_reads`. Used by `tx_restore` where a rejected
+    /// tx should leave no trace at all.
+    ///
+    /// Cleanup of `reads_promoted_to_writes` for the rejected tx is delegated
+    /// to the caller (`tx_restore` walks `reads_promoted_journal` separately
+    /// before invoking this helper).
+    fn walk_storage_writes_journal_no_promote(
```
ElFantasma, line 1554: The promote-variant inline walk in `restore` (lines ~1414-1438) and `walk_storage_writes_journal_no_promote` here share the structural skeleton:

```
while journal.len() > target: pop → get_mut(addr) → get_mut(slot) → truncate → maybe remove slot → maybe remove addr
```

The single semantic delta is whether emptied slots get promoted into `storage_reads`. A unified helper with a callback or `promote: bool` parameter (or even `Option<&mut StorageReads>`) would centralize the loop and make the promotion contract a single function-level decision rather than spread across two siblings:

```rust
fn walk_storage_writes_journal(
    journal: &mut Vec<(Address, U256, usize)>,
    target_len: usize,
    storage_writes: &mut StorageWritesMap,
    promote_to: Option<&mut IndexMap<Address, IndexSet<U256>>>,
) { /* ... */ }
```

Non-blocking — code duplication is small and readable as-is. But the next time a third variant lands (e.g. SELFDESTRUCT cleanup wants to feed through the same shape), the unified form pays off.

crates/common/types/block_access_list.rs
```diff
@@ -1491,62 +1594,47 @@ impl BlockAccessListRecorder {
...
     /// Matches EELS `track_selfdestruct` in state_tracker.py:315.
+    ///
+    /// INVARIANT: must be called only at top-level tx finalization, with no
+    /// outstanding `checkpoint()` pending restore. The `retain` calls below
+    /// mutate the change maps WITHOUT pushing journal entries, so any later
+    /// `restore()` against an older checkpoint would observe a journal whose
+    /// `prev_len`s no longer match the underlying map and silently produce
+    /// incorrect results. Caller is `LEVM::finalize_execution`, which runs
+    /// after every nested call frame has already committed or restored.
     pub fn track_selfdestruct(&mut self, address: Address) {
```
ElFantasma: `restore` and `tx_restore` got runtime-enforced invariants via `debug_assert!(journal.len() >= checkpoint_len, ...)`. `track_selfdestruct` has the analogous invariant ("no outstanding checkpoint, or the `retain` mutations below silently invalidate journal `prev_len`s") documented but **not** enforced.

The invariant here is harder to express precisely — there's no direct "is a checkpoint outstanding?" flag — but a proxy assertion is feasible. Since the only legitimate caller is `LEVM::finalize_execution` running after all nested frames have committed/restored, and `set_block_access_index` is the next thing to happen, any uncleared journal at this point should be the natural accumulation since the last index advance. A weaker but still useful check:

```rust
// All inner-call checkpoints from this tx must have been consumed before
// finalize. We can't detect outstanding tx_checkpoint, but a parallel
// `#[cfg(debug_assertions)]` counter or a callsite assertion in default_hook.rs
// would catch the misuse.
```

The alternative is to drop the doc-only invariant and just push proper journal entries from `track_selfdestruct`'s `retain` paths (one per element removed), so the function becomes journal-safe and the constraint disappears. That's a bigger refactor — non-blocking.

At minimum, mirror the doc invariant text inline at the single call site in `default_hook.rs::finalize_execution` so the constraint is visible from both ends — currently a reader at the call site sees nothing.

ElFantasma (approved):

ilitteri (approved):

PR #6668, opened 2026-05-18
fix(l1): handle skipped intermediate forks in next_fork

edg-l:
## Summary
- `ChainConfig::next_fork` returned `None` on networks that skip intermediate forks (e.g. Osaka @0, no BPOs scheduled, Amsterdam later) because the old if-chain required every fork's immediate successor to be scheduled to advance.
- That made `eth_config` (EIP-7910) report `next: null` on bal-devnet-7, which dora flags as 6 missing fields when comparing client vs expected network config.
- Replaced the if-chain with a scan over timestamp-based forks, picking the one with the smallest activation strictly greater than `block_timestamp`.

## Repro (before fix)
Run any client against bal-devnet-7 (Shanghai/Cancun/Prague/Osaka @0, Amsterdam @1779098127, no BPOs) and call `eth_config` at the genesis block. Old code falls through to the `prague_activated && osaka_time.is_some()` branch returning `Some(Osaka)`, then the final guard `fork > self.fork(ts) == Osaka > Osaka` rejects it.

## Test plan
- [x] `cargo test -p ethrex-common --lib next_fork` (3 new unit tests cover bal-devnet-7, contiguous schedules, and end-of-schedule)
- [x] `cargo test -p ethrex-rpc --lib eth_config` (existing Hoodi-based integration test still passes)
- [ ] Run against bal-devnet-7 kurtosis enclave and confirm dora shows 0 missing fields for the ethrex client

### DIFF
--- a/crates/common/types/genesis.rs
+++ b/crates/common/types/genesis.rs
@@ -508,33 +508,29 @@ impl ChainConfig {
     }
 
     pub fn next_fork(&self, block_timestamp: u64) -> Option<Fork> {
-        let next = if self.is_amsterdam_activated(block_timestamp) {
-            None
-        } else if self.is_bpo5_activated(block_timestamp) && self.amsterdam_time.is_some() {
-            Some(Fork::Amsterdam)
-        } else if self.is_bpo4_activated(block_timestamp) && self.bpo5_time.is_some() {
-            Some(Fork::BPO5)
-        } else if self.is_bpo3_activated(block_timestamp) && self.bpo4_time.is_some() {
-            Some(Fork::BPO4)
-        } else if self.is_bpo2_activated(block_timestamp) && self.bpo3_time.is_some() {
-            Some(Fork::BPO3)
-        } else if self.is_bpo1_activated(block_timestamp) && self.bpo2_time.is_some() {
-            Some(Fork::BPO2)
-        } else if self.is_osaka_activated(block_timestamp) && self.bpo1_time.is_some() {
-            Some(Fork::BPO1)
-        } else if self.is_prague_activated(block_timestamp) && self.osaka_time.is_some() {
-            Some(Fork::Osaka)
-        } else if self.is_cancun_activated(block_timestamp) && self.prague_time.is_some() {
-            Some(Fork::Prague)
-        } else if self.is_shanghai_activated(block_timestamp) && self.cancun_time.is_some() {
-            Some(Fork::Cancun)
-        } else {
-            None
-        };
-        match next {
-            Some(fork) if fork > self.fork(block_timestamp) => next,
-            _ => None,
-        }
+        // Pick the scheduled fork with the smallest activation timestamp strictly
+        // greater than `block_timestamp`. Iterating all timestamp-based forks avoids
+        // bugs when intermediate forks (e.g. BPOs) are skipped in a network's schedule.
+        [
+            Fork::Shanghai,
+            Fork::Cancun,
+            Fork::Prague,
+            Fork::Osaka,
+            Fork::BPO1,
+            Fork::BPO2,
+            Fork::BPO3,
+            Fork::BPO4,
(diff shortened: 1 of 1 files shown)


crates/common/types/genesis.rs
```diff
@@ -508,33 +508,29 @@ impl ChainConfig {
...
+            Fork::BPO5,
+            Fork::Amsterdam,
+        ]
+        .into_iter()
+        .filter_map(|fork| {
+            self.get_activation_timestamp_for_fork(fork)
+                .filter(|&t| t > block_timestamp)
+                .map(|t| (fork, t))
+        })
+        .min_by(|a, b| a.1.cmp(&b.1).then_with(|| (a.0 as u8).cmp(&(b.0 as u8))))
```
ElFantasma: **Tie-breaker via `as u8` assumes enum discriminants match chronological order.** When two forks share an activation timestamp (e.g., a dev chain where multiple BPOs activate at the same time), `min_by` falls through to the tie-breaker `(a.0 as u8).cmp(&(b.0 as u8))`. This works IFF `Fork`'s discriminants are assigned in chronological order — same order as the array two lines above (`Shanghai, Cancun, …, Amsterdam`).

If a future Fork is added in the middle of the enum with `#[repr(...)]` rearranged, or someone reorders the enum variants without renumbering, the tie-breaker can quietly disagree with the array's intent.

Two fixes I'd consider:
1. **Array-position tie-breaker** (most robust, decouples from enum discriminant):
   ```rust
   let forks = [Fork::Shanghai, ...];
   forks.iter().copied().enumerate()
       .filter_map(|(pos, fork)| { ... .map(|t| (fork, t, pos)) })
       .min_by(|a, b| a.1.cmp(&b.1).then_with(|| a.2.cmp(&b.2)))
       .map(|(fork, _, _)| fork)
   ```
2. **Named ordering function** on `Fork` (e.g., `Fork::chronological_order(self) -> u8`) that explicitly encodes "chronological order regardless of enum discriminant".

Low priority — the current code is correct for the current `Fork` enum layout. The concern is durability across future enum edits.

crates/common/types/genesis.rs
```diff
@@ -508,33 +508,29 @@ impl ChainConfig {
...
+            Fork::BPO1,
+            Fork::BPO2,
+            Fork::BPO3,
+            Fork::BPO4,
+            Fork::BPO5,
+            Fork::Amsterdam,
+        ]
+        .into_iter()
+        .filter_map(|fork| {
+            self.get_activation_timestamp_for_fork(fork)
```
ElFantasma, line 534: nit: `get_activation_timestamp_for_fork` returns `Option<u64>` for BOTH timestamp-based forks (Shanghai+) AND block-based forks (Homestead, Paris, …). For block-based forks it returns the BLOCK NUMBER, not a timestamp. This code path only iterates timestamp-based forks (the array starts at `Shanghai`), so the `t > block_timestamp` comparison is always timestamps vs timestamps and the math works out.

But the function name conflates the two concepts under a single `u64` type. If a future refactor adds a block-based fork to the array, or if someone reuses this helper elsewhere expecting timestamps and gets a block number, the silent type collision could produce wrong answers.

Worth either:
1. A one-line doc comment on `get_activation_timestamp_for_fork` calling out the conflation ("returns block number for pre-merge forks, timestamp for post-merge forks; callers must know which they're getting"), or
2. Splitting into `get_activation_timestamp` (Shanghai+) and `get_activation_block` (pre-merge), returning `None` for forks of the wrong kind.

Option 2 makes this code self-documenting (only timestamp-based forks can even be queried this way). Out of scope for this PR — just calling it out.

ElFantasma (approved):
Two small inline observations. Neither blocks.

azteca1998 (approved):

ilitteri (approved):

PR #6669, opened 2026-05-18
perf(l1): lazy BAL cursor for per-tx parallel execution

edg-l:
## Summary

Replaces eager per-tx BAL prefix materialization inside `execute_block_parallel` with an on-read `LazyBalCursor` installed on each per-tx `GeneralizedDatabase` (LEVM's in-memory state cache that the EVM reads accounts/slots through during execution). Each tx materializes only the accounts/slots it actually touches instead of the full BAL prefix.

The two outer sequential `seed_db_from_bal` callers (system-call recovery, post-tx outer seed) are unchanged; the cursor is per-tx only.

## Benchmark

Fixture: `bal-devnet-7-mainnet-mix-460` (460 blocks, ~30 Ggas, transfer/EVM-mix). Single run, `release-with-debug` profile, `import-bench --with-bal`.

| metric            | baseline (parallel, eager seed) | this PR (lazy cursor) | delta              |
|-------------------|---------------------------------:|-----------------------:|--------------------|
| wall time         | 8.58 s                           | 6.85 s                 | **-1.73 s (-20.2%)** |
| agg Ggas/s        | 3.90                             | 5.02                   | **+28.7%**         |
| avg ms / block    | 16.88                            | 13.11                  | -3.77 ms (-22.3%)  |
| p95 ms / block    | 17.73                            | 14.09                  | -3.64 ms (-20.5%)  |
| max ms / block    | 90.06                            | 52.39                  | -37.67 ms          |
| exec avg          | 15.57                            | 11.89                  | **-3.68 ms (-23.7%)** |
| merkle avg        | 0.48                             | 0.44                   | -0.04 ms           |
| store avg         | 0.67                             | 0.63                   | -0.04 ms           |
| warmer avg        | 1.37                             | 1.37                   | flat               |

Win is concentrated in `exec`, which is exactly what the cursor targets; merkle/store/warmer barely move, so the gain is not a measurement shift.

## Changes

- Extract `seed_one_address_info_from_bal` and `seed_one_storage_slot_from_bal` from `seed_db_from_bal` as reusable helpers in `ethrex-levm`. `seed_db_from_bal` becomes a thin loop over these helpers (behavior-preserving).
- Add `Clone` on `BalAddressIndex`.
- Add `lazy_bal: Option<LazyBalCursor>` field on `GeneralizedDatabase`. `LazyBalCursor` holds `Arc<BlockAccessList>`, `bal_index: u32`, `Arc<BalAddressIndex>`.
- `load_account` consults the cursor for account info (balance, nonce, code_hash) on cache miss before falling through to the store. Does not inject `account.storage`.
- `get_storage_value` consults the cursor per-slot on cache miss.
- `execute_block_parallel` sets `tx_db.lazy_bal = Some(...)` per tx instead of calling `seed_db_from_bal` eagerly.
- Per-tx `GeneralizedDatabase` capacity hint drops from `bal_account_count` to `32`.
- `code_from_bal` deduplicated into `gen_db.rs`.

## Invariants

1. Cursor `bal_index = tx_idx + 1`; effective cutoff is `bal_index.saturating_sub(1)`, matching the existing `seed_db_from_bal`'s `max_idx = tx_idx`. `debug_assert!(bal_index >= 1)`.
2. `load_account` only injects account-info fields, never `account.storage`. Storage stays lazy through `get_storage_value`.
3. In `seed_one_address_info_from_bal`, `code_update` is computed before the `&mut LevmAccount` borrow; `db.codes.entry().or_insert()` runs after the borrow is released.
4. In `get_storage_value`, the cursor result is copied to a local before taking `&mut current_accounts_state`.
5. `load_account` `.take()`s the cursor before calling the helper (whose partial-coverage path calls `db.get_account` internally) and restores it after; prevents re-entry into the lazy hook.

## Tests

`test/tests/levm/bal_view_tests.rs`:
- `tx1_sees_tx0_write` ; off-by-one boundary
- `load_account_does_not_inject_storage` ; no storage injection
- `sstore_sees_prior_write` ; SSTORE pre-image flows through cursor
- `lazy_load_account_partial_coverage_does_not_recurse` ; `.take()` guard

## Test plan

- [x] `cargo test -p ethrex-test --features rayon bal_view_tests` (4/4)
- [x] `cargo test -p ethrex-vm -p ethrex-levm -p ethrex-blockchain`
- [x] `make lint`
- [ ] `make -C tooling/ef_tests/state test`
- [ ] `make -C tooling/ef_tests/blockchain test`

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -873,10 +873,11 @@ impl LEVM {
         Ok(updates)
     }
 
-    /// Eager BAL prefix seed — used only by the outer DB path (system-call recovery
-    /// at `:440` and post-tx outer seed at `:463`). Per-tx parallel execution uses
-    /// `LazyBalCursor` in `execute_block_parallel`; see also `seed_one_address_info_from_bal`
-    /// and `seed_one_storage_slot_from_bal` in `ethrex_levm::db::gen_db`.
+    /// Eager BAL prefix seed — used only by the outer DB path (parallel-execution
+    /// fallback recovery and post-tx outer seed before request extraction).
+    /// Per-tx parallel execution uses `LazyBalCursor` in `execute_block_parallel`;
+    /// see also `seed_one_address_info_from_bal` and `seed_one_storage_slot_from_bal`
+    /// in `ethrex_levm::db::gen_db`.
     ///
     /// Pre-seed a GeneralizedDatabase with BAL-derived state for a specific tx.
     ///
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -49,7 +49,7 @@ use ethrex_levm::constants::{
 use ethrex_levm::db::gen_db::GeneralizedDatabase;
 #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
 use ethrex_levm::db::gen_db::{
-    LazyBalCursor, code_from_bal, seed_one_address_info_from_bal, seed_one_storage_slot_from_bal,
+    LazyBalCursor, code_from_bal, post_value_at_or_before, seed_one_address_info_from_bal,
 };
 #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
 use ethrex_levm::db::{Database, gen_db::CacheDB};
@@ -922,9 +922,9 @@ impl LEVM {
                 .get_account_mut(addr)
                 .map_err(|e| EvmError::Custom(format!("seed storage mut: {e}")))?;
             for sc in &acct_changes.storage_changes {
-                let key = ethrex_common::utils::u256_to_h256(sc.slot);
-                if let Some(value) = seed_one_storage_slot_from_bal(bal, acct_idx, key, max_idx) {
-                    acc.storage.insert(key, value);
+                if let Some(value) = post_value_at_or_before(sc, max_idx) {
+                    acc.storage
+                        .insert(ethrex_common::utils::u256_to_h256(sc.slot), value);
                 }
             }
         }
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -48,6 +48,10 @@ use ethrex_levm::constants::{
 };
 use ethrex_levm::db::gen_db::GeneralizedDatabase;
 #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
+use ethrex_levm::db::gen_db::{
+    LazyBalCursor, code_from_bal, seed_one_address_info_from_bal, seed_one_storage_slot_from_bal,
+};
+#[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
 use ethrex_levm::db::{Database, gen_db::CacheDB};
 use ethrex_levm::errors::{InternalError, TxValidationError};
 #[cfg(feature = "perf_opcode_timings")]
@@ -455,6 +459,7 @@ impl LEVM {
             // Withdrawal index is n_txs+1 in BAL; we use n_txs to avoid double-applying
             // withdrawal balances (process_withdrawals handles those below).
             let last_tx_idx = u32::try_from(block.body.transactions.len()).unwrap_or(u32::MAX);
+            // Eager seed retained: lazy_bal cursor is per-tx only; outer DB has no cursor.
             Self::seed_db_from_bal(
                 db,
                 bal,
@@ -737,19 +742,6 @@ impl LEVM {
         ))
     }
 
-    /// Convert BAL into `Vec<AccountUpdate>` for the merkleizer.
-    /// Compute code hash and optional `Code` object from raw bytecode in a BAL entry.
-    #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
-    fn code_from_bal(new_code: &Bytes) -> (H256, Option<Code>) {
-        if new_code.is_empty() {
-            (*EMPTY_KECCACK_HASH, None)
-        } else {
-            let code_obj = Code::from_bytecode(new_code.clone(), &ethrex_crypto::NativeCrypto);
-            let hash = code_obj.hash;
-            (hash, Some(code_obj))
-        }
-    }
-
     ///
     /// For each account in the BAL, extracts the **final** post-block state
     /// (highest `block_access_index` entry per field) and builds an AccountUpdate.
@@ -812,7 +804,7 @@ impl LEVM {
 
             // Final code: last entry or prestate
             let (code_hash, code) = if let Some(c) = acct_changes.code_changes.last() {
-                Self::code_from_bal(&c.new_code)
+                code_from_bal(&c.new_code)
             } else {
                 (prestate.code_hash, None)
             };
@@ -881,6 +873,11 @@ impl LEVM {
         Ok(updates)
     }
 
+    /// Eager BAL prefix seed — used only by the outer DB path (system-call recovery
+    /// at `:440` and post-tx outer seed at `:463`). Per-tx parallel execution uses
+    /// `LazyBalCursor` in `execute_block_parallel`; see also `seed_one_address_info_from_bal`
+    /// and `seed_one_storage_slot_from_bal` in `ethrex_levm::db::gen_db`.
+    ///
     /// Pre-seed a GeneralizedDatabase with BAL-derived state for a specific tx.
     ///
     /// For each BAL-modified account, applies accumulated diffs with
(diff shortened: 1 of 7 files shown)


azteca1998 (approved):

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -160,29 +328,89 @@ impl GeneralizedDatabase {
...
+        {
+            let account = account.clone();
+            if !self.skip_initial_tracking {
+                self.initial_accounts_state.insert(address, account.clone());
+            }
+            return Ok(self
+                .current_accounts_state
+                .entry(address)
+                .or_insert(account));
+        }
```
iovoid: What if the system contracts covered by the shared base are later modified?

edg-l: Yes, that was a bug. Fixed in 91648bb: load_account now consults lazy_bal before shared_base, so any address present in both gets the BAL overlay (system-call effects + prior tx mutations) instead of the stale pre-block snapshot. Regression test in test/tests/levm/bal_view_tests.rs.

crates/vm/backends/levm/mod.rs
```diff
@@ -898,118 +895,38 @@ impl LEVM {
...
-                    }
+            let acc = db
+                .get_account_mut(addr)
+                .map_err(|e| EvmError::Custom(format!("seed storage mut: {e}")))?;
+            for sc in &acct_changes.storage_changes {
+                let key = ethrex_common::utils::u256_to_h256(sc.slot);
+                if let Some(value) = seed_one_storage_slot_from_bal(bal, acct_idx, key, max_idx) {
+                    acc.storage.insert(key, value);
                 }
             }
```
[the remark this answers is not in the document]
edg-l: Valid, this is O(n_slots²) per address in the eager path. The outer loop already has the sc, so the inner find is redundant. Will fix in a follow-up — the eager seed is bounded by accounts_by_min_index slice and per-address slot counts are small in practice, so impact is minor, but worth tightening.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -160,29 +328,89 @@ impl GeneralizedDatabase {
...
+        let state = self.store.get_account_state(address)?;
+        let account = LevmAccount::from(state);
+        if !self.skip_initial_tracking {
+            self.initial_accounts_state.insert(address, account.clone());
+        }
+        Ok(self
+            .current_accounts_state
+            .entry(address)
+            .or_insert(account))
     }
```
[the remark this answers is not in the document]
edg-l: Valid for accounts with large storage_changes vectors. The lazy cursor only fires on cache miss (first read per slot per tx), so amortised cost is OK for typical BALs, but worst case is O(n_slots) per miss. Follow-up: build a slot→idx map once per address when the cursor is constructed.

crates/vm/backends/levm/mod.rs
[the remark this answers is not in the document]
edg-l: Fixed in 64106fe — outer loop now uses post_value_at_or_before(sc, max_idx) on the sc it already holds; no inner find().

crates/vm/levm/src/db/gen_db.rs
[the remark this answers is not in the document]
edg-l: Fixed in 64106fe — added slot_idx_by_account: Vec<FxHashMap<H256, usize>> to BalAddressIndex (built once in build_validation_index). Lazy cursor now resolves slot key → sc_idx in O(1). 8721+93 ef-tests still green.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -24,14 +28,185 @@ use std::collections::hash_map::Entry;
...
+
+/// Select the post-value of a single `SlotChange` up to `max_idx`.
+///
+/// Pure read; returns `Some(value)` if any `slot_changes` entry has
+/// `block_access_index <= max_idx`, `None` otherwise.
+#[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
+pub fn post_value_at_or_before(sc: &SlotChange, max_idx: u32) -> Option<U256> {
+    let pos = sc
+        .slot_changes
+        .partition_point(|c| c.block_access_index <= max_idx);
```
ElFantasma, line 166: **`partition_point` assumes sorted `slot_changes`.** The BAL encoding sorts by `block_access_index` (`block_access_list.rs:367` `encode_sorted_by`), but `RLPDecode for AccountChanges` does **not** re-validate ordering. On the optimistic-merkle / lazy-cursor read path the BAL is consulted *before* `validate_block_access_list_hash` runs — a tampered or malformed BAL with unsorted `slot_changes` would have `partition_point` return wrong indices, producing silently-wrong reads during execution.

The post-execution hash check still catches the tampering (the produced BAL won't match the header hash), so this never lets a bad block validate. But the lazy cursor would deliver wrong values to live execution, which can produce confusing intermediate state (gas accounting drift, storage reads returning wrong values, etc.) before the hash-mismatch error fires.

A `debug_assert!(sc.slot_changes.is_sorted_by_key(|c| c.block_access_index), "slot_changes must be sorted by block_access_index")` here (and the same pattern for `balance_changes` / `nonce_changes` / `code_changes` in `seed_one_address_info_from_bal`) surfaces the violation in dev/CI immediately rather than waiting for the downstream hash check to flag a confusing symptom. Cheap insurance; no release-mode cost.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -24,14 +28,185 @@ use std::collections::hash_map::Entry;
...
+            .ok_or(InternalError::AccountNotFound)?
+            .post_nonce;
+        let code_hash = code_update
+            .as_ref()
+            .map(|(h, _)| *h)
+            .unwrap_or(*EMPTY_KECCACK_HASH);
+        let acc = db
+            .current_accounts_state
+            .entry(addr)
+            .or_insert_with(|| LevmAccount {
```
ElFantasma, line 115: **`has_all_info` shortcut sidesteps `shared_base` / store load.** When the BAL covers balance + nonce + code for an account at `<= max_idx`, this branch creates a fresh `LevmAccount` with `exists: true`, `AccountStatus::Modified`, and `storage: FxHashMap::default()` — skipping the `db.get_account(addr)` that the partial-coverage branch (~line 120+) runs.

In the eager-seed path, `seed_db_from_bal` had the same shortcut, so this is behavior-preserving for that caller. But the **lazy-cursor caller in `load_account`** runs BEFORE the `shared_base` / store check (line 410+). For an address that's in BAL with full-info coverage AND in `shared_base` (e.g. a system-touched predeploy mutated by a prior tx), `shared_base` contains the pre-block snapshot — which the full-info shortcut bypasses entirely. The BAL overlay overwrites balance/nonce/code, but `account.storage` starts empty.

For most txs this is fine — storage is filled lazily through `get_storage_value`, which also consults the cursor. But it means: any code that relies on the FULL `account.storage` map being populated for predeploys (e.g. iterating storage keys, bulk reads, anything that bypasses `get_storage_value`) will silently see an empty map post-shortcut.

Not a regression vs. eager-seed (which also creates empty `storage`), but worth a one-line comment confirming the storage-lazy invariant is the contract, so a future reader who needs full-storage doesn't accidentally rely on this path.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -45,6 +220,9 @@ pub struct GeneralizedDatabase {
     /// Optional tracker for BAL validation: records addresses accessed via load_account.
     /// Enabled only during parallel execution to detect extraneous BAL pure-access entries.
     pub accessed_accounts: Option<FxHashSet<Address>>,
+    /// Optional BAL cursor for lazy per-read prefix materialization.
+    /// When set, account loads and storage reads consult the BAL before hitting the store.
+    pub lazy_bal: Option<LazyBalCursor>,
```
ElFantasma, line 230: nit: `lazy_bal` is declared unconditionally on `GeneralizedDatabase`, but every reader site is `#[cfg(all(feature = "rayon", not(feature = "eip-8025")))]`-gated. Under the eip-8025 feature, a caller could legally set `lazy_bal = Some(...)` and the cursor would be silently inert — no compile error, no runtime panic, just nothing happens.

Two cheap options:
1. cfg-gate the field declaration too, matching the readers — code that doesn't compile under eip-8025 makes the inert-cursor case impossible.
2. Add a `debug_assert!(self.lazy_bal.is_none(), "lazy_bal set under eip-8025")` in the eip-8025 paths where execution happens, so a misuse is caught at runtime in dev builds.

Low priority — current code is correct under all the supported feature combos.

ElFantasma (approved):
Three inline findings, all minor — none blocking.

crates/vm/levm/src/db/gen_db.rs
edg-l: Skipping. The post-exec hash check (recorder emits sorted; mismatch ⇒ block rejected) already catches malformed/tampered BAL, and a debug_assert only fires in dev/CI where we don't construct adversarial BALs. Validating sort at decode or cursor-build would cost real cycles for a class of bug the hash check kills.

crates/vm/levm/src/db/gen_db.rs
edg-l: Good catch. Added a comment at the shortcut documenting the storage-lazy invariant (storage materializes through `get_storage_value`, which also consults the cursor) — f057ddb98.

crates/vm/levm/src/db/gen_db.rs
edg-l: Leaving as-is. The eip-8025 cfg gates are going away soon, so cfg-gating the field is short-lived churn.

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -160,29 +346,96 @@ impl GeneralizedDatabase {
...
-                    }
-                    return Ok(entry.insert(account.clone()));
+
+        // Fast path: already cached.
+        if self.current_accounts_state.contains_key(&address) {
+            return self
+                .current_accounts_state
+                .get_mut(&address)
+                .ok_or(InternalError::AccountNotFound);
+        }
```
iovoid, line 355: Could be a single get_mut

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -160,29 +346,96 @@ impl GeneralizedDatabase {
...
-                if !self.skip_initial_tracking {
-                    self.initial_accounts_state.insert(address, account.clone());
+            } else {
+                None
+            };
+            // Restore the cursor before propagating any error or returning.
+            self.lazy_bal = cursor_opt;
+            if let Some(result) = helper_result {
+                result.map_err(|e| InternalError::Custom(format!("lazy_bal seed: {e}")))?;
+                if self.current_accounts_state.contains_key(&address) {
```
iovoid, line 405: Could be just a get_mut

crates/vm/backends/levm/mod.rs
```diff
@@ -881,6 +873,11 @@ impl LEVM {
         Ok(updates)
     }
 
+    /// Eager BAL prefix seed — used only by the outer DB path (system-call recovery
+    /// at `:440` and post-tx outer seed at `:463`). Per-tx parallel execution uses
```
iovoid: Referencing line numbers is fragile

iovoid (approved):

Issue #6583, opened 2026-05-07
tracking: glamsterdam-devnet-6 alignment


edg-l:
Tracking issue for aligning ethrex's `glamsterdam-devnet-6` branch with the `glamsterdam-devnet@v6.0.0` release. Spec baseline: integration on EELS `forks/amsterdam` (branch `devnets/glamsterdam/6` not yet created), fixture release `tests-glamsterdam-devnet@v6.0.0` (target 2026-06-22, not yet cut). Upstream tracker: [execution-specs#2915](https://github.com/ethereum/execution-specs/issues/2915).

First `glamsterdam-`prefixed devnet. The preceding `bal-devnet-7` alignment (state-gas reservoir + BAL baseline, items 1-10) is **complete and merged to `main`** — see the history of this issue; that work is the baseline glam-6 builds on and is not re-listed here.

## Status — 🟢 EL delta implemented; aligned with upstream merges (2026-06-18)

**Implemented + pushed to `glamsterdam-devnet-6`:**
- `EIP-8282` builder requests EL wiring; addresses swapped to sys-asm#43 vanity (`808b13f4c`)
- `EIP-8037` source-based state-gas refunds (`dead5fc52`) + #3002 create-to-existing-target refund (`7682525da`)
- `EIP-8038 + EIP-2780` Amsterdam gas repricing — confirmed aligned with merged numbers
- `EIP-7997` factory nonce → 1; fixture/CI pin → `tests-glamsterdam-devnet@v6.0.0`

**2026-06-18:** 8037 (#11807/#2999 + #3002), 2780 (#11645), 8038 (#11802) all merged upstream; ethrex re-audited and aligned (numbers kept unchanged). Remaining: 8070 (merge decision); 8282 genesis `alloc` bytecode + address freeze pending sys-asm#43; re-run ef-tests/hive on v6.0.0.

**2026-06-19:** verified every EL PR in the devnet-6 spec sheet against GitHub. All merged ones confirmed merged; ethrex alignment unchanged. Two movements since 06-18, neither needs ethrex code: EIP-8282 left draft (EIPs #11760 open, EELS #2990 non-draft and mergeable, now with contract bins + tests) but #2990 still uses placeholder addresses `0x7732`/`0x7733`, so the 8282 ef-test mismatch caveat below still holds until it adopts the vanity addresses. EELS #2900 (7702 delegate warming) is still open as a draft, not yet closed; ethrex matches the default so no action.

[...]

PR #6671, opened 2026-05-18
feat(l1): bal-devnet-7 v7.2.0

edg-l:
Aligns ethrex with bal-devnet-7 v7.2.0.

Tracking: lambdaclass/ethrex#6583 (item 10).

## Changes

- Port ethereum/execution-specs#2863: inline `state_gas` refunds (SSTORE `0→x→0`, CREATE collision/revert, EIP-7702 auth refills) credit the local frame's reservoir immediately. `state_gas_used` becomes `i64`; the clamp-and-spill scaffolding (`state_gas_refund_pending`, `state_gas_refund_absorbed`, `state_gas_spill_outstanding`, `state_gas_credit_against_drain`, intrinsic-tracking field + frame snapshots) is removed.
- Bump fixtures and `eels_commit` to `tests-bal@v7.2.0` / `a3e5201a5`.
- Preserve intrinsic state gas on top-level error: split out a dedicated `vm.intrinsic_state_gas` so the error path refunds only the execution portion. Without this, EIP-7702 set-code reverts and CREATE-tx reverts underbilled `block_state_gas_used` by `AUTH_TOTAL × CPSB` / `NEW_ACCOUNT × CPSB`.
- `default_hook.rs` `state_refund` i64 cast: propagate `InternalError::Overflow` instead of silently saturating (match `vm.rs`).
- `engine_newPayloadV{4,5}`: return JSON-RPC `-32602` on fork/field mismatch (V4 with BAL field or Amsterdam timestamp; V5 missing BAL field or pre-Amsterdam timestamp). Was returning `PayloadStatus.INVALID`.

## Test plan

- `cargo check`, `clippy`, `fmt`: clean across `ethrex-levm`, `ethrex-rpc`, `ethrex-vm`, `ethrex-blockchain`, `ethrex-l2`.
- `make -C tooling/ef_tests/blockchain test-levm` against `tests-bal@v7.2.0`: 8726 passed, 0 failed.
- Hive `eels/consume-engine` Amsterdam: pending re-run; engine-API fix targets the 2 remaining `test_fork_transition` failures.

### DIFF
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -929,9 +929,12 @@ async fn handle_new_payload_v1_v2(
             "New payload requested but syncer is not initialized".to_string(),
         ));
     };
-    // Validate block hash
+    // Validate block hash. Per engine-API spec, a hash mismatch is a structural
+    // payload error (the sender's hash was computed over header fields the
+    // receiver doesn't know about, or vice versa) and surfaces as JSON-RPC
+    // -32602, not PayloadStatus.INVALID.
     if let Err(RpcErr::Internal(error_msg)) = validate_block_hash(payload, &block) {
-        return Ok(PayloadStatus::invalid_with_err(&error_msg));
+        return Err(RpcErr::WrongParam(error_msg));
     }
 
     // Check for invalid ancestors
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -192,6 +192,15 @@ impl RpcHandler for NewPayloadV4Request {
     }
 
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        // EIP-7928 / Amsterdam: V4 payloads MUST NOT include the BAL field — that
+        // field belongs to V5. Per engine-API spec, structurally-invalid payloads
+        // return JSON-RPC -32602 (Invalid params), not PayloadStatus.INVALID.
+        if self.payload.block_access_list.is_some() {
+            return Err(RpcErr::WrongParam(
+                "block_access_list not allowed in engine_newPayloadV4".to_string(),
+            ));
+        }
+
         // validate the received requests
         validate_execution_requests(&self.execution_requests)?;
 
@@ -211,6 +220,13 @@ impl RpcHandler for NewPayloadV4Request {
         };
 
         let chain_config = context.storage.get_chain_config();
+
+        // Amsterdam-active timestamps must use V5, not V4 (-32602 per engine-API spec).
+        if chain_config.is_amsterdam_activated(block.header.timestamp) {
+            return Err(RpcErr::WrongParam(
+                "engine_newPayloadV4 cannot accept Amsterdam timestamps".to_string(),
+            ));
+        }
 
         if !chain_config.is_prague_activated(block.header.timestamp) {
             return Err(RpcErr::UnsupportedFork(format!(
@@ -295,6 +311,15 @@ impl RpcHandler for NewPayloadV5Request {
     }
 
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        // EIP-7928 / Amsterdam: V5 payloads MUST include the BAL field — its
+        // absence is a structural error, not a block-validity failure. Per
+        // engine-API spec, this returns JSON-RPC -32602 (Invalid params).
+        if self.payload.block_access_list.is_none() {
+            return Err(RpcErr::WrongParam(
+                "block_access_list required in engine_newPayloadV5".to_string(),
+            ));
+        }
+
         validate_execution_payload_v4(&self.payload)?;
 
         // validate the received requests
@@ -322,11 +347,11 @@ impl RpcHandler for NewPayloadV5Request {
 
         let chain_config = context.storage.get_chain_config();
 
+        // Pre-Amsterdam timestamps must use V4, not V5 (-32602 per engine-API spec).
         if !chain_config.is_amsterdam_activated(block.header.timestamp) {
-            return Err(RpcErr::UnsupportedFork(format!(
-                "{:?}",
-                chain_config.get_fork(block.header.timestamp)
-            )));
+            return Err(RpcErr::WrongParam(
+                "engine_newPayloadV5 requires Amsterdam timestamp".to_string(),
+            ));
         }
 
         let bal = self.payload.block_access_list.clone();
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -221,11 +221,15 @@ impl RpcHandler for NewPayloadV4Request {
 
         let chain_config = context.storage.get_chain_config();
 
-        // Amsterdam-active timestamps must use V5, not V4 (-32602 per engine-API spec).
+        // Amsterdam-active timestamps must use V5, not V4. Per engine-API spec
+        // (amsterdam.md): "Client software MUST return -38005: Unsupported fork
+        // if the timestamp of payload is greater than or equal to the Amsterdam
+        // activation timestamp."
         if chain_config.is_amsterdam_activated(block.header.timestamp) {
-            return Err(RpcErr::WrongParam(
-                "engine_newPayloadV4 cannot accept Amsterdam timestamps".to_string(),
-            ));
+            return Err(RpcErr::UnsupportedFork(format!(
+                "{:?}",
+                chain_config.get_fork(block.header.timestamp)
+            )));
         }
 
         if !chain_config.is_prague_activated(block.header.timestamp) {
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -242,7 +242,9 @@ impl RpcHandler for NewPayloadV4Request {
         // EIP-7928 fork-boundary detector: V4 doesn't carry block_access_list_hash
         // in its header schema. If the payload's block_hash matches what a V5-style
         // header (with block_access_list_hash injected) would produce, the sender
-        // used the wrong API version — reject with -32602. Real value-mismatch
+        // used the wrong API version; reject with -32602 (InvalidParams) to match
+        // the EELS fixture test_invalid_pre_fork_block_with_bal_hash_field
+        // [fork_BPO2ToAmsterdamAtTime15k-blockchain_test_engine]. Real value-mismatch
         // tests don't match this alternate and fall through to PayloadStatus.INVALID.
         if block.hash() != self.payload.block_hash {
             let mut alt_header = block.header.clone();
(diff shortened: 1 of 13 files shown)


ilitteri (approved):

crates/networking/rpc/engine/payload.rs
```diff
@@ -212,12 +221,37 @@ impl RpcHandler for NewPayloadV4Request {
...
         }
+
+        // EIP-7928 fork-boundary detector: V4 doesn't carry block_access_list_hash
+        // in its header schema. If the payload's block_hash matches what a V5-style
+        // header (with block_access_list_hash injected) would produce, the sender
+        // used the wrong API version — reject with -32602. Real value-mismatch
+        // tests don't match this alternate and fall through to PayloadStatus.INVALID.
+        if block.hash() != self.payload.block_hash {
+            let mut alt_header = block.header.clone();
+            alt_header.block_access_list_hash = Some(H256::zero());
```
ElFantasma, line 251: **Fork-boundary detector only catches `Some(H256::zero())`.** This recomputes `block.hash()` with the BAL hash injected as `H256::zero()` and checks whether the sender's `payload.block_hash` matches that — but a sender who actually used V5 schema would have computed their hash against `Some(real_bal_hash)` where `real_bal_hash` is the keccak of the real BAL (non-zero for any non-empty BAL). So this detector only fires when the sender either:
1. used `Some(H256::zero())` literally as the BAL hash (test fixture or buggy sender), OR
2. used an empty BAL whose keccak coincides with the zero hash — which it doesn't (RLP-encoded empty list has a known non-zero keccak).

The symmetric V5-side detector at line 382 is more useful: it tests `None` (V4 schema) against the V5 payload's `block_hash`, which catches the much more common "sender omitted the BAL field entirely" case.

If the goal is to give a clearer error to test fixtures specifically (rather than catch real-world misrouting), say so in the comment so future readers don't expect this detector to catch arbitrary V5-from-V4 messages. If the goal IS to catch real-world misrouting, this needs to brute-force more candidate BAL hashes — or just always return `-32602` whenever V4 sees an Amsterdam header schema (`block.header.block_access_list_hash.is_some()` would be a structural signal even on the V4 endpoint).

Actually the simpler fix: just check `block.header.block_access_list_hash.is_some()` directly — that field is present iff the sender used V5 schema, regardless of what value they put there. Same protection, no hash recomputation, no narrowness.

crates/vm/levm/src/vm.rs
```diff
@@ -653,112 +629,70 @@ impl<'a> VM<'a> {
...
+        self.state_gas_used = state_gas_used_at_entry;
+        let net_return = child_state_gas_used
+            .checked_add(child_state_gas_left)
             .ok_or(InternalError::Overflow)?;
+        // net_return is always >= 0 by the spec invariant (reservoir conservation
+        // means a child cannot refund more than its ancestors charged); clamp
+        // defensively and cast — `as u64` is sound because of the `.max(0)`.
+        #[expect(clippy::as_conversions, reason = ".max(0) proves non-negativity")]
+        {
+            self.state_gas_reservoir = net_return.max(0) as u64;
```
ElFantasma, line 694: **`.max(0)` defensive clamp hides spec violations.** The comment correctly notes `net_return >= 0` by the spec invariant (reservoir conservation), but the `.max(0)` silently clamps if that invariant is ever violated by a future change. Same pattern at `vm.rs:822` (top-level error reservoir refill) and the CREATE-collision branch in `default_hook.rs`.

If the invariant is load-bearing for correctness (which it is — silently dropping refund value would underbill state-gas), a `debug_assert!(net_return >= 0, "reservoir conservation violated: ...")` ahead of the clamp would surface a regression in dev/CI without changing release-mode behavior. The `as u64` cast underneath would then provably stay sound since the `.max(0)` precondition is held.

crates/networking/rpc/engine/payload.rs
```diff
@@ -212,12 +221,37 @@ impl RpcHandler for NewPayloadV4Request {
 
         let chain_config = context.storage.get_chain_config();
 
+        // Amsterdam-active timestamps must use V5, not V4 (-32602 per engine-API spec).
+        if chain_config.is_amsterdam_activated(block.header.timestamp) {
+            return Err(RpcErr::WrongParam(
+                "engine_newPayloadV4 cannot accept Amsterdam timestamps".to_string(),
```
ElFantasma: nit: error-type inconsistency on V4. This Amsterdam-mismatch returns `WrongParam` (JSON-RPC `-32602`). The Prague-mismatch check ~10 lines below returns `UnsupportedFork`. Both are "sender used the wrong API version for this fork" — same semantic error, different error codes.

If there's a deliberate spec reason for the distinction (e.g. engine-API spec says Amsterdam-vs-V4 is structural and Prague-vs-V4 isn't), worth a comment. Otherwise consider unifying — picking one and using it for both keeps client-side error handling tractable. (Spec text I can find suggests `-32602` is the right code for both, but that's a guess.)

ElFantasma (commented):
The `i64` transition is everywhere — `try_from(u64)`/`try_from(i64)` conversions are added at every boundary with `InternalError::Overflow` mapping. All bounded by tx `gas_limit` (well below `i64::MAX`), so the conversions are belt-and-suspenders in practice. Worth a one-line `// state_gas_used is i64 throughout; tx gas_limit caps the magnitude well below i64::MAX` comment near the field declaration so future readers don't wonder why the conversions exist.

crates/networking/rpc/engine/payload.rs
edg-l: The narrow `Some(H256::zero())` shape matches EELS fixture `tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py::test_invalid_pre_fork_block_with_bal_hash_field[fork_BPO2ToAmsterdamAtTime15k-blockchain_test_engine]` (line 113: `rlp_modifier=Header(block_access_list_hash=Hash(0))`) — that's exactly the trigger this detector targets, so it's intentional, not a general-purpose V5-from-V4 guard. Will add a comment noting that.

crates/vm/levm/src/vm.rs
edg-l: EELS itself uses the same clamp at `amsterdam/fork.py:1168`: `block_output.block_state_gas_used += Uint(max(0, tx_state_gas))`. `state_gas_used` legitimately goes negative when inline refunds exceed gross charges (EELS lines 1087-1090 merge negative `state_gas_used` into `state_gas_left` on error). The `.max(0)` mirrors spec behavior, not a defensive shim — a `debug_assert!(net_return >= 0)` would fire on legitimate refund-heavy txs. Keeping as-is.

crates/networking/rpc/engine/payload.rs
edg-l: Good catch. Per engine-API spec (amsterdam.md:294): "Client software MUST return `-38005: Unsupported fork` error if the timestamp of payload is greater than or equal to the Amsterdam activation timestamp." Fixed in abb964f31 — Amsterdam-on-V4 now returns `UnsupportedFork`, matching the Prague-mismatch check below. No EELS test exercises this path (framework routes Amsterdam→V5) so zero test risk.

azteca1998 (approved):

crates/networking/rpc/engine/payload.rs
```diff
@@ -322,11 +369,28 @@ impl RpcHandler for NewPayloadV5Request {
 
         let chain_config = context.storage.get_chain_config();
 
+        // Pre-Amsterdam timestamps must use V4, not V5 (-32602 per engine-API spec).
```
ElFantasma: **Asymmetric fix — V5 pre-Amsterdam should also be `UnsupportedFork`, not `WrongParam`.**

The engine-API spec section you cited in the reply on the previous round (amsterdam.md: *"Client software MUST return `-38005: Unsupported fork` error if the timestamp of payload is greater than or equal to the Amsterdam activation timestamp"*) applies symmetrically to both directions of the V4/V5-vs-Amsterdam fork-version mismatch:

1. **V4 + Amsterdam timestamp** → already fixed in `abb964f311` to `RpcErr::UnsupportedFork` ✓
2. **V5 + pre-Amsterdam timestamp** → still `RpcErr::WrongParam` here ❌

Also, the comment at the start of this block (`(-32602 per engine-API spec)`) directly contradicts the spec it cites — the spec's verdict for this category is `-38005`, not `-32602`. (The pre-existing Prague-mismatch check at line 236-240 of this file already uses the right pattern; it's just that this V5 block was left out.)

Two-line code fix mirroring `abb964f311`:
```rust
// Pre-Amsterdam timestamps must use V4, not V5. Per engine-API spec
// (amsterdam.md): symmetric to the V4-Amsterdam case — return -38005:
// Unsupported fork.
if !chain_config.is_amsterdam_activated(block.header.timestamp) {
    return Err(RpcErr::UnsupportedFork(format!(
        "{:?}",
        chain_config.get_fork(block.header.timestamp)
    )));
}
```

**Two related side-notes** (not blocking):

- The V4 narrow alt-hash heuristic at lines 242-258 still has the generic comment from before — your previous-round reply said *"Will add a comment noting that"* (citing the `test_invalid_pre_fork_block_with_bal_hash_field[fork_BPO2ToAmsterdamAtTime15k-blockchain_test_engine]` EELS fixture). That comment hasn't landed yet. Worth doing while you're in this file.
- The symmetric V5 alt-hash heuristic at line 388-394 returns `WrongParam` for "sender sent V4 header to V5 endpoint". That's a structural / version-confusion error, same category as the timestamp check above — also a candidate for `UnsupportedFork`. Consistency-vs-spec call; flagging only.

ElFantasma (commented):
Follow-up on the previous round. The structural header-validator addition in `4884189d05` is good (verified the routing through `validate_block_pre_execution` and the new `InvalidBlockHeaderError` variants), and the V4-Amsterdam-timestamp `UnsupportedFork` fix in `abb964f311` is correct. But the V5 side of the same spec rule didn't get the symmetric update — one inline below.

edg-l:
Good catch, applied in deced659f.

1. V5 + pre-Amsterdam now returns `UnsupportedFork` (-38005); comment cites amsterdam.md verbatim. Symmetric with abb964f31; geth `eth/catalyst/api.go:756-757` does the same.
2. Added EELS fixture citation at the V4 alt-hash heuristic (`test_invalid_pre_fork_block_with_bal_hash_field[fork_BPO2ToAmsterdamAtTime15k-blockchain_test_engine]`) and symmetric one at V5 (`test_invalid_post_fork_block_without_bal_hash_field`).
3. Not flipping V5 alt-hash to `UnsupportedFork`: EELS `test_invalid_post_fork_block_without_bal_hash_field` (`tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py:148`) explicitly asserts `EngineAPIError.InvalidParams` (-32602), so the flip would break the engine fixture. All EELS `engine_api_error_code` sites use `InvalidParams`; none use `UnsupportedFork`.

Verified via hive + consume engine on v7.2.0 fixtures: 21/21 amsterdam fork-transition tests, 6/6 cancun excess_blob_gas fork-transition tests.

ElFantasma (approved):

PR #6673, opened 2026-05-18
feat(l1): pruning of historical blocks/receipts/txs

iovoid:
**Motivation**

Block bodies, receipts and transaction locations grow without bound, and after Glamsterdam they grow faster. The consensus layer is only required to serve blocks over `MIN_EPOCHS_FOR_BLOCK_REQUESTS` — 33024 epochs, roughly 4.8 months — and that is the natural lower bound for execution history: the EL must be able to supply payloads at least as long as the CL has to serve them, and anything beyond that is optional. Following the ACDE discussion (ethereum/pm#2197), that window is now the default retention.

**Description**

Background pruning of block bodies, receipts and transaction locations below a retention barrier. Canonical headers are always kept. A new block-number → known-hashes-at-that-height table lets orphaned blocks be pruned too.

| Flag | Values | Default |
|---|---|---|
| `--history.retention` | `cl-window` \| `all` \| `<N>epochs` | `cl-window` |
| `--history.retention.below-cl-window` | bool | `false` |

Bare numbers and wall-clock durations are rejected. `--history.chain` already takes an absolute block *number*, so a neighbouring flag where a bare number meant a *distance* would read identically and mean something else.

**The barrier is one subtraction on block numbers**, not a timestamp search:

```rust
head - (epochs * SLOTS_PER_EPOCH + CL_WINDOW_SLACK_BLOCKS)
```

A block distance is a sound proxy for a slot distance because EIP-3675 permits at most one block per slot and a missed slot produces no block, so the first canonical block inside the window always sits *above* the barrier. The error is one-sided toward over-retention — missed slots make us keep more, never less. This removes two hazards the earlier timestamp-based cutoff had: it derived the cutoff from the host clock, so a machine skewed a year forward pruned a year of real history, and the bisection assumed every canonical height in range had a header, which is false on a database with a header hole. `HistoryRetention::Blocks` exposes the underlying unit, since an epoch is 32 blocks but the slack alone is 7200, so no epoch value can produce a barrier above zero on a test-sized chain.

**A default that prunes must never delete history a datadir already holds.** So: an explicit flag prunes, the default only prunes what it would have kept anyway. A node holding pre-window history keeps it and warns on every boot until the operator opts in. The staleness probe for the earliest-block pointer reads block **1**, never 0 — genesis always has a body, so probing 0 always succeeds and would conclude a stale pointer was sound, grandfathering every node that snap-synced on a build predating #7024.

`--history.chain` disables pruning by default rather than fighting it: backfilled history would otherwise be deleted as fast as it arrives.

Read paths that could conflate "absent" with "empty" now report unavailable history with JSON-RPC code 4444 instead of returning a plausible-but-wrong value: `eth_getLogs` (both the `blockHash` and range branches, plus the per-transaction receipt fetch) and `log_index_base`, which previously summed a short receipt read and silently shifted every log index in a trace.

Two setters now exist where there was one. The pruner's floor only ever rises, so `advance_earliest_block_number` refuses to regress; backfill fills history *in*, so reconciliation moves the pointer down and needs `set_earliest_block_number`. A single monotone-up setter silently swallowed the second case.

**Not in scope, deliberately**

- **BALs are not pruned.** A BAL miss sends `engine_getPayloadBodiesBy*V2` through a full block re-execution before degrading, so pruning them would put repeated failed re-executions on the consensus-critical path. They grow below the barrier; that is a known, smaller cost than the alternative.
- State is unaffected — only the latest state is kept regardless, so there is no "archive" mode to offer.
- The `trace_*` namespace and `eth_gasPrice`'s 20-block lookback are untouched; the latter is safe only because it sits under the pruner's 256-block near-head floor, which is a constant coincidence worth a follow-up.

**Depends on**

#7207 (`eth_feeHistory` range inversion — an unauthenticated remote abort that pruning makes far easier to hit), #7208 (`debug_getRawReceipts` answering empty for absent receipts) and #7210 (advertising the earliest block we can actually serve, which peers use to decide what to request). This branch's `update.rs` change overlaps #7210 and should keep this version — its error handling is stricter.

**Tests**

- Barrier arithmetic: the default equals `MIN_EPOCHS_FOR_BLOCK_REQUESTS`; `all` never prunes; a chain shorter than the window keeps everything rather than underflowing; and the one-sided property that makes the block-distance proxy safe.
- The pruner's five floors — finality, retention, near-head, persisted-state and the L2 committed-batch cap — each in isolation, plus multi-chunk passes, orphan deletion, genesis never being pruned, and restart resilience.
- CLI parsing of every accepted and rejected spelling, including that an absent flag stays distinguishable from an explicit `cl-window`, since only an explicit one may delete existing history.
- `eth_getLogs` returning 4444 rather than a wrong answer on pruned ranges.

Whole workspace is clean under `cargo clippy --all-targets -- -D warnings`; 127 storage, 121 rpc, 58 blockchain and 967 integration tests pass.

---

**Brought up to date with main**

Four months of drift reconciled (12 conflicted files). Three of the overlaps were
decisions rather than mechanics, so they are called out here for review:

- **Schema version collided.** Both sides added a `migrate_3_to_4`: main's is a
  no-op marker bump for the JUMPDEST bitmap, this branch's rebuilds the
  block-hashes-by-number index by scanning every header. Main's shipped, so this
  branch's became `migrate_4_to_5` and `STORE_SCHEMA_VERSION` is now 5. Git merged
  the two definitions *without* a textual conflict, so this would otherwise have
  compiled as a duplicate symbol — and its tests were passing vacuously against
  main's no-op until they were retargeted.
- **`update.rs` keeps this branch's version**, as the description above states.
  Main's `?` propagates every error including `MissingEarliestBlockNumber`, the
  legitimate unset case; this branch defaults only that to 0 and propagates the
  rest, so a transient backend failure is not reported as "I have everything".
- **`GetBlockByHashRequest` takes main's version.** This branch fetched header and
  body separately to catch a pruned body, but main's `get_block_by_hash` already
  returns `None` when the body is absent, so the simpler form is equivalent.

`eth_getLogs` takes main's validation (a superset — it adds a beyond-head check),
so the pruned-range test now builds a real chain first; without a head above the
range it was tripping the new check instead of the prune check it means to
exercise. `RpcErr` takes main's, which already carries `PrunedHistoryUnavailable`
at 4444 alongside `ResourceNotFound`. CLI, storage and RPC test modules were
additive on both sides and all are kept.

Whole workspace clean under `cargo clippy --all-targets -- -D warnings`; 555 unit
tests green (144 storage, 177 rpc, 109 p2p, 87 blockchain, 38 cmd).

Three CI checks fail for reasons that predate this branch: `Link Check`
(docs.alignedlayer.com returns 404 — unrelated docs), `Hive - Devp2p tests`
(failing on every recent PR head), and `Integration Test`, which aggregates the
latter.


yorickdowne:
Wall clock is an interesting decision. Every other client does either blocks or epochs. Ofc slot time is deterministic and wall clock can be computed from epochs by the user

ilitteri:
> Wall clock is an interesting decision. Every other client does either blocks or epochs. Ofc slot time is deterministic and wall clock can be computed from epochs by the user

This has shifted now, as of what's been discussed in the ACDEs recently, but without removing the wall-clock variant. We'll now support CL block retention window as the default, and either `none`, `<N> epochs`, or wall-clock (`<N>d / <N>h / <N>m`), optionally.

yorickdowne:
`history.retention.dry-run` is great UX. Big kudos for this idea.

Issue #786, opened 2024-10-08
LEVM: (EF Tests) stTransitionTest


iinaki:
To remove from ignored this suites:
        
- `"stTransitionTest".into(),`

PR #6676, opened 2026-05-19
fix(l1): suppress payload build on FCU with finalized-ancestor head

edg-l:
## Summary

execution-apis PR [#786](https://github.com/ethereum/execution-apis/pull/786) revised paris.md: when `forkchoiceState.headBlockHash` references a VALID ancestor of the latest known finalized block, the response MUST be `{payloadStatus: VALID, payloadId: null}` and the client MUST NOT begin a payload build process.

ethrex already had the correct `InvalidForkChoice::NewHeadAlreadyCanonical` detection path in `crates/blockchain/fork_choice.rs` (gate: `head.number <= stored_finalized && head_is_canonical`; the `<=` matches geth's `block.NumberU64() <= finalized.Number.Uint64()` in `eth/catalyst/api.go`, see geth PR [#34767](https://github.com/ethereum/go-ethereum/pull/34767)).

The bug was in `crates/networking/rpc/engine/fork_choice.rs`: the `NewHeadAlreadyCanonical` arm returned `Some(head_header)`, which caused `ForkChoiceUpdatedV3::handle` / `ForkChoiceUpdatedV4::handle` to call `build_payload` and set a non-null `payloadId` when `payloadAttributes` was present.

Fix: return `None` for the head header so the V3/V4 dispatch's `if let (Some(_), Some(_))` guard short-circuits `build_payload`. The `VALID + latestValidHash = head` response is preserved, matching geth's `valid(nil)` return.

## Test plan

- `cargo check -p ethrex-rpc` passes
- `cargo clippy -p ethrex-rpc` passes
- `cargo test -p ethrex-rpc` passes


ilitteri (approved):

crates/blockchain/fork_choice.rs
```diff
@@ -78,7 +78,13 @@ pub async fn apply_fork_choice(
     // execution-apis PR 786: the no-reorg skip is only allowed when there is a known
     // finalized block and the head references a VALID ancestor of it. Skipping for
     // unfinalized canonical ancestors is no longer permitted - those must trigger a reorg.
+    //
+    // The `head.number < latest` guard mirrors geth's `head == current_head` carve-out
+    // (see `eth/catalyst/api.go`): when head IS the current canonical head, the CL may
+    // still want to build a payload on top of it (even if it equals the finalized block),
+    // so we fall through to normal FCU handling instead of suppressing payload build.
     if let Some(stored_finalized) = store.get_finalized_block_number().await?
+        && head.number < latest
```
ElFantasma, line 86: nit: the comment two lines up says "head IS the current canonical head" — that interpretation is correct, but it's the **negation** of this guard (we suppress when head is a strict ancestor, fall through when head IS current). Specifically: `head.number == latest && head_is_canonical` implies `head_hash == canonical_at(latest)` via the canonical-chain invariant, so the carve-out semantically reads "if head IS current head, allow payload build."

The condition itself (`head.number < latest`) is correct. Just consider rephrasing the comment to lead with the guard's actual shape — e.g. "`head.number < latest` is the strict-ancestor check; equality (head IS current canonical head) falls through to normal FCU so the CL can build a payload on top." Future readers won't have to mentally invert the comment to match the code.

No behavior change. Drop or keep at your discretion.

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -319,17 +319,14 @@ async fn handle_forkchoice(
         Err(forkchoice_error) => {
             let forkchoice_response = match forkchoice_error {
                 InvalidForkChoice::NewHeadAlreadyCanonical => {
-                    // The fork-choice was effectively accepted: head is canonical and
-                    // points to a known block. Treat it like the Ok(head) branch:
-                    //   - mark the node synced so eth_syncing reports `false`,
-                    //   - return the head header so the caller can build a payload
-                    //     when payloadAttributes is non-null (engine API spec).
+                    // execution-apis PR 786: when head references a VALID ancestor of
```
ElFantasma, line 322: This fix is consequential to CL coordination (a non-null `payloadId` on a finalized-ancestor head causes the CL to chase a payload that's not coming) and the failure mode is subtle — engine-API spec compliance, easy to regress on the next refactor that touches `handle_forkchoice`.

Worth adding a unit test that drives `handle_forkchoice` (or the RPC handler directly) with a `forkchoiceState.headBlockHash` pointing at a finalized ancestor + non-null `payloadAttributes`, and asserts the response is `{payloadStatus: VALID, payloadId: null}`. The `cargo test -p ethrex-rpc` line in the test plan passes today but doesn't lock the invariant. Hive's `engine` group probably covers it on integration, but a unit test would catch regressions before they reach hive.

Non-blocking — your call whether to add it in this PR or as a follow-up.

ElFantasma (approved):
Just two nits

edg-l:
Thanks for the review @ElFantasma. Both nits addressed:

- Comment reworded to lead with the guard shape (`head.number < latest` strict-ancestor check, equality falls through) — 6d6a49f01
- Regression test added in `test/tests/rpc/fork_choice_tests.rs` driving `engine_forkchoiceUpdatedV3` with `head = finalized ancestor` + non-null `payloadAttributes`, asserts `{VALID, latestValidHash=head, payloadId=null}` per [execution-apis#786](https://github.com/ethereum/execution-apis/pull/786) — 36a12255f

azteca1998 (approved):

PR #6677, opened 2026-05-19
perf(l1): move per-tx BAL validation into the par_iter closure

edg-l:
## Summary

`execute_block_parallel` previously returned `(current_state, codes, shadow_touched, shadow_reads)` per tx and validated them in a serial post-`par_iter` loop across all txs. Validation is per-tx pure work; only the marking of `unread_storage_reads` / `unaccessed_pure_accounts` mutates cross-tx state.

This PR moves `validate_tx_execution` + shadow-touched / shadow-reads checks into the rayon closure (parallel). The closure also precomputes the small `(Vec<(Address, H256)>, Vec<Address>)` inputs the serial pass uses to update the shared sets, so `current_state` + `codes` no longer cross the rayon boundary.

BAL validation errors are deferred via `Option<EvmError>` in the result tuple so the post-`par_iter` gas-limit check still takes priority (preserves `GAS_USED_OVERFLOW` > BAL mismatch on blocks exceeding the gas limit; the BAL is built assuming rejected txs, so miner balance in the BAL won't match execution that ran all txs).

## Changes

- `TxExecResult` tuple shrinks from 8 fields (with two heavy `FxHashMap`s) to 7 fields with `Vec`s only.
- Inside the closure, after `execute_tx_in_block`:
  - Compute `reads_satisfied: Vec<(Address, H256)>` (touched storage keys per non-destroyed account) and `destroyed: Vec<Address>` (selfdestructed accounts whose remaining BAL storage_reads are wiped) from `current_state`.
  - Run `validate_tx_execution` + shadow-touched + shadow-reads checks; capture the first error as `Option<EvmError>` (deferred).
  - Drop `current_state` and `codes` before returning.
- After `par_iter`:
  - Existing 2D gas allowance / breakdowns / block-overflow loop unchanged.
  - **New step**: surface the first deferred BAL error in tx order (only reached if the gas-limit check passes).
  - Apply `reads_satisfied` / `destroyed` to `unread_storage_reads` and `tracked_accounts` to `unaccessed_pure_accounts` (cheap hash-set ops).

## Invariants

1. **Gas-limit error priority preserved**: the serial pass runs gas / 2D allowance / block-overflow before surfacing any deferred BAL error.
2. **Tx-order error reporting preserved**: errors are surfaced in sorted `tx_idx` order; the first failing tx still wins.
3. **Destroyed-account semantics preserved**: the `Destroyed | DestroyedModified` branch in the unread-reads marking is now an explicit `Vec<Address>` from the closure; the serial pass does the same `retain(|&(a, _)| a != addr)` it did before.
4. **No semantic change to validation**: same checks, same error strings, only the location moves.

## Benchmark

Fixture: `bal-devnet-7-mainnet-mix-460` (460 blocks, ~30 Ggas, transfer/EVM mix). `release-with-debug`, `import-bench --with-bal`. Baseline = `feat/import-bench-bal-tooling` (bench tooling only, no perf change).

| metric           | baseline | this PR | delta              |
|------------------|---------:|--------:|--------------------|
| wall time        | 8.575 s  | 7.871 s | **-704 ms (-8.2%)** |
| agg Ggas/s       | 3.901    | 4.291   | **+10.0%**         |
| avg ms / block   | 16.884   | 15.349  | **-1.54 ms (-9.1%)** |
| p95 ms / block   | 17.730   | 16.050  | -1.68 ms (-9.5%)   |
| max ms / block   | 90.060   | 85.000  | -5.06 ms           |
| **exec avg**     | **15.574 ms** | **14.172 ms** | **-1.40 ms (-9.0%)** |
| merkle           | 0.479 ms | 0.412 ms | flat (run variance) |
| store            | 0.667 ms | 0.610 ms | flat (run variance) |
| warmer           | 1.365 ms | 1.315 ms | flat               |

The win is concentrated in `exec` (-9%), which is exactly where the serial post-`par_iter` validation pass used to live. The merkle / store / warmer deltas are within run-to-run variance; this PR doesn't touch those phases.

Compare dashboard: <https://edgl.dev/share/compare-feat-import-bench-bal-tooling@02a5663c-vs-perf-bal-validation-into-par-iter@cc8ba27c.html>

## Test plan

- [x] `cargo check --bin ethrex` (default features)
- [ ] EF blockchain tests (Amsterdam; two-pass parallel runner)
- [ ] Hive `bal` group

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -1028,17 +1028,26 @@ impl LEVM {
         let arc_idx = Arc::new(validation_index.clone());
 
         // 2. Execute all txs in parallel (embarrassingly parallel, BAL-seeded).
-        //    BAL validation is deferred to after the gas limit check (step 3) so that
-        //    blocks exceeding gas limit produce GAS_USED_OVERFLOW before BAL mismatch.
+        //    BAL validation runs INSIDE the par_iter closure (parallel) but its
+        //    errors are deferred via Option<EvmError> so the post-par_iter
+        //    gas-limit check still takes priority (GAS_USED_OVERFLOW must beat
+        //    BAL mismatch on blocks exceeding the gas limit; the BAL is built
+        //    assuming rejected txs, so miner balance in the BAL won't match
+        //    execution that ran all txs).
+        //
+        //    The closure also precomputes the small (Vec<(Address, H256)>,
+        //    Vec<Address>) inputs needed to update the shared
+        //    `unread_storage_reads` / `unaccessed_pure_accounts` sets, so the
+        //    serial pass after par_iter is just hash-set ops; current_state
+        //    and codes never cross the rayon boundary.
         type TxExecResult = (
             usize,
             TxType,
             ExecutionReport,
-            FxHashMap<Address, LevmAccount>,
-            FxHashMap<H256, ethrex_common::types::Code>,
             FxHashSet<Address>,   // accessed_accounts tracker (coarse)
-            Vec<Address>,         // shadow recorder touched_addresses (EIP-7928 exact)
-            Vec<(Address, U256)>, // shadow recorder storage_reads (EIP-7928 exact)
+            Vec<(Address, H256)>, // reads_satisfied: (addr, slot) loaded during this tx
+            Vec<Address>,         // destroyed: accounts selfdestructed during this tx
+            Option<EvmError>,     // deferred BAL validation error
         );
 
         let exec_results: Result<Vec<TxExecResult>, EvmError> = (0..n_txs)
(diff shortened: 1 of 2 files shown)


crates/vm/backends/levm/mod.rs
```diff
@@ -1185,37 +1194,109 @@ impl LEVM {
...
                 ))
             })
             .collect();
 
         let mut exec_results = exec_results?;
 
-        // Sort so gas accounting and validation happen in tx order.
-        exec_results.sort_unstable_by_key(|(idx, _, _, _, _, _, _, _)| *idx);
+        // Sort so gas accounting and error surfacing happen in tx order.
+        exec_results.sort_unstable_by_key(|(idx, _, _, _, _, _, _)| *idx);
```
ElFantasma, line 1214: nit: `sort_unstable_by_key` is defensive but should be a no-op here. `(0..n_txs).into_par_iter()` is an `IndexedParallelIterator` (rayon preserves source-index order through `.map()` and `.collect::<Result<Vec<_>, _>>()`), so `exec_results` arrives sorted by `tx_idx` already.

Three options:
1. Drop the sort and add a `debug_assert!(exec_results.windows(2).all(|w| w[0].0 < w[1].0), "par_iter must preserve tx order")` in dev/CI to lock the invariant.
2. Keep the sort with a one-liner comment: `// defensive: rayon's IndexedParallelIterator preserves order, but locking the invariant here is cheap (O(n log n) but n is small).`
3. Leave as-is.

Low priority — `sort_unstable` on already-sorted input is essentially free. But future readers will wonder whether order is actually undefined here.

crates/vm/backends/levm/mod.rs
```diff
@@ -1185,37 +1194,109 @@ impl LEVM {
                     .take()
                     .map(|mut r| (r.take_touched_addresses(), r.take_storage_reads()))
                     .unwrap_or_default();
+
+                // Precompute the per-tx inputs the serial pass uses to update
+                // the shared unread_storage_reads set. Selfdestruct clears
+                // storage from the final state, so destroyed accounts
+                // satisfy ALL their BAL storage_reads regardless of which
+                // slots remain in `current_state`.
+                let mut reads_satisfied: Vec<(Address, H256)> = Vec::new();
```
ElFantasma: **Vec capacity hints.** `reads_satisfied` and `destroyed` allocate fresh `Vec::new()` each tx; on growth they'll do the usual 2x reallocation walk. Typical tx touches ~5-20 (addr, slot) pairs across `current_state` storage, and `destroyed` is almost always 0 (selfdestruct is rare post-EIP-6780). Worst case is a contract iterating many storage slots.

A cheap improvement, in line with what the rest of the closure does:
```rust
let mut reads_satisfied: Vec<(Address, H256)> =
    Vec::with_capacity(current_state.len() * 4);  // rough avg slots/account
let mut destroyed: Vec<Address> = Vec::with_capacity(2);
```
(or some other hint based on profiling). The `current_state.len() * 4` is a guess; profiling on the bal-devnet-7 fixture would tell you whether `2-3 * current_state.len()` is closer.

Really minor — the realloc cost in the hot path is dwarfed by the validation cost. But since this runs in `par_iter` over thousands of txs/block, the allocator pressure is multiplied. Worth doing if profiling shows allocator time.

ElFantasma (approved):
Two small inline observations + a body note. None blocking.

**Body finding:** the closure's `validate_tx_execution` call is wrapped in `map_err(|e| EvmError::Custom(format!("BAL validation failed for tx {tx_idx}: {e}")))` (line 1236-1238) and the two `Custom(...)` `return Err` paths below it also bake `tx_idx` into the message. When the deferred-error surfacing loop (line 1332-1336) returns the first error, the original `tx_idx` is preserved in the message — good. But: `EvmError::Custom` loses structured information that a future caller might want (e.g., `EvmError::BalValidation { tx_idx, kind, ... }`). Adding a typed variant for BAL errors is out of scope for this PR but worth considering as a follow-up — right now consumers can only string-match.

crates/vm/backends/levm/mod.rs
edg-l: Kept the sort with a comment explaining why; agreed it's a no-op today via `IndexedParallelIterator`, but the cost on already-sorted input is near-linear via pdqsort and the defensive guard is worth keeping for a future refactor that swaps iterators. Applied in 48b69675e.

crates/vm/backends/levm/mod.rs
edg-l: `reads_satisfied` pre-sized to `current_state.len() * 4` per your snippet. Left `destroyed` as `Vec::new()` since selfdestruct is rare post-EIP-6780 and `Vec::new()` doesn't allocate until first push, so it's already optimal for the empty-on-most-blocks case. Applied in 48b69675e.

edg-l:
Re. the body note on `EvmError::Custom(format!(...))` losing structure: agreed, but out of scope for this PR. Opened #6711 as a follow-up to introduce a typed `EvmError::BalValidation { tx_idx, kind }` variant and migrate the BAL call sites.

Rebased on main (PR #6669 landed in the meantime; CHANGELOG conflict resolved by collapsing both `### 2026-05-19` Perf entries under one date). Force-pushed.

azteca1998 (approved):

ilitteri (approved):

iovoid (approved):

PR #6678, opened 2026-05-19
feat(l1): bal prometheus instruments, dashboard panels, and localnet fixture

edg-l:
## Summary

Adds Prometheus instruments for EIP-7928 Block Access Lists and a "BAL (EIP-7928)" row in `ethrex_l1_perf.json` to surface them. Aligns with the metrics subset of the cross-client BAL OTel spec (<https://notes.ethereum.org/@ethpandaops/bal-otel>); OTel span tracing is a separate follow-up.

The instruments live in `crates/blockchain/metrics/bal.rs` using the `register_*` macro / default-registry pattern (same shape as `sync.rs`), so they surface via `gather_default_metrics()` with no `api.rs` coupling. Metric names are bare `bal_*` (no client prefix) to match the cross-client spec.

## Instruments

| name | type | description |
|---|---|---|
| `bal_blocks_total` | IntCounter | cumulative BAL-carrying blocks processed |
| `bal_size_bytes` + `bal_size_bytes_histogram` | Gauge + Histogram | RLP-encoded BAL size and distribution (exponential buckets 1 KiB, 2 KiB, ..., 32 MiB) |
| `bal_account_count` | IntGauge | account count in the latest BAL |
| `bal_slot_count` | IntGauge | unique storage slots (writes + reads), via `BlockAccessList::item_count()` minus per-address rows |

## Dashboard panels

| panel | type | promql |
|---|---|---|
| BAL Blocks Rate | timeseries | `rate(bal_blocks_total[$__rate_interval])` |
| BAL Size (bytes) | timeseries | `bal_size_bytes` |
| BAL Size Distribution | heatmap | `sum by (le)(rate(bal_size_bytes_histogram_bucket[$__rate_interval]))` |
| BAL Account Count | timeseries | `bal_account_count` |
| BAL Slot Count | timeseries | `bal_slot_count` |

The `ethereum-package/src/grafana/ethrex_l1_perf.json` copy is regenerated by `make localnet` from the canonical file.

## Test plan

- [x] `cargo check --bin ethrex`
- [x] `cargo check -p ethrex-blockchain --no-default-features`
- [x] Confirmed live: dev node with an Amsterdam-active genesis (`--dev --network fixtures/genesis/l1-bal.json --metrics`) producing blocks; `curl /metrics` shows all 5 `bal_*` series populating (`bal_blocks_total` incrementing per block, histogram lowest bucket `le="1024"` with no `le=0` bucket). Dev-mode Amsterdam block production required #6731.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -111,6 +111,8 @@ use tokio_util::sync::CancellationToken;
 use vm::StoreVmDatabase;
 
 #[cfg(feature = "metrics")]
+use ethrex_metrics::bal::METRICS_BAL;
+#[cfg(feature = "metrics")]
 use ethrex_metrics::blocks::METRICS_BLOCKS;
 
 #[cfg(feature = "c-kzg")]
@@ -1980,14 +1982,15 @@ impl Blockchain {
             );
         }
 
-        metrics!(if let Some(bal) = produced_bal.as_ref().or(bal) {
-            let account_count = bal.accounts().len() as u64;
-            let slot_count = bal.item_count().saturating_sub(account_count);
-            let size_bytes = bal.length() as f64;
-            METRICS_BLOCKS.inc_bal_blocks_total();
-            METRICS_BLOCKS.set_bal_size_bytes(size_bytes);
-            METRICS_BLOCKS.set_bal_account_count(account_count as i64);
-            METRICS_BLOCKS.set_bal_slot_count(slot_count as i64);
+        metrics!(if let Some(bal_ref) = produced_bal.as_ref().or(bal) {
+            let account_count = bal_ref.accounts().len() as u64;
+            let slot_count = bal_ref.item_count().saturating_sub(account_count);
+            let size_bytes = bal_ref.length() as f64;
+            METRICS_BAL.blocks_total.inc();
+            METRICS_BAL.size_bytes.set(size_bytes);
+            METRICS_BAL.size_bytes_histogram.observe(size_bytes);
+            METRICS_BAL.account_count.set(account_count as i64);
+            METRICS_BAL.slot_count.set(slot_count as i64);
         });
 
         Ok((produced_bal, result))
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -1980,6 +1980,16 @@ impl Blockchain {
             );
         }
 
+        metrics!(if let Some(bal) = produced_bal.as_ref().or(bal) {
+            let account_count = bal.accounts().len() as u64;
+            let slot_count = bal.item_count().saturating_sub(account_count);
+            let size_bytes = bal.length() as f64;
+            METRICS_BLOCKS.inc_bal_blocks_total();
+            METRICS_BLOCKS.set_bal_size_bytes(size_bytes);
+            METRICS_BLOCKS.set_bal_account_count(account_count as i64);
+            METRICS_BLOCKS.set_bal_slot_count(slot_count as i64);
+        });
+
         Ok((produced_bal, result))
     }
 
--- a/crates/blockchain/metrics/blocks.rs
+++ b/crates/blockchain/metrics/blocks.rs
@@ -1,6 +1,6 @@
 use prometheus::Histogram;
 use prometheus::HistogramOpts;
-use prometheus::{Encoder, Gauge, IntGauge, Registry, TextEncoder};
+use prometheus::{Encoder, Gauge, IntCounter, IntGauge, Registry, TextEncoder};
 use std::sync::LazyLock;
 
 use crate::MetricsError;
@@ -37,6 +37,17 @@ pub struct MetricsBlocks {
     warmer_ms: Gauge,
     /// Warmer finished early (positive) or late (negative) relative to exec, in ms
     warmer_early_ms: Gauge,
+    // BAL (EIP-7928) metrics.
+    /// Cumulative count of BAL-carrying blocks processed (post-Amsterdam).
+    bal_blocks_total: IntCounter,
+    /// RLP-encoded size of the most recent BAL, in bytes.
+    bal_size_bytes: Gauge,
+    /// Distribution of RLP-encoded BAL sizes in bytes.
+    bal_size_bytes_histogram: Histogram,
+    /// Number of accounts in the most recent BAL.
+    bal_account_count: IntGauge,
+    /// Unique storage slots (writes + reads) in the most recent BAL.
+    bal_slot_count: IntGauge,
 }
 
 impl Default for MetricsBlocks {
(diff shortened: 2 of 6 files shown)


azteca1998 (approved):

crates/blockchain/metrics/blocks.rs
```diff
@@ -157,6 +168,40 @@ impl MetricsBlocks {
...
+                "RLP-encoded size of the most recent Block Access List, in bytes",
+            )
+            .expect("Failed to create bal_size_bytes metric"),
+            bal_size_bytes_histogram: Histogram::with_opts(
+                HistogramOpts::new(
+                    "bal_size_bytes_histogram",
+                    "Distribution of RLP-encoded Block Access List sizes in bytes",
+                )
+                .buckets({
+                    let mut buckets =
```
ElFantasma: **Histogram bucket choice.** `exponential_buckets(1024, 2.0, 16)` produces buckets at 1 KiB, 2 KiB, 4 KiB, …, 32 MiB (16 buckets), then `buckets.insert(0, 0.0)` prepends a zero floor. The implementation is fine and the comment cites the right range, but the literal `1024.0, 2.0, 16` is easy to misread; future-you wondering "is the top bucket 16 MiB or 32 MiB?" has to do the arithmetic.

Cheap improvement: a named local or constants:
```rust
const BAL_HISTOGRAM_BASE: f64 = 1024.0;  // 1 KiB
const BAL_HISTOGRAM_BUCKETS: usize = 16;  // up to ~32 MiB at base 2
```
or inline-comment the buckets list:
```rust
.buckets({
    // 0, 1 KiB, 2 KiB, 4 KiB, 8 KiB, 16 KiB, 32 KiB, 64 KiB, 128 KiB,
    // 256 KiB, 512 KiB, 1 MiB, 2 MiB, 4 MiB, 8 MiB, 16 MiB, 32 MiB.
    let mut buckets = prometheus::exponential_buckets(1024.0, 2.0, 16)
```

Low priority; just readability.

crates/blockchain/blockchain.rs
```diff
@@ -1964,6 +1964,16 @@ impl Blockchain {
             );
         }
 
+        metrics!(if let Some(bal) = produced_bal.as_ref().or(bal) {
+            let account_count = bal.accounts().len() as u64;
```
ElFantasma: nit: `bal_size_bytes` / `bal_account_count` / `bal_slot_count` are per-block-snapshot Gauges (replaced on every new BAL block). The existing convention at lines ~1997-2000 in the same file uses `latest_*` for the same shape (`set_latest_gas_used`, `set_latest_block_gas_limit`, `set_latest_gigagas`).

If you stick with the manual-Registry pattern in `blocks.rs` (i.e. don't take the body's Body Finding 1 suggestion), worth at least a one-line doc-comment on each field declaration noting it's a snapshot Gauge (not an instantaneous value) so future readers don't expect monotonic semantics:
```rust
/// RLP-encoded size of the most recent BAL, in bytes (per-block snapshot).
bal_size_bytes: Gauge,
```

If you take the body's suggestion and move to a new `bal.rs` module using the macro pattern, the naming question shifts to whether the cross-client OTel BAL spec requires literal `bal_*` or allows an `<client>_bal_*` prefix — see the body. Non-blocking either way.

ElFantasma (commented):
**NOTE 1: this PR is adding to the divergent side of an existing split in how Prometheus is wired** — same shape as the concern I flagged in [#6466](https://github.com/lambdaclass/ethrex/pull/6466#pullrequestreview-4199529604) (review body).

There are two patterns in the codebase today:

- **`register_*` macro / default-registry pattern (2 modules)**: `rpc.rs` (introduced in #5335, 2025-11-24) and `sync.rs` (#6470, 2026-04-20). Each metric is created with `register_int_gauge!` / `register_counter_vec!` / etc., which registers it into Prometheus's default registry at construction time. `api.rs` surfaces them automatically via `gather_default_metrics()`.

- **Manual `Registry::new()` pattern (6 modules)**: `blocks.rs` (this one), `node.rs`, `p2p.rs`, `process.rs`, `transactions.rs`, `l2/metrics.rs`. Each module owns a singleton struct of bare collectors, exposes a `gather_metrics()` method that builds a fresh `Registry::new()` per call, and `api.rs:40` has to know about each module and concatenate their output.

We should be moving towards the first pattern; that's tech debt to be addressed in some future effort. But this PR adds 5 new instruments from scratch — a free moment to land them in the better shape rather than grow `blocks.rs`'s divergent surface by 5 entries. Concrete deltas vs the current diff:

1. **Move BAL metrics to a sibling module** `crates/blockchain/metrics/bal.rs` (mirroring `sync.rs`). Use `register_int_counter!("ethrex_bal_blocks_total", ...)`, `register_gauge!("ethrex_bal_size_bytes", ...)`, `register_histogram_with_buckets!("ethrex_bal_size_bytes", ...)`, `register_int_gauge!("ethrex_bal_account_count", ...)`, `register_int_gauge!("ethrex_bal_slot_count", ...)`. Each macro registers into the default Prometheus registry at construction time. New metrics auto-surface via `gather_default_metrics()` with zero `api.rs` change.

2. **Naming becomes `ethrex_bal_*` (with the `ethrex_` prefix)**, matching the `ethrex_sync_*` / `ethrex_rpc_*` convention used in the macro-pattern modules. The bare `bal_*` names the current diff uses match the cross-client OTel BAL spec literally, but most cross-client dashboards query with regex (`.*bal_.*`) — confirming this against the spec text before merge would lock in the right name. If the spec mandates literal `bal_*` with no client-prefix, the macro pattern doesn't fit and the current shape stays; if it allows a `<client>_` prefix, the macro pattern is the better landing spot.

3. **Drop the `set_bal_*` / `inc_bal_blocks_total` methods on `MetricsBlocks`.** With the new module's instruments being `pub static LazyLock<...>`, the call sites in `blockchain.rs` become `METRICS_BAL.size_bytes.set(size)` / `METRICS_BAL.blocks_total.inc()` — direct collector access, no wrapper methods.

4. **Drop `api.rs`'s explicit `METRICS_BLOCKS.gather_metrics()` call** for the BAL slice. Once the metrics live in the default registry via macros, `api.rs` doesn't need to know they exist. Mirrors the comment at `api.rs:57-58` from #6470: *"METRICS_SYNC registers into the default Prometheus registry at init, so its metrics are already included in `gather_default_metrics()` above."*

The 5 instruments map 1:1 to the current diff — same types, same descriptions, same `produced_bal.as_ref().or(bal)` call site in `blockchain.rs`. Mechanical refactor; the only behavioral change is the metric names gaining the `ethrex_` prefix (point 2 above).

---

**Note 2: test plan + CHANGELOG gaps.** `make localnet against an Amsterdam-active config, confirm the BAL row populates post-fork` is unchecked — that's the validation that proves the dashboard actually shows data. And no CHANGELOG entry for an observability addition; a one-liner under "Observability" would help operators find it later.

The two inline observations below stand independently of the manual-vs-macro choice above — they apply either way.

ilitteri (approved):

edg-l:
Pushed a follow-up addressing review feedback:

- Moved the 5 BAL instruments out of `MetricsBlocks` into a new `metrics/bal.rs` using the `register_*` macro pattern (mirrors `sync.rs`). They now register into the default Prometheus registry and surface via `gather_default_metrics()`, so `blocks.rs` is back to its pre-PR shape and `api.rs` is untouched.
- Metric names kept bare `bal_*` (no `ethrex_` prefix), so dashboard queries are unchanged.
- Dropped the never-populated `le=0` histogram bucket; bucket params are now named constants.
- Gauges documented as per-block snapshots.
- Renamed the shadowed `bal` binding to `bal_ref`.
- Added a CHANGELOG entry.

edg-l:
Smoke-tested live. Ran a dev node with an Amsterdam-active genesis and metrics enabled (needs #6731 for dev-mode Amsterdam block production):

```
ethrex --dev --network fixtures/genesis/l1-bal.json --datadir memory --metrics --metrics.port 3702
```

After a few BAL-carrying blocks, `curl -s localhost:3702/metrics | grep '^bal_'`:

```
bal_account_count 4
bal_blocks_total 12
bal_size_bytes 172
bal_size_bytes_histogram_bucket{le="1024"} 12
bal_size_bytes_histogram_bucket{le="2048"} 12
...
bal_size_bytes_histogram_bucket{le="33554432"} 12
bal_size_bytes_histogram_bucket{le="+Inf"} 12
bal_size_bytes_histogram_sum 2064
bal_size_bytes_histogram_count 12
bal_slot_count 11
```

All 5 instruments populate; `bal_blocks_total` increments per block. The histogram's lowest bucket is `le="1024"` (no `le=0` floor), per the bucket fix.

ElFantasma (approved):

PR #6680, opened 2026-05-19
feat(l1): support engine_newPayloadWithWitnessV5

jsign:
This PR adds support for `engine_newPayloadWithWitnessV5` in the EngineAPI.

The main goal, for now, is having the required endpoint such that we could run execution witness construction correctness checks using the zkevm EEST releases via Hive.


crates/networking/rpc/engine/payload.rs
```diff
@@ -1098,13 +1167,44 @@ async fn try_execute_payload(
...
+        .headers
+        .iter()
+        .map(|header| BlockHeader::decode(header.as_ref()))
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|error| RpcErr::Internal(format!("Failed to decode witness header: {error}")))?;
+    headers.sort_by_key(|header| header.number);
+    let mut codes = rpc_witness.codes;
+    codes.sort_by(|a, b| a.as_ref().cmp(b.as_ref()));
+    let mut state = rpc_witness.state;
+    state.sort_by(|a, b| a.as_ref().cmp(b.as_ref()));
```
jsign, line 1358: [Ordering defined in spec](https://github.com/ethereum/execution-specs/blob/8d7e68f4b7fa58107647164d1749853a6ed00eb3/src/ethereum/forks/amsterdam/stateless_host_exec_witness.py#L175-L176).

jsign:
@ilitteri, I think this is ready for some eyes. Feel free to work on top if this is diverging from your other approaches.

Reg CI, I think the running might ran out of space. :thinking:

crates/networking/rpc/engine/payload.rs
```diff
@@ -295,7 +301,17 @@ impl RpcHandler for NewPayloadV5Request {
...
+    }
+}
+
+impl NewPayloadV5Request {
+    async fn handle_with_witness(
+        &self,
+        context: RpcApiContext,
+        make_witness: bool,
+    ) -> Result<Value, RpcErr> {
+        validate_execution_payload_v5(&self.payload)?;
```
ElFantasma, line 354: **Behavioral change for `engine_newPayloadV5`.** Pre-PR, `NewPayloadV5Request::handle` called `validate_execution_payload_v4`. This PR routes V5 through the new `validate_execution_payload_v5` which adds a `slot_number.is_some()` requirement (the test at line 1334 confirms rejection).

If any in-the-wild CL client was sending V5 payloads without `slot_number`, this PR will start rejecting them with JSON-RPC `-32602` instead of accepting them. The stricter behavior is correct (V5 spec mandates `slot_number`) — but it IS a behavioral change for non-conforming clients, not just an additive feature.

The PR description currently mentions only the new endpoint; consider adding a one-liner noting the V5 validation tightening so this isn't surprising during a CL-EL pairing test. Also worth a release-notes / CHANGELOG entry under "Behavioral changes" or similar.

crates/blockchain/blockchain.rs
```diff
@@ -1892,18 +1916,24 @@ impl Blockchain {
...
                 account_updates,
                 &block,
                 parent_header,
                 &logger,
             )?;
-            self.storage
-                .store_witness(block_hash, block_number, witness)?;
+            if should_store_witness {
+                self.storage
+                    .store_witness(block_hash, block_number, generated_witness.clone())?;
```
ElFantasma: nit: when `should_store_witness=true && force_witness=false` (the original config-driven path), this clones the witness for storage but `generated_witness` is then dropped at the end of the block. Pre-PR the witness was *moved* into `store_witness`, no clone. The clone is only needed when both flags are true (storage + return for RPC).

Cheap fix:
```rust
if force_witness {
    witness = Some(generated_witness.clone());
}
if should_store_witness {
    self.storage.store_witness(block_hash, block_number, generated_witness)?;
}
```
(Reverses the order so the storage path can take ownership when `force_witness=false`.) Or a `match (should_store_witness, force_witness)` for explicitness.

`ExecutionWitness` can be sizable (full block witness with state/code/headers); avoiding the redundant clone on the hot path matters more for the steady-state `precompute_witnesses` flow than for the test endpoint.

crates/networking/rpc/engine/payload.rs
```diff
@@ -1098,13 +1167,44 @@ async fn try_execute_payload(
...
+                    RpcErr::Internal("Payload executed without producing a witness".to_string())
+                })?;
+                status.witness = Some(encode_witness_for_engine_rpc(witness)?);
+            }
+            Ok(status)
         }
     }
 }
 
+fn encode_witness_for_engine_rpc(witness: ExecutionWitness) -> Result<Bytes, RpcErr> {
```
ElFantasma, line 1328: **What spec defines this encoding format?** The function decodes RLP headers back to `BlockHeader` structs just to sort by `header.number`, then re-encodes them via the `Encoder` — a full round-trip on the deserialize side. This works because `BlockHeader` RLP is canonical (the round-trip is byte-identical), but the choice of:
- sort headers by number (vs. e.g. lexicographic byte sort like `codes`/`state`)
- final encoding as a 3-field RLP list of `(headers, codes, state)`
- inner sort orders

…all need to match whatever the zkevm EEST Hive checker expects. A reference (link to the relevant Hive simulator code, EIP draft, or geth implementation of `engine_newPayloadWithWitnessV5`) in the function-level doc comment would make this much easier to verify for future maintenance.

If the format is intended to match geth's `engine_newPayloadWithWitnessV5` response shape, a one-line `// Format: matches geth's <url-to-file>#<line> per <spec>` would prevent future drift.

Also minor: the field order in `Encoder::new(...)` (`headers`, `codes`, `state`) is implicit positional; the test plan doesn't include any round-trip parse via this format, so an EEST format-mismatch would only surface during Hive runs. Worth at least one unit test that hand-builds an `ExecutionWitness` and verifies the bytes.

crates/blockchain/blockchain.rs
```diff
@@ -1820,11 +1828,23 @@ impl Blockchain {
...
     ) -> Result<Option<BlockAccessList>, ChainError> {
-        let (produced_bal, result) = self.add_block_pipeline_inner(block, bal)?;
+        let (produced_bal, _, result) = self.add_block_pipeline_inner(block, bal, false)?;
         result?;
         Ok(produced_bal)
     }
 
+    /// Same as [`add_block_pipeline`] but returns the execution witness produced
+    /// while importing the block.
+    pub fn add_block_pipeline_with_witness(
```
ElFantasma, line 2040: API contract nit: returns `Result<Option<ExecutionWitness>, ChainError>`, but `force_witness=true` is hard-coded internally and the witness *should* always be `Some` here by the upstream invariants (`collect_witness=true` ⇒ `logger` is `Some` ⇒ `accumulated_updates` is `Some` ⇒ witness is computed and assigned). The caller at `try_execute_payload` already treats `None` as `Err("Payload executed without producing a witness")`.

The Option is doing defensive work but the contract is unclear: is `None` a real failure mode, or just downstream paranoia? Two cleaner options:
1. Return `Result<ExecutionWitness, ChainError>` — promote the "witness should be Some" invariant to the type, error explicitly if violated.
2. Document the precondition: `// Returns Some unless ... (specific edge case)`.

Low priority — current behavior is correct — but the API shape will confuse future callers wondering when None is valid.

crates/networking/rpc/engine/payload.rs
iovoid, line 1209: What if we already imported the block without generating the witness?

jsign: Ah, good point. I addressed this as part of https://github.com/lambdaclass/ethrex/pull/6680/commits/b38c77fd52646f3cc44e1174f427043b5f23ae12. So leveraging existing computations in cache or reconstructing if needed.

crates/networking/rpc/engine/payload.rs
jsign: I think there isn't any Amsterdam context that isn't considering the slot number addition to the header at this point. Mainly because it is SFI now.

crates/blockchain/blockchain.rs
jsign: Thanks, that's a good point. Fixed as part of https://github.com/lambdaclass/ethrex/pull/6680/commits/b38c77fd52646f3cc44e1174f427043b5f23ae12

crates/networking/rpc/engine/payload.rs
jsign: Agree. Luckly, the PR I did for Geth was merged just some minutes ago so I could add some better referencing and tests here https://github.com/lambdaclass/ethrex/pull/6680/changes/b38c77fd52646f3cc44e1174f427043b5f23ae12#diff-02a5d59ebf0495a7a5be28b80fcb8ec2b5b7131593c2df039046e3a8de26d39cR1242

crates/blockchain/blockchain.rs
jsign: Yep, agree that since it is forced in theory not receiving it should be better be consdiered an error and avoid this indirection (although changes a bit the caller, but I agree it is clearer). Fixed in https://github.com/lambdaclass/ethrex/pull/6680/commits/60f9ce6130d4d1ce76801ed1cb627e5932977f55

ElFantasma (approved):
LGTM, Thanks for the contribution!

ilitteri (approved):

iovoid (approved):

jsign:
:eyes:

ilitteri:
**Reorg Tests are failing because the node aborts with a stack overflow on the first Engine API call.**

The CI failure surfaces as simulator panics (`assertion failed: res.unwrap().success()` and `hyper::Error(IncompleteMessage)`), but those are symptoms — the node process had already died. Its log (redirected to a file, not shown in the CI output) ends with:

```
thread 'tokio-rt-worker' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

Reproduced locally and captured the crashing thread under lldb: a `tokio-rt-worker` faults on the guard page while handling `engine_forkchoiceUpdatedV3` (the first request the simulator sends), with the stack too deep to even unwind — innermost frame `ethrex_blockchain::fork_choice::apply_fork_choice`.

**Root cause:** `map_engine_requests` is a single `async fn` whose `match` awaits every handler inline. A Rust async fn's future is sized to its *largest* arm, so the `newPayload` arms — enlarged here by threading `make_witness` through `handle_new_payload_v1_v2 → v3 → v4 → v5`, returning `Result<Option<ExecutionWitness>, ChainError>` from `add_block`, and adding the `handle_with_witness` / `payload_status_for_existing_block` / `witness_for_existing_block` async fns — inflate the shared dispatch future. A single poll of *any* engine request, including the lightweight `forkchoiceUpdatedV3`, now exceeds the 2 MB tokio worker stack in the unoptimized debug build CI uses (`cargo build --bin ethrex`). That's why a PR that doesn't touch `fork_choice.rs` or the build path still breaks it, and only in the debug build: the cost is shared through the dispatch future, and Reorg Tests is green on `main` and on this branch's earlier commits.

**Suggested fix:** `Box::pin` the heavy `newPayload` arms so their futures live on the heap instead of bloating the shared stack frame:

```rust
"engine_newPayloadWithWitnessV5" => Box::pin(NewPayloadWithWitnessV5Request::call(req, context)).await,
"engine_newPayloadV5" => Box::pin(NewPayloadV5Request::call(req, context)).await,
"engine_newPayloadV4" => Box::pin(NewPayloadV4Request::call(req, context)).await,
"engine_newPayloadV3" => Box::pin(NewPayloadV3Request::call(req, context)).await,
```

With this change the full reorg suite passes locally (through `test_many_blocks_reorg`).

PR #6683, opened 2026-05-20
feat(l1): add CLI toggles for BAL parallel exec, prefetch, and optimistic trie

### LINKED ISSUE #6684: Add --no-bal-parallel-trie CLI toggle (follow-up to #6683)
Follow-up to #6683, which added `--no-bal-parallel-exec` and `--no-bal-prefetch`. The third BAL perf toggle (\`--no-bal-parallel-trie\`) was intentionally deferred because the optimization it would gate is not yet on \`main\`.

## What this flag would gate
Optimistic / parallel trie merkleization: building the trie updates from BAL-synthesized deltas in parallel with execution, instead of streaming \`AccountUpdate\`s from the executor over a channel after execution finishes. The state root still comes from BAL-derived post-state in both modes; only the timing/parallelism differs.

## Why it's not in #6683
The synthesizer and merkleizer entry point for this path live on \`bal-devnet-7\` and haven't been merged to \`main\`:

- \`synthesize_bal_updates\` (in \`crates/common/types/block_access_list.rs\` on the branch)
- \`handle_merkleization_bal_from_updates\` (in \`crates/blockchain/blockchain.rs\` on the branch)
- The \`optimistic_updates\` branch in \`add_block_pipeline_inner\` that picks between synthesized and streaming merkleization

See branches \`origin/bal-devnet-7\` and \`origin/perf/bal-optimistic-merkleization\`.

## What needs to happen
Once the optimistic-trie path lands on \`main\`, add:

- \`bal_parallel_trie_enabled: bool\` to \`BlockchainOptions\` (default \`true\`)
- \`--no-bal-parallel-trie\` / \`ETHREX_NO_BAL_PARALLEL_TRIE\` CLI arg in \`cmd/ethrex/cli.rs\`, wired through the same construction sites updated in #6683 (\`cmd/ethrex/cli.rs\`, \`cmd/ethrex/initializers.rs\`, \`cmd/ethrex/l2/initializers.rs\`)
- In \`Blockchain::execute_block_pipeline\`, gate the \`synthesize_bal_updates\` / \`optimistic_updates\` assignment with the flag so the merkleizer thread takes the streaming \`handle_merkleization\` branch when disabled. The streaming channel \`(tx, rx)\` already exists on \`main\` for the pre-Amsterdam path; the disabled path needs to keep it wired up on the BAL path too.

Defaults should keep the optimization on; the flag is for benchmarking and cross-client A/B.

## Context
Inspired by the [bal-devnet-6 feature-flags notes](https://notes.ethereum.org/@ethpandaops/bal-devnet-6#Feature-flags) (reth: \`--engine.disable-bal-parallel-state-root\`; geth: \`--bal.executionmode=sequential|nobatchio\`).

edg-l:
## Summary
Adds three CLI flags so operators can disable the BAL-driven optimizations on Amsterdam+ validation, for benchmarking and A/B comparison against other clients:

- `--no-bal-parallel-exec` / `ETHREX_NO_BAL_PARALLEL_EXEC` — disable BAL-driven parallel transaction execution; falls through to the sequential pipeline.
- `--no-bal-prefetch` / `ETHREX_NO_BAL_PREFETCH` — disable the BAL-driven warmer thread.
- `--no-bal-parallel-trie` / `ETHREX_NO_BAL_PARALLEL_TRIE` — disable BAL-driven optimistic trie merkleization; falls back to streaming `AccountUpdate`s from the executor.

Default behavior is unchanged: all three optimizations stay on. Closes #6684.

### Plumbing
- New fields `bal_parallel_exec_enabled`, `bal_prefetch_enabled`, and `bal_parallel_trie_enabled` on `BlockchainOptions` (each default `true`).
- `--no-bal-parallel-exec` is threaded as a bool through `Evm::execute_block_pipeline` → `LEVM::execute_block_pipeline`, where it ANDs into the `if let Some(bal) = header_bal` gate so the function falls through to the existing sequential path (which already handles `is_amsterdam` by recording its own BAL; the caller validates against the header).
- `--no-bal-prefetch` is read in `Blockchain::execute_block_pipeline` and selects the warmer strategy based on both toggles:

| `--no-bal-prefetch` | `--no-bal-parallel-exec` | Warmer does |
|---|---|---|
| off (default) | * | `warm_block_from_bal` (cheap; helps both serial and parallel executor) |
| on | off | skip warming (parallel warmer would just compete with parallel executor) |
| on | on | `warm_block` speculative replay (mirrors pre-Amsterdam: serial executor + parallel warmer = proven win) |

- `--no-bal-parallel-trie` is read in `Blockchain::execute_block_pipeline` and gates the `synthesize_bal_updates` call: when disabled, `optimistic_updates` stays `None`, so the merkleizer thread takes the streaming `handle_merkleization` branch fed by the EVM-side `bal_to_account_updates` send.

### Channel allocation (BAL path)
Two consumers want the streaming channel: the merkleizer thread when `parallel_trie` is off, and `LEVM`'s sequential path when `parallel_exec` is off (its `send_state_transitions_tx` requires a Sender). The channel is now allocated whenever **either** side needs it; only the all-defaults cell can skip it.

| `parallel_exec` | `parallel_trie` | LEVM path | Merkleizer | Channel |
|:---:|:---:|---|---|:---:|
| on | on | parallel; no streaming | optimistic (uses synthesized map) | None |
| on | off | parallel; sends one `bal_to_account_updates` batch | streaming (reads `rx`) | Some |
| off | on | sequential; per-tx `send_state_transitions_tx` | optimistic (channel batch dropped at scope exit) | Some |
| off | off | sequential; per-tx `send_state_transitions_tx` | streaming | Some |

The `off / on` cell was a pre-existing latent bug: the channel was keyed only on whether the merkleizer wanted streaming, so `--no-bal-parallel-exec` alone hit the sequential path's `merkleizer.is_none()` error on Amsterdam+ BAL blocks. Now fixed.

### Context
Part of a wider effort to surface per-optimization toggles for the BAL devnets (see the [bal-devnet-6 feature-flags notes](https://notes.ethereum.org/@ethpandaops/bal-devnet-6#Feature-flags) for what other clients expose: reth's `--engine.disable-bal-parallel-state-root`; geth's `--bal.executionmode=sequential|nobatchio`).

The optimistic-trie path (`synthesize_bal_updates` + `handle_merkleization_bal_from_updates`) landed on main via #6655, which unblocked the third toggle and #6684.

## Test plan
- [ ] `cargo check --workspace --all-targets` (verified locally)
- [ ] `cargo clippy -p ethrex-blockchain -p ethrex-vm -p ethrex` clean (verified locally)
- [ ] `cargo fmt --all -- --check` (verified locally)
- [ ] Manual: run a sync with `--no-bal-parallel-exec` alone and confirm Amsterdam+ blocks still validate (sequential fallback uses the Sender; merkleizer still goes optimistic).
- [ ] Manual: run with `--no-bal-prefetch` alone on an Amsterdam block and confirm the warmer thread no-ops.
- [ ] Manual: run with `--no-bal-prefetch --no-bal-parallel-exec` and confirm `warm_block` speculative replay runs.
- [ ] Manual: run with `--no-bal-parallel-trie` on an Amsterdam block and confirm merkleization takes the streaming path (channel-fed) and the state root still matches the header.
- [ ] Manual: run with all three `--no-*` flags and confirm sequential exec + streaming merkleization + speculative warmer all engage.
- [ ] Hive \`bal-quick\` green with defaults (no regression).

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -515,15 +515,28 @@ impl Blockchain {
                         // Precompile cache lives inside CachingDatabase, shared automatically.
                         let start = Instant::now();
                         if let Some(bal) = bal {
-                            // Amsterdam+: BAL-based precise prefetching (no tx re-execution).
-                            // Skipped when --no-bal-prefetch is set; we don't fall back to
-                            // speculative warm_block here because tx replay would be wasted
-                            // work when the header BAL already drives execution.
-                            if bal_prefetch_enabled
-                                && let Err(e) =
+                            if bal_prefetch_enabled {
+                                // Amsterdam+: BAL-based precise prefetching (no tx re-execution).
+                                if let Err(e) =
                                     LEVM::warm_block_from_bal(bal, caching_store, cancelled_ref)
-                            {
-                                debug!("BAL warming failed (non-fatal): {e}");
+                                {
+                                    debug!("BAL warming failed (non-fatal): {e}");
+                                }
+                            } else if !bal_parallel_exec_enabled {
+                                // --no-bal-prefetch combined with --no-bal-parallel-exec:
+                                // mirror the pre-Amsterdam setup where a parallel speculative
+                                // warmer races ahead of the serial executor. With parallel
+                                // exec still on, we skip warming instead — two parallel passes
+                                // over the same txs would just fight for cores.
+                                if let Err(e) = LEVM::warm_block(
+                                    block,
+                                    caching_store,
+                                    vm_type,
+                                    &NativeCrypto,
+                                    cancelled_ref,
+                                ) {
+                                    debug!("Block warming failed (non-fatal): {e}");
+                                }
                             }
                         } else {
                             // Pre-Amsterdam / P2P sync: speculative tx re-execution
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -233,6 +233,14 @@ pub struct BlockchainOptions {
     /// warmer thread and the executor. Set to false (via `--no-precompile-cache`) to
     /// disable the cache for benchmarking purposes.
     pub precompile_cache_enabled: bool,
+    /// If true (default), Amsterdam+ validation runs transactions in parallel
+    /// using the header BAL to seed per-tx databases. Set to false (via
+    /// `--no-bal-parallel-exec`) to fall back to sequential execution.
+    pub bal_parallel_exec_enabled: bool,
+    /// If true (default), Amsterdam+ validation spawns a warmer thread that
+    /// prefetches accounts, storage slots, and codes listed in the header BAL.
+    /// Set to false (via `--no-bal-prefetch`) to skip prefetching on the BAL path.
+    pub bal_prefetch_enabled: bool,
 }
 
 impl Default for BlockchainOptions {
@@ -244,6 +252,8 @@ impl Default for BlockchainOptions {
             max_blobs_per_block: None,
             precompute_witnesses: false,
             precompile_cache_enabled: true,
+            bal_parallel_exec_enabled: true,
+            bal_prefetch_enabled: true,
         }
     }
 }
@@ -471,6 +481,8 @@ impl Blockchain {
         vm.db.store = caching_store.clone();
 
         let cancelled = AtomicBool::new(false);
+        let bal_parallel_exec_enabled = self.options.bal_parallel_exec_enabled;
+        let bal_prefetch_enabled = self.options.bal_prefetch_enabled;
 
         // Synthesize BAL updates pre-scope so the merkleizer thread can start
         // trie work immediately, in parallel with execution.
@@ -503,9 +515,13 @@ impl Blockchain {
                         // Precompile cache lives inside CachingDatabase, shared automatically.
                         let start = Instant::now();
                         if let Some(bal) = bal {
-                            // Amsterdam+: BAL-based precise prefetching (no tx re-execution)
-                            if let Err(e) =
-                                LEVM::warm_block_from_bal(bal, caching_store, cancelled_ref)
+                            // Amsterdam+: BAL-based precise prefetching (no tx re-execution).
+                            // Skipped when --no-bal-prefetch is set; we don't fall back to
+                            // speculative warm_block here because tx replay would be wasted
+                            // work when the header BAL already drives execution.
+                            if bal_prefetch_enabled
+                                && let Err(e) =
+                                    LEVM::warm_block_from_bal(bal, caching_store, cancelled_ref)
                             {
                                 debug!("BAL warming failed (non-fatal): {e}");
                             }
@@ -540,7 +556,13 @@ impl Blockchain {
                 let execution_handle = std::thread::Builder::new()
                     .name("block_executor_execution".to_string())
                     .spawn_scoped(s, move || -> Result<_, ChainError> {
-                        let result = vm.execute_block_pipeline(block, tx, queue_length_ref, bal);
+                        let result = vm.execute_block_pipeline(
+                            block,
+                            tx,
+                            queue_length_ref,
+                            bal,
+                            bal_parallel_exec_enabled,
+                        );
                         cancelled_ref.store(true, Ordering::Relaxed);
                         let (execution_result, produced_bal) = result?;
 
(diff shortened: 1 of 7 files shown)


ilitteri (approved):

crates/blockchain/blockchain.rs
```diff
@@ -467,25 +477,45 @@ impl Blockchain {
...
-                            {
-                                debug!("BAL warming failed (non-fatal): {e}");
+                            if bal_prefetch_enabled {
+                                // Amsterdam+: BAL-based precise prefetching (no tx re-execution).
+                                if let Err(e) =
+                                    LEVM::warm_block_from_bal(bal, caching_store, cancelled_ref)
+                                {
+                                    debug!("BAL warming failed (non-fatal): {e}");
+                                }
+                            } else if !bal_parallel_exec_enabled {
```
ElFantasma, line 539: **The truth table in the PR body is the contract, and nothing in CI locks it down.** The body's table specifies 3 distinct warmer paths:

| `--no-bal-prefetch` | `--no-bal-parallel-exec` | Warmer |
|---|---|---|
| off | * | `warm_block_from_bal` |
| on | off | skip |
| on | on | `warm_block` speculative |

The code branches at lines 496-531 match this table. The test plan has 3 `[ ]` manual entries covering each row, but no automated test that flips the flags and asserts the right path runs.

If a future refactor of the warmer thread (different parallelism strategy, BAL2-aware prefetcher, etc.) accidentally collapses one of these paths or wires the flags wrong, only the manual run would catch it. Hive `bal-quick` exercises only the default (both enabled).

The minimal locking is a unit test using a test-only counter or metric: spin up a `Blockchain` with each flag combo, drive a block through, assert which warmer fn was called. Or, less invasive: add a `tracing::debug!("warmer: <strategy>")` line at each of the three branches and assert via a tracing-subscriber in the test that the right log fires.

Non-blocking — the flags are stable until someone refactors the warmer. Worth doing as a follow-up to make the truth table CI-enforceable.

crates/vm/backends/levm/mod.rs
```diff
@@ -382,10 +384,13 @@ impl LEVM {
...
         // `execute_block` instead. Adding dummy let to avoid unused warnings.
-        let _ = header_bal;
+        let _ = (header_bal, bal_parallel_exec_enabled);
         #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
-        // When BAL is provided (Amsterdam+ validation path): use parallel execution
-        if let Some(bal) = header_bal {
+        // When BAL is provided (Amsterdam+ validation path): use parallel execution.
+        // `--no-bal-parallel-exec` opts out and falls through to the sequential pipeline below.
+        if let Some(bal) = header_bal
+            && bal_parallel_exec_enabled
```
ElFantasma, line 403: nit: the new conjunct `&& bal_parallel_exec_enabled` joins an existing 1-conjunct gate. As soon as #6671 (bal-devnet-7) lands its `&& is_amsterdam` addition (see the [#6671 review](https://github.com/lambdaclass/ethrex/pull/6671#pullrequestreview-4323453472) inline #3 in the body), this gate becomes 3 conjuncts.

Readability suggestion for whichever PR lands second: pull the gate into a named helper so the intent stays clear as the conjunct list grows:

```rust
fn should_use_parallel_bal_exec(
    header_bal: Option<&BlockAccessList>,
    bal_parallel_exec_enabled: bool,
    is_amsterdam: bool,
) -> bool {
    header_bal.is_some() && bal_parallel_exec_enabled && is_amsterdam
}
```

Then `if let Some(bal) = header_bal.filter(|_| should_use_parallel_bal_exec(...))`. Out of scope for this PR, just flagging the pattern.

ElFantasma (approved):
(pre-existing, dependent on #6671):** the LEVM gate at line 392 is `if let Some(bal) = header_bal && bal_parallel_exec_enabled` — but it doesn't include `is_amsterdam`. A pre-Amsterdam block with a `Some(bal)` would enter the parallel path in release (the inner `debug_assert!(is_amsterdam, …)` in `execute_block_parallel` is a no-op there). #6671's review draft proposed adding `&& is_amsterdam` to this same gate. If #6671 lands first, this PR will need a rebase to keep the `is_amsterdam` conjunct; if #6683 lands first, the #6671 fix will need to add `is_amsterdam` to the new 3-conjunct gate. Worth coordinating the merge order.

azteca1998 (approved):

Issue #6685, opened 2026-05-20
Tracking: deep-reorg support via reverse-diff journal + overlay


edg-l:
## Goal

Lift ethrex's 128-block reorg cap (`crates/blockchain/fork_choice.rs:28`) so we can handle reorgs up to the finalized boundary. Parent: #5564.

Today the cap is an implementation constraint, not a policy choice: the in-memory `TrieLayerCache` retains only the last ~128 blocks of forward diffs, and once those flush to disk, the pre-images are gone. The spec only requires us not to reorg past the finalized prefix, so the physical limit should be `unfinalized window`, not 128.

## Progress

All four PRs are now open and form a stack (each based on the previous):

- PR 1 (journal write path) ; open, in review: #6686 (base `main`)
- PR 2 (overlay + read cascade) ; open, draft, stacked on PR 1: #6687
- PR 3 (deep-reorg orchestration) ; open, draft, stacked on PR 2: #6689
- PR 4 (lift the cap + observability) ; open, draft, stacked on PR 3: #6724

## Approach

Persist a reverse-diff journal alongside each commit, and reconstruct an in-memory overlay representing "disk state minus the path back to the pivot" when a deep reorg is requested. On-disk state stays at the old chain's edge until the first new-chain commit atomically reconciles overlay + new layer.

Heavily informed by the architecture in #6561 (Tomás' first iteration). Closest to erigon's journaling model; geth/nethermind/reth take a different tack (trust CL + path-based state with retained ancestors).

### Pieces

1. **State-history journal** ; new `STATE_HISTORY` CF, keyed by `block_number.to_be_bytes()`. One reverse-diff entry per committed block covering the four state CFs (`ACCOUNT_TRIE_NODES`, `STORAGE_TRIE_NODES`, `ACCOUNT_FLATKEYVALUE`, `STORAGE_FLATKEYVALUE`). Codes not journaled (content-addressed). Hand-rolled compact codec (versioned, varint-prefixed). Written in the same `WriteBatch` as the trie commit. Skipped when `batch_mode == true` (full sync doesn't support reorgs).

2. **Pruning at finality** ; `forkchoice_update_inner` issues `delete_range(STATE_HISTORY, 0, finalized+1)` in the same atomic txn when finalized advances. Journal size bounded by unfinalized window.

[...]

PR #6686, opened 2026-05-20
feat(l1): state-history journal foundation for deep reorgs

edg-l:
> Part of the deep-reorg support effort tracked in #6685. This is PR 1 of 4 and lays the on-disk audit trail; no reorg consumer is wired yet.

## Summary

- Adds a `STATE_HISTORY` column family that records a per-block reverse-diff journal: for every trie node or flat-KV entry overwritten at commit time, the pre-image is stored atomically in the same write batch.
- Finality pruning deletes the journal up to the new finalized block via a single `delete_range` call.
- Supersedes the earlier attempt in #6561.

## What's in this PR

- `crates/storage/journal.rs` (new, 395 LOC): `JournalEntry` struct, hand-rolled varint codec, 8 codec unit tests.
- `crates/storage/api/mod.rs`: `delete_range` method added to `StorageWriteBatch` trait.
- `crates/storage/backend/rocksdb.rs`: `delete_range_cf` impl.
- `crates/storage/backend/in_memory.rs`: iterate-and-retain `delete_range` impl.
- `crates/storage/api/tables.rs`: new `STATE_HISTORY` table constant.
- `crates/storage/layering.rs`: `TrieLayer` carries `(block_number, block_hash)`; `commit()` returns `CommitResult` with block identity instead of bare `Vec`.
- `crates/storage/store.rs`: `apply_trie_updates` opens a read view before `begin_write`, collects pre-images per CF into `FlatDiff`s, stages the `JournalEntry`; `forkchoice_update_inner` prunes the journal when finalized advances. 5 new integration tests at bottom.
- `crates/storage/trie.rs`: extracted `classify_trie_key` helper (was duplicated across three call sites).

## What's NOT in this PR

- Reorg consumer / state rollback (PR 2: overlay + read cascade).
- Lifting the 128-block reorg cap (PR 4).
- Hot-path pre-image reads are O(N) per block; PR 4 can batch via `multi_get_cf`.
- In-memory backend `commit()` is a pre-existing no-op (FIXME); the new `delete_range` inherits that but it is not a regression.
- Journaling is skipped entirely during full sync (`batch_mode == true`).

## Test plan

- 16 unit + integration tests pass on both InMemory and RocksDB backends (`cargo test -p ethrex-storage --features rocksdb`).
- `cargo fmt --check` clean.
- `cargo clippy -p ethrex-storage` clean on all touched files.

Closes part of #6685. Supersedes #6561.


edg-l:
Pushed a fixup addressing all bot findings. Short responses:

**Greptile / Claude — varint 10th-byte truncation** ; fixed with `shift==63 && (b & 0x7e != 0 || b & 0x80 != 0)` guard. Tests added: truncating 10th byte, 11th byte, continuation-at-byte-10, and `u64::MAX` round-trip still works.

**Codex / Claude — `Vec::with_capacity` OOM on malformed count/path_len/value_len** ; new `LengthExceedsRemaining` error variant; all three length prefixes now bounded against `cur.remaining()` before allocating. 3 OOM-attack tests.

**Greptile / Codex — `prev_finalized` silent fallback to 0** ; length mismatch now returns `StoreError::Custom` instead. Test plants a 4-byte value and asserts the error path + that no pruning occurred.

**Codex — `decode()` accepts trailing bytes** ; new `TrailingBytes` variant + end-of-decode check. Test added.

**Claude — `assert_no_journal_entry` 50ms sleep is CI-fragile** ; now polls every 20ms over a 500ms window, fails loudly on any observation.

**Claude — `CommitResult: Default` footgun** ; removed the `Default` derive; doc-comment explains why.

**Claude — `read_slice` subtraction underflow** ; `saturating_sub` form, added `remaining()` helper.

**Claude — `diff_byte_estimate` under-estimates** ; new `varint_len()` helper; estimate now matches encoded length exactly (asserted by test).

**Claude — multi-layer commit attribution** ; added code comment explaining the single-layer invariant.

**Claude — in-memory `commit()` FIXME** ; expanded to call out `delete_range` and the multi-op atomicity gap.

26 storage tests pass on InMemory and RocksDB; clippy + fmt clean.

crates/storage/journal.rs
```diff
@@ -0,0 +1,625 @@
...
+//! `block_hash` (32 bytes), `parent_state_root` (32 bytes), then four
+//! varint-prefixed reverse-diff sections in order: account-trie, storage-trie,
+//! account flat-KV, storage flat-KV. RLP/bincode/postcard are skipped — the
+//! access pattern (write-once, read-on-reorg, large volume) makes encode/decode
+//! cost matter.
+
+use ethrex_common::H256;
+
+/// Current version of the journal entry codec.
+pub const JOURNAL_VERSION: u8 = 1;
```
ElFantasma, line 38: **Version-migration strategy.** `JOURNAL_VERSION = 1` and `decode()` hard-errors on `VersionMismatch`. This is correct for **forward** safety (a v1 binary refuses to interpret v2 entries rather than silently producing a malformed reverse-diff). But the symmetric concern is **backward** safety: when a v2 binary lands, it will see v1 entries on disk written by the previous binary, and the current decoder will reject those too — leaving the v1 journal effectively unreadable post-upgrade.

For the rollback consumer (PR 2 and beyond), that means either:
1. Every codec bump must include a one-shot migration step that re-encodes existing v1 entries into v2, or
2. The decoder grows a version-dispatch table that can decode v1, v2, … (e.g., `match version { 1 => decode_v1(...), 2 => decode_v2(...), other => Err(...) }`), or
3. Treat all pre-bump entries as discardable (acceptable only if rollback-from-pre-bump-state is not a supported scenario).

The doc comment at line 23 calls out the hard-fail design but doesn't pick a strategy. Worth adding one line saying which path PR 2/3/4 will take — "v2 decoder will fall back to `decode_v1` for older entries" or "a finality boundary at the upgrade point drains the v1 journal before v2 entries are written" or similar. Future-you will thank present-you when v2 lands.

crates/storage/layering.rs
```diff
@@ -241,15 +252,36 @@ impl TrieLayerCache {
...
+        // ATTRIBUTION NOTE: In normal block-by-block sync there is only one layer
+        // to commit, so attributing the journal entry to `state_root`'s block is
+        // correct. If a future caller triggers a multi-layer commit (e.g. by
+        // raising the threshold and then forcing a flush), `nodes_to_commit`
+        // below would merge diffs from several blocks while the journal entry
+        // would still be tagged with only the newest block's identity ; the
+        // rollback consumer (PR 2/3) would then be unable to reconstruct
+        // intermediate pre-images. Single-layer commits are an invariant of the
+        // current write path; revisit this if that ever changes.
+        let top_layer = layers_to_commit.first()?;
```
ElFantasma: **Attribution invariant deserves a `debug_assert!`.** The block-comment above (lines 271-279) correctly identifies the multi-layer commit foot-gun: if a future caller flushes >1 layer at once, `nodes_to_commit` merges diffs from several blocks while the journal entry is tagged with only the newest block's identity, leaving the rollback consumer unable to reconstruct intermediate pre-images.

The comment ends with "Single-layer commits are an invariant of the current write path; revisit this if that ever changes" — but nothing structurally enforces it. A future PR that bumps `BATCH_COMMIT_THRESHOLD` or adds a `flush_all()` helper could quietly violate the invariant and corrupt the journal without any test failure (the existing tests use `commit_threshold = 1`, so they only exercise the single-layer path).

Add `debug_assert!(layers_to_commit.len() == 1, "multi-layer commit would corrupt journal attribution; see comment above");` right after the `while`-loop populates `layers_to_commit`. Dev/CI catches the violation immediately; release-mode behavior unchanged. (If multi-layer commits ARE intended to be supported eventually, the `debug_assert!` becomes the explicit re-design trigger.)

crates/storage/store.rs
```diff
@@ -2986,8 +3048,20 @@ fn apply_trie_updates(
...
+    //
+    // NOTE: `StorageReadView` does not currently promise true snapshot isolation
+    // (see the trait docs in `api/mod.rs`). What makes the pre-image read safe here
+    // is the single-writer invariant: `apply_trie_updates` is only ever called from
+    // the trie-update worker thread, and `write_tx` is a buffered batch that does
+    // not become visible until `write_tx.commit()` at the end of the function. So
+    // every `.get()` below sees on-disk state as of the begin_read call. PR 2's
+    // overlay work will need to revisit this if other write paths to trie CFs are
+    // introduced.
+    let read_view = backend.begin_read()?;
```
ElFantasma, line 4071: **Single-writer invariant is structural, not asserted.** The comment block above (lines 3051-3062) is correct: `apply_trie_updates` is only ever called from one trie-update worker thread (verified — line 1656-1668 is the sole call site, inside a single `std::thread::spawn`). And `write_tx` is a buffered batch that doesn't become visible until commit. So `.get()` on `read_view` sees disk-as-of-`begin_read` for the duration.

But the invariant is enforced by *the topology of who calls this function*, not by anything in this function itself. PR 2's overlay work, or any future write path to trie CFs (background compaction helpers, repair tools, …), could quietly violate it without breaking tests. The comment ends with "PR 2's overlay work will need to revisit this if other write paths to trie CFs are introduced" — but "will need to revisit" is easy to forget under deadline pressure.

Cheap insurance: a `static TRIE_WRITER_PRESENT: AtomicBool = AtomicBool::new(false);` claimed via `compare_exchange` at the top of `apply_trie_updates` (with a `defer`/`Drop` guard to release it) would surface a concurrent caller as a panic immediately, while costing one atomic op per call on the hot path. Or a `debug_assert!` checking `std::thread::current().name()` matches the trie-worker thread name.

Not blocking — current correctness holds. The suggestion is to make the invariant runtime-enforced rather than convention-enforced.

ElFantasma (commented):
Three inline findings (all minor) + a body note.

**Body finding:** the body's checklist says "no `STORE_SCHEMA_VERSION` bump (no schema migration required)". Verified — `create_missing_column_families(true)` (rocksdb.rs:34) creates `STATE_HISTORY` on the fly when a v13 binary opens a v12 DB, and existing CFs are untouched. Downgrade (v12 binary on v13 DB) is rejected by the `TABLES.contains(&cf_name)` check at rocksdb.rs:203, which is correct one-way-migration behavior. Worth noting explicitly in the body or CHANGELOG so an operator considering rollback knows the v12 binary won't open a v13 DB.

crates/storage/journal.rs
edg-l: Added a policy note on the JOURNAL_VERSION const in f01b3564da: drain at finality on bump by default, per-version decode arms here if history needs to survive an upgrade.

crates/storage/layering.rs
edg-l: debug_assert! added in f01b3564da. Traced both modes: with the post-commit retain prune, layers_to_commit.len() == 1 holds for non-batch (threshold=128) and batch (threshold=4) on the first commit and steady-state. All 26 storage tests pass with the assertion live.

crates/storage/store.rs
edg-l: Leaving as a comment-enforced invariant for this PR; PR 2 (overlay) is the natural place to add a runtime guard when other write paths to trie CFs are introduced.

ElFantasma (approved):

azteca1998 (dismissed):

crates/storage/journal.rs
```diff
@@ -0,0 +1,635 @@
...
+    pub account_trie_diff: FlatDiff,
+    /// Reverse diff for `STORAGE_TRIE_NODES`. Keys carry the nibble-encoded
+    /// account-hash prefix as written on disk; no separate grouping is needed.
+    pub storage_trie_diff: FlatDiff,
+    /// Reverse diff for `ACCOUNT_FLATKEYVALUE`.
+    pub account_flat_diff: FlatDiff,
+    /// Reverse diff for `STORAGE_FLATKEYVALUE`. Keys carry the nibble-encoded
+    /// account-hash prefix as written on disk.
+    pub storage_flat_diff: FlatDiff,
+}
```
MegaRedHand, line 74: Did we measure how much impact does this have on the DB size when journal entries accumulate for a day?

MegaRedHand (approved):
LGTM

crates/storage/journal.rs
edg-l: yes, it prunes on finality, if a network doesnt finalize for an entire day it can go from 2gb to 7gb, on a finalizing healthy network its not much more than in the order of 10mbs

ElFantasma (approved):
Re-review after clippy Box + docs tweak — still LGTM, no new issues.

iovoid (approved):

PR #6687, opened 2026-05-20
feat(l1): overlay + read cascade for deep reorgs

edg-l:
> Part of the deep-reorg support effort tracked in #6685. This is PR 2 of 4, stacked on #6686. Adds the in-memory consumer of the journal; no orchestration yet.

## Summary

- Adds `Overlay` struct in `crates/storage/layering.rs` that aggregates reverse-diffs from `STATE_HISTORY[D..=T]` (descending, older-entry-wins) to reconstruct virtual state at pivot `T-1`.
- Extends `TrieLayerCache` with optional overlay field and wires `TrieWrapper::get` to cascade: layer cache -> overlay -> disk.
- 14 new unit tests; 30 storage tests total pass on both InMemory and RocksDB backends.

## What's in this PR

- `Overlay` struct: walks the journal range descending, applies older-entry-wins semantics, holds a bloom filter for fast miss short-circuits.
- `OverlayCf` enum + `classify_by_key_length`: routes CF entries to the right column family using the same key-length logic as PR 1's `classify_trie_key`.
- `OverlayError::InvalidRange { from_block, to_block }`: hard error for swapped caller args (not an I/O error variant).
- `TrieLayerCache` set/clear/lookup accessors for the overlay field.
- `TrieWrapper::get` cascade hook integrated into the existing read path.
- 14 unit tests: descending-range load, older-wins, bloom miss, hash mismatch, missing entry, swapped-args hard error, single-entry-at-genesis, empty-but-present round trip, CF classification, no-overlay short-circuit, set/clear, iter_all_entries.

## What's NOT in this PR

- Orchestration: `apply_fork_choice_with_deep_reorg`, `install_overlay_for_reorg`, and the reorg-in-progress flag are in PR 3.
- The 128-block reorg cap is not lifted; that's PR 4.
- `from_block` accessor rename (triggers clippy `wrong_self_convention`, suppressed); deferred to PR 3.
- Swapping `AtomicBloomFilter` for a non-atomic variant during construction for perf; deferred to PR 3 or PR 4.

## Test plan

- `cargo fmt` clean.
- `cargo clippy` clean.
- 14 new tests added in `crates/storage/layering.rs`.
- `cargo test -p ethrex-storage` passes all 30 tests on both InMemory and RocksDB backends.

## References

- Parent issue: #6685
- PR 1 (state history journal): #6686


crates/storage/layering.rs
```diff
@@ -380,9 +446,18 @@ impl TrieDB for TrieWrapper {
...
+        // Read cascade: forward layer cache (new-chain layers above the pivot) ->
+        // overlay (reverse-diff bridge to disk during deep reorgs, if installed) ->
+        // on-disk state. A layer-cache hit pre-empts the overlay because a
+        // side-chain write at this key supersedes the pivot value the overlay holds.
+        // An overlay hit pre-empts disk because disk still reflects the OLD chain's
+        // edge `D`, not the pivot.
         if let Some(value) = self.inner.get(self.state_root, key.as_ref()) {
             return Ok(Some(value));
         }
+        if let Some(overlay_result) = self.inner.lookup_overlay(key.as_ref()) {
```
ElFantasma: Endorsing greptile's P1 with a specific design constraint for PR 3: the `lookup_overlay(key)` call is unconditional on `self.state_root`. Because `TrieLayerCache` is swapped under an outer `RwLock<Arc<...>>` (see `store.rs:183` / `gated_snapshot` line 3172), every TrieWrapper opened **after** `set_overlay` gets an Arc pointing at the overlay-attached cache. For such a TrieWrapper reading at a state_root R that isn't the pivot (e.g. an eth_call at cache-edge D, or a state root within the layer chain on the old-chain side), a key K that (1) is not diffed in the layer chain between R and D and (2) was modified between T and D will short-circuit at the overlay and return the pivot value — where the correct answer is the D-value that disk still holds (since disk is not mutated during the overlay window).

The cascade docstring at line 449-454 reasons about the intended reader (the new-chain forward-execution TrieWrapper) but doesn't distinguish it from unrelated readers of the same shared cache. PR 3 needs one of:

1. Gate `lookup_overlay` on a set of "consuming" state roots (starts as `{pivot}`, grows as new-chain layers are added).
2. Route historical/RPC reads to a separate `Arc<TrieLayerCache>` (or `open_direct_state_trie` — already exists at line 3200 and skips the cache entirely) for the overlay window.
3. Serialize the overlay window with the RPC read path via the reorg-in-progress flag.

Option 2 looks cheapest, but I don't have visibility into PR 3 yet — flagging so it doesn't slip. Not blocking this PR: `set_overlay` is dead-code-gated (`consumed by PR 3`), so the incorrect-cascade path can't be triggered until PR 3 wires it.

edg-l:
The "Overlay Applies Across Roots" P1 is now fixed in PR3 (#6689, commit `915a68dc4`), per your note that the resolution belongs there.

Approach (close to your option 1): the `Overlay` now records `serves_root` — the pivot state it reconstructs, captured from the `parent_state_root` of the journal entry at `to_block`. A new `TrieLayerCache::overlay_serves(root)` gates the read cascade in `TrieWrapper::get`: the overlay is consulted only when the reader's `state_root` is the pivot (`serves_root`) or a new-chain layer root present in the cache (the roots forward replay produces). Every other reader — an eth_call/getProof at the old cache-edge `D`, or any unrelated historical root — falls through to disk, which still holds that root's canonical state during the overlay window.

Note this covers storage reads too: `open_storage_trie` builds its `TrieWrapper` with `state_root` = the account/state root (storage_root goes to `Trie::open`, account_hash is the prefix), so state and storage reads during replay all carry the block's state root and pass the gate. Verified end-to-end: the `deep_reorg_state_oracle` simulator scenario (storage + balance reads after a 150-deep reorg) passes, plus new unit tests `overlay_serves_only_consuming_roots` and `from_journal_captures_serves_root_from_to_block_entry`.

ElFantasma (approved):

ElFantasma (approved):

ilitteri:
Audit of this PR (effective diff `main..feat/state-history-overlay`, head `93769cc88`). The Overlay design itself is solid — oldest-pre-image-wins semantics verified against the per-block journaling, CF classification correctly centralized on `classify_trie_key`, bloom filter safe, unit tests genuinely good. But the Overlay API is dead code until #6689; the only *live* change is in the commit gate, and it regresses a previously fixed case.

## High — layerless safe-commit root now hard-errors instead of no-op'ing (regression of a fixed case; regression tests deleted)

`crates/storage/layering.rs:223-226`, `crates/storage/store.rs:4077-4081`.

Before this PR, branch (c) of `get_commitable` was:

```rust
// (c) ... Still requires a layer: a canonical root need not have one
// (`put_batch` skips blocks whose state root equals their parent's, which on L2 is every empty block).
if parent_state_root == safe_root {
    return self.has_layer(safe_root).then_some(safe_root);
}
```

and `commit_to_disk` began with a `has_layer(root)` bail-out naming the two legitimate layerless cases (empty L2 blocks; already-flushed roots). The PR removes both guards: branch (c) returns `Some(safe_root)` for a layerless root, and `commit_to_disk` turns `commit(root) == None` into a hard `StoreError`. `TrieLayerCache::commit` returns `None` exactly when the root has no layer (`layering.rs:377`). Reachable consequences:

- **L2 steady state:** `compute_safe_commit_root` returns the state root of `head - 128`; on L2 any empty block has `state_root == parent_state_root` and no layer. Every time the safe-commit root lands on such a block, the forkchoice path sends `PersistMessage::Commit(root)` (`store.rs:2873-2876`) straight into the hard error. After a run of ≥128 root-preserving blocks, the error propagates into `last_flush_result` and is delivered in the next block's persist ack (`store.rs:1991-2002`, consumed with `??` at `store.rs:1778-1780`) — `add_block` starts failing on a healthy chain. Previously a documented, correct no-op.
- **L1/L2 re-commit:** after the safe root's layer is flushed (while the cell still holds that root), any payload executed with `parent_state_root == safe_root` (a fork block at exactly `safe+1`, or on L2 any root-repeating block) hits branch (c) and hard-errors, where branch (d) correctly returns `None`.
- The PR **deletes the two regression tests that pinned this behavior** (`parent_equals_safe_root_without_layer_yields_none`, `already_committed_safe_root_yields_none`) instead of updating them.

The invariant the new error message claims ("`root` was returned by a `get_commitable*` gate so `commit(root)` should always return `Some`") is false for branch (c) as rewritten, and was never true for the `PersistMessage::Commit` path, which forwards the cell's root without consulting the cache.

## Medium — the new early return strands the flat-KV generator (Stop without Continue)

`crates/storage/store.rs:4040` (Stop), `4077-4081` (early `?`), `4201` (Continue).

`commit_to_disk` sends `FKVGeneratorControlMessage::Stop` before discovering there is nothing to commit, and the new `?` returns before the matching `Continue` — all later error paths deliberately fall through to 4201 ("We want to send this message even if there was an error"). While initial FKV generation is running, an unmatched Stop parks the generator in `control_rx.recv()` awaiting Continue; the next real commit's Stop then arrives in that slot and the generator thread exits permanently with "Unexpected Stop message" (`store.rs:4325-4329`, thread ends at `store.rs:1955`). Flat-KV generation is dead until restart — and per the finding above, on L2 the triggering error is routine.

## Low — stale `commit_to_disk` docstring

`crates/storage/store.rs:4025` still states "A `root` that is not a layer commits nothing." Post-PR it hard-errors; the stale doc masks the first finding for future readers.

## Suggestion

Either restore the `has_layer` check in branch (c) and the early bail in `commit_to_disk` (keeping the regression tests), or demonstrate the layerless cases are unreachable and update tests/docs accordingly. The Overlay machinery itself needs no changes — noted non-issues: absorb/oldest-wins correctness, cascade ordering (layer → overlay → disk, `Some(None)` suppressing stale disk values), the acknowledged `from_journal`/pruning race (acceptable for dead code, but the PR-3 reorg-in-progress gate is a hard prerequisite for wiring it up).

ilitteri:
Pushed three commits addressing the findings from the review comments above (audits of this PR, #6689, #6724, and #7001):

**`e8477f6a2` — orchestration fixes**
- **Canonical-head pivot (High)**: the head-canonical case installed the overlay serving head's *parent* root while head was never re-executed, so head's state never became readable (unrecoverable FCU loop). The pivot is now `head.number` for that case (unwind `[head+1, edge]`, `serves_root = head.state_root`).
- **Multi-layer reconciliation (High)**: with an overlay installed, `commit_to_disk` now commits only the bottom layer per pass. Previously a backlog flush swept multiple layers, tripping the `debug_assert!` in debug builds and journaling upper layers' pre-images against old-chain disk in release (silent corruption on a future unwind). The backlog drains after the first reconciliation commit clears the overlay.
- **`abort_reorg` rendezvous (Medium)**: it now Ping-rendezvouses the persist worker before swapping the cache, exactly like `install_overlay_for_reorg` (extracted into `rendezvous_persist_worker`).
- **Journal-pruning race (Medium)**: `enter_reorg` now pauses `STATE_HISTORY` pruning for the duration of the apply pass (new `Store::set_journal_pruning_paused`); syncer-driven FCUs bypass the reorg mutex and could prune entries out from under `Overlay::from_journal`. Pruning catches up on the next finality advance.
- **`newPayload` gate (Medium)**: returns SYNCING while a reorg is in progress, closing the window where a concurrent execution could clobber the freshly installed overlay.
- **`ethrex_reorg_depth` histogram (Low)**: now observes `latest - pivot` (matching the `TooDeepReorg` gate's definition) instead of `head - pivot`, which reported ~1 in the canonical-head case regardless of unwind depth.

**`2fef7a6bf` — #7001 (cherry-picked from #7036)**
While an overlay serves the read's state root, all flat-KV fast paths are disabled so reads go through the (always journaled) trie nodes: `TrieWrapper::flatkeyvalue_computed` returns false, the `EMPTY_TRIE_HASH` shortcuts in `get_storage_at_root*` are gated, and `get_account_states_batch_by_root` routes through the trie fallback instead of the direct `ACCOUNT_FLATKEYVALUE` multi_get (which never consulted the overlay at all). Includes the wrapper-gate unit test and the store-level regression test from the issue's acceptance criteria. **This makes #7036 redundant — it can be closed.**

**`0668ac70a` — observability/docs**
- `ethrex_reorg_journal_length` gauge now updated on every successful `apply_fork_choice` (was only set on deep-reorg success; read 0 forever in steady state, contradicting its docs).
- `--max-reorg-depth 0` docs corrected: it rejects **every** reorg with `-38006 TooDeepReorg` (not "restores the pre-stack 128 cap", not "returns SYNCING"). Fixed in the rustdoc, `crates/storage/README.md`, and `docs/l1/running/monitoring.md`; ceiling described as physical, not finality-based.
- Stale finality-ceiling comments in `smoke_tests.rs` / `tooling/reorgs` updated to the physical-ceiling mechanism.

Verified: `ethrex-storage` lib 85✓, `ethrex-blockchain` lib 28✓, `smoke_tests` 12✓, `deferred_persistence` 21✓, clippy clean on the three touched crates.

Not addressed (by design, needs explicit sign-off rather than code): the absence of a default bound on reorg depth/overlay memory when finality stalls (see the #6724 review comment), and the case-1→case-2 journal hash-verification skip.

ilitteri:
Adversarial round on the review fixes (`45abf8477`):

- **One latent bug caught**: reconciliation layer matching by block number (`layer.block_number == overlay.to_block`) lost the bridge fold whenever the bottom committed layer sat above `to_block` — reachable on L2, where root-preserving blocks create no layer. The bridge entries would have been neither written to disk nor journaled before the overlay cleared. Now that the bottom-layer-only rule makes any overlay-backed commit single-layer by construction, the reconciliation layer is identified by overlay presence instead.
- **Two adversarial regression tests added**: `overlay_backed_commit_only_commits_bottom_layer_per_pass` (the 3-layer-backlog shape that panicked the debug_assert pre-fix; asserts bottom-only commit, overlay consumption, bridge keys folded into T's journal entry, correct backlog drain) and `journal_pruning_pauses_during_reorg_and_catches_up_after` (finality advance prunes nothing while paused, prunes cumulatively after release).
- Also verified in this round: the `.await` inside the `metrics!` block type-checks with `--all-features`; case-1's empty replay correctly reaches the final truncating `forkchoice_update`; branch is up to date with `main` (main's tip `ec208dcf0` is already an ancestor).

Suites: storage lib 87, blockchain lib 28, smoke_tests 12, clippy clean.

ilitteri (approved):

PR #6689, opened 2026-05-21
feat(l1): deep-reorg orchestration

edg-l:
> Part of the deep-reorg support effort tracked in #6685. This is PR 3 of 4, stacked on #6687. Wires up the orchestration layer; the deep-reorg path is scaffolded but dormant until PR 4 lifts the cap.

## Summary

- Adds `install_overlay_for_reorg`, `abort_reorg`, `clear_reorg_overlay`, `is_state_in_layer_cache`, and `highest_state_history_block_number` to `store.rs`.
- Section 9 reconciliation in `apply_trie_updates`: on the first new-chain commit after a deep reorg, overlay-only entries are folded into the write batch, the old `STATE_HISTORY[T..=D]` range is deleted, and the overlay is cleared. Disk advances from the OLD chain's edge directly to the new tip in one atomic write.
- `reorg_in_progress: AtomicBool` + `ReorgGuard` RAII in `Blockchain`; concurrent FCUs short-circuit to `SYNCING` while a reorg apply pass is in flight.
- `apply_fork_choice_with_deep_reorg` in `fork_choice.rs`: routes `StateNotReachable` into `reorg_apply_deep`, all other results fall through unchanged. `AbortReorgGuard` resets the cache on any failure or panic.
- Engine API `fork_choice.rs` handler switched to the new wrapper (takes `&context.blockchain`). `try_execute_payload` stashes blocks and returns `ACCEPTED` when the parent is known but its state is not materialized; a later `forkchoiceUpdated` drives the apply.
- Removes stale `#[allow(dead_code)]` markers from layering.rs now that the overlay accessors are consumed.
- 4 storage unit tests for the new primitives; 2 `ReorgGuard` tests including panic-unwind.

## What's in this PR

- `crates/storage/store.rs`: `install_overlay_for_reorg`, `abort_reorg`, `clear_reorg_overlay`, `is_state_in_layer_cache`, `highest_state_history_block_number`, Section 9 reconciliation inside `apply_trie_updates`. 4 new tests.
- `crates/blockchain/blockchain.rs`: `reorg_in_progress: AtomicBool`, `ReorgGuard`, `enter_reorg`, `is_reorg_in_progress`, `store()` accessor. 2 new tests.
- `crates/blockchain/fork_choice.rs`: `apply_fork_choice_with_deep_reorg`, `reorg_apply_deep`, `AbortReorgGuard`, `map_chain_error_for_fcu`. Head is included in the replay iterator via `chain(once((head.number, head_hash)))`, fixing the missing-head bug from #6561.
- `crates/networking/rpc/engine/fork_choice.rs`: switched to `apply_fork_choice_with_deep_reorg`.
- `crates/networking/rpc/engine/payload.rs`: ACCEPTED stash path for unmaterialized parent state.
- `crates/networking/rpc/types/payload.rs`: `PayloadStatus::accepted()` constructor.
- `crates/storage/layering.rs`: drop stale `#[allow(dead_code)]` attrs; `len`/`is_empty` keep theirs (PR 4 target).

## What's NOT in this PR

- The 128-block `REORG_DEPTH_LIMIT` cap is not lifted; the deep-reorg path is unreachable on mainline until PR 4. Behavior for end users is unchanged after this merge.
- `tooling/reorgs` deep-reorg simulator scenario: deferred to PR 4 (requires a CLI knob or cap removal to be end-to-end reachable).
- `map_chain_error_for_fcu` collapses all `ChainError` variants to `StateNotReachable`; a richer mapping with `InvalidAncestor` + last-valid-hash is follow-up.
- Crash-during-reconciliation injection test (open question in #6685): deferred.
- Historical state RPC queries: not in #6685's scope.

## Test plan

- `cargo build --workspace` clean.
- `cargo clippy --workspace --no-deps -- -D warnings` clean.
- `cargo fmt --check` clean.
- `cargo test -p ethrex-storage` (45/45 pass, includes 4 new deep-reorg primitive tests).
- `cargo test -p ethrex-blockchain` (2/2 new `ReorgGuard` tests pass, including panic-unwind).

## References

- Tracking issue: #6685
- PR 1 (state-history journal): #6686
- PR 2 (overlay + read cascade): #6687


crates/blockchain/fork_choice.rs
```diff
@@ -262,3 +274,251 @@ async fn find_link_with_canonical_chain(
...
+pub async fn apply_fork_choice_with_deep_reorg(
+    blockchain: &Blockchain,
+    head_hash: H256,
+    safe_hash: H256,
+    finalized_hash: H256,
+) -> Result<BlockHeader, InvalidForkChoice> {
+    // Short-circuit when a previous deep-reorg apply is still in flight. The CL
+    // retries on SYNCING; once the in-progress reorg completes, the next FCU
+    // is processed normally.
+    if blockchain.is_reorg_in_progress() {
```
ElFantasma, line 310: Endorsing greptile's P1. `is_reorg_in_progress()` at line 300 is a plain `load`, and `enter_reorg()` at blockchain.rs:530 is a plain `store(true)` — not test-and-set. Between line 300 (check), line 305 (`apply_fork_choice`), line 307 (match `StateNotReachable`), and line 339 (`enter_reorg`), a second FCU can observe `false`, drive its own `apply_fork_choice → StateNotReachable → reorg_apply_deep`, and race the first one into `install_overlay_for_reorg`. Both end up executing side-chains against the shared cache; whichever aborts first will `abort_reorg()` and wipe the other's partial state.

Suggestion: make `enter_reorg` a CAS returning `Option<ReorgGuard>`, and shift the check-site to it:

```rust
pub fn try_enter_reorg(&self) -> Option<ReorgGuard<'_>> {
    self.reorg_in_progress
        .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
        .ok()
        .map(|_| ReorgGuard { blockchain: self })
}
```

Then this becomes:

```rust
let Some(_guard) = blockchain.try_enter_reorg() else {
    return Err(InvalidForkChoice::Syncing);
};
```

Applied above the `apply_fork_choice` call (not just above `reorg_apply_deep`) so a concurrent FCU can't even start the shallow-reorg attempt while a deep reorg is racing to install the overlay. Cost: the shallow-reorg path spends the guard, but that's already cheap and the flag is only set for the FCU duration anyway.

(Alternative: keep the plain `store` in `enter_reorg` and just serialise `install_overlay_for_reorg` behind a `tokio::sync::Mutex` inside `reorg_apply_deep`. Slightly cleaner separation of concerns but adds one more sync primitive.)

crates/storage/store.rs
```diff
@@ -4010,6 +4166,23 @@ fn commit_trie_layers(
...
 
+    // Section 9 reconciliation: stage `delete_range` for obsolete OLD-chain
+    // journal entries in `[T, D]` BEFORE the new T entry put, so the put isn't
+    // clobbered by the range delete. T = `overlay.to_block()`, D = `overlay.from_block()`.
+    if result.is_ok()
+        && let Some(overlay) = &overlay_for_reconciliation
+    {
+        let t = overlay.to_block();
+        let d = overlay.from_block();
+        debug_assert_eq!(
```
ElFantasma: Endorsing greptile's P1. `debug_assert_eq!` fires only in debug builds; in release, if `committed_block_number != t`, the code still stages `delete_range(STATE_HISTORY, t..=d)` and then puts a journal entry at `committed_block_number` — potentially outside `[t, d]`. Result: the OLD chain's journal in `[t, d]` is deleted, and the only surviving entry sits at an arbitrary height, leaving the journal with a gap that breaks future overlay construction and finality pruning.

Because of the commit_threshold=128 default vs the 128-block reorg cap, the first commit fires only after the layer chain has accumulated 128 layers on TOP of the reorg — the invariant *should* hold, but the window between overlay install and first commit is long (blocks 128 onward from the deep-reorg tip) and there is no other check ensuring nothing perturbs `trie_cache` in that window (e.g. an `abort_reorg` racing with an in-flight persist as flagged in the ping comment at 3416-3425). This is exactly the kind of invariant that deserves a real check, not a debug-only one:

```rust
if committed_block_number != t {
    return Err(StoreError::Custom(format!(
        "section 9 reconciliation: first overlay-backed commit at block {committed_block_number}, expected pivot T={t}"
    )));
}
```

Placed BEFORE `delete_range` is staged. This keeps the invariant enforced in production and turns a silent-corruption path into a hard, localised error.

crates/storage/store.rs
```diff
@@ -3342,6 +3342,130 @@ impl Store {
...
+                    "install_overlay_for_reorg: failed to ping persist worker: {e}"
+                ))
+            })?;
+        ack_rx.recv().map_err(|e| {
+            StoreError::Custom(format!(
+                "install_overlay_for_reorg: persist worker ping ack failed: {e}"
+            ))
+        })??;
+
+        let mut guard = self.trie_cache.write().map_err(|_| StoreError::LockError)?;
```
ElFantasma, line 3579: Carryover from #6687: the shared `trie_cache` swap makes the overlay visible to every subsequent `TrieWrapper`, not just the deep-reorg's forward-execution reader. Any RPC state read (`eth_call`, `debug_traceCall`, `debug_traceTransaction` at a historical hash) that opens a trie after this swap grabs an `Arc<TrieLayerCache>` with the overlay attached (`gated_snapshot` line 3172 → `.read().clone()`), and its miss-in-layer-chain path falls through to `lookup_overlay` at layering.rs:458 regardless of `self.state_root`. For a key modified between T and D but not present in the freshly-reset layer chain, the RPC read gets the pivot value instead of the D value that disk holds.

`reorg_in_progress` guards concurrent FCUs, but the RPC state-read path doesn't check it. Options for the follow-up:

1. Make the RPC read path check `reorg_in_progress` and either serialise or route through `open_direct_state_trie` (line 3200) while the overlay is alive.
2. Attach a `state_root` (or set of state_roots) to the overlay itself and gate `lookup_overlay` on membership — grows as new-chain layers are appended.
3. Block RPC state serving during the overlay window (simplest, tolerable if deep reorgs are rare and short).

Not blocking this PR — this path is dormant (128-block cap in shallow `apply_fork_choice` won't reach `StateNotReachable` in practice until PR 4). But worth designing in before PR 4 flips the trigger.

ElFantasma (changes requested):

edg-l:
Pushed `638fe2184` addressing the three review P1s:

- **Reorg admission race** (`fork_choice.rs:300` / `blockchain.rs`): `enter_reorg` is now a `compare_exchange` test-and-set returning `Option<ReorgGuard>`. `reorg_apply_deep` short-circuits to `SYNCING` when the slot is already claimed, so two concurrent FCUs can no longer both install/replay against the shared overlay cache. Added a concurrent-acquire assertion to `reorg_guard_sets_and_clears_flag`.
- **Release-build history deletion** (`store.rs:4181`): the Section-9 `delete_range` is now gated on `committed_block_number == t` at runtime, not just `debug_assert`. On mismatch it logs an error and skips the delete rather than dropping new-chain `STATE_HISTORY` entries in release.
- **Invalid replay looks syncing** (`fork_choice.rs:523`): `map_chain_error_for_fcu` now maps `InvalidBlock`/`InvalidTransaction` to `InvalidAncestor(parent_hash)` so the engine responds INVALID with a `latestValidHash`; infra errors (`ParentNotFound`, DB/trie/RLP, bare `EvmError`) stay `StateNotReachable` so a genuinely-missing-state FCU is still retried instead of the branch being wrongly rejected.

edg-l:
Follow-up: also fixed the #6687 "Overlay Applies Across Roots" P1 here (commit `915a68dc4`), since you flagged the resolution belongs in this PR. The `Overlay` records the reconstructed pivot root (`serves_root`) and `TrieLayerCache::overlay_serves(root)` gates the read cascade to consuming roots only (pivot + new-chain layer roots); unrelated readers fall through to disk. Verified via `deep_reorg_state_oracle` (storage + balance correct after a 150-deep reorg) and two new unit tests.

ElFantasma (approved):
LGTM

ElFantasma (approved):

edg-l:
FKV + deep-reorg edge case, mitigation added here.

The FKV generator writes flat-KV entries that aren't journaled, and `commit_to_disk` skips journaling flat-KV leaves past the generator frontier. So a deep reorg can unwind a block whose flat-KV write was never recorded: the overlay misses the key, the read trusts disk flat-KV, and disk holds the value from the chain being reorged away. Stale read, no detection.

Mitigation in `reorg_apply_deep`: gate on `Store::flatkeyvalue_fully_generated()` (durable `[0xff]` sentinel). If generation is still running, return `SYNCING` so the CL retries once it finishes. Tested in `flatkeyvalue_completeness_tests`.

Not a full fix: entries written during generation stay incomplete, so a reorg into them before they're pruned is still exposed. Tracked in #7001.

ilitteri:
Audit of this PR (diff `feat/state-history-overlay...feat/deep-reorg-orchestration`, head `3b44df014`). The architecture is sound — Ping-rendezvous on install, overlay read-gating with regression tests, guard discipline, single-layer reconciliation design verified correct. But two high-severity defects sit in the core orchestration, and an important reachability note first:

**Reachability:** `reorg_apply_deep` runs when the link block's state is unreachable (link < disk edge `D`). After any shallow reorg that lowers `latest`, `D` ends up <128 behind the tip, so a second reorg with pivot < `D` and depth ≤ 128 enters the deep path **today, without waiting for #6724**. The findings below are live-reachable in reorg-after-reorg scenarios, not stack-latent.

## High — "head already canonical" case installs an overlay for the wrong state root; head's state never becomes readable

`crates/blockchain/fork_choice.rs:394-397`, `:446-448`.

For an FCU whose head is already canonical but whose state was flushed past the layer-cache edge, `new_canonical_blocks` is empty, so `pivot_number = head.number - 1`, `to_block = head.number`, and `serves_root` is captured from the journal entry at `to_block` — i.e. the state root of block `head.number - 1`, **head's parent, not head** (`Overlay::from_journal` sets `serves_root = entry.parent_state_root` at `n == to_block`; `CommittedLayer.parent_state_root` is the block's pre-state). Because `head_is_canonical`, the replay iterator is empty — head is never re-executed. After `reorg_apply_deep` returns `Ok(head)`, `has_state_root(head.state_root)` is still false; the FCU handler proceeds to `build_payload` on head, which fails, and every CL retry re-enters the same path and re-installs the same parent-serving overlay — an unrecoverable loop. The comment at `fork_choice.rs:97-104` names exactly this scenario as the reason for the new gate, but as implemented it cannot succeed. The correct pivot for this case is `head.number` itself (overlay range `[head.number+1, edge]`, `serves_root = head.state_root`), or head must be included in the replay.

## High — multi-layer reconciliation contradicts the asserted invariant: persist-worker panic in debug, wrong journal pre-images in release

`crates/storage/store.rs:4252-4256` (`debug_assert!(committed_layers.len() == 1)`), `:4263-4313`.

The assert's premise is that the overlay-backed commit is always a single layer at `T`. But during replay no commits happen at all (the safe-commit gate only fires for canonical roots; replayed blocks aren't canonical yet), so the entire side-chain backlog accumulates. The final `forkchoice_update` computes `safe_commit_root = head − commit_threshold` and, when that lands on a replayed layer (side-chain length ≥ 130 with threshold 128 — the common case once the deep path runs), commits `L − 128` layers at once with the overlay installed. Consequences:

- Debug builds: `debug_assert!` panics **inside the persist worker thread**, silently killing the node's only persistence path.
- Release: disk writes and `delete_range [T, D]` land correctly, **but journal pre-images are wrong for upper layers**: `extra_writes` excludes keys touched by *any* committed layer (`layer_keys` is the union over all layers, `store.rs:4265-4270`), so a key present in the reorg overlay and first touched by layer `T+2` is not bridged, and `T+2`'s journal entry records its pre-image from the old-chain `D` state instead of the correct pivot/bridge value (the intra-batch pre-image map is never seeded with bridge values, `store.rs:4340-4360`). A future deep reorg unwinding into such an entry reconstructs a wrong state — silent, consensus-critical.

Either enforce the single-layer invariant (commit only the bottom layer while an overlay is installed) or seed the intra-batch pre-image map from the overlay; the current code asserts one thing and does another.

## Medium — `abort_reorg` races the persist worker (no rendezvous)

`crates/storage/store.rs:3600-3610`. `install_overlay_for_reorg` rendezvous-pings the persist worker before swapping `trie_cache`; `abort_reorg` (called from `AbortReorgGuard::drop` on every replay failure) swaps with no such synchronization. The live-path ack fires *before* `apply_trie_phase1` installs the layer (`store.rs:1988-2003`), so the worker can still be mid-flight; if phase1's RCU read happens after the abort swap, it installs a side-chain layer into the fresh cache *without* the overlay underneath — reads at that root then cascade layer → old-chain disk state = wrong state. A concurrent `newPayload` parented at such a root could execute against garbage state and return `INVALID` for a valid chain. `abort_reorg` should Ping-rendezvous exactly like install does.

## Medium — the journal-pruning gate the design relies on was never wired up

`crates/blockchain/blockchain.rs:222-227` and the `Overlay::from_journal` SAFETY comment both state that journal pruning in `forkchoice_update_inner` defers while `reorg_in_progress` is set. Nothing in storage reads that flag — `forkchoice_update_inner` (`store.rs:1209-1290`) prunes `STATE_HISTORY` unconditionally when finality advances, and the syncer's direct `forkchoice_update` calls (e.g. `sync/full.rs:690`) aren't gated. `Overlay::from_journal` reads entries one-by-one with no snapshot isolation, so concurrent pruning aborts construction with `MissingEntry` → spurious `SYNCING`. Self-healing but wedges progress under contention, and two safety comments now document guarantees that don't exist.

## Medium — mutual exclusion covers FCU but not `newPayload`

`apply_fork_choice_with_deep_reorg` gates concurrent FCUs, but `try_execute_payload` executes blocks unconditionally. A payload executing concurrently with `install_overlay_for_reorg` can enqueue a `Block` message after the Ping is acked; the worker's phase1 then RCU-reads the pre-swap cache and writes back after the swap, clobbering the freshly installed overlay. Losing the overlay mid-replay makes replay reads fall through to old-chain disk state → wrong execution. Low probability, high impact.

## Low

- **Replay blocks a tokio worker** (`fork_choice.rs:457-468`): the synchronous, EVM-heavy `add_block` loop (up to ~128+ blocks) runs directly on the async FCU handler task — no `spawn_blocking`. A deep replay stalls that runtime worker for seconds to minutes.
- **No orchestration-level tests:** new tests cover the primitives well (`serves_root`, `overlay_serves`, install/abort/clear, FKV sentinel, `ReorgGuard`), but there is no end-to-end test of `reorg_apply_deep` (pivot computation, replay ordering, case-1 vs case-2) and none of `commit_to_disk` reconciliation. A multi-layer reconciliation test would have exposed the `debug_assert!` issue; a case-1 test would have exposed the wrong-`serves_root` issue.
- **Case-1 → case-2 retry skips journal hash verification:** after a case-1 attempt truncates `CANONICAL_BLOCK_HASHES` above head (`store.rs:1228-1236`), the case-2 retry builds `canonical_hashes` from the updated table, gets `None` for every height, and silently skips all `block_hash` verification (`fork_choice.rs:413-418`) — the integrity check is defeated exactly on the path that follows a case-1 attempt.

PR #2006, opened 2025-02-19
fix(l1): support multiple rpc requests in a single request

LeanSerra:
**Motivation**

The hive tests from `engine-withdrawals` were failing with a panic because we were asuming only a single RLPRequest could be received but the test sent an array of requets

**Description**

Introduce a wrapper struct that contains a `Single` type for single `RPCRequest` and `Multiple` for an array of requests, so serde can deserialize into the correct one. 
If we have an array of request we execute them sequentially. Then return a json encoded array of responses.
The function `rpc_response` was also changed to return `serde_json::Value` because it was returning an incorrect value when using it to create the array of responses. 

This PR fixes the following tests from `engine-withdrawals`:

- "Withdrawals Fork on Block 1 - 1 Block Re-Org"
- "Withdrawals Fork on Block 1 - 8 Block Re-Org NewPayload"
- "Withdrawals Fork on Block 8 - 10 Block Re-Org NewPayload"
- "Withdrawals Fork on Canonical Block 8 / Side Block 7 - 10 Block Re-Org"
- "Withdrawals Fork on Canonical Block 8 / Side Block 9 - 10 Block Re-Org"

Advances #1586

PR #6690, opened 2026-05-21
fix(l1): support JSON-RPC batches on the auth RPC port

edg-l:
## Summary

- Engine RPC port (`:8551`) rejected JSON-RPC 2.0 batches with `Invalid request body` because `handle_authrpc_request` deserialized into `RpcRequest` instead of `RpcRequestWrapper`. The public HTTP and WS handlers already supported batches; the auth path was missed when batching was added in #2006.
- Prysm's `execution_payload_envelopes_by_root` handler batches `eth_getBlockByHash` calls to the EL. On glamsterdam-devnet-4, this returned an unmarshal error, lodestar downscored and booted the prysm peers, and many ethrex/prysm slots got orphaned (forking party).
- JWT authentication still runs once per HTTP body. Empty batches return `-32000` `Invalid request body` per JSON-RPC 2.0 §6.

## Test plan

- [x] `cargo test -p ethrex-rpc --lib` (69/69 pass)
- [x] New `authrpc_accepts_batched_engine_requests` regression test: mints a valid JWT, sends a batch of two `engine_exchangeCapabilities` calls, asserts an array of 2 responses with `result` fields and matching ids
- [x] New `authrpc_rejects_empty_batch` test: empty array body returns `-32000` `Invalid request body`
- [x] `cargo fmt` / `cargo clippy -p ethrex-rpc --lib --tests` clean
- [ ] Validate on glamsterdam-devnet-4 that ethrex/prysm pairs no longer orphan slots

### DIFF
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -676,8 +676,8 @@ pub async fn handle_authrpc_request(
     auth_header: Option<TypedHeader<Authorization<Bearer>>>,
     body: String,
 ) -> Result<Json<Value>, StatusCode> {
-    let req: RpcRequest = match serde_json::from_str(&body) {
-        Ok(req) => req,
+    let wrapper: RpcRequestWrapper = match serde_json::from_str(&body) {
+        Ok(w) => w,
         Err(_) => {
             return Ok(Json(
                 rpc_response(
@@ -688,18 +688,41 @@ pub async fn handle_authrpc_request(
             ));
         }
     };
-    match authenticate(&service_context.node_data.jwt_secret, auth_header) {
-        Err(error) => Ok(Json(
-            rpc_response(req.id, Err(error)).map_err(|_| StatusCode::BAD_REQUEST)?,
-        )),
-        Ok(()) => {
-            // Proceed with the request
-            let res = map_authrpc_requests(&req, service_context).await;
-            Ok(Json(
-                rpc_response(req.id, res).map_err(|_| StatusCode::BAD_REQUEST)?,
-            ))
-        }
+
+    if let Err(error) = authenticate(&service_context.node_data.jwt_secret, auth_header) {
+        let id = match &wrapper {
+            RpcRequestWrapper::Single(req) => req.id.clone(),
+            RpcRequestWrapper::Multiple(_) => RpcRequestId::String("".to_string()),
+        };
+        return Ok(Json(
+            rpc_response(id, Err(error)).map_err(|_| StatusCode::BAD_REQUEST)?,
+        ));
     }
+
+    let res = match wrapper {
+        RpcRequestWrapper::Single(req) => {
+            let res = map_authrpc_requests(&req, service_context).await;
+            rpc_response(req.id, res).map_err(|_| StatusCode::BAD_REQUEST)?
+        }
+        RpcRequestWrapper::Multiple(requests) => {
+            if requests.is_empty() {
+                return Ok(Json(
+                    rpc_response(
+                        RpcRequestId::String("".to_string()),
+                        Err(RpcErr::BadParams("Invalid request body".to_string())),
+                    )
+                    .map_err(|_| StatusCode::BAD_REQUEST)?,
+                ));
+            }
+            let mut responses = Vec::with_capacity(requests.len());
+            for req in requests {
+                let res = map_authrpc_requests(&req, service_context.clone()).await;
+                responses.push(rpc_response(req.id, res).map_err(|_| StatusCode::BAD_REQUEST)?);
+            }
+            serde_json::to_value(responses).map_err(|_| StatusCode::BAD_REQUEST)?
+        }
+    };
+    Ok(Json(res))
 }
 
 /// Handle a WebSocket connection.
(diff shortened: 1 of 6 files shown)


edg-l:
Fixed all critical bot review comments

ilitteri (approved):

crates/networking/rpc/rpc.rs
```diff
@@ -676,30 +677,74 @@ pub async fn handle_authrpc_request(
...
-            ))
+            rpc_response(req.id, res).map_err(|_| StatusCode::BAD_REQUEST)?
         }
-    }
+        RpcRequestWrapper::Multiple(requests) if requests.is_empty() => null_id_error(
+            RpcErr::InvalidRequest("empty batch is not a valid Request".to_string()),
+        ),
+        RpcRequestWrapper::Multiple(requests) => {
+            let mut responses = Vec::with_capacity(requests.len());
+            for req in requests {
```
ElFantasma, line 779: **No batch size limit.** A client can send a 100k-request batch and this loop will sequentially process all of them, each `service_context.clone()` per iteration, all under the single JWT-auth check at the top. The auth port is now exposed to the same DoS shape that the public HTTP port has (where the existing handler at line 658-664 has the same pattern with no cap).

Geth caps engine-API batches at 1000 entries by default (`--engine.batchitemlimit`); reth has a similar `RpcServerConfig::set_batch_size_limit` knob. Worth adding a `MAX_AUTHRPC_BATCH_SIZE` constant and rejecting larger batches with `-32600 InvalidRequest`. The fix could be one match arm:
```rust
RpcRequestWrapper::Multiple(requests) if requests.len() > MAX_AUTHRPC_BATCH_SIZE => null_id_error(
    RpcErr::InvalidRequest(format!("batch too large: {} > {MAX_AUTHRPC_BATCH_SIZE}", requests.len())),
),
```

Not strictly blocking the prysm-orphaning fix, but worth pairing — the engine port is auth-gated but it's still attacker-controlled once JWT leaks or in lab setups.

ElFantasma, line 779: **Sequential batch processing.** The HTTP handler (line 660-663) has the same `for req in requests { ... .await; ... }` pattern, so the new auth-rpc handler is consistent — but in both, a batch of N requests blocks the connection for N × per-request latency. For Prysm's typical batched `eth_getBlockByHash` (small N, in-memory lookups), this is fine; for batches that hit storage (snap-sync state queries, multiget receipts), the linear blocking matters.

If the requests in the batch are independent (no ordering dependencies — which JSON-RPC 2.0 §6 explicitly says is fine), `futures::future::join_all` over `map_authrpc_requests` futures would parallelize. The `service_context.clone()` per call is already there; the rewrite is mechanical.

Non-blocking — the current shape matches the existing HTTP handler and the prysm fix is correct as-is. But worth a follow-up issue.

crates/networking/rpc/rpc.rs
```diff
@@ -676,30 +677,74 @@ pub async fn handle_authrpc_request(
...
+    let res = match wrapper {
+        RpcRequestWrapper::Single(req) => {
             let res = map_authrpc_requests(&req, service_context).await;
-            Ok(Json(
-                rpc_response(req.id, res).map_err(|_| StatusCode::BAD_REQUEST)?,
-            ))
+            rpc_response(req.id, res).map_err(|_| StatusCode::BAD_REQUEST)?
         }
-    }
+        RpcRequestWrapper::Multiple(requests) if requests.is_empty() => null_id_error(
```
ElFantasma: **Empty-batch handling is asymmetric with `handle_http_request`.** The auth handler correctly rejects empty `[]` with `InvalidRequest` per JSON-RPC 2.0 §6 ("a Server Response MUST be a single Response object" when the batch isn't a valid Array of values — and empty Array doesn't qualify). The HTTP handler at line 658-664 (same file) treats empty `Multiple([])` as a successful batch and returns... `[]`? Let me re-trace: `responses = Vec::new()`, loop does nothing, returns `serde_json::to_value([])` = `[]`. So the public HTTP port returns an empty array for an empty batch — non-compliant.

This is a pre-existing bug, not a regression of this PR — but since this PR is fixing JSON-RPC 2.0 compliance on the auth path, it'd be natural to either:
1. Mirror the empty-batch rejection on `handle_http_request` (same `if requests.is_empty() => null_id_error(...)` match arm), or
2. Note explicitly in the PR description that the HTTP-side empty-batch case is left for a follow-up.

Clients that explicitly probe `[]` get a different answer from the two endpoints today; that's confusing for a unified RPC surface.

crates/networking/rpc/rpc.rs
edg-l: Skipping the `join_all` rewrite for now ; the mechanical part is cheap but the engine API has implicit ordering that CL clients rely on:

- Prysm/Lighthouse pipeline `engine_newPayload(P)` then `engine_forkchoiceUpdated({head: P.hash})` in a single batch. If FCU executes before newPayload commits, the EL returns SYNCING/ACCEPTED or errors on an unknown hash. JSON-RPC 2.0 §6 permits parallel processing ("MAY process a batch as a set of concurrent tasks"), but the engine spec doesn't, and CL clients don't expect it.
- Back-to-back `newPayload` calls building a chain (child depends on parent) become non-deterministic under parallel execution.
- Geth and reth both serialize engine batches for these reasons.

Prysm's actual batch in this bug is `eth_getBlockByHash` (pure reads, safe to parallelize), but a generic `join_all` doesn't distinguish read from write methods, so it would unsafely apply to `engine_*` too. A safe version would need to split read-only vs state-mutating dispatch, which is well beyond the prysm-orphaning fix scope.

#1 (batch size cap, `MAX_BATCH_SIZE = 1000` matching geth's `--engine.batchitemlimit` default) and #3 (HTTP empty-batch parity) coming in the next push, applied to both `handle_http_request` and `handle_authrpc_request`.

crates/networking/rpc/rpc.rs
```diff
@@ -645,28 +646,78 @@ pub async fn shutdown_signal() {
...
+    serde_json::json!({
+        "jsonrpc": "2.0",
+        "id": Value::Null,
+        "error": meta,
+    })
+}
+
+/// Reject empty and oversize batches before they reach dispatch. Returns
+/// `Some(error_value)` to short-circuit, or `None` to proceed.
+fn reject_invalid_batch(wrapper: &RpcRequestWrapper) -> Option<Value> {
```
iovoid: Should be called validate_batch or similar.

iovoid (changes requested):

ElFantasma (approved):

crates/networking/rpc/rpc.rs
edg-l: Renamed to validate_batch in 4174e2b527.

azteca1998 (approved):

iovoid (approved):

PR #6691, opened 2026-05-21
fix(l1): retry GetPooledTransactions on alternate peer when blob sidecar mismatches

edg-l:
## Bug
When two peers announce the same blob tx hash, only the first becomes the in-flight requester; the second's announcement is dropped. If the first peer responds with bad data, ethrex disconnects it but has no fallback, and the upstream Hive test (`TestBlobTxWithMismatchedSidecar` / `TestBlobTxWithoutSidecar`) times out waiting for a re-request.

A subtler variant of the same bug: the two announcements can carry different `(type, size)` metadata. Peer A advertises a bare blob tx (small size), peer B advertises the full sidecar (large size). Even with a fallback in place, validating B's response against A's announced size would size-mismatch and disconnect the good peer.

## Fix
Track *alternate* announcers per hash in the mempool. When a hash is already in-flight from peer A and peer B announces it, store `Alternate { peer_id: B, tx_type, tx_size }` keyed on B's own announcement. On any failure path (bad blob, validation error, timeout, teardown, send error), pop the next live alternate and re-enqueue the request on that peer's connection, building the retry announcement from the alternate's own metadata so validation lines up.

Bounded by `MAX_ALTERNATES_PER_HASH = 8`, FIFO eviction, stale entries pruned every 6 min with a 10 min TTL.

## Test plan
- [x] 8 unit tests for the mempool alternates logic.
- [x] `cargo clippy -p ethrex-blockchain -p ethrex-p2p --all-targets -- -D warnings`
- [x] Hive devp2p: `TestBlobTxWithMismatchedSidecar` and `TestBlobTxWithoutSidecar` pass across 4 consecutive runs.
- [ ] Hive engine + sync + rpc-compat suites: confirm no regressions in CI.

### DIFF
--- a/crates/blockchain/mempool.rs
+++ b/crates/blockchain/mempool.rs
@@ -2,6 +2,7 @@ use std::{
     collections::{BTreeMap, VecDeque, hash_map::Entry},
     sync::RwLock,
     sync::atomic::{AtomicU64, Ordering},
+    time::{Duration, Instant},
 };
 
 use rustc_hash::{FxHashMap, FxHashSet};
@@ -25,6 +26,27 @@ use ethrex_storage::error::StoreError;
 use ethrex_vm::{intrinsic_gas_dimensions, intrinsic_gas_floor};
 use tracing::warn;
 
+/// Maximum number of alternate announcers tracked per hash. Bounds the memory
+/// used by the alternates map and prevents pathological peers from filling it.
+///
+/// TODO(#6691-fu): expose this through `BlockchainOptions` / CLI like the
+/// other mempool ceilings (`max_mempool_size`, RBF price-bumps). 8 is
+/// conservative; high-fan-in benchmarks and Hive adversarial-mempool scenarios
+/// might want to raise it. FIFO eviction keeps the cap safe regardless.
+pub const MAX_ALTERNATES_PER_HASH: usize = 8;
+
+/// An alternate announcer for a known-in-flight transaction hash. Carries the
+/// announcer's own announced type and size so the eventual retry can validate
+/// the response against the alternate's metadata (which may differ from the
+/// primary announcer's, e.g. when one peer advertises a bare blob tx while
+/// another advertises the full sidecar).
+#[derive(Debug, Clone, Copy)]
+pub struct Alternate {
+    pub peer_id: H256,
+    pub tx_type: u8,
+    pub tx_size: usize,
+}
+
 #[derive(Debug, Default)]
 struct MempoolInner {
     broadcast_pool: FxHashSet<H256>,
@@ -34,6 +56,15 @@ struct MempoolInner {
     /// but whose responses haven't arrived yet. Used to avoid sending duplicate
     /// requests when multiple peers announce the same transaction.
     in_flight_txs: FxHashSet<H256>,
+    /// For each announced hash, the queue of *alternate* announcers that also
+    /// advertised it while the hash was already in-flight from someone else.
+    /// Each entry carries the announcer's own announced type and size so the
+    /// retry can validate the response against the alternate's metadata (which
+    /// may differ from the primary's). Used as a fallback list when an in-flight
+    /// request fails or the responding peer disconnects. The `Instant` records
+    /// the last time the entry was touched so a periodic pruner can drop stale
+    /// entries.
+    alternates: FxHashMap<H256, (VecDeque<Alternate>, Instant)>,
     /// Maps blob versioned hashes to transaction hashes that include them and a position inside
     /// blob bundle where blob and its adjacent data is available.
     blobs_bundle_by_versioned_hash: FxHashMap<H256, FxHashMap<H256, usize>>,
@@ -177,6 +208,7 @@ impl Mempool {
             .insert((sender, transaction.nonce()), hash);
         inner.transaction_pool.insert(hash, transaction);
         inner.broadcast_pool.insert(hash);
+        inner.alternates.remove(&hash);
         // Drop the write lock before notifying to avoid holding it while waking waiters
         drop(inner);
         // Bump `tx_seq` *after* releasing the write lock. The payload builder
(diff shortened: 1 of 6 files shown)


crates/blockchain/mempool.rs
```diff
@@ -343,6 +421,45 @@ impl Mempool {
...
+        if entry.0.is_empty() {
+            inner.alternates.remove(&hash);
+        }
+        Ok(popped)
+    }
+
+    /// Drop the entire alternates list for the given hashes. Called when a
+    /// transaction has been successfully accepted into the pool or when we
+    /// otherwise no longer need to retry.
+    pub fn clear_alternates(&self, hashes: &[H256]) -> Result<(), StoreError> {
```
ElFantasma: **`clear_alternates` is `pub` but has zero production callers.** I grep'd `crates/` and `test/` for `clear_alternates` and the only hit outside this file is the unit test `clear_alternates_drops_entries`. Nothing calls it from the success path.

Consequence: on a healthy fetch (primary peer returns valid data, tx lands in the pool), every alternate announcement for that hash stays in the map until `prune_alternates` sweeps it via the 10-min TTL. For a busy node receiving the same blob tx announcement from many peers — exactly the workload this PR is targeting — that's potentially `MAX_ALTERNATES_PER_HASH * <fan-in>` entries lingering across the prune window.

The map self-bounds at 8 alternates × ~41 bytes per entry, so it's not a runaway leak (cap ≈ 3MB for 10k hashes). And `reserve_unknown_hashes` correctly avoids registering NEW alternates once the tx is in the pool (line 389: `inner.transaction_pool.contains_key(hash)`). But the existing alternates from before the pool insertion don't get dropped.

The natural wire-up is in the GetPooledTransactions response handler: after a successful add, call `clear_alternates(&accepted_hashes)`. That keeps the alternates map tight on the happy path; the TTL stays as a safety net for hashes that never resolve.

If the intent is "rely on TTL only, the success-path cleanup isn't worth the lock cost", make `clear_alternates` `pub(crate)` and add a comment noting that the TTL is the only reaper. Either way the current state — a public unused method — is a fence we don't know which side of.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1616,3 +1698,73 @@ async fn flush_pending_tx_requests(state: &mut Established) -> Result<(), PeerCo
...
+        loop {
+            let alt = match blockchain.mempool.pop_alternate(*hash) {
+                Ok(Some(a)) => a,
+                Ok(None) => break,
+                Err(e) => {
+                    warn!(error = %e, "pop_alternate failed");
+                    break;
+                }
+            };
+            match peer_table.get_peer_connection(alt.peer_id).await {
```
ElFantasma, line 1744: **Double `peer_table.get_peer_connection` lookup with a race window.** The popping loop here calls `get_peer_connection(alt.peer_id)` to verify the peer is alive (and skips dead ones), then ~30 lines down (line 1764) `get_peer_connection(alt_peer_id)` runs again to actually enqueue. Between the two lookups, the connection can close — the second lookup falls into `Ok(None)` and the `enqueue_tx_requests` is silently skipped. The alternate slot is already popped, so the hash retry just fizzles.

Not a correctness bug — the hash is still in flight nowhere, and the next failure path / teardown would pop the next alternate. But it burns a fallback slot for nothing.

Cleaner: pop+lookup in one shot and stash the `PeerConnection` handle in the `by_peer` map alongside the metadata. Then the enqueue site just iterates the handles. Re-grouping by `alt_peer_id` after the fact still needs a HashMap, but the value can be `(PeerConnection, Vec<(H256, u8, usize)>)` instead of re-looking-up. One round-trip per peer, no race.

Low priority; current code degrades gracefully.

crates/blockchain/mempool.rs
```diff
@@ -24,6 +25,22 @@ use ethrex_storage::error::StoreError;
 use ethrex_vm::{intrinsic_gas_dimensions, intrinsic_gas_floor};
 use tracing::warn;
 
+/// Maximum number of alternate announcers tracked per hash. Bounds the memory
+/// used by the alternates map and prevents pathological peers from filling it.
+pub const MAX_ALTERNATES_PER_HASH: usize = 8;
```
ElFantasma, line 36: nit: `MAX_ALTERNATES_PER_HASH = 8` is a hardcoded `pub const` (not configurable via `BlockchainOptions` or CLI). For most realistic deployments 8 is plenty — but it's the kind of value that's nice to bump from config when running Hive's adversarial-mempool scenarios or a high-fan-in benchmark.

Not urgent — the cap is conservative and the FIFO eviction is sound. Flagging only because the rest of the mempool's bounds are configurable (`max_mempool_size`, `mempool_max_size`, etc.) and this one stands out as the only hardcoded ceiling.

test/tests/blockchain/mempool_tests.rs
```diff
@@ -548,3 +548,154 @@ fn blobs_bundle_insert_and_remove() {
...
+    }
+
+    #[test]
+    fn prune_alternates_drops_stale_entries() {
+        let mp = Mempool::new(64);
+        let tx = h(0xaa);
+        reserve(&mp, &[tx], h(1));
+        reserve(&mp, &[tx], h(2));
+        // Sleep a tick so prune_alternates sees these entries as stale.
+        std::thread::sleep(Duration::from_millis(2));
```
ElFantasma: nit: `std::thread::sleep(Duration::from_millis(2))` + `prune_alternates(Duration::from_millis(1))` is a very tight margin (1ms TTL, 2ms sleep). On a busy CI runner where the scheduler hands a 2ms sleep back as 0.7ms, the test would flake by reporting entries-not-stale and the prune leaves them. Same shape as several CI flake reports we've eaten on other tests.

Bump to something like `sleep(20ms) + ttl(5ms)`; the test still finishes in <50ms but tolerates ±10ms of CI jitter.

crates/blockchain/mempool.rs
ilitteri: +1

ilitteri (approved):

crates/blockchain/mempool.rs
edg-l: Dropped clear_alternates + its test in 22089af4e6; the inline remove in add_transaction is the real path and covers every add site.

test/tests/blockchain/mempool_tests.rs
edg-l: Already in: sleep(20ms) + ttl(5ms) on the current head (mempool_tests.rs:691-701).

crates/networking/p2p/rlpx/connection/server.rs
edg-l: Deferred as TODO(#6691-fu) in 27598c9b60 next to the lookup site.

crates/blockchain/mempool.rs
edg-l: Deferred as TODO(#6691-fu) at mempool.rs:31.

crates/networking/p2p/rlpx/connection/server.rs
edg-l: Actually fixed in a3dd7bf190; stash the PeerConnection at probe time and reuse it instead of looking up twice. Drops the deferred TODO too.

azteca1998 (approved):

ElFantasma (approved):

PR #6693, opened 2026-05-21
feat(l1): implement debug_getBadBlocks RPC endpoint

### LINKED ISSUE #6640: Implement debug_getBadBlocks
Implement `debug_getBadBlocks`. This is the only [OpenRPC-specified](https://github.com/ethereum/execution-apis/blob/main/src/debug/getters.yaml) `debug_*` method ethrex is missing. Returns a list of recently observed bad blocks. No hive fixture exists, so coverage is unit/integration tests.

Requires:
- A bounded `BadBlocks` cache populated on block validation failure (see [geth `eth/api_debug.go`](https://github.com/ethereum/go-ethereum/blob/master/eth/api_debug.go) for the `BadBlockArgs { hash, block, rlp }` shape).
- A hook in the validation path to record failures into that cache.
- A handler entry in `map_debug_requests` ([rpc.rs:1035](https://github.com/lambdaclass/ethrex/blob/main/crates/networking/rpc/rpc.rs#L1035)).

Parent: #6572

azteca1998:
## Summary
Adds the `debug_getBadBlocks` RPC endpoint as specified in the [execution-apis OpenRPC spec](https://github.com/ethereum/execution-apis/blob/main/src/debug/getters.yaml).

**This is a stub that always returns `[]`.** A full implementation is not currently possible because ethrex lacks the required storage infrastructure:

- Geth's response requires the full block RLP, header, body, and a human-readable error reason per bad block entry.
- ethrex only stores `(bad_block_hash → latest_valid_hash)` in the `INVALID_CHAINS` table (used by the fork-choice handler). It does **not** persist the full block data or the validation error when a block is rejected.
- Implementing this properly would require:
  1. A new storage table / ring buffer for bad block metadata (full block + error string).
  2. Writes at every block rejection site (engine fork-choice, full sync, etc.).
  3. An eviction policy (geth keeps the last N entries).

Returning `[]` is spec-valid — it's the correct response for a node that has not encountered any bad blocks. The endpoint is registered so that tooling that probes for it (e.g. Blockscout, hive tests) gets a well-formed response instead of a `method not found` error.

Part of #6572

## Test plan
- [x] `debug_getBadBlocks` returns `[]`
- [x] Calling with unexpected params returns error
- [x] e2e integration test (`trace_get_bad_blocks`)

### DIFF
--- a/crates/networking/rpc/debug/get_bad_blocks.rs
+++ b/crates/networking/rpc/debug/get_bad_blocks.rs
@@ -24,3 +24,37 @@ impl RpcHandler for GetBadBlocksRequest {
         Ok(Value::Array(vec![]))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::RpcHandler;
+    use serde_json::json;
+
+    #[test]
+    fn parse_no_params() {
+        let result = GetBadBlocksRequest::parse(&None);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn parse_empty_params() {
+        let result = GetBadBlocksRequest::parse(&Some(vec![]));
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn parse_with_params_fails() {
+        let result = GetBadBlocksRequest::parse(&Some(vec![json!("extra")]));
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
+    async fn handle_returns_empty_array() {
+        let req = GetBadBlocksRequest;
+        let storage = crate::test_utils::setup_store().await;
+        let context = crate::test_utils::default_context_with_storage(storage).await;
+        let result = req.handle(context).await.unwrap();
+        assert_eq!(result, Value::Array(vec![]));
+    }
+}
--- a/crates/networking/rpc/debug/get_bad_blocks.rs
+++ b/crates/networking/rpc/debug/get_bad_blocks.rs
@@ -1,0 +1,26 @@
+use serde_json::Value;
+
+use crate::{RpcApiContext, RpcErr, RpcHandler};
+
+pub struct GetBadBlocksRequest;
+
+impl RpcHandler for GetBadBlocksRequest {
+    fn parse(params: &Option<Vec<Value>>) -> Result<Self, RpcErr> {
+        if let Some(params) = params
+            && !params.is_empty()
+        {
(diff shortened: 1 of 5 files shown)


crates/networking/rpc/debug/get_bad_blocks.rs
```diff
@@ -0,0 +1,60 @@
...
+                "Expected no params and {} were provided",
+                params.len()
+            )));
+        }
+        Ok(GetBadBlocksRequest)
+    }
+
+    async fn handle(&self, _context: RpcApiContext) -> Result<Value, RpcErr> {
+        // Geth maintains a ring buffer of bad blocks encountered during chain
+        // insertion. ethrex does not currently track bad blocks, so we return
```
ElFantasma: nit: "ethrex does not currently track bad blocks" understates what's there — `INVALID_CHAINS` already persists `bad_block_hash → latest_valid_hash` (`Store::set_latest_valid_ancestor`, `store.rs:1125`, consumed for the engine API's `latestValidHash`). The reason this can't be a real implementation isn't a lack of *tracking*, it's that the geth response schema needs the full block RLP + body + a human-readable error reason per entry, none of which is persisted.

Suggest matching the (accurate) explanation from the PR description:
```rust
// ethrex records rejected-block hashes in INVALID_CHAINS (hash -> latest_valid_hash)
// but does not persist the full block RLP / body / error reason that geth's
// debug_getBadBlocks response requires, so we return the spec-valid empty array.
// A real implementation needs a bad-block metadata table + eviction policy.
```
Keeps a future implementer from assuming they're starting from scratch.

ElFantasma (approved):
**Always-`[]` can mislead a debugging operator on a node that *has* rejected blocks.** This is the cost of the stub: someone calling `debug_getBadBlocks` precisely because they suspect bad blocks will be told "none" even when `INVALID_CHAINS` is non-empty. That's acceptable for now given the schema gap, but it argues for a *dedicated* follow-up issue rather than just "Part of #6572" — the real work (bad-block metadata table + writes at each rejection site + ring-buffer eviction) is concrete and shouldn't get lost once this stub merges and the box looks ticked. A linked sub-issue would keep it visible.

edg-l:
Superseded by #6948, which implements real bad-block tracking (persists rejected blocks from engine_newPayload and returns full block JSON + RLP) instead of the empty-array placeholder here.

PR #6694, opened 2026-05-21
feat(l1): implement noopTracer for debug trace endpoints

### LINKED ISSUE #6645: Implement noopTracer for debug_trace*

azteca1998:
## Summary
- Adds `noopTracer` variant to the tracer type enum
- Returns `{}` for `debug_traceTransaction`, and `{txHash, result: {}}` per tx for `debug_traceBlockByNumber`
- Useful for benchmarking raw execution overhead without tracing cost
- Matches [geth's noopTracer](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/native/noop.go)

Closes part of #6572

## Test plan
- [ ] `debug_traceTransaction` with `{"tracer": "noopTracer"}` returns `{}`
- [ ] `debug_traceBlockByNumber` with `{"tracer": "noopTracer"}` returns array of `{txHash, result: {}}`

Closes #6645

### DIFF
--- a/crates/blockchain/tracing.rs
+++ b/crates/blockchain/tracing.rs
@@ -218,6 +218,58 @@ impl Blockchain {
         Ok(traces)
     }
 
+    /// Re-executes the target transaction (and any preceding txs in its block) with no
+    /// tracer attached, then discards the result. Used by `noopTracer` so the RPC layer
+    /// validates that the tx exists and replays it through the EVM without paying any
+    /// tracing collection cost, matching the benchmarking purpose of geth's `noopTracer`.
+    /// May need to re-execute blocks in order to rebuild the transaction's prestate, up
+    /// to the amount given by `reexec`.
+    pub async fn noop_trace_transaction(
+        &self,
+        tx_hash: H256,
+        reexec: u32,
+        timeout: Duration,
+    ) -> Result<(), ChainError> {
+        let Some((_, block_hash, tx_index)) =
+            self.storage.get_transaction_location(tx_hash).await?
+        else {
+            return Err(ChainError::Custom("Transaction not Found".to_string()));
+        };
+        let tx_index = tx_index as usize;
+        let Some(block) = self.storage.get_block_by_hash(block_hash).await? else {
+            return Err(ChainError::Custom("Block not Found".to_string()));
+        };
+        let mut vm = self
+            .rebuild_parent_state(block.header.parent_hash, reexec)
+            .await?;
+        // Run every tx up to AND including the target — no tracer attached.
+        let stop = tx_index.saturating_add(1);
+        timeout_trace_operation(timeout, move || vm.rerun_block(&block, Some(stop))).await
+    }
+
+    /// Re-executes every transaction in the block with no tracer attached and returns the
+    /// transaction hashes in order. Used by `noopTracer` for `debug_traceBlockByNumber` so
+    /// the RPC layer can emit `{txHash, result: {}}` per tx without paying tracing cost.
+    /// May need to re-execute blocks in order to rebuild the block's prestate, up to the
+    /// amount given by `reexec`.
+    pub async fn noop_trace_block(
+        &self,
+        block: Block,
+        reexec: u32,
+        timeout: Duration,
+    ) -> Result<Vec<H256>, ChainError> {
+        let mut vm = self
+            .rebuild_parent_state(block.header.parent_hash, reexec)
+            .await?;
+        let block = Arc::new(block);
+        let block_for_exec = block.clone();
+        timeout_trace_operation(timeout, move || {
+            vm.rerun_block(block_for_exec.as_ref(), None)
+        })
+        .await?;
+        Ok(block.body.transactions.iter().map(|tx| tx.hash()).collect())
+    }
+
     /// Rebuild the parent state for a block given its parent hash, returning an `Evm` instance with all changes cached
     /// Will re-execute all ancestor block's which's state is not stored up to a maximum given by `reexec`
     async fn rebuild_parent_state(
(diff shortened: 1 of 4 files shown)


crates/networking/rpc/tracing.rs
```diff
@@ -59,6 +59,11 @@ enum TracerType {
     /// `structLogger` wrapper shape (`{failed, gas, returnValue, structLogs}`).
     /// Selected via `"tracer": "opcodeTracer"`.
     OpcodeTracer,
+    /// No-op tracer that re-executes the transaction(s) through the EVM with no tracer
```
ElFantasma: nit: "matching geth's `noopTracer`" is true for the *output shape* but not the *mechanism*. geth's noopTracer attaches a tracer with empty hooks, so the EVM still dispatches into each (no-op) hook per opcode; this implementation runs with no tracer attached at all (`rerun_block` with no tracer), skipping hook dispatch entirely.

For the benchmarking goal this is arguably *better* (it isolates pure execution), but the doc could be precise so future readers don't assume the numbers are directly comparable to geth's noopTracer:

```rust
/// No-op tracer: re-executes the transaction(s) through the EVM with no tracer
/// attached and returns empty JSON objects. The output shape matches geth's
/// `noopTracer` (`{}` per tx), but note the mechanism differs — geth runs an
/// empty-hook tracer (paying per-opcode dispatch), whereas this skips tracer
/// hooks entirely, isolating raw execution cost. Any `tracerConfig` is ignored.
/// Selected via `"tracer": "noopTracer"`.
```
Non-blocking — just so a future benchmark comparison isn't misread.

crates/blockchain/tracing.rs
```diff
@@ -218,6 +218,58 @@ impl Blockchain {
...
+        let tx_index = tx_index as usize;
+        let Some(block) = self.storage.get_block_by_hash(block_hash).await? else {
+            return Err(ChainError::Custom("Block not Found".to_string()));
+        };
+        let mut vm = self
+            .rebuild_parent_state(block.header.parent_hash, reexec)
+            .await?;
+        // Run every tx up to AND including the target — no tracer attached.
+        let stop = tx_index.saturating_add(1);
+        timeout_trace_operation(timeout, move || vm.rerun_block(&block, Some(stop))).await
```
ElFantasma, line 247: Optional consistency port from the duplicate #6657. Here the target tx is folded into the replay by stopping at `tx_index + 1`. The sibling tracers (`trace_tx_prestate`, `trace_tx_opcodes`) instead replay `[0, tx_index)` and then run the target tx through a dedicated step — and #6657's noop path follows that same shape (`rerun_block(Some(tx_index))` + a separate `trace_tx_noop(target)`).

For a noop the result is identical either way, so this is purely about keeping the module uniform:
```rust
// replay preceding txs, then run the target through the same dedicated
// step the other tracers use — keeps the tracing module's shape uniform.
vm.rerun_block(&block, Some(tx_index))?;
timeout_trace_operation(timeout, move || vm.trace_tx_noop(&block, tx_index)).await
```
Non-blocking; adopt only if you want parity with the prestate/opcode paths. The current `Some(tx_index + 1)` form is correct as-is.

ElFantasma (commented):
**Mechanism differs from geth's `noopTracer`, which matters for the stated benchmarking goal.** geth's `noopTracer` is a real tracer with empty hook implementations — the EVM runs *with* the tracer attached, so every opcode still dispatches into the (empty) hook. This PR instead runs `rerun_block` with **no tracer at all**, skipping hook dispatch entirely. The JSON *output* matches geth (`{}` per tx), but the *measurement* doesn't: geth's noopTracer number includes per-opcode hook-dispatch overhead, while ethrex's measures pure execution. For "benchmarking raw execution overhead without tracing cost" the no-hook approach is arguably the cleaner number — but anyone comparing ethrex-noop vs geth-noop side by side would be comparing different things. Worth a one-line doc clarification (inline).

---

**Duplicate of #6657 — and one structural idea worth porting from it before this lands.** There's an existing open PR for the exact same feature: [#6657](https://github.com/lambdaclass/ethrex/pull/6657) (`noopTracer`, opened 2026-05-14 by @ElFantasma) implements the same `TracerType::NoopTracer` wired through the same two RPC methods. This PR is the stronger of the two — it ships real tests where #6657 ships none, and the whole-block `rerun_block(None)` path here is both simpler and a cleaner benchmark (no per-tx `setup_env`+`VM::new` reconstruction polluting the measurement). So the recommendation is to land this one and close #6657.

The one thing #6657 does that's worth salvaging: its **single-tx path mirrors the sibling tracers** (`trace_tx_prestate` / `trace_tx_opcodes`) by replaying `[0, N-1]` and then executing the target tx through a dedicated `trace_tx_noop` step, instead of folding the target into `rerun_block(Some(N+1))`. For a noop the observable result is identical, but matching the established replay-then-trace-target shape keeps the tracing module uniform and makes the next tracer that *does* collect per-tx data drop in without a special case. See the inline suggestion on `noop_trace_transaction`. (Non-blocking — adopt if you value the consistency; the current form is correct.)

**Issue linkage:** the description says *"Closes part of #6572"*. #6572 is the umbrella debug-RPC tracker; the granular sub-issue for this exact feature is **[#6645](https://github.com/lambdaclass/ethrex/issues/6645)** ("Implement noopTracer for debug_trace*"). Please change the body to `Closes #6645` so GitHub auto-closes the specific sub-issue on merge — a bare "Closes part of #6572" closes nothing and leaves the umbrella issue to be reconciled by hand later.

**Coordination:** @azteca1998 — since #6657 is the duplicate and this PR supersedes it, suggest closing #6657 once this is finalized (Esteve to action). Flagging so two reviewers don't sink time into both.

azteca1998:
Closing in favour of #6657 which implements the same noopTracer (sub-issue #6645).

PR #6695, opened 2026-05-21
feat(l1): implement debug_traceCall RPC endpoint

### LINKED ISSUE #6641: Implement debug_traceCall
Implement `debug_traceCall`. Same shape as `debug_traceTransaction` but takes a tx call object + block + optional state/block overrides — runs the tx against the post-state of `block` rather than replaying a previously included tx.

References:
- [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-tracecall)
- [geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go)

Implementation notes:
- Reuse tracer dispatch from `trace_transaction_*` in `crates/blockchain/tracing.rs`.
- Apply `stateOverrides` / `blockOverrides` (same shape as `eth_call`) before running the trace.
- Register handler in `map_debug_requests`.

Parent: #6572

azteca1998:
## Summary
- Adds `debug_traceCall` which traces an arbitrary call at a given block state without creating a transaction on-chain
- Supports all existing tracers: `callTracer`, `prestateTracer`, `opcodeTracer`
- Adds trace-call infrastructure at each layer (LEVM → Evm → Blockchain → RPC)
- Accepts the same call object format as `eth_call` plus an optional trace config
- Matches [geth's debug_traceCall](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-tracecall)

Closes part of #6572

## Test plan
- [ ] `debug_traceCall` with callTracer returns nested call frames
- [ ] `debug_traceCall` with prestateTracer returns pre/post state
- [ ] `debug_traceCall` with opcodeTracer returns EIP-3155 step trace
- [ ] Block identifier parameter works (latest, specific number, hash)
- [ ] Default block (latest) works when block param omitted

Closes #6641

### DIFF
--- a/crates/networking/rpc/tracing.rs
+++ b/crates/networking/rpc/tracing.rs
@@ -9,7 +9,13 @@ use ethrex_vm::tracing::OpcodeTracerConfig;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
-use crate::{rpc::RpcHandler, types::block_identifier::BlockIdentifier, utils::RpcErr};
+use ethrex_common::types::GenericTransaction;
+
+use crate::{
+    rpc::RpcHandler,
+    types::block_identifier::{BlockIdentifier, BlockIdentifierOrHash},
+    utils::RpcErr,
+};
 
 /// Default max amount of blocks to re-excute if it is not given
 const DEFAULT_REEXEC: u32 = 128;
@@ -23,6 +29,12 @@ pub struct TraceTransactionRequest {
 
 pub struct TraceBlockByNumberRequest {
     block: BlockIdentifier,
+    trace_config: TraceConfig,
+}
+
+pub struct TraceCallRequest {
+    transaction: GenericTransaction,
+    block: Option<BlockIdentifierOrHash>,
     trace_config: TraceConfig,
 }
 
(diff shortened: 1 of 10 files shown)


crates/networking/rpc/tracing.rs
```diff
@@ -26,6 +32,34 @@ pub struct TraceBlockByNumberRequest {
...
+/// `debug_traceCall` — trace an arbitrary call against the state at a given
+/// block, matching geth's [debug_traceCall][geth]. The call is executed in
+/// memory only; nothing is persisted.
+///
+/// Params: `[callObject, blockIdOrHash?, traceConfig?]`. Omitting the block
+/// defaults to `"latest"`. The call object follows the `eth_call` shape; the
+/// trace config follows the same shape used by `debug_traceTransaction`.
+///
+/// **Known divergence from geth**: geth's `traceConfig` accepts
+/// `stateOverrides` and `blockOverrides` for hypothetical-state debugging.
```
ElFantasma, line 45: The doc honestly flags that `stateOverrides` / `blockOverrides` / `reexec` are silently ignored, which is the friendly choice for callers who don't use them but unfriendly for callers who *do* and don't notice their override never ran. Geth-equivalent clients (which is what most callers will assume on the `debug_traceCall` JSON-RPC surface) honour both override fields, so a caller debugging a hypothetical state will see their override silently dropped and the trace will reflect the real state.

Two options:
1. **Reject unknown/unsupported fields**: on `parse`, peek at the raw `Value` and return `RpcErr::BadParams("stateOverrides not supported — see issue #XYZ")` if present. Loud, but matches operator expectations.
2. **Warn at log level**: `warn!("debug_traceCall: stateOverrides field present but not honoured")` so it surfaces in node logs without breaking callers.

Option 1 is more defensible for an early-stage endpoint. Either works; silently-ignored is the only shape I'd avoid. Non-blocking, but worth deciding before this lands in operator hands.

crates/vm/backends/levm/tracing.rs
```diff
@@ -158,6 +159,92 @@ impl LEVM {
...
+        db: &mut GeneralizedDatabase,
+        block_header: &BlockHeader,
+        tx: &GenericTransaction,
+        only_top_call: bool,
+        with_log: bool,
+        vm_type: VMType,
+        crypto: &dyn Crypto,
+    ) -> Result<CallTrace, EvmError> {
+        let mut env = env_from_generic(tx, block_header, db, vm_type)?;
+        env.block_gas_limit = i64::MAX as u64;
```
ElFantasma, line 176: `env.block_gas_limit = i64::MAX as u64` reads as magic without context. The reason — `debug_traceCall` simulates a tx outside any block's gas budget, matching geth's `eth_call`-style relaxation — is non-obvious to anyone reviewing the trace stack later.

One-line comment above (and at the two sibling sites at lines 202 / 240) would land it:
```rust
// debug_traceCall simulates a tx outside the parent block's gas budget — same
// relaxation as eth_call. Geth uses `evm.GasPool = MaxUint64` for the same reason.
env.block_gas_limit = i64::MAX as u64;
```
Non-blocking. Same comment applies symmetrically to the prestate (line 202) and opcode (line 240) variants.

crates/vm/backends/levm/mod.rs
```diff
@@ -2812,6 +2812,53 @@ fn vm_from_generic<'a>(
     VM::new(env, db, &tx, LevmCallTracer::disabled(), vm_type, crypto)
 }
 
+/// Like `vm_from_generic` but accepts a custom `LevmCallTracer`.
+pub(crate) fn vm_from_generic_with_tracer<'a>(
```
ElFantasma, line 2823: The new `vm_from_generic_with_tracer` duplicates the `GenericTransaction → Transaction` conversion logic from `vm_from_generic` (the EIP7702-vs-EIP1559 branching + access_list mapping + authorization_list mapping). Two near-identical blocks 30 lines apart.

Extracting a `fn generic_to_transaction(tx: &GenericTransaction) -> Result<Transaction, VMError>` would let both callers share the conversion; the per-call sites then just thread their tracer:
```rust
fn vm_from_generic_with_tracer<'a>(...) -> Result<VM<'a>, VMError> {
    let tx = generic_to_transaction(tx)?;
    let vm_type = adjust_disabled_l2_fees(&env, vm_type);
    VM::new(env, db, &tx, tracer, vm_type, crypto)
}
```
Low-effort cleanup, prevents future drift between the two paths. Non-blocking.

crates/networking/rpc/tracing.rs
```diff
@@ -358,3 +392,215 @@ impl RpcHandler for TraceBlockByNumberRequest {
...
+                        &self.transaction,
+                        timeout,
+                        config.only_top_call,
+                        config.with_log,
+                    )
+                    .await
+                    .map_err(|err| RpcErr::Internal(err.to_string()))?;
+                let top_frame = call_trace
+                    .into_iter()
+                    .next()
```
ElFantasma, line 471: `call_trace.into_iter().next().ok_or_else(...)` silently picks the first frame and returns `Internal("Empty call trace")` otherwise. For `trace_call_*` the underlying VM always returns exactly one top-frame (single call has one trace root), so the `.next()` is always sufficient — but the implicit "if there's more than one, the rest is dropped" behaviour is opaque to a reader.

If the invariant is "exactly one frame, always", a `debug_assert!(call_trace.len() == 1)` before the `.next()` makes the contract explicit and catches a future LEVM change that adds parallel sub-traces. If "first frame is the right one even if there are more", a one-line comment saying so. Non-blocking nit.

edg-l:
closing in favour of https://github.com/lambdaclass/ethrex/pull/6931

PR #6698, opened 2026-05-21
feat(l1): implement debug_accountRange RPC endpoint

### LINKED ISSUE #6649: Implement debug_accountRange
Implement `debug_accountRange`. Returns a paginated slice of the state trie at a given block.

References:
- [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-accountrange)
- [geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/api_debug.go)

Implementation notes:
- Trie iteration over the state trie from a given hashed key, bounded by `maxResults`.
- Per-account: balance, nonce, code hash, storage root; optionally include code + storage when flags request it.
- Shares iteration primitives with `debug_storageRangeAt` — coordinate to avoid duplicated traversal helpers.

Parent: #6572

azteca1998:
## Summary
- Adds `debug_accountRange` which iterates accounts in the state trie from a given starting hash
- Returns up to `maxResults` entries with a `next` hash for pagination
- Each entry includes balance, nonce, storage root, and code hash
- Uses block's final state root; txIndex-based reconstruction is a future enhancement
- Matches [geth's debug_accountRange](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-accountrange)

Closes part of #6572

## Test plan
- [ ] Returns account entries starting from a given hash
- [ ] Pagination via `next` field works correctly
- [ ] Returns all-zero `next` when no more accounts
- [ ] Invalid block hash returns error

Closes #6649

### DIFF
--- a/crates/networking/rpc/debug/account_range.rs
+++ b/crates/networking/rpc/debug/account_range.rs
@@ -80,6 +80,11 @@ impl RpcHandler for AccountRangeRequest {
             .map_err(|e| RpcErr::BadParams(format!("invalid start hash: {e}")))?;
         let max_results: usize = serde_json::from_value(params[3].clone())
             .map_err(|e| RpcErr::BadParams(format!("invalid maxResults: {e}")))?;
+        if max_results == 0 {
+            return Err(RpcErr::BadParams(
+                "maxResults must be greater than 0".to_owned(),
+            ));
+        }
         Ok(AccountRangeRequest {
             block,
             tx_index,
@@ -93,7 +98,7 @@ impl RpcHandler for AccountRangeRequest {
             .block
             .resolve_block_header(&context.storage)
             .await?
-            .ok_or_else(|| RpcErr::Internal("Block not found".to_string()))?;
+            .ok_or_else(|| RpcErr::WrongParam("Block not found".to_string()))?;
 
         // `tx_index` is parsed and validated but not honoured yet. Negative
         // values (geth's `-1` "end of block" sentinel) and any value >= the
--- a/crates/networking/rpc/debug/account_range.rs
+++ b/crates/networking/rpc/debug/account_range.rs
@@ -1,0 +1,93 @@
+use std::collections::BTreeMap;
+
+use ethrex_common::{H256, U256};
+use serde::Serialize;
+use serde_json::Value;
+
+use crate::{RpcApiContext, RpcErr, RpcHandler};
+
+pub struct AccountRangeRequest {
+    block_hash: H256,
+    tx_index: usize,
+    start: H256,
+    max_results: usize,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct AccountRangeResult {
+    accounts: BTreeMap<H256, AccountEntry>,
+    #[serde(rename = "next")]
+    next: H256,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct AccountEntry {
+    balance: U256,
+    nonce: u64,
+    root: H256,
+    code_hash: H256,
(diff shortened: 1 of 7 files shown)


crates/networking/rpc/debug/account_range.rs
```diff
@@ -0,0 +1,225 @@
...
+            .resolve_block_header(&context.storage)
+            .await?
+            .ok_or_else(|| RpcErr::WrongParam("Block not found".to_string()))?;
+
+        // `tx_index` is parsed and validated but not honoured yet. Negative
+        // values (geth's `-1` "end of block" sentinel) and any value >= the
+        // block's tx count are equivalent to "end of block" — which is what
+        // we always return, so no error needed. Other values would require
+        // mid-block state reconstruction; documented on the type.
+        let _ = self.tx_index;
```
ElFantasma: `let _ = self.tx_index;` silently discards a parsed parameter. The type-level doc is honest about it ("`txIndex` is currently ignored — the response is always the state at the *end* of the given block"), but a caller passing `txIndex = 0` expecting state *before* the first transaction of the block will get state *after* the last transaction instead — silently. That's the exact same friendliness concern I left on `debug_traceCall` (#6695) for `stateOverrides` / `blockOverrides`.

Negative values (geth's `-1` end-of-block sentinel) and any value `>= block.tx_count` are already equivalent to the current behaviour. The values that *aren't* equivalent are `0 <= txIndex < tx_count` — those would need mid-block state reconstruction. Three reasonable shapes:
1. **Silent and documented** (current): keep, but worth tightening the doc to spell out *which* txIndex values are silently coerced.
2. **Reject when txIndex is in the un-supported range**: `if 0 <= self.tx_index && self.tx_index < (block.tx_count as i64) { return Err(RpcErr::Unsupported(...)) }`. Loud and operator-friendly.
3. **Warn at log level** when the un-supported range is hit, keep the response shape.

I'd lean (2) for an early-stage endpoint where users haven't formed habits yet. Non-blocking — your call.

crates/networking/rpc/debug/account_range.rs
```diff
@@ -0,0 +1,225 @@
...
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct AccountEntry {
+    balance: U256,
+    nonce: u64,
+    root: H256,
+    code_hash: H256,
+    /// Original (unhashed) address. Always `null` — ethrex has no preimage store.
+    key: Option<H256>,
```
ElFantasma, line 65: Every emitted `AccountEntry` will serialize as `"key": null` because ethrex has no preimage store. That's documented and correct — but every entry having `"key": null` is noise on the wire for callers who know to expect it.

`#[serde(skip_serializing_if = "Option::is_none")]` on this field would omit the `key` entirely when unknown, matching the geth-when-preimage-missing shape (geth skips emitting the field). If a future preimage store lands, `Some(addr)` would naturally start appearing.

Non-blocking; current shape is technically conformant. Just leaner on the wire and friendlier to grep through.

test/tests/rpc/helpers.rs
```diff
@@ -0,0 +1,191 @@
+//! Shared setup helpers for the rpc integration tests. Each test file builds
```
ElFantasma, line 1: Heads up: this exact same file (`test/tests/rpc/helpers.rs`) is also being added by #6695 (`debug_traceCall`). Both PRs introduce the module with the same name, similar shape, and partially-overlapping helpers. Whichever lands second will hit a merge conflict — and that's actually the good outcome; the worse one is both landing with subtly-different copies and drifting from there.

ElFantasma (approved):
Three non-blocking comments

PR #6701, opened 2026-05-21
feat(l1): implement debug_dumpBlock endpoint

### LINKED ISSUE #6650: Implement debug_dumpBlock
Implement `debug_dumpBlock`. Full state dump at a given block — effectively `debug_accountRange` without pagination.

References:
- [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debug-dumpblock)
- [geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/api_debug.go)

Implementation notes:
- Should be implemented on top of the iterator from `debug_accountRange` (i.e. depends on or shares code with it).
- Output is unbounded — document the resource implications and consider whether to gate behind a flag.

Parent: #6572

azteca1998:
## Summary
- Implement `debug_dumpBlock` RPC endpoint that dumps the entire state at a given block
- Returns the state root and all accounts with their balance, nonce, storage root, and code hash
- Uses `iter_accounts` to iterate the state trie at the block's state root

Closes part of #6572

Closes #6650

### DIFF
--- a/crates/networking/rpc/debug/dump_block.rs
+++ b/crates/networking/rpc/debug/dump_block.rs
@@ -100,11 +100,11 @@ impl RpcHandler for DumpBlockRequest {
             .block
             .resolve_block_number(&context.storage)
             .await?
-            .ok_or(RpcErr::Internal("Block not found".to_string()))?;
+            .ok_or(RpcErr::WrongParam("Block not found".to_string()))?;
         let header = context
             .storage
             .get_block_header(block_number)?
-            .ok_or(RpcErr::Internal("Block header not found".to_string()))?;
+            .ok_or(RpcErr::WrongParam("Block header not found".to_string()))?;
 
         let state_root = header.state_root;
         let start = self.config.start.unwrap_or_else(H256::zero);
--- a/crates/networking/rpc/debug/dump_block.rs
+++ b/crates/networking/rpc/debug/dump_block.rs
@@ -7,11 +7,16 @@ use serde_json::Value;
 
 use crate::{RpcApiContext, RpcErr, RpcHandler, types::block_identifier::BlockIdentifier};
 
-/// Default cap on accounts emitted by `debug_dumpBlock` when the caller does
-/// not specify `maxResults`. Picked to bound response size on large states —
-/// geth's `debug_dumpBlock` is unbounded; ethrex diverges here to protect the
-/// RPC server. Callers wanting more can paginate via the config object.
+/// Default number of accounts emitted by `debug_dumpBlock` when the caller does
+/// not specify `maxResults`. Callers wanting more can paginate via the `start`
+/// cursor in the config object.
 const DEFAULT_MAX_RESULTS: usize = 100_000;
+
+/// Hard ceiling on `maxResults` — callers cannot exceed this even via an
+/// explicit parameter. Picked to bound response size on large states; geth's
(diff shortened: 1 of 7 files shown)


crates/networking/rpc/debug/dump_block.rs
```diff
@@ -0,0 +1,199 @@
...
+            .await?
+            .ok_or(RpcErr::Internal("Block not found".to_string()))?;
+        let header = context
+            .storage
+            .get_block_header(block_number)?
+            .ok_or(RpcErr::Internal("Block header not found".to_string()))?;
+
+        let state_root = header.state_root;
+        let start = self.config.start.unwrap_or_else(H256::zero);
+        let max_results = self.config.max_results.unwrap_or(DEFAULT_MAX_RESULTS);
```
ElFantasma: `DEFAULT_MAX_RESULTS` only acts as a *default*, not a *ceiling* — a caller passing `{"maxResults": <huge>}` overrides it entirely and dumps the full state into one in-memory `BTreeMap`, which is exactly the unbounded-response DoS the constant's doc says it's protecting against:

> Picked to bound response size on large states ... ethrex diverges here to protect the RPC server.

If the protection is meant to hold regardless of caller input, clamp to a hard ceiling:
```rust
/// Absolute cap; callers cannot exceed this even via `maxResults`.
const MAX_RESULTS_CEILING: usize = 100_000;
...
let max_results = self
    .config
    .max_results
    .unwrap_or(DEFAULT_MAX_RESULTS)
    .min(MAX_RESULTS_CEILING);
```
Then large states are walked via the `next` cursor across multiple calls instead of one unbounded response. (If unbounded single-shot dumps are intentionally allowed for trusted debug operators, a one-line doc note saying so would resolve the apparent contradiction with the "protect the RPC server" rationale.)

ElFantasma (commented):
**Caller-supplied `maxResults` bypasses the DoS protection the default is there to provide.** The doc says `DEFAULT_MAX_RESULTS` exists "to protect the RPC server" against unbounded dumps — but `self.config.max_results.unwrap_or(DEFAULT_MAX_RESULTS)` lets a caller pass `{"maxResults": 18446744073709551615}` and dump the entire state into one un-paginated `BTreeMap` in memory, defeating that protection. If the protection is real, the value should be clamped to a hard ceiling rather than only used as a default — see inline. (Mitigating context: `debug_*` is typically gated behind an access-controlled namespace, so the exposure is limited to operators who deliberately enabled the debug RPC. Still worth a ceiling, since the stated intent is server protection.)

--------

**Also, PR should point out that it closes the sub-issue instead of the tracking issue**

crates/networking/rpc/debug/dump_block.rs
```diff
@@ -0,0 +1,207 @@
...
+            .get_block_header(block_number)?
+            .ok_or(RpcErr::WrongParam("Block header not found".to_string()))?;
+
+        let state_root = header.state_root;
+        let start = self.config.start.unwrap_or_else(H256::zero);
+        let max_results = self
+            .config
+            .max_results
+            .unwrap_or(DEFAULT_MAX_RESULTS)
+            .min(DEFAULT_MAX_RESULTS);
```
ElFantasma: `.unwrap_or(D).min(D)` collapses to just `.unwrap_or(D)` when both defaults are the same constant

ElFantasma (approved):
Left a comment on a redundant `min`

PR #6705, opened 2026-05-21
feat(l1): implement flatCallTracer

### LINKED ISSUE #6647: Implement flatCallTracer for debug_trace*
Implement `flatCallTracer` for `debug_traceTransaction` / `debug_traceBlock*`. Parity-style flat call tree (one entry per call frame with a `traceAddress` path) instead of the nested structure produced by `callTracer`.

[geth source](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/native/call_flat.go)

Implementation notes:
- New `TracerType::FlatCallTracer { config: { convertParityErrors, includePrecompiles } }`.
- Reuses much of `callTracer`'s frame tracking but emits a flat array with `traceAddress` paths.
- `convertParityErrors` rewrites geth-style error strings to Parity-style names for compatibility.

Parent: #6572

azteca1998:
## Summary
- Add `flatCallTracer` tracer type for `debug_traceTransaction` and block tracing endpoints
- Returns a flat array of call frames following the Parity/OpenEthereum trace format
- Each frame includes `traceAddress` (position in call tree as array of indices) and `subtraces` (direct child count)
- Uses the `action`/`result` sub-object structure: `action.callType`, `action.from`, `action.to`, etc.
- Supports frame types: `"call"`, `"create"`, and `"suicide"`
- Implemented by post-processing the existing call tracer output (depth-first recursive flattening)

Closes part of #6572

Closes #6647

### DIFF
--- a/crates/networking/rpc/tracing.rs
+++ b/crates/networking/rpc/tracing.rs
@@ -503,3 +503,143 @@ fn flatten_recursive(
         flatten_recursive(sub_call, &child_address, result);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::rpc::RpcHandler;
+    use serde_json::json;
+
+    // --- TracerType deserialization tests ---
+
+    #[test]
+    fn deserialize_tracer_type_flat_call() {
+        let t: TracerType = serde_json::from_value(json!("flatCallTracer")).unwrap();
+        assert!(matches!(t, TracerType::FlatCallTracer));
+    }
+
+    #[test]
+    fn deserialize_tracer_type_unknown_fails() {
+        assert!(serde_json::from_value::<TracerType>(json!("unknownTracer")).is_err());
+    }
+
+    // --- flatCallTracer parse test ---
+
+    #[test]
+    fn parse_trace_tx_flat_call_tracer() {
+        let params = Some(vec![
+            json!("0x0000000000000000000000000000000000000000000000000000000000000001"),
+            json!({"tracer": "flatCallTracer"}),
+        ]);
+        let req = TraceTransactionRequest::parse(&params).unwrap();
+        assert!(matches!(req.trace_config.tracer, TracerType::FlatCallTracer));
+    }
+
+    // --- flatten_call_trace tests ---
+
+    #[test]
+    fn flatten_single_frame() {
+        let frame = CallTraceFrame {
+            call_type: CallType::CALL,
+            from: Address::zero(),
+            to: Address::from_low_u64_be(1),
+            gas: 21000,
+            gas_used: 21000,
+            input: Bytes::new(),
+            output: Bytes::new(),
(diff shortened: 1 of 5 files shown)


test/tests/rpc/helpers.rs
```diff
@@ -0,0 +1,245 @@
+//! Shared setup helpers for the rpc integration tests. Each test file builds
```
ElFantasma, line 1: Heads up: this file is also being added by #6695 (`debug_traceCall`) and #6698 (`debug_accountRange`). All three copies share the core scaffolding (`setup_store`, `rpc_call`, `build_and_execute_block`, `setup_single_transfer_block`) but each PR has bolted on its own extras — #6695 adds `setup_genesis_only`, this one adds `create_deploy_tx` + `setup_single_deploy_block`, and #6698 has its own subset. The three head copies are 6737, 6169, and 8071 bytes respectively, so they're already divergent.

Whichever PR lands first wins the file; the other two will hit merge conflicts that need resolving by hand, and the risk is that two of the three end up landing with subtly-different copies of the shared helpers and drift from there. Coordination with @azteca1998 on the other two PRs to land `helpers.rs` once via the earliest-merging branch (or as a separate prep PR) and have the feature PRs just `use` from it would avoid that. Non-blocking on this PR's correctness.

crates/networking/rpc/tracing.rs
```diff
@@ -355,6 +380,396 @@ impl RpcHandler for TraceBlockByNumberRequest {
...
+                output: frame.output.clone(),
+            },
+        })
+    };
+
+    result.push(FlatCallFrame {
+        action,
+        error: frame.error.clone(),
+        result: result_value,
+        subtraces: frame.calls.len(),
```
ElFantasma, line 581: Soft invariant worth pinning down: for the suicide shape to be Parity-compliant, the SELFDESTRUCT frame must always be a leaf — `subtraces: 0` and no children to recurse into. Today that holds because the producer side (`opcode_handlers/system.rs`) calls `tracer.enter(SELFDESTRUCT, ...)` immediately followed by `tracer.exit_early(0, None)`, so `frame.calls` is guaranteed empty.

But the unconditional `subtraces: frame.calls.len()` + `for sub_call in frame.calls.iter()` here doesn't reflect that contract — if a future producer-side change ever pushed a child under a SELFDESTRUCT frame, the suicide frame would silently emit `subtraces > 0` and the children would inherit a `traceAddress` rooted at a `"type": "suicide"` parent, which neither Parity nor geth clients expect.

Cheap insurance: `debug_assert!(!matches!(frame.call_type, CallType::SELFDESTRUCT) || frame.calls.is_empty(), "suicide frames must be leaves")` here, or just a `// SELFDESTRUCT frames are leaves by producer contract (see exit_early in OpSelfdestruct).` comment alongside the suicide arm. Non-blocking.

PR #6707, opened 2026-05-21
perf(l1): replace index based jump targets with bit vector

benbencik:
**Motivation**
The motivation for this PR are EEST tests, that crash because of exceeding memory limits on zkVM.
```
test_jumpdest_analysis[fork_Osaka-blockchain_test-5b-benchmark-gas-value_60M]
test_jumpdest_analysis[fork_Osaka-blockchain_test-605b5b-benchmark-gas-value_60M]
```
Currently the jump targets are stored as a vector of `u32` indices, this change switches to storing it as a bit vector.

**Description**

This change fixes the crashing tests. However the test is constructed in an unrealistic adversarial setting where most of the operations are just jumps. The question is whether this is actually practical for normal execution. 

I check the storage required for `jump_target` on 500 blocks 20260424 - 24950286, by counting the jump instructions compared to length of bytecode. The total storage for index-based `jump_target` across all contracts is 21.37MB which is ~10% higher than 19.29MB for the bit vector approach. There are contracts for which bit vector consumes more storage but on average it seems to be more efficient. 

In addition, the bit vector is faster because lookups are constant time, compared to binary search. I benchmarked the costs on Zisk, and there is a slight ~2% improvement. But primarily, the point of this change is to resolve the possible attack vector.

It is worth mentioning that other clients also seem to use the bit vector approach:
- [JumpTable](https://docs.rs/revm-bytecode/11.0.0/revm_bytecode/struct.JumpTable.html) in Reth
- [BitVec](https://github.com/ethereum/go-ethereum/blob/efe58eac003940e736717ab2207f6db22fdbd2e6/core/vm/analysis_legacy.go#L28-L31) in Geth

The new format defined by `encode_code` is breaking! It might be reasonable to recompute and store `Vec<u32>` for backwards compatibility. I leave this up to discussion.

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -851,7 +851,7 @@ impl Store {
         let code = Code {
             hash: code_hash,
             bytecode,
-            jump_targets: <Vec<u8>>::decode(targets)?,
+            jump_targets: Arc::from(<Vec<u8>>::decode(targets)?),
         };
 
         // insert into cache and evict if needed
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -851,7 +851,7 @@ impl Store {
         let code = Code {
             hash: code_hash,
             bytecode,
-            jump_targets: <Vec<u32>>::decode(targets)?.into(),
+            jump_targets: <Vec<u8>>::decode(targets)?,
         };
 
         // insert into cache and evict if needed
@@ -3527,9 +3527,7 @@ pub fn receipt_key(block_hash: &BlockHash, index: u64) -> Vec<u8> {
 }
 
 fn encode_code(code: &Code) -> Vec<u8> {
-    let mut buf = Vec::with_capacity(
-        6 + code.bytecode.len() + std::mem::size_of_val::<[u32]>(&code.jump_targets),
-    );
+    let mut buf = Vec::with_capacity(6 + code.bytecode.len() + code.jump_targets.len());
     code.bytecode.encode(&mut buf);
     // `Arc<[u32]>` (the in-memory share) has no `RLPEncode` impl; encode through an
     // owned `Vec` on this cold DB-write path (code is persisted once per hash).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@
 
 - Two-CF receipts migration: copy old RLP-keyed receipts to `receipts_v2` with fixed-width keys; old CF auto-dropped on restart [#6598](https://github.com/lambdaclass/ethrex/pull/6598)
 
+### 2026-05-21
+
+- Two-CF receipts migration: copy old RLP-keyed receipts to `receipts_v2` with fixed-width keys; old CF auto-dropped on restart [#6598](https://github.com/lambdaclass/ethrex/pull/6598)
+
 ### 2026-04-27
 
 - Reduce peak disk usage during snap sync by moving SST files into the temp DB instead of copying [#6532](https://github.com/lambdaclass/ethrex/pull/6532)
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -34,7 +34,7 @@ pub struct Code {
     pub hash: H256,
     pub bytecode: Bytes,
     // Jump targets stored as a bit vector: bit i is set iff position i is a valid JUMPDEST
-    // The size is exactly the bytecode length
+    // The size is ceil(bytecode.len() / 8) bytes
     pub jump_targets: Vec<u8>,
 }
 
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -33,13 +33,9 @@ pub struct Code {
     // endpoints to access that hash, saving one expensive Keccak hash.
     pub hash: H256,
     pub bytecode: Bytes,
-    // `Arc<[u32]>` so cloning `Code` (hot: every message-call resolves and clones
-    // the callee's code) is a refcount bump instead of deep-copying the table.
-    // Serializes via serde's `rc` feature (enabled workspace-wide).
-    // The valid addresses are 32-bit because, despite EIP-3860 restricting initcode size,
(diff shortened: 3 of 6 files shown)


crates/storage/store.rs
```diff
@@ -714,7 +714,7 @@ impl Store {
         let code = Code {
             hash: code_hash,
             bytecode,
-            jump_targets: <Vec<_>>::decode(targets)?,
+            jump_targets: <Vec<u8>>::decode(targets)?,
```
ElFantasma: **Decode-side pivot point for the three options.**
- **A** (current diff): keep this as `<Vec<u8>>::decode(targets)?`, plus bump `STORE_SCHEMA_VERSION` so old `Vec<u32>` rows don't reach this line.
- **B**: revert this line to `<Vec<u32>>::decode(targets)?` (today's format), then convert the indices to the bitmap before constructing `Code`. No schema interaction.
- **C**: remove this read entirely. `Code::from_bytecode_unchecked(bytecode, code_hash)` already computes `jump_targets` from `bytecode`, so the decode path just needs to tolerantly skip the trailing legacy field for old rows (and the bump handles the rest).

crates/storage/store.rs
```diff
@@ -3264,9 +3264,7 @@ fn snap_state_key(index: SnapStateIndex) -> Vec<u8> {
 }
 
 fn encode_code(code: &Code) -> Vec<u8> {
-    let mut buf = Vec::with_capacity(
-        6 + code.bytecode.len() + std::mem::size_of_val(code.jump_targets.as_slice()),
-    );
+    let mut buf = Vec::with_capacity(6 + code.bytecode.len() + code.jump_targets.len());
     code.bytecode.encode(&mut buf);
     code.jump_targets.encode(&mut buf);
```
ElFantasma: **Encode-side pivot point.**
- **A** (current diff): write the bitmap as today's `code.jump_targets.encode(&mut buf);` — paired with the schema bump.
- **B**: convert the bitmap → `Vec<u32>` of set-bit positions and `encode` that. On-disk bytes match what `main` produces; no DB cares.
- **C**: drop this line entirely. The codes table loses the trailing field; new rows are bytecode-only and the decoder recomputes `jump_targets` on load.

ElFantasma (commented):
The bit-vector approach is the right call: it fixes the adversarial `JUMPDEST`-saturated OOM, gives O(1) lookup vs binary search, and matches geth's `bitvec` / reth's `JumpTable`. I verified the correctness equivalence — `is_valid_jump_target(target)` is exactly the old two-part `(byte == JUMPDEST) && binary_search(target).is_ok()` check, since `compute_jump_targets` only sets bits at real `0x5B` opcodes outside PUSH data and padding bits in the final byte are never written.

You flagged the on-disk format question in the PR description and opened it for discussion — I think there are three coherent options here, and surfacing them so the team can pick. All three preserve the in-memory bitmap, so the adversarial OOM fix and O(1) lookup hold under any of them. The choice is essentially: *do we want a derived cache persisted, and if so, in which form?*

**Option A — current PR shape: bitmap in memory + bitmap on disk**
- One change beyond what you already have: bump `STORE_SCHEMA_VERSION` from 2 to 3 (it's currently 2 on `main`; #6737 is also bumping it to 3, so coordination there). The migration gate at `store.rs:1473` then catches existing DBs instead of mis-decoding their `Vec<u32>` rows as `Vec<u8>`.
- **Wins**: smallest delta on top of the existing diff; codes table shrinks ~10% on average per your measurement.
- **Costs**: a schema bump now (cheap mechanism, but it is a migration); any future change to the bitmap layout (word size, packing, RLE, …) would need another bump.

**Option B — bitmap in memory only, on-disk format unchanged**
- Memory representation is the bitmap (so OOM + O(1) lookup wins hold). At the disk boundary, `encode_code` converts bitmap → `Vec<u32>` of set-bit positions; `decode_code` reads `Vec<u32>` and converts to bitmap before constructing `Code`. The on-disk format is bit-identical to what's on `main` today.
- **Wins**: zero migration concern by construction — old DBs decode unchanged, no `STORE_SCHEMA_VERSION` interaction at all.
- **Costs**: codes table doesn't shrink; two linear conversion passes at the disk boundary (microseconds per code, but pure overhead with no upside); two representations to maintain (memory ≠ disk, so a future reader has to track which form they're holding).

**Option C — don't persist `jump_targets` at all (recompute on load)**
- `compute_jump_targets(&bytecode)` is already called in `Code::from_bytecode` / `from_bytecode_unchecked`, so it's a pure function of bytecode. `encode_code` writes just the bytecode; `decode_code` runs `compute_jump_targets` once and constructs `Code` with the resulting bitmap.
- **Wins**: smallest codes table (no `jump_targets` field on disk at all); removes a derived field from the persistence layer permanently, so future bitmap-layout changes are zero-disk-impact; matches geth and reth (both recompute on load, no persisted JUMPDEST cache).
- **Costs**: schema bump still needed because old rows carry the trailing legacy `jump_targets` field — decode has to either tolerantly skip it or migrate; one `compute_jump_targets` pass per cold code-load (microseconds per code; dwarfed by EVM exec, and the in-memory code cache means cold loads are infrequent in practice).

My lean: **A or C**. A is fine and minimizes change vs what you've written; C is a touch nicer architecturally (don't store what you can compute). B's only real upside (no migration) doesn't move the needle since the bump is cheap to do, and it carries two costs forever (the per-boundary conversions and the two-representation maintenance load) — I'd put it last.

Inlines below mark the two lines in `store.rs` where Options A/B/C diverge, so reviewers can see the pivot points in code.

benbencik:
I agree that option B is not useful. Both A and C seem reasonable, I slightly prefer A. But I will let you (@ElFantasma, @ilitteri) decide, perhaps you have more context.

benbencik:
Settled on the **option A** and added the migration function. Rebased on main,  so our migration is now `migrate_3_to_4`.

PR #6709, opened 2026-05-21
feat(l1): move getBlockAccessList from debug to eth namespace

edg-l:
## Summary

[execution-apis](https://github.com/ethereum/execution-apis) defines `getBlockAccessList` under the `eth_` namespace, not `debug_`. This PR aligns ethrex with the spec.

- Rename `debug_getBlockAccessList` → `eth_getBlockAccessList`.
- Move `rpc/debug/block_access_list.rs` → `rpc/eth/block_access_list.rs`; dispatch moves from `map_debug_requests` to `map_eth_requests`.
- Return the structured JSON shape defined by [`BlockAccessList`](https://github.com/ethereum/execution-apis/blob/main/src/schemas/block-access-list.yaml) (array of `AccountAccess` with camelCase fields, `hash32`/quantity hex per spec) instead of an RLP-encoded hex string.
- Fast-path: read from `storage.get_block_access_list(hash)` (populated at block import in `blockchain.rs:1936`) before falling back to `generate_bal_for_block`.
- Spec conformance: unknown blocks now return `null` (`notFound` schema) instead of an Internal RPC error.
- Update REPL command catalog accordingly.

No back-compat alias for the old `debug_` name.

## Spec references

- Method: `execution-apis/src/eth/block.yaml:245`
- Schema: `execution-apis/src/schemas/block-access-list.yaml`

## Perf notes

The previous handler always re-executed the full block, which is O(block-execution-time) — hundreds of ms to several seconds per call for a full block. The new fast-path is an RLP decode of the stored BAL when the block was imported in this process lifetime (the common case for post-Amsterdam tip lookups since `add_block` writes the BAL on import).

The slow re-execute path still applies for:
- Historical post-Amsterdam blocks that were imported before BAL storage existed.
- L2's `blockchain.add_block` path, which does not currently call `store_block_access_list` (separate follow-up).

A small LRU on top is a possible future optimization but kept out of scope here.

## Test plan

- [ ] `cargo check -p ethrex-rpc -p ethrex-repl`
- [ ] `cargo clippy -p ethrex-rpc -p ethrex-repl --no-deps`
- [ ] `cargo fmt --all -- --check`
- [ ] Manual: call `eth_getBlockAccessList` against a post-Amsterdam block on a devnet and check JSON shape matches the spec example.
- [ ] Manual: call against a pre-Amsterdam block; expect `null`.
- [ ] Manual: call against an unknown block hash; expect `null` (not an error).


crates/networking/rpc/eth/block_access_list.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+}
+
+/// Serializes a BlockAccessList into the JSON shape defined by execution-apis
+/// `eth_getBlockAccessList` (EIP-7928): an array of AccountAccess objects with
+/// camelCase fields and per-spec hex encodings (hash32 = full 32-byte hex,
+/// quantities = no-leading-zero hex).
+fn bal_to_json(bal: &BlockAccessList) -> Value {
+    Value::Array(bal.accounts().iter().map(account_to_json).collect())
```
ElFantasma, line 65: **Worth a unit test against the BlockAccessList → JSON conversion.** The whole point of this PR is wire-shape conformance to execution-apis, and `bal_to_json` is where any drift would land. A small test would catch every format mistake before deploy:

```rust
#[test]
fn bal_to_json_matches_spec_shape() {
    use ethrex_common::types::block_access_list::{AccountChanges, BalanceChange, BlockAccessList, NonceChange, SlotChange, StorageChange};
    let mut acc = AccountChanges::new(Address::from_low_u64_be(0x42));
    acc.balance_changes.push(BalanceChange::new(1, U256::from(1_000)));
    acc.nonce_changes.push(NonceChange::new(2, 7));
    acc.storage_reads.push(U256::from(5));
    let mut sc = SlotChange::new(U256::from(3));
    sc.slot_changes.push(StorageChange::new(4, U256::from(0xdead)));
    acc.storage_changes.push(sc);
    let bal = BlockAccessList::new(vec![acc]);

    let j = bal_to_json(&bal);
    assert_eq!(j[0]["address"], "0x0000000000000000000000000000000000000042");
    assert_eq!(j[0]["balanceChanges"][0], json!({"index": "0x1", "value": "0x3e8"}));
    assert_eq!(j[0]["nonceChanges"][0], json!({"index": "0x2", "value": "0x7"}));
    assert_eq!(j[0]["storageReads"][0], "0x0000000000000000000000000000000000000000000000000000000000000005");
    assert_eq!(j[0]["storageChanges"][0]["key"], "0x0000000000000000000000000000000000000000000000000000000000000003");
    assert_eq!(j[0]["storageChanges"][0]["changes"][0]["value"], "0x000000000000000000000000000000000000000000000000000000000000dead");
}
```
This is the cheap version of every checkbox in the PR test plan. Re-runs forever and doesn't need a devnet.

crates/networking/rpc/eth/block_access_list.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        // Slow path: re-execute the block. Returns None for pre-Amsterdam blocks.
+        let Some(block) = context.storage.get_block_by_hash(block_hash).await? else {
+            return Ok(Value::Null);
+        };
+
+        let bal = context
+            .blockchain
+            .generate_bal_for_block(&block)
+            .map_err(|e| RpcErr::Internal(format!("Failed to generate BAL: {e}")))?;
```
ElFantasma, line 51: nit: spec asymmetry. `generate_bal_for_block` returning `Ok(None)` is treated as `Value::Null` (the `match bal` arm below); returning `Err` becomes `RpcErr::Internal`. The boundary between the two is `generate_bal_for_block`'s implementation detail — pre-Amsterdam is `Ok(None)`, but "post-Amsterdam block whose state is missing from disk" might be either depending on how the function fails. 

If the intent is "known block, can't generate → also null" (which matches the spirit of `notFound`), this branch could swallow more error variants. If the intent is "known block, real error → Internal", a one-line doc-comment here explaining the split would save the next reader from having to chase down `generate_bal_for_block`. Non-blocking; just calling out the boundary.

crates/networking/rpc/eth/block_access_list.rs
```diff
@@ -0,0 +1,133 @@
...
+    async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
+        // Per execution-apis, unknown blocks map to the `notFound` schema (null).
+        let Some(block_number) = self.block_id.resolve_block_number(&context.storage).await? else {
+            return Ok(Value::Null);
+        };
+
+        let Some(header) = context.storage.get_block_header(block_number)? else {
+            return Ok(Value::Null);
+        };
+        let block_hash = header.hash();
```
ElFantasma: nit on the hash round-trip. For a `BlockIdentifier::Hash(h)` input, this code does `resolve_block_number(h) → get_block_header(number) → header.hash()` — three lookups to reproduce the hash the caller already gave us. For `BlockIdentifier::Number`/`Tag` it's the natural path.

Not worth restructuring just for this method, but if `BlockIdentifier` ever grows a `resolve_block_hash(&self, storage) -> Result<Option<H256>>` helper that short-circuits the `Hash` variant, this handler (and any others doing the same dance) gets cheaper for free. Filing under "future-work" rather than "change now."

crates/networking/rpc/rpc.rs
```diff
@@ -1021,6 +1021,7 @@ pub async fn map_eth_requests(req: &RpcRequest, context: RpcApiContext) -> Resul
             GetTransactionByBlockHashAndIndexRequest::call(req, context).await
         }
         "eth_getBlockReceipts" => GetBlockReceiptsRequest::call(req, context).await,
+        "eth_getBlockAccessList" => BlockAccessListRequest::call(req, context).await,
```
ElFantasma, line 1108: Worth a one-line `CHANGELOG.md` entry for the breaking RPC rename. The PR is correct in not adding a `debug_getBlockAccessList` alias (the body explains why), but any external tooling that integrated against the `debug_` name will fail with `MethodNotFound` after upgrade. A "Breaking: `debug_getBlockAccessList` renamed to `eth_getBlockAccessList` (now spec-conformant per execution-apis EIP-7928)" line would surface this for downstream integrators reading release notes.

ilitteri (approved):

azteca1998 (approved):

iovoid (commented):
Should this be marked as a breaking change?

ElFantasma (approved):
Approving it, but I still suggest to add the `CHANGELOG.md` entry for this.

Issue #764, opened 2024-10-08
LEVM: (EF Tests) EIP198 Modexp precompile


iinaki:
To remove from ignored this suites:

- `"eip198_modexp_precompile".into(),`

PR #6710, opened 2026-05-21
feat(l1): engine REST/SSZ API

iovoid:
**Motivation**

We want to reduce the overhead of JSON serialization/deserialization for the large objects handled by the engine API. One of the proposals to do so is https://github.com/ethereum/execution-apis/pull/764

**Description**

Most of the PR consists of boilerplate: defining RPC objects, route handlers and serialization/deserialization logic.

**Benchmarks**

telegram-cloud-photo-size-1-4936507802709068985-y

telegram-cloud-photo-size-1-4936507802709068986-y


edg-l (changes requested):
Major:

https://github.com/lambdaclass/ethrex/blob/6e808f9d0d77422b6a5bfe83d1dd773fccbcb3ee/crates/networking/rpc/engine_rest/handlers/payloads.rs#L366-L380

`get_payload_v5` returns `GetPayloadResponseV5` whose `execution_payload` is `ExecutionPayloadV3` (no `block_access_list`, no `slot_number`), but `new_payload_v5` at line 147 accepts `ExecutionPayloadV4` and gates on `is_amsterdam_activated`. The JSON-RPC `engine_getPayloadV5` correctly emits BAL via `ExecutionPayload::from_block(block, block_access_list)`. A CL building a block over REST will get a payload missing the BAL and cannot reproduce the block hash. Either bump `GetPayloadResponseV5.execution_payload` to `ExecutionPayloadV4` and call `json_to_execution_payload_v4`, or align the V5/V6 split with the spec.

https://github.com/lambdaclass/ethrex/blob/6e808f9d0d77422b6a5bfe83d1dd773fccbcb3ee/crates/networking/rpc/engine_rest/handlers/blobs.rs#L24-L31

Off-by-one in `check_count`: `if n >= MAX_BLOB_HASHES_REQUEST` rejects exactly 128 hashes, making the effective max 127. Compare with the bodies handler which correctly uses `>`. Change to `if n > MAX_BLOB_HASHES_REQUEST`.

Minor:

https://github.com/lambdaclass/ethrex/blob/6e808f9d0d77422b6a5bfe83d1dd773fccbcb3ee/crates/networking/rpc/engine_rest/auth.rs#L25-L31

`format!("auth failed: {e:?}")` returns debug-formatted JWT crate errors in the 401 body. Use a fixed string like `"authentication failed"` to match the JSON-RPC auth path.

https://github.com/lambdaclass/ethrex/blob/6e808f9d0d77422b6a5bfe83d1dd773fccbcb3ee/crates/networking/rpc/engine_rest/mod.rs#L17-L19

The 256 MB `DefaultBodyLimit` is correct for `newPayload` but lets a peer force 256 MB allocations on tiny endpoints like `capabilities` or `client/version`. Per-route caps from #764 §Security are not enforced; worth a tracking issue.

Nit:

https://github.com/lambdaclass/ethrex/blob/6e808f9d0d77422b6a5bfe83d1dd773fccbcb3ee/test/tests/rpc/engine_rest_tests.rs#L1

Tests cover happy paths, auth bypass, malformed SSZ, missing Content-Type, and pre-Shanghai withdrawal rejection. Missing: expired/future JWT, oversized blob-hash batch (which would have caught the off-by-one), and a V3/V4/V5 `newPayload` end-to-end. Worth filling in before relying on REST in production.

edg-l:
Re-reviewed at 56499614a. Previous five findings all addressed (BAL/Amsterdam split on getPayloadV5, off-by-one in `check_count`, auth body, per-route caps, tests). Two new findings, both spec divergences from JSON-RPC (the SSZ transport should be behaviorally equivalent):

**Major: `new_payload_v5` doesn't reject empty BAL**
https://github.com/lambdaclass/ethrex/blob/56499614a518b0b7bb2ffb456633b6f1ef447d88/crates/networking/rpc/engine_rest/handlers/payloads.rs#L147-L188

JSON-RPC returns `-32602 WrongParam("block_access_list required in engine_newPayloadV5")` (engine/payload.rs:344). REST treats empty bytes as `raw_bal_hash = None`, builds a block with `block_access_list_hash = None`, and surfaces `PayloadStatus::INVALID` instead. EIP-7928 makes BAL mandatory in V5 — its absence is structural, not a validity failure. Fix: return `EngineRestError::bad_request(\"block_access_list required in engine_newPayloadV5\")` before `ssz_payload_v4_to_block` when `req.execution_payload.block_access_list.is_empty()`. Update `new_payload_v5_pre_amsterdam_returns_422` to use a non-empty BAL, and add an `empty_bal_returns_400` test.

**Major: EIP-7928 fork-boundary detector missing on REST V4/V5**
The alt-header rebuild at engine/payload.rs:396-406 distinguishes a V5-shaped block misrouted to V4 (`-32602`) from a true value mismatch (`INVALID`). The REST `new_payload_v4` (and `new_payload_v5` when `block_access_list_hash = None`) collapses both into `INVALID`. EELS fixture `test_invalid_post_fork_block_without_bal_hash_field [fork_BPO2ToAmsterdamAtTime15k-blockchain_test_engine]` expects `-32602`; REST conformance fails without it. The revert commit (56499614a) already leans on the JSON-RPC detector as authoritative — REST needs the same one. Fix: port the alt-header trick into `new_payload_v4` and mirror it in `new_payload_v5` for the `block_access_list_hash = None` case.

Minor/nits unchanged from offline notes (`decode_ssz` debug-formats error to body, `error_kind_from_status(422)` mislabels `UnsupportedFork` metrics, `MAX_BLOB_HASHES_REQUEST` boundary test only covers V1, `require_osaka_tip` returns 500 on empty store).

edg-l (approved):

azteca1998 (approved):

crates/networking/rpc/engine_rest/handlers/bodies.rs
```diff
@@ -0,0 +1,283 @@
...
+            .map(|(body_opt, key)| {
+                let Some(body) = body_opt else {
+                    return Ok(None);
+                };
+                let Some(header) = fetch_header(&storage, &key)? else {
+                    return Ok(None);
+                };
+                let block = Block { header, body };
+                let bal = blockchain
+                    .generate_bal_for_block(&block)
```
ElFantasma, line 183: `generate_bal_for_block` re-executes the block in the EVM, and `assemble_blocks_with_bal` runs that loop for *every* requested body, sequentially inside `spawn_blocking`. With `MAX_PAYLOAD_BODIES_REQUEST = 32`, a single `/engine/v2/payloads/bodies/by-hash` call from the CL can synchronously re-execute 32 full blocks before the response goes out — and the CL may issue this every slot during catch-up. Worst-case it's pinning a blocking thread for a noticeable fraction of a slot.

Post-Amsterdam blocks already carry `block_access_list_hash` in their header, so the BAL itself is presumably persisted by the block-import path. If it can be fetched from storage instead of re-derived (`storage.get_block_access_list(hash)?`), only the pre-Amsterdam-but-on-an-Amsterdam-chain case would need re-generation — and even there, returning `bal: None` (which the SSZ V2 body shape already encodes nullable) is spec-valid for blocks predating EIP-7928. Worth checking whether storage-side BAL retrieval exists before paying the re-execution cost. Non-blocking; correctness is fine.

crates/networking/rpc/engine_rest/handlers/blobs.rs
```diff
@@ -0,0 +1,210 @@
...
+        return Err(EngineRestError::payload_too_large(format!(
+            "request exceeds MAX_BLOB_HASHES_REQUEST ({MAX_BLOB_HASHES_REQUEST})"
+        )));
+    }
+    Ok(())
+}
+
+fn copy_blob(b: &[u8]) -> Blob {
+    let mut out = [0u8; BLOB_SIZE];
+    out.copy_from_slice(b);
```
ElFantasma, line 37: `copy_from_slice` panics if `b.len() != BLOB_SIZE`, and this runs inside the axum handler — a producer-side regression that ever yielded a short or oversized buffer would crash the request task (`PoisonError` on shared state in the worst case, but at minimum a 500 with no useful diagnostic in the response body). The current callers (`mempool.get_blobs_data_by_versioned_hashes`) presumably guarantee `BLOB_SIZE` by type, but the panic is a long-range coupling to that invariant.

A defensive bounds check is cheap:
```rust
fn copy_blob(b: &[u8]) -> Result<Blob, EngineRestError> {
    if b.len() != BLOB_SIZE {
        return Err(EngineRestError::internal(format!(
            "blob size {} != {BLOB_SIZE}", b.len(),
        )));
    }
    let mut out = [0u8; BLOB_SIZE];
    out.copy_from_slice(b);
    Ok(out)
}
```
Three call sites in this file would need to `?` it. Non-blocking — but a panic in a JWT-authed handler is a sharper edge than a 500.

ElFantasma (approved):
Two non-blocking inlines.

iovoid:
Closed in favor of #6770 since execution-apis/793 was chosen over execution-apis/764.

PR #6712, opened 2026-05-22
perf(l1): batch account-state prefetch via rocksdb multi_get_cf

edg-l:
## Summary

The BAL prewarmer's `CachingDatabase::prefetch_accounts` issues one rocksdb point-get per address, parallelized via rayon. For a BAL with thousands of addresses, that is thousands of independent column-family lookups, each going through bloom + block-cache + (cache miss) SST read.

This change collapses the prefetch into a single `batched_multi_get_cf` call on `ACCOUNT_FLATKEYVALUE` for addresses past the FKV generator's cursor, falling back to per-address trie walks for paths not yet swept by the generator. The diff-layer cache is checked first per address, matching the existing `Trie::get` semantics.

Plumbing:

- `StorageReadView::multi_get` (default loops `get`; RocksDB override uses `batched_multi_get_cf`)
- `Store::get_account_states_batch_by_root` ; FKV multi_get + trie fallback
- `VmDatabase::get_account_states_batch` ; trait method, overridden on `StoreVmDatabase`
- `LevmDatabase::get_account_states_batch` ; trait method, overridden on `DynVmDatabase`
- `CachingDatabase::prefetch_accounts` ; dispatches to the batch path; default impl loops for non-rocksdb backends

## Benchmark

Fixture: `bal-devnet-7-mainnet-mix-460` (460 blocks, ~30 Ggas, transfer/EVM mix). `release-with-debug`, `import-bench --with-bal`. Baseline = `b0b9a11c5d` (no perf change).

| metric          | baseline | this PR  | delta          |
| --------------- | -------- | -------- | -------------- |
| wall time       | 4.632 s  | 3.636 s  | -996 ms (-21.5%) |
| agg Ggas/s      | 8.016    | 11.066   | +38.0%         |
| avg ms / block  | 8.217    | 5.952    | -2.27 ms (-27.6%) |
| exec            | 7.286 ms | 5.073 ms | -2.21 ms (-30.4%) |
| warmer          | 6.777 ms | 2.369 ms | -4.41 ms (-65.0%) |
| store           | 0.656 ms | 0.660 ms | ~flat          |

Compare dashboard: https://edgl.dev/share/compare-bal-devnet-7-baseline-vs-fkv-multi_get.html

The warmer phase collapsing to a third of baseline is the smoking gun ; `warm_block_from_bal → prefetch_accounts` was the dominant cost there, and is now a single multi_get_cf for the FKV-covered subset.

## Notes

- Only applies to the BAL-driven parallel execution path (`warm_block_from_bal`); non-BAL imports go through `warm_block` and don't touch this method.
- Same optimization on storage prefetch was tried and reverted: each storage lookup is already a single rocksdb get (no trie walk to collapse), so par_iter point-gets beat multi_get's batching amortization on the warmer's wall-clock budget. See PR discussion for the perf data.

## Test plan

- [x] `cargo check --workspace`
- [x] `make -C tooling/ef_tests/blockchain test` ; 8819 passed, 0 failed
- [ ] Hive bal group
- [ ] CI green


crates/storage/store.rs
```diff
@@ -2460,6 +2460,95 @@ impl Store {
...
+                results[*slot] = Some(AccountState::decode(&encoded)?);
+            }
+        }
+
+        if !trie_indices.is_empty() {
+            // Fall back to the regular trie path for any addresses whose path
+            // hasn't been swept by the FKV generator yet.
+            let state_trie = self.open_state_trie(state_root)?;
+            for &i in &trie_indices {
+                results[i] = self.get_account_state_from_trie(&state_trie, addresses[i])?;
```
ElFantasma: **Sequential trie fallback (was parallel in `CachingDatabase::prefetch_accounts`).** The old `par_iter` got free parallelism on per-address trie walks for *every* address; the new path serializes them when `trie_indices` is non-empty.

For the bal-devnet-7 benchmark this is invisible (almost every address hits FKV), but during initial sync — when the FKV cursor is small and `trie_indices.len() >> fkv_indices.len()` — this is a regression vs `main` for the prefetch hot path.

If `Trie` (from `open_state_trie`) is `Send + Sync` (or clonable), the cheap version is:
```rust
use rayon::prelude::*;
let results_pairs: Result<Vec<_>, _> = trie_indices
    .par_iter()
    .map(|&i| self.get_account_state_from_trie(&state_trie, addresses[i]).map(|s| (i, s)))
    .collect();
for (i, s) in results_pairs? { results[i] = s; }
```
If `Trie` isn't `Send`, this needs a small refactor (each parallel branch opens its own trie tx). Worth a quick check on the worst-case workload (low FKV cursor) before merge.

crates/storage/store.rs
```diff
@@ -2460,6 +2460,95 @@ impl Store {
...
+        // point; see `flatkeyvalue_generator`).
+        let fkv_cursor = Nibbles::from_hex(last_written);
+        for (i, path) in leaf_paths.iter().enumerate() {
+            if let Some(value) = trie_cache.get(state_root, path.as_slice()) {
+                if !value.is_empty() {
+                    results[i] = Some(AccountState::decode(&value)?);
+                }
+                continue;
+            }
+            let path_nibbles = Nibbles::from_hex(path.clone());
```
ElFantasma: nit: `Nibbles::from_hex(path.clone())` allocates a fresh 65-byte `Vec<u8>` per address inside the loop, just to wrap it in `Nibbles` for the `>=` comparison. Since `Nibbles::from_hex` is a const wrapper (no normalization) and `PartialOrd<Nibbles>` is `Vec<u8>` lex compare under the hood, comparing slices avoids the alloc:
```rust
if fkv_cursor.as_ref() >= path.as_slice() {
    fkv_indices.push(i);
}
```
or expose `Nibbles::cmp_slice(&self, &[u8])`. For a BAL with ~5k addresses this is 5k tiny allocations; not material on the benchmark you posted, but free to fix and tightens the hot loop.

crates/storage/api/mod.rs
```diff
@@ -59,6 +59,18 @@ pub trait StorageReadView: Send + Sync {
...
 
+    /// Retrieves multiple values by key from the specified table.
+    /// Returns results in the same order as the input keys.
+    /// Backends that support batched reads (e.g. RocksDB `multi_get_cf`)
+    /// should override this for better throughput.
+    fn multi_get(
+        &self,
+        table: &'static str,
+        keys: &[&[u8]],
+    ) -> Vec<Result<Option<Vec<u8>>, StoreError>> {
```
ElFantasma, line 73: Small doc-callout: the default impl is a serial `get` loop, so for non-RocksDB backends (in-memory, etc.) callers see no speedup from `multi_get` vs N `get`s — only the alloc/dispatch shape changes. The current rustdoc says *"Backends that support batched reads ... should override this for better throughput"* which implies the converse correctly. Worth one extra line for callers:

> Callers should not assume `multi_get` is asymptotically faster than `get`; it is only an optimization for backends that have a batched read primitive. In particular, the in-memory backend's `multi_get` has the same cost as N independent `get`s.

Non-blocking; just a hint to future readers picking between `get` and `multi_get` for new code paths.

crates/storage/store.rs
edg-l: Good catch, parallelized with par_iter in fdb876d5fc; no threshold, rayon handles small N fine.

crates/storage/store.rs
edg-l: Fixed in fdb876d5fc; compare fkv_cursor and path as raw slices, drops the per-iter clone.

crates/storage/api/mod.rs
edg-l: Added the callout in fdb876d5fc.

azteca1998 (dismissed):

ElFantasma (approved):

crates/vm/db.rs
```diff
@@ -12,6 +12,19 @@ pub trait VmDatabase: Send + Sync + DynClone {
...
+    /// the flat key-value table) should override this.
+    fn get_account_states_batch(
+        &self,
+        addresses: &[Address],
+    ) -> Result<Vec<Option<AccountState>>, EvmError> {
+        addresses
+            .iter()
+            .map(|a| self.get_account_state(*a))
+            .collect()
+    }
```
MegaRedHand, line 27: I'm not sure I like this trait growing even more.

crates/vm/levm/src/db/mod.rs
```diff
@@ -182,13 +194,24 @@ impl Database for CachingDatabase {
 
     #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
```
MegaRedHand: This doesn't require rayon anymore, so we can remove it, right?

crates/vm/db.rs
edg-l: Both levm::Database and VmDatabase need it because the batched read is in StoreVmDatabase but the cache it fills (CachingDatabase) is a layer up; DynVmDatabase bridges the two. Both have a default-loop impl so no other backend changes. The only way to drop the trait method is downcasting via as_any, which adds a trait method anyway plus a runtime check and fallback. Can do that if you'd rather, but this seemed lighter.

ilitteri (approved):

iovoid (approved):

PR #6713, opened 2026-05-22
fix(l1): announce local IP when --p2p.addr is unspecified

edg-l:
## Summary

When ethrex is launched with `--p2p.addr=0.0.0.0` (or `::`) and no `--nat.extip`, the announce/enode/ENR address ended up as the bind address verbatim, i.e. `enode://…@0.0.0.0:30303`. Peers cannot dial back to that, so the node sits at 0 inbound peers. Spotted on bal-devnet-7 (the deployment passes `--p2p.addr=0.0.0.0` without `--nat.extip`).

This patch:

- Factors the address-resolution logic out of `get_local_p2p_node` into a small pure function `resolve_p2p_endpoints`, so it can be unit-tested.
- When `--p2p.addr` is given and resolves to an unspecified address, the announced address now falls back to the auto-detected local IP (the same `local_ip()` / `local_ipv6()` call the no-flag branch already used). `--nat.extip` still overrides; specific bind addresses are unchanged.
- If no local IP can be detected, emits a loud warning rather than silently advertising the unspecified address.

This is the minimal fix. A follow-up could implement devp2p endpoint prediction (learning the external IP from the `to` field on incoming PINGs/PONGs, à la geth's `UDPEndpointStatement`), which would remove the need for operators to set `--nat.extip` at all on cloud hosts. Out of scope for this PR.

## Test plan

- [x] `cargo fmt -p ethrex`
- [x] `cargo clippy -p ethrex --all-targets -- -D warnings`
- [x] `cargo test -p ethrex --lib initializers::` — 6 new unit tests pass
- [ ] Deploy onto a bal-devnet ethrex node with `--p2p.addr=0.0.0.0` and confirm `admin_nodeInfo` reports the host IP in the enode/ENR and inbound peers > 0.

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -360,27 +360,27 @@ pub fn get_signer(datadir: &Path) -> SecretKey {
     }
 }
 
-pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkConfig) {
-    let tcp_port = opts.p2p_port.parse().expect("Failed to parse p2p port");
-    let udp_port = opts
-        .discovery_port
-        .parse()
-        .expect("Failed to parse discovery port");
-
-    let local_public_key = public_key_from_signing_key(signer);
-
-    // Determine bind and external addresses.
-    //
-    // --nat.extip sets the address announced to peers (for nodes behind NAT).
-    // --p2p.addr sets the bind address (defaults to the auto-detected local IP
-    //   when --nat.extip is not given, or to the unspecified address when it is:
-    //   0.0.0.0 for IPv4, :: for IPv6).
-    let (bind_addr, external_addr): (IpAddr, IpAddr) = match (&opts.p2p_addr, &opts.nat_extip) {
+/// Decide the bind and externally-announced addresses for the P2P endpoint.
+///
+/// Precedence:
+/// - `--nat.extip` wins for the announced address; bind comes from `--p2p.addr` if given,
+///   else the unspecified address of the matching family.
+/// - `--p2p.addr` alone is used for both bind and announce, except when it's an unspecified
+///   address (`0.0.0.0` / `::`). In that case the announced address falls back to the
+///   auto-detected local IP of the matching family; this avoids advertising `0.0.0.0` in
+///   the ENR, which would make the node unreachable for inbound connections. Operators
+///   behind NAT still need `--nat.extip` for that case to resolve correctly.
+/// - With neither flag set, the auto-detected local IP is used for both bind and announce.
+fn resolve_p2p_endpoints(
+    p2p_addr: Option<&str>,
+    nat_extip: Option<&str>,
+    local_v4: Option<IpAddr>,
+    local_v6: Option<IpAddr>,
+) -> (IpAddr, IpAddr) {
+    match (p2p_addr, nat_extip) {
         (_, Some(extip)) => {
             let external: IpAddr = extip.parse().expect("Failed to parse --nat.extip address");
-            let bind: IpAddr = opts
-                .p2p_addr
-                .as_deref()
+            let bind: IpAddr = p2p_addr
                 .map(|a| {
                     let addr: IpAddr = a.parse().expect("Failed to parse p2p address");
                     assert!(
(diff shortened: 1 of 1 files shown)


ilitteri (approved):

cmd/ethrex/initializers.rs
```diff
@@ -399,16 +399,60 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
...
         }
         (Some(addr), None) => {
-            let ip: IpAddr = addr.parse().expect("Failed to parse p2p address");
-            (ip, ip)
+            let bind: IpAddr = addr.parse().expect("Failed to parse p2p address");
+            if bind.is_unspecified() {
+                let external = if bind.is_ipv6() {
+                    local_v6.or(local_v4)
+                } else {
+                    local_v4.or(local_v6)
```
ElFantasma: **Cross-family fallback may advertise an unreachable address.** When `bind` is `0.0.0.0` and `local_v4` is `None` but `local_v6` is `Some`, this falls back to the IPv6 address — but the v4 socket can't accept inbound v6 connections, so peers reaching the ENR endpoint will fail to connect. Same in the v6 branch (line 405) when only `local_v4` is available.

In practice this only matters on single-stack hosts where the *other* stack's local-IP probe somehow succeeds (rare), so it's not a regression vs. `main`. But the warn-then-fallback path below might be the better default when only the wrong-family IP is found: keep `bind` as the announced address and emit the warning, since at least operators won't think peering is healthy when it isn't. Consider:
```rust
let external = if bind.is_ipv6() { local_v6 } else { local_v4 };
```
(drop the cross-family `.or(...)`)

Non-blocking; corner of a corner. Flagging because the test matrix doesn't cover the v6 path of this branch either — see body.

cmd/ethrex/initializers.rs
```diff
@@ -399,16 +399,60 @@ pub fn get_local_p2p_node(opts: &Options, signer: &SecretKey) -> (Node, NetworkC
...
+                            "--p2p.addr is unspecified; announcing auto-detected local IP. Set --nat.extip to override."
+                        );
+                        (bind, ext)
+                    }
+                    None => {
+                        warn!(
+                            bind = %bind,
+                            "--p2p.addr is unspecified and no local IP could be detected; \
+                             announcing the unspecified address. Inbound peer connections will fail. \
+                             Set --nat.extip to fix this."
```
ElFantasma: nit: the message says "Set --nat.extip to fix this" but the branch is only reached when `local_ip()` and `local_ipv6()` both failed — which usually means the host genuinely has no routable IP detectable. An operator hitting this path doesn't necessarily know what extip to pass either. `--p2p.addr=<explicit-ip>` is the more general alternative.

Suggest:
```rust
"--p2p.addr is unspecified and no local IP could be detected; \
 announcing the unspecified address. Inbound peer connections will fail. \
 Set --nat.extip or use --p2p.addr=<ip> to fix."
```
Low priority — just operator UX.

cmd/ethrex/initializers.rs
```diff
@@ -812,3 +856,63 @@ pub async fn regenerate_head_state(
...
+        assert_eq!(bind, local);
+        assert_eq!(ext, local);
+    }
+
+    #[test]
+    fn extip_only_uses_unspecified_bind() {
+        let (bind, ext) = resolve_p2p_endpoints(None, Some("203.0.113.5"), None, None);
+        assert_eq!(bind, ip("0.0.0.0"));
+        assert_eq!(ext, ip("203.0.113.5"));
+    }
```
ElFantasma, line 916: Worth adding a test for the IPv6 unspecified path to mirror `p2p_addr_unspecified_v4_announces_local_ip`:

```rust
#[test]
fn p2p_addr_unspecified_v6_announces_local_ipv6() {
    let local6 = ip("fe80::1");
    let (bind, ext) = resolve_p2p_endpoints(Some("::"), None, None, Some(local6));
    assert_eq!(bind, ip("::"));
    assert_eq!(ext, local6);
}
```

The v6 arm of the `if bind.is_ipv6() { local_v6.or(local_v4) }` switch is currently uncovered. Also missing: a `#[should_panic]` test for the `--p2p.addr` / `--nat.extip` family-mismatch assertion. Both are cheap.

ElFantasma (approved):

cmd/ethrex/initializers.rs
edg-l: Dropped the cross-family .or() in 3e871a9417; v4 bind only falls back to local_v4, v6 to local_v6, otherwise warn-and-keep-unspecified.

cmd/ethrex/initializers.rs
edg-l: Updated in 3e871a9417 to suggest both --nat.extip=<ip> and --p2p.addr=<ip>.

cmd/ethrex/initializers.rs
edg-l: Added p2p_addr_unspecified_v6_announces_local_ipv6 and family_mismatch_panics in 3e871a9417 (8/8 pass).

azteca1998 (approved):

Issue #6688, opened 2026-05-20
perf(l1): `eth_getTransactionByHash` does unbounded RocksDB scan, makes node vulnerable to RPC retry storms


ElFantasma:
# `eth_getTransactionByHash` performs unbounded RocksDB scan, makes node vulnerable to RPC retry storms

## Summary

`Store::get_transaction_location` ([`crates/storage/store.rs:596-602`](https://github.com/lambdaclass/ethrex/blob/main/crates/storage/store.rs#L596-L602)) iterates a RocksDB prefix iterator but **never breaks when the prefix stops matching**. On a fully-synced mainnet node, every single `eth_getTransactionByHash` call therefore scans from its seek point to the end of the `TRANSACTION_LOCATIONS` column family — hundreds of GB of compressed SSTs — taking seconds to minutes per call regardless of whether the transaction exists.

Verified live on mainnet-2 (idle): 9.57 s for a hash near the start of the keyspace; 0.0003 s for a hash near the end. Linear relationship between hash position and scan time.

This caused a real production incident on mainnet-1 on 2026-05-20: a single client opening ~25k connections triggered an avg latency of 6+ hours on `eth_getTransactionByHash` and saturated the Tokio blocking pool to the point that local Prometheus exporters also timed out.

## Root cause

`prefix_iterator_cf` in `rust-rocksdb` seeks to the prefix but iterates forward indefinitely — the caller must break manually when the key no longer matches. The current loop:

```rust
// crates/storage/store.rs:596-602
while let Some(Ok((key, value))) = iter.next() {
    if key.len() == 64 && &key[0..32] == tx_hash_bytes {
        transaction_locations.push(<(BlockNumber, BlockHash, Index)>::decode(&value)?);
    }
    // ← missing: else { break; }
}
```

Once a non-matching key is reached, all subsequent keys also won't match (RocksDB keys are stored in lexicographic order, so all entries with the same `tx_hash` prefix are contiguous). The `if` filter is correct but doesn't stop iteration; the loop continues reading to the end of the column family.

`TRANSACTION_LOCATIONS` on mainnet is ~3.75 billion entries (≈25M blocks × ~150 txs/block) at ~64-byte keys + ~70-byte values, compressed to ~150-200 GB on disk. Worst-case scan: the entire column family.

[...]

PR #6718, opened 2026-05-22
refactor(l1): key TRANSACTION_LOCATIONS by tx hash

### LINKED ISSUE #6708: refactor(l1,storage): simplify TRANSACTION_LOCATIONS schema to key by tx_hash only (follow-up to #6688)
## Motivation

[#6688](https://github.com/lambdaclass/ethrex/issues/6688) was caused by an unbounded prefix-iterator scan in `get_transaction_location`. The minimal fix landed in [#6702](https://github.com/lambdaclass/ethrex/pull/6702): a `break` on prefix mismatch. After that fix, `eth_getTransactionByHash` is sub-millisecond.

The root reason the prefix iterator existed at all is that the `TRANSACTION_LOCATIONS` table uses a *composite key* of `tx_hash + block_hash` to model a one-to-many relationship (one tx hash → potentially multiple block hashes during reorgs). The original author flagged this at [`store.rs:552`](https://github.com/lambdaclass/ethrex/blob/main/crates/storage/store.rs#L552) with `// FIXME: Use dupsort table` — i.e., a `key → Vec<value>` shape instead of `(key, suffix) → value`. This issue proposes following through on that.

## Access pattern audit

Comprehensive scan of `crates/` for `TRANSACTION_LOCATIONS` consumers:

| Caller | Purpose | Direction |
|---|---|---|
| `eth_getTransactionByHash` (`rpc/eth/transaction.rs:245`) | tx hash → location | tx_hash → canonical (block, index) |
| `eth_getTransactionReceipt` (`rpc/eth/transaction.rs:296`) | tx hash → receipt lookup | same |
| L2 RPC (`l2/networking/rpc/l2/messages.rs:46`) | same | same |
| `Store::get_transaction_by_hash` (`store.rs:854`) | internal wrapper | same |
| `debug_trace*` (`blockchain/tracing.rs:30, 100, 160, 228`) | tracer needs the (block, index) for a hash | same |

**Every read does the same query: tx_hash → one canonical `(BlockNumber, BlockHash, Index)`.** No caller iterates entries by `block_hash`, asks for "all txs in block X" via this table, or uses any other access pattern.

"Txs in block X" lookups go through the `BODIES` table (which stores the full ordered tx list per block — see `Store::get_transaction_by_location` at `store.rs:867`). The 32-byte `block_hash` suffix on the current composite key is purely a disambiguator for the rare reorg case where the same tx appears in multiple blocks; it's never iterated.

Write surface:
- `add_blocks` (`store.rs:307`) — bulk import path
- `store_block_updates` (`store.rs:1442`) — block execution pipeline
- `add_transaction_location` / `add_transaction_locations` (`store.rs:545, 563`) — helpers (currently no in-repo callers)

Deletes: **none**. Table is append-only; reorg'd entries stay forever and the read path filters by canonical chain.

## Proposed schema

```
Key:   tx_hash (32 bytes)
Value: Vec<(BlockNumber, BlockHash, Index)>  // RLP-encoded
```

In the common case (no reorgs), the value is a 1-element list. During/after a reorg, entries from orphaned blocks accumulate (same as today).

### Read path

```rust
pub async fn get_transaction_location(&self, transaction_hash: H256)
    -> Result<Option<(BlockNumber, BlockHash, Index)>, StoreError>
{
        let tx = db.begin_read()?;
        let Some(bytes) = tx.get(TRANSACTION_LOCATIONS, transaction_hash.as_bytes())? else {
            return Ok(None);
        };
        let locations: Vec<(BlockNumber, BlockHash, Index)> = Vec::decode(&bytes)?;
        // Canonical filter — same logic as today.
        for (block_number, block_hash, index) in locations {
            let canonical_hash = tx.get(CANONICAL_BLOCK_HASHES, block_number.to_le_bytes().as_slice())?
                .map(|b| H256::decode(b.as_slice())).transpose()?;
            if canonical_hash == Some(block_hash) {
                return Ok(Some((block_number, block_hash, index)));
            }
        }
        Ok(None)
    }).await.map_err(|e| StoreError::Custom(format!("Task panicked: {e}")))?
}
```

Single `get_cf` instead of any iteration. `O(1)` regardless of table size, no dependency on prefix iterator hygiene.

### Write path

Each `put(tx_hash, ...)` becomes read-modify-write: load existing list (if any), append, write back. For bulk paths (`add_blocks`, `store_block_updates`) which import a whole block at a time, the typical patterns:

1. Within a block-import RocksDB transaction, batch all reads first, then all writes.
2. In the common case (no prior entry for any tx in the block), all reads return `None` and the writes are 1-element lists.

The RMW cost is one extra read per tx-write, balanced against eliminating the iterator setup + per-call O(matches+1) scan + per-call `get(CANONICAL_BLOCK_HASHES, ...)` lookup on every read query forever after.

## Secondary benefits

- **Smaller per-entry storage**: drops the redundant 32-byte tx_hash from the key when there are duplicates (rare). For the common single-entry case: 32-byte key + ~70-byte value = 102 bytes, vs current 64-byte key + ~70-byte value = 134 bytes. ~24% smaller.
- **Bloom-filter friendliness**: a fixed-32-byte key works with RocksDB's standard block-based bloom filter without needing a prefix extractor. Negative lookups (tx not present) get O(1) bloom-filter rejection instead of any SST file access.
- **No reliance on iterator hygiene**: future call sites won't be able to introduce a regression of [#6688](https://github.com/lambdaclass/ethrex/issues/6688) on this table.

## Out of scope (separate concerns)

- **The same iterator-hygiene gotcha elsewhere**: `prefix_iterator` callers other than the (already-fixed) `get_transaction_location` need their own audits. The other current call site (`EXECUTION_WITNESSES` cleanup at `store.rs:2034`) already breaks correctly.
- **Defensive API**: a typed bounded-iterator API at the storage trait layer (e.g., `prefix_iterator` that automatically calls `set_iterate_upper_bound`) is a separate good idea. The schema redesign here makes the existing `prefix_iterator` usage on `TRANSACTION_LOCATIONS` simply unnecessary, but doesn't preclude or replace the defensive-API work.

## Migration

Use the DB migration framework added in [#6519](https://github.com/lambdaclass/ethrex/pull/6519):

1. Bump `STORE_SCHEMA_VERSION` in `crates/storage/lib.rs` (currently `1` → `2`).
2. Add `fn migrate_1_to_2(backend: &dyn StorageBackend) -> Result<(), StoreError>` to `crates/storage/migrations.rs`.
3. Append it to the `MIGRATIONS` array (index 0). The compile-time assert in that file enforces array length matches the schema version.
4. On startup, `Store::new` will detect on-disk schema v1 and run the migration before opening for normal use, with fsync-per-step crash safety.

The migration function itself: iterate the existing `TRANSACTION_LOCATIONS` CF, group entries by tx_hash into `Vec<(BlockNumber, BlockHash, Index)>`, write back under the new schema (either same CF with destroy+recreate, or a fresh CF with an atomic swap once complete), then drop the old data.

Sizing: `TRANSACTION_LOCATIONS` is ~150–200 GB compressed on a fully-synced mainnet node. The migration is a one-time full-table rewrite, so a synced node will be unavailable for the duration of the migration on first restart after upgrade — this should be flagged in release notes. No resync required.

## Estimated scope

- Storage layer changes (encoding, 1 read + 4 writes): ~150 lines of diff
- Migration function + registration: ~50–100 lines (most of the lift is in correctness/testing of the rewrite logic, not framework wiring)
- No caller of `get_transaction_location` needs to change — the function's signature stays identical
- Tests: unit tests for the encoding round-trip + multi-block / reorg integration coverage + a migration unit test that builds a small old-schema DB and verifies post-migration reads

## References

- [#6688](https://github.com/lambdaclass/ethrex/issues/6688) — the underlying perf issue
- [#6702](https://github.com/lambdaclass/ethrex/pull/6702) — the `break` fix that's already merged
- [#6519](https://github.com/lambdaclass/ethrex/pull/6519) — the migration framework to use
- Existing FIXME at [`crates/storage/store.rs:552`](https://github.com/lambdaclass/ethrex/blob/main/crates/storage/store.rs#L552)

ElFantasma:
**Motivation**

Follow-up to [#6688](https://github.com/lambdaclass/ethrex/issues/6688). The minimal `break`-on-prefix-mismatch fix in [#6702](https://github.com/lambdaclass/ethrex/pull/6702) made `eth_getTransactionByHash` sub-millisecond again, but the underlying schema (composite key `tx_hash || block_hash` over a table that holds a one-to-many relationship) is the reason a prefix iterator was needed in the first place. The original author left a `// FIXME: Use dupsort table` on the write helper. This PR follows through on that.

Reads now do a single `get_cf` on a fixed 32-byte key — `O(1)`, no iteration, no dependency on bounded-iterator hygiene. Future call sites can't accidentally reintroduce the unbounded-scan footgun from #6688 on this table.

**Description**

Schema change for the `TRANSACTION_LOCATIONS` column family (v2 → v3, on top of the receipts migration from [#6598](https://github.com/lambdaclass/ethrex/pull/6598)):

| | Before (v2) | After (v3) |
|---|---|---|
| Key | `tx_hash ‖ block_hash` (64 bytes) | `tx_hash` (32 bytes) |
| Value | `(BlockNumber, BlockHash, Index)` | `Vec<(BlockNumber, BlockHash, Index)>` |
| Read path | `prefix_iterator` + per-entry filter | single `get` + decode |

The value is a `Vec` because, in the rare reorg case, the same tx hash may appear in multiple blocks. The canonical-chain filter (the second half of `get_transaction_location`) is unchanged.

**Write paths.** The four write sites (`add_blocks`, `store_block_updates`, `add_transaction_location`, `add_transaction_locations`) now go through a small `stage_tx_location` helper that does read-modify-write into a `HashMap<H256, Vec<location>>`. Within a batch, re-touching the same tx hash replaces any prior entry with the same `block_hash` — preserving the dedupe semantics that the old composite key gave us for free. The codebase already implicitly serializes writers to this table (the blockchain pipeline is sequential), and the new RMW preserves that assumption; it doesn't introduce a new concurrency requirement.

**Migration.** A new `migrate_2_to_3` is registered as the second entry in the migration framework's `MIGRATIONS` array (the receipts migration from #6598 stays as entry 0, v1→v2). It streams the v2 table in lex order (composite keys with the same 32-byte prefix are adjacent), groups them by tx hash, and flushes each group as an atomic `put (32-byte key) + delete (64-byte keys)` batch. Commits are chunked (50k groups per commit) to bound memory. Per-tx atomicity means crash-resume sees either a fully-migrated tx hash (32-byte key present, 64-byte keys absent) or a still-pending one (32-byte key absent, 64-byte keys present) — never a mix — so resuming after a crash just re-runs cleanly.

`TRANSACTION_LOCATIONS` is ~150–200 GB on a fully-synced mainnet node; first restart after upgrade will be unavailable for the duration of the migration. This should be flagged in release notes.

**Secondary benefits.**
- 32-byte fixed keys work cleanly with RocksDB's standard block-based bloom filter — negative lookups (tx not present) get O(1) bloom-filter rejection. Adding the filter is out of scope for this PR but is now an easy follow-up.
- ~24% smaller per-entry storage in the common single-block case.

**Tests.** Four new migration unit tests (empty table, single-entry-per-hash, multi-block-per-hash for reorg, idempotent partial-restart). All existing storage and RPC tests pass.

Closes #6708

**Checklist**

### DIFF
--- a/crates/storage/api/tables.rs
+++ b/crates/storage/api/tables.rs
@@ -43,13 +43,12 @@ pub const RECEIPTS: &str = "receipts";
 pub const RECEIPTS_V2: &str = "receipts_v2";
 
 /// Transaction locations column family: [`Vec<u8>`] => [`Vec<u8>`]
-/// - [`Vec<u8>`] = Composite key
-///    ```rust,no_run
-///     // let mut composite_key = Vec::with_capacity(64);
-///     // composite_key.extend_from_slice(transaction_hash.as_bytes());
-///     // composite_key.extend_from_slice(block_hash.as_bytes());
-///    ```
-/// - [`Vec<u8>`] = `(block_number, block_hash, index).encode_to_vec()`
+/// - Key: `transaction_hash.as_bytes()` (32 bytes)
+/// - Value: `Vec<(block_number, block_hash, index)>.encode_to_vec()`
+///
+/// The value is a list because, in the rare case of a reorg, the same
+/// transaction may appear in multiple blocks. Readers must filter by the
+/// canonical chain to pick the right `(block_number, block_hash, index)`.
 pub const TRANSACTION_LOCATIONS: &str = "transaction_locations";
 
 /// Chain data column family: [`Vec<u8>`] => [`Vec<u8>`]
--- a/crates/storage/lib.rs
+++ b/crates/storage/lib.rs
@@ -85,7 +85,7 @@ pub use store::{
 /// When bumping this version, add a corresponding migration function to
 /// `migrations::MIGRATIONS`. The migration framework will automatically
 /// upgrade existing databases instead of requiring a full resync.
-pub const STORE_SCHEMA_VERSION: u64 = 2;
+pub const STORE_SCHEMA_VERSION: u64 = 3;
 
 /// Name of the file storing the metadata about the database.
 ///
--- a/crates/storage/migrations.rs
+++ b/crates/storage/migrations.rs
@@ -1,14 +1,16 @@
 use std::io::Write;
 use std::path::Path;
 
-use crate::api::StorageBackend;
-use crate::api::tables::{RECEIPTS, RECEIPTS_V2};
+use ethrex_common::H256;
+use ethrex_common::types::{BlockHash, BlockNumber, Index};
+use ethrex_rlp::decode::RLPDecode;
+use ethrex_rlp::encode::RLPEncode;
+
+use crate::api::tables::{RECEIPTS, RECEIPTS_V2, TRANSACTION_LOCATIONS};
+use crate::api::{StorageBackend, StorageWriteBatch};
 use crate::error::StoreError;
 use crate::store::receipt_key;
 use crate::{STORE_METADATA_FILENAME, STORE_SCHEMA_VERSION};
-
-use ethrex_common::H256;
-use ethrex_rlp::decode::RLPDecode;
 
 use super::store::StoreMetadata;
 
@@ -27,7 +29,7 @@ pub type MigrationFn = fn(backend: &dyn StorageBackend) -> Result<(), StoreError
 ///
 /// **Invariant**: `MIGRATIONS.len() == (STORE_SCHEMA_VERSION - 1) as usize`
 /// (empty when `STORE_SCHEMA_VERSION == 1`, one entry when it's 2, etc.)
-pub const MIGRATIONS: &[MigrationFn] = &[migrate_1_to_2];
+pub const MIGRATIONS: &[MigrationFn] = &[migrate_1_to_2, migrate_2_to_3];
 
 // Compile-time check: the number of migration functions must match the number
 // of version gaps (i.e. STORE_SCHEMA_VERSION - 1).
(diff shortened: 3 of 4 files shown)


ElFantasma:
Re Greptile's "RMW is now a correctness requirement" P2:

Correct call — under the old composite-key schema, two concurrent writers on the same `tx_hash` (e.g. competing reorg blocks) wrote to different keys (`tx_hash||hash_a` vs `tx_hash||hash_b`) so both survived. The new schema can't do that; the read snapshot and write batch are separate operations.

The serialized-writers assumption holds in ethrex today because the blockchain pipeline (fork-choice + block import) runs one block at a time, and the four affected write paths (`add_blocks`, `store_block_updates`, `add_transaction_location[s]`) all live inside that pipeline. I've added an explicit "Concurrency invariant" section to the `stage_tx_location` doc comment in commit `f361dc064` so this isn't a surprise to future contributors and so the right place to add a lock is obvious if the assumption ever breaks.

A defensive lock could be added now, but it'd be a no-op under current call patterns and would cost a cross-thread sync on the hottest tx-import path. Leaving the doc-comment-only guard for now — happy to revisit if you'd prefer the explicit lock.

edg-l (approved):
Minor (non-blocking):

https://github.com/lambdaclass/ethrex/blob/a631c69b36b635bf6b12a54732e84d324d9e7ca5/crates/storage/migrations.rs#L216-L219

After the mid-loop `write_batch.commit()?`, the batch is reused without `begin_write()`. Works today (RocksDB impl uses `mem::take`, InMemory commit is a no-op), but it leans on an undocumented post-commit reuse contract on `StorageWriteBatch`. Either re-acquire with `write_batch = backend.begin_write()?;` after the commit, or document the reuse semantics on the trait.

Nits:

https://github.com/lambdaclass/ethrex/blob/a631c69b36b635bf6b12a54732e84d324d9e7ca5/crates/storage/migrations.rs#L226-L231

`groups_in_batch` isn't reset after the final flush; no functional impact, just asymmetric.

https://github.com/lambdaclass/ethrex/blob/a631c69b36b635bf6b12a54732e84d324d9e7ca5/crates/storage/store.rs#L272-L291

Concurrency comment on `stage_tx_location` is correct; consider adding a tracking-issue ref so it isn't lost if the pipeline ever goes parallel.

ElFantasma:
Thanks for the review! Addressed in `02d053569`:

- **`write_batch` reuse** → re-acquire with `write_batch = backend.begin_write()?;` after each mid-loop `commit()`, with an inline comment explaining why (post-commit reuse isn't a documented contract on `StorageWriteBatch`).
- **`groups_in_batch` asymmetric on final flush** → documented inline why we don't touch it after the post-loop flush (commit immediately follows, variable goes out of scope). Cleaner than adding a dead write/reset.
- **Tracking-issue ref on concurrency comment** → filed [#6727](https://github.com/lambdaclass/ethrex/issues/6727) and the `stage_tx_location` doc-comment now links to it.

ElFantasma:
Superseded by #6737. Same goal (key `TRANSACTION_LOCATIONS` by tx hash, single-`get_cf` reads), but this RMW approach added a per-tx `get_cf` on the block-import write path that measured a 2–8× `store_ms` regression on mainnet-10. A bloom filter only partially mitigated it (irreducible memtable-traversal floor + lagging deep-level coverage). #6737 uses a RocksDB merge operator instead, removing the per-tx read entirely — write cost matches the current schema (~2.5 ms, stationary) while keeping the read-path cleanup. Closing in favor of #6737.

PR #6720, opened 2026-05-22
[WIP] Add Gnosis support

iovoid:
**Motivation**

We want to be able to execute blocks in the Gnosis network.

**Description**

Gnosis is very similar to ethereum mainnet, and so the differences are rather small.

To test, start a consensus node as explained on the gnosis [documentation](https://docs.gnosischain.com/node/#client-setup), and start ethrex setting the network to 'gnosis'.

```
cargo run --release -- --network gnosis --authrpc.jwtsecret /path/to/jwtsecret
```


yorickdowne:
Note that Gnosis will become a based L2 in the EEZ, likely early 2027. Support should probably target that, not Gnosis as an Ethereum code L1.

PR #6721, opened 2026-05-22
fix(l1): statetest CLI follow-ups (optional _info, Osaka, BLOCKHASH convention)

ElFantasma:
**Motivation**

Follow-ups to [#6663](https://github.com/lambdaclass/ethrex/pull/6663) (statetest CLI for goevmlab fuzzing), addressing three issues raised in Slack feedback after the PR landed:

1. `_info` is required in the statetest fixture deserializer, but goevmlab-generated fixtures don't necessarily include it.
2. `default_forks` (the allow-list of forks the runner will execute) is missing Osaka, so any fixture with an Osaka post-state was silently skipped or surfaced as an error.
3. `BLOCKHASH(0)` returns ethrex's synthesized genesis hash rather than the EF state-test convention `keccak256(decimal_string(n))` that geth uses. This trips the goevmlab fuzzer on the very first block-hash lookup.

**Description**

Two commits, reviewable independently:

**Commit 1 — `fix(l1): make statetest _info optional and add Osaka to runnable forks`**

- `Test._info` becomes `Option<Info>`; `build_test` no longer errors when `_info` is absent.
- Both readers (`runner::run_tests` for EIP filtering and `report::generate_report` for the description/reference lines) are updated to handle `None` with sensible defaults — empty `reference_spec`, "No description or comment" for the report.
- `DEFAULT_FORKS` gets `"Osaka"` inserted between Prague and Amsterdam, matching the fork ordering used elsewhere.
- Two unit tests pin the new contract: fixture-without-`_info` parses, fixture-with-`_info` still parses.

**Commit 2 — `fix(l1): apply EF state-test BLOCKHASH convention in statetest harness`**

The bug: at block N=1, the test contract `0x60004060005500` (`PUSH1 0 / BLOCKHASH / PUSH1 0 / SSTORE / STOP`) calls `BLOCKHASH(0)`. ethrex returned the genesis hash it derived from the test's pre-state (`0x86a0…`); geth (and the EF spec) returns `keccak256("0")` = `0x0448…116d`. The state root differs from the very first SSTORE, which kills any differential fuzzing.

Fix: in `load_initial_state` (shared by the statetest CLI and `runner.rs`'s EF state-test runner), wrap the levm `Database` in a thin `StatetestDatabase` shim that delegates everything except `get_block_hash`, which returns `keccak256(decimal_string(n))`. Matches geth's `vmTestBlockHash` in `tests/state_test_util.go`. Three unit tests pin the contract (n=0, n=1, n=10 to distinguish decimal vs hex encoding).

`block_runner.rs` has a separate pre-existing problem: its phase-3 real-block re-execution path uses the actual genesis hash because it goes through `add_block_pipeline`, not `load_initial_state`. Out of scope here — orthogonal code path, separate fix.

**Checklist**

### DIFF
--- a/tooling/ef_tests/state_v2/src/modules/block_runner.rs
+++ b/tooling/ef_tests/state_v2/src/modules/block_runner.rs
@@ -23,7 +23,26 @@ use crate::modules::{
 };
 
 pub async fn run_tests(tests: Vec<Test>) -> Result<(), RunnerError> {
+    // Fusaka EIPs that block-mode supports; mirrors the allowlist in runner.rs.
+    // TODO: drop once all Fusaka EIPs land.
+    let fusaka_eips_to_test: Vec<&str> =
+        vec!["eip-7594", "eip-7939", "eip-7918", "eip-7892", "eip-7883"];
+
     for test in &tests {
+        // Apply the same gating runner.rs uses so we don't unconditionally run
+        // every Osaka fixture in block mode. Fixtures without `_info` (e.g.
+        // goevmlab-generated) bypass the filter — we can't read the EIP list,
+        // so silently dropping them would be wrong.
+        if test.path.to_str().unwrap().contains("osaka")
+            && let Some(spec) = test
+                ._info
+                .as_ref()
+                .and_then(|info| info.reference_spec.as_deref())
+            && !fusaka_eips_to_test.iter().any(|eip| spec.contains(eip))
+        {
+            continue;
+        }
+
         println!("Running test group: {}", test.name);
         for test_case in &test.test_cases {
             let res = run_test(test, test_case).await;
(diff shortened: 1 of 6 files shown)


ElFantasma:
Re Codex finding 3 (parse `previousHash` from the fixture):

The hardcoded `keccak256(decimal_string(n))` formula is intentional. It matches geth's `vmTestBlockHash` in `tests/state_test_util.go` exactly, and the EF state-test spec defines the formula — not the fixture field — as the source of truth.

`env.previousHash` is informational: it's always equal to `keccak256(decimal_string(currentNumber - 1))` by construction, so reading it would be a no-op for any fixture that follows the spec. If a fixture ever provided a `previousHash` that *didn't* match the formula, geth would still return the formula's value, so honoring `previousHash` instead would *diverge from geth* — the opposite of what this PR is trying to fix.

So: not parsing `previousHash` is the deliberate behavior here.

ilitteri (approved):

edg-l (approved):

azteca1998 (approved):

PR #6722, opened 2026-05-22
feat(l1): add LambdaVM zkVM backend for the L1 prover

ilitteri:
**Motivation**

LambdaVM (`yetanotherco/lambda_vm`) was open-sourced as a verifiable RV64IM zkVM with a transparent STARK-over-Goldilocks proof system, 128-bit security, and LogUp lookups. Bringing it into ethrex as a first-class zkVM backend closes the loop on a vertically integrated stack and gives the LambdaVM team a real EVM workload to harden against.

A proof-of-concept guest already exists in the lambda_vm repo at `executor/programs/rust/ethrex/`, but it was written against an older revision of the guest-program API (before the `Crypto` trait was extracted) and won't compile against current ethrex `main`. This PR brings the integration into ethrex itself, alongside the existing SP1, RISC0, OpenVM, and Zisk backends.

**Description**

Wires LambdaVM into the existing zkVM backend framework, scoped to L1 local proving (mirrors the current Zisk stance — no L2 prover wiring, no on-chain verifier).

**Guest side**

- `crates/guest-program/bin/lambdavm/` — detached workspace, custom RV64IM target spec (`riscv64im-lambda-vm-elf`), pinned `nightly-2026-02-01` toolchain with `rust-src`, thin `main.rs` mirroring the Zisk wrapper.
- `crates/guest-program/src/crypto/lambdavm.rs` — `LambdaVmCrypto` is the LambdaVM team's own provider implementation, kept byte-for-byte in sync with `crypto/ethrex-crypto` in `yetanotherco/lambda_vm` (only the struct name and module docs differ; when changing it, change both). It overrides `keccak256` (a sponge over the `keccak_permute` syscall) and `secp256k1_ecrecover` (the recovery's 2-term linear combination evaluated through the ECSM `ecsm_mul` syscall, reconstructing the full point from x-only queries, with a pure-Rust `lincomb` fallback for host/degenerate inputs), and routes BLS12-381 (EIP-2537) via the portable pure-Rust `bls12_381` backend (the trait defaults require `blst`, which guest builds compile out — same shape as the SP1/RISC0/OpenVM adapters). `recover_signer` is deliberately not overridden: the trait default routes it through the accelerated `secp256k1_ecrecover` after the EIP-2 high-s check, so all sender recovery is ECSM-accelerated too. Every other `Crypto` method inherits the trait default (ark-bn254, malachite, p256). KZG point-evaluation is unsupported in the guest: `kzg-rs` pulls in SP1-specific symbols that don't link for the RV64IM target. The provider's host-side unit tests (keccak sponge KATs + differential vs the trusted `keccak` crate, ecrecover known-answer vectors, x-only ECSM reconstruction vs software `lincomb`) are ported under `src/crypto/lambdavm/tests/`.
- `lambda-vm-syscalls` is target-gated to `cfg(target_arch = "riscv64")` (mirroring the upstream crate layout): host builds with the `lambdavm` feature never pull the riscv-only crates, and every use in the provider sits under the same cfg. The guest `[profile.release]` uses thin LTO, matching the upstream guest profile exactly (upstream measured ~2.3% fewer executed cycles vs cargo's release default; the target JSON already pins `panic-strategy: abort`).
- `build.rs` builds the guest ELF under the `lambdavm-build-elf` feature. Sysroot path resolved from `LAMBDA_VM_SYSROOT_DIR` (default `/opt/lambda-vm-sysroot`); inner toolchain pinned via `RUSTC=$(rustc_path("nightly-2026-02-01"))` with `CARGO` / `RUSTUP_TOOLCHAIN` / `RUSTFLAGS` / `CARGO_ENCODED_RUSTFLAGS` cleared so the outer cargo's env doesn't force the inner build onto stable.

**Host side**

- `crates/prover/src/backend/lambdavm.rs` — `LambdaVmBackend` shells out to LambdaVM's CLI binary (resolved as `lambda-vm-cli` on `$PATH`, overridable via `LAMBDA_VM_CLI`) for `execute`, `prove`, `verify`. Inputs serialized as `[4-byte LE length][rkyv bytes]` matching LambdaVM's `PRIVATE_INPUT_START` memory layout. `prove_timed` parses `Proving time: <float>s` from `prove --time` stdout, falls back to wall-clock if missing. `ProofFormat::Compressed` yields the native STARK proof; `Groth16` returns `NotImplemented` (no L1-cheap wrapper yet).
- `BackendType::LambdaVM` registered, `FromStr` arm added, dispatch site in `start_prover` wired.

**CI / release plumbing**

- `.github/actions/install-lambdavm/` — composite action: installs `nightly-2026-02-01` + `rust-src`, downloads the sysroot tarball (SHA256-verified against the pinned hash from the lambda_vm Makefile, cached under `$HOME` with the hash in the cache key), builds the CLI from the pinned LambdaVM commit (cached), installs it on `$PATH` as `lambda-vm-cli`.
- `.github/workflows/pr-main_l2_prover.yaml` — `lambdavm` added to the `lint_zk` matrix, conditional install step added.
- Root `Makefile` — `bin/lambdavm/Cargo.toml` added to `update-cargo-lock` and `check-cargo-lock` (metadata-only check, toolchain-free pattern matching ZisK).
- `docs/developers/release-process.md` — `bin/lambdavm/Cargo.toml` in the version-bump list, `bin/lambdavm/Cargo.lock` in the expected-changes list.
- `crates/guest-program/Makefile` — `lambdavm` / `l2-lambdavm` build targets, help text, `clean` rule extended.

**Docs**

- `docs/zkvm-integrations.md` — LambdaVM section with status, key features, integration details, limitations, build/run instructions.
- `crates/guest-program/README.md` — "Integrating a New zkVM" guide expanded from 6 to 11 steps, covering the missing host backend, CI install action, root-Makefile cargo-lock plumbing, release-process doc, and zkvm-integrations.md updates that future zkVM vendors will also need.

**Source pinning**

LambdaVM is pinned by commit hash (`9ccdaf28ab373da0a4a2eb79a16b90479f6686fd`) from both `crates/guest-program/Cargo.toml` (the `lambda-vm-syscalls` git dep) and `.github/actions/install-lambdavm/action.yml` (the `cli` checkout target). Bumping the version is a single-commit operation touching exactly those two locations; inline comments cross-reference them.

**Limitations**

Only `keccak_permute` and `ecsm_mul` (keccak + ECDSA recovery) are accelerated today. BN254, BLS12-381, sha256, and modexp run via pure-Rust crates, and KZG point-evaluation (precompile 0x0a) is unsupported in the guest — real EVM block proving will be substantially slower than SP1/RISC0 until LambdaVM lands more precompiles. Acknowledged in `docs/zkvm-integrations.md`.

**How to Test**

Local build verified end-to-end on macOS arm64:

```bash
# One-time sysroot setup (also handled by .github/actions/install-lambdavm in CI)
mkdir -p $HOME/.lambda-vm-sysroot
curl -sL https://lambda.alignedlayer.com/lambda-vm-sysroot-rv64im.tar.gz \
  | tar -xz -C $HOME/.lambda-vm-sysroot --strip-components=1

# Build guest ELF (`brew install llvm` for RISC-V clang)
LAMBDA_VM_SYSROOT_DIR=$HOME/.lambda-vm-sysroot \
  PATH=/opt/homebrew/opt/llvm/bin:$PATH \
  cargo check -p ethrex-guest-program --features lambdavm-build-elf

# Verify ELF type
file crates/guest-program/bin/lambdavm/out/riscv64im-lambda-vm-elf
# => ELF 64-bit LSB executable, UCB RISC-V, soft-float ABI, lp64,
#    statically linked, not stripped

# Build host backend (embeds the ELF via include_bytes!)
cargo check -p ethrex-prover --features lambdavm
```

Regression: `cargo check -p ethrex-prover --features sp1` / `--features risc0` still pass; default builds pull no new deps. `cargo clippy -p ethrex-prover --features lambdavm` and `-p ethrex-guest-program --features lambdavm` are warning-free on the LambdaVM files.

The new `lambdavm` entry in the `lint_zk` matrix will exercise the install action, the guest ELF build, and clippy on every PR.

**Guest parity validation (vs the lambda_vm repo's own guest)**

Both guests were executed in the LambdaVM executor over the four fixtures committed in `yetanotherco/lambda_vm` (`executor/tests/*.bin`), regenerated for the current `ProgramInput` rkyv layout with the upstream `tooling/ethrex-fixtures` generator (deterministic, offline). Cycle counts are deterministic instruction counts:

| Fixture | lambda_vm guest cycles | this PR's guest cycles | committed output |
|---|---|---|---|
| empty block | 706,930 | 601,731 | byte-identical |
| 1 transfer | 1,218,591 | 1,085,681 | byte-identical |
| 10 transfers | 4,120,569 | 3,693,103 | byte-identical |
| 4 transfers (distinct senders) | 2,477,349 | 2,234,772 | byte-identical |

ECSM syscall counts are exactly 4 × tx count in both guests (the x-only ecrecover signature), and all five decoded output fields (initial/final state hash, last block hash, chain id, tx count) match per fixture. The ~10% cycle advantage comes from this branch embedding a newer guest-program than the rev lambda_vm pins. Relative to the previous revision of this PR (pure-Rust k256 ecrecover), tx-bearing blocks got 3.8×–9.4× cheaper (~3.4M → ~309k cycles per transfer).

**Checklist**

  - **Not applicable** — no storage schema changes; the LambdaVM backend is purely on the prover side.

**Follow-ups (not blocking this PR)**

- Run `LambdaVmBackend::execute` / `prove` / `verify` against a real L1 block on a developer machine with the LambdaVM `cli` installed.
- Capture proving-time numbers (execution cycle baselines are in this description) and add them to `docs/zkvm-integrations.md`.
- Coordinate with the LambdaVM team to either retire the `executor/programs/rust/ethrex/` PoC in their repo or update it to point at this in-ethrex guest. The crypto provider is now maintained byte-for-byte in both repos, so changes must land in both; two small shared cleanups are queued: gate the `keccak_hash` import to non-riscv64 (silences a guest-build warning) and switch the keccak sponge to the in-place one in `lambda-vm-syscalls`.


azteca1998 (approved):

crates/prover/src/backend/lambdavm.rs
```diff
@@ -0,0 +1,244 @@
...
+/// `.github/actions/install-lambdavm`.
+#[derive(Default)]
+pub struct LambdaVmBackend;
+
+impl LambdaVmBackend {
+    pub fn new() -> Self {
+        Self
+    }
+
+    fn write_elf_file() -> Result<(), BackendError> {
```
ElFantasma, line 52: `write_elf_file` runs the stat + (if size differs) full read + compare on **every** `execute` / `prove` / `verify` call. For a several-MB ELF that's wasted I/O per request once the on-disk copy matches. A `static WRITTEN: OnceLock<()> = OnceLock::new();` flipped after the first successful write would skip the dance for the lifetime of the process. The tmp+rename is still right for first-time correctness; this just avoids the per-call cost.

crates/prover/src/backend/lambdavm.rs
```diff
@@ -0,0 +1,244 @@
...
+};
+
+use ethrex_common::types::prover::{ProofFormat, ProverOutput, ProverType};
+use ethrex_guest_program::{ZKVM_LAMBDAVM_PROGRAM_ELF, input::ProgramInput};
+
+use crate::backend::{BackendError, ProverBackend};
+
+const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/lambdavm_input.bin");
+const PROOF_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/lambdavm_proof.bin");
+const ELF_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/zkvm-lambdavm-program");
```
ElFantasma, line 15: `INPUT_PATH` / `PROOF_PATH` / `ELF_PATH` are global `CARGO_MANIFEST_DIR`-relative constants. Multiple `LambdaVmBackend` instances in the same process (or two prover processes from the same install) will race on the same files — `serialize_input` writes `INPUT_PATH`, then `prove_core` reads `PROOF_PATH`, with no isolation between them. L1-local-proving today so not actionable, but the L2 / multi-prover line in `docs/zkvm-integrations.md` ("L2 prover integration is planned") is the moment this becomes load-bearing. Even a `tempfile::NamedTempFile` per call, or a per-instance subdir under `CARGO_TARGET_TMPDIR`, would future-proof it.

crates/prover/src/backend/lambdavm.rs
```diff
@@ -0,0 +1,244 @@
...
+        let output = Command::new("cli")
+            .args([
+                "prove",
+                ELF_PATH,
+                "-o",
+                PROOF_PATH,
+                "--private-input",
+                INPUT_PATH,
+                "--time",
+                "--cycles",
```
ElFantasma: `--cycles` is passed to `cli prove` but the only stdout line we parse is `Proving time: <float>s` (line 236). If `--cycles` produces additional stdout without proving overhead, fine — but if the CLI charges measurable time for the cycle count, we're paying for output that's never consumed. Worth confirming with the LambdaVM team whether `--cycles` is free or whether we should drop it until something here reads the result.

crates/prover/src/backend/lambdavm.rs
```diff
@@ -0,0 +1,282 @@
...
+    sync::OnceLock,
+    time::{Duration, Instant},
+};
+
+use ethrex_common::types::prover::{ProofFormat, ProverOutput, ProverType};
+use ethrex_guest_program::{ZKVM_LAMBDAVM_PROGRAM_ELF, input::ProgramInput};
+
+use crate::backend::{BackendError, ProverBackend};
+
+const INPUT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/lambdavm_input.bin");
```
ElFantasma, line 13: Thanks for the round of fixes — re-checked all three of my June comments against the current file:

- **`write_elf_file` re-reading the ELF every call** — fixed, `ELF_WRITTEN.get().is_some()` at `:53`.
- **`--cycles` vs. the parsed stdout** — moot, the flag is gone; it's `--time` + `Proving time: <float>s` now.
- **Shared paths** — half fixed, and I think the open half is the one that matters more.

The ELF write is now safe:

```rust
let tmp_path = format!("{ELF_PATH}.{}.tmp", std::process::id());
std::fs::rename(&tmp_path, ELF_PATH)
```

PID-unique temp plus atomic rename — a concurrent reader sees either the old or the new file, never a torn one.

`INPUT_PATH` and `PROOF_PATH` didn't get the same treatment. They're still fixed `CARGO_MANIFEST_DIR` constants, `serialize_input` writes straight to `INPUT_PATH`, and there's no `Mutex`/`RwLock`/lock anywhere in the file. Two `prove`/`execute` calls overlapping — two threads in one process, or two prover processes from the same install — race on both.

**Why this half is worse than the ELF half:** a torn ELF fails loudly. A clobbered `INPUT_PATH` does not. Sequence: A serializes its input; B overwrites it; A shells out and proves **B's block**; A gets back a well-formed proof and attributes it to its own request. Same shape on `PROOF_PATH` in the other direction — whoever reads last wins. Nothing in the flow would surface it.

Two options depending on how much concurrency you expect:

1. **Per-invocation paths** — `format!("{}/lambdavm_input.{}.{}.bin", …, std::process::id(), counter)` with an `AtomicU64`, cleaned up after. Handles both the intra- and inter-process case, matching what you already did for the ELF.
2. **A backend-level `Mutex`** — simpler, but only serialises within one process; two prover processes from the same install still collide.

If single-threaded, single-process is a real invariant of how this backend is driven, that's a fine answer too — worth stating it in a comment on these constants so the next person doesn't parallelise the caller and hit this silently.

PR #6723, opened 2026-05-26
fix(l2): fix critical bugs in bench_metrics.sh

operagxoksana:
Fix three bugs in scripts/bench_metrics.sh: replace ((ms < min)) && min=$ms and ((ms > max)) && max=$ms with if (( )); then fi to prevent silent crashes under set -e when the arithmetic condition evaluates to false; replace the magic sentinel min=999999999 with min=${ms_arr[0]} to avoid incorrect results for large proving times; and add a local guard before min/max initialization to make the non-empty array invariant explicit rather than relying on a distant check earlier in the script.

### DIFF
--- a/scripts/bench_metrics.sh
+++ b/scripts/bench_metrics.sh
@@ -135,15 +135,25 @@ gpu=$(detect_gpu)
     done
 
     # Summary stats.
-    count=0; total=0; min=999999999; max=0; total_gas=0; total_txs=0
+    if [[ ${#ms_arr[@]} -eq 0 ]]; then
+        echo "No batches to summarize." >&2
+        exit 0
+    fi
+
+    min=${ms_arr[0]}
+    max=${ms_arr[0]}
+    count=0
+    total=0
+    total_gas=0
+    total_txs=0
     for i in "${!batches[@]}"; do
         ms=${ms_arr[$i]}
         gas=${gas_arr[$i]}
         txs=${txs_arr[$i]}
         count=$((count + 1))
         total=$((total + ms))
-        ((ms < min)) && min=$ms
-        ((ms > max)) && max=$ms
+        if (( ms < min )); then min=$ms; fi
+        if (( ms > max )); then max=$ms; fi
         [[ "$gas" != "-" && -n "$gas" ]] && total_gas=$((total_gas + ${gas%%.*}))
         [[ "$txs" != "-" && -n "$txs" ]] && total_txs=$((total_txs + ${txs%%.*}))
     done
--- a/scripts/bench_metrics.sh
+++ b/scripts/bench_metrics.sh
@@ -49,7 +49,7 @@ gas_arr=()
 txs_arr=()
 blocks_arr=()
 
-while read -r line; do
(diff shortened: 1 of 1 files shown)


scripts/bench_metrics.sh
```diff
@@ -135,20 +135,24 @@ gpu=$(detect_gpu)
...
+        total=0
+        total_gas=0
+        total_txs=0
+        for i in "${!batches[@]}"; do
+            ms=${ms_arr[$i]}
+            gas=${gas_arr[$i]}
+            txs=${txs_arr[$i]}
+            count=$((count + 1))
+            total=$((total + ms))
+            if (( ms < min )); then min=$ms; fi
```
ElFantasma, line 151: The `set -e` diagnosis is correct and the `if (( )); then; fi` form is the right rewrite. Worth noting in passing that the alternative one-liner `(( ms < min )) && min=$ms || true` works too — and is more compact when this pattern shows up in dozens of places — but the explicit `if` is more readable for a one-off, so the choice here is fine.

Not blocking, but file this if the pattern ever proliferates: the `|| true` trick avoids the rewrite and reads cleanly inline.

scripts/bench_metrics.sh
```diff
@@ -135,20 +135,24 @@ gpu=$(detect_gpu)
...
-        total=$((total + ms))
-        ((ms < min)) && min=$ms
-        ((ms > max)) && max=$ms
-        [[ "$gas" != "-" && -n "$gas" ]] && total_gas=$((total_gas + ${gas%%.*}))
-        [[ "$txs" != "-" && -n "$txs" ]] && total_txs=$((total_txs + ${txs%%.*}))
-    done
-
-    if [[ $count -gt 0 ]]; then
+    if [[ ${#ms_arr[@]} -gt 0 ]]; then
+        min=${ms_arr[0]}
```
ElFantasma: Just a scoping note for the changelog/PR title — "fix **critical** bugs" reads strong for what's actually shipping:
- Bug #1 (`set -e` + `((expr)) &&`): real, hits any CI run where the script exits on the first batch where `ms >= min`. This *is* a real failure mode, fair to call critical.
- Bug #2 (`min=999999999` sentinel): only manifests when proving time exceeds ~278 hours, which we won't see today.
- Bug #3 (the guard rewrite + `IFS=` on read): defensive cleanup, not a bug per se.

No code change needed — just consider toning the PR title ("fix bug in bench_metrics.sh set -e crash + defensive cleanup" or similar). Non-blocking.

ElFantasma (approved):

scripts/bench_metrics.sh
operagxoksana: Thanks for the feedback! I will take this advice into account in the future

operagxoksana:
Hey @ElFantasma when you have time, you can pay attention to this PR

azteca1998 (approved):

operagxoksana:
Thanks. Looks like the remaining workflow is waiting for maintainer approval

PR #6724, opened 2026-05-26
feat(l1): lift reorg cap + observability (PR 4)

### LINKED ISSUE #6685: Tracking: deep-reorg support via reverse-diff journal + overlay

edg-l:
> Part 4 of 4 in the deep-reorg support effort tracked in #6685. Stacks on
> #6689 (orchestration). Removes the hard 128-block cap, adds Prometheus
> observability, and ships a Kurtosis-driven integration test harness.

## Summary

- Replaces `REORG_DEPTH_LIMIT = 128` with `compute_reorg_ceiling`, a **physical** limit: the ceiling is the deeper of the in-memory layer-cache retention and the reverse-diff journal reach (`latest - (lowest_state_history_block - 1)`), capped at `latest` and by the operator override. Finality is deliberately **not** an input, matching geth/nethermind/reth: `-38006 TooDeepReorg` means "deeper than the client can reconstruct," not "crosses the finalized block." Finality is still enforced by the spec's no-reorg-past-finalized skip and the safe/finalized connectivity checks.
- `--max-reorg-depth` CLI flag (env: `ETHREX_MAX_REORG_DEPTH`), default `None`. Plumbed through L1 + L2 init.
- O(1) `Store::highest/lowest_state_history_block_number` via new `StorageReadView::first_key`/`last_key` primitives.
- 8 Prometheus metrics under `ethrex_reorg_*` / `ethrex_deep_reorg_*`, registered with the scrape registry: 3 gauges (overlay entries/bytes, journal length), 2 histograms (depth, reconcile duration), 3 counters (attempts/success/aborts).
- Kurtosis enclave config + assertoor scenario + manual GitHub workflow for end-to-end network-partition driven deep-reorg testing.

## What changed where

| Phase | Commit | Scope |
|---|---|---|
| 1 | `bb7d3e5a7e`, `adb57933c2` | O(1) state-history edge lookup |
| 2 | `dcd4741dd5`, `40b4f54750`, `c694beacb` | Lift the cap, physical (layer-cache + journal reach) |
| 3 | `61c8722503`, `d0e5692b07` | Prometheus metrics + registry wiring |
| 4 | `ef26cac986` | Rustdoc surface, operator docs |
| 5 | `031461cf63` | Kurtosis + simulator harness |
| 6 | tbd | Salt sibling chains in the simulator so chain B truly diverges from chain A |

## Simulator test calibration

The first version of `deep_reorg_beyond_128` produced byte-identical chains because `Chain::get_next_payload_attributes` derives `prev_randao` and `parent_beacon_block_root` deterministically from the parent hash. Both `base_chain.fork()` calls yielded the same blocks, so chain B never actually diverged ; the FCU to "chain B" was a no-op reorg back to chain A's tip, and the EL got stuck in cascading deep-reorgs because `engine_newPayload` short-circuited on `HEADERS` already containing each block.

Fix: `Chain` now carries a `salt: u8` mixed into the payload-attributes seed, set on chain B via `base_chain.fork().with_salt(1)`. Existing single-line tests (salt stays at 0) are byte-compatible. Doc comment on the field spells out the trap so the next person doesn't reintroduce it.

With the salt, exactly one deep_reorg fires (chain B's first FCU back to genesis), then chain B extends to 200 blocks via the normal shallow path. Test passes in ~1.8 s.

## Simulator scenarios

Two deep-reorg scenarios ship in `tooling/reorgs`:

- **`deep_reorg_beyond_128`** ; single node. Builds chain A canonical to 150, then FCUs to a salted chain B that forks at genesis. The first FCU is a 150-deep reorg, well past the legacy 128 cap. Asserts the reorg is accepted and chain B advances to 200.
- **`deep_reorg_side_chain_replay`** ; two nodes, exercises the `ACCEPTED` stash path end to end. Node 0 builds a 100-block common prefix; node 1 replays it, then builds a salted chain B to 300. Node 0 extends chain A to 300 (evicting the common ancestor's state from cache), then ingests chain B's 200-block divergent suffix via `newPayload` only (each block stashes as `ACCEPTED`, no FCU). A final FCU to chain B's tip drives one `reorg_apply_deep` with `side_chain_len = 200`.

## Bug 4 closeout

A "stack overflow in engine_newPayload on overlay-backed state" was tracked as an open issue from the prior handoff. Root cause: LEVM's `Stack::default()` materializes a 32 KB `[U256; 1024]` temp on the stack before `Box::new` moves it to the heap. In debug builds (which the simulator was using) this consumes ~32 KB per call; combined with the deep-reorg async chain it tipped the 2 MB tokio worker stack. Release builds placement-new directly into the heap, so production is unaffected. Documented as a debug-build artifact ; no EL code change in this PR.

## What's not in this PR

- Removing the journal entirely (full reorg-to-finality is now physically possible; pruning at finality still applies).
- Per-batch journaling in full-sync mode (`batch_mode == true` continues to skip the journal).
- Hardening `engine_newPayload`'s "header already known" short-circuit to verify the state is materialized before replying VALID. Only surfaces when the CL re-sends a known-hash block after a cache-wiping reorg ; deferred to a follow-up.

## EELS consume-enginex

The `consume-enginex` sim reuses one client across tests and resets it to genesis before each, so a test that leaves a deep chain turns the next reset into a deep canonical rewind. That is the case #6985 targets.

Ran the minimal failing pair against a release build of this branch (hive master `9207a978`, pinned `tests@v20.0.0` fixtures). `256_empty_blocks` leaves the client at block 257, then `INVALID-debug` resets to genesis, a 257-deep rewind:

```
./hive --client-file ../fixtures/hive/clients.yaml --client ethrex \
  --sim ethereum/eels/consume-enginex --sim.buildarg fixtures=tests@v20.0.0 \
  --sim.limit '.*fork_Osaka-blockchain_test_engine_x-(256_empty_blocks|test_program_program_INVALID-debug).*'
```

Result: 2/2 pass at default settings (no `ETHREX_MAX_REORG_DEPTH`), and 2/2 with `=512`. The reset goes through `reorg_apply_deep` instead of being rejected with `-38006`: the physical ceiling (journal reach) covers a full-height rewind (a fixed 128 cap does not), and the overlay replay rebuilds the evicted state.

Sim and fixtures are upstream on hive master, no custom fork. Only the minimal pair was run, not the full suite, and this is not wired to CI here.

## Test plan

- [x] `cargo test -p ethrex-blockchain -p ethrex-test -p ethrex-storage --features rocksdb` ; 465 pass.
- [x] `cargo clippy -p ethrex-blockchain -p ethrex-storage -p ethrex-metrics --all-targets --no-deps -- -D warnings`.
- [x] In-process simulator against a **release** ethrex binary (run from `tooling/reorgs`, `reorgs` is not a workspace member so `-p reorgs` does not apply):
  - `cargo run -- deep_reorg_beyond_128 /abs/path/target/release/ethrex` ; one deep_reorg fires, chain B advances to 200, exit 0 in ~1.8 s.
  - `cargo run -- deep_reorg_side_chain_replay /abs/path/target/release/ethrex` ; ACCEPTED-stash path then one `reorg_apply_deep` (side_chain_len 200), exit 0 in ~3.5 s.
- [x] `cargo doc -p ethrex-storage -p ethrex-blockchain --no-deps` ; zero `missing documentation` on new public surface.
- [x] EELS `consume-enginex` deep-reorg pair passes on hive master + `tests@v20.0.0`, at default settings and with `ETHREX_MAX_REORG_DEPTH=512` (see section above).
- [ ] Kurtosis enclave run (manual; see `tooling/reorgs/disruptoor/README.md`).
- [ ] CI green.

Closes #6685 alongside the prior three PRs.


crates/l2/sequencer/block_producer.rs
```diff
@@ -223,7 +223,7 @@ impl BlockProducer {
             .await?;
 
         // Make the new head be part of the canonical chain
-        apply_fork_choice(&self.store, block_hash, block_hash, block_hash).await?;
+        apply_fork_choice(&self.store, block_hash, block_hash, block_hash, None).await?;
```
ElFantasma, line 226: Blocking — endorse greptile's P1 with the concrete arithmetic. `apply_fork_choice(store, block_hash, block_hash, block_hash, None)` after `store_block` (which does not update `CANONICAL_BLOCK_HASHES`) hits the new finality-bounded check as follows:

- `head_is_canonical` = false (store_block didn't set canonical).
- `latest` = `block.number` (just stored).
- `new_canonical_blocks.last()` = `Some((block.number, block_hash))` → `canonical_link_height = block.number - 1` → `reorg_depth = 1`.
- `compute_reorg_ceiling(store, &block_hash, block.number, None)` → case 1 fires because finalized_hash = block_hash is locally synced (it IS the block we just stored) → `finality_ceiling = latest - h.number = 0` → returns 0 regardless of `max_reorg_depth`.
- `1 > 0` → `TooDeepReorg { reorg_depth: 1, limit: 0 }`.

Every L2 sequencer block will fail to canonicalize after this PR merges. Two fix directions worth considering:

1. **Callsite**: pass `finalized_hash = H256::zero()` here (or the actual last finalized) so `compute_reorg_ceiling` routes to case 2 (journal-based ceiling), which trivially clears depth 1.
2. **Ceiling formula**: give `finality_ceiling` a floor of 1 in `compute_reorg_ceiling` so canonical-extend FCUs are never rejected. That covers any future callsite that self-finalizes and matches geth's `head==current_head` carve-out at engine API level.

Either is fine; option 1 is a smaller diff and localises the L2 semantics. The Reorg Tests / Integration Test / Hive Engine tests failing in CI are likely downstream of this — worth verifying after fixing.

crates/blockchain/fork_choice.rs
```diff
@@ -13,21 +13,81 @@ use crate::{
...
+        return Ok(0);
+    }
+
+    if !finalized_hash.is_zero() {
+        // Case 1: finalized block is known AND locally synced.
+        // If the CL sends a finalized hash the EL hasn't synced yet (plausible after
+        // snap sync or a recent restart), fall through to the journal/operator path
+        // rather than treating finalized_number as 0 (which would let any depth pass).
+        if let Some(h) = store.get_block_header_by_hash(*finalized_hash)? {
+            let finality_ceiling = latest.saturating_sub(h.number);
```
ElFantasma: Root-cause note for the L2 breakage flagged on `block_producer.rs:226`: `finality_ceiling = latest.saturating_sub(h.number)` collapses to 0 whenever `finalized_hash == head_hash` (which any callsite doing "canonicalize this tip" pattern will pass). `max_reorg_depth` can only make the ceiling smaller (`min(finality_ceiling, d)`), so the operator override cannot rescue this either.

Suggestion — enforce a floor of 1 here so canonical-extend FCUs always clear:

```rust
let finality_ceiling = latest.saturating_sub(h.number).max(1);
```

Rationale: a depth-1 "reorg" is either (a) making a newly-stored tip canonical, or (b) replacing exactly one canonical tip with a sibling — neither requires unwinding past finality, so it should never be rejected on finality-bound grounds. This also matches geth's `head==current_head` carve-out at the engine-API level.

cmd/ethrex/cli.rs
```diff
@@ -449,6 +449,14 @@ pub struct Options {
         env = "ETHREX_PRECOMPUTE_WITNESSES"
     )]
     pub precompute_witnesses: bool,
+    #[arg(
+        long = "max-reorg-depth",
```
ElFantasma, line 464: Non-blocking coordination note: #6985 (open, danceratopz/EEST) also adds `--max-reorg-depth` with the same flag name, same env var, but different semantics — there it configures the store's `commit_threshold` (state-history retention) via `StoreConfig::max_reorg_depth`, defaulting to `Some(128)` on RocksDB / `Some(10000)` on InMemory. Here it's an operator override on top of a finality-bounded ceiling, defaulting to `None`.

Whichever lands first, the other will need adaptation. The two concepts (retention floor vs. finality-cap override) probably want to coexist under different flag names, or #6985's version should adopt this PR's semantics + wire the operator override through `BlockchainOptions` instead of `StoreConfig`.

ElFantasma (changes requested):

edg-l:
Rebased onto the updated #6689 base and pushed `59490f26f`.

**Blocking fix — Self-Finalized L2 Reorgs** (`block_producer.rs:226`): took option 2 from your analysis and floored `finality_ceiling` at 1 in `compute_reorg_ceiling` (case 1). A depth-1 canonical extend/switch has live parent state and is never a rewrite of finalized history, so it must never be rejected. This fixes the `TooDeepReorg { reorg_depth: 1, limit: 0 }` that would otherwise reject every L2 sequencer block. Added regression test `self_finalized_canonical_extend_is_allowed` (smoke_tests.rs) that reproduces the exact `store_block` + `apply_fork_choice(head, head, head, None)` sequence and asserts three self-finalized extends canonicalize.

Note: this PR now also carries the three orchestration P1 fixes via the #6689 rebase (admission-race test-and-set, release-build history-deletion guard, invalid-replay INVALID mapping).

Re the P2 **Known Blocks Re-Execute** (`payload.rs:1249`): left as-is — the `add_block` re-execution on evicted-state-but-reachable-parent is deliberate here (rebuilds the layer so subsequent FCUs don't loop back into `reorg_apply_deep`), and hardening the header short-circuit is already listed under 'What's not in this PR' as a follow-up. Happy to pull it forward if you'd prefer it in this PR.

edg-l:
Update: rebased onto the latest #6689, so this PR now also carries the #6687 overlay-cross-roots P1 fix (`overlay_serves` gate). Full deep-reorg simulator suite green against a release binary: `deep_reorg_state_oracle` (depth 150, storage+balance verified), `deep_reorg_side_chain_replay` (replay 200), `deep_reorg_beyond_128` — all exit 0.

edg-l:
Fixed the CI regression in the Hive Paris engine re-org tests (`d68d727ea`).

Root cause: `compute_reorg_ceiling` case 3 (no finalized block, empty journal) returned `max_reorg_depth.unwrap_or(0)`, so on a short unfinalized chain every reorg was rejected with `TooDeepReorg { limit: 0 }`. The `engine-withdrawals` suite hits exactly this (9 N-block re-org cases on short chains with no finality set), and the old fixed 128 cap had allowed them.

Fix: with nothing finalized there is no lower bound to protect, so any reorg the node can serve from its retained in-memory diff-layers is legal. Case 3 now returns the layer-cache retention (`Store::reorg_retention`, i.e. the pre-PR-4 cap) capped at `latest`, still bounded by the operator override. Cases 1 (finality) and 2 (journal) are unchanged, and their tests still pass.

Verified locally against a release build: `ethereum/engine --sim.limit engine-withdrawals` now passes 35/35 (was 9 failing). Added unit test `unfinalized_shallow_reorg_with_empty_journal_is_allowed`. CI will confirm the Paris/Cancun engine suites, which share this root cause.

edg-l:
Reworked the reorg ceiling to be **physical, not finality-bounded** (`c694beacb`), which fixes the remaining Hive engine re-org failures.

The Paris/Cancun `engine-api` failures (`Safe Re-Org to Side Chain`, `Re-org to Previously Validated Sidechain Payload`) and the earlier `engine-withdrawals` ones were all the same root cause: using finality distance as the `-38006` limit rejected legal shallow reorgs the CL is entitled to. geth/nethermind/reth do not reject reorgs by finality, and `-38006` means "deeper than the client can reconstruct," not "crosses the finalized block."

`compute_reorg_ceiling` now drops finality entirely and returns the deeper of the layer-cache retention and the journal reach, capped at `latest` and by the operator override. Finality is still enforced by the no-reorg-past-finalized skip and the safe/finalized connectivity checks. My earlier floor-at-1 and case-3 patches are subsumed by this.

Removed the two tests that asserted finality/journal as an upper bound; kept the operator-cap rejection test and the physical-allow tests.

Verified locally against a release build: `ethereum/engine` `engine-api` **129/0**, `engine-withdrawals` **35/0**; unit reorg suite green; `consume-enginex` deep-reorg pair still passes. CI re-running.

ElFantasma (approved):

ilitteri:
Audit of this PR (diff `feat/deep-reorg-orchestration...feat/deep-reorg-cap-lift`, head `cc2c9fe58`). No critical/high-severity bugs found in the changed code — ceiling math, `first_key`/`last_key` primitives, feature wiring, metrics lifecycle, and crash consistency (reconcile fold + journal `delete_range` + new entry in one batch) all verified. Test coverage is genuinely good: the state-oracle end-to-end scenario, journal proptests, and cap-lift regression tests are exactly the right checks. The items below are mostly design exposures and observability/doc fixes, cheap to address in the observability PR.

## Medium — no default bound on reorg depth or overlay memory (design tradeoff, worth explicit sign-off)

`crates/blockchain/fork_choice.rs:37-61`, `crates/storage/store.rs` (`install_overlay_for_reorg`). The default ceiling is bounded only by journal reach, which is bounded by finality advance. If finality stalls (inactivity leak, or a no-finality devnet), the journal — and therefore the accepted reorg depth — grows without bound. A single FCU can then force `Overlay::from_journal` to materialize the entire reverse-diff for the reverted range in memory (O(depth × state churn), no size check anywhere) plus O(depth) block re-execution, during which all other FCUs get SYNCING. The only mitigation is the opt-in `--max-reorg-depth`. This matches geth/reth's "trust the CL" stance, but besu (90k) and erigon (96) keep static caps — ethrex will have *none* by default. Recommend considering a large static safety cap or an overlay byte budget as defense-in-depth.

## Medium — cap lift widens exposure to #7001 (incomplete-journal window)

`crates/blockchain/fork_choice.rs:425-436` documents that journal entries written while the flat-KV generator was running remain permanently incomplete, so a later reorg unwinding into them (before they finalize and are pruned) is served stale flat-KV values — a wrong post-reorg state root, consensus-split class. The in-progress case is deferred with SYNCING, but the after-the-fact case is unguarded. Under the old 128 cap, unwinding into a long regeneration window was often rejected outright; this PR removes that incidental protection. #7001 should be treated as blocking before this stack serves mainnet deep reorgs.

## Low — `ethrex_reorg_depth` histogram records the wrong quantity in the head-canonical case

`crates/blockchain/fork_choice.rs:467-471`. The observed value is `head.number - pivot_number`. When head is already canonical (case 1: `pivot_number = head.number - 1`), the histogram records **1** no matter how many canonical blocks are actually unwound (`latest - head.number`, potentially thousands — precisely the deepest reorgs this metric exists to detect). In the side-chain case it records the new-branch length, while the `TooDeepReorg` gate (`fork_choice.rs:175`) measures reverted depth (`latest - canonical_link_height`) — same metric name, two definitions. Suggest observing `store.get_latest_block_number() - pivot_number`.

## Low — `ethrex_reorg_journal_length` gauge is never updated in steady state

`crates/blockchain/fork_choice.rs:611-619`. The gauge is only set inside the deep-reorg success path; on any node that never deep-reorgs it reads 0 forever, contradicting its doc ("shrinks as finality pruning runs") and `docs/l1/running/monitoring.md` ("should roughly track head - finalized"). Either update it periodically (O(1) now) or fix the docs.

## Low — `--max-reorg-depth 0` docs are misleading

`crates/blockchain/blockchain.rs:330-335` and `crates/storage/README.md` say `Some(0)` restores "the pre-PR-4 behaviour". Pre-stack behavior was a 128-block cap; `Some(0)` rejects *every* reorg — including routine 1–2 block mainnet reorgs — with `-38006`. `docs/l1/running/monitoring.md` additionally claims it "returns SYNCING", which is wrong. An operator following these docs would brick their node's fork-choice. The flag's behavior is self-consistent; the docs/comments are not.

## Low — ceiling's retention component reflects configured threshold, not actual reach

`crates/storage/store.rs:3537-3541`. `reorg_retention()` returns `commit_threshold` (128) unconditionally. Right after a restart (cold layer cache) or in batch-mode full sync (which skips journaling), the node can physically serve far less; such FCUs are admitted by the ceiling, then fail physically and loop as SYNCING instead of being cleanly rejected. Benign in practice, but the "physical ceiling" claim is approximate.

## Low — new `newPayload` re-execution path widens unguarded concurrency with deep-reorg replay

`crates/networking/rpc/engine/payload.rs:1235-1255`. The fall-through re-executes known-header blocks without checking `is_reorg_in_progress()`. During a deep reorg, a duplicate `newPayload` for a new-chain block can execute concurrently with the replay loop's direct `add_block` calls against the same freshly-swapped cache. Executions are deterministic and keyed by the same state root, so practical risk is low — worth a guard or a comment justifying safety.

## Low — stale test comments describe a finality-based ceiling that no longer exists

`test/tests/blockchain/smoke_tests.rs:297-298, 334, 461, 487, 495, 527` and `tooling/reorgs/src/main.rs:495-500` still explain the ceiling as "`latest - finalized_number`" / "finality ceiling 129". The merged code computes a purely physical ceiling; the tests pass via the in-memory retention floor (10000), not the mechanism the comments describe.

## Verified non-issues (checked and cleared)

Ceiling math matches the overlay's reconstruction range and "entry N unwinds N→N-1" semantics; `Some(0)` check precedes `StateNotReachable`; `TooDeepReorg` → `-38006`; journal contiguity holds (single-writer commit, bottom-up pruning, atomic reconciliation); all `STATE_HISTORY` keys are 8-byte BE so lexicographic == numeric for `first_key`/`last_key`; metrics cfg-gating compiles in both modes; metric reset paths cover all exits; abort touches only in-memory state; the `newPayload` change preserves the snap-sync pre-pivot fast path; all `apply_fork_choice` callers updated; L2 paths pass `None` (correct for a trusted sequencer).

PR #6725, opened 2026-05-26
build(l1): modernize Dockerfile with BuildKit + shrink context

edg-l:
## Summary

Bring the Dockerfile up to current best practices and stop leaking gigabytes into the build context.

### Build context size (vs `main`)

| Path | Size on main | Status after this PR |
|---|---|---|
| `.git/` | 1.1 GB | excluded; `vergen-git2` reads branch/SHA from `VERGEN_GIT_BRANCH` / `VERGEN_GIT_SHA` env (passed via build ARG) |
| `tooling/ef_tests/blockchain/vectors_zkevm/` | **2.3 GB** | excluded — **was leaking** (sibling `vectors/` was excluded, this one was missed) |
| `tooling/*` (non-`repl`/`monitor`) | ~50 MB | not copied; only `tooling/repl` and `tooling/monitor` are referenced as path deps by the root workspace's `Cargo.toml`. The rest of `tooling/` is a separate cargo workspace never used by the `ethrex` build. |
| `fixtures/blockchain/` | 340 MB | excluded; pure test data, never read at build time |
| `fixtures/{blobs,cache,rsp,hive}/` | ~10 MB | excluded; not read at build time |
| **Total transferred** | **~2.81 GB** | **~9.5 MB** (measured) |

### Image size

| Tag | Size |
|---|---|
| `ethrex:main` (current) | 180 MB |
| **`ethrex:test` (this PR)** | **126 MB** (~30% smaller) |

Savings come from the slim builder base, removing the redundant runtime copy of `cmd/ethrex/networks` (those JSONs are `include_str!`-embedded in the binary), and `--no-install-recommends` + apt cache mounts not leaking into the layer.

### Build speed

- `# syntax=docker/dockerfile:1.10` + BuildKit cache mounts on `~/.cargo/registry`, `~/.cargo/git`, and `target/` (survive across builds even when `recipe.json` changes).
- Apt cache mounts on `/var/cache/apt` and `/var/lib/apt`.
- `cargo binstall cargo-chef` (prebuilt binary) replaces `cargo install` source build, saves ~2 min cold.
- `rust:1.91-slim-bookworm` builder base (~700 MB smaller than `rust:1.91`).
- `COPY --link` throughout: independent layers can be copied in parallel and their cache survives base-image / earlier-step changes.

### Reliability

- `CARGO_NET_GIT_FETCH_WITH_CLI=true` + `git` in the builder image. The slim base doesn't ship `git`, so libgit2 was used by default and could hang inside the container on cold git-dep fetches. Forcing the CLI sidesteps the hang and uses the user's normal git behavior.

### Image / runtime

- Add OCI labels (`title`, `description`, `source`, `licenses`, `revision`, `version`).
- `ARG TARGETARCH` drives solc download (multi-platform-ready) instead of shell `uname -m`.
- Runtime still `ubuntu:24.04` + `libssl3` running as root, to keep snapsync compose named-volume mounts working.

### Metadata wiring

`GIT_SHA` / `GIT_BRANCH` / `VERSION` now flow consistently through:

- `Makefile` (new `TAG` variable; default `local`, override with `make build-image TAG=foo`)
- `.github/actions/build-docker/action.yml` (from `github.sha` / `github.head_ref || github.ref_name`)
- `.github/actions/snapsync-run/action.yml` (from local git)
- `tooling/sync/docker_monitor.py` (from local git)

Result (verified locally):

```
$ docker inspect ethrex:test --format '{{json .Config.Labels}}' | jq
{
  "org.opencontainers.image.revision": "626c2f2533bdcdfb92641720914ab5d305356f3b",
  "org.opencontainers.image.version":  "v13.0.0-14-g626c2f253",
  ...
}

$ docker run --rm ethrex:test --version
ethrex/v13.0.0-perf/dockerfile-buildkit-626c2f25.../x86_64-unknown-linux-gnu/rustc-v1.91.1
```

Previously the image baked in whatever `vergen-git2` read from `.git`; the metadata was correct but cost 1.1 GB of context per build.

## Test plan

- [x] `make build-image TAG=test` succeeds locally; build context transferred ~9.5 MB
- [x] `docker inspect ethrex:test` shows real `revision` + `version` labels
- [x] `docker run --rm ethrex:test --version` reports the correct branch + full SHA
- [ ] PR-main_l2 workflow (L1 + L2 image builds) green
- [ ] Daily snapsync compose still mounts `/data` as root (no perm regressions)
- [ ] First CI build after merge is slow (cache-from miss expected); subsequent builds use new cache


edg-l:
Addressed bot reviews in `3d3e9e7b4`:

- `VERGEN_GIT_SHA` now passed to builder (was already set in `626c2f253`)
- `fixtures/keys` copied so `include_str!` for `private_keys_l1.txt` resolves (was already set in `b84c33459`)
- `cargo-binstall` pinned to `v1.19.1` via `ARG` (no more `releases/latest`)
- Builder stage now also `COPY test ./test` to match planner
- Makefile stamp file tracks `tooling/repl` and `tooling/monitor`

Skipped: `TARGETARCH` fallback to `uname -m` — `# syntax=docker/dockerfile:1.10` mandates BuildKit, where `TARGETARCH` is always set.

Dockerfile
```diff
@@ -1,100 +1,143 @@
...
+# libgit2 hangs on some hosts/networks inside containers; the CLI also supports
+# single-commit fetches for rev-pinned deps.
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
+
+# Install cargo-chef via prebuilt binary (cargo-binstall) — avoids ~2 min source build.
+# cargo-binstall pinned for reproducibility; bump deliberately.
+ARG CARGO_BINSTALL_VERSION=v1.19.1
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/local/cargo/git \
+    curl -fsSL https://github.com/cargo-bins/cargo-binstall/releases/download/${CARGO_BINSTALL_VERSION}/cargo-binstall-$(uname -m)-unknown-linux-musl.tgz \
```
ElFantasma, line 30: `$(uname -m)` here is correct and should stay — but it reads as inconsistent next to the solc download below, which the PR (rightly) switched to `$TARGETARCH`. The difference is intentional: cargo-binstall installs cargo-chef, a **build-platform** tool that runs *during* the build, so it must match the stage's execution architecture (`uname -m`), not the image's target arch (`$TARGETARCH`). Using `$TARGETARCH` here would download the wrong binstall for emulated cross-builds (e.g. building `linux/arm64` on an amd64 runner).

Worth a one-line comment so a future "consistency" pass doesn't change it:
```dockerfile
# uname -m (NOT $TARGETARCH): cargo-chef is a build-platform tool, must match
# the stage's execution arch, not the target image arch.
```

Makefile
```diff
@@ -42,15 +42,29 @@ clean: clean-vectors ## 🧹 Remove build artifacts
...
+# Git metadata baked into the image via Dockerfile ARGs. Falls back to "unknown"
+# / "dev" if not in a git checkout.
+GIT_SHA    := $(shell git rev-parse HEAD 2>/dev/null || echo unknown)
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null || echo unknown)
+VERSION    := $(shell git describe --tags --always --dirty 2>/dev/null || echo dev)
+
 STAMP_FILE := .docker_build_stamp
-$(STAMP_FILE): $(shell find crates cmd -type f -name '*.rs') Cargo.toml Dockerfile
-	docker build -t ethrex:local .
+$(STAMP_FILE): $(shell find crates cmd tooling/repl tooling/monitor -type f -name '*.rs') Cargo.toml Dockerfile
```
ElFantasma: The stamp-file prerequisites don't include the git metadata, so the new `GIT_SHA`/`GIT_BRANCH`/`VERSION` args can go stale. If you switch to a branch whose tracked `.rs`/`Cargo.toml`/`Dockerfile` are identical (or just dirty the tree so `git describe --dirty` flips), `$(STAMP_FILE)` is still considered up-to-date and `make build-image` skips the rebuild — reusing an image baked with the previous SHA/version labels.

Low-impact in practice (cross-branch source usually differs), but since the whole point of this PR is accurate metadata, consider making the stamp depend on the resolved SHA, e.g. write `$(GIT_SHA)` into a tracked `.git_sha_stamp` and add it as a prerequisite, or just drop the stamp-file indirection for `build-image` and always rebuild (BuildKit layer cache makes the no-op rebuild cheap). Non-blocking.

azteca1998 (approved):

ElFantasma (approved):
**VERSION format diverges between local and CI.** Local/snapsync/monitor use `git describe --tags --always --dirty` → `v13.0.0-14-g626c2f253`; the build-docker action uses `dev-<full-sha>` for non-tag builds. Not wrong, but a `git describe`-style value in the action would make the OCI `version` label consistent across all build paths. Also worth confirming the "runtime runs as root" choice (documented as intentional for snapsync volume mounts) is acceptable to whoever owns the deployment security posture — it's the one thing here that's a deliberate tradeoff rather than a pure win.

ilitteri (approved):

PR #6731, opened 2026-05-27
fix(l1): support Amsterdam block production in dev mode

edg-l:
Dev mode fetched payloads via `engine_getPayloadV5` (which returns the EIP-7928 Block Access List) but submitted them via `engine_newPayloadV4`. V4 rejects any payload carrying a `block_access_list` field and returns `-38005 Unsupported fork` for Amsterdam-timestamped blocks, so dev mode could not produce a single Amsterdam block.

Added `engine_new_payload_v5` to `EngineClient` in `crates/networking/rpc/clients/auth/mod.rs`. The BAL travels inside `execution_payload.block_access_list`, so no extra argument is needed. `block_producer.rs` now routes BAL-carrying payloads to V5 and falls back to V4 for pre-Amsterdam blocks.

Discovered while verifying #6678 (BAL Prometheus metrics): `bal_*` metrics were never populated because dev mode failed every slot with `engine_newPayloadV4: block_access_list not allowed`. After this fix, a dev node with an Amsterdam-active genesis produces blocks and `bal_*` metrics populate as expected.

## Genesis fixture

Also adds `fixtures/genesis/l1-bal.json`, an Amsterdam-active dev genesis (mirrors `l1.json` with `amsterdamTime: 0` plus osaka/amsterdam blob schedules), so dev mode can produce BAL-carrying blocks out of the box:

```bash
ethrex --dev --network fixtures/genesis/l1-bal.json --datadir memory --metrics
```

## Test plan

- `cargo check -p ethrex-rpc -p ethrex-dev` builds clean
- `ethrex --dev --network fixtures/genesis/l1-bal.json --metrics`: previously failed every slot with `engine_newPayloadV4: block_access_list not allowed`; now produces Amsterdam blocks each slot


edg-l:
Smoke test:

```
ethrex --dev --network fixtures/genesis/l1-bal.json --datadir memory --metrics
```

Before this fix, every slot failed:

```
INFO  - Amsterdam: @0
ERROR Failed to produce block: error sending engine_newPayloadV4: Field 'block_access_list not allowed in engine_newPayloadV4' is incorrect or has an unknown format
```

After:

```
INFO  - Amsterdam: @0
INFO Produced block 0x7947adbfa320601a04b4b612ba931d41fa55b16fce106cd4d4b8e332edd82856
INFO Produced block 0xa2736186af7210d774228f87c05a46fecc5910687a19495a9a71a0b6cbb4ca32
INFO Produced block 0xec8c17667cf746bd726aa7c0cbcfc5e2d021df6e95cfa7af20f725a1d94af96d
INFO Produced block 0x3dc037335efb1b558b0e9e1ed2a237859ea4382464d8d4db0305ba2501b96b42
```

BAL-carrying Amsterdam blocks produced each slot, which also unblocks the `bal_*` metrics in #6678.

azteca1998 (approved):

crates/blockchain/dev/block_producer.rs
```diff
@@ -81,34 +81,49 @@ pub async fn start_block_producer(
...
+            engine_client
+                .engine_new_payload_v5(
+                    execution_payload,
+                    versioned_hashes,
+                    parent_beacon_block_root,
+                )
+                .await
+        } else {
+            engine_client
+                .engine_new_payload_v4(
```
ElFantasma, line 118: Out-of-scope follow-up observation: `tooling/reorgs/src/simulator.rs:360` still calls `engine_new_payload_v4(...)` unconditionally. If anyone uses the new `fixtures/genesis/l1-bal.json` (Amsterdam-at-genesis) with the reorg simulator, they'll hit the same V4-vs-V5 mismatch this PR fixes for the dev producer.

Not for this PR — same fix shape (`if payload.block_access_list.is_some() { v5(...) } else { v4(...) }`), or pull the routing into a helper on `EngineClient` so callers don't have to remember the branch. Worth a tracking issue if the team intends the BAL fixture to be useful in the simulator too.

crates/blockchain/dev/block_producer.rs
```diff
@@ -81,34 +81,49 @@ pub async fn start_block_producer(
...
+                    execution_payload,
+                    versioned_hashes,
+                    parent_beacon_block_root,
+                )
+                .await
+        };
+        let payload_status = match new_payload_result {
             Ok(response) => {
-                tracing::debug!("engine_newPayloadV4 response: {response:?}");
+                tracing::debug!("engine_newPayload response: {response:?}");
```
ElFantasma: nit: the log strings used to identify the exact endpoint (`engine_newPayloadV4`); after this change both branches log `engine_newPayload`, so when a failure is reported the version that was used (and therefore where to look in the server logs) is no longer in the message.

Cheap fix: bind the endpoint name at the branch point and interpolate, e.g.
```rust
let endpoint = if execution_payload.block_access_list.is_some() { "engine_newPayloadV5" } else { "engine_newPayloadV4" };
```
then `tracing::debug!("{endpoint} response: ...")` / `tracing::error!("... error sending {endpoint}: ...")`. Non-blocking.

crates/networking/rpc/clients/auth/mod.rs
```diff
@@ -146,6 +146,39 @@ impl EngineClient {
...
+        &self,
+        execution_payload: ExecutionPayload,
+        expected_blob_versioned_hashes: Vec<H256>,
+        parent_beacon_block_root: H256,
+    ) -> Result<PayloadStatus, EngineClientError> {
+        let request = NewPayloadV5Request {
+            payload: execution_payload,
+            expected_blob_versioned_hashes,
+            parent_beacon_block_root,
+            execution_requests: vec![],
```
ElFantasma, line 165: `execution_requests: vec![]` is consistent with the existing V4 dev-mode path (line 130 above) — the dev producer doesn't synthesize deposits/withdrawals/consolidations. Worth a one-line doc note here saying that explicitly: anyone trying to use the new `l1-bal.json` fixture to drive a deposit-bearing Amsterdam test will silently get blocks with no execution requests, which is surprising if you assume "dev mode block producer" means "valid Amsterdam blocks of every kind."

Not blocking — same limitation as V4 today. Just a doc nudge so the next reader doesn't re-discover it.

ilitteri (approved):

ElFantasma (approved):
LGTM. Left some nits, though

Issue #6729, opened 2026-05-26
perf(l1): BAL warmer prefetch experiments; bytecode batch wins, parallel phases don't


edg-l:
## Context

Round of perf experiments on the BAL parallel-execution warmer
(`warm_block_from_bal` in `crates/vm/backends/levm/mod.rs`). The
warmer runs three sequential prefetch stages before EVM execution:
account states, then storage slots, then contract code. Each stage
already uses `rayon::par_iter` over its inputs internally.

## Methodology

- Fixture: `bal-devnet-7-mainnet-mix-460` (460 blocks, 30 Ggas,
  transfer/EVM mix).
- Binary: `release-with-debug`, `import-bench --with-bal`,
  `--removedb` between runs.
- 4 independent runs per branch, second-of-pair from each bench script
  invocation. Run-to-run variance on this fixture is significant
  (~2.7% peak-to-peak), so single-run deltas were misleading; 4 runs
  is the minimum to characterize.

Baseline (main, 4 runs, total ms): `4738.86, 4741.58, 4614.21, 4673.82`
(mean 4691.6, median 4705.7).

## Experiments

| # | idea | branch | mean Δ vs main |
|---|---|---|---|
| A | dedup storage slots before the storage prefetch call (FxHashSet over (addr, slot) pairs to skip read+write duplicates) | (discarded; not pushed) | +0.8% |
| B | new `Database::prefetch_codes(&[H256])` with the lock-once batched pattern (mirrors `prefetch_accounts` / `prefetch_storage`); dedup hashes before calling, so factory-cloned contracts don't refetch the same bytecode N times | [`perf/bal-prefetch-codes-multi-get`](https://github.com/lambdaclass/ethrex/tree/perf/bal-prefetch-codes-multi-get) | -2.43% |
| C | run the storage-slot prefetch and the contract-code prefetch concurrently via `rayon::join` instead of back-to-back (both only depend on the account-state prefetch that ran first; they don't depend on each other) | [`perf/bal-warmer-phase23-parallel`](https://github.com/lambdaclass/ethrex/tree/perf/bal-warmer-phase23-parallel) | -0.38% (inside noise) |
| D | stack B + C: parallel storage/code prefetch where the code side also uses the new batched `prefetch_codes` | [`perf/bal-warmer-codes-parallel-stacked`](https://github.com/lambdaclass/ethrex/tree/perf/bal-warmer-codes-parallel-stacked) | -0.01% |

Per-branch run totals (ms):

[...]

PR #6732, opened 2026-05-27
perf(l1): prefetch all BAL storage synchronously before execution

edg-l:
## What

The BAL warmer and the parallel executor run concurrently, sharing one `CachingDatabase`. The warmer prefetched the block's storage slots from a background thread, and `prefetch_storage` only populates the cache after its whole parallel fetch completes. So the executor started against a cold cache and raced the warmer to the trie for its SSTORE original values, losing for the early transactions. A perf-record profile on the fixture below showed those SSTORE original-value trie walks were ~22% of total CPU, with ~72% of `CachingDatabase` storage reads falling through to the trie.

This change prefetches all BAL storage slots synchronously, before the executor thread starts. The prefetch is parallel (one `par_iter` across all cores), so it is fast, and it lets execution run fully warm with no trie fallthrough for storage and no warmer contending with the executor for CPU and the cache lock. The concurrent warmer is reduced to account states and contract code, which overlap execution well.

Storage is warmed in natural account order (slots grouped per account for storage-trie locality). Account prefetch is intentionally left concurrent: warming accounts synchronously too was measured slower, since account reads already overlap execution and a synchronous pass adds serial latency plus a double-fetch against the warmer.

## Why synchronous, not on the concurrent warmer?

The question is not "warmer thread vs exec thread", it is synchronous (before exec) vs concurrent (racing exec). The warmer and executor are spawned together and start at the same instant; `prefetch_storage` publishes to the cache only after its whole parallel fetch completes (a single write-lock at the end, no incremental visibility). So while the warmer is fetching, the executor sees a cold cache and reaches the trie itself for early-transaction SSTORE original values, paying a synchronous trie walk on the critical path.

Because the executor always starts with zero lead over the warmer, no ordering or chunking of the concurrent prefetch removes that race; concurrent variants topped out around -7% to -13%. The two threads also compete for CPU cores and the cache `RwLock`. Doing the storage prefetch up front, fully in parallel, costs only a few ms/block and lets exec run against a fully warm cache with no contention, for ~ -24%.

The intuition that overlapping the warmer with exec is free does not hold here: on this workload the overlap was a net loss (race plus contention). Only storage is moved; accounts and code stay on the concurrent warmer because they already overlap exec well (moving accounts synchronously regressed).

Tradeoff to note: on a live single-block `newPayload` the prefetch is now on the critical path before exec, so it can add a little latency versus the fully overlapped warmer; it is parallel and removes more fallthrough than it adds, so it should still net positive, and import / sync (back-to-back blocks) clearly win.

## Benchmark

Fixture: `bal-devnet-7-mainnet-mix-460` (460 blocks, ~30 Ggas, transfer/EVM mix). `release-with-debug`, `import-bench --with-bal`. Baseline = `main`. 4 runs each.

| metric         | baseline (main) | this PR   | delta              |
|----------------|-----------------|-----------|--------------------|
| wall time      | 4687.9 ms       | 3577.9 ms | -1110 ms (-23.7%)  |
| exec total     | 4263.2 ms       | 3166.4 ms | -1097 ms (-25.7%)  |
| warmer total   | 3148.3 ms       | 864.8 ms  | warmer now covers accounts + code only |
| warmer overlap | 97.0%           | 97.8%     | preserved          |

## Correctness

The prefetch only fills a read cache with values fetched from the underlying DB (`or_insert`, never overwriting executor writes); any cache miss falls through to the trie as before. Per EIP-7928 the BAL covers every storage slot accessed during the block, so coverage is complete. import-bench validates each block's state root and ran clean.

Related: #6729.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -544,12 +544,12 @@ impl Blockchain {
         // prefetching, skip the synchronous storage warm too. The warmer thread
         // below already honors the same toggle.
         #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
-        if self.options.bal_prefetch_enabled {
-            if let Some(bal) = bal {
-                let slots = LEVM::bal_storage_slots(bal);
-                if !slots.is_empty() {
-                    let _ = caching_store.prefetch_storage(&slots);
-                }
+        if self.options.bal_prefetch_enabled
+            && let Some(bal) = bal
+        {
+            let slots = LEVM::bal_storage_slots(bal);
+            if !slots.is_empty() {
+                let _ = caching_store.prefetch_storage(&slots);
             }
         }
 
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -539,11 +539,17 @@ impl Blockchain {
         // block) it adds serial latency the old overlapped warmer would have hidden.
         // The benchmarks above are cold-cache batch import, not single-block live
         // tail latency; the tradeoff is deliberate and favors throughput.
+        //
+        // Gated by `--no-bal-prefetch`: when the operator disables BAL-driven
+        // prefetching, skip the synchronous storage warm too. The warmer thread
+        // below already honors the same toggle.
         #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
-        if let Some(bal) = bal {
-            let slots = LEVM::bal_storage_slots(bal);
-            if !slots.is_empty() {
-                let _ = caching_store.prefetch_storage(&slots);
+        if self.options.bal_prefetch_enabled {
+            if let Some(bal) = bal {
+                let slots = LEVM::bal_storage_slots(bal);
+                if !slots.is_empty() {
+                    let _ = caching_store.prefetch_storage(&slots);
+                }
             }
         }
 
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -514,6 +514,39 @@ impl Blockchain {
             None
         };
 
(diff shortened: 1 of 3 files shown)


azteca1998 (approved):

crates/blockchain/blockchain.rs
```diff
@@ -488,6 +488,39 @@ impl Blockchain {
...
+        // longer overlapped with it. With a warm cache the reads hit and it is a
+        // no-op; on a genuinely cold block (first slot after restart, account-heavy
+        // block) it adds serial latency the old overlapped warmer would have hidden.
+        // The benchmarks above are cold-cache batch import, not single-block live
+        // tail latency; the tradeoff is deliberate and favors throughput.
+        #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
+        if let Some(bal) = bal {
+            let slots = LEVM::bal_storage_slots(bal);
+            if !slots.is_empty() {
+                let _ = caching_store.prefetch_storage(&slots);
```
ElFantasma: nit: `let _ = caching_store.prefetch_storage(&slots);` silently discards the error, whereas the old warmer path propagated it via `.map_err(...)?`. For a prefetch this is actually the *right* call — a failed cache-warm should degrade to trie reads during execution, not abort the whole block, and if the failure is a genuine storage problem it resurfaces with better context when the executor hits the same slot. But the silent `let _ =` reads like an oversight next to the old `?`. Worth a one-line comment making the intent explicit, e.g.:
```rust
// Best-effort warm: on error, fall through to trie reads during exec rather
// than aborting the block. A real storage fault resurfaces at execution.
let _ = caching_store.prefetch_storage(&slots);
```
Non-blocking.

ElFantasma (approved):
Just one nit

iovoid (approved):

PR #6733, opened 2026-05-27
fix(l1): clearer sync diagnostics during catch-up

edg-l:
## Motivation

A home validator reported their node "not finalizing" after 2 days of snap sync. ethrex was healthy and keeping pace; the consensus client was ~2 weeks behind and only feeding forkchoice heads a few blocks ahead of execution. The execution logs made this hard to see, and `eth_syncing` actively misled them.

## Changes

1. **`eth_syncing.highestBlock` returned garbage.** It reported `last_fcu_head().to_low_u64_be()`; the low 64 bits of the forkchoice head *hash*, not a block number. Now resolves the head hash to a block number via storage, then the pending-block store, falling back to `current_block` only when the target isn't downloaded yet (mid snap-sync).

2. **Sync target + staleness hint.** On the first header batch of a cycle, log the consensus-provided forkchoice head and how far behind the local head is. If the head is more than 8 blocks ahead and older than 30 min, hint that the consensus client may still be catching up. Gated on distance so a synced node stays quiet. The target is read from the pre-rewind forkchoice head, not the rewound ancestor we request headers from.

3. **"Caught up" log.** When a cycle that started behind finishes, log that it reached the consensus-provided head and is waiting for the next forkchoice update, so idle-waiting is distinguishable from a hang.


azteca1998 (approved):

crates/networking/rpc/eth/client.rs
```diff
@@ -58,13 +58,22 @@ impl RpcHandler for Syncing {
...
+            // Resolve it to a block number; if the head's header isn't stored yet (e.g.
+            // mid snap-sync), fall back to the current block instead of reporting garbage.
+            let head_hash = syncer
+                .get_last_fcu_head()
+                .map_err(|error| RpcErr::Internal(error.to_string()))?;
+            let highest_block = context
+                .storage
+                .get_block_number(head_hash)
+                .await?
+                .unwrap_or(current_block);
```
ElFantasma: The PR description says this resolves the head *"via storage, then the pending-block store, falling back to `current_block`"* — three tiers — but this is two: `get_block_number(head_hash)` (which only reads the `BLOCK_NUMBERS` table; I checked) then `current_block`. There's no pending-block-store tier.

That middle tier is exactly the interesting case: a forkchoice head that's been received but not yet executed (the mid-snap-sync scenario this PR is about) isn't in `BLOCK_NUMBERS` yet, so `highest_block` falls straight to `current_block` and `eth_syncing` reports `highestBlock == currentBlock` — distance 0, i.e. "looks synced" while still catching up. That's a milder version of the garbage you're fixing, but the same failure mode.

Two options:
- If the pending/fork-choice store can resolve a received-but-unexecuted head to a number, add that lookup between the two (matches the body).
- Otherwise, reword the PR body to the honest 2-tier behavior — the inline comment above already describes it correctly, so it's just the description that overstates.

Non-blocking, but worth closing the gap since accurate `eth_syncing` during catch-up is the whole point.

crates/networking/p2p/sync/full.rs
```diff
@@ -92,6 +123,35 @@ pub async fn sync_cycle_full(
             first_header.number,
             last_header.number,
         );
+
+        // On the first batch, report the distance to the consensus-provided head and
+        // warn if that head is stale (a strong signal the consensus client is behind).
+        if !sync_target_logged {
+            sync_target_logged = true;
+            let target = first_header.number;
```
ElFantasma: nit: `target = first_header.number` is the first header of the first fetched batch, which is a good-enough proxy for the diagnostic. The PR body says the target is read from *"the pre-rewind forkchoice head, not the rewound ancestor we request headers from"* — but `first_header` *is* the header we request from, so after a rewind to a common ancestor this `target` (and therefore the `behind` distance) understates the true forkchoice head the consensus client advertised.

For a log line that's fine — it errs toward under-warning, not false alarms. But if you want the distance to reflect the actual FCU head (matching the body's claim), resolve `sync_head` to its number once at the top of the cycle and use that. Low priority.

ilitteri (approved):

ElFantasma (approved):
**`highestBlock` fallback is 2-tier, but the PR description claims 3-tier.** The body says it resolves the head hash *"via storage, then the pending-block store, falling back to `current_block`"* — but the code is `get_block_number(head_hash).unwrap_or(current_block)`, and `get_block_number` only reads the `BLOCK_NUMBERS` table (verified — no pending/fork-choice store lookup). The missing middle tier matters precisely for the case this PR targets: a forkchoice head that's been *received but not yet executed* (mid snap-sync) won't be in `BLOCK_NUMBERS`, so `highestBlock` silently collapses to `current_block` and `eth_syncing` reports distance 0 — which is the same class of "looks synced when it isn't" the PR set out to fix. Either (a) add the pending-store lookup the body promises, or (b) correct the description to match the honest 2-tier code + the inline comment (which already says only "if the head's header isn't stored yet … fall back to the current block"). See inline.

PR #6737, opened 2026-05-27
refactor(l1): key TRANSACTION_LOCATIONS by tx hash via merge operator

### LINKED ISSUE #6708: refactor(l1,storage): simplify TRANSACTION_LOCATIONS schema to key by tx_hash only (follow-up to #6688)

ElFantasma:
**Motivation**

Follow-up to [#6688](https://github.com/lambdaclass/ethrex/issues/6688) / [#6702](https://github.com/lambdaclass/ethrex/pull/6702) and the redesign tracked in [#6708](https://github.com/lambdaclass/ethrex/issues/6708). **Supersedes [#6718](https://github.com/lambdaclass/ethrex/pull/6718)**, which used a read-modify-write (RMW) write path that measured a real per-block storage regression on mainnet. This PR keys `TRANSACTION_LOCATIONS` by tx hash *without* that regression, using a RocksDB merge operator.

**Description**

`TRANSACTION_LOCATIONS` moves from the v2 composite-key schema (`key = tx_hash ‖ block_hash`, `value = (BlockNumber, BlockHash, Index)`, read via a prefix iterator) to v3:

| | v2 (current) | v3 (this PR) |
|---|---|---|
| Key | `tx_hash ‖ block_hash` (64B) | `tx_hash` (32B) |
| Value | one `(BlockNumber, BlockHash, Index)` | `Vec<(BlockNumber, BlockHash, Index)>` |
| Read | `prefix_iterator` + per-entry filter | single `get_cf` + canonical filter |
| Write | one `put` per tx | one **`merge_cf`** per tx |

The read path becomes a single `get_cf` on a fixed 32-byte key — O(1), no iteration, no risk of reintroducing the unbounded-scan footgun from #6688.

**Why a merge operator instead of read-modify-write.** The one-to-many shape (a tx hash maps to one location, or several across reorgs) needs the write to merge into any existing value. The obvious approach — read the existing `Vec`, append, write back — adds a per-tx `get_cf` on the hot block-import path. On mainnet that measured a **2–8× `store_ms` regression** (the per-tx point lookup has an irreducible memtable-traversal floor that a bloom filter can't remove, and bloom coverage of deep levels lags until full compaction). The merge operator removes the read entirely: each write is a `merge_cf` append, and RocksDB folds operands into the `Vec` at read/compaction time. As a bonus it's atomic at the storage layer, so there's no "writers must be serialized" assumption.

**Measurements (mainnet-10, same hardware, head-of-chain `store_ms`):**

| approach | mean | p99 | stationary |
|---|---|---|---|
| v2 composite + break (current main) | 2.58 ms | 5.9 ms | yes |
| v3 RMW, no bloom (#6718) | 5.6–19.8 ms | up to 62 ms | no |
| v3 RMW + bloom filter | 14.8 ms | 33.9 ms | no |
| **v3 merge operator (this PR)** | **~2.5 ms** | **~4.8 ms** | **yes** |

Read latency (`eth_getTransactionByHash`) is ~0.4 ms across all approaches. So the merge operator matches the current schema's write cost while delivering the read-path cleanup. (Final mainnet validation of this exact branch is pending review.)

**Migration.** `migrate_2_to_3` is registered in the migration framework from [#6519](https://github.com/lambdaclass/ethrex/pull/6519) (after the receipts `migrate_1_to_2`). It streams the v2 table, groups composite keys by tx hash, and rewrites each group atomically (put 32-byte key + delete 64-byte keys), chunked for bounded memory and idempotent on crash-resume. On a snap-synced mainnet node it runs in ~10–30 s; archive nodes scale with table size.

**Merge correctness.** The merge operator is *associative*, which requires the operand encoding to equal the value encoding — both are `Vec<(BlockNumber, BlockHash, Index)>` (operands are single-element). This is essential: during compaction RocksDB folds operands together (PartialMerge) and re-feeds the result as an operand; if the formats differed the result would fail to decode and entries would be silently dropped. Covered by unit tests for the associative composition plus a real-RocksDB integration test that writes operands across flushed SSTs, forces a compaction, and asserts nothing is dropped.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) — bumped 2 → 3 with the corresponding `migrate_2_to_3`.

Closes #6708


ilitteri (approved):

ElFantasma:
Thanks all — addressed in `e5b8ff915`.

**Migration resume robustness (Codex, Claude).** `flush_tx_location_group` now writes the rewritten value with `merge` instead of `put`. Since the operand and the stored value are the same `Vec` type, a re-processed group **unions** (dedupe by `block_hash`) with any existing v3 value rather than overwriting it — so the "crash-safe resume" guarantee now holds by construction, not by relying on per-group batch atomicity. Added `migrate_2_to_3_unions_same_hash_mixed_state` covering the exact pathological case (a tx_hash with both a v3 value and leftover v2 siblings → all locations survive). Re: Claude's InMemory-ordering note — the InMemory `prefix_iterator` explicitly sorts, so adjacency grouping is safe on both backends; added a comment to that effect.

**`RocksDBWriteTx::merge` CF validation (Codex, Greptile).** Now fails fast: merging any table other than `TRANSACTION_LOCATIONS` returns an error immediately, matching the trait contract and the InMemory backend's behavior, instead of deferring to read/compaction time.

**Corrupt operand handling — switched to fail-loud (Claude, Greptile P2).** `tx_locations_merge` now returns `None` on any RLP decode failure (base value or operand) rather than skipping and committing a partial result. `None` makes RocksDB surface a merge/corruption error on the affected key. We'd rather fail loud than silently land in a corrupted DB without knowing. Tests updated: `malformed_operand_aborts_merge`, `malformed_base_value_aborts_merge`.

**Bloom-filter omission (Claude).** Added an inline comment in the CF config explaining why there's intentionally no bloom filter (merge removes the hot-path negative gets; remaining negative reads are rare user queries; benchmarked as not helping the RMW variant either).

Not changed: the post-loop no-op `commit()` on an empty table (harmless) and the `total_groups`/`groups_in_batch` post-loop asymmetry (documented intentional).

The two failing checks are unrelated to the diff: **Validate** was the PR-title scope (`storage` isn't a valid scope — fixed the title to `refactor(l1)`), and **Test** is a flaky nix-cache `403` in `make image.raw` (will re-run). This branch is storage-crate Rust only, no `Cargo.lock` change.

ElFantasma:
## Authoritative measurement — merge operator (this PR) vs current main, same node

Ran the reviewed commit on mainnet-10, then the old-schema `main` build on the **same node / same DB / same hardware** (mainnet-10 was freshly wiped + snap-synced for a clean baseline). This removes the cross-node hardware variance that confounded earlier comparisons.

**Write path — `store_ms` (last 300 head-of-chain blocks):**

| variant | mean | p50 | p90 | p99 | max |
|---|---|---|---|---|---|
| current `main` (v2 composite key + #6702 break) | 2.36 ms | 2.30 | 3.27 | 4.57 | 4.95 |
| **this PR (v3 merge operator)** | **2.26 ms** | 2.13 | 3.37 | 4.81 | 5.60 |

**Read path — `eth_getTransactionByHash` (mean / p99 ms):**

| variant | hot | mid | nonexistent |
|---|---|---|---|
| current `main` | 0.374 / 0.512 | 0.485 / 0.846 | 0.254 / 0.501 |
| this PR | 0.416 / 0.565 | 0.539 / 0.977 | 0.245 / 0.581 |

**Parity on both paths** — write `store_ms` is indistinguishable (this PR marginally lower on the mean, within noise), and read latency matches (after #6702, the old prefix-iterator read is already fast for the common 1-entry case, so the single-`get_cf` read is a wash on latency; the win is code-safety + no unbounded-scan footgun).

For contrast, the earlier read-modify-write attempt (#6718) regressed `store_ms` to **5–20 ms** on this same node — the per-tx point lookup on the write path, which a bloom filter only partially mitigated (deep-level coverage lags, and the memtable-traversal floor is untouched). The merge operator removes that lookup entirely, which is why it lands back at the old schema's cost.

**Migration:** v2→v3 ran in ~0.2 s on this snap node (347K entries); it scales with table size on archive nodes. **Decode errors: 0** through warmup — the associative operand-format fix + fail-loud merge held. Leaving it running overnight as a final compaction soak.

edg-l (approved):

azteca1998 (approved):

PR #6740, opened 2026-05-27
fix(l1): align to latest EIP-8025 spec and unskip stateless validation tests

iovoid:
**Motivation**

The stateless execution zkEVM fixtures have been updated to reflect bal-devnet-7 spec (which we currently implement) and some other minor changes to the wire format.

**Description**

Unskips all of the EIP-8025 tests, implements the latest changes to the ProgramInput, and closes validation gaps.

Since we were using the old witness format, a small refactor is needed to keep support.

In the process we found a divergence with [the spec](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/src/ethereum/forks/amsterdam/vm/instructions/system.py#L712) in how CALLs order gas checks and state reads. Fixing this was necessary since we were requiring a code to be in the witness when that wasn't needed.


crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -81,6 +95,16 @@ impl OpcodeHandler for OpCallHandler {
...
 
+        if is_delegation_7702 {
+            vm.current_call_frame.check_gas(
+                static_cost
+                    .checked_add(create_cost)
+                    .ok_or(ExceptionalHalt::OutOfGas)?
+                    .checked_add(eip7702_gas_consumed)
+                    .ok_or(ExceptionalHalt::OutOfGas)?,
+            )?;
+        }
```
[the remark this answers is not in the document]
iovoid: False for several reasons:
- To know the account is empty (and thus needs creation) you need to read it and therefore include a proof of non-existence in the witness
- The code in fact only uses the creation cost when is_delegation is True. See https://github.com/ethereum/execution-specs/blob/forks/amsterdam/src/ethereum/forks/amsterdam/vm/instructions/system.py#L407-L432

ilitteri (approved):

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -81,6 +95,16 @@ impl OpcodeHandler for OpCallHandler {
             create_cost,
         );
 
+        if is_delegation_7702 {
+            vm.current_call_frame.check_gas(
```
edg-l, line 99: EELS Amsterdam `call()` L454 does `check_gas(extra_gas + extend_memory.cost)` where `extra_gas = access + transfer + delegation_access_cost`; `create_cost` is not included. It is charged later from the separate state-gas pool via `charge_state_gas` (L465-468), not regular gas.

Including `create_cost` in the regular-gas preview here can OOG on regular gas in the delegation + empty-account + value-transfer case where EELS would proceed and pay create_cost from the state-gas reservoir. CI is green so no v0.4.1 fixture triggers it, but the asymmetry vs the non-delegation branch (no second `check_gas`) is worth a comment or dropping `create_cost` from the preview. Intentional?

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -48,27 +48,41 @@ impl OpcodeHandler for OpCallHandler {
...
+            .max(calculate_memory_size(return_offset, return_len)?);
+        let address_was_cold = !vm.substate.is_address_accessed(&callee);
+        let memory_expansion_cost =
+            memory::expansion_cost(new_memory_size, vm.current_call_frame.memory.len())?;
+        let access_gas_cost = if address_was_cold {
+            gas_cost::COLD_ADDRESS_ACCESS_COST
+        } else {
+            gas_cost::WARM_ADDRESS_ACCESS_COST
+        };
+        let static_cost = memory_expansion_cost
```
edg-l: Static-gas block (`new_memory_size` + `address_was_cold` + `static_cost` + `check_gas(static_cost)`) is duplicated across CALL/CALLCODE/DELEGATECALL/STATICCALL (also at lines 238, 356, 474); only `value_cost` varies. Extract a helper returning `(new_memory_size, address_was_cold, static_cost)` after `check_gas` so the static-then-state ordering invariant lives in one place. Can be a follow-up.

tooling/ef_tests/blockchain/test_runner.rs
```diff
@@ -100,21 +102,24 @@ pub async fn run_ef_test(
     // Two-pass approach: pass 1 collects the BAL produced by sequential execution, pass 2
     // re-executes using that BAL to drive parallel (BAL-warmed) execution and verifies the
     // same final state is reached.
+    // Skipped under `stateless`: `eip-8025` on `ethrex-blockchain` drops the merkleizer
+    // Sender the parallel path needs; the non-stateless runs still cover this.
+    #[cfg(not(feature = "stateless"))]
```
edg-l, line 125: The current wording suggests this is a temporary skip waiting on a fix, but I think it's actually a permanent design boundary: the stateless harness doesn't drive the `blockchain.add_block` pipeline that `run_two_pass_parallel` exercises, and parallel BAL-driven merkleization gives no benefit in zkVM guest builds (single-threaded proof targets). The parallel path is already covered by the non-stateless runs against the same fixtures.

Could you reword along the lines of "two-pass parallel exercises the blockchain pipeline's BAL-driven merkleization, which the stateless harness doesn't use; non-stateless runs already cover it on the same fixtures" so future readers don't treat this as a regression to chase?

crates/common/types/block_execution_witness.rs
```diff
@@ -589,24 +626,18 @@ impl GuestProgramState {
...
-                    "Missing bytecode for hash {} in witness. Defaulting to empty code.", // If there's a state root mismatch and this prints we have to see if it's the cause or not.
-                    hex::encode(code_hash)
-                );
-                Ok(Code::default())
-            }
-        }
+        self.codes_hashed
+            .get(&code_hash)
+            .cloned()
+            .ok_or(GuestProgramStateError::MissingBytecode(code_hash))
```
edg-l, line 640: `get_account_code` now hard-errors with `MissingBytecode` instead of silently returning empty. Looks correct for stateless validation, but can you confirm no non-stateless caller relied on the prior silent-empty fallback? A quick `rg get_account_code` over the workspace would lock it down. Test matrix is green but worth eyeballing.

edg-l (commented):
Reviewed against EELS `tests-zkevm@v0.4.1` (`forks/amsterdam/vm/instructions/system.py`). Gas-charge ordering rewrite matches spec; raising a few items inline.

Verified separately:
- `.fixtures_url_amsterdam` (bal@v7.2.0) untouched.
- `vectors_zkevm/` stays a separate extraction root from `vectors/`. This isolation is correct and worth keeping as a design invariant: zkevm and bal fixture bundles version on independent cadences, so when the zkevm bundle drifts out of date relative to the current bal spec (as happened with zkevm@v0.3.3 vs bal-devnet-7 on main) we need the two trees to stay disjoint. Without that separation, extracting one would clobber the other and the suite the maintainers haven't bumped yet would silently break.
- `Test - {ubuntu-22.04, ubuntu-22.04-arm, macos-15}` (which runs `make -C tooling/ef_tests/blockchain test`) is green on all three runners; both bal@v7.2.0 (`test-levm`) and the re-broadened zkevm@v0.4.1 (`test-stateless`) are exercised.
- Pre-Amsterdam forks unchanged: `gas_cost::call` still bakes `CALL_TO_EMPTY_ACCOUNT` into `call_gas_costs` when `fork < Amsterdam` (`gas_cost.rs:748`).

Approving once the delegation-branch `check_gas` question and the two-pass-parallel comment are addressed.

tooling/ef_tests/blockchain/test_runner.rs
iovoid: Improved in [79a8532](https://github.com/lambdaclass/ethrex/pull/6740/commits/79a8532032fd475a41676d0aaa2b2cd4cc8d5371)

crates/common/types/block_execution_witness.rs
iovoid: Confirmed get_account_code isn't used elsewhere. Other methods (RPC, etc) use the Store and not the VM to access bytecode.

iovoid:
You are right about the gas, I was looking at the forks/amsterdam branch which I assumed was more up-to-date, but actually doesn't include EIP-8037.

crates/vm/levm/src/opcode_handlers/system.rs
iovoid: The change was short, so done in bf87942ef0

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -48,27 +48,32 @@ impl OpcodeHandler for OpCallHandler {
...
+            args_offset,
+            args_len,
+            return_offset,
+            return_len,
+            callee,
+            value_cost,
+        )?;
+
+        vm.substate.add_accessed_address(callee);
+        let address_is_empty = vm.db.get_account(callee)?.is_empty();
```
ElFantasma: `get_account(callee)?` unconditionally reads the callee account. But `address_is_empty` is only consumed at lines 69 (`create_cost`, gated on `!value.is_zero()`) and 111 (`needs_state_gas`, same gate). So when `value.is_zero()`, the read result is computed and immediately thrown away.

That's the *same class* of issue this PR fixes for `eip7702_get_code`: a state read that EELS doesn't perform, which forces witnesses to include records that EELS-conformant witnesses can legitimately omit. EELS' spec only checks `target.balance + value == 0` (the empty-via-create case) when value > 0; for value = 0 calls, the account is not loaded for this purpose.

Low-effort fix:
```rust
let address_is_empty = if value.is_zero() {
    false  // unused in either gated path
} else {
    vm.db.get_account(callee)?.is_empty()
};
```
Worth verifying against the unskipped EIP-8025 tests — if none of them currently catch this, it's because no witness fixture is sparse enough yet; that'll be a future witness regression.

Non-blocking but same family as the bug this PR fixes.

crates/common/types/block_execution_witness.rs
ElFantasma, line 640: Removing the silent "default to empty Code" fallback is the right call — it was masking exactly the kind of bug the system.rs CALL-reorder is fixing (witness misses a code entry; we silently default to empty; downstream execution diverges from EELS).

Worth a sanity check that **every** code-hash path the VM walks during execution is now preceded by an addressed-cold check or the witness invariant guarantees inclusion. If `get_account_code` can be reached for a code hash that wasn't already established as touched, this change converts the previous silent-but-wrong behavior into a hard `MissingBytecode` failure — possibly during stateless validation of an otherwise-valid block.

Based on the PR description ("unskipping tests"), the unskip is the regression net here. If those tests cover the once-fallback cases, we're good. Non-blocking.

crates/common/types/block_execution_witness.rs
```diff
@@ -231,6 +214,58 @@ impl RpcExecutionWitness {
...
+/// state-root check in `execute_blocks`.
+///
+/// Call before any sort/dedup, since reordering hides violations.
+pub fn validate_witness_headers_chain(
+    headers: &[BlockHeader],
+    crypto: &dyn Crypto,
+) -> Result<(), GuestProgramStateError> {
+    for pair in headers.windows(2) {
+        let (prev, next) = (&pair[0], &pair[1]);
+        if next.number != prev.number.saturating_add(1)
```
ElFantasma: nit: `prev.number.saturating_add(1)` silently lets a contiguous chain pass at `u64::MAX → u64::MAX`. Astronomically irrelevant, but the safer expression is `prev.number.checked_add(1)` which returns `None` and lets the comparison fail loudly — matches the spirit of the function ("validate chain linkage, surface violations").

crates/guest-program/src/l1/input.rs
```diff
@@ -28,6 +30,45 @@ impl ProgramInput {
     }
 }
 
+/// Input for the L1 stateless validation program (EIP-8025 build).
+///
+/// `Direct` carries in-memory blocks + witness (test path). `Wire` carries an
+/// already-decoded EIP-8025 stateless input from spec wire bytes.
+#[cfg(feature = "eip-8025")]
+pub enum ProgramInput {
```
ElFantasma, line 43: The same identifier `ProgramInput` is a `struct` (line 15) without the `eip-8025` feature and an `enum` (here) with it. Every consumer of `ProgramInput::new(blocks, witness)` works under both because both have a matching `new` method — but `ProgramInput::default()` returns different shapes (struct-default vs `Direct { ..Default::default() }`), and any code that pattern-matches on `ProgramInput::Wire(...)` or `ProgramInput::Direct { .. }` only compiles under one cfg.

The shape is consistent enough that this is probably fine, but a one-line module-level doc note ("this type's *kind* is feature-gated; cross-feature destructuring will not compile in the disabled branch") would save the next contributor a confused 10 minutes. Non-blocking.

ElFantasma (approved):
LGTM, but left some non-blocking comments. The `get_account(callee)` may be the more important one

edg-l (approved):
Reviewed for Amsterdam devnet impact. Only `opcode_handlers/system.rs` touches the live execution path, and the CALL-family reorder is semantically equivalent to main:

- `create_cost` and `needs_state_gas` collapse to the same `(empty && value!=0)` conditions (new code forces `address_is_empty = false` when value==0).
- The new `check_gas` calls only assert, never consume; authoritative deduction stays in the unchanged `increase_consumed_gas`/`increase_state_gas`. Final success/OOG outcome is identical.
- Moved warming is unobservable since frames revert substate on OOG.

Everything else (witness reconstruction, guest-program, prover, ef_tests) is stateless/proving or test infra, feature-gated behind eip-8025/stateless and off the live node path.

CI green including Hive Consume Engine Amsterdam and EF Tests (bal@v7.2.0). Bot findings are non-blocking spec-fidelity notes, not devnet regressions. LGTM.

PR #6741, opened 2026-05-28
feat(l1): add experimental REST/SSZ /new-payload-with-witness endpoint

developeruche:
This PR introduces an experimental `POST /new-payload-with-witness` REST endpoint to Ethrex.

**Motivation**
Witness retrieval today requires two round-trips , `engine_newPayload` followed by `debug_executionWitness`, with the response serialized as JSON. For large witnesses, this encoding alone inflates payload size by roughly **2×**. On Ethrex, this combined flow currently takes ~8 seconds for a 500 MB witness block.

By serving the endpoint over **HTTP with SSZ serialization** and merging both operations into one, we reduce that to **~1.2 seconds**, a ~6.5× improvement.

**Description**
- Adds a REST-based `POST /new-payload-with-witness` route outside the JSON-RPC layer
- Uses SSZ for both inbound payload and outbound witness encoding
- Validates the payload and produces the execution witness in a **single atomic operation**
- Marked as **experimental**, consistent with how `debug_executionWitness` was introduced

It is also important to note that there are ongoing discussions around introducing SSZ + HTTP to the Engine API more broadly, replacing JSON-RPC entirely, including proposals to combine "new-payload" execution and witness retrieval into a single operation. Once these discussions mature, we can standardize around a unified approach. However, this is a significant architectural change and will likely take time to reach consensus.

Engine: add Rest-SSZ spec
https://github.com/ethereum/execution-apis/pull/793
Add SSZ to Engine API
https://github.com/ethereum/execution-apis/pull/764
Proposal: opt-in witness retrieval via flag on existing engine methods
https://github.com/ethereum/execution-apis/pull/799


crates/networking/rpc/engine/payload.rs
```diff
@@ -1024,6 +1097,30 @@ async fn handle_new_payload_v3(
     handle_new_payload_v1_v2(payload, block, context, bal).await
 }
 
+pub async fn handle_new_payload_v3_with_witness(
```
ElFantasma, line 1100: The duplication concern: `handle_new_payload_v3_with_witness`, `handle_new_payload_v4_with_witness`, `handle_new_payload_v1_v2_with_witness`, `try_execute_payload_with_witness`, and `validate_ancestors_with_witness` are near-line-by-line copies of the non-witness counterparts above. Every future change to one path now has to be mirrored in the other or the two will silently drift.

Two shapes to think about:
1. **Parameterize the existing path** with `compute_witness: bool` (you've already added that flag to `add_block_pipeline_inner`'s signature, this PR adds a parallel function instead of consolidating). The pipeline returns `Option<RpcExecutionWitness>` and the callers thread the flag through. Eliminates ~250 lines of duplicate code.
2. **Wrap `PayloadStatus` in a trait** with the witness as an associated type, so both code paths share the executor and only diverge at the response-building step.

(1) is simpler. The cost is the existing JSON-RPC path's hot path grows a never-true branch, but it's a single `if compute_witness && status == Valid { ... }` at the response-building site and the rest is identical. Non-blocking but the maintenance cliff at the next engine-API spec change is real.

crates/networking/rpc/engine/rest.rs
```diff
@@ -0,0 +1,437 @@
...
+fn ssz_response(resp: SszNewPayloadWithWitnessResponse) -> axum::response::Response {
+    let mut buf = Vec::with_capacity(resp.encoded_len());
+    resp.ssz_append(&mut buf);
+    let bytes = Bytes::from(buf);
+
+    Response::builder()
+        .status(StatusCode::OK)
+        .header(header::CONTENT_TYPE, "application/octet-stream")
+        .body(Body::from(bytes))
+        .expect("failed to build SSZ response")
```
ElFantasma, line 390: Panicking in an axum handler is a sharper edge than returning a 500 — it crashes the per-request task (and depending on tower configuration, can poison the connection). `Response::builder()` only fails if the header value is invalid, which for `application/octet-stream` it never is — but `expect` here makes that an unenforceable invariant rather than a checked one.

Map to an error response instead:
```rust
Response::builder()
    .status(StatusCode::OK)
    .header(header::CONTENT_TYPE, "application/octet-stream")
    .body(Body::from(bytes))
    .unwrap_or_else(|_| json_error_response(-32603, "failed to build SSZ response"))
```
Or factor through `IntoResponse` so the type system enforces it. Non-blocking — today's input is fixed and safe — but it's an `unsafe`-shaped surface in safe code.

crates/networking/rpc/engine/rest.rs
```diff
@@ -0,0 +1,437 @@
...
+    }
+}
+
+/// `POST /new-payload-with-witness-v4`
+pub async fn handle_new_payload_with_witness_v4(
+    State(context): State<RpcApiContext>,
+    auth_header: Option<TypedHeader<Authorization<Bearer>>>,
+    body: axum::body::Bytes,
+) -> axum::response::Response {
+    if let Err(_err) = authenticate(&context.node_data.jwt_secret, auth_header) {
```
ElFantasma, line 185: JWT auth is replicated inline at each handler (here at line 185, and again at line 271 in the v5 handler). #6710 / #6770 use an axum middleware (`engine_auth_middleware`) so every route in the engine-REST sub-router gets auth uniformly, without each handler having to remember.

For a 2-route PR this is fine. But if you keep the two-stacks-in-parallel design, the next REST endpoint added by someone unfamiliar with the auth contract is one missing line away from an unauthenticated endpoint on the engine API port. A `Router::new().route(...).route(...).layer(middleware::from_fn_with_state(secret, auth_middleware))` shape would close that gap. Same argument as #6710's auth design. Non-blocking.

ElFantasma (commented):
**Cross-PR**: #6710 (iovoid's first engine REST/SSZ API) was closed and replaced by the now-open #6770 — both attempt the broader SSZ-REST surface with proper SSZ extractors, axum-middleware JWT auth, and a `Content-Type` guard. This PR adds a *parallel* REST handler that duplicates JWT auth (line 173-176) + SSZ response building (line 381-392) inline instead of riding on top of that infrastructure. If #6770 lands first, you'll want to rebase this onto its `engine_rest` module rather than maintain a second REST stack. Worth syncing with @iovoid on which lands first and whether the foundations should merge.

PR #6743, opened 2026-05-28
chore(l1,l2): migrate tooling/ to ethrex-tooling repo

azteca1998:
## Summary

Supersedes #6487 — fresh branch from main implementing the same migration.

- Delete the in-repo `tooling/` directory (~130 files) and move all development tooling to [`lambdaclass/ethrex-tooling`](https://github.com/lambdaclass/ethrex-tooling)
- `ethrex-monitor` and `ethrex-repl` become git deps with `rev = "215859a3"` pinning
- Add `[patch."https://github.com/lambdaclass/ethrex"]` section so tooling's transitive deps resolve to local workspace crates
- All CI workflows checkout `ethrex-tooling` into `tooling/` before steps that need it
- Dockerfile drops `COPY tooling` and handles `cargo chef cook` failures gracefully
- Makefile removes tooling manifest entries from `update-cargo-lock` / `check-cargo-lock`
- `tooling/` added to `.gitignore` to prevent accidental commits of cloned checkout
- Docs and README updated to reference the ethrex-tooling repo
- `.cargo/config.toml.example` added for local cross-repo development

### Reviewer feedback addressed (from #6487)

| Reviewer | Feedback | Resolution |
|----------|----------|------------|
| iovoid | README should mention ethrex-tooling | Added Tooling section to README |
| iovoid | `tooling/` should be gitignored | Added to `.gitignore` |
| iovoid | Makefile and docs references | Updated Makefile, docs, and LEVM README |
| ElFantasma | `\|\| true` too broad in Dockerfile | Logs warning but still allows failure |
| ElFantasma | Use `rev` instead of `branch` | Using `rev = "215859a3"` |
| ElFantasma | `[patch]` section needs maintenance docs | Comment explains new crates must be added |
| ElFantasma/avilagaston9 | Stale path triggers in LEVM workflow | Removed `tooling/ef_tests/state` paths |

## Test plan

- [ ] `cargo build --workspace` passes
- [ ] Docker build succeeds: `docker build -t ethrex:local .`
- [ ] CI workflows pass (ethrex-tooling checkout steps are correct)
- [ ] Verify no `tooling/` directory in the repo (except as gitignored checkout target)

Closes #6487


edg-l:
Overall the migration is sound and the git-dep + `[patch]` approach is correct. A few things to address before merge, plus coordination with [#6725](https://github.com/lambdaclass/ethrex/pull/6725).

### 1. Pin to a real commit, not `main` — and bump it after [lambdaclass/ethrex-tooling#3](https://github.com/lambdaclass/ethrex-tooling/pull/3) merges
Cargo.toml pins `rev = "215859a3"` ([lambdaclass/ethrex-tooling#1](https://github.com/lambdaclass/ethrex-tooling/pull/1)), but every workflow checks out `ref: main`, which is already ahead. So the binary links monitor/repl from `215859a3` while CI builds/tests tooling from a newer `main` — they already disagree.

Concrete steps:
1. Merge [lambdaclass/ethrex-tooling#3](https://github.com/lambdaclass/ethrex-tooling/pull/3) first (it forwards `GIT_SHA`/`GIT_BRANCH`/`VERSION` to `docker build`, ported from [#6725](https://github.com/lambdaclass/ethrex/pull/6725)). That merge advances ethrex-tooling `main` to a new HEAD commit — call it `<NEW_SHA>`.
2. In this PR, set the monitor/repl git deps to `rev = "<NEW_SHA>"` (full 40-char SHA, not the short form).
3. Replace every workflow's `ref: main` with that same `<NEW_SHA>`.

Pinning to `<NEW_SHA>` guarantees the rev contains the monitor/repl crates *and* the docker_monitor fix, and that CI tests exactly what the binary links. Define the SHA once (shared workflow/repo variable) so the dep `rev` and the CI `ref` can't drift apart again.

### 2. Coordinate with [#6725](https://github.com/lambdaclass/ethrex/pull/6725) (Dockerfile modernization) — rebase after it merges
[#6725](https://github.com/lambdaclass/ethrex/pull/6725) rewrites the Dockerfile (BuildKit, cache mounts, multi-arch) and also touches `.dockerignore`, `.gitignore`, `Makefile`, and edits `tooling/sync/docker_monitor.py` (which this PR deletes). To keep its improvements, merge it first, then rebase this PR on top and apply these Dockerfile edits:
- **Delete** the two `COPY --link tooling/repl ./tooling/repl` and `COPY --link tooling/monitor ./tooling/monitor` lines in *both* the planner and builder stages — those crates are git deps now, so the COPY would fail on a missing path.
- **Keep** `ENV CARGO_NET_GIT_FETCH_WITH_CLI=true` — now load-bearing, since monitor/repl are fetched as rev-pinned git deps inside the container.
- **Keep** the `[patch."https://github.com/lambdaclass/ethrex"]` section so tooling's transitive `ethrex-*` deps resolve to the local workspace.
- That `[patch]` redirects to `./crates/...` which aren't present at the `cargo chef cook` stage, so cook can't fully resolve. Guard it:
  ```dockerfile
  cargo chef cook --profile $PROFILE --recipe-path recipe.json $BUILD_FLAGS \
    || echo "WARNING: cook skipped (git-patched deps need full tree); full build below reuses the target cache mount"
  ```
  The `|| echo` swallows all cook errors, but it's tolerable here because [#6725](https://github.com/lambdaclass/ethrex/pull/6725) mounts `target/` as a persistent cache (`--mount=type=cache,target=/ethrex/target`), so `cargo build` still gets incremental reuse even when cook is a no-op.
- `.dockerignore`: keep [#6725](https://github.com/lambdaclass/ethrex/pull/6725)'s `.git/` + `fixtures/*` excludes; replace its granular `tooling/ef_tests/...` lines with a blanket `tooling/`.
- `.gitignore` / `Makefile`: take [#6725](https://github.com/lambdaclass/ethrex/pull/6725)'s versions as the base, then re-apply this PR's tooling-path removals on top.
- `tooling/sync/docker_monitor.py`: resolve as **deleted** here — the file lives in ethrex-tooling now, and [#6725](https://github.com/lambdaclass/ethrex/pull/6725)'s edit to it is ported via [lambdaclass/ethrex-tooling#3](https://github.com/lambdaclass/ethrex-tooling/pull/3).

### 3. `[patch]` section has no guard
If a workspace crate used transitively by tooling is missing from the list, you get a silently divergent/duplicate build. Consider a small `cargo metadata` CI assertion to enforce it.

### 4. Local-dev UX
The Makefile still points at in-repo paths (`tooling/load_test/Cargo.toml`, `tooling/genesis`, `tooling/ef_tests/blockchain/.fixtures_url`). A dev who hasn't cloned ethrex-tooling into `tooling/` gets cryptic failures. A preflight check would help; non-blocking since docs are updated.

### 5. Ongoing maintenance note
Any ethrex crate API change now requires a lockstep `rev` bump of the tooling dep plus a matching CI `ref`. Worth calling out as a conscious two-repo workflow.

iovoid (changes requested):
`docs/developers/release-process.md` still references tooling/Cargo.lock

We still need to think about how tooling and other repos fit into the release process.

.github/workflows/pr-main_l1.yaml
```diff
@@ -20,6 +20,8 @@ env:
   # Work around frequent libgit2/submodule fetch flakiness in CI.
   CARGO_NET_GIT_FETCH_WITH_CLI: "true"
   CARGO_NET_RETRY: "10"
+  # ethrex-tooling commit; keep in sync with the Cargo.toml rev (docs/updating-ethrex-tooling.md)
+  TOOLING_REV: 20d259d2260824ba665fd5a01064ed3e83e21864
```
ElFantasma, line 24: `TOOLING_REV` is hardcoded here and in six other workflows (`pr-main_levm.yaml`, `pr_loc.yaml`, `pr_lint_license.yaml`, `pr_upgradeability.yaml`, `daily_hive_report.yaml`, `daily_loc_report.yaml`), for eight synchronization points total including Cargo.toml — but only the Makefile derives its rev from Cargo.toml via `sed -nE 's/.*ethrex-tooling".*rev = "([0-9a-f]+)".*/\1/p'`. If someone bumps Cargo.toml's rev but misses a workflow, local `make setup-tooling` and cargo builds pick up the new rev while CI's `tooling/` checkout stays on the old one — and nothing (including `check_tooling_patch.sh`, which only guards the patch section) will catch the divergence. Two low-cost options: extract the checkout into `.github/actions/checkout-ethrex-tooling/action.yaml` with one hardcoded rev, or read the rev from Cargo.toml at CI time (a one-line `grep`/`sed` in a step) so Cargo.toml stays the single source of truth.

iovoid:
Seems outdated. Also I think since it's a separate workspace, the patch directive can't reach it.

PR #6744, opened 2026-05-28
fix(l1): use 0x80 sentinel for missing eth/71 BAL per EIP-8159

edg-l:
## Summary

eth/71 `BlockAccessLists` (0x13) returns the wrong sentinel byte for missing BALs. EIP-8159 §\"BlockAccessLists (0x13)\" mandates the RLP empty string (\`0x80\`); our implementation emitted the RLP empty list (\`0xc0\`).

\`0xc0\` is a valid encoding for an empty \`BlockAccessList\` (a block with no state changes), so using it as the \"unavailable\" sentinel makes \"missing BAL\" indistinguishable from \"valid empty BAL\" on the wire. The spec uses \`0x80\` because it can never alias a real BAL.

## Why this matters

go-ethereum's eth/71 handler uses \`rlp.EmptyString\` (= \`0x80\`) at \`eth/protocols/eth/handlers.go:693\`, with the comment:

> The signal for missing BAL is the empty string, because an empty list is also a valid BAL.

With ethrex on \`0xc0\` and geth on \`0x80\`:

- **ethrex → geth**: geth decodes our \`0xc0\` as a zero-element list, interprets the slot as a valid empty BAL. Silent data confusion.
- **geth → ethrex**: ethrex's decode checks for \`0xc0\`, falls through to \`BlockAccessList::decode_unfinished\` which errors on \`0x80\`.

Found while cross-checking the snap/2 (EIP-8189) PR against go-ethereum.

## Changes

- \`crates/networking/p2p/rlpx/eth/block_access_lists.rs\`: \`OptionalBal\` encode and decode now use \`0x80\`.
- Added \`block_access_lists_none_uses_0x80_sentinel\` test asserting \`0x80\` is present and \`0xc0\` absent in a single-None response payload.

## Test plan

- [x] \`cargo test -p ethrex-p2p --lib block_access_lists\` — 8 passed (7 existing + 1 new).
- [x] \`cargo test -p ethrex-p2p --lib\` — 35 passed.
- [x] \`cargo clippy -p ethrex-p2p -- -D warnings\` — clean.
- [x] \`cargo fmt --all\` — clean.

### DIFF
--- a/crates/networking/p2p/rlpx/eth/block_access_lists.rs
+++ b/crates/networking/p2p/rlpx/eth/block_access_lists.rs
@@ -23,9 +23,9 @@ pub const BLOCK_ACCESS_LIST_LIMIT: usize = 1024;
 /// empty string is the only sentinel that can never alias a real BAL.
 /// `Some(bal)` is encoded as the BAL's normal RLP list encoding.
 ///
-/// INVARIANT: `BlockAccessList` always encodes as an RLP list (first byte
-/// >= 0xc0), so `0x80` is unambiguously the `None` sentinel; keep this true
-/// if `BlockAccessList`'s encoding is ever refactored.
+/// INVARIANT: `BlockAccessList` always encodes as an RLP list (first byte is
+/// `0xc0` or greater), so `0x80` is unambiguously the `None` sentinel; keep
+/// this true if `BlockAccessList`'s encoding is ever refactored.
 #[derive(Debug, Clone)]
 struct OptionalBal(Option<BlockAccessList>);
 
--- a/crates/networking/p2p/rlpx/eth/block_access_lists.rs
+++ b/crates/networking/p2p/rlpx/eth/block_access_lists.rs
@@ -26,8 +26,12 @@ pub const BLOCK_ACCESS_LIST_LIMIT: usize = 1024;
 /// INVARIANT: `BlockAccessList` always encodes as an RLP list (first byte is
 /// `0xc0` or greater), so `0x80` is unambiguously the `None` sentinel; keep
 /// this true if `BlockAccessList`'s encoding is ever refactored.
+///
+/// Public so the byte-level sentinel encoding can be asserted from the test
+/// crate (`test/tests/p2p/rlpx/block_access_lists_tests.rs`); message-level
+/// roundtrips run the bytes through snappy and can't see the raw `0x80`.
 #[derive(Debug, Clone)]
-struct OptionalBal(Option<BlockAccessList>);
+pub struct OptionalBal(pub Option<BlockAccessList>);
 
 impl RLPEncode for OptionalBal {
     fn encode(&self, buf: &mut dyn BufMut) {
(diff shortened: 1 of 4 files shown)


ilitteri (approved):

crates/networking/p2p/rlpx/eth/block_access_lists.rs
```diff
@@ -244,4 +249,31 @@ mod tests {
...
         assert_eq!(decoded.block_access_lists.len(), BLOCK_ACCESS_LIST_LIMIT);
     }
+
+    /// Locks EIP-8159 §"BlockAccessLists (0x13)" sentinel: missing BALs encode
+    /// as RLP empty string (`0x80`), NOT empty list (`0xc0`). Empty list is a
+    /// valid BAL encoding (block with no state changes), so the empty string
+    /// is the only byte that can't alias a real BAL. geth uses the same
+    /// sentinel (`rlp.EmptyString` in `eth/protocols/eth/handlers.go`); any
+    /// drift here is silent interop breakage.
+    #[test]
```
iovoid: This should be on the test/ folder

iovoid (changes requested):

crates/networking/p2p/rlpx/eth/block_access_lists.rs
```diff
@@ -244,4 +249,31 @@ mod tests {
...
+        let msg = BlockAccessLists::new(0, vec![None]);
+        let mut buf = Vec::new();
+        msg.encode(&mut buf).unwrap();
+        let decompressed = snappy_decompress(&buf).unwrap();
+        assert!(
+            decompressed.contains(&0x80),
+            "decompressed payload must contain the 0x80 None sentinel"
+        );
+        assert!(
+            !decompressed.contains(&0xc0),
```
ElFantasma: `!decompressed.contains(&0xc0)` is correct *today* but happens to hold via RLP-encoding luck. The outer `[request_id, [bals]]` wrapper is encoded with a `0xc0 + length` prefix; for `(request_id=0, bals=[None])` the lengths come out to 0xc4 / 0xc1 (not 0xc0). The moment someone changes the test fixture in a way that makes the *outer* list have exactly zero payload bytes, the prefix becomes 0xc0 and the test fails spuriously even though the sentinel is right.

More robust assertion: target `OptionalBal` directly rather than the whole message.
```rust
let mut bytes = Vec::new();
OptionalBal(None).encode(&mut bytes);
assert_eq!(bytes, vec![0x80]);
```
Keeps the intent ("None encodes as exactly 0x80") and survives unrelated encoding changes elsewhere.

Non-blocking — the current test is correct, just brittle.

crates/networking/p2p/rlpx/eth/block_access_lists.rs
```diff
@@ -16,30 +16,34 @@ use ethrex_rlp::{
...
+            None => 1, // empty string = 0x80 per EIP-8159
             Some(bal) => bal.length(),
         }
     }
 }
 
 impl RLPDecode for OptionalBal {
     fn decode_unfinished(rlp: &[u8]) -> Result<(Self, &[u8]), RLPDecodeError> {
-        if rlp.first() == Some(&0xc0) {
+        if rlp.first() == Some(&0x80) {
```
ElFantasma, line 54: Just confirming this is now consistent with the snap/2 codec in #6544 — that PR's `Snap2OptionalBal::decode_unfinished` does the same `rlp.first() == Some(&0x80)` check (codec.rs:376 there). I left an inline on the snap/2 side asking for an invariant comment about why `0x80` is unambiguous (BAL is always RLP-encoded as a list, so `>= 0xc0`). The same comment is worth landing here so a future refactor of `BlockAccessList`'s encoding doesn't silently break this decoder.

One-line suffix on the existing comment block at lines 19–24 is enough — something like "INVARIANT: `BlockAccessList` always encodes as an RLP list (first byte ≥ 0xc0), so `0x80` is unambiguously the None sentinel." Non-blocking.

ElFantasma (approved):
Just two nits

iovoid (approved):

PR #6751, opened 2026-05-29
fix(levm): keep system-call storage write when coinbase is the called contract

ElFantasma:
## Summary

`generic_system_contract_levm` snapshots the block `coinbase` account before a system call (EIP-2935 / 4788 / 7002 / 7251) and restores it afterwards, so the call leaves no fee/value side effect on the fee recipient. That restore was unconditional.

When a block's fee recipient (`coinbase`) is set to the system contract being called, the restore reverts that contract's account to its pre-call snapshot — dropping the storage write the system call just made (e.g. the EIP-2935 parent-hash slot, or an EIP-7002/7251 request) from the emitted state updates. The result is a post-state root that diverges from other clients for an otherwise-valid block, since `feeRecipient` is freely chosen by the block producer.

## Fix

Skip the coinbase restore when `block_header.coinbase == contract_address`. In that case the only change to the account is the system call's own storage write, which must persist; other clients commit the system-call diff and never restore the callee. Ordinary fee recipients are unaffected. A single guard covers all four system contracts, since they share this helper.

## Test

Adds `system_call_coinbase_tests`, which drives the real EIP-2935 system call and asserts the history-contract storage write is emitted both for an ordinary fee recipient and when the fee recipient equals the history contract (the regression case). Verified the regression test fails on the pre-fix code.


JereSalo (commented):
Good news, we don't actually need a backup for any of the accounts. The only change that we have to revert when calling a system contract is the nonce incremental that is done to the `SYSTEM_ADDRESS` and happens on every transaction independently of its outcome, but there's nothing more that needs to be restored (thanks to gas price being set to zero).
I'd lean towards simplifying things here and forget about the backups entirely and after execution we can just decrement the nonce by one to the sender and that's it. Also, there's no need to remove something from the cache (`current_accounts_state`) if it wasn't there in the first place, it doesn't produce any harm not to do so, and in fact it may even be beneficial performance-wise (though probably negligible, so it's just a side-effect) if we reuse the same cached accounts for executing multiple blocks because we are not going to be removing accounts from it unnecessarily and reloading them on every block.


The fix I propose is simply removing the backup of both coinbase and system account and just add these lines at the end to decrease the nonce of the sender, that's it:
```rust
if let Some(account) = db.current_accounts_state.get_mut(&system_address) {
    account.info.nonce = account.info.nonce.saturating_sub(1);
}
```

I made the change in another branch in case you want to check it out. Feel free to incorporate it if you agree on the approach:
https://github.com/lambdaclass/ethrex/compare/fix/syscall-coinbase-collision...refactor/syscall-nonce-decrement

ElFantasma:
Thanks Jere — great call, this is a much cleaner fix than my guard. Incorporated your commit as-is (kept your authorship): cherry-picked it on top after merging `main`, so the branch now drops both backups and just decrements the `SYSTEM_ADDRESS` nonce.

I confirmed it against the regression test I'd added (`system_call_coinbase_tests`) and it stays green — the EIP-2935 history write survives in both the ordinary-coinbase and `coinbase == history contract` cases. And I double-checked the reasoning holds: with `gas_price = 0` there's no coinbase fee and no sender charge, so the nonce bump is the only residue; and `get_state_transitions` already filters empty/unchanged accounts (the EIP-161 handling in `gen_db.rs`), so the old "remove if not cached" branches were redundant — leaving an empty `SYSTEM_ADDRESS` or a warmed coinbase in the cache produces no trie change.

One small thing I added on top (`425e09466`): a `debug_assert!` right after the env is built, asserting `gas_price`/`base_fee_per_gas` are zero, with a comment that the post-call cleanup depends on it. Since we're now leaning on the zero-price assumption (the backups used to tolerate any value), I figured it's worth making that invariant explicit so it can't silently regress. Happy to drop it if you think it's overkill.

JereSalo (approved):
Great, perhaps in the future we could drop the tests added here in case tests with same behavior are added to the [execution-specs](https://github.com/ethereum/execution-specs) test suite. In the meantime it might be okay to keep them here.

iovoid (approved):

MegaRedHand (approved):
LGTM

crates/vm/backends/levm/mod.rs
```diff
@@ -2538,6 +2533,15 @@ pub fn generic_system_contract_levm(
...
     };
 
+    // Invariant relied upon below: with a zero gas price a system call charges no
+    // gas to the SYSTEM_ADDRESS sender and pays no fee to the coinbase, so the only
+    // state change left to undo afterwards is the sender's nonce bump. If this ever
+    // becomes non-zero, the post-call cleanup must be revisited.
+    debug_assert!(
+        env.gas_price.is_zero() && env.base_fee_per_gas.is_zero(),
+        "system calls must run with a zero gas price"
+    );
```
MegaRedHand, line 2543: Tests make this redundant, I think

PR #6752, opened 2026-05-29
fix(l1): reject L2-only transaction types (0x7d/0x7e) in L1 blocks

ElFantasma:
## Summary

ethrex uses one shared `Transaction` decoder for L1 and L2. It accepts the L2-only transaction types `0x7d` (`FeeToken`) and `0x7e` (`PrivilegedL2Transaction`) with no L1/L2 gate, and a privileged transaction's sender is its unsigned, caller-chosen embedded `from` (no signature recovery). Nothing on the L1 import path (`validate_block_pre_execution`, `validate_block_body`, the default LEVM hook) rejects these types, and `engine_newPayload` decodes through the same path.

As a result, an L1 block containing a type-`0x7e` transaction is decoded, accepted, and executed — using the attacker-chosen `from` as the sender. geth/reth and the execution spec reject these types outright, so ethrex would both diverge consensus (accept a block the rest of the network rejects) and let a block forge a sender on L1.

## Fix

Reject L2-only transaction types on the L1 import path:

- `TxType::is_l2_only()` (in `crates/common/types/transaction.rs`) classifies each transaction type. The match is **exhaustive (no wildcard)**, so adding a new `TxType` variant won't compile until it is explicitly classified — an L2-only type can never be silently accepted on L1 by omission.
- `Blockchain::validate_l1_transaction_types`, gated on `BlockchainType::L1`, rejects any block transaction whose type is L2-only. It is called right after `validate_block_pre_execution` at all three pre-execution sites (`execute_block`, `execute_block_pipeline`, `execute_block_from_state`), so every L1 import/execute path is covered.
- New error `InvalidBlockError::UnsupportedTransactionType(u8)`.

L2 is unaffected (the check early-returns when not `BlockchainType::L1`); the zkVM prover and L2 sequencer paths are untouched.

## Test

Adds `l1_rejects_privileged_l2_transaction_block`, which builds an L1 block containing a type-`0x7e` transaction (post-state consistent so it reaches type acceptance) and asserts `add_block` rejects it. Confirmed red on the pre-fix code (`add_block` returned `Ok(())`) and green after the fix.

## Scope

This closes the consensus / block-import vector. The mempool/RPC ingress (`eth_sendRawTransaction` admitting these types on L1) is tracked separately under the `fee-token-l1-tx` finding.


crates/blockchain/blockchain.rs
```diff
@@ -3426,3 +3448,173 @@ fn collect_trie(index: u8, mut trie: Trie) -> Result<(Box<BranchNode>, Vec<TrieN
     };
     Ok((root, nodes))
 }
+
+#[cfg(test)]
+mod l1_transaction_type_tests {
```
iovoid: Should be on the test/ folder

crates/blockchain/blockchain.rs
```diff
@@ -403,6 +403,25 @@ impl Blockchain {
         }
     }
 
+    /// L1 blocks must not contain L2-only transaction types (`FeeToken` 0x7d,
+    /// `Privileged` 0x7e). Those carry an unsigned, caller-chosen `from` (no signature
```
iovoid: nit: only Privileged has no signature check

iovoid (changes requested):

iovoid (approved):

azteca1998 (approved):

edg-l (approved):
LGTM. Verified the L1 import path is fully covered:

- `is_l2_only()` is exhaustive (no wildcard) over all 7 `TxType` variants — a new variant won't compile until classified.
- Guard is gated on `BlockchainType::L1` and early-returns on L2, so L2 is untouched.
- All three L1 execution functions are guarded, and every import entrypoint funnels through one of them: `add_block`→`execute_block`, `add_block_pipeline_inner`→`execute_block_pipeline`, `add_blocks_in_batch`→`execute_block_from_state`. `engine_newPayload` goes through `add_block`/pipeline.
- The only other `validate_block_pre_execution` callers are the zkVM guest program and the L2 sequencer — both correctly out of scope.
- Guard runs before `vm.execute_block`, so a forged-sender privileged tx never executes.
- Regression test round-trips through `decode_canonical` and builds a post-state consistent with execution, so pre-fix the block fully passed (red), making the type guard the only new rejection post-fix.

Minor optional note: the guard is duplicated at three call sites rather than centralized. Reasonable given `validate_block_pre_execution` is a free fn without `BlockchainType` access; could fold it in there later for future-path safety, but not a blocker.

crates/blockchain/blockchain.rs
ElFantasma: Done — moved to `test/tests/blockchain/l1_tx_type_tests.rs`. No inline `#[cfg(test)]` left in `blockchain.rs`.

crates/blockchain/blockchain.rs
ElFantasma: Right — the doc comment now scopes that to `Privileged` only: "`Privileged` **additionally** takes its sender from an unsigned, caller-chosen `from` (no signature recovery)". `FeeToken` is rejected purely as an unknown L1 type.

crates/blockchain/blockchain.rs
```diff
@@ -403,6 +403,25 @@ impl Blockchain {
...
+        if !matches!(self.options.r#type, BlockchainType::L1) {
+            return Ok(());
+        }
+        for tx in &block.body.transactions {
+            if tx.tx_type().is_l2_only() {
+                return Err(ChainError::InvalidBlock(
+                    InvalidBlockError::UnsupportedTransactionType(tx.tx_type() as u8),
+                ));
+            }
+        }
```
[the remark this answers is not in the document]
ElFantasma: Leaving as-is: `TxType` is `Copy`, so the second call is a trivial field read with no allocation or perf cost, and the two uses read clearly in place (predicate vs. error payload). Not worth a re-push + full CI cycle on an approved PR for a cosmetic binding.

ElFantasma:
@edg-l on the optional centralization note: keeping the guard as a `Blockchain` method called at the three execution entrypoints rather than folding it into `validate_block_pre_execution`.

`validate_block_pre_execution` is a free fn with no `BlockchainType` access, and it's also called by the zkVM guest program and the L2 sequencer — both out of scope for this L1-only rejection. Centralizing there would mean threading a `BlockchainType`/`is_l1` flag through its signature and updating those out-of-scope callers to opt out, spreading the L1/L2 distinction into shared (incl. `no_std` guest) code. The method form already has clean `BlockchainType` access, and you verified all current import paths funnel through the three guarded functions. If a future execution path makes this a real gap, a compiler-enforced single chokepoint would be the better fix than param-threading — happy to do that as a separate refactor if/when it comes up.

PR #6753, opened 2026-05-29
fix(l1): bound EIP-7702 authorization y_parity to a single byte at decode

ElFantasma:
## Summary

EIP-7702 bounds an authorization tuple's `y_parity` to `< 2**8` (it must fit in a single byte); geth models it as a `uint8` and rejects an out-of-range value at RLP decode, which makes the whole transaction — and any block containing it — invalid.

ethrex typed `y_parity` as `U256` and decoded it with no bound, so a value like `256` decoded successfully. At execution `eip7702_recover_address` returns `Ok(None)` for any non-0/1 `y_parity` and the authorization tuple is silently skipped (the transaction stays valid). The result: a type-4 transaction with `y_parity >= 2**8` is accepted and executed by ethrex but rejected at decode by other clients → an L1 block carrying it splits ethrex from the network.

Note the bound is purely a field-size rule: `y_parity` values in `2..=255` are in range, fail signature recovery, and are skipped (tuple ignored, tx valid) — matching geth. Only `>= 2**8` diverges, and only that is changed here.

## Fix

Keep the field as `U256` and add an explicit `y_parity > u8::MAX` rejection at RLP decode (`AuthorizationTuple::decode_unfinished` in `crates/common/types/tx_fields.rs`):

```rust
let (y_parity, decoder) = decoder.decode_field("y_parity")?;
if y_parity > U256::from(u8::MAX) {
    return Err(RLPDecodeError::Custom(
        "EIP-7702 authorization y_parity must be < 2**8".to_string(),
    ));
}
```

This is the minimal change: it adds a value-range check on a field whose representation is unchanged, so the **JSON-RPC** (hex quantity `"0x1"`), **RLP** and **rkyv** wire formats stay byte-for-byte identical. Canonical-RLP behavior is already provided by the `U256` decoder (it rejects leading-zero scalars such as `0x8100`), so no change is needed there.

> An earlier revision typed the field as `u8` to make the bound a type invariant. That is arguably the cleaner model, but `AuthorizationTuple` is deserialized directly from JSON on the `eth_sendTransaction` path (where `y_parity` is a hex quantity) and is part of the rkyv prover input — a bare `u8` changes both surfaces. Per review (@iovoid) this PR keeps `U256` and bounds at decode instead, which fixes the consensus issue with zero wire-format churn.

The `2..=255`→skip behavior is unchanged.

## Test

`test/tests/common/eip7702_authorization_tests.rs` pins the surfaces a type change would have silently affected, plus the canonical-RLP guarantee:

- `rlp_rejects_y_parity_at_or_above_256` — hand-builds the RLP of a tuple with `y_parity = 256` and asserts decode rejects it (the chokepoint a type-4 tx's `authorization_list` decodes through). **Red** on pre-fix code, **green** after.
- `rlp_accepts_in_range_y_parity` — `0`, `1`, `255` still decode and round-trip.
- `rlp_rejects_noncanonical_y_parity` — non-minimal `0x8100` encoding of `0` is rejected (canonical-RLP, provided by `U256`).
- `json_deserializes_y_parity_as_hex_quantity` — `AuthorizationTuple` deserializes from `{"yParity":"0x1",...}` to `U256::one()`, pinning the JSON-RPC surface kept intact by staying `U256`.


ElFantasma:
Ran the EIP-7702 Hive EELS suite against this branch (image built from the branch):

**`ethereum/eels/consume-rlp` — `.*7702.*` → 1210 passed / 0 failed** (Prague + Osaka, `tests/prague/eip7702_set_code_tx/...`).

This is the RLP block-import path that this fix touches. Notable coverage in the run:
- Invalid-authorization-tuple *skip* semantics — `test_valid_tx_invalid_auth_signature`, `test_set_code_all_invalid_authorization_tuples`, `test_set_code_multiple_valid_authorization_tuples_first_invalid_same_signer`, `test_account_warming[...single_invalid_authorization...]`. These passing confirms the change does **not** regress the "skip the invalid tuple, keep the tx" behavior for in-range, non-0/1 `y_parity` (the `2..=255` case we intentionally leave alone).
- `test_signature_s_out_of_range` — the field-range analogue of this fix for the `s` field — passes.
- Broad 7702 conformance (delegation, gas, nonce validity, chain-id, precompiles) intact.

Note: v5.3.0 doesn't appear to ship a fixture named specifically for `y_parity >= 2**8` rejection, so the exact out-of-bounds decode rejection is covered by the unit test added here; Hive confirms surrounding conformance and no regression.

Also running `consume-engine` (`newPayload` path); will follow up with that result.

ElFantasma:
Follow-up: `ethereum/eels/consume-engine` — `.*7702.*` → **1210 passed / 0 failed** (engine `newPayload` path). Both consensus entry points (RLP import + newPayload) are green for the EIP-7702 fixtures with this branch.

crates/common/types/tx_fields.rs
```diff
@@ -41,8 +41,9 @@ pub struct AuthorizationTuple {
         deserialize_with = "crate::serde_utils::u64::deser_hex_or_dec_str"
     )]
     pub nonce: u64,
-    #[rkyv(with = crate::rkyv_utils::U256Wrapper)]
-    pub y_parity: U256,
+    // EIP-7702 bounds y_parity to < 2**8; a u8 makes that a type invariant and lets
+    // RLP decoding reject any out-of-range value, matching geth's `uint8`.
+    pub y_parity: u8,
```
iovoid: Should 0x00, 0x80 and 0x8100 all be considered valid encodings for y_parity=0?

iovoid (changes requested):
I think the new encoding might break deserialization in JSON endpoints using it (such as ethrex_sendTransaction)

crates/common/types/tx_fields.rs
ElFantasma: Only `0x80` (the empty string) is valid for `y_parity = 0`. The other two are non-minimal scalar encodings and are rejected by the `U256` decoder before this bound check even runs:

- `0x80` → empty string → `U256(0)` ✅ (canonical)
- `0x00` → 1-byte string `[0x00]` → `static_left_pad` rejects leading-zero bytes → `MalformedData` ❌
- `0x8100` → 1-byte string `[0x00]` → same rejection ❌

So canonical-RLP is already enforced by `U256`; the new check only adds the *value-range* bound (`> 2**8`). The `rlp_rejects_noncanonical_y_parity` test pins the `0x8100` case so it cannot regress.

ElFantasma:
Reworked to address the JSON-deserialization concern. Switched from the `u8` typing (Option A) to **keeping `U256` and bounding at decode** (Option B):

```rust
let (y_parity, decoder) = decoder.decode_field("y_parity")?;
if y_parity > U256::from(u8::MAX) {
    return Err(RLPDecodeError::Custom(
        "EIP-7702 authorization y_parity must be < 2**8".to_string(),
    ));
}
```

This was exactly right — `AuthorizationTuple` is deserialized directly from JSON on the `ethrex_sendTransaction` path (where `y_parity` is a hex quantity like `"0x1"`), and a `u8` field would reject that. It is also part of the rkyv prover input. Keeping `U256` leaves the **JSON-RPC, RLP and rkyv** wire formats byte-for-byte identical; only an out-of-range *value* is now rejected.

Added `test/tests/common/eip7702_authorization_tests.rs` pinning all the surfaces a representation change would have touched:
- `rlp_rejects_y_parity_at_or_above_256` (the fix; red→green)
- `rlp_accepts_in_range_y_parity` (0/1/255 round-trip)
- `rlp_rejects_noncanonical_y_parity` (`0x8100` → rejected)
- `json_deserializes_y_parity_as_hex_quantity` (`"0x1"` → `U256::one()`, pinning the JSON surface)

The greptile note about the opaque `TryFrom … Error = String` is moot now — Option B drops the `TryFrom` entirely and keeps the original `U256` field, so no conversion error type is introduced.

azteca1998 (approved):

iovoid (approved):

edg-l (approved):

PR #6755, opened 2026-05-30
fix(l1): reject missing withdrawals when withdrawals root is present

### LINKED ISSUE #6754: Ethrex accepts a block body with missing withdrawals when withdrawals_root is the empty withdrawals root
## Summary

`validate_block_body` currently accepts blocks where:

- `header.withdrawals_root = Some(EMPTY_WITHDRAWALS_HASH)`
- `body.withdrawals = None`

This effectively treats a missing withdrawals field as equivalent to an empty withdrawals list.

However, this behavior appears to differ from other execution clients and may allow malformed Shanghai-era block bodies to pass validation.

## Current behavior

In `validate_block_body`:

```rust
(Some(withdrawals_root), None) => {
    if withdrawals_root != *EMPTY_WITHDRAWALS_HASH {
        return Err(InvalidBlockBodyError::WithdrawalsRootNotMatch);
    }
}
```

This means the following combination is considered valid:
- `header.withdrawals_root = Some(EMPTY_WITHDRAWALS_HASH)`
- `body.withdrawals = None`

## Why this may be problematic

The presence of `EMPTY_WITHDRAWALS_HASH` only indicates that the withdrawals root corresponds to an empty withdrawals list.

It does not prove that the withdrawals field itself is present in the block body.

Conceptually, these represent different states:
```
Some(vec![])  -> withdrawals field exists and is empty
None          -> withdrawals field is missing
```

Treating them as equivalent makes it impossible to distinguish between:

1. A valid Shanghai block with an empty withdrawals list.
2. A malformed or incomplete block body where the withdrawals field is absent.

## Comparison with other clients

Both Geth and Reth appear to require the withdrawals field to be present whenever a withdrawals root exists in the header.
```
header.withdrawals_root = Some(...)
body.withdrawals = None
```
is rejected, regardless of whether the root equals the empty withdrawals root.

## Reproduction
The fixture contains:
```
withdrawals_root = EMPTY_WITHDRAWALS_HASH
withdrawals = null
```
When executed against multiple clients:
```
Geth and Reth rejects
Ethrex accepts

### PR DESCRIPTION
**Pull request by emirongrr**
**Motivation**

While running a custom Hive fixture across multiple Ethereum execution clients, I observed a behavioral difference between Ethrex, Geth, and Reth.

The fixture constructs a block where the header contains a withdrawals root equal to EMPTY_WITHDRAWALS_HASH, while the block body does not contain a withdrawals field.

Ethrex accepted the block, whereas both Geth and Reth rejected it.

current imp:
```rust
    match (block_header.withdrawals_root, &block_body.withdrawals) {
        (Some(withdrawals_root), Some(withdrawals)) => {
            let computed_withdrawals_root = compute_withdrawals_root(withdrawals, crypto);
            if withdrawals_root != computed_withdrawals_root {
                return Err(InvalidBlockBodyError::WithdrawalsRootNotMatch);
            }
        }
        (Some(withdrawals_root), None) => {
            if withdrawals_root != *EMPTY_WITHDRAWALS_HASH {
                return Err(InvalidBlockBodyError::WithdrawalsRootNotMatch);
            }
        }
        (None, None) => {}
        _ => return Err(InvalidBlockBodyError::WithdrawalsRootNotMatch),
    }
```
The current validation logic treats a missing withdrawals field (None) as equivalent to an empty withdrawals list when the header contains `EMPTY_WITHDRAWALS_HASH.` This makes it impossible to distinguish between an empty withdrawals list and a missing withdrawals field.

**Description**
This PR updates block body validation to require the withdrawals field to be present whenever a withdrawals root is present in the block header.

More specifically:

- Rejects blocks where withdrawals_root exists in the header but withdrawals is missing from the body.
- Preserves support for empty withdrawals lists by validating them through the normal withdrawals root computation path.

Maintains a clear distinction between:
- Some(vec![]) (empty withdrawals list)
- None (missing withdrawals field)

This behavior aligns Ethrex with the behavior observed in Geth and Reth when validating the same Hive fixture.

**Checklist**

If this behavior is intentional or aligns with Ethrex's design goals, please feel free to disregard this change. Likewise, if there is a more appropriate way to handle this validation path, I would greatly appreciate any feedback or suggestions.

Closes #6754


crates/common/types/block.rs
```diff
@@ -646,6 +647,10 @@ pub enum InvalidBlockHeaderError {
 pub enum InvalidBlockBodyError {
     #[error("Withdrawals root does not match")]
     WithdrawalsRootNotMatch,
+    #[error("Withdrawals are not present")]
+    WithdrawalsNotPresent,
```
ElFantasma: nit: `WithdrawalsPresent` / `WithdrawalsNotPresent` read ambiguously without context — present *where*? When an operator hits one of these in a log they have to cross-reference the error variant against the function to understand the direction of the mismatch.

More self-documenting names would tie each variant to which side of the (header, body) pair is the unexpected one:
```rust
WithdrawalsMissingFromBody,   // (Some root in header, None in body)
WithdrawalsUnexpectedInBody,  // (None in header, Some list in body)
```
And for the header-side variants at line 629-631:
```rust
WithdrawalsRootMissingPostShanghai,  // (Shanghai active, header.withdrawals_root.is_none())
WithdrawalsRootUnexpectedPreShanghai, // (Pre-Shanghai, header.withdrawals_root.is_some())
```
The match-arm comments in `validate_block_body` already spell this out — the variant names just have to follow. Non-blocking; the current names are correct, just less helpful to a tired operator at 3 a.m.

azteca1998 (approved):

PR #6760, opened 2026-06-01
fix(l2): restrict TDX signer registration to timelock

Radovenchyk:
Restrict TDX signer registration to timelock-controlled governance flow instead of allowing any sequencer to replace the global trusted signer

Note: TDX quote freshness constraints may make the standard delayed timelock flow impractical for production signer registration, since quotes can expire before execution. In practice, signer rotation may need to go through the Security Council emergency path or a dedicated governance-controlled registration mechanism designed around quote validity windows.


crates/l2/tee/contracts/src/TDXVerifier.sol
```diff
@@ -67,10 +67,9 @@ contract TDXVerifier {
     /// @param quote The TDX quote, which includes the address being registered
     function register(bytes calldata quote) external {
         require(
-            timelock.isSequencer(msg.sender),
-            "TDXVerifier: only sequencer can update keys"
+            msg.sender == address(timelock),
```
ElFantasma: **This breaks the existing off-chain registration path.** `crates/l2/sequencer/setup.rs:50` (`register_tdx_key`) calls `register(quote)` directly from a sequencer-owned `private_key` (signer constructed at line 86). Post-merge, that call always reverts with the new `"only timelock can update keys"` message, so a sequencer can no longer publish its TDX quote — the chain doesn't end up with an `authorizedSignature`, and downstream `verify(...)` keeps rejecting.

The PR description honestly flags that the standard timelock delay isn't operationally viable for TDX quote freshness, but doesn't propose the replacement path. As a result this PR lands a strict-on-paper restriction at the cost of a regression in the working flow.

Three shapes that would unblock this:
1. **Pair with the replacement**: land the Security-Council-emergency or dedicated-governance path in the same PR (and update `register_tdx_key` to use it).
2. **Adapter pattern**: keep the new check, but have `register_tdx_key` build a timelock proposal + execute via the timelock's emergency function (whatever exists today).
3. **Phase the change**: introduce a constructor / setter flag (`bool requireTimelock`) that defaults to the *old* permissive check; flip it on after the off-chain path is updated. Less ideal because it leaves the laxer gate available, but it's a clean staging mechanism.

I'd hold the PR on this — the security property is desirable, but shipping it ahead of the replacement gates the L2's signer-rotation operationally. Happy to be wrong if the off-chain side is already updated elsewhere and I missed the reference.

crates/l2/tee/contracts/src/TDXVerifier.sol
```diff
@@ -67,10 +67,9 @@ contract TDXVerifier {
...
+            "TDXVerifier: only timelock can update keys"
```
ElFantasma: A Solidity access-control change is exactly the kind of fix that needs a test, especially since there are none under `crates/l2/tee/contracts/`. Two cases minimally:

1. `register(quote)` called from a non-timelock EOA reverts with the new message.
2. `register(quote)` called from the timelock succeeds in dev-mode (`isDevMode = true` skips quote verification so the test is cheap).

A third case worth pinning: confirm a *previously-authorized sequencer* who is no longer the timelock now reverts — this is the regression we're guarding against.

Foundry under `crates/l2/contracts/` (where the existing L1 tests live) is the natural home for these. Non-blocking but the security property of this PR is exactly what tests guard against silent regression.

crates/l2/tee/contracts/src/TDXVerifier.sol
Radovenchyk: Thanks for catching the regression. Reworked the fix, instead of routing through timelock directly, added a dedicated authorizedSequencer that's set at deploy time and rotatable via timelock. This keeps the automated prover flow working while closing the inter-sequencer overwrite vector

crates/l2/tee/contracts/src/TDXVerifier.sol
Radovenchyk: done

ElFantasma (commented):
The `authorizedSequencer` rework is exactly the right shape — deploy-time set + timelock-gated rotation closes the inter-sequencer overwrite vector while keeping `register_tdx_key` working. Thanks for the careful redesign.

One follow-up on the test inline: I see `done` in the reply but the current diff only touches `Makefile` and `TDXVerifier.sol` — no test files. Did the tests land in a separate commit / different branch that didn't push through, or is that a deferred follow-up? Three cases worth pinning before merge:

1. Non-`authorizedSequencer` reverts with the new message.
2. `authorizedSequencer` succeeds in dev-mode (`isDevMode = true` skips quote verification, makes the test cheap).
3. `setAuthorizedSequencer` rotation works when called from timelock, reverts otherwise.

The Solidity access-control gate is the entire security property of this PR — having it under test protects against a future refactor silently dropping the check.

Happy to approve once tests land (or if you'd rather defer them to a follow-up, a tracking issue link + a CHANGELOG entry would do).

ElFantasma (approved):
The `authorizedSequencer` rework is the right shape — deploy-time set + timelock-gated rotation closes the inter-sequencer overwrite vector without breaking the off-chain `register_tdx_key` path. And thanks for going further than the defer-to-a-tracking-issue path I'd  offered: bootstrapping the Foundry harness and landing all four access-control cases means the security property is under test from day one. 

LGTM.

Radovenchyk:
Hey @ElFantasma сan you push this commit?

Radovenchyk:
Hey @ilitteri Could you take a look and approve it?

PR #6761, opened 2026-06-01
fix(l1): reject forkchoice v3 payload attributes after Amsterdam

### LINKED ISSUE #6874: Previous-fork Engine API methods do not reject Amsterdam-active timestamps
## Summary

The Amsterdam Engine API specification requires previous-fork Engine API methods to reject Amsterdam-active payload timestamps.

The existing validation checked that requests were past the required fork activation (Cancun or Osaka), but did not yet enforce the Amsterdam upper boundary introduced by the Amsterdam Engine API specification.

As a result, requests that should be rejected with `-38005: Unsupported fork` may be accepted after Amsterdam activation.

## Specification

From [`amsterdam.md`](https://github.com/ethereum/execution-apis/blob/main/src/engine/amsterdam.md#update-the-methods-of-previous-forks):

> For `engine_newPayloadV4`, `engine_getPayloadV5`, and `engine_forkchoiceUpdatedV3`, client software MUST return `-38005: Unsupported fork` if the timestamp of payload is greater than or equal to the Amsterdam activation timestamp.

This means previous-fork methods are only valid before Amsterdam becomes active.

### engine_forkchoiceUpdatedV3

`validate_attributes_v3` only checked the Cancun lower boundary.

The valid range should be:

```text
Cancun <= timestamp < Amsterdam
```

### engine_getPayloadV5

`engine_getPayloadV5` only checked that the payload timestamp was Osaka-active.

The valid range should be:

```text
Osaka <= timestamp < Amsterdam
```

## Expected Behavior

All previous-fork Engine API methods covered by the Amsterdam specification should reject payload timestamps greater than or equal to the Amsterdam activation timestamp and return:

```text
-38005: Unsupported fork
```

## Fixes

* #6761 — Reject Amsterdam-active payload attributes in `engine_forkchoiceUpdatedV3`
* #6767 — Reject Amsterdam-active payloads in `engine_getPayloadV5`

emirongrr:
**Motivation**

This PR fixes engine_forkchoiceUpdatedV3 payload attribute validation so that V3 is rejected once Amsterdam is active.

Previously, validate_attributes_v3 only checked the lower Cancun boundary:

```rust
if !chain_config.is_cancun_activated(attributes.timestamp)
```
This accepts Amsterdam payload timestamps because Amsterdam is also post-Cancun. However, engine_forkchoiceUpdatedV3 must only be valid for the Cancun-era payload timeframe. Once Amsterdam is active, the CL should use engine_forkchoiceUpdatedV4.

**Description**

Before:

```rust
    if !chain_config.is_cancun_activated(attributes.timestamp) {
        return Err(RpcErr::UnsupportedFork(
            "forkChoiceV3 used to build pre-Cancun payload".to_string(),
        ));
    }
```

After:

```rust
    if chain_config.is_amsterdam_activated(attributes.timestamp) {
        return Err(RpcErr::UnsupportedFork(
            "forkChoiceV3 used to build Amsterdam payload".to_string(),
        ));
    }
    if !chain_config.is_cancun_activated(attributes.timestamp) {
        return Err(RpcErr::UnsupportedFork(
            "forkChoiceV3 used to build pre-Cancun payload".to_string(),
        ));
    }
```

**Spec Ref**

The Amsterdam Engine API specification requires previous-fork Osaka API methods to reject Amsterdam-active payload timestamps.

From [`amsterdam.md`](https://github.com/ethereum/execution-apis/blob/main/src/engine/amsterdam.md#update-the-methods-of-previous-forks):

> For `engine_newPayloadV4`, `engine_getPayloadV5`, and `engine_forkchoiceUpdatedV3`, client software MUST return `-38005: Unsupported fork` if the timestamp of payload is greater than or equal to the Amsterdam activation timestamp.

This PR implements that requirement for `engine_forkchoiceUpdatedV3` by rejecting V3 payload attributes when `attributes.timestamp` is Amsterdam-active.

Closes #6874


crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -425,6 +425,11 @@ fn validate_attributes_v3(
             "Attribute parent_beacon_block_root is null".to_string(),
         ));
     }
+    if chain_config.is_amsterdam_activated(attributes.timestamp) {
+        return Err(RpcErr::UnsupportedFork(
```
ElFantasma, line 439: The test module at lines 594–640 has good fixture patterns for `validate_attributes_v2` / `validate_attributes_v2_pre_shanghai` rejections — adding a matching one for this new Amsterdam gate would lock in the spec compliance and protect against a future refactor accidentally swallowing the check.

Minimal shape, following the existing test style:
```rust
#[test]
fn forkchoice_updated_v3_rejects_amsterdam_payload() {
    // ChainConfig with Amsterdam active at timestamp=1; attributes at timestamp=2.
    let cfg = ChainConfig { amsterdam_time: Some(1), ..Default::default() };
    let attributes = PayloadAttributesV3 {
        timestamp: 2,
        parent_beacon_block_root: Some(H256::zero()),
        ..Default::default()
    };
    let head_block = BlockHeader { timestamp: 1, ..Default::default() };
    // Need the chain_config-aware signature here; current sig takes
    // `&context` so the fixture has to thread it.

    let err = validate_attributes_v3(&attributes, &head_block, &context).unwrap_err();
    assert!(matches!(err, RpcErr::UnsupportedFork(_)));
}
```
The existing tests use a simpler 2-arg signature, so this one has the extra setup cost of building a `RpcApiContext`. If that's too much fixture for a 5-line patch, fair to defer — but worth raising while we're here.

emirongrr: Good point. @ElFantasma 

I was also thinking that this boundary deserves a dedicated test, since it is easy to miss during future refactors.

That said, as you mentioned, the current test setup has some extra cost because `validate_attributes_v3` reads the chain config through `RpcApiContext`. I was considering leaving that cleanup for a follow-up PR, but I’d like to get your opinion here.

One possible follow-up refactor would be to narrow the dependency and pass `&ChainConfig` directly:

```rust
fn validate_attributes_v3(
    attributes: &PayloadAttributesV3,
    head_block: &BlockHeader,
    chain_config: &ChainConfig,
) -> Result<(), RpcErr>
```

That would make this kind of test much cheaper to write. For now, I think we can add the regression test using the existing `RpcApiContext` fixture, and then propose this signature cleanup in a separate PR. What do you think?

ElFantasma (approved):
Test addresses the regression coverage cleanly — locks the new gate in. Approving.

On the signature cleanup follow-up: yes, narrowing to &ChainConfig is the right move. Both validate_attributes_v3 and validate_attributes_v4 only use the context for context.storage.get_chain_config(), so threading the whole RpcApiContext is wider than needed. I'd suggest doing v3 and v4 together in that follow-up PR (validate_attributes_v4 has the same &RpcApiContext shape at line 519), so the signatures stay symmetric and we don't have to circle back for v4 separately.

Happy to review that cleanup PR when it lands. Thanks for picking this up.

azteca1998 (approved):

ilitteri (approved):

emirongrr:
Hi @edg-l , could you please review this PR when you have time?

ilitteri:
> Hi @edg-l , could you please review this PR when you have time?

Hello! We really appreciate your contributions. The missing reviews will land as soon as possible.

As a recommendation for future contributions, PRs addressing actual tickets will have more review priority than PRs that are not related to any filed ticket.

If you happen to find gaps, inconsistencies, bad UX, etc. on ethrex, I encourage you to first file an issue for us to review if we want that feature or not.

emirongrr:
> > Hi @edg-l , could you please review this PR when you have time?
> 
> Hello! We really appreciate your contributions. The missing reviews will land as soon as possible.
> 
> As a recommendation for future contributions, PRs addressing actual tickets will have more review priority than PRs that are not related to any filed ticket.
> 
> If you happen to find gaps, inconsistencies, bad UX, etc. on ethrex, I encourage you to first file an issue for us to review if we want that feature or not.

I'll keep it in mind for future contributions. Thanks again for the guidance

edg-l:
The test currently calls the private validate_attributes_v3 directly. Could you route it through the public ForkChoiceUpdatedV3 handler instead and place it in test/tests/rpc/fork_choice_tests.rs? That also exercises the full validation path, closer to real behavior.

(Also you do not need to merge main so many times, it also makes us need to re-review the PR in case you modified some code in some merge, unless there are merge conflicts that is)

emirongrr:
> The test currently calls the private validate_attributes_v3 directly. Could you route it through the public ForkChoiceUpdatedV3 handler instead and place it in test/tests/rpc/fork_choice_tests.rs? That also exercises the full validation path, closer to real behavior.
> 
> (Also you do not need to merge main so many times, it also makes us need to re-review the PR in case you modified some code in some merge, unless there are merge conflicts that is)

Thanks for the review. Updated, it now exercises the full RPC and validation path.

emirongrr:
Hey @iovoid  this PR is just waiting on the final review. I know everyone has a lot on their plate, but I'd really appreciate it if you could take a look when you have some time. Thanks!

iovoid (approved):

PR #6762, opened 2026-06-01
fix(l1): accept number-encoded terminalTotalDifficulty in genesis

edg-l:
## What

`ChainConfig::terminal_total_difficulty` failed to deserialize when a genesis file encodes `terminalTotalDifficulty` as a bare decimal number (geth/reth style), e.g. `58750000000000000000000`. Genesis loading aborts with `Failed to deserialize u128 number`.

## Why

The value exceeds `u64::MAX`, so serde_json (default features) stores it as `f64`. The old code did `n.to_string().parse::<u128>()`, where `to_string()` yields `"5.875e22"`, which doesn't parse as `u128`. Hex-string form (`"0x..."`) worked; bare numbers didn't.

This reads the integer directly: exact via `as_u64()` for `<= u64::MAX`, with an `as_f64()` cast fallback above that. Hex-string form is unchanged.

## Precision

The `as_f64()` cast is lossy above `u64::MAX` (`58750000000000000000000` → `58749999999999996329984`). This is acceptable because `terminal_total_difficulty` is only used as a post-merge sentinel (`!= Some(0)` to detect PoS); it never enters block hashing, execution, or consensus arithmetic, so the approximation is functionally identical.

## Why not a bit-exact parser

serde_json discards the precision before any custom code runs:
- `deserialize_any` (and `Value`) downcast `> u64::MAX` integers to `f64` (verified: it calls `visit_f64`, never `visit_u128`).
- `deserialize_u128` is exact but errors on strings, so it can't also accept the hex-string form.
- `raw_value` (read the raw token) is exact but breaks `from_reader`/`from_value` for the field; genesis is loaded via `from_reader`.
- `arbitrary_precision` is exact but global, and breaks `#[serde(flatten)]` / `#[serde(untagged)]` with numbers — both used across transaction and RPC types here.

For a sentinel field, none of those is worth the blast radius.

### DIFF
--- a/crates/common/serde_utils.rs
+++ b/crates/common/serde_utils.rs
@@ -317,6 +317,13 @@ pub mod u128 {
         }
     }
 
+    /// Deserializes an `Option<u128>` from either a `0x`-hex string or a bare
+    /// JSON number, for geth/reth genesis compatibility.
+    ///
+    /// **Precision**: bare numbers above `u64::MAX` are read through an `f64`
+    /// cast, losing ~3 decimal digits. Acceptable for sentinel-only fields like
+    /// `terminalTotalDifficulty`; if you need bit-exact `u128` here, add a new
+    /// deserializer rather than reusing this one.
     pub mod hex_str_opt {
         use serde::Serialize;
 
@@ -348,7 +355,9 @@ pub mod u128 {
                         // "PoS active" for TTD; reject them above instead.
                         f as u128
                     } else {
-                        return Err(D::Error::custom("Failed to deserialize u128 number"));
+                        return Err(D::Error::custom(format!(
+                            "u128 value must be a finite, non-negative number; got {n}"
+                        )));
                     };
                     Ok(Some(v))
                 }
--- a/crates/common/serde_utils.rs
+++ b/crates/common/serde_utils.rs
@@ -343,7 +343,9 @@ pub mod u128 {
                     // used as a post-merge sentinel, so f64 imprecision is moot).
                     let v = if let Some(u) = n.as_u64() {
                         u as u128
-                    } else if let Some(f) = n.as_f64() {
+                    } else if let Some(f) = n.as_f64().filter(|f| f.is_finite() && *f >= 0.0) {
+                        // `f as u128` saturates negatives to 0, which would mean
+                        // "PoS active" for TTD; reject them above instead.
                         f as u128
                     } else {
                         return Err(D::Error::custom("Failed to deserialize u128 number"));
--- a/crates/common/serde_utils.rs
+++ b/crates/common/serde_utils.rs
@@ -335,11 +335,21 @@ pub mod u128 {
         {
             let value: Option<serde_json::Value> = Option::deserialize(d)?;
             match value {
-                Some(serde_json::Value::Number(n)) => n
-                    .to_string()
-                    .parse::<u128>()
-                    .map(Some)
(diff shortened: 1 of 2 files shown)


edg-l:
Addressed both in the latest commit: negative/non-finite numbers now error instead of saturating to `Some(0)`, and the number-path test asserts the exact f64 approximation (`58749999999999996329984`) plus a negative-input rejection case.

ilitteri (approved):

crates/common/serde_utils.rs
```diff
@@ -335,11 +335,23 @@ pub mod u128 {
         {
             let value: Option<serde_json::Value> = Option::deserialize(d)?;
             match value {
-                Some(serde_json::Value::Number(n)) => n
-                    .to_string()
-                    .parse::<u128>()
-                    .map(Some)
-                    .map_err(|_| D::Error::custom("Failed to deserialize u128 number")),
+                Some(serde_json::Value::Number(n)) => {
+                    // Values above u64::MAX (e.g. mainnet's TTD) are stored by
```
ElFantasma, line 346: The localized comment ties the f64-lossiness trade-off specifically to TTD's sentinel role, which is exactly right *today*. But this lives in a generic `u128::hex_str_opt` deserializer, and the next field that adopts it (because "u128 from JSON, geth/reth-compatible" is a magnet) will silently get the lossy cast above `u64::MAX` without seeing this comment.

Move (or duplicate) the warning to the `pub mod hex_str_opt` doc at line 320, e.g.:
```rust
/// **Precision**: values above `u64::MAX` are read through an `f64` cast,
/// so they lose ~3 decimal digits of precision. Acceptable for sentinel-only
/// fields like `terminalTotalDifficulty`; if you need bit-exact u128 here,
/// add a new deserializer rather than reusing this one.
pub mod hex_str_opt { ... }
```
Makes the trade-off visible at the type level, not just at the call site. Non-blocking.

crates/common/types/genesis.rs
```diff
@@ -784,6 +784,46 @@ mod tests {
...
+
+        let small: ChainConfig = serde_json::from_str(&format!(
+            r#"{{"chainId":1,"terminalTotalDifficulty":17000000000000000,{dca}}}"#
+        ))
+        .expect("small number TTD should parse");
+        assert_eq!(small.terminal_total_difficulty, Some(17000000000000000u128));
+
+        // A negative bare number must error, not silently saturate to Some(0).
+        let negative = serde_json::from_str::<ChainConfig>(&format!(
+            r#"{{"chainId":1,"terminalTotalDifficulty":-1,{dca}}}"#
```
ElFantasma, line 822: The negative-rejection test asserts only `is_err()`, but the actual error here is the generic `"Failed to deserialize u128 number"` — which is the same message you'd get from a NaN or a non-numeric-Value input. If a future operator hits this on a real genesis file, they get a one-line error with no hint that the *sign* was the issue.

Two cheap improvements (either):
- Make the error message more specific in `serde_utils.rs:351`, e.g. `"u128 value must be a finite, non-negative number; got {n}"`, and assert on the message contents here.
- Or split the negative case into its own match arm with a `"u128 must be non-negative"` error.

Non-blocking, but the operator-friendliness improvement is cheap.

ElFantasma (approved):
LGTM. Just two nits

azteca1998 (approved):

PR #6763, opened 2026-06-01
fix(l1): only warn for genuinely pre-merge genesis

edg-l:
## What

The genesis loader logs `Invalid fork, only post-merge networks are supported.` for normal post-merge networks (e.g. mainnet-fork snapshots), so the warning fires on exactly the networks it's meant to accept.

## Why

The old check warned whenever *all* of `terminal_total_difficulty != 0`, `merge_netsplit_block != 0`, `shanghai_time != 0`, `cancun_time != 0`, `prague_time != 0` held — i.e. whenever the post-merge forks were scheduled at real (non-zero) timestamps. That's true for any mainnet-style genesis, so it false-fires. The rejection it guards (`return Err(InvalidFork)`) is already commented out, so it only logs noise.

## Change

Warn only for a genuinely pre-merge PoW genesis: no merge configured (`terminalTotalDifficulty` / `mergeNetsplitBlock` both unset) **and** non-zero genesis difficulty.

- mainnet-style (TTD set) → no warning
- post-merge devnet (difficulty 0) → no warning
- bare PoW genesis (no merge config, difficulty > 0, e.g. the Hive minimal file) → still warns

Extracted into `is_unsupported_pow_genesis` with a unit test covering all four cases.

Draft: opening for discussion on whether the warning should stay (advisory) or the dead `InvalidFork` path be removed entirely.


crates/common/types/genesis.rs
```diff
@@ -100,6 +92,19 @@ impl TryFrom<&Path> for Genesis {
...
 }
 
+/// Returns true for a genesis that describes a pre-merge PoW chain with no
+/// merge configured. A real post-merge genesis either configures the merge
+/// (terminal total difficulty or merge-netsplit block) or starts post-merge
+/// (genesis difficulty 0). The previous heuristic warned whenever the
+/// post-merge forks were scheduled at non-zero timestamps, which false-fired
+/// on every mainnet-style genesis.
+fn is_unsupported_pow_genesis(genesis: &Genesis) -> bool {
+    let merge_configured = genesis.config.terminal_total_difficulty.is_some()
```
ElFantasma, line 106: `terminal_total_difficulty.is_some()` treats `Some(0)` as "merge configured" — which is correct, because `TTD = 0` is the in-code sentinel for "PoS active from genesis" used elsewhere in this codebase (see the discussion on #6762 about the same field). But that's a load-bearing convention that isn't obvious from this function's name or doc.

Worth a one-line addition to the doc block:
```rust
/// Note: `terminal_total_difficulty = Some(0)` is the sentinel for "PoS active
/// from genesis" and counts as merge-configured here.
```
Keeps the next reader from re-deriving why `Some(0)` doesn't trigger the warning. Non-blocking.

crates/common/types/genesis.rs
```diff
@@ -75,18 +75,10 @@ impl TryFrom<&Path> for Genesis {
         let genesis_reader = BufReader::new(genesis_file);
         let genesis: Genesis = serde_json::from_reader(genesis_reader)?;
 
-        // Try to derive if the genesis file is PoS
-        // Different genesis files have different configurations
+        // ethrex only supports post-merge (PoS) networks.
         // TODO: Remove once we have a way to run PoW chains, i.e Snap Sync
```
ElFantasma: Now that the rejection path is gone (and you tightened the warn to be specific + actionable), the `TODO: Remove once we have a way to run PoW chains, i.e Snap Sync` reads as out of date — Snap Sync isn't what's blocking PoW support; consensus & block-production are. Cheap cleanup: either drop the TODO entirely, or rephrase to something accurate, e.g.
```rust
// ethrex only supports post-merge (PoS) networks. PoW execution is not
// planned.
```
Non-blocking, but the current TODO sends future readers down the wrong rabbit hole.

ElFantasma (approved):
Just two non-blocking comments

azteca1998 (approved):

ilitteri (approved):

iovoid (approved):

PR #6764, opened 2026-06-01
fix(l1): persist BAL and serve V2 payload bodies from store

edg-l:
## Problem

On `bal-devnet-7`, Lighthouse logged ~15 errors/hr from a single node:

```
WARN Execution engine call failed
  error: ServerMessage { code: -32603, message: "Internal Error: EVM error: DB error: state root missing" }
```

Triggered during `beacon_blocks_by_range` payload reconstruction. The CL calls `engine_getPayloadBodiesByRangeV2`/`ByHashV2`, whose handlers regenerate each block's BAL by re-executing the block against its parent state (`generate_bal_for_block` -> `StoreVmDatabase::new`). A snap-synced node does not hold that historical state, so re-execution fails with `state root missing` and the CL can't serve blocks to peers.

The BAL is already validated at block execution and there is a `block_access_lists` table, but the persistence site only stored `produced_bal`, which is `None` on the Amsterdam header-driven path, so live blocks never got their BAL stored.

## Fix

- Persist the validated BAL at execution: `produced_bal.as_ref().or(bal)`, covering the Amsterdam path where `produced_bal` is `None`.
- Serve V2 payload bodies from the store via a `bal_for_block` helper that reads the persisted BAL first and only falls back to regeneration when absent.

Serving no longer re-executes, so it never depends on historical state.

## Notes

- BAL lifecycle stays coupled to the block body; no pruning added.
- Forward-looking: only blocks executed after this change get a stored BAL. Blocks pulled in purely via the non-final batch full-sync path (never live-executed) still lack a stored BAL; tracked in #6765.

## Test

`cargo test -p ethrex-test --test ethrex_tests block_access_list` passes.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -2074,8 +2074,10 @@ impl Blockchain {
                 .store_witness(block_hash, block_number, witness)?;
         };
 
-        // Store the produced BAL (present on Amsterdam+ blocks) so peers can request it
-        if let Some(bal) = &produced_bal {
+        // Store the block's BAL so peers can request it later without re-execution.
+        // On the Amsterdam+ validation path the BAL is supplied via the header and
+        // `produced_bal` is None, so fall back to the validated incoming `bal`.
+        if let Some(bal) = produced_bal.as_ref().or(bal) {
             let block_hash = block.hash();
             if let Err(err) = self.storage.store_block_access_list(block_hash, bal) {
                 warn!("Failed to store block access list for block {block_hash}: {err}");
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -1982,7 +1982,9 @@ impl Blockchain {
     }
 
     /// Same as [`add_block_pipeline`] but also returns the BAL produced during execution.
-    /// On Amsterdam+ blocks the returned value is `Some(bal)`, otherwise `None`.
+    /// On the Amsterdam+ validation path the BAL comes from the header and drives
+    /// execution rather than being rebuilt, so the returned value is `None`; the
+    /// pre-Amsterdam / streaming paths rebuild and return `Some(bal)`.
     pub fn add_block_pipeline_bal(
         &self,
         block: Block,
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -690,13 +690,14 @@ impl Blockchain {
                             &chain_config,
                             &execution_result.requests,
                         )?;
-                        // Amsterdam validation path uses the header BAL directly to drive
-                        // execution; it doesn't rebuild a BAL, so produced_bal is None.
+                        // The parallel Amsterdam validation path uses the header BAL directly
+                        // to drive execution; it doesn't rebuild a BAL, so produced_bal is None.
                         // BAL correctness on that path is enforced inside
                         // execute_block_pipeline (header-BAL index/size/withdrawal-index
                         // checks plus unread_storage_reads / unaccessed_pure_accounts).
-                        // Pre-Amsterdam / streaming paths return Some(produced_bal) and
-                        // the hash check below still runs.
+                        // The sequential Amsterdam path rebuilds a BAL and returns
+                        // Some(produced_bal), so the hash check below runs. Pre-Amsterdam
+                        // blocks never record a BAL, so produced_bal is None there too.
                         if let Some(bal) = &produced_bal {
                             validate_block_access_list_hash(
                                 &block.header,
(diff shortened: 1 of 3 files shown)


crates/networking/rpc/engine/payload.rs
```diff
@@ -737,6 +737,25 @@ fn build_payload_body_response(bodies: Vec<Option<BlockBody>>) -> Result<Value,
...
+fn bal_for_block(
+    context: &RpcApiContext,
+    block: &Block,
+) -> Result<Option<BlockAccessList>, RpcErr> {
+    if let Some(bal) = context.storage.get_block_access_list(block.hash())? {
+        return Ok(Some(bal));
+    }
+    context
+        .blockchain
+        .generate_bal_for_block(block)
```
ElFantasma, line 756: Once the fallback regenerates a BAL successfully, the result is sent to the client and dropped — so the next V2 request for the same block triggers `generate_bal_for_block` again, which re-executes the block. For old Amsterdam blocks (pre-PR ones that aren't in the store yet) that's the slow path on every poll.

Cheap optimization: opportunistically persist the regenerated BAL back to the store, e.g.
```rust
let generated = context
    .blockchain
    .generate_bal_for_block(block)
    .map_err(|e| RpcErr::Internal(e.to_string()))?;
if let Some(bal) = &generated {
    let _ = context.storage.store_block_access_list(block.hash(), bal);
}
Ok(generated)
```

One consistency consideration before doing this: the validated-at-execution BAL and the regenerated-from-state BAL should be byte-identical, but if there's ever a determinism bug where they diverge, this write-back masks it. Reasonable counter: by the time we're regenerating, the block has long been accepted, so byte-equality of the stored vs regenerated BAL isn't load-bearing anymore. Probably OK to write back, but worth a 1-line comment justifying.

Non-blocking.

test/tests/rpc/block_access_list_tests.rs
```diff
@@ -94,3 +109,87 @@ async fn eth_get_block_access_list_unknown_hash_returns_null() {
...
+    let got = request.handle(context).await.expect("rpc ok");
+
+    let expected =
+        serde_json::json!([
+            serde_json::to_value(ExecutionPayloadBodyV2::from_body_with_bal(
+                BlockBody::default(),
+                Some(bal)
+            ))
+            .unwrap()
+        ]);
```
ElFantasma, line 193: Both new tests pre-store the BAL, which proves the store-hit path but doesn't exercise the regenerate fallback. The fallback's pre-Amsterdam short-circuit is the bit that prevents re-execution on snap-synced nodes for pre-Amsterdam blocks — worth at least one test that locks that property in.

Minimal shape:
```rust
#[tokio::test]
async fn payload_bodies_v2_pre_amsterdam_returns_none_without_re_execution() {
    let storage = Store::new("temp.db", EngineType::InMemory).unwrap();
    // Block with pre-Amsterdam timestamp; no BAL stored, no state trie either.
    let block = Block {
        header: BlockHeader { number: 1, timestamp: 0, ..Default::default() },
        body: BlockBody::default(),
    };
    storage.add_block(block).await.unwrap();

    let context = default_context_with_storage(storage).await;
    let request = GetPayloadBodiesByHashV2Request { hashes: vec![block_hash] };
    let got = request.handle(context).await.expect("rpc ok");
    // No BAL stored + pre-Amsterdam → fallback returns None without touching state.
    let expected = serde_json::json!([
        serde_json::to_value(ExecutionPayloadBodyV2::from_body_with_bal(BlockBody::default(), None)).unwrap()
    ]);
    assert_eq!(got, expected);
}
```
Locks in the snap-sync-safety property for the pre-Amsterdam case, which is the primary motivation of the PR. Non-blocking.

ElFantasma (approved):
Just two non-blocking comments

azteca1998 (approved):

crates/blockchain/blockchain.rs
```diff
@@ -1982,7 +1982,9 @@ impl Blockchain {
     }
 
     /// Same as [`add_block_pipeline`] but also returns the BAL produced during execution.
-    /// On Amsterdam+ blocks the returned value is `Some(bal)`, otherwise `None`.
+    /// On the Amsterdam+ validation path the BAL comes from the header and drives
+    /// execution rather than being rebuilt, so the returned value is `None`; the
+    /// pre-Amsterdam / streaming paths rebuild and return `Some(bal)`.
```
iovoid: Why would pre-Amsterdam paths compute a BAL?

edg-l: Good catch — they don't. Traced it: `db.take_bal()` only returns `Some` when `is_amsterdam` (BAL recording is gated on `is_amsterdam`), so `produced_bal` is always `None` for pre-Amsterdam blocks, and the incoming `bal` is also `None` since pre-Amsterdam headers carry no BAL. The `.or(bal)` is correct (both operands are `None` → nothing stored), but the comment wording was wrong.

Fixed the comments in aded28c07: the rebuild-vs-header distinction is between the **parallel** Amsterdam path (header BAL drives execution → `produced_bal` is `None`) and the **sequential** Amsterdam path (block production / `--no-bal-parallel-exec` → rebuilds, returns `Some`), not Amsterdam vs pre-Amsterdam. Pre-Amsterdam never records a BAL.

iovoid (approved):

PR #6767, opened 2026-06-01
fix(l1): reject getPayloadV5 after Amsterdam

emirongrr:
**Motivation**

`engine_getPayloadV5` is a previous-fork Osaka Engine API method. According to the Amsterdam Engine API specification, previous-fork Osaka methods must reject Amsterdam-active payload timestamps.

Spec reference: [`amsterdam.md`](https://github.com/ethereum/execution-apis/blob/main/src/engine/amsterdam.md#update-the-methods-of-previous-forks)

For `engine_newPayloadV4`, `engine_getPayloadV5`, and `engine_forkchoiceUpdatedV3`, client software must return `-38005: Unsupported fork` if the payload timestamp is greater than or equal to the Amsterdam activation timestamp.

Before this change, Ethrex only checked that the built payload timestamp was Osaka-active. Since Amsterdam is after Osaka, Amsterdam payloads also satisfy `is_osaka_activated(...)`, allowing `engine_getPayloadV5` to accept a payload that should be served through the Amsterdam method version instead.

**Description**

This PR tightens the fork boundary for `engine_getPayloadV5`.

The validation now rejects payloads whose timestamp is either:

- before Osaka activation
- greater than or equal to Amsterdam activation

This makes `engine_getPayloadV5` valid only for the Osaka payload timeframe:

This follows the same Amsterdam previous-fork boundary rule addressed in [`lambdaclass/ethrex#6761`](https://github.com/lambdaclass/ethrex/pull/6761), which made `engine_forkchoiceUpdatedV3` reject Amsterdam-active payload attributes. This PR applies the same spec requirement to `engine_getPayloadV5`.

```text
Osaka <= timestamp < Amsterdam
```

Related #6874


crates/networking/rpc/engine/payload.rs
```diff
@@ -580,14 +580,16 @@ impl RpcHandler for GetPayloadV5Request {
         let payload_bundle = get_payload(self.payload_id, &context).await?;
         let chain_config = &context.storage.get_chain_config();
 
-        if !chain_config.is_osaka_activated(payload_bundle.block.header.timestamp) {
+        if !chain_config.is_osaka_activated(payload_bundle.block.header.timestamp)
+            || chain_config.is_amsterdam_activated(payload_bundle.block.header.timestamp)
```
ElFantasma, line 619: Same test-coverage observation I left on #6761 applies here — V5 now has a new fork boundary that no existing test asserts. The cheapest shape is to mirror whatever pattern #6761 lands with (it has the same `RpcApiContext`-fixture issue as the FCUv3 case). Worth at least one test pinning that an Amsterdam-active payload timestamp returns `RpcErr::UnsupportedFork`. Non-blocking, but the three-way Amsterdam previous-fork guard now lives in three places — without tests, a future refactor can quietly drop one and reintroduce the original silent-accept bug.

ElFantasma (approved):

azteca1998 (approved):

ilitteri (approved):

emirongrr:
Hi @edg-l or @iovoid  , could you please review this PR when you have time?

edg-l:
Thanks for these, the fork-boundary logic is correct and matches the Amsterdam spec exactly (-38005 when timestamp >= amsterdam_time).

One request before merge: in this repo we keep RPC tests under the test/ root rather than in inline #[cfg(test)] modules. There's already test/tests/rpc/fork_choice_tests.rs (and the rpc test module) that drives the engine handlers through the public RpcHandler::handle path — that's where these new cases belong.

emirongrr:
> Thanks for these, the fork-boundary logic is correct and matches the Amsterdam spec exactly (-38005 when timestamp >= amsterdam_time).
> 
> One request before merge: in this repo we keep RPC tests under the test/ root rather than in inline #[cfg(test)] modules. There's already test/tests/rpc/fork_choice_tests.rs (and the rpc test module) that drives the engine handlers through the public RpcHandler::handle path — that's where these new cases belong.

Thanks for the review.  the new cases are now under `test/tests/rpc/fork_choice_tests.rs`

edg-l (approved):

PR #6770, opened 2026-06-01
feat(l1): engine REST/SSZ API

iovoid:
**Motivation**

We want to reduce the overhead of JSON serialization/deserialization for the large objects handled by the engine API. One of the proposals to do so is https://github.com/ethereum/execution-apis/pull/793

**Description**

Most of the PR consists of boilerplate: defining RPC objects, route handlers and serialization/deserialization logic.

**Benchmarks**

Screenshot 2026-06-10 at 4 47 58 PM
Screenshot 2026-06-10 at 4 48 13 PM
Screenshot 2026-06-10 at 4 48 29 PM

---

**Status (updated 2026-09-22)**

execution-apis#793 is no longer a proposal: it merged upstream on 2026-09-02, so `src/engine/refactor.md` and `src/engine/refactor-ssz.md` are now in `main` there. Both files have exactly one commit and no edits since, and the last substantive change to the spec was 2026-06-29 (moving the fork out of the path and into `Eth-Execution-Version`), which this branch already implements. In other words the target is stable and this branch is built against the current text.

**What the line count is made of**

The diff is large but only about a third of it is the API itself. Roughly 10,650 added lines across 57 files break down as:

| Part | Lines | Files |
| - | - | - |
| Core: `crates/networking/rpc/engine_rest/` | +3,784 | 27 |
| Core: wiring into the existing rpc crate | +104 / -52 | 5 |
| Tests (`test/tests/rpc/engine_rest_tests.rs`) | +2,880 | 2 |
| Benchmarks (`crates/networking/rpc/benches/`) | +1,881 | 2 |
| Tooling (`tooling/engine_bench`) | +1,827 | 12 |
| Lockfiles, regenerated | +106 / -5 | 4 |
| Manifests and CI config | +66 / -3 | 4 |

So the reviewable surface is the ~3,900 lines of core, of which the largest files are `handlers/payloads.rs` (717), `types/conversions.rs` (397), `handlers/forkchoice.rs` (353) and `handlers/bodies.rs` (337). The remaining ~6,600 lines are tests, benchmarks and the standalone `engine_bench` harness, which can be read separately from the protocol logic. One further line is a caller update in `tooling/reorgs` for the `into_block` to `to_block` rename.

Brought up to date with `main` after a long gap. Beyond the textual conflicts, three things needed repair: `get_latest_block_number` became synchronous, so two handlers no longer compiled; `default_context_with_storage` now returns a `TestContext` guard whose `Drop` joins the executor thread, so the test call sites are split between `into_detached()` where the router outlives the guard and `clone()` where it does not; and the branch carried its own `libssz = "0.2.2"` pins in the tooling workspace while the root workspace had moved to 0.3.0, which made `engine_bench` import one version of `SszEncode` while `ethrex-rpc` implemented another.

Endpoint coverage matches the spec: `/payloads`, `/payloads/{id}`, `/forkchoice`, `/bodies/hash`, `/bodies`, `/blobs/v1..v4`, `/capabilities`, `/identity`, all under `/engine/v1` with the fork selected by `Eth-Execution-Version` and errors returned as RFC 7807 `application/problem+json`.

Two known conformance gaps, small enough to be follow-ups rather than blockers:

- No fallback handler, so an unknown URL under `/engine/v1` returns a bare 404 instead of `/engine-api/errors/method-not-found`.
- `/engine-api/errors/parse-error` is unimplemented; `ssz-decode-error` covers the SSZ case, which leaves the JSON diagnostic endpoints without it.

#7297 stacks the payload witness endpoint (execution-apis#885) on top of this branch and is waiting on it.


crates/networking/rpc/engine_rest/handlers/payloads.rs
```diff
@@ -0,0 +1,679 @@
...
+            }
+        }
+        EngineCall::V1V2 | EngineCall::V3 { .. } => {}
+    }
+
+    // 4. Dispatch to the appropriate handle_new_payload_* helper. The helpers
+    //    only need the expected block_hash from the payload, so we pass it
+    //    directly and skip the `JsonExecutionPayload::from_block` intermediate.
+    let result = match call {
+        EngineCall::V1V2 => handle_new_payload_v1_v2(expected_block_hash, block, ctx, None).await,
```
ElFantasma: REST path drops the `expectedBlobVersionedHashes` cross-check that the JSON-RPC `NewPayloadV3/V4/V5` path runs (engine/payload.rs:142, 265, 408 now pass `Some(self.expected_blob_versioned_hashes.clone())`; REST passes `None` here and on V3/V4/V5 dispatch below).

If #793 intentionally removed the field from the SSZ envelope on the theory that the EL can derive blob hashes from the transactions itself, that's a design choice and this is correct — but the safety net that catches a CL miscomputing blob hashes is gone on this transport. Worth confirming with the spec and adding a one-line comment here noting *why* we pass `None` (it currently reads as forgotten rather than intentional). If the spec actually still requires the check via a header or trailer, the right place to enforce it is right here.

crates/networking/rpc/engine_rest/handlers/bodies.rs
```diff
@@ -0,0 +1,275 @@
...
+                };
+                entries.push(entry);
+            }
+            match TryInto::<BodiesResponseShanghai>::try_into(entries) {
+                Ok(resp) => SszBody(resp).into_response(),
+                Err(_) => bodies_overflow().into_response(),
+            }
+        }
+        Fork::Amsterdam => {
+            let mut entries: Vec<BodyEntryAmsterdam> = Vec::with_capacity(blocks.len());
```
ElFantasma, line 182: Amsterdam `bodies_by_hash` / `bodies_by_range` does `spawn_blocking` **per block** in sequence — for `BODIES_MAX_COUNT = 32` that's 32 serialized blocking-task hops. With the new stored-BAL fast path (`generate_bal_for_block` reads `get_block_access_list` first), each one is cheap, so the cost in steady state is mostly dispatch overhead. But on snap-synced nodes where pre-Amsterdam-cutover blocks lack a stored BAL, every miss falls back to full re-execution serially — a single missing BAL in a 32-block range can multi-second the response.

Not blocking — BAL persistence is the long-term answer. Two cheap mitigations to consider: (a) batch the 32 spawn_blocking dispatches into a single one that walks the slice; (b) for the re-execution fallback path, fail fast with `503 Service Unavailable` instead of stalling the response (the CL can retry against another EL or drop the body request).

crates/networking/rpc/Cargo.toml
```diff
@@ -43,9 +48,14 @@ jemalloc_pprof = { version = "0.8.0", optional = true, features = [
...
 
-# EIP-8025 dependencies (optional)
-libssz-merkle = { workspace = true, optional = true }
-libssz-types = { workspace = true, optional = true }
+# EIP-8025 dependencies
+libssz = { workspace = true }
+# libssz-merkle provides the `HashTreeRoot` trait used by the SSZ derives
+# in engine_rest/types/. The libssz-derive macros expand to impls that
+# reference this trait, so it must be a direct unconditional dep.
+libssz-merkle = { workspace = true }
```
ElFantasma, line 60: These four ssz crates were previously `optional = true, eip-8025`-gated; now all four are unconditional rpc-crate deps. That's necessary (engine_rest is unconditional in `lib.rs`), but it does mean every default rpc build picks them up. The added compile time isn't dramatic, but the change widens the dependency surface for non-eip-8025 builds (think `make dev`, EF-tests, etc.). Worth confirming that's OK — the alternative is feature-gating `pub mod engine_rest;` in `lib.rs` behind a feature so consumers who don't need REST/SSZ don't pay for it.

iovoid: The libssz crate isn't a problem, it has a small footprint.

crates/networking/rpc/engine_rest/handlers/bodies.rs
iovoid: Fixed, now we check store.get_block_access_list(hash)

crates/networking/rpc/engine_rest/handlers/payloads.rs
ElFantasma: is the SSZ envelope explicitly without expectedBlobVersionedHashes per the spec? If so, +1 — could you add a one-liner at the dispatch site mirroring the helper's comment?

ElFantasma (approved):

edg-l (approved):

PR #6771, opened 2026-06-02
fix(l1): accept 128 getBlobs hashes

emirongrr:
**Motivation**

`engine_getBlobsV1` requires execution clients to support request sizes of at least 128 blob versioned hashes.

Spec reference: [`engine_getBlobsV1`](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine_getblobsv1)

Before this change, Ethrex returned `-38004: Too large request` when the request length was exactly `128` because the boundary check used `>= GET_BLOBS_V1_REQUEST_MAX_SIZE`. That rejects a size the spec says clients must support.

Reference clients use the expected upper-bound behavior:

- Reth rejects only requests greater than the limit: [`engine_api.rs`](https://github.com/paradigmxyz/reth/blob/8b07219eb87163b58763a712a422aa66c5a59486/crates/rpc/rpc-engine-api/src/engine_api.rs#L943)
- Geth also treats the limit as a supported maximum rather than rejecting exactly `128`: [`api.go`](https://github.com/ethereum/go-ethereum/blob/fdf99d9883c71036518fc4a243f3620539f8150b/eth/catalyst/api.go#L574)

**Description**

This PR fixes the getBlobs request-size boundary.

The validation now rejects requests only when the number of requested blob versioned hashes is greater than `GET_BLOBS_V1_REQUEST_MAX_SIZE`, allowing exactly `128` hashes as required by the Engine API spec.

The same boundary fix is applied to:

- `engine_getBlobsV1`
- the shared `engine_getBlobsV2` / `engine_getBlobsV3` helper path

Regression coverage was added for both paths:

- `blobs_v1_accepts_max_hashes`
- `blobs_v3_accepts_max_hashes`

One small naming cleanup may be worth discussing in a follow-up or in this PR: `GET_BLOBS_V1_REQUEST_MAX_SIZE` is currently used by V1, V2, and V3. Since the limit is shared across getBlobs versions, a version-neutral name such as `MAX_BLOB_LIMIT` or `MAX_GET_BLOBS_REQUEST_SIZE` may describe the intent more clearly.


crates/networking/rpc/engine/blobs.rs
```diff
@@ -64,7 +64,7 @@ impl RpcHandler for BlobsV1Request {
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
         debug!("Received new engine request: Requested Blobs");
 
-        if self.blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE {
+        if self.blob_versioned_hashes.len() > GET_BLOBS_V1_REQUEST_MAX_SIZE {
```
ElFantasma, line 87: If you'd rather not rename `GET_BLOBS_V1_REQUEST_MAX_SIZE` in this PR (keeping the change purely behavioral), the minimum-friction option is a doc comment at the constant's declaration (line 16):
```rust
/// Maximum number of blob versioned hashes accepted by `engine_getBlobsV{1,2,3}`.
/// Per the Engine API spec, clients must support at least 128. Despite the
/// V1 in the name, V2 and V3 share this limit (see `get_blobs_and_proof`).
const GET_BLOBS_V1_REQUEST_MAX_SIZE: usize = 128;
```
Keeps the constant's reach honest until a rename PR lands.

ElFantasma (approved):

azteca1998 (approved):

ilitteri (approved):

emirongrr:
Closing as superseded by #6800(https://github.com/lambdaclass/ethrex/pull/6800), which was merged with an equivalent fix for this issue. Thanks for the review. @ilitteri @ElFantasma

Issue #796, opened 2024-10-09
Investigate when transactions should be removed from the mempool


fmoletta:

fmoletta:
Closed in favour of #797

PR #6773, opened 2026-06-02
feat(l1): implement FCUv4 targetGasLimit (execution-apis#796)

edg-l:
Adds `target_gas_limit: u64` to `PayloadAttributesV4` in `types/fork_choice.rs`. `build_payload_v4` in `engine/fork_choice.rs` uses the CL-supplied value as `BuildPayloadArgs.gas_ceil`; the EIP-1559 1/1024 envelope in `calc_gas_limit` still clamps it relative to the parent.

Per [execution-apis#796](https://github.com/ethereum/execution-apis/pull/796) the field is **required**. `parse_v4` rejects a present-but-malformed V4 attributes object (e.g. one missing `targetGasLimit`) with `InvalidPayloadAttributes` rather than silently dropping it; an absent/null attributes parameter still yields no attributes (a plain forkchoice update). There is no `--builder.gas-limit` fallback: an Amsterdam CL that omits `targetGasLimit` is non-compliant and is rejected, keeping the EL aligned with #796 for upcoming devnets.

Also fixes a latent `BuildPayloadArgs::id()` collision: `slot_number` was not hashed and `gas_ceil` is now caller-supplied, so two FCUv4 calls differing only in `targetGasLimit` or `slotNumber` would produce the same `payload_id`. Both fields are now part of the Keccak input.

Tests cover: `PayloadAttributesV4` serde (present parses; absent and explicit `null` are rejected); FCUv4 e2e accepts a present `targetGasLimit`, rejects an absent one with `InvalidPayloadAttributes`, and still rejects pre-Amsterdam timestamps; `BuildPayloadArgs::id()` distinguishes `gas_ceil` and `slot_number`; `calc_gas_limit` step-up, step-down, and clamp. First commit cherry-picked from ilitteri.


azteca1998 (dismissed):

crates/blockchain/payload.rs
```diff
@@ -117,6 +117,18 @@ impl BuildPayloadArgs {
...
+        // gas_ceil so two FCUv4 calls that differ only in CL-supplied
+        // targetGasLimit (or in slot_number) do not collide on payload_id.
+        // Scoped to V4: for V1/V2/V3 gas_ceil is always the static
+        // --builder.gas-limit (no collision to disambiguate), so hashing it
+        // there would change those IDs without fixing anything.
+        if let Some(slot_number) = self.slot_number {
+            hasher.update(slot_number.to_be_bytes());
+        }
+        if self.version >= 4 {
+            hasher.update(self.gas_ceil.to_be_bytes());
```
ElFantasma, line 130: Asymmetry worth pinning down: `gas_ceil` hashing is explicitly gated on `self.version >= 4`, but `slot_number` hashing is gated implicitly via `if let Some(slot_number)`. Today the two are equivalent because every V1-V3 construction site in `fork_choice.rs` sets `slot_number: None` (verified at line 461) — only the V4 path passes `Some(attributes.slot_number)` (line 563). So V1-V3 payload IDs are unchanged in practice.

But the contract is a soft invariant. If a future V5/V6/V7 path keeps setting `slot_number: Some(_)` (which is correct) AND someone adds slot to a V3 fork-choice call site by mistake (e.g. via a helper that defaults it), the V3 ID would silently start including the slot in the hash, breaking ID stability for any caller cached against the old IDs.

Two cheap lock-ins:
1. Symmetric explicit gate: `if self.version >= 4 && let Some(slot_number) = self.slot_number { ... }` so both V4+-only inputs are gated the same way and the version is the single source of truth.
2. Or add a regression test that V1-V3 IDs stay stable when `slot_number` is None (your existing `payload_id_stable_when_inputs_match` uses version=4; a parallel one for version=3 with slot_number=None would lock it in).

(1) is one extra `&&`; (2) is ~10 lines of test. Either works. Non-blocking — current code is correct.

ElFantasma (approved):
Just one non-blocking observation

ilitteri (approved):

iovoid (approved):

PR #6774, opened 2026-06-02
perf(l1): route remaining hash calls through precompile

benbencik:
**Motivation**
Previous change in #6549 applied the hash function precompiles to computing the root hash. This refactor applies the precompiles to *(almost all)* other hash function calls. 

**Description**
In call sites outside of the guest program, use `NativeCrypto`. To my understanding is equivalent to the current state. The hash function calls that were omitted are in `calculate_create_address` and `calculate_create2_address`. 

**Results**  
Profiling in ZisK v0.16.1 on mainnet blocks 24949787 - 24949811:
- Average cost (main `c8c4d4735`): 45.17B
- Average cost (this change): 43.54B


azteca1998 (approved):
There are some bot review comments about hash caching, I think we could adapt it so that those are still used with the new `Crypto` trait instead of recomputing every time.

For example, refactoring so that instead of `.hash()` we do `.hash(crypto)` and compute the value using the `Crypto` impl.

crates/common/types/transaction.rs
```diff
@@ -1465,11 +1465,11 @@ impl Transaction {
         }
     }
 
-    fn compute_hash(&self) -> H256 {
+    pub fn compute_hash(&self, crypto: &dyn Crypto) -> H256 {
         if let Transaction::PrivilegedL2Transaction(tx) = self {
             return tx.get_privileged_hash().unwrap_or_default();
         }
-        crate::utils::keccak(self.encode_canonical_to_vec())
+        H256(crypto.keccak256(&self.encode_canonical_to_vec()))
```
ElFantasma, line 1475: Caching divergence worth pinning down: `Transaction::hash()` (line 1486) memoizes via `inner_hash.get_or_init(|| compute_hash(&NativeCrypto))`, but `compute_hash(crypto)` is uncached. In a guest run where LEVM calls `tx.compute_hash(precompile_crypto)` (per `levm/mod.rs:260`), the precompile always runs fresh — and if something else also calls `tx.hash()` for the same tx (logging, receipt construction, RLP encoding paths), you end up paying both: NativeCrypto for the cached value AND the precompile per LEVM call.

The two will always agree (both compute `keccak256(canonical_encoding)`), so this is a perf concern, not a correctness one. Three shapes:
1. **Leave it** — the 1.4% win you measured is the net of paying both. Document it inline so future readers know it's intentional.
2. **Replace `tx.hash()` callers in guest paths with `tx.compute_hash(crypto)`** so the precompile is the only impl that runs there.
3. **Cache differently** — a second OnceCell slot, or have `inner_hash` accept the first impl that fills it regardless of crypto choice (deterministic, so the cached value is correct for both).

Non-blocking — but worth a sentence in the PR body or a code comment so the divergence is intentional, not accidental.

crates/vm/backends/levm/mod.rs
```diff
@@ -255,7 +255,11 @@ impl LEVM {
 
             let report = Self::execute_tx(tx, tx_sender, &block.header, db, vm_type, crypto)?;
 
-            tx_gas_breakdowns.push(TxGasBreakdown::from_report(tx_idx, tx.hash(), &report));
+            tx_gas_breakdowns.push(TxGasBreakdown::from_report(
+                tx_idx,
+                tx.compute_hash(crypto),
```
ElFantasma: Tiny consistency note: this call site was previously `tx.hash()` (cached, NativeCrypto). The switch to `tx.compute_hash(crypto)` is the new pattern, but only this call site in LEVM was migrated. Other tx-hash call sites in the VM path that I can find still use `tx.hash()`. If the goal is "all guest-side hashing routes through `crypto`", a quick grep for `tx.hash()` in LEVM / guest-program would confirm whether anything got missed. Non-blocking — this PR doesn't claim to cover every site (the title says "remaining hash calls" but the exclusion list isn't enumerated).

crates/common/types/transaction.rs
```diff
@@ -1465,11 +1465,11 @@ impl Transaction {
         }
     }
 
-    fn compute_hash(&self) -> H256 {
+    pub fn compute_hash(&self, crypto: &dyn Crypto) -> H256 {
```
iovoid: Ideally this should be avoided, and hash() should receive Crypto. Given transactions are only hashed twice outside error paths, it's acceptable to leave as a TODO but that should be mentioned.

benbencik: I was hesitant of rewriting `.hash()` into `.hash(Crypto)` becuase it is a larger refactor. But since you suggested it, I did the rewrite and it seems like a good call.

crates/vm/backends/levm/mod.rs
benbencik: Now `hash` get `crypto`, reverted back to using `hash`.

crates/common/types/transaction.rs
benbencik: Right, the cache was not used properly. After the new changes, I recorded 3.6% improvement from the baseline (`c8c4d4735`) measured on the same blocks as before.

ElFantasma (approved):
LGTM!

iovoid (approved):

ilitteri (approved):

edg-l (approved):

PR #6775, opened 2026-06-02
perf(l1): short-circuit KECCAK256 on empty input

edg-l:
## Motivation

Benchmarkoor (ethrex `bal-full` suite) shows ethrex is at **parity with reth** for non-empty KECCAK256 (msg_size 32/256/1024 ≈ 1.0x throughput ratio), but **~19x slower** on zero-length input:

| msg_size | ethrex MGas/s | reth MGas/s | gap |
|---|---|---|---|
| 0 | ~600 | ~11,500 | **~19x** |
| 32 | ~680 | ~605 | 0.9x |
| 256 | ~710 | ~720 | 1.0x |
| 1024 | ~510 | ~510 | 1.0x |

The gap is isolated to running the keccak permutation on empty input. `keccak256("")` is a constant; fast clients return it directly.

## Change

In the `KECCAK256` handler, return the precomputed `keccak256("")` value as a `U256` when `len == 0`, skipping both the memory load and the permutation. Gas accounting is unchanged. A unit test asserts the constant equals `NativeCrypto.keccak256(&[])`.

## Testing

- `cargo test -p ethrex-levm --lib opcode_handlers::keccak` passes (const verified against the hasher).
- `cargo clippy -p ethrex-levm` clean.

### DIFF
--- a/crates/vm/levm/src/opcode_handlers/keccak.rs
+++ b/crates/vm/levm/src/opcode_handlers/keccak.rs
@@ -11,7 +11,17 @@ use crate::{
     utils::size_offset_to_usize,
     vm::VM,
 };
+use ethrex_common::U256;
 use ethrex_common::utils::u256_from_big_endian;
+
+/// `keccak256("")` as a `U256`. Returned directly for zero-length input so the
+/// hot path skips the permutation entirely (matches what other clients do).
+const EMPTY_KECCAK_U256: U256 = U256([
+    0x7bfad8045d85a470,
+    0xe500b653ca82273b,
+    0x927e7db2dcc703c0,
+    0xc5d2460186f7233c,
+]);
 
 pub struct OpKeccak256Handler;
 impl OpcodeHandler for OpKeccak256Handler {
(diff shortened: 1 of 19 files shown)


MegaRedHand (approved):

crates/vm/levm/src/opcode_handlers/keccak.rs
```diff
@@ -11,8 +11,18 @@ use crate::{
     utils::size_offset_to_usize,
     vm::VM,
 };
+use ethrex_common::U256;
 use ethrex_common::utils::u256_from_big_endian;
 
+/// `keccak256("")` as a `U256`. Returned directly for zero-length input so the
+/// hot path skips the permutation entirely (matches what other clients do).
+const EMPTY_KECCAK_U256: U256 = U256([
```
ElFantasma, line 19: We now have **two** constants for `keccak256("")` — `EMPTY_KECCACK_HASH: LazyLock<H256>` in `crates/common/constants.rs:38` (note the pre-existing **typo** — extra `C`) and this new `EMPTY_KECCAK_U256: U256`. They have to stay in lock-step forever, in two files, under two different names. That's the kind of duplication that tends to drift quietly.

The LazyLock was only needed because the existing definition uses `hex::decode(...).expect(...)` at runtime. A true `const H256` is straightforward — `H256` is just `H256([u8; 32])` (verified: `H256([2; 32])` construction is already used elsewhere in the tree), and `hex_literal::hex!` is already a workspace dep (used by `crates/common/crypto/provider.rs` and others). With that, both shapes become real `const`s side-by-side:
```rust
// in crates/common/constants.rs
pub const EMPTY_KECCAK_HASH: H256 = H256(hex_literal::hex!(
    "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
));
pub const EMPTY_KECCAK_U256: U256 = U256([
    0x7bfad8045d85a470, 0xe500b653ca82273b,
    0x927e7db2dcc703c0, 0xc5d2460186f7233c,
]);
// + a #[test] asserting the two encode the same bytes — catches drift.
```

Two paths I can see:
1. **In this PR**: define both constants together in `common::constants`, deprecate (or rename) `EMPTY_KECCACK_HASH`. The rename touches ~17 files (`git grep -l EMPTY_KECCACK_HASH` returns: vm.rs, account.rs, block.rs, block_execution_witness.rs, healing/state.rs, snap_sync.rs, types/block.rs, store.rs, levm db.rs/mod.rs/tracing.rs, levm/account.rs, levm/db/gen_db.rs, store_tests.rs, archive_sync, ef_tests test_runner.rs). Mechanical `s/EMPTY_KECCACK_HASH/EMPTY_KECCAK_HASH/g`.
2. **Land this as-is, file a follow-up**: this PR stays surgical (1 const, 1 test, hot-path fix); the unification + rename PR can stand on its own. If you go this way, a drift-catching test extension would be the cheapest in-PR guard: `assert_eq!(EMPTY_KECCAK_U256, u256_from_big_endian(EMPTY_KECCACK_HASH.as_bytes()));` next to the existing assertion catches divergence if either ever drifts.

Your call. Non-blocking on the perf merit of this PR either way.

ElFantasma (approved):
Approved with a note... It's up to you to fix it or file a follow up issue.

azteca1998 (approved):

PR #6776, opened 2026-06-02
feat(l1): eth/72 sparse blobpool (EIP-8070)

edg-l:
Implements [EIP-8070](https://eips.ethereum.org/EIPS/eip-8070) (eth/72 sparse blobpool / PeerDAS sampling) and the corresponding [devp2p `caps/eth.md`](https://github.com/ethereum/devp2p/blob/master/caps/eth.md) eth/72 spec. The EIP is Draft (a Glamsterdam candidate, not SFI)

## Changes

- **eth/72 protocol** (extends eth/71): `cell_mask` field on `NewPooledTransactionHashes`; blob payloads elided to samplers in `PooledTransactions`; new `GetCells (0x14)` / `Cells (0x15)`. Capability negotiation is version-gated, so eth/71 peers are unaffected.
- **Blobpool state machine** (`mempool.rs`, `sampling.rs`): per-tx custody columns, cell verification against EIP-7594 proofs before storage, and pseudo-random provider role assignment.
- **Engine API**: `forkchoiceUpdatedV4` accepts optional `custodyColumns`; new `engine_getBlobsV4` returns `BlobCellsAndProofsV1`. Custody expansion re-samples pending blob txs.

## Wire format

Matches the canonical devp2p schemas, including the eth/66+ `request-id` wrapper that the EIP's shorthand omits:

- `GetCells`: `[request-id, [txhash…], cells: B_16]`
- `Cells`: `[request-id, [txhash…], [[cell…]…], cells: B_16]`

`Cells` payloads are packed index-major: for each set index, one cell from every blob of the transaction, in blob order.

`cell_mask` is little-endian: column `i` maps to byte `i/8`, bit `i%8`. EIP-8070 types the field as `B_16`/`uint128` and describes it as a bitarray of length `CELLS_PER_EXT_BLOB` without pinning a byte order, but both readings resolve to the same bytes under SSZ, so there is no real ambiguity:

- `BitVector[N]` ([consensus-specs `ssz/simple-serialize.md`](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#bitvectorn)) serializes as `array[i // 8] |= value[i] << (i % 8)`.
- `UintN` serializes as `value.to_bytes(N // 8, "little")`.

Since `custodyColumns` *is* the CL's custody set, a client choosing big-endian would disagree with its own CL. This also matches geth's `types.CustodyBitmap` (`result[i/8] |= 1 << (i % 8)`). The same layout is applied at every conversion site: the wire `B_16`, FCU v4 `custodyColumns`, and `engine_getBlobsV4`'s `indices_bitarray`.

A `Cells` response is validated against the `GetCells` it answers: an unknown request id, an unrequested transaction hash, or a cell bitmap that isn't a subset of the request's is a subprotocol violation and disconnects the peer, before anything reaches the cell store. Requests are chunked at the recommended 64-hash soft limit.

## Default behavior

Sampling is **off by default**. `--blob-sampling` (`ETHREX_BLOB_SAMPLING`) activates the sampler/provider state machine; without it the node is a full-replication provider (p=1.0) and default builds are behaviorally unchanged.

Eager-provider mode (p=1.0 for every blob tx) latches on permanently the first time the CL requests a payload build, so a validator running `--blob-sampling` cannot propose blobs it only sampled. The latch is one-way, so it does not leak the proposer schedule through the fetch pattern. `--blob-eager-provider` forces it from startup for nodes that build outside the Engine API path.

## Testing

`make lint` and the full test suite pass. New coverage: `mempool_cells_tests`, `sampling_tests`, `eip8070_kzg_tests`, `p2p/rlpx/eth72_tests`, `rpc/eth72_engine_tests`.

## Follow-ups

- Cell pruning on custody contraction is deferred; cells are retained on shrink (overcustody is harmless).


LEVM_DISPATCH_INVESTIGATION.md
```diff
@@ -0,0 +1,83 @@
+# levm dispatch overhead analysis (ISZERO microbenchmark gap vs revm)
```
ElFantasma: This investigation doc is unrelated to EIP-8070 — it's an LEVM dispatch overhead analysis (ISZERO microbenchmark vs revm/reth). It looks like a stray scratchpad file that got committed; please drop it from this PR (or move it under `docs/` in its own PR if it's meant to land). Carrying 83 lines of off-topic notes into the eth/72 PR makes the review harder and confuses the history of where the analysis came from.

crates/networking/p2p/rlpx/eth/eth72/transactions.rs
```diff
@@ -0,0 +1,437 @@
...
+    error::{RLPDecodeError, RLPEncodeError},
+    structs::{Decoder, Encoder},
+};
+use ethrex_storage::error::StoreError;
+use tracing::debug;
+
+/// Convert u128 to a fixed 16-byte big-endian array (B_16 per EIP-8070).
+/// Isolated here so endianness is a one-line change before cross-client devnet.
+pub fn u128_to_b16(v: u128) -> [u8; 16] {
+    v.to_be_bytes()
```
ElFantasma: This is the highest-risk interop item in the PR and the body acknowledges it: the `cell_mask` wire format is hard-coded to big-endian here without a cross-client confirmation. EIP-8070 defines the field as `B_16` (16-byte fixed-length byte array) — which is endianness-agnostic on the EIP side — and the actual convention is whatever geth/reth ship. If they pick LE the masks will be byte-swapped on the wire and every blob announcement will look like "node holds completely different columns than it actually does", silently breaking sampling-vs-provider pairing.

Please nail this down before any merge — even just an inline ref to a geth/reth PR or devnet capture that fixes BE would be enough. The `u128_to_b16` / `b16_to_u128` isolation here is good: it really is a one-line flip if needed, and it would be worth a unit test pinning the expected wire-byte sequence for a non-trivial mask (e.g. `0x00000000_00000000_00000000_00000001` ↔ `[0; 15] ++ [1]` for BE, `[1] ++ [0; 15]` for LE) to make accidental regressions impossible.

The same concern applies to `parse_indices_bitarray` / `parse_custody_columns` in the engine API — those also assume BE. Worth aligning the choice across the wire and the JSON-RPC layer in one go.

edg-l: fixed, checked out geth source

ElFantasma (approved):

crates/networking/p2p/rlpx/eth/cells.rs
```diff
@@ -0,0 +1,303 @@
...
+    /// tx has a full blob payload (blobs non-empty), all 128 columns are available
+    /// and we compute cells on demand via `cells_for_columns`. When only sampled
+    /// cells are held, the TxCells mask is used.
+    ///
+    /// Consequence of the uniform mask: if ANY requested hash is unknown, private
+    /// or we hold no cells for it, the intersection collapses and we serve zero cells
+    /// for the whole batch. This is a protocol-level limitation of the single
+    /// per-message `cell_mask`, not a bug; callers should request hashes they
+    /// expect us to hold together.
+    pub fn handle(&self, mempool: &Mempool) -> Cells {
```
iovoid, line 304: There should be a cap, otherwise this becomes a DoS vector.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1868,6 +2407,120 @@ async fn flush_pending_tx_requests(state: &mut Established) -> Result<(), PeerCo
...
+            .insert(request_id, (announcement, chunk.to_vec(), Instant::now()));
+    }
+
+    Ok(())
+}
+
+/// EIP-8070: flush buffered cell requests as batched GetCells messages.
+/// Mirrors flush_pending_tx_requests_72 but sends GetCells instead of
+/// GetPooledTransactions.
+async fn flush_pending_cell_requests(state: &mut Established) -> Result<(), PeerConnectionError> {
```
iovoid, line 2584: Shouldn't this only happen on eth/72-enabled peers?

crates/networking/p2p/rlpx/eth/eth72/transactions.rs
```diff
@@ -0,0 +1,460 @@
...
+        use ethrex_rlp::encode::encode_length;
+
+        // Build each tx's canonical bytes: blob txs use elided encoding.
+        let mut txs_bytes = vec![];
+        for tx in &self.pooled_transactions {
+            let canonical = match tx {
+                P2PTransaction::EIP4844TransactionWithBlobs(wrapped) => {
+                    encode_elided_canonical(wrapped)
+                }
+                other => other.encode_canonical_to_vec(),
```
iovoid: Shouldn't this use encode_field like the eth/71 path?

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1868,6 +2407,120 @@ async fn flush_pending_tx_requests(state: &mut Established) -> Result<(), PeerCo
...
+    const MAX_HASHES_PER_REQUEST: usize = 256;
+    for (i, chunk) in all_hashes.chunks(MAX_HASHES_PER_REQUEST).enumerate() {
+        let offset = i * MAX_HASHES_PER_REQUEST;
+        let chunk_len = chunk.len();
+        let chunk_types = &all_types[offset..offset + chunk_len];
+        let chunk_sizes = &all_sizes[offset..offset + chunk_len];
+        let chunk_masks = &all_masks[offset..offset + chunk_len];
+
+        // Compute the mask for this chunk only: OR of masks for hashes in this chunk.
+        let chunk_cell_mask: Option<u128> =
```
iovoid: I think None=sampler, Some=provider is error-prone.

iovoid (changes requested):

edg-l:
@iovoid all four addressed in 1f6b85d and 868d526.

- **Cap on `GetCells`**: responses now stop at the devp2p 2 MiB `Cells` soft limit and omit trailing transactions; both `GetCells` and `Cells` decode with a bounded snappy length instead of unbounded. Our own requests are sized to fit that budget and the 64-hash request soft limit.
- **eth/72 gating**: correct, the custody sweep buffered cell requests for every peer and the flush sent them unconditionally. Both gate on the negotiated version now.
- **`encode_field`**: done, `PooledTransactions72::encode` goes through `Encoder` with an `Elided` newtype for the blob payload.
- **None=sampler / Some=provider**: worse than error-prone, it was broken. `flush_pending_tx_requests_72` OR-merged the chunk masks, so any tick carrying both a provider batch (`Some(u128::MAX)`) and a sampler batch (`None`) collapsed to `Some(MAX)`, and the `is_none()` check then skipped the cell fetch for every sampler hash in it. Replaced with an explicit `BlobFetchRole`, grouped by role before chunking.

Cross-checking against devp2p `caps/eth.md` turned up one more: it types `cells` as `B_16` and lets a peer send it when no blob tx is announced, so a conformant peer's non-blob announcements were overwriting its recorded availability with 0 and stopping us sampling from it. Now gated on the announcement actually carrying a type-3 tx.

crates/networking/p2p/rlpx/eth/eth72/transactions.rs
```diff
@@ -0,0 +1,462 @@
...
+    fn encode(&self, buf: &mut dyn BufMut) {
+        let canonical = match self.0 {
+            P2PTransaction::EIP4844TransactionWithBlobs(wrapped) => {
+                encode_elided_canonical(wrapped)
+            }
+            other => other.encode_canonical_to_vec(),
+        };
+        <[u8] as RLPEncode>::encode(&canonical, buf);
+    }
+}
```
iovoid, line 326: legacy transactions also need special handling

crates/blockchain/mempool.rs
```diff
@@ -181,6 +238,19 @@ struct MempoolInner {
...
+
+    // ── EIP-8070 / PeerDAS fields ────────────────────────────────────
+    /// Bitmask of column indices this node is custodying (set via Engine API
+    /// FCU v4 `custodyColumns`; defaults to 0 when sampling is disabled).
+    custody_columns: u128,
+    /// Verified cells for blob transactions that are in the pool.
+    cells: FxHashMap<H256, TxCells>,
+    /// For each blob tx hash, the set of peer IDs that have announced it
+    /// via NewPooledTransactionHashes72 (sampler role tracking).
+    provider_announcers: FxHashMap<H256, FxHashSet<H256>>,
```
iovoid, line 250: Seems able to grow without bound.

crates/networking/p2p/rlpx/eth/eth72/transactions.rs
```diff
@@ -0,0 +1,462 @@
...
+                    debug!(
+                        peer=%node,
+                        "Rejecting blob transaction in L2 mode",
+                    );
+                    continue;
+                }
+                // Blobs are elided in eth/72; store commitments+proofs.
+                // Full KZG validation deferred until blobs are fetched via GetCells.
+                if let Err(e) = blockchain
+                    .add_blob_transaction_to_pool(itx.tx, itx.blobs_bundle)
```
iovoid, line 408: I think the eth/71 path isn't ignoring the eth/72 transactions (which it can't provide blobs for if asked)

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -626,6 +704,73 @@ impl PeerConnectionServer {
...
+                .blockchain
+                .mempool
+                .prune_alternates(INFLIGHT_TX_TIMEOUT)
+            {
+                warn!(error = %e, "prune_alternates failed during sweep");
+            }
+            // EIP-8070: when the custody set changed (Engine API FCU v4) or eager
+            // mode latched on, fetch the now-wanted columns this peer can serve for
+            // pending blob txs. Inert unless sampling is enabled, and only ever
+            // asked of an eth/72 peer.
```
iovoid, line 744: This queues the same sample several times if different peers have them, and doesn't check if we already have it.

crates/networking/p2p/rlpx/eth/eth72/transactions.rs
```diff
@@ -0,0 +1,462 @@
...
+        // so receivers know we can serve every requested column for the whole batch.
+        // u128::MAX is advertised when we hold the full blob payload (all 128
+        // columns derivable); otherwise the sampled-cells mask is used.
+        let cell_mask = if has_blob_tx {
+            let mask = transaction_hashes
+                .iter()
+                .zip(transaction_types.iter())
+                .filter(|&(_, &ty)| ty == 3)
+                .fold(u128::MAX, |acc, (hash, _)| {
+                    acc & blockchain.mempool.available_cell_mask(*hash)
```
iovoid, line 133: Wouldn't this usually end up as 0?

crates/blockchain/sampling.rs
```diff
@@ -0,0 +1,69 @@
...
+/// Returns `None` only when `custody_mask` already covers all 128 columns
+/// (which is the provider case; samplers always have spare columns).
+pub fn pick_random_extra_column(custody_mask: u128, tx_hash: H256) -> Option<u32> {
+    let available = !custody_mask; // bits NOT in custody
+    let count = available.count_ones();
+    if count == 0 {
+        return None;
+    }
+    // Use 4 bytes of tx_hash for a more uniform index across up to 128 columns.
+    let b = tx_hash.as_bytes();
```
iovoid: This doesn't seem very random

iovoid (changes requested):
- In the provider mode, we don't obtain blobs through GetCells as the spec requires.
- is_data_request doesn't consider GetCells

ilitteri (approved):

PR #6780, opened 2026-06-02
feat(l1): default omitted block to latest on eth state methods

MysticRyuujin:
## Summary

Two related JSON-RPC changes aligning ethrex with **ethereum/execution-apis#812** by **@manusw7** (Block parameter optional, default `latest`):

1. **Default an omitted block parameter to `latest`** on the state methods that previously required it. `eth_getBalance`, `eth_getCode`, `eth_getStorageAt`, `eth_getTransactionCount`, and `eth_getProof` did a strict `params.len() != N` check and returned `-32000 "Expected N params"` when the block was omitted. They now accept the block as optional and default it to `latest` via a new `impl Default for BlockIdentifierOrHash` (latest).
2. **Implement `eth_getStorageValues`** (previously `-32601 method not found`): batches storage-slot reads for multiple accounts at a block, mirroring `eth_getStorageAt`, with the same optional-block default and a 1024-slot cap. Registered in the `eth` dispatch.

## Motivation

Aligns ethrex with execution-apis#812 (@manusw7) and matches the other clients implementing the same change. Full credit to @manusw7 for the proposal.

## Testing

- `cargo test -p ethrex-rpc --lib` — 64 passed, 0 failed. New parse-layer tests assert each state method resolves an **omitted** block to `latest` (`test_state_methods_default_block_to_latest_when_omitted`) and that `eth_getStorageValues` parses (and rejects an empty request).
- Confirmed via the hive `rpc-compat` simulator that the omitted-block request now produces the **same** response as an explicit `"latest"` request for every method (previously the omitted form errored).

### Note on hive rpc-compat value fixtures (out of scope)

ethrex does not currently pass the rpc-compat *value* fixtures for these state methods in the hive default chain — but this is **pre-existing and unrelated to this change**: a stock ethrex build returns the same results for an explicit `"latest"` request (e.g. `eth_getBalance(addr, "latest")` returns `0x0` where other clients return the funded balance), and `*-blockhash` requests error with "Could not resolve block number". That points to an ethrex/hive default-chain state-serving issue, which is out of scope here. This PR makes the omitted-block path behave identically to the explicit-`latest` path (the spec change), proven by the unit tests above.

## Related

- Source proposal: ethereum/execution-apis#812 (by @manusw7)
- execution-apis spec + conformance tests: ethereum/execution-apis#814
- go-ethereum: ethereum/go-ethereum#35100
- Nethermind: NethermindEth/nethermind#11883
- Erigon: erigontech/erigon#21586
- Besu: besu-eth/besu#10587


MysticRyuujin:
Thanks Copilot — reviewed both:

- **Slot limit enforced after deserialization / `usize` overflow** — not changing. The request body is already size-capped by the RPC server, which bounds the deserialized map; and `total_slots` overflowing `usize` would require ~2^64 slots (≈10^20 bytes of request body), which the body cap makes impossible. This also matches the existing state methods (`getProof`/`getStorageAt` deserialize params before their own checks), so I'd rather keep it consistent than special-case this one.
- **`RpcErr::Internal` for an unresolvable block → should be `BadParams`** — reasonable in the abstract, but this mirrors the existing sibling methods (`GetBalance`/`GetCode`/`GetStorageAt` all return `Internal("Could not resolve block number")`). Reclassifying just `getStorageValues` would make it inconsistent; reclassifying all of them is a separate, codebase-wide change beyond this PR's scope.

MysticRyuujin:
Thanks Greptile — both are accurate observations; my take on each:

- **Response address keys are lowercase** — keeping lowercase. This matches the reference client (go-ethereum returns lowercase `0x…` keys for `eth_getStorageValues`) and the execution-apis conformance fixture, which is keyed lowercase. Returning EIP-55 checksummed keys would actually diverge from the reference and fail the rpc-compat fixture, so lowercase is the correct choice here.
- **Decimal slot values vs `eth_getStorageAt`** — correct that there's an asymmetry, but I'm leaving `eth_getStorageValues` hex-only. The execution-apis storage-key schema is `bytesMax32` (hex); `eth_getStorageAt`'s acceptance of bare decimal strings is a non-standard leniency, and I'd rather a brand-new method adhere to the spec than propagate it. If the team prefers the two methods to accept identical slot formats, that's a quick follow-up (deserialize slots as strings and run each through `serde_utils::u256::deser_hex_or_dec_str`).

azteca1998 (approved):

crates/networking/rpc/eth/account.rs
```diff
@@ -251,6 +278,79 @@ impl RpcHandler for GetProofRequest {
...
+        if params.len() != 1 && params.len() != 2 {
+            return Err(RpcErr::BadParams("Expected 1 or 2 params".to_owned()));
+        };
+        // params[0] is an object mapping address -> array of storage slots.
+        let raw: HashMap<String, Vec<U256>> = serde_json::from_value(params[0].clone())?;
+        let mut requests = HashMap::with_capacity(raw.len());
+        let mut total_slots = 0usize;
+        for (address, slots) in raw {
+            total_slots += slots.len();
+            if total_slots > MAX_STORAGE_VALUES_SLOTS {
```
ElFantasma, line 305: The `MAX_STORAGE_VALUES_SLOTS = 1024` cap is the security boundary on this endpoint — it's the only thing stopping a single request from asking for arbitrary-many slot reads. Worth a unit test that asserts the cap fires:

```rust
#[test]
fn test_get_storage_values_request_rejects_over_cap() {
    let mut slots = Vec::with_capacity(MAX_STORAGE_VALUES_SLOTS + 1);
    for i in 0..=MAX_STORAGE_VALUES_SLOTS {
        slots.push(format!("{i:#x}"));
    }
    let params = Some(vec![json!({
        "0xdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef": slots
    })]);
    let err = GetStorageValuesRequest::parse(&params).unwrap_err();
    assert!(matches!(err, RpcErr::BadParams(ref m) if m.contains("too many slots")));
}
```

And a complementary one for the multi-account-totaling case (e.g. 600 slots on each of two accounts → over cap) since the cap is total-across-accounts, not per-account. The current `parse` test only exercises the happy path + the empty-request rejection; the cap is the un-tested security control. Non-blocking.

crates/networking/rpc/eth/account.rs
```diff
@@ -251,6 +278,79 @@ impl RpcHandler for GetProofRequest {
...
+        if total_slots == 0 {
+            return Err(RpcErr::BadParams("empty request".to_owned()));
+        }
+        Ok(GetStorageValuesRequest {
+            requests,
+            block: params
+                .get(1)
+                .map(|b| BlockIdentifierOrHash::parse(b.clone(), 1))
+                .transpose()?
+                .unwrap_or_default(),
```
ElFantasma, line 323: The optional-block-defaults-to-latest pattern is now duplicated 6 times in this file:

```rust
params
    .get(N)
    .map(|b| BlockIdentifierOrHash::parse(b.clone(), N))
    .transpose()?
    .unwrap_or_default()
```

…with only the index `N` varying. A small helper would centralize the spec-aligned behavior (and the next state method that lands in this file gets it for free):

```rust
/// Parse an optional block parameter at `index`, defaulting to `latest`
/// when omitted (per execution-apis: Block parameter is optional).
fn parse_optional_block(params: &[Value], index: usize) -> Result<BlockIdentifierOrHash, RpcErr> {
    params
        .get(index)
        .map(|b| BlockIdentifierOrHash::parse(b.clone(), index))
        .transpose()
        .map(Option::unwrap_or_default)
}
```

Non-blocking — the duplication is mechanical and reads fine inline.

MysticRyuujin:
Added the requested `eth_getStorageValues` slot-cap tests, including both single-account over-cap and multi-account total-over-cap coverage.

crates/networking/rpc/eth/account.rs
MysticRyuujin: Do you want me to add a helper? I can, up to you.

ElFantasma: You can leave inline, your call but the duplication is local enough that a helper isn't load-bearing.

ElFantasma (approved):

MysticRyuujin:
If there's anything left to do with this PR please let me know, we plan to merge the execution-apis PR soon, this behavior is merged in all other EL clients.

MysticRyuujin:
bump

PR #6781, opened 2026-06-02
fix(l1): reject getBlobsV1 after Osaka

emirongrr:
**Motivation**

The Osaka Engine API spec updates the previous Cancun API and requires `engine_getBlobsV1` to return `-38005: Unsupported fork` once Osaka has activated.

Spec reference: [`execution-apis/src/engine/osaka.md#cancun-api`](https://github.com/ethereum/execution-apis/blob/main/src/engine/osaka.md#cancun-api)

Before this change, ethrex continued serving `engine_getBlobsV1` after Osaka activation. That differs from the Osaka spec and from other execution clients:

- Reth only allows `engine_getBlobsV1` before Osaka and returns `UnsupportedFork` once Osaka is active: [`engine_api.rs`](https://reth.rs/docs/src/reth_rpc_engine_api/engine_api.rs.html#930-941)
- Geth restricts `engine_getBlobsV1` to Cancun/Prague and returns `unsupportedForkErr` otherwise: [`api.go`](https://github.com/ethereum/go-ethereum/blob/master/eth/catalyst/api.go#L568-L573)

**Description**

This PR adds the missing Osaka fork guard to `engine_getBlobsV1`.

`BlobsV1Request::handle` now checks the latest block header timestamp against the chain config and returns `RpcErr::UnsupportedFork` when Osaka is active.

A regression test was added:

- `blobs_v1_rejects_after_osaka`

This keeps `engine_getBlobsV1` aligned with the Osaka transition, where V1 is a pre-Osaka/Cancun API and Osaka blob retrieval should use the newer getBlobs methods.


crates/networking/rpc/engine/blobs.rs
```diff
@@ -64,6 +64,19 @@ impl RpcHandler for BlobsV1Request {
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
         debug!("Received new engine request: Requested Blobs");
 
+        if let Some(current_block_header) = context
```
ElFantasma, line 69: The `if let Some(current_block_header) = …` chain silently bypasses the gate when `get_block_header(get_latest_block_number())` returns `None` — i.e. on a fresh node with no blocks yet. In that state, `engine_getBlobsV1` would be accepted regardless of the configured Osaka activation.

In practice the window is narrow: a CL won't be calling `engine_getBlobsV1` on a node with no chain tip, and a node without any block also can't have an Osaka-active timestamp to gate against. So this isn't a runtime hazard today.

It's worth a one-line decision in the code, though — either:
- A comment noting the fall-through is intentional because pre-block / pre-tip means "no Osaka yet by definition", or
- An explicit `else` that returns `RpcErr::UnsupportedFork` (or `RpcErr::Internal`) for the no-header case, making the contract loud.

Identical concern applies to the matching `if let … &&` at `blobs.rs:160` — they'll want to be consistent.

azteca1998 (approved):

ElFantasma (approved):
Both call sites at blobs.rs:69 and :160 are clarified consistently with the "no tip ⇒ no Osaka by definition" reasoning — fine choice over the explicit-error path. 

LGTM, Thanks!

edg-l (approved):

PR #6782, opened 2026-06-02
perf(levm): reduce allocations and clones during block execution

iovoid:
**Motivation**

Currently we do several allocations and clones that could be avoided.

**Description**

In general:
- pre-allocating vectors with known or empirically obtained (using ethrex-replay) sizes, avoiding reallocations
- passing full vectors when references could work
- reusing the stack on the first level (32KiB)


edg-l (commented):
Reviewed the allocation/clone changes. One correctness concern; the rest checks out.

### `DestroyedModified` accounts can read a stale `0` after a streaming re-fault

`load_account` now clones info-only (`clone_without_storage`) unconditionally on the [initial-state fast path](https://github.com/lambdaclass/ethrex/blob/6c4d8ec6c5d31ad6530080fad8b8700f8f5775c9/crates/vm/levm/src/db/gen_db.rs#L357-L371). On the sequential streaming path the mid-block flush folds a destroyed-and-recreated account into `initial_accounts_state` [wholesale](https://github.com/lambdaclass/ethrex/blob/6c4d8ec6c5d31ad6530080fad8b8700f8f5775c9/crates/vm/levm/src/db/gen_db.rs#L772-L792), preserving `DestroyedModified` **and its storage**. A later tx re-faults it with empty storage.

The bug: in `get_storage_value` the `DestroyedModified` early-return sits **before** the new `initial_accounts_state` fallback, so after an info-only re-fault `current.storage` is empty, an `SLOAD` of a committed slot misses, hits the early-return, and yields `0` where the old full-clone returned the committed value; the fallback is never reached.

https://github.com/lambdaclass/ethrex/blob/6c4d8ec6c5d31ad6530080fad8b8700f8f5775c9/crates/vm/levm/src/db/gen_db.rs#L1011-L1017

https://github.com/lambdaclass/ethrex/blob/6c4d8ec6c5d31ad6530080fad8b8700f8f5775c9/crates/vm/levm/src/db/gen_db.rs#L1047-L1069

The checks can't simply be reordered: the early-return is required for the *within-tx* destroy->recreate case (an unwritten slot must read `0`, not the stale pre-destruction value in `initial`), and `DestroyedModified` can't distinguish that from "destroyed in a prior committed tx".

Scope: streaming executor only (pre-Amsterdam import, or Amsterdam with `--no-bal-parallel-exec`). The parallel BAL path and the builder path don't drain-back mid-block, so the info-only / `initial`-fallback code is inert there. Reachability is narrow (EIP-6780: needs same-tx create+selfdestruct+recreate surviving a flush, then read by a later tx), which likely explains why EF tests don't catch it.

Suggested fix; full-clone the rare destroyed case, keep the optimization for the hot path:

```rust
if let Some(account) = self.initial_accounts_state.get(&address) {
    let clone = match account.status {
        AccountStatus::Destroyed | AccountStatus::DestroyedModified => account.clone(),
        _ => account.clone_without_storage(),
    };
    return Ok(self.current_accounts_state.entry(address).or_insert(clone));
}
```

A targeted test (destroy+recreate+write in tx A, force a flush, `SLOAD` in tx B) would confirm/refute reachability.

### Verified sound
Precompile warmth (`is_warm_precompile` exactly replicates the old pre-seed range + `0x100`, `vm_type`-independent as before; Amsterdam reuses Prague's count); `&'static` opcode tables (fork is block-invariant); `Arc<[u32]>` jump_targets (serde `rc` enabled workspace-wide); `IntrinsicGas` dedup (same byte source/arithmetic, incl. the EIP-7976/7981 Amsterdam floor branches); `tx_calldata` rewrite (byte-identical, overflow -> `OutOfGas` preserved); stack/memory pooling (`reset_for_reuse` consistent with the existing no-read-unwritten invariant); `preserve_top_level_backup` gating.

azteca1998 (approved):

iovoid:
You are right. Added a test to cover that scenario.

crates/vm/levm/src/vm.rs
```diff
@@ -491,20 +538,115 @@ pub struct VM<'a> {
...
+        let mut root_stack = self.current_call_frame.stack;
+        root_stack.clear();
+        stack_pool.push(root_stack);
+        // Reclaim the root memory buffer with its grown capacity. `reset_for_reuse` truncates it
+        // to length 0 (capacity kept) so the next tx's `resize` zero-fills correctly. At this
+        // point only the root frame holds the buffer (sub-frames are dropped), so the `Rc` is
+        // unique on the success path; on the error path the about-to-drop frames are never read
+        // again, so clearing the shared buffer is harmless.
+        let mut root_memory = self.current_call_frame.memory;
+        root_memory.reset_for_reuse();
```
ElFantasma, line 650: Two correctness-invariants are riding on this one comment, and both are silent-failure modes if a future refactor breaks them:

1. **Rc uniqueness on success** — the comment claims "only the root frame holds the buffer." If a future change keeps a sub-frame's `Rc<RefCell<Vec<u8>>>` alive past the root frame (return-data plumbing, tracer-side capture, etc.), the buffer would be shared at `reset_for_reuse` time, the clear would propagate to the other holder, and the next tx's `resize` zero-fill invariant would silently still hold (good) — but the *other* holder would see length-0 mid-execution.
2. **Buffer reuse safety relies on `resize` zero-filling past `buf.len()`** — already documented at `reset_for_reuse`, but it's the same chain.

Cheap insurance:
```rust
debug_assert_eq!(
    Rc::strong_count(&root_memory.buffer), 1,
    "root memory buffer must be Rc-unique at reclaim; \
     a sub-frame is still holding it and would observe the reset_for_reuse clear",
);
```
After `reset_for_reuse` it's too late — by then the clear has already propagated. So the assert needs to come *before* `root_memory.reset_for_reuse()`. Same idea would apply to the root stack if anything ever Rc-shared that, but it doesn't today. Non-blocking.

crates/vm/levm/src/vm.rs
```diff
@@ -537,18 +679,23 @@ impl<'a> VM<'a> {
             (0, 0, 0, 0, 0)
         };
 
+        // L1 block execution installs no `BackupHook` (see `l1_hooks`), so the top-level backup
+        // is unread after restore and can be consumed; L2 / stateless keep it for the hook.
+        let preserve_top_level_backup = matches!(vm_type, VMType::L2(_));
```
ElFantasma: The flag's *true* semantic is "a `BackupHook` is installed and will read the top-level backup in `finalize_execution`," not specifically "VMType::L2." Today the two coincide because `get_hooks(VMType::L2(_))` installs `BackupHook` and `stateless_execute` adds it imperatively (with the corresponding `self.preserve_top_level_backup = true` at vm.rs:1142). But:

- If a future `vm_type` (e.g. an L1 variant with state-replay hooks) adds `BackupHook`, the L1 path would silently skip the clone and the hook would read a stale/dropped backup.
- If `VMType::L2(_)`'s `get_hooks` ever drops `BackupHook`, the L2 path would pay the deep-clone for nothing.

Two shapes that survive the future change:
1. **Derive from `hooks`**: walk `self.hooks` once at construction and set the flag iff any is a `BackupHook`. Adds a downcast or a `Hook::reads_top_level_backup() -> bool` method.
2. **Rename for what it actually means**: `backup_hook_reads_top_level` or similar, and audit that `get_hooks` and `stateless_execute` are the *only* setters. Cheaper; the `VMType` mapping becomes a documented invariant rather than the truth.

Non-blocking — the current code is correct. This is about making the next-person-who-adds-a-hook's job easier.

ElFantasma (approved):
Two non-blocking inlines below — both are about making correctness invariants slightly harder to break with future refactors. Quality of work is high; this is just lock-in.

edg-l (approved):

crates/vm/levm/src/vm.rs
iovoid: Added the reads_top_level_backup method.

crates/vm/levm/src/vm.rs
iovoid: Added the debug_assert.

ilitteri (approved):

PR #6783, opened 2026-06-03
feat(l1): add --skip-genesis-validation to trust a pre-existing datadir

edg-l:
## What

Adds `--skip-genesis-validation` (env `ETHREX_SKIP_GENESIS_VALIDATION`). When the datadir already contains a genesis header, ethrex trusts it instead of recomputing the genesis state root from the genesis `alloc` and rejecting on mismatch.

## Why

On boot, `init_store` → `add_initial_state` recomputes the genesis block hash from the `--network` file, where `state_root = compute_state_root(alloc)`, and compares it to the stored genesis header. If they differ it returns `IncompatibleChainConfig` and refuses to boot.

That blocks booting a datadir built out-of-band — e.g. a tool that writes the state trie straight into RocksDB and ships a genesis file with an empty `alloc` (so it doesn't have to materialize millions of accounts as JSON). The recomputed root is then `EMPTY_TRIE_HASH`, which never matches the real stored root, so ethrex always rejects it.

Reproduced against the current behavior: build a DB from a populated genesis, reopen with the same genesis but `alloc` emptied → `IncompatibleChainConfig`.

reth (`--debug.skip-genesis-validation`), besu (`--genesis-state-hash-cache-enabled`) and nethermind (`stateUnavailable: true`) all expose an equivalent "trust the stored state root" switch for this exact pattern; ethrex had none.

## How

- `Store::add_initial_state_skip_validation` — same as `add_initial_state` but, when a genesis header is already stored, returns early and trusts it. `set_chain_config` still runs; a fresh datadir is built normally.
- `init_store_trusting` + `init_l1` branches on the new flag.
- `--skip-genesis-validation` on `Options`.

The operator opts in explicitly and vouches for the datadir, so the safety check is waived rather than removed.

## Test

Extended `test_genesis_block`: the mismatching genesis that returns `IncompatibleChainConfig` under normal validation now returns `Ok` under skip-validation (stored genesis kept as-is).

## Notes

Draft — opening for early feedback on the flag name/shape before polishing. Happy to add a CHANGELOG entry once the section/format is confirmed.

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -150,6 +150,17 @@ pub async fn init_store(datadir: impl AsRef<Path>, genesis: Genesis) -> Result<S
     Ok(store)
 }
 
+/// Like [`init_store`], but trusts a pre-existing datadir's genesis instead of
+/// validating it against `genesis`. See [`Store::add_initial_state_skip_validation`].
+pub async fn init_store_trusting(
+    datadir: impl AsRef<Path>,
+    genesis: Genesis,
+) -> Result<Store, StoreError> {
+    let mut store = open_store(datadir.as_ref())?;
+    store.add_initial_state_skip_validation(genesis).await?;
+    Ok(store)
+}
+
 /// Initializes a pre-existing Store
 pub async fn load_store(datadir: &Path) -> Result<Store, StoreError> {
     let store = open_store(datadir)?;
@@ -542,7 +553,12 @@ pub async fn init_l1(
     debug!("Preloading KZG trusted setup");
     ethrex_crypto::kzg::warm_up_trusted_setup();
 
-    let store = match init_store(&datadir, genesis).await {
+    let store_result = if opts.skip_genesis_validation {
+        init_store_trusting(&datadir, genesis).await
+    } else {
+        init_store(&datadir, genesis).await
+    };
+    let store = match store_result {
         Ok(store) => store,
         Err(err @ StoreError::IncompatibleDBVersion { .. })
         | Err(err @ StoreError::NotFoundDBVersion) => {
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -152,7 +152,7 @@ pub async fn init_store(datadir: impl AsRef<Path>, genesis: Genesis) -> Result<S
 
 /// Like [`init_store`], but trusts a pre-existing datadir's genesis instead of
 /// validating it against `genesis`. See [`Store::add_initial_state_skip_validation`].
-pub async fn init_store_trusting(
+pub async fn init_store_skip_validation(
     datadir: impl AsRef<Path>,
     genesis: Genesis,
 ) -> Result<Store, StoreError> {
@@ -554,7 +554,7 @@ pub async fn init_l1(
     ethrex_crypto::kzg::warm_up_trusted_setup();
 
     let store_result = if opts.skip_genesis_validation {
-        init_store_trusting(&datadir, genesis).await
+        init_store_skip_validation(&datadir, genesis).await
     } else {
         init_store(&datadir, genesis).await
     };
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -2270,6 +2270,30 @@ impl Store {
     }
 
     pub async fn add_initial_state(&mut self, genesis: Genesis) -> Result<(), StoreError> {
+        self.add_initial_state_inner(genesis, false).await
+    }
+
+    /// Like [`Store::add_initial_state`], but trusts a pre-existing datadir
+    /// instead of validating it against the provided genesis. If a genesis
+    /// header is already stored, it is kept as-is rather than recomputing the
+    /// genesis state root from `genesis.alloc` and rejecting on mismatch.
+    ///
+    /// Intended for booting a datadir produced out-of-band (e.g. by a state
+    /// generator that writes the state trie directly and emits a genesis file
+    /// with an empty `alloc`), where the operator vouches for the stored state
+    /// root. Has no effect on a fresh datadir: the genesis is built normally.
+    pub async fn add_initial_state_skip_validation(
+        &mut self,
+        genesis: Genesis,
+    ) -> Result<(), StoreError> {
+        self.add_initial_state_inner(genesis, true).await
+    }
+
+    async fn add_initial_state_inner(
+        &mut self,
+        genesis: Genesis,
+        skip_genesis_validation: bool,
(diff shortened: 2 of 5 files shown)


azteca1998 (approved):

crates/storage/store.rs
```diff
@@ -2290,6 +2314,13 @@ impl Store {
         }
 
         match self.load_block_header(genesis_block_number)? {
+            Some(header) if skip_genesis_validation => {
```
ElFantasma, line 2325: Subtle: this early-return runs *after* `set_chain_config(&genesis.config)` at line ~2307, which per its own doc ("Ignores previously stored values if present") unconditionally overwrites whatever was stored. So the actual semantic of `--skip-genesis-validation` is:

- Stored genesis block (header + state root): **trusted** ✓
- Stored chain config: **silently overwritten** with `genesis.config` from the supplied file ✗

The PR description frames the flag as "trust a pre-existing datadir's genesis" — an operator reading that would reasonably expect *both* to be trusted. In the documented use case (state-generator ships the same `genesis` minus `alloc`), the chain configs are identical so the overwrite is a no-op. But if the operator mistakenly supplies a `genesis` file with a different `chainId` / fork schedule, the flag silently lets the wrong config win.

Two reasonable shapes:
1. **Symmetric**: also skip `set_chain_config` when the existing genesis header is present in skip-validation mode — trust everything the datadir already has.
2. **Warn-on-divergence**: keep the overwrite but log at WARN level when the supplied `chain_config` differs from the stored one, so the operator gets a chance to notice.

Worth picking one before merge, or at minimum updating the PR / CLI help to say "trust the stored genesis *block*; chain config is still taken from the supplied genesis file."

docs/CLI.md
```diff
@@ -28,14 +28,14 @@ Options:
...
+          
           [env: ETHREX_NETWORK=]
 
       --datadir <DATABASE_DIRECTORY>
           Base directory for the database. For public networks a subdirectory named after the network is appended (e.g. ~/.local/share/ethrex/mainnet). If the value is `memory`, the InMemory Engine is used instead.
-
+          
           [env: ETHREX_DATADIR=]
-          [default: /home/runner/.local/share/ethrex]
+          [default: /home/edgar/.local/share/ethrex]
```
ElFantasma: nit: `/home/edgar/...` leaked from your local machine into the generated doc (the upstream line was `/home/runner/...`). The CLI.md file is auto-regenerated, so this — plus the trailing-whitespace flips on a bunch of unchanged blank lines further down — looks like a regen artifact from a non-CI run.

Either regenerate on a CI-equivalent path / strip trailing whitespace post-regen, or split the genuine `--skip-genesis-validation` doc addition from the noise so the diff stays focused. Doesn't block the feature; just makes the PR easier to review and avoids landing your username in docs.

cmd/ethrex/initializers.rs
```diff
@@ -150,6 +150,17 @@ pub async fn init_store(datadir: impl AsRef<Path>, genesis: Genesis) -> Result<S
     Ok(store)
 }
 
+/// Like [`init_store`], but trusts a pre-existing datadir's genesis instead of
+/// validating it against `genesis`. See [`Store::add_initial_state_skip_validation`].
+pub async fn init_store_trusting(
```
ElFantasma: nit: naming asymmetry — this wrapper is `init_store_trusting` but the underlying store API is `add_initial_state_skip_validation`. Same concept, two different verbs ("trusting" vs "skip_validation"). Picking one — probably `_skip_validation` since it already matches the CLI flag `--skip-genesis-validation` and the env var — would make grep + navigation easier. Non-blocking.

test/tests/storage/store_tests.rs
```diff
@@ -139,9 +139,17 @@ async fn test_genesis_block(mut store: Store) {
...
-    let result = store.add_initial_state(genesis_hive).await;
+    let result = store.add_initial_state(genesis_hive.clone()).await;
     assert!(result.is_err());
     assert!(matches!(result, Err(StoreError::IncompatibleChainConfig)));
+
+    // With validation skipped, the mismatching genesis is trusted instead of
+    // rejected: the stored (kurtosis) genesis is kept as-is and the call
+    // returns Ok rather than IncompatibleChainConfig.
+    store
+        .add_initial_state_skip_validation(genesis_hive)
```
ElFantasma, line 187: Test verifies that skip-validation returns `Ok` instead of `IncompatibleChainConfig`, which is the core property — but it doesn't assert anything about *what* is stored afterwards. Given the `set_chain_config` overwrite I flagged at `store.rs:~2307`, this test silently accepts whichever semantic the implementation picks.

Two low-cost extensions that would pin the behaviour:
1. After the skip-validation call, load the genesis header and assert it still matches the *kurtosis* hash (the stored one), not the *hive* one — proves the genesis block was actually preserved.
2. Load the chain config back and assert which network's config ended up there — this surfaces the overwrite question concretely so the test fixes whatever the agreed semantic is.

Also worth: a fresh-datadir case (`add_initial_state_skip_validation` against a new store with no genesis) that asserts the genesis is built normally — confirms the "has no effect on a fresh datadir" claim in the doc.

crates/storage/store.rs
```diff
@@ -2278,8 +2302,19 @@ impl Store {
...
+        // rather than overwriting it from the supplied (possibly stripped)
+        // genesis file. Overwriting could silently swap the fork schedule or
+        // chainId out from under a datadir we were told to trust.
+        let trust_stored = skip_genesis_validation && stored_genesis_header.is_some();
+
         // Set chain config
-        self.set_chain_config(&genesis.config).await?;
+        if !trust_stored {
+            self.set_chain_config(&genesis.config).await?;
+        }
```
iovoid: This function is also where self.chain_config is set. Leaving it unset would leave the chain config in the default (incorrect) state.

iovoid (changes requested):

crates/storage/store.rs
edg-l: You're right, this was a bug. `Store::new` initializes `chain_config: Default::default()` and nothing reloads the persisted `ChainConfig` into `self.chain_config` on boot — only `set_chain_config` populates it. Skipping it left the store with the wrong chainId and an empty fork schedule.

Fixed in c06ac5f64: `set_chain_config(&genesis.config)` now always runs. Skip-validation only waives the genesis state-root/header check; the `config` section of the genesis file (valid even with a stripped `alloc`) is always applied. Updated the test accordingly — it now asserts the stored genesis *block/state* is trusted while the chain *config* comes from the supplied genesis file.

iovoid (approved):

ElFantasma (approved):

PR #6787, opened 2026-06-04
fix(l1): guard derive_legacy_chain_id against underflow on malformed v

edg-l:
`derive_legacy_chain_id` only guarded against `v == 27 || v == 28`, so any `v < 35` that wasn't 27/28 (e.g. `v = 0` from an unsigned tx) would hit `v - 35` and either panic in debug/test builds or wrap silently in release.

Fix replaces the 27/28 guard with `if v < 35 { None }`. This subsumes pre-EIP-155 values and rejects all sub-35 inputs, none of which carry an EIP-155 chain id (`v = chain_id*2 + 35/36` implies `v >= 35`).

Found via the EIP-7805 execution-spec test `block_with_invalid_signature_pending_il_tx_is_valid`, which passes a `v=0, r=0, s=0` legacy tx through the IL satisfaction check. The bug predates FOCIL (present since ~Oct 2024); this PR targets `main` independently.

`cargo test -p ethrex-common --lib` passes (79 tests).

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -1527,11 +1527,11 @@ impl Transaction {
 
 fn derive_legacy_chain_id(v: U256) -> Option<u64> {
     let v = u64::try_from(v).ok()?;
-    if v == 27 || v == 28 {
-        None
-    } else {
-        Some((v - 35) / 2)
-    }
+    // EIP-155 encodes the chain id as `v = chain_id * 2 + 35` (or 36), so any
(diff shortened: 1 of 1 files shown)


ilitteri (approved):

MegaRedHand (approved):

ElFantasma (approved):
I was about to suggest a regression test locking in the boundary cases (v ∈ {0, 26, 27, 28, 29, 34} → None and v ∈ {35, 36, 37, 38} → Some(0/1)), but it will land via #6817 as `legacy_chain_id_handles_out_of_range_v_without_underflow`, which exercises the same matrix.

PR #6788, opened 2026-06-04
fix(l1): share merkleization pool across rpc test contexts

edg-l:
`Blockchain::default_with_store` builds a fresh 17-thread rayon merkle pool per call (#6344). The rpc test harness was calling it once per test context, and `cargo test` runs tests in parallel, so live thread count = (parallel tests) x (17 merkle + tokio threads). On macOS the per-process task-thread limit is much lower than Linux, so `build_merkle_pool().expect(...)` hits `pthread_create` EAGAIN and panics with "Failed to create merkle thread pool" in a rotating set of unrelated rpc tests (`net_version_test`, `http_api_allowlist_*`, `ws_subscribe_*`, `eth_config_*`). This has been intermittently failing the L1 `Test - macos-15` CI job.

Fix: add a `thread_local!` `MERKLE_POOL` (`OnceCell`) in `test_utils.rs` and route all 4 test-context constructors through `Blockchain::default_with_store_and_pool(store, merkle_pool())`. A per-thread pool satisfies the safety invariant documented on `default_with_store_and_pool`: only one concurrent `in_place_scope` caller per pool; since each cargo test thread runs its tests sequentially, the invariant holds. A single global pool would deadlock under concurrent test threads. This mirrors the pattern added in #6665 for `tooling/ef_tests/blockchain`, `tooling/ef_tests/engine`, and `tooling/ef_tests/state_v2`, which was never applied to the rpc test harness.

`rayon` is added as a non-dev dependency in `crates/networking/rpc/Cargo.toml` because `pub mod test_utils` compiles into the normal lib target.


ilitteri (approved):

ElFantasma (approved):
LGTM

iovoid (approved):

PR #6791, opened 2026-06-04
fix(l1): stop blob-inclusion decay in the mempool

edg-l:
ethrex-built blocks decayed toward zero blobs over time while peers kept including them. Blob txs arrived fine; the loss was in mempool retention/selection. Three fixes:

1. **Dedicated blob sub-pool** (`mempool.rs`): blob txs get their own cap (`MAX_BLOB_MEMPOOL_SIZE`) and eviction path, so a regular-tx flood can't evict them.
2. **Value/nonce eviction** (`remove_worst_blob_transaction`): over cap, drop the highest-nonce/lowest-fee blob first, not the earliest-inserted, keeping the includable low-nonce ones. Cap 512.
3. **Stale-nonce reset** (`remove_stale_blob_txs`): per-block pruning only covers the FCU head block, so blobs from non-head canonical blocks leaked in and got pinned by (2). Drop blob txs below the sender's on-chain nonce on each fork choice.


crates/blockchain/mempool.rs
```diff
@@ -192,17 +241,32 @@ impl Mempool {
...
+        // can't push them out (and vice versa).
+        if is_blob {
+            // The bundle is inserted before the tx (see add_blob_transaction_to_pool),
+            // so the incoming blob is already counted by `blob_tx_count`.
+            if inner.blob_tx_count() > inner.max_blob_mempool_size {
+                inner.remove_oldest_blob_transaction()?;
+            }
+            inner.blob_txs_order.push_back(hash);
+        } else {
+            if inner.regular_tx_count() >= inner.max_mempool_size {
```
ElFantasma, line 323: Small asymmetry worth pinning down in a comment: the blob check on line 260 uses `>` ("already over cap, evict") while the regular check here uses `>=` ("at cap, evict before adding"). Both are correct as written:

- Blob path: the bundle is inserted *before* `add_transaction` runs (line 2578 of blockchain.rs), so `blob_tx_count()` already includes the incoming tx. `> max` means we're now strictly over cap.
- Regular path: the tx is inserted *after* this check (line 273), so `regular_tx_count()` is the count *before* the incoming tx. `>= max` means "we're at cap, make room."

Both land in the right state (cap respected, incoming tx admitted). But the difference in operator (`>` vs `>=`) reads as accidental, and the existing inline comment ("The bundle is inserted before the tx... so the incoming blob is already counted") only explains the blob side. A symmetric comment on the regular branch — "the regular tx isn't in the pool yet, so `>= max` means we're at cap and need to evict" — would make the two paths read uniformly. Non-blocking, code is right.

ElFantasma (commented):
**Cross-PR with #6607** (min-heap tip eviction, also touches mempool.rs): #6607 replaces `txs_order` (VecDeque) with `txs_by_tip` (BinaryHeap). This PR adds `blob_txs_order` as a parallel VecDeque. Whichever lands second will need a small rebase — and the design question to settle is whether the blob sub-pool should *also* go tip-based or stay FIFO. Your PR body argues correctly for type-based-then-FIFO (low-tip blobs are still high value because the builder pays per-blob slot, not per-tip). Worth a one-line acknowledgment in #6607's eventual rebase that the blob sub-pool intentionally stays FIFO.

crates/blockchain/mempool.rs
```diff
@@ -124,22 +143,73 @@ impl MempoolInner {
...
+    /// slowly starve block building of blobs.
+    fn remove_worst_blob_transaction(&mut self) -> Result<(), StoreError> {
+        while self.blob_tx_count() > self.max_blob_mempool_size {
+            // `blobs_bundle_pool` is keyed by blob-tx hash, so its keys are
+            // exactly the blob txs currently held.
+            let worst = self
+                .blobs_bundle_pool
+                .keys()
+                .filter_map(|hash| self.transaction_pool.get(hash).map(|tx| (*hash, tx)))
+                .max_by_key(|(_, tx)| (tx.nonce(), Reverse(tx.max_fee_per_blob_gas())))
```
ElFantasma: **Cross-sender nonce comparison is meaningless as a "worst" key.**

The commit message frames the key as "highest nonce = most likely to sit behind a nonce gap, so not includable yet" — but that intuition only holds *within a sender*. `tx.nonce()` is the on-wire nonce, not a gap-from-state, so comparing nonces across senders just penalizes senders with long histories.

Concrete:
- Alice: on-chain nonce 99, submits blob at nonce 100 (immediately includable, `gap = 0`)
- Bob:   on-chain nonce 0,  submits blob at nonce 1   (also immediately includable, `gap = 0`)

Under this key, Alice's tx looks "worse" than Bob's — but both are equally includable. A high-throughput blob sender (e.g. a rollup sequencer running continuously) accumulates a high tx.nonce() and gets preferentially evicted in favor of brand-new senders, even when the high-nonce sender's txs are the *most* includable.

Alternatives that don't need state reads on the eviction path:
1. **Drop the nonce term entirely** — sort by `Reverse(max_fee_per_blob_gas())` (lowest blob fee first). Geth's blobpool effectively does this (with a price floor). The "stale nonce" concern is now handled directly by `remove_stale_blob_txs` in `66cdfad9`, which is the right place because it has actual on-chain nonces.
2. **Use the per-sender gap** via `txs_by_sender_nonce.range((sender, 0)..(sender, u64::MAX)).count()` — pick the sender with the deepest queue. More complex.

Option 1 is the cleanest given `66cdfad9` already covers staleness. Non-blocking; flagging because the current key has a worst-case adverse effect on exactly the population this PR is trying to protect (active blob users).

crates/blockchain/mempool.rs
```diff
@@ -124,22 +143,73 @@ impl MempoolInner {
...
+                .map(|(hash, _)| hash);
+            match worst {
```
ElFantasma, line 225: **O(N) scan in a `while`-loop on the admission hot path.**

Each iteration walks `blobs_bundle_pool.keys()` (up to N=512 entries) to find a single victim. In steady state — pool sits at-cap, every admission triggers one eviction — that's ~512 hashmap lookups + tx pointer chases per `add_transaction`. Acceptable at 512, but it'd be worth a comment noting the complexity so future cap bumps (the TODO for CLI-exposure) don't accidentally make this quadratic.

A `BinaryHeap<(EvictKey, H256)>` mirroring `transaction_pool` would amortize to O(log N), at the cost of maintenance on `remove_transaction_with_lock`. Probably overkill for 512 entries today; just worth flagging the dimension before the cap is exposed via CLI.

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -290,6 +290,12 @@ async fn handle_forkchoice(
                     context
                         .blockchain
                         .remove_block_transactions_from_pool(&block)?;
+                    // Reset blob sub-pool against on-chain nonces (head-block
+                    // pruning above misses stale blobs from non-head blocks).
+                    context
+                        .blockchain
+                        .remove_stale_blob_txs(block.hash())
```
ElFantasma: **`remove_stale_blob_txs` runs synchronously inside the FCU response path with `?`-propagation.**

For every accepted FCU, this now does up to N state-lookups (`get_account_info_by_hash` per unique blob-tx sender, capped at `MAX_BLOB_MEMPOOL_SIZE = 512`). The in-function HashMap cache deduplicates across blob txs from the same sender (helpful — a rollup sequencer dominates with one entry), but worst case is still 512 cold state reads gating the FCU response to the CL.

Two options:
1. **Fire-and-forget**: `tokio::spawn(blockchain.remove_stale_blob_txs(hash))` after the FCU ack — the cleanup is best-effort housekeeping, not consensus-load-bearing, and the next FCU will catch any leak the spawned task fails on.
2. **Keep awaiting but downgrade errors**: log+continue on `Err` instead of `return Err(...)` — a state-read failure during cleanup shouldn't fail the FCU itself.

Option 1 is the cleaner shape; option 2 is the smaller diff if you'd rather keep this on the FCU's critical path. Either way, propagating with `?` to the CL feels wrong for a sub-pool cleanup.

ElFantasma (commented):
Refreshing this review against the three new commits (`19fdde64` FIFO→value/nonce eviction, `66cdfad9` head-change reset, `10393e10` saturating_sub). My original framing — "FIFO blob sub-pool is fine because builders pay per-blob-slot, not per-tip" — no longer matches the PR; you've moved to value/nonce eviction, which is closer in spirit to #6607's tip-heap. That's a reasonable pivot (FIFO would have starved low-nonce includable blobs under sustained backlog, which `19fdde64`'s commit message correctly identifies).

Three commits land good things: `10393e10` is the right fix for the Codex underflow finding (transient-undercount-as-0 is the safe direction); `66cdfad9`'s head-change reset closes a real leak (per-block pruning only covers the head block of an FCU jump); and the cap drop from 1024 → 512 with the new memory-bound documentation (~0.5 GiB worst case at ~1 MiB/sidecar) is more honest.

Two non-blocking concerns on the new eviction strategy and one operational concern on the FCU path inlined. The cross-PR friction with #6607 (which I flagged originally) is now lower, since both PRs are converging on fee-aware eviction — worth a one-line acknowledgment in #6607's eventual rebase that the blob sub-pool has its own value/nonce key (it does NOT plug into #6607's `txs_by_tip`).

crates/blockchain/mempool.rs
edg-l: Kept nonce-based eviction but made it per-sender, since (as you note) the intuition only holds within a sender: rank by `tx.nonce() - sender_min_pooled_blob_nonce`, tie-broken by lowest blob fee. A sequencer at high nonce now has offset 0 and is no longer penalized vs a fresh low-nonce sender. Didnt drop the nonce term (Option 1) so the includability ordering this PR adds stays intact. Added a regression test (`blob_eviction_offset_is_per_sender_not_cross_sender`).

crates/blockchain/mempool.rs
edg-l: Added a comment noting the O(N) scan and N <= 512, flagging it to revisit before the cap is CLI-exposed. Left it as a linear scan for now (per-sender min map is an extra O(N) pass on top, still fine at this cap).

crates/blockchain/mempool.rs
edg-l: Added the symmetric comment on the regular branch explaining the `>=` (tx not yet inserted = at cap), mirroring the blob branch `>`.

ElFantasma (approved):

ilitteri (approved):

crates/blockchain/mempool.rs
```diff
@@ -35,6 +36,22 @@ use tracing::warn;
...
+/// Maximum number of blob (EIP-4844) transactions retained in the mempool,
+/// independent of `max_mempool_size`. Blob txs live in a dedicated sub-pool so a
+/// flood of regular transactions cannot evict them, and the sub-pool itself is
+/// evicted by value/nonce (see `remove_worst_blob_transaction`), never FIFO, so
+/// the node keeps the (scarce, high-value, includable) blob txs it needs to build
+/// full blocks.
+///
+/// Sized to comfortably hold several blocks' worth of includable blobs (Amsterdam
+/// allows up to 21 blobs/block) while bounding worst-case memory: blobs are held
+/// in RAM, so the bound is this count times the per-tx sidecar (`MAX_BLOB_TX_SIZE`,
```
iovoid: I would remove the word "sidecar", since MAX_BLOB_TX_SIZE is the limit of the whole transaction.

crates/blockchain/mempool.rs
```diff
@@ -35,6 +36,22 @@ use tracing::warn;
...
+/// ~1 MiB) ⇒ ~0.5 GiB worst case.
+///
+/// TODO(#6691-fu): expose through CLI and prefer a byte-based cap (like geth's
```
iovoid: Maybe open an issue? Referencing a closed PR is confusing, and the meaning of '-fu' isn't immediately clear. Same for the other `TODO(#6691-fu)`.

crates/blockchain/blockchain.rs
```diff
@@ -2753,6 +2753,36 @@ impl Blockchain {
         Ok(())
     }
 
+    /// Drop blob txs with nonce below the sender's on-chain nonce at `head_hash`.
+    /// Per-block pruning only covers the head block, so stale blob txs from
+    /// non-head canonical blocks leak in and are never evicted (value/nonce
+    /// eviction pins low nonces). Resetting against on-chain nonces clears them.
+    pub async fn remove_stale_blob_txs(&self, head_hash: BlockHash) -> Result<(), StoreError> {
```
iovoid: This function doesn't need to be async

iovoid (approved):

PR #6792, opened 2026-06-05
perf(l1): replace pure-Rust BLS12-381 with blst in NativeCrypto

edg-l:
The portable `bls12_381` defaults on the `Crypto` trait normalize result points to affine via Fermat inversion (exponentiation by p-2), which dominates wall-clock time on the cheap G1ADD/G2ADD precompiles (375/600 gas) and makes them ~6-7x slower than expected.

`blst` uses assembly-optimized binary-GCD inversion. It is already in the dependency tree via `c-kzg`, so this just wires it in. A new default-on `blst` cargo feature (implying `std`) overrides the 7 BLS12-381 ops on `NativeCrypto`: g1_add, g1_msm, g2_add, g2_msm, pairing_check, fp_to_g1, fp2_to_g2. zkVM/no_std guest builds omit the feature and keep the portable trait defaults unchanged.

New `crates/common/crypto/bls_blst.rs` contains safe blst wrappers with full EIP-2537 semantics: canonical field check, on-curve check, subgroup checks on MSM/pairing, point-at-infinity handling. `native.rs` gains the feature-gated `impl Crypto for NativeCrypto` with the overrides.

### Why it's faster (not just asm)

The main gain is an algorithm swap in the field inversion, not the assembly. A G1/G2 ADD does a cheap projective point addition (~12-16 field muls) and then one modular inverse to convert the result back to affine; that single inverse is the bulk of the work.

- Old path computes `z^(p-2) mod p` (Fermat) over a 381-bit modulus: ~381 squarings + ~190 multiplications, so ~570 full field multiplications.
- `blst` uses Bernstein-Yang "safegcd", a batched extended-GCD that finishes in ~18 word-sized (62-bit) divstep iterations with only a handful of multi-precision multiplications, roughly an order of magnitude fewer big muls.

Assembly limb arithmetic adds maybe another ~1.5-2x on top, but the inversion algorithm change is the dominant factor. ADD shows it most because it pays one full inversion for very little gas; MSM/pairing amortize the inversion and are priced much higher, so the cost was hidden there. EIP-2537 does not require constant-time precompiles, so the old constant-time exponentiation was pure overhead; safegcd is also constant-time but inherently cheaper.

### MSM is called single-threaded

`g1_msm`/`g2_msm` invoke blst's Pippenger directly through the C `blst_p1s_mult_pippenger` / `blst_p2s_mult_pippenger` entry with a local scratch buffer, instead of the safe `MultiPoint::mult` wrapper. That wrapper forks every call across blst's global thread pool. Inside the (already serial) EVM precompile path that fork/join is pure overhead: it adds per-call latency and its worker threads contend with the node's many runnable threads, so on a CPU-loaded node MSM measured *slower* than the pure-Rust path even though blst MSM is 3-16x faster in isolation. Calling the single-threaded entry keeps the fast Pippenger with no thread spawning. Add/pairing/map never touched the pool and are unaffected.

This was caught on benchmarkoor: after the initial blst image, every serial op (add, pairing, map) sped up sharply while only the two MSM ops regressed; the common factor was `MultiPoint::mult` and its thread pool.

### Skip identity points in MSM

A point at infinity is the group identity: it lies in every subgroup and contributes nothing to a multi-scalar multiplication. `g1_msm`/`g2_msm` now detect it immediately after the on-curve decode and skip the term entirely, bypassing both the subgroup check (which the identity trivially passes) and the full 255-bit scalar multiplication. Non-infinity points keep their unconditional subgroup check, so EIP-2537 input validation is unchanged — a non-subgroup point paired with a zero scalar is still rejected.

This removes wasted work whenever an MSM input is the identity. It is easy to hit in practice: multiplying any subgroup point by the group order yields the identity, and a result reused as a later input then stays at infinity. Dropping a single infinity term measures ~131 µs → ~0.3 µs (G1) and ~236 µs → ~0.6 µs (G2) in isolation; real, non-identity terms are unaffected. End-to-end on the benchmark suite this shows up on the all-distinct-input ("uncachable") MSM cases, which reduce to repeated identity operations: ethrex's per-test time drops from ~681 ms to ~5.7 ms for G1MSM (~120x) and from ~780 ms to ~6.0 ms for G2MSM (~130x).

Adds `crates/common/crypto/benches/bls_single_op.rs` (criterion) isolating the per-op cost (full MSM k=1, subgroup check alone, scalar mul alone, identity input, scalar == group order), plus tests covering infinity terms in MSM and the `P·order == identity` case.

### Build gating (verified)

- L1 `ethrex` binary: `blst` is enabled (`cargo tree` resolves `ethrex-crypto` to `blst,c-kzg,default,kzg-rs,secp256k1,std`); `cmd/ethrex` sets `default-features = true`, re-enabling it past the workspace `default-features = false`.
- zkVM guests (sp1/risc0/openvm): `blst` feature off, so `bls_blst.rs` is excluded and the portable defaults are used; `std` alone does not pull `blst`.

Validation: 1986 EIP-2537 EF state-test vectors pass (Prague 975, Osaka 1004, incl. all G1MSM/G2MSM); the engine-format EF suite passes (5598 tests, incl. the BLS not-in-subgroup rejection vectors, confirming the reordered subgroup check); differential tests confirm blst matches the `bls12_381` reference byte-for-byte for add/msm/pairing/map; microbench shows g1_add 7.7x faster (3181 ns vs 24426), g2_add 5.6x faster (4834 ns vs 26830), and single-threaded MSM staying 3-16x faster than the pure-Rust path across k=1..128; clippy clean; no-default-features build compiles.


test/tests/levm/bls12_tests.rs
```diff
@@ -46,3 +46,166 @@ fn pairing_infinity() {
 
     assert_eq!(result.unwrap(), zero);
 }
+
+// ── blst backend differential tests ─────────────────────────────────────────
+//
+// `NativeCrypto` routes BLS12-381 through the blst backend (the `blst` feature,
+// on by default). These tests assert it agrees byte-for-byte with the portable
+// `bls12_381` trait-default implementation (`Ref` below), which already passes
```
iovoid, line 54: The test could end up proving nothing if the blst feature is disabled.

crates/common/crypto/bls_blst.rs
```diff
@@ -0,0 +1,557 @@
...
+#[inline]
+fn p1_msm(points: &[blst_p1_affine], scalars: &[u8]) -> blst_p1 {
+    let npoints = points.len();
+    let p: [*const blst_p1_affine; 2] = [points.as_ptr(), core::ptr::null()];
+    let s: [*const u8; 2] = [scalars.as_ptr(), core::ptr::null()];
+    let mut out = blst_p1::default();
+    // SAFETY: `npoints >= 1`, `scalars` is `npoints * 32` bytes, and scratch is
+    // sized by blst's own helper.
+    unsafe {
+        let mut scratch = vec![0u64; blst_p1s_mult_pippenger_scratch_sizeof(npoints) / 8];
```
ElFantasma: Per-call `vec![0u64; ...]` allocation for each MSM. For EIP-2537 G1MSM/G2MSM the point counts are small (typical k ≤ ~10), so the scratch is only a few KB and the allocation overhead is probably noise compared to the Pippenger work — but it's an unbounded malloc on a hot precompile path. Worth a follow-up to thread-local-pool the scratch (`std::cell::RefCell<Vec<u64>>` with `reserve` + `resize`) so steady-state MSM calls hit zero heap.

Separately: the divisor 8 assumes `sizeof(limb_t) == 8` (true on x86_64/aarch64, which is all ethrex targets). On a hypothetical 32-bit build `blst_p1s_mult_pippenger_scratch_sizeof` returns a count of 4-byte limbs and `/ 8` truncates → undersized scratch → UB. A safer pattern is `(bytes + 7) / 8` or `Vec<MaybeUninit<u8>>` with byte sizing. Non-blocking since the cfg is locked to 64-bit, but worth a one-line guard or a comment locking in the assumption.

Same pattern at line 161 for `p2_msm` — applies there too.

crates/common/crypto/bls_blst.rs
```diff
@@ -0,0 +1,557 @@
...
+// 32-byte little-endian scalars (`SCALAR_BITS` wide). blst's array arguments are
+// null-terminated lists of pointers to contiguous arrays.
+
+#[inline]
+fn p1_msm(points: &[blst_p1_affine], scalars: &[u8]) -> blst_p1 {
+    let npoints = points.len();
+    let p: [*const blst_p1_affine; 2] = [points.as_ptr(), core::ptr::null()];
+    let s: [*const u8; 2] = [scalars.as_ptr(), core::ptr::null()];
+    let mut out = blst_p1::default();
+    // SAFETY: `npoints >= 1`, `scalars` is `npoints * 32` bytes, and scratch is
```
ElFantasma: nit: the SAFETY comment says `npoints >= 1`, but the only caller (`g1_msm`) routes `points.len() == 1` through `p1_scalar_mul` (line ~520) and `points.is_empty()` returns early. So in practice `p1_msm` is only ever called with `npoints >= 2`. The lower bound is true (`>= 2 ⊆ >= 1`) so the comment isn't *wrong*, but tightening it to `npoints >= 2` makes the invariant exact and explains why the single-point fast path exists. Same note at line 158 for `p2_msm`.

crates/common/crypto/bls_blst.rs
edg-l: Fixed the `/ 8` truncation: now `div_ceil(8)` so the buffer is never undersized. Per-call alloc tracked as a follow-up: #6864.

crates/common/crypto/bls_blst.rs
edg-l: Tightened both SAFETY comments to `npoints >= 2` and noted the caller handles the empty and single-point cases.

ilitteri (approved):

ElFantasma (approved):

juanbono (approved):

jrchatruc (approved):

PR #6795, opened 2026-06-05
fix(l1): return SYNCING instead of INVALID when parent state is missing in engine_newPayload

### LINKED ISSUE #6794: Ethrex returns INVALID instead of syncing
The most important part of the log is the following:

```log
FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, if consensus is synced and this is a consistent message it can be fixed by removing the DB and re-syncing the execution client. link_block=0x9599…f9b1 link_number=5355 head_number=5355
2026-06-05T07:04:24.382950Z  WARN Error executing block: DB error: state root missing
2026-06-05T07:04:24.385161Z  WARN FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, if consensus is synced and this is a consistent message it can be fixed by removing the DB and re-syncing the execution client. link_block=0x9599…f9b1 link_number=5355 head_number=5355
2026-06-05T07:04:24.573240Z  INFO 
2026-06-05T07:04:24.573311Z  INFO ╭──────────────────────────────────────────────────────────────────────────────╮
2026-06-05T07:04:24.573332Z  INFO │ SNAP SYNC COMPLETE                                                           │
2026-06-05T07:04:24.573354Z  INFO ├──────────────────────────────────────────────────────────────────────────────┤
2026-06-05T07:04:24.573360Z  INFO │ Total time: 00:00:03                                                        │
2026-06-05T07:04:24.573367Z  INFO ├──────────────────────────────────────────────────────────────────────────────┤
2026-06-05T07:04:24.573372Z  INFO │ Data summary:                                                                │
2026-06-05T07:04:24.573380Z  INFO │   Headers: 0              │  Accounts: 0                                    │
2026-06-05T07:04:24.573388Z  INFO │   Storage: 0              │  Bytecodes: 0                                   │
2026-06-05T07:04:24.573394Z  INFO │   Healed: 0 state paths + 0 storage accounts                                │
2026-06-05T07:04:24.573400Z  INFO ╰──────────────────────────────────────────────────────────────────────────────╯
2026-06-05T07:04:24.573434Z  INFO Snap Peers: 0 / Total Peers: 0
2026-06-05T07:04:24.574849Z  INFO Snap Peers: 0 / Total Peers: 0
2026-06-05T07:04:25.615317Z  INFO External IP detected via PONG voting, updating local ENR protocol="discv5" old_ip=172.18.0.5 new_ip=159.69.118.5
2026-06-05T07:04:54.387681Z  WARN No messages from the consensus layer. Is the consensus client running?
2026-06-05T07:05:24.388456Z  WARN No messages from the consensus layer. Is the consensus client running?
```

There are two separate problems:

1. The switch to full sync happens but the snap sync report is shown anyway. This is a UX issue but it could be a symptom of a race condition happening internally.
2. The new payload message returns invalid instead of syncing. From the consensus side it looks like this:

```log
[2026-06-05T07:04:24.383Z]  WARN eth1_api::execution_service:209: engine_newPayload returned INVALID status (beacon_block_root: 0x147035138f99152c7ad25e2a3327ed01da5740d6630d56b6c93eabeed0e65cd3, block_number: 5356, block_hash: 0x22502acb99f433bb898dea3d9ef2419db0f5d09ec89a8d044cfb15d44f56bb31, response: PayloadStatusV1 { status: Invalid, latest_valid_hash: Some(0x95993bc553bf82c0a4522f53278dc2337238700920071272e1514c7abc4df9b1), validation_error: Some("DB error: state root missing") }) peers=[14/200]
```

We should check the spec, but not having a parent state root is usually not enough reason to mark a block as invalid, it just means our node is not familiar with that state for whatever reason. It could trigger a sync, but at the minimum we shouldn't be returning invalid, even if we're finished syncing.

edg-l:
`engine_newPayload` was returning INVALID when a payload's parent state was absent from the DB. Without the parent state, execution fails with `EvmError::DB("state root missing")`, which was being mapped to INVALID and persisted via `set_latest_valid_ancestor`. This is wrong: per the Engine API spec (`execution-apis/src/engine/paris.md`), missing requisite state is a SYNCING condition, not INVALID. Returning INVALID for a valid block poisons the CL's view of the canonical chain.

This was observed live on glamsterdam-devnet-5: after an EL restart, ethrex regenerated state up to block 5354; Grandine sent `newPayload` for block 5356 (parent 5355) before the EL caught up, got INVALID back, and halted.

Fix in `try_execute_payload` (`crates/networking/rpc/engine/payload.rs`):

- Before executing, check `storage.has_state_root(parent_header.state_root)`. If the parent header exists but its state is absent, return `PayloadStatus::syncing()` and trigger `syncer.sync_to_head(...)`.
- Change the `ChainError::ParentStateNotFound` match arm from returning an internal RPC error to returning SYNCING and triggering sync. Remove the now-incorrect "this is not possible" comment.

The missing-parent-block case (`ParentNotFound`) is unchanged and still returns SYNCING as before. Normal steady-state operation is unaffected; the only added overhead is one `has_state_root` check per `newPayload` call.

Note: the repo has no integration-test harness for `engine_newPayload`, so no regression test is included. A follow-up test using a mocked `RpcApiContext`/syncer that asserts SYNCING when the parent state is absent would be a good addition.

Fixes https://github.com/lambdaclass/ethrex/issues/6794

### DIFF
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -1147,6 +1147,23 @@ async fn try_execute_payload(
         return Ok(PayloadStatus::valid_with_hash(block_hash));
     }
 
+    // A payload whose parent *state* we don't have yet must be answered with
+    // SYNCING, never INVALID: without the parent state we cannot validate it,
+    // so we must not declare it invalid. This happens after a restart, when
+    // state regeneration hasn't caught up to the CL head, or when the CL sends a
+    // newPayload for a block beyond our current state. Without this guard,
+    // execution fails with `EvmError::DB("state root missing")` and gets mapped
+    // to INVALID below, wrongly poisoning the CL's view of a valid block (and
+    // persisting it via `set_latest_valid_ancestor`). The parent block being
+    // entirely absent is handled as `ParentNotFound` by `add_block` below.
+    if let Some(parent_header) = storage.get_block_header_by_hash(block.header.parent_hash)?
+        && !storage.has_state_root(parent_header.state_root)?
+    {
+        debug!(%block_hash, %block_number, "Parent state missing, returning SYNCING and triggering sync");
+        syncer.sync_to_head(block_hash);
+        return Ok(PayloadStatus::syncing());
+    }
+
     // Execute and store the block
     debug!(%block_hash, %block_number, "Executing payload");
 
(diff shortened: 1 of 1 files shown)


MegaRedHand (approved):
LGTM

pablodeymo (approved):

iovoid (approved):

PR #6796, opened 2026-06-05
fix(l1): don't log SNAP SYNC COMPLETE when no snap sync ran

edg-l:
**Motivation**

Fixes a bad UX pattern where we print the snap sync complete when in reality we don't snap sync and just switch to full sync mode on a node restart

**Checklist**

Closes #issue_number


iovoid (approved):

pablodeymo (approved):

MegaRedHand (approved):
LGTM 👍

PR #6799, opened 2026-06-05
feat(l1): verify EIP-8025 sender hints properly

iovoid:
**Motivation**

Currently we verify EIP-8025 sender hints by recovering and then comparing to the hint. This is inefficient and against the spirit of the spec.

**Description**

Implements signature validation and uses that instead.


test/tests/crypto/verify_signature_tests.rs
```diff
@@ -0,0 +1,102 @@
+//! Tests for `Crypto::verify_signature`, the EIP-8025 sender-hint primitive.
+//!
+//! These exercise the native (`secp256k1`) backend, which is what the default
+//! build and the SP1 guest compile. The `k256` backend variant of
```
ElFantasma, line 4: **Test gap on the k256 path.** The header is right that this crate compiles `verify_signature` against the native (`secp256k1`) backend only — but that means the *novel* logic in this PR (the k256 impl's `R'.x as field bytes != &sig[..32]` aliasing fix, the manual lincomb, the parity check via `y_is_odd`) lands without any executable coverage. That's the subtle path; the native path is essentially `recover; compare`.

What would lock it in: add a small `#[cfg(not(feature = "secp256k1"))]` mirror suite under `crates/common/crypto/`, or — simpler — add a differential test that takes a fixed (sig, msg, pk) tuple and asserts both backends agree, gated on a `--no-default-features` CI variant. Without that, a regression in the k256 fix (e.g., someone "simplifying" to reduce R'.x mod n) silently slips through.

Non-blocking for this PR but worth a follow-up issue — the field-bytes-vs-mod-n distinction is precisely the kind of thing a future maintainer might "fix" thinking it's textbook ECDSA verify.

crates/common/types/transaction.rs
```diff
@@ -1289,9 +1292,51 @@ impl Transaction {
...
         crypto.recover_signer(&sig, &msg)
     }
 
+    #[cfg(feature = "eip-8025")]
+    pub fn compute_sender_with_hint(
+        &self,
+        public_key: &[u8; 65],
+        crypto: &dyn Crypto,
+    ) -> Result<Address, CryptoError> {
+        if let Transaction::PrivilegedL2Transaction(tx) = self {
```
ElFantasma, line 1377: **Behavior change worth surfacing.** Pre-PR (`program.rs`) the guest did `derived = address_from_pk(pk_bytes); recovered = tx.sender(...)?; if recovered != derived → Err`. For a `PrivilegedL2Transaction`, `tx.sender()` returned `tx.from`, so the old code *did* check that the witness-provided hint `pk_bytes` matched `tx.from` via address derivation.

New code skips `verify_signature` and the address-from-pk check entirely for PrivilegedL2 (`return Ok(tx.from)` right here), so a malicious or stale `pk_bytes` for a PrivilegedL2 tx in the witness now goes uncontested. The result is currently discarded by the caller (`tx.compute_sender_with_hint(...).map_err(...)?;` ignores the Ok value), so there's no immediate downstream exploit — but the witness no longer commits to the hint for these txs.

Two options:
1. **Intentional**: PrivilegedL2 has no signature, the hint is meaningless — fine, but worth a one-line comment here explaining *why* this branch ignores `public_key` instead of just looking like a missing case.
2. **Regression**: re-add the `address_from_pk(public_key) == tx.from` check before returning, matching the old guard.

Which did you intend?

crates/common/crypto/provider.rs
```diff
@@ -170,6 +171,101 @@ pub trait Crypto: Send + Sync + core::fmt::Debug {
...
+        let u2 = r * s_inv;
+        let big_r = ProjectivePoint::lincomb(&ProjectivePoint::GENERATOR, &u1, &q, &u2).to_affine();
+        if bool::from(big_r.is_identity()) {
+            return false;
+        }
+        // Compare R'.x to r as field bytes, without reducing mod n. A nonce point
+        // with x = r + n (possible only when r < p - n) must be rejected to match
+        // ecrecover, which reconstructs R from x = r exactly; reducing mod n here
+        // would accept that aliasing and diverge from canonical recovery.
+        if big_r.x().as_slice() != &sig[..32] {
```
ElFantasma, line 261: Excellent comment — the field-bytes-vs-mod-n distinction is exactly the kind of subtlety that gets "refactored" away by someone unfamiliar with ecrecover's quirks. Two small reinforcements would harden it further:

1. The comment says *"x = r + n (possible only when r < p - n)"* — adding `// p - n ≈ 2^128` (i.e., the aliasing window is roughly 2⁻¹²⁸ of valid r values, but not negligible enough to ignore) would help a reader gauge "is this a real edge case or theoretical?".
2. The two backends both reject high-s + out-of-range v with identical constants; `SECP256K1_N_HALF` and the `sig[64] > 1` check are duplicated verbatim. If you want to keep the impls separate (for `cfg(feature = "secp256k1")` gating), a free `fn pre_validate(sig: &[u8; 65]) -> bool` outside the trait would dedupe without merging the backends.

Non-blocking — both are polish.

crates/common/types/transaction.rs
iovoid: Indeed, PrivilegedL2 transactions carry no signature _and_ they are prohibited in L1 mode. Added a comment in 1d299b4ce

crates/common/types/transaction.rs
```diff
@@ -1289,9 +1292,45 @@ impl Transaction {
...
+        }
+        let sender = Address::from_slice(&keccak(&public_key[1..])[12..]);
+        // Cache only after verification; any prior `sender()` result must agree.
+        let cached = self.sender_cache().get_or_init(|| sender);
+        if *cached != sender {
+            return Err(CryptoError::Other(
+                "Inconsistent sender recovered".to_string(),
+            ));
+        }
+        Ok(sender)
```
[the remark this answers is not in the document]
iovoid: GLOBAL_SIGNER_CACHE doesn't matter since we are inside a single-purpose guest program, not a full node

crates/common/crypto/provider.rs
```diff
@@ -820,3 +914,179 @@ fn serialize_bls12_g2(point: &bls12_381::G2Affine) -> Result<[u8; 192], CryptoEr
     out[144..192].copy_from_slice(&raw[96..144]); // y_1
     Ok(out)
 }
+
+#[cfg(test)]
+mod tests {
```
iovoid, line 794: Note: we need to have this in order to test both variants. Since the tests crate depends on ethrex-crypto through several paths (some of which enable secp256k1), we would end up not testing the k256 path.

test/tests/crypto/verify_signature_tests.rs
iovoid: Addressed by adding a non-secp256k1 CI job

crates/common/crypto/provider.rs
iovoid: Done

ElFantasma (approved):

PR #6800, opened 2026-06-07
fix(l1): correct off-by-one and constant in engine getBlobs/getPayloadBodies size limits

### LINKED ISSUE #6674: rpc/engine: fix off-by-one and wrong constant in getBlobs / getPayloadBodies request-size validation
## Summary

Two engine-API request-size validation bugs in Ethrex.

1. Off-by-one comparison. The code uses `>= MAX_SIZE` where the spec requires `> MAX_SIZE`. The spec text "Client software MUST support request sizes of at least N" requires N itself to be served; only N+1 and above should be rejected.

2. Wrong constant for `engine_getPayloadBodies`. `GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE = 128`; the Shanghai spec value is 1024.

A spec-compliant CL requesting payload bodies during sync at typical batch sizes 256, 512, or 1024 receives `TooLargeRequest` from Ethrex. The other 5 mainnet ELs (Geth, Reth, Besu, Erigon, Nethermind) use strict `>` against the spec constants. Source-verified at stable release tags on 2026-05-18.

## Affected code

`crates/networking/rpc/engine/blobs.rs`:

* Line 16: `const GET_BLOBS_V1_REQUEST_MAX_SIZE: usize = 128;`. Value matches spec; comparison below does not.
* Line 67 (`BlobsV1Request::handle`): `if self.blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE`
* Line 143 (shared `get_blobs_and_proof` for V2/V3): `if blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE`

`crates/networking/rpc/engine/payload.rs`:

* Line 25: `const GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE: u64 = 128;`. Spec value is 1024.
* Lines 611, 647, 688, 737 (the four `GetPayloadBodies*` handlers): `>= GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE`

## Spec text

Cancun engine-API spec, `engine_getBlobsV1` ([source](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md)):

> Client software **MUST** support request sizes of at least 128 blob versioned hashes. The client **MUST** return -38004: Too large request error if the number of requested blobs is too large.

Shanghai engine-API spec, `engine_getPayloadBodiesByHashV1` ([source](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md)):

> Client software **MUST** support request sizes of at least 1024 block hashes.

"MUST support at least N" requires N to succeed. The boundary should reject `N + 1` and above.

## Cross-EL evidence

| EL | getBlobs check | getPayloadBodies check | OK? |
|---|---|---|---|
| Geth v1.17.3 | `len(hashes) > 128` (`eth/catalyst/api.go:562`, 630) | strict `>` on 1024 | ✓ |
| Reth v2.2.0 | `len() > MAX_BLOB_LIMIT (128)` (`engine_api.rs:873`) | `len() > MAX_PAYLOAD_BODIES_LIMIT (1024)` | ✓ |
| Besu 26.4.0 | `versionedHashes.length > 128` (`EngineGetBlobsV1.java:104`) | `MAX_REQUEST_BLOCKS = 1024` (`AbstractEngineGetPayloadBodies.java:30`) | ✓ |
| Erigon v3.4.1 | `len(blobHashes) > 128` (`engine_server.go:1062`) | `len(request) > 1024` (`engine_server.go:786`, 821) | ✓ |
| Nethermind v1.37.2 | `Length > MaxRequest (128)` (`GetBlobsHandlerV2.cs:21`) | `Count > MaxCount (1024)` (`GetPayloadBodiesByHashV1Handler.cs:23`) | ✓ |
| **Ethrex v11.0.0** | **`len() >= 128`** | **`len() >= 128`, wrong constant** | **✗** |

## Practical impact

Blob requests at exactly 128 hashes are uncommon under per-block caps of 6, 9, 12 across recent forks.

CL sync requests for payload bodies typically batch at 256, 512, or 1024. Ethrex rejects all batches above 127. CLs retry with smaller batches, lowering effective throughput when peering with Ethrex.

## Proposed change

```diff
--- a/crates/networking/rpc/engine/blobs.rs
+++ b/crates/networking/rpc/engine/blobs.rs
@@ -64,7 +64,7 @@ impl RpcHandler for BlobsV1Request {
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
         debug!("Received new engine request: Requested Blobs");
 
-        if self.blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE {
+        if self.blob_versioned_hashes.len() > GET_BLOBS_V1_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
 
@@ -140,7 +140,7 @@ async fn get_blobs_and_proof(
     context: RpcApiContext,
     version: u64,
 ) -> Result<Vec<Option<BlobAndProofV2>>, RpcErr> {
-    if blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE {
+    if blob_versioned_hashes.len() > GET_BLOBS_V1_REQUEST_MAX_SIZE {
         return Err(RpcErr::TooLargeRequest);
     }
```

```diff
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -22,7 +22,8 @@ use crate::{
 };
 
 // -> https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-3
-const GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE: u64 = 128;
+// Spec: clients MUST support request sizes of at least 1024 block hashes.
+const GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE: u64 = 1024;
 
 // ... in each of the four handle() functions:
-        if self.hashes.len() as u64 >= GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
+        if self.hashes.len() as u64 > GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
-        if self.count >= GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
+        if self.count > GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
```

## Testing

Existing tests construct `MAX_SIZE + 1` and assert rejection. Add boundary tests at exactly `MAX_SIZE`:

```rust
#[tokio::test]
async fn test_blobs_v1_accepts_exactly_max_size() {
    let req = BlobsV1Request {
        blob_versioned_hashes: vec![H256::zero(); GET_BLOBS_V1_REQUEST_MAX_SIZE],
    };
    let ctx = build_test_context();
    let result = req.handle(ctx).await;
    assert!(!matches!(result, Err(RpcErr::TooLargeRequest)));
}

#[tokio::test]
async fn test_payload_bodies_by_hash_v1_accepts_exactly_max_size() {
    let req = GetPayloadBodiesByHashV1Request {
        hashes: vec![BlockHash::zero(); GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE as usize],
    };
    let ctx = build_test_context();
    let result = req.handle(ctx).await;
    assert!(!matches!(result, Err(RpcErr::TooLargeRequest)));
}
```

Update the existing `too_large` tests to use `MAX_SIZE + 1` explicitly so both directions are covered.

## Prior art

* [EF execution-apis Shanghai spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md): "MUST support at least 1024 block hashes"
* [EF execution-apis Cancun spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md): "MUST support at least 128 blob versioned hashes"
* Cross-EL audit on 2026-05-18 confirmed Geth, Reth, Besu, Erigon, Nethermind use the strict `>` form against the spec constants.

## Backwards compatibility

The change loosens validation. Previously rejected requests at the spec-required minimum now succeed. No previously accepted request becomes rejected.

0xMars42:
> Credit: the diagnosis, the cross-client audit, the spec citations and the exact proposed diff are
> all @stefa2k's work in #6674. This PR implements that fix and adds boundary tests.

## Motivation

Two engine-API request-size validation bugs (#6674):

1. **Off-by-one.** The size checks use `>= MAX` where the spec requires `> MAX`. The spec text
   "Client software MUST support request sizes of at least N" requires N itself to be served; only
   N + 1 and above should be rejected.
2. **Wrong constant.** `GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE` was `128`. The engine spec only
   mandates supporting request sizes of at least 32 here; the cap is raised to `1024` to match
   [`MAX_REQUEST_BLOCKS`](https://github.com/ethereum/consensus-specs/blob/a84880a47a88700d8dfa451c2a7cd4b3f309bd0d/specs/phase0/p2p-interface.md#configuration),
   the largest batch a conforming CL requests.

A spec-compliant CL requesting payload bodies during sync at typical batch sizes (256 / 512 / 1024)
received `TooLargeRequest`, forcing smaller retries and lowering throughput when peering with
ethrex. The other mainnet ELs (geth, reth, besu, erigon, nethermind) all use a strict `>` against
the spec constants.

## Changes

- `crates/networking/rpc/engine/blobs.rs`: `>=` -> `>` on the two `getBlobs` size checks.
- `crates/networking/rpc/engine/payload.rs`: `GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE` `128` -> `1024`,
  and `>=` -> `>` on the four `getPayloadBodies` handlers.

This only loosens validation: previously-rejected requests at the spec-required minimum now succeed;
no previously-accepted request becomes rejected.

## Tests

Added boundary tests asserting that requests at exactly `MAX_SIZE` are not rejected (they were
before this change), alongside the existing `MAX_SIZE + 1` rejection checks:

- `blobs.rs`: `blobs_v3_accepts_exactly_max_size`, `blobs_v1_accepts_exactly_max_size_before_osaka`.
- `payload.rs`: accept-at-`MAX` and reject-above-`MAX` boundary tests for all four handlers
  (`GetPayloadBodiesByHashV1`, `ByRangeV1`, `ByHashV2`, `ByRangeV2`).

Closes #6674.

### DIFF
--- a/crates/networking/rpc/engine/blobs.rs
+++ b/crates/networking/rpc/engine/blobs.rs
@@ -79,7 +79,7 @@ impl RpcHandler for BlobsV1Request {
             ));
         }
 
-        if self.blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE {
+        if self.blob_versioned_hashes.len() > GET_BLOBS_V1_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
 
@@ -155,7 +155,7 @@ async fn get_blobs_and_proof(
     context: RpcApiContext,
     version: u64,
 ) -> Result<Vec<Option<BlobAndProofV2>>, RpcErr> {
-    if blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE {
+    if blob_versioned_hashes.len() > GET_BLOBS_V1_REQUEST_MAX_SIZE {
         return Err(RpcErr::TooLargeRequest);
     }
 
@@ -373,5 +373,26 @@ mod tests {
 
         let err = request.handle(context).await.unwrap_err();
         assert!(matches!(err, RpcErr::TooLargeRequest));
+    }
+
+    #[tokio::test]
+    async fn blobs_v3_accepts_exactly_max_size() {
+        // Spec: clients MUST support at least MAX hashes, so exactly MAX must not be rejected.
+        let context = context_with_chain_config(true).await;
+        let request = BlobsV3Request {
+            blob_versioned_hashes: vec![H256::zero(); GET_BLOBS_V1_REQUEST_MAX_SIZE],
+        };
+        let result = request.handle(context).await;
+        assert!(!matches!(result, Err(RpcErr::TooLargeRequest)));
+    }
+
+    #[tokio::test]
+    async fn blobs_v1_accepts_exactly_max_size_before_osaka() {
+        let context = context_with_chain_config(false).await;
+        let request = BlobsV1Request {
+            blob_versioned_hashes: vec![H256::zero(); GET_BLOBS_V1_REQUEST_MAX_SIZE],
+        };
+        let result = request.handle(context).await;
+        assert!(!matches!(result, Err(RpcErr::TooLargeRequest)));
     }
 }
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -21,8 +21,9 @@ use crate::types::payload::{
 use crate::utils::RpcErr;
 use crate::utils::{RpcRequest, parse_json_hex};
 
-// Engine API (Shanghai): clients MUST support request sizes of at least 1024 block hashes.
-// -> https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-3
+// The Engine API (Shanghai) only mandates supporting request sizes of at least 32 blocks.
+// Cap at MAX_REQUEST_BLOCKS = 1024, the largest request a conforming consensus client makes.
+// -> https://github.com/ethereum/consensus-specs/blob/a84880a47a88700d8dfa451c2a7cd4b3f309bd0d/specs/phase0/p2p-interface.md#configuration
 const GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE: u64 = 1024;
 
 // NewPayload V1-V2-V3 implementations
--- a/crates/networking/rpc/engine/payload.rs
+++ b/crates/networking/rpc/engine/payload.rs
@@ -21,10 +21,9 @@ use crate::types::payload::{
 use crate::utils::RpcErr;
 use crate::utils::{RpcRequest, parse_json_hex};
 
-// Must support rquest sizes of at least 32 blocks
-// Chosen an arbitrary x4 value
+// Engine API (Shanghai): clients MUST support request sizes of at least 1024 block hashes.
 // -> https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-3
-const GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE: u64 = 128;
+const GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE: u64 = 1024;
 
 // NewPayload V1-V2-V3 implementations
 pub struct NewPayloadV1Request {
@@ -716,7 +715,7 @@ impl RpcHandler for GetPayloadBodiesByHashV1Request {
     }
 
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
-        if self.hashes.len() as u64 >= GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
+        if self.hashes.len() as u64 > GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
         let mut bodies = Vec::new();
@@ -752,7 +751,7 @@ impl RpcHandler for GetPayloadBodiesByRangeV1Request {
     }
 
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
-        if self.count >= GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
+        if self.count > GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
         let latest_block_number = context.storage.get_latest_block_number().await?;
@@ -839,7 +838,7 @@ impl RpcHandler for GetPayloadBodiesByHashV2Request {
     }
 
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
-        if self.hashes.len() as u64 >= GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
+        if self.hashes.len() as u64 > GET_PAYLOAD_BODIES_REQUEST_MAX_SIZE {
             return Err(RpcErr::TooLargeRequest);
         }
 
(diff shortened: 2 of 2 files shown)


0xMars42:
Extended the boundary tests to cover all four getPayloadBodies handlers (range and V2 variants), per the review.

crates/networking/rpc/engine/blobs.rs
```diff
@@ -155,7 +155,7 @@ async fn get_blobs_and_proof(
     context: RpcApiContext,
     version: u64,
 ) -> Result<Vec<Option<BlobAndProofV2>>, RpcErr> {
-    if blob_versioned_hashes.len() >= GET_BLOBS_V1_REQUEST_MAX_SIZE {
+    if blob_versioned_hashes.len() > GET_BLOBS_V1_REQUEST_MAX_SIZE {
```
ElFantasma, line 157: Out of scope for this PR but worth noting: `GET_BLOBS_V1_REQUEST_MAX_SIZE` is also used by V2/V3 (see `get_blobs_and_proof`) — same spec cap, but the constant name now reads as V1-only. A rename to `GET_BLOBS_REQUEST_MAX_SIZE` (without the version) would match its actual scope. Happy to file a follow-up if you'd rather keep this PR focused on the off-by-one.

ElFantasma (approved):
One small observation inlined on the constant name (now serves V1/V2/V3, but reads V1-only). The PR's branch is at the merge-base before https://github.com/lambdaclass/ethrex/pull/6808's V1-Osaka gate landed, so a quick rebase will be needed; the changes don't semantically conflict (both target the same V1 handler) but the textual hunks overlap.

LGTM contingent.

crates/networking/rpc/engine/blobs.rs
0xMars42: Agreed, the constant caps V2/V3 too, so the V1 name no longer matches its scope. I will keep this PR focused on the off-by-one and open the rename as a follow-up once this lands.

0xMars42:
Rebased on current main. The failing checks on the previous run came from the pre-#6815 base (stack overflow in rpc::tests), not from this diff, so they should come back green now. #6808 has not merged yet as of this push; if it lands first I will rebase over it again.

ilitteri (approved):

crates/networking/rpc/engine/payload.rs
```diff
@@ -21,10 +21,9 @@ use crate::types::payload::{
 use crate::utils::RpcErr;
 use crate::utils::{RpcRequest, parse_json_hex};
 
-// Must support rquest sizes of at least 32 blocks
-// Chosen an arbitrary x4 value
+// Engine API (Shanghai): clients MUST support request sizes of at least 1024 block hashes.
 // -> https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#specification-3
```
iovoid: This spec URL is wrong, we probably want https://github.com/ethereum/consensus-specs/blob/a84880a47a88700d8dfa451c2a7cd4b3f309bd0d/specs/phase0/p2p-interface.md#configuration

0xMars42: Good catch, thanks. The engine spec only mandates supporting at least 32 for these endpoints; the 1024 cap actually comes from MAX_REQUEST_BLOCKS in the consensus p2p spec. Fixed the comment to point at the configuration table you linked, and reworded the PR description to match.

iovoid (approved):

ilitteri:
Hey! PR is fully approved, but it fails the CI. Please, fix this so it can be merged.

PR #2918, opened 2025-05-23
refactor(l1): move some tool crates to tooling folder

DiegoCivi:
**Motivation**

Tools used in the project were not in the `tooling` folder but in the `cmd`.

**Description**
- `cmd/hive_report`, `cmd/loc` and `cmd/load_test` were moved into the `tooling` folder
- The genesis tool was moved from `tools` to `tooling/genesis`

Closes #2888

PR #6801, opened 2026-06-07
refactor(l1): move levm tools to tooling/ folder

### LINKED ISSUE #5610: Move levm tools to the tooling folder
Currently we have several tools (the runner, the bench tool and the flamegraph generator) under `crates/vm/levm`. We might want to move them to `tooling` for clarity.

pjdurden:
Closes #5610. Moves the levm runner and revm_comparison benchmark tool from `crates/vm/levm` to `tooling/`, following the layout precedent from #2918 and #5887. `runner` stays a root-workspace member (like `tooling/repl`); `revm_comparison` keeps its own standalone workspace. Updated workspace members/exclude, manifest path-deps, root + levm Makefiles, and release docs. The bench/flamegraph make-targets stay in the levm Makefile (paths updated) since they're interleaved with EF-test profiling targets and CI runs them from that dir — `CARGO_TARGET_DIR` still resolves to `crates/vm/levm/target`, so the benchmark binary path is unchanged. Verified `cargo check` on both relocated crates.


pjdurden:
One decision worth flagging for review: I left the bench/flamegraph make-targets (`revm-comparison`, `revm-comparison-ci`, `build-revm-comparison`, `flamegraph-levm-*`, `render-benches`) in `crates/vm/levm/Makefile` and only updated their paths, rather than relocating them.

Reasoning:
- They're interleaved with the levm EF-test profiling targets in that Makefile, and CI (`pr-main_levm.yaml`, `pr_perf_levm.yaml`) invokes them after `cd crates/vm/levm`.
- `CARGO_TARGET_DIR` is kept resolving to `crates/vm/levm/target`, so the benchmark binary path the perf workflow uploads (`crates/vm/levm/target/release/benchmark`) is unchanged.

Happy to move these targets into a dedicated `tooling/revm_comparison/Makefile` if you'd rather have them live alongside the moved crate , so just let me know which you prefer.

crates/vm/levm/Makefile
```diff
@@ -126,13 +126,13 @@ flamegraph-levm-bench-generic:
 	cargo flamegraph --root --package revm_comparison --bin benchmark --output flamegraph_levm_benchmark/flamegraph_levm_$(name).svg -- levm $(name) $(repetitions) $(iterations)
 
 build-revm-comparison:
-	cd bench/revm_comparison && \
-		CARGO_TARGET_DIR=../../target \
+	cd ../../../tooling/revm_comparison && \
+		CARGO_TARGET_DIR=../../crates/vm/levm/target \
 		cargo build --release --bin benchmark
```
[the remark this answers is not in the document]
pjdurden: Good point. Switched both build targets to use an absolute `LEVM_TARGET_DIR := $(CURDIR)/target` instead of the deep relative path, so the value no longer depends on the `cd` into `tooling/revm_comparison`. It resolves to `crates/vm/levm/target` (where the perf workflow uploads the benchmark binary from) regardless of which directory the recipe builds in.

crates/vm/levm/Makefile
```diff
@@ -125,14 +125,18 @@ flamegraph-levm-bench-generic:
 	mkdir -p flamegraph_levm_benchmark
 	cargo flamegraph --root --package revm_comparison --bin benchmark --output flamegraph_levm_benchmark/flamegraph_levm_$(name).svg -- levm $(name) $(repetitions) $(iterations)
 
+# Absolute target dir so the benchmark binary lands in this crate's target/
+# regardless of the directory the build command cd's into.
+LEVM_TARGET_DIR := $(CURDIR)/target
```
ElFantasma, line 131: nit: the comment claims this is set *"regardless of the directory the build command cd's into"* — true for the recipe's internal `cd ../../../tooling/revm_comparison` (since `LEVM_TARGET_DIR` is expanded before the cd), but the outer invocation **does** matter. `$(CURDIR)` is the cwd at the moment make started, so:

- `cd crates/vm/levm && make build-revm-comparison` → CURDIR = `crates/vm/levm` → target = `crates/vm/levm/target` ✓
- `make -C crates/vm/levm build-revm-comparison` → CURDIR = `crates/vm/levm` ✓
- `make -f crates/vm/levm/Makefile build-revm-comparison` (from repo root, no `-C`) → CURDIR = repo root → target = repo-root/target ✗

The old `CARGO_TARGET_DIR=../../target` was always `crates/vm/levm/target` because it resolved relative to the recipe's post-cd cwd (`bench/revm_comparison/../../target`).

The last case is uncommon and CI almost certainly uses `-C`, so this is a precision nit rather than a real bug. Either tighten the comment to *"… provided make was invoked from or with `-C crates/vm/levm`"*, or rewrite the recipe to use a path computed relative to the Makefile's directory (`$(dir $(lastword $(MAKEFILE_LIST)))target`) for true invocation-independence.

ElFantasma (approved):
Left one nit

crates/vm/levm/Makefile
pjdurden: Thanks for response @ElFantasma ,tightened the comment to scope the claim to the `-C crates/vm/levm` invocation rather than reworking the recipe.

pjdurden:
Rebased on main, conflicts resolved (workspace Cargo.toml + the moved tool paths). Ready for a re-merge whenever, thanks @ElFantasma

pjdurden:
Devp2p failures here are pre-existing on main, not from this refator: same TestBlobTx* eth-suite tests fail on the latest main run (https://github.com/lambdaclass/ethrex/actions/runs/28058390120). Unrelated to moving the levm tools.

pjdurden:
merged latest main, conflicts resolved. `tooling/` became its own workspace upstream in the meantime, but `runner` stays a root-workspace member since it depends on root workspace deps (`rustc-hash`) and `[workspace.lints]`, neither of which `tooling/Cargo.toml` defines. `cargo metadata --locked` passes on root, `tooling/`, and `tooling/revm_comparison`.

pjdurden:
One thing this move breaks: `tooling/` is its own workspace now, so `cd tooling/runner && cargo run` (the README's command) fails to inherit `rustc-hash` from the root workspace, while `cd crates/vm/levm/runner` works on main. Adding `runner` to the tooling workspace's `exclude` does not help, cargo then resolves to the root workspace and errors that `tooling/monitor` is claimed by both workspaces. `tooling/repl` is already broken the same way on main and `cargo run -p runner` from the repo root still works, so I left it as is. If you'd rather I make `runner` a real member of the tooling workspace in this PR, that needs `rustc-hash` and a lints table added there.

Cargo.toml
```diff
@@ -22,12 +22,12 @@ members = [
   "crates/storage",
   "crates/vm",
   "crates/vm/levm",
-  "crates/vm/levm/runner",
+  "tooling/runner",
```
ElFantasma, line 25: Thanks for flagging this rather than quietly leaving it — your diagnosis checks out. Confirmed on `f3a63e416`: this line and L27 claim `tooling/runner` and `tooling/repl` as **root** members, `tooling/Cargo.toml`'s members list contains neither, and its `exclude` only covers `ef_tests/state`. So both directories physically sit inside the `tooling/` workspace while belonging to the root one, and `cd`-ing into either resolves to the workspace that disowns it.

**Suggestion: fix the README here, leave the layout alone.** `tooling/runner/README.md:5` currently documents a bare `cargo run`, which only works from inside the directory — that's the thing this move actually breaks. Replacing it with the root-relative form fixes the user-visible problem in one line and is the idiomatic invocation in a workspace anyway:

```
Example Run: `cargo run -p runner -- --input input_example.json --code mnemonics_example.txt`
```

**Please don't do the membership change in this PR.** Making `runner` a real `tooling/` member means copying `rustc-hash` *and* a `[workspace.lints]` table into `tooling/Cargo.toml` — that's a duplicated dependency version and a duplicated lints table across two workspaces, both of which drift silently, and it would still leave `repl` inconsistent. Not a good trade for a file-move PR, and not your mess to clean up.

The structural oddity is worth a follow-up issue covering `runner` and `repl` together. Happy to open it if you'd rather not.

pjdurden:
Answering my own open question from above rather than leaving it hanging: `runner` now sets `workspace = "../.."` in its manifest, so it stays a member of the root workspace and keeps inheriting `rustc-hash` and `workspace.lints` from the root, even though it now lives under `tooling/`, which became its own workspace in the meantime.

Before, on this branch:

```
$ cd tooling/runner && cargo metadata --no-deps
error: failed to parse manifest at `tooling/runner/Cargo.toml`
Caused by:
  error inheriting `rustc-hash` from workspace root manifest's `workspace.dependencies.rustc-hash`
```

After:

```
$ cd tooling/runner && cargo run -- --input input_example.json --code mnemonics_example.txt
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 36.10s
     Running `ethrex/target/debug/runner --input input_example.json --code mnemonics_example.txt`
Result:
 ExecutionReport { result: Success, gas_used: 21009, ... }
```

That is the command in `tooling/runner/README.md`, so the documented workflow works from the new location.

It is only the one manifest key, no changes to the members or exclude lists of either workspace, so it avoids the "monitor is claimed by both workspaces" error I ran into with the exclude approach. `cargo metadata --locked` still passes at the repo root, in `tooling/`, and in `tooling/revm_comparison`, and `Cargo.lock` is unchanged.

Separately: `tooling/repl` is broken the same way on main today (`current package believes it's in a workspace when it's not`) and needs the same single line. I left it out to keep this PR to the move. Happy to fold it in here or open it as a small follow-up, whichever you prefer.

pjdurden:
Sorry, I pushed the manifest fix before I saw your inline comment. Reverted it in `8a393573a`, so the layout is untouched and your point about duplicating `rustc-hash` and the lints table across two workspaces stands.

Took the README route instead, with one adjustment. The exact line panics when run from the root, because `cargo run` does not chdir into the package, so the bare filenames resolve against the invocation directory:

```
$ cargo run -p runner -- --input input_example.json --code mnemonics_example.txt
thread 'main' panicked at tooling/runner/src/main.rs:91:33:
Input file 'input_example.json' not found
```

So the README now says:

```
Example Run, from the repository root: `cargo run -p runner -- --input tooling/runner/input_example.json --code tooling/runner/mnemonics_example.txt`
```

which runs clean:

```
Result:
 ExecutionReport { result: Success, gas_used: 21009, gas_spent: 21009, ... }
Final Stack (bottom to top): ["0x1", "0x2", "0x3", "0xf", "0x11f"]
```

Plus one line noting the paths are relative to the invocation directory, since that is the part that bites.

Please do open the follow-up issue for `runner` and `repl`. For whatever it is worth there, `workspace = "../.."` in the package manifest keeps both in the root workspace without copying `rustc-hash` or a lints table into `tooling/Cargo.toml`. I verified it works (`cd tooling/runner && cargo run` succeeds with it), but it is a layout decision for the follow-up, not for this PR.

This branch is 5 commits behind main, no conflicts. Happy to merge main in if you want it current before merging.

pjdurden:
merged current main in (`73381cc9`), pushed as `c8f18c46`. it had gone conflicted; the merge itself was clean, no manual resolution, and the PR's own diff is unchanged at 41 files / +28/-19, still almost entirely renames.

@ElFantasma your README route from the last round is in and unchanged: `tooling/runner/README.md:5` documents the root-relative `cargo run -p runner -- --input tooling/runner/input_example.json ...`, plus a line making explicit that the paths are relative to where you invoke from rather than to the package. no membership change, so `tooling/Cargo.toml` still has no duplicated `rustc-hash` or `[workspace.lints]`. still happy to open the follow-up issue covering `runner` and `repl` together if you have not already.

on the three red checks from the last run, none of them look like this PR:

- **Link Check**: two 404s, both on the same dead external URL `https://docs.alignedlayer.com/guides/7_contract_addresses`, in `docs/l2/deployment/aligned.md:56` and `docs/l2/fundamentals/ethrex_l2_aligned_integration.md:544`. this PR touches neither file, and it is link rot on alignedlayer's side.
- **Hive - Devp2p tests**: 30 failing cases, 24 in `eth` and 6 in `snap`, including `BlobTxAvailabilityFailure`. this PR moves files between directories, it does not touch networking.
- **Integration Test**: fails only as the aggregator reporting the Hive job above.

the merge should re-run all three against current main, so if any of them were already fixed upstream they will clear on their own.

PR #6803, opened 2026-06-08
fix(l1): fix full sync wedge on body fetch failures and canonical-but-stateless state gaps

edg-l:
Full sync on degraded networks (e.g. glamsterdam-devnet-5) could wedge in a loop alternating between `No bodies were found` and `state root missing`, freezing the head. This makes the walk-back resilient.

**Body-fetch resilience** (`sync/full.rs`): a single peer failing to return bodies aborted the whole cycle and re-walked all headers from scratch on the next retry. `request_bodies_with_retry` adds backoff retries (mirroring the header-fetch loop); on exhaustion it ends the cycle gracefully and waits for the next forkchoice head instead of propagating a fatal error.

**Resume-point anchoring** (`sync/full.rs`, `is_resume_point` / `first_resume_point_in_batch`): execution may only resume from a block that is canonical AND has its post-state on disk. `apply_fork_choice` can canonicalize a head before its state is computed, so the canonical chain can extend past the executed-state head; anchoring on such a canonical-but-stateless block fails forever with `state root missing`. The walk scans each header batch (newest->oldest, gated to batches reaching down to `local_head`) for the first canonical+stateful block — the true state head, wherever it falls in the batch — resumes there, and keeps the canonical-but-stateless blocks above it for re-execution to backfill the gap. If no resume point exists down to genesis (a fork/deep-reorg head reconciling only to a block whose state was pruned), the cycle pauses gracefully and waits for a reconcilable head rather than re-executing from a missing base.

**Header start-hash validation** (`peer_handler.rs`): `request_block_headers_from_hash` validated only internal parent-hash linkage, so a peer could return an internally-consistent run of headers from its own (fork/minority) chain that does not begin at the requested `start`. Reject responses where `block_headers[0].hash() != start`, penalize the peer, and let the retry re-roll to a peer serving the requested chain.

**Diagnostics**: `state root missing` names the block number and state root; a `WARN` fires when full sync resumes below the canonical head; `eth_peer_count` is logged at cycle start and on fetch failures so `eth_peers=0` (no peers) is distinguishable from `eth_peers>0` (peers not serving).

**Tests** (`test/tests/p2p/full_sync_tests.rs`): `is_resume_point` for canonical+stateful / canonical+stateless / non-canonical; the walk anchoring on the state head; and `first_resume_point_in_batch` merging at a resume point inside the batch, skipping a batch entirely above `local_head`, and returning none for a fully-pruned batch.

Validated on glamsterdam-devnet-5 (lodestar-ethrex-1).

### DIFF
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -71,8 +71,10 @@ impl HeaderFetchOutcome {
     pub fn failure_reason(&self) -> &'static str {
         match self {
             HeaderFetchOutcome::Headers(_) => "headers received",
-            HeaderFetchOutcome::NoPeerAvailable => "no eth peer available to ask",
-            HeaderFetchOutcome::PeerFailed => "peer(s) asked but did not serve headers",
+            HeaderFetchOutcome::NoPeerAvailable => {
+                "no eth-capable peer with a live connection to query (peers may be connecting or recently dropped)"
+            }
+            HeaderFetchOutcome::PeerFailed => "peer(s) queried but did not serve headers",
         }
     }
 }
@@ -148,11 +150,12 @@ impl PeerHandler {
             .await?)
     }
 
-    /// Number of connected peers that advertise the eth capabilities used for sync.
-    /// Sync diagnostics use this to tell "no peers to ask" (discovery/connectivity
-    /// problem) apart from "peers present but not serving" (the chain data is gone).
-    /// Returns 0 on any peer-table error since this is only used for logging.
-    pub async fn eth_peer_count(&self) -> usize {
+    /// Number of peers known to the table that advertise the eth capabilities used for sync.
+    /// NOTE: this counts eth-capable peers regardless of whether they currently have a live
+    /// connection, so it can be greater than the number actually queryable via
+    /// `get_random_peer` (which requires a live connection). Used only for diagnostics; logged
+    /// as `eth_capable_peers`. Returns 0 on any peer-table error.
+    pub async fn eth_capable_peer_count(&self) -> usize {
         self.peer_table
             .peer_count_by_capabilities(SUPPORTED_ETH_CAPABILITIES.to_vec())
             .await
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -462,11 +462,16 @@ impl PeerHandler {
                 })) = response
                 {
                     if block_headers.is_empty() {
-                        // Empty response is valid per eth spec (peer may not have these blocks)
+                        // Empty response is valid per eth spec (peer may not have these blocks),
+                        // so apply only a soft score penalty (`record_failure`) rather than
+                        // ejecting the peer (`set_disposable`): a spec-conformant peer that simply
+                        // lacks a fork's blocks shouldn't be permanently dropped from rotation.
+                        // Genuine misbehavior below (unchained / wrong-chain-start) uses the same
+                        // soft tier, so the distinction stays consistent.
                         debug!(
                             "[SYNCING] Received empty headers from peer {peer_id}, trying another"
                         );
-                        let _ = self.peer_table.set_disposable(peer_id);
+                        self.peer_table.record_failure(peer_id)?;
                         return Ok(HeaderFetchOutcome::PeerFailed);
                     }
                     if are_block_headers_chained(&block_headers, &order) {
--- a/crates/networking/p2p/peer_handler.rs
+++ b/crates/networking/p2p/peer_handler.rs
@@ -470,6 +470,23 @@ impl PeerHandler {
                         return Ok(HeaderFetchOutcome::PeerFailed);
                     }
                     if are_block_headers_chained(&block_headers, &order) {
+                        // Pin the response to the requested `start` hash. `are_block_headers_chained`
+                        // only verifies internal parent-hash linkage, not that the sequence actually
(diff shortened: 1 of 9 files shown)


crates/networking/p2p/sync/full.rs
```diff
@@ -53,6 +56,83 @@ fn humanize_secs(secs: u64) -> String {
...
+    block_headers: &[BlockHeader],
+    local_head: u64,
+) -> Result<Option<usize>, SyncError> {
+    let Some(oldest) = block_headers.last() else {
+        return Ok(None);
+    };
+    if oldest.number > local_head {
+        return Ok(None);
+    }
+    for (index, header) in block_headers.iter().enumerate() {
```
ElFantasma, line 136: Worst-case complexity worth documenting in the doc comment: `first_resume_point_in_batch` is O(N) `has_state_root` lookups per batch, where N is the batch length (256-1024 in production). The `local_head` gate skips batches that are entirely above local head (the common deep-sync case), and in the typical resume case the scan terminates near the top because the state head is at-or-just-below the not-yet-executed prefix — so the per-batch cost is usually 2-5 lookups, not N.

The pathological case is a state head far below a batch's newest header (e.g. a long canonical-but-stateless gap from a long-running FCU-ahead-of-execution window): there the scan does walk the full batch before finding the hit. Not actionable for this PR — devnets exhibit at most a few-block gap, and a `state_root` lookup is a single MPT root probe — but worth a one-line note in the doc comment so the cost dimension is visible if batch sizes ever climb.

crates/networking/p2p/peer_handler.rs
```diff
@@ -429,25 +467,42 @@ impl PeerHandler {
...
+                        // penalize the peer so the caller re-rolls `get_random_peer` and keeps trying until
+                        // it lands a peer actually serving `start`'s chain (whose ancestry is hash-linked
+                        // and therefore bridges down to our canonical head). General hardening, not
+                        // devnet-specific.
+                        if block_headers[0].hash() != start {
+                            warn!(
+                                "[SYNCING] Peer {peer_id} returned headers not starting at the requested hash {start:#x}, penalizing peer"
+                            );
+                            self.peer_table.record_failure(peer_id)?;
+                            return Ok(HeaderFetchOutcome::PeerFailed);
```
ElFantasma, line 495: Minor consistency: this disposition uses `record_failure` while line 470 (empty response from peer) uses `set_disposable`. Both are remedial signals to the peer table, but they have different semantics — `set_disposable` ejects the peer entirely from the rotation, while `record_failure` (I think?) is a soft tally that may eventually disposable them.

Wrong-chain-start is arguably a *stronger* misbehavior signal than empty response (peer is on a different chain, not just transiently unresponsive), so if anything this should be the harsher of the two. Either way, worth a quick audit pass to confirm the failure-tier mapping is what you want — pre-PR the function returned `None` for both cases and the distinction didn't matter.

ElFantasma (approved):
Two non-blocking observations inlined on the resume-scan complexity and the peer-error disposition asymmetry. LGTM contingent — both are notes, neither gates merge.

crates/networking/p2p/sync/full.rs
iovoid: If the previous step didn't finish correctly, we shouldn't execute the pending blocks (which would fail because the parent state is missing)

crates/networking/p2p/sync/full.rs
```diff
@@ -236,11 +403,18 @@ pub async fn sync_cycle_full(
                         blocks.extend(block_batch);
                     }
                     Ok(None) => {
-                        let _ = body_tx.send(Err(SyncError::BodiesNotFound)).await;
```
iovoid, line 273: BodiesNotFound is now never constructed and should be removed

crates/networking/p2p/sync/full.rs
```diff
@@ -53,6 +56,89 @@ fn humanize_secs(secs: u64) -> String {
...
+/// before its state was computed; see `apply_fork_choice`) are NOT resume points: building
+/// on them fails forever with `state root missing`, so full sync must keep and re-execute
+/// them rather than skip them as "already canonical".
+pub fn is_resume_point(store: &Store, header: &BlockHeader) -> Result<bool, SyncError> {
+    Ok(store.is_canonical_sync(header.hash())? && store.has_state_root(header.state_root)?)
+}
+
+/// Index of the first resume point in a single newest->oldest header batch, or `None` if the
+/// batch contains none. The headers before that index are the missing blocks to execute; the
+/// header at that index is our executed/state head (state presence is contiguous from genesis,
```
iovoid: "state presence is contiguous from genesis" seems false, we only keep very recent state

crates/networking/p2p/sync/full.rs
```diff
@@ -101,29 +195,38 @@ pub async fn sync_cycle_full(
...
             .request_block_headers_from_hash(sync_head, BlockRequestOrder::NewToOld)
-            .await?
-        else {
-            if attempts >= MAX_HEADER_FETCH_ATTEMPTS {
+            .await?;
+        let mut block_headers = match outcome {
+            HeaderFetchOutcome::Headers(headers) => headers,
+            // No headers this round: `reason` says whether we couldn't find a peer to ask
+            // ("no eth peer available") or peers were asked and didn't serve ("peer(s) asked
+            // but did not serve headers"), so operators can tell connectivity apart from
```
iovoid: The error messages in quotes are not from ethrex.

iovoid (approved):

ilitteri (approved):

Issue #6805, opened 2026-06-08
sync: measure and act on peer-serving efficiency (full/snap sync)


edg-l:
## Context

Surfaced while triaging ethrex nodes on glamsterdam-devnet-5 (see #6803). When sync stalls, it's currently hard to tell whether ethrex is the problem or peers simply aren't serving data, and ethrex does little to *act* on chronically unhelpful peers. On a thin devnet this is bounded by peer availability, but on mainnet snap/full sync it directly affects sync speed and robustness.

## What ethrex measures today

- Snap *progress* metrics (`crates/blockchain/metrics/sync.rs`): stage, pivot, headers/accounts/storage/bytecodes downloaded/inserted/healed, inflight_requests, eligible_peers. All "how far along", not "how well peers serve".
- Basic per-peer score (`crates/networking/p2p/peer_table.rs`): `score` in [-150, 50], +1 per success, -1 per failure, -150 on a critical (validation) failure. Used only to *down-weight* `get_random_peer` selection. Bad peers are never excluded or evicted, and scoring ignores latency and request type.
- A partial `header_resolution` outcome counter.

Missing: per-request success/empty/timeout/invalid accounting, request latency, served-vs-issued ratio, and any eviction/replacement of non-serving peers.

## Plan (ranked)

**Item 1. Sync peer-efficiency metrics.** Per-peer and aggregate request outcomes (served / empty / timeout / invalid) and latency histograms for headers and bodies. Exposed via Prometheus; per-peer detail kept in `PeerData` (in-node), aggregate-only labels in Prometheus.

**Item 2. Act on efficiency.** (2a) Latency/severity-aware scoring instead of flat +/-1. (2b) Evict chronically non-serving peers to free a slot and let the initiator dial a fresh fork-valid replacement; optionally exclude (not just down-weight) sub-threshold peers when better ones exist. Guard: only evict when replacements are available.

**Item 3. Sync-aware discovery escalation + dial-on-starvation.** Today `get_lookup_interval` (`discovery/server.rs`) keys only on `target_peers_completion` (peers / TARGET_PEERS). When the *useful* pool collapses but raw count looks fine, discovery throttles exactly when it should escalate. Add a sync-starved signal (shared atomic, like `snap_enabled`) that forces fast lookups + proactive dialing when header/body fetch repeatedly fails, with a cooldown.

[...]

PR #6806, opened 2026-06-08
feat(l1): measure block-request outcomes and score peers by serve quality

edg-l:
Part of #6805. First PR in the sync peer-serving efficiency series.

Background: while triaging ethrex nodes on a degraded devnet (#6803) it was difficult to distinguish ethrex bugs from peers simply not serving data, and ethrex did nothing to deprioritize unhelpful peers over time.

**Measurement.** `PeerData` gains per-peer counters (served/empty/timeouts/invalid) and an EWMA of response latency (alpha=0.2). `peer_handler.rs` records a `RequestOutcome` at the header and body request boundaries.

**Scoring.** `record_success`/`record_failure` calls are replaced by `record_request_outcome`, which applies severity-aware score deltas: served +1, empty -1, timeout -2, invalid sets MIN_SCORE_CRITICAL. `get_random_peer` already weights selection by score, so chronically unhelpful peers are deprioritized automatically with no other changes.

**Observability.** Behind `--features metrics`: `ethrex_sync_block_requests_total` counter (labels: kind=headers|bodies, outcome=served|empty|timeout|invalid) and `ethrex_sync_block_request_latency_ms` histogram (label: kind, buckets 1ms-5s). Latency is only observed on Served outcomes. Aggregate labels only; per-peer detail stays in `PeerData`, not Prometheus. The metrics are observability only and do not drive node logic.

**Scoped out** (follow-up in #6805): peer eviction/replacement for chronically non-serving peers, and basing eligibility checks on live-connected peers. Those touch the RLPx reconnect lifecycle and need separate runtime/test validation.

Verified: `cargo check -p ethrex-p2p`, `cargo check -p ethrex-p2p --features metrics`, `cargo clippy -p ethrex-p2p --features metrics` (zero warnings), `cargo test -p ethrex-p2p` (50 passed).


crates/networking/p2p/peer_table.rs
```diff
@@ -651,6 +687,43 @@ impl PeerTableServer {
...
+            .unwrap_or_default()
+            .as_secs();
+        self.peers
+            .entry(msg.node_id)
+            .and_modify(|p| match msg.outcome {
+                RequestOutcome::Served => {
+                    p.score = (p.score + 1).min(MAX_SCORE);
+                    p.served += 1;
+                    p.last_response_time = Some(now);
+                    p.ewma_latency_ms = Some(match p.ewma_latency_ms {
```
ElFantasma: **EWMA + histogram are Served-only by design, but the latency_ms parameter is passed across the actor for every outcome.** Three of the four arms (Empty/Timeout/Invalid) ignore `msg.latency_ms` entirely — it's collected, serialized into the message, and dropped.

Two cleaner shapes:

1. **Always observe latency in the histogram** (and either expose a per-outcome label or keep histogram outcome-agnostic). A 5000ms reading on a Timeout, or a 50ms reading on an Empty, is operationally useful — "empty responses come back instantly" vs "empty responses come back slow" are different stories. The EWMA stays Served-only (an empty/timeout latency isn't a serve-quality signal), but the histogram captures the wire-time distribution regardless.

2. **Drop `latency_ms` from non-Served paths** at the call site. The `RequestOutcome` enum could carry the latency only for `Served { latency_ms: u64 }` — type-level enforcement of the invariant. Less ergonomic but eliminates the silently-discarded argument.

Option 1 is the higher-leverage one; option 2 is the purer one. Non-blocking either way.

crates/networking/p2p/peer_handler.rs
```diff
@@ -429,24 +432,63 @@ impl PeerHandler {
                             "[SYNCING] Received empty headers from peer {peer_id}, trying another"
                         );
                         let _ = self.peer_table.set_disposable(peer_id);
```
ElFantasma: **Double-action on `headers-empty`: `set_disposable` AND `record_request_outcome(Empty)`.** Pre-PR this was just `set_disposable`; the PR keeps the dispose and adds the score decrement on top. Other outcome paths in this PR don't dispose: `headers-invalid` records (-MIN_SCORE_CRITICAL via Invalid) but doesn't dispose; `headers-timeout` records (-2) but doesn't dispose; `bodies-timeout` records (-2) AND disposes at line 593.

The set of arms that dispose-and-record is `{headers-empty, bodies-timeout}`. That's an oddly specific shape — was it deliberate ("these two are unrecoverable enough to evict immediately"), or an inconsistency from migrating piecemeal? If deliberate, worth a one-line comment justifying the policy. If accidental, the cleanest move is either:

- Drop `set_disposable` and let scoring drive eviction (consistent across all failure outcomes), or
- Add `set_disposable` to `Invalid` too (it's the most severe — the peer sent a protocol-violating response).

Non-blocking but worth nailing down before this becomes the next maintainer's puzzle.

crates/networking/p2p/peer_table.rs
```diff
@@ -651,6 +687,43 @@ impl PeerTableServer {
...
+                RequestOutcome::Empty => {
+                    p.score = (p.score - 1).max(MIN_SCORE);
+                    p.empty += 1;
+                }
+                RequestOutcome::Timeout => {
+                    p.score = (p.score - 2).max(MIN_SCORE);
+                    p.timeouts += 1;
+                }
+                RequestOutcome::Invalid => {
+                    p.score = MIN_SCORE_CRITICAL;
```
ElFantasma: nit: `RequestOutcome::Invalid` setting `MIN_SCORE_CRITICAL` makes the existing `record_critical_failure` API a near-duplicate — both end with the same `score = MIN_SCORE_CRITICAL`. There's still one caller of `record_critical_failure` (peer_handler.rs:624, looks like a snap path), so it can't be removed yet, but if the snap path also migrates to outcome-recording in the next PR of the series, `record_critical_failure` becomes dead. Worth a TODO marker or follow-up issue.

ilitteri (approved):

PR #6807, opened 2026-06-08
feat(l1): evict chronically non-serving peers and count live-connected peers

edg-l:
Stacked on #6806 (peer-efficiency measurement + scoring); part of #6805. Must merge after #6806.

Builds on the per-peer request stats from #6806 to act on unhelpful peers and to extend the instrumentation across all sync request types.

## Conservative peer eviction
`should_evict_peer(score, in_flight, served, failures, replacement_available)` (pure, unit-tested) evicts a peer only when it is idle (no in-flight requests) AND a fork-valid replacement contact is available AND it is either a protocol violator (score floored to critical) or has never served us despite >= 8 cumulative failures (empty + timeout + invalid). Wired into the request-outcome handler; eviction is `self.peers.swap_remove` and the initiator auto-dials a replacement once the slot frees (it dials from the connection pool whenever below the peer target). Evaluated on failure outcomes only, so a served request never triggers the check. A non-mutating `has_dialable_contact` helper mirrors the dialer's eligibility (knows_us, not unwanted, fork-id not invalid, not already a peer).

Eviction is intentionally conservative to avoid stranding a node on a thin network: it will not evict if no replacement is available, and the connection teardown reuses the existing remove-peer path.

## Efficiency stats for snap state requests
The snap state-request path (`account_range`, `bytecodes`, `storage_ranges`, and the `state`/`storage` trienode requests in `snap/client.rs`) previously recorded outcomes with the old flat `record_success`/`record_failure`. It now uses `record_request_outcome`, with the outcome (served/empty/timeout/invalid) and latency classified in each worker and threaded back through the result channels. The two synchronous trienode functions take `peer_id`/`peer_table` params (healing callers updated). This brings severity-aware scoring, eviction, and the request-efficiency metrics to the dominant snap-sync phase (state download), not just block fetch.

## Live-connected accounting + metric naming
Adds `PeerDiagnostics.connected` and a `ethrex_sync_connected_eth_peers` gauge, so the actually-askable peer count is visible separately from capable-but-disconnected peers. Since the request-efficiency metrics now cover both block and snap requests, they are renamed from `ethrex_sync_block_requests_total` / `_latency_ms` to `ethrex_sync_peer_requests_total` / `_latency_ms` (helpers `inc_peer_request` / `observe_peer_request_latency`).

## Bound outbound TCP connect
The dialer (`tcp_stream` in rlpx/connection/handshake.rs) used a bare `TcpSocket::connect().await` with no timeout, so dialing an unreachable peer (stale ENR, NAT'd or offline node) blocked until the OS SYN timeout (~127s on Linux). Dead dials piled up in SYN-SENT while the initiator kept spawning new ones (observed ~150 stuck dials vs ~8 real peers on a Hoodi sync, where EL discovery surfaces many dead ENRs). Wrap the connect in a 5s timeout so dead dials are abandoned quickly and the dial pipeline reaches live peers faster.

## Validated live on Hoodi
Ran ethrex + lighthouse through a full Hoodi snap sync with this branch:
- Eviction fired correctly on multiple chronically-non-serving peers (score=-50, served=0, failures>=8), with replacements dialed.
- The snap request-efficiency metrics recorded all kinds live (account_range / storage_ranges / bytecodes / trienodes, served/empty/timeout).
- The connect timeout dropped the stuck-dial pileup from ~150 SYN-SENT to <10 and the node reached the available peer supply faster (grew past the pre-fix ~7-8 plateau to 15+); full snap sync completed in ~18 min on a low peer count, CL-verified.

## Notes
- The evict-and-replace path touches the live-connection RLPx lifecycle (which has a known reconnect race), so it benefits from runtime validation in addition to the unit tests.

## Verification
`cargo check -p ethrex-p2p` (+ `--features metrics`), `cargo clippy -p ethrex-p2p --features metrics` zero warnings, `cargo test -p ethrex-p2p` 56 passed (incl. 6 eviction-predicate tests).


crates/blockchain/metrics/sync.rs
```diff
@@ -171,19 +177,19 @@ impl MetricsSync {
                 &["outcome"]
             )
             .expect("Failed to create ethrex_sync_header_resolution_total"),
-            block_requests: register_int_counter_vec!(
-                "ethrex_sync_block_requests_total",
+            peer_requests: register_int_counter_vec!(
+                "ethrex_sync_peer_requests_total",
                 "Total block requests by kind (headers|bodies) and outcome (served|empty|timeout|invalid)",
```
ElFantasma, line 182: nit: this help string is now stale — the metric receives `account_range`, `bytecodes`, `storage_ranges`, and `trienodes` labels too (wired in `snap/client.rs`), so "Total block requests by kind (headers|bodies)" understates the surface. Same story for `peer_request_latency_ms` at line 188 ("for block requests that received a response"). Maybe:

```rust
"Total peer requests by kind (headers|bodies|account_range|bytecodes|storage_ranges|trienodes) and outcome (served|empty|timeout|invalid)",
```

or — if you'd rather not commit to an enumeration that will drift — "Total peer requests by kind and outcome, where kind covers both block and snap request types". Same fix on the latency line.

Non-blocking, but the labels are what someone grepping Prometheus docs sees first.

crates/networking/p2p/snap/client.rs
```diff
@@ -1051,9 +1125,27 @@ pub async fn request_state_trienodes(
...
+            #[cfg(feature = "metrics")]
+            {
+                use ethrex_metrics::sync::METRICS_SYNC;
+                METRICS_SYNC.inc_peer_request("trienodes", "timeout");
+            }
+            return Err(e);
+        }
+    };
 
     if nodes.is_empty() || nodes.len() > expected_nodes {
```
ElFantasma, line 1247: This branch collapses two semantically different responses into a single `RequestOutcome::Empty`:

- `nodes.is_empty()` — the peer truthfully said "I don't have these" (cheap, not a protocol violation)
- `nodes.len() > expected_nodes` — the peer sent more nodes than we asked for (clear protocol violation; another part of the file scores `Invalid` for similar misbehavior, e.g. `request_storage_trienodes` mapping `Ok(other_msg)` to `Invalid` at l.768)

With the score weights from #6806, `Empty` and `Invalid` aren't equivalent — `Invalid` carries more weight against the peer and feeds the eviction predicate's `failures` count differently in spirit (a buggy peer hurts us more than a polite "nothing here"). Suggested split:

```rust
let (outcome, outcome_str) = if nodes.is_empty() {
    (RequestOutcome::Empty, "empty")
} else {
    (RequestOutcome::Invalid, "invalid")
};
peer_table.record_request_outcome(peer_id, outcome, latency_ms)?;
#[cfg(feature = "metrics")]
{
    use ethrex_metrics::sync::METRICS_SYNC;
    METRICS_SYNC.inc_peer_request("trienodes", outcome_str);
}
```

Non-blocking — the request still fails closed via `SnapError::InvalidData`, so correctness is unchanged. Just a scoring/observability refinement.

crates/networking/p2p/peer_table.rs
```diff
@@ -1335,6 +1389,26 @@ impl PeerTableServer {
         }
     }
 
+    /// Returns true if there is at least one node in the connection pool that
+    /// is not already a connected peer, is not marked unwanted, has acknowledged
+    /// us (knows_us == true), and has a valid (or unknown) fork ID.
+    /// Mirrors the eligibility criteria of `do_get_contact_to_initiate` but is
+    /// read-only and does not mutate `already_tried_peers`.
+    fn has_dialable_contact(&self) -> bool {
+        for (node_id, _) in &self.connection_pool {
```
ElFantasma, line 1403: Cost characterization for future tuning: this loop is O(`connection_pool.len()`) hash lookups in the worst case (`MAX_CONNECTION_POOL_SIZE = 10_000` at peer_table.rs:69), called on every non-served outcome — which during a noisy peer phase can be every request from every snap worker. Common case it returns on first dialable contact, so amortized it's fine; worst case is when the pool is full of stale/unwanted/no-knows-us entries and you scan all 10K.

Not a problem now (early-exit is fast in practice), but if the pool ever fills with unhelpful entries this becomes a hot-path footgun. Two easy mitigations if it shows up in a flamegraph later:

1. Bound the scan: `take(N)` after a small constant like 64 — if you can't find a dialable contact in 64 tries the pool is mostly garbage anyway.
2. Cache the boolean for a short window (e.g. invalidate on contact mutation), since the answer is stable across thousands of failure outcomes.

Purely a perf-defense observation, not a request to change anything in this PR.

ElFantasma (approved):
Left some minor comments, but LGTM

PR #6808, opened 2026-06-08
fix(l1): return null, not -38005, for engine_getBlobsV2/V3 before Osaka

edg-l:
`engine_getBlobsV2` and `engine_getBlobsV3` were returning `-38005 "getBlobsVN engine only supported for Osaka"` whenever the canonical head was pre-Osaka. The `-38005` ("Unsupported fork") rule in the engine-API osaka.md spec goes the other direction: `getBlobsV1` MUST return `-38005` *after* Osaka. V2/V3 have no such pre-fork error; their contract is to return `null` for any blob they cannot serve, including while syncing.

In practice, because the gate keyed on the local head's fork, it fired throughout snap/full sync while the node was still catching up to Osaka. This spammed the consensus client with "Execution engine call failed ... Unsupported fork: getBlobsV2" errors. Found live on a Hoodi sync with lighthouse v8.1.3.

Fix in `crates/networking/rpc/engine/blobs.rs`: before the canonical tip is at Osaka, `get_blobs_and_proof` now returns a `null` entry for every requested hash instead of erroring. The V2 handler collapses to a top-level `null` (one missing blob voids the whole response per the V2 spec); the V3 handler returns the per-hash null array. The `version` parameter the helper no longer needs is dropped. The getBlobsV1 post-Osaka gate is correct and unchanged.

Test `blobs_v3_requires_osaka` renamed to `blobs_v3_returns_null_before_osaka` and updated to assert a null array rather than an error; all 7 blobs tests pass.


ElFantasma (approved):
One inline on the V1 post-Osaka error message phrasing — actively misleading right now. LGTM contingent on that small wording fix.

---

**Inline (V1 message wording):**

Message phrasing is inverted relative to the trigger: the gate fires *after* Osaka, but the message reads "only supported **before** Osaka". When this lands in a CL log a reader will see `"Unsupported fork: getBlobsV1 engine only supported before Osaka"` and wonder why it fired on a pre-Osaka-looking chain — exactly the confusion this PR is trying to remove on the V2/V3 side.

Clearer phrasing options:
- `"getBlobsV1 not supported after Osaka; use getBlobsV2 or V3"`
- `"getBlobsV1 engine method is deprecated after Osaka; use V2/V3"`

The second one points the CL operator at the recovery action.

ilitteri (approved):

iovoid (approved):
Interesting note, but clients disagree on whether returning null (can't answer) or [null; N] (can answer, all missing) is the correct way. Both have the same effect in practice though.

PR #6812, opened 2026-06-08
feat(l1): add testing_buildBlockV1 and debug_setHead RPC methods

edg-l:
Implements `testing_buildBlockV1` (execution-apis `src/testing/testing_buildBlockV1.yaml`) and `debug_setHead` (geth-compatible head rewind), intended for test fixture generation flows.

**testing_buildBlockV1**: builds a block on top of `parentBlockHash` from given `PayloadAttributesV3` and an optional explicit transaction list; returns a `getPayload`-style response (`executionPayload`, `blockValue`, `blobsBundle`, `shouldOverrideBuilder`, `executionRequests`). `transactions: null` pulls from mempool; `[]` produces an empty block; a non-empty array applies exactly those txs in order, no mempool txs. Blob (type-3) txs in the explicit list are accepted: the canonical encoding carries no sidecar, so blob gas is accounted from the tx's versioned hashes and the returned `blobsBundle` is left empty. Callers needing the blob sidecars must produce them out-of-band. Parent must equal the current canonical head; timestamp must be strictly greater than parent timestamp.

**debug_setHead**: calls `apply_fork_choice` with head/safe/finalized all pointing at the target block, rewinding the canonical chain. Limitation: `apply_fork_choice` refuses to rewind below the stored finalized number, so this works correctly only when finalized is genesis (the normal case for test flows).

**New `testing` namespace** (`crates/networking/rpc/utils.rs`, `rpc.rs`, `testing.rs`): disabled by default, opt-in via `--http.api testing`, served on the public HTTP port only, never on the authenticated engine port. Documented in `--help` and the namespace parser error message.

**Blockchain refactor** (`crates/blockchain/payload.rs`): `build_payload` is now a thin wrapper around `build_payload_inner`. New `build_payload_with_transactions` exposes the same pipeline with mempool selection replaced by `fill_explicit_transactions`.

**Tests**: unit tests for param parsing, parent-not-head rejection, a real empty-block build on the genesis head, and a blob-tx build asserting an empty `blobsBundle` with blob gas accounted; a CLI test that `--http.api testing` parses to `RpcNamespace::Testing`.


crates/blockchain/payload.rs
```diff
@@ -783,13 +843,22 @@ impl Blockchain {
...
-            return Err(
-                StoreError::Custom(format!("No blobs bundle found for blob tx {tx_hash}")).into(),
-            );
+        // The blob count drives blob-gas accounting. Normally it comes from the
+        // mempool sidecar; for an explicit build (`testing_buildBlockV1`) there is
+        // no sidecar, so derive it from the tx's versioned hashes and leave the
+        // blobs bundle empty (the EVM only needs the hashes, which are in the tx).
+        let (blob_count, bundle) = match self.mempool.get_blobs_bundle(tx_hash)? {
+            Some(blobs_bundle) => (blobs_bundle.blobs.len(), Some(blobs_bundle)),
+            None if context.explicit_build => ((**head).blob_versioned_hashes().len(), None),
```
ElFantasma, line 864: **Body-vs-code mismatch on blob (type-3) txs:**

The PR body says: *"Blob (type-3) txs in the explicit list are rejected with a clean error since canonical encoding carries no blob data (geth panics there; we don't)."*

The actual code here accepts them — `None if context.explicit_build => ((**head).blob_versioned_hashes().len(), None)` derives blob count from the tx's versioned hashes and uses an empty bundle, the EVM runs against the hashes, and the tx ends up in `payload.body.transactions` with the blob-gas accounting applied. There's no `return Err(...)` path for type-3 txs in `fill_explicit_transactions` or `apply_blob_transaction`.

This is a meaningful semantic gap for the consumer: the resulting `executionPayload` includes a blob tx, but the returned `blobsBundle` is empty — so a caller using `testing_buildBlockV1` to produce a fixture for engine API submission has no way to obtain the sidecars matching the built block. Pick one:

1. **If the intent is to reject**: add an explicit check in `fill_explicit_transactions` (`if matches!(tx, Transaction::EIP4844Transaction(_)) return Err(ChainError::Custom("blob txs not supported in explicit list"))`) and update the doc-comment at line 720 to match the body. Add a unit test asserting the rejection.
2. **If the intent is to accept with an empty bundle** (which is what the code does today): update the PR body to describe the actual behavior — "Blob txs are accepted; blob gas is accounted from the tx's versioned hashes and the bundle is left empty. Callers needing blob sidecars must produce them out-of-band; the canonical encoding accepted here doesn't carry them." Add a unit test asserting a built block with a blob tx returns an empty `blobsBundle`.

Either shape is fine; the gap is that body and code disagree, and a reader can't tell which is authoritative.

ElFantasma (approved):

ilitteri (approved):

iovoid (approved):

Issue #6785, opened 2026-06-03
perf(l1,rpc): add a background-built inverted log index (address/topic → block) for eth_getLogs


ElFantasma:
Part of #6745. Follow-up from the cross-client benchmark (#6749) and hotspot confirmation (#6750).

## Problem

`eth_getLogs` (`crates/networking/rpc/eth/logs.rs`) iterates every block in `[fromBlock, toBlock]`, loads each block's receipts, and filters logs by address/topic **in memory**. There is no `(address, topic)` index, so cost scales with the **block range** regardless of how many logs actually match — an O(blocks × txs/block) full receipt scan.

## Evidence (mainnet-10, recent ranges, LUSD Transfer)

Single-call latency, ethrex vs geth (geth 1.17.3, pruned, same box/workload):

| range | ethrex | geth | geth speedup |
|---|---|---|---|
| 100 blocks | 618 ms | 2.5 ms | ~250× |
| 1,000 blocks | 6,131 ms | 35 ms | ~175× |
| 10,000 blocks | 70,441 ms | 204 ms | ~340× |

Cost is **~6 ms per block scanned**, independent of result size (LUSD barely matches — responses are 0–80 KB). Under a rate sweep ethrex collapses: the 100-block variant drops to ~1.4 rps with 30 s timeouts at 100 rps; 1k/10k-block ranges saturate even at 10 rps. geth stays fast because it maintains a **log index**.

This is the single largest gap found in the cross-client benchmark, on a top-10 traffic method.

## Direction

Build a log index so `eth_getLogs` doesn't rescan receipts per block. Options to evaluate:

- **Per-block-range bloom filters** (geth-style "bloombits"): compact, lets the query skip block ranges that can't contain the address/topic. Lower storage overhead; still scans candidate ranges.
- **Inverted index** keyed by `(address)` and/or `(topic0)` → block numbers (or block-range buckets): faster lookups, larger storage; needs a write-path update on block import.

Considerations:
- Write-path cost on block import (must stay within the block-processing budget — this codebase is sensitive to `store_ms` regressions).
- Backfill/migration for existing data (the repo has a migration framework, see #6519) vs. index-from-now-forward.
- Storage footprint (relevant on snap nodes with limited retention).
- L2 applicability.

## Scope / done when

- [ ] Decide index design (bloom-bits vs inverted index) with a short written trade-off (storage, write-path cost, query speedup).
- [ ] Implement the index + write-path population.
[...]

PR #6813, opened 2026-06-08
perf(l1): skip non-matching blocks in eth_getLogs via header bloom

ElFantasma:
**Motivation**

`eth_getLogs` iterates every block in `[fromBlock, toBlock]`, loads each block's body and every receipt, and filters logs in memory — an O(blocks × txs) full receipt scan whose cost is independent of how many logs actually match (~6 ms/block). The cross-client benchmark measured ethrex at 175–340× slower than geth on this top-10 traffic method.

Every block header already carries a `logs_bloom` over exactly the `(address, topic)` pairs we filter on, but the endpoint never consulted it.

**Description**

First, lowest-cost step toward the log index in #6785: use the header bloom as a prefilter so blocks that provably can't contain a matching log are skipped without loading their body or receipts.

- In `fetch_logs_with_filter`, fetch the header first and run `block_bloom_matches` against `header.logs_bloom` before touching the body/receipts; skip the block on a miss.
- `block_bloom_matches` is a necessary-condition check via `Bloom::contains_input` (which keccak-hashes its input, mirroring how `bloom_from_logs` builds the header bloom): requires at least one requested address present (if any) AND, for each constrained topic position, at least one allowed topic present. Wildcards impose no constraint. Bloom false positives are harmless — exact filtering still runs on non-skipped blocks, so results are unchanged.
- Added unit tests covering the matching semantics (address OR, topic OR-within / AND-across positions, wildcards, combined address+topic).

Zero storage cost, zero write-path cost, no migration — the bloom is already in the header. For sparse queries over large ranges (the benchmark's worst case) this skips nearly every block. A transposed/sectioned bloom index (geth-style bloombits) for dense or very large ranges is tracked as a follow-up in #6785 pending re-benchmark.

Part of #6785.

**Checklist**

- [x] No changes to the `Store`; `STORE_SCHEMA_VERSION` unaffected.


ElFantasma:
## Benchmark: measured on mainnet-10 (warm, synced)

Ran this branch on a synced mainnet node (`vegeta`, recent-data workload, every call pre-validated to return a real result). Comparing `eth_getLogs` on LUSD Transfer ranges vs the pre-fix baseline.

### Single-call latency (warm)

| range | pre-fix | this PR | speedup |
|---|---|---|---|
| 100-block (median) | ~623 ms | **~306 ms** | ~2× |
| 100-block (no matches) | ~618 ms | **~120 ms** | ~5× |
| 1,000-block (median) | ~6,131 ms | **~3,273 ms** | ~1.9× |

### Under concurrent load — 100-block range (the bigger win)

| rate | pre-fix | this PR |
|---|---|---|
| 10 rps | 623 ms, 100% success | 306 ms, 100% |
| 100 rps | **collapses: 10% success, ~1 rps** | **95% success, ~22 rps** |
| 1000 rps | 1% success, ~0 rps | 92% success, ~21 rps |

The header-bloom skip roughly **doubles single-call latency and ~20×'s sustained throughput under load** for small ranges — it no longer collapses at 100 rps. 

### Where it's bounded
- Still **O(blocks-in-range)**: it reads every block's *header* for the bloom (~1.2 ms/block warm) and still loads receipts for *matching* blocks. So large ranges still saturate: the **1,000-block range times out under any concurrency** (5% success at 10 rps, 0% above; single-call ~3.3 s).
- This doesn't reach indexed clients (geth ~2.5/35 ms, reth ~4.8/49 ms, nethermind ~15/97 ms for 100/1000-block). It's a strong complement to a real log index (#6785), not a replacement.

### No regressions
Other endpoints unchanged (this PR only touches `logs.rs`): `eth_call` 0.09 ms, `eth_getBalance` 0.06, `eth_getBlockByNumber` 0.92, `eth_getTransactionReceipt` 1.31 (p50 @1000 rps) — all within noise of the baseline.

### Caveats
- Measured on a 5k-block recent window (freshly-synced node hadn't accumulated more); the pre-fix side full-scans regardless of window, so the comparison holds for the ranges tested.
- The 10,000-block range wasn't measurable yet (needs more accumulated history) — that's where the per-block-header-read ceiling would show most.

**Verdict: solid, low-risk win for small/sparse `eth_getLogs` (and removes the under-load collapse). Large ranges still need the index work in #6785.**

ElFantasma:
Pushed `65bca2a0b` addressing the review findings:

- **Correctness (Codex + Claude):** the bloom prefilter now treats a `Topics([...])` alternatives list containing any `None` as a wildcard for that position (returns early without constraining the bloom), instead of `flatten()`-ing the `None` away. This fixes the false-negative where `topics: [[null, T]]` could skip blocks that match via the wildcard and silently drop logs.
- **Regression test:** added `bloom_match_topics_with_none_element_is_wildcard` covering `Topics([Some(t), None])` against a bloom missing `t` (fails before the fix, passes after).
- **CHANGELOG:** added a `## Perf` entry (fixes the `changelog` check).

The `Vec` allocation can stay or be tidied into iterator chains per Kimi's nit — happy to fold that in if preferred. Perf is unchanged by this fix (it only adds a wildcard short-circuit; doesn't touch the constrained-topic path the benchmarks exercised).

ElFantasma:
Note: the failing **Run benchmark against base branch** check is **not caused by this PR** — it's a pre-existing `main` breakage tracked in #6819 (the `base`/main binary fails to import the stale `l2-1k-erc20.rlp` fixture under #6766's new logs_bloom validation). This PR only touches the RPC `eth_getLogs` read path. Safe to disregard that check here.

ElFantasma:
### Update: 10,000-block range now measured

The earlier results left the 10k-block range pending (the freshly-synced node hadn't accumulated enough history). It has now, so here's the complete single-call `eth_getLogs` picture (warm, LUSD Transfer):

| range | pre-fix | this PR | speedup | geth | reth | nethermind |
|---|---|---|---|---|---|---|
| 100-block | 618 ms | 415 ms | ~1.5× | 2.5 | 4.8 | 15 |
| 1,000-block | 6,131 ms | 3,529 ms | ~1.7× | 35 | 49 | 97 |
| 10,000-block | 70,441 ms | **31,857 ms** | **~2.2×** | 204 | 507 | 1,074 |

The header-bloom skip is a consistent ~1.5–2.2× win that grows with range size. But because it stays O(blocks-in-range) on header reads, the 10k range is still ~32 s — ~30–150× slower than the indexed clients. So this PR is a solid constant-factor improvement (and removes the under-load collapse for small ranges), while large-range `eth_getLogs` still needs a real log index (tracked in #6785).

edg-l (approved):
maybe tests should go to the test/ dir

crates/networking/rpc/eth/logs.rs
```diff
@@ -231,6 +238,52 @@ pub(crate) async fn fetch_logs_with_filter(
...
+    if !address_filter.is_empty()
+        && !address_filter
+            .iter()
+            .any(|address| bloom.contains_input(BloomInput::Raw(address.as_bytes())))
+    {
+        return false;
+    }
+
+    topics.iter().all(|topic_filter| {
+        let allowed: Vec<&H256> = match topic_filter {
```
ilitteri: Can't we avoid the `collect`s here and perform the checks against an iterator?

ElFantasma: addressed in `ea0b2b089`: the topic check no longer collects into a `Vec<&H256>`; it matches each `TopicFilter` and short-circuits directly over iterators (`any(...)`), preserving the `None`-wildcard semantics and the empty-list wildcard. Behavior-identical — existing `bloom_match_*` tests still pass.

ElFantasma:
> maybe tests should go to the test/ dir

We had an intention to move tests to their own folder, but it seems we ended up keeping unit tests inline. `test/` contains mainly **integration** tests, whereas these are **unit** tests of the private `block_bloom_matches` helper — moving them out would mean exposing it (and its module path) `pub` solely for tests, which we'd rather not do. So the convention I'm following is: integration tests → `test/`, unit tests → inline next to the code.
We can move to **all tests to the `test/` dir** but if that's the case there is a lot of refactor to do everywhere.

ilitteri (approved):

iovoid (commented):
This could be made more efficient by constructing the filter once and then calling `Bloom::contains_bloom`

This is what `Bloom::contains_input` does internally.

ElFantasma:
> This could be made more efficient by constructing the filter once and then calling `Bloom::contains_bloom`
> 
> This is what `Bloom::contains_input` does internally.

Good catch, agreed. `contains_input` re-derives each input's bloom via `Bloom::from(..)` (keccak + bit extraction) on every call, and since `block_bloom_matches` runs per block, we're recomputing the same address/topic blooms once per block. For wide-range queries that skip most blocks — the case this prefilter is meant to accelerate — that's the dominant per-block cost. Precomputing the blooms once before the loop and switching to `contains_bloom` removes the redundant hashing.

Since this PR is already approved and is the base of the getLogs stack (#6852 and the inverted-address-index work build on top of it), I've split it into a fast-follow — #6893 — to keep the approvals and avoid cascading a rebase through the stack. Will pick it up once the stack lands.

iovoid (approved):

PR #6814, opened 2026-06-08
fix(l1): add defense-in-depth for fullsync transition on canonical chains with gaps

iovoid:
**Motivation**

When fullsyncing from a snapsynced chain, in the (bugged) case of the canonical chain having gaps, when we need to fullsync we go back indefinitely trying to find a point to merge with the canonical chain.

**Description**

When the starting block of the batch to execute is below the canonical head, try to merge even if that starting block isn't canonical itself.


crates/networking/p2p/sync/full.rs
```diff
@@ -166,7 +171,11 @@ pub async fn sync_cycle_full(
         start_block_number = last_header.number;
 
         sync_head = last_header.parent_hash;
-        if store.is_canonical_sync(sync_head)? || sync_head.is_zero() {
+        let parent_is_canonical = store.is_canonical_sync(sync_head)? || sync_head.is_zero();
+        // The batch has descended to our head: scan it for the canonical ancestor
+        // the lowest-parent check above can miss (e.g. a lone snap-sync pivot).
+        let reached_local_head = last_header.number <= local_latest_number;
+        if parent_is_canonical || reached_local_head {
```
ElFantasma, line 178: When this branch enters via `reached_local_head` (the new exit) and the inner guard *doesn't* merge (no canonical ancestor in the batch), the walk silently keeps going. A `debug!("sync walk descended below local head=N without finding canonical ancestor; continuing to genesis")` here — or right after the inner `if`'s else-path — would make this much easier to triage from logs when it fires. The current code is correct, but the only way to know it took the new path today is to instrument from source.

Also worth a counter or warn-once if the walk gets within, say, 1000 blocks of genesis without finding a common ancestor — that's the "reorg below our entire chain" pathology and surfacing it explicitly beats sync-time confusion.

crates/networking/p2p/sync/full.rs
```diff
@@ -101,6 +101,11 @@ pub async fn sync_cycle_full(
     let mut started_behind = false;
     let mut sync_target_logged = false;
 
+    // Latest canonical block. The backward walk below also stops here: our head
+    // (e.g. a snap-sync pivot) may be a lone canonical block sitting mid-batch,
+    // which the lowest-parent check would skip, running the walk down to genesis.
+    let local_latest_number = store.get_latest_block_number().await?;
```
ElFantasma, line 107: Worth tightening the comment: this is captured *once* outside the loop, and the loop's `store.add_fullsync_batch(...)` call below writes to the separate `FULLSYNC_HEADERS` column family (not `latest_block_header`), so the captured value can't go stale during the walk. That's the invariant the fix relies on — making it explicit here prevents a future refactor of `add_fullsync_batch` (e.g. promoting batches to canonical eagerly) from silently violating it.

Something like:
```
// Captured once: `add_fullsync_batch` below writes to FULLSYNC_HEADERS and does
// not advance `latest_block_header`, so this floor is stable for the duration
// of the walk.
```

ilitteri (approved):

edg-l (approved):

ElFantasma (approved):
LGTM, two minor observations, though.

PR #6816, opened 2026-06-09
fix(l1): validate eth/71 BAL against header commitment and persist BALs during full-sync

edg-l:
Two bugs found on glamsterdam-devnet-5 (block 8501): ethrex served `0xc0` (empty BAL) over eth/71 for a canonical, finalized block whose header commits to a non-empty `block_access_list_hash`. Erigon rejected it; `0x80` (unavailable) would have been acceptable.

**Root cause**: the full-sync batch path executed and stored blocks without fetching or persisting their BALs (only the final batch did). A later eth/71 or engine request fell back to regenerating the BAL by re-executing against the parent state; once that state is pruned, regeneration produces an empty/wrong BAL that was written back unvalidated and then served as `0xc0`.

**Fix 1 (serve guard, spec-mandated):** adds `BlockAccessList::matches_commitment` in `crates/common/types/block_access_list.rs`. The eth/71 serve handler (`crates/networking/p2p/rlpx/connection/server.rs`) and `bal_for_block` in `crates/networking/rpc/engine/payload.rs` now check the stored/regenerated BAL against `block_access_list_hash` before serving or persisting it; any mismatch, regeneration failure, or absence degrades to the `0x80` "unavailable" sentinel. This makes ethrex robust against already-corrupt stored entries regardless of how they got there.

**Fix 2 (retention, root cause):** `full.rs` now fetches BALs for every Amsterdam batch, not just the final one. `add_blocks_in_batch` in `crates/blockchain/blockchain.rs` accepts the per-block BAL slice and persists each entry whose hash matches the header commitment, so peers can serve them over eth/71 without regenerating against pruned state.

Regression test added in `test/tests/p2p/rlpx/block_access_lists_tests.rs` (`matches_commitment_guards_stale_and_empty_bals`); existing RPC/batch tests updated for the new `bals` parameter. BAL/batch/full_sync test suites and workspace build pass.

Refs EIP-8159 (eth/71 Block Access List Exchange).

### DIFF
--- a/crates/networking/rpc/eth/block_access_list.rs
+++ b/crates/networking/rpc/eth/block_access_list.rs
@@ -21,8 +21,14 @@ impl RpcHandler for BlockAccessListRequest {
 
     async fn handle(&self, context: RpcApiContext) -> Result<Value, RpcErr> {
         // Per execution-apis, unknown blocks map to the `notFound` schema (null).
-        let block_hash = match &self.block {
-            BlockIdentifierOrHash::Hash(h) => *h,
+        let (block_hash, commitment) = match &self.block {
+            BlockIdentifierOrHash::Hash(h) => {
+                let commitment = context
+                    .storage
+                    .get_block_header_by_hash(*h)?
+                    .and_then(|header| header.block_access_list_hash);
+                (*h, commitment)
+            }
             BlockIdentifierOrHash::Identifier(id) => {
                 let Some(block_number) = id.resolve_block_number(&context.storage).await? else {
                     return Ok(Value::Null);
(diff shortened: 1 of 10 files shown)


ilitteri (approved):

pablodeymo (approved):

crates/networking/rpc/eth/block_access_list.rs
```diff
@@ -50,7 +60,9 @@ impl RpcHandler for BlockAccessListRequest {
             .generate_bal_for_block(&block)
             .map_err(|e| RpcErr::Internal(format!("Failed to generate BAL: {e}")))?;
 
-        match bal {
+        // Only serve a regenerated BAL that matches the header commitment; a
+        // mismatch means it was re-executed against wrong/incomplete state.
+        match bal.filter(|bal| bal.matches_commitment(commitment)) {
```
ElFantasma, line 65: Asymmetry with the `bal_for_block` engine path: there, both the stored-BAL-mismatch and regenerated-BAL-mismatch cases emit `tracing::warn!`s. Here in the JSON-RPC `eth_getBlockAccessList` handler, both mismatches silently degrade to `Value::Null` — an operator querying this endpoint and getting `null` can't tell whether the block is unknown, has no BAL stored, or has a corrupt-stored-and-mismatching-regenerated BAL.

A `warn!` mirroring the engine path on each of the two mismatch branches (stored-but-mismatches at line 48-50, and regenerated-but-mismatches at line 65) would surface the same diagnostic signal here. Not blocking — the engine path is the higher-traffic surface and already covered — but worth ~3 lines for consistency.

ElFantasma (approved):

PR #6817, opened 2026-06-09
fix(l1): reject transactions with mismatched chain id at block import

ElFantasma:
**Motivation**

A transaction's chain id is bound into its signature — EIP-155 folds it into `v` for legacy txs, and typed transactions (EIP-1559/2930/4844/7702) carry an explicit `chain_id` field that is part of the signing payload. This binds a signature to a single chain so it cannot be replayed elsewhere.

During **block import**, ethrex recovers a typed tx's sender from the tx's *own* chain-id signing domain and never re-checks that chain id against the node's chain. The only chain-id check lives in the mempool (`MempoolError::InvalidChainId`), but blocks arriving via `engine_newPayload` or p2p block gossip/sync never pass through the mempool. As a result a block containing a transaction signed for a *foreign* chain id executes and is **accepted**. geth's signer rejects the same transaction at block validation (`ErrInvalidChainId`), so ethrex would import a block the rest of the network rejects — a consensus divergence reachable on every imported block.

**Description**

- Adds `InvalidBlockError::InvalidTransactionChainId { have, want }`.
- Adds a per-transaction chain-id check to the shared `validate_block_pre_execution` (`crates/common/validation.rs`): every transaction whose `chain_id()` is `Some` and does not equal `chain_config.chain_id` is rejected pre-execution. Because that validator is called by the full-node import path, the L2 block producer, **and** the zkVM guest program, the check is enforced everywhere a block is validated (same approach as #6766 for `logs_bloom`).
- Legacy pre-EIP-155 transactions carry no chain id (`chain_id() == None`) and are intentionally left untouched (replay-allowed by design), mirroring the existing mempool check.
- `PrivilegedL2Transaction` is exempt: its sender comes from an unsigned `from` field, so chain id is not a signature-binding scalar for it, and on L2 it may legitimately name a different *source* chain for cross-chain deposits (`l1_watcher.rs` builds mint txs with `chain_id = source_chain_id`). On L1 privileged txs are already rejected as an unsupported type (#6752), so the exemption opens no L1 gap.

**Test**

`test/tests/blockchain/wrong_chain_id_tests.rs` builds a fully-valid block (correct post-state, roots, gas) whose only defect is an EIP-1559 transaction signed for a foreign chain id, and asserts `add_block` rejects it. Confirmed **red before the fix** — `add_block` returned `Ok(())`, i.e. ethrex *accepted* the block — and green after.

**Checklist**

- [x] Regression test added (red before, green after)
- [x] Enforced on full-node import, L2 block producer, and zkVM guest paths


ElFantasma:
Addressed the review feedback in `6c9eeb0a0`:

**Codex — legacy `v` underflow (fixed).** Good catch. `tx.chain_id()` now runs for every tx on the import path, and `derive_legacy_chain_id` did `(v - 35) / 2` for any `v ∉ {27, 28}` — so a malformed legacy tx with `v < 35` (decoded from RLP unvalidated) underflowed: a panic in debug, a bogus wrapped chain id in release. Hardened it to return `None` for any `v < 35` (subsumes the pre-EIP-155 `27`/`28` case); valid EIP-155 values (`v ≥ 35`) are unchanged. A malformed-`v` tx now yields `chain_id() == None`, so it is skipped by the chain-id check and rejected later at signature recovery — the correct outcome. This also hardens the pre-existing mempool caller. Added a regression test (`legacy_chain_id_handles_out_of_range_v_without_underflow`) covering `v ∈ {0,1,26,27,28,34} → None` and `35/36 → 0`, `37 → 1`; it panics on the pre-fix code (debug) and is green after.

**Claude — `FeeTokenTransaction` not mentioned (clarified).** The exemption is `PrivilegedL2Transaction` only; the comment now states explicitly that the other L2-only type, `FeeTokenTransaction` (0x7d), is signature-recovered and therefore *stays* checked (carries the L2 chain id on L2; rejected as unsupported on L1 by #6752). No behavior change.

ilitteri (approved):

MegaRedHand (approved):

crates/common/types/transaction.rs
```diff
@@ -1530,11 +1530,11 @@ impl Transaction {
...
-    if v == 27 || v == 28 {
-        None
-    } else {
-        Some((v - 35) / 2)
-    }
+    // EIP-155 encodes the chain id in `v` as `chain_id * 2 + 35` (or `+ 36`). Values below
+    // 35 — including the pre-EIP-155 `27`/`28` — carry no chain id; `v` is decoded from RLP
+    // unvalidated, so returning None here (instead of computing `v - 35`) avoids an
+    // underflow panic on malformed input.
+    if v < 35 { None } else { Some((v - 35) / 2) }
```
MegaRedHand: This conflicts with #6787

iovoid (approved):

pablodeymo (approved):

PR #6818, opened 2026-06-09
fix(l1): reject non-canonical RLP encodings at decode

ElFantasma:
**Motivation**

Canonical RLP requires minimal encodings, and strict clients (geth) reject non-canonical input at decode (`ErrCanonSize` / `ErrCanonInt`). ethrex's decoder was lenient in three ways, all reachable on consensus-relevant decode paths (block bodies, transactions, headers — every `decode_field` and the outer list of every struct route through `decode_rlp_item`):

1. **Single-byte form** — a byte in `0x00..=0x7f` wrapped in a 1-byte string (`0x81 0x01` instead of `0x01`) was accepted.
2. **Non-minimal length** — long-form string/list headers (`0xB8+`/`0xF8+`) used for payloads shorter than 56 bytes were accepted.
3. *(Leading-zero length bytes were already rejected by `static_left_pad`.)*

A withdrawal whose `amount` is encoded as `0x81 0x01` decodes to the same value and re-encodes canonically, so `validate_block_body` then matches it against a canonical `withdrawals_root` and the block is accepted — while geth rejects the raw bytes at decode. Same wire bytes, opposite verdicts → a consensus/propagation divergence (reachable over devp2p, where the original byte encoding survives; `engine_newPayload` normalizes scalars through JSON and loses it).

This closes three EF findings sharing this root cause:
- `withdrawal-byte-normalization-root`
- `noncanonical-tx-rlp-root`
- `noncanonical-header-scalar-rlp`

**Description**

Adds canonical-RLP checks in `decode_rlp_item` (`crates/common/rlp/decode.rs`), the single decode primitive all field/scalar/list decoding flows through (`Decoder::new` for the outer list, `decode_field` → `decode_bytes` for each field):

- Short-string branch (`0x80..=0xB7`): reject a length-1 string whose byte is `< 0x80` (non-canonical single byte).
- Long-string branch (`0xB8..=0xBF`) and long-list branch (`0xF8..=0xFF`): reject payloads `< 56` bytes (must use the short form).
- Adds `RLP_SHORT_ITEM_LIMIT = 56`.

ethrex's own encoders are canonical, so internal round-trips are unaffected; only malformed external input is now rejected.

**Test**

`test/tests/common/withdrawal_rlp_tests.rs` builds a withdrawal whose `amount` uses the non-canonical `0x81 0x01` and asserts `Withdrawal::decode` rejects it (red before, green after); plus controls that the canonical `0x01` still decodes and that the already-handled `0x81 0x00` (zero) is rejected.

**Validation** (global decode change)

- In-repo decode-heavy suites all green: `ethrex-rlp`, `ethrex-trie`, `ethrex-common`, `ethrex-levm`, `ethrex-storage`, and `ethrex-test` `p2p::`/`blockchain::`/`common::`.
- **EF blockchain tests: 8726 passed, 0 failed.**
- **EF state tests: 100% across all groups, 0 failures.**

**Checklist**

- [x] Regression test added (red before, green after)
- [x] Validated against EF blockchain + state vectors (no decode regression)


ElFantasma:
Pushed fixes for the CI failure + review feedback (`4b3f58cc3`), and synced with `main`.

**CI failure — `rlp::decode_tests::test_decode_u256` (fixed; it was asserting the bug).** The failing test decoded `vec![RLP_NULL + 1, 0x01]` = `0x81 0x01` — the *non-canonical* encoding of the scalar `1` — and expected it to decode. That is exactly the leniency this PR removes, so the test was codifying the vulnerability. Updated it to the canonical `vec![0x01]` (a single byte `< 0x80` is its own encoding). Good signal that the fix is doing what it should. (My local sweep ran the `p2p`/`blockchain`/`common` modules but not `rlp::` — my miss; now covered.)

**greptile — long-form cases untested (added).** Added `rejects_long_form_string_for_short_payload` and `rejects_long_form_list_for_short_payload` to `decode_tests.rs`, exercising the `0xB8..=0xBF` / `0xF8..=0xFF` branches (long-form header for a `< 56`-byte payload → rejected), with canonical controls.

**Codex — `get_item_with_prefix()` still lenient (intentionally out of scope here).** Correct that `get_item_with_prefix()` (used by `Decoder::get_encoded_item[_ref]` and the trie decode path) keeps the old rules, so its acceptance surface differs from `decode_rlp_item`. For the three findings this PR targets it makes no difference — block/tx/header/withdrawal decoding flows through `decode_rlp_item` (`Decoder::new` for the outer list, `decode_field` → `decode_bytes` for each field), which these tests + the EF suites exercise. In the trie path, a non-canonical node encoding is already caught by hash verification (the node hash is keccak of its bytes), so it cannot be silently accepted. Making `get_item_with_prefix` canonicality-aware is a reasonable consistency hardening, but it touches the trie hot path, so I would rather do it as a separate, independently-benchmarked change than fold it in here. Happy to open a follow-up if the team wants the surfaces unified.

**Validation:** `ethrex-test` `rlp::` (33) + `withdrawal_rlp` (3) green post-merge; previously EF blockchain (8726) + state (100%) clean. The other reviews (Kimi, Claude, greptile summary) were positive.

ilitteri (approved):

iovoid (approved):

MegaRedHand (approved):
LGTM

PR #6821, opened 2026-06-09
ci(l1): regenerate block-execution benchmark fixture with valid logs_bloom

ilitteri:
**Motivation**

The "Benchmark Block execution" job (`pr_perf_blocks_exec.yaml`) started failing with `LogsBloomMismatch` after #6766, which added header `logs_bloom` validation to every import path. The job imports `fixtures/blockchain/l2-1k-erc20.rlp`, a fixture produced by an old ethrex that never populated the header bloom — so every header stored a zero `logs_bloom`, while its ERC20 blocks emit `Transfer` logs. The first such block now fails validation.

**Description**

Regenerates `l2-1k-erc20.rlp` so its headers carry the correct bloom. The bloom can't be patched in place (the header hash commits to it), so the chain is re-derived: each block is re-executed against a fresh `perf-ci.json` state, the aggregate bloom is recomputed from the executed receipts, written into the header, and every `parent_hash` is re-linked to the corrected predecessor. 611 of 1110 headers gained a non-zero bloom; the corrected chain imports cleanly through the normal validating path. The file size is unchanged (bloom and `parent_hash` are fixed-length), so the diff is a clean LFS pointer swap.

This was produced by a one-off script (not kept in-tree). The other ethrex-generated fixtures (`2000-blocks.rlp`, `l2-loadtest.rlp`) were checked the same way and need **no change** — their blocks are plain ETH transfers that emit no logs, so a zero bloom is already correct.

Verified with a binary that includes #6766: original fixture → `Invalid Block: Logs bloom does not match`; regenerated fixture → `Import completed blocks=1110`. No client code touched.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A (no `Store` changes)


tooling/fixture_rebloom/src/main.rs
```diff
@@ -0,0 +1,127 @@
+//! One-off maintenance tool that recomputes the header `logs_bloom` for an
```
MegaRedHand: Should this be moved to ethrex-tooling?

MegaRedHand (approved):

edg-l (approved):

iovoid (approved):

PR #6822, opened 2026-06-09
refactor(l1): improve db migration logging for operators

ilitteri:
**Motivation**

A UX review of the DB migration logging (follow-up to #6820, same methodology) found that the migration paths are close to silent exactly where operators most need feedback:

1. **The `tooling/migrations` CLI (libmdbx → RocksDB) drops almost all of its output.** `main()` never installs a tracing subscriber, so every `tracing` log from the migration, store, and blockchain crates is silently discarded; the tool's entire output is two bare `println!`s. The block-replay loop — re-executing potentially the whole chain — prints nothing, and the tool exits with no completion message, so success is indistinguishable from a hang until the process ends.
2. **The v2 → v3 schema migration logs nothing between start and finish.** It rewrites the entire `TRANSACTION_LOCATIONS` table (one entry per historical transaction), so a node starting up after a client upgrade looks frozen for the whole migration with only `Running migration v2 → v3` on screen.
3. **No upfront context or timing.** The operator isn't told how many migrations will run or that large databases take a while, and completions don't report elapsed time. Message wording also used three different styles (`Running migration v1 → v2` / `Migration v1→v2:` / `TRANSACTION_LOCATIONS migration:` — the last never naming its versions).
4. **`drop_obsolete_cfs` logs its own intentional cleanup at WARN** (plus a redundant success INFO per column family), making the planned final step of a migration look like a problem on every affected startup.

**Description**

Schema migrations (`crates/storage/migrations.rs`):
- Upfront summary before any migration runs: current schema version, target, number of pending steps, and a note that large databases may take a while. Per-step completion now reports elapsed time.
- v2 → v3 gains progress logging: entries processed and rate, throttled to at most one line every 10s.
- v1 → v2 progress switches from per-100k-entries to the same time-based throttle, and gains a rate.
- Unified wording: `Schema migration vX → vY: ...` everywhere; final summaries phrased in operator terms (receipt entries, transaction records) instead of column-family constants.

Obsolete CF cleanup (`crates/storage/backend/rocksdb.rs`):
- One INFO line on successful drop (was WARN pre-announcement + INFO success); failures remain WARN.

libmdbx → RocksDB tool (`tooling/migrations`):
- `main()` installs a `tracing_subscriber` (env-filter, default `info`) — same pattern the `bench_migration` bin already used. Requires adding the workspace `tracing` dependency to the package.
- `println!`s replaced with `tracing::info!`, and the messages now carry the block range and total count.
- Block-replay progress: migrated/total, percentage, current block number, and blocks/s, at most one line every 10s.
- Completion summary: total blocks migrated, elapsed time, and the new head block.

Example output for the tool after this change:

```
INFO Migrating 184231 blocks (1 to 184231) from libmdbx to RocksDB
INFO Migrated 12400/184231 blocks (6.7%), currently at block 12400 (1240 blocks/s)
...
INFO Migration complete: 184231 blocks migrated in 148.3s, head is now block 184231
```

Not changed (possible follow-ups): the tool's `expect`/`panic!`-based error handling (a failure still surfaces as a panic with a message rather than a clean error exit), and the dev-only `seed_migration_test`/`bench_migration` bins, which already produce reasonable println-based reports.

No behavior changes beyond logging. Verified with `cargo clippy --lib --bins -F debug,sync-test -- -D warnings` (workspace, clean), `cargo clippy -p migrations -- -D warnings` in the tooling workspace (clean), `cargo test -p ethrex-storage --lib migrations` (9/9 pass), and `cargo fmt` on both workspaces.

**Checklist**


iovoid (approved):

crates/storage/migrations.rs
```diff
@@ -141,8 +160,12 @@ fn migrate_1_to_2(backend: &dyn StorageBackend) -> Result<(), StoreError> {
             tx.put_batch(RECEIPTS_V2, std::mem::take(&mut batch))?;
             tx.commit()?;
             migrated += count;
-            if migrated.is_multiple_of(100_000) {
-                tracing::info!("Migration v1→v2: migrated {migrated} RECEIPTS entries so far");
+            if last_progress_log.elapsed() >= PROGRESS_LOG_INTERVAL {
+                let rate = migrated as f64 / start.elapsed().as_secs_f64();
```
MegaRedHand: We should check the denominator is not zero. I don't trust floats.

crates/storage/migrations.rs
```diff
@@ -206,6 +231,15 @@ fn migrate_2_to_3(backend: &dyn StorageBackend) -> Result<(), StoreError> {
         }
 
         total_old_entries += 1;
+        if total_old_entries.is_multiple_of(100_000)
+            && last_progress_log.elapsed() >= PROGRESS_LOG_INTERVAL
+        {
+            let rate = total_old_entries as f64 / start.elapsed().as_secs_f64();
```
MegaRedHand: Same for this and other divisions

MegaRedHand (approved):

crates/storage/migrations.rs
```diff
@@ -206,6 +231,15 @@ fn migrate_2_to_3(backend: &dyn StorageBackend) -> Result<(), StoreError> {
         }
 
         total_old_entries += 1;
+        if total_old_entries.is_multiple_of(100_000)
+            && last_progress_log.elapsed() >= PROGRESS_LOG_INTERVAL
```
ElFantasma: **Minor inconsistency with `migrate_1_to_2`'s progress gate.** Here the log fires on `is_multiple_of(100_000) AND elapsed >= 10s`; in `migrate_1_to_2` (line 163) it fires on just `elapsed >= 10s` (after the batch-commit gate at line 158). The two have different effective cadences:

- `v1→v2`: log on first batch commit ≥ 10s after the last log. Cadence floor is `BATCH_SIZE` entries (default ~10k).
- `v2→v3`: log when entry count crosses a 100k boundary AND 10s have passed. For a slow migration (<10k entries/s), the 100k boundary may take > 10s and the gate effectively becomes "every 100k entries". For a fast one, the 10s gate dominates.

If the goal was to avoid calling `Instant::now()` on every iteration (it's per-entry here, not per-batch), the `is_multiple_of` is doing useful work as a cheap pre-filter — worth a one-line comment to that effect ("per-iter check, modulo gate keeps the `elapsed()` syscall off the hot path") so the next reader doesn't normalize them.

If the modulo wasn't intentional perf, drop it and just use the time gate to match `v1→v2`. Non-blocking either way.

crates/storage/migrations.rs
```diff
@@ -54,10 +58,20 @@ pub fn run_pending_migrations(
     db_path: &Path,
     current_version: u64,
 ) -> Result<(), StoreError> {
+    let pending = STORE_SCHEMA_VERSION.saturating_sub(current_version);
+    if pending == 0 {
```
ElFantasma, line 78: nit: the new `saturating_sub` + early-return treats `current_version > STORE_SCHEMA_VERSION` (an older binary against a newer DB) as "no pending" and silently passes. That's the same pre-existing behavior as the old `for version in current_version..STORE_SCHEMA_VERSION` empty-range loop, so this isn't a regression — but the new banner you added two lines below ("Database schema is at v{current_version}, latest is v{STORE_SCHEMA_VERSION}; running {pending} migration(s)") is the ideal spot to flip the message and warn loudly:

```rust
if current_version > STORE_SCHEMA_VERSION {
    tracing::warn!(
        "Database schema is at v{current_version}, ahead of binary's v{STORE_SCHEMA_VERSION}; \
         continuing read-only is unsupported. Upgrade the binary."
    );
}
if pending == 0 {
    return Ok(());
}
```

Low risk of false positives (only fires on actual downgrade), and the operator otherwise gets zero signal that they ran the wrong binary. Non-blocking — file as follow-up if you'd rather keep this PR pure refactor.

ElFantasma (approved):
Two non-blocking inlines — one on a small inconsistency between the two intra-schema migrations' progress-gate patterns, one on a pre-existing edge case the new banner could surface more clearly.

pablodeymo (approved):

PR #6823, opened 2026-06-10
fix(l1): classify actor request timeouts as recoverable in sync error handling

ilitteri:
**Motivation**

Every snap-sync request to the PeerTable actor carries a 5-second timeout hardwired by the `#[protocol]` macro, with no per-call override. Both sync error classifiers treated any `ActorError` as non-recoverable, so a single timed-out `get_best_peer` under mailbox pressure terminated the node with `exit(2)` mid-sync. As sync gains more concurrent callers, the exposure only grows.

**Description**

- Classify `ActorError::RequestTimeout` as **recoverable** in `SyncError::is_recoverable`, so the sync cycle retries instead of exiting.
- Apply the same split to the delegated classifier in `PeerHandlerError::is_recoverable` — fixing only the first leaves the identical exit reachable through every `SyncError::PeerHandler` wrapper.
- Keep `ActorError::ActorStopped` fatal: a stopped actor means p2p is shutting down, and retrying would loop against dead actors (the `RequestPermit` drop path already documents lost messages during shutdown as benign).

Currently soak-testing on a `multisync-loop-auto` runner (triple-network snapsync with state-trie validation).

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — not needed: no storage changes.


iovoid (approved):

crates/networking/p2p/sync.rs
```diff
@@ -416,10 +416,14 @@ impl SyncError {
             | SyncError::RocksDBError(_)
             | SyncError::BytecodeFileError
             | SyncError::NoLatestCanonical
-            | SyncError::PeerTableError(_)
             | SyncError::MissingFullsyncBatch
             | SyncError::Snap(_)
             | SyncError::FileSystem(_) => false,
```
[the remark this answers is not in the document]
iovoid: I think these are unreachable.

crates/networking/p2p/peer_handler.rs
```diff
@@ -740,9 +741,12 @@ impl PeerHandlerError {
             | PeerHandlerError::ReceiveMessageFromPeerTimeout(_)
             | PeerHandlerError::InvalidHeaders
             | PeerHandlerError::NoResponseFromPeer => true,
-            PeerHandlerError::StorageFull
-            | PeerHandlerError::PeerTableError(_)
-            | PeerHandlerError::Snap(_) => false,
+            // A timed-out actor request is transient (mailbox pressure or a
+            // slow handler — the protocol macro hardwires a 5s timeout); a
+            // stopped actor means p2p is shutting down and must stay fatal.
+            PeerHandlerError::PeerTableError(ActorError::RequestTimeout) => true,
```
ElFantasma, line 748: **Retry path is unbounded.** Returning `true` here sends the error back through `sync.rs:163`'s classifier (which delegates to this fn via `SyncError::PeerHandler(e).is_recoverable()`); on `true` the manager in `sync_manager.rs:220-225` re-enters `start_sync` with no backoff and no consecutive-failure counter.

For the bug this PR fixes (occasional 5s timeout under mailbox pressure → ethrex shouldn't `process::exit(2)`), that's right and is the whole point. But for a structural cause (peer-table actor wedged on a deadlocked request, persistent task starvation), the same `RequestTimeout` keeps firing every ~5s and the loop turns into a tight spin: `start_sync` → timeout → log+return → `start_sync` again. Two non-blocking mitigations worth a follow-up:

1. Bound consecutive recoverable `RequestTimeout`s (e.g., 5 in a row without an intervening successful cycle → escalate to fatal).
2. Add a small backoff (linear or 2× capped) in the `sync_manager.rs:220` loop on recoverable errors.

Neither blocks landing this — the fix is correctly scoped to the immediate symptom — but worth filing as a follow-up before someone has to debug a 100% CPU loop in the wild.

crates/networking/p2p/peer_handler.rs
```diff
@@ -740,9 +741,12 @@ impl PeerHandlerError {
             | PeerHandlerError::ReceiveMessageFromPeerTimeout(_)
             | PeerHandlerError::InvalidHeaders
             | PeerHandlerError::NoResponseFromPeer => true,
-            PeerHandlerError::StorageFull
-            | PeerHandlerError::PeerTableError(_)
-            | PeerHandlerError::Snap(_) => false,
+            // A timed-out actor request is transient (mailbox pressure or a
+            // slow handler — the protocol macro hardwires a 5s timeout); a
```
ElFantasma: nit: "the protocol macro hardwires a 5s timeout" — per spawned-concurrency 0.5.0's docstring on `ActorError::RequestTimeout`, that's the *default* timeout used by `request()`, not hardwired. `request_with_timeout(duration)` overrides it. Worth tightening to "5s default" so a future maintainer (or someone bumping spawned-concurrency) doesn't assume the value is invariant.

Same comment block is duplicated at `sync.rs:421-423` — if you re-word here, mirror it there too.

ilitteri: Reworded in 6aa698e (both files).

crates/networking/p2p/peer_handler.rs
ilitteri: Agreed — filed as #6853.

ElFantasma (approved):

edg-l (approved):

PR #6825, opened 2026-06-10
fix(l1): prevent snap-sync restart livelock from a dangling header checkpoint

ilitteri:
**Motivation**

`process_incoming_headers` wrote the header-download checkpoint and the headers themselves in two separate store commits, checkpoint first. A crash between the two leaves a checkpoint pointing at a header that was never stored; on restart the snap cycle resolves the checkpoint to no block number and fails with a recoverable `BlockNumber` error, so the sync manager retries forever against the same unchanged state — the node livelocks, burning CPU and never syncing.

**Description**

Fixes both directions:

- **Prevention** — reorder the writes so headers commit before the checkpoint that references them (`HEADERS` + `BLOCK_NUMBERS` already commit in one transaction). A crash after the headers only re-downloads a suffix.
- **Cure** — validate the checkpoint on resume: if its header is missing (databases that crashed inside the old write window still carry dangling checkpoints), warn and restart header download from the canonical head instead of erroring; the first stored batch overwrites the checkpoint.

Adds regression tests: a dangling checkpoint must fall back to the canonical head (fails without this change), and a valid checkpoint must still be resumed.

Currently soak-testing on a `multisync-loop-auto` runner (triple-network snapsync with state-trie validation).

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — not needed: write-ordering and resume-logic change only, no table/schema changes.

### DIFF
--- a/crates/networking/p2p/sync/snap_sync.rs
+++ b/crates/networking/p2p/sync/snap_sync.rs
@@ -68,13 +68,24 @@ impl SnapBlockSyncState {
     /// Obtain the current head from where to start or resume block sync
     pub async fn get_current_head(&self) -> Result<H256, SyncError> {
         if let Some(head) = self.store.get_header_download_checkpoint().await? {
-            Ok(head)
-        } else {
-            self.store
-                .get_latest_canonical_block_hash()
-                .await?
-                .ok_or(SyncError::NoLatestCanonical)
+            // A checkpoint pointing at a header we never stored (a crash
+            // between the checkpoint write and the header write) would make
+            // every resume fail with the same recoverable error, retrying
+            // forever. Restart header download from the canonical head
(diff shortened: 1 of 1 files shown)


ElFantasma (approved):
LGTM

edg-l (approved):
maybe tests on test/ folder but can be followup

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -1305,3 +1318,51 @@ async fn insert_storages(
 
     Ok(())
 }
+
+#[cfg(test)]
```
iovoid, line 1322: Should be under test/

iovoid (approved):

PR #6828, opened 2026-06-10
perf(l1): route native P256VERIFY through aws-lc-rs

edg-l:
P256VERIFY (RIP-7212 / EIP-7951, secp256r1) was a ~4.5x perf outlier vs the reference client. Root cause: the RustCrypto `p256` verify path does two full constant-time scalar muls with no Shamir trick and no precomputed basepoint table. All inputs are public on-chain data, so constant-time execution is unnecessary overhead.

Override `secp256r1_verify` on `NativeCrypto` to use the assembly-backed `aws-lc-rs` (AWS-LC, p256-nistz asm) backend. Local microbenchmark: 41.8 µs vs 242 µs (~5.8x per call). End-to-end on the benchmark suite, ethrex's P256VERIFY per-test time drops from ~1414 ms to ~302 ms (~4.7x). `aws-lc-rs` was already in the dependency tree transitively via rustls; pinned to 1.17 for the `verify_digest` API. The pure-Rust `p256` trait default is kept for zkVM guests (SP1/RISC0/OpenVM), behind a default-on, std-only `aws-lc-rs` feature. Explicit r/s in-(0,n) range pre-checks are retained before the backend call.

Changed files: `crates/common/crypto/{Cargo.toml, lib.rs, native.rs, p256_awslc.rs}`, `test/{Cargo.toml, tests/crypto/p256_tests.rs}`, `Cargo.lock`, `tooling/Cargo.lock`.

Mirrors the blst precedent for native BLS12-381 precompiles (#6792).

Differential tests in `test/tests/crypto/p256_tests.rs` verify aws-lc-rs and the `p256` default agree across valid signatures, tampered message/signature/key, and adversarial inputs (r/s = 0, r/s = n, r/s = n-1, point at infinity, coordinate at field boundary). All four feature combinations (none / blst / aws-lc-rs / both) compile; zkVM `--no-default-features` keeps the portable default.

**Caveats:** The EF state-test suite carrying the official EIP-7951 vectors has not been run locally yet; CI should gate on that before merge. CHANGELOG entry added.


test/tests/crypto/p256_tests.rs
```diff
@@ -0,0 +1,163 @@
+//! Differential tests for the P256VERIFY (secp256r1) precompile backend.
+//!
+//! `NativeCrypto` overrides `secp256r1_verify` with the native aws-lc-rs path
+//! (when the `aws-lc-rs` feature is on, which it is by default). A `Crypto` impl
+//! with no overrides resolves to the portable pure-Rust `p256` trait default.
```
iovoid: It's easy for the test to silently check nothing if the test is built with aws-lc-rs turned off.

iovoid (approved):

crates/common/crypto/p256_awslc.rs
```diff
@@ -0,0 +1,67 @@
...
+    // SEC1 uncompressed public key: 0x04 || x || y. AWS-LC rejects coordinates
+    // >= p and off-curve points during parsing; (0, 0) is off-curve and the
+    // point at infinity is not encodable in this fixed 65-byte form.
+    let mut sec1 = [0u8; 65];
+    sec1[0] = 0x04;
+    sec1[1..].copy_from_slice(pk);
+
+    // The message is already hashed; import it as a SHA-256-sized digest and
+    // verify directly rather than re-hashing.
+    let Ok(digest) = Digest::import_less_safe(msg, &SHA256) else {
```
ElFantasma, line 54: `Digest::import_less_safe` is named to flag the caller-must-guarantee: the bytes you import must be a SHA-256-shaped digest, not raw data. That's exactly the EIP-7951 contract (`msg` is *defined* as the 32-byte pre-hashed message and the verifier uses `ECDSA_P256_SHA256_FIXED` which expects a SHA-256 digest internally), so it's sound here — but the `less_safe` suffix will look suspicious to anyone auditing the precompile path. Worth one inline-comment line: e.g. *"Safe per EIP-7951 — the precompile's `msg` parameter is defined as a 32-byte SHA-256 prehash, not raw data, so importing it as a `SHA256` digest preserves the verifier's hash-function contract."*

Non-blocking.

test/tests/crypto/p256_tests.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use ethrex_crypto::{Crypto, NATIVE_P256_BACKEND, NativeCrypto};
+use p256::ecdsa::{SigningKey, signature::hazmat::PrehashSigner};
+
+/// Guard: without the native aws-lc-rs backend, `NativeCrypto` resolves to the
+/// same pure-Rust trait default as `PureRust`, so every differential assertion
+/// below compares a backend against itself and passes vacuously. Fail loudly
+/// rather than let the suite silently verify nothing.
+#[test]
+fn native_backend_is_active() {
```
ElFantasma, line 22: The differential suite is good (vacuous-comparison guard, random-valid + tampered + adversarial cases), but it has a blind spot: if both `aws-lc-rs` and the portable `p256` trait default ever share a bug (e.g., both accept a particular off-curve point, both fail to reject a malformed scalar in the same way), all three tests pass green while ethrex is wrong.

Worth a follow-up PR to seed this file with the EIP-7951 reference vectors (the spec's test vectors at the end of the EIP, or the EELS `tests/p256verify` fixtures): those give an external arbiter so a same-bug-in-both-backends scenario fails loudly. The current suite is great for catching divergence; the spec vectors are great for catching shared bugs. Both pay rent.

Non-blocking for landing this — the perf win is real and the differential coverage is already a step up — but worth filing as the next step.

ElFantasma (approved):
Two non-blocking comments, though

test/tests/crypto/p256_tests.rs
edg-l: the eels still run as part of CI in other jobs

ilitteri (approved):

crates/common/crypto/p256_awslc.rs
edg-l: Added an inline comment spelling out the soundness argument (msg is a 32-byte SHA-256 prehash per EIP-7951, so importing it as a SHA256 digest preserves the verifier's hash-function contract). Fixed in 74299d675.

test/tests/crypto/p256_tests.rs
edg-l: Agreed, filing the EIP-7951 reference-vector seeding as a follow-up. As noted above, the EELS p256verify fixtures already run in other CI jobs, so the spec-vector arbiter is covered there for now; the in-crate vectors are the nice-to-have follow-up.

PR #6829, opened 2026-06-10
perf(l1): thread Arc<BlockAccessList> to avoid per-block BAL deep clones

edg-l:
Before this change, `execute_block_parallel` in `crates/vm/backends/levm/mod.rs` called `Arc::new(bal.clone())` and `Arc::new(validation_index.clone())` on the critical path -- a full deep copy of the BAL (potentially multiple MB of account/slot/code data) before a single transaction executed, serially.

Fix: thread `Option<Arc<BlockAccessList>>` and `Arc<BalAddressIndex>` through the entire pipeline: `add_block_pipeline{,_bal,_with_witness,_inner}`, `Blockchain::execute_block_pipeline`, `Evm::execute_block_pipeline`, `LEVM::execute_block_pipeline`, and `execute_block_parallel`. The owned BAL is wrapped in an `Arc` once at the call boundary (`bal.map(Arc::new)`). The validation index is built directly into an `Arc`. All downstream per-thread (warmer/executor) and per-tx (`LazyBalCursor`) uses become pointer clones. The two deep clones become moves.

7 files touched: `cmd/ethrex/cli.rs`, `crates/blockchain/blockchain.rs`, `crates/networking/p2p/sync/full.rs`, `crates/networking/rpc/rpc.rs`, `crates/vm/backends/levm/mod.rs`, `crates/vm/backends/mod.rs`, `tooling/ef_tests/blockchain/test_runner.rs`. Zero behavior change.

`make test` passes (57 binaries, 0 failures). `cargo check --workspace --all-targets` and `cargo clippy` clean. Full EF state/blockchain suites not run locally; they should gate in CI.


ilitteri (approved):

crates/vm/backends/levm/mod.rs
```diff
@@ -1093,8 +1092,8 @@ impl LEVM {
         });
 
         // Small capacity hint — per-tx DBs materialize only touched accounts via lazy_bal cursor.
-        let arc_bal = Arc::new(bal.clone());
-        let arc_idx = Arc::new(validation_index.clone());
+        let arc_bal = bal;
```
ElFantasma, line 1103: nit: the `// Small capacity hint — per-tx DBs materialize only touched accounts via lazy_bal cursor.` comment is now orphaned. In the pre-refactor code these two lines were `let arc_bal = Arc::new(bal.clone()); let arc_idx = Arc::new(validation_index.clone());` — the "small capacity hint" referred to those allocations or the followup HashMap sizing. With the new move-based bindings (`let arc_bal = bal;`) there's no allocation or capacity hint happening here, and the line below is just the start of the `type TxExecResult = ...` declaration. Either drop the comment or move it down to wherever the per-tx-DB capacity is actually being chosen.

iovoid (approved):

crates/vm/backends/levm/mod.rs
edg-l: Good catch. Moved the capacity-hint comment to the actual per-tx DB construction site (the `new_with_shared_base_and_capacity(..., 32)` call) and gave the `arc_bal`/`arc_idx` bindings a comment that fits their new move-based role. Fixed in 5be3dfd7a.

ElFantasma (approved):

PR #6833, opened 2026-06-10
fix(l1): clamp snap GetAccountRange/GetStorageRanges response_bytes to server limit

ElFantasma:
**Motivation**

The snap/1 server handlers honor the requester's `responseBytes`/`bytes` budget verbatim, with no server-side cap. Per the [snap spec](https://github.com/ethereum/devp2p/blob/master/caps/snap.md#getaccountrange-0x00) this is a **soft limit** the server must bound to its own maximum (go-ethereum applies a 2 MiB `softResponseLimit` to **every** `Get*` handler). ethrex has the constant — `MAX_RESPONSE_BYTES = 512 KiB` — but applied it **only when issuing** requests (`snap/client.rs`), never when serving them.

So a single small, unauthenticated snap request can drive unbounded server-side allocation:
- **`GetAccountRange`** / **`GetStorageRanges`** with `limit_hash = 0xff…ff` and `response_bytes = u64::MAX` walk the **entire** state / storage trie into a `Vec` (then copy it for the proof) before encoding any reply — the loop's only termination is the two attacker-controlled maxima. Allocation scales with state size (~190 B/account; >10⁶× request→memory amplification), unbounded by request size. Each runs on its own `spawn_blocking` worker and completes even after the peer disconnects, so the per-peer serve-rate limiter doesn't help (K peers → K concurrent walks). **Demonstrated to OOM-kill** an 8 GiB node on a 4M-account state with 32 peers in ~10 s (reported by EF).
- **`GetByteCodes`** / **`GetTrieNodes`** take an explicit peer-supplied list (`hashes` / `paths`), but with `bytes = u64::MAX` the soft limit never fires, and the list snappy-compresses to almost nothing when entries repeat — so a tiny request naming one large bytecode (or node path) thousands of times expands to a multi-GB response.

**Description**

Clamp the requester-supplied budget to the server maximum at the top of **all four** `Get*` handlers (`crates/networking/p2p/snap/server.rs`), using the clamped local in the loop break conditions:
```rust
let response_bytes = request.response_bytes.min(MAX_RESPONSE_BYTES); // account / storage ranges
let bytes          = request.bytes.min(MAX_RESPONSE_BYTES);          // byte codes
let mut remaining_bytes = request.bytes.min(MAX_RESPONSE_BYTES);     // trie nodes
```
`limit_hash` is intentionally untouched — a full-range request is legitimate and is now bounded by the byte budget; returning fewer bytes than requested is valid for a soft limit, so there's no interop regression (and ethrex's own client already requests exactly `MAX_RESPONSE_BYTES`). This also bounds the related `low/snap-account-range-proof-amplification` finding (the proof is now built over a clamped set).

**Test**

`test/tests/p2p/snap_server_tests.rs`:
- `account_range_clamps_response_bytes` — 20,000-account state, `response_bytes = u64::MAX` → bounded (count < N, byte sum ≤ clamp + one account). Red before (all 20,000), green after.
- `storage_ranges_clamps_response_bytes` — 12,000-slot storage trie. Red before (all 12,000), green after. (Both have a `> 1000` lower-bound guard so a broken fixture can't false-pass.)
- `byte_codes_clamps_response_bytes` — 30 × 40 KiB codes, `bytes = u64::MAX` → bounded set. Red before, green after.
- `GetTrieNodes` gets the identical clamp but no dedicated test: exercising it needs a committed state trie readable through `get_trie_nodes`' `open_state_trie` path, which the direct-write test fixtures don't populate (noted in the test file). The clamp is mechanically identical to the three covered handlers.
- All 12 existing hive `AccountRange` tests are unchanged (their budgets are below the clamp → no behavior change).

**Checklist**

- [x] Regression tests added (red before, green after — account/storage/bytecodes)
- [x] All four snap `Get*` server handlers clamped; existing hive tests unaffected

### DIFF
--- a/crates/networking/p2p/snap/server.rs
+++ b/crates/networking/p2p/snap/server.rs
@@ -118,6 +118,10 @@ pub async fn process_byte_codes_request(
     store: Store,
 ) -> Result<ByteCodes, SnapError> {
     tokio::task::spawn_blocking(move || {
+        // Clamp the peer-supplied budget: `hashes` can be a large (or snappy-inflated,
+        // repeated) list, so an unbounded `bytes` would let one request buffer many large
+        // bytecodes into memory.
+        let bytes = request.bytes.min(MAX_RESPONSE_BYTES);
         let mut codes = vec![];
         let mut bytes_used = 0;
         for code_hash in request.hashes {
@@ -125,7 +129,7 @@ pub async fn process_byte_codes_request(
                 bytes_used += code.len() as u64;
                 codes.push(code);
             }
-            if bytes_used >= request.bytes {
+            if bytes_used >= bytes {
                 break;
             }
         }
@@ -144,7 +148,9 @@ pub async fn process_trie_nodes_request(
 ) -> Result<TrieNodes, SnapError> {
     tokio::task::spawn_blocking(move || {
         let mut nodes = vec![];
-        let mut remaining_bytes = request.bytes;
+        // Clamp the peer-supplied budget so a single request cannot pull an unbounded
+        // number of trie nodes into memory.
+        let mut remaining_bytes = request.bytes.min(MAX_RESPONSE_BYTES);
         for paths in request.paths {
             if paths.is_empty() {
                 return Err(SnapError::BadRequest(
--- a/crates/networking/p2p/snap/server.rs
+++ b/crates/networking/p2p/snap/server.rs
@@ -22,14 +22,14 @@ pub async fn process_account_range_request(
         // `responseBytes` is a soft limit chosen by the requester; clamp it to our server
         // maximum so a peer cannot request an unbounded walk of the whole state trie
         // (a single such request buffers every account into memory before replying — OOM).
-        let response_bytes = request.response_bytes.min(MAX_RESPONSE_BYTES);
+        let byte_budget = request.response_bytes.min(MAX_RESPONSE_BYTES);
         let mut accounts = vec![];
         let mut bytes_used = 0;
         for (hash, account) in store.iter_accounts_from(request.root_hash, request.starting_hash)? {
             debug_assert!(hash >= request.starting_hash);
             bytes_used += 32 + AccountStateSlimCodec(account).length() as u64;
             accounts.push(AccountRangeUnit { hash, account });
-            if hash >= request.limit_hash || bytes_used >= response_bytes {
+            if hash >= request.limit_hash || bytes_used >= byte_budget {
                 break;
             }
         }
@@ -55,7 +55,7 @@ pub async fn process_storage_ranges_request(
     tokio::task::spawn_blocking(move || {
         // Same soft-limit clamp as the account-range handler: bound the attacker-supplied
         // budget so a peer cannot walk a contract's entire storage trie into memory.
-        let response_bytes = request.response_bytes.min(MAX_RESPONSE_BYTES);
+        let byte_budget = request.response_bytes.min(MAX_RESPONSE_BYTES);
         let mut slots = vec![];
         let mut proof = vec![];
         let mut bytes_used = 0;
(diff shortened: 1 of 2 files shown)


crates/networking/p2p/snap/server.rs
```diff
@@ -48,6 +53,9 @@ pub async fn process_storage_ranges_request(
     store: Store,
 ) -> Result<StorageRanges, SnapError> {
     tokio::task::spawn_blocking(move || {
+        // Same soft-limit clamp as the account-range handler: bound the attacker-supplied
+        // budget so a peer cannot walk a contract's entire storage trie into memory.
+        let response_bytes = request.response_bytes.min(MAX_RESPONSE_BYTES);
```
iovoid: Maybe call it max_response_bytes?

iovoid (changes requested):
GetByteCodes/GetTrieNodes have the same issue, they should also be fixed

test/tests/p2p/snap_server_tests.rs
```diff
@@ -823,3 +829,129 @@ fn setup_initial_state() -> Result<(Store, H256), SnapError> {
...
+    );
+
+    // The returned set must fit the server budget. The loop pushes one account before
+    // checking the bound, so allow a single account of overshoot.
+    let bytes_used: u64 = res
+        .accounts
+        .iter()
+        .map(|unit| 32 + AccountStateSlimCodec(unit.account.clone()).length() as u64)
+        .sum();
+    let max_account_bytes =
```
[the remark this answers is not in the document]
ElFantasma: Fixed — replaced the `res.accounts[0]` index with `.first()`, and added a `> 1000` lower-bound assertion to both range tests so an empty/undersized response fails loudly (and can't silently false-pass the upper-bound check) instead of panicking on the index.

ElFantasma:
Addressed the review feedback (`294576d0c`), and fixed CI + synced `main`.

**Codex — incomplete hardening (extended to all four `Get*` handlers).** Good catch, you're right that `GetByteCodes`/`GetTrieNodes` trusting the peer-supplied `bytes` is the same class: the list is peer-controlled and snappy-compresses to ~nothing when entries repeat, so an unclamped budget lets a tiny request pull GBs. Applied the identical `request.bytes.min(MAX_RESPONSE_BYTES)` clamp to both, matching geth's per-handler `softResponseLimit`. Updated the PR description to cover all four handlers and corrected the earlier "bounded by the request list" framing.

Added a `byte_codes_clamps_response_bytes` regression test (30 × 40 KiB codes, `bytes = u64::MAX` → bounded; red before, green after). `GetTrieNodes` gets the same clamp but no dedicated test — exercising it needs a committed state trie readable through `get_trie_nodes`' `open_state_trie` path, which the direct-write test fixtures don't populate (noted in the test file); the clamp is mechanically identical to the three handlers that are covered. Happy to add a trie-nodes test in a follow-up if someone has a committed-trie fixture helper to build on.

**greptile — `res.accounts[0]` panic** (inline): fixed (`.first()` + a `> 1000` lower-bound guard on both range tests).

**CI:** corrected the PR title scope (`p2p` → `l1`), ran `cargo fmt` (the new test code wasn't formatted), and merged current `main`.

Kimi / Claude / greptile-summary were all positive — thanks for the reviews.

crates/networking/p2p/snap/server.rs
ElFantasma: Renamed the clamped local to `byte_budget`, applied consistently across all four handlers (`7e043fe5e`). Went with `byte_budget` over `max_response_bytes` to avoid visual collision with the `MAX_RESPONSE_BYTES` constant, and because the value can be *below* the cap (when the requester asks for less) — it's the effective per-response budget, not a fixed max. Happy to adjust if you'd prefer a different name.

ElFantasma:
@iovoid the `GetByteCodes`/`GetTrieNodes` point is addressed — your review came in just before I pushed the extension. `294576d0c` applies the same `request.bytes.min(MAX_RESPONSE_BYTES)` clamp to both handlers, so all four `Get*` server handlers are now bounded (also flagged by Codex). Added a `byte_codes_clamps_response_bytes` regression test; `GetTrieNodes` gets the identical clamp but no dedicated test (exercising it needs a committed state trie readable through `get_trie_nodes`' `open_state_trie` path, which the direct-write test fixtures don't populate — noted in the test file). The naming nit is also handled (`7e043fe5e`). Could you re-review when you have a moment?

iovoid (approved):

ilitteri (approved):

edg-l (approved):

PR #6834, opened 2026-06-10
fix(l1): reject l2 transaction types in the L1 guest program

iovoid:
**Motivation**

In #6752 we added validations to ensure L2-only transaction types aren't accepted in L1 mode, but those validations aren't used in stateless mode.

**Description**

Adds the validation to the stateless execution path.


iovoid:
We will go for a more general approach. Moved to issue #6837

Issue #6749, opened 2026-05-29
perf(l1,rpc): establish cross-client eth_* benchmark baseline + publish report


ElFantasma:
Part of #6745. Depends on #6748 (harness selected — `flood`, adopted).

## Goal
Produce the **baseline measurement** of ethrex's `eth_*` RPC performance under load on real mainnet state, compared against the major reference clients, and publish it.

## Methodology — same server, sequential runs
To eliminate the cross-node hardware variance that has confounded earlier A/Bs, all clients are benchmarked **one at a time on the same machine (mainnet-10), against the same canonical chain tip**:

1. Deploy + sync **ethrex** on mainnet-10, run the full benchmark suite, save results.
2. Sequentially do the same on the **same box** for **geth**, **reth**, and (maybe) **nethermind** — one client live at a time.
3. Compare apples-to-apples: identical hardware, identical workload, identical RPS curve.

> ⚠️ mainnet-10 is a shared node. Each stop / redeploy / client swap needs explicit confirmation before it happens (standing rule). Confirm disk headroom for the additional client DBs before committing.

## Scope — top-10 methods (93% of mainnet traffic)
`eth_call` · `eth_getBlockByNumber` · `eth_getBalance` · `eth_chainId` · `eth_getTransactionReceipt` · `eth_blockNumber` · `eth_getLogs` · `eth_getTransactionByHash` · `eth_getTransactionCount` · `eth_getCode`
(flood ships templates for 8/10; `eth_chainId` + `eth_blockNumber` are cached on ethrex — add templates or skip.)

## Tasks
- [ ] Deploy + sync ethrex on mainnet-10; confirm it's at a stable tip.
- [ ] Decide **workload source** (open): `ctc` pointed at an archive RPC for realistic address/block sampling, **vs.** a replayed sample of real mainnet RPC traffic.
- [ ] Run each method at `10 / 100 / 500 / 1000` RPS against ethrex; capture per-endpoint, per-rate **p50 / p90 / p99 latency + error-rate**.
- [ ] Sync + benchmark **geth** on the same box (same tip, same workload).
- [ ] Sync + benchmark **reth** on the same box.
- [ ] (Optional) Sync + benchmark **nethermind** on the same box.
- [ ] Publish the comparison report in `docs/perf_reports/`.

## Open questions
- **Workload realism:** ctc-synthetic (realistic sampling from archive) vs. replayed real traffic — do we have access to a traffic sample?
[...]

PR #6835, opened 2026-06-10
perf(l1): add bloombits log index for eth_getLogs

ElFantasma:
> Stacked on #6813 (base branch `perf/getlogs-index`). Review/merge that first; this will be retargeted to `main` once it lands.

**Motivation**

#6813 made `eth_getLogs` skip non-matching blocks via the per-block header bloom, but the query still walks every header in the range — O(blocks-in-range). The benchmark on that PR confirmed the ceiling: ~1.2 ms/block of header reads, so 1,000+ block ranges still saturate under load and stay far from indexed clients (geth/reth/nethermind).

This closes that gap with a real log index.

**Description**

Adds a geth-style **bloombits** log index. Blocks are grouped into sections of 4096; for each of the 2048 bloom bit positions we store one transposed bit-vector across the section. A query ANDs the three bit-vectors an address/topic maps to (OR within an address set / topic position, AND across positions) to get a candidate-block bitmap, turning `eth_getLogs` from O(blocks-in-range) into ~O(matching blocks).

Why bloombits over an inverted index: a snap-synced node retains **all historical headers** (hence all `logs_bloom`) but **not historical receipts** (`snap_sync.rs` skips pre-pivot bodies/receipts). An inverted index built from receipt logs could only ever cover post-pivot history; bloombits is built purely from header blooms, so it indexes the full chain on the nodes people actually run.

Implementation (one commit per stage):
1. **Core module** (`crates/storage/bloombits.rs`) — bit-index derivation (mirrors `ethbloom`), section transpose, and the candidate-bitmap query. Pure logic, no storage.
2. **Storage** — `LOG_BLOOM_BITS` column family (section-major key `section || bit`; all-zero rows omitted) plus `store_bloombits_section` / `get_candidate_blocks`. Coverage tracked in `MISC_VALUES`.
3. **Background indexer** (`cmd/ethrex`) — finalizes sections only once they're buried ≥256 blocks (well past the 128 reorg-depth limit), so stored sections are immutable and **no reorg invalidation is needed**. First run backfills the whole chain; steady state indexes forward. Runs on a blocking thread, off the import hot path → no `store_ms` impact.
4. **Query wiring** — `fetch_logs_with_filter` visits index candidates instead of the full range; the still-unindexed recent tail is returned in full and scanned. Falls back to the #6813 header-bloom scan when the index can't help (unconstrained query, or range above what's indexed). Every visited block is still exact-filtered, so false positives are harmless and results are identical to a full scan.

No new schema version: the column family auto-creates on open, and an unindexed DB simply uses the scan fallback until the background indexer catches up.

Tested: bloombits transpose/query incl. a no-false-negatives brute-force check; storage candidate-narrowing, unindexed-tail, and confirmation-depth/backfill gating; and an end-to-end `eth_getLogs` test that populates the index and asserts the candidate path returns exactly the matching logs.

Benchmarking against #6813 and geth with the #6749 harness is the remaining step before marking ready.

Part of #6785.

**Checklist**

- [x] New `LOG_BLOOM_BITS` column family auto-creates on DB open; no migration and no `STORE_SCHEMA_VERSION` bump required (index is append-only, absence = scan fallback).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## Perf
 
+### 2026-06-10
+
+- Add a geth-style bloombits log index so `eth_getLogs` queries only candidate blocks instead of scanning the whole range; built from header blooms by a background indexer (full historical coverage on snap-synced nodes), with a scan fallback [#6835](https://github.com/lambdaclass/ethrex/pull/6835)
+
 ### 2026-06-09
 
 - Skip non-matching blocks in `eth_getLogs` using the per-block header bloom, avoiding body/receipt loads for blocks that provably cannot match [#6813](https://github.com/lambdaclass/ethrex/pull/6813)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## Perf
 
+### 2026-06-09
+
+- Skip non-matching blocks in `eth_getLogs` using the per-block header bloom, avoiding body/receipt loads for blocks that provably cannot match [#6813](https://github.com/lambdaclass/ethrex/pull/6813)
+
 ### 2026-06-03
 
 - Short-circuit the `KECCAK256` opcode on zero-length input by returning the precomputed `keccak256("")` constant, skipping the permutation [#6775](https://github.com/lambdaclass/ethrex/pull/6775)
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -38,7 +38,7 @@ use std::{
     net::{IpAddr, Ipv4Addr, SocketAddr},
     path::{Path, PathBuf},
     sync::Arc,
-    time::{SystemTime, UNIX_EPOCH},
+    time::{Duration, SystemTime, UNIX_EPOCH},
 };
 use tokio_util::{sync::CancellationToken, task::TaskTracker};
 #[cfg(not(feature = "l2"))]
@@ -685,6 +685,8 @@ pub async fn init_l1(
     )
     .await;
 
+    tracker.spawn(run_bloombits_indexer(store.clone(), cancel_token.clone()));
+
     if opts.metrics_enabled {
         init_metrics(&opts, &network, tracker.clone());
     }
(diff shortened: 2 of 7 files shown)


ElFantasma:
### Parking this as draft — measured no gain yet, and here's why / the plan

Benched on mainnet-10 (reused the synced datadir; the indexer backfilled all 6,175 sections from header blooms in ~10 min, no resync). On LUSD `Transfer` queries the bloombits index measured **identical to the header-bloom prefilter** (#6813):

| range | #6813 (prefilter) | this PR (bloombits) |
|---|---|---|
| 100-block | ~415 ms | ~347 ms |
| 1,000-block | ~3.5 s | ~3.8 s |
| 10,000-block | ~32 s | ~33 s |

A deep-indexed range was also no faster than an unindexed recent-tail range, confirming the index path performs like the scan path here.

**Why (analysis, not yet fully measured):** `getLogs` has two stages — (1) *find* candidate blocks (filtering), (2) *fetch* each survivor's receipts and extract logs. Both the header-bloom prefilter and bloombits only optimize **stage 1**, and they consume the **same** 2048-bit header bloom — which is **saturated** on mainnet for a ubiquitous signature like `Transfer`, so neither narrows much. Meanwhile the **leading suspect for the dominant cost is stage 2** — ethrex loads each survivor's full body and fetches receipts one transaction at a time. The index doesn't touch stage 2, which is likely why it's masked.

**Important:** that stage-2 hypothesis is **not yet confirmed by profiling** — confirming it is the next step before we act on it.

**Plan:**
1. Confirm the real bottleneck (profile the `getLogs` path: filtering vs body-load vs receipt-fetch).
2. If confirmed, fix stage 2 (e.g. bulk per-block receipt reads + skip the full-body load).
3. **Re-bench this PR rebased on that fix**, across a matrix of **{sparse, saturated} × {small, large range}** — bloombits should help most on *sparse* queries over *large* ranges (where most blocks are skipped and avoiding per-block header reads is the whole point); the saturated/LUSD case may still show little gain and might warrant a sharper (non-bloom) index instead.

Keeping this open as a draft so the work is preserved and can be measured properly once stage 2 isn't dominating.

ElFantasma:
Closing as superseded.

Bloombits is the header bloom transposed into bit-major layout — same bits, same saturation. In testing it produced the same candidate set as the header-bloom prefilter already merged in #6813 (no speedup), and for ubiquitous signatures like ERC-20 `Transfer` it's ~100% false positives, which is the common query.

The address case is now handled exactly by the inverted address index (#6884) — built from receipts (no false positives), benchmarked 50–214× on sparse-address queries. The only thing bloombits could add on top is narrowing address-less, topic-only queries, and that's better done as an *exact* `topic → blocks` index (no saturation), already in scope for the tracking issue #6785. geth itself moved off bloombits to filtermaps (EIP-7745) for the same saturation reason.

If topic-only narrowing becomes a real need, we'll track it as exact topic indexing under #6785 rather than revive this.

PR #6842, opened 2026-06-11
fix(l1): bump BAL fixtures to v7.3.2 and fix self-destruct-in-initcode + pre-fork newPayload BAL handling

edg-l:
Bumps the Amsterdam BAL fixture bundle to `tests-bal@v7.3.2` (from v7.2.0) and folds in the correctness fixes that the newer fixtures surfaced.

> **Diff scope:** this branch went v7.3.0 → v7.3.1 → v7.3.2 in-flight. The v7.3.0-era CREATE-in-static gas workaround and the stateless `zkevm@v0.4.1` skips (`tooling/ef_tests/blockchain/tests/all.rs`, `docs/known_issues.md`) were added and then reverted within the branch, so they **net to zero** against `main` and do not appear in the diff. See "Why the CREATE-in-static workaround was reverted" below.

### What actually changes vs `main`

- `ci`: bump fixture pins to `tests-bal@v7.3.2` (eels_commit `d28f7977069d1b71984ff88ed8040b671f6e6a77`) across `amsterdam.yaml` and the blockchain/engine/state `.fixtures_url_amsterdam` files.
- `fix` (`block_access_list.rs`): self-destruct in initcode was recording an intermediate balance in the BAL instead of the final post-tx balance of 0. `track_selfdestruct` now collapses intermediate per-tx balance entries to the final `(idx, 0)`.
- `fix` (`serde_utils.rs`): an empty hex string (`"0x"`) encodes the *absence* of a BAL, not an empty list. Deserialize it as `None` so pre-Amsterdam `newPayload` calls (which send `"0x"`) parse instead of failing RLP decode.
- `fix` (`payload.rs`): drop the bespoke V4 fork-boundary detector. A pre-Amsterdam header carrying `block_access_list_hash` produces a `block_hash` that won't match the one ethrex reconstructs (the field is omitted from the V4 header schema), so the mismatch now surfaces as `PayloadStatus.INVALID` via the normal block-hash check — matching the EELS fixture `test_invalid_pre_fork_block_with_bal_hash_field` (`INVALID_BLOCK_HASH`, no engine API error code).

### Why the CREATE-in-static workaround was reverted

In **v7.3.0**, EELS `generic_create` reserved `create_message_gas` (the forwarded frame) *before* checking `is_static`, so a CREATE/CREATE2 in a static context charged that gas and then halted. To match, ethrex moved its `is_static` check after the reservation, burned the forwarded frame, and added `create_static_regular_spill` to carve that reserved-but-unused gas out of EIP-7778's regular-gas dimension. The stale `zkevm@v0.4.1` stateless bundle (filled against ~v7.2.0) didn't have this behavior, hence the temporary skips + `known_issues.md`.

**v7.3.1** (eels `59b466e` → `a8b705f`) moved the `is_static` check back upfront: `raise WriteInStaticContext` now runs *before* any gas is charged, so `create_message_gas` is never reserved in a static context. That made the workaround unnecessary — the spill field and `refund_sender` subtraction were removed, the check returned to its original upfront position (which already matched ethrex's pre-existing behavior), and the stateless skips were dropped since the divergence they masked no longer exists.


tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -34,8 +34,19 @@ const EXTRA_SKIPS: &[&str] = &[
...
+// predates the tests-bal@v7.3.0 CREATE/CREATE2-in-static-context gas refinement
+// (EIP-7778 regular-gas dimension excludes the reserved create_message_gas), so its block
+// headers still encode the old full-frame burn. Our execution is correct for v7.3.0 and
+// therefore mismatches these stale fixtures. Skip them here until a newer zkevm bundle ships;
+// they pass in the amsterdam (v7.3.0) test-levm suite. See docs/known_issues.md.
 #[cfg(feature = "stateless")]
-const EXTRA_SKIPS: &[&str] = &[];
+const EXTRA_SKIPS: &[&str] = &[
+    "bal_create_in_static_context",
+    "create2check_fields_in_initcode",
```
ElFantasma: **Substring skip is broader than the `known_issues.md` claim.** `parse_and_execute` matches via `skipped.iter().any(|s| test_key.contains(s))` (test_runner.rs:48), so `"create2check_fields_in_initcode"` here skips every data-index variant. The doc you added at `docs/known_issues.md:17-18` says only `(d3, d7)` of `create2check_fields_in_initcode` and `(d1)` of `test_static_call_create.py` actually fail against the stale zkevm v0.4.1 fixtures.

Two possibilities:

1. **All variants fail equivalently** — the docs are imprecise but the broader skip is correct. In that case, drop the parenthetical `(d3, d7)` / `(d1)` from `known_issues.md` so it doesn't suggest a tighter granularity that doesn't exist.
2. **Only the listed variants fail** — then this is over-skipping passing tests. Tighten the substring (e.g. `"create2check_fields_in_initcode-d3"`, `"create2check_fields_in_initcode-d7"`, `"test_static_call_create.py::test_create2_init_d1"` or whatever the actual test_key suffix is) so the rest still run.

For `bal_create_in_static_context` the parenthetical says "CREATE/CREATE2, with/without value" — that suggests 4 variants and likely all 4 fail, so a broad skip there is probably right. The other two need the call.

Non-blocking — easy to follow up — but worth nailing down before the next maintainer assumes the docs are authoritative.

crates/vm/levm/src/opcode_handlers/system.rs
```diff
@@ -796,6 +784,22 @@ impl<'a> VM<'a> {
...
+        // `raise WriteInStaticContext` runs past `gas_left -= create_message_gas`. The user is
+        // still charged the whole forwarded frame (exceptional halt), but per the EIP-7778
+        // shared counter the reserved-but-unused `create_message_gas` is NOT counted toward the
+        // block-level regular-gas dimension. Record it as a regular-gas spill so
+        // `refund_sender` excludes it from `block_gas_used` (header) while the receipt still
+        // reflects the full burn. Pre-Amsterdam is handled above.
+        if self.env.config.fork >= Fork::Amsterdam && self.current_call_frame.is_static {
+            self.create_static_regular_spill = self
+                .create_static_regular_spill
+                .checked_add(gas_limit)
```
ElFantasma: nit: `checked_add` is correct defensive code — `create_static_regular_spill` can only grow within a single tx and the cumulative spill is bounded by `vm.env.gas_limit` (each `gas_limit = max_message_call_gas` is per-frame, and you can only enter this branch from frames that are themselves bounded by tx gas), so overflow is unreachable in practice. Worth a one-liner comment so a future reader doesn't wonder if there's a real overflow vector to worry about, or alternatively `saturating_add` to make "can't overflow" explicit. Either is fine.

crates/common/types/block_access_list.rs
```diff
@@ -1567,17 +1567,28 @@ impl BlockAccessListRecorder {
...
             .copied()
             .unwrap_or_default();
-        if pre_balance.is_zero()
-            && let Some(changes) = self.balance_changes.get_mut(&address)
-        {
+        if let Some(changes) = self.balance_changes.get_mut(&address) {
+            // Drop all intermediate balance changes recorded for this tx.
             changes.retain(|(i, _)| *i != idx);
+            // Record the final post-tx balance of 0 unless it equals the pre-tx balance.
+            if !pre_balance.is_zero() {
```
ElFantasma, line 1593: Under Amsterdam+ EIP-7928 BAL + EIP-6780 SELFDESTRUCT semantics, this `if !pre_balance.is_zero()` branch is unreachable in practice: `track_selfdestruct` is only called for contracts created **and** destroyed in the same tx (per the function-level comment + the `is_account_created` gate in `system.rs:680`), so by definition `initial_balances[address]` was either unset (→ `unwrap_or_default` = 0) or set during the in-tx CREATE with the creation balance (typically 0 — CREATE-with-value sets initial_balance only if `balance > 0`, but that path is the CREATE side, not the SELFDESTRUCT side that reaches here).

The branch is sound defensive code matching the EELS post-state diff exactly, but the doc comment above (lines 1568-1581) reads as though the non-zero pre-balance case is a real scenario. Worth tightening the comment to: *"In practice `pre_balance` is always 0 under EIP-6780 (same-tx-created); the non-zero branch mirrors EELS's general pre/post diff and guards against future spec changes or test fixtures that exercise alternate paths."* Otherwise a future reader will hunt for the scenario.

ilitteri (approved):

ElFantasma (approved):
One of the previous comments is still relevant, but it is non-blocking. LGTM

iovoid (approved):

PR #6845, opened 2026-06-11
perf(l1): parallelize single-account storage-trie merkleization

edg-l:
**Motivation**

On the BAL `amsterdam-stateful` benchmark, `test_sstore_bloated` (~60k storage writes into one contract) runs at ~20 Mgas/s vs Nethermind ~96 (0.41x, last of 7 clients). Phase logs show merkleization is ~91% of wall time (13.9s of 15.2s). Stage B parallelizes by account, so a single hot contract lands entirely on one worker running a serial `trie.insert` loop, each insert cold-faulting nodes from a ~500GB DB.

**Description**

Adds `compute_sharded_storage_root` in `crates/blockchain/blockchain.rs`. Accounts with >= 2048 slot updates bypass Stage B and enter a new Stage B.5: their slots are split into 16 buckets by first nibble of the hashed key, inserted in parallel across 16 scoped threads, then reassembled via the existing `collect_trie` + `collapse_root_node` machinery Stage C already uses for the state trie.

Normal accounts (< 2048 slots) go through Stage B byte-for-byte unchanged. Two degenerate cases fall back to a serial path so the persisted node set is bit-identical to the non-sharded code: (1) <= 1 occupied nibble bucket (no parallelism to gain), (2) the update empties the storage trie entirely (rare full-clear).

`compute_sharded_storage_root` is `#[doc(hidden)] pub` so the `ethrex-test` crate can run equivalence tests directly.

Ported/cherry-picked from the `bal-devnet-7` devnet branch.

**Correctness**

- ef-tests engine conformance suite: 5598 tests / 83143 fixtures pass, 0 failed.
- 7 new equivalence tests in `test/tests/blockchain/storage_sharding_tests.rs` assert `(root_hash, node_set)` is bit-identical to a serial reference for: brand-new storage, all-slots-removed, mixed insert/remove/update, single-nibble concentration, single-slot-collapse, two-bucket branch, and normal spread.

**Checklist**


iovoid (commented):
Can the code from the existing doubly-sharded merkelizer be used? There are several tricky edge cases (mostly related with removals) with merkelization, so I would suggest either using a battle-tested implementation or adding even more thorough testing.

edg-l:
The nibble-sharding does already exist in `handle_subtrie`, but that path is the streaming/channel-routed merkleizer; the BAL path here is the staged Stage A/B/C variant over a pre-synthesized map, so it cant call into it without restructuring one or the other. Stage B already parallelizes across accounts, the only gap is a single hot contract that B cant split, which is all Stage B.5 targets. Factoring out a shared shard+collapse helper is doable but Id rather keep that as a follow-up than reshape this PR.

On the removal cases: I pushed a lot more testing. Randomized insert/remove/update mixes over many seeds plus targeted full-bucket-removal and dup-key cases. One thing that surfaced and is worth flagging: on a full-bucket removal in the parallel path the sharded node set is not bit-identical to the serial path (it emits a couple of redundant/orphan nodes). The root hash always matches though, and since the node DB is path-keyed I added a functional check that applies the sharded node set, reopens the trie at the new root and reads every slot back. That passes across all the removal/random cases, so the divergence is harmless write bloat, not a missing or wrongly-deleted node. Updated the test module to assert that functional criterion rather than strict node-set equality for those cases.

crates/blockchain/blockchain.rs
```diff
@@ -3579,3 +3631,167 @@ fn collect_trie(index: u8, mut trie: Trie) -> Result<(Box<BranchNode>, Vec<TrieN
...
+
+    // All 16 shard threads spawn even for empty buckets (same rationale as
+    // Stage C comment): each thread opens the storage trie and returns the
+    // existing subtree at its nibble so root reassembly has no holes.
+    std::thread::scope(|s| -> Result<(), StoreError> {
+        let handles: Vec<_> = buckets
+            .iter()
+            .enumerate()
+            .map(|(nibble, bucket)| {
+                let bucket = bucket.clone();
```
iovoid: Is this needed? I think we could just do into_iter

crates/blockchain/blockchain.rs
```diff
@@ -3598,3 +3650,166 @@ fn collect_trie(index: u8, mut trie: Trie) -> Result<(Box<BranchNode>, Vec<TrieN
...
+            );
+        }
+    }
+
+    // Degenerate case: with <=1 occupied bucket there is no parallelism to gain,
+    // and the branchify/collapse reassembly of a single subtree would emit an
+    // extra (unreachable) node vs the canonical serial diff. Compute serially so
+    // the persisted node set is bit-identical to the non-sharded path.
+    let occupied = buckets.iter().filter(|b| !b.is_empty()).count();
+    if occupied <= 1 {
```
ElFantasma, line 3843: The design here relies on a key correctness assumption stated in the test file: "Bit-identical is therefore sufficient but stronger than necessary on a *path-keyed* node DB" — orphan nodes from the parallel-removal path are harmless because they're never reachable by path. Worth promoting that assumption into a comment here next to the `occupied <= 1` fallback (or near the function's doc), because the fallback's necessity *and* the choice not to fall back on the broader "removal collapses post-bucketization" case both depend on it. If the storage backend ever moves to a content-addressed / hash-keyed node DB (or grows a reachability-based GC), the parallel-removal path could leak unreachable but persistent nodes — a follow-up's worth flagging that as a load-bearing invariant.

Not a request to change behavior — just a docs-only ask so the assumption is visible at the function level, not only in the test module's preamble.

iovoid (approved):

ilitteri (approved):

ElFantasma (approved):
Approving it... Just check if you want to address the comment suggestion, but it's not mandatory

PR #6850, opened 2026-06-11
ci(l1,l2): enrich Slack failure alerts and drop aggregator job noise

ilitteri:
**Motivation**

Failure alerts on main carry little actionable information: a bare SHA, a redundant "failure" field, and a job list that includes branch-protection aggregator jobs while hiding which step failed and why. In a recent alert ([run 27363569281](https://github.com/lambdaclass/ethrex/actions/runs/27363569281)) both listed jobs traced back to one real failure whose test step had been cancelled ~19 minutes in — none of which was visible in Slack.

**Description**

- Filter aggregator jobs by their `Check if any job failed` step instead of a hardcoded name list, so every workflow's aggregator ("Integration Test", "Integration Test L2", "Test", "Lint") is excluded regardless of its name; fall back to listing them only when no real job failed.
- Enrich the message: the headline links to the run and distinguishes scheduled workflows, timeouts, startup failures, and re-run attempts; a commit line shows the commit title (which embeds the PR number); each failed job links to its job page and names the failing step, how it ended (failed / was cancelled / timed out), and its duration.
- Alert only on `failure`, `timed_out`, and `startup_failure` run conclusions, so manually cancelled runs no longer notify.
- Hygiene: the webhook and untrusted inputs (workflow name, commit message) are passed via `env:`; the job list is capped at 10 with attempt-level dedupe; Slack mrkdwn special characters are escaped; `curl` retries 3 times and propagates failure so lost alerts show up as red runs.

**Examples**

Before — the actual alert for run 27363569281:

```
🚨 Workflow failed on main
Workflow                Conclusion
L2 (without proving)    failure
Commit                  Run
d66f17cb                Open in GitHub
Failed job(s)
- State Reconstruction Tests
- Integration Test L2
```

After — the same run replayed through the new scripts:

```
🚨 L2 (without proving) failed on main
Commit: d66f17cb chore(l1): use rocksdb block cache of 4 GiB in multisync (#6836)
Failed job(s)
- State Reconstruction Tests — step `Run tests` was cancelled after 18m 47s
```

The headline links to the run, the SHA to the commit, and each job name to its job page. "Integration Test L2" is the aggregator job and is no longer listed.

Headline variants:

```
🚨 Scheduled workflow Daily Hive Report timed out
🚨 L1 failed on main (attempt 2)
🚨 L1 failed to start on main
```

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A (CI only)

### DIFF
--- a/scripts/collect_failed_jobs.js
+++ b/scripts/collect_failed_jobs.js
@@ -1,7 +1,87 @@
 'use strict';
 
+// Aggregator jobs exist only to give branch protection a single required
+// check (e.g. "Integration Test L2"); they fail whenever any real job fails,
+// so listing them in the alert would only duplicate information. They are
+// recognized by their check step rather than by name so that every workflow's
+// aggregator is covered regardless of what the job itself is called.
+const AGGREGATOR_STEP_NAME = 'Check if any job failed';
+
+const FAILING_JOB_CONCLUSIONS = new Set([
+  'failure',
+  'timed_out',
+  'action_required',
+]);
+
+// A step killed by a timeout or a runner-side cancellation reports
+// `cancelled` while its job still reports `failure`, so `cancelled` steps
+// must be considered to locate where a failed job actually stopped.
+const FAILING_STEP_CONCLUSIONS = new Set(['failure', 'timed_out', 'cancelled']);
+
+// Slack section blocks are limited to 3000 characters and each job line can
+// take ~200 with its link, so cap the list defensively.
+const MAX_LISTED_JOBS = 10;
+
+const CONCLUSION_VERBS = {
+  failure: 'failed',
+  timed_out: 'timed out',
+  cancelled: 'was cancelled',
+  action_required: 'needs action',
+};
+
+// Escape the characters that are special in Slack mrkdwn text.
+function escapeSlackText(text) {
+  return String(text)
+    .replaceAll('&', '&amp;')
+    .replaceAll('<', '&lt;')
+    .replaceAll('>', '&gt;');
+}
+
+function formatDuration(startedAt, completedAt) {
+  if (!startedAt || !completedAt) {
+    return null;
+  }
+  const ms = new Date(completedAt) - new Date(startedAt);
+  if (!Number.isFinite(ms) || ms <= 0) {
+    return null;
+  }
+  const totalMinutes = Math.floor(ms / 60_000);
+  if (totalMinutes >= 60) {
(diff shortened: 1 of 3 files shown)


MegaRedHand (approved):

edg-l (approved):

scripts/collect_failed_jobs.js
```diff
@@ -1,34 +1,119 @@
 'use strict';
 
+// Aggregator jobs exist only to give branch protection a single required
+// check (e.g. "Integration Test L2"); they fail whenever any real job fails,
+// so listing them in the alert would only duplicate information. They are
+// recognized by their check step rather than by name so that every workflow's
+// aggregator is covered regardless of what the job itself is called.
+const AGGREGATOR_STEP_NAME = 'Check if any job failed';
```
ElFantasma, line 8: The aggregator detection now depends on the *step* being literally named `Check if any job failed` instead of the *job* matching a hardcoded name list. That's a strictly broader catch (works across `Integration Test`, `Integration Test L2`, `Test`, `Lint`, etc. as the comment notes), but it's also a soft contract with whoever writes future workflow files: anyone adding a new branch-protection aggregator must name its check step *exactly* this string, or the new aggregator will be listed in the Slack alert. Worth (a) adding a short note in a CONTRIBUTING/CI doc (or in `common_failure_alerts.yaml`) referencing this constant so the convention is discoverable, and/or (b) considering a small inverse check at workflow time (e.g. a pre-merge lint that flags `aggregator`-like jobs without the expected step name). Non-blocking — the PR is a strict improvement either way; just flagging because the contract is now invisible to people adding new workflows.

ElFantasma (approved):

scripts/collect_failed_jobs.js
```diff
@@ -1,34 +1,119 @@
...
+  }
+  const ms = new Date(completedAt) - new Date(startedAt);
+  if (!Number.isFinite(ms) || ms <= 0) {
+    return null;
+  }
+  const totalMinutes = Math.floor(ms / 60_000);
+  if (totalMinutes >= 60) {
+    return `${Math.floor(totalMinutes / 60)}h ${totalMinutes % 60}m`;
+  }
+  const seconds = Math.round((ms % 60_000) / 1000);
```
iovoid, line 52: Should also be floor, to avoid displaying "60s"

iovoid (approved):

Issue #5564, opened 2025-12-10
Handle reorgs longer than 128 blocks


MegaRedHand:
We currently handle reorgs of up to 128 blocks through in-memory difflayers. We should implement an off-memory alternative for longer reorgs, which should enable us to handle long non-finalization periods.

Related: #5558

PR #6851, opened 2026-06-12
fix(l1): start sync on unreachable-state FCU instead of idling

edg-l:
**Motivation**

When an FCU references a head whose state we can't reach from the DB, the StateNotReachable arm returned PayloadStatus::syncing() but never armed the syncer. The CL keeps resending FCUs we keep ignoring, so a node that has fallen behind never makes progress and stays wedged.

**Description**

Call syncer.sync_to_head(head) like the Syncing arm does. sync_to_head is idempotent (only starts a cycle when the syncer is inactive) and mode-agnostic, so it is safe for both full and snap clients and avoids re-pivot thrash on repeated FCUs.

**Checklist**

Closes #issue_number


ilitteri (approved):

iovoid (approved):

crates/networking/rpc/engine/fork_choice.rs
```diff
@@ -349,7 +349,15 @@ async fn handle_forkchoice(
...
-                    // Ignore the FCU
+                    // We can't reach the head's state from our DB (the nearest
+                    // link block has pruned or not-yet-executed state). Kick off
+                    // a sync toward the head instead of reporting SYNCING while
+                    // sitting idle, which wedges the node: the CL keeps resending
+                    // FCUs we keep ignoring and we never make progress.
+                    // sync_to_head is idempotent (only starts a cycle if the
+                    // syncer is inactive) and mode-agnostic, so this is safe for
+                    // both full and snap clients.
+                    syncer.sync_to_head(fork_choice_state.head_block_hash);
```
ElFantasma, line 360: **Off-diff: stale `warn!` in `crates/blockchain/fork_choice.rs:163-173` contradicts the new behavior.** That warning still says:

> `"FCU head state not reachable from DB state. Ignoring fork choice update. This is expected if the consensus client is currently syncing. Otherwise, if consensus is synced and this is a consistent message it can be fixed by removing the DB and re-syncing the execution client."`

And the comment above it: `"If the state can't be constructed from the DB, we ignore it and log a warning."`

With this PR we no longer ignore the FCU — we kick off a sync. The log will mislead operators looking at logs after a wedge ("why does it say ignored if it's syncing?"), and the remediation advice ("remove DB and re-sync") is now overly aggressive since the auto-resync handles most cases.

Small touch-up worth folding into this PR: change "Ignoring fork choice update" → "Starting sync toward head" (or similar) and drop the manual-resync remediation. Non-blocking but the messaging drift will compound otherwise.

ElFantasma, line 360: nit: `InvalidForkChoice::Syncing` (line 347) and `InvalidForkChoice::StateNotReachable` now produce literally the same body — `syncer.sync_to_head(head); ForkChoiceResponse::from(PayloadStatus::syncing())`. Could be consolidated as:

```rust
InvalidForkChoice::Syncing | InvalidForkChoice::StateNotReachable => {
    syncer.sync_to_head(fork_choice_state.head_block_hash);
    ForkChoiceResponse::from(PayloadStatus::syncing())
}
```

Keeping them separate makes sense if you want per-reason logging in the future (and the doc-comment block here is genuinely useful explaining the wedge), so this is fine as-is. Mentioning only because the `// TODO(#5564): handle arbitrary reorgs` marker sits between them and may make a future reader miss that they're functionally identical.

ElFantasma (approved):

PR #6855, opened 2026-06-13
perf(l1): collect code hashes in a single pass during the account trie build

ilitteri:
**Motivation**

insert_accounts iterated the ingested temp DB twice: a dedicated full pass that only decoded each account to collect its code hash, then the build pass whose inspect closure already decodes the same accounts. On mainnet that is one redundant sequential scan plus RLP decode over the entire account set.

**Description**

Fold code-hash collection into the build pass's closure, via a sync flush variant on CodeHashCollector (skips awaiting the previous disk write; the JoinSet accumulates a handful of dump tasks whose results finish() already collects).

Part of the snap-sync pipelining stack; tested all-in-one on a fresh mainnet sync plus multisync trie-validation cycles.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` — not needed: no storage changes.

### DIFF
--- a/crates/networking/p2p/sync/code_collector.rs
+++ b/crates/networking/p2p/sync/code_collector.rs
@@ -55,6 +55,19 @@ impl CodeHashCollector {
         Ok(())
     }
 
+    /// Sync variant of [`Self::flush_if_needed`], callable from iterator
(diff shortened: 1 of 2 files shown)


ilitteri:

PR #6857, opened 2026-06-13
refactor(l1): replace per-phase sync progress blocks with an all-lane table

ilitteri:
**Motivation**

The periodic progress display rendered one block for whichever phase owned the global step, assuming phases run exclusively — no longer true with pipelined sync, and misleading on resumed syncs (per-cycle numerator over absolute-height denominator).

**Description**

Each tick prints one table with a row per lane: progress, per-lane rate, state (queued/active/idle/done), and the critical-path marker. Rates are computed over the measured interval — the old blocks divided 10s deltas by an assumed 30s, underreporting threefold. Phase separators/completions remain as scrollback event markers.

Stacked on the streamed-bytecodes PR.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` — not needed: no storage changes.


ilitteri:

PR #6858, opened 2026-06-13
perf(l1): emit account FlatKeyValue rows during the snap-sync trie build

ilitteri:
**Motivation**

The post-sync FlatKeyValue generator re-derives account rows by walking the whole trie, hours of background I/O — while the snap-sync trie build streams the exact same sorted leaves with byte-identical values.

**Description**

The build's inspect closure also writes each account leaf into ACCOUNT_FLATKEYVALUE (routed by key length through the existing trie backend), under a construction marker (MISC_VALUES "fkv_prebuilt_accounts"). The read gate (last_written) stays closed and the generator is untouched: rows are inert until a later promotion phase reconciles them — this PR lands the emission mechanism only.

Stacked on the lane-table PR.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` — not needed: new rows live in an existing CF, no schema change.


ilitteri:

PR #6859, opened 2026-06-13
perf(l1): start state download against the probed sync head while headers backfill in the background

ilitteri:
**Motivation**

The header walk fully serialized the start of snap sync: dozens of 500k batches before the first account request, despite the state phases needing only the pivot header — which the sync-head probe already fetches and discarded.

**Description**

When the probe succeeds, state download starts immediately against the probed header as a provisional pivot while a background task backfills the header chain. The two header segments stay separate and concatenate before the forkchoice update (one contiguous chain ending at the pivot): the backfill owns its segment and the header-download checkpoint; pivot extensions never advance the checkpoint (a restart would otherwise resume past headers the backfill never downloaded). Backfill failure ends the cycle without FCU and without clearing snap state — today's wait-for-a-newer-head semantics. Probe failure falls back to the unchanged serial path.

Verified on a fresh mainnet sync: headers and accounts download concurrently from the first minute.

**Checklist**

### DIFF
--- a/crates/networking/p2p/sync.rs
+++ b/crates/networking/p2p/sync.rs
@@ -213,9 +213,11 @@ impl Syncer {
             // barely synced themselves. Pre-checking avoids that. The probe
             // response is intentionally discarded; on the snap path the loop
             // re-fetches headers, which keeps `sync_cycle_snap`'s entry simple.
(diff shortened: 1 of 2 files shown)


ilitteri:

PR #6860, opened 2026-06-13
perf(l1): emit FlatKeyValue rows for leaves healed during snap sync

ilitteri:
**Motivation**

Prebuilt FKV rows from pre-healing range data go stale wherever healing later changes leaves; emitting at heal time keeps the prebuilt set reconcilable for the promotion phase.

**Description**

State healing inserts each healed leaf's full-path row into the same batched write transaction as its node batch; storage healing collects per-account leaf rows and writes them via a new narrow Store::write_storage_flatkeyvalue_batch from the same single background writer task (ordering preserved). Leaf deletions are not observable in healing — documented at the emission sites; the verifier phase reconciles. Read gate and generator untouched; rows stay inert.

**Checklist**


ilitteri:

PR #6861, opened 2026-06-13
perf(l1): download storage ranges concurrently with the account phase

ilitteri:
**Motivation**

Storage downloads waited for the full account phase (download, build, heal) because the downloader resolved storage roots from the built trie — but every account leaf already carries its storage root. This was the largest remaining serialization in snap sync.

**Description**

Account-range responses feed discovered (account, storage root) pairs to a background runner driving the existing storage downloader in waves, concurrently with account download and trie build. Roots travel with the work: the big-account split keeps its group root instead of degrading to a None entry (which required the built trie and exited the node on a miss). Wave leftovers carry into the post-build loop, which keeps heal-based reconciliation and the heal-everything fallback; wave-completed accounts are excluded. Stale waves wait for a fresher pivot published by the account dispatcher rather than re-pivoting. Design doc: storage-root ownership, pull-based handout, bounded dedup loss.

**Checklist**

### DIFF
--- a/crates/networking/p2p/snap/client.rs
+++ b/crates/networking/p2p/snap/client.rs
@@ -26,6 +26,7 @@ use crate::{
 use bytes::Bytes;
 use ethrex_common::{
     BigEndianHash, H256, U256,
+    constants::EMPTY_TRIE_HASH,
     types::{AccountState, BlockHeader},
 };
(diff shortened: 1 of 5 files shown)


ilitteri:

PR #6862, opened 2026-06-13
perf(l1): ingest account snapshot files while the range download is still running

ilitteri:
**Motivation**

Account snapshot SSTs ingested into the temp DB only after the whole download completed — a serialization with no dependency behind it beyond per-file completion.

**Description**

Each finished dump file is sent (with its chunk index) to a background ingestor owning the temp RocksDB, which ingests the contiguous index prefix in order (re-pivoted re-downloads make ordering load-bearing: later files must win). At build time the ingestor is joined and an in-dir sweep ingests any remainder — move_files semantics make 'files still in the dir' exactly the un-ingested set, covering ingestor failure and restarts without double-ingest. The single trie build after the join is untouched.

**Checklist**


ilitteri:

PR #6863, opened 2026-06-13
perf(l1): improve RpcExecutionWitness::into_execution_witness

han0110:
## Motivation

The current `DecodedEip8025::Canonical` makes guest program much more expensive due to `RpcExecutionWitness::into_execution_witness`, this PR improve it to reduce its cost by 3.43x, and overall the overhead of `execution_program` using canonical input vs legacy input is reduced from 2.04x to 1.09x.

### By commits

- https://github.com/lambdaclass/ethrex/commit/8a1500f6af327b1a1fe6e070b7ef41055707aaba - Route mpt keccak through the caller's `Crypto` impl. (this is also presented in #6797)
- https://github.com/lambdaclass/ethrex/commit/be4d0c486693b88819cd1687b27efa01db8c5f11 - Seed each embedded node's memoization slot with the hash it was just resolved by, so later passes reuse it instead of re-hashing.
- https://github.com/lambdaclass/ethrex/commit/8acd9789a5c94379225cf534416468d65426f831 - Swap the witness node map from `BTreeMap` to `FxHashMap`, replacing `O(logN)` key-comparison chains with `O(1)` hashed lookups.
- https://github.com/lambdaclass/ethrex/pull/6863/commits/a16f5a70a6e9fb3934dc5988c742eced0aa2be54 and https://github.com/lambdaclass/ethrex/pull/6863/commits/86e843af0efaeb69a47b4182a70e697fd84b73b3 - Reexport canonical input structures and EIP-8025 execution functions

#### `RpcExecutionWitness::into_execution_witness` comparison of `main` and this PR

into_execution_witness

#### Input to `GuestProgramState` comparison of legacy input and canonical input using this PR

Input to `GuestProgramState` includes:

- Legacy - `read_input → rkyv → ExecutionWitness → GuestProgramState`
- Canonical - `read_input → ssz → CanonicalExecutionWitness →  RpcExecutionWitness → ExecutionWitness → GuestProgramState`

input_to_guest_program_state

To truly close the gap between legacy and canonical, we'd need to stop building the trie a second time (avoid constructing an owned linked tree via `Trie::get_embedded_root`, and avoid the pre-walk that embeds storage tries). Either resolve nodes from the `FxHashMap<H256, Node>` on demand, or materialize once into a cheap representation (minimal per-node allocation/copy).

**Checklist**

### DIFF
--- a/crates/common/trie/trie.rs
+++ b/crates/common/trie/trie.rs
@@ -27,7 +27,7 @@ pub use self::nibbles::Nibbles;
 pub use self::threadpool::ThreadPool;
 pub use self::verify_range::verify_range;
 pub use self::{
-    node::{Node, NodeRef},
+    node::{Node, NodeRef, OnceLock},
     node_hash::NodeHash,
 };
 
@@ -330,6 +330,16 @@ impl Trie {
             TrieError::InconsistentTree(Box::new(InconsistentTreeError::RootNotFound(root_hash)))
         })?;
 
+        /// Creates an embedded node reference with its hash slot pre-seeded.
+        ///
+        /// The caller must guarantee that `hash` is the hash of the referenced
+        /// node, for example because the node was just resolved by looking that
+        /// hash up. Seeding lets later hash computations over the subtree reuse
+        /// the known value instead of re-encoding and re-hashing it.
+        fn node_with_hash(node: Node, hash: NodeHash) -> NodeRef {
+            NodeRef::Node(Arc::new(node), OnceLock::from(hash))
+        }
+
         fn get_embedded_node(
             all_nodes: &BTreeMap<H256, Node>,
             cur_node: &Node,
@@ -344,7 +354,10 @@ impl Trie {
 
                         if hash.is_valid() {
                             *choice = match all_nodes.get(&hash.finalize(crypto)) {
-                                Some(node) => get_embedded_node(all_nodes, node, crypto)?.into(),
+                                Some(node) => node_with_hash(
+                                    get_embedded_node(all_nodes, node, crypto)?,
+                                    hash,
+                                ),
                                 None => hash.into(),
                             };
                         }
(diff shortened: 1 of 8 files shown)


ilitteri (approved):

ElFantasma (approved):

crates/common/trie/trie.rs
```diff
@@ -366,8 +383,8 @@ impl Trie {
             })
         }
 
-        let root = get_embedded_node(all_nodes, root_rlp)?;
-        Ok(root.into())
+        let root = get_embedded_node(all_nodes, root_rlp, crypto)?;
+        Ok(node_with_hash(root, NodeHash::Hashed(root_hash)))
```
iovoid, line 445: This breaks `InMemoryTrieDB::from_nodes`, which assumes `Trie::get_embedded_root` returns a trie where all nodes are uncommitted (no hash cached).

iovoid (changes requested):

crates/common/trie/trie.rs
han0110: Also it assumes the `all_nodes: &FxHashMap<H256, Node>` is well-formed (the key is indeed keccak256 hash of the value). It might be better to wrap the `all_nodes: &FxHashMap<H256, Node>` into a structure that can only be constructed from raw RLP values so it's guaranteed to be well-formed.

If it breaks `InMemoryTrieDB::from_nodes`, will it be possible to ad a new method like `get_embedded_root_committed` so it doesn't affect original usage of `get_embedded_root` (the main purpose is to avoid double hashing).

iovoid: The best long-term solution is probably to drop that API for one that directly takes `&[NodeRLP]`, since that is how it's currently being used. This refactor is probably better left for another PR.

If the simple way to keep it working is to add a get_embedded_root_committed method, then go ahead with that.

han0110: Got it, reverted the changes of get_embedded_root to keep this PR non breaking change, and add a new method `Trie::get_embedded_root_committed` only for `RpcExecutionWitness::into_execution_witness` in https://github.com/lambdaclass/ethrex/pull/6863/commits/396b8912f88c9957b99b7d440af15595c6bfdefb.

iovoid (approved):

PR #6866, opened 2026-06-15
perf(levm): pad bytecode with 33 STOPs to remove hot-path bounds checks

edg-l:
Pad `Code.bytecode` with 33 zero bytes (STOP opcodes) at construction time so `next_opcode` can do an unchecked index and `advance_pc` becomes a plain `pc += n` with no `Result`. The logical length is tracked in `bytecode_len`; `dispatch_buf()` exposes the padded buffer only for the dispatch loop.

Branch removal is real (~8% fewer instructions, ~13% fewer branches on the dispatch path) but perf-neutral on cycles/wall-time since those branches were ~100% predictable. The value is hot-path simplification: no `Result` return from `next_opcode`/`advance_pc`, no pattern-match overhead per opcode.

`Code::default()` is now `[0u8; 33]` so unchecked reads on empty inner CALLs are safe.

Because the physical `bytecode.len()` now differs from the logical length, `Code.bytecode` and `bytecode_len` were made private and accessors added (`code()`, `code_bytes()`, `len()`, `is_empty()`, `dispatch_buf()`). Roughly 12 call sites across consensus-relevant paths were switched to the logical length: EIP-7702 delegation detection, CREATE gas counting, `eth_getCode`, snap-server `GetByteCodes`, BAL code-change recording/validation, witness code collection, tracing, L2 sponsored-tx parsing, and empty-code checks.

### DIFF
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -30,7 +30,7 @@ static EMPTY_JUMP_TARGETS: LazyLock<Arc<[u32]>> = LazyLock::new(|| Arc::from(Vec
 /// padding regardless of which opcode sits at the last real byte.
 pub const BYTECODE_PADDING: usize = 33;
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Code {
     // hash is only used for bytecodes stored in the DB, either for reading it from the DB
     // or with the CODEHASH opcode, which needs an account address as argument and
@@ -154,6 +154,41 @@ impl Code {
     }
 }
 
+/// Serde shadow for [`Code`]. Stores the *logical* (unpadded) bytecode so the
+/// padding is never part of the serialized form. Deserialization re-pads through
+/// [`Code::from_parts_unchecked`], which keeps the dispatch-loop invariant (every
+/// `Code` is padded with [`BYTECODE_PADDING`] trailing STOPs) sound regardless of
+/// where the bytes came from. Deserializing the padded buffer directly would
+/// otherwise let unpadded input through and cause OOB reads during execution.
+#[derive(Serialize, Deserialize)]
+struct CodeSerde {
+    hash: H256,
+    code: Bytes,
+    jump_targets: Arc<[u32]>,
+}
+
+impl Serialize for Code {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        CodeSerde {
+            hash: self.hash,
+            code: self.code_bytes(),
+            jump_targets: self.jump_targets.clone(),
+        }
+        .serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for Code {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let CodeSerde {
+            hash,
+            code,
+            jump_targets,
+        } = CodeSerde::deserialize(deserializer)?;
+        Ok(Self::from_parts_unchecked(hash, &code, jump_targets))
+    }
+}
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct CodeMetadata {
     pub length: u64,
@@ -215,7 +250,7 @@ impl Default for AccountState {
 impl Default for Code {
     fn default() -> Self {
         Self {
-            bytecode: Bytes::from_static(&[0u8; 33]),
+            bytecode: Bytes::from_static(&[0u8; BYTECODE_PADDING]),
             bytecode_len: 0,
             hash: *EMPTY_KECCAK_HASH,
             jump_targets: EMPTY_JUMP_TARGETS.clone(),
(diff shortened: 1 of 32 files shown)


iovoid (commented):
I would put the 33 inside a constant to make it's purpose clearer.

crates/common/types/account.rs
```diff
@@ -93,6 +102,31 @@ impl Code {
...
+        self.bytecode_len
+    }
+
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.bytecode_len == 0
+    }
+
+    #[inline]
+    pub fn dispatch_buf(&self) -> &[u8] {
```
iovoid, line 147: The name could be clearer and/or have some documentation.

crates/common/types/account.rs
```diff
@@ -48,18 +51,24 @@ impl Code {
...
+        Self::from_parts(hash, &code, jump_targets)
     }
 
     pub fn from_bytecode(code: Bytes, crypto: &dyn Crypto) -> Self {
         let jump_targets = Self::compute_jump_targets(&code);
+        let hash = H256(crypto.keccak256(code.as_ref()));
+        Self::from_parts(hash, &code, jump_targets)
+    }
+
+    pub fn from_parts(hash: H256, code: &[u8], jump_targets: Arc<[u32]>) -> Self {
```
MegaRedHand: Should this be `_unchecked`?

crates/vm/levm/src/call_frame.rs
```diff
@@ -534,12 +530,11 @@ impl<'a> VM<'a> {
...
-            .pc
-            .checked_add(count)
-            .ok_or(InternalError::Overflow)?;
-        Ok(())
+    #[expect(
+        clippy::arithmetic_side_effects,
+        reason = "pc bounded by padded bytecode len"
+    )]
+    pub fn advance_pc(&mut self, count: usize) {
+        self.current_call_frame.pc += count;
```
MegaRedHand: Can we remove the `count` parameter here? I see it being used only with `count=1`

MegaRedHand (approved):
LGTM

crates/vm/levm/src/opcode_handlers/push.rs
```diff
@@ -31,31 +31,39 @@ pub struct OpPushHandler<const N: usize>;
 impl<const N: usize> OpcodeHandler for OpPushHandler<N> {
     #[inline(always)]
     fn eval(vm: &mut VM<'_>) -> Result<OpcodeResult, VMError> {
+        // PUSHn reads up to 32 immediate bytes without a bounds check, relying on
+        // the trailing zero padding appended to every bytecode. Keep the unchecked
+        // read below sound if the padding ever shrinks.
+        const { assert!(BYTECODE_PADDING >= 32) };
```
iovoid: If the push reads 32 bytes, that means there needs to be an additional byte for the read to be sound. This should be >32 (or equivalently >=33)

crates/common/types/account.rs
```diff
@@ -176,7 +215,8 @@ impl Default for AccountState {
 impl Default for Code {
     fn default() -> Self {
         Self {
-            bytecode: Bytes::new(),
+            bytecode: Bytes::from_static(&[0u8; 33]),
```
iovoid: Should use BYTECODE_PADDING

crates/common/types/account.rs
```diff
@@ -24,6 +24,12 @@ use crate::constants::{EMPTY_KECCAK_HASH, EMPTY_TRIE_HASH};
 /// placeholder and every EOA / empty-code load would otherwise each allocate.
 static EMPTY_JUMP_TARGETS: LazyLock<Arc<[u32]>> = LazyLock::new(|| Arc::from(Vec::new()));
 
+/// Trailing STOP bytes appended to every bytecode so the dispatch loop can read
+/// the next opcode without a bounds check. 33 is the widest single-opcode advance
+/// (PUSH32: 1 opcode byte + 32 immediate bytes), so `pc` can never step past the
+/// padding regardless of which opcode sits at the last real byte.
+pub const BYTECODE_PADDING: usize = 33;
+
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
```
iovoid: Serialize and Deserialize will act on the padded bytecode. Deserializing and executing a bytecode without padding leads to an OOB read.

iovoid (changes requested):

crates/common/types/account.rs
```diff
@@ -48,18 +57,26 @@ impl Code {
     // or never be read (e.g. for initcode).
     pub fn from_bytecode_unchecked(code: Bytes, hash: H256) -> Self {
         let jump_targets = Self::compute_jump_targets(&code);
-        Self {
-            hash,
-            bytecode: code,
-            jump_targets,
-        }
+        Self::from_parts_unchecked(hash, &code, jump_targets)
```
ElFantasma, line 63: Doc-only nit: neither `from_bytecode` nor `from_bytecode_unchecked` carries an explicit "`code` is the *logical*, unpadded bytecode — padding is added internally by `from_parts_unchecked`" note. The invariant is correctly enforced by the private fields (so external misuse is structurally impossible) and the existing call sites are all correct, but the constructor-level contract is currently only obvious from reading the impl. A one-liner on each public constructor (or just `from_parts_unchecked`, since both delegate) saying "the caller passes the logical bytecode; this function appends `BYTECODE_PADDING` STOP bytes" would future-proof against someone adding a constructor that takes a pre-padded buffer and silently producing a Code with `bytecode_len == padded_len`. Non-blocking.

ElFantasma (approved):
Doc-only nit inline. LGTM

iovoid (approved):

PR #6867, opened 2026-06-15
fix(l2): fix critical bugs in bench_metrics.sh

operagxoksana:
Fix three bugs in scripts/bench_metrics.sh: replace ((ms < min)) && min=$ms and ((ms > max)) && max=$ms with if (( )); then fi to prevent silent crashes under set -e when the arithmetic condition evaluates to false; replace the magic sentinel min=999999999 with min=${ms_arr[0]} to avoid incorrect results for large proving times; and add a local guard before min/max initialization to make the non-empty array invariant explicit rather than relying on a distant check earlier in the script.

This is a resubmission of #6723, accidentally closed when I deleted the fork. @ElFantasma @azteca1998 you already reviewed this one

### DIFF
--- a/scripts/bench_metrics.sh
+++ b/scripts/bench_metrics.sh
@@ -49,7 +49,7 @@ gas_arr=()
 txs_arr=()
 blocks_arr=()
 
-while read -r line; do
+while IFS= read -r line; do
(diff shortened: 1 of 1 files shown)


ElFantasma (commented):
LGTM

PR #6871, opened 2026-06-15
fix(l1): clamp eth_getTransactionCount pending nonce to latest

edg-l:
## Problem

`eth_getTransactionCount` could return a lower nonce for `pending` than for `latest`:

```
latest:   {"result":"0x59"}
pending:  {"result":"0x51"}
```

`pending` must never be below `latest`.

## Root cause

The handler read the pending nonce from the mempool and returned it unconditionally. `mempool.get_nonce` returns `highest_mempool_nonce_for_sender + 1`. When the pool still holds stale txs for a sender (nonces below the account's current on-chain nonce), that value can be lower than the canonical latest nonce, and the handler returned it verbatim.

## Fix

Resolve the on-chain nonce first, then for `Pending` return `mempool_nonce.max(account_nonce)`. The mempool may only advance the pending nonce past the on-chain value, never below it.

## Tests

- `pending_nonce_is_clamped_to_latest` — stale pool tx below on-chain nonce → `pending == latest` (fails on old code).
- `pending_nonce_advances_past_latest` — pool tx at the on-chain nonce → `pending` advances by one.

### DIFF
--- a/crates/networking/rpc/eth/account.rs
+++ b/crates/networking/rpc/eth/account.rs
@@ -163,28 +163,28 @@ impl RpcHandler for GetTransactionCountRequest {
             self.address, self.block
         );
 
-        // If the tag is Pending, we need to get the nonce from the mempool
-        let pending_nonce = if self.block == BlockTag::Pending {
-            context.blockchain.mempool.get_nonce(&self.address)?
-        } else {
-            None
+        // Resolve the canonical nonce for the requested block first. For the
+        // `Pending` tag this resolves to the latest block.
+        let Some(block_number) = self.block.resolve_block_number(&context.storage).await? else {
+            return serde_json::to_value("0x0")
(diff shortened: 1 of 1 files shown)


ElFantasma (approved):

crates/networking/rpc/eth/account.rs
```diff
@@ -291,4 +291,91 @@ mod tests {
...
+                H256::random(),
+                address,
+                MempoolTransaction::new(tx, address),
+            )
+            .unwrap();
+    }
+
+    /// Regression: a stale tx left in the pool with a nonce below the account's
+    /// on-chain nonce must not make `pending` report a value lower than `latest`.
+    #[tokio::test]
```
iovoid, line 347: The tests should be on the test/ folder

iovoid (approved):

ilitteri (approved):

PR #6872, opened 2026-06-15
perf(l1): sharded parallel multi_get for storage-heavy BAL prefetch

edg-l:
Stacked on #6712. Adds `Store::get_storage_values_batch_by_root` and wires it into `CachingDatabase::prefetch_storage` for storage-heavy blocks.

**Problem.** `prefetch_storage` issues one storage read per slot. On cold, storage-heavy blocks the per-slot `par_iter` path caps queue depth at ncpu. A single sorted serial `multi_get` fixes the block-sharing problem but collapses queue depth to ~1 (async_io is off in our build; no io_uring), which regressed `sload_bloated` ~4.5x on the cold Benchmarkoor suite.

**How.** `get_storage_values_batch_by_root` sorts the FKV leaf keys by hashed order (so adjacent slots share 4KB RocksDB data blocks), then splits them into contiguous shards (`KEYS_PER_SHARD=256`, `MAX_SHARDS=64`) read concurrently via `std::thread::scope` sharing one read snapshot. Sorted shards keep data-block sharing while running at high queue depth; both levers at once.

`prefetch_storage` is gated: below `BLOATED_BATCH_THRESHOLD=49152` it keeps the warm-optimal per-slot `par_iter` point-gets (sharded multi_get costs ~28% on a warm batch). The gate sits above the physical max distinct cold slots a block can reach at a 100M gas limit (~47.6k at 2100 gas/SLOAD), so normal mainnet never triggers it even through Glamsterdam's gas-limit increase. It engages on genuinely storage-heavy blocks and hardens validation against storage-bloat DoS.

**Measurements** (local microbench, 6 cores pinned, O_DIRECT cold, one-bloated-account key shape, ~200M-gas-equivalent batch):
- `par_iter` baseline (QD ~= 6): 0.060 Mslots/s
- sharded multi_get (QD ~= 64): 0.255 Mslots/s (~4.25x)
- serial multi_get (the regressing path): 0.017 Mslots/s
- range-scan: 0.033 Mslots/s (over-reads at sparse density, loses)

The win grows as cores shrink because sharded decouples queue depth from core count, which `par_iter` cannot.

**Correctness.** `storage_batch_parity_in_memory` and `storage_batch_parity_rocksdb` in `test/tests/storage/storage_batch_tests.rs` assert the batch path is byte-identical to the per-slot single-get path; the rocksdb test crosses multiple shards (`POPULATED_SLOTS=2048`). EF blockchain LEVM suite: 8758 passed / 0 failed.

**Notes.** This targets storage-heavy cold reads only; the gate keeps it inert on normal/warm blocks. It does not close the throughput gap to other clients on the bloated suite, which is dominated by storage layout (hashed-key FKV => random reads) and the exec/merkle pipeline.


edg-l:

PR #6873, opened 2026-06-16
perf(l1): BAL-driven trie-node prefetch to warm merkle reads

edg-l:
Stacked on #6872. Will retarget to main once that merges.

The existing flat-KV prefetch (from #6872) warms the leaf-value CFs that execution reads. The merkle phase reads from a completely separate set of column families, `STORAGE_TRIE_NODES` and `ACCOUNT_TRIE_NODES`, and every node read there is cold. This adds a complementary prefetch that warms those internal nodes up front from the same BAL working set, before the merkleizer starts.

The key insight is that internal nodes on the path to a leaf all live at prefixes of that leaf's nibble path, so for each BAL leaf we speculatively probe every prefix up to depth 8 (covers tries up to ~16^8 entries near-fully). Prefixes that don't correspond to real nodes are rejected by the CF bloom filter in RAM with no disk I/O. This turns the cold merkle node reads into one sorted, sharded, high-queue-depth `multi_get` per CF, mirroring the approach from #6872. Deduplication collapses shared shallow prefixes so roots and high-level branches are read once regardless of working-set size.

Covers both storage tries (BAL storage slots, via `apply_prefix(hashed_addr, slot_nibbles)`) and the state/account trie (BAL addresses including balance/nonce-only accounts that `bal_storage_slots` doesn't surface). The `bal_accounts` helper in `levm/mod.rs` extracts those. Runs synchronously next to the flat-KV prefetch in `blockchain.rs`: it touches different CFs than execution so it doesn't contend with the executor, and it completes before the merkleizer starts so it doesn't race it.

Gated on `slots.len() + accounts.len() >= 16_384` so ordinary blocks don't pay the probe cost. At ~2100 gas per cold access, 16384 corresponds to roughly 34M gas of cold accesses, well above normal traffic and below the large-state blocks this targets.

The prefetch is best-effort and read-only: missing a node (e.g. one deeper than depth 8) is harmless, the merkleizer cold-reads it. It cannot affect consensus output.

`Store::prefetch_trie_nodes` returns a hit count. Two new tests, `trie_node_prefetch_effectiveness_in_memory` and `trie_node_prefetch_effectiveness_rocksdb`, seed a contract with 2048 storage slots, run the prefetch, and assert the hit count is at least 64. A key-encoding mismatch (wrong nibble order, stray leaf flag, wrong `apply_prefix` wrapping) would turn every probe into a bloom miss and produce a near-zero hit count, so this guards against silent encoding regressions.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -607,6 +607,8 @@ impl Blockchain {
         // Gated by `--no-bal-prefetch`: when the operator disables BAL-driven
         // prefetching, skip the synchronous storage warm too. The warmer thread
         // below already honors the same toggle.
+        // Flat-KV warm for execution (SLOAD and SSTORE original values). Uses the
+        // full BAL access set, since execution genuinely reads every accessed slot.
         #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
         if self.options.bal_prefetch_enabled
             && let Some(bal) = bal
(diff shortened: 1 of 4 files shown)


edg-l:

PR #6875, opened 2026-06-16
perf(l1): run BAL trie-node prefetch concurrently with execution

edg-l:
Stacked on #6873. Retargets to that branch once it merges.

#6873 added a BAL-driven prefetch that warms the MPT internal nodes the merkleizer walks. It ran synchronously before execution, so its entire cost sat on the critical path.

This moves the trie-node prefetch onto its own thread inside the `std::thread::scope` that already holds the executor and merkleizer, so it overlaps execution instead of preceding it.

**Why it is safe to overlap with execution:** the trie-node prefetch reads the trie-node column families, which execution never touches (execution reads only the flat-KV CFs via SLOAD/SSTORE). No contention with the executor. Against the merkleizer, which reads those same trie-node CFs, the prefetch acts as a high-queue-depth batch running in parallel, raising aggregate I/O queue depth so the scattered cold reads land warm before the merkleizer's per-account walk reaches them.

The flat-KV warm (introduced in earlier PRs) stays synchronous before execution, since it serves the executor and must be ready before it starts.

**Safety properties:**

- Best-effort and read-only. A spawn failure is logged via `debug!` and silently skipped (`inspect_err(...).ok()`). Any node the thread misses is simply cold-read by the merkle walk, with no effect on block execution output.
- The scope joins the handle before returning, so all borrows end cleanly. A thread panic is caught via `h.join()`, surfaced with `warn!`, and otherwise ignored.
- No deadlock path: the thread holds a shared ref to `self.storage`, does a single batch read, and returns. It acquires no locks the executor or merkleizer hold.

Gated to `feature = "rayon"` and `not(feature = "eip-8025")`, same as the rest of the BAL prefetch machinery. The 16,384-slot/account threshold that gates the trie prefetch is unchanged, so ordinary blocks see no effect.

EF blockchain LEVM suite: 8758 passed, 0 failed.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -619,38 +619,38 @@ impl Blockchain {
             }
         }
 
-        // Warm the MPT internal nodes the merkleizer will walk (the trie-node CFs,
-        // which execution never reads). Derived from the MERKLE WRITE SET, not the
-        // access set: the merkleizer only walks accounts/slots that actually change,
-        // and `optimistic_updates` (synthesize_bal_updates) already drops read-only
-        // accesses and read-only slots. Warming trie nodes for read-only accesses is
-        // wasted cold reads that compete with execution for the same device; gating
-        // on the access set regressed read-heavy blocks (value-0 CALLs to existing
-        // accounts, bloated SLOADs) that have little or no merkle work.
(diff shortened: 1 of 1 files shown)


edg-l:

PR #6876, opened 2026-06-16
fix(l1): validate block_access_list_hash on the parallel import path

edg-l:
## Motivation

EIP-7928 makes `block_access_list_hash` a block-validity condition: reject the block unless `header.block_access_list_hash == keccak(rlp(canonical_BAL))`.

The default **parallel** import path never checked it. The check is gated on the rebuilt BAL (`produced_bal`), but the parallel Amsterdam path drives execution from the header BAL and returns `produced_bal = None`, so it never fired — a block with a content-valid BAL but a forged commitment was accepted on the parallel path while the sequential/batch paths rejected it. Reachable via P2P full-sync's final-batch import; `engine_newPayloadV5` was already sound.

## Description

When `produced_bal` is `None` but a header BAL drove an Amsterdam block, compare `keccak(rlp(header_bal))` against `header.block_access_list_hash` via `matches_commitment`. Pure hash compare on the BAL already in memory — the no-rebuild parallel optimization is preserved.

## Test

`bal_hash_parallel_skip.rs`: forge only the header commitment, import both paths with the canonical BAL. Post-fix both reject with `BlockAccessListHashMismatch`; positive control confirms a valid block still imports on the parallel path.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -725,14 +725,26 @@ impl Blockchain {
                             &chain_config,
                             &execution_result.requests,
                         )?;
-                        // The parallel Amsterdam validation path uses the header BAL directly
-                        // to drive execution; it doesn't rebuild a BAL, so produced_bal is None.
-                        // BAL correctness on that path is enforced inside
-                        // execute_block_pipeline (header-BAL index/size/withdrawal-index
-                        // checks plus unread_storage_reads / unaccessed_pure_accounts).
-                        // The sequential Amsterdam path rebuilds a BAL and returns
-                        // Some(produced_bal), so the hash check below runs. Pre-Amsterdam
-                        // blocks never record a BAL, so produced_bal is None there too.
+                        // EIP-7928 block_access_list_hash commitment check.
+                        //
+                        // Sequential Amsterdam path: rebuilds a BAL and returns
+                        // Some(produced_bal), so the full hash+index+size check runs here.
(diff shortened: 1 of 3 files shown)


ilitteri (approved):

iovoid (approved):

crates/blockchain/blockchain.rs
```diff
@@ -725,21 +725,38 @@ impl Blockchain {
...
                         if let Some(bal) = &produced_bal {
                             validate_block_access_list_hash(
                                 &block.header,
                                 &chain_config,
                                 bal,
                                 block.body.transactions.len(),
                             )?;
+                        } else if let Some(header_bal) = bal
+                            && chain_config.is_amsterdam_activated(block.header.timestamp)
+                            && !header_bal.matches_commitment(block.header.block_access_list_hash)
```
ElFantasma, line 757: nit: `matches_commitment`'s docstring in `block_access_list.rs:546-550` still says it's used "to gate persisting/serving a BAL ... callers degrade to the `0x80` 'unavailable' sentinel on a `false` here." After this PR, the same `false` outcome here means *reject the block* — a strictly stronger consequence. Worth amending the docstring to mention both call sites (gate-for-serving vs gate-for-block-validity) so a future reader doesn't assume the only failure mode is the 0x80 degradation. Non-blocking, can be a follow-up.

ElFantasma (approved):
Approved, with a nit

PR #6877, opened 2026-06-16
fix(l1): improve witness generation spec conformance

iovoid:
**Motivation**

Currently we are failing several [witness generation tests](https://eth-act.github.io/eest-execution-witness-dashboard/#/group/execution-witness).

**Description**

Fixes the ordering by sorting, and the over-inclusion by:
- Disabling both BAL and speculative warming when an execution witness is requested
  - While disabling BAL warming isn't strictly needed, it's the simpler solution. There are several edge-cases around accounts created within the block that have to be handled carefully.
- Fixing over-accessing in OOG scenarios (call that runs out of gas right after finding out the target is a eip-7702 delegation, or paying the cold access cost, etc)

Also adds a test runner for the execution witness tests. You can run the tests with:

```
cd tooling/ef_tests/blockchain && make test-stateless
cd tooling/ef_tests/engine && make test
```

There are also hive [tests](https://github.com/eth-act/eest-execution-witness-dashboard) which take significantly longer to execute.


edg-l (approved):
Reviewed the diff (read-only, not run locally); CI gates green (EF Tests, Engine EF, Hive, Benchmarks, Build, Test). Looks sound.

- No normal-path perf regression: the warming / parallel-exec / parallel-trie disables are all gated behind `collect_witness` (the prover/stateless path), so normal block execution is unaffected.
- Consensus changes are spec-aligned edge cases: OOG-on-7702 (gas-check-before-delegate-read via `eip7702_peek_delegation`), EXTCODECOPY size=0 reads code, system calls skip the SYSTEM_ADDRESS sender path, coinbase read unconditionally on zero-fee user txs. Net state/gas on the normal path is unchanged; these move the access set / witness to match EELS.
- Code-cache eviction on revert (`inserted_code_hashes`) correctly tracks only newly-inserted hashes (`Entry::Vacant`), so it can't evict pre-existing codes.

Non-blocking heads-up for whoever merges this with the BAL-prefetch stack (`bal-devnet-7` / #6872-#6878): there's a conflict in the warmer region, and our synchronous flat-KV prefetch (`prefetch_storage`) will need `&& !collect_witness` added so it doesn't re-introduce witness over-inclusion (the trie-node and account prefetches are auto-disabled via `optimistic_updates`/the warmer; the synchronous flat-KV one is not). Tracked in #6879. Worth running the stateful/BAL EF suite after that merge, since this shifts the BAL access set.

ilitteri (approved):

crates/vm/levm/src/db/gen_db.rs
```diff
@@ -949,7 +949,16 @@ impl<'a> VM<'a> {
         let acc = self.get_account_mut(address)?;
         let code_hash = new_bytecode.hash;
         acc.info.code_hash = new_bytecode.hash;
-        self.db.codes.entry(code_hash).or_insert(new_bytecode);
+        if let Entry::Vacant(entry) = self.db.codes.entry(code_hash) {
```
ElFantasma, line 952: Worth a brief invariant comment near the `Entry::Vacant` guard explaining that the revert-eviction in `restore_cache_state` is correct *only* because we track exactly the hashes we just inserted (i.e. not ones that were already cached from a pre-existing account or an earlier same-block tx). A future refactor that flips this to unconditional `.or_insert()` + push would silently break the eviction: revert would drop an entry committed by a sibling/earlier frame, hiding a real store read on the next access. The behavior is correct as written — this is a doc-only ask.

crates/vm/levm/src/hooks/default_hook.rs
```diff
@@ -343,6 +365,15 @@ pub fn pay_coinbase(vm: &mut VM<'_>, gas_to_pay: u64) -> Result<(), VMError> {
     // Only pay coinbase if there's actually a fee to pay.
     if !coinbase_fee.is_zero() {
         vm.increase_account_balance(vm.env.coinbase, coinbase_fee)?;
+    } else if !vm.env.is_system_call {
```
ElFantasma, line 367: Perf observation, not a request for change: this adds one `get_account(coinbase)` cache lookup to every zero-fee non-system tx. On `eth_call` (gas_price=0) and any zero-base-fee L2 path that uses LEVM, that's per-call overhead the previous code avoided. The witness-correctness justification is sound (EELS reads coinbase unconditionally in user-tx fee transfer, so witness must record the trie path / exclusion proof), so this is the right tradeoff — flagging so it doesn't surprise anyone profiling `eth_call`-heavy workloads later.

ElFantasma (approved):

PR #6878, opened 2026-06-16
Shard the account-state batch prefetch read

edg-l:
Stacked on #6872. Retargets to main when that merges.

`get_account_states_batch_by_root` read its FKV-covered keys with a single `multi_get` over `ACCOUNT_FLATKEYVALUE`. With async_io off, that runs at queue depth 1, which on cold account-heavy blocks serializes thousands of random reads. The storage FKV batch already used sorted, parallel shards; the account path never got the same treatment.

Fix: sort the FKV keys before reading (adjacent accounts share RocksDB data blocks) then split them into contiguous shards read concurrently via `std::thread::scope`. MAX_SHARDS is 64, accounts are scattered and the batch is smaller than a bloated account's storage, so 64 is the sweet spot; 128 over-fragments.

`prefetch_accounts` in `CachingDatabase` now gates the same way `prefetch_storage` does: cold batches above a threshold go to the sharded batch, small or warm batches go to parallel point-gets. The gate counts missing (uncached) accounts, so warm blocks stay on the point-get path regardless of total account count.

Prefetch is best-effort cache warming and cannot change execution output. Tests in `test/tests/storage/storage_batch_tests.rs` assert the sharded batch is byte-identical to the per-account single-get path in both FKV-unswept (trie fallback) and FKV-swept (sharded multi_get) states, for both in-memory and RocksDB engines.

### DIFF
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -2734,12 +2734,40 @@ impl Store {
         }
 
         if !fkv_indices.is_empty() {
+            // Sort by the prefixed (hashed-order) key so adjacent accounts land
+            // in the same RocksDB data blocks, then read the sorted keys in
+            // contiguous shards concurrently. A single `multi_get` runs the whole
+            // batch at queue depth 1 (async_io is off), which on a cold account-
+            // heavy block serializes thousands of random reads (coldbench: ~13x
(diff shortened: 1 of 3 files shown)


edg-l:

PR #6880, opened 2026-06-16
fix(l1): bound eth_getLogs response size and cap the filter registry (configurable)

ilitteri:
## Motivation

Three `eth_*` JSON-RPC surfaces had no memory bound, all reachable **unauthenticated** on the default HTTP/WS port (`eth` is in `DEFAULT_HTTP_API`):

- `eth_getLogs` / `eth_getFilterChanges` resolved an attacker-chosen `fromBlock..=toBlock` and built the full `Vec<RpcLog>` (plus a second `serde_json::Value` copy) — no range or result cap.
- `eth_newFilter` inserted into a process-global map with no count cap.

Validated empirically on a mainnet node: a single ~300-byte request held the whole chain's logs in RAM, and 500,000 filters were accepted with zero refusals.

## Description

Adds a log-range cap, a log-count cap, and a filter-count cap, each **configurable via CLI** (defaults below; `0` disables). Defaults follow the cross-client consensus (reth/Nethermind/Erigon cap logs at 20,000; Besu/Erigon cap range; the filter-count cap mirrors the existing subscription total). The same limits are threaded into the L2 RPC server. When a cap is hit, the JSON-RPC error **names the flag** that raises or disables it.

### CLI configuration

| Flag | Default | Env var | `0` |
|------|---------|---------|-----|
| `--rpc.max-log-block-range` | 5000 | `ETHREX_RPC_MAX_LOG_BLOCK_RANGE` | unlimited |
| `--rpc.max-logs-per-response` | 20000 | `ETHREX_RPC_MAX_LOGS_PER_RESPONSE` | unlimited |
| `--rpc.max-active-filters` | 10000 | `ETHREX_RPC_MAX_ACTIVE_FILTERS` | unlimited |

## Validation (mainnet node, before → after)

| Exploit | Before | After (defaults) |
|---------|--------|------------------|
| `getLogs` span 6001 | builds full `Vec` | rejected: *"Block range too large… max is 5000"* (no DB access) |
| `getLogs` ~200k-log window | 167 MB / **+759 MB RSS** | rejected at 20,376 logs; **+11 MB RSS** |
| `getLogs` normal (4800 logs) | works | works (unchanged) |
| `getLogs` ×40 concurrent | **+4–6 GB/round** | **+312 MB** |
| filter flood | 500,000 / 0 refused | **10,000 then refused** |

CLI override confirmed live (e.g. `--rpc.max-active-filters 5` → 6th `eth_newFilter` refused), and each error message names its flag, e.g.:

> `Active filter limit reached (5); uninstall unused filters with eth_uninstallFilter, or raise/disable the limit with --rpc.max-active-filters (0 = unlimited)`

## How to test

- `cargo test -p ethrex-rpc` — 3 new unit tests (range rejection, range-boundary pass, filter-capacity refusal) plus the existing suite.
- Manual: start with `--rpc.max-active-filters 5` and create 6 filters via `eth_newFilter`; the 6th returns the limit error above. Likewise `--rpc.max-log-block-range`/`--rpc.max-logs-per-response` reject over-limit `eth_getLogs` queries and name the flag.

## Notes

- `eth_getFilterChanges` shares `fetch_logs_with_filter`, so it inherits the same range/result caps.
- Per-poll TTL refresh on filters is intentional/standard behavior across clients and is unchanged; the count cap is the DoS bound.
- Follow-up (separate): the public HTTP/WS router has no global concurrency limit / request timeout, which amplifies any per-request response builder — worth adding as a defense-in-depth layer.

### DIFF
--- a/crates/networking/rpc/eth/filter.rs
+++ b/crates/networking/rpc/eth/filter.rs
@@ -76,6 +76,7 @@ impl NewFilterRequest {
         &self,
         storage: ethrex_storage::Store,
         filters: ActiveFilters,
+        max_active_filters: usize,
     ) -> Result<serde_json::Value, crate::utils::RpcErr> {
         let from = self
             .request_data
@@ -106,9 +107,9 @@ impl NewFilterRequest {
         // Cap the global filter registry. HTTP filters are not connection-scoped,
         // so this is the only bound on how many a client can accumulate (a polled
         // filter never hits the inactivity TTL).
-        if active_filters_guard.len() >= MAX_ACTIVE_FILTERS {
+        if max_active_filters != 0 && active_filters_guard.len() >= max_active_filters {
             return Err(RpcErr::Internal(format!(
-                "Active filter limit reached ({MAX_ACTIVE_FILTERS}); uninstall unused filters with eth_uninstallFilter before creating new ones"
+                "Active filter limit reached ({max_active_filters}); uninstall unused filters with eth_uninstallFilter before creating new ones"
             )));
         }
         active_filters_guard.insert(
@@ -129,9 +130,10 @@ impl NewFilterRequest {
         req: &RpcRequest,
         storage: Store,
         state: ActiveFilters,
+        max_active_filters: usize,
     ) -> Result<Value, RpcErr> {
         let request = Self::parse(&req.params)?;
-        request.handle(storage, state).await
+        request.handle(storage, state, max_active_filters).await
     }
 }
 
@@ -201,6 +203,8 @@ impl FilterChangesRequest {
         &self,
         storage: ethrex_storage::Store,
         filters: ActiveFilters,
+        max_block_range: u64,
+        max_logs_per_response: usize,
     ) -> Result<serde_json::Value, crate::utils::RpcErr> {
         let latest_block_num = storage.get_latest_block_number().await?;
         // Box needed to keep the future Sync
@@ -237,7 +241,13 @@ impl FilterChangesRequest {
                 // Drop the lock early to process this filter's query
                 // and not keep the lock more than we should.
                 drop(active_filters_guard);
-                let logs = fetch_logs_with_filter(&filter.filter_data, storage).await?;
+                let logs = fetch_logs_with_filter(
+                    &filter.filter_data,
+                    storage,
+                    max_block_range,
+                    max_logs_per_response,
+                )
+                .await?;
                 serde_json::to_value(logs).map_err(|error| {
                     tracing::error!("Log filtering request failed with: {error}");
                     RpcErr::Internal("Failed to filter logs".to_string())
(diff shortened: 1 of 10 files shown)


edg-l (approved):

crates/networking/rpc/eth/filter.rs
```diff
@@ -95,6 +104,14 @@ impl NewFilterRequest {
             filters.clear_poison();
             poisoned_guard.into_inner()
         });
+        // Cap the global filter registry. HTTP filters are not connection-scoped,
+        // so this is the only bound on how many a client can accumulate (a polled
+        // filter never hits the inactivity TTL).
+        if max_active_filters != 0 && active_filters_guard.len() >= max_active_filters {
+            return Err(RpcErr::Internal(format!(
```
ElFantasma, line 111: Type choice: `RpcErr::Internal` reads as a server-side bug ("internal error"), but this is an actionable client-facing condition — the message even tells the caller what to do (`eth_uninstallFilter` or raise the limit). The other two new caps in this PR (block range, logs per response) both return `RpcErr::BadParams` with similarly user-actionable messages. Worth aligning to `BadParams` (or introducing a `ResourceExhausted` variant if you want to distinguish; JSON-RPC server-implementation-defined codes are -32000..-32099 territory). Same applies to the matching `matches!(err, RpcErr::Internal(_))` assertion in the test.

crates/networking/rpc/eth/logs.rs
```diff
@@ -193,6 +224,15 @@ pub(crate) async fn fetch_logs_with_filter(
                 }
             }
         }
+        // Bound peak memory: stop accumulating once the (address-filtered)
+        // result set exceeds the cap and return a paginatable error pointing
+        // at the block reached, so the caller can split the query.
+        if max_logs_per_response != 0 && logs.len() > max_logs_per_response {
```
iovoid, line 230: Considering the CLI option says "per response", we might want to check the count after topic filters are also applied.

crates/networking/rpc/eth/filter.rs
```diff
@@ -95,6 +104,14 @@ impl NewFilterRequest {
...
+                "Active filter limit reached ({max_active_filters}); uninstall unused filters with eth_uninstallFilter, or raise/disable the limit with --rpc.max-active-filters (0 = unlimited)"
```
iovoid, line 112: Should probably return BadParams or similar since this is the client's fault.

crates/networking/rpc/eth/logs.rs
iovoid, line 230: Should be >= instead of >

crates/networking/rpc/eth/filter.rs
```diff
@@ -221,7 +241,13 @@ impl FilterChangesRequest {
...
                 // and not keep the lock more than we should.
                 drop(active_filters_guard);
-                let logs = fetch_logs_with_filter(&filter.filter_data, storage).await?;
+                let logs = fetch_logs_with_filter(
+                    &filter.filter_data,
+                    storage,
+                    max_block_range,
+                    max_logs_per_response,
+                )
+                .await?;
```
iovoid, line 250: If this fails due to the range being too large, events are lost.

PR #6883, opened 2026-06-17
docs(l1,l2): add custom release notes step

iovoid:
**Motivation**

In the last releases we've been adding manual release notes highlighting key details. This practice should be documented.

**Description**

Adds a step to the release process mentioning custom release notes.

### DIFF
--- a/docs/developers/release-process.md
+++ b/docs/developers/release-process.md
@@ -90,7 +90,16 @@ Once the pre-release is created and you want to publish the release, go to the [
 
     ![edit title](../img/publish_release_step_3.png)
 
-4. Set the release as the latest release (you will need to uncheck the pre-release first). And finally, click on `Update release`
+4. Customize the release notes.
+
+    The auto-generated changelog lists every commit, but it doesn't tell users what actually matters in this release. At the start of the notes, add a section with hand-made notes.
+    - If the release contains critical security fixes, start with an `> [!IMPORTANT]` block urging users to upgrade.
+    - The rest of the custom notes should be in a `> [!NOTE]` block that:
+      - Has a **What's new** list highlighting the key changes (new features, important fixes).
+      - Mentions, if there's a migration, that users won't be able to roll back to a previous version.
+      - Mentions whether a resync is needed.
+
+5. Set the release as the latest release (you will need to uncheck the pre-release first). And finally, click on `Update release`
 
     ![set latest release](../img/publish_release_step_4.png)
 
@@ -155,7 +164,7 @@ If hotfixes are needed before the final release, commit them to `release/vX.Y.Z`
 
 ### Failure on "latest release" workflow
 
-If the CI fails when setting a release as latest (step 4), Docker tags `latest` and `l2` may not be updated. To manually push those changes, follow these steps:
+If the CI fails when setting a release as latest (step 5), Docker tags `latest` and `l2` may not be updated. To manually push those changes, follow these steps:
 
 - Create a new Github Personal Access Token (PAT) from the [settings](https://github.com/settings/tokens/new).
 - Check `write:packages` permission (this will auto-check `repo` permissions too), give a name and a short expiration time.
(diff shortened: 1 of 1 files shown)


edg-l (approved):

MegaRedHand (approved):
LGTM

ilitteri:
Nice addition — this closes a real gap. One suggestion to make it a reusable recipe rather than prose that drifts (the last few releases already diverged on `>[!NOTE]` vs `> [!NOTE]` and on where the migration note lands).

Three tweaks:

1. **Give `[!WARNING]` a role.** An irreversible migration ("can't roll back") is a *risk of a negative outcome* — GitHub's own definition of `[!WARNING]`/`[!CAUTION]`. Tucking it into a `[!NOTE]` bullet is how v16.0.0's rollback warning ended up buried mid-list. It deserves its own box.
2. **Pin the syntax.** GitHub requires the space (`> [!NOTE]`, not `>[!NOTE]`) and a blank line between boxes — this also matches the rest of `/docs`, which uses the spaced form everywhere.
3. **Add a copy-paste template** so the format stops drifting release to release.

Map the boxes to operator decisions, in order:
- `> [!IMPORTANT]` — only for critical security/correctness fixes; one line urging upgrade.
- `> [!WARNING]` — only when the upgrade can't be rolled back or needs action (schema migration, required resync, changed flag).
- `> [!NOTE]` — always; **What's new** highlights + whether a resync is needed.

That keeps it to ≤2 boxes in the common case, matching GitHub's "one or two per article" guidance. Suggested step 4 body:

````markdown
4. Customize the release notes.

    The auto-generated changelog lists every commit, but it doesn't tell operators what actually matters in this release. Above the auto-generated changelog, add a hand-written summary using [GitHub alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts). Pick boxes by what the operator needs to decide, in this order:

    - `> [!IMPORTANT]` — **only** when the release carries critical security or correctness fixes. One line stating that upgrading is strongly recommended for all operators.
    - `> [!WARNING]` — **only** when the upgrade can't be cleanly undone or needs operator action (a database schema migration you can't roll back from, a required resync, a changed flag). State what changes and what the operator must do.
    - `> [!NOTE]` — **always**. A **What's new** list of the highlights (new features, important fixes), plus a line saying whether a resync is needed (if not already covered above).

    Keep the space after `>` (`> [!NOTE]`, not `>[!NOTE]`) and leave a blank line between boxes so each renders separately. Drop the `[!IMPORTANT]` / `[!WARNING]` boxes when they don't apply — a routine release needs only the `[!NOTE]`.

    ```markdown
    > [!IMPORTANT]
    > This release contains critical fixes. Upgrading is strongly recommended for all operators.

    > [!WARNING]
    > This release changes the database schema; once you upgrade you can't roll back to a previous version. The migration runs automatically.

    > [!NOTE]
    > **What's new**
    > - <highlight>
    > - <highlight>
    >
    > No resync is needed.
    ```
````

ElFantasma (approved):
One minor consistency observation: in the worked example, the `[!WARNING]` block says "This release changes the database schema; once you upgrade you can't roll back" and the `[!NOTE]` block immediately below says "No resync is needed." Both can technically be true (auto-migration ≠ resync) but for an example, the juxtaposition is a little jarring — an operator scanning the snippet may wonder which signal to act on. A small tweak that would read cleaner:

- Drop the `No resync is needed.` line from the example (the rule earlier already says the resync line is conditional on "if not already covered above", and the schema-migration WARNING arguably covers it).
- Or, swap the WARNING for a non-resync-adjacent example, e.g. a removed CLI flag, so the NOTE's resync line stands on its own without competition.

Either way the surrounding prose is unambiguous; this is just sanding the example down. LGTM either way.

ilitteri (approved):

PR #6884, opened 2026-06-17
perf(l1): inverted address log index to narrow eth_getLogs candidate blocks

ElFantasma:
**Motivation**

`eth_getLogs` was 175–340× slower than geth (issue #6785). Two prior steps closed most of the constant-factor gap but left the algorithmic one:

- #6813 — header-bloom prefilter (skip blocks whose header bloom can't match).
- #6852 — bulk per-block receipt reads (kill the per-tx receipt point-lookup).

Both stay **O(blocks-in-range)**: every block in the range is still visited. For popular signatures (e.g. the ERC-20 `Transfer` topic) the 2048-bit header bloom **saturates** — nearly every block looks like a candidate — so the prefilter can't narrow the range. The remaining gap to indexed clients is **stage-1 candidate selection**, which needs a real index.

**Description**

Adds an inverted **address → blocks** log index and uses it to narrow `eth_getLogs` to candidate blocks.

- **New `log_index` storage module + `LOG_ADDRESS_INDEX` column family.** Key `address (20B) || section (8B BE)`, value = big-endian `u16` in-section block offsets where that address emitted a log. Blocks are grouped into fixed `SECTION_SIZE = 4096`-block sections.
- **Background indexer** (`run_log_indexer`, spawned in `init_l1`). Polls periodically and indexes any section once it is buried `CONFIRMATION_DEPTH = 256` blocks — well clear of the 128-block reorg limit, so indexed sections are immutable and need **no reorg invalidation**. Runs off the block-import path (`spawn_blocking`); backfills on first run, then moves forward. Never touches the hot write path.
- **Query path** (`fetch_logs_with_filter`): when the filter has an address and the range is covered by the index, narrow to candidate blocks via `get_candidate_blocks_by_address`; otherwise fall back to the full-range scan (the header-bloom prefilter still applies). Every visited block is still exact-filtered, so extra or missing index candidates can only ever cost a little time — **results are identical**.
- **No migration / no `STORE_SCHEMA_VERSION` bump.** The CF is registered in `TABLES` and auto-creates; on a DB without the index, or for a range that isn't indexed yet, the query returns `None` and falls back to the existing scan. The most recent <4096 blocks (the current, not-yet-buried section) always serve from the scan path until they're indexed (minutes).

Part of #6785. Stacks on #6852 → #6813; will retarget to `main` as those merge.

**Benchmarks**

`ethrex-mainnet-5`, fully-synced mainnet, LUSD `Transfer` over recent ranges, single-call median latency (same harness/methodology as the earlier runs).

| range | main baseline | #6813 prefilter | #6852 bulk receipts | **this PR (indexed)** | geth (ref) |
|---|---|---|---|---|---|
| 100 blk | 618 ms | 415 ms | 80 ms | **0.3 ms** | 2.5 ms |
| 1,000 blk | 6,131 ms | 3,529 ms | 613 ms | **14.4 ms** | 35 ms |
| 10,000 blk | 70,441 ms | 31,857 ms | 4,272 ms | **195.7 ms** | 204 ms |

- ~270× / 43× / 22× vs the previous best (bulk receipts); ~360–2000× vs the original baseline.
- **First time ethrex `eth_getLogs` is competitive with an indexed client** — faster than geth at 100 & 1,000 blocks, on par at 10,000.

Under load (vegeta rate-sweep, indexed ranges), `eth_getLogs` previously **collapsed** (30 s timeouts at 100 rps; 1k/10k saturated even at 10 rps). Now **100% success, no timeouts**: 100-blk sustains the full 1,000 rps; 1,000-blk ~400 rps; 10,000-blk ~40 rps (the remaining ceiling is CPU-bound stage-2 receipt reads on the matched blocks).

**Checklist**

- [x] Background indexer runs off the block-import path; only finalized (256-block-buried) sections are indexed → no reorg invalidation.
- [x] Correctness preserved: every visited block is exact-filtered; index only selects candidates, with full-scan fallback.
- [x] No `Store` re-sync required: new CF auto-creates; absence / unindexed range → scan fallback.


crates/storage/store.rs
```diff
@@ -1264,6 +1269,154 @@ impl Store {
...
+                        }
+                    }
+                }
+            }
+            let entries = log_index::build_section_entries(section, address_offsets);
+            self.write_batch(LOG_ADDRESS_INDEX, entries)?;
+            self.set_indexed_log_sections(section + 1)?;
+            section += 1;
+            newly_indexed += 1;
+        }
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 9d2111315. A missing canonical hash for a buried block now makes the indexer **defer the section** — it breaks out of the loop, logs a `warn!`, and leaves that section (and everything after it) unindexed rather than propagating an error. Queries fall back to scanning the un-indexed range, so results stay correct; the next poll retries in case it was transient. The section is never marked indexed unless it was built in full, so we can't silently drop a block's logs (which would be a false negative on the indexed path).

crates/storage/store.rs
```diff
@@ -1264,6 +1269,154 @@ impl Store {
...
+                let offset = log_index::offset_in_section(block_number);
+                // Each address contributes this block's offset at most once.
+                let mut seen = HashSet::new();
+                for receipt in &receipts {
+                    for log in &receipt.logs {
+                        if seen.insert(log.address) {
+                            address_offsets.entry(log.address).or_default().push(offset);
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 9d2111315 — added `if !receipt.succeeded { continue; }` so the indexer mirrors the query path exactly. As you note it's a no-op on EIP-658 networks (reverted txs carry no logs), but it keeps the index exact everywhere and makes the docstring's "no false positives" claim hold unconditionally.

crates/storage/store.rs
```diff
@@ -1264,6 +1269,154 @@ impl Store {
...
+            }
+            let entries = log_index::build_section_entries(section, address_offsets);
+            self.write_batch(LOG_ADDRESS_INDEX, entries)?;
+            self.set_indexed_log_sections(section + 1)?;
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 9d2111315. Section entries and the counter now commit in a **single write transaction** (`commit_indexed_log_section`): one `begin_write` → `put_batch(LOG_ADDRESS_INDEX, …)` + `put(MISC_VALUES, counter)` → `commit`. RocksDB write batches span column families, so the two writes are now atomic and there's no crash window at all.

ElFantasma:
Thanks for the reviews. Addressed in `9d2111315` (on top of the index commits):

**Fixed**
- **Indexer stall on missing canonical hash** (greptile P1 / Claude #2) — now defers the section gracefully instead of erroring; unindexed ranges fall back to scan, so results stay correct.
- **`decode_offsets` silent truncation** (Kimi #1 / Claude #3) — now returns an error on an odd-length value instead of dropping the trailing byte; added a unit test.
- **Non-atomic section commit** (greptile P2 / Claude #1) — entries + counter now commit in a single cross-CF write transaction.
- **Missing `receipt.succeeded` filter** (greptile P2) — indexer now skips reverted receipts, matching the query path.
- **Overflow in `last_block + confirmation_depth`** (Kimi #3) — switched to `saturating_add`.
- **Block-0 / underflow reasoning** (Claude design note) — added a comment explaining why `indexed_until - 1` can't underflow.

**Acknowledged, not changing**
- **New read txn per block in `get_receipts_for_block_sync`** (Kimi #2) — backfill is one-time and steady-state is one section per 12 s poll, off the hot path, so this hasn't shown up as a cost; can batch a section's reads into one txn later if profiling says otherwise.
- **N×M index reads for large address lists** (Claude) — fine for typical small address sets; pathological cases still get the correct (if unoptimized) scan fallback.
- **Defensive `starts_with(prefix)` guard** (Claude) — intentional belt-and-suspenders around the prefix iterator; harmless.
- **`CONFIRMATION_DEPTH` location** (Claude) — kept in `initializers.rs` alongside the poll cadence since it's indexer-task policy, whereas `SECTION_SIZE` is the on-disk layout constant in `log_index`.

**Test coverage** (Claude) — the pure helpers are unit-tested and the storage methods are exercised end-to-end by the mainnet benchmark (6184 sections indexed, query results verified). A Store-level integration test for the partial-coverage (`from < indexed_until <= to`) path needs a block+receipt fixture harness that doesn't exist in this crate yet; happy to add it as a follow-up if you'd like it in this PR.

crates/networking/rpc/eth/logs.rs
```diff
@@ -143,12 +143,24 @@ pub(crate) async fn fetch_logs_with_filter(
...
+        };
+
     let mut logs: Vec<RpcLog> = Vec::new();
-    // The idea here is to fetch every log and filter by address, if given.
-    // For that, we'll need each block in range, and its transactions,
-    // and for each transaction, we'll need its receipts, which
-    // contain the actual logs we want.
-    for block_num in from..=to {
+    // For each candidate block, load its receipts (bulk) and collect matching logs.
+    for block_num in block_numbers {
```
iovoid, line 164: This hides the error in the case where `from` is lower than the earliest block we have. Previously we would have tried to obtain the body and failed doing so.

iovoid (changes requested):

crates/networking/rpc/eth/logs.rs
ElFantasma: Fixed in 8ec984293. The index now records the first section it actually covers (`log_index_start_section`), and `get_candidate_blocks_by_address` returns `None` (→ full scan) when the query's `from` is below that covered start. So a range beginning below the node's retained history takes the original scan path — surfacing the missing-body error as before — instead of the index silently reporting no logs for it.

crates/storage/store.rs
```diff
@@ -1261,6 +1273,218 @@ impl Store {
...
+                        if seen.insert(log.address) {
+                            address_offsets.entry(log.address).or_default().push(offset);
+                        }
+                    }
+                }
+            }
+            if !section_complete {
+                break;
+            }
+            // The first section that carries any logs marks where the index's
```
iovoid: The first section will contain the snapsync pivot, which means we won't have data for the full section and incorrectly return empty when the truth is we didn't even have the data.

crates/storage/store.rs
```diff
@@ -1261,6 +1273,218 @@ impl Store {
...
+                    let offsets = log_index::decode_offsets(&bytes)?;
+                    candidates.extend(log_index::offsets_to_blocks(
+                        section, &offsets, from, indexed_to,
+                    ));
+                }
+            }
+        }
+        // Unindexed tail of the range: include every block so it gets scanned.
+        if to >= indexed_until {
+            candidates.extend(from.max(indexed_until)..=to);
```
iovoid: This can loads lots of data into memory on nodes that have lots of unindexed data (any node upgrading to the latest version including this change!).

iovoid (changes requested):

crates/storage/store.rs
ElFantasma: Good catch. Fixed in 808d1b25c: the covered start is now the section *after* the first data-bearing one, which is guaranteed fully post-pivot, so the pivot-straddling section is never claimed as covered — queries into it fall back to a full scan instead of reporting its below-pivot blocks as having no logs. (There's no persisted pivot block number to anchor on more precisely; `get_earliest_block_number` is hardcoded to genesis.) Costs at most one section (~4096 blocks) of acceleration at the oldest edge of the retained range.

crates/storage/store.rs
ElFantasma: Fixed in 808d1b25c. `get_candidate_blocks_by_address` now returns a `Box<dyn Iterator<Item = BlockNumber>>` where the unindexed tail is a lazy `indexed_until..=to` range chained onto the exact indexed candidates — it's never collected into a Vec. So a wide query on a node with lots of unindexed data (mid-backfill after upgrade) no longer materializes millions of block numbers; only the (small, for sparse addresses) indexed candidate set is held. Candidates are all `< indexed_until <= tail`, so ascending block order is preserved.

PR #6886, opened 2026-06-17
docs(l1,l2): add release testing checklist

iovoid:
**Motivation**

Currently we are following an internal checklist for testing releases, but this isn't documented anywhere.

**Description**

Documents the test checklist we use.

### DIFF
--- a/docs/developers/release-process.md
+++ b/docs/developers/release-process.md
@@ -76,6 +76,19 @@ A changelog will be generated based on commit names (using conventional commits)
 
 ## 4th - Test & Publish Release
 
+### Testing checklist
+
+Before publishing the release, run through the following checks using the pre-release binaries:
+
+- [ ] Upgrade `ethrex-ethdocker-mainnet`
+- [ ] Upgrade `ethrex-mainnet-1`
+- [ ] Upgrade `ethrex-minimum-mainnet`
+- [ ] Launch multisync on `ethrex-multisync-main`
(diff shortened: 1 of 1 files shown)


MegaRedHand (approved):

edg-l (approved):

ElFantasma (approved):
One optional thought: each item is an action without a stated success criterion. For someone running the list for the first time (or after a long gap), "Upgrade `ethrex-ethdocker-mainnet`" leaves "and then what" implicit. If you want to make this self-contained, the lightest version would be one trailing clause per row, e.g.:

```markdown
- [ ] Upgrade `ethrex-ethdocker-mainnet` — node restarts cleanly, resumes syncing, no panics/errors in logs
- [ ] Upgrade `ethrex-mainnet-1` — node restarts cleanly, head advances
- [ ] Upgrade `ethrex-minimum-mainnet` — node still fits the hardware envelope, stays at head
- [ ] Launch multisync on `ethrex-multisync-main` — completes a full snap sync, post-sync trie validation passes
- [ ] Upgrade a local L2 created with the previous version and run the integration tests — all pass against the upgraded node
- [ ] Run the L2 integration tests with a SP1 prover on the GPU server (`l2-gpu`) — proofs verify
```

Not a blocker — the current checklist is already strictly more documented than "it lives in someone's head." Just a question of whether the next release-runner should be able to operate it without DMing the previous one.

PR #6892, opened 2026-06-18
fix(l1): align mempool intrinsic-gas and EIP-7702 auth-list validation with execution

### LINKED ISSUE #6889: mempool: align intrinsic-gas validation with execution (EIP-7702 auth-list + EIP-7623 floor)
## Motivation

The mempool's `transaction_intrinsic_gas` (`crates/blockchain/mempool.rs`) computes Berlin-era intrinsic gas and does not match what LEVM enforces at execution. Transactions that LEVM will reject at inclusion are therefore admitted to the pool, wasting payload-builder cycles and able to crowd out valid txs. geth rejects both cases at admission.

This is **not** a consensus issue — LEVM rejects these txs at inclusion — it's purely a mempool-fidelity / cross-client parity gap.

## Description

Two gaps, both in mempool admission only:

1. **EIP-7702 auth-list cost** — `transaction_intrinsic_gas` never reads `tx.authorization_list()`, so it omits `PER_EMPTY_ACCOUNT_COST` per tuple. `validate_transaction` also never rejects an empty type-4 auth list. LEVM enforces both (`get_intrinsic_gas` / `validate_type_4_tx`).
2. **EIP-7623 calldata floor** — intrinsic gas omits the Prague calldata floor (10 gas/token), so an under-floor Prague tx is admitted that LEVM rejects with `IntrinsicGasBelowFloorGasCost`.

## Proposed work

- Add auth-list cost + empty-auth-list rejection to mempool intrinsic-gas validation.
- Apply the EIP-7623 calldata floor (Prague+) in the same helper.
- Tests: admit/reject parity with LEVM for type-4 and high-calldata txs.

## Checklist

- [ ] EIP-7702 auth-list cost + empty-auth-list rejection at admission
- [ ] EIP-7623 calldata floor (Prague+) in `transaction_intrinsic_gas`
- [ ] Tests for auth-list cost + EIP-7623 floor parity with LEVM

ElFantasma:
**Motivation**

On Prague (the active mainnet fork) the mempool's intrinsic-gas check did not match what LEVM enforces at execution, so transactions the VM will reject at inclusion were admitted to the pool — wasting payload-builder cycles and crowding out valid txs. Two gaps (EF findings `eip7702-mempool-intrinsic-gas-gap`, `mempool-calldata-floor-gas-gap`):

- the per-EIP-7702 authorization-tuple cost (`PER_EMPTY_ACCOUNT_COST`) was omitted;
- the EIP-7623 calldata floor was not applied.

Additionally, an EIP-7702 transaction with an empty `authorization_list` (invalid per the EIP, rejected by LEVM in `validate_type_4_tx`) was admitted.

None of these are consensus issues — LEVM independently rejects all three at inclusion — but they let invalid txs pollute the pool, diverging from geth/reth which reject them at admission.

**Description**

- `transaction_intrinsic_gas` now delegates to the VM's own helpers (`intrinsic_gas_dimensions` + `intrinsic_gas_floor`) for every fork instead of a parallel hand-rolled computation. This picks up the EIP-7702 auth-list cost and the EIP-7623 calldata floor automatically, and removes a drift-prone duplicate implementation. The floor is gated to `fork >= Fork::Prague`, matching the VM's default-hook gating.
- `validate_transaction` rejects type-4 txs with an empty `authorization_list` via the new `MempoolError::EmptyAuthorizationList`.
- Tests: added parity/regression tests for the auth-list cost, the EIP-7623 floor, and the empty-auth-list rejection. The pre-Istanbul calldata test was removed (unreachable: `get_fork()` floors at Paris, so EIP-2028 is always in effect, making it a duplicate of the post-Istanbul test); the two Shanghai create tests were corrected to EIP-2028 pricing.

Resolves #6889

**Checklist**

- [x] Added regression tests (intrinsic auth-list cost, EIP-7623 floor, empty-auth rejection)
- [x] `cargo fmt` + `cargo clippy` clean on the touched crates
- [ ] No `Store` changes (no `STORE_SCHEMA_VERSION` bump needed)


ilitteri (approved):

ElFantasma:
Thanks for the reviews. Follow-up commits `8db215ad4` and `579f453be` address the actionable points; summary below.

### Addressed

**Empty-auth check ordering** (Claude, Greptile P2) — moved the EIP-7702 type-4 structural validation ahead of the intrinsic-gas check, so an empty-auth tx now reports `EmptyAuthorizationList` rather than the downstream `TxIntrinsicGasCostAboveLimitError`, matching LEVM's `validate_type_4_tx` ordering. Regression test added.

**Pre-Prague type-4 gate** (Codex #2, Claude #3, Kimi #3) — real gap, good catch. ethrex supports pre-Prague forks (Paris/Shanghai/Cancun), and the mempool was admitting type-4 txs that LEVM rejects at inclusion with `Type4TxPreFork`. Added a `MempoolError::Eip7702TxPreFork` rejection gated on `is_prague_activated`, with a test.

**Lossy intrinsic-gas error mapping** (Claude #5) — the helper failures now map to a new `IntrinsicGasError(String)` that preserves the underlying `VMError` cause instead of collapsing to `TxGasOverflowError`. While there, I also corrected a pre-existing swap in two `MempoolError` Display messages: `TxGasOverflowError` and `TxTipAboveFeeCapError` carried each other's text, so RPC users hit the wrong message on those two rejections.

**Cosmetics** — `floor` → `calldata_floor` (Kimi); updated the now-stale "non-Amsterdam path" doc comment on `prague_config_and_header` (Claude #4).

### False positive

**Pre-Istanbul calldata pricing "regression"** (Codex #1, Claude #1, Kimi #2) — not a regression. `ChainConfig::get_fork()` floors at `Fork::Paris` (there are no sub-Paris fork variants, and the block-import runner rejects pre-Merge forks), so EIP-2028 (16 gas/non-zero byte) is always in effect on any reachable ethrex config — the old 68-gas/byte branch was dead code. The two constants are genuinely distinct (`TX_DATA_NON_ZERO_GAS = 68`, `TX_DATA_NON_ZERO_GAS_EIP2028 = 16`), so the updated test expectations reflect a real value change (e.g. `74556 → 58072`), not a same-value rename — they align the tests with what LEVM actually charges rather than masking anything. The removed `pre_istanbul` test was an exact duplicate of `post_istanbul` once both resolve to Paris.

### Not changed

**`EIP-7981` reference in the new comment** (Kimi #1) — kept intentionally. It mirrors the terminology LEVM already uses in `gas_cost.rs` / `utils.rs` for the Amsterdam access-list-bytes-into-floor rule (`floor_tokens_in_access_list`). If that EIP number is inaccurate it's inaccurate at the source too and should be fixed there in a dedicated change rather than diverging this comment.

**Genesis-setup verbosity in the async test** (Claude) — left as-is; it follows the existing pattern in `l1_tx_type_tests.rs`. A shared helper would be a reasonable cleanup but is out of scope here.

All mempool tests green (32), clippy + fmt clean.

iovoid (approved):

MegaRedHand (approved):

PR #6895, opened 2026-06-19
perf(l1): precompute eth_getLogs bloom inputs once instead of per-block

### LINKED ISSUE #6893: perf(l1,rpc): precompute eth_getLogs bloom inputs once instead of per-block
## Motivation

In the `eth_getLogs` header-bloom prefilter (added in #6813), `block_bloom_matches` is called once per block over the query range. For each address/topic it calls `bloom.contains_input(BloomInput::Raw(x.as_bytes()))`, which internally does `self.contains_bloom(&Bloom::from(input))` — i.e. it **re-derives the input's bloom bits (a keccak + bit extraction) on every block**.

Since the requested addresses and topics are constant for the whole query, this derivation is recomputed once per block unnecessarily. For wide-range queries that skip most blocks — exactly the case the prefilter targets — the bloom check is the *only* per-block work, so a query spanning ~1M blocks performs ~1M redundant keccaks (hundreds of ms of pure hashing).

Raised by @iovoid in review of #6813:
> This could be made more efficient by constructing the filter once and then calling `Bloom::contains_bloom`. This is what `Bloom::contains_input` does internally.

## Description

Hoist the bloom-input derivation out of the `from..=to` loop:

- Before the loop, precompute `Bloom::from(BloomInput::Raw(..))` once for each requested address and for each constrained topic position (preserving the `None`/empty-list position-wildcard semantics — a wildcard position is skipped, otherwise OR over the concrete alternatives).
- In `block_bloom_matches`, replace the per-block `contains_input(..)` calls with `bloom.contains_bloom(&precomputed)` (cheap bit-subset check, no hashing).

Net effect: O(addresses + topics) keccaks per *query* instead of per *block*; everything else unchanged. Behaviour-identical — the existing `bloom_match_*` unit tests should still pass (with constructors updated to the precomputed form).

Fast-follow on the #6813 / #6852 / inverted-index getLogs stack to keep those PRs' existing approvals intact.

## Checklist

- [ ] Precompute address/topic blooms once before the block loop
- [ ] Switch `block_bloom_matches` to `contains_bloom`
- [ ] Update `bloom_match_*` tests

ElFantasma:
## Motivation

In the `eth_getLogs` header-bloom prefilter, the per-block check re-derived each requested address's and topic's bloom bits **on every block**: `Bloom::contains_input(BloomInput::Raw(..))` internally does `Bloom::from(input)` (a keccak + bit extraction) before the subset test. Since the filter is constant for the whole query, that derivation was recomputed once per block — and for wide-range queries that skip most blocks, the bloom check is the *only* per-block work, so a ~1M-block sparse query did ~1M redundant keccaks.

Raised by @iovoid in review of #6813. Closes #6893.

## Description

- Introduce `BloomFilterMatcher`, built once before the block loop, holding each address and each constrained topic position pre-derived into `Bloom`s (a topic position is a wildcard, or an OR over its concrete alternatives — preserving the `None`/empty-list wildcard semantics).
- The per-block check is now `block_bloom.contains_bloom(&precomputed)` — a cheap bit-subset test with no hashing.
- Behaviour-identical: the existing `bloom_match_*` unit tests pass unchanged (rewired through a `bloom_matches` helper that builds the matcher).

## Checklist

- [x] Precompute address/topic blooms once before the block loop
- [x] Per-block check uses `contains_bloom`
- [x] `bloom_match_*` tests pass unchanged


crates/networking/rpc/eth/logs.rs
```diff
@@ -242,45 +246,86 @@ pub(crate) async fn fetch_logs_with_filter(
...
-        && !address_filter
+/// `Bloom::contains_input` internally builds a `Bloom` from the input (a keccak
+/// hash plus bit extraction) before testing it; for wide-range queries that skip
+/// most blocks that derivation dominates the per-block cost. We do it once here.
+struct BloomFilterMatcher {
+    /// One bloom per requested address; empty means no address constraint.
+    addresses: Vec<Bloom>,
+    /// Per constrained topic position: `None` is a wildcard (no constraint),
+    /// `Some(blooms)` requires at least one of its alternatives to be present.
+    topic_positions: Vec<Option<Vec<Bloom>>>,
```
iovoid: Can we get rid of the Option?

ElFantasma: Done in ae3f2b38a. Since a wildcard position imposes no constraint, I drop it during construction instead of storing `None` — the field is now `Vec<Vec<Bloom>>` (one entry per *constrained* position) and `matches` is just `all(positions).any(alternatives)`. The header bloom is position-agnostic, so omitting unconstrained positions is behavior-identical (the `bloom_match_*` tests pass unchanged).

iovoid (approved):

edg-l (approved):

ilitteri (approved):

PR #6900, opened 2026-06-22
fix(l1): stop unbounded P2P connection memory growth (bounded mailbox, lifecycle timeouts, admission cap, map pruning)

### LINKED ISSUE #6340: Memory leak
Hey ethrex team, we noticed a pretty significant memory leak on ethrex v9.0.0 . Over the last month its memory usage increased from about 10GB to 47GB (this is a mainnet node).

Image

Let me know if you're seeing the same or if there's anything else I can provide.

ilitteri:
> Closes #6340 

**Motivation**

A long-running node's RSS grows monotonically over time and never frees, even under light, steady traffic. Heap profiling (jemalloc `inuse_space` diff over many hours) localizes the growth to the **P2P layer** — per-connection actors (`spawned_concurrency` `MessageEnvelope` / `spawn_listener`), `rlpx` decode, and `tx_broadcaster` — not RPC or the state trie. It is per-time (transaction gossip + connection churn), not per-request. Confirmed by an idle-drop: with RPC fully off for ~20 h, in-use heap still climbed ~1.5 GB/day and established connections drifted 347 → 432.

The leak is several reinforcing issues in the P2P connection layer: each connection is a `PeerConnectionServer` actor with an **unbounded mailbox**; the actor drains serially by awaiting `state.sink.send(message)`, so a slow/half-dead peer blocks it while fixed-rate producers keep enqueuing; the handshake had no timeout and pongs were ignored, so stalled/dead peers parked actors + sockets; inbound connections were accepted with no cap; and per-peer index/session maps were never pruned. The net effect is monotonic, never-freed growth that scales with peer churn.

**Description (L1 connection-leak fixes)**

- **Bounded outbound queue + writer task** — the socket sink moves into a per-connection writer task; the actor keeps only a **bounded** `mpsc::Sender` and `send()` is a non-blocking `try_send`, so the actor never blocks on the network and its mailbox can't grow on a slow peer. On a full queue the peer is dropped (`OutboundQueueFull`).
- **Handshake timeout** (`HANDSHAKE_TIMEOUT`, 10 s) — drops stalled-handshake ghost connections.
- **Pong-deadline** (`MAX_MISSED_PONGS`, 3) — drops application-dead-but-TCP-alive peers.
- **Inbound admission cap** (`MAX_INBOUND_CONNECTIONS`, 150) — a bounded semaphore permit, held by each inbound connection actor and released on drop, hard-bounds concurrent inbound connections (the attack surface). Outbound dials are not capped.
- **tx-broadcaster index prune** — a free-list reuses peer indices and removes them on disconnect (clearing the freed bit across `known_txs`), bounding `peer_indexer` and every `PeerMask` width to live peers.
- **discv5 / peer-table session pruning** — `session_ips` gets a TTL (`SESSION_TTL`, 1 h) evicted in the existing cleanup sweep, and the peer-table `sessions` entry is removed on peer drop (both were insert-only).

Together these drop the three ways a peer lingers (wedged / stalled / dead), bound the inbound connection count, and stop the per-peer/per-handshake maps from growing.

**Deliberately out of scope here**
- **Eviction to `TARGET_PEERS`:** `initialize_connection` already gates new connections on `target_peers_reached()`, and the admission cap above hard-bounds inbound connections — so the table/connection count is already bounded and eviction would only trim small race-overshoot. A correct version also needs best-effort actor shutdown of the victim; left as a focused follow-up.
- **L2 `ProofCoordinator`** replicates this same unbounded-mailbox-per-connection pattern — fixed in a separate L2 PR.

**Validation**

- `cargo check` + `clippy -D warnings` + `fmt` clean on `ethrex-p2p` (default and `l2` features).
- Standalone reproduction on the same actor framework + version shows the unbounded-mailbox / awaited-sink pattern leaks without bound; bounded-and-stop stays flat.
- **Live soak (mainnet, jemalloc `inuse_space` heap-diff):**
  - **Phase 1 — ~2 days under sustained load** (full tx-gossip, ~3.4k `NewPooledTransactionHashes`/s, plus replayed RPC): steady-state in-use ~+0.2 GB/day (vs unfixed ~1.5), with the per-connection actor / mailbox sites net-negative; established connections pinned at target (vs unfixed 347 → 432).
  - **Phase 2 — ~4 days idle with RPC fully off** (the decisive condition — the unfixed binary leaked under exactly this): in-use flat (mailbox slope −0.005 GB/day, total flat-to-down), connections stable. Leak rate cut >99%.
- Judge on the per-connection actor/mailbox sites in the diff, not total RSS: on a profiling build jemalloc's own `prof_backtrace_impl` bookkeeping grows several GB over days (absent in non-profiling builds).

**Related**

Upstream framework hardening that would prevent this class of bug at the source: lambdaclass/spawned#170 (opt-in bounded mailbox) and lambdaclass/spawned#171 (cancellation can't interrupt a wedged handler).

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) — N/A (no storage changes).

### DIFF
--- a/crates/networking/p2p/discv5/server.rs
+++ b/crates/networking/p2p/discv5/server.rs
@@ -27,6 +27,14 @@ const MESSAGE_CACHE_TIMEOUT: Duration = Duration::from_secs(2);
 /// per discv5 handshake and (absent this) was never removed for nodes we don't keep as peers.
 const SESSION_TTL: Duration = Duration::from_secs(3600);
 
+/// Source IP a discv5 session was established from, paired with when it was recorded so stale
+/// entries can be evicted (see `SESSION_TTL`).
+#[derive(Debug, Clone)]
+pub struct SessionSource {
+    pub ip: IpAddr,
+    pub established_at: Instant,
+}
+
 /// Discv5-specific state held within the unified DiscoveryServer.
 #[derive(Debug)]
 pub struct Discv5State {
@@ -45,7 +53,7 @@ pub struct Discv5State {
     pub whoareyou_global_window_start: Instant,
     /// Tracks the source IP that each session was established from, with the insertion time
     /// so stale entries can be evicted (see `SESSION_TTL`).
-    pub session_ips: FxHashMap<H256, (IpAddr, Instant)>,
+    pub session_ips: FxHashMap<H256, SessionSource>,
     /// Collects recipient_addr IPs from PONGs for external IP detection via majority voting.
     pub ip_votes: FxHashMap<IpAddr, FxHashSet<H256>>,
     /// When the current IP voting period started. None if no votes received yet.
@@ -112,7 +120,7 @@ impl Discv5State {
 
         let before_sessions = self.session_ips.len();
         self.session_ips
-            .retain(|_node_id, (_ip, inserted_at)| now.duration_since(*inserted_at) < SESSION_TTL);
+            .retain(|_node_id, source| now.duration_since(source.established_at) < SESSION_TTL);
         let removed_sessions = before_sessions - self.session_ips.len();
 
         let total_removed = removed_messages + removed_challenges + removed_sessions;
--- a/crates/networking/p2p/discv5/server.rs
+++ b/crates/networking/p2p/discv5/server.rs
@@ -23,6 +23,9 @@ const IP_VOTE_WINDOW: Duration = Duration::from_secs(300);
 const IP_VOTE_THRESHOLD: usize = 3;
 /// Timeout for pending messages awaiting WhoAreYou response.
 const MESSAGE_CACHE_TIMEOUT: Duration = Duration::from_secs(2);
+/// Max age of a `session_ips` entry before it is evicted. Bounds the map: it is inserted
+/// per discv5 handshake and (absent this) was never removed for nodes we don't keep as peers.
+const SESSION_TTL: Duration = Duration::from_secs(3600);
 
 /// Discv5-specific state held within the unified DiscoveryServer.
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


MegaRedHand (commented):
I think we should verify that bounding the queue is even necessary. Maybe just adding the pruning improvements fixes the memory leak. Adding the queue bound may hide problems in how we choose peers: it could be that the peer is not really slow, but we're just always choosing them and overloading the connections with too many messages, for example.

crates/networking/p2p/rlpx/error.rs
```diff
@@ -30,6 +30,10 @@ pub enum PeerConnectionError {
     NoMatchingCapabilities,
     #[error("Too many peers")]
     TooManyPeers,
+    #[error("Outbound send to peer timed out")]
+    OutboundSendTimeout,
```
iovoid, line 34: Never constructed

crates/networking/p2p/discv5/server.rs
```diff
@@ -40,8 +43,9 @@ pub struct Discv5State {
     pub whoareyou_global_count: u32,
     /// Start of the current global rate limit window.
     pub whoareyou_global_window_start: Instant,
-    /// Tracks the source IP that each session was established from.
-    pub session_ips: FxHashMap<H256, IpAddr>,
+    /// Tracks the source IP that each session was established from, with the insertion time
+    /// so stale entries can be evicted (see `SESSION_TTL`).
+    pub session_ips: FxHashMap<H256, (IpAddr, Instant)>,
```
iovoid: Might be worth using a struct for clarity

iovoid (changes requested):

ilitteri:
@iovoid, both addressed in `ac9d510`:

- **`OutboundSendTimeout` never constructed** — it's now actually produced. The writer task flags a send-timeout exit via a shared `AtomicBool`, and `send()` surfaces `OutboundSendTimeout` (instead of a generic `Disconnected`) when the bounded queue closes after a wedged send. While there, local drop reasons are now attributed in `ethrex_p2p_disconnections` instead of all bucketing under `NetworkError`: `OutboundQueueFull`/`OutboundSendTimeout` → `UselessPeer` (slow peer / we're over-sending), missed-pong → `PingTimeout` — so a queue-full spike is a visible signal rather than hidden.
- **`session_ips` tuple → struct** — now a named `SessionSource { ip, established_at }` instead of `(IpAddr, Instant)`.

edg-l (approved):

iovoid (approved):

jrchatruc (approved):

ilitteri:
> I think we should verify that bounding the queue is even necessary. Maybe just adding the pruning improvements fixes the memory leak. Adding the queue bound may hide problems in how we choose peers: it could be that the peer is not really slow, but we're just always choosing them and overloading the connections with too many messages, for example.

I'll file an issue to continue discussing these concerns. I think it is necessary.

PR #6901, opened 2026-06-22
ci(l1): publish each crate as a separate job via a reusable workflow

ilitteri:
**Motivation**

The publish workflow ran all 16 crates inside one job looping over them, so the run showed a single "Publish crates in dependency order" step — no per-crate visibility, and a failure or re-run couldn't target an individual crate. This refactors publishing so each crate is its own job, without duplicating the setup/publish logic.

**Description**

- **`.github/workflows/publish-crate.yml`** (new, reusable via `workflow_call`): holds the checkout, Rust toolchain, and publish/verify logic **once**, acting on a single crate (inputs: `crate`, `dry_run`).
- **`.github/workflows/publish.yml`** (caller): one job per crate, each calling the reusable workflow. `needs` encodes the crate dependency graph (derived from `cargo metadata`), so a crate publishes only after its dependencies while independent crates (rlp, crypto, sdk-contract-utils) run in parallel. A small `config` job computes the dry-run flag once for all crate jobs.

Behaviour is unchanged from the current single-job version:
- `rustflags: ""` (don't deny warnings during the verify build),
- idempotent re-runs (skip crates already on crates.io: `already (uploaded|exists)`),
- dry-run = `cargo package --no-verify --list` + `cargo check` (validates every crate against workspace path deps, needs nothing published),
- run name shows `(dry run)` for a `workflow_dispatch` dry run.

The per-job `sleep` between publishes is gone: a dependent's job only starts after its dependency's job completes (and `cargo publish` already blocks until the version is in the index), so `needs` handles propagation.

**Trade-off**

Separate jobs each re-checkout and recompile their dependency subtree (no shared build dir across jobs), so total runner time is higher than the single-job loop. The DAG keeps wall-clock down via parallelism; the payoff is per-crate visibility and re-runnability.

**Validation**

Ran the refactored workflow on CI in dry-run mode (temporary push trigger, reverted): the `config` job and all 16 crate jobs passed, confirming the reusable-workflow wiring, `secrets: inherit`, DAG ordering, and dry-run logic. Verified the run name renders `… (dry run)` for a dry run. `actionlint` clean on both files.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A.

### DIFF
--- a/.github/workflows/publish-crate.yml
+++ b/.github/workflows/publish-crate.yml
@@ -1,0 +1,66 @@
+name: Publish a single crate
+
+# Reusable workflow: publishes (or, in dry-run, verifies) exactly one crate.
+# All the shared setup and logic lives here once; the caller (publish.yml) invokes
+# it per crate with `needs:` enforcing dependency order.
+on:
+  workflow_call:
+    inputs:
+      crate:
+        description: "Crate package name (e.g. ethrex-rlp)"
+        required: true
+        type: string
+      dry_run:
+        description: "Verify only (package + compile), do not publish"
+        required: true
+        type: boolean
+
+permissions:
+  contents: read
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment: crates-release-prod
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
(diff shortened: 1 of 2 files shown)


ilitteri:
Not worth it

PR #6902, opened 2026-06-23
fix(l2): count & report skipped storage slots, not just accounts 

0xalizk:
## What

`ethrex migrate` (binary-trie bootstrap from a geth snapshot + preimages) tracks and
reports how many **accounts** were skipped due to a missing keccak preimage, but
**storage slots** with a missing preimage are dropped silently — no counter, no warning.
This makes the storage path symmetric with the account path: count skipped slots, warn
(capped), and report the total.

## Why

Every leaf migrate drops is an account/slot the resulting binary trie won't contain. The
account skip count is currently the only completeness signal at migration time, and it
says nothing about storage — so a structurally incomplete trie (missing slots) can look
clean. We hit this bootstrapping mainnet: the run reported its account skips, but tens of
millions of storage entries had dropped silently, discoverable only by manually diffing
`storage_count` against the input. This surfaces it directly.

## The subtlety: zero-value slots

In the collect phase a storage entry resolves to `None` for two different reasons:
1. a **missing preimage** (a real coverage gap — what we want to count), or
2. a **legitimately zero-valued slot**, which is intentionally not stored.

Counting all `None`s would conflate them and overcount. The new code re-checks the
preimages on the `None` path and counts a slot as skipped only when its address/slot
preimage is genuinely absent; zero-value slots are not counted.

## Changes

- `collect_phase` returns `skipped_accounts` and `skipped_slots` separately (was one `skipped`).
- Storage `None` arm distinguishes missing-preimage from zero-value, increments
  `skipped_slots`, and emits a capped warning mirroring the account path.
- Final report: `"{skipped_accounts} accounts + {skipped_slots} storage slots skipped …"`.
- Live progress line also shows a running `{accounts}+{slots} skipped`.

Observability only — no change to the trie output. One file (`cmd/ethrex/migrate.rs`), +37/−11.

## Testing

Compiles against `eip-7864-plan`. Semantics validated on a full mainnet migration
(~2 B entries): the account skip count matches the existing counter, the new slot counter
surfaces a previously-silent storage gap, and zero-value slots are correctly excluded.

### DIFF
--- a/cmd/ethrex/migrate.rs
+++ b/cmd/ethrex/migrate.rs
@@ -171,11 +171,13 @@ pub async fn migrate_with_preimages(
     store.create_cf(MIGRATION_TEMP)?;
 
     // Step 4: Collect phase - write all (tree_key, value) pairs to temp CF.
-    let (account_count, storage_count, skipped, total_inserts) =
+    let (account_count, storage_count, skipped_accounts, skipped_slots, total_inserts) =
         collect_phase(&preimages, &code_map, snapshot_path, &store)?;
 
-    if skipped > 0 {
-        warn!("{skipped} entries skipped due to missing preimages");
+    if skipped_accounts > 0 || skipped_slots > 0 {
+        warn!(
+            "{skipped_accounts} accounts + {skipped_slots} storage slots skipped due to missing preimages"
+        );
     }
     info!(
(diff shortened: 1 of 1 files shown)


ElFantasma (approved):
LGTM

PR #6903, opened 2026-06-23
fix(l2): bound proof-coordinator connection handlers (read timeout + admission cap)

ilitteri:
**Motivation**

A codebase audit for the unbounded-per-connection accumulation class found in the L1 P2P layer (see the companion P2P PR) surfaced the same pattern in the L2 proof coordinator: it spawns one `ConnectionHandler` actor per inbound TCP connection with **no admission cap**, and each handler reads the request via an **unbounded `read_to_end`**. A prover that connects and then stalls (or never sends a valid message) keeps its handler actor + socket alive indefinitely, and a burst of such connections spawns handler actors without bound.

**Description**

- **Read timeout** (`READ_TIMEOUT`, 120 s): the request read is wrapped in `tokio::time::timeout`, so a stalled/misbehaving prover connection is dropped instead of pinning a handler + socket forever. Generous, so slow-but-honest provers sending large payloads aren't dropped.
- **Admission cap** (`MAX_CONCURRENT_HANDLERS`, 256): a shared atomic counter is incremented when a connection is admitted and decremented in the handler's `stopped()` hook; connections beyond the cap are dropped. Bounds how many handler actors/sockets can accumulate.

**Validation**
- `cargo check` + `clippy -D warnings` + `fmt` clean on `ethrex-l2`.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` — N/A (no storage changes).

### DIFF
--- a/crates/l2/sequencer/proof_coordinator.rs
+++ b/crates/l2/sequencer/proof_coordinator.rs
@@ -29,6 +29,14 @@ use tracing::{debug, error, info, warn};
 #[cfg(feature = "metrics")]
 use ethrex_metrics::l2::metrics::METRICS;
 
+/// Max time to read a complete prover message before dropping the connection. Prevents a
+/// stalled or misbehaving prover from keeping a ConnectionHandler actor (and its socket) alive
+/// forever. Generous, so slow-but-honest provers sending large payloads aren't dropped.
+const READ_TIMEOUT: Duration = Duration::from_secs(120);
+/// Max concurrent ConnectionHandler actors. Bounds how many handler actors/sockets can
+/// accumulate if provers connect but stall or never send a valid message.
+const MAX_CONCURRENT_HANDLERS: usize = 256;
+
 #[protocol]
 pub trait ProofCoordinatorProtocol: Send + Sync {
     fn incoming_connection(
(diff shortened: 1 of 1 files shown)


crates/l2/sequencer/proof_coordinator.rs
```diff
@@ -139,6 +151,24 @@ impl ProofCoordinator {
...
+                .fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
+            debug!(
+                "Proof coordinator handler cap reached, dropping connection from {}",
+                msg.addr
+            );
+            return;
+        }
+        // On any path after a successful start the handler runs its stopped() hook exactly
+        // once, which releases the slot — so we do not decrement here.
         let _ = ConnectionHandler::spawn(self.clone(), msg.stream, msg.addr)
```
ElFantasma, line 172: The comment on the previous line ("On any path after a successful start the handler runs its stopped() hook exactly once") is precise about the precondition — *successful start*. The path where `ConnectionHandler::spawn` returns `Err` is the gap: the admission slot was already taken at l.157-160, the handler never starts, and so `stopped()` never fires to release it. Each failed spawn permanently subtracts one from the cap; under prolonged spawn failures the coordinator stops admitting connections altogether.

Probably rare in practice (spawn failure on a healthy actor system is unusual), but the fix is mechanical — release on the err branch:

```rust
if let Err(err) = ConnectionHandler::spawn(self.clone(), msg.stream, msg.addr).await {
    error!("Error starting ConnectionHandler: {err}");
    self.concurrent_handlers
        .fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
}
```

Or, slightly cleaner, a small RAII guard that takes the slot at admission and releases on drop, with `mem::forget` (or a `disarm()`) on the successful-spawn path so `stopped()` retains ownership. The guard form also means future early-return paths between admission and spawn (none today) can't introduce the same leak.

Non-blocking but worth picking up since the fix is one line.

ElFantasma (approved):

PR #6905, opened 2026-06-24
perf(l1): defer block data persistence

iovoid:
**Motivation**

Currently we persist block data (bodies, receipts, etc) and perform trie maintenance in the store path, blocking `engine_newPayload`. This delay is usually 8~11% of the block time.

**Description**

However, we don't need to _store_ the data to respond whether the block is valid. Because the block data is expected to be available as soon as we answer the RPC call, we need to commit it somewhere.

For block data, this PR maintains an overlay that stores the unprocessed data for the newest blocks until it is fully persisted.

For state tries, we have an atomic "tries pending commit" marker and a "roots pending commit" map. When reading state data we:
- Check if there are any tries pending commit, if not we can proceed as usual (read from the TrieLayerCache)
- If there are, check if the one we need is in the pending pool and wait until it isn't


crates/storage/store.rs
```diff
@@ -3158,72 +3234,337 @@ impl Store {
...
+        trie_mut.put_batch(parent_state_root, child_state_root, new_layer);
+        *trie_cache.write().map_err(|_| StoreError::LockError)? = Arc::new(trie_mut);
+        Ok(())
+    })();
+    // Always clear the pending root, whether or not the swap succeeded: on success
+    // readers see the installed layer; on failure (poisoning) the lock is poisoned
+    // so gated readers error rather than read stale, and we must not leave them
+    // blocked forever.
+    pending_roots.clear(child_state_root);
+    build
```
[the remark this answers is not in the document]
iovoid: If the worker panics, the channel will be dropped and the node will fail soon after. I don't think unblocking the readers to read stale state is worth it.

Not to mention the proposed panic mechanism (alloc failure) can't even happen by default ([the default is aborting the process](https://doc.rust-lang.org/1.96.0/std/alloc/fn.handle_alloc_error.html)).

crates/storage/store.rs
```diff
@@ -3158,72 +3234,337 @@ impl Store {
...
+    if to_flush.is_empty() {
+        return Ok(());
+    }
+    let hashes: Vec<_> = to_flush.iter().map(|b| b.header.hash()).collect();
+    let codes = snapshot.codes_for(&hashes);
+    let mut max_number = snapshot.flushed_upto();
+
+    let mut tx = backend.begin_write()?;
+    for b in &to_flush {
+        let hash = b.header.hash();
```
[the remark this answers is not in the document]
iovoid: The header caches it's own hash.

iovoid:
About the issue mentioned by @greptile-apps, delaying the error response until the next commit is deliberate. Unless there is some unwanted consequence, surfacing the error during the next block seems like the cleanest way to do it.

Also note that we log the error right after it happens, we don't just wait until the next block.

ilitteri (approved):

crates/storage/store.rs
```diff
@@ -1548,78 +1573,74 @@ impl Store {
...
+                blocks: blocks_with_receipts,
+                codes: code_updates,
+                parent_state_root,
+                child_state_root: last_state_root,
+                account_updates,
+                storage_updates,
+                wait_for_flush: batch_mode,
+                ack: ack_tx,
+            }))
+            .map_err(|e| StoreError::Custom(format!("failed to send block persist: {e}")))?;
```
ElFantasma, line 1631: `pending_trie_roots.register(last_state_root)` runs above at line 1598, but if `persist_tx.send(...)` here returns `Err` (e.g. the worker thread has died), the registered root is never cleared: `apply_trie_phase1` is what would have called `pending_roots.clear(child_state_root)`, and the worker never receives the message. Any subsequent `gated_snapshot(last_state_root)` then blocks forever in `wait_until_ready` — the fast-path `count.load() == 0` no longer short-circuits, and nothing will wake the condvar.

Granted, a dead worker is already a fatal store state, but `wait_for_persistence_idle` and any trie open at that root now deadlock rather than failing fast, which is a worse failure mode.

Minimal fix at this site:

```rust
self.persist_tx
    .send(PersistMessage::Block(BlockPersist { /* ... */ }))
    .map_err(|e| {
        self.pending_trie_roots.clear(last_state_root);
        StoreError::Custom(format!("failed to send block persist: {e}"))
    })?;
```

The same window exists at line 1643 (`ack_rx.recv()` failing): the worker may have died after staging but before `apply_trie_phase1`, leaving the root registered. An RAII guard around `register` that disarms once `apply_trie_phase1` runs would close both windows in one shot.

Non-blocking — the cleanest fix probably belongs in a follow-up since the worker-died case has bigger issues anyway, but worth picking up since the failure mode here is deadlock not error.

iovoid: I addressed a similar comment by greptile [here](https://github.com/lambdaclass/ethrex/pull/6905#discussion_r3474761929).

I'm not sure there is a clean way to handle panics in the worker thread without adding complexity (explicit error signaling), risking correctness (letting it read stale data and relying on the consistency checks) or deadlocks.

ElFantasma: The send-error and panic-in-worker cases are different shapes:

- **Panic-in-worker** (greptile's concern, applies inside `apply_trie_phase1`'s closure): closure unwinds past `clear()`. As you noted, the default allocator aborts so this is largely theoretical, and unblocking readers to see stale state isn't an improvement.

- **Send-error here**: `persist_tx.send(...)` returns `Err(SendError(_))` because the worker has dropped the receiver — *no panic involved on this thread*. Cleanup is straightforward because we still own `last_state_root`:

```rust
self.persist_tx
    .send(PersistMessage::Block(BlockPersist { /* ... */ }))
    .map_err(|e| {
        self.pending_trie_roots.clear(last_state_root);
        StoreError::Custom(format!("failed to send block persist: {e}"))
    })?;
```

Three lines, no new machinery. The fatal-node outcome is the same — what changes is that `wait_for_persistence_idle` and `gated_snapshot` fail fast with an error rather than hang on the condvar after the worker is gone, which is easier to diagnose from logs.

Same window exists at the `ack_rx.recv()` failure path; an RAII guard around `register` that's disarmed by `apply_trie_phase1` closes both in one shot if you'd prefer that shape.

Up to you whether to pick this up here or as a follow-up — not a blocker on my end.

edg-l (commented):
**1. Boot can brick permanently after a head reorg + crash**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/store.rs#L3308-L3336

`anchor_to_durable_head` does `head = min(flushed_upto, latest)` then `load_block_header(head)`, which is disk-only and goes through `CANONICAL_BLOCK_HASHES[head]`. The marker tracks the max flushed number, not which hash is canonical at that height. So: 10a gets imported, made canonical, flushed (marker=10, 10a evicted), then a 1-block reorg comes in, `newPayload(10b)` buffers 10b, FCU rewrites `canonical[10]=hash10b` and `latest=10`, node dies before 10b is flushed. On reboot `head=10` but `canonical[10]=hash10b` has no header on disk, so the load is `None` and we hit `MissingLatestBlockNumber`. Every restart hits the same thing. That window is exactly the flush latency this PR adds, and head reorgs are normal. Note this is the canonical hash being repointed to an unflushed block, not "is some block at this number on disk", so it is reachable with the current `flush_block_data`. Fix: when the load is `None`, walk `head` down to the highest number whose canonical hash actually resolves on disk (and lower the marker), instead of erroring out.

**2. `add_blocks` duplicates the `flush_block_data` write loop**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/store.rs#L405-L431

Same headers/bodies/numbers/tx-locations loop lives in both places, and the `// TODO: Same logic in apply_updates` comment now points at code that's gone. A future schema or encoding change has to land in both or they drift. Pull out one `write_block_data(tx, block)` helper.

**3. `evict_flushed` re-hashes every tx**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/block_data_buffer.rs#L160-L174

Eviction recomputes `tx.hash(&NativeCrypto)` twice (lines 169 and 172) to clean up `tx_index`, but those hashes were already computed at insert (line 69). Up to 2·B·T extra keccaks per flush on the worker. Store the hash in the index value and reuse it.

**4. `mutate_block_buffer` deep-clones the whole buffer**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/store.rs#L3361-L3369

`(*buffer.read()?.clone()).clone()` rebuilds all five maps plus the VecDeque on every insert and every evict, i.e. once per newPayload. The comment at `block_data_buffer.rs:25` calling these "O(1) pointer operations" isn't accurate. It's bounded by the cap-2 backpressure today, but grows if flushing falls behind. The worker is the only writer, so a plain `Mutex<BlockDataBuffer>` mutated in place would avoid the clone. At minimum fix the comment.

**5. Live path acks before the trie layer is built**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/store.rs#L1843-L1869

The live path acks after staging but before `apply_trie_phase1`. If the build fails, the block already returned VALID and the error gets carried to the next block's ack. Only realistic failure here is lock poisoning, which is fatal anyway, so low priority, but worth a comment on the ordering.

**6. `pending_trie_roots` leak if the worker send fails**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/store.rs#L1598-L1642

`apply_updates` calls `register(last_state_root)` before sending the Block message. If `send()` returns `Err` (worker thread gone, channel disconnected), it propagates the error without clearing the root. Every later `open_state_trie` / `gated_snapshot` at that root then blocks forever in `wait_until_ready`. Same class as the panic-skips-clear gap on the worker side. Fix: clear the root in the send error arm before returning. A `Drop` guard around the register-to-ack span would cover both this and the panic case.

**7. `code_exists` is not buffer-aware**

https://github.com/lambdaclass/ethrex/blob/8710da08ee3c51b5674f0d996cdd8c22ec2be481/crates/storage/store.rs#L868-L889

`code_exists` checks the LRU cache then goes straight to `ACCOUNT_CODES` on disk, with no fallback to the buffer. So it returns `false` for a contract whose code was introduced in the current block and is still buffered. No external callers today, but it is public API and a correctness trap. `get_code_metadata` has the same disk-first shape but is saved by falling back to the buffer-aware `get_account_code`. Add a `self.buffer()?.get_code(&code_hash).is_some()` check up front.

Minor: the new `StoreError::Custom(format!(...))` spots (channel send, lock, `decode_flushed_upto`) are stringly-typed where the crate otherwise uses typed variants. Pre-existing style, low priority.

Number 1 is the one I'd hold the merge on. The rest are cleanup and perf polish.

ElFantasma (approved):

iovoid:
1- Fixed in deb7eb81a
2- Improved in 37b9680d6
3- Transactions cache their own hashes, so this is not a problem. 
4- It was done this way (RCU) to block readers for the shortest time possible.
5- Added a comment in 37b9680d6
6- It was done this way because it didn't have the correctness risk of letting them read anyway. It seems like most paths do check through `has_state_root`, but not sure if there are hidden risks. What do you think?
7- Fixed in 37b9680d6

edg-l (approved):
Fixes look good. The walk-down boot recovery, the shared write_block_data dedup, and the buffer-aware code_exists are all what I had in mind, and the ack-ordering comment covers the live-path point.

One nit that's fine as a followup: in apply_updates we register the pending root before the persist_tx.send, but the send error arm doesn't clear it. If the worker thread ever dies (panic drops the receiver), the next send fails and the root leaks, so readers at that root block forever in wait_until_ready. Same root cause as the apply_trie_phase1 panic-skips-clear case. It's only reachable once the worker is already gone, so not a blocker. A small followup that clears the root on the send error and lets wait_until_ready bail out when the worker is dead would close it.

PR #6910, opened 2026-06-25
docs(l1,l2): close release-process and upgrade-test gaps found during v18

ElFantasma:
## Motivation

Running the v18.0.0 release manually surfaced several gaps in the release docs that caused avoidable failures and debugging cycles. This folds the fixes back into the docs so the next release-runner doesn't hit them.

## Description

`docs/developers/release-process.md`:
- **Step 2 (Bump version)** — document that the bump must also update the internal crate dependency version pins (root `[workspace.dependencies]` + `crates/vm`, `crates/blockchain`, `crates/l2/sdk`), with a `grep` verification. These pins were added by #6896 but never added to the bump checklist; missing them merges cleanly but breaks the whole workspace build on the release PR (every crate at the new version requiring the previous version of its siblings).
- **Step 4.2 (create tag)** — instruct to set the tag **Target** to the release branch (not `main`) and verify it matches the rc tag's commit. The publish workflow reads `publish.yml` from `main` but checks out the tag's commit, so a tag on the wrong commit silently republishes the previous version (green run, nothing published) — which is exactly what happened.
- **Step 4 (testing checklist)** — add per-host upgrade commands (`ethdocker-mainnet`, `multisync-main` filled in; `l2-gpu`, `mainnet-1`, `minimum-mainnet` left as TODO placeholders since the host roster changes between releases).
- **Step 5 (Homebrew)** — convert the prose steps into copy/paste command blocks templated on a version variable.

`docs/developers/l2/upgrade-test.md`:
- **Step 1.3** — put `--l1.timelock-address` in the command body (required for `VERSION_FROM ≥ v9`) and add `--no-monitor` (the monitor TUI hides logs and can deadlock shutdown on an interactive terminal).
- **Step 2** — remind to `source cmd/.env` when running the drain/stop steps from a fresh terminal.

A follow-up will replace the `TODO_MONITOR` reference in `upgrade-test.md` with the tracking issue for the monitor-shutdown deadlock once it's filed.

## Checklist

- [ ] Tests added/updated (N/A — docs only)
- [ ] Documentation updated (this PR)
- [ ] `make lint` passes


docs/developers/release-process.md
```diff
@@ -82,11 +91,184 @@ Before publishing the release, run through the following checks using the pre-re
...
+#### Local L2 upgrade + integration tests
+
+See [Upgrade test](l2/upgrade-test.md) for the full procedure.
+
+#### L2 integration tests with a SP1 prover (`l2-gpu`)
+
+This validates the release's SP1 **GPU** proving path end to end: bring up an L1 + L2 testnet from the **release binaries** with the SP1 GPU prover, then run the integration test against it. Run it on a host with an NVIDIA GPU (`l2-gpu`).
+
+> [!NOTE]
+> The standard [integration tests](l2/integration-tests.md) run the prover in `exec` mode (no real proofs). This variant swaps in the SP1 GPU backend so the release's proving path is actually exercised. GPU/driver setup (CUDA, `nvidia-container-toolkit`, the `docker` group) is covered in [Prover § GPU mode](l2/prover.md#gpu-mode) and [Run an ethrex L2 SP1 prover](../l2/deployment/prover/sp1.md); the SP1 wrap step runs in a `moongate` Docker container, so `docker run --gpus all` must work.
```
ilitteri: Let's move the step-by-step guide for this section out of the release process docs and reference it like how we do with `Local L2 upgrade + integration tests`.

docs/developers/release-process.md
```diff
@@ -87,6 +96,174 @@ Before publishing the release, run through the following checks using the pre-re
...
+    mkdir -p contracts && tar xzf ethrex-contracts.tar.gz -C contracts
+    ```
+
+    Genesis files and rich-account keys aren't published as assets — copy them from a checkout at the same tag:
+
+    ```bash
+    git show $TAG:fixtures/genesis/l1.json          > l1.json
+    git show $TAG:fixtures/genesis/l2.json          > l2.json
+    git show $TAG:fixtures/keys/private_keys_l1.txt > private_keys_l1.txt
+    ```
```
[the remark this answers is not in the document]
ElFantasma: Addressed. This procedure was moved out of the release process into `docs/developers/l2/sp1-gpu-integration-test.md`, and the `git show $TAG:fixtures/...` step (which ran outside a git repo and would fail) is replaced with a shallow `git clone` + `cp`, reused by the test step. Thanks for catching it.

docs/developers/release-process.md
```diff
@@ -95,10 +272,19 @@ Once the pre-release is created and you want to publish the release, go to the [
...
-2. Manually create the tag `vX.Y.Z`
+2. Manually create the tag `vX.Y.Z`. **Set the tag's _Target_ to the release branch (`release/vX.Y.Z`), not the default `main`** — otherwise the tag lands on `main`'s HEAD, which is not the version-bumped commit.
 
     ![edit tag](../img/publish_release_step_2.png)
 
+    Before finalizing, verify the tag points at the exact commit you tested (the release candidate):
+
+    ```bash
+    git ls-remote origin refs/tags/vX.Y.Z refs/tags/vX.Y.Z-rc.W   # the two SHAs must match
+    ```
```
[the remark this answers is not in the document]
ElFantasma: Addressed. Reworded so the `git ls-remote` check runs **after** clicking *Update release* (the tag only exists on the remote then), and noted that an earlier check falsely "matches" two empty results. Added the move-the-tag fix-up for the mismatch case.

docs/developers/l2/upgrade-test.md
```diff
@@ -154,17 +154,21 @@ set -a; source cmd/.env; set +a
...
   --proof-coordinator.l1-private-key 0x39725efee3fb28614de3bacaffe4cc4bd8c436257e2c8bb887c4b5c4be45e76d \
-  --proof-coordinator.addr 127.0.0.1
+  --proof-coordinator.addr 127.0.0.1 \
+  --no-monitor
 ```
 
-> If the release you are testing requires `--l1.timelock-address`, add `--l1.timelock-address "$ETHREX_TIMELOCK_ADDRESS"`. From v9 onwards this is required for non-based deployments; check the per-release migration guide for the exact set of flags.
+> `--l1.timelock-address` is required for non-based deployments from **v9 onwards** (the committer targets the Timelock, not the OCP directly). Without it the sequencer exits during startup. Drop the flag only if `$VERSION_FROM` predates v9; check the per-release migration guide for the exact set of flags.
+
+> `--no-monitor` disables the terminal monitor UI. Keep it off for this test: the monitor renders a full-screen TUI that hides the startup logs, and it can deadlock node shutdown on an interactive terminal (see [#6911](https://github.com/lambdaclass/ethrex/issues/6911)). With `--no-monitor` the logs stream normally and `Ctrl-C` stops the node cleanly.
```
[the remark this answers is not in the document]
ElFantasma: Dismissed — already resolved. The `TODO_MONITOR` placeholder was replaced with a link to #6911 in an earlier commit; thread resolved.

docs/developers/release-process.md
ElFantasma: Done — moved the step-by-step into `docs/developers/l2/sp1-gpu-integration-test.md` and left a one-line reference here, mirroring the `Upgrade test` link. Added it to `SUMMARY.md` too.

ElFantasma:
Thanks for the Kimi review — suggestions addressed:

- **Cleanup of `nohup`'d processes**: added a cleanup step to the SP1 GPU page (`pkill` the prover/sequencer/L1 + remove the `moongate` container).
- **`tmux` status**: the multisync step now notes `tmux attach -t sync` (detach `Ctrl-b d`).
- **"v9" ambiguity**: clarified to ethrex **v9.0.0**.
- **`$WORK` in a fresh terminal**: the Step 2 env reminder now also points to the Placeholders section to re-export `WORK`/`VERSION_FROM`.
- **`--sp1-vk-path` filename**: kept the warning; the exact filename is being validated against the contracts tarball separately.

(The SP1 GPU procedure these refer to now lives in `docs/developers/l2/sp1-gpu-integration-test.md`.)

ElFantasma:
Thanks for the Codex review — both findings addressed:

1. **`git show` outside a git repo** — fixed. This procedure moved to `docs/developers/l2/sp1-gpu-integration-test.md` and now clones the source + `cp`s the fixtures (the test step reuses that clone), so no `git show` in a non-repo dir.
2. **`cmd/.env` verifier var names + spurious TDX** — correct, the deployer emits `ETHREX_DEPLOYER_*_VERIFIER_ADDRESS` (not `*_CONTRACT_VERIFIER`/`PICO`). Verified the integration test reads **none** of the verifier/aggregator vars (they're deploy-time inputs to `deploy`), so I reduced `cmd/.env` to the only two the test actually reads — `ETHREX_WATCHER_BRIDGE_ADDRESS` and `ETHREX_COMMITTER_ON_CHAIN_PROPOSER_ADDRESS`. That drops the wrong names and the TDX entry entirely.

ElFantasma:
Thanks for the Claude review — all findings addressed:

- **grep `"<previous version>"` false-green** → now `PREV=X.Y.Z; grep -rn "\"$PREV\"" --include=Cargo.toml .`.
- **`git show` outside repo** → fixed (clone + `cp`, in the moved SP1-GPU page).
- **`ls-remote` timing** → reworded to run **after** *Update release*, with the mismatch fix-up.
- **`tmux new-session` duplicate `sync`** → added `tmux kill-session -t sync 2>/dev/null || true` guard (mirrors the Makefile).
- **`$WORK` in a fresh terminal** → Step 2 now points to re-export the [Placeholders](#placeholders).
- **pre-v9 timelock note missing at Step 4.1** → caveat added there too.
- **`TODO_MONITOR`** → resolved to #6911 (earlier commit; thread resolved).
- **hardcoded dev keys** → confirmed dev-chain test keys, no change.

docs/developers/release-process.md
```diff
@@ -82,11 +91,184 @@ Before publishing the release, run through the following checks using the pre-re
...
+    nohup ./ethrex --network l1.json \
+      --http.addr 0.0.0.0 --http.port 8545 \
+      --authrpc.addr 0.0.0.0 --authrpc.port 8551 \
+      --dev --datadir dev_ethrex_l1 > l1.log 2>&1 &
+    ```
+
+3. **Deploy the L1 contracts with SP1 enabled** (`solc`/`forge` must be on `PATH`):
+
+    ```bash
+    COMPILE_CONTRACTS=true ./ethrex-l2 l2 deploy \
```
iovoid: Setting ETHREX_DEPLOYER_ENV_FILE_PATH prevents a panic if the default destination is unwriteable/missing _and_ gives the .env that is used in a later step.

docs/developers/l2/sp1-gpu-integration-test.md
```diff
@@ -0,0 +1,134 @@
...
+    ETHREX_WATCHER_BRIDGE_ADDRESS=<BRIDGE_FROM_DEPLOY>
+    ETHREX_COMMITTER_ON_CHAIN_PROPOSER_ADDRESS=<PROPOSER_FROM_DEPLOY>
+    EOF
+
+    INTEGRATION_TEST_L1_RPC=http://localhost:8545 \
+    INTEGRATION_TEST_L2_RPC=http://localhost:1729 \
+    ETHREX_DEPLOYER_PRIVATE_KEYS_FILE_PATH=~/ethrex_$TAG/private_keys_l1.txt \
+    INTEGRATION_TEST_PRIVATE_KEYS_FILE_PATH=~/ethrex_$TAG/private_keys_l1.txt \
+    INTEGRATION_TEST_BRIDGE_OWNER_PRIVATE_KEY=0x941e103320615d394a55708be13e45994c7d93b932b064dbcb2b511fe3254e2e \
+    cargo test -p ethrex-test l2:: --release --features l2 -- --nocapture --test-threads=1
```
iovoid: This tries to run the shared bridge tests, which require additional setup.

iovoid (changes requested):

docs/developers/release-process.md
ElFantasma: Good call — applied. The deploy now passes `--env-file-path ~/ethrex_${TAG}_src/cmd/.env`, so it writes the addresses to a writable path (no panic) and lands the `cmd/.env` the test reads. That let me drop the manual `cat > cmd/.env` from the test step. (This procedure has since moved to `docs/developers/l2/sp1-gpu-integration-test.md`, where the change lives.)

docs/developers/l2/sp1-gpu-integration-test.md
ElFantasma: Fixed — narrowed the selector from `l2::` to `l2::integration_tests` so it runs only the integration suite, not the shared-bridge tests (which need extra setup). Matches the command we actually ran.

edg-l (approved):

iovoid (approved):

ilitteri (approved):

PR #6913, opened 2026-06-26
fix(levm): match EELS system-call gas accounting under EIP-8037

ilitteri:
**Motivation**

Under Amsterdam, a system contract that runs out of gas is not rejected with `SYSTEM_CONTRACT_CALL_FAILED`. The system-call frame is funded with `SYS_CALL_GAS_LIMIT + TX_BASE_COST` and then charged intrinsic gas during execution. The fixed `+21000` buffer was sized to cancel Prague's flat 21000 intrinsic, but EIP-2780 lowers Amsterdam's intrinsic to 15000, over-funding the frame by 6000 — enough that a system contract engineered to consume exactly `SYS_CALL_GAS_LIMIT + 1` finishes instead of running out of gas, so the block is rejected with a BAL-validation error instead of `SYSTEM_CONTRACT_CALL_FAILED`.

**Description**

- `generic_system_contract_levm`: set the system-call frame gas to `SYS_CALL_GAS_LIMIT` (drop the `+ TX_BASE_COST` padding).
- `DefaultHook::prepare_execution`: charge zero intrinsic for system calls, matching EELS `process_unchecked_system_transaction` (`gas = SYSTEM_TRANSACTION_GAS`, `intrinsic_regular_gas = 0`), while still setting up the EIP-8037 state-gas reservoir.
- `exception_mapper`: map the EIP-7928 "empty change set" BAL error to `INVALID_BLOCK_ACCESS_LIST` so the in-process engine mapper matches the EELS mapper.

Verified against the full Amsterdam suite: EEST blockchain 2907/2907, engine 2908/2908 (including strict exception matching, 23596/23596), state 14673/14673; and hive `ethereum/eels/consume-engine` `fork_Amsterdam` 23596/23596, 0 failed. The `system_contract_out_of_gas` cases are now rejected with `SYSTEM_CONTRACT_CALL_FAILED` while the `reaches_gas_limit` boundary still succeeds.

**Checklist**

### DIFF
--- a/crates/vm/backends/levm/mod.rs
+++ b/crates/vm/backends/levm/mod.rs
@@ -2681,9 +2681,17 @@ pub fn generic_system_contract_levm(
     let config = EVMConfig::new_from_chain_config(&chain_config, block_header);
     let env = Environment {
         origin: system_address,
-        // EIPs 2935, 4788, 7002 and 7251 dictate that the system calls have a gas limit of 30 million and they do not use intrinsic gas.
-        // So we add the base cost that will be taken in the execution.
-        gas_limit: SYS_CALL_GAS_LIMIT + TX_BASE_COST,
+        // EIPs 2935, 4788, 7002 and 7251 dictate that the system calls have a gas
+        // limit of 30 million and they do not use intrinsic gas. EELS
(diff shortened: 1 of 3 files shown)


ilitteri:

PR #6915, opened 2026-06-26
fix(levm): preserve balance in EIP-8246 selfdestruct block access list

ilitteri:
**Motivation**

Under EIP-8246 (Amsterdam) `SELFDESTRUCT` no longer burns ETH, so a contract created and self-destructed in the same transaction keeps its balance while its nonce, code and storage are cleared. `delete_self_destruct_accounts` already preserved the balance in state, but the BAL recorder's `track_selfdestruct` still collapsed the account's balance changes to 0 (pre-EIP-6780 burn semantics), producing an EIP-7928 block access list whose hash did not match the expected commitment.

**Description**

Make `track_selfdestruct` fork-aware: when the balance is preserved (Amsterdam+), keep the recorded balance changes and only drop the now-net-zero nonce/code/storage changes; pre-Amsterdam keeps the burn/collapse behaviour.

Fixes the blockchain ef-test `BlockAccessListHashMismatch` failures under `for_amsterdam` (eip8246/eip8037/eip7928 selfdestruct cases); engine and pre-Amsterdam selfdestruct suites unchanged.

**Checklist**

### DIFF
--- a/crates/common/types/block_access_list.rs
+++ b/crates/common/types/block_access_list.rs
@@ -1560,7 +1560,7 @@ impl BlockAccessListRecorder {
     /// Called after destroy_account for contracts created and destroyed in the same tx.
     /// Removes nonce/code changes, converts storage writes to reads.
     /// Matches EELS `track_selfdestruct` in state_tracker.py:315.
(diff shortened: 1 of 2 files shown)


ilitteri:

PR #6917, opened 2026-06-26
fix(l1): run and correctly grade Amsterdam state ef-tests

ilitteri:
**Motivation**

The Amsterdam state ef-tests were silently skipped (Amsterdam not in the runner's default fork list), and the intrinsic-gas insufficiency grading did not match the spec's single combined exception.

**Description**

- Enable the Amsterdam fork in the state ef-test runner's default fork list so the `state_tests/for_amsterdam` fixtures are executed instead of skipped.
- Accept the intrinsic-gas and calldata-floor insufficiency exceptions interchangeably (`IntrinsicGasTooLow` / `IntrinsicGasBelowFloorGasCost`): the Amsterdam spec raises a single `InsufficientTransactionGasError` for both, so the EEST sub-label is finer than consensus and a spec-faithful rejection must be accepted for either expected label. The intrinsic-gas computation is unchanged.
- Fix the exception-name mapping typo `INTRINSIC_BELOW_FLOOR_GAS_COST` → `INTRINSIC_GAS_BELOW_FLOOR_GAS_COST` to match the canonical EEST name.

**Checklist**

### DIFF
--- a/tooling/ef_tests/state/runner/levm_runner.rs
+++ b/tooling/ef_tests/state/runner/levm_runner.rs
@@ -337,6 +337,17 @@ fn exception_is_expected(
                 TransactionExpectedException::IntrinsicGasBelowFloorGasCost,
                 VMError::TxValidation(TxValidationError::IntrinsicGasBelowFloorGasCost)
             ) | (
+                // The spec raises a single InsufficientTransactionGasError for both the
+                // intrinsic-gas and calldata-floor insufficiency cases (amsterdam
+                // transactions.py: `Insufficient intrinsic gas` / `Insufficient calldata
+                // floor`), so the TooLow vs BelowFloor distinction is finer than the spec
+                // defines. Accept either spec-faithful rejection for the other's label.
+                TransactionExpectedException::IntrinsicGasTooLow,
+                VMError::TxValidation(TxValidationError::IntrinsicGasBelowFloorGasCost)
+            ) | (
+                TransactionExpectedException::IntrinsicGasBelowFloorGasCost,
+                VMError::TxValidation(TxValidationError::IntrinsicGasTooLow)
+            ) | (
                 TransactionExpectedException::InsufficientAccountFunds,
                 VMError::TxValidation(TxValidationError::InsufficientAccountFunds)
             ) | (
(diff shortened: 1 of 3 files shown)


edg-l (approved):

tooling/ef_tests/state/runner/levm_runner.rs
```diff
@@ -336,6 +336,17 @@ fn exception_is_expected(
...
+                // intrinsic-gas and calldata-floor insufficiency cases (amsterdam
+                // transactions.py: `Insufficient intrinsic gas` / `Insufficient calldata
+                // floor`), so the TooLow vs BelowFloor distinction is finer than the spec
+                // defines. Accept either spec-faithful rejection for the other's label.
+                TransactionExpectedException::IntrinsicGasTooLow,
+                VMError::TxValidation(TxValidationError::IntrinsicGasBelowFloorGasCost)
+            ) | (
+                TransactionExpectedException::IntrinsicGasBelowFloorGasCost,
+                VMError::TxValidation(TxValidationError::IntrinsicGasTooLow)
             ) | (
```
ElFantasma, line 350: Scope question: `exception_is_expected` doesn't take the fork as a parameter, so this cross-acceptance applies to every fork the runner iterates — Paris, Shanghai, Cancun, Prague, Osaka, Amsterdam. The comment's spec citation ("Amsterdam transactions.py") is correct for Amsterdam, but pre-Amsterdam forks do distinguish the two errors:

- Pre-Prague: `IntrinsicGasBelowFloorGasCost` shouldn't even be reachable (EIP-7623 lands in Prague). If LEVM ever returns it on Paris/Shanghai/Cancun, this match arm would silently accept what's actually a wrong-error bug.
- Prague/Osaka: the spec keeps them distinct; a label mismatch on a non-Amsterdam test would now pass where it would previously have flagged.

In practice this is probably a no-op (LEVM's pre-EIP-7623 paths can't reach `IntrinsicGasBelowFloorGasCost`, and the Prague EEST suite is unlikely to swap labels), but you can pin the scope by threading the fork into `exception_is_expected` and gating those two cross arms behind `fork >= Amsterdam`. Non-blocking — flagging because the comment locks the rationale to Amsterdam while the code does not.

ElFantasma (approved):
Just one scope question, it LGTM

ilitteri:

PR #6919, opened 2026-06-26
fix(l1): accept v1 blob sidecars pre-osaka and relax pooled-tx size check

ElFantasma:
**Motivation**

The `devp2p` hive suite (`TestBlobTxWithoutSidecar`, `TestBlobTxWithMismatchedSidecar`) started failing main-wide with "unexpected disconnect". The trigger is go-ethereum [#35191](https://github.com/ethereum/go-ethereum/pull/35191) ("V0 Blob Sidecar Support Removal"), which makes geth always produce **v1 (EIP-7594 cell-proof)** blob sidecars — even pre-Osaka — instead of v0 (EIP-4844 blob proofs). The hive simulator builds its `devp2p` test peer from go-ethereum at image-build time, so once that picked up #35191 the peer began sending v1 sidecars, which ethrex rejected, disconnecting an otherwise-valid peer.

**Description**

Two independent issues surfaced, both on the mempool/propagation path (blob sidecars are never included in a block, so neither change affects consensus):

1. **Blob sidecar version acceptance** (`crates/common/types/blobs_bundle.rs`, `validate_cheap`): previously a fork-exact check (`v0` before Osaka, `v1` on Osaka+) rejected v1 sidecars pre-Osaka. The network is mid-migration to cell proofs and peers switch at different times, so pre-Osaka we now accept **both v0 and v1**; Osaka+ still requires v1. Updated the unit tests accordingly (`v1_sidecar_is_accepted_pre_osaka`, `v0_sidecar_is_rejected_on_osaka`).

2. **Pooled-transaction size check** (`crates/networking/p2p/rlpx/eth/transactions.rs`): `validate_requested` required the received transaction's encoded length to *exactly* equal the size announced in `NewPooledTransactionHashes`. geth's `Transaction.Size()` under-counts a v1 sidecar by exactly one byte — it omits the wrapper version byte — so the strict check disconnected geth on every v1 blob-tx announcement. The announced size is a soft hint; we now tolerate up to 8 bytes of skew, matching go-ethereum's tx fetcher (`eth/fetcher/tx_fetcher.go`). ethrex's own announcements already fall within geth's 8-byte tolerance, so the send side needs no change.

Validated by reproducing the exact CI configuration (hive simulator rebuilt with fresh geth incl. #35191): the `devp2p` `eth` suite goes from `failed=3` to **21/21 passing**, including both target tests. `blobs_bundle` unit tests (8/8 with `--features c-kzg`) and clippy on the touched crates are clean.

**Checklist**

- [x] Reproduced the failure against a fresh-geth hive simulator and confirmed the fix (`devp2p`/`eth` 21/21).

### DIFF
--- a/crates/networking/p2p/rlpx/eth/transactions.rs
+++ b/crates/networking/p2p/rlpx/eth/transactions.rs
@@ -19,6 +19,13 @@ use ethrex_rlp::{
 use ethrex_storage::error::StoreError;
 use tracing::debug;
 
+/// Allowed skew between a peer's announced pooled-transaction size and the size we compute
+/// from the received transaction. The announced size is a soft hint that varies slightly
+/// between clients (e.g. geth's `Transaction.Size()` omits the v1 blob-sidecar wrapper
+/// version byte), so we tolerate a few bytes before treating it as a protocol violation.
+/// Matches go-ethereum's tx fetcher (`eth/fetcher/tx_fetcher.go`).
+const POOLED_TX_SIZE_TOLERANCE: usize = 8;
+
 // https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transactions-0x02
 // Broadcast message
 #[derive(Debug, Clone)]
@@ -305,8 +312,14 @@ impl PooledTransactions {
             if tx.tx_type() as u8 != expected_type {
                 return Err(MempoolError::InvalidPooledTxType(expected_type));
             }
+            // The announced size is a soft hint, not an exact value. geth's
+            // `Transaction.Size()` under-counts a v1 (EIP-7594 cell-proof) blob sidecar by
+            // one byte — it omits the wrapper version byte — so a strict equality check
+            // would disconnect geth peers on every v1 blob-tx announcement. Match geth's tx
+            // fetcher, which tolerates up to 8 bytes of skew before treating it as a
+            // violation (go-ethereum eth/fetcher/tx_fetcher.go).
             let tx_size = tx.encode_canonical_to_vec().len();
-            if tx_size != expected_size {
+            if tx_size.abs_diff(expected_size) > POOLED_TX_SIZE_TOLERANCE {
                 return Err(MempoolError::InvalidPooledTxSize);
             }
         }
--- a/crates/common/types/blobs_bundle.rs
+++ b/crates/common/types/blobs_bundle.rs
@@ -201,10 +201,22 @@ impl BlobsBundle {
             return Err(BlobsBundleError::BlobBundleEmptyError);
         }
 
-        // The wrapper version is fork-specific: 0 (blob proofs) before Osaka, 1 (cell
-        // proofs, EIP-7594) on Osaka+. Any other value is invalid.
-        let expected_version = if fork >= Fork::Osaka { 1 } else { 0 };
-        if self.version != expected_version {
+        // Blob sidecar wrapper version: 0 = EIP-4844 blob proofs, 1 = EIP-7594 cell proofs.
+        // The sidecar is a mempool/propagation-only artifact (never included in a block) and
+        // both formats are cryptographically verifiable at any fork.
+        //
+        // Acceptance criteria:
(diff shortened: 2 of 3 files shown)


ElFantasma:
**Re: greptile P1 — "Cell Proof Returned As Blob Proof"**

Good catch — confirmed and fixed in c68d3f4c4.

Accepting v1 (EIP-7594 cell-proof) sidecars pre-Osaka means a v1 bundle (128 cell proofs/blob) can now sit in the mempool before Osaka. `getBlobsV1` — which is gated to pre-Osaka and serves the single EIP-4844 blob proof — was unconditionally taking `proofs[0]`, so for a v1 bundle it would hand the CL the first *cell* proof in the blob-proof field.

Fix: `getBlobsV1` now returns `null` for any bundle that isn't a single-proof (v0) bundle. `null` is the spec's "not available" signal, so the CL falls back to gossip instead of receiving an invalid proof. (Recomputing the v0 blob proof from the blob would preserve the optimization, but the rpc crate doesn't enable `c-kzg`; returning `null` is the safe, minimal fix for the transitional window.)

`getBlobsV2`/`V3` are unaffected — they serve the full cell-proof vec and are Osaka-gated (return `null` pre-Osaka).

---

**Related same-root exposure I found (not flagged by the review): payload building.**

`BlobsBundle`'s `AddAssign` (used by the payload builder at `crates/blockchain/payload.rs:805` via `context.blobs_bundle += blobs_bundle`) extends `proofs` but does not carry `version`. If ethrex builds a pre-Osaka payload that includes a v1 blob tx (now possible), the resulting `blobs_bundle` carries cell proofs while `getPayloadV3`/`V4` advertises them as v0 blob proofs — the same class of mismatch as the `getBlobsV1` bug, on the proposer side.

Unlike `getBlobsV1`, this can't be fixed by returning `null` (it's assembling a block). Options: (a) skip v1 blob txs during pre-Osaka payload building, or (b) recompute v0 proofs for the served bundle. Both are behavior/throughput decisions rather than a one-liner, so I've left them out of this PR pending a call on direction. Flagging here for visibility.

iovoid (approved):

ilitteri (approved):

crates/networking/p2p/rlpx/eth/transactions.rs
```diff
@@ -305,8 +312,14 @@ impl PooledTransactions {
             if tx.tx_type() as u8 != expected_type {
                 return Err(MempoolError::InvalidPooledTxType(expected_type));
             }
+            // The announced size is a soft hint, not an exact value. geth's
+            // `Transaction.Size()` under-counts a v1 (EIP-7594 cell-proof) blob sidecar by
+            // one byte — it omits the wrapper version byte — so a strict equality check
+            // would disconnect geth peers on every v1 blob-tx announcement. Match geth's tx
+            // fetcher, which tolerates up to 8 bytes of skew before treating it as a
+            // violation (go-ethereum eth/fetcher/tx_fetcher.go).
```
MegaRedHand, line 320: This sounds like a bug on geth's side. Also, we should limit the number of bytes we read instead of checking the size after decoding.

MegaRedHand (approved):
LGTM 👍 👍

crates/networking/p2p/rlpx/eth/transactions.rs
ElFantasma: Agreed on both:

- **geth side:** it's a `Transaction.Size()` imprecision — it never counts the v1 wrapper version byte, so it under-reports by 1. But geth knowingly tolerates it: its tx fetcher allows ±8 bytes (*"RLP vs consensus format messyness"*) with a `TODO: get rid of this relaxation when clients are proven stable`. So it won't tighten upstream soon, and matching that 8-byte tolerance is the pragmatic interop choice here regardless.
- **bounding the read:** you're right that this checks size only after decoding the whole payload, so it's a correctness/interop check rather than a resource guard. Bounding the bytes read up front (per-response budget from the sizes we requested + capping the snappy decoded length) is a larger decode-path change, so I've filed it as a follow-up: #6928. Kept this PR scoped to the CI fix.

PR #6920, opened 2026-06-26
ci(l1,l2): make latest-release Docker tag promotion race-free

### LINKED ISSUE #6918: tag_latest.yaml silently skips docker latest/version promotion (and apt publish) and reports success
## Summary

After the v18.0.0 release, `ghcr.io/lambdaclass/ethrex:latest` still pointed at **v17**, `:18.0.0` / `:18.0.0-l2` were **never created**, `:l2` was stale, and the APT package wasn't published — yet the `Ethrex Latest Release` workflow run reported **success**. Operators had to use `:main` to get v18.

## Root cause

The `retag_docker_images` job in `.github/workflows/tag_latest.yaml` is gated by:

```yaml
if: needs.get_latest_release.outputs.latest_tag == github.event.release.tag_name && github.event.changes.tag_name.from
```

In the v18 run, `get_latest_release` logged `Latest release tag: v17.0.0`, so the gate evaluated `v17.0.0 == v18.0.0` → false → both `retag_docker_images` **and** the chained `publish-apt` were **skipped**. Skipped jobs don't fail a run, so the workflow went green.

i.e. at the instant the triggering `release: edited` event fired, GitHub's `/releases/latest` still returned v17 — v18.0.0 wasn't recognized as "latest" yet (the edit landed before "Set as latest" took effect / while still pre-release; possibly an eventual-consistency race between the edit and the workflow's API call seconds later).

This is independent of the wrong-commit tag that broke the crates.io publish — that workflow checks out the tag's commit; this one retags by image *name* (`18.0.0-rc.1` -> `18.0.0`/`latest`) and failed on the `latest_tag` gate.

## Why it's fragile

- Promotion happens **only** if one precise event lands: a `release: edited` that renames the tag from the rc **and** finds the release already "latest".
- Any miss is reported as a **successful** run (skipped != failed) — silent.
- There's **no `workflow_dispatch` re-promotion path**, so it can't be re-run after a miss.
- The documented manual fallback needs **ghcr write creds** that most people don't have.

## Proposed fixes

1. Add a `workflow_dispatch` path (inputs: source rc tag + target version) that runs the retag + `publish-apt` via `GITHUB_TOKEN` (`packages: write`) — credential-free recovery.
2. Make a required-but-skipped promotion **fail** (or post a warning) instead of reporting success.
3. `docker buildx imagetools create` (already used) preserves the multi-arch index — keep it.

## Impact

v18.0.0 shipped with `:latest`/`:l2` on v17 and no `:18.0.0`/`:18.0.0-l2` for ~2 days until manually retagged. Surfaced via an operator's `eth update` report.

ilitteri:
**Motivation**

When v18.0.0 was released, `ghcr.io/lambdaclass/ethrex:latest` never moved off v17.0.0 — and `:18.0.0` was never created at all (HTTP 404). Users on `:latest` kept getting v17; the only newer image was the rolling `:main` tag. The v18 apt package also never published.

The "Ethrex Latest Release" run for v18 reported success, but its `retag_docker_images` and `publish-apt` jobs were **skipped**. The retag job was gated on a live read of GitHub's `/releases/latest`, which excludes prereleases, sorts by `created_at`, and lags a just-saved edit. During the promotion edit that read still returned `v17.0.0`, so `latest_tag == release.tag_name` was false and the job skipped — and a skipped job leaves the whole run green, so nothing flagged it. `publish-apt` skipped in the cascade (`needs: retag`). The v17 release needed three edit events before the gate happened to align; v18 never got that lucky.

**Description**

Decide the promotion from the event payload / manual inputs instead of the API, and make a non-promotion loud:

- **Race-free gate** — remove the `get_latest_release` job and its `/releases/latest` read. A new `resolve` job derives the source RC tag and target version from the edit payload (tag was renamed **and** the release is no longer a pre-release) or from `workflow_dispatch` inputs. No external state, so the race is gone.
- **Manual recovery** — `workflow_dispatch` gains `rc_tag` and `version` inputs that perform the identical promotion deterministically (a one-click replacement for the manual PAT + `docker tag`/`push` procedure).
- **Loud verification** — `verify-latest` fails the run unless the floating tags (`:latest`, `:l2`, `:performance`, `:performance-l2`) resolve to the same digest as `:X.Y.Z` / `:X.Y.Z-l2`. A skipped or partial promotion can no longer pass as a green run.
- **Promotion tripwire** — on `release: released`, `assert-latest-promoted` polls until `:latest` matches the released version, failing loudly when a multi-step promotion means the automatic retag never fired. It polls (not check-once) because a single-edit promotion emits both `edited` and `released` as two concurrent runs with no guaranteed order.
- **Docs** — `release-process.md`: promote in a single edit; use the `workflow_dispatch` recovery as the first-line fix.

The RC image remains the promotion source (reusing the tested build); the final `vX.Y.Z` tag does not build its own image.

**How to roll out (also fixes the live v18 breakage)**

After merge: **Actions → Ethrex Latest Release → Run workflow** with `rc_tag=v18.0.0-rc.1`, `version=v18.0.0`. The RC images are confirmed present in ghcr; this promotes `:latest`/`:l2`/`:performance`/`:18.0.0` from them, publishes the v18 apt package, and `verify-latest` confirms `:latest == :18.0.0`.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` — N/A (CI-only change)

Closes #6918

### DIFF
--- a/.github/workflows/tag_latest.yaml
+++ b/.github/workflows/tag_latest.yaml
@@ -1,52 +1,96 @@
 name: Ethrex Latest Release
 
+# Promotes a tested release candidate to the user-facing Docker tags
+# (latest / l2 / performance / X.Y.Z) and triggers the apt publish.
+#
+# The promotion decision is made purely from the triggering event payload
+# (or the manual inputs) -- it never reads GitHub's /releases/latest endpoint,
+# which excludes prereleases, sorts by created_at, and lags a just-saved edit.
+# That race previously made the retag job skip silently while the run stayed
+# green (e.g. v18.0.0 left :latest pointing at v17.0.0).
+#
+# Paths:
+#   - release: edited   -> a single edit that renames the RC tag to vX.Y.Z and
+#                          unchecks "pre-release" promotes automatically.
+#   - workflow_dispatch -> deterministic manual recovery: supply rc_tag + version.
+#   - release: released -> tripwire that asserts :latest actually moved to the
+#                          released version, failing loudly if it did not.
 on:
   workflow_dispatch:
+    inputs:
+      rc_tag:
+        description: "Source RC image tag to promote, e.g. v18.0.0-rc.1 (the tested release candidate)"
+        required: true
+        type: string
+      version:
+        description: "Final version tag to publish, e.g. v18.0.0"
+        required: true
+        type: string
   release:
     types:
       - edited
+      - released
 
 permissions:
   contents: read
-  packages: write
+  packages: read
 
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
-  get_latest_release:
-    name: "Get latest release tag"
+  resolve:
+    name: "Resolve promotion tags"
     runs-on: ubuntu-latest
+    # Run on a manual dispatch, or on an edit that promotes an RC to a full
+    # release (the tag was renamed AND the release is no longer a pre-release).
+    # A notes-only edit, or an edit while still a pre-release, is skipped.
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'release' && github.event.action == 'edited' && github.event.changes.tag_name.from != '' && github.event.release.prerelease == false) }}
     outputs:
-      latest_tag: ${{ steps.get_latest_release.outputs.latest_tag }}
-
+      prev_tag: ${{ steps.tags.outputs.prev_tag }}
+      new_tag: ${{ steps.tags.outputs.new_tag }}
     steps:
-      - name: Get latest release tag from GH API
-        id: get_latest_release
+      - name: Determine source (RC) and target (version) tags
+        id: tags
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          INPUT_RC_TAG: ${{ inputs.rc_tag }}
+          INPUT_VERSION: ${{ inputs.version }}
+          EDIT_FROM: ${{ github.event.changes.tag_name.from }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
         run: |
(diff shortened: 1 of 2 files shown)


jrchatruc (approved):

ElFantasma (approved):
Nice — this fixes the root cause rather than papering over it (dropping the `/releases/latest` read and deciding promotion from the event payload), and the `verify-latest` + `assert-latest-promoted` tripwires mean a silent skip can't pass as green anymore. The `workflow_dispatch` recovery is exactly the credential-free re-promotion path we wanted. Closes #6918.

A couple of things:

1. **The last-resort PAT fallback still flattens multi-arch.** In the Troubleshooting section, the retained `docker pull --platform linux/amd64 … && docker tag && docker push` path produces an **amd64-only** `:latest`, dropping arm64. Since the `workflow_dispatch` path now supersedes it, I'd either drop that fallback entirely or switch it to `docker buildx imagetools create -t …:latest … …:X.Y.Z-rc.W` (server-side, preserves the full index — same form the retag job uses).

2. **(minor) `verify-latest` doesn't check the `performance` tags.** The retag also creates `:performance` / `:performance-l2`, but verification only covers `latest`/`l2` vs `X.Y.Z`/`X.Y.Z-l2`. Probably fine (those are the user-facing floating tags), but worth a line if you want full coverage.

FYI on docs: I've trimmed the broader release-docs PR (#6910) so it no longer touches the §Publish "Once done…" block or the "Failure on latest release workflow" section — this PR owns the latest-tag docs now, so no conflict. #6910 keeps the unrelated gaps (version-pin bump, tag Target, l2-gpu, Homebrew, arch-name table fixes, upgrade-test).

iovoid (approved):

PR #6921, opened 2026-06-26
fix(l1): serve GetBlockHeaders by non-canonical hash so syncing peers can fetch their head

ilitteri:
## Motivation

On glamsterdam-devnet-6 several ethrex nodes that fell behind got stuck or lagged (`buildoor-prysm-ethrex-1` wedged ~15h; `lighthouse-ethrex-1`/`nimbus-ethrex-1` transient lag + a missed/orphaned proposal). The common trigger: a syncing node requests headers from its **sync-head hash**, and peers answer with `peer(s) queried but did not serve headers`, so full sync aborts after `MAX_HEADER_FETCH_ATTEMPTS`.

Root cause is in how we **serve** `GetBlockHeaders` by hash. `fetch_headers` translated *any* by-hash start to a block number via `get_block_number` — which is populated for **every** stored block regardless of canonicity — then served **by number**, returning the responder's *canonical* block at that height. When the requested hash is non-canonical on the serving peer (routine on a reorg-prone ePBS devnet where a sync head isn't yet canonical everywhere), `headers[0].hash() != requested`, so the requester (`request_block_headers_from_hash`, which pins `headers[0].hash() == start`) rejects the whole batch.

## Description

In `GetBlockHeaders::fetch_headers` (`crates/networking/p2p/rlpx/eth/blocks.rs`):

1. Only translate a by-hash start to a number when that hash is the **canonical** block at its height (`get_canonical_block_hash(n) == requested`); otherwise keep it as a hash and serve the exact block.
2. For a by-hash start with `reverse` (NewToOld) and `skip == 0`, walk **parent hashes** so a non-canonical chain (a peer's fork/sync head) can be served beyond a single header. Ascending/skipping by hash isn't representable, so we stop after the requested header (unchanged from before).

The canonical by-number path is unchanged. Output is internally hash-chained (`headers[i].parent_hash == headers[i+1].hash()`), matching `are_block_headers_chained`, so both ethrex and other clients accept it. This mirrors geth's `serviceNonContiguousBlockHeaderQuery` (reverse ancestor walk).

## How to test

`cargo test -p ethrex-p2p serves_non_canonical_block_by_hash` — stores a canonical block and a non-canonical sibling at the same height; a by-hash NewToOld request for the sibling must return the sibling (then its parent), not the canonical block. Fails against the pre-fix serve path.

## Notes / follow-ups

- The non-canonical parent walk is bounded by `limit` (<= `BLOCK_HEADER_LIMIT`) and by stored data (an unknown parent hash ends it). geth additionally caps non-canonical traversal at `maxNonCanonical = 100`; a similar small cap here would avoid long side-chain scans during reorg storms — happy to add if preferred.
- This is the *trigger*; a separate follow-up addresses the *recovery* side (a wedged sync cycle holding the syncer lock, and the resume-state path that currently requires `ethrex removedb`).


crates/networking/p2p/rlpx/eth/blocks.rs
```diff
@@ -98,15 +98,28 @@ impl GetBlockHeaders {
...
+            // reverse requests in the loop below), which is exactly what a
+            // syncing peer asking for a fork/sync head needs.
+            HashOrNumber::Hash(block_hash) => match storage.get_block_number(block_hash).await {
+                Ok(Some(block_number))
+                    if storage
+                        .get_canonical_block_hash(block_number)
+                        .await
+                        .ok()
+                        .flatten()
+                        == Some(block_hash) =>
```
ElFantasma, line 117: nit: `Store::is_canonical_sync` (`store.rs:3320`) already implements exactly this `get_block_number` + `get_canonical_block_hash` pair. An async wrapper (or `get_block_number_if_canonical(hash) -> Result<Option<BlockNumber>>`) would let this callsite express the intent directly, and would give the storage-error case a single place to `debug!`/propagate — right now `.ok().flatten() == Some(block_hash)` silently absorbs both `Err(_)` and `Ok(None)`. Both silent paths default to keeping the hash + serving by-hash, which is safe, but a real DB read fault or the (rare) inconsistency where `get_block_number` returns `Some(n)` and `get_canonical_block_hash(n)` returns `None` currently disappears without a log.

ElFantasma (approved):
It looks good to me. Just left a nitpick

ilitteri:
Superseded by #7230, which reapplies this exact fix cleanly on current main (this branch drifted ~2 months / 110 files behind main and was unmergeable). The fix is already on `glamsterdam-devnet-8` (783349968); #7230 brings it to main. Closing as superseded.

PR #6924, opened 2026-06-26
fix(l1): release the syncer lock when waiting for a forkchoice head (sync wedge)

ilitteri:
## Motivation

On glamsterdam-devnet-6, a node that fell behind (`buildoor-prysm-ethrex-1`) sat **wedged ~15h with zero sync logs** while its consensus client drove it every slot — it only recovered after a manual `removedb` + resync. This is the *recovery* half of that incident (the *trigger* — peers not serving headers by hash — is #6921).

## Root cause

`SyncManager::start_sync` spawns a task that `try_lock()`s the `syncer` mutex and **holds it for the whole task**. When the forkchoice head is `H256::zero()` (e.g. right after a restart, before the first `forkchoiceUpdated`), the task entered an infinite `log + sleep(5s) + continue` loop, **never releasing the lock**. Since `is_active()` is `self.syncer.try_lock().is_err()` and `sync_to_head()` only spawns when `!is_active()`, a task parked in that spin pinned `is_active() == true` and turned **every** later `sync_to_head()` — from both `forkchoiceUpdated` and `newPayload` — into a silent no-op. Compounded by `set_head()` dropping the fcu-head on `try_lock` contention (leaving the head at zero), the node could never re-arm: a permanent wedge with no sync logs.

## Fix

In the zero-head branch, **return and release the syncer lock** instead of spinning while holding it. The next `forkchoiceUpdated`/`newPayload` calls `sync_to_head()`, finds the syncer idle, and re-spawns `start_sync()` with the head populated. The non-zero-head path and the snap multi-cycle loop (`continue` while `get_header_download_checkpoint()` is `Some`) are unchanged. The restart-resume state is still surfaced at `info` level (now one-shot). Removed the now-unused `Duration`/`sleep`/`debug` imports.

## Correctness / review

Adversarially reviewed (concurrency + regression lenses) — ship-it, no blockers:
- The `try_lock` guard is dropped synchronously on `return` (no `.await` between the zero-check and the return), so `is_active()` correctly flips back to false.
- There is a residual, **transient** TOCTOU: if a concurrent first FCU's `is_active()` check observes the boot task still holding the lock during the few instructions before its `return`, that FCU's spawn is skipped — recovered on the **next** FCU (CLs resend every slot). This is a strict improvement over the baseline, which held the lock *forever*. **It cannot leave a node permanently un-synced** as long as the CL keeps sending forkchoice updates (if it doesn't, no sync design could proceed).

## Testing

`cargo check`/`cargo clippy -p ethrex-p2p` clean. This is a concurrency/lifetime fix (hard to unit-test deterministically); validated by reasoning + review. Empirically, the wedged node recovered after a manual reset, consistent with the diagnosis.

## Scope / follow-up

This is **Q1** of the recovery work. **Q2** — a node with no retained state pausing with "Full sync cannot resume … run `ethrex removedb`" — is left as a separate change: the obvious "re-enable snap from the resume guard" loops on `< MIN_FULL_BLOCKS` devnets (where full sync is force-enabled), so the real fix is an automatic reset-and-full-resync, which deserves its own design.


edg-l (approved):

crates/networking/p2p/sync_manager.rs
```diff
@@ -213,20 +209,19 @@ impl SyncManager {
...
-                            "Resuming sync after node restart, waiting for a forkchoice update from the consensus client"
-                        );
-                        waiting_for_fcu_logged = true;
-                    }
-                    sleep(Duration::from_secs(5)).await;
-                    continue;
+                    info!(
+                        "No forkchoice head yet (e.g. after a node restart); waiting for the next forkchoice update to start syncing"
+                    );
+                    return;
```
ElFantasma, line 283: Non-blocking: with this change, the initial `start_sync()` spawn from `new()` at line 97 (guarded by `has_checkpoint && snap_enabled`) now always short-circuits — cold-restart has no fcu head yet, so the spawned task hits this `return`, logs once, and exits. Harmless (and arguably useful as a boot signal), but the `new()` call is now effectively a no-op-plus-one-log; could be inlined as just the log directly in `new()` in a later cleanup.

ElFantasma (approved):
LGTM. One non-blocking observation inline.

MegaRedHand (approved):

PR #6916, opened 2026-06-26
feat(l1): add EIP-8282 builder predeploys to l1-bal genesis

ilitteri:
**Motivation**

The Amsterdam request-extraction path reads builder deposit/exit requests from the EIP-8282 system contracts on every block; empty code at those addresses invalidates the block (mirroring the EIP-7002/7251 empty-code rule). `l1-bal.json` activates Amsterdam at genesis but did not allocate the builder predeploys, so every Amsterdam block built on the bundled genesis was rejected.

**Description**

Pre-allocate the builder deposit (`0x0000884d2AA32eAa155F59A2f24eFa73D9008282`) and exit (`0x000014574A74c805590AFF9499fc7A690f008282`) contracts with their canonical bytecode (nonce 1), matching the v6.0.0 fixture pre-state and the existing Prague system-contract allocations.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync.

PR #6914, opened 2026-06-26
fix(l1): use block gas limit for the Amsterdam eth_estimateGas ceiling

ilitteri:
**Motivation**

`get_max_allowed_gas_limit` returned `POST_OSAKA_GAS_LIMIT_CAP` (16,777,216) for every fork ≥ Osaka, including Amsterdam. Under EIP-8037 a transaction's gas limit may exceed that flat cap (the excess funds the state-gas reservoir; only the regular dimension is bound by `TX_MAX_GAS_LIMIT`), so capping the `eth_estimateGas` binary-search ceiling at 16.7M prevents convergence for state-heavy creations.

**Description**

Gate the flat EIP-7825 cap to `Osaka..Amsterdam` (mirroring the Osaka-only gating at the other cap sites); Amsterdam+ uses `block_gas_limit`. RPC-only: consensus tx-gas-limit validation is already Amsterdam-exempt.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync.

PR #6929, opened 2026-06-29
feat(l1): glamsterdam-devnet v6.1.0 EVM support + fixture bump

edg-l:
Brings Amsterdam (glamsterdam-devnet) EVM work to main and bumps EF/hive fixtures to `tests-glamsterdam-devnet@v6.1.0` (state, blockchain, engine ef-tests + hive eels_commit pin). EVM-only, self-contained: everything needed for v6.1.0 fixtures to pass.

**EIP work included:**
- EIP-8037: source-based state-gas refunds, 2D gas-allowance
- EIP-8038: state-access gas repricing (SSTORE cold-XOR-warm, full STORAGE_WRITE refund on restore-to-original)
- EIP-2780: reduced intrinsic tx gas (CREATE base + per-new-account state bytes)
- EIP-7702: multi-auth AUTH_BASE refills
- EIP-7708 / EIP-7928 / EIP-8246: BAL handling
- EIP-8282: builder execution-requests EL wiring + genesis builder predeploys (`l1-bal.json`)
- ef-test state grading for devnet-6

**Intentionally excluded** (separate PRs): snap/2 BAL-based state healing (#6544), discv4 IpPredictor (#6811), devnet deploy configs. Some small Amsterdam fixes have standalone PRs (#6913-#6917); they're folded here so the fixture bump is self-contained and reconcile trivially if they land first.

**Reconciliations to review:**
- `mempool.rs`: `transaction_intrinsic_gas` uses main's fork-general form (from #6892) + `sender` arg for Amsterdam `intrinsic_gas_dimensions`; Amsterdam special-case dropped since the fork-general path covers it.
- `fork_choice.rs`: one-line `.await` added for now-async `remove_stale_blob_txs`.
- `eip8038_tests.rs`: 4 SSTORE-refund test expectations corrected (restore-to-original refunds full STORAGE_WRITE = 10000; warm new-slot charges warm access + STORAGE_WRITE = 10100); matches EELS at pinned eels_commit.
- `mempool_tests.rs`: main's #6892 prague/validation tests merged with branch's Amsterdam tests on the 4-arg signature.

**Verification:** `cargo check` + clippy clean; 350/350 integration tests pass; state EF-tests against real v6.1.0 fixtures: 85494/85494 (100%), all forks including Amsterdam.

## Supersedes: these PRs can be closed

This PR contains these standalone Amsterdam fixes line-for-line, so they are now redundant and can be closed:

- #6913: fix(levm): match EELS system-call gas accounting under EIP-8037
- #6914: fix(l1): use block gas limit for the Amsterdam eth_estimateGas ceiling
- #6915: fix(levm): preserve balance in EIP-8246 selfdestruct block access list
- #6916: feat(l1): add EIP-8282 builder predeploys to l1-bal genesis
- #6917: fix(l1): run and correctly grade Amsterdam state ef-tests


crates/vm/levm/src/hooks/default_hook.rs
```diff
@@ -380,46 +434,49 @@ pub fn pay_coinbase(vm: &mut VM<'_>, gas_to_pay: u64) -> Result<(), VMError> {
...
-                    None
-                }
-            })
-            .collect();
-
-        // Sort by address (lexicographical order per EIP-7708)
-        addresses_with_balance.sort_by_key(|(addr, _)| *addr);
-
-        for (addr, balance) in addresses_with_balance {
-            let log = create_burn_log(addr, balance);
```
iovoid, line 403: create_burn_log becomes dead code with this

crates/vm/backends/levm/mod.rs
```diff
@@ -106,36 +106,36 @@ fn check_gas_limit(
...
 ///
 /// Note: `block_gas_used_regular` here equals EELS's `block_output.block_gas_used`
 /// because our `report.gas_used` already reflects `max(raw_regular, calldata_floor)`
 /// per-tx — i.e. the floor is applied before aggregation, not after. Keep this in
 /// sync with the aggregation loop in [`execute_block_parallel`].
 pub fn check_2d_gas_allowance(
     tx: &Transaction,
-    fork: Fork,
+    _sender: Address,
+    _fork: Fork,
```
iovoid: Are we leaving these in case of further changes?

ilitteri (approved):

test/tests/levm/eip8038_tests.rs
```diff
@@ -0,0 +1,1216 @@
...
+//! SSTORE sequences through a full VM and asserting the observable pre-refund
+//! regular gas and net regular refund (`gas_refunded`) for each row of the
+//! EIP-8038 SSTORE table, with `Fork::Osaka` controls proving pre-Amsterdam
+//! accounting is byte-identical.
+//!
+//! Expected values follow the EELS Amsterdam reference
+//! (ethereum/execution-specs `amsterdam/vm/instructions/storage.py::sstore`):
+//! a warm first change costs `COLD_STORAGE_WRITE - COLD_STORAGE_ACCESS`
+//! (= 13000 - 3000 = 10000), and a restore-to-original refunds
+//! `COLD_STORAGE_WRITE - COLD_STORAGE_ACCESS - WARM_ACCESS` (= 9900), so the
```
iovoid: The test below seems to disagree?
> set-then-clear net refund (full STORAGE_WRITE) must be 10000

iovoid (changes requested):

crates/vm/levm/src/hooks/default_hook.rs
edg-l: Right, removed create_burn_log in d3581b8. Kept BURN_EVENT_TOPIC since the tests still use it for the negative assert_no_burn_log check.

crates/vm/backends/levm/mod.rs
edg-l: No reason to keep them, they're leftover from when the check subtracted intrinsic gas. Dropped both params and updated the call sites in d3581b8.

test/tests/levm/eip8038_tests.rs
edg-l: Good catch, the doc was wrong, not the test. At the fixture tag tests-glamsterdam-devnet@v6.1.0 the restore-to-original refund is the full STORAGE_WRITE (10000):

    if original_value == new_value:
        evm.refund_counter += int(GasCosts.STORAGE_WRITE)

The 9900 form (COLD_STORAGE_WRITE - COLD_STORAGE_ACCESS - WARM_ACCESS) only landed later in #2972, after v6.1.0. Fixed the module doc and the stale 9900 section comment to 10000 in d3581b8.

crates/blockchain/blockchain.rs
```diff
@@ -792,22 +792,33 @@ impl Blockchain {
...
+                        // a channel is created (and execution streams per-tx updates into it via
+                        // `send_state_transitions_tx`) in every case except `bal=Some &&
+                        // parallel_exec`. So the optimistic synthesized-updates path is valid ONLY
+                        // when no channel exists (`rx_for_merkle` is None); whenever a channel was
+                        // created we must consume it. Taking the optimistic path while a channel is
+                        // live drops `rx` mid-execution and races execution's later sends (the
+                        // post-requests send especially), surfacing as "sending on a closed channel".
+                        let (account_updates_list, streaming_witness) = match rx_for_merkle {
+                            None => {
+                                let prepared = optimistic_updates.expect(
```
ElFantasma, line 818: Real correctness fix worth calling out — the old `if optimistic_updates.is_some() { use_opt() } else { use_rx() }` branch order took the optimistic path whenever both were present, dropping `rx` mid-execution and racing the per-tx sends that the sequential LEVM path (or any post-requests send) still does. The new `match rx_for_merkle` orientation makes channel presence the decider, which matches the channel-creation condition at l.693 (`optimistic_updates.is_some() && bal_parallel_exec_enabled` → `(None, None)`; everything else → `Some((tx, rx))`). So the `.expect()` here is sound by construction.

One small hardening to make the invariant bidirectional: add a `debug_assert!` at the `(None, None)` arm of the channel-creation site (l.694-695) so any future change to that decision tree fails fast in tests rather than surfacing here as a panic in production:

```rust
let (tx, rx_for_merkle) =
    if optimistic_updates.is_some() && bal_parallel_exec_enabled {
        // Paired with the `.expect` in the merkleizer below: when rx is None,
        // the optimistic path is the only valid one.
        debug_assert!(
            optimistic_updates.is_some() && bal_parallel_exec_enabled,
            "channel skipped only when both optimistic updates and parallel exec are on"
        );
        (None, None)
    } else {
        let (tx, rx) = channel();
        (Some(tx), Some(rx))
    };
```

(The condition is the same as the `if`, so the assert is trivially true today — but if someone later refactors the condition split, the assert codifies what the merkleizer is relying on.)

Non-blocking — the current code is correct.

ElFantasma (approved):
Large Amsterdam/glamsterdam-devnet v6.1.0 EVM bring-up (+6226/-508, 45 files — though ~5000 lines are new EIP test suites). Scope review:

- **Fork gating**: every EIP code path (8037, 8038, 2780, 7702, 7708, 7928, 8246, 8282) is `fork >= Amsterdam`-gated. Pre-Amsterdam behavior is preserved (`tx_base_cost`, `cold_account_access_cost`, `cold_storage_access_cost`, `access_list_*_cost`, `call_positive_value_cost`, `selfdestruct_base`, `balance` etc. all keep the pre-fork constants in their `else` arm). Risk to Prague mainnet is essentially nil.
- **EIP-7778 fix in `refund_sender`**: the `effective_regular = regular_gas.max(floor)` was being used for `ctx_result.gas_used` (block-level), but EELS only floors the user payment (`gas_spent`). Now uses `regular_gas` unfloored for block gas, floor only for the user side. Real correctness fix; matches EELS `tx_regular_gas = tx_gas_used_before_refund - state_gas`.
- **EIP-8246 BAL `track_selfdestruct(preserve_balance)` plumbing**: signature change, single caller in `default_hook::delete_self_destruct_accounts` passes `vm.env.config.fork >= Fork::Amsterdam`. Pre-Amsterdam path is unchanged (`!preserve_balance` collapses balance changes to final 0 as before). Verified.
- **EIP-7928 validation tightening**: `empty slot_changes` is now rejected as malformed at validation time. Good.
- **EIP-8282 predeploys + genesis**: builder deposit/exit predeploys added to `SYSTEM_CONTRACTS` (now 7) and `AMSTERDAM_REQUEST_PREDEPLOYS` mirrors the PRAGUE_SYSTEM_CONTRACTS empty-code-failure rule for Amsterdam+. Request types 0x03/0x04 added. Genesis updates in both `execution-api.json` and `l1-bal.json` use Nick's-method addresses matching sys-asm#43.
- **Merkleizer race fix at `blockchain.rs:803`**: NOT Amsterdam-gated — applies to current main. One inline on this with a small hardening suggestion.
- **Stateless ef-test bundle skip**: honestly documented in `docs/known_issues.md` with a clear re-enable path (when zkevm bundle is filled for v6.1.0). The `fork_Amsterdam` skip is `cfg(feature = "stateless")` only so `test-levm` still covers these EIPs against v6.1.0 fixtures.
- **Supersedes claim verified**: #6913, #6914, #6915, #6916, #6917 are all already CLOSED — the PR description is accurate.

No blockers. The volume here is reviewable mainly because almost every EVM change is one of `if fork >= Fork::Amsterdam { ... }` and the test files are end-to-end coverage rather than fork-cross logic.

iovoid (approved):

PR #6930, opened 2026-06-29
fix(l1): gate trie-layer commit/prune on canonical depth

edg-l:
## Problem

Full-sync ethrex nodes can wedge permanently, looping:

```
Full sync cannot resume: post-state for block 0 is absent (pruned from the layered store, or never executed). ... resume_parent_number=0 local_head=0
```

Observed on glamsterdam-devnet-6 (prysm-paired ethrex nodes). This is the **Q2** half of the incident cluster — see #6921 (trigger: peers not serving headers by hash) and #6924 (Q1: syncer lock held while waiting for a forkchoice head). Q1 (the lock) was masking Q2 (this state-resume wedge); once the syncer isn't pinned, the node actively cycles and hits this.

Root cause: trie diff-layers were committed/pruned by a fixed-depth count (`DB_COMMIT_THRESHOLD`) **regardless of canonicality**. Under a stale consensus forkchoice, non-canonical `newPayload` blocks executed ahead of canonicalization tripped that threshold, flushing non-canonical state to the single path-keyed on-disk root and **overwriting genesis**. Since those blocks never canonicalized (`local_head` stayed 0), the full-sync resume gate (`is_resume_point = is_canonical_sync && has_state_root`) found no canonical+stateful block and paused every cycle.

## Fix

Split the trie diff-layer disk-commit gate by execution mode:

- **Live block-by-block execution (`newPayload`)** commits on **CANONICAL + DEPTH**:
  - In `forkchoice_update` (after canonicalization), the Store computes `safe_commit_root` = the state_root of the canonical block at `head - commit_threshold`, and writes it into a dedicated `Arc<RwLock<H256>>` cell shared with `TrieLayerCache`.
  - The worker's `get_commitable(parent_state_root)` returns that root only if it is found on the executed parent walk; otherwise it commits nothing. `H256::zero()` (chain shorter than the threshold, e.g. genesis init) → no commit, so genesis-on-disk is never gated away.
  - The cell is written only by the FCU path (post-canonicalization) and is disjoint from the worker's RCU cache writes — no lost-update race, no `std::RwLock` guard held across `.await`.
  - **Non-canonical state is never committed** → genesis (and the canonical floor) is preserved → the resume gate always has a canonical+stateful block. Wedge fixed.

- **Batch execution (full sync / block `import`)** commits on **DEPTH** (`BATCH_COMMIT_THRESHOLD`), via `get_commitable_by_depth`:
  - Batch mode inserts one trie layer per ~1024-block batch, so the canonical `head - commit_threshold` (128) safe-commit root never lands on a layer boundary — the canonical gate would never flush, so memory would grow with the whole unfinalized window and a restart would leave a canonical/stateless suffix to re-execute.
  - Committing by depth bounds memory (~`BATCH_COMMIT_THRESHOLD` batch layers) and keeps state durable across restart. This is sound because full sync and `import` only ever extend a **single canonical chain** — there are no competing forks to mis-commit, so the non-canonical hazard the canonical gate guards against cannot arise here.
  - `commit_trie_if_due(.., is_batch)` selects the gate; `is_batch` is the existing `batch_mode`/`wait_for_flush` flag.

- A compile-time `const _: ()` assert requires `DB_COMMIT_THRESHOLD >= REORG_DEPTH_LIMIT`: the committed block (`head - DB_COMMIT_THRESHOLD`) must be at least as deep as the deepest reorg the engine API still accepts, so an accepted reorg can never revert already-committed state.

## Testing

- White-box gate unit tests in `layering.rs`:
  - `live_canonical_gate_holds_while_depth_gate_would_commit` — the wedge regression: with nothing canonicalized (`safe_commit` cell zero) the canonical gate commits nothing while a depth gate would prune genesis.
  - `batch_depth_gate_flushes_without_safe_commit` — the batch regression iovoid flagged: the canonical gate never flushes batch layers while `safe_commit` is zero, but the depth gate flushes the layer `BATCH_COMMIT_THRESHOLD` deep, bounding memory.
  - plus the memory-bound and ancestor-walk assertions.
- Public-API integration test `test/tests/blockchain/canonical_commit_gate_tests.rs`: non-canonical imports (no FCU) never prune genesis.
- `is_resume_point` tests relocated to `test/tests/p2p/`.
- Validated locally with the CI commands: `cargo test --workspace --exclude 'ethrex-l2*' --exclude ethrex-prover --exclude ethrex-guest-program` (green), `make -C tooling/ef_tests/blockchain test` (green — **0** "state root missing for block 0"), `make -C tooling/ef_tests/engine test` (green), `cargo clippy --workspace … -D warnings` (clean).

## Relationship to #6924 / #6921

Complementary, different files (this PR: `storage`/`blockchain`; #6924: `sync_manager.rs`). #6924's description left this exact "Full sync cannot resume … `removedb`" case as a separate change, suggesting an automatic reset-and-resync; this PR takes the **preventive** angle (never strand the resume gate in the first place), which should make a recovery mechanism unnecessary for this cause. Worth aligning on the Q2 direction.


ilitteri (approved):

crates/storage/store.rs
```diff
@@ -261,9 +263,9 @@ pub struct UpdateBatch {
     pub receipts: Vec<(H256, Vec<Receipt>)>,
     /// Contract code updates (code hash -> bytecode).
     pub code_updates: Vec<(H256, Code)>,
-    /// Whether this batch comes from full sync (batch execution mode).
-    /// When true, uses `BATCH_COMMIT_THRESHOLD` (aggressive) instead of
-    /// `DB_COMMIT_THRESHOLD` to bound memory during bulk block import.
+    /// Whether this batch comes from full sync (batch execution mode). When true,
+    /// the persist worker waits for the disk flush (`wait_for_flush`) to bound
+    /// in-flight batches during bulk import.
     pub batch_mode: bool,
```
ElFantasma, line 276: Batch full-sync question. In batch mode, `apply_updates` registers ONE layer per `UpdateBatch` keyed by the end-of-batch `last_state_root` (spans ~1024 blocks). `compute_safe_commit_root(head)` returns the state root of the canonical block at `head - commit_threshold` = `head - 128` — that's per-block granularity, so it rarely lands on a batch boundary. The returned root is then not a layer key, `get_commitable` walks past it, and returns None. Iterating that: batch 1 finishes at block 1024, FCU sets safe_commit to block 896's state_root (mid-batch, no layer) → None; batch 2 to block 1920 → still no layer → None; and so on. Layers accumulate at ~1 GB per batch — up to ~128 layers before any pressure kicks in — whereas the removed `BATCH_COMMIT_THRESHOLD = 4` capped this at ~4 batches.

The description says "Memory is bounded to ~commit_threshold layers in all cases", so I'd like to sanity-check my read. Is there a mechanism that snaps `safe_commit_root` to a batch boundary in batch mode (e.g., the full-sync driver issues FCU only at block numbers where head-128 is on a boundary), or is the batch-mode cadence something that will be looked at in follow-up? Not blocking — just want to make sure I understand the intended behavior.

ElFantasma (approved):

crates/blockchain/fork_choice.rs
```diff
@@ -27,6 +27,15 @@ use crate::{
 /// - geth / nethermind / reth: no engine-API rejection; trust the CL's fork choice
 pub const REORG_DEPTH_LIMIT: u64 = 128;
 
+// Compile-time guard: committed trie layers must never exceed the reorg-depth limit.
+// If DB_COMMIT_THRESHOLD > REORG_DEPTH_LIMIT, a committed layer could be unreachable
+// after a reorg that the engine API still accepts, corrupting state.
+// ethrex-blockchain depends on ethrex-storage (one-way), so the assertion lives here.
+const _: () = assert!(
+    ethrex_storage::DB_COMMIT_THRESHOLD as u64 <= REORG_DEPTH_LIMIT,
```
iovoid: Isn't it the other way around? If DB_COMMIT_THRESHOLD is _lower_ than the reorg limit, we would commit blocks that could be reorged.

iovoid (changes requested):
In batch execution (node offline for days), we never flush since the last layer (1024 blocks away) will never match the head minus the commit threshold (128 blocks away).

In import, we now keep every executed block until the FCU at the end.

edg-l:
Pushed `0d500937a` addressing all three points:

- **Batch execution never flushes / import keeps every block**: `commit_trie_if_due` now selects the gate by `is_batch` (= `batch_mode`). Batch execution (full sync and `import`) commits by depth via `get_commitable_by_depth(BATCH_COMMIT_THRESHOLD)` (4 layers, ~4k blocks) so memory stays bounded and state is durable across restart; live block-by-block execution keeps the canonical `head - 128` safe-commit gate so non-canonical newPayload state is never persisted. Sound because full sync and import only extend a single canonical chain. Added `batch_depth_gate_flushes_without_safe_commit` covering the case where the safe-commit cell is still zero.
- **Assertion direction**: you were right, flipped to `DB_COMMIT_THRESHOLD >= REORG_DEPTH_LIMIT` with the comment corrected: the committed block (`head - threshold`) must be at least as deep as the max accepted reorg, else an accepted reorg could revert already-committed state.

Storage + blockchain suites green, clippy clean.

edg-l:
Heads-up on a collision with the deep-reorg cap-lift PR #6724 (part of the #6685 stack):

This PR adds the compile-time guard `assert!(DB_COMMIT_THRESHOLD >= REORG_DEPTH_LIMIT)`, which references the flat `REORG_DEPTH_LIMIT = 128` const. #6724 **removes** `REORG_DEPTH_LIMIT` entirely and replaces the fixed cap with a *physical* reorg ceiling (the deeper of the layer-cache retention and the reverse-diff journal reach; finality is no longer a limit). So whichever of the two merges second will conflict, and if #6724 lands first this guard references a deleted const.

Resolution direction (whoever rebases onto the other): the guard's intent, do not commit a trie layer that is still inside the reorg window, is preserved by the physical model, since the window is now exactly the layer-cache retention. Once #6724 is in, re-key the guard off the retention value (`Store::reorg_retention` / the layer-cache commit threshold) instead of `REORG_DEPTH_LIMIT`; the two become the same quantity, so the invariant holds by construction. Flagging so it is a deliberate reconcile, not a surprise at merge time.

crates/blockchain/fork_choice.rs
edg-l: Right, flipped it in 0d500937a: assert is now `DB_COMMIT_THRESHOLD >= REORG_DEPTH_LIMIT`.

crates/storage/store.rs
edg-l: Confirmed the read was right; fixed in 0d500937a. Batch mode now commits by depth (`BATCH_COMMIT_THRESHOLD=4`) instead of the canonical head-128 root, which never lands on a ~1024-block layer boundary. Live mode keeps the canonical safe-commit gate.

iovoid (approved):

PR #6931, opened 2026-06-29
feat(l1): add debug_traceCall and debug_traceBlockByHash

edg-l:
Adds two geth-style debug tracing endpoints and tidies up the block-trace path while doing so.

This started as `debug_traceCall` but grew to cover `debug_traceBlockByHash` too, since Dora calls that on every block on the devnet.

> Note: this is an alternative take on the now-closed #6695. Same feature, independent implementation, with the block-trace perf work added.

## What's new

**`debug_traceCall(callObject, blockNrOrHash, traceConfig)`**
- All three tracers: `callTracer`, `prestateTracer`, `opcodeTracer`.
- Block-or-hash plus EIP-1898 identifiers, defaults to `latest`.
- Supports geth's `txIndex` config field (trace the call against the state just before tx `i`). Out-of-range `txIndex` is rejected to match geth instead of silently tracing against a bogus state.
- `stateOverrides` / `blockOverrides` are not supported (documented in the config struct).
- Reuses the existing `eth_call` machinery: the call is an unsigned `GenericTransaction`, so the sender comes from the `from` field with no signature recovery, and fees/base-fee are relaxed the same way `simulate_tx_from_generic` does.

**`debug_traceBlockByHash(blockHash, traceConfig)`**
- We already had `debug_traceBlockByNumber`; this mirrors it for a block hash. The two now share a single `trace_block` dispatch helper.

## Performance

Dora hits these on every block, so two things got optimized:

1. **`debug_traceCall` no longer re-executes the block.** For the default case (no `txIndex`) it reads the block's committed post-state directly, like `eth_call`, instead of rebuilding the parent state and replaying every transaction. Falls back to rebuild + rerun for a specific `txIndex` or when the block's state isn't stored (archive/pruned gap). A default `traceCall` on an N-transaction block drops from ~N+1 EVM executions to 1.

2. **`debug_traceBlock*` traces the whole block in one pass.** Replaced the per-transaction `spawn_blocking` + `Arc<Mutex>` loop with a single blocking pass, and the block-invariant `EVMConfig` is computed once instead of per transaction. Removes N-1 thread handoffs and N config recomputes per block trace. Side effect: the trace timeout now bounds the whole block trace rather than each transaction, which is the more sensible budget for this endpoint anyway.

## Implementation notes

Layered the same way as the existing tracers:
- `crates/vm/backends/levm/tracing.rs`: refactored the three `trace_tx_*` fns to share `run_*` cores, added `trace_call_*` (generic-tx) and `trace_block_*` (batch) variants on top.
- `crates/vm/tracing.rs`: `Evm` wrappers.
- `crates/blockchain/tracing.rs`: `Blockchain::trace_call_*` and the rewritten one-pass `trace_block_*`.
- `crates/networking/rpc/tracing.rs`: `TraceCallRequest`, `TraceBlockByHashRequest`, shared `trace_block` helper.
- `crates/networking/rpc/rpc.rs`: routes both new methods.

## Tests

Three VM-level tests in `test/tests/levm/trace_call_tests.rs` covering the call / opcode / prestate-diff paths, pinning the from-field sender behaviour. The existing opcode and prestate tracer tests still pass after the core refactor.

## Possible follow-up (not in this PR)

On Amsterdam the block carries a BAL (EIP-7928), which records per-transaction state changes. That makes it possible to reconstruct each transaction's pre-state from the parent state plus the BAL without replaying predecessors, which would let `debug_traceBlock*` trace transactions in parallel instead of sequentially. Out of scope here (correctness risk, needs the per-block BAL available at trace time), but worth a separate issue.

## callTracer / `debug_traceCall` geth-compatibility fixes

Follow-up on a glamsterdam-devnet-6 report: execution was consensus-correct, but the callTracer reporting layer diverged from geth's de-facto output. Fixed by comparing against go-ethereum (`eth/tracers/native/call.go`):

- **`gasUsed` (correctness bug):** the top-level frame now reports post-refund gas — matching the receipt and geth's `callstack[0].GasUsed = receipt.GasUsed` — instead of the pre-refund / EIP-7778 block-accounting value. Previously overstated gas on refund txs (e.g. by the full EIP-3529 refund).
- **`debug_traceCall` nonce:** fills the nonce from account state when omitted (matching geth's `ToMessage` and our own `eth_estimateGas`) instead of rejecting the call on a nonce mismatch against a default of 0.
- **`revertReason`:** decodes the ABI `Error(string)` payload (geth's `abi.UnpackRevert`) instead of raw UTF-8, which produced `null` for standard `revert("...")`.
- **error strings:** revert maps to `"execution reverted"`; static halts map to geth's wording (`out of gas`, `write protection`, `contract address collision`, `max code size exceeded`, …).
- **JSON shape:** omit empty `to`/`value`/`output`/`error`/`revertReason`/`calls` like geth's `omitempty`; drop `to` on a failed CREATE; omit `value` for STATICCALL.
- **logs:** clear logs of any frame whose ancestor reverted (geth's `clearFailedLogs`), and emit a block-absolute log `index` (geth's `log.Index`) for `withLog` traces.

**Scope / compatibility:** these changes touch only the **callTracer** (`CallTraceFrame` / `CallLog` / `LevmCallTracer`). The **opcodeTracer** (EIP-3155 `Eip3155Step`, consumed by the goevmlab `statetest` subcommand from #6663 / #6595), the **prestateTracer**, and `VMError`'s `Display` (which feeds the execution-spec-tests / hive error mapping) are all left untouched — verified against branch history, so goevmlab output is byte-for-byte unchanged.

Adds 9 VM-level tests in `test/tests/levm/trace_call_tests.rs` covering post-refund gas, revert-reason decode, geth error mapping, omit-empty shape, and block-absolute log index.


crates/networking/rpc/tracing.rs
```diff
@@ -249,80 +277,291 @@ impl RpcHandler for TraceBlockByNumberRequest {
...
+        // Fill the nonce from account state when the caller omits it, matching geth's
+        // `ToMessage` (`args.Nonce = db.GetNonce(from)`) and `eth_estimateGas`. Without this
+        // the VM's nonce check compares the account's real nonce against a default of 0 and
+        // rejects the call.
+        let transaction = match self.transaction.nonce {
+            Some(_) => self.transaction.clone(),
+            None => {
+                let nonce = context
+                    .storage
+                    .get_nonce_by_account_address(block_number, self.transaction.from)
```
iovoid, line 521: Considering the post-state will lead the transaction to fail due to a nonce mismatch when tx_index is set (ie it isn't added on top of the block).

This isn't a problem in geth because they skip the nonce check.

iovoid (dismissed):

crates/networking/rpc/tracing.rs
```diff
@@ -26,6 +31,29 @@ pub struct TraceBlockByNumberRequest {
...
+/// `debug_traceCall`'s third parameter. Extends [`TraceConfig`] with `txIndex`, the
+/// in-block transaction index whose pre-state the call should run on top of (geth's
+/// `TraceCallConfig`). `stateOverrides`/`blockOverrides` are not supported.
+#[derive(Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+struct TraceCallConfig {
+    #[serde(flatten)]
+    base: TraceConfig,
+    #[serde(default)]
+    tx_index: Option<u64>,
```
ElFantasma, line 58: The doc above says `stateOverrides`/`blockOverrides` "are not supported", but with `#[serde(flatten)]` on `base` and no `deny_unknown_fields` on `TraceCallConfig`, a client that sends either field gets its trace run without the overrides — silently, with no error. Geth's `TraceCallConfig` includes both, so this is a fairly high-probability compat mismatch (a client relying on geth semantics gets a trace against unexpected state). Two loud-fail options: `#[serde(deny_unknown_fields)]` on the whole struct, or explicit `state_overrides: Option<Value>` / `block_overrides: Option<Value>` fields that reject with `BadParams("stateOverrides not supported")` when `Some`. The explicit form has the advantage of surviving geth adding *other* new fields in the future without breaking us.

crates/vm/backends/levm/tracing.rs
```diff
@@ -158,6 +280,160 @@ impl LEVM {
...
+                base_blob_fee,
+            )?;
+            let trace = Self::run_opcode_trace(db, env, tx, cfg.clone(), vm_type, crypto)?;
+            traces.push((tx.hash(crypto), trace));
+        }
+        Ok(traces)
+    }
+}
+
+/// Computes the block-invariant `(EVMConfig, chain_id, base_blob_fee)` once so a
```
ElFantasma: Two docstrings got glued: the "Computes the block-invariant …" sentence describes `block_trace_env_config` (line 411, currently undocumented), while "Recursively counts the logs …" is the one that belongs above `count_call_logs`. Splitting them puts each doc above its function.

ElFantasma (approved):

ElFantasma (approved):

ilitteri (approved):

crates/blockchain/tracing.rs
```diff
@@ -199,24 +192,109 @@ impl Blockchain {
...
-            let block = block.clone();
-            let vm = vm.clone();
-            let tx_hash = block.as_ref().body.transactions[index].hash(&NativeCrypto);
-            let cfg = cfg.clone();
-            let result = timeout_trace_operation(timeout, move || {
-                vm.lock()
-                    .map_err(|_| EvmError::Custom("Unexpected Runtime Error".to_string()))?
-                    .trace_tx_opcodes(block.as_ref(), index, cfg)
-            })
+        timeout_trace_operation(timeout, move || vm.trace_block_opcodes(&block, cfg)).await
```
iovoid, line 195: Now the timeout is on the whole block (which is substantially heavier). If there is a timeout (presumably because there is heavy load) then leaving it working while possibly even more requests to trace are being processed could lead to a big degradation.

crates/vm/backends/levm/tracing.rs
```diff
@@ -158,6 +280,160 @@ impl LEVM {
...
+    let base_blob_fee = get_base_fee_per_blob_gas(header.excess_blob_gas, &config)?;
+    Ok((config, chain_config.chain_id, base_blob_fee))
+}
+
+/// Builds the call-style [`Environment`] and concrete [`Transaction`] for tracing a
+/// synthetic `eth_call`-shaped request (`debug_traceCall`). Mirrors
+/// `simulate_tx_from_generic`: the sender is taken from `tx.from` (no signature
+/// recovery), the block gas limit is disabled, and the base fee is relaxed when no gas
+/// price is provided.
+fn prepare_call_env(
```
iovoid, line 458: Might not work on L2

crates/vm/levm/src/tracing.rs
```diff
@@ -111,17 +125,37 @@ impl LevmCallTracer {
             return Ok(());
         }
         if is_top_call {
-            // After finishing transaction execution clear all logs of callframes that reverted.
-            clear_reverted_logs(self.current_callframe_mut()?);
+            // After finishing transaction execution clear all logs of callframes that reverted,
+            // then assign block-absolute indices to the survivors in emission order.
+            clear_reverted_logs(self.current_callframe_mut()?, false);
```
iovoid: callframe.error isn't set at this point

PR #6940, opened 2026-06-30
perf(l1): use 4KB RocksDB blocks for trie-node and flat-KV CFs

edg-l:
## What

Lower the RocksDB data-block size from 16KB to 4KB on the four execution-read-path column families:

- `ACCOUNT_TRIE_NODES` / `STORAGE_TRIE_NODES`
- `ACCOUNT_FLATKEYVALUE` / `STORAGE_FLATKEYVALUE`

## Why

These CFs are pure exact-key point lookups (trie nodes and flat KV) hit on the hot execution read path; there are no range scans. With a 16KB block, every cold random get pulls a full 16KB block off disk to return a ~32-byte value — ~4x read amplification. A page-sized 4KB block minimizes per-get read + in-block search amplification.

`set_block_size` is a write-time option: it applies to newly flushed/compacted SSTs. Existing SSTs are read with whatever block size they were written at, so the benefit accrues as the DB rewrites (or after a regen/reblock of a snapshot).

## Measurement

A/B on a cold, RAM-exceeding bloated state DB (`sstore_bloated` 10GB-class prestate, 300M-gas block), reblocking a copy of the same DB from 16KB to 4KB:

| metric | 16KB | 4KB | delta |
| --- | --- | --- | --- |
| Mgas/s | 135.4 | 160.1 | **+18%** |
| merkle drain | 1305 ms | 1104 ms | -15% |
| disk read | — | — | **-36%** |

The win is read-amplification driven, so it shows up most on cold, larger-than-RAM state where reads actually hit the device.

## Scope

Bloom filters (`set_bloom_filter(10.0, false)`) and `memtable_prefix_bloom_ratio` were already present on these CFs; this PR only changes the block size. No behavior change to other CFs (headers/bodies, codes, receipts, tx locations stay at their current sizes).


pablodeymo (approved):

iovoid (approved):

crates/storage/backend/rocksdb.rs
```diff
@@ -190,7 +190,12 @@ impl RocksDBBackend {
...
 
                     let mut block_opts = BlockBasedOptions::default();
-                    block_opts.set_block_size(16 * 1024); // 16KB
+                    // 4KB blocks: these CFs are pure exact-key point lookups
+                    // (trie nodes / flat KV) on the execution read path, so a
+                    // page-sized block minimizes per-get read+search amplification
+                    // vs the prior 16KB (scan-tuned). Write-time option: applies to
+                    // newly flushed/compacted SSTs; existing SSTs are read as-is.
+                    block_opts.set_block_size(4 * 1024); // 4KB
                     block_opts.set_bloom_filter(10.0, false); // 10 bits per key
```
ElFantasma, line 256: Non-blocking follow-up thought: given the ~4x index/filter block-count growth on these two large CFs (already flagged in the block-cache-size doc), what do you think about pairing this later with partitioned filters (`set_bloom_filter(10.0, true)` + a two-level partitioned index)? That's the standard mitigation for exactly this pattern — small data block + large SST → filter/index memory blowup — because filter loads become finer-grained under cache eviction instead of the whole per-SST filter block being pulled in. The doc update on `DEFAULT_ROCKSDB_BLOCK_CACHE_SIZE_BYTES` covers this from the sizing side; partitioning would attack the same pressure from the CF-config side. Not for this PR, just checking if you'd already considered it.

ElFantasma (approved):

ilitteri (approved):

PR #6942, opened 2026-06-30
ci(l1): run no_std check against zkVM standard target

iovoid:
**Motivation**

Previously we used the `riscv64imac` target, but the [standard](https://github.com/eth-act/zkvm-standards/blob/main/standards/riscv-target/target.md) is `riscv64im_zicclsm`.

**Description**

Sets the target to `riscv64im_zicclsm`. It's nonstandard, so it has to be built from a a target json.

Importantly, while the target does not support atomics we can enable lowering (turning atomic into non-atomic operations) because zkVMs run in single-core (hart in RISC-V terms).

### DIFF
--- a/.github/workflows/pr_nostd.yaml
+++ b/.github/workflows/pr_nostd.yaml
@@ -33,15 +33,18 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      # Pinned for reproducibility (build-std + json-target-spec are nightly-only);
+      # bump the date deliberately. dtolnay has no per-date ref, so pass it as input.
       - name: Install Rust nightly + rust-src
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@master
         with:
+          toolchain: nightly-2026-06-29
           components: rust-src
 
       - name: Check no_std crates against the zkVM standard target
         run: |
           TARGET=.github/targets/riscv64im_zicclsm-unknown-none-elf.json
-          cargo +nightly -Z json-target-spec -Z build-std=core,alloc check \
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_nostd.yaml
```diff
@@ -24,17 +25,23 @@ env:
...
+    # see .github/targets/README.md.
+    name: no_std build check (riscv64im_zicclsm)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
 
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@1.91.0
+      - name: Install Rust nightly + rust-src
+        uses: dtolnay/rust-toolchain@nightly
```
pablodeymo: you should pin the version to a specific date

iovoid: Pinned in 8f93e3c

ilitteri (approved):

edg-l (approved):

ElFantasma (approved):
Small concern about the nightly pin: `-Z json-target-spec` and `-Z build-std` are both unstable, so any of them could get renamed or gated in a future nightly. Because the workflow only runs on PRs touching `crates/common/` or `crates/vm/`, a breakage in a random nightly bump wouldn't show up until someone unrelated to CI opens such a PR and gets a red X — awkward to debug for whoever triggers it.

What do you think about either (a) adding a weekly scheduled canary that runs this same check on unpinned `nightly` (the repo already has `daily_snapsync` / `daily_hive_report` / `daily_loc_report` so it'd fit), or (b) just handling it on-break by bumping the pin when someone hits it? No strong preference — the canary catches drift early, but on-break bumps are lower-maintenance.

PR #6944, opened 2026-07-01
fix(l1): flush db on graceful shutdown

edg-l:
## Problem

`docker restart -t 0` (and any abrupt exit) leaves the db in an unclean state. The SIGTERM/SIGINT handler only cancelled the token, wrote `node_config.json`, and slept 1s. It never drained the persist worker, never force-flushed the tail of the block-data buffer, and never fsynced RocksDB.

Two problems on exit:
- The tail of the in-memory block-data buffer (blocks queued but below the flush threshold) was dropped, so a block a crash would lose was also lost on a *graceful* stop.
- RocksDB memtables were never flushed, so the next open did WAL recovery.

Result: slow, unclean restarts (WAL replay + re-execution). Users worked around it by waiting ~2 min before killing.

## Fix

Add `Store::shutdown()` and call it from `server_shutdown` after cancelling the token:

1. Sends a `Shutdown` handshake to the persist worker. Being FIFO, it drains every queued block first.
2. Worker force-flushes the block-data buffer (not just the "when due" tail).
3. After the worker acks, `backend.flush()` flushes every CF memtable to SST and syncs the WAL, so the next open needs no recovery.

The in-memory trie diff-layers are **intentionally not** force-committed. The on-disk trie is a single-version, path-based store, so folding the non-finalized tail into it would leave a post-restart reorg unable to reconstruct the overwritten ancestor state (the node would wedge). Those recent (< `DB_COMMIT_THRESHOLD`) layers are simply dropped and re-executed on the next start from the deep, reorg-safe on-disk base, exactly as after any restart today.

Supporting changes:
- `StorageBackend::flush()` trait method (default no-op; RocksDB impl flushes CFs + syncs WAL).
- New `PersistMessage::Shutdown` handshake handled by the persist worker.
- RPC/engine servers now key their graceful-shutdown future on the `cancel_token` (`rpc_shutdown_signal`), not just Ctrl+C. This makes the engine API stop accepting requests on **SIGTERM** (`docker stop`) before the store is flushed.
- `init_l1` returns the `Store` so the shutdown handler can reach it.

## Note

`-t 0` sends SIGKILL with no grace period, which no handler can intercept. Pair this with a real `stop_grace_period` / `docker stop -t N` so the handler runs.

### DIFF
--- a/cmd/ethrex/ethrex.rs
+++ b/cmd/ethrex/ethrex.rs
@@ -5,11 +5,12 @@ use ethrex::{
     utils::{NodeConfigFile, get_client_version, is_memory_datadir, store_node_config_file},
 };
 use ethrex_p2p::{peer_table::PeerTable, types::NodeRecord};
+use ethrex_storage::Store;
 use serde::Deserialize;
 use std::{path::Path, time::Duration};
 use tokio::signal::unix::{SignalKind, signal};
 use tokio_util::sync::CancellationToken;
-use tracing::info;
+use tracing::{error, info};
 
 const LATEST_VERSION_URL: &str = "https://api.github.com/repos/lambdaclass/ethrex/releases/latest";
 
@@ -50,9 +51,19 @@ async fn server_shutdown(
     cancel_token: &CancellationToken,
     peer_table: PeerTable,
     local_node_record: NodeRecord,
+    store: &Store,
 ) {
     info!("Server shut down started...");
+    // Stop feeding new blocks before draining, so the persist queue can't grow.
     cancel_token.cancel();
+    // Drain the persist worker, commit all in-memory trie diff-layers and the
+    // block-data buffer, and fsync the DB. Without this an abrupt exit (e.g.
+    // `docker restart -t 0`) loses up to ~128 blocks of in-memory state and
+    // leaves the DB needing WAL recovery on next start.
+    info!("Flushing database to disk...");
+    if let Err(err) = store.shutdown().await {
+        error!("Failed to flush database on shutdown: {err}");
+    }
     if !is_memory_datadir(datadir) {
         let node_config_path = datadir.join("node_config.json");
         info!("Storing config at {:?}...", node_config_path);
@@ -179,7 +190,7 @@ async fn main() -> eyre::Result<()> {
     info!("ethrex version: {}", get_client_version());
     tokio::spawn(periodically_check_version_update());
 
-    let (datadir, cancel_token, peer_table, local_node_record) =
+    let (datadir, cancel_token, peer_table, local_node_record, store) =
         init_l1(opts, Some(log_filter_handler)).await?;
 
     let mut signal_terminate = signal(SignalKind::terminate())?;
@@ -188,10 +199,10 @@ async fn main() -> eyre::Result<()> {
 
     tokio::select! {
         _ = tokio::signal::ctrl_c() => {
-            server_shutdown(&datadir, &cancel_token, peer_table, local_node_record).await;
+            server_shutdown(&datadir, &cancel_token, peer_table, local_node_record, &store).await;
         }
         _ = signal_terminate.recv() => {
-            server_shutdown(&datadir, &cancel_token, peer_table, local_node_record).await;
+            server_shutdown(&datadir, &cancel_token, peer_table, local_node_record, &store).await;
         }
     }
 
--- a/cmd/ethrex/ethrex.rs
+++ b/cmd/ethrex/ethrex.rs
@@ -56,10 +56,11 @@ async fn server_shutdown(
     info!("Server shut down started...");
     // Stop feeding new blocks before draining, so the persist queue can't grow.
     cancel_token.cancel();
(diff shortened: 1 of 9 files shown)


ElFantasma (approved):
Minor observation, not a change request: `head_state_root` is captured in the caller task before `spawn_blocking`, so an FCU-in-flight that arrives after capture but before the worker processes Shutdown would advance `latest_block_header` after the fact, and that newer block's trie layer would stay in memory (discarded on exit). Since block bodies/headers are flushed by `flush_block_data`, restart re-executes those blocks from the last durably-committed head. That's a graceful degradation, not a bug — and taking a stale snapshot is strictly better than the alternative (racing an unconfirmed head into the trie DB). Worth mentioning in the shutdown docstring so operators know a truncated tail may re-execute on the next start; the tail is bounded by the persist channel capacity + FCU lag.

`StorageBackend::flush()` trait default (no-op) is correct for InMemory and future backends; the RocksDB impl does `flush_cf` over `TABLES` + `flush_wal(true)`, which is the canonical clean-shutdown recipe. 

The body's `-t 0` caveat (SIGKILL bypasses any handler; pair with `stop_grace_period` / `docker stop -t N`) is the right thing to call out for operators.

crates/storage/store.rs
```diff
@@ -392,6 +392,41 @@ impl Store {
         .map_err(|e| StoreError::Custom(format!("wait_for_persistence_idle join: {e}")))?
     }
 
+    /// Flushes all in-memory state to disk for a clean shutdown.
```
ilitteri, line 395: Do we really want to flush all in-memory state to disk for a clean shutdown? Wouldn't state regeneration do the work?

crates/storage/store.rs
```diff
@@ -392,6 +392,41 @@ impl Store {
...
+    ///
+    /// After this returns the persist worker has exited; the store must not be
+    /// used for further writes. Idempotent only in the sense that a second call
+    /// errors on the closed channel — call it exactly once, on shutdown.
+    pub async fn shutdown(&self) -> Result<(), StoreError> {
+        let tx = self.persist_tx.clone();
+        let backend = self.backend.clone();
+        // Anchor the flush to the confirmed canonical head so we never durably
+        // persist a non-canonical sidechain root.
+        let head_state_root = self.latest_block_header.get().state_root;
```
iovoid: This isn't _confirmed_ in the sense of it being finalized.

iovoid (changes requested):

crates/storage/store.rs
edg-l: Net less work: at shutdown we commit already-computed trie diffs; if we skip it, the next boot regenerates that same state by re-executing the blocks (+ WAL replay), which costs more

crates/storage/store.rs
edg-l: Good catch — reworded to "latest canonical head (not finalized)". We anchor on the latest canonical head (`latest_block_header`), which is not a finality guarantee; the tail past the last finalized block can still reorg. Fixed in c24c3fdc3.

ilitteri (approved):

crates/storage/store.rs
edg-l: You were right. I've dropped the trie force-commit. The on-disk trie is a single-version, path-based store, so folding the non-finalized tail into it would leave a post-restart reorg unable to reconstruct the overwritten ancestor state — the node wedges. So state regeneration is exactly the right call for the trie: the recent (< `DB_COMMIT_THRESHOLD`) layers stay in memory and re-execute on next boot from the deep, reorg-safe on-disk base, same as after any restart. Shutdown now only force-flushes the block-data buffer and syncs RocksDB (memtables + WAL) to avoid WAL replay — the parts that are safe to persist.

crates/storage/store.rs
edg-l: Resolved by removing the trie commit entirely rather than rewording. Your point was the actual problem: anchoring on a non-finalized head is unsafe here because the trie is a single-version path store — committing the non-finalized tail makes a post-restart reorg unrecoverable. Anchoring on finalized would be safe but buys almost nothing over normal operation (layers ≥128 deep are already committed), so shutdown no longer force-commits trie layers at all; they re-execute on next boot. It keeps only the block-data flush + WAL/memtable sync.

cmd/ethrex/ethrex.rs
```diff
@@ -50,9 +51,19 @@ async fn server_shutdown(
     cancel_token: &CancellationToken,
     peer_table: PeerTable,
     local_node_record: NodeRecord,
+    store: &Store,
 ) {
     info!("Server shut down started...");
+    // Stop feeding new blocks before draining, so the persist queue can't grow.
     cancel_token.cancel();
+    // Drain the persist worker, commit all in-memory trie diff-layers and the
```
iovoid: No longer the case

iovoid (changes requested):
PR description doesn't reflect the removal of trie flushing logic

iovoid (approved):

PR #6945, opened 2026-07-02
fix(l1,l2)!: harden RPC startup (fail-fast bind, shared HTTP/WS port, honest logs)

ilitteri:
**Motivation**

A mainnet node couldn't sync for over an hour because `--http.port` and `--ws.port` were configured to the same `address:port` (a config carried over from a client that multiplexes WS onto the HTTP port). ethrex opens an independent `TcpListener` per endpoint, so the WebSocket `bind` failed with `EADDRINUSE`. Because bind and serve were fused in one function whose `Result` was discarded by `tracker.spawn`, the already-bound HTTP and Auth-RPC listeners were dropped — closing the Engine API (8551) the consensus client needs — and the failure was completely silent (the startup logs even asserted the servers had started, and the one diagnostic that should have fired was muted). A config typo produced the most silent failure possible.

**Description**

Reworks RPC startup so a misconfiguration fails loudly and the config that triggered the incident simply works:

- **Config-time validation:** conflicting listener addresses (duplicates, or same-port same-family wildcard overlaps like `0.0.0.0:8545` vs `127.0.0.1:8545`) are rejected right after CLI parsing, before anything binds, with an error naming **both** flags to change. Wildcard overlaps matter cross-platform: Linux fails the second bind, but macOS/BSD `SO_REUSEADDR` would let the specific bind silently shadow the wildcard listener. The check is a pure, unit-tested function in the cmd layer.
- **Fail-fast bind:** `start_api` is split into a foreground `bind_api` (returns a typed `RpcStartupError`) and a spawned `serve`; a genuine bind failure (e.g. another process holding the port) aborts the node with an error naming the role, address, and flag to change.
- **Shared HTTP/WebSocket port:** when WS resolves to the same address as HTTP they're served on one listener (`POST`→JSON-RPC, `GET`+Upgrade→WS), like geth/reth/nethermind; WS addr/port inherit HTTP's when unset. Auth-RPC is never merged (its 256 MB engine body limit stays private to the authenticated listener).
- **Honest logging:** listeners are logged only after they bind (`<role> listening on <addr>`).
- **Consensus-liveness fix:** the "no messages from the consensus layer" task no longer busy-loops or silently misses its warning.
- **Cancellation-aware lifecycle:** `main` and every server observe a cancellation token; a fatal subsystem (RPC serve, or the dev block producer) cancels it and the node shuts down orderly, then **exits non-zero** so orchestrators keying on exit status (systemd `Restart=on-failure`, Docker restart policies) can tell a crashed node from a clean stop. Ctrl+C and SIGTERM still exit zero. Metrics failures are logged instead of swallowed. The same bind/serve split and wiring is applied to the L2 RPC crate, threading the sequencer cancellation token so an L2 RPC failure and the sequencer shut down together.

Design notes made explicit for review:
- The `spawn_fatal` wiring on the RPC serve loop is **defensive future-proofing**: axum's serve loop retries accept errors internally and only returns after graceful shutdown, so its error arm is unreachable today; it exists so any future serve error aborts the node instead of being silently dropped.
- A bounded `tracker.wait()` drain was deliberately **not** added: several long-lived tasks (p2p accept loop, peer-stats, metrics) are not yet cancellation-aware, so a drain would burn its full timeout on every shutdown. Making them cancellation-aware is a tracked follow-up.

Verified live (`--dev --datadir memory`):
- Incident config (`--http.port 8545 --ws.enabled --ws.port 8545`, same addr) → boots on **one merged listener**; `POST eth_blockNumber` → `0x1`, `GET`+upgrade → `101 Switching Protocols`, node stays up.
- `ws == authrpc` (`--ws.port 8551`) → **exits 1** with `127.0.0.1:8551 is requested by both the Auth-RPC server and the WebSocket server; change --authrpc.port (or --authrpc.addr) or --ws.port (or --ws.addr).`
- Wildcard overlap (`--http.addr 0.0.0.0 --ws.addr 127.0.0.1`, same port) → **exits 1** with the overlap error naming both flags.
- Default start → separate `HTTP-RPC server listening on …:8545` + `Auth-RPC server listening on …:8551` lines, no phantom WS line.

**⚠️ Breaking change (external only):** enabling WebSocket without an explicit `--ws.port` now shares the HTTP port instead of defaulting to `8546`. No in-repo consumers rely on 8546. Operators who want a separate WebSocket port must pass `--ws.port` explicitly.

Deferred follow-ups (tracked, out of scope here): full L2 shutdown parity (SIGTERM arm + node-fatal token bridge) — including threading the sequencer token into the L2 admin server so it drains on a fatal shutdown rather than only on Ctrl+C (raised in review); making the p2p accept / peer-stats / metrics loops cancel-aware to enable a bounded task-drain; auditing `Syncer::start_sync` cancel-awareness; and the remaining automated tests (log-capture, merged-router dispatch, consensus-loop, integration/lifecycle).

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A: no `Store` changes, no re-sync required.


iovoid (approved):
I feel like the flag-showing system is a bit convoluted, but otherwise looks good.

ElFantasma (approved):
Two observations, both non-blocking:

1. **`FATAL_SHUTDOWN_CAUSE: OnceLock<String>`** is process-global and can never be cleared. In tests running in the same process, a first test that exercises the fatal-shutdown path would leave `fatal_shutdown_cause()` returning `Some(...)` for every subsequent test in that process — so any test asserting on clean-shutdown behavior after a fatal-path test is order-dependent. Not a production concern (process runs once, exits), just something to keep in mind if lifecycle tests get added later.

2. **L2 admin server uses a fresh, never-cancelled `CancellationToken`** (comment acknowledges this: "shuts down on Ctrl+C only, as before"). That's fine functionally since the process is going down anyway, but it means during a fatal-subsystem shutdown of the sequencer the admin listener stays bound until the process exits — an odd asymmetry now that the L2 RPC is wired to the sequencer token. Could easily follow-up-list if you'd rather keep this PR focused on the RPC surface.

edg-l (approved):

PR #6946, opened 2026-07-02
feat(l1): integrate feature-gated hotpath profiler

### LINKED ISSUE #6932: Integrate hotpath profiler (feature-gated, zero-cost when disabled)
I want to try wiring [hotpath](https://github.com/pawurb/hotpath-rs) into ethrex as an optional profiler.

It's a feature-gated Rust profiler that shows where code spends time, burns CPU, and allocates. The parts that look useful for us: timing + CPU + allocation breakdowns per function, async observability (futures, channels, streams) and Mutex/RwLock contention, Tokio runtime stats, a live TUI plus static reports, and a CI mode for catching regressions on PRs. Usage is just attributes:

```rust
#[hotpath::measure]
fn execute_block(...) { ... }

#[tokio::main]
#[hotpath::main]
async fn main() { ... }
```

Right now we lean on ad-hoc perf/flamegraph/hyperfine when something feels slow. Having a profiler always sitting behind a feature flag would let us instrument the obvious hot paths (block execution, levm, trie, RLP, RPC), look at allocations and lock contention without a separate setup, and ideally catch regressions in CI before they land.

The one thing I care about most, and the hard requirement for this to land:

**It has to be genuinely zero cost with the feature off.** No extra deps in the default build, no codegen, the attributes compile down to nothing. hotpath claims this is how it works, but we verify before merging:

- [ ] default `make build` dependency graph is unchanged (hotpath and its transitive deps only show up when the feature is on)
- [ ] no runtime regression on the default profile (benchmark block import or ef_tests with the flag off, confirm it's within noise)
- [ ] the attributes expand to no-ops when disabled
- [ ] both `cargo build` and `cargo build --features <hotpath-feature>` build, and CI's default jobs don't turn it on

Rough plan: add it as an optional dep behind a workspace feature (probably `hotpath` plus the `hotpath-cpu` / `hotpath-alloc` sub-features), put `#[hotpath::main]` on the `cmd/ethrex` entrypoint behind that feature, instrument block execution + levm as a first pass, and write up how to run it.

A few things to check while doing it: whether hotpath needs nightly or a global allocator override for the alloc feature and how that plays with our allocator, making sure it stays fully out of the zkVM guest builds (sp1, risc0), and whether the MCP-server integration is worth bothering with.

edg-l:
## What
Integrates the `hotpath` profiler (crate `hotpath` v0.19) into ethrex as an optional, feature-gated profiler, zero-cost when the feature is off.

Closes #6932

## Design
- Optional dep plus `#[cfg_attr(feature = "hotpath", hotpath::measure)]` instrumentation. `cfg_attr` erases both the attribute and the `dep:hotpath` edge when the feature is off, so the default `cargo tree` is provably unchanged.
- Three cargo features on `ethrex`: `hotpath` (timing), `hotpath-cpu` (CPU sampling), `hotpath-alloc` (allocation tracking). None in `default`.
- `hotpath = "0.19"` added to `[workspace.dependencies]`; optional dep carried by `cmd/ethrex`, `ethrex-blockchain`, `ethrex-levm`; `ethrex-vm` is a pure pass-through forwarder (`hotpath = ["ethrex-levm/hotpath"]`, no dep).
- Allocator handling: `hotpath-alloc` makes hotpath install its own `#[global_allocator]` (a `CountingAllocator<A>` wrapper). ethrex's jemalloc `#[global_allocator]` is cfg-disabled under `hotpath-alloc`, and jemalloc is passed as the backing allocator (`hotpath::main(allocator = tikv_jemallocator::Jemalloc)`), so allocations get counted then routed through jemalloc. Three mutually-exclusive `cfg_attr` arms on `main` cover: jemalloc-backed, System fallback (jemalloc off / msvc), and timing-only (no alloc). No nightly required.

## Instrumented (first pass)
`Blockchain::execute_block`, `Blockchain::execute_block_pipeline`, `Blockchain::store_block` (crates/blockchain/blockchain.rs), and `VM::execute` (crates/vm/levm/src/vm.rs). `measure` sits above the existing `#[instrument]` so it only times the body.

## Verification (all passing)
- Default `cargo tree -p ethrex`: 0 hotpath refs; `--features hotpath`: present.
- Builds: default, `--features hotpath`, `hotpath-cpu`, `hotpath-alloc` all compile clean, no new warnings, no double-global-allocator error (including the `--no-default-features` jemalloc-absent fallback arm).
- zkVM guest isolation: `cargo tree -p ethrex-guest-program` for both `sp1` and `risc0` shows 0 hotpath; the guest crate doesn't define a `hotpath` feature at all.
- **End-to-end block import**: ran a real block import under `hotpath-alloc` and confirmed the instrumented functions show up in the `[hotpath]` report (not just `ethrex::main`), so the `measure` wire-up is confirmed live during actual execution.
- Runtime smoke test: ran the node built with `hotpath-alloc`, sent SIGTERM, got a `[hotpath]` report printed (timing/alloc/thread tables); startup logged `CountingAllocator<jemalloc>`.

## Docs
`docs/developers/l1/profiling.md` gets a hotpath section: flags, run commands, instrumented functions, the multi-thread async-alloc caveat, the jemalloc interaction (including that `hotpath-alloc`'s counting wrapper adds overhead to every alloc path and inflates timing, so use plain `hotpath` for pure timing), follow-up targets, and a guardrail that hotpath must be enabled with `-p ethrex`, never `--workspace` (workspace-wide feature unification would leak it into the guest).

## Testing notes / known gaps
- No performance benchmark was run for the feature-off case; the no-regression claim rests on tree/codegen identity (default build never references hotpath).
- Follow-up instrumentation (trie, rlp, rpc, levm opcode loop) is deliberately out of scope for this first pass.


Makefile
```diff
@@ -70,6 +70,12 @@ dev: ## 🏃 Run the ethrex client in DEV_MODE with the InMemory Engine
 		--dev \
 		--datadir memory
 
+HOTPATH_FEATURES ?= hotpath,hotpath-alloc,hotpath-cpu
+dev-hotpath: ## 🔥 Run ethrex in DEV_MODE with the hotpath profiler (all signals: timing+alloc+cpu). Narrow with HOTPATH_FEATURES=hotpath. Report prints on shutdown.
+	cargo run $(PROFILING_CFG) --release --features $(HOTPATH_FEATURES) -- \
+		--dev \
+		--datadir memory
```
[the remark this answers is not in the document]
iovoid: Valid nit.

cmd/ethrex/ethrex.rs
```diff
@@ -152,6 +167,27 @@ async fn periodically_check_version_update() {
 }
 
 #[tokio::main]
+#[cfg_attr(
```
iovoid, line 187: Might be worth mentioning the order of these is important

docs/developers/l1/profiling.md
```diff
@@ -147,10 +147,113 @@ samply record ./target/release-with-debug/ethrex [ARGS]
...
+This first pass instruments four synchronous hot paths:
+
+- `Blockchain::execute_block` (`crates/blockchain/blockchain.rs`)
+- `Blockchain::execute_block_pipeline` (`crates/blockchain/blockchain.rs`)
+- `Blockchain::store_block` (`crates/blockchain/blockchain.rs`)
+- `VM::execute` (`crates/vm/levm/src/vm.rs`)
+
+### Async / multi-thread allocation caveat
+
+hotpath's allocation tracking is thread-local and assumes a `current_thread` tokio runtime. ethrex runs a multi-thread `#[tokio::main]` runtime, so allocation counts attributed to `async fn`s are unreliable: allocations performed on a spawned worker thread are not attributed back to the `async fn` that awaited the work. Timing and CPU attribution are not affected by this and work correctly on any runtime.
```
iovoid: timing is affected in the sense that hotpath-alloc adds overhead to all alloc paths

ElFantasma (commented):
One outstanding gap worth closing before merge, from the PR description itself: "the 4 instrumented functions haven't been confirmed to appear in a report during real block execution yet." A `hotpath::measure` that silently doesn't fire (name-resolution bug, cfg mismatch, etc.) wouldn't be caught by a warning-free build — the whole point is timing data, so an end-to-end run that imports a few blocks and grep-confirms `execute_block` / `VM::execute` / `store_block` in the printed report would pin the wire-up. Small effort, closes the biggest "did it actually work?" risk of the PR.

Makefile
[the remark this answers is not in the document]
edg-l: Fixed, added `-p ethrex` in 4240810e1.

cmd/ethrex/ethrex.rs
edg-l: Added a comment on the attr stack noting order matters (tokio::main outermost) in 4240810e1.

docs/developers/l1/profiling.md
edg-l: Good point, clarified in the doc that hotpath-alloc's counting wrapper does inflate timing, so use plain hotpath for pure timing. 4240810e1.

ElFantasma (approved):

iovoid (approved):

PR #6948, opened 2026-07-02
feat(l1): implement debug_getBadBlocks with bad-block tracking

edg-l:
## Description

Implements `debug_getBadBlocks` with real bad-block tracking, superseding #6693 (which returned a hardcoded empty array as a placeholder).

Blocks rejected during `engine_newPayload` execution are now persisted and returned by the RPC method, matching geth's `BadBlockArgs` response shape.

## Changes

**Storage (`ethrex-storage`)**
- New `BAD_BLOCKS` column family storing a single-keyed, RLP-encoded `Vec<Block>` list.
- `Store::add_bad_block` — dedups on `(number, hash)`, sorts by descending block number, and caps the list at 16 entries (oldest evicted).
- `Store::get_bad_blocks` — returns the tracked list (empty when unset).

**Population (`engine/payload.rs`)**
- The block is recorded at the two `newPayload` INVALID decision points (`InvalidBlock`, `EvmError`), alongside the existing `set_latest_valid_ancestor` call. These already exclude `ParentNotFound` / `ParentStateNotFound`, which are SYNCING rather than bad.

**RPC (`debug/bad_blocks.rs`)**
- `GetBadBlocksRequest` returns `[{ hash, block, rlp }]` per entry: the full block JSON (via `RpcBlock`) plus the `0x`-hex block RLP. Routed as `debug_getBadBlocks` under the existing `debug` namespace allowlist.

## Testing

- `test/tests/storage/store_tests.rs::test_bad_blocks` covers the empty default, descending ordering, `(number, hash)` dedup, and the 16-entry cap with oldest eviction (runs under both in-memory and rocksdb engines).

## Scope note

Bad-block recording is wired into the `engine_newPayload` path — the canonical "bad blocks seen on the network" source per the execution-apis spec. The full-sync batch/pipeline import paths also reject blocks but would require per-block clones on hot loops to record; left out of scope for this change.

Closes #6693


crates/storage/store.rs
```diff
@@ -1312,6 +1318,40 @@ impl Store {
             .map_err(StoreError::from)
     }
 
+    /// Records a block that failed validation so it can be served by
+    /// `debug_getBadBlocks`. The list is bounded to [`MAX_BAD_BLOCKS`] entries,
+    /// kept sorted by descending block number, with the oldest dropped once the
+    /// bound is exceeded. Duplicate `(number, hash)` entries are ignored.
+    pub async fn add_bad_block(&self, block: Block) -> Result<(), StoreError> {
```
iovoid, line 1365: Nit: non-atomic so it could lose an entry
In practice it won't happen (bad blocks are rare, simultaneously executing two blocks is also rare even on fast networks like bsc).

iovoid (approved):

crates/storage/store.rs
```diff
@@ -1312,6 +1318,40 @@ impl Store {
...
+        let block_hash = block.hash();
+        if bad_blocks
+            .iter()
+            .any(|b| b.header.number == block_number && b.hash() == block_hash)
+        {
+            return Ok(());
+        }
+        bad_blocks.push(block);
+        bad_blocks.sort_by(|a, b| b.header.number.cmp(&a.header.number));
+        bad_blocks.truncate(MAX_BAD_BLOCKS);
```
ElFantasma, line 1377: Eviction semantic worth double-checking against geth. This truncates by descending block number (keeps the 16 highest-numbered bad blocks), whereas geth's bad-block cache is an LRU keyed by insertion time (returns "most recently seen"). Divergence only shows up when a bad block at a lower number arrives after the 16 slots are full — e.g., a losing-fork bad block reaching us after a deep reorg. Under this code it's silently dropped and never appears in `debug_getBadBlocks`; under geth it enters at the head of the LRU. If tooling that consumes the endpoint assumes "most recent by clock" semantics, the mismatch could confuse it. Might be intentional (highest-number is arguably a cleaner proxy for "newest fork"); worth pinning either the semantic choice in a doc comment or the divergence-vs-geth in the PR description.

ElFantasma (approved):

ilitteri (approved):

crates/storage/store.rs
edg-l: geth isn't an LRU here. `WriteBadBlock` (accessors_chain.go:882-909) sorts descending by block number and truncates, same as this. A late lower-numbered block gets dropped in geth too. Only diff is the cap: geth 10, us 16.

PR #6951, opened 2026-07-02
feat(l1): implement eth_simulateV1 RPC method

### LINKED ISSUE #6212: [BAL] Support eth_simulateV1 RPC method
Implement the `eth_simulateV1` JSON-RPC method for Amsterdam fork readiness.

`eth_simulateV1` simulates multiple transactions across multiple blocks without submitting them, supporting state and block overrides. Unlike `eth_call` (single isolated call), it runs transactions inside full block contexts and returns detailed per-block/per-tx results (gas used, logs, return data, status).

**Current state:** ethrex already has `eth_call`, `eth_estimateGas`, and `eth_createAccessList` in `crates/networking/rpc/eth/transaction.rs`, all built on a shared `simulate_tx()` function. `eth_simulateV1` is not registered in the RPC handler (`crates/networking/rpc/rpc.rs`).

**Implementation path:**
1. Add `SimulateV1Request` type with `blockStateCalls` array, state overrides, and block overrides.
2. Extend or wrap `simulate_tx()` to apply state/block overrides before execution.
3. Loop through block/tx combinations, accumulating state changes across calls.
4. Return detailed results per block (hash, number, gas info) and per call (gas used, logs, return data).

**Spec:** [Ethereum execution-apis eth_simulateV1](https://ethereum.github.io/execution-apis/api/methods/eth_simulateV1)

ilitteri:
**Motivation**

`eth_simulateV1` is the multi-block successor to `eth_call` batching (execution-apis `ethSimulate`); wallets, simulators and MEV tooling increasingly depend on it. Builds on the State/Block Override Sets from #6660, which has since merged; this PR now targets `main` directly.

**Description**

Simulates up to 256 chained blocks on top of a base block without committing anything to the store:

- **Per-block overrides and calls**: each `blockStateCalls` entry carries block overrides (number/time/gasLimit/feeRecipient/prevRandao/baseFeePerGas/blobBaseFee/difficulty/withdrawals/beaconRoot), state overrides (reusing #6660's `StateOverrideSet`), and a list of call objects executed as real transactions (variant chosen by fields, EIP-1559 default; auto-filled nonces from the live simulated state; remaining-block-gas defaults; zeroed signatures with senders carried out-of-band).
- **State chaining**: a cumulative `SimulationOverlay` is the single source of truth for both reads (`SimulationVmDatabase`) and per-block state roots (re-applying the cumulative account updates onto the base block's trie), so block N+1 observes block N's execution results. `BLOCKHASH` resolves real ancestors and previously simulated blocks; EIP-4788/EIP-2935 system calls run per simulated block; EIP-7685 requests are extracted per block.
- **Modes**: `validation: true` enforces nonces and a real EIP-1559 base fee; the default behaves like `eth_call` (zero base fee, relaxed checks). `traceTransfers` emits synthetic ERC-20 `Transfer` logs from `0xeeee…eeee` for every ETH transfer, kept out of receipts and the logs bloom. Like geth's tracer it is fork-independent: on Amsterdam these sit alongside the consensus EIP-7708 logs rather than being replaced by them, so the flag keeps its shape across the fork and self-transfers keep a log even though EIP-7708 excludes them. A value-bearing CALLCODE gets one too, naming the callee it entered, which is geth's rule (its `OnEnter` hook excludes DELEGATECALL and nothing else). `returnFullTransactions` switches between tx hashes and full objects.
- **Responses**: full block objects matching `eth_getBlockByHash` output with real state/transactions/receipts roots, extended with per-call results (`status`, `returnData`, `gasUsed`, `maxUsedGas`, `logs` with per-block `logIndex` and `blockTimestamp`, `error` with revert data). Request-level failures use the spec error-code table (-38010..-38026); tx validation failures abort the request while reverts/halts are per-call results with the tx included in the block, matching geth.
- **Precompile relocation and Amsterdam semantics**: `movePrecompileToAddress` is applied, not just validated, so a moved precompile executes at its destination and the vacated address behaves as a normal account. `beaconRoot` is honored rather than refused, because this engine runs each simulated block's system calls and so writes the value into the EIP-4788 ring buffer -- the `eth_call` family has no block to run them in and still refuses it. On Amsterdam a simulated block models EIP-8037 two-dimensional gas (block `gasUsed` is `max(regular, state)`, and inclusion is decided per dimension) and commits to its EIP-7928 Block Access List, so its hash matches an equivalent real block's.
- Semantics verified against geth master `internal/ethapi/simulate.go` and the execution-apis `eth_simulateV1` test corpus (91 `.io` files); ~11 scenarios are ported as integration tests against the in-memory store. A 50M cumulative gas budget (geth's `RPCGasCap` default) bounds the request.

One pinned divergence, covered by tests rather than carried as prose: `maxUsedGas` omits the EIP-7623 calldata-floor term that geth (`settleGas`'s `peakUsed`) and reth (`total_gas_spent().max(floor_gas())`) both apply — despite the Go name, neither tracks an in-flight peak. ethrex agrees with both except for a calldata-heavy call that also earns a refund, where the refund lands on top of the floor instead of being absorbed by it: 35800 against their 31000, pinned in `max_used_gas_over_reports_when_the_calldata_floor_binds`, with `max_used_gas_is_the_pre_refund_figure` pinning the agreeing case against geth's own `TestRefundCappedAt20Percent`. Closing it needs the pre-refund figure surfaced on `ExecutionReport`, since pre-EIP-7778 `gas_used == gas_spent` and the floor has already replaced it. The hive `rpc-compat` CI pin predates the `eth_simulateV1` corpus (newer fixtures use a pre-merge genesis), so CI does not exercise the method yet — conformance is covered by the ported scenarios.

**How to Test**

```
cargo test -p ethrex-rpc --lib simulate
cargo test -p ethrex-blockchain --lib simulate
```

Or against a live node:

```
ethrex --dev --datadir devdata &
curl -s -X POST http://localhost:8545 -H 'Content-Type: application/json' -d '{
  "jsonrpc":"2.0","id":1,"method":"eth_simulateV1",
  "params":[{
    "traceTransfers": true,
    "blockStateCalls": [
      {"stateOverrides": {"0xc000000000000000000000000000000000000000": {"balance": "0xde0b6b3a7640000"}},
       "calls": [{"from": "0xc000000000000000000000000000000000000000", "to": "0xc100000000000000000000000000000000000000", "value": "0x7b"}]},
      {"blockOverrides": {"feeRecipient": "0x000000000000000000000000000000000000beef"},
       "calls": [{"from": "0xc100000000000000000000000000000000000000", "to": "0xc200000000000000000000000000000000000000", "value": "0x1"}]}
    ]
  }, "latest"]
}'
```

Expected: two simulated blocks chained by `parentHash`, both calls `"status": "0x1"`, a `Transfer` log from `0xeeee…eeee` per call, and the second block's `miner` set to `0x…beef`.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A, no Store schema changes.

Closes #6212


crates/networking/rpc/eth/transaction.rs
```diff
@@ -812,15 +812,6 @@ fn convert_state_overrides(
...
 
-/// Caps the estimation ceiling at the gas the sender can actually pay for.
-///
-/// `fee_cap` is the per-gas price the transaction's balance check uses: `max_fee_per_gas`
-/// when the call object carries one, otherwise the legacy `gas_price`.
-///
-/// A zero `fee_cap` is a request that names no fee at all, or names a fee of zero, and in
-/// both cases no balance bounds the gas: the ceiling is returned unchanged. Checked here
-/// rather than at the call site so no caller can reach the division with a zero divisor,
-/// which is the same split this function's `fee_cap` argument exists to close.
```
pablodeymo, line 823: why did you remove this comment?

pablodeymo (changes requested):
please preserve the previous comments

crates/networking/rpc/eth/transaction.rs
juanbono: restored

pablodeymo:
Went through this in depth. The engine is solid: I ran
`cargo test -p ethrex-rpc --lib simulate` (43 pass) and
`cargo test -p ethrex-blockchain --lib simulate` (9 pass) on 61cb43a80d, and
cross-checked the error-code table, gas defaulting/budget capping and the
traceTransfers rules against geth master (`internal/ethapi/simulate.go`,
`logtracer.go`, `errors.go`). Those match. Below is what did not, plus nits.
Line numbers are on 61cb43a80d.

### 1. `traceTransfers` on Amsterdam drops the `0xeee…eeee` logs entirely

`VM::eth_transfer_log_address` (`crates/vm/levm/src/vm.rs:3272`) returns the
SYSTEM_ADDRESS consensus log on Amsterdam+ and only reaches the
`TRACE_TRANSFER_ADDRESS` branch pre-Amsterdam. So with `traceTransfers: true` on
an Amsterdam block the caller gets no `0xeeee…eeee` logs at all, and no log for
a self-transfer, since the EIP-7708 branch excludes `from == to`.

geth's tracer is fork-independent: `OnEnter` calls `captureTransfer` whenever
the opcode is not DELEGATECALL/CALLCODE and `value > 0`, and `captureTransfer`
is gated only on `t.traceTransfers`. On Amsterdam geth emits both the consensus
log and the trace log, including for self-transfers. The `Environment` doc
comment says the consensus logs "take precedence", so I take this as
deliberate, but it changes the response shape a client sees the moment the fork
activates. Is diverging here the intent? If so it belongs in the description
next to the `maxUsedGas` divergence, pinned by a test.

### 2. CALLCODE detection is inferred and misfires under EIP-7702

`crates/vm/levm/src/opcode_handlers/system.rs:1272` and `:1327`:

```rust
let is_callcode = to == msg_sender && code_address != to && !is_delegation_7702;
```

`OpCallCodeHandler` resolves delegation on its target
(`eip7702_peek_delegation` at `system.rs:258`, inside the handler starting at
`:223`) and passes `is_delegation_7702 = true` into `generic_call`. For a
CALLCODE to a delegated address the heuristic evaluates to `false` and a trace
log is emitted for a CALLCODE, which geth never does. Suggest an explicit
`is_callcode: bool` (or a call-kind enum) parameter on `generic_call`, set by
each handler. That also removes the duplicated condition.

### 3. `execute_tx_simulate` builds a receipt the engine discards

`crates/vm/backends/mod.rs:238` clones `execution_report.logs` into a
`Receipt`; `crates/blockchain/simulate.rs:619` destructures it as
`let (_, report)` and rebuilds the receipt with trace logs filtered. Returning
only the `ExecutionReport` removes a clone per call and one of the two receipt
constructions.

### 4. The 5s timeout does not stop the work

`crates/networking/rpc/eth/simulate.rs:188` wraps `spawn_blocking` in
`tokio::time::timeout`. On expiry the caller gets `-32603 "eth_simulateV1
timeout"` but the blocking task keeps running up to the 50M gas cap, while the
comment calls this "a wall-clock cap". Either reword the comment to say it
bounds the response, not the CPU, or thread an `AtomicBool` the engine checks
between blocks.

### 5. Per-block state root is quadratic

`crates/blockchain/simulate.rs:745` re-applies the whole cumulative update set
onto the base trie for every simulated block, and `:557`/`:566` clone the full
overlay into a fresh `Arc` per block. O(blocks × cumulative updates) for up to
256 blocks. The comment explains why incremental roots are not an option, so
fine as is, but together with (4) a 256-block request on a node with deep state
may hit the 5s limit. `Arc<SimulationOverlay>` + `Arc::make_mut` would at least
drop the clones.

### 6. Unlisted validation errors map to `-32603`

`crates/networking/rpc/eth/simulate.rs:421`. A call with explicit `gas` above
the EIP-7825 cap gets an "internal error" code for a user mistake. geth's
`txValidationError` default does the same, so no change needed; a one-line
comment saying so would save the next reader the lookup.

### 7. History

~60 commits including the `feat/state-override-set` merges. Squash on merge.

Only (2) blocks for me; (1) needs a yes/no on intent. The rest are nits.

PR #6952, opened 2026-07-02
feat(l1): enforce movePrecompileToAddress in eth_simulateV1

ilitteri:
**Motivation**

The State Override Set (#6660) parses `movePrecompileToAddress` but nothing consumes it, and `eth_simulateV1` (#6951) documented it as a known gap. This closes the gap; stacked on #6951.

**Description**

- LEVM gains `PrecompileOverrides` (per-block removed set + moved destination→source map) on the `Environment`, resolved through `VM::active_precompile_at` at the three precompile decision points (simple-transfer fast path, top-level dispatch, `generic_call`): a moved precompile executes at its destination address while its source — like any precompile whose account is touched by state overrides — stops behaving as one. Consensus execution is unchanged: without overrides the helper falls through to the static per-fork check.
- The simulate engine validates each block's overrides exactly like geth's `StateOverride.Apply` (address order): moving a non-precompile, moving to an overridden address, or overriding a move destination abort the request with code `-32000` and geth's message shapes; a second move to the same destination silently wins, matching current geth (the historical -38022/-38023 codes are gone from the reference corpus). Moves are scoped to their block.
- Move-only overrides no longer materialize account state (that would have created empty accounts in the simulated state roots).

Known divergence (accepted): the pre-warmed address set stays the canonical per-fork precompile list, so access-warmth of moved precompiles can differ from geth in gas-exact scenarios; revisit with the hive pin bump.

**How to Test**

```
cargo test -p ethrex-rpc --lib simulate
```

New integration tests move/override the identity precompile: the destination echoes input while the source becomes a plain account, per-block scoping restores the canonical layout in the next block, code overrides at a precompile address execute the bytecode, and both error paths return -32000 with the expected messages.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A, no Store schema changes.

### DIFF
--- a/crates/blockchain/simulate.rs
+++ b/crates/blockchain/simulate.rs
@@ -15,18 +15,18 @@ use ethrex_common::{
     Address, U256,
     constants::{DEFAULT_OMMERS_HASH, DEFAULT_REQUESTS_HASH, GAS_PER_BLOB},
     types::{
-        AccountInfo, AccountUpdate, Block, BlockBody, BlockHeader, ChainConfig,
-        EIP1559Transaction, EIP4844Transaction, EIP7702Transaction, ELASTICITY_MULTIPLIER, Fork,
-        GenericTransaction, LegacyTransaction, Log, Receipt, Transaction, TxKind, Withdrawal,
-        bloom_from_logs, calc_excess_blob_gas, calculate_base_fee_per_gas, compute_receipts_root,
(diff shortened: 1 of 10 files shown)


juanbono:
Closing this one since it was already implemented in #6660

PR #6956, opened 2026-07-02
fix(l1): cap snappy-decompressed length before allocating

ElFantasma:
**Motivation**

The RLPx codec caps the *compressed* frame size (`MAX_MESSAGE_SIZE` = `0xFFFFFF`), but `snappy_decompress` places no bound on the *decompressed* length. A snappy stream begins with a peer-controlled varint of the decompressed length, and `snap`'s `decompress_vec` allocates a buffer of that size **before** validating the body — so a tiny frame can declare a very large length (up to ~4 GiB) and force a correspondingly large allocation. go-ethereum bounds `snappy.DecodedLen` against the message-size cap before allocating; ethrex did not.

**Description**

`snappy_decompress` now reads the declared decompressed length via `snap::raw::decompress_len` and rejects it if it exceeds `MAX_SNAPPY_DECOMPRESSED_LEN` (`0xFFFFFF`, matching the compressed-frame cap and geth's maxUint24 bound) **before** allocating. Normal messages are unaffected.

Adds unit tests: a crafted oversized declared-length header is rejected up front, and a normal round-trip still succeeds.

**Checklist**

- [x] `cargo test -p ethrex-p2p` (snappy tests) and `cargo clippy -p ethrex-p2p` pass.


ElFantasma:
Thanks — all non-blocking; notes on each:

- **Constant duplication (`MAX_SNAPPY_DECOMPRESSED_LEN` vs `MAX_MESSAGE_SIZE`)**: kept as a separate named constant deliberately — it's a conceptually distinct limit (decompressed size vs the compressed-frame cap in `codec.rs`) that happens to share geth's `maxUint24` value; the doc comment records that linkage. Reusing the private codec const would couple two independent limits across modules. Can hoist both into a shared constants module if the team prefers a single source of truth.
- **16 MiB doesn't reject legit messages**: confirmed — geth likewise bounds `snappy.DecodedLen` at `maxUint24`, so a compliant peer never sends a >16 MiB decompressed RLPx message.
- **Test assertion / LEB128**: the substring match is intentional (the `snap` crate's error text isn't a stable API and there's no dedicated variant to match); the round-trip test is the positive control.
- **Error variant**: reused `InvalidCompression` as the closest existing variant; if peer-scoring ever needs to distinguish "oversized declared length" from "corrupt stream," a dedicated variant can be split out then.

edg-l (approved):

crates/networking/p2p/rlpx/utils.rs
```diff
@@ -71,9 +73,65 @@ pub fn snappy_compress(encoded_data: Vec<u8>) -> Result<Vec<u8>, RLPEncodeError>
...
+            declared >>= 7;
+        }
+        frame.push(declared as u8);
+        frame.push(0x00); // minimal (invalid) body
+
+        let err =
+            snappy_decompress(&frame).expect_err("oversized declared length must be rejected");
+        let msg = format!("{err}").to_lowercase();
+        assert!(
+            msg.contains("exceed") || msg.contains("too large"),
```
iovoid: why "too large" if that is (seemingly) never used?

iovoid (approved):

crates/networking/p2p/rlpx/utils.rs
ElFantasma: Good point — the error only ever says "exceeds maximum", so the `|| "too large"` branch was dead. Dropped it; the assertion now just checks for "exceed".

ilitteri (approved):

PR #6957, opened 2026-07-02
fix(l1): charge snap serving budget on every lookup attempt

ElFantasma:
**Motivation**

The snap serving handlers (`GetByteCodes`, `GetStorageRanges`, `GetTrieNodes`) clamp the response to a byte budget, but that budget only advances when a lookup **hits**. A missed lookup, a duplicate, or an overlong path (which resolves to an empty node) costs nothing, so a request full of misses/empties never trips the budget and the handler walks the *entire* peer-supplied list (`hashes` / `account_hashes` / `paths`) — O(N) DB/trie probes returning a near-empty response. The per-peer rate limiter counts messages, not items, so it doesn't catch this. This is the request/miss-side complement to the response-side hardening in #6833.

**Description**

Bounds each list-walking serving handler by a hard **per-request lookup count**, `MAX_SERVE_LOOKUPS` (1024), independent of the byte budget. Each item is a DB/trie probe (a trie-node lookup traverses several nodes), so capping the *count* directly bounds disk/CPU work, where the byte budget only bounds response size:

- `process_byte_codes_request` — serve at most `MAX_SERVE_LOOKUPS` hashes.
- `process_storage_ranges_request` — probe at most `MAX_SERVE_LOOKUPS` accounts.
- `process_trie_nodes_request` — truncate the paths to the remaining lookup budget before the trie descent (so an oversized pathset can't blow past the cap in one call).

Partial responses are protocol-legal, so an honest peer whose request exceeds the cap simply re-requests the remainder; `MAX_SERVE_LOOKUPS` sits comfortably above our own client's request batches (`NODE_BATCH_SIZE` = 500, `STORAGE_BATCH_SIZE` = 300). The `MAX_RESPONSE_BYTES` (512 KiB) clamp is retained for the response-*size* axis. `account_range` is a single range query with no per-item walk and is unchanged.

The value **1024** takes geth as a reference point (`maxTrieNodeLookups` / `maxCodeLookups` = 1024, `softResponseLimit` = 2 MB, `maxTrieNodeTimeSpent` = 5 s); geth splits the cap per message type, but a single shared cap is simpler here and serves the same purpose. This supersedes the earlier byte-cost-per-probe approach (`MIN_LOOKUP_COST`) — a count cap is a tighter, more direct bound on the same lookup-walk vector — so `MIN_LOOKUP_COST` and its store-level twin `MIN_TRIE_NODE_LOOKUP_COST` are removed.

Regression tests place a real hit *past* the cap and confirm it is never reached (`byte_codes_bounds_lookup_count`, `storage_ranges_bounds_lookup_count`), plus `byte_codes_bounds_duplicate_hits_by_count`. The existing `*_clamps_response_bytes` tests still pass.

**Checklist**

- [x] `cargo test -p ethrex-test --test ethrex_tests snap_server` (18 tests) and `cargo clippy -p ethrex-p2p -p ethrex-storage` pass.


ElFantasma:
Thanks — the reviews caught a real gap I'd missed. Addressed in 368182a09:

**`GetTrieNodes` nested sub-path amplification (Codex High / Claude Finding 1) — fixed.** Correct: charging per outer pathset was one level too shallow. The per-sub-path work is inside `Store::get_trie_nodes`, whose inner loop advanced `bytes_used` only on a hit (`+= node.len()`), so a single pathset packed with missing/overlong storage sub-paths (each → `Ok(vec![])`, 0 B) walked to completion for 32 B of outer budget. Fixed at the source: that loop now charges `max(node.len(), MIN_TRIE_NODE_LOOKUP_COST)` per sub-path, so it breaks once the per-call `byte_limit` is exhausted regardless of hits.

**Storage-ranges over-charge (greptile P2 #1) — fixed.** Now charges the per-probe minimum only when an account produced no slots (absent account / empty range); hit accounts keep paying 64 B/slot and aren't charged extra. Note the literal `bytes_used += (n_slots*64).max(MIN)` suggestion would double-count — the 64 B/slot is already added in the inner loop — so charging-on-empty is the equivalent that avoids that.

**Trie-nodes miss-path test (greptile P2 #2 / Claude) — acknowledged, not added.** Exercising `get_trie_nodes`' storage loop needs a state trie readable via its snapshot-gated `open_state_trie`, which the direct-write test fixtures don't populate (it goes through the block/layer commit machinery — the same limitation already documented for this handler). The fix mirrors the byte-codes/storage mechanism that *is* unit-tested; I updated the code comment to reflect the new per-sub-path charge and the standing fixture gap. Happy to add a committed-trie fixture as a fast-follow if preferred.

**Minor (greptile):** the storage over-charge above was the substantive half of that note; the accounting is now consistent with the other two handlers.

edg-l (approved):

crates/networking/p2p/snap/server.rs
```diff
@@ -162,9 +187,14 @@ pub async fn process_trie_nodes_request(
...
+            let returned_bytes = trie_nodes
+                .iter()
+                .fold(0u64, |acc, nodes| acc + nodes.len() as u64);
             nodes.extend(trie_nodes.iter().map(|nodes| Bytes::copy_from_slice(nodes)));
-            byte_budget = byte_budget
-                .saturating_sub(trie_nodes.iter().fold(0, |acc, nodes| acc + nodes.len()) as u64);
+            // Charge at least a per-pathset probe cost so overlong/empty-result pathsets
+            // (e.g. a path > 32 bytes, which resolves to an empty node) still consume the
+            // budget instead of letting the loop run over every decoded pathset.
+            byte_budget = byte_budget.saturating_sub(returned_bytes.max(MIN_LOOKUP_COST));
```
iovoid: MAX_RESPONSE_BYTES/MIN_LOOKUP_COST is currently 16k, plus each trie lookup requiring several lookups (since it has to traverse the trie), makes the worst case very resource intensive

iovoid: The cost should be 32 _per path_, not total. Otherwise the user can request as many paths as the protocol supports and if they all miss, only pay once.

iovoid (changes requested):

crates/networking/p2p/snap/server.rs
ElFantasma: Fixed in 21b89eec2: it now charges `MIN_LOOKUP_COST` (32) *per path* — `returned_bytes.max(MIN_LOOKUP_COST * n_paths)` where `n_paths = paths.len()` — so an all-miss pathset with N paths costs 32·N instead of a single 32. This matches how the bytecodes handler already charges per hash.

ElFantasma: With per-path charging (above), total probed paths per request is now bounded by budget/32 — ~16k for the 512 KiB cap. If you want that tighter given each trie lookup traverses several nodes, I can raise the per-lookup cost for trie nodes above the flat 32 (they're pricier than a single bytecode/DB get), e.g. down to roughly geth's `maxTrieNodeLookups` (~1024). Happy to — just say what bound you'd prefer.

ElFantasma: Reworked this to address the concern directly (24ccc1410): instead of the byte-cost proxy, the handlers now enforce a **hard per-request lookup-count cap**, `MAX_SERVE_LOOKUPS` = 1024, so the worst case is 1024 trie descents per request, not ~16k.

I took geth as a reference point here: it caps `GetTrieNodes`/`GetByteCodes` at `maxTrieNodeLookups`/`maxCodeLookups` = **1024** each (plus `softResponseLimit` 2 MB and a `maxTrieNodeTimeSpent` 5 s budget). I used a single shared 1024 cap rather than geth's per-message split since it serves the same purpose and is simpler. 1024 sits above our own client's request batches (`NODE_BATCH_SIZE` 500, `STORAGE_BATCH_SIZE` 300), and snap responses may be partial, so honest peers just re-request the remainder.

Since the count cap is a tighter, more direct bound on the same walk, I dropped `MIN_LOOKUP_COST` (and its store-level twin `MIN_TRIE_NODE_LOOKUP_COST`) entirely.

ElFantasma: Superseded by the count-cap rework (24ccc1410): `process_trie_nodes_request` now truncates each pathset to the remaining `MAX_SERVE_LOOKUPS` budget *before* the trie descent, so a pathset packed with many paths can't over-probe — the per-path byte charge is no longer needed and `MIN_LOOKUP_COST` is gone. See the updated PR description for the full approach.

iovoid (approved):

ilitteri (approved):

PR #6961, opened 2026-07-03
fix(l1): eth_call/eth_estimateGas tolerate a missing header slot_number

ilitteri:
**Motivation**

`env_from_generic` — the environment builder behind `eth_call`,
`eth_estimateGas` and `eth_createAccessList` — was the only environment
constructor that treated a missing `slot_number` as an error under Amsterdam+
rules. `setup_env_with_config`, the block-execution builder, defaults it to `0`
with no fork check at all.

That asymmetry is in the wrong direction. A missing slot on an Amsterdam block
would be a consensus fault, and it is already rejected where it matters:

- `validate_prague_header_fields` rejects an Amsterdam header whose
  `slot_number` is `None` (`SlotNumberNotPresent`), on the import path via
  `validate_block_pre_execution`.
- `engine_newPayloadV5` rejects a payload without the field.
- `validate_attributes_v3` rejects Amsterdam-timestamped payload attributes, so
  a `forkchoiceUpdatedV3`-driven build cannot produce a slot-less Amsterdam
  block — it produces no block at all.
- `Genesis::get_block_header` fills in `Some(0)` when Amsterdam is active, so
  even an Amsterdam genesis header carries a slot.

So the removed guard was unreachable through normal block import: this is not a
fix for an observed failure, and there is no user-visible change on a
well-formed chain. It is a consistency fix. Simulation's job is to predict what
execution does, so a header the executor would happily run with `SLOTNUM`
reading `0` must not make `eth_call` fail. The one way to reach the old
divergence is a devnet whose fork timestamps are moved so that Amsterdam
retroactively covers headers already stored under earlier rules — the store
keeps the old slot-less headers, and every simulation against them fails while
execution does not.

**Description**

`env_from_generic` now defaults a missing `slot_number` to `0`, matching
`setup_env_with_config`. L2 keeps its explicit `0`. Behavior below Amsterdam is
unchanged — that path already defaulted to `0`.

Adds two regression tests on an Amsterdam-activated config. Both call a
contract whose bytecode is `SLOTNUM; PUSH0; MSTORE; PUSH1 0x20; PUSH0; RETURN`
and assert the returned word, rather than only that simulation returns `Ok`:
`0` for a header without a slot, `1234` for a header carrying `1234`.
Mutation-verified in both directions — hardcoding the environment's slot to
zero fails the second test, and restoring the old guard fails the first.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A, no store changes.


test/tests/levm/simulation_env_tests.rs
```diff
@@ -0,0 +1,100 @@
...
+        "simulation must tolerate a missing slot_number like execution does; got {result:?}"
+    );
+}
+
+#[test]
+fn simulation_uses_the_slot_number_when_present() {
+    let (mut db, header) = amsterdam_db_and_header(Some(1234));
+    let result =
+        LEVM::simulate_tx_from_generic(&plain_transfer(), &header, &mut db, VMType::L1, &NativeCrypto);
+    assert!(result.is_ok(), "got {result:?}");
```
ElFantasma: Building on Copilot's callout above: this test's assertion (`result.is_ok()`) is identical to the first test's, so it would pass regardless of whether the slot value flows through at all — nothing here actually distinguishes `Some(1234)` from `None`. Two shapes worth considering: (a) execute a small SLOTNUM-reading contract and assert the returned value equals 1234, which makes the test name honest; or (b) rename to something like `simulation_accepts_a_present_slot_number` so the name matches what's asserted. Non-blocking.

ElFantasma (approved):

test/tests/levm/simulation_env_tests.rs
```diff
@@ -0,0 +1,100 @@
+//! eth_call / eth_estimateGas environment construction.
+//!
+//! Simulation must never be stricter than execution: a header without a
+//! `slot_number` imports and executes fine (the execution env defaults the
+//! slot to zero), so simulating on top of it must work the same way. Headers
+//! legitimately lack the slot on Amsterdam+ chains whose consensus client
+//! drives a pre-V4 engine API (the slot only arrives in
+//! `PayloadAttributesV4`) — including ethrex's own `--dev` mode, which
+//! produces blocks via `engine_forkchoiceUpdatedV3`.
```
iovoid: pre-V4 APIs explicitly reject Amsterdam blocks (see validate_attributes_v3)

edg-l (approved):

crates/vm/backends/levm/mod.rs
```diff
@@ -3761,20 +3761,17 @@ fn env_from_generic(
...
-    // Validate slot_number for Amsterdam+ blocks
-    // For L2 chains, slot_number is always 0
+    // slot_number: tolerate a missing value exactly like the block-execution
+    // path does (see the Environment construction in `run_execution`'s caller).
+    // Headers legitimately carry None even on Amsterdam+ chains when the CL
+    // drives a pre-V4 engine API (the slot only arrives in
+    // PayloadAttributesV4), and blocks with None import and execute fine with
+    // SLOTNUM reading 0. Simulation (eth_call / eth_estimateGas) must never be
+    // stricter than execution — erroring here made every simulation fail on
+    // such chains. For L2 chains, slot_number is always 0.
```
MegaRedHand: > Headers legitimately carry None even on Amsterdam+ chains when the CL drives a pre-V4 engine API

This is wrong. The V4+ engine API rejects headers with missing fields.

```suggestion
    // slot_number: tolerate a missing value exactly like the block-execution
    // path does (see the Environment construction in `run_execution`'s caller).
    // For L2 chains, slot_number is always 0.
```

MegaRedHand (approved):
LGTM

test/tests/levm/simulation_env_tests.rs
ilitteri: a129674c7

test/tests/levm/simulation_env_tests.rs
```diff
@@ -0,0 +1,100 @@
...
+    let result =
+        LEVM::simulate_tx_from_generic(&plain_transfer(), &header, &mut db, VMType::L1, &NativeCrypto);
+    assert!(
+        result.is_ok(),
+        "simulation must tolerate a missing slot_number like execution does; got {result:?}"
+    );
+}
+
+#[test]
+fn simulation_uses_the_slot_number_when_present() {
```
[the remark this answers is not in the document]
ilitteri: 7ff4c7f74

test/tests/levm/simulation_env_tests.rs
ilitteri: 7ff4c7f74

ilitteri:
Applied review feedback:

- module declaration moved into alphabetical order — 1b3e933fa
- `cargo fmt` (this was the actual failing Lint step) — e35217b3a
- test assertions now check the value SLOTNUM reads, not just `is_ok()` — 7ff4c7f74
- corrected rationale in comments and docs — a129674c7

PR #6964, opened 2026-07-06
fix(l1): reject short discv5 packet authdata instead of panicking

ElFantasma:
**Motivation**

The discv5 packet-decode path read fixed-size fields out of the peer-controlled `authdata` before validating its length, so a malformed packet could panic. Because the discv5 server runs as an actor whose runtime catches a handler panic and stops the loop (no restart), a single malformed packet could take the discv5 discovery actor down. The masking key is the destination node id (public), so an ordinary packet can be forged without a session — making the ordinary path reachable unauthenticated.

Two sites:
- `discv5_handle_ordinary` extracted the 32-byte src-id via `H256::from_slice(&authdata)` before the session lookup — panics on any `authdata.len() != 32`. Reachable unauthenticated (primary).
- `WhoAreYou::decode` sliced `authdata[..16]` before checking length — panics on `authdata.len() < 16`. Reachable when a matching pending nonce exists.

**Description**

- Added `Ordinary::src_id(&Packet) -> Result<H256, PacketCodecError>` that returns `InvalidSize` unless `authdata.len() == 32`, and routed `discv5_handle_ordinary` through it (`?`). `Ordinary::decode` now uses the same helper, so its length check is shared.
- `WhoAreYou::decode` now rejects `authdata.len() != 24` (16-byte id-nonce + 8-byte enr-seq) up front, before slicing.

Both return an error instead of panicking; the packet dispatch already logs-and-drops handler errors (`inspect_err`), so a malformed packet is now discarded and the actor keeps running. Handshake decode was already length-guarded.

Adds unit tests: short WHOAREYOU authdata and short ordinary authdata are rejected (previously panicked), and a well-formed 32-byte ordinary authdata still yields the src-id.

**Checklist**

- [x] `cargo test -p ethrex-p2p discv5` (incl. new authdata tests) and `cargo clippy -p ethrex-p2p` pass.


MegaRedHand (approved):

ElFantasma:
Thanks for the reviews — addressed the optional items (`8fc87de8d`):

- **@Kimi** (over-length coverage): added `who_are_you_decode_rejects_over_length_authdata` — a 32-byte authdata must also be rejected by the exact `!= 24` check, not just short ones.
- **@Claude** (confirm `HandshakeAuthdata` is guarded): verified — `HandshakeAuthdata::decode` checks `authdata.len() < HANDSHAKE_AUTHDATA_HEAD` (= 34) *before* the `authdata[..32]`/`[32]`/`[33]` reads, so that path can't panic on short authdata. No change needed.
- **@greptile** (`Ordinary::src_id` computed twice per decrypted packet): keeping it. The handler needs the src-id *before* decryption (to look up the session key), and `Ordinary::decode` recomputes it to stay self-contained as the single source of the length guard. The redundant call is a length check + a 32-byte copy; threading the id through `decode`'s signature to save that would make the guard's ownership fuzzier for no meaningful gain.
- **@Claude** (vestigial `try_into` after the `!= 24` check; comment verbosity): the `?` on the now-infallible `try_into` is harmless and defensive, and the per-site comments document a security invariant (why the length guard exists) so a future edit doesn't drop it — leaving both as-is.

For context: this PR closes the EF finding `discv5-short-authdata-panic` (unauthenticated single-packet DoS that permanently stops the discv5 actor).

ilitteri (approved):

pablodeymo (approved):

PR #6965, opened 2026-07-06
perf(l1): decode engine_newPayload transactions in parallel

ilitteri:
## Motivation

`engine_newPayload` runs on the block-import critical path: the consensus client
waits for the response, and the handler executes before returning `VALID`. Profiling
that path showed a small but real amount of work sits *before* execution and is not
counted by the block-processing pipeline's timing (`gigagas` / `total_ms`): turning the
payload into a `Block`. Part of that is decoding the transactions.

`ExecutionPayload::into_block` decoded the payload's transactions **sequentially**, even
though each transaction is an independent RLP blob.

## Description

Decode the payload's transactions in parallel with `rayon`'s `par_iter`. The result is
identical to the sequential decode:

- `par_iter().map().collect::<Result<Vec<_>, _>>()` is **order-preserving** and
  short-circuits on the first decode error, so the decoded `Vec<Transaction>` and any
  error are exactly what the sequential path produced.
- The `eip-8025` (stateless / zkVM guest) build keeps the deterministic sequential path,
  mirroring the existing `get_transactions_with_sender`.

Scope is intentionally tiny: only the decode loop in `into_block` changes.

## How to test

`cargo test -p ethrex-rpc --lib types::payload` — the `deserialize_payload_into_block`
test decodes a mixed legacy + EIP-1559 payload and now also asserts that the parallel
decode matches a sequential decode (same order, same content).

## Notes

This is a pure, deterministic parallelization (no consensus behavior change). A follow-up
change computes the transactions root directly from the payload's raw canonical bytes
(skipping a decode→re-encode round-trip); it is kept separate as it warrants its own review.


crates/networking/rpc/types/payload.rs
```diff
@@ -379,7 +396,28 @@ mod test {
...
         let payload: ExecutionPayload = serde_json::from_str(json).unwrap();
-        assert!(payload.into_block(Some(H256::zero()), None, None).is_ok());
+        // Reference: decode the transactions sequentially, in order.
+        let sequential: Vec<Vec<u8>> = payload
+            .transactions
+            .iter()
+            .map(|e| e.decode().unwrap().encode_canonical_to_vec())
+            .collect();
+        let block = payload
+            .clone()
```
iovoid: This .clone is redundant

ElFantasma (approved):
LGTM. 

Endorse iovoid's/Copilot's `payload.clone()` nit — the sequential decode above only borrows `payload.transactions` and that borrow is released before `into_block`, so `payload.into_block(...)` works without the clone. Also: `changelog` check is failing — an entry under `## Perf` is missing.

PR #6967, opened 2026-07-06
perf(l1): prewarming using mempool contents

iovoid:
**Motivation**

Some of the transactions that will be included in the next block are already known beforehand. We can use this fact to speculatively load state to save costs during actual block execution.

**Description**

To ensure that malicious mempool transactions can't have an outsized negative impact we:
- use the "block building" logic in intervals, rather than executing transactions as they arrive
- halt the process before the block is expected to arrive

There is a known downside of making measurements more noisy (as they now depend on mempool contents), but the ~15% improvement brought by this PR is worth it.


ElFantasma (approved):

ilitteri (approved):

ilitteri:
PR is good to merge as is — let's get this optimization on main asap. Nice work; the cross-block handoff is sound (the cache holds only parent-state-pure reads, guarded by `parent_hash` + `fork`, discarded on mismatch/witness-collection), and I didn't find any correctness or data-race issues.

A few low-hanging-fruit leftovers (all can be follow-ups, none block merge):

1. **Warm the per-block system contracts** (EIP-4788 beacon roots, EIP-2935 history, request predeploys). `run_pass` only warms user txs via `warm_txs`, so it never runs `prepare_block` — yet those fixed addresses are touched on every block and their slots are fully deterministic from the synthetic child header we already build (EIP-2935 slot `parent.number % 8191` with calldata `parent.hash()`; EIP-4788 slots `timestamp % 8191` and `+8191`). Accounts/code are usually already hot, but the rotating ring slots were last touched ~8191 blocks ago (~27h), so they're plausibly cold — and this also covers the pre-Amsterdam `warm_block` path. Replay the deterministic system ops against `cache_dyn` (discarded writes) and feed the keys through `warm_merkle_paths`.
2. **`select_warm_set` deep-clones every selected `Transaction`** (`mtx.transaction().clone()`) — with `GAS_BUDGET_MULTIPLIER = 6` that's hundreds of full clones per delta pass on a busy mempool. The loop already owns each `mtx`, so return/move `MempoolTransaction` (a cheap `Arc` bump) and borrow at the `warm_txs` call site.
3. **`warm_merkle_paths` re-opens the state trie every delta pass** — `parent.state_root` is fixed for the slot, so open the handle once in `run_pass` and reuse it.
4. **Build hygiene:** `PrewarmedEntry` is constructed only in the rayon-gated `run_pass` but read unconditionally in `execute_block_pipeline` — a possible `dead_code` warning in `--no-default-features` / `eip-8025` builds. Worth a quick check + a matching `cfg_attr` if it fires.

There's also some potential follow-up work to try out:

- **Cheap front-loaded account/code pre-pass** before the heavy speculative exec (à la Nethermind's address warmer / geth's `statePrefetcher`), using the new batched `multi_get_cf` account prefetch — since the last delta is usually cut short by the next block, this guarantees the warm set's top-level sender/`to` accounts are resident even when deep execution doesn't finish.
- **Write-set-only merkle warming** (geth's `noreads`): only written keys' paths feed the state-root recompute, so walking read-only key paths spends idle-window I/O on nodes the merkleizer never reads. `warm_txs` already produces per-tx `AccountUpdates` (currently discarded) — surface the written-key set and pass a write-only predicate into `touched_keys_where`.
- **Adaptive `GAS_BUDGET_MULTIPLIER`** driven by measured prewarm hit-rate (6× warms ~6 blocks of txs while ~1 lands; instrument, then likely lower to 2–3×).
- **Bound the per-seq `filter_transactions` snapshot** (top-of-mempool head per sender) so delta passes stop cloning + sorting the whole pool every ~100ms; also enables tightening the poll cadence near the slot deadline to catch the final arrival burst.
- **Proposer-side reuse**: on the local-build path the fee recipient and withdrawal set *are* known ahead, so `run_pass` could set the real coinbase and warm withdrawal recipients — closing the two gaps that are inherent only to followers.
- **(As EIP-7928 matures)** idle-window BAL-driven storage/trie prefetch, and execution-overlapped proof generation (reth drives both from the block access list / per-tx multiproof targets).

edg-l (approved):
**Delta-pass warming is largely wasted on busy mempools.** `drop_already_warmed` strips a sender's warmed prefix, then `warm_txs` runs the suffix against a fresh parent-state overlay; pass 1's nonce/balance changes never persist (only reads are cached, mutations die with the discarded overlay). So nonce k+1 hits a nonce-mismatch, short-circuits, warms nothing, and still burns the per-sender cap.

**Fix path:** carry a per-sender speculative overlay across delta passes (persist each sender's `GeneralizedDatabase` between passes and replay only the new suffix on top), or simpler: don't drop the prefix, keep each sender's full 0..N sequence in every delta's replay set so nonces resolve. The first is more efficient; the second is a one-line change to skip `drop_already_warmed`'s intra-sender pruning.

No correctness, consensus, or data-race issue; worst case is wasted idle-window I/O, and execution falls back to the cold path. The delta-pass fix can land in this PR or a follow-up; either is fine.

PR #6968, opened 2026-07-07
perf(levm): utilize native u256 zisk arithmetic

benbencik:
**Motivation**
Currently, all zkVM u256 arithmetic in levm runs as a software implementation. ZisK has dedicated 256-bit arithmetic: `syscall_arith256`, `syscall_add256`, `syscall_arith256_mod`, and `fcall_bigint256_div`. Operations using these calls consume fewer rows in the Main SM.  This PR routes the EVM opcodes through the ZisK paths when compiled with the `zisk` feature.

**Description**
Adding a new (feature-gated) module `zisk_u256.rs` that utilizes the u256 arithmetic from ZisK. Operations `checked_div` and `checked_rem` use hints from the host that are verified by asserts. This change assumes that the host is untrusted, a wrong hint panics the computation.

Opcode handlers in `arithmic.rs` are in conditional compile such that non-zisk computation remains unchanged. 

The ziskos dependency is moved to the root so `levm` and the guest program share a single pinned version.

**Results**
The current version of ziskos is pinned to [v0.16.1](https://github.com/lambdaclass/ethrex/blob/567ea6025ed14f36178e000d6975e7374e8dfe3a/crates/guest-program/bin/zisk/Cargo.toml#L17), and I had trouble integrating this with the rest of the stack, which updated to a newer version. For this reason, I was not able to run the zisk profiler but measured just the runtime of the emulator (not AIR-costs). Emulator execution times do not perfectly map to actual proving time costs, but performance improvements should translate.

I measured the average emulator execution time on 100 mainnet (24949787 - 24949886) blocks with ~5,5% improvement: 
- 0ab4d4d2e: 11.331 s
- this change: 10.692 s

Additionally, I also check selected 100M EEST arithmetic U256 benchmarks:
eest-100m-comparission

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1445,6 +1445,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "aurora-engine-modexp"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "518bc5745a6264b5fd7b09dffb9667e400ee9e2bbe18555fac75e1fe9afa0df9"
+dependencies = [
+ "hex",
+ "num",
+]
+
+[[package]]
 name = "auto_impl"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


ilitteri:
This could be useful once we resume https://github.com/lambdaclass/ethrex/pull/6433

ElFantasma (approved):
LGTM

PR #6973, opened 2026-07-07
feat(l1): state-bench read/write benchmarking harness (BAL + streaming, cold + warm)

edg-l:
## What

New tooling crate `tooling/state_bench` (`state-bench`): measures ethrex state read/write performance repeatably and A/B (ethrex-vs-ethrex) on one machine, across **both merkle import paths and both cache regimes**:

- **BAL parallel path** (Amsterdam+ genesis, e.g. `l1-bal.json`) — the optimistic/sharded merkleizer.
- **BAL-less / streaming path** (pre-Amsterdam genesis, e.g. `l1.json`) — the realistic mainnet import path (`handle_subtrie`), no BAL anywhere.
- **cold** (O_DIRECT + small block cache, the default) or **warm** (raise `--block-cache-bytes` so the working set stays cached).

Four steps:

- `gen-state` — synthetic fixture: N small storage accounts + one multi-GB mega account + a shared accessor contract + a funded signer; generates flat-KV; writes a manifest.
- `gen-workload` — real blocks (payload builder) touching cold slots. On an Amsterdam+ genesis it also captures each block's BAL → `chain.rlp` + `bals.rlp`; on a pre-Amsterdam genesis the blocks carry no BAL and `run` drives the streaming merkleizer (`bals.rlp` empty). Built on a throwaway copy; fixture untouched.
- `run` — subprocess-per-run import (O_DIRECT + fresh block cache, optional `--drop-caches`), per-run metrics + a coldness self-check; reset via `--reset checkpoint` (default) or `undo`.
- `compare` — diffs two run logs (mean/median/stddev/CoV, jobs-mismatch guard).

Usage: `docs/developers/l1/state-benchmarking.md` and `tooling/state_bench/README.md`.

## Fixture build: parallel + bounded memory

The mega storage trie builds in parallel across 16 nibble shards, each with per-chunk arena eviction, so peak memory is bounded (20 GB fixture ≈ 10 GB RSS; previously OOM'd past ~5 GB). Shards insert in **sorted key order** so eviction re-reads only the O(depth) spine — the build stays cache-independent and ~4× faster at scale (20 GB in ~13 min vs ~52 min). Flat-KV is written directly from the generated data instead of the store's scan-based generator.

Build/workload knobs: `gen-state --block-cache-gb` (raise the RocksDB cache for faster large builds); `gen-workload --hot-slots N` (draw reads from an N-slot hot working set for temporal locality — needed to benchmark read-through / value caches, e.g. #6979; default 0 = uniform cold); `gen-workload --zipf-s` (mainnet-like read skew); genesis choice selects BAL vs streaming path (above).

## Storage support (feature-gated, production-inert)

Behind a new `bench-harness` feature on `ethrex-storage`, off by default: `RocksDbOpenOpts` (O_DIRECT + statistics) on `RocksDBBackend::open` + `statistics_string()`; gated `Store::from_backend_bench`; two `StoreConfig` flags (default false); `DB_COMMIT_THRESHOLD` made `pub`. `cargo check -p ethrex-storage` without the feature confirms production opens are unchanged. Two additive helpers used only by the harness: `Trie::clear_dirty` and `Store::flush`.

## Notes

- On `main` the store commits trie/flat-KV layers *rolling* during import (at `commit_threshold`=128 depth), so cold **writes** land in `loop_seconds` and `commit_seconds` is tiny; the last ~128 blocks are an unmeasured tail, so use large workloads.
- `--jobs` floors at 17 (the BAL merkleizer needs 16 concurrent workers); scales up only.
- Undo-mode reset has cache-miss variance from async compaction — checkpoint is the reproducible default.

## Testing

End-to-end (`gen-state` → `gen-workload` → `run` → `compare`) on a 20 GB fixture, both reset modes: stable cold metrics, self-check passing. Pre-Amsterdam path validated via `gen-workload --verify-reimport` (streaming merkleizer, `bals=0`). Parallel build gated by `sharded_build` (sorted root == serial root, disk read-back, 20-iter root-collision stress) and `fkv_equivalence` (direct flat-KV byte-identical to the store generator). `state_bench_undo` covers the recording backend + undo log. clippy + fmt clean.

<details>
<summary>Findings from using the harness (validated PR #6986)</summary>

Drove the cold/warm A/B for the streaming-merkleizer trie-node prefetch (PR #6986), pre-Amsterdam fixture (4 GB mega account, `l1.json`), 3 runs each:

| scenario | baseline | +prefetch | Δ |
|---|---|---|---|
| cold, write-heavy → mega (64 MiB cache) | 336.5 s | 171.7 s | **−49% (1.96×)** |
| light-cold (8r/8w, mega_frac 0.5) | 47.1 s | 45.7 s | neutral (prefetch gated off; identical read metrics) |
| warm (heavy workload, 6 GiB cache) | 7.56 s | 6.92 s | neutral, no regression |
| threshold=1 (always-prefetch), light-cold | — | — | neutral, no upside |

All runs byte-identical (every block's state root validated on import).

**Limitation surfaced:** the synthetic workload writes unique fresh slots per block, so it did **not** catch a duplicate-slot last-write-wins bug in the prefetch (a stale value winning after an unstable sort) that the EF fixtures did — motivates a future duplicate-slot workload mode.

</details>


crates/blockchain/blockchain.rs
```diff
@@ -4550,3 +4550,156 @@ pub fn compute_sharded_storage_root(
...
+/// subtrees would emit extra (unreachable) nodes vs the canonical serial diff
+/// (see the `occupied <= 1` note on [`compute_sharded_storage_root`]). With
+/// keccak-uniform keys, `>= 4096` slots occupies all 16 nibbles with
+/// overwhelming probability, so the parallel path always has `> 1` root child —
+/// which is also why the `EMPTY_TRIE_HASH` placeholder passed to
+/// `collapse_root_node` below is never dereferenced.
+///
+/// Exposed (hidden) for `state_bench`; not part of the stable public API.
+#[doc(hidden)]
+pub fn build_sharded_storage_trie_streaming(
```
ElFantasma, line 4590: gating: `#[doc(hidden)] pub fn` still exports this on the crate's public API surface — any downstream crate can call it and skip the layered write path (this uses `open_direct_storage_trie` and writes directly to disk, so calling it on a live store would corrupt state). Consider matching `from_backend_bench`'s pattern and gating behind `#[cfg(feature = "bench-harness")]`:

```rust
#[cfg(feature = "bench-harness")]
#[doc(hidden)]
pub fn build_sharded_storage_trie_streaming(...)
```

That way `state_bench` (which will enable the feature anyway) is the only path that even sees the symbol, and a stray `use ethrex_blockchain::build_sharded_storage_trie_streaming` from any other crate stops compiling instead of corrupting silently. Non-blocking; the doc-comment warning is already explicit.

ElFantasma (approved):

PR #6974, opened 2026-07-07
feat(l1): add ethrex_simulateFrameTransaction RPC

ilitteri:
**Motivation**

Frame transactions (EIP-8141, type `0x06`) are opaque to the standard `eth_call`/`eth_estimateGas`/`eth_simulateV1` call model, whose input is a single flat call with no frames or validation prefix. A client therefore has no way to check whether a frame transaction is valid, or to size its multi-frame gas limits, before submitting it with `eth_sendRawTransaction`.

**Description**

Adds `ethrex_simulateFrameTransaction(rawTx [, block])` under a new `ethrex` JSON-RPC namespace. It dry-runs a canonical type-`0x06` frame transaction — without submitting it — against a throwaway state at the requested head:

1. the EIP-8141 validation-prefix simulation (the same machinery mempool admission runs), reporting validity, the recognized prefix shape, the resolved payer, and the failure reason if any; then
2. if the prefix passes, a full multi-frame execution reporting an accurate total gas figure plus per-frame gas and success.

Response object: `valid`, `prefixShape`, `payer`, `maxCost`, `violation`, `gasUsed`, `frames[]` (each `{gasUsed, succeeded}`), `executionStatus`, and `executionError`. The method is read-only and never touches the mempool or block building.

Namespace choice: the method lives in a new `ethrex` namespace rather than `eth_` (reserved for standardized methods) or `debug_` (namespace exposure is all-or-nothing, so enabling `debug_` publicly would also expose `debug_setHead` and the trace methods). Operators can expose only this method on a public endpoint via `--http.api ethrex`.

Denial-of-service: the full-execution step is bounded, before any EVM runs, by the same per-transaction gas cap `eth_estimateGas` applies (`get_max_allowed_gas_limit`). A frame transaction whose total gas limit exceeds the cap is reported invalid and not simulated, so a public endpoint cannot be driven into unbounded EVM execution by an attacker-chosen frame gas limit. This also adds `FrameTransaction::max_cost` as the single source of truth for the TXPARAM `0x06` max cost; the backend's `frame_tx_max_cost` now delegates to it.

Tests: handler `parse` (accept/reject paths), namespace routing, and `--http.api ethrex` parsing. Docs: an ethrex-extension section in `docs/eip-8141.md`.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A: read-only RPC, no `Store` changes.


ElFantasma (commented):
Copilot and Greptile have already flagged the substantive issues; both look real and worth acting on:

- **`payer` under-reports self-funded senders** (Copilot, `ethrex.rs:197`) — `outcome.accessed_paymaster` is `None` for self-funded prefixes (payer == sender), so today the RPC returns `payer: null` even when the prefix passed and the payer is well-defined. The field doc ("paymaster or self-funded sender") is the intended contract; the fix is to fall back to `frame_tx.sender` when `accessed_paymaster` is `None` (guarded on `outcome.passed` if you want to keep `null` for the failure path).

- **`gasUsed` is pre-refund on Amsterdam+** (Greptile, `ethrex.rs:297`) — `report.gas_used` is the EIP-7778 block-accounting figure post-Amsterdam, while `gas_spent` is what the sender actually pays. For an RPC whose whole purpose is "help the client size the gas limit before submitting", the post-refund figure is the one they'll want. This is the same distinction the debug_traceCall top-frame `gasUsed` fix from #6931 landed on last week (post-refund matches the receipt).

- **Docs for `valid` / `executionError`** (Copilot, `ethrex.rs:47` and `docs/eip-8141.md:679`) — small alignment fixes: `valid` also reports the pre-simulation gas-cap gate, and the docs currently say `executionError` carries the gas-cap violation while the code puts it in `violation`. Worth tightening before landing.

Issue #6976, opened 2026-07-08
prewarm: delta passes warm nothing for deep-queue senders (follow-up to #6967)


edg-l:
## Context

Follow-up from #6967 (mempool prewarming). Not a correctness issue; the whole warming layer is read-only speculative execution and execution falls back to the cold path on any miss. This is a warming-effectiveness gap.

## Problem

Delta-pass warming is largely wasted on busy mempools.

In `crates/blockchain/prewarm.rs`, `drop_already_warmed` strips a sender's already-warmed prefix (nonces `0..k`) before each delta pass. `LEVM::warm_txs` (`crates/vm/backends/levm/mod.rs`) then builds that sender's group from only the suffix (`k+1..`) and executes it against a fresh `GeneralizedDatabase` overlay on parent state. Pass 1's nonce/balance changes never persist into the shared `CachingDatabase` (only reads are cached; state mutations die with the discarded overlay).

Result: nonce `k+1` hits a nonce-mismatch, short-circuits inside `execute_tx_in_block`, warms nothing, and still burns the per-sender cap (`warmed_per_sender` increments on warm-set membership, not on success). Since most included txs arrive moments before their block, this undercuts the main "refreshing delta" mechanism the PR relies on.

## Fix options

1. Carry a per-sender speculative overlay across delta passes: persist each sender's `GeneralizedDatabase` between passes and replay only the new suffix on top. More efficient.
2. Simpler: don't drop the prefix. Keep each sender's full `0..N` sequence in every delta's replay set so nonces resolve. Roughly a one-line change to skip `drop_already_warmed`'s intra-sender pruning.

## Related (from PR review, optional)

- Synthetic child header never updates `gas_limit` (stays at parent's); from Amsterdam intrinsic state gas depends on it, so warming can select a slightly different set than a block that bumps its limit. Not a consensus break. Consider invalidating handoff when `block.gas_limit != parent.gas_limit`.
- `select_warm_set` deep-clones every selected `Transaction`; move the owned `MempoolTransaction` (Arc bump) instead.
- `warm_merkle_paths` re-opens the state trie every delta pass; `parent.state_root` is fixed for the slot, so open once in `run_pass`.

PR #6977, opened 2026-07-08
perf(l1): make prewarm delta passes replay full sender queue

edg-l:
## Motivation

Follow-up to #6967, fixing #6976. The mempool prewarmer's "refreshing delta" passes were largely ineffective on a busy mempool.

## Problem

Each delta pass dropped a sender's already-warmed prefix (`drop_already_warmed`) and handed `LEVM::warm_txs` only the new suffix. `warm_txs` rebuilds each sender group from parent state on every call, so a suffix tx (nonce N+1) ran without its predecessor (nonce N), hit a nonce-mismatch, short-circuited, and warmed none of the state it touches. It also still counted against the per-sender cap (`cap_sender_depth`'s cumulative `warmed_per_sender`), so deep-queue senders exhausted their 16-tx budget on warms that did nothing.

Net effect: only the first pass of a slot warmed a multi-tx sender correctly; every subsequent delta for that sender was wasted CPU. Since most included txs arrive moments before their block, this undercut the main mechanism the feature relies on.

## Change

Replace `drop_already_warmed` + `cap_sender_depth` with a single `select_dirty_senders`:

- Truncate each sender's queue head-first (nonce order) to the cap.
- Keep a sender only if its in-cap queue still has a not-yet-warmed tx; drop senders whose whole in-cap queue is already warmed.
- Surviving senders keep their **full** in-cap queue, so `warm_txs` replays the prefix and later-nonce txs get their nonce/balance predecessors and actually warm.

Re-running an already-warmed prefix is cheap: its reads are served from the shared per-slot `CachingDatabase`. The cap is now a straight per-pass depth truncate, so it can no longer be burned by failed warms; the cumulative `warmed_per_sender` counter is gone.

Strictly read-only speculative warming, unchanged: results are still discarded and never reach shared state.

## Testing

- `cargo test -p ethrex-blockchain prewarm::` (7 pass), including three new `select_dirty_senders` cases: cap truncation head-first, full-queue retention when the prefix is already warmed, and dropping a sender whose only new tx sits beyond the cap.
- `cargo clippy -p ethrex-blockchain` clean.

## Follow-up

A per-sender speculative overlay carried across delta passes (issue option 1) would avoid re-executing the prefix entirely; deferred as a larger change to `warm_txs`. This PR takes the minimal, contained fix.

### DIFF
--- a/crates/blockchain/prewarm.rs
+++ b/crates/blockchain/prewarm.rs
@@ -86,39 +86,31 @@ fn select_warm_set(
 #[cfg_attr(any(not(feature = "rayon"), feature = "eip-8025"), allow(dead_code))]
 const MAX_WARMED_TXS_PER_SENDER_PER_SLOT: usize = 16;
 
-/// Enforces [`MAX_WARMED_TXS_PER_SENDER_PER_SLOT`] across a slot's delta
-/// passes: `warmed_per_sender` counts txs already warmed this slot, and each
-/// sender's snapshot group is truncated to the remaining allowance. Nonce
-/// order within groups is preserved; exhausted senders are removed.
+/// Prepares a delta pass's warm set from a fresh mempool snapshot: truncates
+/// each sender's queue (head-first, nonce order) to `cap`, then keeps only
+/// senders whose truncated queue holds at least one not-yet-warmed tx (by
+/// hash). Senders whose entire in-cap queue is already warmed this slot are
+/// dropped.
+///
+/// The surviving senders keep their **full** in-cap queue, not just the new
+/// suffix. This is required for warming to be effective: `warm_txs` rebuilds
+/// each sender group from parent state on every pass, so a tx only warms the
+/// state it touches if its nonce/balance predecessors run first. Replaying a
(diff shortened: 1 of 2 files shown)


crates/blockchain/prewarm.rs
```diff
@@ -86,39 +86,31 @@ fn select_warm_set(
 #[cfg_attr(any(not(feature = "rayon"), feature = "eip-8025"), allow(dead_code))]
 const MAX_WARMED_TXS_PER_SENDER_PER_SLOT: usize = 16;
```
ElFantasma, line 87: nit: after this change this is a per-pass depth cap, not a hard per-slot cap. Under head-tx replacement (a sender bumps a nonce mid-slot, giving that slot in the window a new hash), the sender's warmed-tx count for the slot can exceed 16. Either the name (e.g. `MAX_SENDER_QUEUE_DEPTH_PER_PASS`) or the docstring above could be updated to reflect that.

ElFantasma (approved):
Fix looks correct; one naming nit inline. Endorse greptile's gas-budget-crowd-out observation as a known non-blocking trade-off.

PR #6979, opened 2026-07-08
perf(l1): read-through cache for committed on-disk state values

edg-l:
## What

Adds a read-through value cache (`CleanCache`) beneath the in-memory diff-layer overlay in the trie read cascade:

```
TrieWrapper::get
  → TrieLayerCache overlay   (last ~128 blocks of writes)
  → CleanCache               (this: committed on-disk values)  ← new
  → disk (RocksDB)           (fill CleanCache on a hit)
```

## Why

Today there are only two things that keep a state read off disk:
- the **128-block overlay**, which holds *writes* from recent blocks, and
- **per-block memoization** (`StoreVmDatabase.account_state_cache`, `CachingDatabase`), rebuilt and thrown away every block.

So an account or slot that is **read a lot but not written** (a popular router/token/proxy) is not in the overlay and the per-block scratch is gone next block, it re-reads disk **every block**. This cache absorbs those repeats. It also composes with the BAL/prefetch work (#6872/#6873/#6878): prefetch handles first-touch-this-block, this handles repeat-across-blocks.

## Correctness

The on-disk state is a **single-version, path-keyed** store and `BackendTrieDB::get` ignores the state root, so a cache entry is just a memoization of that root-independent point get. The overlay always shadows any recently-written key, so a stale entry can never be observed for a recently-written key.

To keep invalidation trivial, the cache is **gated on FKV-complete (synced)** via `clean_cache_if_synced()`. In that regime the only writers of the four state CFs are:
- `commit_trie_layers` — invalidates each committed key (done in the write loop; race-free because the overlay still shadows those keys until the RCU swap);
- `write_storage_trie_nodes_batch` — bulk clear.

Before sync (genesis, snap-sync healing, FKV backfill) other paths mutate on-disk state without invalidating, so the cache simply stays off. Locked/snap-sync tries also never use it.

## Memory (bounded)

Hard-bounded **by bytes, not entry count** (the thing you asked about):
- default budget **512 MiB** (`DEFAULT_CLEAN_CACHE_MAX_BYTES`), split evenly across **64 shards**;
- each shard evicts its LRU tail until under its share; accounted weight = `key.len() + value.len() + 64 B` overhead, so the figure tracks real resident memory, not just payload;
- it is resident memory *in addition to* the RocksDB block cache, but far denser: it stores ~40-130 B decoded-path values instead of whole 4-16 KB SST blocks, so for random hashed-key access it holds far more of the working set per byte.

Observability: `Store::clean_cache_stats() -> (used_bytes, max_bytes, entries)`.

Sharding uses per-shard `Mutex` (keyed by the uniform leading nibbles) to avoid a global lock under the parallel executor + prefetch workers.

## Scope / limits

- Wins on **steady-state mainnet head-following** (Zipfian repeated reads). Does **not** help the `sload_bloated` / large-state suite (read-once slots never hit a cache) — that is what the parallel-prefetch PRs target.
- Positive-only caching in v1 (no negative caching); raw RLP bytes (skips the disk seek, not the decode — a typed decode-skipping cache at `Store::get_*` is a possible follow-up).
- Budget is a `const` for now; `TODO` to move to `StoreConfig`.

## Tests

- `crates/storage/clean_cache.rs`: get/insert/invalidate/clear, **byte-bound never exceeded under hammering**, oversized-entry rejection.
- Full `ethrex-storage` suite passes (27 tests), no read/commit-path regression. `cargo fmt` + `cargo clippy` clean.

## Follow-ups (not in this PR)

- End-to-end synced-path test driving the FKV generator to completion.
- `state-bench` (#6973) scenario: repeated hot-contract reads across N blocks to measure the delta.
- Frequency-pinning + persist-hot-set-on-shutdown to also kill the restart cold-start cliff.

Draft: correctness-reviewed and unit-tested, but not yet benchmarked on a synced node.

## Docs

- `docs/perf/storage-cold-paths.md` (new): reference for perf analyzers on how cold storage access and accounts work — the read tiers (overlay -> FKV point get -> trie-node walk), what makes a read truly cold, the 128-block window, FKV watermark, Store caches, and the RocksDB tuning that moves the cold boundary. Linked from `docs/perf/README.md`.
- `CHANGELOG.md`: entry under Perf.

### DIFF
--- a/crates/storage/layering.rs
+++ b/crates/storage/layering.rs
@@ -6,6 +6,8 @@ use std::{fmt, sync::Arc};
 
 use ethrex_trie::{Nibbles, TrieDB, TrieError};
 
+use crate::clean_cache::CleanCache;
+
 const BLOOM_SIZE: usize = 1_000_000;
 const FALSE_POSITIVE_RATE: f64 = 0.02;
 
@@ -271,6 +273,10 @@ pub struct TrieWrapper {
     pub state_root: H256,
     pub inner: Arc<TrieLayerCache>,
     pub db: Box<dyn TrieDB>,
+    /// Read-through cache of committed on-disk values, consulted after the
+    /// overlay and before disk. `None` disables it (locked/snap-sync tries and
+    /// any not-yet-synced regime, see the gate in `store.rs`).
+    clean: Option<Arc<CleanCache>>,
     /// Pre-computed prefix nibbles for storage tries.
     /// For state tries this is None; for storage tries this is
     /// `Nibbles::from_bytes(address.as_bytes()).append_new(17)`.
@@ -283,12 +289,14 @@ impl TrieWrapper {
         inner: Arc<TrieLayerCache>,
         db: Box<dyn TrieDB>,
         prefix: Option<H256>,
+        clean: Option<Arc<CleanCache>>,
     ) -> Self {
         let prefix_nibbles = prefix.map(|p| Nibbles::from_bytes(p.as_bytes()).append_new(17));
         Self {
             state_root,
             inner,
             db,
+            clean,
             prefix_nibbles,
         }
     }
@@ -322,10 +330,24 @@ impl TrieDB for TrieWrapper {
             Some(prefix) => prefix.concat(&key),
             None => key,
         };
+        // Tier 1: in-memory diff-layer overlay (recent writes).
         if let Some(value) = self.inner.get(self.state_root, key.as_ref()) {
             return Ok(Some(value));
         }
-        self.db.get(key)
+        // Tier 2: read-through cache of committed on-disk values.
+        let Some(clean) = &self.clean else {
+            return self.db.get(key);
+        };
+        let cache_key: Box<[u8]> = key.as_ref().into();
+        if let Some(value) = clean.get(&cache_key) {
+            return Ok(Some(value));
+        }
+        // Tier 3: disk. Fill the cache on a hit (positive-only caching).
+        let value = self.db.get(key)?;
+        if let Some(ref v) = value {
+            clean.insert(cache_key, v);
+        }
+        Ok(value)
     }
 
     fn put_batch(&self, _key_values: Vec<(Nibbles, Vec<u8>)>) -> Result<(), TrieError> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## Perf
 
+### 2026-07-08
+
+- Add a bounded read-through cache of committed on-disk state values beneath the diff-layer overlay, so repeated cold reads of hot-but-unwritten accounts/slots (read every block but not recently written) hit memory instead of re-reading disk each block [#6979](https://github.com/lambdaclass/ethrex/pull/6979)
+
 ### 2026-07-06
 
 - Warm state caches between blocks by speculatively executing top-of-mempool transactions [#6967](https://github.com/lambdaclass/ethrex/pull/6967)
--- a/crates/storage/clean_cache.rs
+++ b/crates/storage/clean_cache.rs
@@ -1,0 +1,295 @@
+//! Read-through value cache for committed on-disk state.
+//!
+//! # What it is
+//!
+//! A bounded, sharded LRU that memoizes the result of a trie/flat-KV point
+//! lookup against the on-disk state (the `*_trie_nodes` / `*_flatkeyvalue`
+//! column families). It sits *beneath* the in-memory diff-layer overlay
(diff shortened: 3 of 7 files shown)


edg-l:
Benched this with the `state-bench` cold-state harness against a **synthetic 20 GB fixture** (one multi-GB storage account + small accounts), A/B = base vs this PR, 3 cold imports of 2000 blocks each, 17 workers. Two workloads that **bracket** the cache's behaviour:

**Hot — reads re-access a small working set (`--hot-slots 2000`):**

| metric | base | this PR | Δ |
|---|---|---|---|
| import time (median) | 181.1 s | 131.5 s | **−27%** |
| import time (mean) | 185.8 s | 138.0 s | −26% |
| sst_read_count | ~215k | ~150k | **~−30%** |
| bytes_read | 19.4 MB | 17.9 MB | −8% |

**Cold — uniform reads, ~no re-access (`--hot-slots 0`):**

| metric | base | this PR | Δ |
|---|---|---|---|
| import time (median) | 139.7 s | 139.2 s | **−0.3% (within noise)** |
| import time (mean) | 138.4 s | 138.3 s | −0.1% |
| block_cache_miss | 931k | 925k | −0.7% |

**Takeaway:** ~26% faster cold-import when reads hit a working set, and **no regression** on cold-miss-heavy workloads. The `sst_read_count` −30% (hot) is the cleanest signal — the read-through cache short-circuits ~a third of the disk lookups.

**Caveats (important):**
- **Neither workload is mainnet-representative.** The fixture (one giant account) and the two access patterns (uniform hot window / uniform cold) are synthetic micro-benchmarks that *bracket* the cache. Real mainnet access is skewed/Zipfian across many contracts, so expect *some* of the hot-case win (not the full 26%); the cold case establishes the floor (no regression).
- 3 runs only; hot-case variance was high (CoV 9–18%, though mean and median agree), cold-case was clean (CoV ~2%). More runs would tighten the hot number.
- Memory: the cache is bounded at 512 MiB, so worst case ~+512 MB RSS during import even when it barely hits (not sampled here).

Repro: `state-bench` harness (branch `perf/state-bench`) — `make gen-state MEGA_GB=20`, `make gen-workload NUM_BLOCKS=2000 MEGA_FRACTION=1.0 HOT_SLOTS=2000` (or `HOT_SLOTS=0` for cold), then `make run-a` / `run-b` / `compare`.

edg-l:
Follow-up: added a **Zipf read distribution** to the harness (`--zipf-s`, `r^-s` power law: a few hot slots + long cold tail) so reads approximate mainnet-like skew instead of uniform. Re-ran the same A/B (2000 blocks, 3 runs, full 135M-slot space, `--zipf-s 1.0`):

| metric | base | this PR | Δ |
|---|---|---|---|
| import time (median) | 146.1 s | 134.7 s | **−7.8%** |
| import time (mean) | 143.7 s | 135.1 s | −6.0% |
| block_cache_miss | 932k | 906k | −2.8% |
| bytes_read | 19.5 MB | 18.2 MB | −6.7% |

Low variance (B CoV <1%). This lands between the earlier brackets — **hot −26% / Zipf −7% / cold −0%** — so **~6–8%** is the more mainnet-representative estimate (skewed access, not the flat hot window). Still synthetic (one mega account, accessor `SLOAD`/`SSTORE`, not real opcode/gas mix), but the access-skew is now realistic.

crates/storage/layering.rs
```diff
@@ -322,10 +330,24 @@ impl TrieDB for TrieWrapper {
...
         if let Some(value) = self.inner.get(self.state_root, key.as_ref()) {
             return Ok(Some(value));
         }
-        self.db.get(key)
+        // Tier 2: read-through cache of committed on-disk values.
+        let Some(clean) = &self.clean else {
+            return self.db.get(key);
+        };
+        let cache_key: Box<[u8]> = key.as_ref().into();
+        if let Some(value) = clean.get(&cache_key) {
```
ElFantasma, line 342: nit / perf: allocating `cache_key: Box<[u8]>` before the cache lookup pays a heap alloc on the hit path — which is the common case for a well-warmed cache. Reorder to allocate only on miss:

```rust
if let Some(value) = clean.get(key.as_ref()) {
    return Ok(Some(value));
}
let cache_key: Box<[u8]> = key.as_ref().into();
let value = self.db.get(key)?;
if let Some(ref v) = value {
    clean.insert(cache_key, v);
}
Ok(value)
```

`Nibbles::as_ref` is a `&[u8]` view, so `clean.get(key.as_ref())` works with no allocation. Non-blocking; the savings compound at hot-key read rates.

ElFantasma (approved):

PR #6980, opened 2026-07-08
perf(l1): sharded + concurrent BAL prefetch warming (flat-KV, accounts, trie-nodes)

edg-l:
Warms cold reads on storage-heavy blocks. `async_io` is off (no io_uring), so a single `multi_get` runs at queue depth ~1 and serializes thousands of random cold reads. Every path below applies the same fix: sort keys by hashed order (adjacent entries share 4KB RocksDB data blocks), then split into contiguous shards read concurrently via `std::thread::scope` on one snapshot. `KEYS_PER_SHARD=256`, `MAX_SHARDS=64`.

## 1. Sharded storage-slot prefetch

`Store::get_storage_values_batch_by_root`, wired into `CachingDatabase::prefetch_storage`. Sorts FKV leaf keys, shards them, reads concurrently. Slots not yet swept by the FKV generator fall back to per-slot trie walks, byte-identical to the single-slot path.

Gated on `BLOATED_BATCH_THRESHOLD=16_384`, counting missing (cold) slots, not total accesses: warm/normal blocks keep the per-slot `par_iter` point-gets (sharding costs ~28% on a warm batch) since warm slots are never counted. 16384 cold slots is ~34M gas of cold reads, above ordinary cold blocks and below the large-state blocks this targets. Also hardens against storage-bloat DoS.

Microbench (6 cores pinned, O_DIRECT cold, one-bloated-account, ~200M-gas batch):
- `par_iter` baseline (QD ~6): 0.060 Mslots/s
- sharded multi_get (QD ~64): 0.255 Mslots/s (~4.25x)
- serial multi_get (regressing path): 0.017 Mslots/s
- range-scan: 0.033 Mslots/s

The win grows as cores shrink, since sharding decouples queue depth from core count.

## 2. Sharded account-state prefetch

`get_account_states_batch_by_root` read its FKV keys with a single `multi_get` on `ACCOUNT_FLATKEYVALUE` at queue depth 1. Same sort + shard treatment as the storage path. `prefetch_accounts` gates like `prefetch_storage`, counting missing (uncached) accounts so warm blocks stay on point-gets regardless of total count. `MAX_SHARDS=64` (accounts are scattered; 128 over-fragments).

## 3. BAL-driven trie-node prefetch

The flat-KV prefetch warms the leaf CFs execution reads. The merkle phase reads `STORAGE_TRIE_NODES` and `ACCOUNT_TRIE_NODES`, a separate set of CFs where every read is cold. This warms those internal nodes from the same BAL working set.

Internal nodes sit at prefixes of a leaf's nibble path, so for each BAL leaf we probe every prefix up to depth 8. Prefixes with no real node are rejected by the in-RAM CF bloom filter, no disk I/O. Result is one sorted, sharded `multi_get` per CF; dedup collapses shared shallow prefixes so roots and top branches read once.

Covers storage tries (BAL slots) and the account trie (BAL addresses, including balance/nonce-only accounts). Derived from the merkle write set, not the access set: `synthesize_bal_updates` already drops read-only accesses. Gated on `slots.len() + accounts.len() >= 16_384` (~34M gas of cold accesses). Best-effort and read-only; a missed node is cold-read by the walk and cannot affect consensus output.

## 4. Concurrent trie-node prefetch

The trie-node prefetch runs on its own thread inside the `std::thread::scope` holding the executor and merkleizer, overlapping execution instead of preceding it.

Safe to overlap: it reads the trie-node CFs, which execution never touches (execution reads only flat-KV via SLOAD/SSTORE), so no executor contention. Against the merkleizer it acts as a high-queue-depth batch, raising aggregate queue depth so scattered cold reads land warm before the per-account walk reaches them. The flat-KV warm stays synchronous before execution since it serves the executor.

Best-effort: spawn failure is logged and skipped; the scope joins the handle before returning; a thread panic is caught via `h.join()` and surfaced with `warn!`. No deadlock path (single batch read, no shared locks).

## Correctness

`test/tests/storage/storage_batch_tests.rs`, all passing (in-memory + rocksdb):
- `storage_batch_parity_*`: sharded storage batch byte-identical to per-slot single-get across FKV-swept and trie-fallback states (rocksdb test crosses shards, `POPULATED_SLOTS=2048`)
- `account_batch_parity_*`: sharded account batch byte-identical to per-account single-get
- `trie_node_prefetch_effectiveness_*`: 2048-slot contract, hit count >= 64; a key-encoding regression turns every probe into a bloom miss and fails this

EF blockchain LEVM suite: 8758 passed / 0 failed.

## Notes

Targets storage/account-heavy cold reads; gates keep every path inert on normal/warm blocks. Does not close the throughput gap to other clients on the bloated suite, which is bound by storage layout (hashed-key FKV random reads) and the exec/merkle pipeline.


crates/vm/backends/levm/mod.rs
```diff
@@ -2269,6 +2269,14 @@ impl LEVM {
             .collect()
     }
 
+    /// Every account address touched in the BAL. Drives the account-trie
+    /// (state-trie) node prefetch — including balance/nonce-only accounts with
+    /// no storage slots (the CALL/BALANCE-family blocks), which
+    /// `bal_storage_slots` does not surface.
+    pub fn bal_accounts(bal: &BlockAccessList) -> Vec<Address> {
```
ElFantasma: This helper looks orphaned — `LEVM::bal_accounts` isn't called anywhere in the tree. `blockchain.rs` builds the trie-prefetch input from `updates.keys()` (the merkle write set from `synthesize_bal_updates`), and switching to `bal_accounts` would undo the deliberate write-set-not-access-set choice made in commit 7427a89 ("prefetch only the merkle write set, not the BAL access set") that regressed read-heavy blocks. The docstring above still claims it "drives the account-trie node prefetch", which no longer matches. Suggest dropping the fn (and correcting the docstring on `bal_storage_slots`, which cross-references it).

crates/blockchain/blockchain.rs
```diff
@@ -696,6 +703,38 @@ impl Blockchain {
...
+        // runs on its own thread inside the scope below (`trie_prefetch_handle`),
+        // overlapping execution rather than preceding it. Gated so ordinary
+        // merkle-light blocks skip the probe cost; the payoff is on blocks that WRITE
+        // many distinct slots or modify many accounts, where merkle walks a large set
+        // of scattered, cold nodes. Tunable.
+        #[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
+        let trie_prefetch_input: Option<TriePrefetchInput> = if self.options.bal_prefetch_enabled
+            && let Some(updates) = optimistic_updates.as_ref()
+        {
+            const MERKLE_PREFETCH_THRESHOLD: usize = 16_384;
```
ElFantasma: nit: this `MERKLE_PREFETCH_THRESHOLD = 16_384` and the two `BLOATED_BATCH_THRESHOLD = 16_384` locals in `crates/vm/levm/src/db/mod.rs` (`prefetch_accounts` line 331, `prefetch_storage` line 375) all encode the same "bloated batch" gate. Three copies of the same magic number in three files is fragile: tuning one almost certainly means tuning all three. Hoisting to a single shared `pub const` (next to `KEYS_PER_SHARD`/`MAX_SHARDS` in `store.rs`, since those are already the shared shard-tuning knobs) would make the coupling explicit.

crates/vm/backends/levm/mod.rs
iovoid: Dead code

crates/storage/store.rs
```diff
@@ -2899,6 +2936,279 @@ impl Store {
...
+        let last_written = self.last_written()?;
+        let trie_cache = self
+            .trie_cache
+            .read()
+            .map_err(|_| StoreError::LockError)?
+            .clone();
+
+        let mut results: Vec<Option<U256>> = vec![None; slots.len()];
+        // Per-slot prefixed FKV leaf paths: account_hash nibbles + separator
+        // (17) + hashed-slot nibbles + leaf flag. Length 131. This is the exact
```
iovoid: Not why it's 131. The way it breaks down is: 64 (account) + 1 (leaf) + 1 (separator) + 64 (storage) + 1 (leaf)

test/tests/storage/storage_batch_tests.rs
```diff
@@ -0,0 +1,372 @@
...
+    store
+        .add_initial_state(genesis)
+        .await
+        .expect("add genesis state");
+
+    // State A: FKV not yet generated -> the batched helper takes the per-slot
+    // trie-walk fallback for every slot. Must match single-get exactly.
+    assert_parity(&store, state_root, contract);
+
+    // State B: FKV fully swept -> the batched helper takes the sorted multi_get
```
iovoid, line 172: There is a third state (FKV partially complete) that is untested. It's very annoying to test, so it's understandable if you don't, but this comment implies all cases are covered.

crates/vm/levm/src/db/mod.rs
```diff
@@ -259,30 +320,71 @@ impl Database for CachingDatabase {
...
-        let states = self.inner.get_account_states_batch(&missing)?;
+        // Same gate as `prefetch_storage`: a large set of distinct COLD accounts is
+        // queue-depth bound. The inner batch path on the rocksdb-backed
+        // StoreVmDatabase used a single multi_get (queue depth 1, async_io off),
+        // which collapses on cold account-heavy blocks (coldbench: ~13x slower than
+        // the sharded batch). Route large/cold sets to the (now sharded) batch and
+        // small/warm sets to parallel point-gets. The gate counts MISSING (cold)
+        // accounts, so warm blocks stay on the point-get path however many accounts
+        // they touch. Tunable.
+        const BLOATED_BATCH_THRESHOLD: usize = 16_384;
```
iovoid: This constant is defined twice. It also has the same value and seemingly same purpose as MERKLE_PREFETCH_THRESHOLD.

iovoid (changes requested):

crates/storage/store.rs
```diff
@@ -2862,12 +2876,45 @@ impl Store {
...
+                    let handles: Vec<_> = keys
+                        .chunks(chunk)
+                        .map(|ck| scope.spawn(move || rv.multi_get(ACCOUNT_FLATKEYVALUE, ck)))
+                        .collect();
+                    handles
+                        .into_iter()
+                        // A panicked shard yields one Err element rather than
+                        // re-panicking the whole node; the consumer's `?` below
+                        // then surfaces it and the best-effort prefetch caller
+                        // swallows it (the slots are simply left uncached).
```
iovoid: bal_to_account_updates does propagate the error

crates/blockchain/blockchain.rs
```diff
@@ -696,6 +699,39 @@ impl Blockchain {
...
+                }
+                let write_accounts: Vec<Address> = updates.keys().copied().collect();
+                if write_slots.len() + write_accounts.len() >= MERKLE_PREFETCH_THRESHOLD {
+                    Some((write_slots, write_accounts))
+                } else {
+                    None
+                }
+            } else {
+                None
+            };
```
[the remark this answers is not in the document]
edg-l: Added the note (see \`blockchain.rs:719-723\`): \`prefetch_trie_nodes\` reads the trie-node CFs directly via \`backend.begin_read()\`, bypassing the witness-recording caching layer, so no \`!collect_witness\` guard is needed. (b83343e5c)

crates/vm/backends/levm/mod.rs
edg-l: Dropped \`LEVM::bal_accounts\` and corrected the docstrings that cross-referenced it. (b83343e5c)

crates/blockchain/blockchain.rs
edg-l: Deduplicated: there is now a single \`pub const BLOATED_BATCH_THRESHOLD\` in \`crates/vm/levm/src/db/mod.rs\`, re-exported via \`ethrex_vm::backends\` and used at all sites (blockchain.rs + both db/mod.rs paths). \`MERKLE_PREFETCH_THRESHOLD\` is gone. (b83343e5c)

crates/vm/backends/levm/mod.rs
edg-l: Removed \`LEVM::bal_accounts\`. (b83343e5c)

crates/storage/store.rs
edg-l: Fixed the breakdown to \`64 (account) + 1 (leaf) + 1 (separator) + 64 (storage) + 1 (leaf)\` (store.rs:2985). (b83343e5c)

test/tests/storage/storage_batch_tests.rs
edg-l: Added a note documenting the untested FKV-partially-swept state and why it is not deterministically reproducible in a test (storage_batch_tests.rs:176-181). (b83343e5c)

crates/vm/levm/src/db/mod.rs
edg-l: Deduplicated into the single shared \`BLOATED_BATCH_THRESHOLD\` (same value/purpose as the former \`MERKLE_PREFETCH_THRESHOLD\`, which is removed). (b83343e5c)

crates/storage/store.rs
edg-l: Reworded: the comment no longer claims the error is swallowed; \`bal_to_account_updates\` propagates via \`?\`. (b83343e5c)

crates/blockchain/vm.rs
```diff
@@ -228,6 +228,59 @@ impl VmDatabase for StoreVmDatabase {
...
+            return Ok(Vec::new());
+        }
+
+        // Resolve the account state (hashed address + storage root) for each
+        // distinct address. This mirrors the per-slot `get_storage_slot` path,
+        // which opens the storage trie from the cached account entry. Slots for
+        // a non-existent account resolve to `None`, exactly as the single-get
+        // path returns `None` when the account entry is missing.
+        let mut entries: FxHashMap<Address, Option<AccountStateCacheEntry>> = FxHashMap::default();
+        for &(addr, _) in keys {
```
ElFantasma, line 251: nit: this per-unique-address account resolution runs serially before the batch multi_get. Fine for the target case (bloated single-contract block → 1 lookup), but under a pathological cross-address bloated batch (16k slots spread over hundreds of distinct addresses), it re-introduces the exact serial-QD1 front the rest of the PR is eliminating — every address hits `get_cached_account_state_entry` in sequence, and each cold miss is a full account-trie walk.

Since the workload targeted here is explicitly "one bloated contract," this is fine as-is; worth a sentence in the doc-comment above making that assumption explicit so a future reader doesn't rework this path assuming the sharded design carries through end-to-end. If cross-address bloat later becomes a concern, `par_iter` on the unique addresses would restore fan-out with no other changes to this function.

crates/vm/levm/src/db/mod.rs
```diff
@@ -45,6 +51,17 @@ pub trait Database: Send + Sync {
             .map(|a| self.get_account_state(*a))
             .collect()
     }
+    /// Batch storage-slot lookup. Default: loop. Backends with a batched read
+    /// path (e.g. rocksdb `multi_get_cf` on the storage flat key-value table)
+    /// should override this and the caching layer above will dispatch to it.
+    fn get_storage_values_batch(
```
ElFantasma, line 57: nit: the doc doesn't state the length invariant `returned.len() == keys.len()`, but the caller relies on it. `prefetch_storage` (line 389) does `for (key, value) in missing.into_iter().zip(values.into_iter())` — if a future override returns a shorter Vec (partial success, dedup at backend level, whatever), zip silently truncates and some keys are cached, others aren't, with no error. The default impl trivially satisfies it (`iter().map().collect()`), and the current sharded rocksdb impl does too, but the contract should be stated so a future override doesn't quietly break the caching invariant. One-line doc note, or a `debug_assert_eq!(values.len(), missing.len())` at the caller.

ElFantasma (approved):
Re-review — both prior inlines cleanly addressed via b83343e5c; LGTM, but left two non-blocking small fresh-eyes nits inline.

crates/storage/store.rs
```diff
@@ -3152,6 +3200,291 @@ impl Store {
...
+    /// The batch is sorted by the prefixed FKV key (hashed order) before the
+    /// `multi_get` so that adjacent slots share RocksDB data blocks, removing
+    /// the random-order read amplification of per-slot lookups. Slots not yet
+    /// swept by the FKV generator fall back to per-slot trie walks via
+    /// [`Self::get_storage_at_root_with_known_storage_root`], which preserves the
+    /// exact FKV / trie / diff-cache ordering of the single-slot path.
+    ///
+    /// Results are returned in the same order as the input slots. A missing slot
+    /// (deleted or never written) yields `None`, identical to the single-get path.
+    pub fn get_storage_values_batch_by_root(
```
iovoid, line 3222: Missing the `cache.overlay_serves(state_root)` checks that the neighboring functions have in order to behave well under a deep reorg.

iovoid (changes requested):

crates/storage/store.rs
edg-l: fixed

iovoid (approved):

ilitteri (approved):

PR #6981, opened 2026-07-08
fix(l1): bound PooledTransactions decode size and reject duplicate pooled txs

### LINKED ISSUE #6928: Bound bytes read when decoding PooledTransactions instead of checking size after decode
**Context**

Follow-up from review of #6919 ([comment](https://github.com/lambdaclass/ethrex/pull/6919#discussion_r3483802454)). That PR relaxed the pooled-transaction size check in `PooledTransactions::validate_requested` (`crates/networking/p2p/rlpx/eth/transactions.rs`) from strict equality to an 8-byte tolerance, matching geth, to stop disconnecting peers over the 1-byte v1-sidecar size skew.

**Problem**

The announced-size check runs *after* the whole `PooledTransactions` message has been snappy-decompressed and RLP-decoded into memory. So it's a correctness/interop check, not a resource guard: a peer can make us decompress and decode a payload far larger than what we requested before the size comparison ever runs. The check rejects the mismatch, but only once the work is already done.

**Goal**

Bound the bytes we read *before/during* decode rather than checking size afterward:

- Enforce a per-response size budget derived from the sizes we requested (we already track `requested_pooled_txs` with the announced sizes/types), plus the tolerance, and reject/abort a response that exceeds it before fully materializing it.
- Cap the snappy *decoded* length before allocating, so a response can't be inflated past the frame budget in the first place.

**Notes**

- The post-decode equality/tolerance check can stay as a final sanity check; this issue is about adding the up-front bound so the decode path can't be used to force oversized allocations.
- Touches the RLPx decode path (`crates/networking/p2p/rlpx/`), so scope/verify against the snappy decompression and message-size handling there.

ElFantasma:
## Motivation

Closes #6928. Follow-up from #6919: `PooledTransactions::validate_requested` checked the announced size *after* the whole message was snappy-decompressed and RLP-decoded into memory, so it was a correctness check, not a resource guard. This adds up-front bounds so the pooled-transaction request/response paths can't be used to force oversized allocations or amplified responses. Also closes the EF findings `pooledtransactions-duplicate-response-amplification`, `getpooledtransactions-duplicate-hit-amplification`, and `getpooledtransactions-duplicate-miss-amplification`.

## Description

**Receive side — `PooledTransactions` responses:**
- **Duplicate guard** in `validate_requested`: a `HashSet` of seen tx hashes rejects the first duplicate (`MempoolError::DuplicatePooledTx`); with the existing "every tx must match a requested hash" check this bounds the response to one tx per requested hash. Checked before blob `validate_cheap` so a duplicate blob tx is rejected before any KZG work.
- **Pre-alloc decode cap**: `snappy_decompress_bounded(msg_data, max_len)` rejects an over-large *declared* decompressed length (via `decompress_len`) before allocating; `PooledTransactions::decode` passes `MAX_POOLED_TRANSACTIONS_BYTES` = 4 MiB. Enforced at the snappy header, not a stream `.take()` — RLPx uses raw block snappy (one-shot); a comment guards against a future "fix" that would break interop.

**Serve side — `GetPooledTransactions::handle`:**
- **Dedup** served hashes (serve each at most once, skipping the lookup for repeats) — closes the duplicate-hit/duplicate-miss amplification findings.
- **Byte cap**: stop before a tx would push the response over `MAX_POOLED_RESPONSE_BYTES` = 2 MiB (geth's `softResponseLimit`), so we never emit more than a peer's inbound cap accepts. Partial responses are protocol-legal.

**Request side:** `Mempool::reserve_unknown_hashes` now dedups within an announcement (it filtered on `in_flight_txs` before the `extend`, so `[h, h]` was reserved/requested twice — an honest echo would then trip the response-side dup guard).

## On the constants

- Global decompressed cap (`MAX_SNAPPY_DECOMPRESSED_LEN` ~16 MiB, = codec `MAX_MESSAGE_SIZE`) landed in #6956 — this is the per-message tightening on top.
- **Serving = 2 MiB** matches geth's `softResponseLimit` exactly (`eth/protocols/eth/handler.go`).
- **Inbound = 4 MiB**: a well-behaved response can reach ~3 MiB (soft-limit + one crossing tx, up to a ~1 MiB blob wrapper, per our #6599 wire cap), so the inbound cap must sit above that; 4 MiB clears it with margin and stays 4× under the frame cap. A tighter value would wrongly reject legit geth blob responses.

## Checklist

- [x] `cargo test -p ethrex-test --test ethrex_tests` (transactions + snap + utils) and `cargo clippy -p ethrex-p2p -p ethrex-blockchain --tests` pass.


crates/networking/p2p/rlpx/eth/transactions.rs
```diff
@@ -294,11 +304,19 @@ impl PooledTransactions {
...
+        // how little we asked for.
+        let mut seen = HashSet::with_capacity(self.pooled_transactions.len());
         for tx in &self.pooled_transactions {
             if let P2PTransaction::EIP4844TransactionWithBlobs(itx) = tx {
                 itx.blobs_bundle.validate_cheap(&itx.tx, fork)?;
             }
             let tx_hash = tx.compute_hash();
+            if !seen.insert(tx_hash) {
+                return Err(MempoolError::DuplicatePooledTx);
+            }
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 6bb68e806 — moved the `compute_hash` + `seen` dedup check ahead of `validate_cheap`, so a duplicate (incl. blob) tx is rejected before any per-tx work. Thanks.

ElFantasma:
Thanks for the reviews — addressed the actionable points in `6bb68e806`.

**@Codex #2 (self-inflicted duplicate request):** confirmed. `Mempool::reserve_unknown_hashes` filtered on `in_flight_txs` *before* the `extend`, so an announcement with `[h, h]` was returned — and requested — twice; an honest responder then echoes the tx twice, which the new response-side guard would wrongly treat as a fault. Fixed by deduping within the announcement (a local `seen` set), plus a regression test `reserve_unknown_hashes_dedups_within_announcement`.

**@greptile (dedup ordering):** moved the `compute_hash` + `seen` check ahead of `validate_cheap`, so a duplicate blob tx is rejected before any per-tx (KZG) work — fail-fast.

**@Codex #1 (inbound 4 MiB stricter than outbound):** confirmed real under blob load — `GetPooledTransactions::handle` serves uncapped and we batch requests by 256-hash count, so an ethrex peer could legitimately emit a `PooledTransactions` above 4 MiB, which a patched peer would then reject. The fix is a serving-side response byte cap (geth `softResponseLimit` style) so we never emit above a peer's inbound cap. I'm settling the exact approach (serving-side cap vs. size-bounded request batching) before implementing — follow-up commit to this PR.

Thanks @Kimi / @Claude for the approvals; `cargo fmt`/`clippy` pass locally.

ElFantasma:
**@Codex #1 (inbound cap stricter than outbound) — fixed in 7e3a60b29.**

`GetPooledTransactions::handle` now bounds the serving path both ways:
- **Byte cap:** accumulates encoded size and stops *before* a tx would cross `MAX_POOLED_RESPONSE_BYTES` = 2 MiB (geth's `softResponseLimit`, verified in `eth/protocols/eth/handler.go`). So we never emit above a peer's inbound decode cap — no more asymmetry. Partial responses are protocol-legal.
- **Dedup:** serves each requested hash at most once (skipping the lookup for repeats), which also closes two EF findings on this path — `getpooledtransactions-duplicate-hit-amplification` and `getpooledtransactions-duplicate-miss-amplification`.

On the values: 2 MiB serving = geth's `softResponseLimit` exactly; the 4 MiB *inbound* cap stays because a well-behaved response can reach ~3 MiB (soft-limit + one crossing tx, up to a ~1 MiB blob wrapper), so a tighter inbound cap would wrongly reject legit geth blob responses.

Tests: `get_pooled_transactions_handle_dedups_requested_hashes`, `get_pooled_transactions_handle_caps_response_bytes`.

crates/networking/p2p/rlpx/eth/transactions.rs
```diff
@@ -231,18 +248,36 @@ impl GetPooledTransactions {
...
             // building after being advertised.
-            .filter_map(|hash| blockchain.get_p2p_transaction_by_hash(hash).ok())
-            .collect::<Vec<_>>();
+            let Ok(tx) = blockchain.get_p2p_transaction_by_hash(hash) else {
+                continue;
+            };
+            // Cap the response size (geth `softResponseLimit`): stop before a tx would push the
+            // response over the budget so we never emit more than a peer's inbound cap accepts.
+            // Always serve at least one tx so a lone oversized (blob) tx still goes out.
+            let tx_size = tx.encode_canonical_to_vec().len();
```
iovoid: You can use encode_canonical_len to avoid the cost of encoding

ElFantasma: Done in 665f3521c. `tx` here is a `P2PTransaction`, which only had `encode_canonical_to_vec()` (allocates), so I added `P2PTransaction::encode_canonical_len()` — mirroring `encode_canonical` and summing each inner variant's `RLPEncode::length()`, exactly like the existing `Transaction::encode_canonical_len`. It's allocation-free and, for the blob variant, correctly includes the sidecar via `EIP4844TransactionWithBlobs::length()`. Swapped it in on the serving hot path (and the one test helper that built the same way); the value is identical, so the response-cap tests are unchanged. Left the pre-existing call sites alone to keep the PR scoped.

iovoid (approved):

ilitteri (approved):

crates/networking/p2p/rlpx/utils.rs
```diff
@@ -76,19 +76,37 @@ pub fn snappy_compress(encoded_data: Vec<u8>) -> Result<Vec<u8>, RLPEncodeError>
...
+/// Like [`snappy_decompress`] but rejects a declared decompressed length above `max_len` before
+/// allocating, for messages with a tighter natural bound than the global frame cap. `max_len` is
+/// clamped to [`MAX_SNAPPY_DECOMPRESSED_LEN`] so it can never exceed the global limit.
+///
+/// RLPx uses *raw* (block) snappy, which is one-shot: the block header declares the full
+/// decompressed length and `decompress_vec` produces the whole buffer in a single allocation —
+/// there is no output stream to wrap in a `Read::take(max_len)`. So the bound is enforced here,
+/// at the one point the format exposes the decoded size (the header), rather than by limiting a
+/// stream. A true streaming `.take()` bound would require snappy *frame* format, which is a
+/// different wire encoding and would break peer interop — don't "upgrade" this into one.
```
MegaRedHand: I think this part can be removed.

```suggestion
///
/// RLPx uses *raw* (block) snappy, which is one-shot: the block header declares the full
/// decompressed length and `decompress_vec` produces the whole buffer in a single allocation.
```

crates/networking/p2p/rlpx/utils.rs
```diff
@@ -76,19 +76,37 @@ pub fn snappy_compress(encoded_data: Vec<u8>) -> Result<Vec<u8>, RLPEncodeError>
 /// Maximum decompressed size accepted for a snappy-compressed RLPx message. Matches the
 /// compressed-frame cap (`MAX_MESSAGE_SIZE` = `0xFFFFFF`, ~16 MiB, in `connection/codec.rs`)
 /// and go-ethereum, which likewise bounds `snappy.DecodedLen` at maxUint24 before allocating.
-const MAX_SNAPPY_DECOMPRESSED_LEN: usize = 0xFF_FFFF;
```
MegaRedHand, line 79: Does this need to be pub?

MegaRedHand (approved):
LGTM

crates/networking/p2p/rlpx/utils.rs
ElFantasma: Applied in 3e5e83569 — doc trimmed to your suggestion.

crates/networking/p2p/rlpx/utils.rs
ElFantasma: Good call — it doesn't. Only used inside `utils.rs` (the `transactions.rs` mention is doc prose, not code). Made it private in 3e5e83569, and de-linked the intra-doc reference so rustdoc doesn't warn about linking a private item.

PR #6982, opened 2026-07-08
feat(l1,l2): zkevm execution benchmark

### LINKED ISSUE #6955: Implement a benchmark for zkEVM execution
After reading [this blog post](https://benbencik.github.io/blogs/zisk-allocator-comparison.html) by benbencik, I think we would find it useful to have a premade benchmark of real blocks and microbenchmarks, so we can optimize for the zkEVMs.

This has the benefit that, since the tests are deterministic, it's not vulnerable to noise and more friendly to autoresearch and similar.

The tool used by benbencik, ziskemu, should generalize decently to other zkEVMs.

iovoid:
**Motivation**

We want to benchmark the performance and memory usage of our client when executing inside a zkVM.

This is useful to detect and fix OOMs, and deterministically measure purported performance improvements

**Description**

This executes our zisk guest using ziskemu, giving information about cost and memory usage. The performance under ZisK should generalize well to other zkVMs.

This tool has three presets: `quick` (a couple of real blocks + selected worst-case tests), `medium` (real blocks + EEST microbenchmarks + selected worst-case tests), and `slow` (everything).

The workloads used are EEST tests to evaluate improvements to specific operations, resource-hogging fixtures (EEST tests-benchmark) to improve pathological cases, and real mainnet blocks to assess typical real-world performance.

The output format is JSON, for easy automated usage. Some future uses could be in the CI (to prevent regressions) and with autoresearch-like tools.

Closes #6955


ElFantasma (changes requested):
Nice tooling scope — deterministic ziskemu-based benchmarking with the three-tier preset structure is a good regression-signal shape. Fixtures correctly LFS-tracked (`.gitattributes` pattern `tooling/zkevm_bench/fixtures/**/*.json.gz`). Endorse greptile's 2 P2s and add one cross-file observation, plus flag the red CI:

- **`zisk_version` hardcoded** at `run.rs:148` — real bug: reports will misattribute across ziskemu upgrades. Greptile's suggested `detect_zisk_version()` via `ziskemu --version` is the right shape.
- **`curate.rs:115` rejects `.json.gz`** — reinforcement for greptile: `run.rs:52` (`discover_stress_workloads`) already accepts both `.json` and `.json.gz` via `name.ends_with(".json.gz") || name.ends_with(".json")`. Same tool, different sibling function, opposite behavior on the same file extension. The `curate` docstring documents the `.gz` skip explicitly (line 100–104: *"`.gz` archives — are skipped"*), so the design intent seems to be "curate is only for uncompressed ethrex-replay caches" — but a user pointing `curate` at `tooling/zkevm_bench/fixtures/blocks/` (where every fixture is `.json.gz`) will silently get 0 rows with no error. Either accept `.json.gz` for consistency with `discover_stress_workloads` + `load_cache` (both already handle both), or add a hard error when `rows.is_empty()` after the scan.
- **CI is red** (Cargo.lock check, Build Docker, and Test failing across ubuntu-22.04 / ubuntu-22.04-arm / macos-15). Author has a merge-fixup commit `9707f1606 chore: fix Cargo.lock versions after merge` at the tip; the check failures may reflect a still-stale lockfile or the newly-added `datatest-stable` / `ef_tests-blockchain` dep additions in `Cargo.lock`. Blocker for merge — worth confirming the lockfile matches `Cargo.toml` cleanly and re-running.

One minor observation on `run.rs:74`: the empty-ELF path prints a warning to stderr and continues — every workload then fails with `guest_output_ok: false`. That's arguably fine for a bench tool (still produces a report), but in a CI context that stderr line is easy to miss. A `--strict-elf` flag (or auto-abort when ELF is empty AND no workloads succeed) would prevent someone from mistakenly using an all-`false` report as a regression signal.

ElFantasma (approved):

ilitteri (approved):

PR #6983, opened 2026-07-09
fix(trie): reject over-32-byte node child reference instead of panicking

han0110:
**Motivation**

A branch/extension child reference is canonically either a 32-byte hash or an inline node under 32 bytes. `decode_child` routed anything that was neither a clean 32-byte hash nor empty into `NodeHash::from_slice`, which panics on slices over 32 bytes, so a malformed child field of over 32 bytes would panic. This is unreachable for nodes read from the DB, which are well-formed by construction, but in the guest program the `ExecutionWitness` is untrusted and may be malformed, so a single crafted node could abort the prover.

**Description**

This PR makes `decode_child` return an error when such a fallthrough child reference is 32 bytes or more (neither a 32-byte hash nor a sub-32-byte inline node), which is not a canonical encoding, instead of passing it to `NodeHash::from_slice` and panicking.

Ran the benchmark `cargo bench -p ethrex-rlp --bench decode -- 'Node::(Extension|Branch)'` and no significant performance change:

```
# Without the fix

decode_trie/Node::Extension/1000
                        time:   [47.607 µs 47.702 µs 47.824 µs]
decode_trie/Node::Branch/1000
                        time:   [255.04 µs 258.11 µs 262.82 µs]

# With the fix

decode_trie/Node::Extension/1000
                        time:   [47.454 µs 47.644 µs 47.830 µs]
decode_trie/Node::Branch/1000
                        time:   [255.80 µs 257.27 µs 258.66 µs]
```

**Checklist**

Closes #issue_number

### DIFF
--- a/crates/common/trie/rlp.rs
+++ b/crates/common/trie/rlp.rs
@@ -1,5 +1,3 @@
-use std::array;
-
 // Contains RLP encoding and decoding implementations for Trie Nodes
 // This encoding is only used to store the nodes in the DB, it is not the encoding used for hash computation
 use ethrex_rlp::{
@@ -12,7 +10,7 @@ use ethrex_rlp::{
 
 use ethrex_crypto::NativeCrypto;
 
-use super::node::{BranchNode, ExtensionNode, LeafNode, Node};
+use super::node::{BranchNode, ExtensionNode, LeafNode, Node, NodeRef};
 use crate::{Nibbles, NodeHash};
 
 // SAFETY: `NativeCrypto` is used here instead of a `&dyn Crypto` parameter because
@@ -132,7 +130,7 @@ impl RLPDecode for Node {
                     // Decode as Extension
                     ExtensionNode {
                         prefix: path,
-                        child: decode_child(rlp_items[1].expect("we already checked the length"))
+                        child: decode_child(rlp_items[1].expect("we already checked the length"))?
                             .into(),
                     }
                     .into()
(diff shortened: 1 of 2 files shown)


crates/common/trie/rlp.rs
```diff
@@ -161,10 +161,13 @@ impl RLPDecode for Node {
...
+fn decode_child(rlp: &[u8]) -> Result<NodeHash, RLPDecodeError> {
     match decode_bytes(rlp) {
-        Ok((hash, &[])) if hash.len() == 32 => NodeHash::from_slice(hash),
-        Ok((&[], &[])) => NodeHash::default(),
-        _ => NodeHash::from_slice(rlp),
+        Ok((hash, &[])) if hash.len() == 32 => Ok(NodeHash::from_slice(hash)),
+        Ok((&[], &[])) => Ok(NodeHash::default()),
+        _ if rlp.len() < 32 => Ok(NodeHash::from_slice(rlp)),
+        _ => Err(RLPDecodeError::Custom(
+            "invalid trie node child reference".to_string(),
```
ElFantasma: nit: `RLPDecodeError` has an `InvalidLength` variant which reads more naturally here than `Custom("invalid trie node child reference")` — this branch is exactly a length-invariant violation (child ref is neither a 32-byte hash nor an inline node under 32 bytes).

han0110: Fixed in https://github.com/lambdaclass/ethrex/pull/6983/changes/3db918ad9a3ff6f8aae161a7d2874104c8fab073

PR #6985, opened 2026-07-10
feat(l1): make the engine-API max reorg depth configurable (--max-reorg-depth)

danceratopz:
## TLDR

We added the `enginex` simulator to the [Hive Generic dashboard](https://hive.ethpandaops.io/#/group/generic?suites=eels%2Fconsume-enginex%2Ceels%2Fconsume-engine) this week. This simulator runs multiple tests against a single client instance; for each test, the simulator FCUs to genesis and then after each payload in the test. This triggers reorgs larger than ethrex's 128-block limit. More details in the [enginex docs](https://github.com/danceratopz/execution-specs/blob/a5da2157f765ef3b64f556f40438a1dbe01ec72d/docs/running_tests/running.md#enginex). The simulator's FCU behavior was designed to be consistent with the original `engine` simulator; an alternative would be to change the simulator to only verify payloads without FCU, but I think it's cleaner as it is currently (per docs above).

Feel free to close this PR/re-implement as you see fit; whatever's easiest. Thanks!

## Motivation

EEST's `consume enginex` simulator (via hive) reuses one client per pre-allocation group across many tests, and each test starts by sending a forkchoiceUpdated with `head = genesis` to reset the shared client:

- That reset is a canonical-ancestor rewind whose depth equals the chain height left by the previous test. As soon as one test's chain grows deeper than 128 blocks (e.g. `test_genesis_hash_available[...-256_empty_blocks]` leaves the head at block 257), the reset is refused with `-38006 Too deep reorg: Reorg depth 257 exceeds the client's limit of 128`.
- Because the refused rewind leaves the head in place, every subsequent test on that client fails identically — a single deep test permanently poisons the shared client. Importing the deep chain itself is fine; only the deep canonical-ancestor rewind trips the limit.

Erigon has the same kind of cap but reads it from the `MAX_REORG_DEPTH` environment variable, which is how hive already works around it for enginex runs (ethereum/hive#1568, raising it to 512); go-ethereum is gaining the equivalent knob in ethereum/go-ethereum#35335. ethrex currently offers no knob.

## Why not just bump the constant

Unlike geth's cap (an arbitrary safety margin), ethrex's 128 is anchored to real state retention: the store keeps one in-memory trie diff-layer per block (`TrieLayerCache`, commit threshold 128) and folds older layers into the single-version on-disk trie, after which the replaced ancestor states are unrecoverable. Raising only the fork-choice constant would let a deep rewind pass the depth check and then fail the `has_state_root` reachability check, degrading the FCU into `SYNCING` and a sync toward the head that can never complete in an isolated test network. The knob has to raise the retention and the fork-choice limit **together**.

## Changes

- Derive the fork-choice limit from the store's actual retention, and make the retention configurable — a single source of truth, so the limit and the retention cannot drift apart:
  - `StoreConfig::max_reorg_depth: Option<usize>` sets the trie diff-layer commit threshold. `None` (the default) keeps the existing per-backend defaults: 128 for RocksDB, 10000 for the in-memory backend.
  - `Store::max_reorg_depth()` exposes the configured retention; `apply_fork_choice` now reads the limit from the store instead of the hardcoded `REORG_DEPTH_LIMIT` const (removed). No call-site signature changes anywhere.
- New CLI flag `--max-reorg-depth <BLOCKS>` (env `ETHREX_MAX_REORG_DEPTH`, values ≥ 1), plumbed through `StoreConfig` like the existing `--rocksdb.block-cache-size`.
- Added `reorg_depth_limit_follows_store_retention` (test/tests/blockchain/smoke_tests.rs): a 10-block chain on a store configured with `max_reorg_depth = 8` refuses the genesis reset with `TooDeepReorg { reorg_depth: 10, limit: 8 }` and leaves the head untouched (the poisoning dynamic); a rewind of exactly the configured depth succeeds end-to-end (the retained state is really served, not just depth-checked); an unconfigured store reports its backend default as the limit.

### Semantics

- Default behavior is unchanged: a persistent (RocksDB) node still refuses rewinds deeper than 128 with `-38006 Too deep reorg`.
- `--max-reorg-depth 512` retains 512 diff-layers and accepts rewinds up to 512 blocks. Each retained layer costs roughly one block's worth of trie updates in memory, so large values are intended for testing environments.
- In-memory (`--datadir memory`) nodes now report their true retention (10000) as the limit instead of the over-strict 128 — they were already physically able to serve those rewinds.
- There is no "0 = unlimited" mode (unlike geth): unlimited retention is not free in ethrex's architecture, so the limit is always explicit.
- Batch import (`ethrex import`, full sync) intentionally keeps its aggressive flush threshold; the retention setting governs live per-block execution, which is what the engine-API scenario exercises.

## Notes

- Hive-side companion (required for enginex runs to benefit): ethereum/hive#1571 sets `ETHREX_MAX_REORG_DEPTH=512` in `clients/ethrex/ethrex.sh` when the simulator requests deep reorgs (`HIVE_EXPECT_DEEP_REORGS`, set by the enginex multi-test client environment as of ethereum/execution-specs#3145), mirroring the erigon fix in ethereum/hive#1568. clap only reads env vars for args it declares, so ethrex builds without this flag silently ignore the variable — safe to set for older releases.

## Verification

Verified against EEST's `consume enginex` in hive `--dev` mode and via the standalone hive command, using the minimal failing pair: a 257-block test followed by any test on the same shared client (the second test's reset-to-genesis forkchoiceUpdated is a depth-257 canonical rewind). Note the pair differs from the one used to verify the equivalent geth fix (ethereum/go-ethereum#35335): geth's limit is 32, so a 37-block chain trips it; ethrex's 128-block limit needs the deeper `test_genesis_hash_available[...-256_empty_blocks]` chain.

Setup, in a hive checkout: build this branch on the host (`cargo build --release --locked --bin ethrex`), copy the binary to `clients/ethrex/ethrex-local/ethrex`, add a `clients/ethrex/Dockerfile.binlocal` that COPYs it into the stock ethrex client image (same layout as `clients/reth/Dockerfile.binlocal`; the image's glibc must be >= the host's), add `export ETHREX_MAX_REORG_DEPTH=512` to `clients/ethrex/ethrex.sh` (this is what ethereum/hive#1571 does, gated on the simulator opting in via `HIVE_EXPECT_DEEP_REORGS` — that PR is required for enginex runs to pick this fix up), and create `ethrex-local.yaml`:

```yaml
- client: ethrex
  nametag: maxreorg-local
  dockerfile: binlocal
  build_args:
    local_path: ethrex-local
```

Fastest: run hive in `--dev` mode and drive it with consume from an execution-specs (EEST) checkout, filling the two fixtures locally instead of downloading the release:

```bash
./hive --dev --client ethrex --client-file ethrex-local.yaml \
  --docker.output --dev.addr 127.0.0.1:3012
```

```bash
uv run fill --generate-all-formats -m blockchain_test_engine_x --output=fixtures/ --clean \
  --regex='.*fork_Osaka.*(256_empty_blocks|program_INVALID-debug).*' \
  tests/frontier/opcodes/test_blockhash.py tests/frontier/scenarios/test_scenarios.py -v

export HIVE_SIMULATOR=http://127.0.0.1:3012
uv run consume enginex --input=fixtures/ -v
```

Alternatively, run the two tests with a single standalone hive command (the first run builds the simulator image, which downloads the `tests@v20.0.0` fixture release):

```bash
./hive --sim ethereum/eels/consume-enginex \
  --sim.buildarg fixtures=tests@v20.0.0 \
  --client ethrex --client-file ethrex-local.yaml \
  --sim.limit '.*fork_Osaka-blockchain_test_engine_x-(256_empty_blocks|test_program_program_INVALID-debug).*' \
  --docker.output
```

Both fixtures land in the same pre-allocation group (`0x8d9f55744d74a1de`), so one client is reused across them, with the 256-block test running first.

Results (identical in both modes):

- Control (this branch, `ETHREX_MAX_REORG_DEPTH` unset, default 128 — identical to current main): `test_genesis_hash_available` passes leaving the head at block 257; `test_scenarios[...program_INVALID-debug]` fails its initial forkchoice update with `-38006 Too deep reorg: Reorg depth 257 exceeds the client's limit of 128`.
- With `ETHREX_MAX_REORG_DEPTH=512`: both tests pass.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — not needed: no on-disk format change; the retention knob only affects how many in-memory diff-layers are kept, and restarts already drop and re-execute the non-flushed tail.

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -740,6 +740,7 @@ pub async fn init_l1(
 
     let store_config = StoreConfig {
         rocksdb_block_cache_size: opts.rocksdb_block_cache_size,
+        max_reorg_depth: opts.max_reorg_depth,
         ..StoreConfig::default()
     };
     let store_result = if opts.skip_genesis_validation {
--- a/crates/storage/store.rs
+++ b/crates/storage/store.rs
@@ -61,8 +61,10 @@ use tracing::{debug, error, info, warn};
 /// Maximum number of execution witnesses to keep in the database
 pub const MAX_WITNESSES: u64 = 128;
 
-// We use one constant for in-memory and another for on-disk backends.
-// This is due to tests requiring state older than 128 blocks.
+// Default trie diff-layer retention per backend, overridable via
+// `StoreConfig::max_reorg_depth`. We use one constant for in-memory and
+// another for on-disk backends because tests require state older than
+// 128 blocks.
 // TODO: unify these
 #[allow(unused)]
 const DB_COMMIT_THRESHOLD: usize = 128;
@@ -99,6 +101,18 @@ pub struct StoreConfig {
     /// blocks — that is the backpressure that throttles `newPayload`.
     /// Clamped to `max(1)` at construction (0 would make a rendezvous channel).
     pub persist_channel_capacity: usize,
+    /// Maximum reorg depth the store can serve, in blocks. Sets the in-memory
+    /// trie diff-layer retention (the [`TrieLayerCache`] commit threshold):
+    /// during live per-block execution one layer holds one block's state diff,
+    /// and once a layer is folded into the single-version on-disk trie the
+    /// ancestor state it replaced can no longer be unwound. Fork choice derives
+    /// its -38006 "too deep reorg" limit from this value (see
+    /// [`Store::max_reorg_depth`]), so the limit and the retention cannot drift
+    /// apart. `None` keeps the backend default: `DB_COMMIT_THRESHOLD` (128) for
+    /// RocksDB, `IN_MEMORY_COMMIT_THRESHOLD` (10000) for in-memory. Each
+    /// retained layer costs memory (one block's trie updates), so raising this
+    /// is mainly intended for testing environments with deep reorgs.
+    pub max_reorg_depth: Option<usize>,
 }
 
 impl Default for StoreConfig {
@@ -106,6 +120,7 @@ impl Default for StoreConfig {
         Self {
             rocksdb_block_cache_size: DEFAULT_ROCKSDB_BLOCK_CACHE_SIZE_BYTES,
             persist_channel_capacity: DEFAULT_PERSIST_CHANNEL_CAPACITY,
+            max_reorg_depth: None,
         }
     }
 }
@@ -1688,8 +1703,7 @@ impl Store {
     pub fn new_with_config(
         path: impl AsRef<Path>,
         engine_type: EngineType,
-        // `config` only feeds the RocksDB backend; without that feature it is unused.
-        #[cfg_attr(not(feature = "rocksdb"), allow(unused_variables))] config: StoreConfig,
+        config: StoreConfig,
     ) -> Result<Self, StoreError> {
         let db_path = path.as_ref().to_path_buf();
 
@@ -1740,7 +1754,7 @@ impl Store {
                     return Self::from_backend(
                         backend,
                         db_path,
-                        DB_COMMIT_THRESHOLD,
+                        config.max_reorg_depth.unwrap_or(DB_COMMIT_THRESHOLD),
                         config.persist_channel_capacity,
                     );
                 }
@@ -1762,7 +1776,7 @@ impl Store {
                 Self::from_backend(
                     backend,
                     db_path,
-                    DB_COMMIT_THRESHOLD,
+                    config.max_reorg_depth.unwrap_or(DB_COMMIT_THRESHOLD),
                     config.persist_channel_capacity,
                 )
             }
@@ -1771,7 +1785,7 @@ impl Store {
                 Self::from_backend(
                     backend,
                     db_path,
-                    IN_MEMORY_COMMIT_THRESHOLD,
+                    config.max_reorg_depth.unwrap_or(IN_MEMORY_COMMIT_THRESHOLD),
                     config.persist_channel_capacity,
                 )
             }
@@ -3264,6 +3278,20 @@ impl Store {
             Some(account_hash),
         );
         Ok(Trie::open(Box::new(trie_db), storage_root))
+    }
+
+    /// Maximum reorg depth this store can serve: the number of in-memory trie
+    /// diff-layers retained (one per block during live sync) before they are
+    /// folded into the single-version on-disk trie. Ancestor states deeper than
+    /// this are unrecoverable, so fork choice rejects deeper rewinds with
+    /// -38006 (see `apply_fork_choice`). Configurable via
+    /// [`StoreConfig::max_reorg_depth`] (`--max-reorg-depth`).
+    pub fn max_reorg_depth(&self) -> Result<u64, StoreError> {
+        Ok(self
+            .trie_cache
+            .read()
+            .map_err(|_| StoreError::LockError)?
+            .commit_threshold() as u64)
     }
 
     pub fn has_state_root(&self, state_root: H256) -> Result<bool, StoreError> {
--- a/test/tests/blockchain/smoke_tests.rs
+++ b/test/tests/blockchain/smoke_tests.rs
@@ -12,7 +12,7 @@ use ethrex_common::{
     H160, H256,
     types::{Block, BlockHeader, DEFAULT_BUILDER_GAS_CEIL, ELASTICITY_MULTIPLIER},
 };
-use ethrex_storage::{EngineType, Store};
+use ethrex_storage::{EngineType, Store, StoreConfig};
 
 #[tokio::test]
 async fn test_small_to_long_reorg() {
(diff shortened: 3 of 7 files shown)


edg-l:
A config flag will not be needed once our deep reorg support from https://github.com/lambdaclass/ethrex/pull/6724 merges, as discussed on discord

crates/storage/store.rs
```diff
@@ -3266,6 +3280,20 @@ impl Store {
...
+    /// Maximum reorg depth this store can serve: the number of in-memory trie
+    /// diff-layers retained (one per block during live sync) before they are
+    /// folded into the single-version on-disk trie. Ancestor states deeper than
+    /// this are unrecoverable, so fork choice rejects deeper rewinds with
+    /// -38006 (see `apply_fork_choice`). Configurable via
+    /// [`StoreConfig::max_reorg_depth`] (`--max-reorg-depth`).
+    pub fn max_reorg_depth(&self) -> Result<u64, StoreError> {
+        Ok(self
+            .trie_cache
+            .read()
```
ElFantasma, line 3292: This reads through the trie-cache `RwLock` on every FCU, but the value is set once at construction (line 1757/1779/1788) and never mutated by any code path currently in-tree. Two options:

1. Cache the resolved value on `Store` itself as `max_reorg_depth: u64`, populated in `new_with_config` / `from_backend`. Drops the per-FCU lock read and makes the invariant explicit at the type level.
2. Keep the getter but store as `AtomicU64` (also invariant) if you want future flexibility for runtime tuning.

Option 1 is smaller and matches how `commit_threshold` is treated (immutable-at-construction). Deferrable / optional.

cmd/ethrex/initializers.rs
```diff
@@ -740,6 +740,7 @@ pub async fn init_l1(
 
     let store_config = StoreConfig {
         rocksdb_block_cache_size: opts.rocksdb_block_cache_size,
+        max_reorg_depth: opts.max_reorg_depth,
```
ElFantasma, line 743: L1 init is plumbed here; the L2 init path at `cmd/ethrex/l2/initializers.rs:223` isn't:

```rust
let store_config = StoreConfig {
    rocksdb_block_cache_size: opts.node_opts.rocksdb_block_cache_size,
    ..StoreConfig::default()   // ← max_reorg_depth defaults to None
};
```

An L2 sequencer therefore always gets the backend default (128 for RocksDB, 10000 for in-memory) regardless of what the operator passes on the L1 CLI. Two reasonable directions:

1. If the L2 sequencer shouldn't accept this knob (its FCU pattern rarely reorgs deeply), leave it as-is — but add a docstring note on `StoreConfig::max_reorg_depth` clarifying the flag only takes effect on L1.
2. If L2 should honor the flag too, pass `opts.node_opts.max_reorg_depth` at line 224 as well. One-line fix.

Even if L2 doesn't reorg deeply in production, the hive-style deep-rewind tests you cite in the long help would fail against an L2 sequencer under the current wiring.

test/tests/blockchain/smoke_tests.rs
```diff
@@ -290,10 +290,10 @@ async fn latest_block_number_should_always_be_the_canonical_head() {
 async fn unfinalized_reorg_deeper_than_32_is_allowed() {
     // Per execution-apis PR 786 point 6, -38006 TooDeepReorg fires when the reorg
     // depth exceeds the implementation-specific limit. ethrex defines that limit as
-    // its state-history retention (REORG_DEPTH_LIMIT = 128), matching the stance of
-    // Erigon/Nethermind/Besu/geth — the EL trusts the CL's fork choice and only
-    // rejects when it physically cannot unwind. A 33-block reorg from genesis is
-    // well under the cap and must succeed.
+    // its state-history retention (`Store::max_reorg_depth`, 128 by default on
+    // RocksDB), matching the stance of Erigon/Nethermind/Besu/geth — the EL trusts
+    // the CL's fork choice and only rejects when it physically cannot unwind. A
```
ElFantasma, line 295: Great catch updating this docstring for the removed constant. Same function has a second stale reference just below — line 331 (in the FCU comment `// FCU to chain B head: reorg depth = 33, well under REORG_DEPTH_LIMIT (128).`) is unchanged by the diff and still points at the removed symbol. Consider updating that line too, e.g.:

```rust
// FCU to chain B head: reorg depth = 33, well under the store's max_reorg_depth (128 on RocksDB).
```

test/tests/blockchain/smoke_tests.rs
```diff
@@ -345,6 +345,66 @@ async fn unfinalized_reorg_deeper_than_32_is_allowed() {
...
+    );
+    // The refused rewind must leave the head untouched.
+    assert_eq!(store.get_latest_block_number().await.unwrap(), 10);
+
+    // A rewind of exactly the configured depth is allowed (the check is exclusive)
+    // and the target state is still retained, so the FCU succeeds end-to-end.
+    apply_fork_choice(&store, hashes[1], H256::zero(), H256::zero())
+        .await
+        .expect("rewind at the configured limit should succeed");
+    assert_eq!(store.get_latest_block_number().await.unwrap(), 2);
```
ElFantasma, line 400: The `latest_block_number == 2` assertion proves fork-choice accepted the FCU, but doesn't verify the state at the rewound tip is actually reachable — which is the load-bearing invariant this PR is establishing ("the reorg depth check tracks what the store can physically serve"). Consider adding one more assertion:

```rust
let block_2_header = store.get_block_header_by_hash(hashes[1]).unwrap().unwrap();
assert!(store.has_state_root(block_2_header.state_root).unwrap(),
    "state at rewound tip must be reachable after the FCU");
```

Without it, a future regression that lets fork-choice accept a rewind but leaves the state unreachable (i.e. the FCU succeeds but a follow-up `eth_call` at the new head fails) would pass this test. Cheap belt-and-suspenders.

ElFantasma (commented):
Solid PR overall — four inlines below on things existing reviewers didn't cover; all non-blocking.

PR #6986, opened 2026-07-10
perf(l1): breadth-first batched trie-node prefetch in merkle storage-root path

edg-l:
## What

Adds `Trie::prefetch_sorted`: a breadth-first, level-batched pre-resolution of trie nodes, wired into the per-block merkle write path -- both the **storage-root recomputation** and the **state (account) trie update** (Stage C). Gated on write-set size so small writes are untouched. State root is byte-identical.

## Why

When a block's storage root is recomputed, each written slot calls `Trie::insert`, which walks the trie from the root and resolves one `NodeRef::Hash` child per level via a single `db.get` -- dependent reads at **queue depth 1**. On a state larger than the block cache these are cold disk reads, and under a small cache the path even re-reads shared upper nodes evicted between successive keys. This is pointer-chasing that flat-KV cannot cover, because computing the new root requires materializing the real internal trie nodes.

## How

Before the serial insert/remove loop for an account, `prefetch_sorted` walks the sorted write keys breadth-first and issues **one batched `multi_get` per trie level** for every node needed at that level. Each fetched node is decoded and installed into the in-memory node arena exactly the way lazy resolution does -- `NodeRef::Node(Arc::new(node), OnceLock::from(hash))`, preserving the stored hash. The subsequent inserts then find every node already resolved in memory and never touch disk.

Two effects stack:
1. Queue-depth-1 dependent reads become batched per-level reads (RocksDB parallelizes the SST I/O).
2. Each cold node is read once and reused from the arena, eliminating the re-reads the baseline incurs under a small block cache.

New surface:
- `TrieDB::multi_get` (default = map over `get`; overridden by the RocksDB-backed `BackendTrieDB` with per-table partition/scatter, and by `TrieWrapper` with overlay-probe-then-batched-fill). Order-preserving.
- `Trie::prefetch_sorted(&mut self, sorted_paths: &[Nibbles])`.
- Gated call sites across the merkle write path: the storage-root paths (parallel `bal_storage_worker`, `compute_sharded_storage_root` per shard, `serial_storage_root` fallback) and the Stage C state-trie update (per shard). Thresholds: 64 slots/account for storage, 4 accounts/shard for the per-nibble shard paths.

## Correctness

Pre-resolution is hash-preserving and content-neutral -- it only converts `NodeRef::Hash` to `NodeRef::Node` with the stored hash, never mutating node fields -- so `collect_changes_since_last_hash` emits an identical node set and the root is byte-identical. This is verified by:
- a trie-level equivalence test (`trie_prefetch_tests`): serial-built root == prefetch-then-serial root, and identical collected node bytes;
- an order-preservation test against a real RocksDB store across all four state column families (`multi_get_order_tests`);
- the full engine EF fixture suite on the RocksDB backend with the gate forced to fire on every storage write: **74,331 fixtures passed, 0 failed** across all forks including Amsterdam.

## Measurement

Measured with the state-bench cold-state harness (O_DIRECT, 64 MiB block cache, fresh subprocess per run, `--reset checkpoint`, coldness self-check). Workload: 1000 blocks each writing 64 brand-new storage slots into a >RAM storage trie (~135M slots), reads disabled to isolate the write path. Baseline = this branch with the prefetch removed.

| metric | before | after | change |
| --- | --- | --- | --- |
| loop_seconds (mean) | 474.06 | 306.68 | **-35.3%** |
| throughput (Ggas/s) | 0.0132 | 0.0204 | **+54.6%** |
| block_cache_miss | 712,574 | 504,646 | -29.2% |
| bytes_read | 88.0 MB | 1.86 MB | -97.9% |

Variance is well under 0.5% CoV on both sides; the 167 s gap is far outside noise. The drop in `bytes_read`/`block_cache_miss` reflects the eliminated re-reads: the baseline repeatedly re-reads nodes evicted from the small cache, while pre-resolution reads each node once and serves the rest from the arena.

### Warm regime: no regression

Re-run of the same workload with a large block cache (16 GiB) and OS page cache enabled (`--direct-reads false`), so reads are cache-resident. Fully-warmed steady state:

| metric | before | after | change |
| --- | --- | --- | --- |
| loop_seconds | 7.007 | 6.984 | -0.3% (within noise) |
| throughput (Ggas/s) | 0.894 | 0.897 | +0.3% |
| block_cache_hit | 697,037 | 424,291 | fewer cache lookups |
| bytes_read | 88.0 MB | 1.88 MB | -97.9% |

When the working set is cached the prefetch is neutral -- the extra breadth-first `multi_get` pass costs nothing measurable because the arena pre-resolution removes the serial re-descents that would otherwise re-hit the cache (424k vs 697k cache hits). So the change helps when cold and does not regress when warm.

### State trie (Stage C)

Separate fixture with a cold, larger-than-cache **account trie** (~4M accounts, ~700 MB of account-trie nodes), workload writing 64 accounts/block spread across it (storage prefetch inert here, so the delta isolates the state-trie prefetch). Baseline = no prefetch.

| metric | before | after | change |
| --- | --- | --- | --- |
| loop_seconds (mean) | 581.5 | 334.1 | **-42.5%** |
| throughput (Ggas/s) | 0.0108 | 0.0188 | **+74.1%** |
| block_cache_miss | 801,418 | 674,238 | -15.9% |
| bytes_read | 72.9 MB | 19.8 MB | -72.9% |

Same mechanism, same byte-identical guarantee; the account trie benefits as much as (here, more than) the storage tries when it is the cold bottleneck.

## Consensus

The full engine EF fixture suite passes on the RocksDB backend with **both** prefetch paths forced to fire on every merkle write: 74,331 fixtures, 0 failed, across all forks including Amsterdam.

## Scope / follow-ups

- The 64-slot / 4-account thresholds are first cuts and can be tuned.


crates/blockchain/blockchain.rs
```diff
@@ -4066,18 +4170,23 @@ fn handle_subtrie(
...
-                            )?;
-                            if value.is_zero() {
-                                trie.remove(hashed_key.as_bytes())?;
-                            } else {
-                                trie.insert(hashed_key.as_bytes().to_vec(), value.encode_to_vec())?;
+                            // Local storage: buffer for a prefetched batch flush.
+                            let buf = storage_buffer
+                                .entry(prefix)
+                                .or_insert_with(|| (storage_root, Vec::new()));
+                            buf.1.push((hashed_key, value));
```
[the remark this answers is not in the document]
edg-l: Not a bug. `get_or_open_storage_trie` opens each prefix trie once and reuses it — the eager per-slot path does the same (the Occupied arm ignores the passed root). During routing an account's storage_root is the parent block root and is constant; it only changes at StorageShard finalization, which runs after the flush. On removed_storage/DeleteStorage the buffer is cleared alongside the trie reset. All 3018 Amsterdam blockchain tests pass, including duplicate-slot and storage-reset cases.

crates/common/trie/trie.rs
```diff
@@ -691,6 +779,188 @@ impl Trie {
...
+}
+
+/// Mutable counterpart to [`collect_prefetch_targets`] used by
+/// [`Trie::prefetch_sorted`]: repeats the identical descent/routing so it
+/// visits `NodeRef::Hash` children in the exact same order, installing the
+/// corresponding (same-order) `multi_get` result into the arena
+/// hash-preservingly, exactly as `NodeRef::get_node_mut` installs a single
+/// resolved node. On a `None`/empty result the ref is left untouched (absent
+/// path); the subsequent `insert` will create it.
+fn install_prefetch_targets(
```
ElFantasma, line 903: Correctness of this pair depends on `collect_prefetch_targets` (line 789) and `install_prefetch_targets` visiting `NodeRef::Hash` children in *exactly* the same order — otherwise `results.next()` binds to the wrong slot. Structurally identical today (branch iterates buckets 0..15, extension matches prefix, recurse on `Node`, emit/consume on `Hash`), but they're 60+ line siblings that must stay in lockstep across future edits. A `#[cfg(debug_assertions)]` guard that has `install` record its per-`Hash` descent path and asserts it matches what `collect` produced would catch silent drift without runtime cost in release. Optional, resilience investment.

ElFantasma (approved):

crates/common/trie/trie.rs
```diff
@@ -731,6 +819,188 @@ impl Trie {
...
+    node: &mut Node,
+    path: &Nibbles,
+    key_indices: &[usize],
+    sorted_paths: &[Nibbles],
+    results: &mut std::vec::IntoIter<Result<Option<Vec<u8>>, TrieError>>,
```
ElFantasma: `no_std build check (riscv64im_zicclsm)` is failing on `f8f433f4a` ([run](https://github.com/lambdaclass/ethrex/actions/runs/30428632843/job/90500543652)), and all five errors are in this file's prefetch additions. `ethrex-trie` has to compile for the zkVM target — `pr_nostd.yaml` says in so many words that the check exists to keep trie code no_std-clean.

Three are `E0433: cannot find module or crate 'std'`:

- L908 (this line) `std::vec::IntoIter` → `alloc::vec::IntoIter`
- L839 and L913 `std::array::from_fn` → `core::array::from_fn`

The other two are `E0277: the size for values of type '[u8]' cannot be known at compilation time` at L938 and L981, both on `bytes.filter(|b| !b.is_empty())` and pointing into `Option::filter`. Those two look like they could be fallout from `Vec` failing to resolve once the `std::` paths above error out, so it's probably worth re-running the check after the three renames before treating them as separate problems.

All five lines are added by this branch (diffed against the merge-base), so the fix belongs here rather than upstream.

ElFantasma (commented):
Heads-up: the latest commit turned the no_std check red, and all five errors are on lines this branch adds (inline).

crates/blockchain/blockchain.rs
```diff
@@ -133,6 +133,28 @@ use ethrex_common::types::BlobsBundle;
...
+// ~1/16 of them. The prefetch gate is therefore scaled down from
+// `STORAGE_PREFETCH_THRESHOLD` by 16 (same reasoning as
+// `STATE_SHARD_PREFETCH_THRESHOLD`): a trie with ~64 total touched slots then
+// clears the gate on each of its shards. Used only on the BAL-less / pre-Amsterdam
+// streaming path.
+const STREAM_STORAGE_PREFETCH_THRESHOLD: usize = 4;
+// Cap on the per-worker, per-trie slot buffer before an opportunistic mid-routing
+// flush (one sorted `prefetch_sorted` + the inserts). Bounds transient memory for
+// very heavy single-account blocks; smaller buffers just flush once at RoutingDone.
+const STREAM_STORAGE_FLUSH_BATCH: usize = 1024;
```
iovoid, line 159: 1024 per worker is 16k updates total, which seems a bit too much. have you tried other values?

ilitteri (approved):

ElFantasma (approved):
Approving. The no_std fix reads correct — `alloc::vec::IntoIter` and `core::array::from_fn` are the right paths for the guest build. 

Leaving my trie.rs:903 comment open rather than blocking on it: I re-checked at the current head and `collect_prefetch_targets` / `install_prefetch_targets` are still in exact lockstep, so nothing is wrong as written. Still worth a `debug_assert` when convenient — if they ever drift, `results.next()` binds a fetched node to the wrong slot and you get a wrong state root with nothing panicking. 4845140ea0 editing `install_` on its own was order-neutral, but it's the shape of edit that would do it. 

iovoid's question about the 1024-per-worker batch (16k updates in flight) also looks unanswered.

ilitteri (approved):

Issue #6623, opened 2026-05-12
Tracking: mempool-hardening


ilitteri:
## Context

This umbrella tracks the work to harden ethrex's mempool, closing a set of admission, lifecycle, and operator-surface gaps identified by a cross-client comparison (geth / reth / nethermind / erigon / besu).

The PRs below were validated against an external cross-client audit and reviewed.

**Status (2026-07-31):** 7 merged — #6599, #6600, #6603, #6606, #6609, #6889 (via #6892), #6871. 11 open. Ownership of the remaining tranche moved from @ilitteri to @ElFantasma.

### Admission gates

- [x] #6599 — per-transaction wire-size cap (128 KiB / 1 MiB blob wire-wrapper)
- [x] #6600 — EIP-3607 contract-sender rejection with EIP-7702 exception
- [ ] #6601 — RBF percentage fee bump 10% / 100% with type-change rejection — *review threads addressed, mergeable*
- [ ] #6624 — reject blob-tx replacements that shrink the blob count
- [x] #6603 — per-sender queued (future/nonce-gapped) tx cap (default 64), atomic with insertion; executable txs uncapped (geth AccountQueue model)
- [ ] #6630 — cap EIP-7702-delegated EOAs at 1 in-flight tx — ⚠️ **needs design rework, see below**
- [ ] #6604 — minimum priority-fee floor on raw `gas_tip_cap`
- [x] #6606 — cumulative balance check across sender's pending txs
- [x] #6609 — reject gapped-nonce txs when pool ≥ 90% full
- [x] #6889 — align intrinsic-gas validation with execution (EIP-7702 auth-list + EIP-7623 floor) — merged via #6892

### Lifecycle and eviction

- [ ] #6607 — min-heap eviction keyed by `gas_tip_cap` with lazy deletion — ⚠️ **needs design rework, see below**
- [ ] #6625 — `punishSpammer`: drop highest-nonce half on per-sender cap breach — ⚠️ **obsoleted by the #6603 re-scope, see below**
- [ ] #6610 — periodic TTL + nonce-gap dormancy sweep tasks
- [ ] #6631 — reorg re-injection for mempool transactions and blob sidecars

### Origin and propagation

- [ ] #6576 — `--mempool.private`: non-propagating local txs
- [ ] #6608 — `TxOrigin::{Local, External}` + `--mempool.nolocals`

### RPC fidelity

- [x] #6871 — clamp `eth_getTransactionCount` pending nonce to latest

[...]

PR #6989, opened 2026-07-13
fix(l1): re-check per-sender mempool admission gates atomically under the insertion lock

### LINKED ISSUE #6938: mempool: TOCTOU between per-sender admission gates and insertion
## Summary

The per-sender mempool admission gates added in the mempool-hardening umbrella (#6623) share a TOCTOU race: `Blockchain::validate_transaction` reads mempool/sender state under a **read lock**, then the actual insertion (`add_transaction` / `add_transaction_no_broadcast`) happens later under a **write lock**. Two concurrent submissions from the same sender can both pass validation against the same stale snapshot, then both insert — exceeding the per-sender budget/cap the gate is meant to enforce.

This affects the gates whose purpose is precisely per-sender rate-limiting, so the adversarial "parallel submissions from one sender" case is exactly their threat model:

- **#6606** — cumulative balance check across a sender's pending txs
- **#6603** — per-sender pending-tx cap
- **#6609** — gapped-nonce rejection under pressure
- **#6600** — EIP-3607 contract-sender rejection (sender becomes a contract mid-flight)

## Concrete scenario (cumulative balance, #6606)

Sender balance = 10, existing pending cost = 5. Two parallel cost-3 submissions:
- Task A: `sum_cost_for_sender` → 5, `+3 = 8 ≤ 10` → pass → insert.
- Task B (concurrent): sees 5 (A hasn't inserted yet), `+3 = 8 ≤ 10` → pass → insert.
- Final: 5 + 3 + 3 = 11 > 10 — cumulative budget exceeded.

## Severity

Low / hardening. The window is small, only adversarial parallel same-sender submission exploits it, and it is not a consensus or fund-loss bug (block execution re-validates each tx at inclusion). But it does weaken the per-sender gates against the exact case they target, so it's worth closing.

## Proposed fix (pick one; do it once at the validate→insert boundary)

1. **Per-sender admission mutex** — serialize validate+insert per sender. Cheap; only contended for an adversarial sender hammering one address.
2. **Re-validate under the write lock** — move the same-sender-dependent checks into the write-lock-holding `add_transaction` path (re-run `sum_cost_for_sender` etc. after acquiring the write lock).

A single boundary-level fix closes it for all the gates above; the individual PRs then drop their TOCTOU caveats.

## References

Flagged in review on #6600, #6606, and #6609.

ElFantasma:
**Motivation**

The per-sender mempool admission gates added under umbrella #6623 share a TOCTOU race: `Blockchain::validate_transaction` reads sender/pool state under a **read lock**, but the insertion in `add_transaction` happens later under a **write lock**. Two concurrent submissions from the same sender can both pass validation against the same stale snapshot, then both insert — exceeding the per-sender budget each gate is meant to enforce.

Example (cumulative balance): balance 10, existing pending cost 5, two parallel cost-3 submissions each see `5 + 3 = 8 ≤ 10`, both insert → final 11 > 10.

**Description**

Generalizes the existing atomic `QueuedCap` pattern into a single `SenderAdmission` guard so all mempool-state per-sender gates are re-checked **atomically under the insertion write lock**:

- **`SenderAdmission`** carries the precomputed inputs — `account_nonce`, `queued_max` (#6603), `gap_threshold` (#6609), `is_replacement`, and `balance_check { tx_cost, sender_balance }` (#6606, `None` for frame txs).
- **`MempoolInner::check_admission`** runs the queued-cap, gapped-nonce, and cumulative-balance gates under the write lock in `add_transaction`, before any mutation, so each gate reads the live pool state the insert will mutate.
- `occupancy_pct` / `sum_cost_for_sender` now delegate to lock-free `_inner` cores that `check_admission` reuses while holding the write lock. The unlocked checks in `validate_transaction` remain as a cheap **pre-filter** (mirrors the existing frame-tx gate's "pre-filter + authoritative locked re-check"); the guard is built there and returned so both call sites just forward it.

Behavior is preserved exactly: replacements bypass the queued-cap and gap gates; frame txs are not balance/gap-gated on nonce/balance; non-frame replacements are removed before insertion so the locked cumulative sum already excludes them.

**Scope note:** #6600 (EIP-3607 "sender becomes a contract mid-flight") is intentionally **not** covered — its race is a change to the sender's account code in storage, not to mempool state, so the mempool write lock cannot close it. It is a distinct, lower-severity account-state race.

This is the reusable seam later per-sender gates (#6625, #6630, #6608) plug into by adding a field, rather than each re-introducing the race.

Tests: 5 new inline unit tests exercise the locked path directly (cumulative reject/accept, gap reject/accept, replacement bypass); the existing behavioral suites (cumulative-balance, gap-admission, replacement-bypass) continue to pass through the new atomic path.

Resolves #6938

**Checklist**

- [x] `cargo fmt` + `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `ethrex-blockchain` mempool unit tests (17) + `ethrex-test` mempool suite (68) pass


crates/blockchain/mempool.rs
```diff
@@ -305,6 +333,110 @@ impl MempoolInner {
...
+    /// pool, and any replaced tx has already been removed by the caller.
+    fn check_admission(
+        &self,
+        sender: Address,
+        tx_nonce: u64,
+        admission: &SenderAdmission,
+    ) -> Result<(), MempoolError> {
+        // The queued cap and gap gate exempt replacements: a same-`(sender,
+        // nonce)` replace neither grows the queue nor introduces a nonce gap.
+        if !admission.is_replacement {
```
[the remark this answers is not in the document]
ElFantasma: Fixed in 1f6a01990. `is_replacement` was a precomputed value; it's now derived **live under the write lock** in `check_admission` (`txs_by_sender_nonce.get(&(sender, tx_nonce))`), and the replaced tx is removed inside `add_transaction` under that same lock (mirroring the frame-tx path) instead of by the caller beforehand. So replacement detection, the queued-cap/gap gates, the excluded-cost sum, the removal, and the insert are now one atomic scope — a concurrent removal in the validate→insert window can no longer let a now-fresh gapped/future tx skip the gates.

ElFantasma:
Thanks all — Codex, Claude, Greptile, and Kimi converged on the same thing, addressed now.

**Replacement atomicity (Codex; Claude #1 High; Greptile P1).** All three flagged that removing the replaced tx in `blockchain.rs` *before* `add_transaction` takes the write lock left a gap: a concurrent same-sender op between `validate_transaction` and the locked insert could make the locked re-check (or the live gates) reject a replacement whose predecessor was already removed — losing the tx / skipping gates. Fixed in `1f6a01990` (pushed after these reviews, which were all against `6abfd199d`):

- The same-nonce removal now happens **inside `add_transaction`, under the write lock, only after `check_admission` passes** — exactly mirroring the frame-tx path. Callers no longer pre-remove.
- Replacement-ness is derived **live** from `txs_by_sender_nonce` under that lock (no precomputed `is_replacement`), and the predecessor's hash is the `exclude` for the cumulative sum.
- So replacement detection, all three gates, the removal, and the insert are one atomic scope. A rejected fee-bump can no longer leave the sender with neither tx.

**Regression test (Codex + Claude both asked).** Added `rejected_balance_replacement_keeps_original` (`4fc8db7bf`): a replacement that fails the locked cumulative-balance re-check leaves the original same-nonce tx in the pool — the non-frame analogue of the frame `fee_bump_rejected_leaves_original_intact` test, and it also exercises the `is_replacement` + `balance_check` interaction Claude noted was uncovered.

**Gap check clarity (Kimi).** `check_admission` now uses `tx_nonce > account_nonce` instead of `!=` — equivalent here (`nonce < account_nonce` is already `NonceTooLow`) but states the forward-jump intent.

**Doc (Claude #2.3).** The `check_admission` doc no longer claims "removed by the caller"; it now describes the removal happening after the check under the same lock.

All green after the changes: fmt + clippy `-D warnings` clean, 18 inline + 68 behavioral mempool tests pass.

crates/blockchain/blockchain.rs
```diff
@@ -2904,22 +2904,12 @@ impl Blockchain {
...
         let sender = transaction.sender(&NativeCrypto)?;
 
-        // Validate transaction
-        let (tx_to_replace, frame_reservation, sender_account_nonce) =
+        // Validate transaction. The returned `sender_admission` carries the
+        // per-sender gate inputs, re-checked atomically inside `add_transaction`,
+        // which also removes any same-nonce tx being replaced under the same lock
+        // (#6938) — so no separate pre-removal here.
+        let (_tx_to_replace, frame_reservation, sender_admission) =
             self.validate_transaction(&transaction, sender).await?;
```
iovoid: validate_transaction's first return value is no longer used anywhere

ElFantasma: Good catch — removed. Since the same-nonce removal now happens live inside `add_transaction` (not by the caller), the returned `tx_to_replace` hash was dead: both call sites bound it to `_`. `validate_transaction` now returns `(Option<FramePaymasterReservation>, Option<SenderAdmission>)`; it still computes the replace-hash internally for the RBF/cumulative pre-filters, just doesn't surface it. Done in 648df1940 (rebased onto the latest branch).

iovoid (approved):

ilitteri (approved):

MegaRedHand (approved):

Issue #5509, opened 2025-12-03
OpenVM toolchain installation is failing


ilitteri:
**Context**

https://github.com/lambdaclass/ethrex/actions/runs/19897838369/job/57042034872?pr=5141

**Description**

Fix and re-enable OpenVM linter and guest release once fixed.

PR #6990, opened 2026-07-13
chore(l1,l2): bump c-kzg to the latest version

iovoid:
**Motivation**

There's a new version of c-kzg.

**Description**

Bumps the version.

### DIFF
--- a/.github/workflows/pr-main_l2_prover.yaml
+++ b/.github/workflows/pr-main_l2_prover.yaml
@@ -46,7 +46,10 @@ jobs:
     strategy:
       matrix:
         # Note: openvm backend linting is disabled due to a problem in the installation https://github.com/lambdaclass/ethrex/issues/5509
-        backend: ["sp1", "risc0", "zisk"]
+        # Note: risc0 backend linting is temporarily disabled: the c-kzg 2.1.8 floor exceeds the
+        # highest risc0 c-kzg fork tag (v2.1.7-risczero.0), so the guest workspace can't resolve.
+        # Re-add "risc0" once risc0/c-kzg-4844 publishes a >=2.1.8 tag.
+        backend: ["sp1", "zisk"]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v6
--- a/.github/workflows/tag_release.yaml
+++ b/.github/workflows/tag_release.yaml
@@ -52,12 +52,15 @@ jobs:
             os: macos
             stack: l2
             features: l2,l2-sql
+          # risc0 temporarily dropped from features: the c-kzg 2.1.8 floor exceeds the highest
+          # risc0 c-kzg fork tag (v2.1.7-risczero.0), so the risc0 guest can't build. Re-add
+          # ",risc0" once risc0/c-kzg-4844 publishes a >=2.1.8 tag.
           - platform: ubuntu-22.04
             stack: l2
-            features: l2,l2-sql,sp1,risc0
+            features: l2,l2-sql,sp1
           - platform: ubuntu-22.04
             stack: l2_gpu
-            features: l2,l2-sql,sp1,risc0,gpu
+            features: l2,l2-sql,sp1,gpu
           - platform: ubuntu-22.04-arm
             stack: l2
             features: l2,l2-sql,sp1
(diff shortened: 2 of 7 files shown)


edg-l (approved):

ilitteri (approved):

crates/common/crypto/Cargo.toml
```diff
@@ -13,7 +13,7 @@ path = "./lib.rs"
 
 [dependencies]
 # TODO(#1102): Move to Lambdaworks in the future
-c-kzg = { version = "2.1.1", default-features = false, optional = true }
+c-kzg = { version = "2.1.8", default-features = false, optional = true }
```
ElFantasma, line 16: Endorsing greptile's P1 with concrete cost/benefit framing: c-kzg 2.1.8 is a minor release whose notable changes are Zig bindings + minor stuff (see [v2.1.8 release notes](https://github.com/ethereum/c-kzg-4844/releases/tag/v2.1.8)) — nothing security-critical, nothing that a Rust consumer of the bindings actually needs. Meanwhile the bump forces disabling the risc0 backend across CI + release artifacts because the `[patch.crates-io]` in `crates/guest-program/bin/risc0/Cargo.toml:40` pins to `c-kzg/v2.1.1-risczero.0`, and the highest published risc0-fork tag is `v2.1.7-risczero.0`.

Cost of the bump = risc0 backend is silently broken for anyone building `cargo build --features risc0` from source; release archives lose the risc0 vk/ELF (see the two inlines on `tag_release.yaml`); and no tracking issue is linked for restoration.

Options worth considering (not blocking, author's call):

1. **Revert to c-kzg 2.1.7** until `risc0/c-kzg-4844` publishes `>=2.1.8`. Preserves risc0. Costs nothing observable — we don't gain any Rust-consumer functionality from 2.1.8.
2. **Keep the bump but hard-gate the `risc0` feature**: add a `compile_error!` under `#[cfg(feature = "risc0")]` in this crate pointing to the same tracking issue, so `cargo build --features risc0` fails loudly with a clear message instead of the current "cargo can't resolve" symptom.
3. **Deprecate risc0 entirely**: remove the `risc0` feature, `crates/guest-program/bin/risc0/`, `ethrex-l2/prover` risc0 gates, and downstream docs. Silent-broken code paths are worse than removed ones.

Whichever direction: at minimum, please add a tracking issue ("Re-enable risc0 backend after c-kzg fork catches up to 2.1.8+") and link it in the workflow comments, so "temporarily disabled" doesn't become permanent by accident.

.github/workflows/tag_release.yaml
```diff
@@ -127,7 +130,7 @@ jobs:
         if: ${{ matrix.platform == 'ubuntu-22.04' && matrix.stack == 'l2_gpu' }} # Run only once
         run: |
           mkdir -p ./verification_keys
-          mv crates/guest-program/bin/risc0/out/riscv32im-risc0-vk verification_keys/ethrex-riscv32im-risc0-vk
+          # risc0 vk temporarily not produced (risc0 disabled; see build-ethrex features note).
```
ElFantasma, line 133: Endorsing greptile's P1 on the archive contents change: `ethrex-contracts.tar.gz` (the release artifact this vk copy feeds) no longer contains `ethrex-riscv32im-risc0-vk`. Downstream automation that initializes contracts from the release archive and expects that file will break silently on the next tagged release.

Minimum ask: mention this drop explicitly in the changelog / release notes so downstream consumers can adjust. Better if there's a migration path noted (e.g., "pin previous release tag until risc0 is restored").

.github/workflows/pr-main_l2_prover.yaml
```diff
@@ -46,7 +46,10 @@ jobs:
     strategy:
       matrix:
         # Note: openvm backend linting is disabled due to a problem in the installation https://github.com/lambdaclass/ethrex/issues/5509
-        backend: ["sp1", "risc0", "zisk"]
+        # Note: risc0 backend linting is temporarily disabled: the c-kzg 2.1.8 floor exceeds the
+        # highest risc0 c-kzg fork tag (v2.1.7-risczero.0), so the guest workspace can't resolve.
+        # Re-add "risc0" once risc0/c-kzg-4844 publishes a >=2.1.8 tag.
+        backend: ["sp1", "zisk"]
```
ElFantasma, line 52: The `# Re-add "risc0" once ...` comment carries the intent, but there's no tracking issue linked. In practice "temporarily disabled" workflows that lack a follow-up ticket tend to become permanent — see the openvm comment on the line above (referencing #5509, which has been in that state for a while). Linking a fresh tracking issue here (e.g. `#XXXX`) makes the temporary vs permanent status observable and prevents the risc0 backend from silently rotting.

ElFantasma (approved):
LGTM, but left some follow-up notes

PR #6995, opened 2026-07-14
fix(l1): close EIP-7928 supplied-BAL validation gaps

edg-l:
Closes supplied-BAL validation gaps on the parallel block-import path (`validate_tx_execution` + `execute_block_parallel` in `crates/vm/backends/levm/mod.rs`), where a non-canonical BAL that the sequential/witness path rejects was being accepted.

## Gaps closed

- **No-op entries (post == start-of-tx value):** reject a balance/nonce/code/storage change whose advertised post-value equals the value at the start of the tx. Non-canonical; a spec-conformant recorder filters these out. Covers both the arm where execution touched the account/slot **and** the absent (`None`) arm — a no-op entry for an account/slot execution never touched leaves the state root unchanged, so nothing downstream catches it; the parallel path must reject it to match the sequential path.
- **Missing storage write:** reject a slot execution wrote that's absent from both `storage_changes` and `storage_reads`. Previously silently skipped, so it was invisible to validation and the BAL-derived state root.
- **Write mis-declared as a read:** reject a slot execution genuinely wrote (value diverges from start-of-tx) that the BAL declares under `storage_reads` instead of `storage_changes`. The shadow-reads cross-check only verifies a slot is present in *some* list, never that a changed value landed in the correct one, so the value comparison now runs regardless of which list declared the slot.
- **Selfdestruct phantom reads:** `execute_block_parallel` no longer blanket-clears a selfdestructed account's `storage_reads`. A phantom (never-actually-read) declared read on a destroyed account now fails validation; a genuinely-read-then-destroyed slot is still accepted (reconciled via the shadow recorder's observed reads).
- **CREATE `code_changes`:** adds a regression test for a related gap (missing CREATE `code_changes` entry) already fixed on main in 80b5951347f5.

Went with surgical per-gap checks that reuse already-computed per-tx recorder data instead of a full canonical-BAL rebuild + RLP byte-equality, to keep the `bal_parallel_exec` fast path intact. The full-rebuild guarantee already exists on the sequential/witness path.

## Performance

The no-op checks add a start-of-tx ("seeded") lookup per validated BAL entry. Balance/nonce resolve in-memory against the pre-state snapshot, but storage would otherwise hit the store once per storage-change entry (the snapshot clones account info without storage). Two changes keep the parallel path lean:

- **Pure-read skip:** the execution→BAL check skips genuinely-read slots (they cannot diverge from their pre-tx value), computed as observed reads minus written slots.
- **Seed from captured pre-values:** `seeded_storage` consults the recorder's `tx_initial_storage` first — the start-of-tx value the EVM already captured during execution (`capture_pre_storage`) — turning the per-storage-change store read into an O(1) hashmap hit. First-write-wins + `restore()` never touching that map means the value is correct even across reverted writes (bound by recorder revert-safety tests).
- Also swapped a linear `storage_reads.contains()` scan in the Group B read check for a binary search (the list is validated strictly-ascending).

## Testing
- Unit tests (`test/tests/blockchain/bal_validate_tx_execution_tests.rs`) calling `validate_tx_execution` directly: no-op rejection for all four dimensions in both the present and absent arms, missing-write / mis-declared-write rejection, read-only acceptance, and storage-seed fast-path (proven by an erroring store + empty snapshot so only `tx_initial` can satisfy the lookup).
- e2e content-validation tests (`bal_content_validation_tests.rs`, + `fixtures/genesis/l1-bal-content.json`), each asserting the parallel path rejects a mutated BAL and accepts the unmutated control — including a real-pipeline write-mis-declared-as-read case.
- Recorder revert-safety unit tests (`crates/common/types/block_access_list.rs`).
- Full ef-test engine gate (`make -C tooling/ef_tests/engine test`, 10921 fixtures / 74462 total) passes. fmt/clippy clean.

These are consensus-behavior changes on the future Amsterdam fork only, not mainnet-active.


ilitteri:
The no-op rejection only fires in the `Some(...)` arm of `validate_tx_execution` (account/slot present in execution state). The `None`/absent arm still accepts a spurious no-op entry: the balance/nonce/code branches only error when `expected != seeded`, and the storage branch's `actual.is_none() || actual_value.is_none()` path does `if expected_value == seeded { continue; // Extraneous entry }` — accepting exactly the no-op case the `Some` arm now rejects. The `// the state root will catch any true discrepancy` rationale doesn't hold here: a post==pre entry leaves the state root unchanged, so nothing catches it.

So a supplied BAL carrying a no-op entry for an account execution never touches is accepted on the parallel path while the sequential path rejects the same block. Repro against this branch (extends the `bal_hash_parallel_skip` harness):

- build a valid Amsterdam block + its canonical BAL
- append an `AccountChanges` for an untouched address with `balance_changes: [BalanceChange::new(0, U256::zero())]`
- set `header.block_access_list_hash = spurious_bal.compute_hash(&NativeCrypto)`
- import with `bal_parallel_exec_enabled` toggled

```
parallel   = Ok(None)                                        // accepted
sequential = Err(InvalidBlock(BlockAccessListHashMismatch))  // rejected
```

This is the "or seeded pre-state when absent" case from the original finding, so it leaves `bal-spurious-account-changes` and `bal-spurious-storage-change` partially open. Suggestion: mirror the `Some`-arm no-op check in the `None`/absent arms for all four dimensions — when an entry resolves to a no-op (`expected == seeded`) for an account/slot not in execution state, reject it rather than accepting it as extraneous. A `None`-arm variant of each new no-op test would lock it in.

ElFantasma (approved):

ilitteri:
Thanks — the `None`-arm fix closes what I flagged; `parallel_path_rejects_noop_balance_entry` confirms an absent-from-execution account with a no-op change is now rejected.

A sibling path in the same family is still open, though. The no-op/absent rejection lives in PART A, which is only reached for accounts at a `bal_idx` that a real tx references. An extraneous account carrying a change entry at a `bal_idx` **no validated tx touches** is never checked — the per-tx loop skips it, and since it has a change it isn't "pure", so the leftover `unaccessed_pure_accounts` check misses it too. With `produced_bal = None` nothing rebuilds, so the parallel path accepts it while the sequential path rejects.

Repro on the current head (extends `bal_hash_parallel_skip`):
- build a valid **empty** Amsterdam block + its canonical BAL
- append an `AccountChanges` for an untouched address with `balance_changes: [BalanceChange::new(0, U256::zero())]`
- set `header.block_access_list_hash = spurious_bal.compute_hash(&NativeCrypto)`
- import with `bal_parallel_exec_enabled` toggled

```
parallel   = Ok(_)                                           // accepted
sequential = Err(InvalidBlock(BlockAccessListHashMismatch))  // rejected
```

An empty block is the simplest trigger, but the same holds for any block where the extraneous account's `bal_idx` falls outside the validated tx set.

Suggestion: after PART A/B, assert every supplied-BAL account with changes was actually covered by a validated tx — reject any whose change entries were never validated (or extend the leftover check to flag non-pure, never-touched accounts, not just pure ones). A no-tx-index variant of the None-arm test would lock it in.

ilitteri (approved):

iovoid (approved):

PR #6996, opened 2026-07-14
fix(l1): evict deterministically-invalid transactions during block building

ilitteri:
**Motivation**

A regular transaction that fails to apply while building a payload is kept in the mempool, on the assumption that the failure is transient (nonce ordering, balance, fees) and a later block will resolve it. That assumption is wrong for a transaction that fails an *intrinsic* validity check: gas limit below the minimum intrinsic cost or below the EIP-7623 calldata floor, initcode over the size cap, or gas limit over the per-transaction cap. Such a transaction keeps failing at the same nonce, so keeping it pooled lets it re-occupy the head of its sender's queue on every build and starve that sender's other transactions indefinitely. Both payload builders `pop()` the whole sender queue on failure, so the stall is per-sender, not per-transaction. The symptom is long stretches of empty blocks broken by a sudden burst once the stuck head transaction is finally displaced.

**Description**

Adds `is_deterministic_invalid` and evicts matching transactions from the pool during block building, mirroring how deterministic frame-transaction failures are already handled. Four error classes are matched:

| Class | Why it can reach the builder despite mempool admission |
|---|---|
| `IntrinsicGasTooLow` | Admission mirrors LEVM's `gas_limit >= max(intrinsic, calldata_floor)`, but has no counterpart for the Amsterdam EIP-8037 cap on `regular.max(floor)` — that one is unscreenable at admission and fires with no fork boundary involved. |
| `IntrinsicGasBelowFloorGasCost` | Admission covers it against the fork active at insertion; a later fork boundary can invalidate an already-pooled transaction. |
| `InitcodeSizeExceeded` | Same: the cap is fork-dependent. |
| `TxMaxGasLimitExceeded` | Admission's cap check is gated `is_osaka_activated && !is_amsterdam_activated`. On L1 a transaction admitted before Osaka then fails every build. On L2 the hook applies the cap from **Prague** onward, ahead of that gate, so a fee-token transaction over the cap is admitted and then rejected on every build with no fork boundary at all. |

Classification is by LEVM `Display` substring because the typed variant does not survive the trip to `ChainError` — `impl From<VMError> for EvmError` collapses it into `EvmError::Transaction(String)`. A test pins every substring through the real `TxValidationError -> VMError -> EvmError -> ChainError` path, so a reworded error breaks the test rather than silently disabling eviction.

Substring matching only works if each variant has a single determinism story, and `IntrinsicGasTooLow` did not: the L2 hook's `reserve_l1_gas` reused it for a shortfall against the reserved `l1_gas`, which tracks the L1 fee config and the block's gas price and is therefore transient. This PR splits that off into `TxValidationError::L1GasReservationTooLow`. Every remaining producer of `IntrinsicGasTooLow` is deterministic given the transaction's bytes and the active fork, so the predicate needs no chain-type argument and the L2 case is fixed rather than excluded.

Also:

- The **L2 block producer** had the same starvation bug — its `Err` arm popped the sender queue without touching the pool. It now evicts through the same predicate.
- Evictions are **logged**; previously the `Err` arm was the only silent drop path in that loop.
- The doc comment records which variants mempool admission already rejects (and why the obvious-looking omissions — `SenderNotEOA`, the type-3 structural faults, `nonce == u64::MAX` — are either transient or unreachable), plus the rule that every arm must be deterministic at *every* producing site.

**Determinism is bounded by the active fork, not absolute.** A fork can relax the very limit that failed: Amsterdam raises the initcode cap, lowers the intrinsic base cost, and removes the per-transaction gas cap. A transaction evicted just before such a fork would have become includable just after. The window is one fork boundary and the sender can resubmit, which is a better trade than starving that sender's queue on every build until then. The doc comment states this bound explicitly rather than claiming the failures are permanent.

**How to Test**

```
cargo test -p ethrex-blockchain --lib payload::
```

Three tests: the nonce-mismatch pin, the deterministic-invalid pin across all four classes, and `l1_gas_reservation_too_low_is_not_deterministic`, which asserts the transient L2 variant is *not* evicted, the permanent one still is, and the two do not share a `Display` string.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A (no storage/schema change)


crates/blockchain/payload.rs
```diff
@@ -1019,6 +1034,25 @@ fn is_nonce_mismatch(e: &ChainError) -> bool {
...
+/// below the minimum intrinsic cost or below the EIP-7623 calldata floor) and
+/// the EIP-3860/7907 initcode size cap. There is no typed variant at the
+/// `ChainError` level, so it is detected by the stable Display substrings; the
+/// `deterministic_invalid_detected_from_chain_error` test pins them through the
+/// real conversion path so a reworded error breaks the test, not eviction.
+///
+/// Such a tx must be evicted rather than kept pooled: otherwise it re-occupies
+/// its sender's queue head on every payload build and starves that sender's
+/// other transactions (a payload-inclusion stall).
+fn is_deterministic_invalid(e: &ChainError) -> bool {
```
ElFantasma: nit: the three-error set is narrower than the broader "deterministically invalid at this nonce" class the doc comment implies. Off the top: `SenderNotEOA` (EIP-7702 setcode from a non-EOA), `Type3TxContractCreation` (blob-tx create), `Type3TxZeroBlobs` / `Type3TxBlobCountExceeded`, nonce-is-max — all deterministic given the tx bytes and sender.

Fine as an initial fix — the current three are the ones actively causing head-of-queue stalls in the wild. Just worth calling out in the doc that this is a narrowly-scoped starter set rather than the whole class, so the next reader who trips over `SenderNotEOA` starving a sender's queue doesn't assume it's already covered. Optional.

crates/blockchain/payload.rs
```diff
@@ -1019,6 +1034,25 @@ fn is_nonce_mismatch(e: &ChainError) -> bool {
...
+    let msg = e.to_string();
+    msg.contains("gas limit lower than the minimum gas cost")
+        || msg.contains("gas cost floor for calldata tokens")
+        || msg.contains("Initcode size exceeded")
+}
```
ElFantasma, line 1158: nit / forward-looking: the fragility this test guards against also disappears entirely if `ChainError` (or the intermediate `VMError`/`TxValidationError`) surfaces a typed `IsDeterministicallyInvalid` marker (either a trait method or a dedicated variant class), turning the substring match into an `.is_deterministic_invalid()` call. Non-blocking; useful when this pattern grows past 3 substrings.

ElFantasma (approved):
Non-blocking observations inline.

edg-l (commented):
Ran it locally: clippy and fmt clean, both payload tests pass. Diagnosis holds, `pop()` drops the whole sender queue so a permanently failing head tx starves that sender on every build:

https://github.com/lambdaclass/ethrex/blob/b126af39066a4eddb1023cbe64933f813a521805/crates/blockchain/payload.rs#L1222-L1227

Substring matching is justified, the typed error is stringified before it reaches `ChainError`:

https://github.com/lambdaclass/ethrex/blob/b126af39066a4eddb1023cbe64933f813a521805/crates/vm/errors.rs#L24-L32

and all three strings are unique in the tree.

**The L2 block producer has the same bug.** Its `Err` arm pops the sender and never touches the pool:

https://github.com/lambdaclass/ethrex/blob/b126af39066a4eddb1023cbe64933f813a521805/crates/l2/sequencer/block_producer/payload_builder.rs#L280-L294

Same error class, same starvation. Make `is_deterministic_invalid` visible outside the module and add the same call there, or say in the description why it's split out.

**"Can never become valid at its nonce" is wrong across a fork boundary.** Amsterdam drops base cost from 21000 to 12000:

https://github.com/lambdaclass/ethrex/blob/b126af39066a4eddb1023cbe64933f813a521805/crates/vm/levm/src/utils.rs#L876-L896

and raises the initcode cap:

https://github.com/lambdaclass/ethrex/blob/b126af39066a4eddb1023cbe64933f813a521805/crates/vm/levm/src/hooks/default_hook.rs#L692-L709

Evicting is still safe, but the comment should say "for as long as the active fork's rules hold".

**The eviction is silent.** Every other drop path in that loop logs, the `Err` arm doesn't:

```rust
if evict {
    debug!("Evicting deterministically-invalid transaction {tx_hash}: {e}");
    self.remove_transaction_from_pool(&tx_hash)?;
}
```

Non-blocking: I checked the full `TxValidationError` set against the admission gate and `NonceIsMax` is the only permanently invalid variant that's neither rejected there nor evicted here. `SenderNotEOA` is correctly excluded since 7702 delegation can be revoked, and the type 3 cases are caught by `blobs_bundle.validate`.

crates/blockchain/payload.rs
MegaRedHand: This is horrible 😀

MegaRedHand (approved):

crates/blockchain/payload.rs
ilitteri: f5b65826f

crates/blockchain/payload.rs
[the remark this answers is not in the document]
ilitteri: Not applying this one. `PriorityGreaterThanMaxFeePerGas` can't reach the payload builder: mempool admission rejects it at `crates/blockchain/blockchain.rs:3409` with `MempoolError::TxTipAboveFeeCapError`, so such a tx never enters the pool and can't occupy a sender's queue head.

`f5b65826f` documents which variants admission already covers, so this doesn't get re-raised.

ilitteri: Agreed on the discomfort, but the typed marker isn't reachable from here: `impl From<VMError> for EvmError` (`crates/vm/errors.rs:24-32`) collapses the variant into `EvmError::Transaction(String)` before it gets to `ChainError`, so a typed check needs that conversion reworked first — bigger than this PR.

`f5b65826f` records that the strings are load-bearing and adds the rule that any new arm must have exactly one producer, since a variant raised from two sites with different determinism (as `IntrinsicGasTooLow` is) can't be classified by string at all.

ilitteri: Agreed on the discomfort, but the typed marker isn't reachable from here: `impl From<VMError> for EvmError` (`crates/vm/errors.rs:24-32`) collapses the variant into `EvmError::Transaction(String)` before it gets to `ChainError`, so a typed check needs that conversion reworked first — bigger than this PR.

`f5b65826f` records that the strings are load-bearing and adds the rule that any new arm must have exactly one producer, since a variant raised from two sites with different determinism (as `IntrinsicGasTooLow` is) can't be classified by string at all.

ilitteri:
Applied: `c05006d21` (L1-only `IntrinsicGasTooLow`), `0954e3e12` (`TxMaxGasLimitExceeded` + test), `739aff693` (eviction log), `d61cbd68e` (L2 block producer), `6ba07ecbc` (fork-bounded wording), `64be27f8c` (EIP-7954), `f5b65826f` (scope docs).

Two refinements worth noting. The L2 guard is narrower than suggested — only `IntrinsicGasTooLow` is chain-type gated, because `IntrinsicGasBelowFloorGasCost` and `InitcodeSizeExceeded` have a single producer each (`default_hook`) and stay deterministic on L2. A blanket L2 exclusion would have made the L2-producer fix a no-op. And `TxMaxGasLimitExceeded` needs no guard: both LEVM sites derive it from `tx.gas_limit()` and the fork alone.

Not applying `NonceIsMax`: admission already rejects `nonce == u64::MAX` in both branches — `blockchain.rs:3459` for existing accounts and `:3525` for the fresh/frame-sender path, both returning `MempoolError::NonceTooLow` — so it can't reach the pool. Happy to be corrected if there's a path I missed.

crates/blockchain/payload.rs
```diff
@@ -1019,6 +1035,69 @@ fn is_nonce_mismatch(e: &ChainError) -> bool {
...
+    let msg = e.to_string();
+    // `IntrinsicGasTooLow` is only deterministic on L1. The L2 hook raises the SAME
+    // variant from `reserve_l1_gas` when gas_limit cannot cover `floor + l1_gas`, and
+    // `l1_gas` is derived from the L1 fee config and the block's gas price — so a tx
+    // can fail it in one block purely because L1 data availability was expensive and
+    // succeed in the next. Both paths stringify identically, so the only safe reading
+    // on L2 is "transient": keep the tx pooled. The other arms below have no such
+    // second producer (only `default_hook` raises them, from the tx's own bytes), so
+    // they stay deterministic on both chain types.
+    let intrinsic_gas_too_low = matches!(chain_type, BlockchainType::L1)
```
pablodeymo: The L1-only gate is the right call given the string collision, but I think it leaves the *only* arm that fires in steady state unfixed on L2.

Admission (`validate_transaction`) already enforces `gas_limit >= max(intrinsic, calldata_floor)` (blockchain.rs:3428), the initcode cap, and the Osaka per-tx gas cap — so three of the four arms here can only fire when a fork activates between admission and build. The arm that fires *without* a fork boundary is `IntrinsicGasTooLow` from the Amsterdam EIP-8037 check in `validate_min_gas_limit` (default_hook.rs:664-670, `regular_gas.max(floor_cost_by_tokens) > TX_MAX_GAS_LIMIT_AMSTERDAM`), which `transaction_intrinsic_gas` has no counterpart for.

I checked that it's actually reachable rather than size-capped away — 5000 access-list addresses (EIP-7981 folds their bytes into both arms):

```
n=5000  encoded_len=115044 (MAX_TX_SIZE=131072)  regular=21415000  floor=6415000  cap=16777216
```

Under the 128 KiB wire cap, `regular` is ~1.28× the cap. Admission only wants `gas_limit >= max(intrinsic, floor)` and `gas_limit <= header.gas_limit`, both satisfiable on a 45M-gas chain — so the tx is admitted and then fails every build for as long as Amsterdam holds. That's exactly the head-of-queue stall this PR is about, and on L2 it stays open.

Would a distinct variant for the transient producer unblock it? Something like `TxValidationError::L1GasReservationTooLow` raised from `reserve_l1_gas` instead of reusing `IntrinsicGasTooLow`: the two stop stringifying identically, the `BlockchainType` parameter and both L2 test cases go away, and L2 gets the fix. Contained to levm, and it's the same "one producer per arm" rule this doc comment already asks future arms to satisfy.

Cheaper complement, if you'd rather not touch levm here: `transaction_intrinsic_gas` already computes both `regular` and `calldata_floor`, so mirroring the cap at admission (`fork >= Amsterdam && regular.max(floor) > TX_MAX_GAS_LIMIT_AMSTERDAM` → reject) keeps these out of the pool entirely on both chain types. That also matches that function's own stated invariant — "Mempool admission must charge the same intrinsic gas LEVM enforces at execution, or we admit txs the VM later rejects". Happy for either to be a follow-up rather than a blocker on this PR.

crates/blockchain/payload.rs
```diff
@@ -1019,6 +1035,69 @@ fn is_nonce_mismatch(e: &ChainError) -> bool {
...
+///
+/// `SenderNotEOA` is excluded on purpose for a different reason: an EIP-7702
+/// delegation can be revoked, so that failure is transient, not permanent.
+///
+/// Note for anyone extending this: the substrings are load-bearing. They are
+/// matched against `Display` output because the typed variant does not survive the
+/// trip to `ChainError` — `impl From<VMError> for EvmError` collapses it into
+/// `EvmError::Transaction(String)`. Reworded LEVM errors must therefore break the
+/// pinning test rather than silently disable eviction, which is what that test is
+/// for. Before adding an arm, check it has exactly one producer: a variant raised
```
pablodeymo: Small correction to the producer count, since this paragraph makes producer-counting the rule for anyone adding an arm: `TxMaxGasLimitExceeded` is also raised outside `default_hook`, at l2_hook.rs:596 and :607 (the fee-token path). Both are deterministic (`gas_limit > cap`, straight off the tx), so the conclusion holds — but the stated reason doesn't.

Worth spelling out, because that second producer is what makes the `"gas limit exceeds maximum"` arm load-bearing: it fires from **Prague** onward, while admission's cap check is gated `is_osaka_activated && !is_amsterdam_activated`. So on an L2 at Prague-but-not-Osaka, a fee-token tx with `gas_limit > 16_777_216` is admitted and then rejected on every build — a case this PR genuinely fixes, and the clearest justification for that arm existing.

crates/blockchain/payload.rs
```diff
@@ -754,7 +754,23 @@ impl Blockchain {
...
+                    // pooled lets it re-occupy the sender's queue head on every
+                    // build and starve that sender's other txs indefinitely.
+                    // Evict those too.
+                    let evict = if is_frame {
+                        !is_nonce_mismatch(&e)
+                    } else {
+                        is_deterministic_invalid(&e, &self.options.r#type)
+                    };
+                    if evict {
+                        debug!("Evicting deterministically-invalid transaction {tx_hash}: {e}");
```
pablodeymo: nit: the log sits outside the `is_frame` split, so a frame tx evicted for the pre-existing reason (any non-nonce-mismatch failure) also gets reported as "deterministically-invalid" — which is a different claim, and specifically *not* `is_deterministic_invalid`. Either move the `debug!` into each branch with its own wording, or make this one neutral (e.g. "Evicting transaction {tx_hash} from the pool: {e}").

pablodeymo (commented):
Diagnosis is right and the L1 fix is sound — `pop()` really does drop the whole sender queue, so a permanently-failing head tx starves that sender on every build. Verified locally: the 3 `payload::tests` pass, `cargo clippy -p ethrex-l2 --all-targets` is clean, and the branch still merges cleanly with `main` (only #7005 touched `payload.rs` since the merge-base, and its `EvmError::Custom` message doesn't collide with any of the matched substrings).

My one substantive concern is about what the L2 half actually buys — inline. Two process notes:

**PR description is out of sync** with the last four commits: the EIP-7825/8037 per-tx gas-cap arm (`0954e3e`), the L1-only gating of `IntrinsicGasTooLow` (`c05006d`), and the L2 block-producer change (`d61cbd6`) aren't mentioned. The `Description` section still reads as the original three-substring version.

**Both approvals predate the substantive commits.** ElFantasma approved 07-16, MegaRedHand 07-27 14:07Z, and edg-l's local verification was 07-27 07:32Z — but `c05006d`…`f5b6582` (chain-type awareness + L2 eviction) landed 07-27 15:57–17:10Z. Might be worth re-pinging, since the L2 path is the part I'd want a second pair of eyes on. Also flagging that Hive, `Test - ubuntu-22.04`, `Test - macos-15`, Integration, Reorg and Lint L2 were still pending when I looked.

crates/blockchain/payload.rs
ilitteri: `436b4dcd2` — went with the distinct variant.

Confirmed the reachability before touching anything: `regular` = 21,415,000 for n=5000 (12,000 base + 3,000 recipient + 5000×3000 access-list + 400,000×16 EIP-7981 data cost), floor 6,415,000, both against a 16,777,216 cap at 115,044 bytes encoded. EIP-7702 auth tuples get there too at 7816 regular gas per ~90 bytes, so it isn't an access-list artifact.

`reserve_l1_gas`'s two sites now raise `TxValidationError::L1GasReservationTooLow`. That leaves `IntrinsicGasTooLow` with three producers — `validate_min_gas_limit`'s `gas_limit < intrinsic` check, its EIP-8037 cap, and the equivalent budget check in `add_intrinsic_gas` — all deterministic, so the `BlockchainType` parameter is gone from the predicate and both call sites. Net effect is less machinery than before `c05006d`, and L2 gets the fix instead of the exclusion.

No EF exposure: `reserve_l1_gas` is L2-only and `IntrinsicGasTooLow`'s own string is untouched, so the exception mappers need no new arm (verified by checking both the root and `tooling` workspaces).

The pinning test now asserts the transient variant is not evicted, the permanent one still is, and the two don't share a `Display` string. Load-bearing check: rewording the new variant to contain the intrinsic-gas substring fails it.

Left the admission-side mirror out — it's a real gap in `transaction_intrinsic_gas`'s stated parity invariant, but it wouldn't remove the builder-side need (a tx admitted pre-Amsterdam still needs eviction once Amsterdam activates), so it's better as its own change.

crates/blockchain/payload.rs
ilitteri: `8da4d3cc1` — right, the count was wrong: three producers, two of them in the fee-token path.

Rewrote the rule to say "every producing site" rather than counting within `default_hook`, and recorded the Prague-vs-Osaka gap as the arm's justification, since that's the case that fires with no fork boundary. Also qualified the neighbouring claim that admission covers the per-tx gas cap — it only holds while `is_osaka_activated && !is_amsterdam_activated`.

crates/blockchain/payload.rs
ilitteri: `be74a3fbd` — took the neutral wording. The L2 block producer's log keeps the specific phrasing, since only the deterministic branch exists there.

ilitteri:
Applied: `436b4dcd2` (distinct `L1GasReservationTooLow` variant, drops the chain-type gate), `8da4d3cc1` (producer count + Prague-vs-Osaka gating), `be74a3fbd` (neutral eviction log). Description rewritten to cover all four arms, the variant split, the L2 producer fix and the fork-boundary bound — it had been stale since the original three-substring version.

On the review-timing note: no re-ping needed for now, since `REVIEW_REQUIRED` already blocks the merge and these three commits change the L2 path again anyway. Worth flagging that the CI observation is stale — all 60 check-runs were green on `f5b65826f` before this push.

edg-l (approved):

PR #6997, opened 2026-07-15
fix(l1): harden supplied-BAL RPC input (0x prefix + decode alloc cap)

edg-l:
Two small hardening fixes for the supplied EIP-7928 block access list on the Engine API / decode path. Both are strictness/decoding issues on the future Amsterdam fork (JWT-gated, no consensus split), but worth tightening.

## `0x`-prefix strictness
`blockAccessList` is a spec DATA field (`^0x[0-9a-f]*$`), so the `0x` prefix is mandatory. The engine `newPayload` paths used `trim_start_matches("0x")`, silently accepting unprefixed (schema-invalid) values that strict clients reject. Now the raw-hash path (`payload.rs`) and both BAL RLP deserializers (`serde_utils::block_access_list::rlp_str` / `rlp_str_opt`) require the prefix. `"0x"` still denotes an absent BAL (unchanged). Scoped to the BAL-specific helpers — the shared numeric/hex serde utils are untouched.

## Capacity-hint cap in `build_validation_index`
A supplied BAL padded with many (empty) accounts drove `FxHashMap::with_capacity(inner.len())` / `Vec::with_capacity(inner.len())` allocations up front, before the per-account "never accessed" check and the size cap both reject it. Those capacity args are only pre-allocation hints, so they're now capped (`min(PREALLOC_CAP)`); the collections still grow to whatever a genuine block needs. Error precedence of the post-execution size cap is deliberately left unchanged.

## Testing
- Serde tests (`test/tests/common/bal_input_tests.rs`): prefixed accepted, unprefixed rejected, `"0x"`/null → absent.
- `build_validation_index` test: a block with more accounts than the cap is still indexed in full (no truncation).
- Full ef-test engine gate (`make -C tooling/ef_tests/engine test`, 10921 fixtures / 74462 total) passes — real 0x-prefixed BALs unaffected. fmt/clippy clean.


test/tests/common/bal_input_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+                .with_balance_changes(vec![BalanceChange::new(1, U256::from(i as u64 + 1))])
+        })
+        .collect();
+    let bal = BlockAccessList::from_accounts(accounts);
+    let index = bal.build_validation_index();
+    assert_eq!(
+        index.addr_to_idx.len(),
+        n,
+        "every account must be indexed despite the capacity-hint cap"
+    );
```
ElFantasma, line 96: Nit — non-blocking: the assertion covers `addr_to_idx.len() == n` but three collections are capacity-capped in `build_validation_index`: `addr_to_idx`, `slot_idx_by_account`, and the per-account `slot_map`. Extending this test to also assert `index.slot_idx_by_account.len() == n` (and, optionally, one point lookup like `assert!(index.addr_to_idx.contains_key(&accounts.last().unwrap().address))`) locks in that none of the three caps silently truncates. Same test scaffold, marginal cost.

ElFantasma (approved):
One non-blocking test-strengthening observation inline.

ilitteri (approved):

iovoid (approved):

PR #6999, opened 2026-07-15
perf(l1): execute precondition transactions during mempool prewarming

iovoid:
**Motivation**

Currently if the same sender sends two transactions at different points in time, we execute the first once which (due to nonce issues) prevents the second from executing.

**Description**

This PR always executes all the previous transactions from the same sender (which is cheaper since they are already warmed). This prevents nonce errors and reduces state-related errors (eg. approve + swap).

It also tunes the warming deadline using real data for a further minor improvement.

The performance improvement is only ~4%, but leaving the bug unfixed could unknowingly prevent future optimizations from working properly.


ilitteri (approved):

crates/blockchain/prewarm.rs
```diff
@@ -374,17 +414,14 @@ fn run_pass(blockchain: &Blockchain, pool: &rayon::ThreadPool, req: PrewarmReque
...
-            let fresh = cap_sender_depth(
-                fresh,
-                &warmed_per_sender,
-                MAX_WARMED_TXS_PER_SENDER_PER_SLOT,
-            );
-            let warm_set = select_warm_set(fresh, base_fee, gas_budget);
+            // Keep only each sender's ready contiguous prefix (drops stale +
+            // gapped txs that would fail the nonce check and warm nothing).
+            let ready = filter_ready(txs_by_sender, cache_dyn.as_ref());
+            let capped = cap_sender_depth(ready, MAX_WARMED_TXS_PER_SENDER_PER_SLOT);
```
ElFantasma: nit: `MAX_WARMED_TXS_PER_SENDER_PER_SLOT` is now misleading — after this PR it's a per-*snapshot/pass* cap (the doc comment on `cap_sender_depth` even calls this out: "no cross-pass accounting"). A future reader looking for the slot-level ceiling will double-count. Consider `MAX_WARMED_TXS_PER_SENDER_PER_PASS` to keep the semantic honest. Non-blocking.

crates/blockchain/prewarm.rs
```diff
@@ -86,38 +101,63 @@ fn select_warm_set(
...
+    }
+    ready
+}
+
+/// Trims each sender's snapshot to its ready contiguous prefix (see
+/// [`trim_to_ready`]) using the account nonce from parent state. The nonce
+/// read goes through the warming cache, so it also warms the sender account;
+/// on a read error the sender is left unfiltered. Empty senders are removed.
+#[cfg(all(feature = "rayon", not(feature = "eip-8025")))]
+fn filter_ready(
```
ElFantasma, line 137: nit: no direct test for `filter_ready` (only `trim_to_ready`). The read-error branch (`if let Ok(...)`) leaves the sender unfiltered — worth pinning that behavior with a test that fails `get_account_state` for one sender and passes for another, to prevent a future refactor from accidentally dropping the unfiltered sender. Non-blocking; `trim_to_ready` covers the meat.

ElFantasma (approved):
Two non-blocking nits inline.

MegaRedHand (approved):

PR #7000, opened 2026-07-16
refactor(l1,l2): remove v0 blob sidecar support

### LINKED ISSUE #6927: Remove v0 blob sidecar support (follow go-ethereum#35191): produce & store only v1 cell-proof sidecars
**Context**

[go-ethereum#35191](https://github.com/ethereum/go-ethereum/pull/35191) removed v0 blob sidecar support entirely — geth now always produces and gossips v1 (EIP-7594 cell-proof) sidecars, even pre-Osaka, and rejects v0. ethrex still creates and stores v0 sidecars:

- `BlobsBundle::create_from_blobs` defaults to v0 (`wrapper_version.unwrap_or(0)`).
- The L2 committer passes `Some(0)` (`crates/l2/sequencer/l1_committer.rs:1213`).
- `validate_cheap` accepts both v0 and v1 pre-Osaka (after #6919).

All networks ethrex targets are already post-Osaka (mainnet `osakaTime` ~2025-12-03; sepolia/hoodi earlier), so v0 sidecars are legacy weight.

**Goal**

Move ethrex to v1-only sidecars internally, matching geth:

- `create_from_blobs` → always v1 (cell proofs); drop the v0 branch.
- L2 committer (`crates/l2/sequencer/l1_committer.rs:1213`, `Some(0)`) → v1.
- `validate_cheap` → require v1 (remove the pre-Osaka v0 acceptance added in #6919); delete the now-dead v0 verify/length branches.
- Audit all `create_from_blobs` callers (defaults / `None` args).

**Pre-Osaka Engine API paths (no recompute needed)**

This is a standalone cleanup — it does **not** require deriving v0 proofs on demand. geth removed v0 from the pool without any v1→v0 conversion path (it explicitly prohibits proof conversion as too expensive), and handles the pre-Osaka Engine API by **declining**:

- `engine_getBlobsV1` — geth returns all-null for a v1-only pool; ethrex already does the same after #6919 (`getBlobsV1` returns `null` for v1 bundles). No change needed here.
- `engine_getPayloadV3/V4` — geth copies cell proofs through unconverted and relies on Osaka being live so pre-Osaka blocks aren't built. ethrex is in the same position (all target networks post-Osaka), so this is a latent path unreachable in production. If we want it correct for pre-Osaka CI/private chains, the geth-consistent option is to **skip v1 blob txs when building a pre-Osaka payload** (upholds geth's "pre-Osaka packs only v0" invariant) — not recompute.

(Supersedes the earlier plan in #6926, which assumed on-demand v0 derivation was required; that was closed as not-planned once geth's decline-don't-derive behavior was confirmed.)

**Verification**

unit + EF state/blockchain + hive (devp2p, engine getBlobs, execute-blobs) + L2 integration.

NikhilSharmaWe:
Closes #6927

## Summary

Stop producing and accepting v0 blob sidecars. Match geth (#35191):
always use v1 (EIP-7594 cell-proof) sidecars.

## What changed

- `create_from_blobs` always builds v1 sidecars
- mempool validation rejects v0
- dead v0 proof/verify paths removed
- payload merge uses a fallible v1 append (avoids version/label mismatch)
- L2 store/committer regenerate v1 from stored blobs
- prover still gets a separate single EIP-4844 proof where needed
- `engine_getBlobsV1` stays decline-don’t-derive (null for v1 pool data)
- `based/1` `BatchSealed` wire format unchanged
- L2 networking requires `c-kzg`
- Osaka CLI flags kept as deprecated no-ops (warn + ignore)
- regression tests for admission, merge, prover proof, and BatchSealed

## Tested

- blob-bundle unit tests
- mempool v0 rejection tests
- Engine `blobs_v1` tests
- BatchSealed wire/legacy-shape tests
- `cargo check` on affected crates (incl. `ethrex-p2p` with `l2`)

Not run yet: Hive / EF / full L2 integration from #6927.

### DIFF
--- a/cmd/ethrex/Cargo.toml
+++ b/cmd/ethrex/Cargo.toml
@@ -106,6 +106,7 @@ jemalloc_profiling = [
 sync-test = ["ethrex-p2p/sync-test"]
 
 l2 = [
+  "c-kzg",
   "ethrex-l2",
   "ethrex-l2-common",
   "ethrex-l2-rpc",
--- a/cmd/ethrex/l2/command.rs
+++ b/cmd/ethrex/l2/command.rs
@@ -149,7 +149,7 @@ pub enum Command {
         #[arg(
             short = 'o',
             long,
-            help = "Whether Osaka fork is activated or not. If None, it assumes it is active."
+            help = "Deprecated compatibility option; v1 blob sidecars are always used."
         )]
         osaka_activated: Option<bool>,
     },
@@ -381,6 +381,11 @@ impl Command {
                 store_path,
                 osaka_activated,
             } => {
+                if osaka_activated.is_some() {
+                    tracing::warn!(
+                        "--osaka-activated / -o is deprecated and ignored; ethrex always uses v1 (EIP-7594) blob sidecars"
+                    );
+                }
                 #[cfg(feature = "rocksdb")]
                 let store_type = EngineType::RocksDB;
 
@@ -501,16 +506,7 @@ impl Command {
                     let blob = blobs_bundle::blob_from_bytes(Bytes::copy_from_slice(&blob))
                         .expect("Failed to create blob from bytes; blob was just read from file");
 
-                    let wrapper_version = if let Some(activated) = osaka_activated
-                        && !activated
-                    {
-                        None
-                    } else {
-                        Some(1)
-                    };
-
-                    let blobs_bundle =
-                        BlobsBundle::create_from_blobs(&vec![blob], wrapper_version)?;
+                    let blobs_bundle = BlobsBundle::create_from_blobs(&vec![blob])?;
 
                     let batch = get_batch(
                         &store,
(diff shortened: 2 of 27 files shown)


NikhilSharmaWe:
@ElFantasma A few decisions I’d like to confirm before marking this ready:

1. **L2 upgrade order**  
   New followers soft-ignore legacy one-proof-per-blob `BatchSealed` under `based/1`.  
   Is a documented sequencer-first rollout enough, or do we need `based/2`?

2. **L2 ingress KZG checks**  
   We now validate structure + cell proofs before queue/rebroadcast.  
   `validate_signature` is still a no-op, so this adds some CPU cost/DoS surface.  
   Keep it in this PR and harden auth/DoS later, or split it out?

3. **Compat surface**  
   Osaka CLI flags remain deprecated no-ops.  
   `get_l1_active_fork` was removed. OK as-is, or keep a deprecated shim?
   
Please let me know

PR #7002, opened 2026-07-16
fix(l1): reject non-canonical 0x00-typed transaction envelopes (EIP-2718)

ElFantasma:
## Motivation

Closes the EF finding `typed-zero-legacy-transaction`. Per [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718), a legacy transaction is a bare RLP list; a *typed* envelope is `TransactionType || payload` where the type is `0x01..=0x7f`. Type `0x00` is **not** assigned, and geth rejects `0x00 || …` (`ErrTxTypeNotSupported`).

ethrex instead mapped a leading `0x00` byte to a legacy transaction in **four** decoders (each with a literal `// TODO: check if this is a real case scenario`). Because `compute_transactions_root` re-encodes each tx via `encode_canonical_to_vec` — which for a legacy tx emits `rlp(legacy)`, dropping the `0x00` prefix — a block body carrying `0x00 || rlp(legacy)` is committed by ethrex as `rlp(legacy)`. ethrex derives a different transactions-root than the raw bytes represent and can accept a block a strict client rejects: a **cross-client consensus-split risk**. The same non-canonical encoding also entered via `eth_sendRawTransaction` and P2P tx gossip.

## Description

Introduces a single source of truth for envelope type-byte validity:

- `TxType::from_type_byte(byte) -> Result<TxType, RLPDecodeError>` — accepts only assigned typed types (`0x01, 0x02, 0x03, 0x04, 0x06, 0x7d, 0x7e`) and **rejects `0x00` and any unassigned byte**. (`TxType::from_u8` is left unchanged — its `0x00 => Legacy` mapping is correct for contexts like receipts, where the type is already known valid.)

The four typed-envelope decoders now route their type dispatch through it (matching the returned `TxType` enum, so a future type is added in one place and the match is compiler-checked per site):

- `Transaction::decode_canonical` (block import / engine / RPC)
- `Transaction::decode_unfinished` (RLP decode of block-body txs)
- `P2PTransaction::decode_unfinished` (eth/6x tx gossip — additionally rejects `Privileged`, which is never gossiped)
- `SendRawTransactionRequest::decode_canonical` (`eth_sendRawTransaction`)

Genuine legacy txs are untouched — they arrive as RLP lists and take the existing list branch, never the typed-envelope match. Nothing in ethrex ever *emits* `0x00 || …`, so the removed arm only ever matched malformed/hostile input.

## Tests

`test/tests/blockchain/typed_zero_transaction_tests.rs` (written test-first, observed red before the fix):
- `0x00`-typed envelope is **rejected** by `Transaction::decode_canonical`, `Transaction::decode`, and `P2PTransaction::decode`.
- End-to-end: a **block** whose body carries a `0x00`-typed tx fails `Block::decode` (the P2P block-import path: block RLP -> `BlockBody` -> `Vec<Transaction>`); the malicious tx is injected via `Encoder::encode_raw` since the encoders can't produce it.
- Controls: a genuine legacy tx (bare RLP list) and a valid typed tx (`0x02` EIP-1559) still decode, and a well-formed block with a legacy tx still round-trips through `Block::decode`.

## Checklist

- [x] `cargo test -p ethrex-test --test ethrex_tests typed_zero` (5 tests) + `cargo test -p ethrex-common transaction` (66, no regression) + `cargo clippy -p ethrex-common -p ethrex-rpc` pass.


ElFantasma:
Thanks for the reviews — all four approve (Codex "no findings", greptile 5/5, Kimi + Claude approve). Addressed the two actionable nits in `03c68704c`:

- **@Kimi (#6):** split the merged `TxType::Privileged | TxType::Legacy` arm in `P2PTransaction::decode_unfinished` into two arms with distinct messages — `privileged transactions are not sent over p2p` vs `legacy transactions are not typed envelopes` — since they violate different constraints.
- **@Claude (#3):** added the "never returned by `from_type_byte` … RLP list, not a typed envelope" comment to the rpc `TxType::Legacy` arm, matching the two arms in `common/types/transaction.rs`.

**Not changed (out of scope):**
- **@Claude (#1):** the `< 0x7f` (`Transaction::decode_canonical`) vs `<= 0x7f` (`SendRawTransactionRequest::decode_canonical`) guard mismatch is pre-existing and now moot for correctness — `from_type_byte` rejects `0x7f` either way, so both paths error (just via different branches). Leaving it out of this consensus-focused fix; worth a separate cleanup PR to unify the guards.

The dead `TxType::Legacy => Err(..)` arms are intentional — they satisfy the exhaustive match (`from_type_byte` never returns `Legacy`), which is the compiler-checked property that makes adding a future tx type safe across all four decoders.

iovoid (approved):

ilitteri (approved):

crates/common/types/transaction.rs
```diff
@@ -534,38 +543,42 @@ impl RLPDecode for Transaction {
...
+                TxType::Privileged => Transaction::PrivilegedL2Transaction(
+                    PrivilegedL2Transaction::decode(tx_encoding)?,
+                ),
+                // Never returned by `from_type_byte` (0x00 is rejected); a legacy tx arrives as
+                // an RLP list on the branch below, not as a typed envelope.
+                TxType::Legacy => {
+                    return Err(RLPDecodeError::Custom(
+                        "legacy transactions are not typed envelopes".to_string(),
+                    ));
+                }
```
MegaRedHand, line 579: We should use `unreachable` then, or use another enum for the return type

MegaRedHand (approved):
LGTM. Left a comment for a following refactor

crates/common/types/transaction.rs
ElFantasma: Good catch — the `Legacy` arm is unreachable (`from_type_byte` never returns it), so that `Err` is misleading dead code. Opened #7014 to do the type-safe version — a `from_type_byte` return type without `Legacy`, dropping the arm across all 4 decoders — as a follow-up rather than expanding this PR. Leaning to the dedicated enum over `unreachable!()` to avoid adding panics on the decode path.

PR #7003, opened 2026-07-16
fix(l1): box forkchoiceUpdated engine handlers

edg-l:
`map_engine_requests` awaits each match arm inline, so its future is sized to the largest arm and shared by every engine request. The `forkchoiceUpdated` arms weren't boxed like the `newPayload` arms; box them so a lightweight `forkchoiceUpdated` poll doesn't have to reserve the full frame.

### DIFF
--- a/crates/networking/rpc/rpc.rs
+++ b/crates/networking/rpc/rpc.rs
@@ -1457,10 +1457,10 @@ pub async fn map_engine_requests(
 ) -> Result<Value, RpcErr> {
     match req.method.as_str() {
         "engine_exchangeCapabilities" => ExchangeCapabilitiesRequest::call(req, context).await,
-        "engine_forkchoiceUpdatedV1" => ForkChoiceUpdatedV1::call(req, context).await,
-        "engine_forkchoiceUpdatedV2" => ForkChoiceUpdatedV2::call(req, context).await,
-        "engine_forkchoiceUpdatedV3" => ForkChoiceUpdatedV3::call(req, context).await,
-        "engine_forkchoiceUpdatedV4" => ForkChoiceUpdatedV4::call(req, context).await,
+        "engine_forkchoiceUpdatedV1" => Box::pin(ForkChoiceUpdatedV1::call(req, context)).await,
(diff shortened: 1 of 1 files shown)


iovoid (approved):

ilitteri (approved):

crates/networking/rpc/rpc.rs
```diff
@@ -1457,10 +1457,10 @@ pub async fn map_engine_requests(
...
     match req.method.as_str() {
         "engine_exchangeCapabilities" => ExchangeCapabilitiesRequest::call(req, context).await,
-        "engine_forkchoiceUpdatedV1" => ForkChoiceUpdatedV1::call(req, context).await,
-        "engine_forkchoiceUpdatedV2" => ForkChoiceUpdatedV2::call(req, context).await,
-        "engine_forkchoiceUpdatedV3" => ForkChoiceUpdatedV3::call(req, context).await,
-        "engine_forkchoiceUpdatedV4" => ForkChoiceUpdatedV4::call(req, context).await,
+        "engine_forkchoiceUpdatedV1" => Box::pin(ForkChoiceUpdatedV1::call(req, context)).await,
+        "engine_forkchoiceUpdatedV2" => Box::pin(ForkChoiceUpdatedV2::call(req, context)).await,
+        "engine_forkchoiceUpdatedV3" => Box::pin(ForkChoiceUpdatedV3::call(req, context)).await,
+        "engine_forkchoiceUpdatedV4" => Box::pin(ForkChoiceUpdatedV4::call(req, context)).await,
```
ElFantasma, line 1463: nit: the explanatory comment at line 1464-1470 ("newPayload handlers carry the largest futures…") is now scoped narrowly — the same rationale applies to the fcu arms this PR is boxing, but the doc still talks only about newPayload. Consider moving the comment above the first boxed arm (`engine_forkchoiceUpdatedV1`) and generalizing the wording to "the engine handlers below", so the next reader understands why *both* families are boxed rather than assuming fcu is boxed for a different reason. Non-blocking.

ElFantasma (approved):

PR #7004, opened 2026-07-16
fix(l1): catch EIP-8141 up to the current spec revision

ilitteri:
**Motivation**

ethrex's EIP-8141 implementation tracked spec commit `fe0940cae2` (2026-06-03). The spec was substantially rewritten after that: 320 lines added, 136 removed. Signature canonicality moved into consensus validation, `ARBITRARY` was introduced as a scheme with its own gas cost, the skipped-frame receipt status was renumbered, atomic batches stopped admitting `VERIFY` frames, default code's signature index became scope-dependent, storage refunds became transaction-scoped, and the EIP-7623 calldata floor started applying to frame and signature data. Spec-conformant nodes therefore reject frame transactions ethrex produces, and ethrex accepts transactions they reject.

This PR brings the implementation up to the current spec revision (`0af4e0044d`, 2026-07-28) and repins `docs/eip-8141.md` to it. Every `spec commit <hash>` and `spec line NNN` citation in the code is now stale (line numbers all moved, and several hashes were being cited as authority for values that had since changed), so those are replaced with section references.

**Description**

*Signature semantics*

- **Scheme numbering**: add `FRAME_SIG_SCHEME_ARBITRARY = 0`; renumber `SECP256K1 = 1`, `P256 = 2`. Verification, gas, the SIGPARAM handlers and static validation all reference the named constants, so they follow.
- **Empty signer**: `FrameSignature.signer` becomes `Option<Address>`, RLP-encoded via `TxKind` like `Frame.target`. `ARBITRARY` requires an empty signer; for `SECP256K1`/`P256` an absent signer resolves to `tx.sender`, for verification and for introspection alike.
- **Canonicality is now a consensus rule.** `validate_frame_signatures` rejects `v > 1`, `r == 0`, `r >= n`, and high-s (`s > n/2`), for both `SECP256K1` and `P256`. The former mempool-only policy (`frame_signatures_are_low_s`) and its `MempoolError::FrameTxMalleableSignature` variant are deleted, since high-s is no longer merely unrelayable.
- **`v` is a bare recovery id.** The spec fixes `v ∈ {0, 1}`; ethrex was forwarding it to the `ecrecover` precompile unchanged, which accepts only 27/28, so it required exactly the encoding the spec forbids. `+27` is now applied when building the precompile calldata, not carried on the wire.
- **`ARBITRARY` verify gas**: 0 → 100 per entry.

*Opcodes*

- **`SIGPARAM 0x04`**: copy an `ARBITRARY` signature's raw bytes into memory with `CALLDATACOPY` gas and semantics, zero-filling past the end; other schemes halt.
- **`SIGPARAM 0x00`**: an exceptional halt for `ARBITRARY` entries, which the protocol assigns no resolved signer to. It previously returned `tx.sender`.
- **`FRAMEPARAM 0x05`**: reports the renumbered skipped status.

*Execution and approval*

- **`APPROVE_PAYMENT` ordering**: scope `0x1` reverts while `sender_approved == false`; payment must not precede the sender's execution approval. Scope `0x3` is exempt, since it sets the flag atomically. This closes an over-permissiveness where a paymaster VERIFY frame could bind the payer and debit the max cost before the sender approved anything.
- **Default code signature index**: `sig_index = 0` when the allowed scope includes `APPROVE_EXECUTION`, else `sig_index = 1`. It was a whole-list scan, then a hardcoded index 0; neither matches a payment-only default VERIFY frame.
- **Atomic batches exclude `VERIFY` frames**: static validation rejects `ATOMIC_BATCH_FLAG` on a `VERIFY` frame and requires the frame it batches with to be non-`VERIFY` too.
- **Skipped-frame receipt status**: `0x3` → `0x2`.

*Gas accounting*

- **Transaction-scoped storage refunds**: EIP-3529 refunds accumulate across frames into one counter and are applied once at the end, capped at `gas_used_before_refund / 5`. The counter already lived in the substate that frame reverts and batch unrolls discard; it was simply never applied on the frame path, so refunds were silently dropped.
- **EIP-7623 calldata floor**: the data cost now covers `frame.data` and each signature's `signer`, `msg` and `signature` only, not the RLP framing or the scalar fields. The floor (`64` gas per byte, per EIP-7976; frame transactions exist only from Hegota, which is after Amsterdam, so the raised floor always applies) is floored against execution in the charged gas, and reserved as a static validity check so a transaction cannot carry data it could never pay for.

*Static constraints*

- Versioned hashes must carry the EIP-4844 KZG version byte.
- `max_fee_per_blob_gas` must be zero when the transaction carries no blobs.
- `frame.flags & APPROVE_EXECUTION` requires an empty target or `tx.sender`.

*Cleanup*

- Remove the dead `opcode_handlers/frame.rs`: 444 lines of pre-spec draft with a name-colliding `APPROVE` handler, never declared in `mod.rs`.
- `docs/eip-8141.md` repinned and its stale sections rewritten to the current behavior.

*Tests*

Updated the golden RLP and `sig_hash` vectors for the scheme byte, and added coverage for the new rules: canonicality rejections (high-s, zero and out-of-range `r`, `v > 1`, the 27/28 encoding), `ARBITRARY` and empty-signer handling, the three new static constraints, data cost and floor arithmetic, floor reservation, scope-dependent default-code signature index (both directions, with real signatures so only the index differs), and cross-frame refund application.

`SIGPARAM` had no execution-level coverage at all, only unit tests that reimplemented the handler's logic without invoking the opcode. Added real bytecode tests for `0x01`, `0x03`, the `0x00` `ARBITRARY` halt, and `0x04` copy and zero-fill. The `0x04` non-`ARBITRARY` halt guard is still uncovered; exercising it needs a P256 signature that verifies against the transaction's own sig hash.

Several existing fixtures were genuinely invalid under the new rules rather than merely stale: paymaster VERIFY frames using scope `0x3` on a third-party target, and transactions whose frame gas limits could not reserve their own calldata floor.

The spec also newly specifies the `Receipts` p2p encoding for frame receipts, as `[tx-type, cumulative-gas, payer, [[status, gas-used, logs], ...]]`. ethrex already emits exactly that on eth/69 and later via `Receipt::encode_inner`, and Hegota rides eth/70 and eth/71, which both serve `Receipts70` through that impl; no code change was needed, so this adds a test pinning the wire bytes and the message round-trip. eth/68 still serves the raw-prefix consensus payload instead, which is unreachable for frame transactions since eth/68 predates the fork; that is noted in the docs rather than changed.

**Consensus / compatibility**

Frame transactions are gated behind Hegota and are not live on any production fork, so there is no mainnet consensus impact. Three changes are breaking for existing Hegota devnets and tooling:

- The scheme byte and the signer encoding are part of the `compute_sig_hash` preimage, so devnets must be rebuilt and any contract hardcoding a scheme number read via `SIGPARAM` needs a redeploy.
- Signers must emit `v` as a bare recovery id (0 or 1). The previous 27/28 encoding is now consensus-invalid, and high-s signatures are rejected at block execution rather than only at admission.
- The skipped-frame status renumber moves receipts-trie bytes.

One open question worth a cross-client check before this is relied on: the spec does not say whether a prior frame's storage write counts as the `original` value for a later frame's EIP-3529 refund. A set-then-clear across two frames currently yields an applied refund of 5200, which matches neither reading of a transaction-boundary `original`. The test asserts only what the spec does guarantee, that a refund is applied and respects the one-fifth cap.

Deliberately out of scope, as a p2p change rather than an execution-semantics fix: **blob-carrying frame transactions**, which the spec now defines fully (the payer takes the EIP-4844 blob-fee role, the EIP-7594 per-transaction blob limit applies, `BLOBHASH` resolves in every frame, and carrying blobs is optional). ethrex still rejects them at admission. Much of the execution half already works incidentally, `compute_tx_max_cost` covers the blob cost, the refund burns the base-rate blob fee, `tx_blob_hashes` is wired so `BLOBHASH` resolves in frames, and block import counts frame blobs; what is missing is the `PooledTransactions` EIP-7594 wrapper, its validation, blobpool indexing, announce-only gossip, and blob accounting in the payload builder. Follow-up PR.

**Checklist**

- [x] `STORE_SCHEMA_VERSION` unchanged; no `Store` schema change, so no re-sync is required.


crates/vm/levm/src/vm.rs
```diff
@@ -769,7 +771,14 @@ pub fn validate_frame_signatures(
                     return false;
                 }
                 let recovered = ethrex_common::Address::from_slice(&result[12..]);
-                if recovered == ethrex_common::Address::zero() || recovered != sig.signer {
+                if recovered == ethrex_common::Address::zero() {
+                    return false;
+                }
+                // An empty signer (None) resolves to the recovered key; a pinned
+                // signer must equal it.
+                if let Some(expected) = sig.signer
```
iovoid: empty signers are read by `SIGPARAM 0x00` as the empty address, not set to the recovered value

iovoid: The spec says
> For SECP256K1 and P256, the signer is a 20-byte Ethereum address. If absent, tx.sender is used (for introspections as well)

crates/vm/levm/src/opcode_handlers/frame_tx.rs
iovoid, line 556: The spec added a 0x04 operation that can copy signature data from `ARBITRARY` signatures.

iovoid: No longer true for `ARBITRARY` signatures

iovoid (changes requested):
docs/eip-8141.md should be updated

iovoid (approved):

ElFantasma (approved):
LGTM

edg-l (approved):

PR #7006, opened 2026-07-17
fix(l1): amsterdam header + BAL selfdestruct conformance fixes

edg-l:
Three small Amsterdam-fork conformance fixes, each dormant until fork activation (no mainnet/consensus impact today).

## 1. SELFDESTRUCT: don't record warm-but-unread slots as BAL reads (EIP-7928)
On `SELFDESTRUCT`, ethrex folded the entire warm access set (`get_accessed_storage_slots`) into the block access list as storage reads. That set includes EIP-2930 access-list slots that were never actually read, whereas the spec records a read only on an actual `get_storage` access. The extra entries diverge the `block_access_list_hash` from conformant clients. Genuine reads are already captured on `SLOAD`/`SSTORE` via `record_storage_slot_to_bal`, so the per-selfdestruct loop is removed (both fork branches). Touched-address and initial-balance recording is unchanged.

## 2. Require `slot_number` in Amsterdam headers (EIP-7843)
`validate_prague_header_fields` enforced `block_access_list_hash` presence on Amsterdam but never `slot_number`, accepting headers that omit the trailing field. Now Amsterdam requires `slot_number` present, and the pre-Amsterdam validators (prague-pre-amsterdam / cancun / pre-cancun) reject it when present, mirroring the existing BAL-hash checks.

## 3. Avoid u64 overflow in `excess_blob_gas` reserve-price check (EIP-7918)
`U256::from(BLOB_BASE_COST * parent_base_fee_per_gas)` multiplied two `u64`s before widening, overflowing when the parent base fee exceeds `u64::MAX / 8192`. Widen both operands to `U256` before the multiply.

## Testing
- `test/tests/levm/bal_selfdestruct_reads_tests.rs`: a SELFDESTRUCT tx that warms a slot via access list but only SLOADs a different slot — asserts the read slot is in the BAL and the warm-unread slot is not.
- `test/tests/common/amsterdam_header_tests.rs`: slot_number required/rejected across forks; `calc_excess_blob_gas` no longer overflows at large base fees.
- clippy clean; existing selfdestruct/BAL suites (111 tests) and common lib tests (139) pass.

Note: the full ef-test engine gate (`make -C tooling/ef_tests/engine test`) is the authoritative conformance check for #1 and should run in CI.


ElFantasma (approved):
LGTM

ilitteri (approved):

iovoid (approved):

PR #7009, opened 2026-07-17
refactor(l1): unify P2PTransaction into PooledTransaction

iovoid:
**Motivation**

Currently we duplicate a lot of deserialization logic between P2PTransaction and Transaction, the difference being that the former carries the blob bundle for blob transactions instead of just the commitment.

**Description**

Unifies, turning P2PTransaction into PooledTransaction which holds a transaction and an optional blob bundle.


ElFantasma (approved):
One micro-observation, non-blocking: `encode_canonical` derives `wrapper_version` as `(bundle.version != 0).then_some(bundle.version)`, which drops a *stored* `wrapper_version: Some(0)` back to `None` on re-encode. The old `WrappedEIP4844Transaction` stored `wrapper_version` independently. In practice `0` is the pre-Osaka default and peers set `wrapper_version = None` for it (matching the derivation), so this is likely intentional lossless — worth a sanity check that no code path actually persists an explicit `Some(0)` upstream. LGTM.

PR #7016, opened 2026-07-21
fix(l1): avoid duplicate process metrics

catwith1hat:
The Prometheus default registry already exposes the built-in process collector. Registering a second collector in METRICS_PROCESS appended duplicate samples to each response, and values gathered at different times could make Prometheus reject the scrape.

That duplicate metrics are present can be verified with:

```
curl -s http://localhost:9090/metrics | grep '^process_.* ' | sort
```


MegaRedHand (approved):
LGTM

`prometheus` v0.14 does the same initialization we do here if we compile it with the "process" feature, which we do.

Here's the relevant code:

https://github.com/tikv/rust-prometheus/blob/e07efb4f372f1245bf7410b71e822c69877bcb32/src/registry.rs#L280-L292

ElFantasma (approved):
Good catch. `prometheus = { features = [\"process\"] }` on Linux auto-registers a ProcessCollector into the default registry, so `gather_default_metrics()` already emits `process_cpu_seconds_total`, `process_resident_memory_bytes`, etc. Re-registering `ProcessCollector::for_self()` in this local Registry and concatenating that output produced the duplicate samples the PR body describes. Dropping the manual registration is the right fix; the remaining `datadir_size_bytes` emission still has a reason to live in this function.

LGTM.

ilitteri (approved):

ilitteri:
Hey! The PR has met all the needed approvals, yet this repository requires all commits to be signed for it to be merged.

catwith1hat:
Signed the commit. Should be ready for merging now.

catwith1hat:
No response from maintainers.

PR #7018, opened 2026-07-21
fix(l1): bound single-canonical-chain re-execution memory via depth-gated commit

ElFantasma:
**Motivation**

Startup state regeneration (`regenerate_head_state`) can OOM. When the node restarts with a canonical head ahead of the last on-disk state root (e.g. the CL forkchoice canonicalized blocks whose state was never committed, or a checkpoint-then-restart), the node re-executes the canonical-but-stateless gap block by block. Each re-executed block produces one in-memory trie diff-layer, and these layers are only flushed to disk by the canonical safe-commit gate (`get_commitable`, keyed on `head - DB_COMMIT_THRESHOLD`). At startup nothing has been canonicalized yet, so `safe_commit_root` is zero, `get_commitable` returns `None`, and **no layer is ever committed** — the backlog grows with the size of the gap until the process runs out of memory (~60 GB observed on an ~11k-block regeneration, heap-profiled to `regenerate_head_state`).

The same latent unbounded-accumulation exists in the other single-canonical-chain re-execution paths that use the canonical gate: full-sync block-by-block resume (`run_blocks_pipeline`) and block `import` (which executes every block and only issues a single forkchoice update at the very end, so nothing commits during the import itself).

The batch path already solved the equivalent problem by committing trie layers *by depth* rather than by the canonical gate (`BATCH_COMMIT_THRESHOLD`, introduced in #6231, refined in #6930). This PR extends that same sanctioned mechanism to the per-block re-execution paths.

**Description**

Generalize the per-batch commit strategy from a boolean into an explicit depth:

- `UpdateBatch.batch_mode: bool` → `commit_depth: Option<usize>`:
  - `None` — live path (`newPayload`): commit by the canonical `head - DB_COMMIT_THRESHOLD` safe-commit root, ack after staging (no flush wait). **Behavior unchanged.**
  - `Some(depth)` — single-canonical-chain execution: commit every trie layer deeper than `depth` (`get_commitable_by_depth`), ack after flush to bound in-flight memory.
- Threaded `commit_depth` through `apply_updates` → `BlockPersist` → the persist worker → `commit_trie_if_due`. `store_block` / `add_block_pipeline` / `add_block_pipeline_bal` keep their signatures and pass `None`, so the live path is byte-for-byte unchanged. Added `store_block_with_depth` and `add_block_pipeline_bounded` for the re-execution paths.
- Routed the single-canonical-chain re-execution call sites through the depth gate at `DB_COMMIT_THRESHOLD` (128): `regenerate_head_state` (the OOM fix), `run_blocks_pipeline` (full-sync resume), and `import_blocks` / `import_blocks_bench`.
- `add_blocks_in_batch` now passes `Some(BATCH_COMMIT_THRESHOLD)` (equivalent to the old `batch_mode: true`).

Why depth `128` for the re-execution paths (not the batch value `4`): batch layers hold ~1024 blocks each, so a depth of 4 keeps ~4096 blocks resident; per-block re-execution produces one layer per block, so the same window requires a depth of 128. `128` matches `DB_COMMIT_THRESHOLD` and preserves the recent-state window that snap serving and historical-state queries depend on — committing more aggressively (e.g. depth 4 = 4 blocks) would prune the window and let historical queries read a clobbered on-disk root (the concern raised in #6573). Depth-gated commit is only sound because these paths extend a single canonical chain with no competing forks to mis-commit; live `newPayload` (which can receive non-canonical blocks) keeps the canonical gate.

Regression test (`bounded_reexec_without_fcu_bounds_memory_and_serves_window`): re-executes 135 real blocks (> 128) through `add_block_pipeline_bounded` with **no** forkchoice update — the exact regeneration shape — and asserts (a) the backlog is flushed forward without any FCU (genesis root pruned off disk), and (b) the recent window stays resident and serves *correct* historical state (a balance query at a mid-window root returns that block's value, not the latest root's). Verified to fail on assertion (a) when the call is reverted to the canonical gate, so it is a non-vacuous guard.

The live block-execution path (`newPayload`) is unaffected: it never calls the bounded entry, so its gate, ack timing, and latency are identical to before.

Note: `import_blocks_bench` now waits-for-flush per block (in-flight bounded to ~1), so its throughput numbers will shift from the previous fire-and-forget behavior. This is intentional (bounded memory over unbounded); the benchmark will be re-run under the new behavior.


crates/blockchain/blockchain.rs
```diff
@@ -2225,10 +2244,29 @@ impl Blockchain {
...
     }
 
+    /// Same as [`add_block_pipeline`] but for single-canonical-chain re-execution
+    /// (startup state regeneration, full-sync block-by-block, block import): commits
+    /// trie layers by depth (`commit_depth`) instead of the canonical safe-commit gate,
+    /// bounding in-memory layers to ~`commit_depth`. Only sound because these paths extend
+    /// a single canonical chain with no competing forks. Must NOT be used for live
+    /// `newPayload`. Returns the BAL produced during execution (see
+    /// [`add_block_pipeline_bal`]).
+    pub fn add_block_pipeline_bounded(
```
MegaRedHand, line 2224: I don't like this solution. It sounds to me like we should always limit the number of layers at the DB, instead of having two different paths.

ElFantasma:
## Review responses

Thanks all. One fix, one doc cleanup, the rest answered.

### Codex #1 + Claude #1 — per-block wait-for-flush coupling (fixed)

Correct and well-spotted by both. `commit_depth: Some(_)` was doing double duty: selecting the depth gate **and** forcing ack-after-flush. Those are independent concerns, and coupling them serialized CPU execution with disk I/O on the per-block re-execution paths (regen / full-sync fallback / import tail) for no memory benefit — the trie-layer bound comes from `commit_trie_if_due`'s depth gate (which runs every block regardless of ack timing) and the persist channel capacity already bounds in-flight work.

Decoupled in cc598b3c3: added an explicit `wait_for_flush` to `UpdateBatch`/`BlockPersist`, independent of `commit_depth`:

| path | `commit_depth` (gate) | `wait_for_flush` (ack) |
|---|---|---|
| live `newPayload` | `None` (canonical) | `false` (after staging) — unchanged |
| regen / full-sync fallback / import tail | `Some(128)` (depth) | `false` (after staging) — **now pipelined** |
| bespoke batch (`add_blocks_in_batch`) | `Some(4)` (depth) | `true` (after flush) — one message ≈ 1024 blocks / ~1 GB, so bounding in-flight to ~1 still matters |

This is exactly the ack-after-staging behavior the live path already used safely (deferred flush-error via `last_flush_result`, final durability via the idle/`Ping` handshake). Memory stays bounded (depth gate + channel cap); the `bounded_reexec_without_fcu_bounds_memory_and_serves_window` regression test — which drives the now-ack-after-staging bounded path — still passes.

### Codex #2 — `add_block_pipeline_bounded` visibility / footgun

Fair concern for consensus-critical code. It has to stay `pub`: the callers are in other crates (`cmd/ethrex`, `crates/networking/p2p`), so visibility can't be reduced. The current guard is the doc contract plus the fact that it's only reachable from single-canonical-chain contexts. The stacked follow-up #7008 introduces a `SyncCommitMode` enum that encapsulates the depth choice behind an intent-revealing type rather than a raw `usize` — moving in the direction you suggest. Open to tightening further (e.g. a dedicated `CommitDepth` newtype) if the team prefers a stronger type-level guard.

### Claude #2 — same latent pattern in L2 `l1_committer.rs`

Good catch, and out of scope for this l1 fix. `execute_batch_to_generate_checkpoint` has the same no-FCU per-block re-execution shape, though L2 checkpoint batches are far smaller than the ~11k-block L1 regeneration gaps that triggered this, so it's low-risk today. Will file a follow-up issue to route it through the depth-gated path too.

### Claude #3 — stale "batch mode" naming in `get_commitable_by_depth` (fixed)

Updated the doc comment (cc598b3c3) to describe it as the single-canonical-chain re-execution gate (full sync / import / regeneration), not just "batch mode".

### Kimi — verification asks

- `newPayload` untouched: confirmed. `store_block` → `store_block_with_depth(.., None)` keeps the canonical gate; no live-path call site (`rpc.rs`, `l2_connection.rs`, `l1_committer.rs` main commit, `block_producer.rs`) was switched to depth-gating.
- 128 resident layers for the bounded paths: intended and acceptable — it matches `DB_COMMIT_THRESHOLD` / `REORG_DEPTH_LIMIT`, i.e. the recent-state window reorg handling and snap/historical serving depend on.

MegaRedHand (approved):
LGTM, but we should simplify this solution in the future

crates/blockchain/blockchain.rs
ElFantasma: Agree, and thanks for the review. The two paths exist for a specific reason: live `newPayload` uses the canonical gate so non-canonical speculative blocks are never committed to the path-keyed disk (the wedge #6930 fixed), while single-chain re-execution has no FCU to advance `safe_commit_root`, so it needs the depth gate to stay bounded.

But you're right that it can collapse into one rule: a block ≥`REORG_DEPTH_LIMIT` behind head can't be reorged, so committing by depth is safe on the live path too — the canonical gate is really a conservative special case of a distance-from-head rule. #7023 (stacked on this) already generalizes the sync side to distance-gated commit; extending that same rule to live would retire the split entirely. Filed #7027 to track that unification.

ilitteri (approved):

pablodeymo (approved):

PR #7019, opened 2026-07-22
fix(l1): don't pay for a full extra block build in engine_getPayload under sustained mempool inflow

ilitteri:
**Motivation**

ethrex on glamsterdam-devnet-7 misses block proposals because `engine_getPayload` breaches the Engine API 1s deadline: `engine_getPayloadV6` p90 = 834ms / p99 = 986ms over a 4-hour window, with Prysm missing 6 of 31 proposal duties. An engine-API capture on `prysm-ethrex-1` ties one 1000ms `getPayload` response (consensus-client disconnect) to the FCU for slot 50774, which dora shows as Missed. Full analysis: [ethrex getPayload deadline report](https://panda-uploads-production.devops-539.workers.dev/panda/uploads/f8f36e/ethrex-getpayload-deadline-report.html).

**Description**

`PayloadBuildTask::finish()` cancels `build_payload_loop` and awaits it. After the loop exits, the final mempool re-drain introduced in #6596 runs one more **full block build**, gated only on `tx_seq() > last_built_seq`. Under sustained mempool inflow (spamoor keeps ~5,000 txs pending on the devnet) that guard never short-circuits, so every `engine_getPayload` pays for one complete extra block build after cancellation — 0.5–1.0s on 300M-gas EIP-8037 blocks, sitting directly on the proposal critical path.

This PR skips the post-cancellation re-drain when the payload already in hand contains transactions. The re-drain still runs when the held payload is empty — which is exactly the race #6596 closed (a quick FCU → `getPayload` sequence with the tx landing mid-build must not return an empty payload). The slot-timeout path is unchanged. Returning the best-so-far payload on `getPayload` matches the reference clients: geth's `Payload.Resolve()` never waits on or triggers another build once a full payload exists, and reth resolves `PayloadKind::Earliest` without blocking, with an in-source comment citing the 1s deadline as the rationale.

Two deterministic regression tests drive the `build_payload_loop` future manually (no sleeps, no timing assumptions):

- `cancelled_loop_still_drains_mempool_when_payload_is_empty` locks the #6596 property;
- `cancelled_loop_returns_held_payload_without_final_rebuild` fails on current `main` (a tx that lands strictly after cancellation shows up in the payload, proving the extra rebuild ran) and passes with the fix.

An ignored, manual latency reproduction (`getpayload_latency_under_sustained_mempool_inflow`) simulates the devnet's inflow regime: on unfixed `main`, `get_payload` costs one full build (8.4ms vs 8.2ms per build locally; 0.5–1.0s on the devnet's workload); with the fix it returns in ~0.2ms with the same payload size.

`TODO(#5011)` (the detached `spawn_blocking` build keeps running after cancellation) is adjacent but out of scope: this PR removes the extra build from the response path; making in-flight builds cancellable remains follow-up work.

**Checklist**

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -467,7 +467,20 @@ impl Blockchain {
         // an in-progress rebuild via `run_until_cancelled`, and (b) the slot-
         // timeout `select!` arm winning over a simultaneous `tx_added`
         // notification near the slot boundary.
-        if self.mempool.tx_seq() > last_built_seq {
+        //
+        // When the loop was cancelled, `engine_getPayload` is blocked on this
+        // task and must answer within the Engine API deadline (1s), while a
+        // full rebuild of a large block can exceed that deadline on its own —
+        // under sustained mempool inflow `tx_seq() > last_built_seq` is true
+        // on essentially every call, so an unconditional rebuild here puts one
+        // whole extra block build on the proposal critical path. In that case
+        // only rebuild when the payload we hold is empty: returning a slightly
+        // stale block is fine (geth/reth return best-so-far too), but
+        // proposing an empty block while the mempool has transactions is not
+        // (race (a) with no completed rebuild yet).
+        if self.mempool.tx_seq() > last_built_seq
+            && (!cancel_token.is_cancelled() || res.payload.body.transactions.is_empty())
+        {
             let blockchain = self.clone();
             match tokio::task::spawn_blocking(move || blockchain.build_payload(payload)).await {
                 Ok(Ok(final_res)) => res = final_res,
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -478,6 +478,17 @@ impl Blockchain {
         // stale block is fine (geth/reth return best-so-far too), but
         // proposing an empty block while the mempool has transactions is not
         // (race (a) with no completed rebuild yet).
+        //
+        // TODO(#5011): two paths below can still put a full build on the
+        // critical path, so this narrows the exposure rather than removing it.
+        // First, the empty-payload case still rebuilds over the whole mempool;
+        // that has to stay correct (it is the race this guard exists for), but
+        // a large mempool of txs that never make it into a block — low-fee or
+        // nonce-gapped — can make even that approach the deadline. Second, this
+        // rebuild is not cancellable: unlike the in-loop build it runs as a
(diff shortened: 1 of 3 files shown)


ElFantasma (approved):

edg-l (approved):

crates/blockchain/payload.rs
```diff
@@ -467,7 +467,20 @@ impl Blockchain {
...
+        // full rebuild of a large block can exceed that deadline on its own —
+        // under sustained mempool inflow `tx_seq() > last_built_seq` is true
+        // on essentially every call, so an unconditional rebuild here puts one
+        // whole extra block build on the proposal critical path. In that case
+        // only rebuild when the payload we hold is empty: returning a slightly
+        // stale block is fine (geth/reth return best-so-far too), but
+        // proposing an empty block while the mempool has transactions is not
+        // (race (a) with no completed rebuild yet).
+        if self.mempool.tx_seq() > last_built_seq
+            && (!cancel_token.is_cancelled() || res.payload.body.transactions.is_empty())
```
MegaRedHand, line 493: I think the `!cancel_token.is_cancelled()` here is unnecessary, since once we leave the loop above the token was cancelled or the build is for an old slot.

MegaRedHand (approved):

crates/blockchain/payload.rs
ilitteri: The check isn't redundant as written — the loop has two exits that leave the token uncancelled: the `while start.elapsed() < SECONDS_PER_SLOT` condition failing, and the `_ = tokio::time::sleep(remaining) => break` arm. So dropping it wouldn't be a no-op; it would stop the slot-timeout path from rebuilding on new txs, which is race (b) in the comment above.

Keeping it deliberately, to confine the behaviour change to the `getPayload` path this PR is about. But your instinct points at something real that I'd missed, and it argues your way: **this final rebuild isn't cancellable.** Unlike the in-loop build it runs as a plain `spawn_blocking(..).await` rather than through `run_until_cancelled`, so a `getPayload` arriving while a slot-timeout rebuild is in flight waits for it to finish. The latency hazard therefore isn't fully confined to the cancelled path, and gating on `is_empty()` alone would shrink the worst case further than this PR does.

I've left that out of scope here and recorded both remaining paths under the existing `TODO(#5011)` in `4575e3b57` — the other being that the empty-payload case still rebuilds over the whole mempool. Happy to do the simplification instead if you'd rather take the timeout-path behaviour change in this PR.

PR #7022, opened 2026-07-22
fix(l1): stop wasting full sync cycles on the ePBS FCU/newPayload ordering race

ilitteri:
**Motivation**

On glamsterdam-devnet-7 every ethrex node paired with an ePBS consensus client burns a 12–17s peer-download sync cycle roughly every 3–4 minutes (measured on `nimbus-ethrex-1`: 100 cycles per 6h window, ~22 minutes of wasted peer traffic, each ending in `tried to insert a state_root that's already inserted` after re-downloading and re-executing a block the engine had already applied). Each episode also answers the consensus client with SYNCING, pushing it toward optimistic (`NOT_VALIDATED`) head status, which degrades attestation and proposal behavior.

The trigger is an ordering race that ePBS makes routine rather than exceptional: the consensus client learns the payload's block hash from the *bid* and moves its forkchoice head there before the payload itself reaches the EL through `engine_newPayload` (envelope gossip lags the block). ethrex reacts to the FCU-for-unknown-head by immediately starting a full sync cycle — but peers cannot serve the block either, because the payload has not propagated anywhere yet. Measured over 6h: 132 of 132 SYNCING forkchoice heads arrived through `engine_newPayload` 10–13s later; none were servable by peers earlier.

**Description**

Two independent layers:

1. `sync_cycle_full` ends as soon as the head is executed locally. New predicate `sync_head_executed` (header known + post-state on disk), checked at cycle entry, between header-fetch retries, and before the body-download pipeline. A cycle that raced `engine_newPayload` now stops the moment the engine applies the block, instead of re-downloading headers/bodies and re-executing it.

2. The sync manager gives `engine_newPayload` a bounded window before starting a cycle at all. When the node's latest executed block is recent (≤30s old and above genesis — i.e. it was following the tip moments ago), the manager polls up to 15s for the forkchoice head to become locally executed before launching a peer-download cycle, re-reading the head each poll so newer FCUs are accounted for. Cold starts, restarts, and genuinely-behind nodes skip the wait entirely, so initial-sync latency and the hive sync suites are unaffected; the wait is bounded, so a head that never arrives through the engine still gets a real cycle.

Also adds `sync_cycles_started` to `SyncDiagnostics`: a node following the tip should see it stay flat, making this failure mode observable in production.

Regression tests drive a real `SyncManager` over an in-memory chain:

- `fcu_before_newpayload_race_starts_no_sync_cycle` — the production race costs zero sync cycles (fails on the unfixed manager with `sync_cycles_started = 1`);
- `cold_node_starts_sync_cycle_without_heal_delay` — the wait does not apply off-tip;
- `missing_head_still_starts_sync_cycle_after_heal_wait` — the wait is bounded;
- `sync_head_executed_reflects_local_state` — predicate unit coverage.

**Checklist**


crates/networking/p2p/sync/full.rs
```diff
@@ -407,6 +448,16 @@ pub async fn sync_cycle_full(
         store.add_fullsync_batch(block_headers).await?;
         single_batch = false;
     }
+    // Last chance before paying for body downloads and execution: the head having
+    // post-state means the whole chain below it is executed too, so the work this
+    // cycle was about to do already happened through the engine.
+    if sync_head_executed(&store, fcu_head_hash)? {
+        info!(
+            ?fcu_head_hash,
+            "Sync head executed via engine_newPayload during header fetch; ending sync cycle"
```
ElFantasma, line 454: nit: this check runs after the header-fetch loop completes, right before body downloads — the `"during header fetch"` phrasing (copied from the sibling check at line 273) no longer matches. Consider `"before body download"` or dropping the phase qualifier entirely.

ElFantasma (approved):
One non-blocking log-message nit inline; core fix and tests look solid.

edg-l:
Went through this end to end. The two-layer approach makes sense and the devnet numbers justify it. One blocker plus a few smaller things.

### 1. `was_recently_at_tip` keys off the canonical head instead of the executed head (blocking)

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/crates/networking/p2p/sync_manager.rs#L294-L307

`get_latest_block_number()` returns `latest_block_header`, which is the canonical pointer. The doc you added on `executed_head` in this same PR says that pointer can run ahead of executed state, and full sync already logs exactly that situation:

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/crates/networking/p2p/sync/full.rs#L430-L438

So a node whose FCU canonicalized past its state has a fresh canonical timestamp with nothing behind it, `was_recently_at_tip` returns true, and we put 15s in front of a cycle that genuinely has work to do. That is the inverse of what the gate is for. Checking that the state is actually there should be enough:

```rust
let Ok(Some(header)) = store.get_block_header(latest) else {
    return false;
};
if !matches!(store.has_state_root(header.state_root), Ok(true)) {
    return false;
}
```

### 2. Nothing re-checks the head during body download and execution

The guards sit at cycle entry, between header retries, and right before the pipeline. But the cycles you measured were 12 to 17s and header fetch is the cheap part; most of that is bodies plus execution, which is precisely the 10 to 13s window the payload lands in. A cycle that clears the third guard by a few hundred ms still pays for the whole thing.

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/crates/networking/p2p/sync/full.rs#L583-L603

A `sync_head_executed(&store, fcu_head_hash)?` at the top of that loop is one header read plus one trie root lookup per batch, and it covers the part of the cycle that actually costs something.

### 3. Two of the guards log the same line, and one of them is wrong

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/crates/networking/p2p/sync/full.rs#L270-L276

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/crates/networking/p2p/sync/full.rs#L454-L460

Byte identical strings, so from logs you cannot tell which one fired. The second runs after the header loop finished and the batches were persisted, so "during header fetch" is misleading there. Something like "after header fetch, before body download" would do it.

### 4. `missing_head_still_starts_sync_cycle_after_heal_wait` passes whether or not the wait ran

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/test/tests/p2p/sync_race_tests.rs#L190-L214

It only asserts a cycle starts within 20s, which is also true if the wait was skipped entirely. And it will be skipped if CI is slow enough that 30s elapse between `chain_at_fresh_tip()` and the assert, since block 1 is stamped with `now_secs()` and `RECENT_TIP_MAX_AGE_SECS` is 30. Asserting the negative first would pin the actual property:

```rust
tokio::time::sleep(Duration::from_secs(5)).await;
assert_eq!(sync_manager.get_sync_diagnostics().await.sync_cycles_started, 0);
```

### 5. The new tests sleep around 25s of wall clock

All three are `multi_thread`, so `tokio::time::pause()` is out. There is already a precedent in sync.rs for making a constant tunable under a test feature:

https://github.com/lambdaclass/ethrex/blob/aa20677589c74611643dabd010656fdbff29b409/crates/networking/p2p/sync.rs#L47-L55

Giving `NEWPAYLOAD_HEAL_WAIT` and `NEWPAYLOAD_HEAL_POLL` the same treatment would let these run against a 1s window instead of 15, which also brings the flake risk down.

Rest looks good to me. The comment explaining why canonicality is deliberately not required in `sync_head_executed` is appreciated, that would have been a confusing read otherwise. Also worth a rebase, the branch is 7 commits behind main.

MegaRedHand (approved):

edg-l (approved):

PR #7024, opened 2026-07-22
feat(l1): optional historical chain backfill + DB observability

ilitteri:
**Motivation**

A snap-synced ethrex node keeps only block *headers* below the sync pivot, so historical RPC queries for pre-pivot blocks (`eth_getBlockByNumber`, `eth_getBlockReceipts`, `eth_getTransactionByHash`, `eth_getLogs`, …) return empty. This PR adds an opt-in way to backfill the missing bodies and receipts so the node can serve those queries, plus the DB metrics needed to observe the backfill (and the RocksDB store in general) while it runs.

**Description**

Historical chain backfill (opt-in, off by default):

- `--history.chain {off|postmerge|all|BLOCK}` — after initial sync, backfill block bodies + receipts below the snap pivot, down to the merge (Paris) block (`postmerge`), as far back as receipts are decodable (`all`, best-effort), or down to an explicit block number so an operator can keep just a recent slice of history. `off` keeps today's headers-only behavior.
- Every mode is clamped at the **Byzantium** block: before EIP-658 a receipt's first field is a post-state root rather than a status flag, a form `Receipt` cannot represent, so those receipts can never be fetched. `all` therefore means "down to Byzantium", not genesis.
- `--history.transactions N` — how far back to keep the transaction-lookup index (mirrors geth's flag; `0` = the whole backfilled range).
- Runs as a low-priority background task: reverse-fills one bounded batch at a time, validates bodies/receipts against the already-synced headers (recomputing the logs bloom eth/69 onward omits), and tracks progress in `earliest_block_number` (the frontier). Receipt requests use the form the connection negotiated — eth/68/69 take the original `GetReceipts`, while eth/70 (EIP-7975) replaced it with a paginated form that eth/71 also uses; a trailing partial block is dropped so every stored block stays root-verified.
- Crash-safe: each batch — bodies, receipts, tx-index entries and the frontier — is committed as a single atomic RocksDB write, so an interrupted backfill resumes from the persisted frontier with no holes. A one-time startup reconciliation corrects a stale frontier (e.g. a node that snap-synced before this feature existed).
- Not an archive node: this restores historical *chain* data (blocks/txs/receipts/logs), not historical *state*.

`earliest_block_number` now means "lowest block with a stored body" and is set (as a lower bound) on every snap sync, not only when backfill is enabled. That touches five existing readers: it **corrects three** — `export` reads it precisely to clamp and warn about pruned blocks, `eth_feeHistory` reaches into `body.transactions` and so was already asking for bodies a snap node doesn't have, and fork choice gets a tighter walk-back bound — and **changes two**: `eth_getBlockByNumber("earliest")` now returns the pivot rather than genesis on a snap-synced node, and `eth_syncing`'s `startingBlock` follows the backfill frontier (it descends as backfill progresses rather than marking where the current sync attempt started).

DB observability:

- A storage-backend `db_stats()` snapshot: per-column-family live/total SST, estimated live data, key/file counts, blob + memtable bytes, pending-compaction debt; plus DB-wide block-cache usage/capacity/pinned, running compactions, and — behind `ETHREX_ROCKSDB_STATISTICS` — block-cache hit/miss.
- New `ethrex_db_*` Prometheus gauges (including the history-backfill frontier), populated by a 15s collector spawned when `--metrics` is enabled. No-op for the in-memory backend.
- A provisioned Grafana dashboard (`db_dashboard.json`) covering per-column-family sizes, growth over time, the backfill frontier vs the merge floor with speed/ETA/on-disk-size panels, and read/compaction internals.

Docs: new "Historical chain backfill" and "DB observability" pages (plus a pointer from the Sync modes page), and the CLI reference regenerated for the new flags.

Note on scope: an experiment to parallelize the backfill (concurrent fetches across multiple peers) is present in the branch history and then **reverted**. Measurements on a mainnet node showed throughput is bounded by how fast peers serve post-history-expiry data, not by client-side concurrency, so the branch ships the simple serial backfill. The reverted commits are kept for the record.

**Validation**

A mainnet node completed a full `--history.chain postmerge` run: **9,993,456 blocks** backfilled from its sync pivot down to the merge block (25,530,850 → 15,537,394) in about **14 days**. The resulting data was then verified three ways.

*1. Root re-derivation over 100% of the range.* Every backfilled block had its `transactionsRoot`, `ommersHash`, `withdrawalsRoot` and `receiptsRoot` recomputed from the bytes on disk and compared against the consensus-validated header:

```
range              15537394..=25530850
blocks checked     9,993,457
missing header     0
missing body       0
BAD BODY ROOTS     0
BAD RECEIPT ROOTS  0
elapsed            8168s (136.1 min)
```

This is stronger than the fetch-time validation the backfill already does, because it recomputes from what is stored *now* and so also rules out post-write corruption. It ran against a hard-link snapshot of the datadir, so the node kept serving throughout; total downtime for the whole exercise was ~25s across two brief stops to take the snapshot. The tool is `tooling/history_validator`.

*2. Independent cross-check.* 3,000 blocks sampled uniformly over the backfilled range were compared against a public mainnet node on block hash, `receiptsRoot` and the full ordered transaction-hash list: **0 mismatches across 562,625 transactions**. Zero defects in 3,000 samples bounds the defect rate below 0.1% at 95% confidence, which the exhaustive check above then settles outright.

*3. Historical RPC.* On block 16,000,000 (deep pre-pivot, previously empty): `eth_getBlockByNumber` returns the block (211 txs), `eth_getBlockReceipts` 211 receipts, `eth_getTransactionByHash`/`eth_getTransactionReceipt` resolve, and `eth_getLogs` returns 1,404 logs over a 6-block range.

**Measured cost** (now in the hardware-requirements page, replacing the earlier provisional estimate): ~105 KiB per block, ~1.0 TB of history on top of ~459 GB of state, 1,490 GiB total. Note this came in *below* the mid-run extrapolation of ~125 KiB/block, because the blocks sampled then were the newest and largest. Backfill inverts where the database spends space: ~82% state / ~18% history becomes ~31% state / ~69% history. `all` remains unmeasured.

Rebased onto `main`: no longer stacked on #7013, and the diff is only this feature.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A: no new column families and no on-disk format change; backfill writes to existing CFs and repurposes `earliest_block_number`.


edg-l (commented):
Read through this with the eth wire specs open. One blocker, then smaller stuff. Leaving it as a comment rather than a change request since it's still draft.

The backfill design itself holds up well: atomic per-batch write of data plus frontier, top-down header ordering so a truncated response stays contiguous, the `min(bodies, receipts)` prefix. Crash-resume reasoning is sound.

## eth/71 receipt requests use the wrong wire format

This comment is out of date:

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/peer_handler.rs#L641-L646

EIP-8159 (eth/71) declares `requires: [7928, 7975]`, and 7975 is eth/70. eth/71 doesn't skip eth/70, it builds on it. That's what #7021 fixed on main four days ago, and that commit is already in this branch's merge base.

So the eth/68 form is wrong for eth/71 peers:

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/peer_handler.rs#L656-L664

`Message::encode` only uses the negotiated version for the code offset; the payload comes from the enum variant. The peer decodes 0x0f as `GetReceipts70` (`message.rs:249-252` matches `V70 | V71`) and errors on the missing `firstBlockReceiptIndex`.

The response path fails the same way. An eth/71 peer's reply decodes locally as `Receipts70`, which isn't matched, so it falls to the catch-all and the peer gets penalized and marked disposable:

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/peer_handler.rs#L674-L683

The capability list doesn't guard against this. `get_random_peer` matches peers advertising any of the listed caps, but the connection negotiates the highest mutual version (`connection/server.rs:1121-1145`) and we advertise up to 71. A peer selected via `eth(68)` that also speaks 71 is on 71. On mainnet today that's most of them, so receipt fetches fail across the board and burn through the peer table on the way.

Fix is to branch on the negotiated version: send `GetReceipts70 { first_block_receipt_index: 0, .. }` for V70/V71 and accept `Receipts70` back. `PeerConnection` needs to expose the negotiated cap; the state is already tracked at `connection/server.rs:267`.

One detail on the response side. When `lastBlockIncomplete == 1` the trailing list is partial and can't be checked against the header root, and EIP-7975 asks for a tx-count check, a per-receipt size check against gaslimit/8, and a total size bound. For a backfill task I'd just drop the trailing partial list and treat the response as a shorter prefix. The `min(bodies.len(), receipts.len())` logic already handles short prefixes, so every block stays root-verified, no new validation path, no buffering unvalidated input. Costs one block per truncated response, which is nothing over millions.

## earliest_block_number changes for everyone, not just backfill users

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/sync/snap_sync.rs#L659-L666

This runs on every snap sync regardless of `--history.chain`, and four existing callers read that value:

- `block_identifier.rs:41`, so `eth_getBlockByNumber("earliest")` returns the pivot instead of genesis. geth returns genesis there even when snap-synced, and our genesis is fully stored, so the old answer wasn't wrong.
- `eth/client.rs:114`, `eth_syncing.startingBlock`
- `fee_market.rs:220`, the `eth_feeHistory` lower bound
- `fork_choice.rs:250`, the walk-back bound in `find_link_with_canonical_chain`

Faster for fork choice on a snap node, but it's a user-visible RPC change riding along with an opt-in feature. Worth calling out in the description at minimum.

Separately, it's an unconditional set rather than a floor. `min(current, pivot)` would stop a second snap cycle from pushing the frontier above blocks that already have data. Reachable on pre-merge networks with no `merge_netsplit_block`, where `is_synced` is hardcoded false (`sync_manager.rs:71`), and `reconciled` is a once-per-process latch so nothing corrects it until restart.

## Receipts batch size

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/sync/backfill.rs#L115-L117

128 mainnet post-merge blocks of receipts is well past the 10 MiB soft limit; that overflow is the whole reason EIP-7975 exists. Truncation is handled, but you're paying a 128-hash request to get a handful of blocks back each round. Probably wants its own smaller constant separate from the bodies limit.

## Smaller things

The floor gets recomputed every batch, roughly 25 header reads every 500ms for a value that never moves:

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/sync/backfill.rs#L231-L246

Hoisting it into `run_history_backfill` next to `reconciled` would fix it. Same for the `format!` on `config.mode` a few lines down, that string never changes.

If bodies come back but receipts don't, the bodies are thrown away and refetched next round:

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/networking/p2p/sync/backfill.rs#L268-L276

Requesting receipts first makes the discard cheaper.

The metrics collector has no cancellation token, and every other long-lived task in that file takes one:

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/cmd/ethrex/initializers.rs#L159-L195

It's also doing about nine property reads per CF synchronously on the runtime every 15s. Cheap individually, but `spawn_blocking` would match the rest of the storage layer.

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/storage/backend/rocksdb.rs#L101-L107

`is_ok()` means `ETHREX_ROCKSDB_STATISTICS=0` turns statistics on. Worth parsing the value.

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/storage/backend/rocksdb.rs#L374-L376

Levels 0..=6 is right for the default `num_levels = 7`, but silently undercounts if that ever gets tuned.

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/crates/blockchain/metrics/db.rs#L119-L131

Cumulative counters exported as `IntGauge` with a `_total` suffix, which Prometheus reserves for counters. Either drop the suffix or switch to `IntCounter`.

https://github.com/lambdaclass/ethrex/blob/5e177a5bfd13a95e094ac0b6da076aee74a34b41/Cargo.toml#L161-L170

From #7013, still open. Needs a rebase before merge so the diff is only this feature.

## Tests

The pure and storage-level parts are well covered. The bisect tests and the four `add_backfilled_blocks` cases are good, the gap-free resume one especially.

Nothing covers `backfill_step` or `request_receipts` though, so the truncation prefix, the `min(bodies, receipts)` pairing, receipts-root rejection, and per-version request encoding are all untested. A GetReceipts encode/decode matrix across V68/V69/V70/V71, like the ones #7021 added at `message.rs:499-537`, would have caught the wire format issue.

## One thing that's better than the comment claims

`first_pos_block` describes `difficulty == 0` as reusing the same discriminator the block validator uses. It's stronger than that. EIP-3675 pins `difficulty` to the constant 0 from `TRANSITION_BLOCK` onward and makes `difficulty != 0` a validity failure, so the predicate is monotonic by consensus on any TTD-merged chain and the bisect precondition holds. Worth citing 3675 in the doc comment instead.

pablodeymo:
Went through this independently with @edg-l's review already in hand. Confirming their eth/71 blocker, and I found a second one in the same area. CI is green across the board; the design and the docs are genuinely good — the atomic batch + frontier-as-resume-cursor reasoning holds, and the durability section is the clearest writeup of that pattern I've read in this repo.

## Blocker: ethrex cannot decode pre-Byzantium receipts, so `--history.chain all` has a hard wall at block 4,370,000

`Receipt` has no field for the pre-EIP-658 post-state root:

```rust
pub struct Receipt {
    pub tx_type: TxType,
    pub succeeded: bool,     // <-- EIP-658 status byte
    pub cumulative_gas_used: u64,
    pub logs: Vec<Log>,
    pub payer: Option<Address>,
    pub frame_receipts: Option<Vec<FrameReceipt>>,
}
```

Before Byzantium (mainnet block `4_370_000`), field 0 of a receipt is a 32-byte post-state root, not a status. `Receipt::decode_inner_with_bloom` decodes it as a `bool`, and `bool::decode_unfinished` (`crates/common/rlp/decode.rs:41-55`) accepts only `0x80` and `0x01` — a `0xa0`-prefixed 32-byte string returns `RLPDecodeError::MalformedBoolean`. So `Receipts68::decode` fails on every pre-Byzantium block, in every eth version, and there's no encoding of a pre-Byzantium receipt that ethrex could round-trip even if it decoded.

This is new surface: full sync derives receipts by execution, so `request_receipts` is the first path in the client that decodes receipts off the wire.

Effect in `all` mode: backfill descends from the merge to 4,370,000 fine, then every `request_receipts` call fails and every peer that answered *correctly* gets `record_failure` + `set_disposable`. On top of that, the failure is a codec error, which brings me to:

**The decode failure also kills the peer's inbound listener.** `codec.rs:232` returns the `Message::decode` error to `FramedRead`, and tokio-util 0.7.19 terminates the stream on the poll after a decoder error (`framed_impl.rs:161-168`, `if state.has_errored { ... return Poll::Ready(None) }`). So this in `connection/server.rs:879-887`:

```rust
Err(e) => {
    debug!(error=?e, "Error receiving RLPx message");
    // Skipping invalid data
    None
}
```

doesn't skip and continue — the next poll yields `None`, the `spawn_listener` stream ends, and the connection is read-dead until a ping notices. That comment is wrong today (pre-existing, and worth a separate issue), but it means both this finding and @edg-l's eth/71 one cost a whole connection per attempt, not a penalty point.

The docs claim the graceful version of this:

> **`all`**: ... this can stall at a block it cannot fetch (it reports the stall rather than failing the node).

What actually happens is silent peer-table attrition with a `warn!` every 10s.

The floor is already in the chain config, so the fix is small:

```rust
HistoryChain::All => store
    .get_chain_config()
    .byzantium_block
    .unwrap_or(0),
```

(mainnet genesis already carries `byzantiumBlock: 4370000`.) Then document `all` as "down to Byzantium — pre-Byzantium receipts use the pre-EIP-658 format ethrex does not represent", rather than "down to genesis, best-effort".

## Confirming the eth/71 blocker

Independently reached the same conclusion. `message.rs:249-252` and `260-263` both gate on `V70 | V71`, so an eth/71 peer decodes code `0x0f` as `GetReceipts70` and needs the `firstBlockReceiptIndex` our eth/68-form request doesn't carry. And `request_receipts_inner`'s match only accepts `Receipts68`/`Receipts69`, so an eth/71 peer's valid reply hits the catch-all and gets disposed. `SUPPORTED_ETH_CAPABILITIES` advertises through 71 (`p2p.rs:17-22`), so on mainnet this is most peers.

The doc page propagates the same misconception and needs the same fix:

> (Only the rare, skipped eth/70 — whose `GetReceipts` is paginated — is unused.)

eth/71 doesn't skip eth/70's receipt format; EIP-8159 `requires: [7928, 7975]` and 7975 *is* eth/70. Same as what #7021 fixed on main four days ago.

## A completed backfill never stops working

```rust
Ok(BackfillProgress::Complete | BackfillProgress::Waiting) => BACKFILL_IDLE_INTERVAL,
```

`Complete` sleeps 10s and loops, forever, for the life of the process. Each of those iterations re-runs `resolve_postmerge_floor`, which on mainnet (`merge_netsplit_block == None`) is a ~25-header bisect — so a finished node does ~150 header reads/minute and a `diagnostics.write()` indefinitely, having nothing left to do. @edg-l flagged the per-batch recompute; the post-completion case is the same cost with no end.

Worth splitting the two things `Complete` currently means, since only one of them is worth retrying:

- `frontier <= floor` → genuinely done, `return` from the loop after the `info!`.
- `resolve_postmerge_floor() == None` (chain hasn't merged) → keep idling, that can change.

## `--history.transactions` doesn't match the geth semantics it claims

Docs say it's "mirroring geth's flag of the same name" and that a non-zero `N` "indexes only blocks within `N` of the chain head". Two gaps:

1. **No pruning.** geth's flag also drops index entries that fall outside the window. Here nothing is ever unindexed, and blocks above the pivot are indexed unconditionally by the normal sync path — so any `N` smaller than `head - pivot` has no effect on the range where the index is actually large.
2. **The boundary drifts.** `head` is re-read every batch and `index_transactions` is computed as `head.saturating_sub(header.number) < horizon`. A mainnet `postmerge` backfill runs "on the order of weeks" by the doc's own estimate, so head moves ~50k blocks over the run and the cutoff moves with it. The indexed set isn't `[head - N, head]` for any single `head`.

Hoisting `head` once at task start (next to `reconciled`) makes it deterministic, and the no-pruning caveat belongs in the doc's Limitations section.

## Peers that legitimately lack expired history get a *critical* failure

`backfill_step` reasons that both responses are prefixes:

```rust
// Only take blocks that have BOTH a body and receipts; because both responses
// are prefixes of the top-down header list, their common prefix is contiguous
```

geth's `ServiceGetBlockBodiesQuery` skips hashes it doesn't have rather than truncating at the first miss, so a peer holding a *subset* of the requested range returns a non-prefix list. The good news: nothing corrupts — a misaligned body fails `validate_block_body` against the header's tx/ommers/withdrawals roots, and the one case that slips through (two blocks with byte-identical empty bodies) stores the correct bytes anyway. The cost is that `request_block_bodies` answers with `record_critical_failure`, and the peers being burned are exactly the partial-history peers `all` mode depends on. Post-history-expiry that's the common case, not the adversarial one.

## Two nuances on the `earliest_block_number` change

@edg-l lists four consumers; there's a fifth, and the picture is more mixed than "user-visible RPC change riding along":

- `cmd/ethrex/cli.rs:1232` (the `export` subcommand) reads it specifically to warn about pruned blocks and clamp the start — the new value is what it always wanted.
- `fee_market.rs` isn't just a lower bound: `calculate_percentiles_for_block` reads `block.body.transactions`, so with `earliest == 0` a snap node's `eth_feeHistory` was already reaching for bodies it doesn't have. `earliest == pivot` fixes that.

So it's a regression for the `earliest` tag, a fix for `feeHistory` and `export`, and a perf change for fork choice. Still deserves a line in the description, but the framing should be "corrects three call sites, changes one" — and the `min(current, pivot)` floor @edg-l suggests is right regardless.

## Test gaps beyond the ones already noted

Agreed that `backfill_step` and `request_receipts` are uncovered. Two more specific ones:

- `resolve_postmerge_floor` — the public entry point — has no tests; only the pure `first_pos_block` helper does. The `merge_netsplit_block` short-circuit is the branch Sepolia (`1735371`) and Hoodi (`0`) actually take, and it's the one that isn't exercised. Mainnet is the only bundled network that reaches the bisect.
- `reconcile_frontier` covers the snap-node and full-node shapes but not `head == 0` or the "no body at head" early return, both of which write a frontier value.

A `GetReceipts`/`Receipts` encode-decode matrix across V68/V69/V70/V71 (like `message.rs:499-537` from #7021) catches the eth/71 issue; a pre-Byzantium receipt fixture in that matrix catches the first one.

## Rebase

`Cargo.toml`'s `[patch.crates-io]` block is #7013, still open and not on main, so it's in this diff. Needs the rebase before merge as the description says.

ilitteri:
Thanks — both reviews were right on the wire-format blocker, and it was worse than "on rebase": #7021 is already in this branch's merge base, so receipt fetches are broken on the PR as it stands. Every finding is addressed below.

## Blockers

**Receipt wire format per negotiated version.** eth/70 (EIP-7975) replaced `GetReceipts` with the paginated form, and eth/71 (`requires: [7928, 7975]`) builds on eth/70 rather than skipping it — the comment claiming otherwise was stale, and the doc page repeated it. Both are corrected.

Since `Message::encode` only derives the code offset from the negotiated version, the caller has to choose the payload. Peer selection now returns the peer's advertised capabilities (`SelectedPeer`), and `negotiated_eth_version` mirrors the handshake's highest-mutual rule to pick the form and to accept `Receipts70`. Capability-based selection alone can't express this: a peer advertising eth/68 that also speaks eth/71 negotiates 71, so the version has to be derived rather than assumed.

The `lastBlockIncomplete` suggestion was the right call and is what's implemented — the trailing partial list is dropped and the response treated as a shorter prefix, so the existing `min(bodies, receipts)` pairing keeps every stored block root-verified with no second validation path and no buffering of unvalidated input.

**Pre-Byzantium receipts.** Confirmed exactly as described: field 0 is a 32-byte post-state root, the non-frame path decodes it as `succeeded: bool`, and `bool::decode_unfinished` accepts only `0x80`/`0x01`. Every mode is now clamped at `byzantium_block`, so `all` means "down to Byzantium", not genesis, and the docs say so.

One extension the reviews couldn't have seen: this branch also gained `--history.chain <BLOCK>`, which lets an operator aim at a pre-Byzantium floor *deliberately*. The clamp covers that path too, with a warning rather than a silent no-op.

Small correction for the record: the failure surfaces as `Custom("Error decoding field 'succeeded' … MalformedBoolean")` — the field decoder wraps it — so the new test asserts structurally instead of on the variant: two receipts identical except for field 0, the status-flag one decodes and the post-state-root one doesn't.

## Everything else

- **`Complete` looped forever.** Split into the two things it meant: reaching the floor now ends the task; `postmerge` on an unmerged chain keeps idling, since that can change.
- **Per-batch recompute.** Floor, tx-index reference head and mode label are resolved once per run in `BackfillPlan`. Pinning the head also fixes the drifting `--history.transactions` boundary.
- **`--history.transactions` semantics.** Both gaps documented: it governs only what backfill writes (nothing is ever un-indexed, and post-pivot blocks are indexed unconditionally by normal import), and the window is pinned at task start.
- **Receipts batch size.** 128 → 64, with a compile-time assert against the `GetBlockBodies` limit. Receipts bind, not bodies.
- **Fetch order.** Receipts first, so a failure discards less.
- **Partial-history peers.** Confirmed against geth: `ServiceGetBlockBodiesQuery` appends only what it finds, so the response is compacted rather than truncated. Both paths now keep the longest leading run that validates and charge a normal failure only when nothing validated. Everything stored is still root-verified.
- **`earliest_block_number`.** Now `min(current, pivot)`. Agreed on the framing — it corrects three call sites and changes one: `export` reads it precisely to clamp and warn, and `eth_feeHistory` isn't only a lower bound (`calculate_percentiles_for_block` reads `body.transactions`, so `earliest == 0` was already reaching for bodies a snap node doesn't have). The `earliest` tag is the one regression. Going into the description.
- **Statistics gating.** `is_ok()` meant `=0` enabled it; now parsed numerically.
- **Level count / `_total` suffix / collector lifecycle.** Levels behind a named constant; `_total` dropped (the values are RocksDB ticker absolutes, so a gauge is right and the suffix was reserved); collector now selects on a cancellation token and samples under `spawn_blocking`.
- **EIP-3675.** Good point that the predicate is monotonic *by consensus*, not convention — cited in the doc comment.

## Tests added

A `GetReceipts`/`Receipts` form matrix across eth/68–71 (the one that would have caught the blocker), a pre-Byzantium receipt fixture pinning why the clamp exists, the `merge_netsplit_block` short-circuit, and `reconcile_frontier`'s `head == 0` and no-body-at-head paths. 291 tests green, `clippy --all-targets -D warnings` clean.

`backfill_step` end-to-end and `request_receipts` against a live peer are still uncovered by unit tests — there's no mock-peer harness — so those remain validated on a live mainnet node rather than in CI.

## Still open

- **Rebase.** `[patch.crates-io]` traces to a branch-only commit from #7013; a rebase drops it. Not removing it by hand, since the workspace may need those forks.
- **The listener-termination bug is worth its own issue.** Confirmed: the decoder error reaches `FramedRead`, tokio-util ends the stream on the next poll, and the "Skipping invalid data" arm therefore doesn't skip — the connection goes read-dead until a ping notices. Pre-existing, but it's what turned both blockers into a lost connection per attempt rather than a scoring penalty. (Lockfile is tokio-util 0.7.18, not 0.7.19 — same behaviour.)

ilitteri:
Small correction to how I characterized the decode-error problem above: I framed it as the connection going read-dead until a ping notices. That's true but it isn't the defect — a peer that sends us a message we can't parse is a bad peer, so losing that connection is fine.

The actual defect is that the peer is neither penalized nor discarded. Tracing it through:

- The connection actor stays alive, so `PeerData.connection` remains `Some(..)` and the peer still looks connected.
- Nothing on that path records a failure or marks the contact disposable, so its **score is unchanged** — weighted selection doesn't merely still allow the peer, it keeps *preferring* it for the whole ~30–40s window (`PING_INTERVAL` 10s × `MAX_MISSED_PONGS` 3) while it cannot answer anything.
- When ping liveness finally reclaims it, `remove_peer` only `swap_remove`s the entry without setting `disposable`, so discovery can re-add and re-dial the same peer straight into the identical failure.

So the fix isn't to make the codec tolerant and skip the message, which is what I suggested — it's to treat the peer as bad: record the failure, mark it disposable, and stop the actor deliberately with a distinct disconnect reason so the drop is attributable rather than landing in the generic bucket.

Filed with that framing in #7035. Still out of scope for this PR and unchanged in substance for the review findings — it's the amplifier that made both wire-format mismatches expensive, not a cause of them.

ElFantasma (approved):

edg-l (approved):

crates/networking/p2p/sync/backfill.rs
```diff
@@ -0,0 +1,688 @@
...
+    // Don't compete with initial sync; wait until it has finished.
+    if snap_enabled.load(Ordering::Relaxed) {
+        return Ok(BackfillProgress::Waiting);
+    }
+
+    // One-time correction of `earliest_block_number` for nodes that synced
+    // before this feature existed (where it was left at genesis) or otherwise
+    // drifted from the true lowest-full-data block. Without this, backfill would
+    // see `frontier == 0` and conclude there is nothing to do.
+    if !*reconciled {
```
MegaRedHand, line 329: Can we represent this as a migration instead? That would turn this runtime check (which we pay from this day going forward), into a one-time migration. It would also allow us to simplify `reconcile_frontier`, which could be using the stored `earliest_block_number` instead of doing a bisect on the full block range each time.

crates/networking/p2p/sync/backfill.rs
```diff
@@ -0,0 +1,688 @@
...
+
+use crate::peer_handler::{MAX_BLOCK_BODIES_TO_REQUEST, PeerHandler};
+
+use super::{HistoryChain, SyncDiagnostics, SyncError};
+
+/// Resolve the floor block for `--history.chain postmerge`: the merge (Paris)
+/// activation block, i.e. the first proof-of-stake block.
+///
+/// ethrex has no per-network merge-block constant and mainnet merged by TTD
+/// (`merge_netsplit_block == None`), so this is a hybrid:
```
MegaRedHand, line 26: Can't we just set `merge_netsplit_block` for mainnet?

crates/networking/p2p/sync/backfill.rs
```diff
@@ -0,0 +1,688 @@
...
+/// there is no post-merge segment to backfill.
+///
+/// Precondition: the canonical header chain is present from genesis to head,
+/// which holds once snap sync has completed.
+async fn resolve_postmerge_floor(store: &Store) -> Result<Option<BlockNumber>, SyncError> {
+    if let Some(merge_block) = store.get_chain_config().merge_netsplit_block {
+        return Ok(Some(merge_block));
+    }
+    let head = store.get_latest_block_number().await?;
+    first_pos_block(head, |n| is_pos_block(store, n))
```
MegaRedHand: We should embed this closure into the function instead. The function is only called here anyways.

crates/networking/p2p/peer_handler.rs
```diff
@@ -608,21 +627,170 @@ impl PeerHandler {
...
-            let mut validation_success = true;
             for (header, body) in block_headers[..block_bodies.len()].iter().zip(block_bodies) {
                 if let Err(e) = validate_block_body(header, &body, &NativeCrypto) {
-                    debug!("Invalid block body error {e}, discarding peer {peer_id} and retrying");
-                    validation_success = false;
-                    self.peer_table.record_critical_failure(peer_id)?;
+                    debug!(
+                        "Block body for {} did not validate ({e}); keeping the {} verified before it",
+                        header.number,
+                        res.len()
```
MegaRedHand: Keep the log message a static, so it's easier to grep

```suggestion
                        err,
                        block_number=header.number,
                        bodies_kept=res.len(),
                        "Block body did not validate; keeping the bodies verified before it",
```

MegaRedHand (commented):
Left some suggestions

pablodeymo:
Re-reviewed after the two fix rounds. CI is green. Every item from my last pass is addressed, and I checked the two I couldn't take on faith:

- **eth/70 vs eth/71 request form.** `negotiated_eth_version` (`peer_handler.rs:95`) is a faithful mirror of the handshake: `state.capabilities` is the peer's raw `Hello` list (`connection/server.rs:1155`) and it reaches the peer table verbatim (`:822`), so `max` over the intersection with `SUPPORTED_ETH_CAPABILITIES` reproduces exactly what `exchange_hello_messages` computed. The V68–V71 encode/decode matrix pins it.
- **Pre-Byzantium clamp.** `resolve_floor` clamps every mode, including the new explicit-block one, and the receipt test asserts structurally (two receipts differing only in field 0) rather than on error text — that will survive a decoder refactor.

Also confirmed: `Complete` now ends the task, the floor/head/mode label are resolved once in `BackfillPlan` (which fixes the drifting tx-index window), batch is 64 with the compile-time assert, receipts are fetched first, `[patch.crates-io]` is gone, and #7035 covers the listener bug. All the new tests are in `cargo test --workspace`, so unlike the `ethrex-test` rocksdb-gated ones they actually run in CI.

Five things came out of the fixes themselves.

## The bodies path now never penalizes an invalid body — and rewards the peer

`request_block_bodies` (`peer_handler.rs:630-658`) drops the old `record_critical_failure` and only penalizes when *nothing* validated:

```rust
if !res.is_empty() {
    return Ok(Some(res));
}
// Nothing usable: penalize and re-roll onto another peer.
self.peer_table.record_failure(peer_id)?;
```

But `request_block_bodies_inner` already called `record_success` (`:601`) before any validation ran. So a peer answering a 128-hash request with `[valid_body_0, garbage×127]` nets **+1** every round trip, climbs to `MAX_SCORE`, and gets *more* traffic from the score-weighted selection. Previously that peer was pinned at `MIN_SCORE_CRITICAL` and effectively out.

This isn't scoped to backfill — `request_block_bodies` is the body-download path for full sync and for snap's pivot fetch, so the leniency I asked for on the backfill caller landed on the hot path too.

The two cases are distinguishable without giving up the tolerance: compute each remaining header's `(transactions_root, ommers_hash, withdrawals_root)` once, and when a body fails at position *k*, check whether it matches any header still in the request. If it does, it's a compacted response from a partial-history peer — stop, keep the prefix, no penalty. If it matches nothing, the peer invented it: `record_critical_failure`. That keeps the fix's intent and restores the teeth. (Cheap alternative: a `partial_ok: bool` on the request so only backfill gets the lenient path.)

## The empty-response fix went into receipts but not bodies

79083944c1 correctly reasons that an empty `Receipts` is spec-conformant for a peer holding none of the range (`:734`, soft penalty only). `request_block_bodies_inner` still does the old thing for the identical situation:

```rust
if !block_bodies.is_empty() && block_bodies.len() <= block_hashes_len { ... }
debug!("Didn't receive block bodies from peer, penalizing peer {peer_id}");
self.peer_table.record_failure(peer_id)?;
let _ = self.peer_table.set_disposable(peer_id);   // :607
```

geth's `ServiceGetBlockBodiesQuery` appends only what it finds, so a peer with no pre-merge bodies answers `BlockBodies { block_bodies: [] }` and we mark it disposable. That doesn't kill the live connection — `do_get_random_peer` doesn't look at `disposable` — but on the next prune tick the contact is dropped from its k-bucket and a replacement is promoted (`peer_table.rs:1249`, driven by the discovery server's `PRUNE_INTERVAL`), and it stops being revalidated (`is_validation_needed` returns false for disposable contacts). So the peer serves head-following until the connection happens to drop, and then it's no longer in the routing table to re-dial.

Since `all` mode draws a fresh random peer for the bodies request each batch, it walks the table doing this to honest peers — the churn 79083944c1's own commit message argues against, on the other half of the request pair. Same split as receipts: empty → soft, oversized → disposable.

## "Waits until initial sync finishes" is really "waits until snap finishes"

`backfill_step`'s gate is `snap_enabled` (`backfill.rs:290`). That flag is false whenever the node isn't snap-syncing — including the auto-switch in `sync_manager.rs:73` ("Node has synced state, switching to full sync"), which fires on **every restart** of an already-synced node, and `--syncmode full` generally. So a snap-synced node restarted 300k blocks behind starts backfilling immediately while it is full-syncing hard toward the head, competing for the same peers. The docs' "it waits until initial sync finishes … does not slow down block processing" only holds for the first-ever sync.

Same gate also fixes `plan.head` (`:322`) to whatever the stale head was at that moment, so the `--history.transactions` window is anchored below the real head — the drift the plan was introduced to remove, in the other direction.

Not severe on its own — the cost is peer/bandwidth competition during a catch-up, and the head-pinning only bites if `--history.transactions` is non-zero — but both go away by gating on `blockchain.is_synced()` (already used in `connection/server.rs`), which is also the predicate the doc describes. `SyncManager::new` has the `Blockchain` handle to thread in.

## The frontier invariant is off by one: the pivot block has no receipts

The docs state *"every block in `[frontier, head]` has full bodies and receipts"*, and after 4a4ed661e6 the frontier lands exactly on the pivot. But snap sync fetches only the pivot's **body** (`snap_sync.rs:633`) and `add_block` (`:640`) writes header/body/`BLOCK_NUMBERS`/tx-locations — no receipts, and nothing else fills them: backfill's batch is `[batch_lo, frontier - 1]`, strictly below the frontier.

Net effect on a fresh snap node, for the pivot block only:

- `eth_getTransactionByHash(tx)` → the tx (it *is* indexed by `write_block_data`)
- `eth_getTransactionReceipt(same tx)` → `null`
- `eth_getBlockReceipts(pivot)` → `[]`, indistinguishable from an empty block

Nothing later covers it: `add_receipts` has no callers anywhere in the tree, the only other receipt writer is the layered flush of *executed* blocks, and full sync won't re-execute the pivot — `is_resume_point` is `canonical && has_state_root`, both true for the pivot, so the walk-back stops there and execution starts at pivot+1.

This is pre-existing, but it's this PR that advertises that block as the lowest full-data block. Cheapest fix is to make the first batch inclusive of the frontier (`batch_lo..=frontier`) so backfill picks up the pivot's receipts on its way down; otherwise the invariant needs an "except the pivot's receipts" caveat.

## A stall is documented as reported, but nothing reports it

The docs say `all` "can stall at a block it cannot fetch (it reports the stall rather than failing the node)" twice, and the log table lists `warn` only for `History backfill step failed`. A genuine stall — no peer serving that height — returns `BackfillProgress::Waiting` from `request_receipts`/`request_block_bodies` returning `None`, which logs *nothing* and sleeps 10s forever. It's indistinguishable from a healthy idle unless someone is watching `ethrex_db_backfill_frontier_block` flatline. A `warn!` after N consecutive non-advancing steps, or a `backfill_stalled` field next to `backfill_complete` in `SyncDiagnostics`, would make the doc true.

Related and worth one line in Limitations: `first_block_receipt_index` is hardcoded to `0` and the incomplete tail is dropped, so a block whose receipts alone exceed the 10 MiB soft limit can never be fetched — backfill would stall there permanently rather than paginate through it. Not reachable at today's gas limits, but that block is precisely why EIP-7975 added the parameter.

## Nits

- `negotiated_eth_version` re-derives what the connection already knows: `Established.negotiated_eth_capability` is computed at handshake and then dropped on the floor instead of being handed to `new_connected_peer`. Propagating it would make the two copies impossible to desync; right now a future change to the handshake rule silently breaks the wire form with no test coverage on the peer-table side.
- `ETHREX_ROCKSDB_STATISTICS=true` parses as non-numeric and silently disables statistics. Numeric-only was the right call for `=0`, but `true`/`yes` are what an operator will reach for — worth accepting them or logging when the value is unparseable.

ilitteri:
Second round pushed (`77c3a8a`..`e2c113b`). Every finding from the re-review is addressed; two suggestions I'd argue against rather than quietly skip, with reasoning below.

## Fixed

**The bodies path rewarded bad peers.** This one was mine, from the previous round's leniency fix, and it was worse than described: `record_success` fired in `request_block_bodies_inner` before any validation ran, so a peer answering with one valid body and 63 invented ones netted +1 per round trip and climbed to the maximum score. Success is now recorded by the caller, after validation. Tolerance and teeth coexist by distinguishing the two cases: on a mismatch at position `k`, check whether that body validates against any header still in the request. If it does, the peer omitted blocks it lacks and the response is compacted, so keep the verified prefix and charge nothing; if it matches no requested header, the peer invented it and earns a critical failure.

**The frontier was off by one.** Confirmed exactly as described: snap stores the pivot's body but never its receipts, and the batch was strictly below the frontier so nothing could repair it. The first batch of a run now includes the frontier block itself. Re-fetching one body is the price of making the invariant true.

**The gate was wrong.** `snap_enabled` is clear on every restart of an already-synced node and under `--syncmode full`, so a node restarted behind the head would backfill while full-syncing toward the tip and pin the tx-index head to a stale value. Now gated on `Blockchain::is_synced`, which is what the docs already claimed.

**Empty bodies no longer dispose the peer**, matching what the receipts path already did; returning more bodies than requested still does, since that is a protocol violation rather than a partial-history peer.

**Stalls are now reported.** After several consecutive non-advancing steps the task warns once and sets `backfill_stalled` in the sync diagnostics, and logs again when progress returns. The 10 MiB single-block pagination limit is now a documented limitation.

**Cache hit/miss are counters.** They were gauges holding RocksDB's absolute tickers while the dashboard graphed them with `rate()`, so every DB reopen made the series step backwards. Now `IntCounter`s advanced by delta, treating a decrease as a reset, with `_total` restored since the type earns it.

Also adopted: the one-use helper inlined, the log message made static with structured fields, and `ETHREX_ROCKSDB_STATISTICS` now takes `1`/`0` or `true`/`false` and warns rather than staying silent on anything else.

## Two I'd push back on

**Setting `merge_netsplit_block` for mainnet would break peering.** It is not inert config: `gather_forks` (`genesis.rs:675`) feeds `ForkId::new` (`status.rs:58`), which goes out in the `Status` handshake. Setting it on mainnet inserts an extra fork block into our fork-ID hash chain, giving us a different fork ID than every other mainnet client, and peers would reject us. The difficulty bisect avoids that entirely and costs ~25 header reads once per process.

**Reconciliation as a migration trades away its main property.** A migration runs once and bakes in whatever it computed at that moment — including for a node that happened to be mid-sync, which is exactly the case `reconcile_frontier`'s "no body at head" guard exists for. The runtime check is idempotent and self-corrects on every start. On the cost concern: it is once per process, not per batch, so the bisect is ~25 reads at startup rather than an ongoing tax. Happy to revisit if you still prefer the migration, but the current shape is deliberate.

## Validation

The reference mainnet run finished: 9,993,456 blocks backfilled to the merge block in ~14 days, then verified three ways — full root re-derivation over all 9,993,457 blocks from disk (0 bad roots, 0 missing, 136 min), a 3,000-block cross-check against an independent public node (0 mismatches across 562,625 transactions), and historical RPC on a deep pre-pivot block. Details and the measured disk figures are in the PR description; the re-derivation tool is `tooling/history_validator`.

Measured cost came in *below* the mid-run extrapolation, ~105 KiB/block against ~125 KiB/block, because the blocks sampled mid-run were the newest and largest.

## Still not covered

`backfill_step` and `request_receipts` have no unit tests, so the compacted-vs-fabricated discrimination above is exercised only in production. That needs a mock-peer harness, which does not exist in the tree; worth its own issue rather than being smuggled into this PR.

tooling/history_validator/src/main.rs
```diff
@@ -0,0 +1,136 @@
...
+                let Ok(Some(header)) = store.get_block_header(n) else {
+                    counts.missing_header.fetch_add(1, Ordering::Relaxed);
+                    continue;
+                };
+                let Ok(Some(body)) = store.get_block_body(n).await else {
+                    counts.missing_body.fetch_add(1, Ordering::Relaxed);
+                    continue;
+                };
+                // Covers transactions_root, ommers_hash and withdrawals_root.
+                if validate_block_body(&header, &body, &crypto).is_err() {
```
MegaRedHand: Our `validate_block_body` function checking that the block body has no uncles, which breaks in pre-merge blocks. Also, our `validate_block_body` doesn't check the `ommers_hash`

crates/networking/p2p/peer_handler.rs
```diff
@@ -608,21 +643,198 @@ impl PeerHandler {
...
+                    self.peer_table.record_failure(peer_id)?;
+                    return Ok(None);
+                }
+                if receipts.len() > block_hashes_len {
+                    debug!("Received oversized receipts from peer {peer_id}, penalizing");
+                    self.peer_table.record_failure(peer_id)?;
+                    let _ = self.peer_table.set_disposable(peer_id);
+                    return Ok(None);
+                }
+                self.peer_table.record_success(peer_id)?;
```
MegaRedHand: Here we're recording a success before validating the receipts

crates/networking/p2p/sync/snap_sync.rs
```diff
@@ -655,6 +656,19 @@ pub async fn snap_sync(
...
+    // pivot is headers-only. Record the true frontier — the lowest block with a
+    // stored body — so RPC (`earliest` tag, feeHistory) reflects what is
+    // actually available instead of genesis. Historical backfill, when enabled,
+    // lowers this frontier further as it fills bodies/receipts downward.
+    // Recomputing rather than writing the pivot keeps the value a lower bound:
+    // a later snap cycle must not push the frontier above blocks that already
+    // have bodies (e.g. backfilled ones), and a fresh node must not keep the
+    // genesis-init placeholder (0) that `min` with the pivot would preserve.
+    store
+        .update_earliest_block_number(reconcile_frontier(store).await?)
```
MegaRedHand, line 670: Calling `eth_getBlockByNumber` with `"earliest"` as a parameter now returns the pivot instead of genesis (as do other client implementations).

crates/blockchain/metrics/db.rs
```diff
@@ -0,0 +1,228 @@
...
+            self.block_cache_hits.inc_by(
+                block_cache_hits
+                    .saturating_sub(last_hits)
+                    .min(block_cache_hits),
+            );
+            self.block_cache_misses.inc_by(
+                block_cache_misses
+                    .saturating_sub(last_misses)
+                    .min(block_cache_misses),
+            );
```
MegaRedHand: This returns 0 when statistics are reset. We should instead do: `if block_cache_hits < last_hits { block_cache_hits } else { block_cache_hits - last_hits }`, same for `block_cache_misses` and `last_misses`.

Maybe there's a one-liner that returns the same, but I think this one's clearer.

crates/networking/p2p/sync_manager.rs
```diff
@@ -86,6 +93,21 @@ impl SyncManager {
             datadir,
             diagnostics.clone(),
         )));
+
+        // Spawn the optional historical-chain backfill task. It idles until
+        // initial sync finishes, then fills bodies/receipts below the pivot down
+        // to the configured floor, resuming across restarts via the persisted
+        // frontier. A no-op when `--history.chain off`.
+        if backfill_config.mode != HistoryChain::Off {
+            tokio::spawn(run_history_backfill(
```
MegaRedHand: Should this use `tracker.spawn`?

crates/networking/p2p/sync/backfill.rs
```diff
@@ -0,0 +1,734 @@
...
+    // just below it. On a snap-synced node the frontier is the pivot, and snap only
+    // stored the pivot's *body* (`add_block` writes no receipts and the block is
+    // never executed), so its receipts would otherwise never be filled and
+    // `eth_getBlockReceipts`/`eth_getTransactionReceipt` would stay wrong for that
+    // one block while the frontier advertised it as complete. Re-fetching one body
+    // is the cost of repairing it.
+    let batch_hi = if plan.first_batch {
+        frontier
+    } else {
+        frontier - 1
```
MegaRedHand: There's an edge case here, where if `frontier == floor` after snap-sync, the floor could end up without receipts, since the `frontier <= floor` check runs before checking `plan.first_batch`

MegaRedHand (approved):
Left some comments. Those can be addressed in further PRs

PR #7029, opened 2026-07-27
ci(l1,l2): make the cargo cache restorable, and cut duplicate builds in the L2 dev test

edg-l:
**Motivation**

Nothing in this repo restores a cargo cache. Every job logs `No cache found`, including on main, so every CI leg builds from scratch. The `Test - ubuntu-22.04` leg of the last completed L1 run spent 29 minutes on that, and the L2 Dev integration test 44 minutes, of which ~36 was rustc (21 min building ethrex, 15.5 min building the test binary, 6.5 min actually testing).

**Description**

*The cache is never readable.* An entry saved on a `refs/pull/N/merge` ref can only be restored by that same PR; `?ref=refs/heads/main` returns `count=0`, so there is nothing on a ref that other branches may read. `setup-rust` now saves only on main.

*Nothing fits in the budget either.* Counting matrix legs, 32 `setup-rust` calls run on main across six workflows, and rust-cache keys per job, so that is 32 writers at 1.2 to 2.75 GB each. The default budget is 10 GB with LRU eviction (admin-configurable, billed above that); `actions/cache/usage` currently reports 17 entries totalling 31.37 GiB. Either way 32 writers do not coexist, so main would thrash exactly as PR refs do now.

`setup-rust` gains `cache-pool`, forwarded to rust-cache's `shared-key`, which replaces the per-job segment of the key. Four pools, each with exactly one writer, picked as the widest dependency set in it:

| pool | writer | serves |
| --- | --- | --- |
| `l1` | `pr-main_l1::test` | root workspace at dev, plus `tooling` |
| `l1-reorg` | `pr-main_l1::reorg-tests` | root workspace at `release-fast` |
| `l2` | `pr-main_l2::integration-test` | `-F l2,l2-sql` at release |
| `prover` | `main_prover::test` | the SP1 host stack |

Everything else restores only, so a lean job cannot overwrite an entry with a subset and clippy-only target dirs never claim a slot. `reorg-tests` gets its own pool because it is the only job on main building the root workspace at `release-fast`, which the dev-profile `l1` entry cannot serve. That is 6 entries (the `l1` writer is a 3-platform matrix) instead of 32.

*The second workspace was never cached.* rust-cache's `workspaces` defaults to the repo root, and `tooling` is a separate workspace with its own target dir, so the ef_tests, reorgs, loc and hive_report builds were all cold regardless of anything above. That includes a step inside the `l1` writer's own job, since `pr-main_l1::test` runs `make -C tooling/ef_tests/blockchain test`. `setup-rust` now lists both workspaces. Entries get bigger for it; the count does not change.

Writing is opt-in: `save-cache` defaults to `'false'` and only the four jobs above pass `'true'`. It has to be opt-in rather than opt-out, because `schedule` and `workflow_dispatch` have no head ref of their own and fire on the default branch, so `github.ref` is `refs/heads/main` and `save-if` evaluates true. `daily_loc_report` and `daily_hive_report` were writing the `l1` entry nightly on that path, and `deep_reorg`, `pr_lint_readme`, `pr_perf_levm` and `pr_perf_build_block_bench` could. Cache keys are immutable, so the first of those to save wins and the pool owner's richer save is skipped, not merged.

The remaining PR-only workflows (`pr_loc`, `pr_check_l2_genesis`, `pr_perf_blocks_exec`) get a pool name too. They restore only; naming a pool is what makes main's entry readable for them, at no storage cost.

*The L2 test binary was a near total duplicate of the node build.* cargo fingerprints include the resolved feature set, and the two invocations diverged on 18 shared packages, among them sha3, tokio, hyper and idna, which sit low enough in the graph that the second build restarted from libc. Most of it came from `ethrex-test`'s `l2` feature being an empty marker while the crate dev-depends on `ethrex` with default features, so cmd/ethrex and the whole L2 stack compiled twice in different configurations. Forwarding `l2` to `ethrex/l2` and `ethrex/l2-sql` brings the divergence down to 6 packages and makes cmd/ethrex resolve identically in both builds. What still rebuilds is the three crates `ethrex-test` turns `testing`/`test-utils` on for, plus their dependents.

Three other jobs (`main_prover`, and two in `pr-main_l2`) prebuilt the test binary without `--features l2` and then ran with it. Harmless while the feature was empty, but now it flips the whole L2 stack config, so they get the flag too.

*The L2 build was hidden inside the backgrounded `cargo run`.* A compile error did not fail the job, it let the readiness loop poll for 25 min first, and the test binary was compiled while the node and prover were competing for the same 4 cores. Both binaries are now built in their own steps before the node starts, and the job uses `release-fast` instead of `release`. The test waits on 15s batch commits and L1 confirmations rather than on EVM throughput, so a slightly slower binary costs very little here.

*The deposit test raced the dev pre-funding.* Removing the implicit sleep exposed a pre-existing bug. A node started with `--dev` funds the rich accounts by depositing half of each L1 balance through the bridge, and the deployer awaits only the L1 receipts; the credits reach L2 through the L1 watcher a batch later. The port-3900 readiness gate opens ~20s before that, a gap the test binary's own compile time used to cover. So the suite read a zero initial L2 balance and `test_deposit` failed by exactly the 5e26 wei pre-funding. `l2_integration_test` now polls every test account's balance until it is non-zero before asserting any delta, which fixes local `make init` runs too.

Expect no improvement on this PR itself, since main has to run once to populate the pools.

**Not in this PR**

Jobs building the root workspace at plain `release` (the `pr_perf_*` benchmarks, `deep_reorg`) have no writer on main, so they restore `~/.cargo` but no matching target dir. Same for `revm_comparison`, which is excluded from the workspace and has a third target dir. Giving them writers costs entries, and the budget is not measurable from outside the org.

`aligned-sdk` is a non-optional dependency of `ethrex-l2` and pulls the whole SP1 host stack (~4 min of the critical path) into every `-F l2` build even with no zkVM backend enabled. Gating it behind a feature needs config and release-plumbing changes.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. N/A: CI and feature wiring only.


ilitteri (approved):

test/Cargo.toml
```diff
@@ -9,7 +9,10 @@ path = "src/lib.rs"
 
 [features]
 rocksdb = ["ethrex-storage/rocksdb"]
-l2 = []
+# Matches the feature set the L2 node itself is built with (`-F l2,l2-sql`), so
+# the test binary reuses those artifacts instead of compiling the whole L2 stack
+# a second time in a different configuration.
+l2 = ["ethrex/l2", "ethrex/l2-sql"]
```
ElFantasma, line 15: Blocking, and I don't think it's a flake: the `L2 Dev` / Integration Test job this PR optimizes fails on **both** commits of this branch, with the identical assertion —

```
test/tests/l2/integration_tests.rs:1509
assertion `left == right` failed: Deposit recipient L2 balance didn't increase as expected after deposit
```

`bbe312b35` (07-27) and `21bacb457` (07-29) both die there, while `L2 Dev` on `main` is 8/8 green over the same window — including two runs *after* the latest failure here. The L2 deposit receipt is found and its `status` is true; only the balance-delta assertion fails, so the deposit lands but not with the expected effect at the point the test reads it.

Two changes in this PR could produce that, and they want separating:

1. **This line.** Flipping `l2` from `[]` to `["ethrex/l2", "ethrex/l2-sql"]` is not only a fingerprint/artifact-reuse change — it changes what the *test binary* links. `ethrex` is a dev-dependency (line 53), previously compiled with default features; it now carries the full L2 stack plus `l2-sql`, so any `#[cfg(feature = ...)]` path in the linked lib is live in-test for the first time. "Harmless while the feature was empty" holds for the prebuild steps, but not necessarily for test behaviour.
2. **Build ordering** (`pr-main_l1_l2_dev.yaml`). The test binary used to compile *after* the node started — ~15.5 min in which the node and prover incidentally warmed up. Now the node is ready in 5 s, the prover starts 5 s later, and the test begins ~10 s after node start. Anything in the deposit path that leaned on that window (L1-watcher poll cycles, the first `--committer.commit-time 15000` batch, prover connection) now races.

Cheapest way to tell them apart is to land this as two PRs — cache + build-ordering first, the `test/Cargo.toml` feature change second — because the fixes differ completely. If it turns out to be (2), the durable fix is gating `Run test` on the real precondition (e.g. first committed batch) rather than leaning on compile time as an implicit sleep.

.github/workflows/pr-main_l1_l2_dev.yaml
```diff
@@ -70,20 +81,20 @@ jobs:
 
       - name: Wait for ethrex L2
         run: |
-          for i in {1..100}; do
+          for i in {1..120}; do
```
ElFantasma, line 86: Non-blocking, and flagging it because the opposite change was suggested elsewhere on this PR: this loop is not the flakiness risk it looks like. In the latest run the `Wait for ethrex L2` step completed in **5 seconds** (15:24:17 → 15:24:22), so 120 × 5 s is roughly 120× the observed startup need. Raising the iteration count would only lengthen the failure path for a genuinely dead node; the 25 min budget it replaces was sized when this step was also absorbing the node's *compile* time, which now happens in its own step above.

(Relevant to the deposit failure too: if that turns out to be startup-timing related, a longer port poll won't help — port 3900 opens in 5 s either way. The gate would need to be on the actual precondition.)

.github/actions/setup-rust/action.yml
```diff
@@ -23,6 +23,13 @@ runs:
...
 
+    # Only main writes the cargo cache. A repo has a 10GB Actions cache budget
+    # with LRU eviction, and a single release target dir here is ~2.75GB, so
+    # letting every PR job save means the entries evict each other and almost
+    # every run builds cold. Caches saved on PR refs are also unreadable from
+    # any other branch, whereas main's are readable everywhere.
     - name: Add Rust Cache
       uses: Swatinem/rust-cache@v2
+      with:
+        save-if: ${{ github.ref == 'refs/heads/main' }}
```
ElFantasma: Direction is right — PR-scoped entries aren't readable from other branches, so those writes were pure eviction pressure. Worth re-checking the arithmetic on what's left, though: `save-if` still leaves every job that runs on `push: main` as a writer, and that's ~18 `setup-rust` invocations across the five workflows triggered on main (`pr-main_l2` 7, `pr-main_levm` 5, `pr-main_l1` 4, `main_prover` 1, `pr-main_l1_l2_dev` 1). rust-cache keys per job, so ~18 entries still compete for the 10 GB budget at the ~2.75 GB/entry this repo actually observes (the 3 entries / 8.25 GB in the description) — about 5× oversubscribed, so main entries will keep LRU-evicting each other.

That doesn't undercut the change, since main entries are at least restorable everywhere. But it does mean "~16 to 20 min once main repopulates it" may not hold steadily for this job — its entry can be evicted by the other 17 between main runs. The `shared-key` pooling the description already earmarks as follow-up for `pr_loc`/`pr_perf_*` is the same lever here: collapsing these onto a few shared keys gets the entry count under budget, rather than only changing who writes.

ElFantasma (changes requested):

.github/actions/setup-rust/action.yml
edg-l: Did the arithmetic, and it's worse than either of us had it. Pushed the shared-key pooling in 7fae6538c.

First, a correction to my own description: the "3 entries / 8.25 GB" came from `GET /actions/caches`, which is partial. It returns 6 entries right now; `GET /actions/cache/usage` reports **17 entries / 31.37 GiB**. So the budget picture in the description understates it by ~4x.

Second, your writer count is low, and not by rounding. `tag_release.yaml` also triggers on `push: branches: ["main"]` (alongside tags) with no `if:` guard on either job, so it contributes 11 more: 9 `build-ethrex` matrix legs plus 2 `build-ethrex-guest`. Counting matrix legs, main-triggered `setup-rust` calls come to **32 distinct keys**, not ~18. At the 1.2-2.75 GB/entry this repo actually produces that's 40-88 GB of writes against the budget, so you're right that `save-if` alone just moves the thrash from PR refs onto main.

Third, the 10 GB in my description is only the *default*. Per the [caching reference](https://docs.github.com/en/actions/reference/dependency-caching-reference) it "can be increased by enterprise owners, organization owners, or repository administrators", billed. I can't read this repo's configured value (`orgs/lambdaclass/actions/cache/usage-by-repository` needs `admin:org`), but 31.37 GiB sitting flat while every PR job writes reads like a ceiling near 32 GB rather than 10. Someone with org admin could confirm. Either way pooling behaves correctly at any budget, so I didn't wait on it.

Worth noting the misses we actually observe are a *scoping* failure, not a size one, which is why `save-if` is the right core fix regardless: `?ref=refs/heads/main` returns `count=0`, every existing entry is on a `refs/pull/N/merge` ref, and the last completed L1 run's `Test - ubuntu-22.04` job logged

```
Restore Key: v0-rust-test-Linux-x64-1e811033
... Restoring cache ...
No cache found.
```

then built cold for 29 minutes. Not even a prefix hit, because there is nothing on a ref it may read.

What 7fae6538c does, which is the follow-up you earmarked:

- `setup-rust` gains `cache-pool` (forwarded to rust-cache's `shared-key`, which replaces the per-job segment of the key) and `save-cache`. An empty `cache-pool` falls back to the per-job key — `src/config.ts` does `if (sharedKey)` — so any call site I didn't touch is unaffected.
- Three pools: `l1` (pr-main_l1 lint/test/engine-ef-tests/reorg-tests + all five pr-main_levm jobs), `l2` (all seven pr-main_l2 jobs, pr-main_l1_l2_dev, and tag_release's 11 legs), `prover` (main_prover).
- One writer per pool — `pr-main_l1::test`, `pr-main_l2::integration-test`, `main_prover::test` — chosen as the widest dependency set in each. Everyone else restores only, so a lean job can't overwrite the pool entry with a subset, and clippy-artifact target dirs never claim a slot.

That takes 32 keys to ~5 entries: 3 for `l1` (its writer is the 3-platform matrix), 1 `l2`, 1 `prover`.

Two things this deliberately does not fix. rust-cache's `workspaces` defaults to the repo root, so `tooling/target` is never cached at all — the levm `ef-test*` jobs and `engine-ef-tests` only ever restore `~/.cargo`, which is part of why they're restore-only here rather than pool writers. Fixing that needs `workspaces: . + tooling` and more budget than we can currently account for. And per-key rotation still doubles the steady-state count for a while after any `Cargo.toml`/lockfile change, since the old entry lingers until it's evicted or hits the 7-day unused cutoff.

The `test/Cargo.toml` feature change and the deposit failure are untouched here — that split is a separate call.

.github/workflows/daily_loc_report.yaml
```diff
@@ -33,6 +33,8 @@ jobs:
         uses: actions/checkout@v6
       - name: Setup Rust Environment
         uses: ./.github/actions/setup-rust
+        with:
+          cache-pool: l1
```
ElFantasma, line 37: `save-cache` defaults to `"true"` in the action, so a call site that omits it is a pool **writer**, not just a reader.

The `refs/heads/main` half of `save-if` does cover the PR-only workflows this commit targets — on a PR `github.ref` is `refs/pull/N/merge`, so nothing saves. But this one isn't PR-only. It's `schedule: cron "0 0 * * *"`, which runs on main, so `save-if` evaluates true and this job writes the `l1` entry every night.

That's the case the design explicitly rules out:

> One writer per pool — `pr-main_l1::test`, `pr-main_l2::integration-test`, `main_prover::test` [...] Everyone else restores only, so a lean job can't overwrite the pool entry with a subset, and clippy-artifact target dirs never claim a slot.

A LOC counter is precisely that lean job. It's also stickier than a plain overwrite: Actions cache keys are immutable, so whichever job writes a given key first wins and the later, richer save from `pr-main_l1::test` is skipped rather than replacing it. The thin entry then persists until the key rotates on a `Cargo.lock`/toolchain change or hits the 7-day unused cutoff, and `pr-main_l1::test` restores a subset and rebuilds most of what pooling was meant to save.

`save-cache: 'false'` here — the way `tag_release.yaml` already has it — restores the intended invariant.

.github/workflows/daily_hive_report.yaml
```diff
@@ -204,6 +204,8 @@ jobs:
         uses: actions/checkout@v6
       - name: Setup Rust Environment
         uses: ./.github/actions/setup-rust
+        with:
+          cache-pool: l1
```
ElFantasma, line 208: Same missing guard as `daily_loc_report.yaml`: `schedule: cron "0 3 * * 1,2,3,4,5"` runs on main, so with `save-cache` defaulting to `"true"` this claims the `l1` slot on weekday mornings. Wants `save-cache: 'false'`.

The `pull_request:` trigger on this workflow is fine as-is — the `refs/heads/main` condition in `save-if` already stops PR runs from saving.

.github/workflows/deep_reorg.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
 
       - name: Setup Rust Environment
         uses: ./.github/actions/setup-rust
+        with:
+          cache-pool: l1
```
ElFantasma, line 29: Same missing `save-cache: 'false'`. `workflow_dispatch` carries no ref guard of its own and defaults to the default branch, so a manual run — the normal way this one is invoked — writes the `l1` entry from main.

Lower frequency than the two daily crons, but it competes for the same slot, and a reorg-test target dir is not the dependency set `pr-main_l1::test` needs restored.

ElFantasma (commented):
Follow-up on the pooling commits: three main-running workflows are missing `save-cache: 'false'`, so they claim the l1 pool slot the one-writer design reserves.

.github/workflows/daily_loc_report.yaml
edg-l: Fixed in 21536a4e6, and inverted the default rather than patching the call site: `save-cache` now defaults to `'false'` and the four pool owners opt in.

You're right that this one fires. Last 100 runs of `daily_loc_report` are all `schedule` on `refs/heads/main` — it wrote the `l1` entry at 00:47 UTC today. Same for `daily_hive_report` at 04:14.

The audit found three more the same way, all reachable via `workflow_dispatch`: `pr_lint_readme`, `pr_perf_levm` (both jobs) and `pr_perf_build_block_bench`. Those are latent rather than observed — their main-ref history is entirely `pull_request` from fork branches named `main`, which are `refs/pull/N/merge` and never saved. `pr_perf_build_block_bench` is the sharp one: its `pull_request:` trigger is commented out, so dispatch is now its only trigger and every future run of it is on main.

Which is the argument for opt-in over opt-out. The invariant shouldn't depend on reading nine trigger blocks correctly; a forgotten flag now yields a reader.

.github/workflows/daily_hive_report.yaml
edg-l: Confirmed and fixed in 21536a4e6 — this one ran on main at 04:14 UTC today. Covered by the default inversion rather than a per-site guard; see the reply on `daily_loc_report`.

Agreed on `pull_request:` being fine as-is.

.github/workflows/deep_reorg.yaml
edg-l: Fixed in 21536a4e6. Worth noting this one is latent, not observed: `deep_reorg` has zero runs in its history, so it has never claimed the slot. Still reachable exactly as you describe, and it's a reader now.

test/Cargo.toml
edg-l: Root-caused it, and it's (2). Fixed in 406604684, so I don't think the split buys anything — the `test/Cargo.toml` change is not implicated.

The mechanism, from the last failing run:

```
11:55:46  node starts
11:55:51  nc -z localhost 3900 succeeds  ->  "Wait for ethrex L2" exits
11:55:53  L2 RPC listening on 127.0.0.1:1729
11:56:03  test reads initial L2 balance -> "initial L2 balance=0 wei"
11:56:12  "Added 192 privileged to the batch"  <- the pre-funding lands, batch 2, blocks 2-4
```

`--dev` runs the deployer with `deposit_rich: true`, which deposits half of each rich account's L1 balance through the bridge for all 192 keys in `private_keys_l1.txt` (`cmd/ethrex/l2/deployer.rs:1514`). L1 genesis funds each with 1e27 wei, so each L2 account is credited 5e26. `make_deposits` awaits only the **L1** receipt of the last deposit; nothing waits for the L1 watcher to apply them on L2.

That accounts for the assertion exactly: `left: 500001000000000000000000000` = 5e26 pre-funding + 1e21 deposit, `right: 1000000000000000000000` = 0 + 1e21. Off by precisely the pre-funding, not by anything fee- or config-shaped. The L2 genesis does not fund these accounts at all, so a feature flag flipping a `cfg` path in the linked lib can't produce this number.

So the port-3900 gate was never the right precondition — it opens ~20s before the state the test depends on exists, and the 15.5 min compile was covering the gap. `l2_integration_test` now polls every test account's L2 balance until non-zero before the first delta assertion. All 21 keys in `private_keys_tests.txt` are a subset of the 192 pre-funded ones, so the precondition is well defined. Put it in the test rather than the workflow because local `make init` runs have the identical race.

On the readiness loop in your other comment: agreed, and left as-is.

ElFantasma (approved):

MegaRedHand (approved):

PR #7030, opened 2026-07-27
fix(l1): do not journal a trie layer that aggregates several blocks

ilitteri:
**Motivation**

A `STATE_HISTORY` entry describes exactly one block: it is keyed by a block number and
carries that block's hash and pre-state root. The write path added in #6686 decides whether
to journal per persist **message**, but a commit is per **layer** — `TrieLayerCache::commit`
removes the target layer *and every ancestor*, so one non-batch message can commit layers
that an earlier batch message created.

A batch layer aggregates ~1024 blocks and carries only its **last** block's number and hash.
Journaling it writes a single entry keyed at block N whose reverse diff actually undoes
N-1023..N.

Nothing downstream detects this:

- the entry's `parent_state_root` is self-consistent with the over-wide revert, so a root
  check inside the entry passes;
- the rollback consumer's `block_hash` verification in `Overlay::from_journal` (#6687) passes
  too, because a batch layer's hash **is** the canonical hash at that key.

What the consumer does notice is the resulting *gap*: `from_journal` walks contiguously
descending and returns `OverlayError::MissingEntry` for the missing N-1. So today the
practical outcome is a deep reorg that aborts. The silent case is a walk with
`from_block == to_block == N`, which meets no gap, passes the hash check, and over-reverts by
~1023 blocks.

Reachable via `ethrex import` and at the full-sync-to-live handoff, where residual batch
layers sit below the per-block commit gate. Latent so far only because nothing reads
`STATE_HISTORY` yet.

Two notes on scope:

- **The canonical safe-commit gate does not make this safe by itself.** Pointing the safe
  root at a batch boundary reproduces the same entry. The gate avoids it in practice because
  `head - 128` does not land on a batch boundary — an accident the write path currently
  depends on, not an invariant it enforces.
- **Depth-gated commits make it systematic**, because a depth gate returns whatever layer
  sits at the requested depth, batch or per-block, indiscriminately. #7018 introduces
  depth-gating on the full-sync and import tails, so this moves from incidental to ~4 bad
  entries per full-sync completion.

**Description**

Threads the layer's block count from the persist message into the trie layer and out through
`CommittedLayer`, and journals a layer only when it covers exactly one block. The journaling
decision moves from per-message (`!is_batch`) to per-layer (`!is_batch && block_count == 1`),
which is the granularity a commit actually operates at.

Skipping leaves a gap rather than a wrong entry, so the consumer's existing `MissingEntry`
detection turns a wrong-but-plausible unwind into a fail-closed abort. Making deep reorgs
actually *succeed* across an aggregate region needs span-aware entries on both the write and
read sides, which is out of scope here.

Also corrects the `BlockPersist::block_number` doc, which said that using the batch's last
block is "harmless for batch updates, since journal writes are skipped when `wait_for_flush`
(batch mode) is set". That holds for the message that *creates* the layer, not for whoever
later *commits* it — and it is what made this defect invisible on review.

Adds a regression test that fails without the guard: two three-block batches, then one
per-block commit whose gate reaches back into them, asserting no entry lands on either batch
boundary. Verified red before the change (an entry appeared at block 3 carrying all three
blocks' keys) and green after.

Follow-up to #6686. Relevant to #6687, which defines and consumes the one-entry-per-block
invariant, and to #7018, which widens the exposure.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A. No schema change: the entry format is untouched and this only omits entries that were previously malformed.

### DIFF
--- a/crates/storage/layering.rs
+++ b/crates/storage/layering.rs
@@ -23,6 +23,15 @@ struct TrieLayer {
     block_number: BlockNumber,
     /// Hash of the block whose post-state this layer represents.
     block_hash: H256,
+    /// How many blocks this layer's diff aggregates. One on every per-block path; the
+    /// bespoke batch path inserts a single layer for a whole batch (~1024 blocks).
+    ///
+    /// A journal entry describes exactly one block, so the journal write path uses this
+    /// to refuse layers that aggregate several (see `commit_to_disk`). Without it a
+    /// batch layer looks like a single block: it carries only its LAST block's number
+    /// and hash, so an entry written for it would claim to undo one block while
+    /// actually undoing the whole batch.
+    block_count: u64,
 }
 
 /// In-memory cache of trie diff-layers, one per block (or batch in full sync), forming a
@@ -227,6 +236,7 @@ impl TrieLayerCache {
         state_root: H256,
         block_number: BlockNumber,
         block_hash: H256,
+        block_count: u64,
         key_values: Vec<(Nibbles, Vec<u8>)>,
     ) {
         if parent == state_root && key_values.is_empty() {
@@ -260,6 +270,7 @@ impl TrieLayerCache {
             id: self.last_id,
             block_number,
             block_hash,
+            block_count,
         };
         self.layers.insert(state_root, Arc::new(entry));
     }
@@ -336,6 +347,7 @@ impl TrieLayerCache {
             .map(|layer| CommittedLayer {
                 block_number: layer.block_number,
                 block_hash: layer.block_hash,
+                block_count: layer.block_count,
                 parent_state_root: layer.parent,
                 nodes: layer.nodes.into_iter().collect(),
             })
@@ -353,6 +365,10 @@ impl TrieLayerCache {
 pub struct CommittedLayer {
     pub block_number: BlockNumber,
     pub block_hash: H256,
+    /// How many blocks this layer's diff aggregates; see [`TrieLayerCache::put_batch`].
+    /// Only a layer with `block_count == 1` may be journaled, since a journal entry
+    /// describes exactly one block.
+    pub block_count: u64,
     pub parent_state_root: H256,
     pub nodes: Vec<(Vec<u8>, Vec<u8>)>,
 }
@@ -458,7 +474,7 @@ mod tests {
         let mut parent = H256::zero();
         for i in 1..=n {
             let root = h256(i);
-            cache.put_batch(parent, root, i as u64, root, vec![(key(i), vec![i])]);
+            cache.put_batch(parent, root, i as u64, root, 1, vec![(key(i), vec![i])]);
             roots.push(root);
             parent = root;
         }
@@ -510,7 +526,7 @@ mod tests {
         let orphan = h256(7);
         let (mut cache, _cell) = cache_with_cell(4, orphan);
         // An empty block: parent == state_root, so `put_batch` inserts nothing.
-        cache.put_batch(orphan, orphan, 7, orphan, vec![]);
+        cache.put_batch(orphan, orphan, 7, orphan, 1, vec![]);
         assert!(
             !cache.has_layer(orphan),
             "empty-block root must not create a layer"
@@ -555,8 +571,8 @@ mod tests {
         let (mut cache, _cell) = cache_with_cell(4, h256(99));
         // Insert C (parent B) then B (parent C); neither key pre-exists, so both insert,
         // forming the cycle B <-> C. The safe root (h256(99)) is absent on purpose.
-        cache.put_batch(b, c, 21, c, vec![(key(21), vec![21])]);
-        cache.put_batch(c, b, 20, b, vec![(key(20), vec![20])]);
+        cache.put_batch(b, c, 21, c, 1, vec![(key(21), vec![21])]);
+        cache.put_batch(c, b, 20, b, 1, vec![(key(20), vec![20])]);
         // Walking from C must terminate (start-of-walk + bounded-walk guards) and yield None.
         assert_eq!(cache.get_commitable(c), None);
     }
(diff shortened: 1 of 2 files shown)


edg-l (approved):

edg-l:
Maybe check claude review in this PR

ilitteri:
Closing in favour of #7008, which removes the hazard structurally rather than guarding against it.

This PR guards the journal write path against layers that aggregate several blocks: `TrieLayerCache::commit` sweeps the target layer and every ancestor, so a per-block commit could journal a batch layer, writing one entry keyed at block N whose reverse diff actually undoes N-1023..N.

#7008 deletes the only producer of such layers — the multi-block `UpdateBatch` in `add_blocks_in_batch`, along with the only `commit_depth: Some(BATCH_COMMIT_THRESHOLD)` and `wait_for_flush: true` sites. After it, every `UpdateBatch` in production carries exactly one block, so `put_batch` can only build single-block layers and the guard here becomes a check on a condition that cannot be false. The `block_count` field threaded through `BlockPersist` → `apply_trie_phase1` → `put_batch` → `CommittedLayer` would be dead weight, and the regression test would exercise a shape production no longer produces.

There is also no harmful window in the meantime: the rollback consumer (#6687) is unmerged, so nothing reads `STATE_HISTORY` yet.

One piece is worth keeping regardless, so I have split it out: the `BlockPersist::block_number` doc claims using the batch's last block is "harmless for batch updates, since journal writes are skipped when `wait_for_flush` (batch mode) is set". That reasoning is wrong independent of #7008 — it holds for the message that *creates* a layer, not for whoever later *commits* it — and it is what made this defect invisible on review.

For the record, in case batching is ever reintroduced: the canonical safe-commit gate does not make this safe by itself. Pointing the safe root at a batch boundary reproduces the same entry; the gate avoids it only because `head - 128` does not land on one. Reproducer is on this branch.

ilitteri:
Audit of this PR (diff `main...fix/journal-skip-aggregate-layers`, head `8e00fd75d`). The `block_count` plumbing and the per-layer gate are the right shape of fix, and single-block/empty/genesis boundary cases all check out — but the gate is applied one level too broadly, and it regresses pre-image correctness in exactly the mixed-commit scenario the PR targets. (Noting this is still worth fixing or carrying into #7008's design review even if this PR is closed in favor of #7008.)

## High — mixed commits now record stale pre-images for journaled layers above a skipped batch layer

`crates/storage/store.rs:4169` and `store.rs:4207` (head).

The PR reuses `journal_this_layer` for two different decisions. The intra-commit overlay exists so that in a multi-layer commit, each block's journal entry records its pre-image as of the *previous committed layer's* write, not the pre-commit disk value:

```rust
let prev_value = if journal_this_layer {        // was: if !is_batch
    match overlay.get(key) { ... read_view.get(table, key) ... }
} else { None };
...
if let Some(prev) = prev_value {
    bucket.push((key.clone(), prev));
    overlay.insert(key.clone(), new_value);     // skipped when !journal_this_layer
}
```

Before this PR, in a live commit (`is_batch == false`) *every* swept layer maintained the overlay. After this PR, a skipped aggregate layer (`block_count > 1`) neither reads nor advances the overlay. A mixed commit — a live commit sweeping leftover batch layers (oldest) plus newer single-block layers in one `commit_to_disk` call — then records, for any key written by both the batch layer and a later journaled layer, a pre-image read from `read_view` (opened before the write batch, so blind to the batch layer's staged writes). The journaled entry's reverse diff restores the *pre-batch* value instead of the *post-batch* value.

This is reachable in normal operation: at the full-sync→live transition up to `BATCH_COMMIT_THRESHOLD - 1` batch layers remain uncommitted, and the first canonical-gate commit after head advances past them sweeps them together with the single-block layers above. Key overlap is effectively guaranteed because upper state-trie node paths are rewritten by every layer. The wrong entries persist until finality pruning; a rollback applying them would leave on-disk state inconsistent with the batch tip's state root — silent state corruption on the deep-reorg path. The PR's own comment (`store.rs:4124-4132`) acknowledges mixed commits happen but only considers the entry-identity problem, missing the overlay interaction.

Fix is simple: compute `prev_value` and maintain `overlay` under the old per-commit condition (`!is_batch`), and use `journal_this_layer` only for the entry write at `store.rs:4216`.

## Low — new test's block-6 assertion is vacuous; the mixed-commit case is untested

`crates/storage/store.rs:4877`. In `aggregate_batch_layer_is_not_journaled_by_a_per_block_commit`, the safe-commit root is set to block 3's state root, so the live commit triggered by block 7 commits only the batch-1 layer (commit sweeps the target and *ancestors*; block 6's layer is a descendant of the commit point and is never committed). `assert_no_journal_entry(&backend, 6)` would pass even against pre-fix code; only the block-3 assertion is a real regression check. More importantly, no test covers the scenario behind the finding above: a journaled single-block layer committed in the same sweep as a skipped aggregate layer with an overlapping key, asserting the recorded pre-image is the intermediate value.

## Low — the "fail-closed gap detection" claim is not delivered by this change

`crates/storage/store.rs:4129-4132` states the rollback consumer "detects [the gap] and refuses to unwind across". No rollback consumer exists on this branch or its base — the comment is a forward-looking claim about a future stack PR; nothing here enforces it.

## Verified non-issues

Single-block layers, empty layers, and genesis handled correctly (empty blocks create no layer; genesis bypasses layering); only one production `put_batch` caller and one `BlockPersist` construction site; finality pruning unaffected by gaps; `block_count` is in-memory only (no schema concern); locking/threading unchanged.

**Strengths:** clear, honest comments explaining why `wait_for_flush` cannot substitute for `block_count`; correct per-layer rather than per-message reasoning; the block-3 leg of the new test is a genuine regression test that fails pre-fix; good boundary handling for 1-block batches.

PR #7033, opened 2026-07-27
fix(l1): reject post-Shanghai block bodies missing the withdrawals field

ilitteri:
## Summary

`validate_block_body` accepted a post-Shanghai block body that **omits** the `withdrawals` field, as long as the header's `withdrawals_root` equaled the empty-withdrawals root. Every reference client rejects this shape:

- geth: `"missing withdrawals in block body"` (`core/block_validator.go`)
- reth: `WithdrawalsRootUnexpected` + `BodyWithdrawalsMissing`
- nethermind: `MissingWithdrawals` (`BlockErrorMessages.cs`)
- erigon: `"body is missing withdrawals"` (`execution/types/block.go`)
- besu: `AllowedWithdrawals` presence check + `checkArgument`

The RLP schema for a Shanghai+ body *includes* the withdrawals field (possibly an empty list); a body without it is malformed, not "empty".

## Change

`crates/common/types/block.rs::validate_block_body`: the `(Some(_), None)` match arm now rejects unconditionally instead of accepting when the root matches `EMPTY_WITHDRAWALS_HASH`. Also drops the now-unused `EMPTY_WITHDRAWALS_HASH` import.

## Impact

No live-network impact (post-merge payloads always carry the field via the engine API, and the devp2p path was state-identical), but this removes a block-validation divergence from all other EL clients — relevant for hive / cross-client conformance and for p2p body serving.

## Tests

Added `test_validate_block_body_rejects_missing_withdrawals_field` covering:
- `(Some(root), None)` → rejected, even with the empty root
- `(Some(root), Some([]))` → accepted
- `(None, Some([]))` → rejected
- `(None, None)` → accepted

`cargo test -p ethrex-common --lib types::block` — 26/26 green.

### DIFF
--- a/crates/blockchain/payload.rs
+++ b/crates/blockchain/payload.rs
@@ -198,7 +198,15 @@ pub fn create_payload(
     let body = BlockBody {
         transactions: Vec::new(),
         ommers: Vec::new(),
-        withdrawals: args.withdrawals.clone(),
+        // Post-Shanghai the withdrawals field is part of the body schema: emit
+        // an explicit (possibly empty) list, matching the header's
+        // withdrawals_root above, instead of omitting the field. Omitting it
+        // produces blocks that fail `validate_block_body` — the L2 block
+        // producer passes `withdrawals: None`, which previously leaked through
+        // as a body without the field under an empty-root header.
+        withdrawals: chain_config
+            .is_shanghai_activated(args.timestamp)
+            .then(|| args.withdrawals.clone().unwrap_or_default()),
     };
 
     // Delay applying withdrawals until the payload is requested and built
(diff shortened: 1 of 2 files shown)


MegaRedHand (approved):

jrchatruc (approved):

edg-l (approved):

ilitteri:
**L2 integration test fix** (commit 9a6f19455)

Root cause of the 6h L2 hangs: the L2 block producer calls `create_payload` with `withdrawals: None`, which the old code leaked into the body while the header still committed to the empty withdrawals root — exactly the shape this PR now rejects in `validate_block_body`. Every sequenced L2 block then failed body validation on import, so batches never advanced and the integration test hit the job timeout.

Fix: `create_payload` now mirrors the header's Shanghai condition in the body — post-Shanghai it emits an explicit (possibly empty) withdrawals list; pre-Shanghai the field stays absent. Production paths that pass real withdrawals (engine FCU) are unchanged.

Regression test added: `create_payload_emits_explicit_empty_withdrawals_post_shanghai` asserts a `withdrawals: None`-args payload produces `body.withdrawals == Some(vec![])` and passes `validate_block_body`. `cargo test -p ethrex-blockchain --lib` green (20/20).

PR #7034, opened 2026-07-27
docs(l1,l2): correct stale network, sync-mode and GPU binary references

ilitteri:
**Motivation**

Several operator-facing docs describe behavior the CLI no longer has. Following them verbatim fails: `--network holesky` does not start, `ethrex --sync <mode>` is not a valid flag, and two documented download URLs 404. One of the dead links also fails the Link Check CI job.

**Description**

- **Holesky is not a recognized network.** `Network::from` (`crates/common/config/networks.rs:48`) maps only `mainnet`, `hoodi` and `sepolia`; anything else becomes a genesis *file path*, so `--network holesky` tries to open a file named `holesky` and fails. There are zero `holesky` references left in the Rust source. Dropped it from the supported-networks table and the startup example, and retargeted the syncing walkthrough, profiling guide and importing-blocks note to Hoodi. `docs/CLI.md`'s `ethrex l2` section also advertised holesky — that block is an outdated snapshot of the same `long_help` in `cmd/ethrex/cli.rs:75`, which is already correct (it is outside the CI-verified `BEGIN_CLI_HELP` markers, which is why the drift went unnoticed).
- **Also unblocks Link Check.** `checkpoint-sync.holesky.ethpandaops.io` no longer resolves in DNS, and `lychee.toml` accepts only 2xx/429/503 with no exclusion for that host, so any PR touching `docs/**` fails on it.
- **`--syncmode` defaults to `snap`, not `full`** (`cmd/ethrex/cli.rs:123`). Fixed the inverted claim in `startup.md` and dropped the now-redundant `--syncmode snap` from the mainnet/sepolia examples.
- **`configuration.md` documented the flag as `--sync`**, which the CLI rejects.
- **GPU binary URLs pointed at assets that are never published.** The release workflow builds `-gpu` binaries only for the `l2` stack, so the real names are `ethrex-l2-linux-{x86_64,aarch64}-gpu`. Confirmed the documented URLs return 404 and the corrected ones resolve.
- Renamed `syncing_holesky.md` → `syncing_hoodi.md`. The page is absent from `SUMMARY.md` and nothing links to it, so no references break.

Verified with `mdbook build` (html + linkcheck2 backends) using the CI-pinned preprocessor versions — clean, no errors.

Left alone deliberately: `docs/crate_reviews/ethrex_common_review.md` (a dated, commit-pinned review snapshot referencing a `holesky_test_cases` fn that no longer exists) and `docs/roadmaps/forks-roadmap.md` (historical roadmap citing a since-removed genesis path). Both are point-in-time records, not instructions.

Separate issue noticed but not touched here: `docs` is not in the `Makefile`'s `.PHONY` list, so `make docs` resolves against the `docs/` *directory*. Locally it printed `make: 'docs' is up to date.` and built nothing — `mdbook build` had to be invoked directly. In CI this depends on checkout mtimes relative to `mermaid-init.js`/`mermaid.min.js`, so the Lint job may not actually be building the book. Worth a one-line Makefile fix in its own PR.

**Checklist**

- [ ] ~Updated `STORE_SCHEMA_VERSION`~ — not applicable, docs-only change.

### DIFF
--- a/docs/l1/running/configuration.md
+++ b/docs/l1/running/configuration.md
@@ -7,12 +7,12 @@ This page covers the basic configuration options for running an L1 node with eth
 Ethrex supports different sync modes for node operation:
 
 - **full**: Downloads and verifies the entire chain.
-- **snap**: Fast sync using state snapshots (recommended for most users).
+- **snap**: Fast sync using state snapshots (default, recommended for most users).
 
 Set the sync mode with:
 
 ```sh
-ethrex --sync <mode>
+ethrex --syncmode <mode>
 ```
 
 ## File Locations
--- a/docs/l1/running/configuration.md
+++ b/docs/l1/running/configuration.md
@@ -6,8 +6,8 @@ This page covers the basic configuration options for running an L1 node with eth
 
 Ethrex supports different sync modes for node operation:
 
+- **snap** (default): Fast sync using state snapshots (recommended for most users).
 - **full**: Downloads and verifies the entire chain.
-- **snap**: Fast sync using state snapshots (default, recommended for most users).
 
 Set the sync mode with:
 
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -315,7 +315,7 @@ Options:
 
 Node options:
       --network <GENESIS_FILE_PATH>
-          Alternatively, the name of a known network can be provided instead to use its preset genesis file and include its preset bootnodes. The networks currently supported include holesky, sepolia, hoodi and mainnet. If not specified, defaults to mainnet.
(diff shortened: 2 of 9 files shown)


MegaRedHand (approved):
I think we still have some other Holesky mentions in our tooling

jrchatruc (approved):

docs/l1/running/configuration.md
```diff
@@ -7,12 +7,12 @@ This page covers the basic configuration options for running an L1 node with eth
 Ethrex supports different sync modes for node operation:
 
 - **full**: Downloads and verifies the entire chain.
-- **snap**: Fast sync using state snapshots (recommended for most users).
+- **snap**: Fast sync using state snapshots (default, recommended for most users).
```
ManuelBilbao: I'd put the default first

ilitteri: Done

ManuelBilbao (approved):

PR #7037, opened 2026-07-28
fix(l1): refresh the eth forkid ENR entry when the chain crosses a fork

edg-l:
**Motivation**

The `eth` ENR entry (EIP-2124 fork id) is derived once, in `DiscoveryServer::spawn`. A node that is already running when a fork activates keeps advertising the pre-fork id until it restarts.

geth re-derives it on every chain head event (`StartENRUpdater` in `eth/protocols/eth/discovery.go`) and filters dial candidates on it (`NewNodeFilter`), skipping records whose `eth` entry fails to load.

This matters on glamsterdam-devnet-8, which runs EL discovery discv5-only, so the ENR is the only fork signal a peer has before the RLPx handshake.

**Description**

`refresh_fork_id` re-derives the fork id from the current head on the existing 5s prune tick, and re-signs with a bumped seq when it changed. Unchanged ids return early so the seq does not churn.

Startup now warns instead of silently publishing a record with no `eth` entry; that is the case geth's filter drops, and it was invisible in the logs. The periodic refresh self-heals it on the next tick.

Not a hard peering break today: geth accepts a well-formed stale pre-fork id under EIP-2124 rule #2 (remote is syncing), since the advertised `FORK_NEXT` still matches the fork we passed. The record is still wrong for anything that reads it before handshaking.

Tests cover both directions: crossing the fork re-derives the hash, drops `fork_next` to 0 and bumps the seq; an unchanged id leaves the record byte-identical.

**Checklist**

### DIFF
--- a/crates/networking/p2p/discovery/server.rs
+++ b/crates/networking/p2p/discovery/server.rs
@@ -28,7 +28,7 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 use thiserror::Error;
 use tokio::net::UdpSocket;
 use tokio_util::udp::UdpFramed;
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info, trace, warn};
 
 use super::{DiscoveryConfig, codec::DiscriminatingCodec, lookup_interval_function};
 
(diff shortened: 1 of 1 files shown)


ilitteri (approved):

ElFantasma (approved):

ilitteri:
Superseded by #7225 (merged), a from-scratch redo that respects main's discovery restructure. Closing as superseded.

PR #7040, opened 2026-07-28
fix(levm): enforce EIP-8141 canonical low-s for frame signatures

AnkushinDaniil:
## Rationale

EIP-8141 commits the raw signature bytes to the transaction identity hash while eliding them from `sig_hash`. Without a canonical-s rule, a malleated `(v^1, r, n-s)` counterpart of a valid SECP256K1 signature (and the `n-s` counterpart for P256) still verifies against the same `sig_hash` but yields a second valid tx hash for the same logical transaction, bypassing pool dedup and any tx-hash-based tracking.

This PR enforces the canonical low-s form (`s <= n/2`) at block validation, so a high-s frame signature invalidates the whole transaction, and again at mempool admission as an early reject so a malleated duplicate never occupies a pool slot.

It also maps the EIP-8141 signature encoding, where `v` is the raw recovery id (0 or 1), to the 27/28 form expected by the ECRECOVER precompile; any other `v` byte is rejected outright.

Code comments are intentionally minimal; this description and the commit message carry the full rationale.

## Testing

```
cargo test -p ethrex-test --test ethrex_tests levm::eip8141_tests -- --nocapture
```

Result: 76 passed, 0 failed, 0 ignored (751 filtered out). Includes a new test, `frame_tx_high_s_signature_is_rejected_at_block_validation`, which asserts that the low-s and high-s forms share a `sig_hash` but have distinct tx hashes, that the low-s tx is accepted, and that the high-s tx is rejected at block validation with no state leakage.

## Status

Draft-only: cross-client Kurtosis interop evidence is still pending.

### DIFF
--- a/crates/blockchain/blockchain.rs
+++ b/crates/blockchain/blockchain.rs
@@ -3340,9 +3340,9 @@ impl Blockchain {
                 return Err(MempoolError::InvalidFrameSignature);
             }
 
-            // Local anti-malleability policy (NOT consensus): reject high-s
-            // signatures at admission. EIP-8141 accepts high-s at block execution
-            // (`signer == ecrecover`), but the raw signature bytes are committed to
+            // Anti-malleability policy (now also consensus): reject high-s
+            // signatures early at admission. EIP-8141 `verify_signature` (spec commit
+            // fc016d59c5) now rejects high-s too, and the raw signature bytes are committed to
             // the tx identity hash while elided from the sig hash, so a malleated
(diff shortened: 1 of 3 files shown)


edg-l:
Superseded by #7004, which landed on main while this was open. `validate_frame_signatures` now rejects `v > 1` and enforces low-s against `SECP256K1_N_HALF` for secp256k1 and `SECP256R1_N_HALF` for P256, so the canonical-signature rule this PR adds is already in place. Closing as fixed; thanks for spotting it.

PR #7041, opened 2026-07-28
fix(l1): journal reconciliation pre-images at the reorg pivot

edg-l:
**Motivation**

A benchmarkoor run of the `ethrex-bal-full` instance wedged after 628 of 4773 Amsterdam engine fixtures: `628 passed, 4145 failed`, and every failure from #629 on was the same line, 21 856 times:

```
ERROR deep-reorg: side-chain block execution failed number=2
  error=EVM error: DB error: state root missing for block 1 (state_root 0xca4a…9e52)
```

That workload replays every fixture from the same genesis, so each fixture switch is a deep reorg through a pivot near block 1. Once one reconciliation commit had run through that pivot, no later reorg could reach it again and the node never recovered.

**Description**

`commit_to_disk`'s reconciliation path advances disk from the old chain's edge `D` straight to the new chain's `T = pivot + 1` in a single batch, folding the overlay's bridge entries in. `T`'s journal entry is the only route back to the pivot afterwards; the old chain's `[T, D]` entries are `delete_range`d in the same batch.

Its pre-images were read from `read_view`, which is disk *as of `D`*. For every key the old chain rewrote in `[T, D]`, that records `D`'s value instead of the pivot's, so reversing `T` lands on the old chain's state rather than the pivot's. The reconstructed root then does not hash to the pivot's state root, `has_state_root` returns false, and `StoreVmDatabase::new` rejects the replay with `state root missing for block <pivot>`.

The overlay *is* the pivot-vs-`D` difference, so it is exactly the right source: consult it first for the reconciliation layer, fall back to `read_view` for keys it does not carry (unchanged since the pivot, so disk is already correct). `None` in the overlay means "absent at the pivot", which is already the journal's "delete on rollback" entry.

Bridge (`extra_writes`) keys fall out correctly for free: their pivot value is the value being written, so the entry becomes a no-op on rollback.

**Testing**

Two RocksDB integration tests in `test/tests/blockchain/deep_reorg_state_tests.rs`, both driving real execution through `apply_fork_choice_with_deep_reorg`. RocksDB because it is the backend that reaches the commit cadence (`DB_COMMIT_THRESHOLD` = 128); the InMemory threshold of 10 000 is out of reach for a test.

- `reconciliation_journal_entry_reverses_to_pivot_state` builds two 130-block forks off block 1, canonicalizes one so block 1's state is evicted from both disk and the layer cache, deep-reorgs onto the other (which triggers the reconciliation commit at `T = 2`), then rebuilds the overlay from that single journal entry and asserts the pivot's state root is serveable again. This is the step the run above died on.
- `canonical_head_with_evicted_state_becomes_readable` is regression cover for the already-shipped canonical-head pivot (`None if head_is_canonical => head.number`): a forkchoice update onto a canonical head whose state was evicted must expose head's own post-state, not its parent's. There was no end-to-end test for it.

Verified non-vacuous: forcing `pivot_overlay` to `None` fails the first test with the expected assertion, and reverting the canonical-head arm fails the second. `ethrex-storage` (87) and `ethrex-blockchain` (30) suites pass; fmt and clippy clean.

Both tests were dark on CI, so the L1 unit-test job now runs with `--features rocksdb`. The `cfg` reads `ethrex-test`'s own feature, and another crate enabling `ethrex-storage/rocksdb` does not satisfy it, so the file compiled to nothing under a plain `cargo test --workspace`. The flag unlocks 11 tests in total, the two here plus `canonical_commit_gate_tests` and `test/tests/storage/{store_tests,deferred_persistence_tests}`, which sat behind the same gate. It needs no new job and no extra dependency compile: the `rocksdb` crate is already in the `ethrex-test` unit graph via its `ethrex` dev-dependency, and the four deep-reorg tests run in 1.38s.

`release-fast` also gains `debug-assertions` and `overflow-checks`, since test targets are its only consumer. `reorg-tests` was forcing both through per-job `CARGO_PROFILE_RELEASE_FAST_*` env keys, which are now dead config and dropped.

**Not in this PR**

`PersistMessage::Commit(root)` still carries a root snapshot taken off the persist worker. The `has_layer` guard makes a stale root a logged no-op rather than an error, but the flush it was meant to trigger is still dropped rather than re-anchored to the live safe-commit root. Separate from this bug; worth a follow-up.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. N/A: no on-disk schema change. Journal entries written by an unfixed build stay decodable; their pre-images are simply wrong for reconciliation entries, so a node that already wedged needs its pivot re-derived by re-execution, not a re-sync.


ilitteri (approved):

ElFantasma:
Possibly useful while validating this: #7023 reproduces the bug this fixes **reliably in CI**, which may be a stronger signal than the unit test alone.

`tooling/reorgs`' `deep_reorg_state_oracle` fails there with:

```
post-reorg: chain A's slot must be reverted
  left: 1163
 right: 0
```

i.e. the revert landed on the old chain's state rather than the pivot's — the symptom described in this PR's message. It has failed 2/2 CI runs on that branch, while passing intermittently on other `main`-based branches and passing locally on both `main` and that branch (so it looks timing-sensitive, and something about that branch's commit cadence lines the race up consistently).

#7023 changes only the commit *depth policy* for the single-canonical-chain paths (full sync / import), leaving the live path on the canonical gate, so it doesn't appear to be a second bug — just a more reliable trigger for this one. If a CI-level reproducer would help confirm the fix, running the reorg suite with that branch merged on top of this one should exercise it; happy to do that and report back if useful.

test/tests/blockchain/deep_reorg_state_tests.rs
```diff
@@ -0,0 +1,232 @@
+//! Deep-reorg state reachability: after a deep-reorg apply the pivot's state must stay
+//! reconstructible, and an already-canonical head whose state was evicted must become
+//! readable again.
+//!
+//! Both properties are about what `STATE_HISTORY` + disk can reproduce once the layer
+//! cache has been swapped out, so they need the backend that actually reaches the commit
+//! cadence: RocksDB (`DB_COMMIT_THRESHOLD` = 128). The InMemory threshold is 10 000,
+//! far past what a test can build.
+
+#![cfg(feature = "rocksdb")]
```
ElFantasma, line 10: **Neither of these tests runs in CI**, so the regression cover this PR is built around doesn't actually guard the fix.

`test/Cargo.toml` declares `rocksdb = ["ethrex-storage/rocksdb"]` with **no `default`** entry, and the L1 unit-test job runs:

```
cargo test --workspace --exclude 'ethrex-l2*' --exclude ethrex-prover --exclude ethrex-guest-program
```

(`.github/workflows/pr-main_l1.yaml:122`) — no `--features rocksdb`, and there's no `--all-features` anywhere in `.github/workflows/` or the `Makefile`. Feature unification doesn't help either: another crate enabling `ethrex-storage/rocksdb` turns on *that* crate's feature, not the `ethrex-test` crate's own `rocksdb` feature that this `cfg` reads. So the whole file compiles to nothing on CI and `628 passed, 4145 failed` can come back silently.

To be clear this isn't something the PR invented — `canonical_commit_gate_tests.rs` (same directory) and `test/tests/storage/{store_tests,deferred_persistence_tests}.rs` sit behind the same dark gate. But this PR is the one whose stated value is "two RocksDB integration tests ... verified non-vacuous", so it's worth resolving here rather than inheriting the gap.

Adding `--features rocksdb` to that job would light up all four files at once. If there's a workflow I've missed that already does this, ignore me.

crates/storage/store.rs
```diff
@@ -4504,6 +4504,17 @@ fn commit_to_disk(
...
         };
 
+        // Pre-images for the reconciliation layer must be taken at the pivot, not at the
+        // old chain's edge `D`. `T`'s journal entry has to reverse disk back to the pivot
+        // state, but this batch is built while disk still holds `D`, so `read_view` yields
+        // `D`'s values for every key the old chain rewrote in `[T, D]`. The overlay *is*
+        // the pivot-vs-`D` difference, so it wins over `read_view` for any key it carries.
+        // `None` in the overlay means "absent at the pivot", which is exactly the journal's
+        // "delete on rollback" entry.
+        let pivot_overlay = is_reconciliation_layer
```
ElFantasma: nit: this reduces to `overlay_for_reconciliation.as_ref()`.

`is_reconciliation_layer` is defined as `overlay_for_reconciliation.is_some()` (line 4500), so the two arms are:
- `Some` → `true.then_some(Some(&ov))` → `.flatten()` → `Some(&ov)`
- `None` → `false.then_some(None)` → `None` → `.flatten()` → `None`

which is exactly what `as_ref()` yields. The bool round-trip reads like it's adding a condition that isn't there.

Also loop-invariant — it and `extra` (4501) only depend on `overlay_for_reconciliation`, so both could be hoisted above `'layers` next to `extra_writes`. Given the `debug_assert` on 4437 pins this to one layer it makes no practical difference; purely readability.

edg-l:
Confirmed working against the glamsterdam devnet-7 benchmarkoor suite (state-actor stateful-engine, 4773 tests). Every test in that suite rewinds to block 1 and rebuilds on top of it, so it hammers this path a few thousand times per run.

Before, image at `072c0c705` ([run](https://benchmarkoor.core.ethpandaops.io/runs/1785237891_b3485055_ethrex-bal-full)), 628/4773 passed:

```
4145x  deep-reorg apply succeeded head_number=1 pivot_number=0 side_chain_len=1
21857x newPayload status is ACCEPTED, expected VALID
21856x forkchoiceUpdated status is SYNCING, expected VALID
```

That is exactly what `canonical_head_with_evicted_state_becomes_readable` covers: the overlay gets built one block below the head, head's own post-state stays unreachable, every payload on top is stashed as ACCEPTED, and the next FCU bounces through another deep reorg. One deep reorg per test for the remaining 4145 tests.

After, image carrying this fix (`2da4139bb`) ([run](https://benchmarkoor.core.ethpandaops.io/runs/1785296808_f891a2e7_ethrex-bal-full)), 4715/4773 passed:

```
55x deep-reorg apply succeeded head_number=1 pivot_number=1 side_chain_len=0
```

and then it stops recurring, since the pivot state stays reachable. Zero ACCEPTED/SYNCING mismatches.

Two caveats so the numbers aren't read as more than they are. That image also carries the rest of the deep-reorg hardening from main, so the whole 628 -> 4715 delta is not this PR alone; what is attributable is `pivot_number=0` -> `pivot_number=1` and the ACCEPTED/SYNCING storm disappearing. And the 58 tests still failing are unrelated to this one: `TrieWrapper::multi_get` skipped the overlay stage of the read cascade, so the merkle prefetch read old-chain nodes off disk during the overlay window and produced a wrong state root with gas, receipts and BAL all matching. Fixed in #6986.

ElFantasma:
Ran the CI-level check I offered earlier — this fix resolves the failure on a reliable reproducer.

[#7078](https://github.com/lambdaclass/ethrex/pull/7078) (now closed) was #7023's head with this PR's commit cherry-picked on top, nothing else. On #7023 alone, `tooling/reorgs`' `deep_reorg_state_oracle` fails 3/3 in CI with `chain A's slot must be reverted; left: 1163, right: 0`. With this fix applied it **passes** (10.4s), and the rest of the reorg suite stays green.

That branch appears to line the race up consistently where other `main`-based branches only hit it intermittently, so if you want an integration-level regression signal for this fix beyond the unit test, running the reorg suite against it is a dependable trigger.

test/tests/blockchain/deep_reorg_state_tests.rs
edg-l: You're right, and I checked the whole chain rather than just taking it: `#![cfg(feature = "rocksdb")]` is a file-level inner attribute, `test/Cargo.toml` has no `default`, `pr-main_l1.yaml:122` and `make test` both pass no `--features`, and nothing in the repo uses `--all-features`. Your unification point holds a fortiori: nothing in the workspace depends on `ethrex-test` at all (it's a leaf), so no other crate can turn on its `rocksdb` feature.

Fixed the way you suggested, on the existing job rather than a new one:

```
cargo test --workspace --features rocksdb --exclude 'ethrex-l2*' ...
```

That lights up all four files, 11 tests. It costs nothing extra to compile, which surprised me until I looked: `ethrex-test` dev-depends on `ethrex` with default features, and `rocksdb` is in that default set, so `ethrex-storage/rocksdb` and the `rocksdb` crate are *already* in this test binary's unit graph (`cargo tree -p ethrex-test -e features` confirms `ethrex-storage feature "rocksdb"`). The flag only flips `ethrex-test`'s own cfg. The four deep-reorg tests run in 1.38s in dev profile, 0.48s in release-fast.

I first wrote it as a separate job on `release-fast`, then threw that away. Its only justification would have been test runtime, and at 1.38s there isn't any; it would have added a cold ~30 min workspace build plus another cache key, which is the opposite of what #7029 is trying to do.

crates/storage/store.rs
edg-l: Applied, it's `overlay_for_reconciliation.as_ref()` now. The bool round-trip was left over from an earlier shape where the condition wasn't the same predicate, and you're right that it reads as though it still adds one.

Left the hoist alone. `is_reconciliation_layer` is still what selects `extra`, and moving it plus the two comment blocks above `'layers` is ~20 lines of churn for something the `debug_assert!` at 4437 already pins to one layer. Happy to do it if you'd rather have it in this PR.

ilitteri (approved):

PR #7042, opened 2026-07-28
fix(l1): restore deep-reorg pivot state reachability

edg-l:
Targets `glamsterdam-devnet-7`. Same reconciliation fix as #7041 (against `main`), plus two backports this branch is missing because it sits 39 commits behind.

**Motivation**

The benchmarkoor `ethrex-bal-full` run on the `glamsterdam-devnet-7` image (`072c0c705`) wedged after 628 of 4773 Amsterdam engine fixtures: `628 passed, 4145 failed`. Every failure from #629 on was the same line, 21 856 times:

```
ERROR deep-reorg: side-chain block execution failed number=2
  error=EVM error: DB error: state root missing for block 1 (state_root 0xca4a…9e52)
```

plus three of these earlier in the run:

```
ERROR commit_to_disk failed: commit(0x5b30…b6e7) returned None; layer cache invariant violated
```

That workload replays every fixture from the same genesis, so each fixture switch is a deep reorg through a pivot near block 1. Three distinct defects compound:

1. **Reconciliation journal pre-images are taken at the old chain's edge, not the pivot** — the wedge. Still present on `main`; fixed here and in #7041.
2. **The pivot for an already-canonical head is computed as `head - 1`** — the overlay serves the parent's state and head's own state is never re-materialized, so `reorg_apply_deep` reports success while leaving head unreachable. Every payload built on head is then stashed as ACCEPTED and drives another deep reorg. Visible in the log as `deep-reorg apply succeeded head_number=1 pivot_number=0` immediately followed by `newPayload status is ACCEPTED`. Already fixed on `main`.
3. **A commit whose target root has no layer errors instead of skipping** — the `PersistMessage::Commit(root)` path forwards the safe-commit root without consulting the cache, so a root that was already flushed (or whose cache a reorg replaced) produces the hard error above. Already fixed on `main`.

**Description**

The reconciliation fix (1): `commit_to_disk` advances disk from the old chain's edge `D` straight to `T = pivot + 1` in one batch and `delete_range`s the old `[T, D]` journal entries, so `T`'s entry is the only route back to the pivot. Its pre-images came from `read_view` — disk as of `D` — so reversing `T` landed on the old chain's state and the reconstructed root did not hash to the pivot's. The overlay is precisely the pivot-vs-`D` difference, so it now supplies the pre-image for any key it carries, with `read_view` as the fallback for keys unchanged since the pivot.

(2) and (3) are copied verbatim from `main` (`None if head_is_canonical => head.number` in `reorg_apply_deep`, and the `has_layer` / `layer_count` skip guard at the top of `commit_to_disk`) so merging `main` into this branch later is conflict-free.

**Testing**

Branch gate — `make -C tooling/ef_tests/engine test`:

```
fixture result: ok. 74362 passed; 0 failed; 0 skipped; 0 filtered; 74362 total
```

Two new RocksDB integration tests in `test/tests/blockchain/deep_reorg_state_tests.rs`, both driving real execution through `apply_fork_choice_with_deep_reorg`:

- `reconciliation_journal_entry_reverses_to_pivot_state` — two 130-block forks off block 1; canonicalize one so block 1's state is evicted from disk and cache, deep-reorg onto the other (triggering the reconciliation commit at `T = 2`), then rebuild the overlay from that single journal entry and assert the pivot's state root is serveable. This is the step the run above died on.
- `canonical_head_with_evicted_state_becomes_readable` — pins (2): a forkchoice update onto a canonical, state-evicted head must expose head's own post-state.

Plus `has_layer_tracks_layer_residency` in `crates/storage/layering.rs` covering the guard from (3).

Verified non-vacuous: forcing `pivot_overlay` to `None` fails the first test, and reverting the canonical-head arm fails the second. `ethrex-storage` (81) and `ethrex-blockchain` suites pass; fmt and clippy clean.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A: no on-disk schema change.

### DIFF
--- a/crates/blockchain/fork_choice.rs
+++ b/crates/blockchain/fork_choice.rs
@@ -461,6 +461,13 @@ async fn reorg_apply_deep(
 
     let pivot_number = match new_canonical_blocks.last() {
         Some((n, _)) => n.saturating_sub(1),
+        // Case 1 (head already canonical): the fix is unwinding the blocks ABOVE
+        // head, so the pivot is head itself — the overlay range [head+1, edge]
+        // captures `serves_root = head.state_root`, making head's state readable
+        // again. Using head - 1 here would serve head's PARENT root while head is
+        // never re-executed, leaving head's state unreachable forever.
+        None if head_is_canonical => head.number,
+        // Case 2 (parent canonical, head not): serve the parent's root and replay head.
         None => head.number.saturating_sub(1),
     };
 
--- a/crates/storage/layering.rs
+++ b/crates/storage/layering.rs
@@ -261,6 +261,19 @@ impl TrieLayerCache {
         self.commit_threshold
     }
 
+    /// Whether `state_root` has a diff layer. Not every canonical root does: [`Self::put_batch`]
+    /// skips blocks whose state root equals their parent's, and flushed roots are pruned. Callers
+    /// holding a root from outside the cache must check this before treating it as committable.
+    pub fn has_layer(&self, state_root: H256) -> bool {
+        self.layers.contains_key(&state_root)
+    }
+
+    /// Number of diff layers held in memory. Diagnostic only: distinguishes a pruned root from
+    /// an empty cache.
+    pub fn layer_count(&self) -> usize {
+        self.layers.len()
+    }
+
     fn create_filter(expected_items: usize) -> AtomicBloomFilter<FxBuildHasher> {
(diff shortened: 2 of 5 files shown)


edg-l:
Landing directly on `glamsterdam-devnet-7` instead.

PR #108, opened 2024-07-01
feat: implement `Pong` message encoding

ricomateo:
This PR depends on #92 

**Motivation**

Having some basic functionality of discv4 

**Description**

Enables `Pong` message encoding.

Closes #80

PR #96, opened 2024-06-28
feat: add struct Decoder helper

MegaRedHand:
**Motivation**

Decoding structs isn't straightforward enough.

**Description**

This PR adds a helper for decoding structs from RLP bytes.

PR #52, opened 2024-06-19
feat: add mocked data to be able to integrate to a kurtosis localnet.

mpaulucci:
**Motivation**
We want to be able to run a localnet with kurtosis

**Description**
Implemented the minimum amount of rpc endpoints to be able to participate in a Kurtosis localnet.

PR #7043, opened 2026-07-28
fix(levm): convert EIP-8141 recovery id to EVM ecrecover v at precompile boundary

AnkushinDaniil:
## Problem

EIP-8141 defines `signature[0]` of a SECP256K1 frame signature as the recovery id, 0 or 1, "matching EIP-2718 typed transactions", and validates with `resolved_signer == ecrecover(msg, v, r, s)` (spec, Signature Validation pseudocode). The EVM `ecrecover` precompile, which ethrex uses to implement this check, expects `v` in {27, 28}.

ethrex forwarded the recovery id byte to the precompile verbatim. A spec-conformant signature with `v = 0` or `v = 1` therefore failed recovery (empty result) and was rejected, while nothing rejected out-of-range values explicitly. This diverges from clients that implement the spec encoding (observed against Nethermind in devnet interop).

## Change

- `validate_frame_signatures` now rejects `v > 1` up front and converts `v` from the spec's 0/1 recovery id to 27/28 only at the precompile call boundary.
- Tests cover: spec encoding 0/1 accepted, legacy EVM encoding 27/28 rejected, and out-of-range values (2, 3, 26, 29, 35, 36, 0xFF) rejected, using real k256 signatures for both parities.

## Evidence

Commands run on this branch:

```
# 78 passed; 0 failed; 0 ignored; 751 filtered out
```

RED/GREEN: before this commit `secp256k1_spec_recovery_id_is_accepted` fails (spec-encoded `v = 0` and `v = 1` are rejected by recovery returning empty) and `secp256k1_evm_v_encoding_is_rejected` fails (`v = 27`/`28` authenticate); the pre-existing `secp256k1_positive_and_tampered` was itself written against the 27/28 encoding and is rewritten here onto the spec encoding. After the commit all three pass together with the rest of the suite.

## Cross-client interop evidence

Mixed-client Kurtosis enclave, Nethermind + ethrex, two lighthouse v8.2.1 CLs, chainId 3151908.

- Images: `ethrex:local` `sha256:c3f3cae6c5014ebbc4bd39eae115f5c3e24db4c41d398face442ba530f9d0866`, built from ethrex `503855613`, which has this PR's commit `aed92df38` as an ancestor; `nethermind:frames-local` `sha256:e0d4aee7651f5a181e2cf07a516c922921316a1a2552eb9cf837f7d9db578d13`.
- Reference frame transaction `0x6f505ce40ce8e2c61b88a8dd10a17f3b7396b3fa823aa290d4ee1f91686528fa` (single SECP256K1 low-s signature, spec recovery id) was accepted at RPC by both clients, returning the same hash, and included in block `#52` `0x8a0fafff525f2cd75f2cf9edf34c18357bb95ec7222b07f2ff8d057d315cff4c` with status `0x1` and `gasUsed 21260` on both.
- Block `#52` matches field-for-field across clients: `hash`, `parentHash`, `stateRoot`, `receiptsRoot`, `transactionsRoot`, `logsBloom`, `gasUsed`, and `blockAccessListHash 0xa6648aab4360a8fc68c2b84125dc67c8208a57a4977459e716cc82bfe0a17bbd`.
- Mutual inclusion in both directions: 12 Nethermind-proposed frame-tx blocks `#96`-`#108` were executed and accepted by ethrex over the engine API; the ethrex-proposed blocks `#52` and a fresh `#181` `0xb74c10cbe1ded12588f6fa36ecef5dfea75504350fe21bc1f57698394b67c401` were accepted by Nethermind.
- 25-slot soak: all 14 frame-tx blocks stayed canonical on both clients with heads in lockstep, and neither EL logged an invalid block over the whole run.

## Consensus risk / limits

Consensus-affecting: it flips acceptance of frame signatures with `v` in {0, 1} (now valid) and {27, 28} (now invalid) on the block-validation path.

Limits of the evidence above:

- The replay image carries the full EIP-8141 fix stack (siblings below), not this commit in isolation, so it shows the fixed encoding interoperating, not this change attributed alone.
- The interop run exercises only the positive direction: frames signed with the spec recovery id are now accepted by both clients. The negative cases (`v` in {27, 28} and out-of-range `v`) are covered by unit tests only; there is still no cross-client conformance fixture for them.
- One mid-run chain split was traced to the two CLs losing P2P connectivity (peer count 0/0), not to EL disagreement: no invalid-block marker appeared on either side, and the chains reconverged after resyncing one beacon node.

## Stack

Targets `frames-devnet-0`. Part of a series of independent EIP-8141 fixes, one PR per logical change, each standing alone against the same base. Siblings: #7040 (canonical low-s), #7044 (intrinsic gas over payload bytes), #7045 (arbitrary-scheme verification gas), #7046 (frame gas refunds), #7047 (frame receipt storage encoding), #7048 (skipped-frame status 2), #7049 (EIP-7623 calldata floor).

### DIFF
--- a/crates/vm/levm/src/vm.rs
+++ b/crates/vm/levm/src/vm.rs
@@ -748,6 +748,16 @@ pub fn validate_frame_signatures(
                 let v = sig.signature[0];
                 let r = &sig.signature[1..33];
                 let s = &sig.signature[33..65];
+                // EIP-8141 defines `signature[0]` as the RECOVERY ID (0 or 1),
+                // matching EIP-2718 typed transactions - not the EVM `ecrecover`
+                // `v` (27/28). Anything above 1 is invalid, so reject it here
+                // rather than letting the precompile's own 27/28 mapping decide:
+                // forwarding the byte verbatim both rejected conformant 0/1
+                // signatures and accepted non-conformant 27/28 ones, which is a
+                // consensus split against a spec-conformant client.
+                if v > 1 {
+                    return false;
+                }
                 // EIP-8141 defines verification as `signer == ecrecover(msg, v, r, s)`
                 // and does NOT mandate EIP-2 low-s, so a high-s frame signature is
                 // spec-valid and MUST be accepted here (this is the consensus
@@ -757,7 +767,13 @@ pub fn validate_frame_signatures(
                 // client that accepts it.
                 let mut calldata = vec![0u8; 128];
                 calldata[..32].copy_from_slice(&msg);
-                calldata[63] = v;
+                // Translate the recovery id to the EVM `ecrecover` `v` only at
+                // the precompile boundary. `v` is 0 or 1 (checked above), so
+                // this mapping is total and exact.
+                calldata[63] = match v {
+                    0 => 27,
+                    _ => 28,
+                };
                 calldata[64..96].copy_from_slice(r);
                 calldata[96..128].copy_from_slice(s);
                 let Ok(result) = crate::precompiles::ecrecover(
--- a/test/tests/levm/eip8141_tests.rs
+++ b/test/tests/levm/eip8141_tests.rs
@@ -2757,6 +2757,139 @@ mod frame_sig_validation_tests {
         }
     }
 
+    /// Sign `msg_hash` with a fixed devnet key and return
+    /// `(recovery_id, r || s, signer)`. `sign_prehash_recoverable` normalizes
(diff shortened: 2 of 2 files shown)


edg-l:
Superseded by #7004. Main converts the bare recovery id at the precompile boundary (`calldata[63] = v.saturating_add(27)` in `validate_frame_signatures`), which is the fix described here. Closing as fixed; thanks.

PR #7044, opened 2026-07-28
fix(l1): charge EIP-8141 intrinsic gas over payload bytes only

AnkushinDaniil:
## Problem

EIP-8141 (Gas accounting) defines the calldata component of a frame transaction's intrinsic gas over the payload byte strings only:

```
frame_data_cost = sum(calldata_cost(frame.data) for frame in tx.frames)
signature_data_cost = sum(
    calldata_cost(sig.signer) + calldata_cost(sig.msg) + calldata_cost(sig.signature)
    for sig in tx.signatures)
```

with `calldata_cost` per standard EIP-7623 rules (4 per zero byte, 16 per nonzero byte).

`FrameTransaction::total_gas_limit` instead RLP-encoded the whole `frames` and `signatures` lists and charged calldata gas over every byte of the encoding, including RLP length prefixes and structural fields (`gas_limit`, `value`, `to`, `mode`, `flags`, `scheme`). That over-charges every frame transaction. In devnet interop replay against Nethermind, the 162-byte reference type-0x06 transaction came out at intrinsic 20662 in ethrex vs the spec value 19778 (delta 884), which makes ethrex reject exact-gas blocks other clients accept.

## Change

Replace whole-RLP charging with per-field charging over the payload byte arrays (`frame.data`, and each signature's `signer`, `msg`, `signature`), keeping saturating arithmetic. Adds regression tests that pin the reference transaction's exact intrinsic gas and assert the intrinsic cost is insensitive to structural frame fields (`gas_limit`, `value`).

## Evidence

Commands run on this branch:

```
cargo check -p ethrex-common                                # pass
# 77 passed; 0 failed; 0 ignored; 751 filtered out
```

## Stack

Targets `frames-devnet-0`. Part of a series of independent EIP-8141 fixes, one PR per logical change, each standing alone against the same base. Siblings: #7040 (canonical low-s), #7043 (recovery id at ecrecover boundary), #7045 (arbitrary-scheme verification gas), #7046 (frame gas refunds), #7047 (frame receipt storage encoding), #7048 (skipped-frame status 2), #7049 (EIP-7623 calldata floor).

## Consensus risk / limits

Consensus-affecting: it lowers intrinsic gas for every frame transaction, changing gas accounting and block validity. Covered by unit tests pinning the reference vector; no cross-client conformance fixture yet.

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2038,21 +2038,24 @@ impl FrameTransaction {
             .sum()
     }
 
-    /// Compute total gas limit: intrinsic + calldata cost (frames + signatures)
-    /// + signature verification cost + sum of frame gas limits.
+    /// Computes the total gas limit defined by EIP-8141.
     pub fn total_gas_limit(&self) -> u64 {
-        let mut calldata_gas: u64 = 0;
-        // RLP-encode frames to compute calldata cost
-        let mut frames_buf = Vec::new();
-        self.frames.encode(&mut frames_buf);
-        for byte in &frames_buf {
-            calldata_gas = calldata_gas.saturating_add(if *byte == 0 { 4 } else { 16 });
+        fn add_calldata_gas(calldata_gas: u64, bytes: &[u8]) -> u64 {
+            bytes.iter().fold(calldata_gas, |acc, byte| {
+                acc.saturating_add(if *byte == 0 { 4 } else { 16 })
+            })
         }
-        // RLP-encode signatures to compute their calldata cost
-        let mut sigs_buf = Vec::new();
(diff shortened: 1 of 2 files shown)


edg-l:
Superseded by #7004, which changed the calldata charge to `data_fields()` over `frame.data` plus each signature's `signer`/`msg`/`signature`, dropping the RLP framing. That is what this PR set out to do.

Worth salvaging though: main has no intrinsic-gas-accounting tests, so the `intrinsic_gas_accounting_tests` module here is still useful coverage over code main already has. A tests-only PR against current main would be welcome. Closing the implementation half.

AnkushinDaniil:
Opened the tests-only PR against main: #7059

PR #7045, opened 2026-07-28
feat(l1): charge EIP-8141 arbitrary frame signatures their own gas

AnkushinDaniil:
## Problem

The EIP-8141 signature scheme table assigns a verification gas cost to every scheme, including `ARBITRARY`:

| scheme byte | scheme | verification gas |
|---|---|---|
| 0x0 | ARBITRARY | 100 |
| 0x1 | SECP256K1 | 2800 |
| 0x2 | P256 | 6700 |

`FrameTransaction::signature_verification_cost` charged 0 for `ARBITRARY` entries (it was written against an older spec snapshot that predates the 100-gas entry). That under-charges `total_gas_limit` by 100 per `ARBITRARY` signature and also under-counts the validation-prefix verify-gas budget, so the `MAX_VERIFY_GAS` boundary sits 100 gas away from where a conformant client puts it.

## Change

- Charge 100 gas per `ARBITRARY` signature in `signature_verification_cost`.
- Tests: `total_gas_limit` differs by exactly `2800 - 100` between an otherwise-identical SECP256K1 and ARBITRARY transaction (with an RLP-shape guard so the comparison isolates verification cost), and the prefix verify-gas budget flips from valid to `VerifyGasBudgetExceeded` exactly at the 100-gas boundary.

## Evidence

Commands run on this branch:

```
cargo check -p ethrex-common                                # pass
cargo test -p ethrex-common transaction::tests
# 69 passed; 0 failed; 74 filtered out
# 75 passed; 0 failed; 751 filtered out
```

## Stack

Targets `frames-devnet-0`. Part of a series of independent EIP-8141 fixes, one PR per logical change, each standing alone against the same base. Siblings: #7040 (canonical low-s), #7043 (recovery id at ecrecover boundary), #7044 (intrinsic gas over payload bytes), #7046 (frame gas refunds), #7047 (frame receipt storage encoding), #7048 (skipped-frame status 2), #7049 (EIP-7623 calldata floor).

## Consensus risk / limits

Consensus-affecting: it raises `total_gas_limit` (and the verify-gas budget accounting) for any frame transaction carrying `ARBITRARY` signatures. Covered by unit tests only; no cross-client conformance fixture yet.

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2023,14 +2023,13 @@ impl FrameTransaction {
         keccak(&buf)
     }
 
-    /// Per EIP-8141 (spec commit fe0940cae2): 2800 gas per SECP256K1 signature,
-    /// 6700 per P256. Unknown schemes are rejected by static validation, so
-    /// they are treated as 0 here (validation runs first).
+    /// Per EIP-8141: 100 gas per ARBITRARY signature, 2800 per SECP256K1, and
+    /// 6700 per P256. Unknown schemes are rejected by static validation.
     pub fn signature_verification_cost(&self) -> u64 {
         self.signatures
             .iter()
             .map(|s| match s.scheme {
-                FRAME_SIG_SCHEME_ARBITRARY => 0u64,
+                FRAME_SIG_SCHEME_ARBITRARY => 100u64,
                 FRAME_SIG_SCHEME_SECP256K1 => 2800u64,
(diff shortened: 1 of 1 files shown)


edg-l:
Superseded by #7004: `signature_verification_cost` on main already charges `FRAME_SIG_SCHEME_ARBITRARY => 100`, and main covers the scheme gas table in its own tests. Closing as fixed; thanks.

PR #7046, opened 2026-07-28
fix(levm): apply frame transaction gas refunds

AnkushinDaniil:
## Problem

EIP-8141 (Gas refunds) requires EIP-3529 storage refunds to be applied to frame transactions:

```
gas_used_before_refund = tx_gas_limit - tx_unused_gas
applied_refund = min(refund_counter, gas_used_before_refund // 5)
gas_used = gas_used_before_refund - applied_refund
```

with the refund counter accumulated across frames and the deltas of reverted frames (or frames unrolled by a failed atomic batch) discarded, and the refund "returned to the gas payer ... and added back to the block gas pool".

ethrex's frame-tx finalization ignored the accumulated `refunded_gas` counter entirely: the payer was charged for the gross gas used, the coinbase fee was computed on gross gas, and the execution report's `gas_used`/`gas_spent` were gross. A frame transaction that clears storage therefore reports a higher `gas_used` than a conformant client and charges the payer more, which breaks receipt cumulative-gas and block-gas accounting against other clients. `gas_refunded` in the report was also overloaded to mean "unused frame gas" rather than the EIP-3529 refund.

## Change

- Apply `applied_refund = min(refunded_gas, total_gas_used / MAX_REFUND_QUOTIENT)` at frame-tx finalization; substate checkpoints already discard refund deltas from reverted frames and unrolled batches.
- Charge the payer and pay the coinbase priority fee on the net (post-refund) gas.
- Report `gas_used`/`gas_spent` net of the refund and `gas_refunded` as the applied EIP-3529 refund.

Tests cover: a successful frame netting payer balance, coinbase fee and block gas; refund deltas of a reverted frame discarded; refund deltas of a reverted atomic batch discarded; and the `gas_used / 5` cap binding.

## Evidence

Commands run on this branch:

```
cargo check -p ethrex-levm                                  # pass
cargo test -p ethrex-test --test ethrex_tests levm::eip8141_tests
# 79 passed; 0 failed; 0 ignored; 751 filtered out
cargo fmt --all -- --check                                  # pass
git diff --check                                            # pass
```

## Stack

[...]

PR #7047, opened 2026-07-28
fix(l1): decode stored EIP-8141 frame receipts without MalformedBoolean

AnkushinDaniil:
## Problem

EIP-8141 gives frame transactions their own receipt shape (`ReceiptPayload = [cumulative_gas_used, payer, [frame_receipt, ...]]`, no top-level `succeeded`), so ethrex stores frame receipts with a dedicated storage encoding: every receipt write path uses `Receipt::encode_storage()` and every read path uses `Receipt::decode_storage()`.

The deferred-persistence flush path (`flush_block_data`) was the one exception: it wrote receipts to `RECEIPTS_V2` with `encode_to_vec()`, the consensus encoding. For legacy receipt types the two encodings happen to coincide, so this went unnoticed. For a frame receipt they differ, and reading the flushed receipt back (e.g. `eth_getTransactionReceipt` after the block is flushed from the deferred buffer) fails to decode with `RLPDecodeError::MalformedBoolean`.

## Change

- Use `receipt.encode_storage()` in `flush_block_data`, matching the direct write paths.
- Add a `testing`-gated helper to buffer a block with receipts, and a regression test that round-trips a frame receipt through the deferred buffer and a flush, asserting it decodes back unchanged.

## Evidence

Commands run on this branch:

```
cargo check -p ethrex-storage                               # pass
cargo test -p ethrex-test --test ethrex_tests storage::deferred_persistence_tests
# 16 passed; 0 failed; 0 ignored; 811 filtered out
```

## Stack

Targets `frames-devnet-0`. Part of a series of independent EIP-8141 fixes, one PR per logical change, each standing alone against the same base. Siblings: #7040 (canonical low-s), #7043 (recovery id at ecrecover boundary), #7044 (intrinsic gas over payload bytes), #7045 (arbitrary-scheme verification gas), #7046 (frame gas refunds), #7048 (skipped-frame status 2), #7049 (EIP-7623 calldata floor).

## Consensus risk / limits

Not consensus-affecting: this is a local storage-encoding bug; block execution and receipt roots are unchanged. It affects RPC availability of frame receipts once a block leaves the deferred buffer. Covered by an in-memory store round-trip test; not exercised against a RocksDB-backed store here.


edg-l:
Thanks, this looks like a real gap: main's `store.rs` has no frame-receipt-specific decode handling, so nothing has addressed it.

AnkushinDaniil:
The red check is the Reorg Tests job, and it looks like an infrastructure flake rather than a regression from this PR.

What the CI log actually shows is the ethrex node process exiting non-zero mid-simulation (`assertion failed: res.unwrap().success()` in `tooling/reorgs/src/simulator.rs:160`), after which the RPC call fails with a connection abort against `localhost:8594`. So the node went away, not an assertion about reorg behaviour.

I ran the same suite locally on this branch's head (`cargo build --bin ethrex` then `cd tooling/reorgs && cargo run`, exactly what the job does) and it exits 0 with every test passing.

I also checked that the change itself cannot break receipt reads: every writer goes through `Receipt::encode_storage` and every reader through `Receipt::decode_storage`, and this PR only brings the deferred-flush write onto the same codec the direct write already used. A codec mismatch would break the direct path too, which is covered by existing tests.

I do not have rerun rights on this repo, so could someone re-trigger that job?

AnkushinDaniil:
Closing this. The base branch has since landed the receipt decode fix itself (Receipt::decode_storage in store.rs) and independently reworked the EIP-8141 signature-scheme paths these commits touched, so five of the six commits here are now redundant. The only net-new piece left is the regression test frame_receipt_survives_deferred_flush_and_eviction, which I will reopen as a focused single-commit PR against the current base if it is still wanted.

PR #7048, opened 2026-07-28
fix(l1): report status 2 for skipped frames and keep rolled-back batch frames' gas

AnkushinDaniil:
## Problem

EIP-8141 uses status `0x2` for frames skipped by a failed atomic batch. #7004 already moved `FRAME_RECEIPT_STATUS_SKIPPED` to 2 on main, but a few doc comments still say 3, and no test pins the consensus bytes or the FRAMEPARAM(0x05) value.

## Change

- Fix the stale doc comments: rpc `types/receipt.rs`, levm `errors.rs`, levm `vm.rs`, and the test comment in common `receipt.rs`. No production code changes on main.
- The receipt RLP round-trip test now pins the consensus bytes of a skipped frame receipt literally (`[0xc3, 0x02, 0x80, 0xc0]`), so a status-code regression fails at the encoding layer.
- New end-to-end test: a five-frame transaction where an atomic batch fails asserts the skipped frame's `frame_results` status is 2 and that FRAMEPARAM(0x05) observed from a later frame reads 2.

## Evidence

Commands run on this branch (rebased on current main):

```
cargo test -p ethrex-common --lib types::receipt
# 17 passed; 0 failed
# 85 passed; 0 failed
cargo check -p ethrex-rpc -p ethrex-levm    # pass
cargo fmt --check                           # pass
```

## Notes

Originally targeted `frames-devnet-0`, retargeted to main as requested. frames-devnet-0 still has the constant at 3 (it predates #7004), so a cherry-pick of this PR alone will not fix the value there - the new tests would fail on that branch until it picks up #7004's constant change.


edg-l:
Confirmed, and main is the one that's wrong here. Current EIP-8141 specifies `0x2` for frames skipped by a failed atomic batch, in both the receipt section and the TXPARAM status table. Main still uses `3` (see the doc comment on `frame_results` in `vm.rs`), a stale spec revision that #7004 did not catch. Good find.

edg-l (approved):

edg-l:
If possible, this pr should be targeted to main, and then we can cherry pick it to the frames-devnet-0 branch

AnkushinDaniil:
Done, retargeted to main and rebased. Note that #7004 already sets the constant to 2 on main, so here this PR is now the stale doc comments plus the regression tests. frames-devnet-0 still has 3, so a cherry-pick there also needs the constant change (or a rebase onto main).

ilitteri (approved):

test/tests/levm/eip8141_tests.rs
```diff
@@ -568,6 +568,102 @@ fn frameparam_reads_frame_index_from_stack_top() {
...
+    // (not 0x3). Layout:
+    //   frame0: VERIFY on the sender -> APPROVE(3) -> payer + sender_approved.
+    //   frame1: DEFAULT, atomic-batch flag, reverts -> unrolls the batch.
+    //   frame2: DEFAULT, atomic-batch flag -> SKIPPED.
+    //   frame3: DEFAULT, no flag -> batch terminator, also SKIPPED.
+    //   frame4: DEFAULT, no flag -> reads FRAMEPARAM(0x05, 2) and stores it.
+    let reverter = Address::from_low_u64_be(0xD1);
+    let stop_ct = Address::from_low_u64_be(0xD2);
+    let reader = Address::from_low_u64_be(0xD3);
+    let tx = frame_tx_with_frames(vec![
```
ElFantasma, line 604: This layout covers the settled half of batch-revert status but not the contested half.

In `vm.rs`, `skip_until_batch_end` assigns `FRAME_RECEIPT_STATUS_SKIPPED` only to frames *at or after* the failure (through the terminator). Frames that already executed-and-succeeded earlier in the same batch never go through that branch — per `docs/eip-8141.md` §11 they are recorded as `0x0` (FAILURE) and charged their full `gas_limit`. Here frame1 *is* the failing frame, so there's no succeeded-then-rolled-back frame in the tx at all.

That matters because §11 marks exactly this as open:

> **Status — UNRESOLVED, pending cross-client confirmation.** […] The canonical per-frame **gas** […] and **status** (`0x0` FAILURE vs `0x2` SKIPPED) for rolled-back batch frames are a consensus decision that must be aligned with the reference / other clients before being pinned.

Inserting a flagged STOP frame *before* the reverter would cover it — asserting `0x0` + full `gas_limit` today documents the known divergence from the single-frame path, and that assert is what will fail when §11 gets resolved, which is the useful signal. A `docs/eip-8141.md §11` reference in the test's doc comment would also help whoever resolves it find this test.

test/tests/levm/eip8141_tests.rs
```diff
@@ -568,6 +568,102 @@ fn frameparam_reads_frame_index_from_stack_top() {
...
+        ),
+    ];
+    let (result, db) = run_frame_tx(&accounts, tx);
+    let report = result.expect("a DEFAULT-frame batch revert must not invalidate the tx");
+
+    let frame_results = report
+        .frame_results
+        .expect("frame tx report must carry per-frame results");
+    assert_eq!(
+        frame_results[2].0, FRAME_RECEIPT_STATUS_SKIPPED,
```
ElFantasma: Two cheap assertions worth adding while you're here, both on things the frame layout already sets up:

- **frame3, the batch terminator.** The comment above says "no flag -> batch terminator, also SKIPPED", but nothing checks it. That's a distinct branch in `vm.rs` (`if frame_idx == end_idx` — the terminator is included in the skip, then `skip_until_batch_end` is cleared), and terminator inclusion is the subtle part of batch semantics. `assert_eq!(frame_results[3].0, FRAME_RECEIPT_STATUS_SKIPPED)` pins it.
- **`gas_used == 0` on skipped frames.** `vm.rs` records `(SKIPPED, 0, vec![])` and deliberately does *not* add the frame's `gas_limit` to `total_gas_used`, which feeds the block header `gasUsed`. So `frame_results[2].1` is consensus-visible and currently unasserted — one line pins the refund half of the behaviour alongside the status half.

ElFantasma (commented):
Two test-coverage suggestions inline, both non-blocking.

test/tests/levm/eip8141_tests.rs
AnkushinDaniil: Done. The layout now carries a flagged frame that executes and succeeds before the reverter, so the test covers both halves: that frame is rewritten as `0x0` FAILURE and charged its full `gas_limit`, which is what `docs/eip-8141.md` §11 still marks UNRESOLVED, and the doc comment on the test says so and names §11 so whoever resolves it finds this assertion. If §11 lands on SKIPPED-and-refunded instead, this test fails loudly rather than the behaviour changing quietly.

test/tests/levm/eip8141_tests.rs
AnkushinDaniil: Both added: the terminator is now asserted as SKIPPED (that branch clears `skip_until_batch_end`), and `gas_used == 0` is asserted on both skipped frames since it feeds the header `gasUsed`. FRAMEPARAM now reads the shifted index.

test/tests/levm/eip8141_tests.rs
```diff
@@ -568,6 +568,140 @@ fn frameparam_reads_frame_index_from_stack_top() {
     );
 }
 
+/// FRAMEPARAM(param=0x05, frameIndex=3) -> status of frame[3], then SSTORE at slot 0.
+/// Bytecode: PUSH1 0x05 (param), PUSH1 0x03 (frameIndex - top), FRAMEPARAM (0xB3),
+///           PUSH1 0x00 (slot key), SSTORE (0x55), STOP (0x00).
```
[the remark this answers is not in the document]
AnkushinDaniil: Reworded: the doc comment now states that FRAMEPARAM takes frameIndex from the top of the stack and param below it, so 0x03 (the frameIndex) is pushed last. The "- top" shorthand is gone.

crates/common/types/receipt.rs
```diff
@@ -744,16 +744,25 @@ mod test {
 
     #[test]
     fn test_frame_receipt_skipped_status_rlp_roundtrip() {
-        // Spec line 137: status code 0x3 marks frames skipped by a failed atomic batch.
+        // Spec line 137: status code 0x2 marks frames skipped by a failed atomic batch.
         let fr = FrameReceipt {
```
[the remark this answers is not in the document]
AnkushinDaniil: Replaced with a section reference (EIP-8141 §Receipt Format) rather than a line number, for the reason you give: this PR is about stale doc comments, so it should not add a reference that goes stale on the next spec edit.

ElFantasma (approved):

AnkushinDaniil:
Pushed a second fix on top, found by running the two clients against each other on the frames devnet.

The chain split at block 717 with Nethermind rejecting the ethrex block: `HeaderGasUsedMismatch: expected 80265, got 21825`. The transaction was a three-frame one whose atomic batch reverted. On batch revert this branch rewrote every frame in the batch as FAILURE and charged it its full `gas_limit`, so `total_gas_used` (and with it the header `gasUsed`, the receipts trie and the payer refund) counted gas the transaction never spent.

The EIP settles this in Behavior: "Logs emitted by frames that executed before the failure are discarded together with their state changes when the batch is unrolled. Those frame receipts retain their execution status and gas used, with empty logs." So the rolled-back frames keep their status and gas, and only their logs go away. That also removes the inconsistency `docs/eip-8141.md` §11 flagged, where the same REVERTing frame was billed differently inside a batch than outside one; §11 is updated accordingly.

The test in this PR pinned the old behaviour, so it moves with the fix: the rolled-back frame now emits a log and is asserted to keep SUCCESS, keep the gas it actually used, and lose its log. Verified in reverse — with the fix reverted the test fails. 85 eip8141 tests pass.

edg-l (approved):
Re-approving now that f86c50d1 is in, since my earlier approval was on the docs and tests only.

The batch revert change looks right. Checked it against the EIP text and the Behavior section is explicit that frames which ran before the failure keep their execution status and gas used and only lose their logs, so this matches. Also traced the refund counter: `push_backup` snapshots `refunded_gas` at batch entry and `revert_backup` restores it, so keeping the frames' gas doesn't leave us handing back an SSTORE refund for state that got unrolled. Receipt side is a straight pass through into the frame receipts, no surprises there.

One nit, not blocking: `docs/eip-8141.md` around line 400 still has the old parenthetical saying the atomic batch revert path "currently charges full gas_limit uniformly, which is inconsistent with this single-frame rule and is spec-silent, see Spec Gaps §11". That now contradicts the §11 rewrite further down in the same file. Worth deleting while you're in there.

Would also be nice to assert `report.gas_used` in the new test, since the header gasUsed is the thing Nethermind actually rejected and right now it's only pinned indirectly through the per frame values. Fine to skip.

Needs a rebase, main moved a few commits but nothing touching these files.

edg-l:
@AnkushinDaniil hi, sorry for the trouble but our repo requires all commits to have verified signatures, if you could sign all commits we can merge this then

AnkushinDaniil:
Sorry about that, signing was not set up on my side. I added an SSH signing key and re-signed the commits on all nine open PRs, force-pushed just now. GitHub shows them verified.

One commit here is still unverified and it is not mine: the `Merge branch 'main'` commit by @estebandh, signed with a key GitHub does not recognise for that account. Same on 7059. Let me know if you want me to rebase those away.

ElFantasma:
> One commit here is still unverified and it is not mine: the `Merge branch 'main'` commit by @estebandh, signed with a key GitHub does not recognise for that account. Same on 7059. Let me know if you want me to rebase those away.

Hi @AnkushinDaniil , I could not find the commit you mention here. But I did find it on #7059. I think the problem is that you re-signed my `Merge branch 'main'` commit with your own key, and as such, it's not recognized by github as mine. Let me see if I can re-re-sign myself to let it verify.

PR #7049, opened 2026-07-28
feat(levm): apply the EIP-7623 calldata floor to frame transactions

AnkushinDaniil:
## Problem

EIP-8141 (Gas accounting) applies the EIP-7623 calldata floor to a frame transaction's frame and signature data:

```
calldata_tokens = sum(tokens_in(frame.data) for frame in tx.frames)
                + sum(tokens_in(sig.signer) + tokens_in(sig.msg) + tokens_in(sig.signature)
                      for sig in tx.signatures)

charged_gas = mandatory_costs + max(frame_data_cost + signature_data_cost + execution_gas,
                                    TOTAL_COST_FLOOR_PER_TOKEN * calldata_tokens)
```

and "A transaction whose `tx_gas_limit` is below `FRAME_TX_INTRINSIC_COST + len(tx.frames) * FRAME_TX_PER_FRAME_COST + signature_verification_cost + TOTAL_COST_FLOOR_PER_TOKEN * calldata_tokens` is invalid, so the floor is reserved independently of execution".

ethrex applied no calldata floor to frame transactions at all: a data-heavy frame tx with little execution paid only the standard per-token calldata cost, and a tx whose gas limit cannot cover the floor was accepted. Both diverge from a conformant client (reference tx gasUsed target: 18750 mandatory + 10 * 251 tokens = 21260).

## Change

- New `frame_tx_floor_charged_gas` helper in levm utils: `mandatory + TOTAL_COST_FLOOR_PER_TOKEN * calldata_tokens`, with tokens counted over the decoded payload byte fields (frame data; signature signer/msg/signature) per EIP-7623's `tokens_in`, at 10 gas per token (the EIP-7623 constant, not the EIP-7976 override; frame-tx path only).
- Reject a frame tx whose `total_gas_limit()` cannot reserve the floor, on both the execution path and the admission/simulation path (`IntrinsicGasBelowFloorGasCost`).
- At finalization, charge `max(total_gas_used, floor_charged_gas)` before the refund split.

Tests cover a floor-dominant tx charging exactly the floor, an execution-dominant tx unaffected by the floor, and a sub-floor gas limit being invalid.

Note on the floor-dominant test: its sanity guard (proving the execution arm really is the smaller one) originally derived the un-floored charge from `gas_refunded`. That derivation is circular against this branch's report semantics and also does not hold once #7046 redefines `gas_refunded` as the EIP-3529 refund. The guard here computes the un-floored charge from the per-frame gas in the execution report instead, which is well defined in both worlds. The assertions on the charged gas are unchanged.

## Evidence

Commands run on this branch:

```
# 78 passed; 0 failed; 0 ignored; 751 filtered out (incl. all 3 floor tests)
```

## Stack

Targets `frames-devnet-0`. Part of a series of independent EIP-8141 fixes, one PR per logical change, each standing alone against the same base. Siblings: #7040 (canonical low-s), #7043 (recovery id at ecrecover boundary), #7044 (intrinsic gas over payload bytes), #7045 (arbitrary-scheme verification gas), #7046 (frame gas refunds), #7047 (frame receipt storage encoding), #7048 (skipped-frame status 2).

## Consensus risk / limits

Consensus-affecting: it changes gas charged for data-heavy frame transactions and rejects sub-floor gas limits. The floor interacts with the refund change in #7046: there the floored charge feeds the EIP-3529 refund split; on this standalone branch the max() applies to the unrefunded charge (the cherry-pick onto `frames-devnet-0` was resolved to keep only the floor logic, dropping the #7046 hunks it was stacked on). Covered by unit tests only; no cross-client conformance fixture yet.

### DIFF
--- a/crates/vm/levm/src/utils.rs
+++ b/crates/vm/levm/src/utils.rs
@@ -1035,6 +1035,81 @@ pub fn intrinsic_gas_floor(tx: &Transaction, sender: Address, fork: Fork) -> Res
         .ok_or(InternalError::Overflow.into())
 }
 
+/// EIP-7623 `tokens_in`: `zero_bytes * 1 + nonzero_bytes * 4`, which is exactly
+/// the weighted calldata cost divided by `STANDARD_TOKEN_COST`.
+fn floor_tokens_in_bytes(data: &[u8]) -> Result<u64, VMError> {
+    Ok(gas_cost::tx_calldata(data)? / STANDARD_TOKEN_COST)
+}
+
+/// EIP-8141 Gas Accounting (spec commit c0c9f639): the floor arm of a frame
+/// transaction's `charged_gas`, i.e. the minimum gas the payer can be charged.
+///
+/// ```text
+/// calldata_tokens = sum(tokens_in(frame.data) for frame in tx.frames)
+///                 + sum(tokens_in(sig.signer) + tokens_in(sig.msg)
+///                       + tokens_in(sig.signature) for sig in tx.signatures)
+///
+/// returns FRAME_TX_INTRINSIC_COST
+///       + len(tx.frames) * FRAME_TX_PER_FRAME_COST
+///       + signature_verification_cost
+///       + TOTAL_COST_FLOOR_PER_TOKEN * calldata_tokens
+/// ```
+///
+/// `tokens_in` and `TOTAL_COST_FLOOR_PER_TOKEN` are the EIP-7623 definitions the
+/// spec names verbatim (weighted tokens, 10 gas each) - NOT the EIP-7976
+/// Amsterdam overrides that other tx types use at this fork.
+///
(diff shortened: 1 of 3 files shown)


edg-l:
Superseded by #7004, which added the floor to main: `calldata_tokens`, `calldata_floor_gas`, the static reservation check, and the `max()` in finalization.

One note on the approach here, since it matters if this is revisited. This PR uses EIP-7623's 10/40 model (weighted tokens, `TOTAL_COST_FLOOR_PER_TOKEN = 10`). EIP-7976 redefines those parameters: `floor_tokens_in_calldata = (zero + nonzero) * 4` with the per-token cost at 16, i.e. 64 gas per byte for both zero and non-zero bytes. EIP-7976 is in Glamsterdam and frame transactions exist only from Hegota onward, so 64/64 is what applies, and that is what main implements. Merging this as written would take the floor back to 10/40. Closing.

PR #7050, opened 2026-07-29
ci(l1): build the reorg-test ethrex binary with release-fast

edg-l:
Reorg Tests builds `ethrex` unoptimized, and an unoptimized node overflows the 2 MiB tokio worker stack while serving `engine_getPayload`:

```
INFO Requested payload with id=0x0309d34b7a476e9f
thread 'tokio-rt-worker' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The node aborts, the harness sees a non-zero exit, and the job fails. Which scenario trips it varies per run, hence the flakiness.

Not a recursion bug: the 150/200/300-block deep-reorg scenarios pass in every configuration, so depth does not scale with chain length. It is debug frame bloat.

| Build | Worker stack | Result |
|---|---|---|
| `dev` | 2 MiB | overflows |
| `dev` | `RUST_MIN_STACK=32M` | passes |
| `release-fast` | 2 MiB | passes |

Reproduced on `main` itself, independently of any feature branch. `release-fast` also cuts build time (`opt-level = 3`, no thin-LTO). `deep_reorg.yaml` is unaffected, it already builds `--release`.

### DIFF
--- a/.github/workflows/pr-main_l1.yaml
+++ b/.github/workflows/pr-main_l1.yaml
@@ -504,9 +504,13 @@ jobs:
       # Built with release-grade optimization: an unoptimized ethrex overflows
       # the 2 MiB tokio worker stack while building a payload, since every
       # `.await` in the block-building chain keeps its locals live in a future
-      # that no inlining collapses. `release-fast` keeps opt-level 3 without
-      # thin-LTO, so the job also spends far less time compiling.
+      # that no inlining collapses. The profile keys re-enable the checks that
+      # `release-fast` inherits as off from `release`, so reorg invariants
+      # guarded by `debug_assert!` still panic instead of being skipped.
       - name: Compile ethrex binary
+        env:
+          CARGO_PROFILE_RELEASE_FAST_DEBUG_ASSERTIONS: "true"
+          CARGO_PROFILE_RELEASE_FAST_OVERFLOW_CHECKS: "true"
         run: cargo build --profile release-fast --bin ethrex
 
       - name: Run reorg tests
--- a/.github/workflows/pr-main_l1.yaml
+++ b/.github/workflows/pr-main_l1.yaml
@@ -501,8 +501,13 @@ jobs:
       - name: Setup Rust Environment
         uses: ./.github/actions/setup-rust
 
+      # Built with release-grade optimization: an unoptimized ethrex overflows
+      # the 2 MiB tokio worker stack while building a payload, since every
+      # `.await` in the block-building chain keeps its locals live in a future
+      # that no inlining collapses. `release-fast` keeps opt-level 3 without
+      # thin-LTO, so the job also spends far less time compiling.
       - name: Compile ethrex binary
-        run: cargo build --bin ethrex
+        run: cargo build --profile release-fast --bin ethrex
 
       - name: Run reorg tests
-        run: cd tooling/reorgs && cargo run
+        run: cd tooling/reorgs && cargo run -- ../../target/release-fast/ethrex


ilitteri (approved):

.github/workflows/pr-main_l1.yaml
```diff
@@ -501,8 +501,13 @@ jobs:
...
         uses: ./.github/actions/setup-rust
 
+      # Built with release-grade optimization: an unoptimized ethrex overflows
+      # the 2 MiB tokio worker stack while building a payload, since every
+      # `.await` in the block-building chain keeps its locals live in a future
+      # that no inlining collapses. `release-fast` keeps opt-level 3 without
+      # thin-LTO, so the job also spends far less time compiling.
       - name: Compile ethrex binary
-        run: cargo build --bin ethrex
+        run: cargo build --profile release-fast --bin ethrex
```
ElFantasma, line 514: Endorse the debug-assertions tradeoff already raised — one specific casualty is worth naming, because it lands squarely in this job's blast radius. `crates/storage/store.rs:4578`:

```rust
debug_assert_eq!(
    layer.block_number, t,
    "first new-chain commit must be at the pivot's T height (overlay.to_block)"
);
```

Its own comment states the intent: the assert "catches this loudly in tests; in release we skip (and log) rather than corrupt STATE_HISTORY". That's the deep-reorg journal-reconciliation invariant — precisely what `deep_reorg_side_chain_replay` and `deep_reorg_state_oracle` exist to exercise. Since `release-fast` inherits `release` (`debug-assertions = false`), this job now takes the silent-skip-and-log branch instead of panicking, so a reconciliation regression would go green in CI.

This doesn't have to be traded against build time — the profile keys can be overridden per invocation, without touching `Cargo.toml` or falling back to the slower `release-with-debug-assertions`:

```yaml
      - name: Compile ethrex binary
        env:
          CARGO_PROFILE_RELEASE_FAST_DEBUG_ASSERTIONS: "true"
          CARGO_PROFILE_RELEASE_FAST_OVERFLOW_CHECKS: "true"
        run: cargo build --profile release-fast --bin ethrex
```

Keeps opt-level 3 (so the stack overflow stays fixed) while restoring both checks.

.github/workflows/pr-main_l1.yaml
```diff
@@ -501,8 +501,13 @@ jobs:
       - name: Setup Rust Environment
         uses: ./.github/actions/setup-rust
 
+      # Built with release-grade optimization: an unoptimized ethrex overflows
+      # the 2 MiB tokio worker stack while building a payload, since every
+      # `.await` in the block-building chain keeps its locals live in a future
+      # that no inlining collapses. `release-fast` keeps opt-level 3 without
+      # thin-LTO, so the job also spends far less time compiling.
```
ElFantasma: nit: this clause measures against a baseline the job never used. What's being replaced is `cargo build` at the **`dev`** profile (opt-level 0); `release-fast` is opt-level 3, so this compile step gets *slower*, not faster. `release-fast` is only "far less time" relative to `--release` (thin-LTO, `codegen-units = 1`) — which this workflow never ran. The stack-overflow rationale above is the real justification and stands on its own, so dropping this clause would avoid a future maintainer reading it as a speedup that isn't there.

(Scope note in case it's useful: the harness build is unaffected either way — `tooling/reorgs` belongs to the separate `tooling/` workspace with its own `tooling/target`, so `cargo run` keeps its own dev-profile compile of the `ethrex` lib regardless of this flag. The nodes under test are spawned as subprocesses from the path passed in on line 513, which is why building *that* binary optimized is what fixes the overflow.)

MegaRedHand (approved):

.github/workflows/pr-main_l1.yaml
```diff
@@ -501,8 +501,17 @@ jobs:
...
+      # Built with release-grade optimization: an unoptimized ethrex overflows
+      # the 2 MiB tokio worker stack while building a payload, since every
+      # `.await` in the block-building chain keeps its locals live in a future
+      # that no inlining collapses. The profile keys re-enable the checks that
+      # `release-fast` inherits as off from `release`, so reorg invariants
+      # guarded by `debug_assert!` still panic instead of being skipped.
       - name: Compile ethrex binary
-        run: cargo build --bin ethrex
+        env:
+          CARGO_PROFILE_RELEASE_FAST_DEBUG_ASSERTIONS: "true"
```
ElFantasma, line 512: This does the right thing and the variable names are correct (`CARGO_PROFILE_<PROFILE>_<KEY>`, `release-fast` → `RELEASE_FAST`) — thanks for adding `overflow-checks` on top of what I asked for.

One robustness point on the mechanism. If either name is ever mistyped, or Cargo's handling of profile env vars shifts, **the build proceeds silently without debug assertions** — which is exactly the failure this commit exists to prevent, and it would fail open with no signal in the log. The reorg invariants would stop being checked and the job would still go green.

A named profile fails loudly instead — an unknown `--profile` is a hard error:

```toml
[profile.release-fast-with-debug-assertions]
inherits = "release-fast"
debug-assertions = true
overflow-checks = true
```

```yaml
      - name: Compile ethrex binary
        run: cargo build --profile release-fast-with-debug-assertions --bin ethrex
```

It also matches what the repo already does — `Cargo.toml` defines `release-with-debug-assertions` (`inherits = "release"`, `debug-assertions = true`) and `daily_snapsync.yaml` defaults to it. Neither existing profile gives the combination you need here (opt-3 + no LTO + assertions), so a fourth one is the consistent way to express it rather than a special case.

Non-blocking — the env vars work today. Mostly about which failure mode you'd rather have if this drifts.

ElFantasma (approved):
Both earlier comments are addressed; one follow-up on how the assertions are re-enabled (inline).

PR #7052, opened 2026-07-29
feat(l1): make the frame-tx mempool verify-gas budget configurable

edg-l:
## What

Adds `BlockchainOptions::max_verify_gas` plus a `--mempool.max-verify-gas` CLI flag (env `ETHREX_MEMPOOL_MAX_VERIFY_GAS`), and threads it to the three places that previously read the hardcoded constant:

- the cheap signature-verification-cost pre-filter at mempool admission
- `FrameTransaction::validate_prefix_structure` (prefix frame gas limits + sig cost)
- `LEVM::simulate_frame_validation_prefix` (simulated prefix gas)

## Default is unchanged

`DEFAULT_MAX_VERIFY_GAS` is `FRAME_TX_MAX_VERIFY_GAS`, i.e. the EIP-8141 spec value of 100_000. The constant stays exported.

## Why configurable

EIP-8141 defines `MAX_VERIFY_GAS = 100_000` as mempool policy (SHOULD), not consensus, so it is a peer-policy knob rather than a validity rule. Operators running frame-tx devnets need a larger budget; ours runs 500k.

Covered by a new test in `test/tests/blockchain/mempool_tests.rs` asserting the same tx is rejected under the default budget and admitted under a raised one.


ilitteri (approved):

cmd/ethrex/cli.rs
```diff
@@ -257,6 +257,15 @@ pub struct Options {
         env = "ETHREX_MEMPOOL_MAX_QUEUED_TXS_PER_ACCOUNT"
     )]
     pub mempool_max_queued_txs_per_account: usize,
+    #[arg(
+        help = "EIP-8141 MAX_VERIFY_GAS: maximum gas spent validating signatures and simulating a frame transaction's validation prefix at mempool admission. Mempool policy, not consensus.",
```
ElFantasma, line 289: Worth saying something about the upper end here. This value is the only thing bounding admission-time crypto and simulation work for frame txs, and the three call sites are ordered so the bound applies *before* the work: the sig-cost pre-filter gates `validate_frame_signatures`, `validate_prefix_structure` caps Σ(prefix frame gas limits), and only then does `simulate_frame_validation_prefix` run. That ordering is what makes the knob safe to expose at all.

At the spec 100_000 the pre-filter allows ~35 secp256k1 or ~14 P-256 recoveries per tx (2_800 / 6_700 each). The flag scales that linearly, and with no upper clamp the effective ceiling becomes `MAX_TX_SIZE` instead — 131_072 bytes holds on the order of a thousand P-256 signature entries, so a mistyped value turns admission into a self-DoS on a public-facing node. A `value_parser` upper bound, or one clause in the help text noting the value bounds per-tx signature-verification work, would keep that visible. 500k for the devnet is nowhere near the problem range.

PR #7053, opened 2026-07-29
fix(l1): gate the initcode admission cap on the fork ordinal

edg-l:
Mempool admission chose the EIP-7954 initcode cap from `is_amsterdam_activated(timestamp)`, while levm `validate_init_code_size` chooses it from `fork >= Fork::Amsterdam`.

The two disagree on a chain that schedules a fork after Amsterdam without setting an explicit `amsterdamTime`: the resolved fork is already past Amsterdam so execution allows the larger cap, but the unset field makes admission apply the legacy 49152 cap and reject contract deployments block execution would accept. Both of our Hegota devnets are configured that way (`hegotaTime` set, no `amsterdamTime`).

Admission now resolves the fork once and gates on the ordinal. The EIP-7825 per-tx gas cap a few lines below had the same defect, so it uses `fork >= Osaka && fork < Amsterdam`, matching `default_hook`.

Regression test in `test/tests/blockchain/mempool_tests.rs`: a creation tx with 49153 bytes of initcode on a Hegota-scheduled chain with no `amsterdamTime` must be admitted. It fails with the old field-based gate (`TxMaxInitCodeSizeError`) and passes with the fix.


ilitteri (approved):

crates/blockchain/blockchain.rs
```diff
@@ -3320,9 +3320,18 @@ impl Blockchain {
...
+        // fork after Amsterdam without setting an explicit `amsterdamTime`, the
+        // ordinal is already `>= Fork::Amsterdam` while the field is unset, so a
+        // field-based gate would over-reject transactions block execution accepts.
+        let fork = config.fork(header.timestamp);
+
         // Check init code size
-        // [EIP-7954] - Amsterdam increases the limit
-        let max_initcode_size = if config.is_amsterdam_activated(header.timestamp) {
+        // [EIP-7954] - Amsterdam increases the limit.
+        // Mirrors levm's `validate_init_code_size`.
```
ElFantasma, line 3335: The two gates this hunk rewrites now resolve the fork by ordinal, but three sibling gates in this same function still use the field checks. `is_*_activated` is a pure field test (`self.X_time.is_some_and(|t| t <= ts)`), so each one carries the identical defect:

- **L3339 `is_shanghai_activated`** — the guard on the very statement `max_initcode_size` feeds. levm gates `validate_init_code_size` on `fork >= Fork::Shanghai`, so the new "Mirrors levm's `validate_init_code_size`" comment holds for the cap *value* but not for the activation test. A chain scheduling a post-Shanghai fork without `shanghaiTime` skips the cap at admission while execution enforces it — the mirror image of the bug being fixed, and the worse direction: the tx enters the pool, gets built into a payload, and fails during execution.
- **L3375 `is_prague_activated`** — live on the same genesis shape this PR documents as what Hegota devnets use. `setup_hegota_store_without_amsterdam_time` sets only `shanghai_time` and `hegota_time`, so `prague_time` is `None` while `config.fork()` returns `Fork::Hegota` (ordinal 26 ≥ `Prague`'s 18). A type-4 tx on such a chain is rejected with `Eip7702TxPreFork` at admission even though execution accepts it.
- **L3226 `is_hegota_activated`** — same shape, latent only because Hegota is currently the newest fork, so nothing can be scheduled past it yet.

`fork` is already computed once above, so these become `fork >= Fork::Shanghai` / `>= Fork::Prague` / `>= Fork::Hegota`.

crates/blockchain/blockchain.rs
```diff
@@ -3334,8 +3343,10 @@ impl Blockchain {
             return Err(MempoolError::TxMaxInitCodeSizeError);
         }
 
-        if config.is_osaka_activated(header.timestamp)
-            && !config.is_amsterdam_activated(header.timestamp)
+        // EIP-7825's flat per-tx gas cap applies from Osaka until Amsterdam, which
+        // supersedes it with the EIP-8037 gas model. Mirrors levm's `default_hook`.
+        if fork >= Fork::Osaka
```
ElFantasma, line 3350: The EIP-7825 half of the change has no test. On the genesis the new test uses, old and new gates agree — `osaka_time` is `None`, so `is_osaka_activated` is false and `fork >= Fork::Osaka` is true but `fork < Fork::Amsterdam` is false; both skip the cap. The config that actually changed behavior is `osakaTime` + `hegotaTime` with no `amsterdamTime`: there the old `is_osaka_activated && !is_amsterdam_activated` applied the flat cap while execution (fork = Hegota) did not, and the ordinal form correctly skips it. A second case on that genesis with `gas_limit > POST_OSAKA_GAS_LIMIT_CAP` would pin this half of the fix.

ElFantasma (commented):
Fix is correct; three sibling field-based gates in the same function still carry the same defect (inline).

crates/blockchain/blockchain.rs
edg-l: Both fixed.

All three siblings now go through the ordinal: `fork >= Fork::Shanghai` for the initcode cap, `fork < Fork::Prague` for type-4, `fork < Fork::Hegota` for the frame-tx gate. The last one needed the `fork` binding moved up next to `config`, since it runs earlier in the function; the comment moved with it and now covers both directions, not just over-rejection. Removed a duplicate `config.fork(header.timestamp)` in the frame-signature call while there.

Added the EIP-7825 case on the config you named: `osakaTime` + `hegotaTime`, no `amsterdamTime`, `gas_limit = POST_OSAKA_GAS_LIMIT_CAP + 1`, admitted. Plus one per converted gate: type-4 admitted without `pragueTime`, and initcode over the cap rejected without `shanghaiTime`. The last uses Cancun rather than Hegota because `AMSTERDAM_MAX_INITCODE_SIZE == MAX_TX_SIZE`, so the wire cap fires first and the Amsterdam initcode cap is unreachable at admission; Cancun's 49152 is the only cap an admissible tx can exceed.

PR #7055, opened 2026-07-29
fix(l1): require base_fee_per_gas presence in header validation

ilitteri:
## Summary

`base_fee_per_gas` is RLP-optional in ethrex's header decoder (`decode_optional_field`, since pre-London headers lack it), so a header omitting it decodes fine at any height. `validate_block_header` then validated the missing field as `INITIAL_BASE_FEE` (1 gwei) via `unwrap_or` instead of rejecting the header as malformed.

`base_fee_per_gas` is mandatory for every fork ethrex validates (Merge+), and every other fork-gated header field already has an explicit presence/absence check (`withdrawals_root`, `blob_gas_used`, `excess_blob_gas`, `parent_beacon_block_root`, `requests_hash`, `block_access_list_hash`, `slot_number`). `base_fee_per_gas` was the only one without.

## Change

`crates/common/types/block.rs::validate_block_header`: new `InvalidBlockHeaderError::BaseFeePerGasNotPresent`, checked up front before the expected-fee computation. The existing `unwrap_or` remains as a defensive fallback but is now unreachable for validated headers.

## Impact

No live-network impact (engine payloads always carry `baseFeePerGas`; the RLP-optional path only exists on devp2p sync, where the canonical chain is CL-driven), but this closes the last schema-leniency gap in header validation and matches other clients, which reject the shape at RLP decode.

## Tests

Added `test_validate_block_header_requires_base_fee_present` (presence accepted, absence rejected with `BaseFeePerGasNotPresent`). `cargo test -p ethrex-common --lib types::block` — 26/26 green.

### DIFF
--- a/crates/common/types/block.rs
+++ b/crates/common/types/block.rs
@@ -625,6 +625,8 @@ pub enum InvalidBlockHeaderError {
     GasUsedGreaterThanGasLimit,
     #[error("Gas limit changed more than allowed from the parent")]
     GasLimitTooFarFromParent,
+    #[error("Base fee per gas is not present")]
+    BaseFeePerGasNotPresent,
     #[error("Base fee per gas is incorrect")]
     BaseFeePerGasIncorrect,
     #[error("Timestamp is not greater than parent timestamp")]
@@ -689,6 +691,14 @@ pub fn validate_block_header(
 ) -> Result<(), InvalidBlockHeaderError> {
     if header.gas_used > header.gas_limit {
         return Err(InvalidBlockHeaderError::GasUsedGreaterThanGasLimit);
+    }
+
+    // `base_fee_per_gas` is RLP-optional (pre-London headers lack it) but
+    // mandatory for every fork ethrex validates (Merge+). Enforce presence
+    // explicitly instead of silently defaulting to INITIAL_BASE_FEE below —
+    // every other fork-gated header field already has a presence check.
+    if header.base_fee_per_gas.is_none() {
+        return Err(InvalidBlockHeaderError::BaseFeePerGasNotPresent);
     }
 
     let expected_base_fee_per_gas = if let Some(base_fee) = calculate_base_fee_per_gas(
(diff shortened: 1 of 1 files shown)


crates/common/types/block.rs
```diff
@@ -691,6 +693,14 @@ pub fn validate_block_header(
         return Err(InvalidBlockHeaderError::GasUsedGreaterThanGasLimit);
     }
 
+    // `base_fee_per_gas` is RLP-optional (pre-London headers lack it) but
+    // mandatory for every fork ethrex validates (Merge+). Enforce presence
+    // explicitly instead of silently defaulting to INITIAL_BASE_FEE below —
+    // every other fork-gated header field already has a presence check.
+    if header.base_fee_per_gas.is_none() {
```
ElFantasma, line 700: Checked the pre-London concern from the automated reviews (Codex flags it High, Claude Code Review Critical) — it doesn't hold, and gating this on `is_london_activated` would cost `ChainConfig` plumbing for a branch that can't be taken:

- `validate_block_header` already rejects every pre-merge header further down via `!header.difficulty.is_zero()`. That check is on `main`, not new here, so pre-London headers were never getting past this function anyway.
- `ChainConfig::get_fork()` bottoms out at `Fork::Paris`, so fork resolution can't name London or earlier.
- `is_london_activated` has exactly two occurrences in the tree: its own definition, and `genesis.rs:745` deciding whether the *genesis* header carries a base fee. Zero uses in validation.

The one observable change for a pre-merge header is the error label — `BaseFeePerGasNotPresent` rather than `DifficultyNotZero`, since this check sits earlier in the function.

crates/common/types/block.rs
```diff
@@ -691,6 +693,14 @@ pub fn validate_block_header(
...
+        return Err(InvalidBlockHeaderError::BaseFeePerGasNotPresent);
+    }
+
     let expected_base_fee_per_gas = if let Some(base_fee) = calculate_base_fee_per_gas(
```
ElFantasma, line 704: nit: with the presence check above, the two `unwrap_or(INITIAL_BASE_FEE)` fallbacks around here now look alike but mean different things.

- L708 `parent_header.base_fee_per_gas.unwrap_or(…)` is live — `genesis.rs:743-747` only gives the genesis header a base fee when London is active at block 0, so a genesis parent can legitimately be `None`. This one should stay.
- L716 `header.base_fee_per_gas.unwrap_or(…)` is now unreachable.

Binding it once makes the invariant structural instead of commented:

```rust
let base_fee_per_gas = header
    .base_fee_per_gas
    .ok_or(InvalidBlockHeaderError::BaseFeePerGasNotPresent)?;
```

Then compare against `base_fee_per_gas` at L716 and the dead fallback goes away with it.

PR #7056, opened 2026-07-29
fix(l1): use abs_diff for engine JWT iat window check

ilitteri:
## Summary

`invalid_issued_at_claim` computed `(now as i64 - iat as i64).abs()`, which **overflows for adversarial `iat` values above 2^63 + now** — a panic in debug builds, a wrap in release.

Reachability: the JWT HMAC signature is verified before the claims arithmetic runs, so this is only triggerable by the trusted consensus client — not a remote vector. But a misbehaving or misconfigured CL should get a clean `InvalidIssuedAtClaim` rejection, not a panic path.

## Change

`crates/networking/rpc/authentication.rs`: `u64::try_from(iat).unwrap_or(u64::MAX)` + `now.abs_diff(iat) > 60` — panic-free for the full `usize` range, same ±60 s semantics as the engine API spec.

## Tests

Added `test_iat_extreme_values_rejected_without_panic` (`usize::MAX`, `2^63 + 1000`). Full auth suite: 6/6 green.

### DIFF
--- a/crates/networking/rpc/authentication.rs
+++ b/crates/networking/rpc/authentication.rs
@@ -62,7 +62,11 @@ fn invalid_issued_at_claim(token_data: TokenData<Claims>) -> Result<bool, Authen
         .duration_since(UNIX_EPOCH)
         .map_err(|_| AuthenticationError::InvalidIssuedAtClaim)?
         .as_secs();
-    Ok((now as i64 - token_data.claims.iat as i64).abs() > 60)
+    // `abs_diff` is panic-free for any `iat` (including adversarial values near
+    // usize::MAX): the previous `now as i64 - iat as i64` subtraction could
+    // overflow for `iat > 2^63 + now`, panicking in debug builds.
+    let iat = u64::try_from(token_data.claims.iat).unwrap_or(u64::MAX);
+    Ok(now.abs_diff(iat) > 60)
 }
 
 #[cfg(test)]
@@ -205,5 +209,32 @@ mod tests {
         .unwrap();
         let res_future = validate_jwt_authentication(&token_future, &secret);
         assert!(res_future.is_ok());
+    }
+
+    #[test]
+    fn test_iat_extreme_values_rejected_without_panic() {
+        // Regression test: `iat` values above 2^63 + now overflowed the old
+        // i64 subtraction (debug panic / release wrap). They must now be
+        // rejected cleanly as InvalidIssuedAtClaim.
+        let secret = Bytes::from("my_secret_key");
+        for iat in [usize::MAX, (1usize << 63) + 1000] {
+            let claims = Claims {
+                iat,
+                id: None,
+                clv: None,
+            };
+            let token = encode(
+                &Header::default(),
+                &claims,
+                &EncodingKey::from_secret(&secret),
+            )
+            .unwrap();
+            let res = validate_jwt_authentication(&token, &secret);
+            assert_eq!(
+                res.unwrap_err(),
+                AuthenticationError::InvalidIssuedAtClaim,
+                "iat {iat} must be rejected"
+            );
+        }
     }
 }


crates/networking/rpc/authentication.rs
```diff
@@ -62,7 +62,11 @@ fn invalid_issued_at_claim(token_data: TokenData<Claims>) -> Result<bool, Authen
         .duration_since(UNIX_EPOCH)
         .map_err(|_| AuthenticationError::InvalidIssuedAtClaim)?
         .as_secs();
-    Ok((now as i64 - token_data.claims.iat as i64).abs() > 60)
+    // `abs_diff` is panic-free for any `iat` (including adversarial values near
+    // usize::MAX): the previous `now as i64 - iat as i64` subtraction could
+    // overflow for `iat > 2^63 + now`, panicking in debug builds.
```
ElFantasma, line 67: The stated condition is inverted. Overflow happens for `iat` in **`[2^63, 2^63 + now]`**, not for `iat > 2^63 + now`.

The old expression is `now as i64 - (iat as i64)`. Once `iat >= 2^63`, `iat as i64` reinterprets as negative, so the subtraction becomes `now + (2^64 - iat)`. That exceeds `i64::MAX` only while `2^64 - iat > i64::MAX - now`, i.e. `iat < 2^63 + now + 1`. Above that band, `iat as i64` is a *small* negative number and the result is comfortably in range.

Concretely, with `now ≈ 1.785e9`:

| `iat` | `iat as i64` | `now as i64 - iat as i64` | overflow? |
|---|---|---|---|
| `2^63` | `i64::MIN` | `now + 9223372036854775808` | yes |
| `2^63 + 1000` | `-9223372036854774808` | `≈ 9.2233720386e18` > `i64::MAX` | yes |
| `usize::MAX` | `-1` | `now + 1` | no |

Suggested wording:

```suggestion
    // overflow for `iat` in `[2^63, 2^63 + now]`, panicking in debug builds.
```

crates/networking/rpc/authentication.rs
```diff
@@ -206,4 +210,31 @@ mod tests {
...
         assert!(res_future.is_ok());
     }
+
+    #[test]
+    fn test_iat_extreme_values_rejected_without_panic() {
+        // Regression test: `iat` values above 2^63 + now overflowed the old
+        // i64 subtraction (debug panic / release wrap). They must now be
+        // rejected cleanly as InvalidIssuedAtClaim.
+        let secret = Bytes::from("my_secret_key");
+        for iat in [usize::MAX, (1usize << 63) + 1000] {
```
ElFantasma, line 220: Following from the comment above: `usize::MAX` is not a regression input. It reinterprets as `-1`, so the old code computed `now + 1` — no overflow, no panic, and it was already rejected. Only `(1usize << 63) + 1000` sits inside the overflow band and actually reproduces the debug panic this PR fixes.

Both values are correctly rejected either way, so the test passes; the issue is that it would still pass against the *unfixed* code for half its inputs, and the loop reads as though both are boundary cases.

Worth bracketing the real boundary instead — `1usize << 63` is the strongest case (`i64::MIN`, overflows for any `now >= 1`), and keeping `usize::MAX` as an explicitly-labelled non-overflow control:

```suggestion
        // `1<<63` and `1<<63 + 1000` sit in the overflow band and panicked in
        // debug before the fix; `usize::MAX` reinterprets as -1 and never did,
        // kept here as a control that huge values are still rejected.
        for iat in [1usize << 63, (1usize << 63) + 1000, usize::MAX] {
```

crates/networking/rpc/authentication.rs
```diff
@@ -62,7 +62,11 @@ fn invalid_issued_at_claim(token_data: TokenData<Claims>) -> Result<bool, Authen
         .duration_since(UNIX_EPOCH)
         .map_err(|_| AuthenticationError::InvalidIssuedAtClaim)?
         .as_secs();
-    Ok((now as i64 - token_data.claims.iat as i64).abs() > 60)
+    // `abs_diff` is panic-free for any `iat` (including adversarial values near
+    // usize::MAX): the previous `now as i64 - iat as i64` subtraction could
```
iovoid, line 66: The values need to be positive, otherwise the 'as i64' will just turn them into negative numbers. `usize::MAX` would be equivalent to -1.

PR #7057, opened 2026-07-29
fix(l1): reject short ECIES handshake messages without underflow

ilitteri:
## Summary

`decrypt_message` (RLPx ECIES handshake) computed `rest.len() - 32` after splitting off the 65-byte public key and 16-byte IV. For 81–112-byte handshake messages — fewer than 32 bytes of ciphertext+MAC — the subtraction **underflows**: a panic in debug builds; in release it wraps to a huge `usize` and `split_at_checked` returns `None`, producing the correct `InvalidMessageLength` error only by accident.

## Change

Explicit `rest.len() < 32` bound before the split, so the malformed message is rejected with `InvalidMessageLength` in all build profiles.

## Impact

None on production nodes (release builds already degrade gracefully); removes a debug-profile panic and makes the bound explicit instead of incidental.

## Tests

Added `decrypt_message_rejects_short_mac_section_without_panic` (valid pubkey + IV + 5 bytes of trailing data → `InvalidMessageLength`, no panic).


crates/networking/p2p/rlpx/connection/handshake.rs
```diff
@@ -377,6 +377,12 @@ fn decrypt_message(
...
         .split_at_checked(16)
         .ok_or_else(|| PeerConnectionError::InvalidMessageLength)?;
+    // Explicit bound: after the public key and IV, the message must still hold
+    // at least the 32-byte MAC. `rest.len() - 32` underflows for 81..112-byte
+    // messages (debug panic / release wrap) without this check.
+    if rest.len() < 32 {
+        return Err(PeerConnectionError::InvalidMessageLength);
+    }
     let (c, d) = rest
         .split_at_checked(rest.len() - 32)
```
ElFantasma, line 387: nit: with the new guard above, this `split_at_checked` can no longer return `None` — `rest.len() >= 32` implies `rest.len() - 32 <= rest.len()` — so the `ok_or_else` on the next line is now a dead error path. Harmless as belt-and-braces, but `checked_sub` expresses the same invariant in one expression and keeps a single fallible path:

```rust
let mac_start = rest
    .len()
    .checked_sub(32)
    .ok_or(PeerConnectionError::InvalidMessageLength)?;
let (c, d) = rest.split_at(mac_start);
```

That drops the separate `if` and makes the no-underflow intent the reason the error exists, rather than having the bound stated in one place and enforced (unreachably) in another. Either shape is fine — flagging it only because the redundancy is easy to misread later as "this can still fail".

ElFantasma (approved):
Just one non-blocking nit inline.

PR #7058, opened 2026-07-29
fix(l1): persist frame receipts with the storage codec, and make Amsterdam-scheduling devnets boot

edg-l:
## What

Three defects that together made a fresh Amsterdam-scheduled devnet unusable, found by standing one up rather than by reading the code.

### 1. Frame receipts were unreadable once flushed to disk

`flush_block_data` wrote receipts to `RECEIPTS_V2` with `encode_to_vec` (the consensus codec) while every reader uses `decode_storage`. For an EIP-8141 frame receipt the layouts differ — the storage codec carries `succeeded` and the aggregated top-level logs that the consensus layout deliberately omits — so reading a row back put `cumulative_gas_used` where `succeeded` was expected:

```
eth_getTransactionReceipt -> -32603
  Internal Error: Error decoding field 'succeeded' of type bool: MalformedBoolean
```

Every frame transaction's receipt was affected. Non-frame receipts were not, because `encode_storage` falls through to the same encoding for them, and reads are served from the in-memory block buffer until the block flushes — which is what kept this latent.

Observed on a live local devnet: the transaction mined correctly in block `0x59` and `eth_getTransactionByHash` served it fine, but its receipt could never be fetched.

The regression test drives the store's real buffer-and-flush path. A codec-only round-trip would pass either way, since both codecs are individually correct and only the call site was wrong.

### 2. The Amsterdam-scheduling devnet config could not boot

`eip8141-devnet.yaml` scheduled Gloas at genesis, which makes the beacon genesis state a Gloas state. Lighthouse rejects the one the genesis generator writes and never starts:

```
CRIT Failed to start beacon node
     reason: "Built-in genesis state SSZ bytes are invalid: OffsetsAreDecreasing(0)"
```

Gloas moves to epoch 1 and Heze to epoch 2, which keeps the genesis state Fulu and matches what every `bal-devnet-*` config already does. It also gives the chain an `amsterdamTime`, which Hegotá needs since it is defined on top of Amsterdam.

That in turn forces the CL pin. Once Amsterdam is active ethrex rejects `forkchoiceUpdatedV3` payload attributes, and a **stable** Lighthouse does not drive V4 for a Gloas payload even though its binary carries the method string — block production stops dead at the boundary:

```
-38005 "Unsupported fork: forkChoiceV3 used to build Amsterdam payload"
ERROR Proposer prepare routine failed ... slot: 88
```

So the CL moves to `ethpandaops/lighthouse:glamsterdam-devnet-7`, the pairing the Hegotá devnet config already uses. Gloas ≥ epoch 1 **and** a Glamsterdam-capable CL are both required; either alone leaves the config broken.

### 3. The EIP-8282 predeploys were missing from generated genesis

EIP-8282 (Builder Execution Requests, part of Glamsterdam via EIP-7773) specifies that if either predeploy has no code once the EIP is active, **every block from activation onward is invalid** — the same rule EIP-7002/EIP-7251 apply. `ethereum-genesis-generator` only deploys them from **6.1.4**; the pinned revision shipped 6.0.7.

Symptom: `getPayload` failed on every slot from the Amsterdam boundary on, the CL reported `Unknown payload`, and the chain froze on the last pre-Amsterdam block with all ELs stuck at the same height.

```
-32603 Internal Error: EVM error: System call failed:
       System contract: 0x0000…8282 has no code after deployment
```

The pin moves to a revision carrying 6.1.4. Verified locally: the predeploys now come from the generator with no per-config preloading.

### Also

`ChainConfig::unscheduled_predecessor_forks` + a `display_config` warning for a gapped fork schedule. Fork rules resolve from the fork ordinal, so scheduling a fork silently activates the rules of every fork below it while each check keyed on an individual fork's own timestamp stays inactive. That mismatch is why both Hegotá devnets ran Amsterdam EVM rules under a pre-Amsterdam header schema for a month, and it is the class #7053 fixes one instance of. Now it is visible in the first screen of node output.

Plus the stale `3 = skipped` comment on `RpcFrameReceipt::status`, corrected to `2`.

## Testing

- `cargo test -p ethrex-test` — 911 passed, 0 failed
- `cargo clippy --workspace --all-targets` clean, `cargo fmt` clean
- Each new test verified to fail against the unfixed code first. The receipt tests fail with the exact production error; the non-frame test passes throughout, confirming the bug is frame-only.
- End to end on a local kurtosis devnet, for both the Hegotá config and the `eip8141` config changed here: the chain crosses Osaka → Amsterdam (→ Hegotá), `slotNumber` and `blockAccessListHash` appear exactly at the Amsterdam boundary and not before, the EL logs confirm the CL drives `forkchoiceUpdatedV4`, `requestsHash` stays the empty-keccak with empty 8282 queues, all three ELs agree on every block hash, and a real frame transaction mines with per-frame receipts.

## Upstream follow-up (not fixed here)

Generator 6.1.4 arms the `EXCESS_INHIBITOR` in slot 0 for the builder **exit** predeploy but omits it for the **deposit** one (`apps/el-gen/system-contracts.yaml` gives `eip8282_exit` a `storage` block and `eip8282_deposit` none). Per EIP-8282 both should be armed at deployment so no request can be enqueued before the first system call. Harmless on a devnet with no builder deposits, but worth a one-line PR upstream.


crates/storage/store.rs
```diff
@@ -4152,10 +4174,14 @@ fn flush_block_data(
...
         for (index, receipt) in b.receipts.iter().enumerate() {
+            // `RECEIPTS_V2` holds the internal storage codec, which for frame
+            // receipts carries `succeeded` and the aggregated logs that the
+            // consensus layout omits. Identical to `encode_to_vec` for non-frame
+            // receipts.
             tx.put(
                 RECEIPTS_V2,
                 &receipt_key(&hash, index as u64),
-                &receipt.encode_to_vec(),
+                &receipt.encode_storage(),
```
ElFantasma, line 4532: The fix is right — `decode_storage` is what every reader uses, so `encode_storage` is the correct pairing. But this changes the on-disk representation of a table that already has rows, and I don't see anything covering the ones written before the fix.

Rows flushed by the old code are frame receipts encoded with the consensus codec. After upgrading they're still in that codec, so they still fail exactly as the PR describes:

```
eth_getTransactionReceipt -> -32603
  Internal Error: Error decoding field 'succeeded' of type bool: MalformedBoolean
```

The failure mode that worries me isn't the broken row itself — it's that the symptom is *identical to the bug being fixed here*. An operator who upgrades a running devnet, queries a frame tx from before the upgrade, and sees the same `-32603` will reasonably conclude the fix didn't work.

The repo already has the machinery: `STORE_SCHEMA_VERSION` is `3` in `crates/storage/lib.rs:91`, and `RECEIPTS_V2` has migration history — `crates/storage/migrations.rs:176,193` is the `RECEIPTS` → `RECEIPTS_V2` move. Note `migrations.rs:31` asserts `MIGRATIONS.len() == (STORE_SCHEMA_VERSION - 1)`, so a bump needs a matching entry.

I don't think a migration is obviously required — EIP-8141 is devnet-only today, so the affected population may be zero chains anyone intends to keep. Two ways to close it:

1. Bump to `4` with a migration that re-encodes frame receipts in `RECEIPTS_V2`, or
2. State explicitly in the PR body (and ideally the commit message) that pre-fix frame receipts stay unreadable and affected devnets must be re-genesised — so the next person hitting that `-32603` knows it's stale data rather than a regression.

The PR template's `STORE_SCHEMA_VERSION` checklist item applies here either way, since this does touch the Store.

Not flagging the non-frame path: your claim that `encode_storage` falls through to the same encoding there matches what the readers expect, and the test that stays green throughout backs it up.

ilitteri (approved):

PR #7059, opened 2026-07-29
test(l1): intrinsic gas accounting coverage for EIP-8141 frame transactions

AnkushinDaniil:
## Why

Salvaged from #7044. The implementation half was superseded by #7004, but as @edg-l noted there: "main has no intrinsic-gas-accounting tests, so the intrinsic_gas_accounting_tests module here is still useful coverage over code main already has. A tests-only PR against current main would be welcome."

## Change

Tests only, no production code. Adds the `intrinsic_gas_accounting_tests` module to `test/tests/levm/eip8141_tests.rs`:

- `interop_reference_tx_intrinsic_gas_charges_payload_bytes_only`: decodes the 162-byte reference type-0x06 tx from the cross-client replay and pins its intrinsic gas to the payload-byte accounting (`FRAME_TX_INTRINSIC_COST` + 2 * `FRAME_TX_PER_FRAME_COST` + secp256k1 verification + calldata gas over the signature bytes), plus the resulting `total_gas_limit()`.
- `intrinsic_gas_is_insensitive_to_structural_frame_fields`: changing frame `gas_limit`/`value` must not change intrinsic gas - only payload bytes are charged.

## Evidence

```
cargo test -p ethrex-test --test ethrex_tests intrinsic_gas_accounting_tests
# 2 passed; 0 failed
# 86 passed; 0 failed
cargo fmt --check   # pass
```

## Notes

Targets main per the request above. These tests assert the #7004 accounting, so they only pass on branches that include it (frames-devnet-0 currently does not).


edg-l (approved):

test/tests/levm/eip8141_tests.rs
```diff
@@ -3604,3 +3604,79 @@ fn storage_refund_from_a_later_frame_reduces_reported_gas() {
...
+        assert_eq!(tx.frames.len() as u64, FRAME_COUNT);
+        assert_eq!(intrinsic_gas(&tx), expected_intrinsic_gas);
+        assert_eq!(
+            tx.total_gas_limit(),
+            expected_intrinsic_gas + FRAME_GAS_LIMITS
+        );
+    }
+
+    #[test]
+    fn intrinsic_gas_is_insensitive_to_structural_frame_fields() {
```
ElFantasma: This test cannot fail, for any values you put in those fields.

`total_gas_limit()` is derived, not stored (`crates/common/types/transaction.rs:2147`):

```rust
mandatory_gas() + data_cost() + Σ frames.map(|f| f.gas_limit)
```

and the helper computes `total_gas_limit() - Σ frames.map(|f| f.gas_limit)`. The frame gas limits cancel:

```
intrinsic_gas(tx) ≡ mandatory_gas() + data_cost()
```

So `gas_limit` is *algebraically* absent from the result. Setting `frames[0] = 1` and `frames[1] = 4_000_000_000` changes `total_gas_limit` by exactly the amount the helper then subtracts back off. The assertion would hold even if intrinsic gas genuinely did depend on frame gas limits — which is the thing the test is named for.

The `value` half has the same problem unless `mandatory_gas()` or `data_cost()` reads `value`, which as far as I can see neither does.

Two ways to give it teeth, depending on which property you actually want:

**(a) Pin the decomposition directly** — no helper, no cancellation:

```rust
let baseline = interop_reference_tx();
let mut restructured = interop_reference_tx();
restructured.frames[0].gas_limit = 1;
restructured.frames[1].value = U256::from(1u64);

assert_eq!(restructured.data_cost(), baseline.data_cost());
assert_eq!(restructured.mandatory_gas(), baseline.mandatory_gas());
```

**(b) Assert the delta relationship**, which *is* a real invariant and does constrain `total_gas_limit`:

```rust
let before = baseline.total_gas_limit();
restructured.frames[0].gas_limit = baseline.frames[0].gas_limit + 1_000;
assert_eq!(restructured.total_gas_limit(), before + 1_000);
```

No objection to the first test — asserting the concrete expected value against `FRAME_TX_INTRINSIC_COST`, the per-frame cost, the secp256k1 verify cost and the payload byte gas genuinely pins `mandatory_gas() + data_cost()`, and it would catch a regression in any of them.

test/tests/levm/eip8141_tests.rs
```diff
@@ -3604,3 +3604,79 @@ fn storage_refund_from_a_later_frame_reduces_reported_gas() {
...
+        }
+    }
+
+    fn intrinsic_gas(tx: &FrameTransaction) -> u64 {
+        let frame_gas = tx
+            .frames
+            .iter()
+            .map(|f| f.gas_limit)
+            .fold(0u64, |acc, g| acc.saturating_add(g));
+        tx.total_gas_limit().saturating_sub(frame_gas)
```
ElFantasma: nit, and related to the comment above: `saturating_sub` here can't ever clamp, because the value being subtracted is one of the terms `total_gas_limit()` is built from — the difference is `mandatory_gas() + data_cost()`, always ≥ 0.

That's harmless today, but it means if the relationship between the two ever changes the helper silently returns `0` instead of failing. A plain `-` would panic in debug and surface it. Minor either way — if the tests move to asserting `mandatory_gas()`/`data_cost()` directly, the helper disappears.

test/tests/levm/eip8141_tests.rs
AnkushinDaniil: You are right, and it was worse than untested: the frame gas limits cancel algebraically, so the assertion held by construction. Rewritten along the lines of (a): the two tests now assert `mandatory_gas()` and `data_cost()` directly, and the frame gas limit is pinned by the delta relationship instead (`+1_000` in, `+1_000` out). The helper is gone with it, which also removes the `saturating_sub` from the note below.

test/tests/levm/eip8141_tests.rs
AnkushinDaniil: Gone with the helper.

test/tests/levm/eip8141_tests.rs
```diff
@@ -3604,3 +3604,79 @@ fn storage_refund_from_a_later_frame_reduces_reported_gas() {
...
+            + SECP256K1_VERIFY_GAS
+            + payload_calldata_gas;
+
+        let tx = interop_reference_tx();
+        assert_eq!(tx.frames.len() as u64, FRAME_COUNT);
+        assert_eq!(intrinsic_gas(&tx), expected_intrinsic_gas);
+        assert_eq!(
+            tx.total_gas_limit(),
+            expected_intrinsic_gas + FRAME_GAS_LIMITS
+        );
```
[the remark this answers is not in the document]
AnkushinDaniil: Done: the fixture properties that make one signature the whole payload are now asserted (two frames with empty data, a single secp256k1 signature with no signer and no msg), and the zero/non-zero byte counts are derived from `tx.signatures[0].signature` rather than hard-coded.

test/tests/levm/eip8141_tests.rs
```diff
@@ -3604,3 +3604,79 @@ fn storage_refund_from_a_later_frame_reduces_reported_gas() {
...
+    }
+
+    #[test]
+    fn intrinsic_gas_is_insensitive_to_structural_frame_fields() {
+        let baseline = interop_reference_tx();
+
+        let mut restructured = interop_reference_tx();
+        restructured.frames[0].gas_limit = 1;
+        restructured.frames[1].gas_limit = 4_000_000_000;
```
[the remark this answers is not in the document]
AnkushinDaniil: Added an explicit length assertion before the indexing.

ElFantasma (approved):
LGTM. Thanks!

ilitteri (approved):

ElFantasma:
Heads up on why this is showing as blocked — it's a signature thing, and it traces back to me rather than anything you did.

When I hit "Update branch" on Jul 31, GitHub created a merge commit (`f6c611d`) authored by me and signed with GitHub's own web-flow key, so it verified fine at the time. Your force-push on Aug 4 rewrote the branch, and since you have SSH commit signing enabled, git re-signed that commit with your key on the way past. GitHub now sees a commit authored by me but signed with a key registered to your account, so it can't attribute the signature and marks it "unknown key".

That's just what a rebase does when it passes over someone else's commit — but `main` has a required-signatures rule that checks every commit in the PR, so this one commit is what's blocking the merge button.

Simplest fix is to drop the merge commits entirely with a rebase. They aren't buying us anything anyway, since the repo squash-merges:

```bash
git fetch upstream main        # or whatever you call the lambdaclass remote
git rebase upstream/main
git push --force-with-lease
```

That leaves just your two commits — both already verified — and picks up current main at the same time.

Sorry for the detour.

PR #7061, opened 2026-07-29
fix(l1): allow EIP-8141 VERIFY frame to target an approved non-sender account

AnkushinDaniil:
**Motivation**

ethrex rejects a sponsored EIP-8141 frame transaction whose `pay` frame targets an account other than `tx.sender`, with `frame N: VERIFY frame target is not tx.sender`. This blocks the canonical `[only_verify, pay]` sponsorship prefix, which exists precisely so that an account other than the sender can pay.

This was reproduced cross-client in a Nethermind + ethrex Kurtosis run: for a sponsored transaction (`only_verify` + `pay` targeting a sponsor EOA + a SENDER transfer, sponsor signature at index 1), Nethermind accepted and executed it - all frames succeeded and the sponsor was charged as payer - while ethrex rejected it at RPC admission with the error above.

**Description**

`FrameTransaction::validate_prefix_structure` applied the target-must-be-`tx.sender` check to every VERIFY frame of the validation prefix. The current EIP-8141 structural rules restrict the target this way only for `self_verify` and `only_verify` frames (rule 3: they "must execute in `VERIFY` mode, target `tx.sender` (either explicitly or via a null target)"). Rule 4 places no target requirement on the `pay` frame: it "must execute in `VERIFY` mode, have flags set to `APPROVE_PAYMENT`, and must successfully call `APPROVE(APPROVE_PAYMENT)`". The `APPROVE(APPROVE_PAYMENT)` semantics set `payer = resolved_target` with no sender-equality requirement, so a non-sender sponsor targeted by the `pay` frame approves payment when the frame executes.

Changes:

- `crates/common/types/transaction.rs`: in `validate_prefix_structure`, exempt the `pay` frame of the `OnlyVerifyPay` / `DeployOnlyVerifyPay` shapes from the `VerifyTargetNotSender` check. `self_verify` and `only_verify` frames (including the `pay_index` frame of the `SelfVerify` shapes, which is the `self_verify` frame) keep it. No other structural check is touched.
- `test/tests/levm/eip8141_tests.rs`: regression test `sponsored_pay_frame_may_target_non_sender` - an `OnlyVerifyPay` prefix whose `pay` frame targets a non-sender sponsor must pass `validation_prefix` + `validate_prefix_structure` and simulate to a passing outcome with the sponsor identified as the paymaster. The test fails on `main` with `VerifyTargetNotSender { frame_index: 1 }` and passes with this change.

Verification (all local):

- `cargo check -p ethrex-levm` - clean
- `cargo test -p ethrex-test --test ethrex_tests levm::eip8141_tests` - 85 passed, 0 failed
- `cargo test -p ethrex-test --test ethrex_tests common::frame_tx_validation_tests` - 25 passed, 0 failed
- `cargo fmt --all -- --check` and `git diff --check` - clean

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2494,8 +2494,10 @@ impl FrameTransaction {
     /// - Deploy frame (if any) is at index 0 and uses DEFAULT execution mode.
     /// - At most one deploy frame exists in the prefix.
     /// - self_verify / only_verify / pay frames use VERIFY execution mode.
-    /// - Resolved target of each VERIFY frame matches `tx.sender` (target == None
-    ///   means sender; a non-None target must equal sender).
+    /// - Resolved target of each self_verify / only_verify frame matches
+    ///   `tx.sender` (target == None means sender). The pay frame of the
+    ///   OnlyVerifyPay / DeployOnlyVerifyPay shapes has no target restriction
+    ///   (EIP-8141 structural rule 4): it may target a non-sender sponsor.
     /// - Scope restriction matches the frame's role:
     ///   self_verify → `APPROVE_EXECUTION_AND_PAYMENT`, only_verify → `APPROVE_EXECUTION`,
     ///   pay → `APPROVE_PAYMENT`.
@@ -2542,10 +2544,19 @@ impl FrameTransaction {
                         });
                     }
 
-                    // Resolved target must be tx.sender (None means sender).
+                    // EIP-8141 structural rule 3 restricts the target to
+                    // tx.sender (None means sender) only for self_verify /
+                    // only_verify frames. Rule 4 places no target requirement
+                    // on the pay frame: it may target a non-sender sponsor,
+                    // which approves payment via APPROVE(APPROVE_PAYMENT)
+                    // when the frame executes.
+                    let is_pay_frame = matches!(
+                        prefix.shape,
+                        PrefixShape::OnlyVerifyPay | PrefixShape::DeployOnlyVerifyPay
+                    ) && prefix.pay_index == Some(idx);
                     let target_ok = match frame.target {
                         None => true,
-                        Some(addr) => addr == self.sender,
+                        Some(addr) => addr == self.sender || is_pay_frame,
                     };
                     if !target_ok {
                         return Err(FrameValidationError::VerifyTargetNotSender {
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2491,6 +2491,11 @@ impl FrameTransaction {
                     // on the pay frame: it may target a non-sender sponsor,
                     // which approves payment via APPROVE(APPROVE_PAYMENT)
                     // when the frame executes.
+                    // The shape guard is load-bearing: every shape populates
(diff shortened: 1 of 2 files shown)


crates/common/types/transaction.rs
```diff
@@ -2483,10 +2485,19 @@ impl FrameTransaction {
...
                     }
 
-                    // Resolved target must be tx.sender (None means sender).
+                    // EIP-8141 structural rule 3 restricts the target to
+                    // tx.sender (None means sender) only for self_verify /
+                    // only_verify frames. Rule 4 places no target requirement
+                    // on the pay frame: it may target a non-sender sponsor,
+                    // which approves payment via APPROVE(APPROVE_PAYMENT)
+                    // when the frame executes.
+                    let is_pay_frame = matches!(
```
ElFantasma, line 2558: Confirmed correct against `docs/eip-8141.md` rather than just the cross-client result — the spec attaches the restriction to the **scope**, not the frame's position:

| scope | doc | target requirement |
|---|---|---|
| `0x1 APPROVE_PAYMENT` | L196 | none |
| `0x2 APPROVE_EXECUTION` | L197 | `resolved_target == tx.sender` |
| `0x3 APPROVE_EXECUTION_AND_PAYMENT` | L198 | `resolved_target == tx.sender` |

and L217 restates it: *"Scopes 0x2 and 0x3 require `resolved_target == tx.sender`"*. The pay frame carries `0x1`, so exempting it is right.

**The `matches!` guard is load-bearing — please don't let it get simplified away later.** All four shapes populate `pay_index`, including the self-verify ones:

```
DeployOnlyVerifyPay → non_expiry[2]      OnlyVerifyPay → non_expiry[1]
DeploySelfVerify    → non_expiry[1]      SelfVerify    → non_expiry[0]
```

For `SelfVerify`/`DeploySelfVerify`, `pay_index` points at the single self-verify frame, which carries scope `0x3` and *does* require `target == tx.sender`. Reducing this to `prefix.pay_index == Some(idx)` alone would exempt it and let a self-verify frame target a non-sender. A one-line comment saying so would protect it.

**Optional:** since the rule is a property of the scope, keying off it directly is closer to the spec and locally self-evident:

```rust
let is_pay_frame = frame.scope_restriction() == APPROVE_PAYMENT;
```

Equivalent today only because the scope check below (L2509-2521) enforces the role↔scope mapping — but that runs *after* this one, so as written the target check's correctness depends on a downstream validation. Fine either way; the current form just isn't self-contained.

ElFantasma (approved):

edg-l (approved):

crates/common/types/transaction.rs
AnkushinDaniil: Added a comment on the guard saying exactly that: every shape populates `pay_index`, and for `SelfVerify` / `DeploySelfVerify` it points at the self-verify frame, which carries `APPROVE_EXECUTION_AND_PAYMENT` and does require the sender as its target, so matching on `pay_index` alone would exempt it. I left the shape match rather than keying off the scope, since the scope check runs after this one and I would rather not make the target rule depend on a later validation.

ElFantasma (approved):

PR #7064, opened 2026-07-29
ci(l1,l2): catch stale blob fixtures at PR time, not on main

ilitteri:
## Problem

#7033 changed the block body schema/validation in `crates/common` and `crates/blockchain`, which silently invalidated `fixtures/blobs/`. The L2 workflow's `detect-changes` filter only watches `crates/l2/**`, `test/tests/l2/**` and `fixtures/**` — so every L2 check, including the fast **Validate Blob Fixtures** job, was skipped on the PR (visible as `Integration Test L2 — skipping`). The State Reconstruction Tests job only surfaced the staleness [on main after merge](https://github.com/lambdaclass/ethrex/actions/runs/30471087953/job/90652696564).

## Guards (defense in depth)

1. **`crates/common/tests/blob_fixtures.rs`** — decodes every fixture blob, runs `validate_block_body` on each block, and checks genesis linkage. It lives in the workspace test suite (`pr-main_l1.yaml`), which runs on every non-docs PR **regardless of path filters**. Verified: fails on the stale fixtures currently on main ("`block 1 … fails validate_block_body: Withdrawals root does not match`"), passes on the migrated fixtures from #7063.
2. **Filter broadening** — `pr-main_l2.yaml` `detect-changes` now also triggers on `crates/common/types/**` and `crates/blockchain/**`, the crates that own block schema, encoding, validation and production.
3. **Docs/Makefile** — `regenerate-blobs.md`, `state-reconstruction-blobs.md` and `crates/l2/Makefile` now list block schema/validation changes (not just genesis changes) as staleness triggers, and note that encoding-only changes can be migrated in place without the full Docker regeneration flow.

## Notes

- Stacked on **#7063** (contains the migrated fixtures the new test expects; merge that first).
- The new test needs no Docker, no store, no node — ~0.3 s.


edg-l (approved):

crates/common/tests/blob_fixtures.rs
```diff
@@ -0,0 +1,98 @@
...
+    let root = workspace_root();
+
+    // Genesis linkage: the first block of the first blob must point at the
+    // L2 genesis block hash (catches fixtures left stale by a genesis change).
+    let genesis_file = fs::File::open(root.join("fixtures/genesis/l2.json"))?;
+    let genesis: Genesis = serde_json::from_reader(genesis_file)?;
+    let genesis_hash = genesis.get_block().hash();
+
+    let blobs_dir = root.join("fixtures/blobs");
+    let mut entries: Vec<_> = fs::read_dir(&blobs_dir)?
```
ElFantasma, line 40: `file_idx` indexes **every** directory entry, but non-`.blob` files are `continue`d a few lines below — so if entry 0 ever isn't a blob, the genesis-linkage assertion at `:68` silently never runs, and the test keeps passing.

Not live today: `fixtures/blobs/` currently holds exactly `1-1.blob` … `6-1.blob` and nothing else, so index 0 is a blob and the check does fire. It becomes live the moment anyone adds a non-blob file that sorts first — and a dotfile is the likely one, since `.` (0x2E) sorts before `1` (0x31). A `.gitattributes` marking these as binary is a very natural thing for someone to add to a directory of 128 KiB blobs.

The same slip weakens the assert just above: `!entries.is_empty()` is satisfied by *any* entry, so a directory containing only a README would pass it with zero blobs read. `total_blocks > 0` at the end does catch that case, so it is belt-and-braces rather than a hole — but filtering before you index makes both checks mean what they say:

```rust
let mut entries: Vec<_> = fs::read_dir(&blobs_dir)?
    .collect::<Result<Vec<_>, _>>()?
    .into_iter()
    .filter(|e| e.path().extension().and_then(|x| x.to_str()) == Some("blob"))
    .collect();
```

then drop the `continue` inside the loop.

Worth saying the rest of the test holds up well: `blocks_count > 0` per blob and `total_blocks > 0` at the end make it genuinely non-vacuous, which is not a given — and the error strings point straight at `regenerate-blobs.md`, which is what makes a guard like this actually usable at 3am.

One optional extra while you are in here: the test validates each body independently and checks only the *first* block's parent. It would not notice a missing or reordered blob in the middle. Asserting `block.header.parent_hash == previous.hash()` across the whole sequence would close that, and it is the same staleness class the PR is about.

PR #7067, opened 2026-07-30
fix(l1): drive the Amsterdam engine-API versions from dev mode

edg-l:
## Motivation

`--dev` cannot run an Amsterdam+ chain. The dev block producer picks `newPayload`
V4/V5 by fork, but pins `forkchoiceUpdated` to V3 and `getPayload` to V5, and both are
rejected once Amsterdam is active:

- FCU V3 with attributes → `Unsupported fork: forkChoiceV3 used to build Amsterdam payload`
- `getPayload` V5 serves Osaka only; Amsterdam+ needs V6

Every slot fails, and the producer is fatal after three consecutive failures, so the node
exits a second after startup:

```
ERROR Failed to produce block: error sending engine_forkchoiceUpdatedV3 ...
ERROR block producer failed: EngineClient System Failed after: 3; shutting down the node
```

I hit this reproducing a devnet issue locally on a Hegotá genesis.

## Description

Select the version by fork at the block timestamp, mirroring the `newPayload` switch that
was already there:

| call | pre-Amsterdam | Amsterdam+ |
|---|---|---|
| `forkchoiceUpdated` | V3 | **V4** |
| `getPayload` | V5 | **V6** |
| `newPayload` | V4 | V5 (unchanged) |

`EngineClient` gains `engine_forkchoice_updated_v4` and `engine_get_payload_v6`.

V4 attributes need two fields a CL would supply, so the producer stands in: the EIP-7843
`slotNumber` (one slot per production attempt, starting from the head's slot so it stays
monotonic across restarts) and the execution-apis#796 `targetGasLimit` (held at the head's
gas limit).

`start_block_producer` would have reached 9 positional parameters, tripping
`clippy::too_many_arguments` under `-D warnings`, so its arguments move into a
`BlockProducerConfig`.

## How to Test

```
cargo run --release --features dev -- --dev --network fixtures/genesis/l1-hegota.json --datadir memory
```

That genesis is added here, because selecting the right engine-API versions is not enough on
its own. The EIP-8282 builder deposit/exit predeploys come from genesis, like the
EIP-7002/EIP-7251 ones, and their empty code invalidates every Amsterdam+ block, so the
producer still died after three attempts:

```
ERROR Failed to produce block: System contract: 0x0000…8282 has no code after deployment
ERROR block producer failed: EngineClient System Failed after: 3; shutting down the node
```

`fixtures/genesis/l1-hegota.json` is `l1.json` plus those two predeploys, with `amsterdamTime`
and `hegotaTime` at 0. The Hegotá predeploys (`0x…8141`, `0x…8250`) need no entry: the client
installs them at the fork boundary. `docs/developers/l1/dev-mode.md` documents the flag.

With it, a chain that produced no blocks at all before mines steadily, and headers carry both
the EIP-7843 `slotNumber` and the EIP-7928 `blockAccessListHash`:

```
{ "number": "0x8", "slotNumber": "0x8", "blockAccessListHash": "0x4c61de18e8…" }
```

Verified further by driving frame transactions through the resulting node: the EIP-8141
admission rules, EIP-8250 keyed nonces and a sponsored transaction all behave, 47 blocks with
no errors.

Pre-Amsterdam is unchanged: the same run on an Osaka-only genesis still produces blocks via
V3/V5 with `slotNumber: null`.

`make lint-l1` clean; `cargo test -p ethrex-rpc --lib` 98 passing.

## Checklist

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. _(N/A — dev-mode block production and engine-API client only.)_


ilitteri (approved):

crates/blockchain/dev/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+ethrex-common.workspace = true
```
ElFantasma, line 12: This edge also has to land in `tooling/Cargo.lock`, which is what `Check Cargo.lock` is failing on:

```
error: the lock file /home/runner/work/ethrex/ethrex/tooling/Cargo.lock needs to be updated but --locked was passed to prevent this
make: *** [Makefile:269: check-cargo-lock] Error 101
```

`tooling/` is a separate workspace with its own lockfile, and `Makefile:269` runs `cargo metadata --locked --manifest-path tooling/Cargo.toml`. Its `ethrex-dev` entry still carries the pre-PR dep set:

```
name = "ethrex-dev"
dependencies = [
 "bytes",
 "envy",
 "ethereum-types",
 "ethrex-rpc",
 ...
]
```

Running `cargo metadata --manifest-path tooling/Cargo.toml` without `--locked` should regenerate it. The root `Cargo.lock` change in this PR is already correct.

ElFantasma (approved):
Engine-version dispatch is consistent across all three calls; one blocking item — `tooling/Cargo.lock` is missing this new dependency edge, which is the `Check Cargo.lock` failure.

PR #7068, opened 2026-07-30
fix(l1): reconcile an unrolled atomic batch's access list, gas and status

edg-l:
## Motivation

Two defects on the EIP-8141 atomic-batch unroll path. They are separate rules and are
called out separately below, because the second one is consensus-visible beyond the
receipts trie.

**1. The unroll leaves the block access list holding writes the block does not contain.**
An atomic batch with a reverting member halts block production. The unroll reverts the
batch's substate, state gas, approvals and logs — but not the EIP-7928 recorder. The
builder seals a block whose access list claims changes the block does not contain, and
import rejects it:

```
WARN BAL validation failed for tx 0: account 0x422a…373f storage slot 0
     mismatch at index 1: BAL=1, exec=Some(9)
```

`BAL=1` is the in-batch write; `exec=Some(9)` is the prior value the unroll correctly
restored. The builder re-selects the same tx every slot, so it rebuilds the identical
rejected block forever. Nothing is logged above `WARN`.

**2. The unroll re-prices every batch frame at its full `gas_limit`.** The spec now
states the rule: "Logs emitted by frames that executed before the failure are discarded
together with their state changes. Those frame receipts retain their execution status and
gas used, with empty logs." ethrex instead rewrote each frame as `FAILURE` charged its
whole `gas_limit`, which over-bills the payer and mislabels frames that ran.

## Description

**The access list.** Checkpoint the recorder at batch entry — where the substate backup
and approval snapshot are already taken — and restore it in the unroll, as the EIP-7906
body revert does a few lines above. `BlockAccessListRecorder::restore` is written for
this: it truncates changes made after the checkpoint, re-files a freshly-written slot as
a **read**, and leaves `touched_addresses` alone. Per EIP-7928 a reverted call's accesses
still belong in the list; only its changes do not.

**The gas and status.** Stop rewriting the batch's frame results; drop only their logs.
Frames that ran keep their status and their gas, the failing frame keeps what the
single-frame rule already gave it (actual `gas_used` for a `REVERT`, full `gas_limit` for
an exceptional halt), and the remaining members stay `SKIPPED` with their gas refunded.
`total_gas_used` therefore stands as executed.

This half is consensus-visible through more than the receipts trie: per-frame gas feeds
`total_gas_used` → the header `gasUsed` and the payer refund. `docs/eip-8141.md` §11
previously marked the choice **UNRESOLVED, pending cross-client confirmation**. That
confirmation has since arrived from the reference implementation: the
`tests-frames-devnet@v0.0.0` release covers this shape in
`atomic_batch_rollback`, `bal_atomic_batch_write` and
`bal_atomic_batch_skipped_frame_absent`, and each of the three rejects the
full-`gas_limit` behaviour on the block header `gasUsed` — the exact quantity the old
note was worried about. §11 is updated to record that, and the two halves' status is no
longer a unilateral choice.

Scope is the batch-unroll path. A batch whose members all succeed is unaffected.

## How to Test

```
cargo test -p ethrex-test --test ethrex_tests atomic_batch
```

`atomic_batch_revert_drops_the_batch_writes_from_the_bal` executes a frame tx whose batch
writes a slot then reverts, with BAL recording enabled as the builder and importer do. It
asserts both halves of the EIP-7928 rule: the reverted **change** is gone, and the
reverted **access** stays — the address is still in the list and the slot is re-filed as
a read. A slot in neither list is a slot the block accessed and the BAL omits, which
re-execution rejects; asserting only the first half would pass in that direction.

`atomic_batch_unroll_keeps_frame_status_and_gas_but_drops_logs` pins the second half:
the frame that ran keeps `SUCCESS` and its actual gas, the failing frame is charged less
than its `gas_limit`, the remaining member is `SKIPPED` at 0, and no log from the batch
reaches the transaction log set.

Both fail without the corresponding fix. The three spec fixtures named above also pass
with this branch and fail without it, on `gasUsed`.

On a local Amsterdam + Hegotá dev node the tx that previously stalled the chain now mines
with zero BAL failures.

`eip8141` 86, `eip7928` 57, `eip8037` 46, `eip8038` 39, `eip7906` 34, `eip8272` 8,
`eip8250` 3, `ethrex-blockchain` 30 — all passing. Clippy `-D warnings` and `fmt` clean in
both workspaces.

## Checklist

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. _(N/A — BAL recording and frame-receipt gas only.)_


test/tests/levm/eip8141_tests.rs
```diff
@@ -3604,3 +3604,224 @@ fn storage_refund_from_a_later_frame_reduces_reported_gas() {
...
+        "the batch unrolled, so the write must not survive; got {live}"
+    );
+
+    let bal = db
+        .bal_recorder
+        .take()
+        .expect("BAL recording was enabled")
+        .build();
+    let writer_entry = bal.accounts().iter().find(|acc| acc.address == writer);
+    if let Some(entry) = writer_entry {
```
ElFantasma: This assert is skipped when the writer is absent from the BAL entirely, so the test passes in exactly the regression direction that matters.

EIP-7928 has two halves and the fix depends on both: the reverted *change* goes, the reverted *access* stays. `restore` delivers the second by leaving `touched_addresses` alone — but nothing here pins it. If the writer entry vanished, `writer_entry` is `None`, the `if let` body never runs, and this goes green while the builder emits a BAL missing an address the block accessed — import rejects it, the same class of stall this PR fixes, just from the other side.

That's not hypothetical: `block_access_list.rs` exposes a second checkpoint type, `TxCheckpoint`, which *does* capture `touched_addresses_len` and roll it back. Swapping one for the other at the call site is a plausible edit, and this test would not notice.

Asserting the entry exists, and that the slot is now a read, pins both halves:

```rust
let entry = bal
    .accounts()
    .iter()
    .find(|acc| acc.address == writer)
    .expect("a reverted batch's accesses still belong in the BAL (EIP-7928)");
assert!(
    entry.storage_changes.iter().all(|change| change.slot != slot),
    "the BAL must not record a storage change for a slot whose write was unrolled"
);
assert!(
    entry.storage_reads.iter().any(|read| *read == slot),
    "the unrolled write must be re-filed as a read, not dropped"
);
```

(`AccountChanges` is `{ address, storage_changes, storage_reads: Vec<U256>, .. }`, so this lines up with the `change.slot != slot` comparison already here.)

docs/eip-8141.md
```diff
@@ -652,13 +653,15 @@ The top-level `succeeded` is **derived identically on encode and decode**: `true
 
 **Rationale**: Re-approving is likely a bug in the frame transaction construction. Failing loudly prevents subtle issues like a second payer accidentally overriding the first.
 
-### 11. Atomic-batch revert: gas and status of rolled-back frames
+### 11. Atomic-batch revert: gas and status of rolled-back frames (resolved)
 
-**Spec gap**: When a frame in an atomic batch fails, the spec rolls back the batch's state to the point before the batch began and skips the remaining frames (refunding their gas). It is silent on: (a) the gas charged to frames that already executed-and-succeeded *earlier* in the same batch; (b) the gas charged to the failing frame when it failed via `REVERT` (vs an exceptional halt); and (c) the receipt `status` of rolled-back-but-succeeded frames.
+**Formerly a spec gap**, now settled by the spec:
```
ElFantasma, line 672: Worth separating this from the BAL fix, or at least calling it out in the PR title/description — right now both say only "drop an unrolled atomic batch's writes from the block access list", and the second commit's subject disappears on squash-merge.

The two changes have quite different risk profiles. The BAL fix is a contained bug fix that unblocks a chain stall. This one re-prices an unrolled batch: per-frame gas and status feed the receipts trie, `total_gas_used` → header `gasUsed`, and the payer refund. The version of this doc on `main` called §11 "**UNRESOLVED, pending cross-client confirmation**" and said the choice "must be aligned with the reference / other clients before being pinned" — this PR pins it from a spec commit alone.

The direction looks right to me and the new behavior is self-consistent (I checked: skipped members are `(SKIPPED, 0, [])`, `total_gas_used` is left as executed, `all_logs.truncate(batch_logs_start)` drops the tx-level logs). The question is just whether the cross-client check the old note asked for has happened — if it has, saying so here would close it out; if not, landing the BAL half first keeps an urgent fix from waiting on a consensus decision.

Marchhill:
Tested this on a cross-client (ethrex ↔ another frame-tx client) devnet with this fix merged in, using a reverting atomic-batch spammer shape. **This fixes the validation/execution BAL but not the block-building BAL for unrolled atomic batches.**

- **Validation path — fixed.** ethrex now accepts every peer-proposed block containing a reverting atomic batch (0 rejections; 61 reverting-batch frame txs mined on the shared chain, both clients hash-for-hash). The `bal_recorder` checkpoint/restore on unroll works on the re-execution path.
- **Block-building path — still mismatches.** When ethrex *proposes* a block containing a reverting atomic batch, the sealed block's `BlockAccessListHash` still includes the unrolled batch member's storage write (which the reverted post-state doesn't contain). A spec-correct peer re-executes, drops the reverted write per EIP-7928, and rejects the block with `InvalidBlockLevelAccessListHash` (e.g. `Expected 0xe8dd75a6…, got 0x7b925ceb…`). Observed 68× on ethrex-proposed reverting-batch blocks; 0 on ethrex non-reverting blocks and 0 on peer blocks.

So the checkpoint/restore added on the unroll path is reached during validation/re-execution, but the block-production path seals the BAL without the same unroll restore — worth extending this PR to cover the builder path.

Repro shape: `[VERIFY(ExecPay, sender), SENDER(ATOMIC_BATCH, SSTORE writer), SENDER(reverting terminator)]` — the terminator reverts, the batch unrolls, and the writer's SSTORE must be dropped from the block-level access list.

ilitteri (approved):

PR #851, opened 2024-10-14
fix(l2): gracefully stop `ProofDataProvider`

ManuelBilbao:
**Motivation**

When `Ctrl+C` is sent to the process, the `ProofDataProvider` listener is still opened waiting for new connections, causing the process to not finish.

**Description**

Now the listener is launched in a new thread, and the previous thread have a signal handler that sends a message under a `mpsc` channel to the listener so it stops.

Closes #809

PR #7069, opened 2026-07-30
feat(l1): add debug_getRawBlockAccessList and align BAL getter errors

edg-l:
## Motivation

[execution-apis#851](https://github.com/ethereum/execution-apis/pull/851) — **merged 2026-08-24**, now in `src/eth/block.yaml` and `src/debug/getters.yaml` — pins the semantics of the two EIP-7928 block-access-list getters, and glamsterdam-devnet-8 lists it as required. Two gaps against it:

- `debug_getRawBlockAccessList` is not implemented at all.
- `eth_getBlockAccessList` returns `null` for a block that predates Amsterdam, where the
  spec requires `-32001: Resource not found`. A block whose access list cannot be produced
  is also indistinguishable from one that has none.

`null` for a pre-fork block is the wrong signal: it tells a caller the block is unknown
when the truth is that block access lists do not exist yet at that height.

## Description

Both getters now share one resolver:

| case | `eth_getBlockAccessList` | `debug_getRawBlockAccessList` |
|---|---|---|
| unknown block, or `pending` | `null` | `-32001` |
| predates Amsterdam | `-32001` | `-32001` |
| Amsterdam+, history to answer is pruned | `4444` | `4444` |

`debug_getRawBlockAccessList` returns the RLP encoding and accepts a block number, tag or
hash (`BlockNumberOrTagOrHash`, widened by #851 from number-or-tag).

The `4444` row covers three concrete situations: the block body is pruned, a regenerated
list does not match the header commitment, and — following review — the parent header or
parent state needed for re-execution is pruned (previously that fell into the blanket
`-32603 Internal` mapping, which is exactly the failure mode of a snap-synced node asked
about a block below its pivot).

> [!IMPORTANT]
> The `4444` rows are a **behaviour change to the existing `eth_getBlockAccessList`
> contract**, not just new surface: these cases previously returned `null`. A caller that
> treated `null` as "no BAL" now gets a hard error distinguishing "unavailable" from
> "unknown", which is the point of the spec change.

Where the codes come from (per review, now documented on the `RpcErr` variants):
`4444: Pruned history unavailable` was introduced across the eth/debug getters by
[execution-apis#636](https://github.com/ethereum/execution-apis/pull/636) — EIP-4444
deliberately left RPC behaviour out of scope — and #851 requires it for the BAL getters.
`-32001: Resource not found` is likewise per #851.

Two more review follow-ups:

- An Amsterdam **genesis** block is served its empty-by-construction access list instead of
  an error (it has no parent to re-execute from; `0xc0` is the spec's own no-accesses raw
  example).
- Both getters enforce an exact arity of 1, consistent with the rest of the
  `debug_getRaw*` family.

The Amsterdam boundary is decided from the chain config rather than from the presence of
the header's BAL commitment, so a header missing the field for another reason is not
silently reported as pre-fork.

`eth_getBlockAccessList` already emitted all six account fields with empty lists, which
#851 makes explicit in the `AccountAccess` schema; that behaviour is unchanged.

## Testing

Eight cases in `test/tests/rpc/block_access_list_tests.rs`: the raw getter serving RLP by
hash, the raw getter's `-32001` on an unknown block, the JSON getter's `-32001` on a
pre-Amsterdam block, the unknown-hash-returns-`null` case, both `4444` paths (body pruned;
parent state pruned), the genesis empty list served as `0xc0`, and extra-argument
rejection.

Also exercised against a dev node on an Amsterdam genesis: both getters over a block
carrying an `SSTORE` transaction (9 accounts, all six fields present), the raw getter by
both hash and tag, and each error path above.


crates/networking/rpc/utils.rs
```diff
@@ -205,6 +209,16 @@ impl From<RpcErr> for RpcErrorMetadata {
                 data: None,
                 message: format!("Proof generation unavailable: {context}"),
             },
+            RpcErr::ResourceNotFound(context) => RpcErrorMetadata {
+                code: -32001,
+                data: Some(context),
+                message: "Resource not found".to_string(),
+            },
+            RpcErr::PrunedHistoryUnavailable(context) => RpcErrorMetadata {
+                code: 4444,
```
ElFantasma, line 225: `-32001` is cited to execution-apis#851, but `4444` isn't cited to anything — could you add the source?

Two things make it worth pinning down rather than taking on trust:

- **It's the only positive code in the enum.** Every other arm here is negative (`-32000`, `-32015`, …), and grepping `crates/` finds no prior use of `4444` anywhere. JSON-RPC 2.0 reserves `-32768..-32000` and leaves the rest to the application, so a positive code is legal — just unprecedented in this codebase.
- **`RpcErr` is shared.** Once `PrunedHistoryUnavailable` exists, any handler can return it, so a wrong code propagates repo-wide rather than staying local to these two getters.

If it comes from the EIP-4444 history-expiry discussion rather than a merged execution-apis spec, that's fine — worth saying so in the doc comment so the next person doesn't have to re-derive it. If #851 or a sibling PR does pin it, a link in the PR body settles it.

crates/networking/rpc/eth/block_access_list.rs
```diff
@@ -1,79 +1,159 @@
...
+        .blockchain
+        .generate_bal_for_block(&full_block)
+        .map_err(|e| RpcErr::Internal(format!("Failed to generate BAL: {e}")))?;
+
+    // Only serve a regenerated BAL that matches the header commitment; a
+    // mismatch means it was re-executed against wrong/incomplete state.
+    match bal.filter(|bal| bal.matches_commitment(commitment, &NativeCrypto)) {
+        Some(bal) => Ok(ResolvedBal::Found(bal)),
+        None => Err(RpcErr::PrunedHistoryUnavailable(format!(
+            "block access list for {block_hash:#x} cannot be reconstructed"
```
ElFantasma, line 178: Neither `4444` path has a test. The four new cases cover `-32001` (pre-Amsterdam, and unknown block on the raw getter), the raw getter serving RLP, and the retained unknown-hash-`null` — but nothing asserts `code == 4444` for either this commitment-mismatch branch or the missing-body branch above it.

That matters more than usual here because `4444` is a **behaviour change for `eth_getBlockAccessList`, not just new surface**. Both of these previously returned `Ok(Value::Null)`; they now return a hard error. A client that treated `null` as "no BAL for this block" now gets a failure instead, and the PR body presents that under "neither stored nor reconstructable → 4444" without calling out that it's a change to the existing getter's contract.

The missing-body case looks straightforward to cover with the existing harness — store a header without its body and assert the code. The commitment-mismatch case is more work, so testing just the first would already close most of the gap.

Separately: the pre-Amsterdam check reading `is_amsterdam_activated(header.timestamp)` rather than sniffing `block_access_list_hash` is the right call, and the doc comment explains why. Worth keeping that reasoning where it is.

ilitteri (approved):

PR #852, opened 2024-10-14
fix(l2): contracts version

ManuelBilbao:
**Motivation**

Forge throws an error about `solc` version (although you have the needed one)

**Description**

Specify the `solc` version in `foundry.toml` and flex the contracts version using `^0.8.27`

Closes #839

PR #7070, opened 2026-07-30
feat(l1): report EIP-8037 two-dimensional gas in callTracer output

edg-l:
## Motivation

[execution-apis#852](https://github.com/ethereum/execution-apis/pull/852) defines the
gas-related part of the `callTracer` output, and glamsterdam-devnet-8 lists it as required.
It adds the EIP-8037 two-dimensional gas breakdown to the top-level frame, which ethrex
does not report, so a caller cannot see how a transaction's gas split between the
execution and state dimensions.

## Description

`CallTraceFrame` gains three optional fields, emitted on the top-level frame only:

| field | meaning |
|---|---|
| `regularGasUsed` | gross pre-refund regular gas, the transaction's contribution to `block_regular_gas_used` (EIP-7778) |
| `stateGasUsed` | gross pre-refund state gas, net of source-based state-gas refunds |
| `gasRefund` | the applied EIP-3529 refund |

They are stamped in `run_call_trace` from the `ExecutionReport` when the fork is Amsterdam
or later. Before the fork all three stay `None` and are skipped during serialization, so
they are absent from the JSON rather than present as zero, which #852 requires. Sub-frame
attribution is unspecified, so sub-frames are never stamped.

The spec states the identity `regularGasUsed + stateGasUsed == gasUsed + gasRefund`, which
holds unless the EIP-7976 calldata floor binds.

## Review follow-ups & rebase onto merged #6931 (2026-08-27)

#6931 has landed, so the stack note above is gone: main was merged in and every conflict resolved by taking main's version and re-applying this PR's own commit on top — the branch's net diff against main is now exactly this feature (+342/−3 across three files).

Two review findings are addressed:

- **Frame transactions are skipped** (greptile P1, confirmed real): an EIP-8141 frame transaction's `ExecutionReport.gas_refunded` holds *unused frame gas*, not the EIP-3529 refund `gasRefund` advertises, so tracing one would publish a wrong value and break the accounting identity. The report marks these transactions itself (`frame_results` is `Some`), and the three fields are now omitted for them — consistent with the spec defining no aggregate attribution for frame transactions.
- **The gas split asserts its invariant**: `state_gas_used ≤ gas_used` holds by construction, and the `saturating_sub` would silently report a plausible-looking zero if it ever broke. A `debug_assert!` now turns that into a signal, mirroring the payload builder's pattern for the same split.

The doc claim questioned in review — `regularGasUsed` being the transaction's contribution to `block_regular_gas_used` — is confirmed literal: the payload builder accumulates `report.gas_used.saturating_sub(report.state_gas_used)` per transaction (`crates/blockchain/payload.rs`, the EIP-8037 gas-tracking block), the same expression this tracer stamps.

## Testing

Three cases in `test/tests/levm/trace_call_tests.rs`, all driving a real execution rather
than hand-built frames:

- an Amsterdam transaction that writes a fresh slot and clears a pre-existing one, so both
  `stateGasUsed` and `gasRefund` are non-zero, asserting the identity above and the hex
  wire values;
- the same bytecode under an Osaka chain config, asserting all three are absent from the
  serialized JSON;
- a transaction whose contract `CALL`s a second account, asserting the sub-frame omits all
  three while the top frame carries them.

Also checked against a dev node on an Amsterdam genesis. A fresh `SSTORE` reports
`stateGasUsed` of 97,920 (`STATE_BYTES_PER_STORAGE_SET` x `CPSB`) and satisfies the
identity exactly; clearing the slot reports a 5,627 refund, the EIP-3529 cap, and again
satisfies it. A contract creation reports 183,600 (`STATE_BYTES_PER_NEW_ACCOUNT` x `CPSB`)
and is the documented floor-binding exception, with `regularGasUsed` equal to the floor.


crates/vm/backends/levm/tracing.rs
```diff
@@ -273,9 +273,19 @@ impl LEVM {
...
+        let fork = vm.env.config.fork;
+
+        let mut callframe = vm.get_trace_result()?;
 
-        let callframe = vm.get_trace_result()?;
+        // EIP-8037 two-dimensional gas, reported on the top-level frame only from
+        // Amsterdam onwards (execution-apis `CallFrame`).
+        if fork >= Fork::Amsterdam {
+            callframe.regular_gas_used =
+                Some(report.gas_used.saturating_sub(report.state_gas_used));
```
ElFantasma, line 328: The decomposition is sound — I checked the basis rather than assuming. `errors.rs:251-263` documents `gas_used` as pre-refund on EIP-7778+ and `state_gas_used` as *"State gas portion of gas_used"*, so both are on the same pre-refund footing and `gas_used - state_gas_used` is a genuine split, not a mix of gross and net. Gating on `fork >= Fork::Amsterdam` also correctly carries forward to Hegotá and later.

Two things:

**1. `saturating_sub` masks the invariant it depends on.** `state_gas_used ≤ gas_used` holds by construction — it's a portion of it. If that ever stopped holding, this silently reports `regularGasUsed: 0x0`, which is a *plausible-looking* value served straight to an RPC caller doing gas analysis, rather than a failure anyone would notice. A checked subtraction (or a `debug_assert!` alongside) would turn a broken invariant into a signal instead of a quiet zero.

**2. A doc claim I couldn't confirm.** The field comment says `regularGasUsed` is *"its contribution to `block_regular_gas_used` (EIP-7778)"*. I couldn't locate where `block_regular_gas_used` actually accumulates per transaction — `payload.rs` only reads it (`:881`, `:1057`, `:1186`, and the `max()` at `:323`) in the lines I sampled, so I can't tell whether the accumulated expression is literally `gas_used - state_gas_used`. If it is, ignore this. If it differs even slightly, the comment asserts an equivalence that doesn't hold, and that's the kind of thing a consumer of this field will rely on.

Worth noting for landing order: base is `feat/debug-trace-call-alt`, so this needs #6931 to merge first. #6931 is approved but still open.

ilitteri (approved):

PR #7071, opened 2026-07-30
fix(l1): reject historical state reads when the block's state is unavailable

### LINKED ISSUE #7028: Historical-state RPC methods silently return latest state for unavailable roots
> **Updated 2026-07-28** after the deep-reorg stack (#6686 / #6687 / #6689 / #6724) merged. Verified against current `main`: the new read cascade does **not** fix this, and `has_state_root` is unchanged — the issue stands as written. But the reverse-diff journal makes a *better* fix possible than the one originally suggested; see "Option B" below.

### Summary

Several historical-state RPC methods open the state trie at the requested block's root **without verifying that root's state is actually available**. Because the on-disk trie is path-keyed and single-version (only the latest committed root is correct), a query for a block whose state is not resident in the trie-layer window silently returns **the latest committed state instead of an error** — a wrong answer with a 200 response.

### Affected methods (unguarded)

- `eth_getBalance` / `eth_getCode` / `eth_getTransactionCount` — resolve `header.state_root` and read via `open_state_trie` with no availability check (`crates/networking/rpc/eth/account.rs`)
- `eth_getStorageAt` — same, via `get_storage_at_root` (`crates/storage/store.rs`)
- `eth_getProof` — same, via `get_account_proof`; there is a `// TODO: check state root` next to it in `store.rs`

### Correctly guarded methods (for contrast)

`eth_call`, `eth_estimateGas`, `eth_createAccessList` and `debug_trace*` go through `StoreVmDatabase::new`, which checks `has_state_root(header.state_root)` and returns a clean "state root missing for block ..." error (`crates/blockchain/vm.rs`); tracing re-executes from the nearest stateful ancestor.

### The deep-reorg read cascade does not cover this

`TrieWrapper::get` now cascades **layer cache → overlay → disk**, but the overlay stage is not a historical-state server:

- An overlay exists only while a deep-reorg apply pass is in flight (`install_overlay_for_reorg`), and it is torn down on the first new-chain commit.
- Even then, `overlay_serves` answers only for the pivot `serves_root` or a root already in the (new-chain) layer cache. An arbitrary historical root falls straight through to disk, exactly as before.
- `has_state_root` is unchanged: it opens the trie at the root, re-hashes the root node, and compares.

So in steady state the cascade is a no-op for these methods and the silent-wrong-answer path is intact.

### When it bites

- On a **fully-synced node**: any query for a block older than the resident window (`head - DB_COMMIT_THRESHOLD`, 128) returns head-window state silently.
- On a **syncing node**: the RPC server is up and answers, so historical queries into ranges whose state was pruned or never resident return current disk state silently. Other clients return "missing trie node" / "state not available" here.
- Worth noting the window is not a constant: with #7023 the resident window shrinks to ~the tip while a node is catching up far from head, which widens the silently-wrong range during that phase (it returns to the normal 128-block window on arrival at head). The same residency assumption produced a separate reporting bug — `eth_syncing.currentBlock` reading stale during catch-up — fixed in #7023, which is a good indication that "callers assume the window is wide" is a recurring class rather than a one-off.

### Suggested fix

**Option A (minimum, honest errors).** Apply the same `has_state_root` guard the guarded methods already use before serving a state read at a historical root, and return a standard "state not available / pruned" JSON-RPC error instead of wrong data. `eth_getProof` should additionally resolve its `store.rs` TODO. This is small and strictly better than today.

**Option B (now possible: actually answer).** The reverse-diff journal spans `[finalized+1, latest]` — pruned only when finality advances — and `Overlay::from_journal` already reconstructs virtual state at an arbitrary pivot in that range. That machinery could back these reads, letting the node **serve correct historical state throughout the unfinalized window** instead of erroring for anything older than 128 blocks. It costs an overlay build per query range (the journal walk is `O(entries)` with a bloom-filtered lookup), so it likely wants caching or a bounded range, but it turns a limitation into a capability.

Suggested sequencing: land Option A first so the node stops returning wrong answers, then evaluate Option B on top — the guard from A is what tells you when the journal path needs to be consulted.

### Context

Surfaced while mapping trie-layer-window consumers during review of #7018 / #7023: the recent-state window is the only place older roots resolve correctly, and these methods implicitly assume the queried block is inside it without enforcing that. Related: #6686 (journal), #6687 (overlay + read cascade), #7023 (residency during catch-up).

ElFantasma:
**Motivation**

`eth_getBalance`, `eth_getCode`, `eth_getTransactionCount`, `eth_getStorageAt` and `eth_getProof` open the state trie at the requested block's root without checking that the state is actually there. The on-disk trie is path-keyed and single-version, so opening it at a root that is no longer retained does not fail — it resolves against whatever the latest committed root holds. The node therefore answers a query about an old block with **another block's state and a success response**, which the caller has no way to detect.

`eth_getProof` is the worst case: it returns a proof that verifies against a root the caller never asked about.

This is already known in-tree. `test/tests/l2/state_reconstruct.rs` carries:

```rust
// TODO: Historical state is not supported in the DB currently by the client.
// This is due to the newest path-based trie implementation.
// ... The commented tests below make no sense until then.
// test_state_block(&addresses, 0, 0).await;
// test_state_block(&addresses, 6, 50).await;
```

i.e. the tests were commented out rather than the wrong answers fixed. Other clients return an error here (geth: `missing trie node …`).

Resolves #7028.

**Description**

Check state availability before serving these five reads, and return an error instead of a wrong value:

- New `RpcErr::StateNotAvailable`, mapped to JSON-RPC code **`-32000`** — the generic server-error code geth and erigon use for this, with a message that leads with `missing trie node <root>` so tooling that already matches on that phrasing keeps working.
- A shared `ensure_state_available` helper in `eth/account.rs` runs `has_state_root` on the resolved block's header, mirroring the guard `eth_call`/`estimateGas`/`debug_trace*` already get via `StoreVmDatabase::new`.

The guard is deliberately narrow: it errors **only** when the block resolves to a header whose state root is gone — the case where the node would otherwise lie. If the block can't be resolved to a header at all, the existing behaviour is kept, so `pending`, `latest` and unknown-block handling are unchanged (for `Pending` the tag resolves to the latest block, whose state is resident, and the mempool nonce logic is untouched).

It is applied at the RPC boundary rather than in the store on purpose. The store's lenient behaviour is relied on elsewhere: snap serving reads arbitrary pivot roots, and `crates/storage/store.rs` has a test asserting that unserved roots keep the flat-KV fast path. Guarding inside the store would break both.

**Behaviour change worth flagging:** on a node that has advanced more than `DB_COMMIT_THRESHOLD` (128) blocks, `eth_getBalance(addr, "earliest")` and other genuinely-historical queries now return an error rather than a plausible-looking wrong number. That is the point of the change, and it matches other clients, but it is visible to anything that was unknowingly consuming the wrong values.

Not addressed here: actually *serving* historical state. The reverse-diff journal added by #6686 spans `[finalized+1, latest]` and `Overlay::from_journal` can already reconstruct any pivot in that range, so a follow-up could answer these queries correctly across the unfinalized window instead of erroring. #7028 tracks that as Option B; this PR is Option A, and the guard is what tells a future implementation when to consult the journal.

**Testing**

`test/tests/rpc/historical_state_guard_tests.rs`, driven through the public JSON-RPC dispatch (`map_eth_requests`) so it exercises the real wire path:

- all five methods return `StateNotAvailable` for a canonical block whose state root the store does not hold;
- all five still succeed when the state is present, so the guard cannot be silently over-strict.

Verified non-vacuous: with the guard disabled the first test fails with `eth_getBalance answered "0x0" from another block's state instead of erroring` — the exact bug.

`ethrex-rpc` unit tests (98, including the existing `latest`/`pending` account tests) and the RPC integration tests (43) pass; fmt and clippy clean.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A: no storage change.


ElFantasma:
## Review responses

Thanks — three of the findings were right and are fixed in 9e6f8b1db.

### Codex #1 / Greptile — EIP-1898 `blockHash` selectors (fixed)

Correct, and it is the same bug class this PR exists to fix. `resolve_block_number` turns a hash into a height via `BLOCK_NUMBERS`, which maps *any* stored block — canonical or not — to its number, and every read below is canonical-by-number. So a hash naming a non-canonical block was answered with its canonical sibling's state.

The guard now rejects that: if the identifier is a hash and it is not the canonical hash at that height, the request errors instead of answering about a different block.

I deliberately did not go the other way (plumbing hash-based reads end-to-end). That needs store APIs that read at an arbitrary block's state root, and it would change the `requireCanonical` posture that `BlockIdentifierOrHash::parse` documents as intentionally permissive — a bigger change than this PR, and one that should be argued on its own. Refusing is strictly better than the status quo of silently serving the sibling's state; serving non-canonical state correctly can come later.

New test `state_reads_reject_a_non_canonical_block_hash` covers it, and I verified it fails when the canonicality check is neutralised.

### Kimi #1 — `async` with nothing awaited (fixed)

Right: `get_block_header` and `has_state_root` are both sync. The helper is now a plain `fn` and the five call sites lost their `.await`.

### Kimi #2 — redundant header lookup in `eth_getProof` (fixed)

Also right. The helper now returns the header it validated, so `GetProofRequest` consumes that instead of fetching it a second time — the double lookup is gone rather than just acknowledged.

### Greptile — guard/read TOCTOU

Fair, and worth stating explicitly: the availability check and the read are separate operations, so a commit or reorg landing between them can still change what the read sees. That window is inherent to check-then-read without a snapshot, and closing it means holding a consistent view across both — larger than this change. Worth noting that the failure mode in that race is the *current* behaviour (an answer from a newer root), so this is no worse than today; the guard removes the systematic case, not the racy one.

### Kimi #3 — `"temp.db"` with `EngineType::InMemory`

Left as is: that is the existing convention in this test directory (e.g. `block_access_list_tests.rs`, `fork_choice_tests.rs` all pass `"temp.db"` to an in-memory store). Changing it here alone would make this file the odd one out; a sweep would be a separate cleanup.

### `claude-review` CI job

Infrastructure, not a finding — it fails during setup without producing review content (same as on #7008). The other three reviewers ran fine.

PR #7072, opened 2026-07-30
fix(l2): bound trie-layer memory in committer re-execution paths

### LINKED ISSUE #7026: L2 checkpoint regeneration can accumulate unbounded trie layers (same pattern as #7018)
### Summary

The L2 checkpoint-regeneration path re-executes blocks via `add_block_pipeline(block, None)`, which gates trie-layer commits on the canonical `head - DB_COMMIT_THRESHOLD` safe-commit root. As in the L1 case fixed by #7018, these blocks are re-executed without a forkchoice update, so `safe_commit_root` never advances to cover them and the depth-based commit never fires — the in-memory trie diff-layer backlog grows with the number of re-executed blocks.

### Locations

- `crates/l2/sequencer/l1_committer.rs:678` — `execute_batch_to_generate_checkpoint`
- `crates/l2/sequencer/l1_committer.rs:1672`

### Risk

Low today: L2 checkpoint batches are far smaller than the ~11k-block L1 regeneration gaps that triggered the L1 OOM (#7018). This is a latent hazard, not an active failure, and it is not externally triggerable with arbitrary size — batch sizes are internally bounded. It would matter only if L2 batch sizes grow large.

### Suggested fix

Route these re-execution loops through the depth-gated bounded path (`add_block_pipeline_bounded(block, None, DB_COMMIT_THRESHOLD)`), the same mechanism #7018 applied to the L1 regeneration / full-sync / import paths. Since checkpoint regeneration extends a single canonical chain (no competing forks), the depth gate is sound here.

### Context

Surfaced during review of #7018 (bound single-canonical-chain re-execution memory via depth-gated commit).

ElFantasma:
**Motivation**

Three L2 committer paths re-execute blocks through `add_block_pipeline(block, None)`, which commits trie layers on the canonical safe-commit gate. That gate keys on `head - DB_COMMIT_THRESHOLD` and only advances on a forkchoice update, so during a re-execution loop it never fires: every block adds an in-memory trie diff-layer and nothing is ever committed. This is the same defect #7018 fixed for the L1 paths, where it grew to ~60 GB over an ~11k-block regeneration and OOM'd the node.

`regenerate_state` (`crates/l2/sequencer/l1_committer.rs`) is the one that matters. It is the L2 analogue of L1's `regenerate_head_state`, and its own doc comment describes the setup precisely:

> blocks behind the head block while the state of the blocks in between are kept in in-memory-diff-layers. After the node is shut down, those in-memory layers are lost […] the state needs to be regenerated by re-applying the blocks from the last known state root up to the head block.

Its loop runs `last_state_number + 1..=target_block_number` — the full regeneration gap. That is not bounded by batch size; it grows with how far committed state lags the head, exactly as on L1. (#7026 assumed this was batch-bounded and therefore low risk; it isn't.)

Resolves #7026.

**Description**

Route the three single-canonical-chain re-execution sites through the depth-gated entry point added in #7018, `add_block_pipeline_bounded(.., DB_COMMIT_THRESHOLD)`, so the resident layer count stays bounded regardless of how many blocks are replayed:

- `regenerate_state` — startup/state-gap regeneration (unbounded range; the important one).
- the one-time checkpoint recreation loop — replays a batch's blocks.
- `produce_batch`'s re-execution fallback — replays candidate blocks when the execution cache misses.

Committing by depth is sound in all three: each replays a single canonical chain with no competing forks, which is the precondition the depth gate documents. Live `newPayload` is untouched and keeps the canonical gate.

No behaviour change beyond memory: the same blocks are executed in the same order, and every commit still writes its `STATE_HISTORY` journal entry (journaling follows `wait_for_flush`, not `commit_depth`), so reorg reach is unaffected.

**Testing**

No new test. The bound itself is already covered on the storage side by `bounded_reexec_without_fcu_bounds_memory_and_serves_window` (added in #7018), which drives >128 blocks through `add_block_pipeline_bounded` with no forkchoice update and asserts both that old layers are committed and that the recent window stays resident. Reproducing that at the L2 committer layer would need a full sequencer harness to re-test the same mechanism.

`cargo build -p ethrex-l2 -p ethrex`, `cargo clippy -p ethrex-l2` and `cargo fmt` are clean; the L2 integration suite exercises these paths in CI.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — N/A: no storage change.


ElFantasma:
### CI: `State Reconstruction Tests` is not from this PR

Two independent reasons:

**1. The same job fails at the same step on unrelated branches.** `State Reconstruction Tests` → `step 10: Run tests` is currently failing on `feat/l1-eip8070` and `eip-7906` as well. Three different changes, one failure point. (It is intermittent rather than uniformly broken — #7071 passed this job twice today.)

**2. This change cannot affect that test.** The reconstruction fixture is ~29 blocks (blobs cover 1–21; the assertion reads block 29). `get_commitable_by_depth` only returns `Some` once `counter >= threshold`, and the threshold here is `DB_COMMIT_THRESHOLD` (128), so with ~29 layers it returns `None` and **nothing is committed** — byte-for-byte the previous behaviour. The gate can only change anything past 128 resident layers, which this test never reaches.

Every check this PR can influence is green (33 pass at time of writing). I'll re-run the job once the workflow finishes and confirm it flips.

ElFantasma:
## Review responses

Thanks all — Codex found nothing, Greptile 5/5, and the two actionable items are handled.

### Kimi — import grouping (applied)

Grouped into `use ethrex_storage::{DB_COMMIT_THRESHOLD, EngineType, Store};` in eeb438024.

On the error-handling difference you noted between the three sites: agreed, and kept as is — each site matches the convention of the function it lives in (`?` with `From` propagation in the first two, `if let Err(err)` mapping to `CommitterError::FailedToCreateCheckpoint` in `regenerate_state`).

### Claude — `l2_connection.rs` `process_blocks_on_queue` (checked; not the same bug)

Good lead, and your hypothesis is right — confirmed in the code rather than assumed. That loop calls `apply_fork_choice(...)` immediately after every `add_block_pipeline` (`crates/networking/p2p/rlpx/l2/l2_connection.rs`, the FCU right below the add). So the canonical head advances on every iteration, `safe_commit_root` follows at `head - DB_COMMIT_THRESHOLD`, and the canonical gate fires continuously — layers cannot accumulate the way they do in the three loops this PR fixes, which run with no forkchoice update at all.

It should also *stay* on the canonical gate: that path is an L2 follower applying blocks received from a peer, i.e. the live-equivalent path, which is exactly where the canonical gate's "never persist non-canonical state" guarantee is needed. Switching it to the depth gate would be a regression, not an extension of this fix.

Thanks also for independently checking the single-canonical-chain precondition at all three sites — that each of them builds a fresh isolated checkpoint blockchain rather than sharing the live one is the property the depth gate depends on, and it is worth having verified by someone other than the author.

### CI

`State Reconstruction Tests` is unrelated to this change — same job and same `step 10: Run tests` is failing on `feat/l1-eip8070` and `eip-7906` too, and the fixture chain here is ~29 blocks while the depth gate cannot commit anything below 128 layers, so this PR is a strict no-op for it. Detail in the comment above; a re-run is queued for when the workflow finishes.

ElFantasma:
Confirmed: CI is fully green now (34 pass, 0 failures). The earlier `State Reconstruction Tests` failure was transient, as expected — it was failing on unrelated branches at the same step, and this change cannot affect that test (its fixture chain is ~29 blocks, and the depth gate commits nothing below 128 layers).

PR #7073, opened 2026-07-30
fix(l1): settle frame-tx fees on max_gas and the base blob rate

edg-l:
EIP-8141 changed its gas accounting in [ethereum/EIPs#11969](https://github.com/ethereum/EIPs/pull/11969) ("apply `charged_gas` to fee settlement"), merged 2026-07-28. Our pinned spec commit predates it. Two consensus-visible gaps.

## `max_cost` charges blobs at the base rate

> The transaction is only valid for inclusion in a block if `tx.max_fee_per_blob_gas >= blob_base_fee` of that block. `max_fee_per_blob_gas` is used only for this inclusion check. The payer is charged `blob_gas * blob_base_fee`; no additional blob fee is collected or refunded.

We charged `max_fee_per_blob_gas` at APPROVE and refunded the difference at the end, so the payer's final balance was already correct. What was wrong is what the chain can observe in between:

- `TXPARAM(0x06)` returned the max-rate figure, and contracts read it.
- A frame running after APPROVE saw the payer's balance debited by that figure via `BALANCE`/`SELFBALANCE`.

Now the blob terms in `max_cost` and `charged_fee` are identical, so they cancel: the burn is collected once, at the base rate, and never refunded. `FrameTxContext` carries the block's blob base fee, captured at transaction entry, since `load_tx_param` has no `Environment` handle.

## The reservation is `max_gas`, not `standard_gas_limit`

```
max_gas = max(standard_gas_limit, calldata_floor_gas)
```

Both sides carry the mandatory costs. A transaction whose EIP-7623 data floor exceeds what it declared for execution now **reserves the floor** rather than being rejected, so the static floor-reservation check is gone. `total_gas_limit()` returns `max_gas`; `standard_gas_limit()` and `calldata_floor_total()` are the two inputs.

The mempool paymaster reservation uses the same definition, so the reservation ceiling and `TXPARAM(0x06)` remain one number.

## Reachability

Only blob-carrying frame transactions see the `max_cost` change, and only data-heavy ones see the floor change. Our mempool rejects blob-carrying frame transactions today (no sidecar transport), but block import accepts them, so a block from another builder would have diverged.

## Testing

- `max_cost_is_max_fee_times_limit_plus_base_rate_blob_cost` sets `max_fee_per_blob_gas` a thousandfold above the base fee, so the old formula cannot pass.
- `max_gas_reserves_the_calldata_floor_instead_of_rejecting` and `max_gas_takes_the_calldata_floor_when_it_exceeds_the_standard_limit` pin both arms of the `max` and the removed rejection.
- Full suite green; `cargo clippy --workspace --all-targets` clean.


crates/vm/backends/levm/mod.rs
```diff
@@ -3199,7 +3205,7 @@ impl LEVM {
             .saturating_mul(U256::from(frame_tx.total_gas_limit()));
         let blob_cost = U256::from(frame_tx.blob_versioned_hashes.len())
             .saturating_mul(U256::from(131072u64))
-            .saturating_mul(frame_tx.max_fee_per_blob_gas);
+            .saturating_mul(blob_base_fee);
```
ElFantasma: Charging the blob fee at the base rate is right for consensus, but this helper does double duty and the mempool half loses a guarantee here.

The old term, `max_fee_per_blob_gas`, was a true upper bound: a block whose blob base fee exceeds it cannot include the transaction, so no execution could ever charge more than the reservation. `blob_base_fee` is read from the header passed into `simulate_frame_validation_prefix`, which on the admission path (`blockchain.rs:3572`) is the **current head** — while execution charges at the base fee of whatever later block includes the transaction. Blob base fee moves per block, so the reserved amount can come in under the eventual charge.

That matters because `max_cost` is the paymaster reservation, not just a display value — `blockchain.rs:3588` takes it, `:3635`/`:3644` gate admission on `available < max_cost`, and `:3661` stores it as `reserved_cost`. Under-reserving lets a paymaster accumulate pending frame txs whose true cost exceeds its balance, which is the thing the reservation exists to stop.

The comment just below still asserts the old contract:

> `we compute a reservation ceiling for the mempool, so saturating to U256::MAX on overflow is conservative — it just makes the reservation larger, not smaller.`

After this change that only holds for the gas term.

Bounded in practice (blob base fee moves ~12.5% per block and the gas term is usually dominant), so non-blocking. But the two callers now want different quantities: consensus wants the base rate, the mempool reservation wants `max_fee_per_blob_gas`. Simplest split is to keep `blob_base_fee` for the TXPARAM path and pass `frame_tx.max_fee_per_blob_gas` on the reservation path — or, if the relaxation is deliberate, say so in the comment so the next reader doesn't rely on "ceiling".

ElFantasma (approved):
Consensus side of both changes checks out; one non-blocking inline where the shared helper stops being the reservation ceiling its comment claims.

crates/vm/backends/levm/mod.rs
edg-l: Went with the split. The helper is now `frame_tx_reservation_ceiling(frame_tx)` and prices the blob term at `max_fee_per_blob_gas`; consensus keeps `blob_base_fee` in `compute_tx_max_cost` / TXPARAM 0x06.

Also renamed `FrameValidationOutcome.max_cost` to `reservation_ceiling`, since the old name was what made "ceiling" look true. Its doc comment now says it is not the TXPARAM 0x06 value and why: the including block's base fee is unknown at admission, and `max_fee_per_blob_gas >= blob_base_fee` is an inclusion condition, so the declared rate bounds every block that can include the tx.

Pinned by `reservation_ceiling_prices_blobs_at_the_declared_max_rate`: one blob, `max_fee_per_blob_gas = 1000`, head base fee 1, so the old code reserved 131072 and the ceiling is now 131072000.

ilitteri (approved):

Issue #7051, opened 2026-07-29
Branch layout for the Hegota frame-transaction EIPs


edg-l:
Informational, so the layout does not have to be rediscovered.

## Branches

| Branch | PR | Contents |
|---|---|---|
| `main` | | EIP-8141 frame transactions |
| `frames-devnet-0` | no PR | frames only, devnet config on top of 8141. Requested by STEEL for a future devnet |
| `eip-8250` | #6906 | keyed nonces |
| `eip-8272` | #6907 | recent roots |
| `eip-7906` | #6891 | transaction assertions |
| `focil` | #7039 | EIP-7805 (FOCIL) inclusion lists only |
| `hegota-devnet` | no PR | integration branch: 8141 + 8250 + 8272 + 7906, plus a modified EIP-7805 soon |

`frames-devnet-0` and `hegota-devnet` are deployment branches and are not meant to be merged, so they have no PR.

`frames-devnet-0` is the one STEEL asked us for, targeting a future devnet, which is why it has to stay a faithful frames-only reference and cannot pick up the other frame-transaction EIPs.

Its devnet config is `fixtures/networks/eip8141-devnet.yaml`, and it does **not** activate frame transactions at genesis: `fulu_fork_epoch: 0` and `gloas_fork_epoch: 0` put Osaka and Amsterdam at genesis, `heze_fork_epoch: 1` puts Hegotá at epoch 1, which at `seconds_per_slot: 6` is 192 seconds in. Frame transactions are rejected with `FrameTxPreFork` until then. That fixture also pairs `gloas_fork_epoch: 0` with stable `sigp/lighthouse:v8.1.3`, where `hegota-devnet` uses `ethpandaops/lighthouse:glamsterdam-devnet-7` with `gloas: 1`; gloas-at-genesis is known to make the beacon genesis a Gloas state that the glamsterdam image refuses to decode, and whether stable v8.1.3 accepts it has never been tested. Boot it before relying on it.

## Container images

`ethpandaops/eth-client-docker-image-builder` auto-builds `ethpandaops/ethrex` from `lambdaclass/ethrex` on every push, for the branches listed in its `branches.yaml`:

- `main`, `bal-devnet-4`, `focil` (the `el.clients.ethrex.branches` list)
- `glamsterdam-devnet-7`, `glamsterdam-devnet-8` (inherited from `el.common_branches`)

[...]

PR #7075, opened 2026-07-31
fix(l1): enforce the EIP-8141 post-prefix and replacement/eviction mempool rules

edg-l:
**Motivation**

Three EIP-8141 §Mempool rules were not enforced, and one of them is a DoS hole: a VERIFY frame placed after the validation prefix is admitted today, yet a reverting VERIFY frame invalidates the whole transaction wherever it sits. That lets a sender tie transaction validity to third-party state the prefix simulation never inspects, which is exactly the unbounded invalidation the public mempool rules exist to prevent.

**Description**

- Structural rule 8: reject a VERIFY frame after the validation prefix (`VerifyFrameAfterPrefix`).
- §Expiry Verifier Frame: an expiry verifier frame may only be the first frame of the list (`ExpiryFrameNotFirst`). They were treated as transparent anywhere.
- §Replacement and Eviction, replacement: a frame-tx replacement must raise both `max_fee_per_gas` and `max_priority_fee_per_gas` by at least `FRAME_TX_MIN_FEE_BUMP_PERCENT` (10%, the spec's conventional default). Frame txs previously took the shared "either fee dimension rises" path, and since `gas_price()` returns `max_fee_per_gas` for them, a replacement could raise the max fee while lowering the priority fee. One-wei bumps were free, and each replacement re-runs a validation-prefix simulation.
- §Replacement and Eviction, eviction: when arrival order picks a frame tx as the victim, the spec order decides which one goes, nearest expiry then lowest priority fee (`Mempool::worst_frame_tx`). Arrival order still decides whether a frame tx is dropped at all, so frame txs are neither starved nor privileged. The "already invalid against the head" tier is served by post-block revalidation, which has the state view the eviction path lacks.

Two existing tests encoded the old permissive behavior and now assert the rejections: `prefix_rejection_wrong_scope_only_verify_pay` (trailing pay frame) and `prefix_shape_deploy_self_verify_with_expiry_verifier_between` (misplaced expiry frame).

**Verification**

- `cargo test -p ethrex-test --test ethrex_tests`: 906 passed, 0 failed.
- `cargo check --workspace --all-targets`, `cargo clippy` on the touched crates, `cargo fmt --all`: clean.
- 7 new tests: rule-8 rejection and the non-VERIFY-frames-after-prefix case, expiry placement both ways, the replacement matrix (sub-10% bump, unbumped priority fee, lowered priority fee, valid bump), and the two eviction orderings.

Note for reviewers: the same fixes are cherry-picked onto `hegota-devnet` per #7051, where the replacement rule lives in `is_fee_bump` so the keyed (EIP-8250) replacement path is covered too. `main` has no such helper, so here it sits in `find_tx_to_replace`.


test/tests/blockchain/mempool_tests.rs
```diff
@@ -2958,3 +2958,172 @@ mod cumulative_balance_tests {
...
+    // EIP-8141 §Replacement and Eviction: a replacement is accepted only if it
+    // raises BOTH max_fee_per_gas and max_priority_fee_per_gas by at least the
+    // node's minimum increment. Each frame-tx replacement re-runs the
+    // validation-prefix simulation, so one-wei bumps would be free EVM work.
+    let mempool = Mempool::new(MEMPOOL_MAX_SIZE_TEST);
+    let sender = Address::from_low_u64_be(0xF00D);
+    let pooled = priced_frame_tx(sender, 0, 100, 10, None);
+    insert_frame_tx(&mempool, sender, pooled);
+
+    let cases: [(u64, u64, bool, &str); 5] = [
```
ElFantasma, line 3022: All five cases run against a non-zero baseline (`pooled` is `100`/`10`), so the `old_fee == 0` regime that the guard was specifically written for isn't covered.

`fee_is_bumped` is `new_fee > old_fee && new_fee * 100 >= old_fee * 110`, and its doc comment names the strict-greater conjunct as the part that "carries the zero case, where the percentage bound alone would admit an equal bid". That reasoning is correct: at `old_fee == 0` the percentage bound degenerates to `new_fee * 100 >= 0`, which holds for every `new_fee`, including `0`.

The exposure is that the conjunct *reads* as redundant — `new * 100 >= old * 110` does imply `new > old` for every positive `old`, so a later simplification that drops it keeps the whole suite green. After that, a pooled frame tx with `max_priority_fee_per_gas = 0` (perfectly legal) accepts unlimited equal-bid replacements, each re-running the validation-prefix simulation. That is precisely the DoS this PR closes, reintroduced silently.

It needs its own small test rather than a sixth row here, since this matrix's `pooled` is `(100, 10)` and a `(110, 0)` row would just re-test the already-covered "lowered priority fee" path:

```rust
let pooled = priced_frame_tx(sender, 0, 100, 0, None); // priority fee 0
// (110, 0) must still be UnderpricedReplacement: equal bid, no bump
```

crates/blockchain/mempool.rs
```diff
@@ -459,21 +475,65 @@ impl MempoolInner {
...
+    /// deadline sort last.
+    ///
+    /// The order's first tier — transactions already invalid against the current
+    /// head — is served by `Blockchain::revalidate_frame_txs_after_block`, which
+    /// has the head-state view this path lacks.
+    ///
+    /// Scans the pool, so it runs only when arrival order already picked a frame
+    /// transaction for eviction.
+    fn worst_frame_tx(&self) -> Option<H256> {
+        self.transaction_pool
```
ElFantasma, line 513: This scans the entire `transaction_pool`, where the blob-eviction analogue just below deliberately scans only its sub-pool, and documents the reasoning:

> `O(N) scan over the blob sub-pool (N <= max_blob_mempool_size, 512 today). Fine at this cap; revisit (e.g. a priority index) before exposing a much larger cap via CLI.`

Both conditions that comment sets for "revisit" are already true here on day one: N is `max_mempool_size`, which defaults to `10_000` (`MAX_MEMPOOL_SIZE_DEFAULT`, ~20x the blob cap) and *is* CLI-exposed, via `ETHREX_MEMPOOL_MAX_SIZE`.

The doc comment above bounds how *often* this runs — "only when arrival order already picked a frame transaction for eviction" — which is a fair argument, but that predicate is attacker-chosen. Keeping the arrival-order head a frame tx makes the scan fire on essentially every admission once the pool is full, and a full pool is the steady state under load. That's a ~10k-entry scan per admitted transaction, reachable by exactly the party the rest of this PR hardens against.

Non-blocking, and a full scan may well be the right first cut while frame txs are new. But it's worth either recording that tradeoff in the doc comment the way the blob path does, or keying frame txs in their own index so the scan is over the frame subset instead of the whole pool.

ElFantasma (approved):
Two non-blocking inlines (an untested zero-fee edge, and an eviction scan wider than the 512-capped precedent it mirrors).

ilitteri (approved):

PR #7076, opened 2026-07-31
ci(l1): seed sepolia snapsync jobs with reachable bootnodes

ilitteri:
**Motivation**

Sepolia's five EF execution-layer bootnodes stopped answering UDP discovery on 2026-07-30, in a window between 12:39 and 15:44 UTC, and have not come back since.

Every snapsync job starts from an empty datadir, so the bootnodes compiled into the binary are its only route into the DHT. With them silent a fresh node never finds a single peer, snap sync cannot even fetch the sync-head header, and the job just runs until the assertoor deadline expires — about 3h30m per sepolia job.

Both sepolia jobs now fail on every run. Run `30594206688` is representative: `sepolia - Lighthouse` and `sepolia - Prysm` each burnt 3h31m and failed, while both hoodi jobs passed normally. That is over 7h of the workflow's 6h cron window, so runs are queuing behind each other and the backlog compounds (`cancel-in-progress: false`).

The failure is not a regression and not a fork mismatch. Commit `2f1593f2` passed sepolia twice before failing on the same SHA; sepolia's `eth_config` reports its current fork with `next: null`, matching our genesis exactly. Reproduced independently on a separate host and network: ~9.9k RLPx handshake timeouts and **zero** inbound discv4 messages from those five addresses, against a hoodi node peering normally at the same moment. The hosts themselves are alive — they answer ICMP, and one still serves consensus-layer discv5 — so only the execution-layer discovery service is gone. Upstream `eth-clients/sepolia` still lists all five unchanged, so nothing has been retired on paper.

**Description**

Pass a set of reachable sepolia nodes, taken from the network's public DNS node list, as `--bootnodes` for the sepolia matrix entries in `.github/actions/snapsync-run/action.yml`.

- Only sepolia is affected. Hoodi's bootnodes are healthy and it ships 16 of them rather than 5, so its jobs are left exactly as they were.
- Verified against the current `main` image with DNS-based discovery unavailable: peers appear within 30 seconds, where the default bootnodes give none after several minutes.
- The generated `network_params.generated.yaml` was validated for both networks by running the step's script body and parsing the result, so the injected list cannot silently break the hoodi path.

This is deliberately a stopgap. The seeded nodes are ordinary community nodes rather than dedicated infrastructure, so they will go stale, and hardcoding host addresses is the exact failure mode being worked around here. For that reason they are **not** added to the client's bundled bootnode list.

The durable fix is for ethrex to read that DNS node list itself (EIP-1459), which is what the companion PR does. This block should be deleted once that lands.

**Checklist**

Not applicable — this changes only CI configuration, with no effect on the client or its storage.

### DIFF
--- a/.github/actions/snapsync-run/action.yml
+++ b/.github/actions/snapsync-run/action.yml
@@ -83,7 +83,27 @@ runs:
         ETHREX_TAG: ${{ inputs.ethrex_tag }}
         CL_TYPE: ${{ inputs.cl_type }}
         CL_IMAGE: ${{ inputs.cl_image }}
+        # Community sepolia nodes taken from all.sepolia.ethdisco.net on
+        # 2026-07-31. Only used to seed discovery; see the note below.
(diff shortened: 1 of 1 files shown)


ilitteri:
Durable fix: #7077. Once that lands this seeding block should be deleted.

ilitteri:
Superseded by #7079, which fixes the same outage in the bundled sepolia bootnode list instead — that covers CI, the multisync host and users in one change, rather than the snapsync action alone. Suggest closing this in favour of it.

ilitteri:
Closing in favour of #7079, which fixes the same outage in the bundled sepolia bootnode list instead. That covers CI, the multisync host and users in one change, whereas this one patched the snapsync action only — and so would not have helped the multisync host, which builds from `tooling/sync/docker-compose.multisync.yaml` and never reads this action.

#7079 has since been soaked on ethrex-multisync-main with trie validation enabled: sepolia went from zero peers and no state downloaded in eight hours to peering and syncing normally.

PR #7077, opened 2026-07-31
feat(l1): add EIP-1459 DNS discovery for bootstrap resilience

ilitteri:
**Motivation**

Discovery bootstrapping relies entirely on the bootnodes compiled into the binary: **4** for mainnet, 5 for sepolia, 16 for hoodi. When those hosts stop answering, a node whose peer table is empty has no path into the DHT at all — it re-pings the same dead addresses forever and sits at zero peers.

That is not hypothetical. All five of sepolia's EF execution-layer bootnodes stopped answering UDP discovery on 2026-07-30 and have not come back, and every fresh sepolia node has been stuck at zero peers since, unable to start snap sync even though the network itself is healthy and the consensus client peers normally. Reproduced on two unrelated hosts and networks: ~9.9k RLPx handshake timeouts and zero inbound discv4 messages from those five addresses, while a hoodi node on the same box peered normally.

Every other client survives this because it ships a DNS node list as a second, independent bootstrap source. ethrex was the only one without one. Note that mainnet is the *more* exposed network here, with one fewer bootnode than sepolia.

A bundled list of discv5 ENRs is **not** an equivalent fallback, even though it looks like one. go-ethereum's `V5Bootnodes` and Nethermind's `discv5-bootnodes.json` are the same 15 records in both clients, and all 15 are consensus-layer nodes: an `eth2`/`attnets` key, no `eth` key, and a TCP port of either 0 or a beacon port. None of them can ever become an execution-layer peer. A DNS node list is the only bootstrap source that is both independent of the bootnodes and actually made of execution-layer nodes. This is recorded in the module docs so the option isn't mistaken for a fix later.

**Description**

Implements EIP-1459 DNS-based node discovery in a new `crates/networking/p2p/discovery/dns.rs`.

A node list is a DNS-hosted Merkle tree of ENRs described by an `enrtree://<base32-pubkey>@<domain>` URL, so it can be refreshed without shipping a new binary:

- The root TXT record is **signature-verified** against the tree's public key, so a hijacked resolver cannot inject nodes. The signed preimage is the record without its `sig=` field; a test pins this against the real sepolia root so a change in how it is constructed fails loudly.
- Every interior record is addressed by `base32(keccak256(record)[..16])`, and that label is **verified on fetch**, which makes the whole tree tamper-evident under the one root signature.
- Each ENR's own signature is verified before its node is accepted.
- Both the node count and the DNS request count per sync are bounded, so a cyclic or adversarial tree cannot walk us indefinitely. Links to other trees are followed one level deep.

The walk runs as a background task rather than gating startup, since a slow resolver must never delay the node coming up. Nodes are added to the peer table exactly the way bootnodes are, so they get pinged, validated and dialed through the existing paths. A sync that yields nothing retries after a minute and backs off toward the 30-minute steady-state interval — when the bootnodes are unreachable this is the only way in, so waiting a full interval over a momentary resolver failure would leave the node peerless for that whole time.

Trees for mainnet, sepolia and hoodi are enabled by default. `--discovery.dns` / `ETHREX_DISCOVERY_DNS` overrides them; passing an empty string disables the feature.

Deliberately uses `new_contacts` rather than `new_contact_records`: the record path runs `evaluate_fork_id`, which returns `Some(false)` for an ENR carrying no `eth` key, and `is_fork_id_valid == Some(false)` marks a contact for pruning — routing DNS-discovered ENRs through it would silently discard them.

Adds one dependency, `hickory-resolver` (default features off, `system-config` + `tokio`), for TXT lookups. It sits behind a `TxtResolver` trait, so the tree walk is tested without DNS and the backend is a one-line swap.

**Testing**

16 unit tests covering base32 round-tripping and label lengths, enrtree URL parsing, root parsing and signature verification against the real sepolia root, rejection of a root signed by another key and of a tampered root, label-substitution detection, real ENR leaf parsing plus signature verification, an end-to-end walk over a locally-signed tree, request-budget termination on a cyclic tree, one unreachable tree not blocking the others, and the retry backoff schedule.

Validated A/B on sepolia while the bootnodes were still unreachable:

| | contacts from DNS | peers @30s | peers @35min |
|---|---|---|---|
| DNS discovery on | 250 | 3 | 5 |
| off (`--discovery.dns ""`) | — | 0 | 0 |

**Checklist**

Not applicable — no storage format change, and no resync is required.

---

Stopgap that unblocks CI while this is reviewed: #7076. That PR should be reverted once this lands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2252,6 +2252,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4235,6 +4246,7 @@ dependencies = [
  "futures",
  "hex",
  "hex-literal 0.4.1",
+ "hickory-resolver",
  "hkdf",
  "hmac",
  "indexmap 2.14.0",
@@ -5012,6 +5024,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "r-efi 6.0.0",
+ "rand_core 0.10.1",
  "wasip2",
  "wasip3",
 ]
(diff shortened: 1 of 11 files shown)


ilitteri:
Stopgap that restores sepolia syncing while this is reviewed: #7079. Once this lands, the hardcoded addresses added there stop being load-bearing.

MysticRyuujin:
I tested this PR at commit `74c0e063`.

It looks good generally speaking. `SystemResolver` walked the live trees to completion with nothing rejected: 3000 nodes from `all.mainnet.ethdisco.net`, 250 from sepolia and 433 from hoodi, which are the counts go-ethereum's `dnsdisc` reads from the same trees. I found a few issues worth mentioning:

**1. The tests could use an update, because `stops_at_request_budget_on_a_cyclic_tree` always passes.**

The test puts a bare `enrtree-branch:` record in the mock DNS. That record is an empty branch. The walk therefore stops after two lookups. It never spends the request budget.

Two experiments show the gap. First, I replaced the budget check in `sync_tree` with `if false`. All 16 tests still passed. Second, I made `visited_labels.insert` always report a new label. All 16 tests still passed. No test covers either protection.

One way to cover the budget is a chain of 40 branches, where each branch names the next one. A resolver that counts the lookups can then assert that the count stops at the budget. A diamond, where two branches name the same leaf, would cover the visited set.

The test name may also be worth a second look. A tree cannot hold a cycle, because a cycle needs a record that contains its own label, and the label is the hash of that record. The budget protects against a large tree, not against a cycle.

**2. The description of `new_contacts` may be worth updating, because its reason does not apply to your default trees.**

The description says that the record path discards an ENR with no `eth` key. I read every record in the three default trees. They all carry an `eth` key that decodes: 3000 of 3000, 250 of 250, and 433 of 433.

`new_contacts` still looks like the correct choice, for a different reason. A DNS contact that answers a ping then gets a discv4 ENR request. `discv4_validate_enr_fork_id` reads the record and the fork verdict from the node itself. The DNS copy only has to supply an address to ping.

PR #7080, opened 2026-08-01
fix(l1,l2): stop removedb from faking success on the wrong datadir

### LINKED ISSUE #6925: removedb silently no-ops and reports success when run without the node's --network
## Summary

`ethrex removedb` resolves its target directory the same way the node does — `compute_effective_datadir(datadir, network, dev)` = `datadir.join(network.datadir_suffix())` (`cmd/ethrex/cli.rs`). For `--network=<genesis.json>`, `datadir_suffix()` returns `chain-<chain_id>` (`crates/common/config/networks.rs`). So a node started with `--datadir=/data --network=/path/genesis.json` keeps its DB at `/data/chain-<chain_id>`.

`removedb` must therefore be run with the **same** `--datadir` **and** `--network`. If `--network` is omitted (or differs), `compute_effective_datadir` returns a *different* path. `remove_db` then calls `init_datadir(effective)` — which **creates** that directory — followed by `remove_dir_all(effective)`, and logs **"Database removed successfully."** Net effect: it creates and deletes an empty/unrelated directory, reports success, and **leaves the real `chain-<chain_id>` DB untouched.**

## Reproduction

1. Run a node: `ethrex --datadir /data --network /path/genesis.json …` → data lands in `/data/chain-<chain_id>`.
2. Stop it, then run `ethrex removedb --datadir /data --force` (note: **no `--network`**).
3. It logs `Database removed successfully`, but `/data/chain-<chain_id>` still holds the full DB.
4. Restart the node. If the DB was in a no-retained-state condition it then crash-loops on:
   `Error: Unknown state found in DB. Please run `ethrex removedb` and restart node`
   — i.e. `removedb` reported success, yet the very error it is meant to clear persists. The actual remedy is `rm -rf /data/chain-<chain_id>`, or rerunning `removedb` **with** the node's `--network`.

(Encountered while recovering a glamsterdam-devnet-6 node.)

## Suggested fix

Make `removedb` resistant to this footgun:
- Don't `init_datadir` (create) the target inside `remove_db` — only remove a dir that **already exists**, and report the resolved path + whether it actually held data.
- And/or warn/refuse when the computed effective dir is empty/missing while a sibling `chain-*` dir under `--datadir` contains data (mirroring the existing migration warning in the `RemoveDB` handler).
- At minimum, log the resolved effective path and avoid printing "Database removed successfully" when nothing of substance was removed.

NikhilSharmaWe:
## Summary
Fixes #6925.

`removedb` could resolve the wrong network datadir (e.g. default `mainnet` instead of `chain-<id>`), create that empty directory, delete it, and print success while the real DB was left alone.

This change:
- never creates the target path just to delete it
- refuses the wipe when the resolved path has no valid DB but a sibling under `--datadir` does (`chain-*`, public network dirs, or the old unsuffixed layout)
- keeps `--force` as “skip confirm” only
- leaves `import --removedb` soft when the target dir is missing, so a fresh network under a shared base still works

Also updates L2 `removedb` to use the same honest delete path, and adds unit tests for the main cases.

## Test plan
- [x] `cargo test -p ethrex --lib -- remove_db_ require_datadir guard_ find_candidates explicit_removedb import_style`
- [ ] Smoke: `removedb` without the node’s `--network` should refuse and leave `chain-*` alone
- [ ] Smoke: same command with the matching `--network` should actually remove the DB

### DIFF
--- a/cmd/ethrex/Cargo.toml
+++ b/cmd/ethrex/Cargo.toml
@@ -140,3 +140,6 @@ thiserror = { workspace = true, optional = true }
 
 [lints]
 workspace = true
+
+[dev-dependencies]
+tempfile.workspace = true
--- a/cmd/ethrex/cli.rs
+++ b/cmd/ethrex/cli.rs
@@ -865,8 +865,12 @@ impl FromStr for LogColor {
 }
 
 /// Find directories under `base` that look like real ethrex databases
-/// (`metadata.json` present), including the unsuffixed base, known public-network
-/// suffixes, and `chain-*` custom-genesis / L2 dirs.
+/// (`metadata.json` with a recognized schema version).
+///
+/// Checks the unsuffixed base and every immediate subdirectory. Name-based
+/// filters (public-network lists, `chain-*`) are intentionally avoided so the
+/// removedb guard fails closed if a future network suffix is added without
+/// updating a hand-maintained list.
 pub fn find_valid_datadir_candidates(base: &Path) -> Vec<PathBuf> {
     let mut candidates = Vec::new();
 
@@ -874,20 +878,8 @@ pub fn find_valid_datadir_candidates(base: &Path) -> Vec<PathBuf> {
         candidates.push(base.to_path_buf());
     }
 
-    for suffix in Network::all_datadir_suffixes() {
-        let path = base.join(suffix);
-        if has_valid_db(&path) {
-            candidates.push(path);
-        }
-    }
-
     if let Ok(entries) = std::fs::read_dir(base) {
         for entry in entries.flatten() {
-            let name = entry.file_name();
-            let name = name.to_string_lossy();
-            if !name.starts_with("chain-") {
-                continue;
-            }
             let path = entry.path();
             if path.is_dir() && has_valid_db(&path) {
                 candidates.push(path);
@@ -930,9 +922,9 @@ pub fn guard_remove_datadir(base: &Path, effective: &Path) -> eyre::Result<()> {
             "refusing to remove {effective:?}: no valid database there, but found one at the old \
              unsuffixed location {base:?}.\n\
              Re-passing the same `--datadir` will not help while `--network` still appends a \
-             subdirectory. Migrate first, remove the old directory manually \
-             (e.g. `rm -rf {base:?}`), or run removedb with `--network local-devnet` (without \
-             `--dev`) so no subdirectory is appended and the target is {base:?}.\n\
+             subdirectory. Migrate first (start the node once so the DB moves into the network \
+             subdirectory), then re-run removedb with the matching `--network`, or remove the old \
+             directory manually (e.g. `rm -rf {base:?}`).\n\
              Databases found under {base:?}:\n{alternate_list}"
         );
     }
@@ -1558,7 +1550,7 @@ mod tests {
             err.contains("Re-passing the same `--datadir` will not help"),
             "{err}"
         );
-        assert!(err.contains("local-devnet"), "{err}");
+        assert!(err.contains("Migrate first"), "{err}");
         assert!(err.contains("rm -rf"), "{err}");
         assert!(has_valid_db(base));
     }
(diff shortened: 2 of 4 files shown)


cmd/ethrex/Cargo.toml
```diff
@@ -140,3 +140,6 @@ thiserror = { workspace = true, optional = true }
 
 [lints]
 workspace = true
+
+[dev-dependencies]
+tempfile.workspace = true
```
ElFantasma, line 145: `tempfile` lands in the root `Cargo.lock`, but `tooling/` is a second workspace with its own lockfile and this PR doesn't touch it. Its `ethrex` entry still carries the pre-PR dep set, so `Check Cargo.lock` should fail on:

```
cargo metadata --locked --manifest-path tooling/Cargo.toml
```

(`Makefile:269`.) Cargo records dev-dependencies in the lockfile's `dependencies` array alongside normal ones, so a dev-only dep still rotates both lockfiles - I checked, and `tempfile` is present in the root `ethrex` block and absent from the `tooling/` one.

Running `cargo metadata --manifest-path tooling/Cargo.toml` without `--locked` regenerates it. #7067 hit exactly this and it cost a round-trip, so flagging it before CI finishes rather than after.

The `.workspace = true` form itself is right, and the root lockfile entry is correct.

cmd/ethrex/cli.rs
```diff
@@ -850,30 +864,150 @@ impl FromStr for LogColor {
...
-            if input.trim().eq_ignore_ascii_case("y") {
-                std::fs::remove_dir_all(datadir).expect("Failed to remove data directory");
-                println!("Database removed successfully.");
-            } else {
-                println!("Operation canceled.");
+    if has_valid_db(base) {
+        candidates.push(base.to_path_buf());
+    }
+
+    for suffix in Network::all_datadir_suffixes() {
```
ElFantasma: This makes `all_datadir_suffixes()` load-bearing for a safety decision, and it's a hand-maintained list whose own comment says "Update this when adding new PublicNetwork variants."

Before this PR that list drove migration detection, where missing an entry means a directory doesn't get migrated - annoying, recoverable. Here a missing entry means `find_valid_datadir_candidates` doesn't see that sibling, `guard_remove_datadir` finds no alternates, and it returns `Ok(())`. **The guard fails open**, which on this code path means proceeding with a `remove_dir_all` the guard exists to prevent.

Worth noting the coverage is currently exact - I checked `datadir_suffix()` against it and every arm is reachable here: `mainnet`/`hoodi`/`sepolia` from the static list, `chain-*` from the `read_dir` scan (covering both `L2Chain` and `GenesisPath`), `dev`, and bare `base`. So this is about the invariant's durability, not a present gap.

Two options, either fine:

- Derive the list from the same match `datadir_suffix()` uses, so adding a variant can't silently skip it.
- Or invert the default: if the resolved target has no valid DB and the base directory contains *any* subdirectory that looks like a datadir, refuse regardless of whether it matched a known suffix. Fails closed, which is the right direction for a destructive command.

A test that adds a hypothetical unknown-suffix sibling and asserts the guard still refuses would pin whichever you pick.

cmd/ethrex/Cargo.toml
NikhilSharmaWe: Looked into it and I don't think we need to update `tooling/Cargo.lock` for this change.

`tempfile` is only added under `[dev-dependencies]` on `ethrex`. In the tooling workspace, `ethrex` is used as a path dependency, not as a workspace member, so Cargo does not include its `[dev-dependencies]` in that lockfile.

I verified locally that `cargo metadata --locked --manifest-path tooling/Cargo.toml` passes, and running it without `--locked` does not change the lockfile. #7067 needed the tooling lock update because it added a normal dependency on `ethrex-dev`, which is different from this case.

Happy to take another look if I missed something.

cmd/ethrex/cli.rs
NikhilSharmaWe: Went with fail-closed.

Candidate discovery now looks at any immediate subdirectory with a valid ethrex DB, not the hand-maintained suffix list. Added a test with an unknown sibling name (`weirdnet`) to lock that in.

NikhilSharmaWe:
cc @ilitteri

PR #7083, opened 2026-08-01
fix(levm): reserve the calldata floor instead of rejecting the frame transaction

AnkushinDaniil:
EIP-8141 resolves a frame transaction whose frames reserve less gas than its own calldata floor by raising the reservation, not by rejecting the transaction: the payer's maximum cost is taken against `max(standard_gas_limit, calldata_floor_gas)`.

`validate_static_constraints` rejected that transaction instead. A client that reserves the max accepts a block containing one, so the same block is valid for that client and invalid here — a consensus split rather than a difference in error text. Nethermind had the same bug independently; it is fixed there in NethermindEth/nethermind#12594.

This drops the rejection and adds `FrameTransaction::max_gas()`, which the reservation and the end-of-transaction refund read. The frames keep their own `gas_limit` allocations, so the extra headroom is never spendable — it is reserved and refunded.

The static-validation test that asserted the rejection is replaced by one asserting the raised reservation.

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2155,6 +2155,18 @@ impl FrameTransaction {
             )
     }
 
+    /// EIP-8141 `max_gas`: what the payer's maximum cost is reserved against.
+    ///
+    /// A transaction whose frames reserve less than its own calldata floor is not
+    /// invalid — the spec raises the reservation to the floor instead, so the payer
+    /// can always cover the floor-priced charge that settlement applies. The extra
+    /// headroom is never spendable by a frame (each keeps its own `gas_limit`); it
+    /// is reserved and refunded.
+    pub fn max_gas(&self) -> u64 {
+        self.total_gas_limit()
+            .max(self.mandatory_gas().saturating_add(self.calldata_floor_gas()))
+    }
+
     /// The expiry deadline (8-byte big-endian) of this transaction's expiry
     /// verifier frame, if one exists with well-formed data.
     pub fn expiry_deadline(&self) -> Option<u64> {
@@ -2320,15 +2332,6 @@ impl FrameTransaction {
                     Some(_) => {}
                 }
             }
-        }
-        // Per EIP-8141, the EIP-7623 calldata floor must be reserved independently
-        // of execution: the derived `tx_gas_limit` has to cover the mandatory costs
-        // plus the floor, or the transaction cannot pay for the data it carries.
-        let floor_gas = self.calldata_floor_gas();
-        if self.total_gas_limit() < self.mandatory_gas().saturating_add(floor_gas) {
-            return Err(format!(
-                "Total gas limit does not reserve the calldata floor of {floor_gas}"
-            ));
         }
         Ok(())
     }
--- a/crates/vm/levm/src/opcode_handlers/frame_tx.rs
+++ b/crates/vm/levm/src/opcode_handlers/frame_tx.rs
@@ -40,7 +40,7 @@ pub fn u256_to_offset(value: U256) -> Option<usize> {
 
 /// Compute the transaction's MAXIMUM cost (EIP-8141 §Gas Accounting: APPROVE must
 /// "collect the transaction's maximum cost from payer"):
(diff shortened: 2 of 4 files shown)


crates/common/types/transaction.rs
```diff
@@ -2321,15 +2333,6 @@ impl FrameTransaction {
                 }
             }
         }
-        // Per EIP-8141, the EIP-7623 calldata floor must be reserved independently
-        // of execution: the derived `tx_gas_limit` has to cover the mandatory costs
-        // plus the floor, or the transaction cannot pay for the data it carries.
-        let floor_gas = self.calldata_floor_gas();
```
ElFantasma, line 2327: Heads-up that this is not the only PR removing this block — **#7073** (`fix(l1): settle frame-tx fees on max_gas and the base blob rate`) deletes these exact lines and fixes the same consensus split. Both are open, so whichever lands second will conflict here and in `FrameTxContext`.

The two take different shapes:

| | #7083 (this) | #7073 |
|---|---|---|
| new API | adds `max_gas()` alongside `total_gas_limit()` | renames the old body to `standard_gas_limit()`, redefines `total_gas_limit()` as the `max()` |
| reaches | `FrameTxContext` / `compute_tx_max_cost` only | every existing `total_gas_limit()` caller |
| also does | — | blob fee at the base rate (separate EIP-8141 §Gas accounting fix) |

Not a judgement on which shape is better — the additive `max_gas()` is arguably clearer about intent, and #7073 carries an unrelated blob change that could just as well be split out. But they can't both merge as-is, and the second one through needs more than a conflict resolution, since the naming choices are mutually exclusive. Worth coordinating with @edg-l before either goes further.

crates/common/types/transaction.rs
```diff
@@ -2155,6 +2155,18 @@ impl FrameTransaction {
...
     }
 
+    /// EIP-8141 `max_gas`: what the payer's maximum cost is reserved against.
+    ///
+    /// A transaction whose frames reserve less than its own calldata floor is not
+    /// invalid — the spec raises the reservation to the floor instead, so the payer
+    /// can always cover the floor-priced charge that settlement applies. The extra
+    /// headroom is never spendable by a frame (each keeps its own `gas_limit`); it
+    /// is reserved and refunded.
+    pub fn max_gas(&self) -> u64 {
```
ElFantasma, line 2165: `max_gas()` is read only by `FrameTxContext`, so the two places that *reserve* against the payer still use the un-raised `total_gas_limit()` — and for a floor-dominated transaction, which is exactly the case this PR exists to admit, those differ.

- **Block gas pool.** `Transaction::gas_limit()` (`transaction.rs:1412`) still returns `tx.total_gas_limit()` for frame txs. EIP-8141 reserves `max_gas` from the block pool before execution, so a tx whose floor exceeds its frame allocations is charged `max_fee * max_gas` at APPROVE while the block only accounted for the smaller figure.
- **Mempool paymaster reservation.** `frame_tx_max_cost` (`crates/vm/backends/levm/mod.rs:3199`) multiplies `max_fee_per_gas` by `total_gas_limit()`. That value becomes `reserved_cost` and gates admission (`blockchain.rs:3588`/`:3635`/`:3661`), so the pool reserves less than execution will actually debit — under-reserving the paymaster the reservation exists to protect.

The refund path is fine, for what it's worth: `charged` comes from `compute_tx_max_cost` (now `max_gas`) and `total_gas_used = mandatory + max(data_and_execution, floor) <= max_gas`, so `charged >= owed` still holds.

Pointing `Transaction::gas_limit()` at `max_gas()` and using it in `frame_tx_max_cost` closes both. Worth noting #7073 avoids this by construction — folding the `max()` into `total_gas_limit()` means every existing caller picks it up — which is the main practical argument between the two shapes.

ElFantasma (commented):
Same bug and same deleted block as #7073, and one gap: `max_gas` doesn't reach the two sites that reserve against it.

edg-l:
Closing in favour of https://github.com/lambdaclass/ethrex/pull/7073

PR #7084, opened 2026-08-02
fix(l1): price the empty EIP-8272 reference list byte

AnkushinDaniil:
`recent_root_calldata()` returns an empty vector when `recent_root_references` is empty, so an empty list is charged nothing. On the wire the field is still present: `rlp([])` is the single byte `0xc0`, and EIP-8141 prices the whole payload as transaction data.

The gap is 64 gas on the EIP-7623 floor, which is a consensus split against a client that charges the byte — that is how it surfaced, as a `-64007402496 wei` balance divergence on a two-client devnet.

The existing unit test is updated: the expected calldata cost gains one non-zero token.

### DIFF
--- a/crates/common/types/transaction.rs
+++ b/crates/common/types/transaction.rs
@@ -2343,12 +2343,9 @@ impl FrameTransaction {
     /// EIP-8272 `recent_root_calldata`: `rlp(recent_root_references)`. The bytes
     /// this EIP adds to the payload are priced exactly as EIP-8141 prices frame
     /// and signature data, so they enter both `data_cost` and `calldata_tokens`.
-    /// Empty when no reference is declared, so a reference-free transaction's
-    /// gas is exactly the EIP-8141 figure.
+    /// An empty list still encodes as `0xc0` and occupies that byte on the wire,
+    /// so it is priced; only `recent_root_reference_intrinsic_gas` is waived.
     pub fn recent_root_calldata(&self) -> Vec<u8> {
-        if self.recent_root_references.is_empty() {
-            return Vec::new();
-        }
         let mut buf = Vec::new();
         self.recent_root_references.encode(&mut buf);
         buf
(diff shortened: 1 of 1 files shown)


crates/common/types/transaction.rs
```diff
@@ -5773,6 +5772,13 @@ mod tests {
         );
     }
 
+    #[test]
+    fn empty_recent_root_reference_list_is_charged_the_byte_it_occupies() {
+        let mut tx = make_test_frame_tx();
+        tx.recent_root_references = vec![];
+        assert_eq!(tx.recent_root_calldata(), vec![0xc0]);
```
ElFantasma, line 5782: nit: this asserts the helper's *encoding*, not that the byte is charged, so it doesn't guard the bug the name describes. Restoring the early return in either consumer — `data_cost` at :2301 or `calldata_tokens` at :2324, e.g. `if refs.is_empty() { 0 } else { ... }` — leaves `recent_root_calldata()` untouched and this test green, while the 64-gas divergence is back.

The real guard is already in the updated `recent_root_references_are_charged_calldata_gas`: its `calldata - 16` only holds because `base` now pays for the `0xc0`, so reintroducing the skip breaks that delta. So this is a naming/strength point, not a coverage hole.

Asserting the charge directly would make it match the name and cover the consumers too:

```rust
// 0xc0 is one non-zero byte: +16 on data_cost, and
// 1 * STANDARD_TOKEN_COST(4) * TOTAL_COST_FLOOR_PER_TOKEN(16) = +64 on the floor
let with_empty = tx.calldata_floor_gas();
tx.recent_root_references = vec![/* one ref */];
assert_eq!(tx.calldata_floor_gas() - with_empty, (refs_len - 1) as u64 * 64);
```

Or simply rename it to say what it checks (`empty_recent_root_reference_list_encodes_as_one_byte`).

AnkushinDaniil: The test now measures the charge instead of the encoding: it takes `calldata_floor_gas` with an empty list, adds one reference, and asserts the delta is exactly the added bytes times `FRAME_TX_STANDARD_TOKEN_COST` times `FRAME_TX_TOTAL_COST_FLOOR_PER_TOKEN`. Putting the `is_empty` skip back into `calldata_tokens` fails it, which I checked before and after.

ElFantasma (approved):

PR #7085, opened 2026-08-02
fix(levm): charge the EIP-8250 first-use surcharge on the default-code approval

AnkushinDaniil:
`consume_keyed_nonces` deducted the 20k first-use surcharge from the current call frame. The default-code approval path runs outside an ordinary frame, so the deduction landed nowhere and the surcharge was silently free.

The function now returns the surcharge instead of deducting it, and the default-code path charges it explicitly before applying the approval. Observed cross-client as a `16688117199824 wei` balance divergence on a keyed-nonce transaction whose sender has no code.

### DIFF
--- a/crates/vm/levm/src/opcode_handlers/frame_tx.rs
+++ b/crates/vm/levm/src/opcode_handlers/frame_tx.rs
@@ -254,8 +254,22 @@ impl OpcodeHandler for OpApproveHandler {
                 vm.current_call_frame.memory.len(),
             )?)?;
 
+        let approval_snapshot = vm
+            .frame_tx_context
+            .as_ref()
+            .map(|ctx| (ctx.payer_address, ctx.sender_approved));
         let surcharge = apply_approve(vm, scope_val, frame_target)?;
-        vm.current_call_frame.increase_consumed_gas(surcharge)?;
+        // The approval context is not database-backed, so the frame rollback that undoes the debit
+        // and the nonce consumption would leave the payer recorded and refunded at end of tx.
+        if let Err(err) = vm.current_call_frame.increase_consumed_gas(surcharge) {
+            if let (Some((payer, sender_approved)), Some(ctx)) =
+                (approval_snapshot, vm.frame_tx_context.as_mut())
+            {
+                ctx.payer_address = payer;
+                ctx.sender_approved = sender_approved;
+            }
+            return Err(err.into());
+        }
 
         let ctx = vm
             .frame_tx_context
--- a/crates/vm/levm/src/opcode_handlers/frame_tx.rs
+++ b/crates/vm/levm/src/opcode_handlers/frame_tx.rs
@@ -898,13 +898,13 @@ fn execute_default_verify(
     }
 
     // The surcharge comes out of this frame's budget like any other gas, so a frame that cannot
-    // afford it approves nothing.
-    let surcharge = vm.keyed_nonce_first_use_surcharge()?;
-    if surcharge > frame.gas_limit {
+    // afford it approves nothing. The affordability gate has to run before any state is touched,
+    // hence the estimate here; what is charged below is what consumption actually took.
+    if vm.keyed_nonce_first_use_surcharge()? > frame.gas_limit {
         return Ok((false, frame.gas_limit, Vec::new()));
     }
 
-    apply_approve(vm, allowed_scope, target)?;
+    let charged = apply_approve(vm, allowed_scope, target)?;
 
     let ctx = vm
         .frame_tx_context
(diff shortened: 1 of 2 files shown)


crates/vm/levm/src/opcode_handlers/frame_tx.rs
```diff
@@ -251,7 +254,8 @@ impl OpcodeHandler for OpApproveHandler {
                 vm.current_call_frame.memory.len(),
             )?)?;
 
-        apply_approve(vm, scope_val, frame_target)?;
+        let surcharge = apply_approve(vm, scope_val, frame_target)?;
+        vm.current_call_frame.increase_consumed_gas(surcharge)?;
```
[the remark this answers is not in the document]
AnkushinDaniil: Valid. `apply_approve` records the payer and mutates database state, and the surcharge charge that follows can return OutOfGas; the frame rollback restores the database-backed debit and nonce consumption but not `FrameTxContext`, which is not database-backed, so the failed approval survived and the payer was refunded at end of transaction. The approval context is now snapshotted before `apply_approve` and restored when the surcharge charge fails.

crates/vm/levm/src/vm.rs
```diff
@@ -1697,7 +1697,30 @@ impl<'a> VM<'a> {
                 current,
             )?;
         }
-        Ok(())
+        Ok(surcharge)
+    }
+
+    /// The surcharge `consume_keyed_nonces` will charge, without consuming anything.
+    pub(crate) fn keyed_nonce_first_use_surcharge(&mut self) -> Result<u64, VMError> {
```
ElFantasma, line 1841: This reimplements the slot derivation from `consume_keyed_nonces` character-for-character - same `[0u8; 64]` preimage, same `[12..32]`/`[32..64]` layout, same `keccak_hash`. Two copies of a consensus-critical derivation that must agree: if either drifts, the pre-check prices a different slot than the one actually consumed, and the default-code path charges a surcharge unrelated to what it did.

Worth collapsing to one private helper, something like `fn keyed_nonce_slot(sender: Address, key: &U256) -> H256`, called from both. Cheap now, and it makes the invariant structural instead of something a future edit has to remember.

crates/vm/levm/src/opcode_handlers/frame_tx.rs
```diff
@@ -862,6 +880,13 @@ fn execute_default_verify(
         return Ok((false, 0, Vec::new()));
     }
 
+    // The surcharge comes out of this frame's budget like any other gas, so a frame that cannot
+    // afford it approves nothing.
+    let surcharge = vm.keyed_nonce_first_use_surcharge()?;
```
ElFantasma: `apply_approve` now returns the surcharge it actually consumed, but this path discards it and reports the pre-computed estimate instead (`Ok((true, surcharge, ...))` below). The two agree today, though only because of an invariant that lives in another crate: `validate_static_constraints` rejects `nonce_keys` unless strictly ascending (`transaction.rs:2445`, `w[0] >= w[1]`), so no two keys can hash to the same slot and the pre-check cannot count a first-use that consumption then finds already set.

That is a real guarantee, not a gap - I verified it before writing this. But it is load-bearing for gas accounting and nothing here says so. Relax that ordering rule to allow duplicates and the estimate over-charges by 20k per repeated key, silently.

The pre-check has to stay where it is, since the affordability gate must run before any state is touched. But the *charge* could come from `apply_approve`'s return value rather than the estimate, which makes this locally correct regardless of what the ordering rule does later:

```rust
let surcharge = vm.keyed_nonce_first_use_surcharge()?;   // affordability gate only
if surcharge > frame.gas_limit {
    return Ok((false, frame.gas_limit, Vec::new()));
}
let charged = apply_approve(vm, allowed_scope, target)?; // what was actually consumed
...
Ok((true, charged, Vec::new()))
```

At minimum, a comment naming the strict-ordering rule as the reason the estimate is trustworthy.

crates/vm/levm/src/vm.rs
AnkushinDaniil: Collapsed into `keyed_nonce_slot(sender, key)`, called from both `consume_keyed_nonces` and `keyed_nonce_first_use_surcharge`.

crates/vm/levm/src/opcode_handlers/frame_tx.rs
AnkushinDaniil: Changed as you wrote it. The pre-check stays where it is as the affordability gate, and the value reported is now what `apply_approve` returned rather than the estimate, so gas accounting here no longer rests on the strict-ascending rule in another crate.

PR #7086, opened 2026-08-02
fix(levm): perform the EIP-8272 native write on every path into the predeploy

AnkushinDaniil:
`RECENT_ROOT_ADDRESS` carries no runtime bytecode, so the 64-byte `salt || root` write is executed natively. That native handler was only reachable from the CALL opcode: a transaction sent straight to the predeploy took the codeless-recipient fast path in `execute()` and succeeded without writing anything, which makes it impossible for a root source to publish a root the ordinary way.

Three parts:

- `is_simple_transfer_fast_path` excludes the predeploy, so the call reaches `run_execution`.
- `run_execution` dispatches a top-level call into the predeploy to the native write, next to the precompile branch, and records the address as BAL-touched.
- The write reports `gas_limit - gas_remaining` rather than the bare write cost, so intrinsic gas is included. Reporting only `RECENT_ROOT_WRITE_GAS` undercharged the transaction by exactly its intrinsic gas and produced a block-level access-list balance mismatch.

Verified on a Nethermind/ethrex devnet: after this change a root written through a plain transaction commits the same entry hash under the same storage key on both clients, and the block carrying it has the same hash.

There is no unit coverage here because levm has no in-crate VM harness to drive a top-level transaction, so the coverage is cross-client instead: two scenarios on a two-client devnet, one writing a root through a plain transaction and one through an EIP-8141 SENDER frame targeting the predeploy. Both compare the receipt, the block hash, the state root and the block-level access-list hash between the clients, and both fail without this change.

A fourth change rides along, and it is a separate defect from the native write. When an EIP-7906 assertion discards the body, the transaction-level log set was truncated but the per-frame receipt logs were left in place, so the header bloom committed to logs the receipt no longer carried. `frame_results` now drops them too. Found by inspection while reading the unroll path, not by an observed divergence, and there is no unit coverage for it, so it is the part most worth a second opinion.

### DIFF
--- a/crates/vm/levm/src/vm.rs
+++ b/crates/vm/levm/src/vm.rs
@@ -3544,7 +3544,9 @@ impl<'a> VM<'a> {
 
         // EIP-8272: the predeploy is codeless, so a transaction sent straight to it would
         // otherwise run as a transfer to an ordinary account and write nothing. Same native
-        // write the CALL-opcode and frame paths perform.
+        // write the CALL-opcode and frame paths perform. Only the top-level entry reaches
+        // here: a frame targeting the predeploy is codeless, so it takes the default-code
+        // path, where `execute_default_code` performs the write and never calls back in.
         if self.env.config.fork >= Fork::Hegota
             && self.current_call_frame.to == ethrex_common::types::frame_tx_recent_root()
         {
--- a/crates/vm/levm/src/vm.rs
+++ b/crates/vm/levm/src/vm.rs
@@ -2058,6 +2058,7 @@ impl<'a> VM<'a> {
         // depth-driven rather than a single revert/commit.
         let mut body_substate_depth: Option<usize> = None;
         let mut body_logs_start: usize = 0;
+        let mut body_frame_start: usize = 0;
         let mut state_gas_used_at_body_entry: i64 = 0;
         let mut post_tx_reverted = false;
 
@@ -2165,6 +2166,7 @@ impl<'a> VM<'a> {
                 body_substate_depth = Some(self.substate.backup_depth());
                 body_backup.bal_checkpoint = self.db.bal_recorder.as_ref().map(|r| r.checkpoint());
                 body_logs_start = all_logs.len();
+                body_frame_start = frame_idx;
                 state_gas_used_at_body_entry = self.state_gas_used;
             }
             absorbing_body = is_body_frame;
@@ -2693,8 +2695,16 @@ impl<'a> VM<'a> {
                 }
                 crate::utils::restore_cache_state(self.db, mem::take(&mut body_backup))?;
                 // Logs the body emitted are gone with its state. The prefix's logs
-                // (an APPROVE-side EIP-7708 transfer log, say) survive.
+                // (an APPROVE-side EIP-7708 transfer log, say) survive. The per-frame
+                // receipts keep their status and gas but lose those logs too — the
+                // consensus receipt carries only them, so the header bloom is built from
+                // them and would otherwise commit to logs that no longer happened.
                 all_logs.truncate(body_logs_start);
+                if let Some(ctx) = self.frame_tx_context.as_mut() {
+                    for (_, _, logs) in ctx.frame_results.iter_mut().skip(body_frame_start) {
+                        logs.clear();
+                    }
+                }
                 // EIP-8037: the body was unrolled, so it created no state and owes no
                 // state gas. Mirrors the atomic-batch unroll, which drops the state
                 // gas accumulated since batch entry for the same reason. EIP-7906 does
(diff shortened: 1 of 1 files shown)


crates/vm/levm/src/vm.rs
```diff
@@ -3123,6 +3184,15 @@ impl<'a> VM<'a> {
         // `refill_frame_state_gas`). Set in-region by `prepare_execution`.
         let top_frame_new_account_charged = self.value_new_account_charged;
 
+        // EIP-8272: the predeploy is codeless, so a transaction sent straight to it would
+        // otherwise run as a transfer to an ordinary account and write nothing. Same native
+        // write the CALL-opcode and frame paths perform.
+        if self.env.config.fork >= Fork::Hegota
+            && self.current_call_frame.to == ethrex_common::types::frame_tx_recent_root()
+        {
+            return self.run_top_level_recent_root_write();
```
ElFantasma, line 3553: `run_execution` isn't entered only for top-level transactions — it has three other call sites in this same file:

- `:1558` — the top-level transaction, the case you're targeting
- `:2261` — inside the frame-tx execution loop
- `:3016` — the other frame path

So this guard fires for **any** frame whose `to` is the predeploy, not just a transaction sent straight to it. An EIP-8141 frame that calls `RECENT_ROOT_ADDRESS` now routes here instead of wherever it went before, which is a behaviour change the description doesn't claim and the function name (`run_top_level_...`) actively denies.

The gas comment inside makes the same assumption explicit:

> `gas_remaining` enters this frame already net of intrinsic gas, so what the transaction owes is always measured against `gas_limit`, never the write cost alone.

That holds for the top-level frame, where `gas_limit` spans intrinsic. For a nested frame entered at `:2261`/`:3016`, `gas_limit` is the frame's own allocation and `gas_remaining` starts equal to it, so `gas_limit - gas_remaining` collapses to just the write cost — a different quantity than the one the comment describes, reached through the same code.

I haven't traced what a predeploy-targeted frame did before this change, so I can't say whether the new path is wrong for it or merely undocumented. Worth confirming, since "undercharged by exactly its intrinsic gas" is the bug this PR exists to fix and the frame path is where the asymmetry would hide. If it is meant to be top-level only, the guard needs a depth/call-site condition to match the name.

crates/vm/levm/src/vm.rs
```diff
@@ -2498,8 +2500,16 @@ impl<'a> VM<'a> {
...
                 // Logs the body emitted are gone with its state. The prefix's logs
-                // (an APPROVE-side EIP-7708 transfer log, say) survive.
+                // (an APPROVE-side EIP-7708 transfer log, say) survive. The per-frame
+                // receipts keep their status and gas but lose those logs too — the
+                // consensus receipt carries only them, so the header bloom is built from
+                // them and would otherwise commit to logs that no longer happened.
                 all_logs.truncate(body_logs_start);
+                if let Some(ctx) = self.frame_tx_context.as_mut() {
+                    for (_, _, logs) in ctx.frame_results.iter_mut().skip(body_frame_start) {
+                        logs.clear();
```
ElFantasma, line 2705: This is a fourth change, and the description lists three. It's also the one with the widest blast radius: clearing per-frame receipt logs on a body unroll changes what the consensus receipt carries, and your own comment says the header bloom is built from those logs.

That's a separate bug from the EIP-8272 native write — different EIP, different failure mode (a bloom committing to logs that didn't survive, versus a root that never gets written). It stands on its own and reads correct to me: `all_logs.truncate(body_logs_start)` already dropped them from the transaction-level set, so leaving them in `frame_results` was an inconsistency between the two views.

Two asks, no rework implied:

1. Add it to the description. A reviewer comparing the diff against "three parts" will assume they've miscounted, and a consensus change to receipt contents shouldn't arrive unannounced.
2. Say whether it was observed or found by inspection. The EIP-8272 half has cross-client evidence on the Nethermind devnet; if this half also produced a divergence, that's worth recording, and if it didn't, it's the part most worth a second opinion since there's no unit coverage here either.

ElFantasma (commented):
Two things: the dispatch isn't top-level-only as the name asserts, and there's a fourth change here the description doesn't mention.

crates/vm/levm/src/vm.rs
AnkushinDaniil: The guard is reachable only from `:1558`. A frame targeting the predeploy never gets to `run_execution`: the predeploy is codeless, so both frame call sites take the `bytecode.is_empty()` branch into `execute_default_code`, which intercepts `frame_tx_recent_root()` on its first line and returns `execute_recent_root_frame` without re-entering. Reaching the CallFrame branch needs non-empty bytecode or a 7702 delegation indicator, and a system account has neither. So the gas comment holds on every path that actually arrives here. I added a comment saying so, since nothing in the function did.

crates/vm/levm/src/vm.rs
AnkushinDaniil: Added to the description. It was found by inspection while reading the unroll path, not from an observed divergence, and it has no unit coverage, so it is the least supported part of the diff.

PR #7087, opened 2026-08-02
fix(l1): bound snap sync retries on PeerTable request timeouts

### LINKED ISSUE #6853: Sync retry loop has no backoff or consecutive-failure cap for recoverable errors
**Context:** #6823 classifies `ActorError::RequestTimeout` from the PeerTable actor as recoverable, so a transient mailbox-pressure timeout no longer terminates the node (`exit(2)`) mid-sync.

**Problem:** When a sync cycle fails with a recoverable error, `SyncManager::start_sync`'s loop re-enters `Syncer::start_sync` immediately as long as a snap header-download checkpoint exists — with no backoff and no consecutive-failure counter. That's correct for one-off transient timeouts, but if the cause is structural (peer-table actor wedged on a deadlocked request, persistent task starvation), the same `RequestTimeout` fires every ~5s and the node spins in a retry loop indefinitely instead of failing loudly.

**Proposed mitigations** (from review discussion on #6823):
1. Bound consecutive recoverable `RequestTimeout`s — e.g. 5 in a row without an intervening successful cycle escalates to fatal.
2. Add a small capped backoff (linear or 2× capped) in the `sync_manager.rs` retry loop on recoverable errors.

**Related:** in full sync (no snap checkpoint) a recoverable error breaks the manager loop and sync idles until the next FCU — pre-existing behavior for all recoverable error classes; a retry-semantics rework could cover both paths.

NikhilSharmaWe:
Closes #6853

## Summary

After #6823, PeerTable `RequestTimeout` is recoverable. With a snap header-download checkpoint, the manager could retry immediately forever if that timeout kept firing.

This change:

- returns a cycle outcome from `Syncer::start_sync`
- backs off on recoverable failures in the snap manager loop (1s, doubling, capped at 30s)
- exits after 5 consecutive PeerTable `RequestTimeout`s with no successful cycle in between
- leaves full sync alone (still stops until the next FCU when there's no checkpoint)

**Scope note:** backoff fixes the *immediate* retry for every recoverable error. The consecutive-failure cap (and shutdown) applies only to `RequestTimeout`. A persistent non-timeout recoverable error can still retry indefinitely at the 30s cap - intentional for #6853, not a hard give-up for `RecoverableOther`.

## Test plan

- [ ] `cargo test -p ethrex-p2p --lib sync::tests`
- [ ] `cargo test -p ethrex-p2p --lib sync_manager::tests`

### DIFF
--- a/crates/networking/p2p/sync_manager.rs
+++ b/crates/networking/p2p/sync_manager.rs
@@ -18,8 +18,105 @@ use tracing::{debug, error, info, warn};
 
 use crate::{
     peer_handler::PeerHandler,
-    sync::{SyncDiagnostics, SyncMode, Syncer},
+    sync::{SyncCycleOutcome, SyncDiagnostics, SyncMode, Syncer},
 };
+
+/// Maximum consecutive PeerTable `RequestTimeout` cycles without an intervening
+/// successful cycle before the snap manager loop escalates to `exit(2)`.
+/// The counter resets on each successful cycle. Value matches the pivot-update
+/// rotation budget in `snap_sync`.
+const MAX_CONSECUTIVE_REQUEST_TIMEOUTS: u64 = 5;
+
+/// Initial backoff after a recoverable sync-cycle failure in the manager loop.
+const INITIAL_RETRY_DELAY: Duration = Duration::from_secs(1);
+
+/// Upper bound for exponential backoff between recoverable sync-cycle retries.
+const MAX_RETRY_DELAY: Duration = Duration::from_secs(30);
+
+/// Action produced by [`SyncRetryState`] after observing a cycle outcome.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum SyncRetryAction {
+    /// Continue the manager loop; optionally sleep before the next cycle.
+    Continue { sleep: Option<Duration> },
+    /// Too many consecutive PeerTable request timeouts; escalate to fatal.
+    Fatal,
+}
+
+/// Tracks consecutive PeerTable request timeouts and backoff for the snap
+/// manager retry loop.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct SyncRetryState {
+    consecutive_timeouts: u64,
+    backoff: Duration,
+}
+
+impl SyncRetryState {
+    fn new() -> Self {
+        Self {
+            consecutive_timeouts: 0,
+            backoff: INITIAL_RETRY_DELAY,
+        }
+    }
+
+    fn on_outcome(&mut self, outcome: SyncCycleOutcome) -> SyncRetryAction {
+        match outcome {
+            SyncCycleOutcome::Success => {
+                self.consecutive_timeouts = 0;
+                self.backoff = INITIAL_RETRY_DELAY;
+                SyncRetryAction::Continue { sleep: None }
+            }
+            SyncCycleOutcome::RecoverableTimeout => {
+                self.consecutive_timeouts += 1;
+                if self.consecutive_timeouts >= MAX_CONSECUTIVE_REQUEST_TIMEOUTS {
+                    return SyncRetryAction::Fatal;
+                }
+                let sleep_for = self.backoff;
+                self.backoff = next_backoff(self.backoff);
+                SyncRetryAction::Continue {
+                    sleep: Some(sleep_for),
+                }
+            }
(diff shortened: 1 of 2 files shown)


crates/networking/p2p/sync_manager.rs
```diff
@@ -229,16 +327,36 @@ impl SyncManager {
...
+                    .is_some();
+                let action = retry_state.on_outcome(outcome);
+                match next_sync_loop_step(action, will_continue) {
+                    SyncLoopStep::ExitFatal => {
+                        error!(
+                            consecutive_timeouts = retry_state.consecutive_timeouts,
+                            max = MAX_CONSECUTIVE_REQUEST_TIMEOUTS,
+                            "Sync cycle failed with {MAX_CONSECUTIVE_REQUEST_TIMEOUTS} consecutive PeerTable request timeouts without a successful cycle; exiting"
+                        );
+                        std::process::exit(2);
```
ElFantasma: A hard `process::exit` from inside a spawned task skips the node's shutdown sequence entirely — no store flush, no recorded cause, destructors never run. On a node with a live RocksDB and a block-data buffer that is drained by an explicit flush, that means dropping whatever the buffer was holding, on a path that fires precisely when the network is already unhealthy.

The codebase has machinery for this. `spawn_fatal` (`cmd/ethrex/initializers.rs:243`) exists so a subsystem can declare the node dead: it logs loudly, records the cause, cancels the node, and `main` exits non-zero after the shutdown sequence runs. Its own doc comment describes exactly this situation.

The catch is layering — `spawn_fatal` is in `cmd/ethrex`, and `crates/networking/p2p` can't depend upward on the binary, so it isn't callable from here. That makes this a design question rather than a one-line swap:

- surface the fatal condition to the caller (the manager task could return it, or set a flag the node loop polls) and let `cmd/ethrex` invoke its existing path, or
- hand `SyncManager` a cancellation token / shutdown handle at construction and trigger that instead of exiting.

Either keeps the fatal semantics you want while going through the flush. There are already a few `process::exit` calls under `crates/networking/p2p` (`snap_sync.rs` has three), so this isn't unprecedented — but those predate this PR and adding a fourth on a newly-reachable path is worth pausing on, especially since exit code 2 here is otherwise undocumented.

crates/networking/p2p/sync_manager.rs
```diff
@@ -18,9 +18,106 @@ use tracing::{debug, error, info, warn};
...
+                if self.consecutive_timeouts >= MAX_CONSECUTIVE_REQUEST_TIMEOUTS {
+                    return SyncRetryAction::Fatal;
+                }
+                let sleep_for = self.backoff;
+                self.backoff = next_backoff(self.backoff);
+                SyncRetryAction::Continue {
+                    sleep: Some(sleep_for),
+                }
+            }
+            SyncCycleOutcome::RecoverableOther => {
```
ElFantasma, line 80: `RecoverableOther` backs off but never increments `consecutive_timeouts`, and only `Success` resets the backoff — so a persistent non-timeout recoverable error retries forever at the 30s cap.

That's a deliberate choice and the comment says so, and it matches #6853's scope. Worth being explicit in the PR description though, because the stated problem is "could retry immediately forever": this fixes *immediately* for every recoverable error but only fixes *forever* for `RequestTimeout`. A node stuck on a persistent `RecoverableOther` still loops indefinitely, just at 30s intervals rather than hot.

Bounded resource use, so not a blocker. But if the intent is that any persistent recoverable failure eventually gives up, a separate cap on total consecutive failures (reset by `Success`, same as the timeout counter) would cover it without conflating the two error classes.

crates/networking/p2p/sync_manager.rs
NikhilSharmaWe: This was relevant. Switched to your second option:

- Keep the construction-time `CancellationToken` on `SyncManager`
- On the timeout cap, `cancel()` instead of `process::exit`

That lets L1 run `server_shutdown` and flush the store.

crates/networking/p2p/sync_manager.rs
NikhilSharmaWe: Updated the PR description to make this explicit:

- Backoff applies to **all** recoverable errors
- The consecutive-failure shutdown applies only to `RequestTimeout`
- A persistent `RecoverableOther` can still retry at the 30s cap

PR #7089, opened 2026-08-03
fix(l1): align SIGPARAM copy operands with CALLDATACOPY

edg-l:
EIP-8141 changed the `SIGPARAM` copy form's operand order so it matches `CALLDATACOPY`: the three operands below `signatureIndex` and `param` are now `memOffset`, `dataOffset`, `length` (ethereum/EIPs@4a9ad32c, "align SIGPARAM copy operand order with CALLDATACOPY"). We still popped them in the old `length`, `dataOffset`, `memOffset` order, so an ARBITRARY-signature verifier copying its own bytes read the wrong three words — consensus-visible, since `SIGPARAM 0x04` is the only way to reach those bytes from the EVM.

- `OpSigParamHandler` pops `[mem_offset, data_offset, length]`.
- Doc comment and `docs/eip-8141.md` state the operand order top-first, like `FRAMEDATACOPY`.
- `sigparam_copy_code` pushes in the new order, and a new test pins it with three distinct operands and a destination past the first word, so a reversed read lands the bytes elsewhere instead of coincidentally passing.

The two existing copy tests use `memOffset = 0`; they now exercise the new order too.

Note for the frame-tx satellite branches: `eip-7906`, `eip-8250` and `eip-8272` all touch `OpSigParamHandler` and will need this one-line rebase.


ilitteri (approved):

iovoid (approved):

crates/vm/levm/src/opcode_handlers/frame_tx.rs
```diff
@@ -457,10 +457,11 @@ impl OpcodeHandler for OpFrameParamHandler {
 /// SIGPARAM (0xB4) -- signature-scoped metadata and data copy (EIP-8141).
 /// Metadata (params 0x00-0x03): stack `[param, signatureIndex]` with
 /// `signatureIndex` on top; gas 2; returns one word (0x00 effective signer,
-/// 0x01 scheme, 0x02 msg, 0x03 len(signature)). Copy (param 0x04): stack
-/// `[memOffset, dataOffset, length, param, signatureIndex]` with `signatureIndex`
-/// on top; CALLDATACOPY gas; copies an ARBITRARY signature's raw bytes into
-/// memory (zero-filled past the end) and pushes nothing — any other scheme halts.
+/// 0x01 scheme, 0x02 msg, 0x03 len(signature)). Copy (param 0x04): takes
+/// `[signatureIndex, param, memOffset, dataOffset, length]` from the stack,
```
ElFantasma: nit: this list reads top-first (`signatureIndex` first, because it's on top), but the metadata sentence two lines up lists `[param, signatureIndex]` bottom-first and spells out "`signatureIndex` on top". Same operand, same position, opposite ends of the list — and the copy form is the one that never says which end is the top.

`FRAMEDATALOAD` at :266 has the same top-first shape as this line and does state it ("with offset on top (popped first)"), so borrowing that phrasing keeps it unambiguous:

```rust
/// 0x01 scheme, 0x02 msg, 0x03 len(signature)). Copy (param 0x04): takes
/// `[signatureIndex, param, memOffset, dataOffset, length]` with `signatureIndex`
/// on top (popped first), matching `CALLDATACOPY`'s operand order; CALLDATACOPY
```

Worth tightening precisely because misreading this list in the other direction is the bug being fixed here.

docs/eip-8141.md
```diff
@@ -316,8 +316,9 @@ signature-scoped metadata for `signatures[signatureIndex]`:
 | 0x02 | msg (0 when empty — the canonical sig-hash case — else the 32-byte digest) |
 | 0x03 | len(signature) |
 
-**Copy (param 0x04)** — Stack input `[memOffset, dataOffset, length, param,
-signatureIndex]` (signatureIndex on top); no stack output; gas cost matches
+**Copy (param 0x04)** — Stack input `[signatureIndex, param, memOffset, dataOffset,
+length]` (signatureIndex on top, then the `CALLDATACOPY` operand order); no stack
```
ElFantasma: nit: same notation clash as the Rust doc comment, and a bit sharper here — the metadata paragraph above says `[param, signatureIndex]` **(signatureIndex on top)** and this one says `[signatureIndex, param, memOffset, dataOffset, length]` **(signatureIndex on top)**. Identical parenthetical, opposite ends of the list, four lines apart; a reader can reasonably conclude the two forms pop the index from different ends.

Making the copy line say it's listed top-first (or reordering it to match the metadata line) resolves it either way.

ElFantasma (approved):
Two doc nits, both non-blocking.

PR #7091, opened 2026-08-03
fix(l1): drop peers on undecodable RLPx inbound frames

### LINKED ISSUE #7035: A peer sending an undecodable RLPx message is neither penalized nor discarded, and keeps getting selected
## Summary

A peer that sends an RLPx message we cannot decode is neither penalized nor discarded. Its peer-table entry survives with a live connection handle and an unchanged score, so peer selection keeps handing it out — while its inbound stream is already dead and it can never answer anything.

An undecodable message is grounds for dropping that peer. Today we neither drop it nor score it down.

## What happens

`RLPxCodec::decode` propagates a `Message::decode` failure to `FramedRead`:

https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/rlpx/connection/codec.rs#L232

`tokio-util` treats a decoder error as terminal — it sets `has_errored`, and the next poll returns `Poll::Ready(None)` (0.7.18, `src/codec/framed_impl.rs`, citing tokio-rs/tokio#3976). The stream is finished from that point on.

`spawn_listener` reacts to a finished stream by simply leaving its loop:

```rust
None => {
    tracing::trace!("Stream finished");
    break;
}
```

Nothing else happens. In particular:

- The connection actor stays alive, so `PeerData.connection` remains `Some(..)` and the peer still looks connected.
- No `record_failure`, no `record_critical_failure`, no `set_disposable` on this path — the peer's **score is exactly what it was**, so score-weighted selection in `do_get_random_peer` / `get_best_peer` still favours it.
- The teardown that calls `remove_peer` runs on actor stop, which a finished stream does not trigger.

So for the next ~30-40 seconds (`PING_INTERVAL` 10s × `MAX_MISSED_PONGS` 3) the peer is repeatedly selectable as a healthy, well-scored peer, and every request routed to it burns a full `PEER_REPLY_TIMEOUT` with no possible reply. Only then does the ping liveness check `ctx.stop()` it.

And when it is finally reclaimed, `remove_peer` just `swap_remove`s the entry:

https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/peer_table.rs#L570-L579

It does not mark the contact `disposable`, so discovery is free to re-add and re-dial the same peer, which sends the same undecodable message, and the cycle repeats.

The `// Skipping invalid data` comment describes an intent that isn't achievable here — the stream is already gone — and it hides the fact that no peer bookkeeping happens at all:

https://github.com/lambdaclass/ethrex/blob/main/crates/networking/p2p/rlpx/connection/server.rs#L880-L886

## Impact

A decode-level incompatibility becomes silent peer-table churn instead of a bad peer being dropped: requests are handed to a peer that cannot answer, time out, and the same peer can be re-dialed later and fail identically. Nothing surfaces beyond a single `debug!`.

This is pre-existing and not specific to any message type. It showed up while working on receipt decoding, where wire-format mismatches were expensive for exactly this reason — the peer kept being retried rather than being discarded.

## Suggested fix

Treat a codec/decode error as what it is: this peer is unusable to us. On that path:

1. Record a failure against the peer (a critical one is defensible — it sent us something we cannot parse), so score-based selection stops preferring it immediately rather than after a ping timeout.
2. Mark the contact `disposable` so it is not re-dialed straight after being removed.
3. Stop the connection actor deliberately with a distinct disconnect reason, so teardown/`remove_peer` runs promptly instead of waiting on missed pongs, and the drop is attributable in `ethrex_p2p_disconnections` rather than landing in the generic bucket.

The `// Skipping invalid data` comment should go with the change, since skipping is not what we want here.

Worth noting for whoever picks this up: an undecodable message does not always mean a malicious peer — it can equally mean our decoder is behind on a wire format. The handling is the same either way (that peer is no use to us), but penalizing it makes the incompatibility visible in metrics/logs instead of silent, which is the more useful outcome of the two.

## Test

Feed a connection a frame with a valid header and an undecodable body, then assert the peer is no longer returned by peer selection (and its score dropped), rather than remaining selectable until the ping timeout.

NikhilSharmaWe:
Closes #7035

## Summary

A peer that sends an undecodable RLPx message kills the inbound `Framed` stream, but we only logged and skipped the error. The connection actor stayed alive, so peer selection kept handing that peer out until ping timeout (~30s), and every request burned a full reply timeout.

This change:

- notifies the connection actor on inbound `Framed`/codec `Err` via `InboundStreamFailed` (removes the fake "Skipping invalid data" path)
- stops the actor immediately with `ProtocolError` (decode/frame/crypto) or `NetworkError` (IO), so `remove_peer` runs and the peer is no longer selectable
- sends a wire Disconnect on `ProtocolError` when outbound may still work

## Test plan

- [ ] `cargo test -p ethrex-p2p --lib inbound_`
- [ ] `cargo clippy -p ethrex-p2p --all-targets -- -D warnings`

### DIFF
--- a/crates/networking/p2p/rlpx/connection/server.rs
+++ b/crates/networking/p2p/rlpx/connection/server.rs
@@ -130,6 +130,14 @@ pub trait PeerConnectionServerProtocol: Send + Sync {
         announcement: NewPooledTransactionHashes,
         hashes: Vec<H256>,
     ) -> Result<(), ActorError>;
+    /// Inbound RLPx stream failed (codec/decode/IO). The connection must be dropped:
+    /// after a `Framed` decoder error the read half is finished and the peer can no
+    /// longer answer requests.
+    fn inbound_stream_failed(
+        &self,
+        reason: DisconnectReason,
+        detail: String,
+    ) -> Result<(), ActorError>;
 }
 
 #[cfg(feature = "l2")]
@@ -358,6 +366,13 @@ pub struct PeerConnectionServer {
 impl PeerConnectionServer {
     #[started]
     async fn started(&mut self, ctx: &Context<Self>) {
+        // Unit tests may start the actor already in `Established` to exercise teardown
+        // without a full RLPx handshake. Production always starts as Initiator/Receiver.
+        #[cfg(test)]
+        if matches!(self.state, ConnectionState::Established(_)) {
+            return;
+        }
+
         // Set a default eth version that we can update after we negotiate peer capabilities
         // This eth version will only be used to encode & decode the initial `Hello` messages.
         let eth_version = Arc::new(RwLock::new(EthCapVersion::default()));
@@ -718,6 +733,40 @@ impl PeerConnectionServer {
         }
     }
 
+    #[send_handler]
+    async fn handle_inbound_stream_failed(
+        &mut self,
+        msg: peer_connection_server_protocol::InboundStreamFailed,
+        ctx: &Context<Self>,
+    ) {
+        if let ConnectionState::Established(ref mut established_state) = self.state {
+            debug!(
+                peer=%established_state.node,
+                reason=?msg.reason,
+                detail=%msg.detail,
+                "Inbound RLPx stream failed, dropping peer",
+            );
+            // Preserve an earlier reason (e.g. peer-sent Disconnect) if one is already set.
+            // Use the stored reason for both metrics and any wire Disconnect so they agree.
+            let reason = *established_state
+                .disconnect_reason
+                .get_or_insert(msg.reason);
+            // Protocol breach: outbound may still be writable, so tell the peer why.
+            // Network/IO failures usually mean the socket is already gone.
+            if reason == DisconnectReason::ProtocolError {
+                send_disconnect_message(established_state, Some(reason)).await;
+            }
+            ctx.stop();
+        } else {
+            debug!(
+                reason=?msg.reason,
+                detail=%msg.detail,
+                "Inbound RLPx stream failed while connection was not Established",
+            );
+            ctx.stop();
+        }
+    }
+
     fn process_cast_error(
         state: &ConnectionState,
         result: Result<(), PeerConnectionError>,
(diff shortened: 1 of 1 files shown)


crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -358,6 +366,13 @@ pub struct PeerConnectionServer {
 impl PeerConnectionServer {
     #[started]
     async fn started(&mut self, ctx: &Context<Self>) {
+        // Unit tests may start the actor already in `Established` to exercise teardown
+        // without a full RLPx handshake. Production always starts as Initiator/Receiver.
+        #[cfg(test)]
```
ElFantasma: This puts test-only control flow into a production lifecycle hook: under `cfg(test)` an actor whose state is already `Established` skips `started()` entirely — no eth-version setup, none of the initialization below.

So the tests for this fix drive an actor that was never started the way a real one is. For a change about *when the connection actor stops*, that's the part you most want the tests to be faithful about: `handle_inbound_stream_failed` calls `ctx.stop()`, and whether stopping is correct depends on what `started()` spawned and registered.

The usual alternative is to construct the actor through a real (or in-memory) startup path in the test rather than branching in the hook — a test-only constructor, or a trait/param that lets the test supply a stub stream, so production code has no `cfg(test)` in it. If the seam has to stay, it's worth a comment saying which tests need it and why, because as written the next person can't tell whether it's load-bearing or leftover scaffolding.

crates/networking/p2p/rlpx/connection/server.rs
```diff
@@ -1083,6 +1153,20 @@ fn match_disconnect_reason(error: &PeerConnectionError) -> Option<DisconnectReas
...
+/// dropping the peer. Used for the established-connection read-listener path;
+/// distinct from [`match_disconnect_reason`], which is for handshake/`connection_failed`.
+fn disconnect_reason_for_inbound_error(error: &PeerConnectionError) -> DisconnectReason {
+    match error {
+        PeerConnectionError::RLPDecodeError(_)
+        | PeerConnectionError::InvalidMessageFrame(_)
+        | PeerConnectionError::CryptographyError(_)
+        | PeerConnectionError::InvalidMessageLength => DisconnectReason::ProtocolError,
+        PeerConnectionError::IoError(_) => DisconnectReason::NetworkError,
+        _ => DisconnectReason::NetworkError,
```
ElFantasma: `IoError` and the `_` arm both produce `NetworkError`, so the explicit arm is redundant — but the catch-all is the part worth reconsidering.

The direction it defaults matters here. `ProtocolError` means "the peer misbehaved", and per `handle_inbound_stream_failed` it's the only reason that sends a wire Disconnect. `NetworkError` is the benign reading. So a `PeerConnectionError` variant added later — including a decode- or frame-shaped one, which is exactly the class this function exists to catch — silently classifies as benign and the peer is dropped without being told why.

Dropping the `_` and matching exhaustively makes the compiler force that decision at the point a variant is added, which is the same fail-closed argument as the `all_datadir_suffixes` list on #7080. If some variants genuinely can't reach the inbound path, an explicit arm listing them with `unreachable!()` or a `NetworkError` default plus a comment documents the reasoning instead of hiding it.

crates/networking/p2p/rlpx/connection/server.rs
NikhilSharmaWe: Removed the `cfg(test)` early-return in `started()`. Reason bookkeeping is covered via `apply_inbound_stream_failure`, and peer removal is covered by a regression that runs a real RLPx handshake before sending `InboundStreamFailed`.

crates/networking/p2p/rlpx/connection/server.rs
NikhilSharmaWe: Dropped the redundant `IoError` arm and the `_` catch-all. The match is now exhaustive: decode/frame/crypto failures map to `ProtocolError`, and everything else (including variants not expected on this path) maps to `NetworkError` explicitly.

NikhilSharmaWe:
cc @ilitteri

PR #7093, opened 2026-08-03
fix(l1): size the RocksDB block cache from available memory

### LINKED ISSUE #7092: OOM while syncing glamsterdam-devnet-7: memory grows to host limit (~14.6GB) within 40 minutes, container killed
## Description

While syncing glamsterdam-devnet-7 (checkpoint-synced CL driving ethrex via the engine API), ethrex's memory usage grows monotonically until it hits the host limit and the container is OOM-killed. In ethpandaops' syncoor CI, **all 6 checkpoint-sync test pairings crashed** the same way on 16.8GB runners:

| Pairing | Crash block | Last sampled memory | Duration |
|---|---|---|---|
| ethrex + teku | 45,056 | 12.1 GB | 46 min |
| ethrex + nimbus | 82,944 | 15.0 GB | 69 min |
| ethrex + lighthouse | 81,920 | 14.6 GB | 61 min |
| ethrex + grandine | 81,920 | 14.8 GB | 56 min |
| ethrex + prysm | 81,920 | 14.8 GB | 74 min |
| ethrex + lodestar | 78,848 | 13.7 GB | 64 min |

Memory trajectory of the ethrex+lighthouse run (30s sampling via docker stats; runner: 16.8 GB RAM, 8-core Xeon Platinum 8280):

```
t+  0min block=      0 mem=  0.2GB disk= 0.0GB cpu=  2%
t+  7min block=  16384 mem=  4.4GB disk= 1.8GB cpu=145%
t+ 15min block=  38912 mem=  7.1GB disk= 1.9GB cpu=120%
t+ 22min block=  40960 mem=  8.8GB disk= 2.4GB cpu=104%
t+ 30min block=  43008 mem= 11.9GB disk= 2.5GB cpu=126%
t+ 38min block=  44032 mem= 14.2GB disk= 3.4GB cpu=102%
t+ 45min block=  52224 mem= 14.3GB disk= 4.0GB cpu=149%
t+ 53min block=  71680 mem= 14.5GB disk= 5.5GB cpu=100%
t+ 58min block=  81920 mem= 14.6GB disk= 5.6GB cpu=134%   ← container killed shortly after
```

In the genesis-sync (non-checkpoint) test directory the picture is similar: 2 crashes, 2 × 9h50m timeouts, and a single success that took 5.6h peaking at 13.4 GB — so the sync occasionally survives on a 16GB box but is right at the edge.

The chain being synced is ~130k blocks with a 300M gas limit and blobs (glamsterdam devnet), so per-block state is heavy — but memory growing linearly with import progress until the host limit suggests something (state layers? caches?) is accumulating without bound rather than a fixed working set.

Full test reports and per-30s metrics: https://syncoor.glamsterdam-devnet-7.ethpandaops.io (e.g. test `sync-test-1785509155155920039` under `cl-full-checkpoint-sync`).

## Steps to reproduce (syncoor + kurtosis)

Prereqs: docker, [kurtosis](https://docs.kurtosis.com/install) ≥ 1.20.0, Go ≥ 1.26, a host with ≤16GB available to containers reproduces the crash (more RAM will just plateau higher).

```bash
git clone https://github.com/ethpandaops/syncoor && cd syncoor
go run ./cmd/syncoor sync \
  --network glamsterdam-devnet-7 \
  --el-client ethrex --cl-client lighthouse \
  --el-image ethpandaops/ethrex:glamsterdam-devnet-7 \
  --cl-image ethpandaops/lighthouse:glamsterdam-devnet-7 \
  --check-interval 30s --run-timeout 120m --report-dir ./reports
```

Watch `docker stats` on the `el-1-ethrex-lighthouse` container — memory climbs several hundred MB/min during import. On OOM the syncoor run exits with "execution client container crashed" and the generated report contains the full memory timeline.

## Versions

- ethrex: `ethrex/v20.0.0-glamsterdam-devnet-7-aac6736867152879c42174432a1febdae342e03d/x86_64-unknown-linux-gnu/rustc-v1.91.1` (image `ethpandaops/ethrex:glamsterdam-devnet-7`)
- Network: glamsterdam-devnet-7 (~130k blocks, 300M gas limit, blobs enabled)
- CL: lighthouse v8.2.0 (also reproduced with teku, nimbus, grandine, prysm, lodestar)

edg-l:
## Motivation

`DEFAULT_ROCKSDB_BLOCK_CACHE_SIZE_BYTES` was a flat 12 GiB regardless of the machine. With `cache_index_and_filter_blocks` enabled that constant is the effective ceiling on RocksDB's resident memory, so on a 16 GiB host we were telling RocksDB it could claim **71% of the box** — leaving nothing for the in-memory trie-layer backlog, block execution, the mempool, peer buffers and allocator slack.

This showed up on ethpandaops' syncoor runners for glamsterdam-devnet-7 (16.8 GB, 8 cores), where ethrex is OOM-killed during sync. The dominant term there is a separate unbounded-trie-layer bug, but even with that fixed a 12 GiB cache ceiling on a 16.8 GB host has no headroom: the one genesis-sync run that survived peaked at 13.4 GB, right at the edge.

The problem is worse in containers than the number suggests: `/proc/meminfo` reports the **host's** memory, not the container's limit, so a 4 GiB-limited container also saw "12 GiB is fine".

## Description

Derive the default from the memory the process may actually use instead of hardcoding it:

- `MAX_ROCKSDB_BLOCK_CACHE_SIZE_BYTES` (12 GiB) is now documented as the **ceiling**, not the default. Its value and its tuning rationale are unchanged.
- `default_rocksdb_block_cache_size()` returns `ROCKSDB_BLOCK_CACHE_MEMORY_PERCENT` (40%) of the detected limit, clamped to `MIN_ROCKSDB_BLOCK_CACHE_SIZE_BYTES` (512 MiB) ..= 12 GiB.
- The detected limit is `min(physical memory, cgroup memory limit)` — `MemTotal` from `/proc/meminfo` against `memory.max` (cgroup v2) / `memory.limit_in_bytes` (v1) — so a container is sized against its own limit rather than the machine it lands on. cgroup v1's "unlimited" sentinel is discarded by the `min`; v2's literal `max` fails to parse and is skipped.
- Detection failure (non-Linux, unreadable `/proc`) falls back to the 12 GiB ceiling, so **no host regresses relative to today**.

Effect: ≥30 GiB hosts keep 12 GiB (unchanged). A 16 GiB host gets ~6.4 GiB and keeps >half the machine for the node. That is below the ~8 GiB filter-thrash floor the ceiling's docs describe, which is the deliberate tradeoff — reduced throughput on a small host beats being OOM-killed on it.

The blast radius is deliberately small: with a 40% factor clamped to 12 GiB, the default only moves below ~30 GiB of detected memory, and `docs/getting-started/hardware_requirements.md` puts the RAM minimum at 32 GB for every network — so **any spec-compliant host sees no change**. Behaviour changes only for under-spec machines and containers, which is the population this PR exists for.

Per review, the derivation is no longer invisible: startup logs the detected limit, which detector won (physical memory, cgroup limit, or detection failure falling back to the ceiling), and the resolved cache size, so an operator debugging memory can tell whether a cgroup limit was seen and whether the clamp fired without reproducing the environment.

No new dependencies; the detection is ~25 lines of `std::fs`.

### CLI

`--rocksdb.block-cache-size` becomes `Option<usize>` and resolves at store construction. Explicitly setting it (or `ETHREX_ROCKSDB_BLOCK_CACHE_SIZE`) still overrides in either direction.

It is deliberately **not** `default_value_t`: clap would then print the resolving machine's byte count into `--help`, and CI diffs `--help` against `docs/CLI.md`, so the check would pass or fail depending on the runner's RAM. The verbose `long_help` is dropped in favor of one sentence; the sweep rationale now lives on the constants.

### Known limits (kept out of scope)

- **Detection is `/proc`-based**, so the 40% default applies on Linux only; elsewhere (macOS/Windows release binaries included) the default stays at the flat 12 GiB ceiling. The help text, `docs/CLI.md` and the CHANGELOG now say so explicitly.
- **Each `Store` gets its own cache.** The L2 committer opens additional checkpoint stores via `Store::new`, each of which resolves its own 40% default, and the CLI override is not plumbed to them (pre-existing on both counts — they previously each got a flat 12 GiB, so this PR strictly improves the absolute numbers). Sharing one cache across stores, and plumbing the flag into `init_native_rollup_l2` / `ethrex import`, is follow-up work.

## Testing

`test/tests/storage/rocksdb_block_cache_tests.rs` covers the pure clamp via `rocksdb_block_cache_size_for(Option<usize>)`, so it is machine-independent:

- undetectable limit → the ceiling (the no-regression path)
- 32/64/1024 GiB → clamped to the ceiling
- 16 GiB → exactly 40%, below the ceiling, >half the host left for the node
- 1 GiB and 0 → the 512 MiB floor
- the resolved default on the running machine lands inside the clamp

Also run locally: `cargo clippy --workspace --all-targets` clean, `cargo fmt --all --check` clean, storage suite 33 passed, and `--help` verified byte-identical to `docs/CLI.md` under CI's own `diff -ubB` (modulo the `$HOME`-dependent datadir default line, which CI generates as `/home/runner`).

Fixes https://github.com/lambdaclass/ethrex/issues/7092


crates/storage/store.rs
```diff
@@ -84,7 +84,131 @@ pub const BATCH_COMMIT_THRESHOLD: usize = 4;
...
+/// Default RocksDB shared block cache size, derived from the memory this process
+/// is actually allowed to use.
+///
+/// A fixed default cannot serve both a 64 GiB validator and a 16 GiB CI runner: the
+/// ceiling alone is 71% of a 16 GiB host, which leaves no headroom for the rest of
+/// the node. Sizes to [`ROCKSDB_BLOCK_CACHE_MEMORY_PERCENT`] of the detected limit,
+/// clamped to [`MIN_ROCKSDB_BLOCK_CACHE_SIZE_BYTES`] ..=
+/// [`MAX_ROCKSDB_BLOCK_CACHE_SIZE_BYTES`]. Falls back to the ceiling when the limit
+/// cannot be detected, preserving the previous behavior.
+pub fn default_rocksdb_block_cache_size() -> usize {
```
ElFantasma, line 133: Nothing logs the detected limit or the size this returns, and that is the one thing the PR'''s own motivation argues for.

The old value was a constant: wrong on small hosts, but knowable from the source. This replaces it with a number derived from `/proc/meminfo`, cgroup v2 `memory.max`, or cgroup v1 `memory.limit_in_bytes`, whichever is smallest — and `None` from all of them silently restores the old 12 GiB. So the cache size is now environment-dependent and invisible, on a code path that exists because a mis-sized cache was only discovered via an OOM kill.

Concretely, an operator debugging memory today cannot answer: was a cgroup limit detected at all, or did we fall back to physical memory? Did detection fail entirely and hand back 12 GiB? Was the result clamped? The clamp matters because it is silent in both directions — below ~1.3 GiB of detected limit, `MIN_ROCKSDB_BLOCK_CACHE_SIZE_BYTES` (512 MiB) wins and the cache is more than 40% of the limit.

One line at startup would close it:

```
info!(detected_limit_bytes = ?limit, source = %source, cache_bytes = size, "sized RocksDB block cache");
```

Worth noting the blast radius is small, which is a point in the PR'''s favour and not stated: with a 40% factor clamped to 12 GiB, the default only moves below ~30 GiB of detected memory. `docs/getting-started/hardware_requirements.md` puts the RAM minimum at 32 GB for every network, so on any spec-compliant host this is a no-op (32 GB -> 12.8 GiB -> clamped to 12 GiB, unchanged). It changes behaviour only for under-spec machines and containers — which is exactly the intent, and worth saying explicitly so reviewers know a spec-compliant deployment sees no change.

ilitteri (approved):

PR #3293, opened 2025-06-24
fix(l1, levm): fix remaining blockchain test

JereSalo:
**Motivation**

- Fix last blockchain test, it was failing for both LEVM and REVM

**Description**

- The test destroyed an account that had non-empty storage and then re-created it with CREATE2. When getting the AccountUpdates we just said that the account had no added storage (which is true) but we don't have a way to directly communicate that the account was destroyed and then created again, so even though it exists its old storage should be cleared.
  - For this I implemented an ugly solution. For both LEVM and REVM in get_account_updates if I see that an account was Destroyed but now exists what I'll do is I'll push 2 Account Updates, one that removes the account and another one with the new state of the account, so that the whole account is removed (and therefore, its storage) and then we write to the database the new state of the account with it's new storage. I think a cleaner solution would be to have an attribute `removed_storage` (or similar) in `AccountUpdate` that will tell the client to remove the storage of the existing account without removing all the account and then we don't have to do messy things like the one I implemented. The downside that I see on this new approach is that we'll have an attribute that we'll hardly ever use, because it's an edge case.
  - Then, for LEVM I had to implement a `destroyed_accounts` in `GeneralizedDatabase` so that in `get_state_transitions()` we can check which accounts were destroyed and now exist so that we do the procedure that I described above. This and many other things would be way nicer if we used in LEVM our own Account struct instead of reusing the one in Ethrex. I'm seriously considering making that change because it seems worth doing so, there are other reasons to pull the trigger on that.

Questions:
1. Should we add a `removed_storage` to `AccountUpdate` instead? Or this way of implementing it (removing account and then writing it) is good enough? Created #3321 
2. Should we use our own Account type in LEVM so that we don't rely on external HashSets and HashMaps for some things? For this I opened #3298 

Closes #3283

PR #7094, opened 2026-08-03
test(l1): bump Amsterdam fixtures to glamsterdam-devnet-8

edg-l:
Bumps the Amsterdam fixtures from `tests-glamsterdam-devnet@v7.2.0` to `v8.1.0` and makes the client pass them.

The bundle landed as v8.0.0 and then v8.1.0, which revised the EIP-8038 gas values (execution-specs [#3293](https://github.com/ethereum/execution-specs/pull/3293)). Both pins are at v8.1.0; `devnets/glamsterdam/8` and the v8.1.0 tag are the same commit, `864fa03a`, so this is the head of the devnet-8 spec.

### Gas schedule

devnet-8 moved two of the repricing EIPs, and v8.1.0 revised the EIP-8038 numbers again (`COLD_STORAGE_ACCESS` 2100, `ACCOUNT_WRITE` 9000, with `CALL_VALUE`, `CREATE_ACCESS` and the storage-clear refund derived from them):

- **EIP-2780**: a value transfer is now a single `TX_VALUE_COST` of 6000 covering both the recipient balance write and the EIP-7708 transfer log, replacing the 4244 + 1756 split. A creation carrying value pays no value charge at all, since `CREATE_ACCESS` already covers the balance write — create-with-value drops from 24756 to 23000, and the calldata-floor anchor follows because it shares `recipient_regular_gas`.
- **EIP-8038**: access-list entries cost the cold access minus `WARM_ACCESS`, so prepaying an entry is gas neutral with the cold access it replaces, and EIP-2930's extra 100 discount is not restored. At v8.1.0 the two entry kinds diverge, since they derive from different cold costs: an address is 3000 − 100 = 2900, a storage key is 2100 − 100 = 2000. (Under v8.0.0 both were 2900, because `COLD_STORAGE_ACCESS` was still 3000.)

### Engine API

`engine_newPayloadV1..V4` accepted payloads carrying `slotNumber` or `blockAccessList`, which are defined only from V5. A pre-fork block with a slot number was executed instead of being rejected with `-32602`. V3 previously doubled as both the pre-Amsterdam shape and the base of the Amsterdam one, so the shared Cancun-onwards field checks are split out and V5 keeps requiring what the earlier versions now forbid.

### Fixtures and CI

- `.fixtures_url_amsterdam` and `.github/config/hive/amsterdam.yaml` (`fixtures` + `eels_commit`) are both at v8.1.0, and `AMSTERDAM_FIXTURES_BRANCH` at `devnets/glamsterdam/8`. The two release pins have to move together: the ef-test side went to v8.1.0 while hive stayed on v8.0.0, so hive graded a v8.1.0 client against v8.0.0 expectations and failed 2168 cases in Consume Engine Amsterdam — concentrated in `eip8037_state_creation_gas_cost_increase`, `eip2780`, `eip7708` and `eip7928`, exactly the gas-sensitive fixtures. `Integration Test` failed only as the aggregate gate reporting that job. A comment in the yaml now records the coupling for the next bump.
- The bundle is overlaid onto an existing `vectors/eest`, so the subtrees are cleared before extracting. Fixtures renamed between devnet releases (EIP-8037 regular gas → execution gas) otherwise survived as stale leftovers and ran against the client.
- Also extract `for_bpo2toamsterdamattime15k`. It holds the fork-activation fixtures — a header field appearing mid-chain, the gas schedule switching between blocks — and exists only in this bundle, so omitting it left the whole fork boundary untested. The hive limits become `fork_.*Amsterdam` to sweep them too, and the daily report keys the Amsterdam bundle off an explicit flag rather than a substring of the limit, which no longer contains a literal fork name.
- Fixture downloads are idempotent: the artifact targets are order-only on the vectors dir, since building it runs `clean-vectors` and would otherwise wipe the tarball between the curl and the extract. The state suite keeps its tarball, hence the new `.gitignore` entry. This is what recovers the Amsterdam state fixtures noted below — previously a second clean destroyed them before the tests ran.
- The state suite's `run-evm-ef-tests-ci` and `test-levm` now depend on `amsterdam-vectors`, so they cover Amsterdam without a separate download step.
- The stateless Amsterdam+ skip is retargeted at the real cause: `tests-zkevm@v0.6.2` is filled against devnet-7 and its gas expectations no longer match a devnet-8 client. Documented in `docs/known_issues.md`.

New tests pin the EIP-8037 two-dimensional inclusion gate and the per-tx execution-gas cap, including the cap on `max(intrinsic, calldata floor)`.

### Verification

- Engine ef-tests after a full `clean-vectors`: 10993 tests / 75478 fixtures, 0 failed.
- `cargo fmt --check` and `cargo clippy --all-targets -D warnings` clean.

### Notes for review

- The ef-test comparison shows 56628 → 72432 fixtures, both at 100%. That +15804 is not devnet-8 churn: on `main` the state suite's `Run tests` step re-triggers `clean-evm-ef-tests`, because the `$(VECTORS_DIR)` rule deletes the tarball and make then rebuilds it, wiping `vectors/` and with it the Amsterdam fixtures the `Download EF Tests` step had just extracted. `main` was running zero Amsterdam state fixtures. The idempotency change fixes that, so they now actually execute.
- The `EF Tests (no_std crypto)` job pulls no Amsterdam fixtures at all, so the no_std crypto path has zero Amsterdam coverage. Pre-existing on `main`, left alone here to keep this PR scoped; worth its own change.
- The roadmap lists EIP-8070 (eth/72) as in review rather than done, since it lands with #6776.


ilitteri (approved):

edg-l:
On the Codex point about `for_bpo2toamsterdamattime15k` reaching the stateless path: it does not, for two independent reasons.

The stateless harness reads a different corpus. `tests/all.rs` sets `TEST_FOLDER = "vectors_zkevm/"` under the `stateless` feature, while the `AMSTERDAM_SUBTREES` overlay this PR adds extracts into `$(SPECTEST_VECTORS_DIR)` = `vectors/eest`. Nothing this PR overlays is visible to the stateless run.

And the zkevm extraction is narrowed to one subtree: `$(ZKEVM_VECTORS_DIR)` untars only `fixtures/blockchain_tests/for_amsterdam` (Makefile:82). The v0.6.2 bundle does ship a `for_bpo2toamsterdamattime15k` directory, but it is never extracted, so it is not in `vectors_zkevm/` either.

The ordering observation itself is correct and intentional: `BPO2ToAmsterdamAtTime15k` sits before `Amsterdam`, the same way `CancunToPragueAtTime15k` sits before `Prague`. So `test.network >= Fork::Amsterdam` means "fixtures whose whole chain is Amsterdam", which is the right predicate for a corpus that only contains `for_amsterdam`. It would become too narrow only if the zkevm extraction were widened to match `AMSTERDAM_SUBTREES`, which would also need the devnet-8 fill to exist first.

crates/vm/levm/src/gas_cost.rs
```diff
@@ -278,7 +284,7 @@ pub fn access_list_address_cost(fork: Fork) -> u64 {
     }
 }
 
-/// Per-storage-key access-list cost. EIP-8038 raises this from 1900 to 3000 at Amsterdam.
+/// Per-storage-key access-list cost. EIP-8038 raises this from 1900 to 2900 at Amsterdam.
```
ElFantasma, line 291: This says 2900, but the constant it documents evaluates to **2000**.

```
ACCESS_LIST_STORAGE_KEY_COST_AMSTERDAM = COLD_STORAGE_ACCESS_AMSTERDAM - WARM_ADDRESS_ACCESS_COST
                                       = 2100 - 100
                                       = 2000
```

The 2900 figure is the *address* case, where `COLD_ACCOUNT_ACCESS_AMSTERDAM` is 3000 — that line above is correct. Both comments were moved from 3000 to 2900 together, but the storage-key one also needed recomputing because this PR drops `COLD_STORAGE_ACCESS_AMSTERDAM` from 3000 to 2100. So the value is right and the formula is applied consistently; only the number in the prose is stale.

Worth fixing rather than leaving, because on a gas constant the comment is what a cross-client check reads. Someone diffing ethrex against the EIP or another client sees 2900, finds 2000 in the fixtures, and has to work out which of the two is lying. `1900 to 2000` keeps the sentence shape.

Same paragraph is also the one place `COLD_STORAGE_ACCESS_AMSTERDAM: 3000 -> 2100` shows up — the PR description covers the EIP-2780 and access-list moves but not that one, nor `ACCOUNT_WRITE` 8000 -> 9000, `CALL_VALUE` -> 11300 or `CREATE_ACCESS` -> 12000. All plausibly EIP-8038 repricing, but a reader reconciling against devnet-8 would have an easier time if the description listed them.

crates/vm/levm/src/gas_cost.rs
```diff
@@ -210,23 +210,29 @@ pub const BLOB_GAS_PER_BLOB: u64 = 131072;
 pub const ACCESS_LIST_STORAGE_KEY_COST: u64 = 1900;
 pub const ACCESS_LIST_ADDRESS_COST: u64 = 2400;
 
-// ===== EIP-8038 Amsterdam values (merged EIPs#11802) =====
+// ===== EIP-8038 Amsterdam values =====
 pub const COLD_ACCOUNT_ACCESS_AMSTERDAM: u64 = 3000;
-pub const COLD_STORAGE_ACCESS_AMSTERDAM: u64 = 3000;
-pub const ACCESS_LIST_ADDRESS_COST_AMSTERDAM: u64 = 3000;
-pub const ACCESS_LIST_STORAGE_KEY_COST_AMSTERDAM: u64 = 3000;
+pub const COLD_STORAGE_ACCESS_AMSTERDAM: u64 = 2100;
```
ElFantasma, line 215: The title is `test(l1): bump Amsterdam fixtures to glamsterdam-devnet-8`, but this hunk changes the Amsterdam gas schedule and `engine/payload.rs` changes which payload versions accept `slotNumber`/`blockAccessList`. Only 2 of the 30 files are fixtures.

Flagging the type specifically because it outlives the PR: on a squash merge the title becomes the commit subject in `main`, so the durable record of a consensus gas change reads `test(l1)`. Anyone later scanning history or generating release notes for behaviour changes filters it out. A stale *description* is ephemeral and I would not raise it; a stale *type* is not.

`fix(l1)` looks right — the body frames these as conformance corrections against devnet-8, and the engine half fixes payloads that were wrongly accepted.

PR #7095, opened 2026-08-03
perf(l1): cut the cost of cold contract-code access

edg-l:
**Motivation**

ethrex is red on `COLD_ACCOUNT_CODE_ACCESS` in the [EIP-8038 repricing fit](https://misilva73.github.io/eip-8038-repricing/): 7680 gas against that page's 3000 goal, second worst of six clients. Every other cold-access parameter is green.

The binding fixture is EEST `test_account_access` with `AccountMode.EXISTING_CONTRACT_DIFF_MAX`, where every target address holds a unique max-size runtime that is 24544 of 24576 bytes `JUMPDEST`. Splitting the measurement by account mode isolates the cost: a shared 24KB code runs as fast as a 1-byte code, so the code cache hits by code hash, and `EXTCODEHASH`/`BALANCE` stay fast, so the account path is fine (ethrex is the fastest of the six on `NON_EXISTING`). All of it is the cost of ingesting a newly seen bytecode, ~84 us per unique 24KB contract against ~17.6 for reth, ~18.5 besu, ~21.3 geth, ~25.3 nethermind.

**Description**

1. **Jump destinations are a bitmap, not an RLP list of `u32` offsets.** The list was ~4x the size of the code it described and was rebuilt on every code-cache miss. Per 24KB jumpdest-dense contract, the stored value goes from 97,906 B (3.98x code) to 27,654 B (1.13x), a load from 224 us to 0.48 us (17.4 us for a legacy value, rebuilt from the bytecode), and a `JUMP` validity check from 12.3 ns to 0.23 ns. No data rewrite is needed: the RLP item header tells a list from a byte string, so existing values still read. The store schema version is bumped with a no-op migration for the other direction, since an older binary cannot decode a bitmap and should warn rather than fail on the first code read. Building the bitmap is also cheaper than building the offset list, which matters for initcode analysis on every CREATE, 17.0 us against 49.8 us per 24KB of dense JUMPDESTs. The bits of a byte are accumulated in a register and stored once, so the scan loop carries neither a read-modify-write nor a bounds-check panic path.

2. **The code cache counts the bytecode it holds.** `Code::size()` counted `size_of::<Bytes>()` instead of the heap allocation, so the byte budget excluded the only thing that makes an entry large. The budget goes from 64 MiB of jump tables to 256 MiB of bytecode, and re-inserting a cached hash no longer inflates the counter.

3. **Bloom filter for `account_codes` and `account_code_metadata`, 4KB blocks for the former.** `account_codes` was the only exact-key point-lookup CF on the execution read path without a filter, and its 32KB blocks buy nothing when the SST value is just a blob reference. The metadata CF has no blob indirection, so it keeps 16KB; the bytecode CF also skips its last-level filter, whose lookups are almost always positive because the hash comes from an account that references it.

4. **`EXTCODESIZE` reads the code-length table.** It was loading the full bytecode to return `.len()`, while `ACCOUNT_CODE_METADATA` was plumbed through every layer and reached only from a mempool check. Witnesses stay complete: a metadata read records the code hash, so the bytecode still lands in the witness and stateless validation recomputes the length from it.

**Measured**

Stateful suite `3f6a0898955dff4f`, run `1785777240_7cd53c4f` (2026-08-03T17:14Z) against the run before it, refit through the same pipeline the repricing site uses (`benchmarkoor-fetch` 0.3.1 and `evm-gasfit` 0.3.0 on the site's own `fit.yaml`, scoped to ethrex):

| param | before | after |
| --- | --- | --- |
| `COLD_ACCOUNT_CODE_ACCESS` | 7736 (103.1 us) | **4652 (62.0 us)** |
| `COLD_ACCOUNT_CODE_WRITE` | 10415 (138.9 us) | **6355 (84.7 us)** |
| `COLD_ACCOUNT_NOCODE_ACCESS` | 1377 (18.4 us) | 1389 (18.5 us) |

Binding model n=11, R² 0.997, confidence intervals disjoint (98.8-105.7 us against 59.0-63.8 us). Refitting the pre-fix run gives 7736 where the site publishes 7680, so these are comparable to it. In raw wall time on the 300M `DIFF_MAX` fixture every code-loading opcode drops ~38% (`EXTCODESIZE` 68%, since (4) removes its bytecode read) while `BALANCE` and `EXTCODEHASH`, which never materialize the bytecode, sit flat at ~1940 and ~1950 ms across three runs.

The `COLD_ACCOUNT_ACCESS` goal is 3000 and needs both the CODE and NOCODE variants under it, so ethrex goes from 2.58x the goal to 1.55x and the goal stays unmet. Clearing it means 40.0 us per access at this fit's anchor, against the 62.0 measured here. What is left is the ~105 us blob read per unique contract, plus the 17.4 us legacy rebuild on this snapshot only. The blob read wants the BAL code prefetch issued as batched `multi_get`s rather than 22.9k point gets, which is a separate change.

**Notes**

- (3) is a write-time option: it applies to newly flushed or compacted SSTs, and benchmarkoor gives every instance a copy of a pinned data directory, so it will show nothing there until that snapshot is regenerated. (1), (2) and (4) land on the existing snapshot.
- (2) is a candidate to derive from the detected memory limit, like the block cache in #7093, once that lands.
- The measured run predates a01c66792, so it still carries the sparse-analysis regression that commit removes.
- That run also carries #7093's block-cache sizing, which is not part of this PR and is what moves the storage parameters in the same refit. The account-path controls above are flat, so it does not account for the code-access delta.
- Verified on `glamsterdam-devnet-7` with the same commits: 11616 EEST blockchain tests pass, plus clippy and the workspace and storage suites.


crates/vm/levm/src/db/gen_db.rs
```diff
@@ -492,29 +492,18 @@ impl GeneralizedDatabase {
...
-                    };
-
-                    code.len() as u64
-                };
-
-                let metadata = CodeMetadata {
-                    length: code_length,
+                let length = match self.codes.get(&code_hash) {
+                    Some(code) => code.len() as u64,
+                    None => self.store.get_code_metadata(code_hash)?.length,
```
ElFantasma, line 502: This is the change that makes `ACCOUNT_CODE_METADATA` an execution-path read, and there's a pre-existing wrinkle behind it worth knowing about before merge.

`Store::get_code_metadata` (`crates/storage/store.rs`, the `else` branch on a table miss) auto-migrates: it loads the full code, derives the length, and fires a **`tokio::task::spawn`** to write the metadata back. That code is on `main` already, so it isn't yours — but until now it was, in your words, "reached only from a mempool check". This PR puts it under `EXTCODESIZE`.

Two things change character as a result:

**Runtime context.** `tokio::task::spawn` panics if no reactor is running on the calling thread. A mempool check is comfortably inside the async runtime; block execution is not always — anything reached via `spawn_blocking` or a sync execution path would panic on a metadata miss rather than degrade. Worth confirming every caller of `EXTCODESIZE` is on a runtime thread.

**Miss volume.** A miss is the steady state on any node whose codes predate the metadata table, and the burst is worst on exactly the shape this PR targets: the `DIFF_MAX` fixture is unique max-size contracts, so a block of them is one spawned write task per contract, concurrent with execution's own writes. Fine once migrated, potentially a large unbounded burst on the first blocks after upgrade.

Neither is a reason to hold the perf work, and the fallback is correct. But since it's newly hot, a bounded or synchronous backfill — or just confirming the panic case is unreachable — seems worth doing in this PR rather than discovering it on a mainnet upgrade. A one-off backfill at startup would also make the fallback cold again.

ElFantasma (approved):

crates/vm/levm/src/db/gen_db.rs
edg-l: Good catch, the panic case was reachable: execution runs on rayon workers, which have no runtime, so a metadata miss aborted the node. Fixed in 6dd6ca857, the spawn is gated on `Handle::try_current()` and skipping it just costs one code read next time.

Added a regression test that is deliberately a plain `#[test]` rather than `#[tokio::test]`, since the absence of a runtime is the condition. Verified it fails without the guard.

That also bounds the burst you describe: on the `DIFF_MAX` shape the reads come from rayon workers, so no task is spawned at all. A startup backfill would still be worth having to make the fallback cold, but it is out of scope here.

ElFantasma (approved):

crates/common/types/account.rs
```diff
@@ -42,13 +42,19 @@ pub struct Code {
...
+    /// [`Code::is_valid_jumpdest`] reads a missing byte as "no jump destination".
+    //
+    // `Arc<[u8]>` so cloning `Code` (hot: every message-call resolves and clones
+    // the callee's code) is a refcount bump instead of deep-copying the bitmap.
     // Serializes via serde's `rc` feature (enabled workspace-wide).
-    // The valid addresses are 32-bit because, despite EIP-3860 restricting initcode size,
-    // this does not apply to previous forks. This is tested in the EEST tests, which would
-    // panic in debug mode.
-    pub jump_targets: Arc<[u32]>,
+    jumpdests: Arc<[u8]>,
```
iovoid, line 57: This changes the serialization of `Code` and therefore of `AccountUpdate`, breaking `store_account_updates_by_block_number`.

iovoid (changes requested):

crates/common/types/account.rs
edg-l: Right, thanks. Fixed in 9a4c1fd84 by dropping the jump destinations from the wire format entirely and recomputing them on deserialize: they are a pure function of the bytecode, so carrying them coupled `AccountUpdate`s stored format to how they happen to be represented. The format is now hash plus bytecode, so a future representation change cannot break it again, and the payload loses a byte per eight bytes of code. Added a test asserting the field set.

Note this still does not make rows written by an older binary readable, and the rollup store has no schema version to gate on. Since those rows are consumed at batch commit the window is small, but tell me if you want a version guard instead.

ilitteri (approved):

crates/storage/store.rs
```diff
@@ -5229,15 +5264,27 @@ pub fn receipt_key(block_hash: &BlockHash, index: u64) -> Vec<u8> {
...
 
+/// Decodes the JUMPDEST bitmap stored after the bytecode in an [`ACCOUNT_CODES`] value.
+///
+/// Values written before the bitmap representation hold an RLP *list* of `u32` offsets
+/// there instead. The RLP item header distinguishes a list from a byte string, so both
+/// forms are readable: for the older one the bitmap is rebuilt from the bytecode, which
+/// is cheaper than decoding the list.
+fn decode_jumpdests(code: &[u8], encoded: &[u8]) -> Result<Arc<[u8]>, StoreError> {
+    let (is_list, payload, _) = decode_rlp_item(encoded)?;
+    if is_list {
```
iovoid, line 5530: No migration path

crates/storage/migrations.rs
```diff
@@ -44,6 +44,19 @@ fn migration_for_version(version: u64) -> MigrationFn {
...
+/// `ACCOUNT_CODES` values carry their JUMPDEST positions as a bitmap rather than an RLP
+/// list of `u32` offsets. Both forms are readable, so a v3 database needs no rewriting
+/// and this migration only moves the version marker.
+///
+/// The bump exists for the other direction: a v3 binary cannot decode a bitmap, and
+/// `run_pending_migrations` warns that the database is ahead of the binary instead of
+/// letting it fail on the first code read.
+fn migrate_3_to_4(_backend: &dyn StorageBackend) -> Result<(), StoreError> {
+    Ok(())
+}
```
iovoid, line 58: Since there is no migration, the "recompute everything" fallback will be used on every cold read. This is likely going to cause a big performance hit.

crates/common/types/account.rs
iovoid: I would rather have a migration and not break things.

We could simply keep the field on CodeSerde and do the migration later if you want. For the encode path, transforming from a bitmap to a list should be easy.

iovoid (changes requested):

ilitteri (approved):

PR #7098, opened 2026-08-03
fix(l1,l2): make eth_getLogs topics optional and honour blockHash

jrchatruc:
Two spec deviations in `eth_getLogs`, both in `crates/networking/rpc/eth/logs.rs`.

## `topics` was mandatory

Every field of the filter object is optional in [`execution-apis/src/schemas/filter.yaml`](https://github.com/ethereum/execution-apis/blob/main/src/schemas/filter.yaml), and an absent `topics` means match any topic. `LogsFilter::parse` read it with `.ok_or_else(|| RpcErr::MissingParam("topics"))`, so a request omitting it got `-32000 Expected parameter: topics is missing`, while `fromBlock`, `toBlock` and `address` immediately above all tolerate absence via `.map(...).transpose()?`. Everything downstream already handled the empty case (`topics: topics_filters.unwrap_or_else(Vec::new)`, and the `if filter.topics.is_empty()` short-circuit in `fetch_logs_with_filter`), so this is the parse step being written in a different shape than its neighbours. `"topics": null` behaves as before.

geth agrees: `GetLogs` only rejects `len(crit.Topics) > maxTopics` and never requires the field.

## `blockHash` was silently ignored

[EIP-234](https://eips.ethereum.org/EIPS/eip-234) lets a filter name one block by hash instead of a range, and the schema models the two as mutually exclusive shapes. `LogsFilter` had no field for it, and since the parser reads keys by name and ignores the rest, `{"blockHash": "0x…"}` was not rejected but dropped, leaving `from_block`/`to_block` at their `latest` defaults. The request appeared to succeed and returned the head block's logs instead of the requested block's.

That failure mode matters precisely where the hash form is used. Naming a block by hash rather than by height exists so a reorg produces an error instead of substituting a different block at the same height; returning the head's logs defeats the guarantee and hands back wrong data with no signal. A chain-following consumer at the tip is the main caller of this shape.

So `block_hash: Option<H256>` is now a real field: parsed, `WrongParam` when malformed, and honoured in `fetch_logs_with_filter`, which resolves the hash to a block number and scans exactly that block, erroring with `BadParams` on an unknown hash. Combining it with `fromBlock` or `toBlock` is `BadParams`, matching geth's `errBlockHashWithRange` ("can't specify fromBlock/toBlock with blockHash").

`eth_newFilter` shares this parser but tracks a moving range of blocks, and the spec defines `blockHash` for `eth_getLogs` only, so `NewFilterRequest::parse` rejects the field rather than accepting something meaningless.

## Tests

Added to the `logs` module: an absent `topics` key, a fully empty filter object, `blockHash` alone, `blockHash` combined with either range field rejected, and a malformed `blockHash` rejected. The existing tests all passed `topics` explicitly and never passed a `blockHash`, which is why neither gap was caught.

`cargo test -p ethrex-rpc --lib` is green (103 tests), `cargo fmt --check` and `cargo clippy -p ethrex-rpc --lib` clean.

Draft because the `blockHash` path is only covered by parse-level tests here: it would be worth an integration test against a store with a known block, and a hive/spec-test run, before merging.


ilitteri (approved):

edg-l (approved):

crates/networking/rpc/eth/logs.rs
```diff
@@ -125,16 +153,30 @@ pub(crate) async fn fetch_logs_with_filter(
...
-        .to_block
-        .resolve_block_number(&storage)
-        .await?
-        .ok_or(RpcErr::WrongParam("toBlock".to_string()))?;
+    // A `blockHash` filter covers exactly that block; the range fields are not
+    // part of that form of the request (see LogsFilter::block_hash).
+    let (from, to) = match filter.block_hash {
+        Some(block_hash) => {
+            let block_number = storage
+                .get_block_number(block_hash)
```
ElFantasma: `get_block_number` resolves *any* stored block hash, not only canonical ones — `write_block_data` writes `BLOCK_NUMBERS` unconditionally alongside `HEADERS`/`BODIES`, so a side-chain block that was executed and stored, or one that a reorg since demoted, still has an entry.

The scan below then goes by number: `for block_num in from..=to { storage.get_block_header(block_num) }` resolves the **canonical** block at that height. So for a non-canonical hash this returns a different block's logs under the requested hash, with no error — which is the substitution the PR is fixing, surviving in exactly the case `blockHash` exists for:

> Naming a block by hash rather than by height exists so a reorg produces an error instead of substituting a different block at the same height.

Before this PR the hash was ignored and you got the head's logs. After it, you get the canonical block at the right *height*. Both are the wrong block; the second is harder to notice because the block number looks right.

geth avoids it by resolving the header by hash and never going through the number. The smallest fix here is to confirm canonicality after resolving, since the helpers already exist (`get_canonical_block_hash`, `is_canonical_sync`):

```rust
let block_number = storage.get_block_number(block_hash).await?
    .ok_or_else(|| RpcErr::BadParams(format!("Unknown block hash {block_hash:#x}")))?;
if storage.get_canonical_block_hash(block_number).await? != Some(block_hash) {
    return Err(RpcErr::BadParams(format!("Block {block_hash:#x} is not canonical")));
}
```

That also closes the narrower race where a reorg lands between the hash→number and number→header lookups.

Worth a test alongside the existing ones: store a block, reorg past it, and assert `blockHash` on the demoted hash errors rather than returning the replacement's logs.

crates/networking/rpc/eth/logs.rs
```diff
@@ -50,9 +50,14 @@ pub struct LogsFilter {
     /// Up to which block to stop retrieving logs.
     /// Will default to `latest` if not provided.
     pub to_block: BlockIdentifier,
+    /// Restricts the filter to a single block, by hash. The spec defines this as
+    /// an alternative to the `fromBlock`/`toBlock` range, so the two forms are
+    /// mutually exclusive and this one takes precedence when present.
+    pub block_hash: Option<H256>,
```
ElFantasma, line 59: nit: `block_hash` arrives without a doc comment while `from_block` and `to_block` above both have one, and the `/// Which topics to filter.` line on `topics` was dropped in the same hunk.

`block_hash` is the field that most needs one — its mutual exclusion with `fromBlock`/`toBlock` and its single-block semantics aren't guessable from the type. Something like "EIP-234: match exactly this block instead of a range; mutually exclusive with `from_block`/`to_block`" would carry the rule next to the field enforcing it.

crates/networking/rpc/eth/logs.rs
```diff
@@ -125,16 +153,30 @@ pub(crate) async fn fetch_logs_with_filter(
...
-        .ok_or(RpcErr::WrongParam("toBlock".to_string()))?;
+    // A `blockHash` filter covers exactly that block; the range fields are not
+    // part of that form of the request (see LogsFilter::block_hash).
+    let (from, to) = match filter.block_hash {
+        Some(block_hash) => {
+            let block_number = storage
+                .get_block_number(block_hash)
+                .await?
+                .ok_or_else(|| RpcErr::BadParams(format!("Unknown block hash {block_hash:#x}")))?;
+            (block_number, block_number)
```
iovoid: That returns the canonical block at the height of the block hash, which aren't necessarily the same.

iovoid (changes requested):

crates/networking/rpc/eth/logs.rs
jrchatruc: Addressed here [e3c81f4](https://github.com/lambdaclass/ethrex/pull/7098/commits/e3c81f418e27929ddc6b2ad2e14886a4428a4b14)

crates/networking/rpc/eth/logs.rs
jrchatruc: Addressed here [e3c81f4](https://github.com/lambdaclass/ethrex/pull/7098/commits/e3c81f418e27929ddc6b2ad2e14886a4428a4b14)

crates/networking/rpc/eth/filter.rs
```diff
@@ -59,6 +59,14 @@ pub struct PollableFilter {
 impl NewFilterRequest {
     pub fn parse(params: &Option<Vec<serde_json::Value>>) -> Result<Self, RpcErr> {
         let filter = LogsFilter::parse(params)?;
+        // A polled filter tracks a moving range of blocks, so a single-block
+        // `blockHash` filter is meaningless here. The spec only defines it for
```
iovoid: It _is_ defined by EIP-234.

iovoid (approved):

ElFantasma (approved):

PR #7099, opened 2026-08-04
perf(l1): batch and stream the BAL code prefetch

edg-l:
> Stacked on #7095 (`perf/cold-contract-code-access`), which adds the `ACCOUNT_CODES` decode helpers this builds on. Retarget to `main` once that merges.

**Motivation**

On a block whose accounts hold code the executor has not seen, the BAL warmer does not get ahead of execution. Measured on the benchmarkoor stateful suite by parsing the client's own per-block log (`warmer` against `exec`), grouped by block gas:

| block set | n | exec median | warmer/exec median | ratio >= 0.95 | finished after exec |
| --- | --- | --- | --- | --- | --- |
| 0-50 Mgas | 3133 | 26.7 ms | 0.268 | 26% | 282 |
| 50-150 Mgas | 401 | 580.8 ms | 0.968 | 61% | 101 |
| 150-250 Mgas | 476 | 1101.2 ms | 0.994 | 75% | 164 |
| 250-400 Mgas | 243 | 1638.7 ms | 0.996 | 76% | 92 |
| exec >= 1000 ms | 525 | 1576.4 ms | 0.998 | 97% | 256 |

Light blocks are fine at 0.268. Once a block carries real code-loading work the warmer consumes the whole execution window and lands at the wire or later: on the slowest blocks (~6100 ms exec) it finishes with 15-24 ms of slack, and on 256 of the 525 blocks over 1000 ms it finishes *after* execution. So the executor is racing the warmer for bytecode instead of being shielded by it, and warmer throughput is the constraint.

Two structural reasons, both in `LEVM::warm_block_from_bal`:

1. **A whole-block barrier.** Code hashes come from account states, so Phase 1 read *every* access-list account before Phase 2 issued the first bytecode read. For the entire duration of Phase 1 the executor faults in code itself, however early the warmer finishes in aggregate.
2. **One point get per code hash.** `code_hashes.par_iter().for_each(|h| store.get_account_code(h))` is a `get` per hash, capped at core count in flight, where the accounts alongside it already collapse into a single batched read. There was no batched bytecode read on the `Database` trait at all.

**Description**

- `get_account_codes_batch` on the VM `Database` trait, default impl looping `get_account_code` so non-rocksdb backends are unaffected, threaded through `DatabaseLogger`, `DynVmDatabase`, `CachingDatabase` and `StoreVmDatabase`.
- `Store::get_account_codes_batch` resolves the block-data buffer and the code LRU first, then reads what is left. Cold reads here are latency-bound, so the read path is chosen by which mechanism gets more requests in flight for the batch at hand: a rayon fan-out of point gets reaches queue depth ~= core count for free, while contiguous shards of the sorted keys on blocking threads reach queue depth ~= shard count and share RocksDB blocks within a shard, but pay a spawn each. It shards only once the batch forms more 256-key shards than there are cores, so ordinary blocks keep the fan-out and bloated ones get the depth. Each shard decodes its own values, keeping the jumpdest bitmap rebuild parallel. The LRU is locked once per batch rather than twice per code.
- `warm_block_from_bal` walks the access list in chunks instead of two whole-block phases, so code reads start after the first slice of addresses. The chunk is `256 * 64`, the smallest that still saturates the shard path, so chunking cannot trade the executor's head start for shallower reads.
- Code hashes now come out of the account read. `CachingDatabase::get_account_states_batch` returns the states it warms, which removes a `get_account_state` per access-list account (~115k on a bloated block) that existed only to recover hashes Phase 1 had already fetched.

**Notes**

- The prefetch stays best-effort. The batched read is all-or-nothing, so on error the warmer falls back to per-hash reads rather than leaving a whole chunk cold because one code hash in an access list is absent from the database.
- Witnesses are unaffected: `DatabaseLogger` records every hash in the batch before delegating, matching the per-hash path, and the warmer does not run when witness collection is on.
- `prefetch_accounts` keeps its own implementation rather than delegating to the states batch, because its other caller passes a whole write set and would pay for a result vector it discards.
- #6980 hoists these shard-tuning constants into a shared `shard_count` helper and adds the storage-side batch this mirrors. The constants are local here to keep this independent of that PR; worth folding into the helper whenever the two meet.
- Not yet measured. The bottleneck above is measured, and the redundant per-account lookups and lock traffic this removes are countable, but the size of the win depends on how well `MultiGet` batches blob-backed values, which only a run on the bloated fixtures will show. `warmer/exec` on heavy blocks is the direct readout: if that median does not drop below 1.0 with real slack, the batching was not the constraint.

**How was it tested?**

`test/tests/storage/account_code_batch_tests.rs` checks the batched read against the per-hash read the executor uses, since the prefetch warms a cache execution then trusts: same bytecode and same jumpdest bitmap per hash, `None` for absent hashes, results aligned to the caller's order rather than the internal read order, and every position of a hash repeated within one batch. Sized from `available_parallelism` so the parity check crosses into the sharded path, with a small-batch case covering the fan-out.

Plus the workspace suites and `-D warnings` clippy.


crates/storage/store.rs
```diff
@@ -943,6 +943,132 @@ impl Store {
...
+        //   spawn per shard.
+        //
+        // So shard only once sharding can actually beat the fan-out, i.e. once the batch
+        // is wide enough to form more shards than there are cores. Below that the
+        // fan-out is both deeper and cheaper, and a single serial `multi_get` would be
+        // far worse than either.
+        const KEYS_PER_SHARD: usize = 256;
+        const MAX_SHARDS: usize = 64;
+        let parallelism = std::thread::available_parallelism().map_or(8, |p| p.get());
+        let shards = missing.len().div_ceil(KEYS_PER_SHARD).min(MAX_SHARDS);
```
ElFantasma: `shards` is capped at `MAX_SHARDS` (64) but the branch condition is `shards > parallelism`, so on a machine with 64 or more cores the shard path is unreachable **at any batch size** — `shards` can never exceed 64, so it can never exceed `parallelism`.

| cores | smallest batch that reaches the shard path |
| --- | --- |
| 8 | ~5k hashes |
| 16 | ~5k |
| 32 | ~20k |
| 64+ | never |

The outcome isn't wrong — rayon fan-out on a 64-core box already reaches queue depth ~64, roughly what 64 shards would give — but two things the description promises quietly don't happen there: "bloated ones get the depth" is a no-op, and you lose the "share RocksDB blocks within a shard" benefit, which is the part fan-out can't replicate. It's also the class of machine where the bloated blocks this targets actually get executed.

If the intent is depth-plus-block-sharing rather than depth alone, the condition probably wants to key off batch size directly (`missing.len() >= KEYS_PER_SHARD * something`) rather than comparing a capped shard count against core count. Alternatively, raising `MAX_SHARDS` above the largest core count you care about restores the intended behaviour, at the cost of more spawns.

Worth a note either way, since the current form reads as "shard when the batch is big" and actually means "shard when the batch is big *and* the host is small".

crates/storage/store.rs
```diff
@@ -943,6 +943,132 @@ impl Store {
...
+                    .collect()
+            };
+            std::thread::scope(|scope| {
+                let handles: Vec<_> = missing
+                    .chunks(chunk)
+                    .map(|ck| scope.spawn(move || read_shard(ck)))
+                    .collect();
+                handles
+                    .into_iter()
+                    .flat_map(|h| h.join().expect("account code shard panicked"))
```
ElFantasma: `.expect` here turns a shard-thread panic into a panic on the warmer's thread, which sits oddly with the design note that "the prefetch stays best-effort [...] on error the warmer falls back to per-hash reads".

That fallback covers `Err`, not `panic!`. Nothing in `read_shard` looks likely to panic — `multi_get`, `decode_bytes` and `decode_jumpdests` all return `Result` — so this is about the failure mode rather than a bug I can point at: an allocation failure or a panic inside the backend escalates a best-effort prefetch into taking down whatever runs the warmer, instead of degrading to the per-hash path that already exists for exactly this situation.

Mapping the `join()` error into `StoreError` instead would let the existing fallback handle it and keep the best-effort promise total:

```rust
.map(|h| h.join().map_err(|_| StoreError::Custom("account code shard panicked".into())))
```

Cheap, and it means one bad shard costs the prefetch rather than the caller.

ElFantasma (changes requested):

crates/storage/store.rs
edg-l: Confirmed, and the table is right: with the cap a constant 64 the shard path was unreachable at 64+ cores, which is exactly where the bloated blocks run. Fixed in e32a3aee3, the cap now derives from the core count (`parallelism * 2`), so there is always a batch size where sharding is selected and the block-sharing benefit is not lost on big hosts.

Also re-derived the warmer chunk size in the same commit. It was sized from the old shard geometry, which was doubly wrong since a chunk bounds accounts while the code fetch keys on distinct uncached hashes.

crates/storage/store.rs
edg-l: Agreed, the fallback covered `Err` and not `panic`, so the best-effort promise was not total. Fixed in d3a603a9f.

Took your suggestion with one change: a panicked shard yields one `Err` per key it covered rather than a single `Err`. A single one makes the result vector shorter than the request, and the `zip` that assembles the output then silently misaligns. Worth noting the same construct in #6980 has that arity issue; it is not observable there because the first `Err` short-circuits before any misaligned write, but they should agree when the two land.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 ### 2026-08-04
 
 - Cut the cost of a cold contract-code access: store jump destinations as a 1-bit-per-byte bitmap instead of a persisted RLP list of `u32` offsets, count the bytecode in the code cache's byte budget, answer `EXTCODESIZE` from the code-length table instead of materializing the bytecode, and give the account-code column families a bloom filter and 4KB data blocks. `COLD_ACCOUNT_CODE_ACCESS` drops from 7736 to 4652 gas in the EIP-8038 repricing fit, and `COLD_ACCOUNT_CODE_WRITE` from 10415 to 6355 [#7095](https://github.com/lambdaclass/ethrex/pull/7095)
+- Batch and stream the BAL contract-code prefetch: warm accounts and their code in chunks instead of reading every access-list account before the first bytecode, take code hashes from the account read rather than a second lookup per account, and add a batched bytecode read that resolves the buffer and code cache first, then either fans out parallel point gets or shards the remainder across concurrent `multi_get`s, whichever reaches the greater read queue depth for the batch size [#PRNUM](https://github.com/lambdaclass/ethrex/pull/PRNUM)
```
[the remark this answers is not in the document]
edg-l: Already fixed. The placeholder was filled in once the PR number existed; the entry links #7099.

ElFantasma (approved):

iovoid (approved):

ilitteri (approved):

ilitteri (approved):

Issue #7097, opened 2026-08-03
Update our fork of the TDX DCAP/PCSS contracts


iovoid:
The automata TDX contracts have several updates, we should rebase our changes (deployment code) on top of the latest version.

PR #7102, opened 2026-08-04
chore(l2): pin tdx contract dependencies

iovoid:
**Motivation**

We want to update our TDX dependencies (#7097), but upstream has some breaking changes but since there are three repos involved (ethrex, [pccs](https://github.com/lambdaclass/automata-on-chain-pccs) and [dcap](https://github.com/lambdaclass/automata-dcap-attestation)), upgrading all three at once is fragile.

**Description**

Pins our TDX-related contract repos to a specific commit.

Required to solve #7097


ilitteri (approved):

edg-l (approved):

crates/l2/tee/contracts/Makefile
```diff
@@ -4,13 +4,46 @@ DEPLOYMENT_PATH := deploydeps/automata-dcap-attestation/evm/deployment
...
+# OpenZeppelin/openzeppelin-contracts v5.7.0
+OPENZEPPELIN_REV = cab19933c33c2ad1d4c7a84864a3601dddfd16f3
+# automata-network/automata-dcap-qpl, main tip as of 2026-08-03
+DCAP_QPL_REV = 3877d9aab99c090ad63bc71c18878e7659e26695
+# automata-network/pccs-reader-rs, main tip as of 2026-08-03
+PCCS_READER_REV = 926025f12cb63b6c1c50d5aad735223e71ed853b
+
+# Clone if absent, then re-assert and verify the pin on every run, so a stale or
+# half-checked-out tree is corrected rather than silently reused. After editing a
+# revision above, `make clean` to rebuild.
```
ElFantasma: `make clean` does not reset `lib/openzeppelin-contracts`, which this PR newly pins:

```make
clean:
	rm -rf deploydeps cache out deployment/* automata-dcap-qpl
```

The local `.gitignore` lists exactly three clone trees — `lib/`, `deploydeps/` and `automata-dcap-qpl/` — and `clean` removes two of them. `lib/` is the one it misses, and it is now a pinned tree like the others.

In the normal case this costs nothing, because `pin-openzeppelin` is `.PHONY` and re-asserts the checkout on every run, so bumping `OPENZEPPELIN_REV` takes effect without any clean at all. The gap only bites in the situation you would actually reach for `clean`: a tree the pin macro cannot re-assert — local edits in `lib/`, an interrupted checkout, a clone left pointing at a different remote. There the macro fails loudly with `lib/openzeppelin-contracts: not at <rev>` (good), and the documented escape hatch does not clear it, so the fix is an undocumented manual `rm -rf lib`.

Adding `lib` to the `clean` recipe makes the instruction true for all three:

```make
	rm -rf deploydeps lib cache out deployment/* automata-dcap-qpl
```

Worth doing here rather than later, since this PR is what turns `lib/` into a pinned artifact and the header comment above frames these revisions as "the only in-tree record of what a deployed verifier was built from".

ElFantasma (approved):
Pins all verified against their claimed sources; one nit where the documented `make clean` escape hatch misses `lib/`.

PR #7103, opened 2026-08-04
ci(l1,l2): key pooled rust caches by runner image

ilitteri:
**Motivation**

The v24.0.0-rc.1 release build failed with:

```
libparity_scale_codec_derive-*.so: /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.39' not found
```

Cache pools (`rust-cache`'s `shared-key`) are keyed only by `runner.os` ("Linux"), so `ubuntu-latest` (24.04) and `ubuntu-22.04` jobs in the same pool share one entry. The day's main push saved the `l2` pool entry from a 24.04 guest-build job; the tag run's 22.04 `build-ethrex` job restored it and could not load the glibc-2.39 proc-macro dylibs. Actions cache keys are immutable, so the poisoned entry persisted until deleted by hand (which is how the release was unblocked).

**Description**

Append `ImageOS` (`ubuntu22` / `ubuntu24` / `macos14`; empty on self-hosted runners) to the pooled `shared-key`, so different runner images never share dylib-bearing caches. The suffix is only applied when `cache-pool` is set — an empty `shared-key` must stay empty so pool-less jobs keep rust-cache's per-job keys.

Cost: one extra entry per (pool × image) actually in use; entries only save on `main`, unchanged.

**How to Test**

After merge, a main push saves `v0-rust-l2-ubuntu24-…` and 22.04 jobs restore/miss independently of it. The failure mode reproduced on [the v24.0.0-rc.1 run](https://github.com/lambdaclass/ethrex/actions/runs/30917441500) (attempt 1) and cleared after deleting cache key `v0-rust-l2-Linux-x64-1e811033-ba3f8f5a` and rerunning.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -58,7 +58,15 @@ runs:
     - name: Add Rust Cache
       uses: Swatinem/rust-cache@v2
       with:
-        shared-key: ${{ inputs.cache-pool }}
+        # ImageOS (e.g. ubuntu22, ubuntu24, macos14) keeps runner images from
+        # sharing one pool entry: rust-cache's default key only carries
+        # runner.os ("Linux"), so an ubuntu-latest job could save proc-macro
+        # dylibs linked against a newer glibc that ubuntu-22.04 jobs then
+        # restore and fail to load ("version `GLIBC_2.39' not found").
+        # Empty on self-hosted runners, which preserves the old key shape.
+        # Only pooled entries get the suffix: an empty shared-key must stay
+        # empty so pool-less jobs keep their per-job keys.
+        shared-key: ${{ inputs.cache-pool != '' && format('{0}-{1}', inputs.cache-pool, env.ImageOS) || '' }}
         save-if: ${{ inputs.save-cache == 'true' && github.ref == 'refs/heads/main' }}
         workspaces: |
           .


edg-l (approved):

.github/actions/setup-rust/action.yml
```diff
@@ -58,7 +58,15 @@ runs:
...
-        shared-key: ${{ inputs.cache-pool }}
+        # ImageOS (e.g. ubuntu22, ubuntu24, macos14) keeps runner images from
+        # sharing one pool entry: rust-cache's default key only carries
+        # runner.os ("Linux"), so an ubuntu-latest job could save proc-macro
+        # dylibs linked against a newer glibc that ubuntu-22.04 jobs then
+        # restore and fail to load ("version `GLIBC_2.39' not found").
+        # Empty on self-hosted runners, which preserves the old key shape.
+        # Only pooled entries get the suffix: an empty shared-key must stay
+        # empty so pool-less jobs keep their per-job keys.
+        shared-key: ${{ inputs.cache-pool != '' && format('{0}-{1}', inputs.cache-pool, env.ImageOS) || '' }}
```
ElFantasma, line 69: The whole fix rests on `env.ImageOS` resolving **inside a composite action**, and if it does not, this degrades to a silent no-op that is indistinguishable from success.

`ImageOS` is a runner-image environment variable, not something set by an `env:` key in this repo. The `env` context is documented as containing "variables set in a workflow, job, or step", and composite actions are where env-context inheritance is most surprising. I could not settle it from here: there is no other use of `ImageOS` anywhere in the repo to compare against, and the job logs I pulled came back truncated before the rust-cache key line.

It matters more than a normal uncertainty because of the failure shape. If `env.ImageOS` is empty, `format` still returns a truthy `"l2-"`, so the ternary does not fall through, the key stays valid, rust-cache still works, and every job still passes — with ubuntu-22.04 and ubuntu-24.04 sharing `v0-rust-l2--...` exactly as before. The GLIBC failure would only resurface on the next release tag, which is where it hurt last time. Nothing in CI asserts the key carries an image.

Cheapest confirmation is the rust-cache log line on the first main push after merge: `v0-rust-l2-ubuntu24-...` means it works, `v0-rust-l2--...` means it silently did not. If you would rather not wait for a main push to find out, a one-line `echo "suffix: $ImageOS"` step surfaces it immediately — in a shell that is a plain process env var and definitely resolves, unlike the expression context.

Everything else checks out. The diagnosis matches rust-cache's key composition (shared-key + runner.os + rustc/deps hashes, so two ubuntu images with the same host triple genuinely collide). The ternary is safe: with `cache-pool` set, `format` always yields a non-empty string, so it cannot fall through to the empty branch and accidentally un-pool a job. And keeping the suffix off the empty case correctly preserves per-job keys for pool-less callers.

ElFantasma (changes requested):

ilitteri:
Superseded by #7154, which fixes the same defect (rust-cache keying only on runner.os/runner.arch, so ubuntu-22.04 and ubuntu-24.04 jobs share one entry and a proc-macro dylib built on 24.04 fails in a 22.04 job with `GLIBC_2.39 not found`) more thoroughly:

- It applies the image family via rust-cache's `key` input, which separates entries for pooled *and* non-pooled jobs; this PR only suffixes the pooled branch, leaving same-name jobs across images colliding.
- It reads `ImageOS` through a step output instead of `env.ImageOS` in `with:`, because the `env` context is not reliably available to a `uses:` step inside a composite action — the mechanism this PR relies on.
- It also folds the read into the existing version-extraction step, so `pr_loc.yaml` (two checkouts in one job) does not hit the runner's "Index was out of range" post-step bookkeeping error.

Both edit the same `shared-key:` hunk, so they cannot both land without a manual resolution, and #7154 is already approved with checks clean. Closing in its favour; branch left in place.

PR #7104, opened 2026-08-04
perf(l1): give branch-node hashing a monomorphic RLP encoder

diegokingston:
**Motivation**

`RLPEncode::encode` takes `&mut dyn bytes::BufMut`. Encoding a branch node writes ~41 items through that trait object (16 children — each an RLP length byte plus a 32-byte hash — plus the header and value), so each one is a vtable dispatch.

That lands on the hashing path: `BranchNode::compute_hash_no_alloc` encodes into a `Vec<u8>` before keccak, and every `memoize_hashes` walk re-encodes each dirty branch. Note `encode_to_vec` directly below it is *already* monomorphic (`buf.push` / `extend_from_slice`) — it just allocates a fresh buffer each call, so the hashing path can't use it.

**Description**

Add `BranchNode::encode_into_vec(&self, buf: &mut Vec<u8>)`, a concrete-typed sibling that appends into a caller-owned `Vec<u8>`, and point `compute_hash_no_alloc` at it. `RLPEncode::encode` is untouched for every other caller.

Also resolve each child's hash once rather than twice — the payload-length pass and the encode pass each called `compute_hash_ref`, so a 16-choice branch performed 32 resolutions.

`cargo test -p ethrex-trie` (61 tests), `fmt` and `clippy` are clean.

**Effect**

The win scales with how poorly indirect calls are amortised. These numbers come from the [LambdaVM](https://github.com/yetanotherco/lambda_vm) zkVM guest executing mainnet block 25368371, where every RISC-V instruction is proven and there is no out-of-order core to hide a vtable dispatch:

| | cycles |
|---|---|
| `BranchNode::encode` + `put_slice` + `put_u8` | 2,122,427 → 1,312,210 (**−38%**) |
| `memcpy` | 908,885 → 586,969 |
| whole-block execution | 20,544,222 → 19,408,617 (**−5.53%**) |

`memcpy` drops as well because `extend_from_slice` into a `Vec` with known capacity beats the trait object's `put_slice`.

Native gains will be considerably smaller — a modern core predicts and pipelines these calls well — but the change strictly removes indirection, so it should not regress. Happy to add a native benchmark before merging if you'd like a number for that path.

Companion to #7101 (same guest, same profiling run).

### DIFF
--- a/crates/common/trie/rlp.rs
+++ b/crates/common/trie/rlp.rs
@@ -27,14 +27,19 @@ use crate::{Nibbles, NodeHash};
 // where `NativeCrypto` is the correct provider.
 impl RLPEncode for BranchNode {
     fn encode(&self, buf: &mut dyn bytes::BufMut) {
+        // Resolve each child's hash once: the length pass and the encode pass
+        // both needed it, so a 16-choice branch was paying 32 resolutions.
+        let hashes: [&NodeHash; 16] =
+            array::from_fn(|i| self.choices[i].compute_hash_ref(&NativeCrypto));
+
         let value_len = <[u8] as RLPEncode>::length(&self.value);
-        let payload_len = self.choices.iter().fold(value_len, |acc, child| {
-            acc + RLPEncode::length(child.compute_hash_ref(&NativeCrypto))
-        });
+        let payload_len = hashes
+            .iter()
+            .fold(value_len, |acc, hash| acc + RLPEncode::length(*hash));
 
         encode_length(payload_len, buf);
-        for child in self.choices.iter() {
-            match child.compute_hash_ref(&NativeCrypto) {
+        for hash in hashes {
+            match hash {
                 NodeHash::Hashed(hash) => hash.0.encode(buf),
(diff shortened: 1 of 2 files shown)


crates/common/trie/rlp.rs
```diff
@@ -27,14 +27,19 @@ use crate::{Nibbles, NodeHash};
 // where `NativeCrypto` is the correct provider.
 impl RLPEncode for BranchNode {
     fn encode(&self, buf: &mut dyn bytes::BufMut) {
```
MegaRedHand, line 29: I think we should try dropping BufMut from here first instead. encode_to_vec is duplicating logic from here

iovoid (approved):

crates/common/trie/rlp.rs
ilitteri: This could be done as a follow-up right?

ilitteri (approved):

MegaRedHand (approved):

crates/common/trie/rlp.rs
MegaRedHand: Yes. We should do it as a follow-up

PR #7106, opened 2026-08-04
fix(l2): normalize legacy omitted-withdrawals bodies read from trusted history

ilitteri:
**Motivation**

Older ethrex versions stored post-Shanghai blocks with `withdrawals: None` while the header committed to the empty-withdrawals root. `validate_block_body` rejects that shape, so re-reading such a block from local history fails with `WithdrawalsRootNotMatch`.

The normalization helper and its first call sites already reached `main` by another route, so what remains here are the follow-ups that came out of review, and one of them is load-bearing for a failure that only shows up inside the prover:

`generate_batch_prover_input` hands `ProverInputData.blocks` to the guest, and the guest re-encodes *those exact blocks* and KZG-checks the result against the committed blob (`crates/guest-program/src/l2/blobs.rs`). `BlockBody::encode` uses `encode_optional_field`, so an absent `withdrawals` field is omitted while an empty one writes `0xc0`. Because `prepare_batch_from_block` normalizes *before* `generate_blobs_bundle`, the blob for a batch containing legacy blocks holds the canonical encoding — so a prover-input path that did not normalize would differ by one byte and fail as `InvalidBlobProof` inside the zkVM. That is the concrete failure behind the review comment asking for `generate_batch_prover_input` to normalize too.

**Description**

- **One read path for stored blocks.** `read_trusted_block` (`crates/l2/sequencer/utils.rs`) is now the only place a block is read from the store and normalized, and both `prepare_batch_from_block` and `fetch_blocks_with_respective_fee_configs` go through it. Previously the two duplicated the store read and each carried its own normalization call, so correctness depended on two sites staying identical with nothing pinning them together — and the penalty for drift is a one-byte RLP mismatch surfacing as a KZG failure inside the guest. Routing both through one helper makes the coupling structural. It returns `Ok(None)` for a missing body, preserving the committer's "no new block to commit" loop exit instead of turning it into an error.
- Going through the helper also covers the reads that feed witness generation and the based-mode commit calldata, which previously read the store directly.
- **The helper's doc comment states the real invariant.** It previously claimed the point was agreeing on the canonical body the header commits to; for the prover-input path canonicality is incidental and the requirement is byte-equality with the bodies that went into the published blob. Stating the weaker property invited exactly the simplification the blob check punishes.
- **`ethrex import` deliberately left strict.** It still rejects legacy bodies, so an RLP export from a pre-upgrade node cannot be re-imported. That is the mirror image of the `reconstruct` blob-dir case, which this work does normalize, and the asymmetry is intentional: `reconstruct` reads blobs the node itself published to L1 and re-checks them against the committed root, whereas `import` takes an arbitrary file whose provenance the node cannot establish. Widening that boundary should be its own change with its own argument, not a side effect of this one.
- **Upgrade test now exercises the case.** Step 2 drains commits before the swap, so if block production stops there too, no batch ever spans blocks written by the old version and committed by the new one — the only situation where this normalization changes anything, meaning the test passed without touching it. Step 2.1 now leaves uncommitted blocks behind on purpose and records the L2 head; a new step 4.3 waits for a batch covering that head to be **verified** rather than merely committed, since committing only shows the sequencer accepted the blocks while the encoding is what the prover checks.

Noted while reviewing, not addressed here: the based follower applies L1-calldata blocks through `blockchain.add_block` (`crates/l2/based/block_fetcher.rs`), which never calls `validate_block_body` — that path is unaffected by any of this, but it also body-validates nothing today.

**Checklist**

- [x] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — n/a, no `Store` changes


edg-l (approved):

ElFantasma (approved):

iovoid (commented):
generate_batch_prover_input should also normalize, otherwise the data commited to the blob and given to the prover might diverge

pablodeymo:
Went through this one deeply, mostly to convince myself the shared-helper move is safe for the paths that don't re-execute. It is, and it's load-bearing rather than cosmetic — but for a reason the description doesn't mention, and that has a couple of consequences.

### Why this delta matters (worth putting in the description)

`generate_batch_prover_input` feeds `ProverInputData.blocks` to the guest, and the guest re-encodes *those exact blocks* and KZG-checks the result against the committed blob:

```rust
// crates/guest-program/src/l2/blobs.rs:30-42
blob_data.extend(len.to_be_bytes());
for block in blocks { blob_data.extend(block.encode_to_vec()); }
...
if !verify_blob_kzg_proof(blob_data, commitment, proof)? { return Err(L2ExecutionError::InvalidBlobProof) }
```

`BlockBody::encode` uses `encode_optional_field` (`crates/common/types/block.rs:414`), so `None` omits the field and `Some(vec![])` writes `0xc0`. Since `prepare_batch_from_block` already normalizes at `l1_committer.rs:878`, *before* `generate_blobs_bundle`, the blob for a v24-committed batch containing legacy blocks holds the canonical encoding. Without normalizing on the prover-input read too, those batches would fail in the guest as `InvalidBlobProof` — so @iovoid's comment was pointing at a real failure, and this fixes it.

(For what it's worth, pre-upgrade committed-but-unverified batches aren't a concern here: `upgrade-test.md:196-234` already requires draining to `lastCommittedBatch == lastVerifiedBatch`, and the vkey is keyed by commit hash anyway.)

### 1. The blob-consistency invariant now lives in two places, with nothing pinning them together

Correctness now depends on `l1_committer.rs:878` (blob path) and `utils.rs:216` (prover-input path) normalizing *identically*. Nothing documents or tests that coupling, and the two sites duplicate the same store read (`l1_committer.rs:856-880` vs `utils.rs:203-217`). Drop or move either one and every non-validium batch containing a legacy block fails as a KZG mismatch **inside the zkVM** — about the worst place to debug a one-byte RLP difference.

```rust
// crates/l2/sequencer/utils.rs — one read path, one normalization site
pub async fn read_trusted_block(store: &Store, number: BlockNumber) -> Result<Block, StoreError> {
    let header = store.get_block_header(number)?.ok_or(...)?;
    let body = store.get_block_body(number).await?.ok_or(...)?;
    let mut block = Block::new(header, body);
    normalize_legacy_withdrawals(&block.header, &mut block.body);
    Ok(block)
}
```

Then both `fetch_blocks_with_respective_fee_configs` and the `potential_batch_block` read in `prepare_batch_from_block` go through it. If you'd rather not touch `prepare_batch_from_block` in this PR, a comment on each site pointing at the other, plus a test asserting the helper's blocks encode identically to what `generate_blobs_bundle` is handed, would cover the drift risk.

### 2. The new doc comment states the wrong invariant

> Everything derived from these blocks -- witness generation, based-mode commit calldata, checkpoint re-execution -- therefore agrees on the canonical body the header commits to.

For the prover-input path canonicality is incidental; the requirement is byte-equality with the bodies that went into the *published blob*. A future reader taking this at face value could simplify one of the two sites without realizing the blob check pins them together. Something closer to:

```rust
/// Bodies come from immutable local history, so the legacy omitted-withdrawals
/// shape older ethrex versions stored is normalized here rather than by each
/// caller (see [`normalize_legacy_withdrawals`]). This must stay in sync with the
/// block read in `prepare_batch_from_block`: the guest re-encodes these blocks and
/// KZG-checks them against the committed blob, so the bodies fed to the prover input
/// have to be byte-identical to the ones `generate_blobs_bundle` encoded.
/// Never use this for blocks received from the network.
```

Also, the list of callers by name will rot as callers come and go, and the `--` should be `—` to match the rest of the file.

### 3. Testing — the scenario this delta changes isn't exercised

The only case where the delta changes behavior is *legacy blocks committed by a v24 sequencer*. Step 2 of the upgrade test drains commits before the swap, so if block production stops there too, no batch ever mixes v23 blocks with a v24 commit and this path stays untested. For the rc.2 run: keep producing L2 blocks while the committer is stopped, then check the first v24-committed batch spanning those blocks gets **proven and verified**, not just committed. That's the only real check for point 1.

### Minor

- `ethrex import` (`cmd/ethrex/cli.rs:1002` and `:1173`) still rejects legacy bodies, so an RLP export taken from a pre-v24 node can't be re-imported. That's the mirror image of the `reconstruct` blob-dir case, which this PR does normalize — worth a deliberate call on where the file-on-disk trust boundary sits.
- Unrelated, just noting it: the based follower applies L1-calldata blocks via `blockchain.add_block` (`crates/l2/based/block_fetcher.rs:241-243`), which never calls `validate_block_body` — only `add_block_pipeline` does (`blockchain.rs:728`). So that path is unaffected by all of this, but it also body-validates nothing today.
- Description is now slightly stale: it lists five sites including the checkpoint re-execution as its own, but the checkpoint path is covered *through* the helper now, which also newly covers the prover-input and based-calldata reads.

Nothing above blocks merging; 1 and 3 are what I'd want settled before this gets cherry-picked to `release/v24.0.0`.

pablodeymo (approved):

PR #7113, opened 2026-08-05
fix(l1): apply RocksDB max_bytes_for_level_base on state CFs

### LINKED ISSUE #7096: RocksDB: column-family options set on the DB-level `Options` never reach any CF (`max_bytes_for_level_base` 2 GB is a no-op)
## Summary

In `RocksDBBackend::open` (`crates/storage/backend/rocksdb.rs`), the DB-level `Options` sets nine column-family-scoped values that reach no column family. The one with a real effect is `max_bytes_for_level_base`: the comment says `// 2GB L1`, but every CF runs with RocksDB's 256 MB default.

Found while checking ethrex's RocksDB config against a third-party tool that mirrors it; the configs match, so this is purely an ethrex-side observation.

## Mechanism

`Options` in the `rocksdb` crate holds both DBOptions and ColumnFamilyOptions. `open_cf_descriptors(&opts, path, descriptors)` takes only the DBOptions half from `opts`; each CF's column-family options come from its own descriptor (`rocksdb-0.24.0/src/db.rs:679-691` passes `cf.options.inner` per descriptor, and the C API builds `DBOptions(db_options->rep)` + `ColumnFamilyOptions(cf_options[i]->rep)` separately).

Line 121 starts each CF from `Options::default()`, inheriting nothing:

```rust
for cf_name in &all_cfs_to_open {
    let mut cf_opts = Options::default();
```

And every CF gets a descriptor — `all_cfs_to_open = existing_cfs ∪ TABLES`, and `existing_cfs` always contains `default` (either `list_cf` returns it, or the fallback is `vec!["default"]`). So no CF falls through to the DB-level values.

## What is affected (line numbers at 0cc4572c5)

| Line | Setting | Consequence |
|---|---|---|
| 52-54 | `level_zero_*_trigger` 2/10/16 | none — 123-125 sets 4/20/36 per CF |
| 55 | `target_file_size_base` 512 MB | none — every arm sets its own |
| 61-62 | `write_buffer_size` 128 MB, `max_write_buffer_number` 4 | none — every arm sets its own |
| 76 | `compression_type` None | none — set per CF at 128/130 |
| 57 | `max_bytes_for_level_multiplier` 10.0 | none — equals the default |
| 58 | `level_compaction_dynamic_level_bytes` true | none — default is already true since 8.4 |
| **56** | **`max_bytes_for_level_base` 2 GB** | **no CF sets it → all get the 256 MB default** |
| **63** | **`min_write_buffer_number_to_merge` 2** | **only the 4 trie/FKV CFs set it; the rest get 1** |

Lines 47-50 and 60-75 are genuine DBOptions and do apply.

## Why 256 MB looks wrong for the state CFs

Under `level_compaction_dynamic_level_bytes`, `max_bytes_for_level_base` caps the base level target (`advanced_options.h`: the base target lands in `(max_bytes_for_level_base / multiplier, max_bytes_for_level_base]`, and the base level moves up one level each time it exceeds it). So it decides how many levels the LSM uses.

At ~490 GB of state with multiplier 10:

- base 256 MB → targets 256 MB / 2.5 GB / 25 GB / 256 GB below the last level, ~5 populated levels
- base 2 GB → 2 GB / 20 GB / 200 GB, ~4 populated levels

One fewer level. On the rough `write_amp ≈ 1 + multiplier × (levels - 1)` model that is ~41 → ~31, i.e. roughly 20-25% less compaction write volume. **Estimate, not measured.**

The sharper concern is the L0-to-base ratio on `account_trie_nodes` / `storage_trie_nodes` / `account_flatkeyvalue` / `storage_flatkeyvalue`. Those use `write_buffer_size` 512 MB, `min_write_buffer_number_to_merge` 2, `level0_file_num_compaction_trigger` 4, so L0 accumulates several GB before compacting — into a base level targeted at 256 MB. When L0's trigger size greatly exceeds the base target, each L0→base compaction rewrites most of the base level and immediately pushes it over target, cascading down.

(That L0-versus-base argument is reasoning from the level-sizing math. I recalled a RocksDB recommendation to set `max_bytes_for_level_base` ≈ total L0 size but could not find it on the Leveled-Compaction wiki page, so I am not citing it as a rule.)

Point lookups should be largely unaffected — those CFs carry 10-bit blooms with `cache_index_and_filter_blocks(true)`, so an extra level is an in-cache bloom check. The cost lands on compaction volume, and to a lesser degree on iterators (the FKV generator's trie walk, snap `GetAccountRange`).

The metadata CFs are the opposite case: 64 MB buffers × 4 ≈ 256 MB of L0, which matches the default. Nothing to fix there.

## Suggested fix

Move the values into the per-CF `match` arms, since they only take effect there — and the right value differs per CF, so a single DB-level number was the wrong shape regardless:

- set `max_bytes_for_level_base` explicitly in the `ACCOUNT_TRIE_NODES | STORAGE_TRIE_NODES` and `ACCOUNT_FLATKEYVALUE | STORAGE_FLATKEYVALUE` arms, sized against their multi-GB L0 footprint
- leave the metadata CFs on the 256 MB default
- either set `min_write_buffer_number_to_merge` in the arms that want it, or drop line 63
- delete the remaining dead setters (52-55, 57-58, 61-62, 76) so the file stops implying they do something

Worth measuring before picking a number: `max_bytes_for_level_base` is `SetOptions()`-changeable at runtime, and RocksDB's per-level compaction stats print a W-Amp column, so an A/B on one importing node would replace the estimate above with a real figure.

NikhilSharmaWe:
Closes #7096

**Summary**
`open_cf_descriptors` only applies DBOptions from the top-level `Options`. CF setters on that object never reach any column family, so `max_bytes_for_level_base(2GB)` was a no-op and every CF kept RocksDB’s 256 MiB default. State CFs (large memtables / L0) were especially exposed to excess compaction write amp.

This change:

- sets `max_bytes_for_level_base(2GB)` on the four state CFs (`account_trie_nodes`, `storage_trie_nodes`, `account_flatkeyvalue`, `storage_flatkeyvalue`)
- removes the ineffective CF options from DB-level `opts`
- leaves real DBOptions unchanged

**Benchmarking**
A/B tooling is on [`perf/rocksdb-level-base-bench`](https://github.com/NikhilSharmaWe/ethrex/tree/perf/rocksdb-level-base-bench) (stacked on this fix): tunable `--rocksdb.max-bytes-for-level-base`, compaction settle after `import-bench`, and state-CF `rocksdb.cfstats` dump. Primary metric is Sum W-Amp / Write(GB).

Compare **256 MiB vs 2 GiB vs 4 GiB** on a frozen DB with a long `chain.rlp` (≥5k–10k blocks), fresh copy per run.

I could not run that A/B locally (no frozen bench DB; machine too small to sync one). I smoke-tested the tooling on a 30-block `perf-ci` import: flag, settle, and cfstats dump all worked. RocksDB unit tests pass.

**Test plan**
- [x] `cargo test -p ethrex-storage --features rocksdb -- backend::rocksdb::tests`
- [ ] import-bench A/B on frozen DB via `perf/rocksdb-level-base-bench`

### DIFF
--- a/crates/storage/backend/rocksdb.rs
+++ b/crates/storage/backend/rocksdb.rs
@@ -49,18 +49,9 @@ impl RocksDBBackend {
 
         opts.set_max_background_jobs(8);
 
-        opts.set_level_zero_file_num_compaction_trigger(2);
-        opts.set_level_zero_slowdown_writes_trigger(10);
-        opts.set_level_zero_stop_writes_trigger(16);
-        opts.set_target_file_size_base(512 * 1024 * 1024); // 512MB
-        opts.set_max_bytes_for_level_base(2 * 1024 * 1024 * 1024); // 2GB L1
-        opts.set_max_bytes_for_level_multiplier(10.0);
-        opts.set_level_compaction_dynamic_level_bytes(true);
-
+        // Only DBOptions belong here. Column-family options must be set on each
+        // ColumnFamilyDescriptor — open_cf_descriptors ignores CF fields on `opts`.
         opts.set_db_write_buffer_size(1024 * 1024 * 1024); // 1GB
-        opts.set_write_buffer_size(128 * 1024 * 1024); // 128MB
-        opts.set_max_write_buffer_number(4);
-        opts.set_min_write_buffer_number_to_merge(2);
 
         opts.set_wal_recovery_mode(rocksdb::DBRecoveryMode::PointInTime);
         opts.set_max_total_wal_size(2 * 1024 * 1024 * 1024); // 2GB
@@ -73,7 +64,6 @@ impl RocksDBBackend {
         opts.set_enable_write_thread_adaptive_yield(true);
         opts.set_compaction_readahead_size(4 * 1024 * 1024); // 4MB
         opts.set_advise_random_on_open(false);
-        opts.set_compression_type(rocksdb::DBCompressionType::None);
 
         let compressible_tables = [
             BLOCK_NUMBERS,
@@ -187,6 +177,9 @@ impl RocksDBBackend {
                     cf_opts.set_max_write_buffer_number(6);
                     cf_opts.set_min_write_buffer_number_to_merge(2);
                     cf_opts.set_target_file_size_base(256 * 1024 * 1024); // 256MB
+                    // Base level target ≈ expected L0 size under memory-pressure flushes
+                    // (db_write_buffer_size often flushes before a full 512MB×2 merge).
+                    cf_opts.set_max_bytes_for_level_base(2 * 1024 * 1024 * 1024); // 2GB
                     cf_opts.set_memtable_prefix_bloom_ratio(0.2); // Bloom filter
 
                     let mut block_opts = BlockBasedOptions::default();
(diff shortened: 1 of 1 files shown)


NikhilSharmaWe:
cc @edg-l

PR #7122, opened 2026-08-09
fix(l1): reject BAL under-declaration outside transaction execution

### LINKED ISSUE #7109: Ethrex accepts a block whose EIP-7928 BAL omits a system-call storage read
A block whose Block Access List omits a storage slot read by one of the block's system calls is imported as valid, from the parallel exec path, which is the default (`bal_parallel_exec_enabled=true`). The sequential exec path works as expected, issue only with parallel exec path.  Reads made by transactions are caught by the per-tx shadow recorder. The pre-block and post-block system call phases have no equivalent.

The malicious entity i.e. builder can create such a payload where they drop a read-only entry.

## Expected

Ethrex should regenerate the canonical BAL during execution and compare it against the supplied one, rejecting on mismatch. That is what the sequential path does. Hence it should reject.

## Actual

Ethrex accepts (unexpected), while other EL clients like reth, geth, erigon reject (as expected).

Dropping a read-only entry does not change `state_root` or `gas`. So `block_access_list_hash` and the block hash are recomputed over the "trimmed" BAL. Hence the commitment check passes.

## Reproduction

1. Download [repro-main.patch](https://github.com/user-attachments/files/30747068/repro-main.patch) (it simply adds two tests that fail)
2. `git apply repro-main.patch`
3. `cargo test --release -p ethrex-test --test ethrex_tests bal_system_phase_under_declaration`

The tests added in the patch were verified to fail on main @ 26f748fdc.

zemse:
Closes #7109

**Motivation**

Ethrex behaves different from other clients like reth, geth, Erigon etc.

**Description**

The parallel import path only catches BAL under-declaration for transactions, via the per-tx shadow recorder. Accounts and slots touched by the pre-block system calls or the post-block requests and withdrawals phase were unchecked, so a block whose BAL omitted one of them was accepted while geth, reth and erigon rejected it. Both phases now run their own shadow recorder, as the per-tx path does, and the supplied BAL must match what they recorded: every entry present with the same values, and nothing extra declared at the phase's own index. The import path also runs `validate_ordering` now, which before only the Engine API did.

**Checklist**


zemse:
@ilitteri @edg-l been couple of weeks so bringing it to your attention. A builder can submit a block whose BAL is structurally valid and self-consistent but omits a slot read by a pre-block system call. Ethrex's parallel path accepts it, but geth/reth/erigon reject and it cause chain split. Found this while working on Deathstar in the EPF.

ilitteri (changes requested):
Thanks for the contribution and for identifying this validation gap. The approach preserves parallel execution while adding the missing system-phase checks.

I’m requesting changes because whole-account omission during pre-block execution appears to remain unchecked.

validate_bal_covers(..., "system calls", false) skips accounts absent from the supplied BAL. The comment relies on the state-root check to catch omitted writes, but the parallel merkleizer derives its root from that same supplied BAL. A builder can recompute both the BAL commitment and the header’s state root to match the omission.

For example, an empty Amsterdam block could omit the entire EIP-2935 history-contract entry. validate_bal_pre_exec_index only checks supplied entries, the new coverage check skips the missing account, and there are no transactions to catch the omission.

Could you add a regression that removes that entry, recomputes both commitments using the existing forge_state_root helper, and asserts rejection on the parallel path, alongside an unmodified positive control? I haven’t executed this reproduction; this finding comes from tracing the validation path.

The fix should enforce coverage of actual pre-block accesses and state changes while explicitly handling internal loads such as the expiry-verifier probe. Simply setting require_bal_entry = true for the whole cache could introduce the valid-block rejection described in the comment.

This can retain parallel execution and BAL-driven merkleization; the missing piece is independent completeness validation.

ilitteri:
Following up on my earlier review with a reproduction and two more findings after rebasing the branch onto current main locally.

**1. The whole-account omission I described is confirmed empirically.**
Empty Amsterdam block on `fixtures/genesis/l1-bal.json`, drop the entire EIP-2935 history-contract entry (1 storage write at index 0), recompute `block_access_list_hash` over the trimmed BAL, and forge `state_root` with the `forge_state_root` helper from `bal_content_validation_tests.rs`:

- sequential path (`bal_parallel_exec_enabled: false`): `Err(InvalidBlock(BlockAccessListHashMismatch))`
- parallel path (default): `Ok(None)` — the block is accepted.

So with this PR applied the pre-block phase is still under-declarable at account granularity. The doc comment on `validate_bal_covers` says the state root catches this "since the BAL feeds the merkleizer"; that is the reason it does *not*: the merkleizer's root is a pure function of the supplied BAL, so the builder recomputes it.

**2. `not(feature = "eip-8025")` no longer exists on main.**
#7115 removed the feature from `crates/vm/Cargo.toml`. Rebased, root `cargo clippy -- -D warnings` (the L1 CI gate) fails with `unexpected cfg condition value: eip-8025` at the new function. The gate should just be `#[cfg(feature = "rayon")]`.

**3. Suggested shape for the fix**, reusing the rule `validate_bal_withdrawal_index` already applies for the post-block phase: drop the `require_bal_entry` flag and require an entry when the cached account is modified or has any cached storage key:

```rust
None if !acct.is_unmodified() || !acct.storage.is_empty() => {
    return Err(/* ... missing from BAL */);
}
None => continue,
```

- pre-block writers (4788, 2935) are `Modified` -> required;
- a system contract that only SLOADs has storage keys -> required (the recorder records the read, so an honest BAL has the entry);
- the EIP-8141 expiry-verifier probe on later Hegota blocks is an unmodified info-only load with no storage -> skipped, which is the false-positive your comment guards against; on the first Hegota block the install marks it `Modified` and the recorder records the code/nonce change, so the entry exists.

Locally this makes the reproduction above reject with `BAL validation failed: account 0x…2935 was accessed during system calls but is missing from BAL`, and the three tests in this PR plus the other `bal_*` modules still pass. Please add the reproduction as a fourth test (with the unmodified block as the positive control, as you did for the others).

Minor:
- `first_read_only` in the first test picks whichever read-only entry sorts first; on this genesis that is the EIP-7002 withdrawal-request predeploy (4 storage reads at the post-block index), so the test covers the post-block half only and the second test is the one pinning the pre-block half. A one-line comment saying so would help the next reader.
- The control block is rebuilt from `build_store` instead of cloning `(block, bal)` before the trim; harmless, just redundant.

CI never ran on this PR (external-contributor workflows need approval), so the clippy failure above would only have shown up after approval.

zemse:
Thanks for the review. Rebased on current main and addressed all three points. Using codex/claude found a few more issues while testing the fix.

**1. Whole-account omission.** Confirmed: with your repro the parallel path returned `Ok(None)`. It's now a test. I didn't keep the `!acct.is_unmodified() || !acct.storage.is_empty()` rule, because inferring what a phase touched from the cache shape has a hole: only the request predeploys are guarded against empty code, so a call to a codeless 2935/4788 records the target but leaves only an info-only load in the cache, and that rule skips it. Instead both system phases run their own shadow recorder, and the supplied BAL is checked against what it recorded. The expiry-verifier probe then needs no exemption, since it records nothing.

**2. `eip-8025` cfg.** Dropped, gate is now `#[cfg(feature = "rayon")]`.

**3. Minor.** Documented what `first_read_only` picks, and the control block is cloned before the trim.

**Also fixed**, each with a regression test that returns `Ok(None)` without the fix:
- a write demoted to a read at index 0 (EIP-2935 writes every block, so membership checks miss it)
- an extra no-op change at the post-block index whose value matches the cache, which `validate_bal_withdrawal_index` accepts
- a bare entry for the unrecorded EIP-8141 probe, because `unaccessed_pure_accounts` was cleared from the cache
- a forged no-op change on `SYSTEM_ADDRESS` after a withdrawal to it
- duplicate changes at one index: `find_exact_change_*` reads the first, `synthesize_bal_updates` merkleizes the last, and `validate_ordering` only ran in the Engine API

The ordering fix is its own commit, so it's easy to split out if you'd rather take it separately.

Locally: clippy (root and workspace), fmt, `ethrex_tests` (1154) and the EF blockchain suite (11819) pass. The cases from ethereum/execution-specs#3551 are covered by the tests here, but those fixtures aren't in a release yet (pinned v8.1.4), so I haven't run them directly.

CI hasn't run on this PR yet (external contributor), could you approve it?

ilitteri (approved):

PR #7134, opened 2026-08-12
fix(l1): resolve repl against the root workspace

### LINKED ISSUE #7107: chore(l1): tooling/repl and tooling/runner resolve to the workspace that disowns them
`tooling/repl` is listed as a member of the **root** workspace (`Cargo.toml:27`) but physically lives inside the `tooling/` workspace, which lists neither it nor an exclusion for it. Cargo resolves the nearest workspace root, finds `tooling/Cargo.toml`, and fails:

```
$ cargo metadata --manifest-path tooling/repl/Cargo.toml --no-deps
error: current package believes it's in a workspace when it's not:
current:   /…/tooling/repl/Cargo.toml
workspace: /…/tooling/Cargo.toml

this may be fixable by adding `repl` to the `workspace.members` array of the manifest
located at: /…/tooling/Cargo.toml
```

Any `cargo` invocation with `tooling/repl` as the manifest — or run from inside that directory — hits this. Only root-relative invocations (`cargo run -p ethrex-repl`) work.

State on `main`:

- `Cargo.toml:27` claims `tooling/repl` as a root member
- `tooling/Cargo.toml` `members` omits it; `exclude` is only `["ef_tests/state"]`
- `tooling/repl/Cargo.toml` has no `[package] workspace` key

**#6801 adds a second instance.** It moves `runner` from `crates/vm/levm/runner` to `tooling/runner` while keeping it a root member, putting it in exactly the same position. That PR works around the symptom by documenting root-relative invocation in its README, which is the right call for a file-move PR — this issue is the follow-up it defers to.

## Suggested fix

One line per package:

```toml
[package]
workspace = "../.."
```

This names the root workspace explicitly and duplicates nothing.

The alternative — adding the packages to `tooling/Cargo.toml`'s `members` — would require copying `rustc-hash` and a `[workspace.lints]` table into the second workspace, giving two copies of each to drift apart.

questfever:
**Motivation**

`tooling/repl` is declared as a member of the root workspace, but Cargo discovers `tooling/Cargo.toml` first when the repl manifest is used directly.

Because the tooling workspace does not include `repl`, commands such as:

```console
cargo metadata --manifest-path tooling/repl/Cargo.toml --no-deps
```

fail with:

`current package believes it's in a workspace when it's not`

**Description**

Explicitly set the package workspace to the repository root with:

`workspace = "../.."`

This makes manifest-relative Cargo invocations resolve `tooling/repl` against the workspace that already declares it as a member.

`tooling/runner` is not included in this change because it is not present under `tooling/` on the current main branch.

**Checklist**

Closes #7107

### DIFF
--- a/tooling/repl/Cargo.toml
+++ b/tooling/repl/Cargo.toml
@@ -1,4 +1,5 @@
 [package]
+workspace = "../.."
 name = "ethrex-repl"
 version.workspace = true
 edition.workspace = true


questfever:
@ilitteri @iovoid Could you please take a look when you have a chance? Thanks!

PR #639, opened 2024-10-02
feat: (L1) Unify crates under same directory

belfortep:
**Motivation**

Reduce the ammount of crate folders in the project

**Description**

Issue #627, opened 2024-10-02
LEVM: Transaction types


lazcanoluca:
- Define the transaction types (EIP-1559, Legacy, etc), but with `sender` instead of `r`, `s`, `parity`.

PR #7142, opened 2026-08-13
fix(l1): populate blockTimestamp on RPC log objects

ilitteri:
**Motivation**

Log objects returned by `eth_getBlockReceipts`, `eth_getTransactionReceipt` and `eth_getLogs` omit `blockTimestamp`. Comparing a devnet block's receipts against geth: same receipt count, same receipt-level keys, but **156 differing fields across 226 receipts** — all from one missing key inside each log.

```
receipt[8].logs[0]  blockTimestamp:  geth "0x6a7cf6d0"   ethrex <absent>
```

geth, besu, nethermind, reth and erigon all populate it on every log object. Indexers that read the timestamp straight off a receipt's logs get nothing from ethrex and must fall back to a separate block lookup per receipt.

The field is defined on the `Log` object in the execution-apis schema (`src/schemas/receipt.yaml`) but only `transactionHash` is required, so omitting it is schema-legal. This is a convention divergence rather than a spec violation — the argument for fixing it is the per-receipt lookup every consumer otherwise pays.

**Description**

Adds `block_timestamp` to `RpcLog`. One struct change covers all three endpoints, since `eth_getLogs` and the receipt getters build the same type.

The timestamp is threaded in as an explicit parameter rather than read from `RpcReceiptBlockInfo`. That struct is `#[serde(flatten)]`-ed into `RpcReceipt`, so sourcing it from there would also have emitted a **receipt-level** `blockTimestamp` that no other client sends — and the receipt-level keys already match geth exactly. `blockTimestamp` belongs on the log object and nowhere else.

Decoding defaults the field, so a peer that legitimately omits it (the schema permits it) still parses instead of failing the whole response.

**How to test**

```
cargo test -p ethrex-rpc --lib receipt
```

- `serialize_receipt` — updated expected JSON, pinning `blockTimestamp` inside the log object.
- `block_timestamp_is_on_the_log_and_not_on_the_receipt` — new; asserts both halves, so the flatten hazard above cannot regress silently.

End to end, against a synced node:

```
curl -s -X POST -H 'content-type: application/json' \
  --data '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["latest"]}' \
  http://localhost:8545 | jq '.result[0].logs[0].blockTimestamp'
```

`blockTimestamp` should be present on every log and equal the block's `timestamp`, and must **not** appear at the receipt level.

**Checklist**

- [x] No `Store` schema change, so `STORE_SCHEMA_VERSION` is untouched.

### rpc-compat: eight cases excluded, permanently

The first CI run failed 8 rpc-compat cases — exactly the ones whose recorded response contains at least one log object (`eth_getLogs/no-topics` with 15 logs down to the three single-log `eth_getTransactionReceipt` cases). All 13 cases with an empty log array passed, which places the cause on the one added key rather than on a wrong value.

hive's rpc-compat compares responses byte-exactly: `runTest` requires `jsondiff.FullMatch`, and the lenient `checkJSONStructure` path applies only to cases upstream marks `speconly`. The corpus is pinned to execution-apis `d08382ae` (2025-02-10), whose recordings predate `blockTimestamp` — it entered the schema in [#639](https://github.com/ethereum/execution-apis/pull/639) and those fixtures in [#846](https://github.com/ethereum/execution-apis/pull/846). So the field cannot match, no matter that the other five clients all send it.

**The pin cannot move.** It sits one commit before [#627](https://github.com/ethereum/execution-apis/pull/627), which moved the test chain to a pre-merge genesis — the current corpus carries ~36 proof-of-work blocks before its terminal total difficulty. ethrex does not support pre-merge chains and will not, so that `chain.rlp` fails to import at block 1 (`validate_block_header` has no pre-London base-fee path). I tried the bump: the job went from 8 failures to `tests=1 failed=1`, the single failure being `client launch`, with no method test executing at all. Every revision that has `blockTimestamp` in its fixtures also has that chain, so no revision satisfies both, and the corpus cannot be patched locally either — rpc-compat's Dockerfile clones `ethereum/execution-apis` by hard-coded URL, so the `branch` buildarg cannot point at a fork.

So the eight cases are counted out in `KNOWN_EXCLUDED_TESTS`, which prints `Ignoring N known-excluded test(s)` rather than passing silently, and the reasoning is recorded in `docs/known_issues.md` — which CI posts as a sticky comment on every PR.

Worth being explicit about the cost: all four `eth_getLogs` cases are in the set, so that method keeps no rpc-compat coverage. The field itself is pinned instead by `block_timestamp_is_on_the_log_and_not_on_the_receipt`, which asserts it is present on each log and absent from the receipt level.

### DIFF
--- a/.github/scripts/check-hive-results.sh
+++ b/.github/scripts/check-hive-results.sh
@@ -59,7 +59,23 @@ mkdir -p "${failed_logs_root}"
 
 # Tests excluded from the failure count (substring match against test case
 # name).
+#
+# The rpc-compat cases below are the ones whose recorded response contains at
+# least one log object. That corpus is pinned to execution-apis d08382ae, which
+# predates `blockTimestamp` on log objects, and rpc-compat compares byte-exactly,
+# so emitting the field (as every other client does) cannot match. The pin cannot
+# move: the field only reaches those fixtures in revisions that also carry a
+# pre-merge test chain, which ethrex does not support by design. See
+# docs/known_issues.md.
 KNOWN_EXCLUDED_TESTS=(
+  "eth_getLogs/contract-addr"
+  "eth_getLogs/no-topics"
+  "eth_getLogs/topic-exact-match"
+  "eth_getLogs/topic-wildcard"
+  "eth_getBlockReceipts/get-block-receipts-latest"
+  "eth_getTransactionReceipt/get-access-list"
+  "eth_getTransactionReceipt/get-blob-tx"
+  "eth_getTransactionReceipt/get-dynamic-fee"
 )
 
 # Build a jq filter that excludes the known-excluded tests.
--- a/crates/networking/rpc/eth/block.rs
+++ b/crates/networking/rpc/eth/block.rs
@@ -356,6 +356,7 @@ pub async fn get_all_block_rpc_receipts(
         .map_err(|_| RpcErr::Internal("blob_base_fee does not fit in u64".to_owned()))?;
     // Fetch receipt info from block
     let block_hash = header.hash();
+    let block_timestamp = header.timestamp;
     let block_info = RpcReceiptBlockInfo::from_block_header(header);
     // Fetch receipts: only up to target_index+1 when set, otherwise all
     let fetch_count = target_index
@@ -395,6 +396,7 @@ pub async fn get_all_block_rpc_receipts(
             tx_info,
             block_info.clone(),
             current_log_index,
+            block_timestamp,
         );
         last_cumulative_gas_used += gas_used;
         current_log_index += receipt.logs.len() as u64;
--- a/crates/networking/rpc/eth/logs.rs
+++ b/crates/networking/rpc/eth/logs.rs
@@ -316,6 +316,7 @@ async fn collect_block_logs(
                         transaction_index: tx_index as u64,
                         block_number: block_num,
                         block_hash,
+                        block_timestamp: block_header.timestamp,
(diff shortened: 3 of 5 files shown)


.github/scripts/check-hive-results.sh
```diff
@@ -59,7 +59,23 @@ mkdir -p "${failed_logs_root}"
...
+# docs/known_issues.md.
 KNOWN_EXCLUDED_TESTS=(
+  "eth_getLogs/contract-addr"
+  "eth_getLogs/no-topics"
+  "eth_getLogs/topic-exact-match"
+  "eth_getLogs/topic-wildcard"
+  "eth_getBlockReceipts/get-block-receipts-latest"
+  "eth_getTransactionReceipt/get-access-list"
+  "eth_getTransactionReceipt/get-blob-tx"
+  "eth_getTransactionReceipt/get-dynamic-fee"
```
MegaRedHand, line 78: Are we sure these are only pre-merge cases?

MegaRedHand (approved):

.github/scripts/check-hive-results.sh
ilitteri: Yes. For all 8, the entire diff is one line (`-- "blockTimestamp"`), present in ours and absent from the recording. 

The tradeoff is that if we pin a newer commit, we would leave the RPC with no coverage at all. So we decided to keep this and ignore these 8.

PR #7147, opened 2026-08-13
refactor(l1): build the local ENR outside the discovery server

MegaRedHand:
**Motivation**

`DiscoveryServer::spawn` derived the record it advertises from `local_node`,
then stamped our own fork id onto it. Neither is discovery's decision to make.
What a node publishes about itself belongs to the caller: which entries the
record carries beyond the addresses, and what `seq` it resumes from. Deriving it
inside `spawn` left no way to say either.

It also forced a `panic` on a failure the caller reports as an error, via two
`expect`s on the signing path.

**Description**

`spawn` takes a `NodeRecord` instead of building one, and the construction moves
to `start_network`, which already holds the node, the signer and the store.

The resulting record is byte-identical to what `spawn` produced: `from_node` at
`INITIAL_ENR_SEQ` followed by `set_fork_id`, which bumps `seq` once. No entry,
sequence number or signature changes.

That was the last reader of the `Store` `spawn` took. #7145 moved the fork-id
screening behind `PeerFilter` and removed the `DiscoveryServer::store` field it
fed; with the ENR built elsewhere, discovery no longer touches our chain at all,
so the parameter goes too. Neither change could drop it alone, which is why this
sits on top of #7145 rather than on `main`.

The two `expect`s become a `NetworkError`. `init_network` ends in
`.expect("Network starts")`, so an unsignable record still aborts startup rather
than leaving a node running without discovery; it now does so from one place and
with the underlying `NodeError` attached. A chain that cannot supply a fork id
still leaves the entry out and starts, since the record is usable for discovery
and a peer that requires `eth` simply passes us over until we republish.

**What this unblocks**

`cmd/ethrex` already builds a local record, in `get_local_node_record`. That one
resumes from a `seq` persisted in the node config file (or a unix timestamp on
first run), and is the record `admin_nodeInfo` reports and that shutdown writes
back. Discovery never saw it, and gossiped its own starting from
`INITIAL_ENR_SEQ`, so the ENR peers receive and the ENR the node reports have
different sequence numbers, and only the gossiped one carries `eth`.

Restarting therefore re-announces at `seq` 2 no matter how high the previous
run climbed, and a peer holding a cached higher `seq` treats the new record as
stale. Handing discovery the record that already exists fixes that, but it is a
behaviour change and wants its own PR; this one only makes it expressible.

**Tests**

No new ones: the record is built from the same two calls in the same order, so
there is no new behaviour to pin. Existing `p2p` unit and integration tests pass.

**Checklist**

### DIFF
--- a/crates/networking/p2p/discovery/server.rs
+++ b/crates/networking/p2p/discovery/server.rs
@@ -106,16 +106,6 @@ impl std::fmt::Debug for DiscoveryServer {
 impl DiscoveryServer {
     /// Starts the discovery actor, advertising `local_node_record` as this
     /// node's ENR.
-    ///
-    /// The record is supplied rather than derived from `local_node`, because
-    /// what a node publishes is the caller's to decide: which entries it
-    /// carries beyond the addresses, and what `seq` it resumes from. Deriving
-    /// it here also forced a panic on a failure the caller is better placed to
-    /// report.
-    ///
-    /// This was the last reader of the `Store` `spawn` used to take: with the
-    /// fork-id screening moved behind `PeerFilter`, discovery no longer touches
-    /// our chain at all.
     pub async fn spawn(
         local_node: Node,
         local_node_record: NodeRecord,
--- a/crates/networking/p2p/discovery/server.rs
+++ b/crates/networking/p2p/discovery/server.rs
@@ -8,11 +8,10 @@ use crate::{
         server::{Discv5Message, Discv5State, update_local_ip},
     },
     peer_table::{DiscoveryProtocol, PeerTable, PeerTableServerProtocol as _},
-    types::{INITIAL_ENR_SEQ, Node, NodeRecord},
+    types::{Node, NodeRecord},
 };
 use bytes::BytesMut;
 use ethrex_common::utils::keccak;
-use ethrex_storage::Store;
 use futures::StreamExt;
 use secp256k1::SecretKey;
 use spawned_concurrency::{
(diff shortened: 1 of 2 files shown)


crates/networking/p2p/discovery/server.rs
```diff
@@ -105,9 +104,21 @@ impl std::fmt::Debug for DiscoveryServer {
...
+    ///
+    /// The record is supplied rather than derived from `local_node`, because
+    /// what a node publishes is the caller's to decide: which entries it
+    /// carries beyond the addresses, and what `seq` it resumes from. Deriving
+    /// it here also forced a panic on a failure the caller is better placed to
+    /// report.
+    ///
+    /// This was the last reader of the `Store` `spawn` used to take: with the
+    /// fork-id screening moved behind `PeerFilter`, discovery no longer touches
+    /// our chain at all.
```
MegaRedHand:
```suggestion
```

crates/networking/p2p/discovery/server.rs
```diff
@@ -105,9 +104,11 @@ impl std::fmt::Debug for DiscoveryServer {
 
 #[actor(protocol = DiscoveryServerProtocol)]
 impl DiscoveryServer {
+    /// Starts the discovery actor, advertising `local_node_record` as this
+    /// node's ENR.
     pub async fn spawn(
-        storage: Store,
         local_node: Node,
+        local_node_record: NodeRecord,
```
MegaRedHand, line 111: These should be deduped in the future

ilitteri (approved):

PR #3356, opened 2025-06-27
perf(levm): reduce handle_debug runtime cost

edg-l:
**Motivation**

With this handle_debug disappears from samplies on push/memory heavy benches.

**Description**

Closes #3355

Issue #3292, opened 2025-06-24
Embed dev-mode genesis in the client


MegaRedHand:
Whenever a user specifies `--dev`, we need them to manually specify a genesis JSON. We could, instead, embed in the client the genesis file we are using, as a default. That way, users can start an L1 local node with just `ethrex --dev`

MegaRedHand:
The same thing happened with public networks. That was fixed for public networks in #3372, and can be done in a similar way for the dev-mode genesis.

PR #7153, opened 2026-08-14
fix(l1,l2): correct the stateless guest's Amsterdam blob schedule and move EF vectors to zkevm@v0.8.0

### LINKED ISSUE #7150: L1 CI: stateless vectors are regenerated from execution-specs on every PR, on all three runners
Since #7115, `make -C tooling/ef_tests/blockchain test` reaches `test-stateless`, which
depends on `stateless-vector`, which runs `scripts/gen_stateless_vectors.sh`. That script:

- `git clone https://github.com/ethereum/execution-specs`
- `uv venv --python 3.12`
- two `uv pip install` of the spec source tree
- an EEST `fill` of the whole `eip8025_optional_proofs` suite

`.github/workflows/pr-main_l1.yaml` runs that job on
`[ubuntu-22.04, ubuntu-22.04-arm, macos-15]`, so every L1 PR pays it three times,
against two external networks (GitHub and PyPI), with no caching and no retry.

The script's own header notes the spec pulls in `coincurve`, which builds from source
via scikit-build-core. A macOS/arm64 wheel failure, a PyPI blip, or a GitHub outage
therefore fails L1 CI on changes that have nothing to do with stateless validation —
the same class of flake that `.github/scripts/download.sh` was introduced to stop for
release downloads.

The vectors are needed; generating them per-PR-per-runner is what costs. Options,
roughly in order of preference:

1. Generate once and publish as a release asset, then fetch it through
   `download.sh` the way the Amsterdam and zkEVM bundles already are. Restores a
   hermetic, retrying, cacheable download and drops the Python toolchain from L1 CI
   entirely.
2. Generate once in a dedicated job and pass it to the matrix legs as an artifact.
3. Cache the clone and the uv venv keyed on the pinned execution-specs commit.

Option 1 also removes the `uv` prerequisite from the local `make test` path.

Context: raised while reviewing #7115. Not a blocker for that PR — the coverage it adds
is wanted, the delivery mechanism is what needs changing.

### LINKED ISSUE #7152: Bump stateless fixtures to tests-zkevm@v0.8.0 and drop the local vector generator
[`tests-zkevm@v0.8.0`](https://github.com/ethereum/execution-specs/releases/tag/tests-zkevm%40v0.8.0) (2026-08-12) is based on [`tests-glamsterdam-devnet@v8.1.0`](https://github.com/ethereum/execution-specs/releases/tag/tests-glamsterdam-devnet%40v8.1.0), which is the base `.fixtures_url_amsterdam` already points at. That closes the gap that forced the current workaround in #7115, and it should let us delete the local vector generator entirely.

Upstream skipped `v0.7.x` to align the zkEVM minor with the `tests-glamsterdam-devnet` major, so `v0.8.0` is the successor to `v0.6.2`, not a jump over a release we missed.

## Why this matters now

`make test-stateless` currently reads a locally generated vector set rather than a released bundle, because no published revision had both the EIP-8025 stateless schema and the devnet-8 gas schedule:

- `scripts/gen_stateless_vectors.sh` pins execution-specs `3c3b6f4af` — the [#3278](https://github.com/ethereum/execution-specs/pull/3278) merge commit — which carries the **devnet-7** gas schedule.
- ethrex implements **devnet-8**, so 45 of 105 fixtures fail in `add_block_pipeline` with `GasUsedMismatch` / `ReceiptsRootMismatch` before stateless validation runs, and sit in `EXTRA_SKIPS` in `tooling/ef_tests/blockchain/tests/all.rs`.
- Generating locally means every L1 PR clones execution-specs and runs an EEST `fill` under `uv` on three runners — tracked in #7150.

`v0.8.0` resolves all three. Verified against the tag:

| Constant | `3c3b6f4af` (current pin) | `tests-zkevm@v0.8.0` | ethrex |
| --- | --- | --- | --- |
| `COLD_STORAGE_ACCESS` | 3000 | 2100 | 2100 |
| `ACCOUNT_WRITE` | 8000 | 9000 | 9000 |
| `CALL_VALUE` | 10300 | 11300 | 11300 |
| access-list address / key | full cold cost | cold − `WARM_ACCESS` | cold − `WARM_ACCESS` |

## What is actually new relative to our pin

Our pin *is* the #3278 merge, so four of the six changes in the release notes are already in the vectors we generate today: #3179 (2026-07-15), #3234 (2026-07-26), #3248 (2026-07-29), #3278 (2026-08-03). The delta is two PRs.

### [#3356](https://github.com/ethereum/execution-specs/pull/3356) — `SszList` → `ProgressiveList` for state, codes and public keys

A wire-format change we must follow in `crates/common/types/stateless_ssz.rs`:

| Field | Current | `v0.8.0` |
| --- | --- | --- |
| `state` | `SszList<SszList<u8, MAX_BYTES_PER_WITNESS_NODE>, MAX_WITNESS_NODES>` | `ProgressiveList<SszList<u8, MAX_BYTES_PER_WITNESS_NODE>>` |
| `codes` | `SszList<SszList<u8, MAX_BYTES_PER_CODE>, MAX_WITNESS_CODES>` | `ProgressiveList<SszList<u8, MAX_BYTES_PER_CODE>>` |
| `public_keys` | `SszList<SszVector<u8, PUBLIC_KEY_BYTES>, MAX_PUBLIC_KEYS>` | `ProgressiveList<SszVector<u8, PUBLIC_KEY_BYTES>>` |

`headers` stays `SszList<_, MAX_WITNESS_HEADERS>` — it is explicitly *not* in scope upstream. `MAX_WITNESS_NODES`, `MAX_WITNESS_CODES` and `MAX_PUBLIC_KEYS` are deleted upstream and become unused on our side.

### [#3292](https://github.com/ethereum/execution-specs/pull/3292) — preserve block `extraData`

Mostly fixture construction, but it also touches `forks/amsterdam/execution_engine/validation_helpers.py`, so it needs a read before we assume the client is unaffected. Fixtures will now carry non-empty `extraData` where it was previously dropped.

## The schema ID does not distinguish these bodies

> [!WARNING]
> At the `v0.8.0` tag, `STATELESS_INPUT_SCHEMA_REVISION` is still `0x01` and the fork index still `0x15`, so `STATELESS_INPUT_SCHEMA_ID` is **`0x1501` — unchanged**, even though #3356 altered the body encoding.

That makes `0x1501` the identifier for a third distinct body: v0.6.2's, our current pin's, and v0.8.0's. It is the same trap `docs/known_issues.md` already records for v0.6.2 ("reuses schema id 0x1501 for the OLD body, so its vectors misparse"), and our `only_amsterdam_schema_id_decodes` check accepts `0x1501` on its face.

Mixing bundles will not be caught by the schema prefix — only by a decode failure or a wrong root downstream. Worth raising upstream, and worth a comment on our constant either way.

## Work

- [ ] Point `.fixtures_url_zkevm` at the `v0.8.0` `fixtures_zkevm.tar.gz`
- [ ] Adopt #3356: `ProgressiveList` for `state`, `codes` and `public_keys`; drop the three now-unused bound constants
- [ ] Read #3292 and decide whether `validation_helpers.py` implies any client change
- [ ] Repoint `TEST_FOLDER` in `tests/all.rs` from `vectors_stateless_3278/` back to the downloaded bundle, and restore `zkevm-vectors` as a `download-test-vectors` prerequisite
- [ ] Delete `scripts/gen_stateless_vectors.sh`, `STATELESS_VECTORS_ROOT`, `stateless-vector` and the `uv` / Python 3.12 prerequisite — closes #7150
- [ ] Empty `EXTRA_SKIPS` and confirm the suite is green; any fixture still failing is a real bug, not spec skew, and must not be re-skipped without a written reason
- [ ] Rewrite the "Stateless EF tests: 45 of 105 fixtures skipped" section of `docs/known_issues.md`
- [ ] Re-check the five original skips specifically — they failed against `3c3b6f4af` even while ethrex was on devnet-7, so they are the ones most likely to be genuine

## Acceptance

`make test-stateless` passes on a downloaded `v0.8.0` bundle with an empty `EXTRA_SKIPS`, and no L1 CI job runs an EEST `fill`.

Context: skips added in #7115; generator cost tracked in #7150.

ilitteri:
**Motivation**

`tests-zkevm@v0.8.0` is based on `tests-glamsterdam-devnet@v8.1.0`, the same base `.fixtures_url_amsterdam` already points at. It is the first zkEVM release carrying both the EIP-8025 stateless schema and the devnet-8 gas schedule — earlier releases had at most one of the two, which is why the stateless vectors were being filled locally from execution-specs `3c3b6f4af` and 45 fixtures were skipped for a devnet-7/devnet-8 gas split.

Moving to the released bundle removes the generator, its `uv` / Python 3.12 prerequisite, the EEST fill that ran on every L1 PR across three runners, and the whole skip list.

Upstream skipped `v0.7.x` to align the zkEVM minor with the `tests-glamsterdam-devnet` major, so `v0.8.0` is the successor to `v0.6.2`.

**Description**

Four commits. The first two are the substance and are deliberately separate; the last two are a build fix and review follow-up.

**`ac2a7efbe` — the guest resolved the wrong blob schedule at Amsterdam.** `amsterdam_chain_config` scheduled every *named* fork at 0 but left `bpo1_time`..`bpo5_time` as `None`. EIP-7892 gives named forks no blob params of their own, so `get_fork_blob_schedule` walks the BPO chain and Amsterdam inherits the highest activated entry — with none activated it fell through to Osaka's schedule instead of BPO2's:

| | Osaka (what it used) | BPO2 (correct) |
| --- | --- | --- |
| `target` | 6 | 14 |
| `max` | 9 | 21 |
| `base_fee_update_fraction` | 5007716 | 11684671 |

The guest therefore recomputed `excess_blob_gas` against the wrong target and rejected valid blob-bearing blocks with `ExcessBlobGasIncorrect`, committing `successful_validation = 0`. Only the stateless path was affected — everywhere else the config comes from genesis with the BPO times set, which is why the same fixtures pass statefully and fail through the guest. It cost 18 fixture files across `eip4844_blobs` and `eip7918_blob_reserve_price`, each a single flipped byte in the committed result.

Only BPO1 and BPO2 are activated, matching the `AMSTERDAM_CONFIG` the fixtures are filled against. BPO3-5 have no default `blob_schedule` entry, and a BPO time set without one is rejected as invalid genesis config, so scheduling them would be both wrong and unresolvable. The added test pins the *resolved* schedule rather than the field values, so it also catches a change to the fallthrough order.

**`227ae1b80` — read the bundle instead of generating it.** Four of the six changes in the release are already in the vectors generated today, since the pinned commit is the #3278 merge. The delta this adopts:

- [#3356](https://github.com/ethereum/execution-specs/pull/3356) moves `state`, `codes` and `public_keys` from `SszList` to `ProgressiveList`. `headers` is deliberately not in scope upstream and stays a bounded `SszList`. The element-count bounds disappear with it, so `MAX_WITNESS_NODES`, `MAX_WITNESS_CODES` and `MAX_PUBLIC_KEYS` are deleted and the outer list conversions in the L2 native-rollup advancer — which builds the witness on the producing side — become infallible. The per-element byte caps stay.
- [#3292](https://github.com/ethereum/execution-specs/pull/3292) only changes how fixtures are constructed; nothing on the client side.

`EXTRA_SKIPS` is now empty, and the stateless run covers all 3218 `for_amsterdam` fixture files rather than the 100 the generated set carried — the whole subtree is exercised statelessly, not just `eip8025_optional_proofs`. That wider coverage is what surfaced the blob-schedule bug.

`docs/known_issues.md` gains a section recording that the schema id does not identify the encoding: upstream has now shipped three incompatible bodies under `0x1501`, so a stale bundle fails in decode or on a mismatched root rather than being rejected on its prefix. Worth raising upstream — a revision field that does not move across a body change provides no version negotiation.

**`78f3ecfc5`** fixes a call site behind the `l2` feature that a default-feature `--all-targets` check does not compile; it broke five checks, all through the `ethrex_tests` target. **`95ce3cf16`** applies review feedback: the tree states what is true now rather than how it got here, and the "Stateless EF tests: no skips" section is removed from `known_issues.md` since an empty skip list is not a known issue.

**How to test**

```bash
make -C tooling/ef_tests/blockchain test-stateless   # downloads the v0.8.0 bundle
cargo test -p ethrex-common amsterdam_chain_config
```

To see the bug the first commit fixes, revert `bpo1_time`/`bpo2_time` in `amsterdam_chain_config`: `resolves_the_bpo2_blob_schedule_at_amsterdam` fails, and the stateless run drops 18 files.

**Checklist**

Closes #7152
Closes #7150


tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -30,75 +26,11 @@ const SKIPPED_BASE: &[&str] = &[
...
-    "test_witness_state_sstore_into_empty_storage_omits_post_state_nodes",
-    "test_witness_state_sstore_new_slot_omits_post_state_nodes",
-    "test_witness_state_sstore_without_explicit_read_contains_storage_proof",
-];
-#[cfg(not(feature = "stateless"))]
+// Nothing extra to skip in either run. tests-zkevm@v0.8.0 shares the devnet-8
+// gas schedule this client implements, so the spec-base skew that forced a
+// 45-entry stateless skip list against the old generated set is gone. Anything
+// failing is a real bug — do not re-add a skip without recording why in
+// docs/known_issues.md.
```
MegaRedHand:
```suggestion
// Do not re-add a skip without recording why in docs/known_issues.md.
```

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -2,19 +2,15 @@ use ef_tests_blockchain::test_runner::parse_and_execute;
...
-// directory, neither of which this harness can parse.
-//
-// It must be the generated set, not the downloaded `vectors_zkevm/` one. That
-// bundle is tests-zkevm@v0.6.2, which predates execution-specs #3278 while
-// reusing schema id 0x1501, so its `statelessOutputBytes` are 69 bytes against
-// the 43 this client now produces — every wire-path comparison would fail.
+// v0.8.0 is the first zkEVM release that carries both the EIP-8025 stateless
+// schema (#3248 progressive SSZ, #3278 `ChainConfig` removal, #3356
+// `ProgressiveList` witness fields) and the glamsterdam-devnet-8 gas schedule
+// this client implements. Earlier bundles had at most one of the two.
```
MegaRedHand:
```suggestion
```

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -2,19 +2,15 @@ use ef_tests_blockchain::test_runner::parse_and_execute;
 use std::path::Path;
 
 // test-levm reads snobal-devnet-6 + legacy from `vectors/`; test-stateless reads
-// the generated #3248+#3278 conformance vectors from a separate root so the two
-// do not overlay each other. See the Makefile for how each is populated.
+// tests-zkevm@v0.8.0 from `vectors_zkevm/` so the two do not overlay each other.
+// See the Makefile for how each is populated.
 //
```
MegaRedHand:
```suggestion
```

tooling/ef_tests/blockchain/tests/all.rs
```diff
@@ -2,19 +2,15 @@ use ef_tests_blockchain::test_runner::parse_and_execute;
 use std::path::Path;
 
 // test-levm reads snobal-devnet-6 + legacy from `vectors/`; test-stateless reads
```
MegaRedHand: Is this comment up to date?

docs/known_issues.md
```diff
@@ -33,62 +33,28 @@ marks these cases `speconly` so they are type-checked instead of compared byte-f
...
+Until `tests-zkevm@v0.8.0` there was no published bundle carrying both the
+EIP-8025 stateless schema and the glamsterdam-devnet-8 gas schedule, so the
+vectors were generated locally from execution-specs `3c3b6f4af` — which is
+devnet-7. 45 fixtures failed in `add_block_pipeline` with `GasUsedMismatch` or
+`ReceiptsRootMismatch` on the gas difference alone (`COLD_STORAGE_ACCESS`
+3000 vs 2100, `ACCOUNT_WRITE` 8000 vs 9000, `CALL_VALUE` 10300 vs 11300,
+access-list entries at full cold cost vs cold − `WARM_ACCESS`) and were skipped
+by name. v0.8.0 is filled against `tests-glamsterdam-devnet@v8.1.0`, the same
+base this client targets, so the split is gone and the generator, its `uv` and
+Python 3.12 prerequisites, and the whole skip list were removed with it.
```
MegaRedHand: I think we can remove this

```suggestion
```

docs/known_issues.md
```diff
@@ -33,62 +33,28 @@ marks these cases `speconly` so they are type-checked instead of compared byte-f
 
 ---
 
-### Stateless EF tests: 45 of 105 fixtures skipped
+### Stateless EF tests: no skips
```
MegaRedHand: We should remove this whole section, since it's no longer an issue

MegaRedHand (approved):
LGTM. Highlighted some docs that can be removed/simplified.

PR #7158, opened 2026-08-18
feat(l1): add ethrex-evm t8n CLI for filling EELS tests

spencer-tb:
**Motivation**

Let EELS fill test fixtures through LEVM (`fill --evm-bin=ethrex-evm`).

**Description**

Adds an `ethrex-evm` binary with a geth compatible `t8n` subcommand, built on the same payload pipeline as `testing_buildBlockV1` via a new skip and record build entry point. A separate commit fixes the block blob cap in explicit payload builds, which was unenforced because sidecar less blob transactions never grew the blobs bundle.

**Why?**

Our benchmark test releases need to be filled outside of EELS as our t8n/filling is too slow, especially once we get to 100Ms of block gas limit. Currently we fill our benchmark test releases (compute) with geth's t8n tool, the same for zkevm benchmark test releases.

Before we can fill new devnet **benchmark** tests, the client we fill our tests with must pass all the new denvet **spec** tests. Given that ethrex is always the first to pass these tests, and is faster to fill vs other clients, it would be great to use you as the defacto client for filling compute benchmark tests for future releases.

The added benefit here is testing block building of the **spec** tests, as the t8n tool uses the testing build block function as an entry point for filling the tests.

**Checklist**


ilitteri:
LGTM! One minor organizational change, let's move this binary to the `tooling` workspace.

PS: We have goevmlab integration as a `tooling/ef_tests/state_v2` subcommand. If you feel like it's a good place for this to live as a subcommand of `statetest` add it there, and if not, just add it as `tooling/ethrex-evm`.

spencer-tb:
Moved to `tooling/ethrex-evm`! Thanks :)

ilitteri (approved):

spencer-tb:
CI should be fixed on latest commit

PR #7160, opened 2026-08-19
fix(l1): return the executable minimum from eth_estimateGas

ilitteri:
**Motivation**

Differential testing across the six clients found `eth_estimateGas` reporting above the
executable minimum on every path:

| case | minimum | ethrex reported | over |
| --- | --- | --- | --- |
| EIP-2780 self-send | 12 000 | 12 156 | +156 |
| zero-value → existing | 15 000 | 15 159 | +159 |
| transfer → existing | 21 000 | 21 165 | +165 |
| transfer → fresh | 204 600 | 207 391 | +2 791 |

The report attributed the margin to `COLD_ACCOUNT_ACCESS`, but it is not repricing-related.
The tests below reproduce the identical +165 on the Osaka test genesis, which has neither
EIP-2780 nor EIP-8038, and replaying the search arithmetic reproduces all four numbers
exactly. The cause is that the search stopped once its bracket was within
`ESTIMATE_ERROR_RATIO` (1.5%) of the upper bound and then returned that upper bound. The
error is proportional, so it was simply most visible where EIP-2780 made the intrinsic cost
small — which is that EIP's whole point.

geth runs the same loop with the same constant, and returns the same 207 391 on the last
row. What hides it on the first three is a short circuit before the estimator:

```go
// If the transaction is a plain value transfer, short circuit estimation and
// directly try 21000.
if len(call.Data) == 0 {
    if call.To != nil && opts.State.GetCodeSize(*call.To) == 0 {
        failed, _, err := execute(ctx, call, opts, params.TxGas)
        if !failed && err == nil {
            return params.TxGas, nil, nil
        }
    }
}
```

That is a hardcoded `params.TxGas`, which is exactly what EIP-2780 invalidated: it is why
geth answers a self-send with 21 000 when the minimum is 12 000, a 75% overestimate. So the
report's "legacy 21000 floor" is right about the symptom but not the mechanism — nothing
clamps upward, the estimator simply never runs.

**Description**

Nothing can succeed below what an unconstrained run consumed, and most transactions execute
identically at their own `gas_used` as with the whole block's gas available. So the estimate
re-runs once at exactly `gas_used` and returns it when that succeeds: two simulations, exact
by construction, no search. Measured on a plain transfer:

| | simulations | answer |
| --- | --- | --- |
| before | 5 | 21 165 |
| this PR | **2** | **21 000** |

Cheaper than the behaviour it replaces, and correct — which answers the review question
about added iterations.

This is the same idea as geth's short circuit, generalised. geth probes a fixed
`params.TxGas` and only for a data-less call to a codeless account; probing the run's own
`gas_used` costs the same single simulation, is correct at any fork, and covers every
transaction shape.

The fallback search keeps geth's tolerance. `ESTIMATE_ERROR_RATIO` is deliberate there, and
geth writes down why: a perfect estimate is not worth extra execution when callers bump by
20-25% anyway, and for a transaction that inspects its own remaining gas it "probably
doesn't want to return the lowest possible gas limit for these cases anyway". Those are
precisely the transactions that reach the fallback, so it bisects and stops within the ratio
rather than converging.

Also pins `eth_createAccessList` returning an empty list for an access that touches an
account but no storage slot. Under EIP-8038 an entry costs the cold charge minus
`WARM_ACCESS`, so prepaying is gas-neutral for a single access while the entry's floor
tokens still raise the EIP-7623 floor — the minimal list is the empty one, and the same
testing found four of six clients returning a costlier list. ethrex already returns the
empty list, but incidentally: `Substate::make_access_list` derives entries from
`accessed_storage_slots`, so an address with no storage access cannot appear, and no
profitability heuristic tracks the pricing. The test is what keeps that true if the pricing
moves again.

A third finding from the same testing is not addressed here: `eth_simulateV1` with
`traceTransfers` double-counts EIP-7708 transfer logs across clients. ethrex answers
`-32601` because the method is unimplemented (#6212), so there is nothing to fix yet, but it
is a design requirement for #6951 — and the reporter is right that EIP-7708 and the
`traceTransfers` flag need an explicit rule about which log wins.

**L2: estimating against the fees the transaction will pay**

Making the estimate exact surfaced an older bug, as `test_fee_token: Deploy transaction
failed` across the L2 integration jobs. It took two goes to find, and the first answer was
only half of it.

`EthClient::estimate_gas` hand-builds its call object and sent only
`to`/`input`/`from`/`value` (plus blobs and nonce). With no fee field the server backfills
`env.gas_price` to zero — and on an L2 a zero gas price makes `adjust_disabled_l2_fees` drop
the fee configs outright, so the estimate is computed with **no L1 data fee gas at all**.
`build_generic_tx` submits the estimate verbatim, so the transaction ran out of gas at
exactly its own estimate. The call object now carries `maxFeePerGas`,
`maxPriorityFeePerGas`, `maxFeePerBlobGas` and a non-zero `gasPrice`; a zero `gasPrice` is
still omitted rather than sent as `0x0`, so a genuinely fee-less call keeps reaching the
relaxed path that exists for it.

The second half is the size that fee is computed from. The L2 prices its L1 data fee from
`Transaction::length()`, but `generic_tx_to_transaction` built the simulated transaction with
`..Default::default()` for everything except `to`/`value`/`data`/`access_list`, and a zero
`U256` is one RLP byte where a real signature component is 33 — a plain deploy encoded **15
bytes against the signed transaction's 93**. That only becomes load-bearing once the fee
configs are enabled at all, which is why fixing it first changed nothing. The envelope is
now carried over and the signature stamped at full width, deliberately at the maximum: a
real component can be shorter after leading-zero trimming, and over-reserving the L1 fee is
safe where under-reserving is not.

Populating `gas_limit` regressed `eth_call` before a test caught it, which is worth naming:
the hooks read `vm.tx.gas_limit()` for the EIP-7825 per-transaction cap, so a call passing
more than 16 777 216 gas — tools routinely pass the block gas limit — began failing with
`TxMaxGasLimitExceeded` where it used to simulate. That cap gates a transaction's admission,
not its execution, so it now honours the same `disable_gas_allowance_check` as the
block-level allowance, with the flag's documentation covering both. Real transactions are
unaffected: only the simulation entry points set the flag, and the 18 EIP-7825 fixtures in
the blockchain suite still pass.

**Tests**

`estimate_gas_returns_the_exact_minimum_for_a_transfer` and `..._for_a_zero_value_call` pin
the reported case; both fail before this change with `left: 21165, right: 21000`.

`estimate_gas_for_a_gas_observing_call_is_executable_and_within_the_tolerance` covers the
fallback, so it is not untested. A predeploy running `GAS; PUSH2 0x2710; GT; JUMPI` into
`INVALID` fails at its own `gas_used`, forcing the search, and the test pins the one-sided
contract the search owes: the estimate is executable, never sits below the true minimum —
found by bisection — and stays within the ratio above it.

`create_access_list_omits_an_account_only_access` pins the empty access list.

`simulated_tx_never_encodes_shorter_than_the_signed_one` pins the L2 invariant across an
empty, a small and a 4 KiB payload; without the fix it reports 15 bytes against 93.
`eth_call_tolerates_gas_above_the_per_tx_cap` pins the regression the cap gating avoids.
`carries_the_fees_the_transaction_will_pay` and `carries_gas_price_only_when_set` pin the
estimate's call object, which is where this went wrong unseen — the JSON construction is
split out as `estimate_gas_call_object` so the field set can be asserted on rather than
inspected by eye.

Regression sweep: 11752 blockchain ef-tests (including the 18 EIP-7825 fixtures), 449 levm
and rpc integration tests, 11 vm and 18 levm unit tests, 106 rpc unit tests.

Already on `glamsterdam-devnet-8` (`da0d87c30`).

### DIFF
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -515,6 +515,23 @@ impl RpcHandler for EstimateGasRequest {
         let gas_used = result.gas_used();
         let gas_refunded = result.gas_refunded();
 
+        // Most transactions execute identically at their own `gas_used` as they do with
+        // the whole block's gas available, and nothing can succeed below what an
+        // unconstrained run consumed — so if that one re-run succeeds it *is* the minimum,
+        // and the search can be skipped entirely. Two simulations, exact. Only gas-observing
+        // callers (an explicit `GAS` check, or a subcall needing 63/64 headroom the
+        // consumed total does not imply) fall through to the search below.
+        transaction.gas = Some(gas_used);
+        if let Ok(ExecutionResult::Success { .. }) = simulate_tx(
+            &transaction,
+            &block_header,
+            storage.clone(),
+            blockchain.clone(),
+        ) {
+            return serde_json::to_value(format!("{gas_used:#x}"))
+                .map_err(|error| RpcErr::Internal(error.to_string()));
+        }
+
         // Choose an optimistic start limit. See https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/eth/gasestimator/gasestimator.go#L135
         let optimistic_limit = (gas_used + gas_refunded + CALL_STIPEND) * 64 / 63;
         let mut lowest_gas_limit = gas_used.saturating_sub(1);
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -24,6 +24,13 @@ use serde::Serialize;
 use serde_json::Value;
 use tracing::debug;
 
+/// Allowed upward overestimation before the estimate's binary search stops, matching
+/// geth's `estimateGasErrorRatio` (`internal/ethapi/api.go`). geth's rationale is that a
+/// perfect estimate is not worth extra execution when callers bump by 20-25% anyway, and
+/// that for a transaction which inspects its own remaining gas the true minimum is not
+/// even the useful answer. Only reached when the fast path below cannot answer.
+pub const ESTIMATE_ERROR_RATIO: f64 = 0.015;
+
 pub const CALL_STIPEND: u64 = 2_300; // Free gas given at beginning of call.
 pub const TRANSACTION_GAS: u64 = 21_000; // Per transaction not creating a contract. NOTE: Not payable on data of calls between transactions.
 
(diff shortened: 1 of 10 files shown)


crates/networking/rpc/eth/transaction.rs
```diff
@@ -521,13 +520,14 @@ impl RpcHandler for EstimateGasRequest {
...
 
+        // Converge all the way to `lowest + 1 == highest`, so the returned
+        // `highest_gas_limit` is the least gas the transaction actually executes with.
+        // This used to stop once the bracket was within 1.5% of the upper bound (geth's
+        // `estimateGasErrorRatio`) and return that bound, which reported a plain transfer
+        // as 21165 instead of 21000 and, on the EIP-2780 paths where the intrinsic cost is
+        // lower, a self-send as 12156 instead of 12000. Callers that want headroom can add
+        // it; a caller that wants the minimum cannot recover it from an upper bound.
         while lowest_gas_limit + 1 < highest_gas_limit {
-            if (highest_gas_limit - lowest_gas_limit) as f64 / (highest_gas_limit as f64)
```
MegaRedHand, line 525: Are we sure this doesn't add that many more iterations? Maybe we could instead simulate the transaction with the bare minimum gas first, and early return that estimate if successful, otherwise keeping the old 1.5% error rate estimate.

MegaRedHand (approved):

PR #7162, opened 2026-08-19
perf(l1): track trie path position with a cursor instead of a second vector

diegokingston:
**Motivation**

`Nibbles` keeps the traversal position by moving nibbles between two vectors: `data` holds what is left to match, `already_consumed` what has been walked past. Their concatenation is invariant, so every advance pays to shift bytes between them:

```rust
next()        -> data.remove(0)   // memmove of the whole remaining path
skip_prefix() -> data.drain(..n)  // memmove, plus extend of already_consumed
offset(n)     -> two fresh vectors and two copies
```

`next()` runs once per branch descended, so walking a 64-nibble path does 64 memmoves of up to 64 bytes to consume 64 nibbles — quadratic work for what is logically an index bump.

**Description**

Keep one buffer and a cursor: `data[..cursor]` is consumed (what `current()` returns) and `data[cursor..]` remains. Advancing becomes an integer add, and `offset` allocates once instead of twice.

The public API and its semantics are unchanged. Every method that reads "the nibbles" still means the remaining ones — `len`, `at`, `is_leaf`, `AsRef<[u8]>`, the comparison impls and the compact encoding.

**Measurements**

On zkVM guests, where allocation and memmove are executed work:

- **ZisK**, trie-heavy workload (1200-leaf trie, 1200 lookups, full hash): 6,956,595 → 6,751,333 steps (**−2.95%**), `dma_memcpy` −22.8%, RAM use −7.4%. The keccak count is identical before and after, so no work is skipped, and the guest's committed output is byte-identical.
- **LambdaVM**, transfer blocks: −0.26% to −0.74%. Small, and worth stating plainly: those blocks barely touch the trie and are dominated by signature recovery, so there is little for this to win.
- **Host**, 1200-account trie: 12.5% fewer allocations.

The argument for the change is mostly the algorithmic one — it removes a quadratic memmove from every traversal — rather than the size of the win on transfer-heavy blocks.

**Validation**

The existing suite passes (61 trie tests, plus `ethrex-storage` and `ethrex-common`), `clippy` clean.

Beyond that, a differential harness compared this against a verbatim transcription of the previous implementation: **115,975 randomized operations over 3,161,843 assertions, no divergence**. It covers `next`, `skip_prefix`, `offset`, `slice`, `concat`, `append_new`, `current`, `take`, `prepend`, `extend`, the ordering/equality/hash impls and the compact encoding.

**Note for integrators**

`Nibbles` derives `rkyv::Archive`, so replacing `already_consumed: Vec<u8>` with `cursor: usize` changes the archived layout. Anything holding pre-serialized `ExecutionWitness` bytes has to regenerate them; host and guest built together are unaffected. As a side effect the serialized witness gets slightly smaller.

This touches the same file as #7155, so whichever lands second will need a trivial rebase.

### DIFF
--- a/crates/common/trie/nibbles.rs
+++ b/crates/common/trie/nibbles.rs
@@ -396,9 +396,21 @@ fn count_common_prefix_scalar(a: &[u8], b: &[u8]) -> usize {
 }
 // ─────────────────────────────────────────────────────────────────────────────
 
-// TODO: move path-tracking logic somewhere else
 // PERF: try using a stack-allocated array
 /// Struct representing a list of nibbles (half-bytes)
+///
+/// The path and the traversal position are kept in a single buffer: `data` holds
+/// the whole path and `cursor` splits it into the part already consumed by a
+/// traversal (`data[..cursor]`, see [`Self::current`]) and the part still to be
+/// matched (`data[cursor..]`, the *remaining* nibbles).
+///
+/// Every method that reads "the nibbles" means the remaining ones — `len`, `at`,
+/// `is_leaf`, `AsRef<[u8]>`, the comparison impls and the compact encoding all
+/// operate on `data[cursor..]`.
+///
+/// Keeping one buffer is what makes advancing free: consuming a nibble or a
+/// prefix moves `cursor` instead of shifting the rest of the path down. In the
+/// zkVM guest that matters twice over, since the bump allocator never reclaims.
 #[derive(
     Debug,
     Clone,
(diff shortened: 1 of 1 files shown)


crates/common/trie/nibbles.rs
```diff
@@ -661,50 +690,46 @@ impl Nibbles {
...
 impl RLPEncode for Nibbles {
     fn encode(&self, buf: &mut dyn bytes::BufMut) {
-        Encoder::new(buf).encode_field(&self.data).finish();
+        // Only the remaining nibbles round-trip; `decode` yields them with a
+        // fresh cursor. Avoid the copy in the common, already-normalized case.
+        if self.cursor == 0 {
+            Encoder::new(buf).encode_field(&self.data).finish();
+        } else {
+            Encoder::new(buf)
+                .encode_field(&self.as_ref().to_vec())
```
MegaRedHand, line 722: Does this need the `to_vec`?

crates/common/trie/nibbles.rs
```diff
@@ -396,9 +396,21 @@ fn count_common_prefix_scalar(a: &[u8], b: &[u8]) -> usize {
 }
 // ─────────────────────────────────────────────────────────────────────────────
 
-// TODO: move path-tracking logic somewhere else
```
MegaRedHand, line 399: This is still valid

MegaRedHand (approved):
LGTM. There's probably a better way to address this issue, though

ilitteri (approved):

PR #7165, opened 2026-08-19
feat(l1): track the EEST zkEVM benchmark suite as bench workloads

ilitteri:
**Motivation**

[`tests-zkevm-benchmark@v0.8.2`](https://github.com/ethereum/execution-specs/releases/tag/tests-zkevm-benchmark%40v0.8.2) is upstream's first benchmark fixture release — Amsterdam compute benchmarks at 10M, 30M and 60M gas, produced entirely in the EEST pipeline on the same `tests-zkevm@v0.8.0` base the stateless harness already reads.

The reason to take it is where the witnesses come from. `tooling/zkevm_bench`'s `stress` tier is generated locally using **ethrex's own witness generation**, which makes those numbers self-referential: a witness-generation bug would be baked into the baseline it is measured against. These fixtures carry upstream's witnesses, so the AIR costs are comparable against other zkEVM clients rather than only against ourselves.

**Description**

Pinned by `tooling/ef_tests/.fixtures_url_zkevm_benchmark`, fetched on demand:

```bash
make -C tooling/ef_tests/blockchain zkevm-benchmark-vectors
```

Deliberately **not** a `download-test-vectors` prerequisite — the bundle is ~500MB and only the bench reads it. The extracted root is gitignored, same treatment as `vectors_zkevm`.

Fifteen workloads are wired into `fixtures/manifest.toml`: five areas (keccak, bn128 pairings, p256verify, EIP-7928 block access lists, jumpdest analysis) across the three gas tiers. They reuse the existing `micro` loader, which already reads EEST fixtures, so there is no new loading path. Tiers map onto run modes so a mode's cost stays proportional — 10M in `medium`, 30M and 60M in `slow`.

No client changes were needed. The bundle's schema matches what we already consume: `0x1501` prefix, 43-byte outputs, and all 7390 blocks carrying both `statelessInputBytes` and `executionWitness`.

**Validation**

Ran the release against ethrex before wiring it in:

- **all 3464 tests pass the stateless wire path** — decode `statelessInputBytes`, run the guest, compare the 43-byte `statelessOutputBytes`
- the full 339-file set runs **339 passed / 0 failed** through the EF harness (needs #7164, which stops the harness demanding our generated witness equal the fixture's; without it 9 files fail on a difference the spec explicitly permits)
- tier medians land exactly on 10M / 30M / 60M gas

Two observations reported upstream rather than worked around: some blocks run well above their tier (largest is 613,958,400 gas in the 10M tier, ~61×, all cold-access benchmarks that look like state-priming setup blocks), and `gasLimit` is `1e12` on every header, which leaves gas-limit-derived logic such as EIP-7928's BAL size cap unexercised.

Upstream flagged the release as experimental, so the tag is pinned exactly and churn is expected.

**How to test**

```bash
make -C tooling/ef_tests/blockchain zkevm-benchmark-vectors
cargo run --manifest-path tooling/zkevm_bench/Cargo.toml -- run --mode medium
```

AIR-cost figures need `ziskemu` on x86-64; the validation above was native execution.

**Checklist**

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,6 +29,7 @@ hive/
 ethereum-package/
 tooling/ef_tests/blockchain/vectors
 tooling/ef_tests/blockchain/vectors_zkevm
+tooling/ef_tests/blockchain/vectors_zkevm_benchmark
 tooling/ef_tests/state/vectors
 tooling/sync/multisync_logs/
 dev_ethrex_l1/
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 
 tooling/ef_tests/blockchain/vectors
 tooling/ef_tests/blockchain/vectors_zkevm
+tooling/ef_tests/blockchain/vectors_zkevm_benchmark
 
 tooling/ef_tests/state/vectors
 
--- a/tooling/ef_tests/.fixtures_url_zkevm_benchmark
+++ b/tooling/ef_tests/.fixtures_url_zkevm_benchmark
@@ -1,0 +1,1 @@
+https://github.com/ethereum/execution-specs/releases/download/tests-zkevm-benchmark%40v0.8.2/fixtures_zkevm-benchmark.tar.gz
--- a/tooling/ef_tests/blockchain/Makefile
+++ b/tooling/ef_tests/blockchain/Makefile
@@ -40,11 +40,14 @@ ZKEVM_FIXTURES_FILE := ../.fixtures_url_zkevm
 ZKEVM_ARTIFACT := zkevm-tests.tar.gz
 ZKEVM_URL := $(shell cat $(ZKEVM_FIXTURES_FILE))
 
-# tests-zkevm-benchmark@v0.8.2 — Amsterdam compute benchmarks at 10M/30M/60M gas,
-# filled by the EEST pipeline on the same tests-zkevm@v0.8.0 base, so the schema
-# matches what the stateless harness already reads. Kept in its own root and NOT
-# a `download-test-vectors` prerequisite: it is ~500MB and only the zkevm bench
-# reads it. Fetch on demand with `make zkevm-benchmark-vectors`.
+# tests-zkevm-benchmark@v0.8.2 — Amsterdam compute benchmarks at 10M/30M/60M gas.
+# Filled by the EEST pipeline at execution-specs 117dd1cf with go-ethereum's
+# `evm` as t8n, not the EELS t8n that produced tests-zkevm@v0.8.0 (0695c34c);
+# the two pins move independently despite the shared "zkevm" name.
+#
+# Kept in its own root and NOT a `download-test-vectors` prerequisite: 519MB
+# compressed and ~3.9GB extracted, and only the zkevm bench reads it. Fetch on
+# demand with `make zkevm-benchmark-vectors`.
 ZKEVM_BENCH_VECTORS_ROOT := vectors_zkevm_benchmark
 ZKEVM_BENCH_VECTORS_DIR := $(ZKEVM_BENCH_VECTORS_ROOT)/eest
 ZKEVM_BENCH_FIXTURES_FILE := ../.fixtures_url_zkevm_benchmark
@@ -106,13 +109,41 @@ $(ZKEVM_VECTORS_DIR): $(ZKEVM_ARTIFACT)
 $(ZKEVM_BENCH_ARTIFACT): $(ZKEVM_BENCH_FIXTURES_FILE)
 	$(DOWNLOAD) $(ZKEVM_BENCH_URL) $(ZKEVM_BENCH_ARTIFACT)
 
-# Extracts all three gas tiers (for_amsterdam_at_{0010,0030,0060}M).
-$(ZKEVM_BENCH_VECTORS_DIR): $(ZKEVM_BENCH_ARTIFACT)
-	rm -rf $(ZKEVM_BENCH_VECTORS_DIR)
-	mkdir -p $(ZKEVM_BENCH_VECTORS_DIR)
-	tar -xzf $(ZKEVM_BENCH_ARTIFACT) --strip-components=2 -C $(ZKEVM_BENCH_VECTORS_DIR) fixtures/blockchain_tests
+# Only the subtrees the manifest references, not all 339 fixture files: the
+# whole tree unpacks to ~3.9GB for the 15 workloads actually run.
+ZKEVM_BENCH_SUBTREES := \
+  compute/instruction/keccak \
+  compute/precompile/alt_bn128 \
+  compute/precompile/p256verify \
+  compute/eip7928_block_level_access_lists/block_access_list \
+  compute/scenario/mix_operations
+ZKEVM_BENCH_TIERS := for_amsterdam_at_0010M for_amsterdam_at_0030M for_amsterdam_at_0060M
 
-zkevm-benchmark-vectors: $(ZKEVM_BENCH_VECTORS_DIR) ## 📥 Download the zkEVM benchmark fixtures (~500MB)
+# Keyed on the release URL, like `amsterdam-vectors`, and written only after a
+# successful untar. Two things this buys: `clean-vectors` runs from the
+# `$(SPECTEST_ARTIFACT)` rule on any `.fixtures_url` bump, which would otherwise
+# silently discard this opt-in multi-GB download; and an interrupted extraction
+# leaves no stamp, so the next run redoes it instead of reporting "nothing to be
+# done" while workloads fail on missing files.
+ZKEVM_BENCH_STAMP := $(ZKEVM_BENCH_VECTORS_DIR)/.zkevm_benchmark_url
+
+zkevm-benchmark-vectors: ## 📥 Download the zkEVM benchmark fixtures (519MB compressed, ~3.9GB extracted)
+	@have=$$(cat $(ZKEVM_BENCH_STAMP) 2>/dev/null); \
+	if [ "$$have" = "$(ZKEVM_BENCH_URL)" ]; then \
+		echo "zkEVM benchmark fixtures already extracted from $(ZKEVM_BENCH_URL)"; \
+	else \
+		$(MAKE) $(ZKEVM_BENCH_ARTIFACT) && \
+		rm -rf $(ZKEVM_BENCH_VECTORS_DIR) && \
+		mkdir -p $(ZKEVM_BENCH_VECTORS_DIR) && \
+		members=""; \
+		for tier in $(ZKEVM_BENCH_TIERS); do \
+			for sub in $(ZKEVM_BENCH_SUBTREES); do \
+				members="$$members fixtures/blockchain_tests/$$tier/$$sub"; \
+			done; \
+		done; \
+		tar -xzf $(ZKEVM_BENCH_ARTIFACT) --strip-components=2 -C $(ZKEVM_BENCH_VECTORS_DIR) $$members && \
+		echo "$(ZKEVM_BENCH_URL)" > $(ZKEVM_BENCH_STAMP); \
+	fi
 
 zkevm-vectors: $(ZKEVM_VECTORS_DIR)
 
--- a/tooling/ef_tests/blockchain/Makefile
+++ b/tooling/ef_tests/blockchain/Makefile
@@ -1,4 +1,4 @@
-.PHONY: download-test-vectors clean-vectors test test-levm test-stateless amsterdam-vectors zkevm-vectors
+.PHONY: download-test-vectors clean-vectors test test-levm test-stateless amsterdam-vectors zkevm-vectors zkevm-benchmark-vectors
 
 VECTORS_ROOT := vectors
 FIXTURES_FILE := ../.fixtures_url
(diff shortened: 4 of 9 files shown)


MegaRedHand (commented):
Traced the 15 new workloads end to end and checked the pinned bundle against the release itself. Two issues block this as it stands; the rest are ranked after them.

## 1. `tier = "slow"` makes the manifest unloadable in every run mode

`Tier::parse` (`tooling/zkevm_bench/src/manifest.rs:48`) accepts only `"quick"` and `"medium"`, and `load_manifest` hard-bails on anything else rather than defaulting. Ten of the new workloads declare `tier = "slow"`, and `manifest.rs` is unchanged in this PR, so there is no `Tier::Slow` variant to accept it.

`run_bench` calls `load_manifest` at `run.rs:126`, before any mode filtering, so this is not scoped to the slow entries:

```
main:      LOADED OK: 20 workloads
this PR:   LOAD FAILED: workload "eest_bench_keccak_30M" has unknown tier "slow"
                        (expected "quick" or "medium")
```

Every `run` invocation fails, `--mode quick` included, and so does the `How to test` command in the PR description. CI does not catch it: the prover workflow only runs `cargo clippy` on the crate and diffs `Cargo.lock`, nothing executes the bench.

Affected lines: `manifest.toml` 207, 214, 221, 228, 235, 242, 249, 256, 263, 270.

Worth noting `includes_tier`'s defensive branch already does what you want here (an unparseable tier is visible only under `RunMode::Slow`), but `load_manifest` rejects the value before it can ever reach that branch. So the fix is either a real `Tier::Slow` variant wired through `parse` and `includes_tier`, or retagging those ten as `medium`.

## 2. None of the 15 workloads executes the block it is meant to measure

`micro.rs:66` passes `bal: None`, and for Amsterdam blocks that is fatal well before the EVM runs:

```
micro.rs:66          build_ssz_stateless_input(..., bal: None)
  -> l1_advancer.rs:263  bal_to_ssz_block_access_list(None) = empty ProgressiveList
  -> program.rs:160      `if !payload.block_access_list.is_empty()` is false, so
                         header.block_access_list_hash stays None
  -> execution.rs:121    validate_block_pre_execution runs BEFORE vm.execute_block
  -> block.rs:849        is_amsterdam_activated && hash.is_none()
                         => Err(BlockAccessListHashNotPresent)
```

Checked against a real fixture from the pinned bundle (`for_amsterdam_at_0060M/compute/instruction/control_flow/jumpis.json`, `network: Amsterdam`). The data needed is present in band and simply not deserialized by `MicroBlock`:

| Field on each block | Read by `MicroBlock`? |
| --- | --- |
| `blockAccessList` (9 entries here) | no |
| `blockHeader.blockAccessListHash` | no |
| `statelessInputBytes` | no |
| `rlp`, `executionWitness` | yes |

It also fails silently. `run_stateless_guest` is documented to never panic and never return an error, committing `successful_validation` as an output byte instead, and `execute_profiled` (`prover/src/backend/zisk.rs:144`) only checks ziskemu's exit status. So ziskemu exits 0, `run.rs:167` records `guest_output_ok: true`, and the report carries small, stable AIR costs that reflect witness load plus header validation rather than keccak, bn128, p256verify, BAL or jumpdest work at 10M/30M/60M gas. `compare` will then gate on those numbers.

Both halves of this are pre-existing in `micro.rs` and `run.rs`, not introduced here, and the single existing `micro` entry is Amsterdam too, so it has the same problem. Flagging it on this PR because the PR is built entirely on that path and takes the affected surface from 1 workload to 16, and because the numbers it advertises as comparable against other zkEVM clients are the ones affected.

Two ways out, both already available in the tree:

- deserialize `blockAccessList` and pass it through to `build_ssz_stateless_input`, or
- consume the fixture's own `statelessInputBytes` and check the committed validity byte, which is what `run_stateless_from_input_bytes` and `parse_expected_valid_flag` (`tooling/ef_tests/blockchain/test_runner.rs:943` and `:977`) already do.

The second also removes a class of divergence rather than one instance of it: those bytes are exactly what another client would consume, so re-deriving them host side is what let the missing BAL slip in.

## 3. The provenance claim does not match what the bundles say

The comment states the bundle is "filled by the EEST pipeline on the same `tests-zkevm@v0.8.0` base". Pulled each bundle's own `fixtures/.meta/fixtures.ini`:

| Bundle | commit | t8n |
| --- | --- | --- |
| `tests-zkevm@v0.8.0` (what `.fixtures_url_zkevm` pins) | `0695c34c` | `2.19.0` (EELS) |
| `tests-zkevm@v0.8.2` | `117dd1cf` | `2.19.0` (EELS) |
| `tests-zkevm-benchmark@v0.8.2` (this PR) | `117dd1cf` | `evm 1.17.6-unstable` (go-ethereum) |

So the matching base is v0.8.2, not v0.8.0, and even against v0.8.2 the witnesses come from geth's t8n rather than EELS. Schema compatibility clearly holds in practice given your validation run, so this is a provenance and drift point rather than breakage: nothing ties the benchmark pin to the zkevm pin, so they can move independently while the comment says they cannot. Same claim appears at `manifest.toml:159` and `README.md:35-36`.

## 4. The 10M entries put an opt-in download in the default mode

The five 10M workloads are `tier = "medium"`, and `medium` is the clap default. But `zkevm-benchmark-vectors` is deliberately not a `download-test-vectors` prerequisite, unlike `zkevm-vectors`, which is what makes the existing `micro` entry work there.

So after a plain `make download-test-vectors` plus `run`, those five hit missing files, get `guest_output_ok: false`, and `compare_reports` (`compare.rs:48`) forces `regressed: true` for each. A developer cannot distinguish "did not download 500MB of fixtures" from a real AIR-cost regression. Tagging all fifteen slow (once a slow tier exists) or documenting the download as a `medium` prerequisite would both resolve it.

## 5. The disk figure is understated by roughly 8x

"~500MB" is the compressed asset. Measured from the release: the asset is 518,692,757 bytes, and the gzip trailer's ISIZE pins the uncompressed tar to 3.94 GB. Two checks select that value over the 4 GiB-wraparound alternatives: streaming a 60MB prefix gives 10.95x, and exact tar header sizes for 72 files (mean 9.21 MB, all from the largest tier) land near the 11.63 MB mean that 3.94 GB implies, whereas 8.24 GB would need the remaining 267 files to average 27 MB.

So the real cost is 519 MB downloaded plus about 3.9 GB extracted, with the tarball retained alongside it. Worth stating both numbers in the Makefile comment and README, since the surrounding text is what justifies keeping this out of the default path.

Also: `tar` is given the whole `fixtures/blockchain_tests` tree (339 fixture files) when the manifest references 15. Passing the member paths, or the five leaf directories per tier, would cut this substantially.

## 6. `clean-vectors` discards the bundle on an unrelated URL bump

`clean-vectors` now removes `vectors_zkevm_benchmark` and `zkevm-benchmark-tests.tar.gz`, and the `$(SPECTEST_ARTIFACT)` rule runs `$(MAKE) clean-vectors` whenever `.fixtures_url` changes. So a routine main-fixtures bump silently throws away the expensive opt-in download that the comment three lines up justifies keeping out of `download-test-vectors`, with no message saying why. A URL-keyed stamp, the way `amsterdam-vectors` handles this, would avoid it.

## 7. Only one test per file is measured, chosen by name ordering

`micro_to_program_input` parses into `BTreeMap<String, MicroTest>` and takes `.next()`, so it measures whichever pytest id sorts first. The referenced files are not single-case:

| workload | cases in file |
| --- | --- |
| keccak (each tier) | 18 |
| jumpdest_analysis | 9 |
| mixed_dependency_graph | 7 to 8 |
| p256verify | 3 |
| bn128_pairings_amortized | 1 |

`eest_bench_keccak_*` therefore measures `mem_update_False-offset_0-mem_alloc_b''`, the no-memory-growth variant, and ignores the other 17. Since you expect churn on an experimental tag, an upstream parameter rename that sorts earlier silently changes what a workload name means, and `compare` matches on name alone, so the swap would surface as an unexplained AIR-cost delta and read as a client regression. Selecting by explicit test id, or encoding the variant in the workload name, would pin it.

## 8. Smaller things

- All fifteen omit `gas` while encoding the tier in `category`, so `WorkloadResult.gas` serializes as absent even though the gas target is the defining axis here and is readable from the fixture. Without a denominator there is no AIR-cost-per-gas normalization, which is the natural cross-client comparison. `category` also collapses the operation axis: keccak, bn128, p256verify, bal_mixed and jumpdest at 10M all share `bench-10M`, so nothing groups one operation across tiers.
- `README.md:37` says the 30M and 60M tiers run in `slow`, while the unchanged text at 104 and 113 defines the tier vocabulary as `quick`/`medium` and defines `slow` mode as medium plus `--stress-dir`. The manifest header at line 19 makes the same RunMode/Tier conflation, which is what invited issue 1.
- README Setup step 2 still says `micro` workloads need only `make zkevm-vectors`, so following it verbatim leaves the five default-mode workloads failing.
- An interrupted multi-GB extraction leaves a directory make treats as complete, and later runs report "nothing to be done" while workloads fail on missing files. Note `.DELETE_ON_ERROR` will not reliably help for a directory target; a stamp file written after a successful untar would.
- The 105 lines of near-duplicate TOML differ only in the tier segment, category and tier, so any change to the set has to be made three times. `discover_stress_workloads` plus `--stress-dir` already does directory discovery; generalizing that to `micro`, or supporting a glob in `source`, would make the tier axis data instead of copy-paste. There is also no recorded criterion for why these 5 of the available benchmark names per tier were chosen.

## Verified sound

- The pinned tag, asset and URL resolve correctly: `fixtures_zkevm-benchmark.tar.gz`, 494 MiB, marked prerelease, consistent with the experimental note.
- All 15 declared paths exist. Checked against the bundle's `.meta/index.json` (339 files, 3464 cases, matching the description), and the case each workload actually selects carries the gas value its name claims.
- The tar layout matches `--strip-components=2` with `fixtures/blockchain_tests`, and only the three tier directories live under it, so "extracts all three gas tiers" is accurate.
- The artifact is already covered by the existing `*tests*.tar.gz` ignore rule, and the new vectors root is added to both `.gitignore` and `.dockerignore`.
- Missing fixtures degrade per workload rather than aborting the run, which is the right call even though issue 4 turns it into a forced regression.

MegaRedHand (commented):
Re-checked the last round against `fefd4e3e7`. Issues 1, 3, 4, 5 and 7 are addressed, and I verified the manifest now loads in every mode (`quick` 0 / `medium` 20 / `slow` 35, so the default set is unchanged and the fifteen new entries are opt-in). Three things are left, and the first two are the halves of issue 2 that did not travel.

## 1. `bal: None` is fixed in `micro.rs`, still live in `cache.rs`

Consuming the fixture's own `statelessInputBytes` does fix the fifteen new workloads. But the defect was in the shared re-derivation pattern, not in `micro.rs` specifically, and `cache.rs:78` still passes `None`:

```
cache.rs:78            build_ssz_stateless_input(..., bal: None)
  -> l1_advancer.rs:263  bal_to_ssz_block_access_list(None) = empty ProgressiveList
  -> program.rs:160      `if !payload.block_access_list.is_empty()` is false, so
                         header.block_access_list_hash stays None
  -> block.rs:849        is_amsterdam_activated && hash.is_none()
                         => Err(BlockAccessListHashNotPresent)
```

`validate_stateless_execution` (`program.rs:386`) derives the `ChainConfig` from `(chain_id, Amsterdam)`, so `is_amsterdam_activated` is unconditionally true no matter which fork the cached block belongs to. The committed stress fixtures are Osaka (`amsterdamTime: null` in their embedded `chain_config`), so they cannot carry a BAL hash and cannot pass.

Ran the guest natively over the committed fixtures:

| source | validity byte | reason |
| --- | --- | --- |
| `fixtures/stress/stress_keccak_150M.json.gz` | 0 | `InvalidHeader(BlockAccessListHashNotPresent)` |
| `fixtures/blocks/cache_mainnet_25087668.json.gz` | 0 | `InvalidHeader(BlockAccessListHashNotPresent)` |
| `fixtures/cache/rpc_prover/cache_hoodi_1265656.json` | 0 | `versioned hashes mismatch between NewPayloadRequest and transactions` |

That covers all twenty `real-block` and `stress` entries, which is the whole default `medium` set. So after this PR the manifest is fifteen workloads measuring the intended block and twenty measuring header rejection, and `compare` gates on both without distinguishing them.

The hoodi cache fails for a different reason and deserves its own look: it is the fixture behind the one `cache.rs` unit test, and that test asserts only the two-byte schema id, so it passes green on an input the guest rejects.

## 2. The silent-failure mechanism is unchanged

The input half of issue 2 is fixed for `micro`; the reporting half is untouched, and it is the half that made the first one invisible.

`execute_profiled` (`prover/src/backend/zisk.rs:144`) checks only ziskemu's exit status and never reads the committed output, and `run.rs:172` maps "the closure returned `Ok`" straight to `guest_output_ok: true`. A guest that commits `successful_validation = 0` still lands in the report as a valid measurement carrying a small, stable AIR cost. That is what hid the BAL omission through the original validation run, and it is what still hides the twenty above.

The expected output is now deserialized (`micro.rs:19`) but used only as a skip filter (`micro.rs:93-104`); nothing compares it against what the guest actually committed.

Cheapest fix that needs no new backend plumbing: run `run_stateless_guest` on the host before the profiling run and compare all forty-three bytes, which is what `run_stateless_from_input_bytes` (`test_runner.rs:955`) already does. That also covers the `cache.rs` workloads, which have no fixture-supplied expected output but do have a validity byte worth asserting on.

Failing that, `guest_output_ok` should be renamed to say what it actually means (the emulator exited zero), because as written it reads as a validity signal and `compare` treats it as one.

## 3. `micro.rs` measures only the first block of the selected test

`micro_to_program_input` loops at `micro.rs:80` and returns at the first block carrying `statelessInputBytes` and an expected-valid output. `case` pins which pytest id is measured; nothing pins which block inside it.

By the PR description's own figures the bundle carries 7390 blocks across 3464 cases, roughly 2.13 blocks per case, so multi-block is the norm here rather than the exception. And the description's observation that some 10M-tier blocks reach 613,958,400 gas and "look like state-priming setup blocks" is exactly the shape where the first executable block is the primer and the benchmark is the one after it.

Two ways to pin it, both nearly free:

- `cache.rs:70-76` already refuses to guess for the same wire format, erroring with "cache holds N blocks; the stateless input wire format carries exactly one". `micro.rs` silently takes the first. Making that an error unless the workload opts in would surface any multi-block fixture on the first run instead of leaving it to be inferred from the numbers.
- `micro_to_program_input` receives `gas` and discards it (`micro.rs:56`, `_gas`), while the manifest now declares the tier gas for all fifteen. Asserting the selected block's gas against the declared value would catch a wrong-block pick, and needs one more header field rather than a new code path.

Unrelated but adjacent: `micro.rs:97` treats any validity byte other than `1` as "expected failure, skip", where `parse_expected_valid_flag` (`test_runner.rs:989`) errors on anything that is not `0` or `1`. A malformed byte is currently indistinguishable from a legitimately expected failure.

MegaRedHand (approved):
The issues are pre-existing. We can address them after merging

Issue #3389, opened 2025-06-30
Move discv4 tasks into spawned process


ElFantasma:
We should use `spawned` `GenServer`s to migrate discv4 tasks in the same fashion we are doing that for p2p `RLPxConnection`. That way all communication between `RLPxConnection`s and discv4 processes would be done through spawned messages and concurrency problems will be handled on `spawned` side. Ideally, all tokio references should disappear from ethrex codebase

ElFantasma:
There is already a version of Discv4 on `spawned`  `Genserver`s but it needs some improvements.

Refactor is in progress in `refactor_discv4` branch (it branches from `snap_sync_final`).

PR #3248, opened 2025-06-19
Move ef_tests to tooling folder

guha-rahul:
As part of the effort to clean up the cmd directory, moved the ef_tests directory to tooling. Updated all references to the old path.

**Motivation**

**Description**

Closes #3208

PR #7167, opened 2026-08-20
test(l1): bump zkevm tests to v0.8.2

ilitteri:
**Motivation**

Move the zkEVM EF-test pin to [`tests-zkevm@v0.8.2`](https://github.com/ethereum/execution-specs/releases/tag/tests-zkevm%40v0.8.2), the current release of the stream. Its release notes list three PRs and make no spec-change claims; the tag-to-tag diff confirms them (verified, not assumed — see below). No `v0.8.1` was ever published, so `v0.8.2` is the direct successor to `v0.8.0`.

**Description**

**What the release actually changes.** From the `tests-zkevm@v0.8.0...tests-zkevm@v0.8.2` diff (9 commits, 22 files), read hunk-by-hunk rather than taken from the notes:

- [#3372](https://github.com/ethereum/execution-specs/pull/3372) — renames the Python SSZ classes `Ssz*` → `SSZ*` (`SszStatelessInput` → `SSZStatelessInput`, the `SszList` import alias, etc.) in `stateless_{ssz,guest,host}.py`, the consume-simulator RPC types, and one docstring. SSZ encoding is positional and structural, so class names never reach the wire: no container shape, bound, field order, or schema-id change. The schema stays `0x1501` in the same post-#3356 dialect ethrex already speaks.
- [#3366](https://github.com/ethereum/execution-specs/pull/3366) — CI only (geth branch used for bench-fill comparison).
- [#3389](https://github.com/ethereum/execution-specs/pull/3389) — release automation that introduces a separate `tests-zkevm-benchmark@` release stream. ethrex consumes nothing from that stream; the zkevm_bench micro workloads read the same `vectors_zkevm/` tree this pin populates, so they move together.

Nothing in the diff touches test logic, gas schedules, or fork configuration, which is why this PR contains **no client-code changes** — only the pin and the version-bearing comments/docs around it.

**Fixture-content verification.** Extracting the `for_amsterdam` subtree (the one the stateless harness runs) from both release tarballs and comparing: identical file lists, 3218 JSON files each, and a key-by-key comparison of every fixture finds differences only in `_info.url` — the fill-commit hash in each fixture's source link (`0695c34c1` → `117dd1cf9`, matching each tarball's `.meta/fixtures.ini`). Zero differences outside `_info`; per-fixture hashes match. v0.8.2 ships the same test content as v0.8.0, which is what a CI-and-renames-only diff should produce.

**Base compatibility.** `v0.8.2`'s notes don't name a base; `v0.8.0`'s notes name `tests-glamsterdam-devnet@v8.1.0`, and the 9 commits between the tags merge nothing from the base branch, so the fill content is still the v8.1.0 line. `.fixtures_url_amsterdam` points at `v8.1.1` (#7159), which satisfies the matching-or-newer constraint from #7153.

**Doc corrections riding along.** The v0.8.0 transition (#7153) left three spots stating the pre-v0.8.0 world; this bump would have made them staler, so they now state what is true (per the review principle from #7153: the tree states what is true now rather than how it got here):

- `docs/eip-8025.md` "Target schema" still said the newest release was `v0.6.2` and that conformance vectors are *generated* from a pinned spec commit — the generator was removed in #7153. The section now describes the released wire format (#3248 + #3278 + #3356), the three-dialects-under-`0x1501` note covers all three bodies, and the spec links are anchored to the `tests-zkevm@v0.8.2` tag instead of a pre-#3356 commit.
- `docs/eip-8025.md` "Measured baseline" still reported the generated 100-file set with five skips; it now reports the released bundle: 3218/3218 with `EXTRA_SKIPS` empty. The follow-up sentence about the two historical stateless divergences is re-anchored to "conformance vectors" so its referent stays true across pin bumps.
- `docs/roadmaps/forks-roadmap.md` still listed "devnet-8 zkEVM bundle" as a pending item claiming the stateless run skips every Amsterdam+ fixture — resolved by #7153, row removed.

`crates/common/types/stateless_ssz.rs`'s dialect history comment and the Makefile's bundle comment now name `v0.8.2` (comment-only changes).

**Results**

| Suite | Command | Result |
|---|---|---|
| blockchain (LEVM) | `make -C tooling/ef_tests/blockchain test-levm` | 14880 passed, 0 failed, 0 ignored (97s) |
| stateless | `make -C tooling/ef_tests/blockchain test-stateless` | 3218 passed, 0 failed, 0 filtered out (19s) |

No skips added; `EXTRA_SKIPS` stays empty and the five pre-existing `SKIPPED_BASE` name filters are untouched. The stateless count equals the extracted file count (one datatest case per JSON file), so the suite demonstrably ran the new bundle rather than an empty or stale directory — and since the release ships identical test content, "no new fixtures" is the verified expectation here, not an assumption.

**How to Test**

```bash
make -C tooling/ef_tests/blockchain test-levm        # re-downloads nothing zkevm-related
make -C tooling/ef_tests/blockchain test-stateless   # fetches the v0.8.2 bundle
```

**Checklist**

Not needed: the PR changes a fixtures pin, comments, and docs; no `Store` schema or on-disk layout is touched.

### DIFF
--- a/crates/common/types/stateless_ssz.rs
+++ b/crates/common/types/stateless_ssz.rs
@@ -415,8 +415,10 @@ const PUBLIC_KEY_BYTES: usize = 65;
 /// does **not** identify the encoding. Three dialects have shipped under it:
 /// `tests-zkevm@v0.6.2`, then #3248 + #3278, then #3356 (which moved `state`,
 /// `codes` and `public_keys` to `ProgressiveList`). ethrex speaks the last one,
-/// matching `tests-zkevm@v0.8.0`. A bundle from an older dialect will not be
-/// caught by this prefix — it fails later, in decode or on a mismatched root.
+/// matching `tests-zkevm@v0.8.0` and unchanged in `v0.8.2` (upstream #3372 only
+/// renamed the Python classes; SSZ encoding is positional, so the wire is
+/// identical). A bundle from an older dialect will not be caught by this
+/// prefix — it fails later, in decode or on a mismatched root.
 pub const STATELESS_INPUT_SCHEMA_ID: u16 = 0x1501;
 
 /// Byte length of the big-endian [`STATELESS_INPUT_SCHEMA_ID`] prefix.
--- a/docs/eip-8025.md
+++ b/docs/eip-8025.md
@@ -15,22 +15,26 @@ rather than building its own ethrex guest (issue
 
 ### Target schema
 
-The wire format tracks **execution-specs master at commit
-`3c3b6f4af315b268a61e20d5a4da8aa4f24c91f0`**, which is two merged PRs:
+The wire format is the one shipped in `tests-zkevm@v0.8.0` and unchanged in
+`v0.8.2`, the release the vectors are pinned to (the v0.8.0 → v0.8.2 upstream
+delta is release automation plus `Ssz*` → `SSZ*` Python renames,
+[#3372](https://github.com/ethereum/execution-specs/pull/3372) — SSZ encoding
+is positional, so nothing on the wire moved). It is the sum of three merged PRs:
 
 | PR | merged | change |
 |---|---|---|
 | [#3248](https://github.com/ethereum/execution-specs/pull/3248) | 2026-07-29 | "change StatelessInput SSZ serialization to be EIP-7688 aligned" — introduces `ProgressiveContainer` / `ProgressiveList` |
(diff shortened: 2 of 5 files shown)


ilitteri:
Superseded by #7182: renaming the head branch closed this PR (GitHub closes PRs whose head branch is renamed); the same commits continue there under the new branch name.

PR #7171, opened 2026-08-20
feat(l1): support frames-devnet-0 (tests-frames-devnet@v0.0.0)

ilitteri:
**Motivation**

First main-branch test track for the frames stream. Until now the EIP-8141 spec-test work lived only on the `frames-devnet-0` dev branch; main carries the frame-transaction implementation (#6326, #7004, #7068, #7073, #7089) but nothing runs the upstream fixtures against it. This PR pins the first released bundle, `tests-frames-devnet@v0.0.0`, and ports from the dev branch exactly the implementation the released fixtures exercise.

**Description**

The release refills the *whole* suite at the `Bogota` pseudo-fork (Amsterdam + EIP-8141, activated by `bogotaTime` — ethrex's `Fork::Hegota`), so the new `for_bogota/` subtree re-covers every other Amsterdam EIP with frame transactions active, not just the 8141 tests: 3,137 fixture files carrying 24,916 blockchain test cases (counted as top-level case keys), every one at `network: Bogota`.

Track wiring, following the Amsterdam overlay's shape:

- `tooling/ef_tests/.fixtures_url_frames` pins the release (`fixtures_frames-devnet.tar.gz`).
- `make frames-vectors` overlays `for_bogota/` onto `vectors/eest`, keyed on the release URL like the Amsterdam overlay, fetching through the retrying `download.sh`. `download-test-vectors` and `test-levm` depend on it, so the existing "Run Blockchain EF tests" CI job covers the track with no workflow change — the same way the Amsterdam bundle is covered.
- Harness: a `Bogota` fork config (Amsterdam + `hegota_time`), type-0x06 fixture deserialization (including the `"0x"`-as-absent encoding of an implicit frame target and an ARBITRARY signature's signer), and mappings for the exception names the release introduces: `TYPE_6_INVALID_{FRAME_FORMAT,SIGNATURE}` and the three overflow names, which surface at RLP decoding because the fixture values do not fit the fields (`GASPRICE_OVERFLOW`/`PRIORITY_OVERFLOW`: 33-byte fees on type-0x06 transactions, beyond the 2**256 bound EIP-8141 puts on fees and their `U256` fields; `GASLIMIT_PRICE_PRODUCT_OVERFLOW`: a legacy-tx price beyond `u64`). The two-pass parallel BAL check now runs for Amsterdam and later rather than Amsterdam exactly.
- Hive: `make run-hive-eels-frames{,-rlp,-quick}` plus `patch-hive-frames-fork`, which adds `bogotaTime` to hive's ethrex genesis mapper in the clone (the mapper stops at Amsterdam upstream; without it every EIP-8141 payload runs pre-fork and fails).

Implementation ported from `frames-devnet-0`, each commit keeping the dev branch's reasoning; everything is reachable only through type-0x06 execution or the Hegota-gated frame path, so no pre-Hegota behavior changes:

- **Frame entry access charge** (dev `666703940`): EIP-8141 Rationale — the target's cold/warm access cost is charged inside the frame's own gas limit, plus the EIP-7702 delegation access behind it. Warmth lands in the frame's substate backup so a failed frame leaves nothing warm; the EIP-7928 recorder learns of the target read regardless of how the frame ends.
- **BAL reconciliation when a frame reverts on its own** (dev `b24600017`): a reverting non-batch frame restores a recorder checkpoint, mirroring the atomic-batch unroll, so the block access list agrees with the state the block carries (write-then-read re-files as a read).
- **EIP-8037 NEW_ACCOUNT state cost when a frame revives a dead target** (dev `8817d7911`): matches EELS `charge_value_transfer_to_non_alive_account` at frame entry, found as a `gasUsed` chain split against Nethermind.
- **Run a frame's target through the EVM unless it is a codeless VERIFY target** (dev `d39f4f547`): only a codeless VERIFY target takes the protocol default code; a SENDER/DEFAULT frame to a precompile must run it. Also orders the EIP-7702 delegate read after the entry charge is known affordable (`eip7702_peek_delegation`), so an unaffordable frame files no delegate in the BAL.
- **Frame-tx fee fields widened to `U256`** (dev `999bb95dd`): EIP-8141 bounds the fee fields at 2**256 and the release's `admission_constraints[max_cost_within_bound]` uses the room (a 2**232 `max_fee_per_gas` on a transaction that must be *valid*); ethrex's `u64` fields could not decode it. The two shared accessors widen with the fields; `GenericTransaction` and the `feeHistory` reward still narrow, saturating rather than panicking. The L2 fee-arithmetic test narrows at its read site (dev `c2161f3f2`).
- Clippy conformance on the entry charge (`saturating_add`, dev `8ce49c97b`) and `.gitignore` for `*.tar.gz.part` partial downloads (dev `1d8a5ed58`).

Deliberately left on the dev branch (not needed by any released fixture):

- Post-v0.0.0 EIP revisions (dev `57a44764b`, `1e4f60cf5`): approval-context restore on frame revert, atomic-batch approval-scope static rejection, the relaxed banned-opcode list, payer warming at APPROVE. The released suite passed on the dev branch before these landed; they follow with the next release that fills them.
- Mempool-admission work (validation-prefix SLOTNUM ban, post-prefix rules, replacement/eviction, sponsored-VERIFY targeting) and RPC receipt `to` reporting: invisible to blockchain fixtures.
- The pre-release local `uv fill` pipeline and the `frames-devnet-0.yaml` kurtosis config (depends on the dev-only glamsterdam-devnet-8 base config).

`STORE_SCHEMA_VERSION` is untouched: nothing here changes a storage codec — the frame-receipt storage encoding and its flush call sites were already on main.

CI housekeeping picked up along the way: `grafana.com` and `leastauthority.com` joined `lychee.toml`'s bot-blocked-hosts exclude list. Both serve fine in a browser but 403/time out from CI runners, and they were failing the docs Link Check on pages this PR does not touch (the monitoring docs and the audits index).

**Local test run**

- `make -C tooling/ef_tests/blockchain test-levm` (Bogota overlay included): 18,017 / 18,017, 0 ignored — 14,880 pre-existing + 3,137 `for_bogota` (isolated `-- for_bogota` re-run: 3,137 / 3,137, matching the dev branch's count for this release).
- `make -C tooling/ef_tests/blockchain test-stateless`: 3,218 / 3,218.
- `make -C tooling/ef_tests/engine test`: 11,020 / 11,020 (75,641 fixture cases), 0 failed.
- `cargo test --workspace --exclude 'ethrex-l2*' --exclude ethrex-prover --exclude ethrex-guest-program`: 1,870 / 1,870 across 48 suites.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean in both workspaces; `cargo fmt --all -- --check`: clean in both workspaces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,9 @@ jwt.hex
 /tmp
 
 *tests*.tar.gz
+# Partial downloads of the fixture bundles; the Makefile fetches to `.part`
+# first so a failed download cannot truncate the previous bundle.
+*.tar.gz.part
 tooling/ef_tests/state/test.tar.gz
 
 .env
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 .PHONY: build lint test clean run-image build-image clean-vectors \
 		setup-hive test-pattern-default run-hive run-hive-debug clean-hive-logs \
 		load-test-fibonacci load-test-io run-hive-eels-blobs run-hive-eels-amsterdam \
-		run-hive-eels-bal-quick run-hive-build-block bench-rlp zkevm-bench-setup
+		run-hive-eels-bal-quick run-hive-build-block bench-rlp zkevm-bench-setup \
+		patch-hive-frames-fork run-hive-eels-frames run-hive-eels-frames-rlp \
+		run-hive-eels-frames-quick
 
 help: ## 📚 Show help for each of the Makefile recipes
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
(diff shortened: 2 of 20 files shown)


ilitteri:
Superseded by #7183: renaming the head branch closed this PR (GitHub closes PRs whose head branch is renamed); the same commits continue there under the new branch name.

PR #7180, opened 2026-08-20
perf(l1): remove the trait object from RLPEncode and backpatch list prefixes

MegaRedHand:
### Motivation

`RLPEncode::encode` took `&mut dyn BufMut`, so every `put_u8`/`put_slice` in every one of the 113 impls was an indirect call. The cost was already known in the tree: `BranchNode` carried a hand-written `encode_into_vec` whose comment said `RLPEncode::encode` "must take a trait object, so hashing a branch paid a vtable dispatch for each of its ~41 `put_u8`/`put_slice` calls".

Nothing actually needed the polymorphism. The one thing that did was `length()`, whose default ran `encode` against a counting `BufMut` so that impls without an explicit `length()` still got a correct one. Only 3 of 113 impls override it, so that default was load-bearing for ~110 types.

Once the sink is concrete, a better option opens up that a trait object cannot express: derive the list length from the buffer instead of computing it in advance. That removes `Encoder`'s per-list allocation, which turned out to cost more than the dispatch did.

### Description

Reviewable in order, with merges from `main` interleaved.

**1. Concrete sink.** `encode(&self, buf: &mut Vec<u8>)`. Mechanical across 42 files. Callers outside the impls are unchanged, since `Vec<u8>` is what nearly everyone already passed; the exceptions are the discv4/discv5 paths that owned a `BytesMut`.

`#[inline]` on the leaf impls is load-bearing, not cosmetic: with a concrete sink the vector-growth code inlines into each `encode` body, which pushes them past ThinLTO's cross-crate import threshold and stops them being inlined at call sites. Without them a tight `u64` encode loop regresses by more than 2x; with them it is flat.

**2. Backpatching.** `Encoder` records `start = buf.len()`, writes fields straight through, and closing the list inserts the prefix in front of the payload. `Vec<T>::encode` does the same instead of running a `length()` pre-pass over every element. Trades a malloc plus a copy for one memmove, per list, at every nesting level.

**3. Review follow-ups.** The last 55 `BufMut` calls converted, `encode_length` renamed to `encode_list_prefix`, two missing `#[inline]`s, one doc comment that had become false, and a clippy fix for `Encoder::finish` in release builds.

**4. Drop the branch-node child-hash cache.** #7104 gave `BranchNode` two things: the monomorphic `encode_into_vec` above, and a 16-element array caching each child's resolved hash so the length pass and the encode pass did not each call `compute_hash_ref`. Commit 1 already removed the first, since with a concrete sink there is no vtable left to dodge. The cache is the leftover, and reverting it measures as free: across the 34 guest workloads below it moves AIR cost and step count by less than 0.06% in either direction with no consistent sign, and leaves peak guest RAM byte-identical on every one. Folding over `self.choices` directly is 5 lines shorter.

**5. Remove the `encode_into_buf` benchmark group.** It encoded into one reused pre-grown buffer to isolate the write path from allocation. In practice it isolated something else: how the buffer argument is passed at the ABI level. A variant that threaded an owned `Vec<u8>` through the encoder instead of a reference moved that group by +44% on `BlockHeader/100`, purely because a `Vec<u8>` is 24 bytes and so returns through memory rather than in registers. The same change moved guest AIR cost by +0.058%. A microbenchmark that mis-ranks changes by two orders of magnitude is worse than no microbenchmark.

**6. Close the list on drop.** `Encoder`'s `Drop` panicked when an encoder was dropped without `finish`. See the note below; it now writes the prefix instead.

### Benchmarks

Measured with `tooling/zkevm_bench` under `ziskemu` v1.1.0-alpha, which runs the guest as a RISC-V program: every instruction is proven and there is no out-of-order core to hide an indirect call or an allocation. One freshly built guest ELF per side (the build script only declares `rerun-if-changed=build.rs`, so it does **not** rebuild on a branch switch; the two reports' `guest_elf_sha256` differ, `850dd74e3344` vs `203c299a79c0`).

Baseline is the merge-base `f76460224`, not current `main`, so nothing but this PR is in the delta.

`--mode medium`, 20 workloads:

| metric | baseline | this PR |
|---|---:|---:|
| total AIR cost | 87,427,966,616 | **-0.172%** |
| steps | 513,975,784 | **-0.330%** |
| peak guest RAM | 403,974,895 | **-1.510%** |

`--mode slow`, 34 workloads:

| metric | baseline | this PR |
|---|---:|---:|
| total AIR cost | 275,528,361,231 | **-0.101%** |
| steps | 1,810,411,865 | **-0.178%** |
| peak guest RAM | 807,851,331 | **-0.851%** |

Nothing regresses; the worst single workload in either mode is `mainnet_25087563_midhigh` at -0.001%.

Both aggregates are dominated by precompile-bound fixtures that barely touch RLP (`bn128_pair_60M` alone is 17% of the slow total), which is why slow reads smaller than medium. The per-workload numbers are the signal. Where RLP encoding is a real share of the work the change is worth much more, and it scales with block size:

| workload | AIR cost | steps |
|---|---:|---:|
| `eest_bench_bal_mixed_60M` | **-4.27%** | **-18.74%** |
| `eest_bench_bal_mixed_30M` | -2.89% | -16.25% |
| `eest_bench_bal_mixed_10M` | -2.07% | -14.05% |
| `eest_amsterdam_witness_7702_delegation_chain` | -1.57% | -11.01% |
| `stress_ecrecover_150M` | -0.49% | -3.50% |

Peak guest RAM falls across the board, most on the allocation-heavy fixtures: `stress_alt_bn128_150M` -16.98%, `stress_modexp_150M` -7.61%, `eest_bench_bn128_pair_60M` -5.34%, `stress_mix_operations_150M` -4.45%. That win comes entirely from removing `Encoder`'s per-list scratch allocation.

`eest_bench_jumpdest_60M` is excluded from the slow set: it panics inside ziskemu itself (`Mem::read() section not found for addr: 0=0`) identically on the baseline and on this branch, so it is an emulator bug rather than a difference.

### Notes for reviewers

- **`Encoder` is write-through, and closing the list is what writes its prefix.** Dropping an encoder closes its list, so an early return leaves well-formed RLP rather than a payload with no prefix in front of it. `finish` remains the explicit form: it takes `self`, so it decides *where* the list closes and ends the borrow on the output buffer, which matters when encoders nest and their order fixes the byte layout.
- **This replaced a `Drop` that panicked.** The old guard reported a missed `finish` but did not prevent it, and it was `#[cfg(debug_assertions)]`, so release builds had no protection at all. It was also unsound in the ordinary sense: a panic raised inside any `encode` unwound through a live `Encoder`, hit the guard, and aborted the process with `panic in a destructor during cleanup`, taking every other test in the binary with it and burying the original message. Writing the prefix instead fixes all three.
- **`Vec<T>` prefixes now come from the bytes actually written**, not from `sum(item.length())`. If any type's `length()` disagreed with its `encode()`, its encoding changes. One such type exists: `NodeHash::length()` returns 33 for the hashed case while `RLPEncode::encode` routes through `Vec<u8>` and emits an RLP list of individual bytes. It is unreachable (no `Vec<NodeHash>` or `[NodeHash; N]` anywhere) and the new bytes would be the well-formed ones, but it is a real latent inconsistency and worth a follow-up.
- **The default `length()` allocates now.** For the 52 `Encoder`-based impls this is still cheaper than `main`, which allocated a `temp_buf` per nesting level. It is worse only for the 14 direct-write impls with no override. The block-access-list encoders were the one hot path where this bit, at three levels of nesting, and they are converted to backpatching in commit 2.
- `BranchNode::encode_into_vec` is deleted: it existed only to dodge the vtable, and `RLPEncode::encode` is now byte-identical to it. Commit 4 removes the rest of #7104 for the same reason, with measurements above.
- **Overlaps with #7161**, which pre-sizes the same `Encoder` scratch buffer to 512 rather than removing it. Benchmarked head to head on the same 34 workloads, the two aggregate to nearly the same AIR cost (-0.101% here, -0.086% there), but they diverge where it counts. On the RLP-heavy fixtures this branch is close to 3x better (`eest_bench_bal_mixed_60M` is -4.27% AIR and -18.74% steps here, against -1.53% and -6.85% there), and on memory they move in opposite directions (peak guest RAM -0.85% here, +0.47% there, since pre-sizing holds a 512-byte buffer many workloads never fill). They are alternatives rather than complements.
- The staging buffer added to `Discv4Codec::encode` costs an allocation per packet, but `Discv4Codec` currently has no users; the live send path is `udp_socket.send_to` in `discv4_handlers.rs`.

Tests: full workspace suite green (1942 tests). New coverage for the 55/56 short/long boundary, every size where the length field grows a byte, nested encoders, and the drop-closes-the-list behaviour (a dropped encoder produces the same bytes as a finished one, `finish` does not double-write, and a dropped encoder still nests correctly and leaves pre-existing buffer contents alone). The prefix tests assert bytes written out from the RLP spec rather than comparing the two code paths against each other, which was verified by mutation.


crates/common/rlp/structs.rs
```diff
@@ -165,74 +163,93 @@ fn field_decode_error<T>(field_name: &str, err: RLPDecodeError) -> RLPDecodeErro
...
+#[cfg(debug_assertions)]
+impl Drop for Encoder<'_> {
+    fn drop(&mut self) {
+        panic!(
+            "`Encoder` dropped without `finish()`: {} bytes of RLP payload are in \
+             the output buffer with no list prefix in front of them",
+            self.buf.len() - self.start
+        );
     }
 }
```
MegaRedHand: Not really sure about this. I think there should be a better way to do this. Maybe making the encoder take ownership of the buffer.

ilitteri (approved):

PR #7196, opened 2026-08-21
fix(l1): recover instead of stalling when the head's post-state is unreachable

ilitteri:
**Motivation**

A devnet node stopped following the chain and could not recover, not even across restarts. A deep reorg unwound past the window of retained state and left the canonical chain and the state history pointing at different branches, so no canonical block had a post-state we could still read.

From there three independent paths all declined to make progress:

- Full sync walked to genesis looking for a stateful parent, found none, and returned `Ok` to pause "until a reconcilable forkchoice head arrives". That head can never arrive — the state is gone, not merely unreferenced — so the wait never ends.
- Every `forkchoiceUpdated` naming a finalized block we had not downloaded was rejected with `-38002` through an arm that, unlike `Syncing` and `StateNotReachable`, never starts a sync. The blocks we were missing were therefore never fetched, and each later FCU failed identically.
- On restart, `regenerate_head_state` runs the same walk and refuses to boot.

The node logged 2836 forkchoice rejections over nine hours while the chain moved ~6500 blocks ahead, then died and crash-looped. The trigger was a CL-side bug producing repeated invalid heads, but the reorg churn only exposed the condition; nothing in the execution client could get out of it.

**Description**

- Full sync reports an exhausted walk as `SyncError::StateUnrecoverable` instead of pausing, and `sync_cycle` escalates to snap sync — the only in-protocol way to obtain state we do not hold. The guard fires only once the walk has bottomed out with no stateful parent anywhere, so it cannot trigger while a usable base still exists.
- A safe/finalized block we simply do not have is now treated as missing data: start a sync and answer `SYNCING`. `-38002` is kept for elements we do hold but that are ordered wrongly (`Unordered`) or sit on a disjoint branch (`Disconnected`). Without this, the escalation above is unreachable, since no sync is ever started.
- `check_order` reports the element the caller actually looked up. It hardcoded `Finalized`, so a missing *safe* block was also reported as "Finalized" — meaning those 2836 log lines may not have been about the finalized block at all.
- Corrected the comment at the deep-reorg cache-edge bail, which claimed `apply_fork_choice` "should have succeeded as a shallow reorg" when that path having failed is precisely how execution reaches it. The bail itself is right: with an empty overlay range the state is genuinely absent locally, and deferring to sync is the honest answer — it just needed a sync that can recover.

Deliberately out of scope: the boot-time refusal in `regenerate_head_state`. A node that dies before the escalation completes still needs `ethrex removedb`, so that path is worth a follow-up.

**Tests**

Unit tests for `check_order` cover the missing-element label for both call sites, known blocks in the wrong order, and equal block numbers.

The escalation path itself has no automated coverage — reaching it requires a store whose retained state has been pruned out from under a canonical chain plus live peers to snap-sync from, which the current sync tests cannot set up. It is exercised by `cargo clippy`/`check` only. Reproducing it in CI would need a p2p harness that does not exist yet.


crates/networking/p2p/sync.rs
```diff
@@ -296,34 +296,59 @@ impl Syncer {
...
+            // canonical chain and the state history pointing at different branches.
+            //
+            // Full sync cannot dig itself out: it needs a stateful parent and there is none,
+            // so every later cycle repeats the same walk and pauses again while the chain
+            // moves on — the node goes quiet indefinitely. Snap sync is the only in-protocol
+            // way to obtain state we do not have, so escalate rather than stall until an
+            // operator notices and runs `ethrex removedb`.
+            warn!(
+                %sync_head,
+                "Full sync has no reachable state to resume from; escalating to snap sync"
```
MegaRedHand:
```suggestion
                "Full sync has no reachable state to resume from; switching to snap sync"
```

crates/networking/p2p/sync.rs
```diff
@@ -296,34 +296,59 @@ impl Syncer {
...
+            );
+            self.snap_enabled.store(true, Ordering::Relaxed);
+            self.run_snap_cycle(sync_head, store).await
```
MegaRedHand, line 353: Should we check here if the sync-mode was set to snap-sync/default? If the sync-mode is full-sync, switching to snap-sync here can result in lost data.

crates/networking/p2p/sync/full.rs
```diff
@@ -406,12 +406,16 @@ pub async fn sync_cycle_full(
...
+                     not reconcile to a block whose state we retain, so there is nothing to execute \
+                     on top of; handing over to snap sync to rebuild state from a recent pivot."
                 );
                 store.clear_fullsync_headers().await?;
-                return Ok(());
+                // Returning Ok here would pause the cycle and wait for a forkchoice head that
+                // reconciles to state we retain. When the walk has already bottomed out at
+                // genesis no such head exists — the state is gone, not merely unreferenced —
+                // so the wait never ends and the node stops following the chain entirely.
+                // Report it so the caller can escalate to snap sync.
```
MegaRedHand, line 417:
```suggestion
                // Returning Ok here would pause the cycle.
                // Report it so the caller can switch to snap sync.
```

crates/blockchain/fork_choice.rs
```diff
@@ -99,11 +99,15 @@ pub async fn apply_fork_choice(
     let head_res = store.get_block_header_by_hash(head_hash)?;
 
     if !safe_hash.is_zero() {
-        check_order(&safe_res, &head_res)?;
+        check_order(&safe_res, &head_res, error::ForkChoiceElement::Safe)?;
```
MegaRedHand: Let's import `ForkChoiceElement` here

crates/blockchain/fork_choice.rs
```diff
@@ -766,3 +776,57 @@ fn map_chain_error_for_fcu(err: ChainError, last_valid_hash: H256) -> InvalidFor
         | ChainError::UnknownPayload => InvalidForkChoice::StateNotReachable,
     }
 }
+
+#[cfg(test)]
+mod tests {
```
MegaRedHand, line 777: Should we put this in the `tests` directory?

MegaRedHand (approved):

PR #7200, opened 2026-08-24
fix(l1): expose blockTimestamp on transaction objects

ilitteri:
**Motivation**

Differential RPC testing against the other execution clients on glamsterdam-devnet-8 found ethrex's transaction objects were the only ones without `blockTimestamp`. Measured live against geth on a local two-client devnet:

| | fields on a mined tx | `blockTimestamp` |
|---|---|---|
| geth | 21 | present |
| ethrex | 20 | **absent** |

Same result via `eth_getBlockByNumber` with full transactions and via `eth_getTransactionByHash`. Receipts and logs already carried the timestamp; transaction objects were missed, so a consumer wanting to date a transaction needs a second round trip to fetch its block.

**Description**

Adds `block_timestamp` to `RpcTransaction` and threads the block's timestamp through every construction site: the full-block body, both by-block-and-index getters, and by-hash. It is `None` for a pending transaction, which has no block yet — matching how `block_number` and `block_hash` already behave there.

The by-hash-and-index handler previously read only the block body, so it now loads the header as well.

**Tests**

Three tests covering all four getters: full-block transactions, `eth_getTransactionByHash`, and both by-block-and-index spellings. They assert the exact timestamp the test harness stamps (`0x3e8`) rather than merely asserting presence.


crates/networking/rpc/eth/transaction.rs
```diff
@@ -212,6 +213,10 @@ impl RpcHandler for GetTransactionByBlockHashAndIndexRequest {
             Some(block_body) => block_body,
             _ => return Ok(Value::Null),
         };
+        let block_header = match context.storage.get_block_header(block_number)? {
+            Some(block_header) => block_header,
+            _ => return Ok(Value::Null),
+        };
```
pablodeymo, line 227: just a comment to reduce lines of code and improve readibility:

This code (and all the let in here) could be replaced by let ... else:
        let Some(block_header) = match context.storage.get_block_header(block_number)? else {
            return Ok(Value::Null)
        };

pablodeymo (approved):

MegaRedHand (approved):

PR #7203, opened 2026-08-24
docs(l2): run the release upgrade test with a real prover, and fix the verification-key instructions

ilitteri:
**Motivation**

The release checklist's L2 upgrade test ran locally with `--backend exec`. That backend produces no proof, so it never consults a verification key — and the verification key is precisely what an ethrex upgrade changes. The test could not observe the main way an upgrade breaks.

Running it with a real SP1 prover for the first time (v24.0.0 → v25.0.0) surfaced a requirement that has applied since **v9.0.0** and whose written procedure has been unusable for just as long. Two changes landed together in that release: per-commit-hash verification keys, and the Timelock becoming the OnChainProposer's owner.

Each batch is committed under `keccak` of the git sha of the binary that committed it, and `commitBatch` rejects a commit hash it holds no key for. So after upgrading, a deployment that verifies real proofs commits **nothing** until the new build's key is registered — every commit reverts with `MissingVerificationKeyForCommit()` (`0xf6b9798e`) while the L2 keeps producing blocks, which reads as a stuck committer rather than a missing key.

`docs/l2/fundamentals/upgrades.md` has documented the mechanism since it landed, but both of its instructions stopped working in that same release:

| The doc says | Measured against a live deployment |
| --- | --- |
| hash the "reduced" git commit, e.g. `9219410` → `b9105485…` | that keccak is `4732af28…`, so the example does not reproduce. The value is the **full 40-char sha as ASCII** — hashing v24.0.0's full sha yields exactly the key its own deployer registered |
| send `upgradeSP1VerificationKey` from the OnChainProposer owner account | reverts with `OwnableUnauthorizedAccount` (`0x118cdaa7`); the Timelock has owned the contract since v9.0.0, so the call has to be routed through it |

**Description**

Move the release upgrade test onto the GPU server with `--backend sp1`, and document the step it was missing.

- `docs/developers/release-process.md` — the checklist item now names `l2-gpu` and SP1, explains why `exec` hides this failure class, and notes both L2 checks share ports and datadirs so they run sequentially. Keeps the fresh-deploy SP1 test as well: a release can deploy cleanly and still fail to upgrade, and the reverse.
- `docs/developers/l2/upgrade-test.md` — deploy `VERSION_FROM` with `--sp1 true` so proofs are actually required, use the `-gpu` asset (the others have no CUDA, so `--backend sp1` silently falls back to CPU and a batch looks like a hang), both provers on `--backend sp1`, a fourth stated goal covering on-L1 verification, and a new **Step 3.6** that registers `VERSION_TO`'s key.
- `docs/l2/deployment/upgrades.md` — a standing per-release entry, since this applies to any real-proof deployment and not only to the release test.
- `docs/l2/fundamentals/upgrades.md` — corrected. This page stays the single canonical explanation and the other two link to it; two divergent copies are what let the stale version go unnoticed.

All four now point at `rex l2 register-vk` (lambdaclass/rex#249), which derives the commit hash, routes through the Timelock and reads the key back, rather than a hand-assembled `cast`/`rex send` call.

**Validation**

Every claim here was measured on a v24.0.0 → v25.0.0 upgrade on `l2-gpu`, not derived from reading the contracts:

- the two releases ship **different** verification keys, so registration is genuinely required
- `keccak(full ascii sha)` reproduces the key the `VERSION_FROM` deployer registered for itself — confirming the derivation
- v24.0.0 with `--sp1 true`: batches verified with real GPU proofs
- v25.0.0 **before** registration: could not commit at all, reverting with `0xf6b9798e`
- after registering through `Timelock.emergencyExecute`: the committer's own retry recommitted the pending batch, and the batch carrying the v24.0.0-produced blocks verified under v25.0.0's key
- the integration suite then passed against the upgraded stack, `1 passed; 0 failed`

**Checklist**

### DIFF
--- a/docs/developers/l2/upgrade-test.md
+++ b/docs/developers/l2/upgrade-test.md
@@ -26,11 +26,13 @@ mkdir -p "$WORK"
 > `--backend exec`. The `exec` backend produces no proof, so it never consults a
 > verification key — and the verification key is the thing an upgrade changes.
 > Batches are committed under `keccak(VERGEN_GIT_SHA)`, baked into the binary at
-> build time, and `OnChainProposer.verificationKeys[commitHash][SP1_VERIFIER_ID]`
-> is read at verification, so `$VERSION_TO`'s batches cannot be verified until
-> its key is registered ([Step 3.6](#36-register-the-version_to-verification-key)).
-> Under `exec` that failure is invisible: the test passes and the real upgrade
-> stalls at the first batch.
+> build time, and `commitBatch` refuses a commit hash it holds no key for. Until
+> `$VERSION_TO`'s key is registered
+> ([Step 3.6](#36-register-the-version_to-verification-key)) the upgraded
+> sequencer cannot commit **anything**: every `commitBatch` reverts with
+> `MissingVerificationKeyForCommit()` (selector `0xf6b9798e`) and the L2 keeps
+> producing blocks locally while nothing reaches L1. Under `exec` that failure is
+> invisible — the test passes and the real upgrade stops at the first batch.
 
 ## Prerequisites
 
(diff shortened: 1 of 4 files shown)


ilitteri:
Superseded by #7284, which combines this with the other L2 release-test docs PR. Both extend the same release checklist, so they conflicted with each other; the combined branch carries every commit from this one.

PR #7206, opened 2026-08-24
fix(l1): recap `eth_estimateGas` by the fee the balance check uses

jrchatruc:
## Motivation

`eth_estimateGas` fails with `Insufficient account funds` for an EIP-1559 call object whenever the sender's balance is below `maxFeePerGas * block_gas_limit`, even when the transaction it is estimating costs a few tens of thousands of gas and the sender can pay for it many times over. The same call succeeds against other clients.

The estimation ceiling is recapped against the sender's balance only when the legacy `gasPrice` field is set:

```rust
if !transaction.gas_price.is_zero() {
    highest_gas_limit = recap_with_account_balances(...)
}
```

A 1559 call object states its fee cap in `maxFeePerGas` and does not carry `gasPrice`, and `GenericTransaction::gas_price` is `#[serde(default)]`, so it deserializes to zero. The guard is false, the recap is skipped, and `highest_gas_limit` stays at the whole block's gas limit.

The balance check that the simulation then runs reads a *different* field. `default_hook::validate_sufficient_balance` requires:

```rust
vm.env.tx_max_fee_per_gas.unwrap_or(vm.env.gas_price).checked_mul(gas_limit.into())
```

and `tx_max_fee_per_gas` *is* populated from `maxFeePerGas`. So with the ceiling left at the block gas limit, the check demands a full block's worth of gas at the caller's fee cap: at a 60M block limit and a 10 gwei cap, 0.6 ETH, before the transaction has been simulated at all.

That asymmetry is the bug. The recap reads `gas_price`; the check reads `max_fee_per_gas`. geth caps by `GasFeeCap` first and falls back to `GasPrice`, so it computes a real allowance and estimates fine.

## Description

Pick the fee cap the balance check will actually be measured against, and recap by that:

- `maxFeePerGas` when the call object carries one, otherwise the legacy `gasPrice`. Written as the same expression the check uses rather than an equivalent one, so the two sites cannot drift: a call object setting *both* fields, which `GenericTransaction` accepts with no conflict check, would otherwise be capped against one fee and checked against the other.
- A request stating no fee at all, or an explicit fee of zero, is left uncapped: no balance bounds the gas. That check now lives inside `recap_with_account_balances`, which returns the ceiling unchanged, so no caller can reach the division with a zero divisor.
- `recap_with_account_balances` takes the fee cap explicitly rather than reaching for `transaction.gas_price`.

Four regression tests over an in-memory store whose sender holds 0.01 ETH, against a 25M genesis gas limit and a 1 gwei fee cap (1 gwei * 25M = 0.025 ETH, so an uncapped ceiling cannot be afforded):

- a 1559 request estimates `0x5348` instead of failing (fails on `main`),
- a request setting both fees is capped by the max fee (fails against a `gasPrice`-first recap),
- a legacy request still estimates `0x5208`,
- a request with no fee at all is not recapped and does not divide by zero.

The 1559 tests carry calldata deliberately. The plain-transfer short circuit above the recap answers `TRANSACTION_GAS` without reaching the cap, and a test that asserts on the recap should not depend on missing it.

Two things mask the bug, which is why it is not more visible: a sender holding more than a full block's gas at its own fee cap, and a recipient absent from state entirely, which takes the short circuit above the recap and never reaches it.

`maxFeePerBlobGas` is the one term of the balance check still missing from the recap; that is #7209, stacked on this.

### DIFF
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -547,33 +547,29 @@ impl RpcHandler for EstimateGasRequest {
         };
 
         // The cap has to be computed against the same fee the balance check will be
-        // measured against. `default_hook` requires `tx_max_fee_per_gas * gas_limit`,
-        // falling back to `gas_price` only when the former is absent, so recapping by
-        // `gas_price` alone leaves a 1559 request — where the legacy field is absent and
-        // deserializes to zero — uncapped at the block gas limit, and the simulation then
-        // fails `InsufficientAccountFunds` for any sender that cannot afford the whole
-        // block's gas at its fee cap.
-        let fee_cap = if transaction.gas_price.is_zero() {
-            transaction
-                .max_fee_per_gas
-                .map(U256::from)
-                .unwrap_or_default()
-        } else {
-            transaction.gas_price
-        };
+        // measured against, and by the same rule: `default_hook` requires
+        // `tx_max_fee_per_gas * gas_limit`, falling back to `gas_price` only when the
+        // former is absent. Recapping by `gas_price` alone left a 1559 request — where the
+        // legacy field is absent and deserializes to zero — uncapped at the block gas
+        // limit, and the simulation then failed `InsufficientAccountFunds` for any sender
+        // that could not afford the whole block's gas at its fee cap. Written as the same
+        // expression rather than an equivalent one, so the two sites cannot drift: a call
+        // object setting both fields (which `GenericTransaction` accepts, and which
+        // `calculate_gas_price_for_generic` resolves legacy-first) would otherwise cap
+        // against one fee and be checked against the other.
+        let fee_cap = transaction
+            .max_fee_per_gas
+            .map(U256::from)
+            .unwrap_or(transaction.gas_price);
 
-        // Zero means no fee was specified at all, which is the caller asking not to be
-        // capped; it also guards the division below.
-        if !fee_cap.is_zero() {
-            highest_gas_limit = recap_with_account_balances(
-                highest_gas_limit,
-                &transaction,
-                fee_cap,
-                storage,
-                block_header.number,
-            )
-            .await?;
-        }
+        highest_gas_limit = recap_with_account_balances(
+            highest_gas_limit,
+            &transaction,
+            fee_cap,
+            storage,
+            block_header.number,
+        )
+        .await?;
 
         // Check whether the execution is possible
         let mut transaction = transaction.clone();
@@ -649,8 +645,13 @@ impl RpcHandler for EstimateGasRequest {
 
 /// Caps the estimation ceiling at the gas the sender can actually pay for.
 ///
-/// `fee_cap` must be the per-gas price the transaction's balance check uses: the legacy
-/// `gas_price` when it is set, otherwise `max_fee_per_gas`. It must be non-zero.
+/// `fee_cap` is the per-gas price the transaction's balance check uses: `max_fee_per_gas`
+/// when the call object carries one, otherwise the legacy `gas_price`.
+///
+/// A zero `fee_cap` is a request that names no fee at all, or names a fee of zero, and in
+/// both cases no balance bounds the gas: the ceiling is returned unchanged. Checked here
+/// rather than at the call site so no caller can reach the division with a zero divisor,
+/// which is the same split this function's `fee_cap` argument exists to close.
 async fn recap_with_account_balances(
     highest_gas_limit: u64,
     transaction: &GenericTransaction,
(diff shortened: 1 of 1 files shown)


ilitteri:
Verified the diagnosis before reviewing it: rebuilt the parent commit's `transaction.rs` with only
this PR's test module appended, and `estimate_gas_recaps_a_1559_request_by_its_max_fee_per_gas`
fails there and passes here. The recap/check asymmetry is real and this closes it for the 1559 case.

Four things, ordered by how much I'd insist.

### 1. The precedence is inverted, and one input still fails

`validate_sufficient_balance` reads `tx_max_fee_per_gas.unwrap_or(vm.env.gas_price)`: max fee wins.
This picks `gas_price` first. The two agree on every input except a call object setting both fields,
which ethrex accepts. `GenericTransaction` has no conflict check, `env_from_generic` fills
`env.gas_price` from the legacy-first `calculate_gas_price_for_generic`, and `env.tx_max_fee_per_gas`
comes straight from `maxFeePerGas`, so the two can hold different numbers.

`gasPrice: 2 gwei` + `maxFeePerGas: 3 gwei`, against the 0.01 ETH sender the new tests use:

| | result |
|---|---|
| this branch | `Err("Invalid Transaction: Insufficient account funds")` |
| with the diff below | `Ok("0x5208")`, three new tests still green |

```rust
let fee_cap = transaction
    .max_fee_per_gas
    .map(U256::from)
    .unwrap_or(transaction.gas_price);
```

Same expression as the check, so the two sites can't drift. It also drops the `if/else`, the
`unwrap_or_default()` and the separate zero case.

Why the case exists at all: geth's `ToMessage` sets `gasFeeCap = gasPrice` whenever the caller gives
the legacy field, and reth inherits the property from revm's single `TxEnv::gas_price`, so neither
client can hold two disagreeing fee caps. geth, reth and nethermind also reject `gasPrice` alongside
`maxFeePerGas`. ethrex does neither. Worth a separate look, though the one-liner above removes the
dependency on it.

### 2. Blob gas sits in the balance requirement but not in the recap

`validate_sufficient_balance` adds `max_fee_per_blob_gas * BLOB_GAS_PER_BLOB * n_blobs` to what the
sender must hold. `recap_with_account_balances` subtracts only `value`. geth subtracts the blob usage
from `available` before dividing, for this reason. A call object with `blobVersionedHashes` and
`maxFeePerBlobGas` reaches it: `validate_sender_balance` runs ahead of `validate_4844_tx` in
`prepare_execution`.

`maxFeePerGas: 1 gwei` + `maxFeePerBlobGas: 45 gwei` + one versioned hash, same sender: still
`Insufficient account funds` here. Subtracting the blob cost in the recap gives `0x5208`. Fine as a
follow-up. It's the last piece of the same asymmetry.

### 3. The new tests reach the recap only because of a second bug

The plain-transfer short circuit above the recap:

```rust
let code = account_info.map(|info| storage.get_account_code(info.code_hash));
if code.is_none() {
```

`get_account_code` is sync and returns `Result<Option<Code>, StoreError>`, so `code` is an
`Option<Result<..>>`. It goes `None` only when the recipient account is missing, and nothing looks
inside the `Result`. The three new tests reach the recap because `to = SENDER` is a genesis account.
Point the same 1559 request at an address absent from genesis and even the pre-fix code answers
`0x5208`.

Correct that short circuit to mean "recipient has no code", which geth
(`len(call.Data) == 0 && GetCodeSize(to) == 0`), reth (`input.is_empty()` plus an empty bytecode
hash) and nethermind (`To is not null && Data.IsEmpty`) all check, and these three tests will
short-circuit to `0x5208` and keep passing without touching the recap. Calldata on one of them buys
insurance no short circuit can satisfy:

```rust
"input": "0x0102030405060708",
```

which estimates `0x5348` against `fixtures/genesis/l1.json`.

### 4. Minor

- The doc comment says `fee_cap` "must be non-zero" while the guard enforcing it sits in the caller,
  the same split this PR closes one line above. Move the zero check into
  `recap_with_account_balances` and return `highest_gas_limit` unchanged, and no caller can reach the
  division with a zero divisor.
- "Zero means no fee was specified at all" misses `maxFeePerGas: "0x0"`, which lands in the same
  branch. Behaviour is right, the comment overstates.
- Description: sending to an address that doesn't exist yet also masks the bug, per (3), not just a
  well-funded sender. Changes who would have noticed.

Checked while here, out of scope: `eth_call` and `eth_createAccessList` don't recap. geth does the
same, its `CallDefaults` setting `gas = RPCGasCap` and stopping there; reth diverges and caps by the
caller allowance when the request omits `gas`. nethermind never recaps and turns the failed first
`CallAndRestore` into an `InputError`, so it errors for an underfunded sender the way ethrex did
before this PR. "The same call succeeds against other clients" holds for geth and reth.

ilitteri (approved):

PR #7211, opened 2026-08-24
perf(l1): make `eth_estimateGas`'s plain-transfer short circuit fire

jrchatruc:
Stacked on #7209, which is stacked on #7206. Merge those first. Raised in #7206's review.

## Motivation

`eth_estimateGas` has a short circuit that answers `TRANSACTION_GAS` immediately when the request is a plain value transfer, skipping the binary search. Its condition:

```rust
let code = account_info.map(|info| storage.get_account_code(info.code_hash));
if code.is_none() {
```

`get_account_code` is sync and returns `Result<Option<Code>, StoreError>`, so `code` is an `Option<Result<..>>` and nothing ever looks inside the `Result`. It goes `None` only when the recipient **account is absent from state**, which is close to the opposite of the case this exists for: an ordinary funded wallet has an account, so every transfer to one ran the full search.

## Description

Check what the other clients check: no calldata, and a recipient with no code. geth is `len(call.Data) == 0 && GetCodeSize(to) == 0`; reth (`input.is_empty()` plus an empty bytecode hash) and nethermind (`To is not null && Data.IsEmpty`) agree.

Both halves matter:

- calldata to a code-less account is 21000 **plus** its per-byte cost, so it is not a transfer,
- an empty call to a contract runs that contract's fallback, so it is not one either.

Compared against the code hash rather than fetching the code through `get_account_code`: the hash is already in hand, and the code itself is not needed to know it is empty. An account absent from state has no code either, so it keeps taking the same path.

## Testing

Four tests for the short circuit itself, in their own module: a transfer to an ordinary funded account, a transfer to an absent account, a call carrying calldata, and an empty call to the l1 genesis deposit contract.

They do not pin the condition, and cannot: the short circuit returns only when its own simulation at `TRANSACTION_GAS` succeeds, so a request the old condition wrongly admitted fell through to the search anyway. Dropping both halves still passes. The change is invisible in the answer, which is what makes it safe and also what leaves it unpinned by any test asserting on output. Observing it would take a count of `simulate_tx` calls.

Making the short circuit fire also retired three of the stack's regression tests, which is fixed here since this is the PR that causes it. They send to `SENDER`, a code-less genesis account, with no calldata — exactly the shape now answered before the recap is reached. Measured by reintroducing the bugs those tests exist to catch:

| mutation | before this PR | on this branch | with the fix |
|---|---|---|---|
| legacy-first fee precedence | `..._when_both_fees_are_set` fails | passes | fails again |
| drop the `fee_cap.is_zero()` guard | `..._without_any_fee_is_not_recapped` fails | passes | fails again |

They now carry the same calldata the 1559 test already did, as shared `CALLDATA`/`WITH_CALLDATA` constants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 ## Perf
 
+### 2026-08-24
+
+- Make `eth_estimateGas`'s plain-transfer short circuit fire. Its condition tested whether the recipient account existed rather than whether it had code, so every transfer to an ordinary funded wallet ran the full binary search instead of returning `TRANSACTION_GAS` at once [#7211](https://github.com/lambdaclass/ethrex/pull/7211)
+
 ### 2026-07-22
 
 - Unify full-sync batch import onto the per-block execution pipeline, validating every block's state root and reusing the pipeline's BAL-driven parallel execution instead of the bespoke "execute all, apply once" batch path [#7008](https://github.com/lambdaclass/ethrex/pull/7008)
--- a/crates/networking/rpc/eth/transaction.rs
+++ b/crates/networking/rpc/eth/transaction.rs
@@ -12,7 +12,7 @@ use crate::{
 use ethrex_blockchain::{Blockchain, vm::StoreVmDatabase};
 use ethrex_common::{
     H256, U256,
-    constants::GAS_PER_BLOB,
+    constants::{EMPTY_KECCAK_HASH, GAS_PER_BLOB},
     types::{AccessListEntry, BlockHash, BlockHeader, BlockNumber, GenericTransaction, TxKind},
 };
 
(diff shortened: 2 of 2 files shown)


ilitteri:
The condition is right and matches what the other clients check. Two things about the tests, and the
first one has to land here because this is the PR that causes it.

### 1. This PR retires three of the stack's regression tests

`to: SENDER` is a code-less genesis account, so every recap test that carries no calldata now
short-circuits at 21000 and never reaches the recap it was written to pin. Measured by putting the
legacy-first fee precedence back — the bug #7206 exists to fix — at each point in the stack:

| | #7206 | this branch |
|---|---|---|
| `estimate_gas_recaps_by_the_max_fee_when_both_fees_are_set` | `Err("Insufficient account funds")`, catches it | passes |

All ten pass here with that bug reintroduced. Same story for `..._a_legacy_request_by_its_gas_price`.
And deleting the `fee_cap.is_zero()` guard inside `recap_with_account_balances` no longer fails
`estimate_gas_without_any_fee_is_not_recapped`, the test named for it — the division by zero is
caught instead by two of the tests added here, which happen to state no fee.

One line each, the same insurance the 1559 test already carries:

```rust
"input": "0x0102030405060708",
```

`0x5208` becomes `0x5348` in all three (21000 + 8 bytes at the 7623 floor). With that applied:
12/12 green, and both mutations above fail again — `..._when_both_fees_are_set` on the precedence,
`..._without_any_fee_is_not_recapped` on the divisor.

### 2. The four new tests can't fail, so they don't pin the condition

Dropping **both** halves — `input.is_empty()` and the code-hash check, i.e. short-circuiting every
`Call` — still passes 12/12. The condition is invisible to the answer, because the short circuit only
returns when its own simulation at `TRANSACTION_GAS` succeeds, and anything carrying calldata or real
code dies on intrinsic gas there and falls through. Checked on the two shapes that would be the
expensive misses if it didn't:

| call object | estimate |
|---|---|
| `accessList`, one address + one key, no calldata | `0x62d4` = 21000 + 2400 + 1900 |
| `authorizationList`, one tuple, no calldata | `0xb3b0` = 21000 + 25000 |

Both fall through, both correct. That guard is where the correctness lives, which is what makes the
change safe — but it also means "the tests pin the intended semantics so a future edit cannot quietly
reintroduce either half" isn't so: `code.is_none()` could come back tomorrow and CI would stay green.
Pinning it needs the *work* observed rather than the answer — a count of `simulate_tx` calls — or the
sentence should come out of the description, since the four tests then read as coverage the diff
doesn't have.

### 3. Minor

- The new tests live in `estimate_gas_fee_cap_tests`, which is now two subjects in one module.
- The description restates local `cargo test` / `fmt` / `clippy` results; CI reports those.

jrchatruc:
> c4c3560

Addressed here [c4c3560](https://github.com/lambdaclass/ethrex/pull/7211/commits/c4c3560af3f9cf3c2e7b0e9870e634ed63608b14)

ilitteri (approved):

PR #7222, opened 2026-08-26
perf(l1): store jump destinations as a bitmap

ilitteri:
> [!WARNING]
> **Blocked on #7093.** `Code::size()` here accounts its allocation against a budget
> derived from `host_memory_limit_bytes()`, which #7093 introduces and `main` does not
> yet have. CI will fail to compile until #7093 merges. The bitmap logic itself is
> complete and reviewable now; I deliberately did not vendor #7093's commits into this
> branch, because resolving another PR's conflicts on its behalf would make both harder
> to review.

**Motivation**

The last feature on `glamsterdam-devnet-8` genuinely missing from `main`. EthPandaOps need devnet-8 to mirror `main`, so it is being upstreamed rather than carried on the branch.

Jump destinations were a sorted `Arc<[u32]>` of offsets persisted next to the bytecode. For jumpdest-dense contracts that is roughly 4× the code size — a measured 97,906 B of value for 24,576 B of code — and ~10 ns per entry to rebuild on every code-cache miss, so 246 µs for a max-size runtime that is almost all `JUMPDEST`s.

**Description**

One bit per bytecode byte, set when that offset holds a `JUMPDEST` that is not part of a `PUSH` immediate.

| | offset list | bitmap |
|---|---|---|
| size | ~4× code size at high density | `len/8` at **any** density |
| decode | 246 µs (max-size, dense) | 0.5 µs (a memcpy) |
| jump validation | 12.9 ns (binary search) | 0.37 ns (bit test) |

**No migration.** Values in the older form are still readable: the RLP item header distinguishes a list from a byte string, and for a list the bitmap is rebuilt from the bytecode — 26.6 µs, still ~9× cheaper than decoding the list.

`Code::size()` now counts the bytecode allocation it previously excluded, so the code cache honours its byte budget instead of holding orders of magnitude more than it accounted for. That is the line that pulls in #7093.

**The second commit is not optional.** Indexing the bitmap put a bounds-check panic path in the scan loop that cost more than the scan: jumpdest-free 24 KB initcode went to 55.4 µs against 11.2 µs for the offset-list version it replaced, and `benchmarkoor`'s `test_jumpdest_analysis[00]` lost 69–78 % throughput. Accumulating a byte's bits in a register and storing once per byte fixes it — the monotonically increasing index makes the single store safe. Net against the offset-list version, per 24 KB: dense 49.8 → 17.0 µs, jumpdest-free 11.1 → 11.8 µs, 1-in-32 12.5 → 12.0 µs, PUSH-heavy unchanged.

**Tests**

Adds `test/tests/common/jumpdest_bitmap_tests.rs`. Running on `glamsterdam-devnet-8` since 2026-08-03.

All performance figures above are from the original commit messages; I have not reproduced any of them. The jumpdest-free row is a small regression (11.1 → 11.8 µs) that the author accepted, which is worth a second look given how common jumpdest-free initcode is.

### DIFF
--- a/crates/common/types/account.rs
+++ b/crates/common/types/account.rs
@@ -95,16 +95,31 @@ impl Code {
 
     /// Builds the [`Code::jumpdests`] bitmap: one pass over the bytecode, setting the
     /// bit for every `JUMPDEST` while skipping `PUSH` immediates.
+    ///
+    /// The bits of a byte are accumulated in a register and written once the scan leaves
+    /// that byte, which the monotonic `i` makes safe. Reading the bitmap back inside the
+    /// loop instead would turn each `JUMPDEST` into a read-modify-write, and indexing it
+    /// would put a bounds-check panic path in the loop; either costs more than the scan
+    /// itself for bytecode with few jump destinations.
     pub fn compute_jumpdests(code: &[u8]) -> Arc<[u8]> {
         let mut bitmap = vec![0u8; code.len().div_ceil(8)];
         let mut any = false;
+        let mut current_byte = usize::MAX;
+        let mut bits = 0u8;
         let mut i = 0;
         while i < code.len() {
             // TODO: we don't use the constants from the vm module to avoid a circular dependency
             match code[i] {
                 // OP_JUMPDEST
                 0x5B => {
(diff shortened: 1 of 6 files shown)


ilitteri:

PR #7224, opened 2026-08-26
test(l1): bump Amsterdam fixtures to glamsterdam-devnet v8.1.2

ilitteri:
**Motivation**

[`tests-glamsterdam-devnet@v8.1.2`](https://github.com/ethereum/execution-specs/releases/tag/tests-glamsterdam-devnet@v8.1.2) is out, superseding v8.1.1 on `glamsterdam-devnet-8`. It is announced as a test-only release: "the spec equal to v8.1.1 (minus non-consensus breaking EIP-7610 removals), no repricing or EIP changes", carrying new EIP-8037, EIP-7928 and EIP-7702 coverage and dropping the EIP-7610 create-collision tests.

**Description**

Moves `tooling/ef_tests/.fixtures_url_amsterdam` and `.github/config/hive/amsterdam.yaml` (`fixtures` + `eels_commit`) to v8.1.2. Both, together: grading one release's client against another release's bundle fails every fixture the two disagree on, which is what produced 2168 hive failures when only the ef-test pin moved to v8.1.0. `eels_commit` is `50a7f6ec`, which is both the v8.1.2 tag and the tip of `devnets/glamsterdam/8`.

Also refreshes the two places that document the pin: the `glamsterdam-devnet-8` block in `docs/roadmaps/forks-roadmap.md`, and the two `hive.md` lines that still named v8.0.0 — stale since the v8.1.0 bump.

**Scope, stated against an explicit base.** This branch is based on `glamsterdam-devnet-8` and delivers there by direct push, so that branch is the review base. Measured against it, the change is exactly four files — `tooling/ef_tests/.fixtures_url_amsterdam`, `.github/config/hive/amsterdam.yaml`, `docs/roadmaps/forks-roadmap.md` and `docs/developers/l1/testing/hive.md` — two pins plus the two documents that record them, +24/−18 lines over two commits. No Rust, and no client code of any kind, is modified relative to that base. `git diff --stat origin/glamsterdam-devnet-8..HEAD` shows the whole of it.

That claim is scoped to that base on purpose, and should not be read against `main`. `glamsterdam-devnet-8` is a long-lived branch, 108 commits and 159 files ahead of `origin/main`, so `git diff origin/main...HEAD` renders the entire pre-existing devnet baseline — `CHANGELOG.md`, `Cargo.lock`, `cmd/ethrex/*`, `crates/blockchain/*` and the rest — none of which this branch touches. The check that settles it: the file set of `origin/main...HEAD` is *identical* to the file set of `origin/main...origin/glamsterdam-devnet-8`, which is the same diff computed without these two commits. These commits add no file to it, and every client file in it is byte-identical between `HEAD` and the `glamsterdam-devnet-8` tip.

**Verified the release's claims rather than assuming them.** The `v8.1.1..v8.1.2` diff is 15 commits. Under `src/ethereum/forks/amsterdam/` exactly three files move, and none of them reprices anything:

- `fork.py`: a single `# TODO: This function has not been tested rigorously` comment deleted from `get_last_256_block_hashes`, in every fork ([specs#3397](https://github.com/ethereum/execution-specs/pull/3397) added unit tests for it). No behaviour.
- `state_tracker.py`: `account_deployable` loses its storage check and `account_has_storage` is deleted outright, here and in every other fork, along with the `PreState.account_has_storage` protocol member in `state.py`/`state_mpt.py`/`state_db.py` ([specs#3417](https://github.com/ethereum/execution-specs/pull/3417)). A zero-nonce, empty-code account that has storage is now deployable in EELS.
- `vm/instructions/system.py`: as a direct consequence, `generic_create` stops refunding the account-creation state gas on a collision.

```python
     # On a collision the child's execution-gas grant is consumed and no
-    # account is created; a storage-only collision target is
-    # non-existent: charged above, refilled here.
+    # account is created. A collision target has code or a nonce, so
+    # the account-creation charge above was never taken.
     if not account_deployable(tx_state, contract_address):
         increment_nonce(tx_state, sender_address)
-        if new_account_charged:
-            credit_state_gas_refund(evm.gas_meter, StateGasCosts.NEW_ACCOUNT)
         push(evm.stack, U256(0))
         return
```

There is also a structural argument that no repricing can be hiding anywhere in this release, which is stronger than checking the hunks one by one. Across the whole of `src/`, the `v8.1.1..v8.1.2` diff is **+2/−767** over 52 files, and the only file with *any* added lines is `src/ethereum/forks/amsterdam/vm/instructions/system.py` (+2/−4) — the two lines of replacement comment shown above. A gas-constant change requires an added line, and there are exactly two added lines in the entire specification tree, both of them comment text. So "no repricing" is not merely unobserved here, it is structurally impossible. (Confirmed against the GitHub compare API for the two tags, independent of any local checkout: 15 commits, 102 files, and exactly three touched under `src/ethereum/forks/amsterdam/` — `fork.py` +0/−1, `state_tracker.py` +0/−27, `system.py` +2/−4.)

The one other thing that could have been a repricing is not: `tests/amsterdam/eip8037_state_creation_gas_cost_increase/spec.py` bumps its EIP-8037 reference-spec hash from `a12902ae` to `5a8c8089`, which is [EIPs#11328](https://github.com/ethereum/EIPs/pull/11328) "Update EIP-8037: clarify reservoir mechanics" — 9 added and 8 deleted lines of prose (a stray `<-TODO->` removed, the calldata-floor/refund interaction documented, `refund_counter` clarified as a single undimensioned counter, and the EIP-7702 authorization refund stated to land in `state_gas_reservoir`). No gas constant moves. So "no repricing" holds.

**EIP-7610: ethrex keeps the storage collision check, deliberately.** [specs#3417](https://github.com/ethereum/execution-specs/pull/3417) removes it because "EIP-7610 is declined for inclusion in Glamsterdam ([EIP-7773](https://eips.ethereum.org/EIPS/eip-7773)): no mainnet account has non-empty storage with empty code and nonce, and clients are dropping the storage collision check." The replacement matrix in [specs#3425](https://github.com/ethereum/execution-specs/pull/3425) deliberately excludes the zero-nonce/empty-code cells at param generation, because "with non-empty storage the behavior is undefined in protocol". So no fixture in this release mandates dropping `LevmAccount::has_storage` from `create_would_collide` (`crates/vm/levm/src/account.rs:113`), and the whole bundle passes with it in place. That is checked against the shipped fixtures and not just against the upstream generator. Parsing the top-level case keys of all four `frontier/create/create_collision` files in both the blockchain and engine bundles gives 170 cases: 160 from the parameterised matrix, plus 10 `balance_only_target` cases. Bucketing the 160 by the nonce/code/storage tokens in their keys shows the matrix realises exactly **6 of the 8** possible `nonce × code × storage` cells, and the two it omits are `nonce_0 + no_code + storage` and `nonce_0 + no_code + no_storage` — precisely the pair upstream's `if nonce != 0 or code != b""` filter drops. The bundle therefore reproduces that source-level exclusion independently of reading the generator.

Of those two omitted cells, the storage-free one is still covered, by the dedicated `balance_only_target` tests: a target holding balance only is zero-nonce and empty-code, ethrex's `create_would_collide` returns false for it because all three of its terms are false, EELS finds it deployable for the same reason, and those 10 cases pass. The remaining cell — zero nonce, empty code, **non-empty** storage — is the one place ethrex and v8.1.2 EELS disagree, and the number of shipped cases exercising it is **0**. Nothing in this release grades that behaviour in either direction. Removing it is also not a fork-gated change in ethrex: `create_would_collide` applies to every fork, so dropping it would change behaviour for historical mainnet blocks too. That is a coordinated consensus decision that deserves its own PR and its own review, not a side effect of a fixture bump. It is now flagged in the EIP table in `docs/roadmaps/forks-roadmap.md`.

ethrex's state-gas side already agrees with the new `generic_create` for every case EELS still defines. ethrex charges `NEW_ACCOUNT` conditionally on `target_alive = !account.is_empty()` rather than charging and refunding, so the refund at `crates/vm/levm/src/opcode_handlers/system.rs:1001` needs `create_would_collide() && !target_alive` to fire. `AccountInfo::is_empty()` is the EIP-161 predicate — zero balance, zero nonce, empty code — so `!target_alive` forces all three to zero, which leaves `has_storage` as the only term that can still make `create_would_collide()` true. The refund is therefore reachable only on a storage-only, zero-balance target: precisely the cell upstream calls undefined, and precisely the cases upstream deleted it for. It stays because it is the coherent partner of keeping the check — it fires only for the storage-only target that ethrex still treats as a collision — and it should be removed together with the check if and when that happens.

**What the bundle actually changed.** Counted by parsing top-level case keys out of each fixture JSON, not by grepping:

| subtree | v8.1.1 files / cases | v8.1.2 files / cases |
|---|---|---|
| `blockchain_tests/for_amsterdam` | 3133 / 24920 | 3137 / 24965 |
| `blockchain_tests/for_bpo2toamsterdamattime15k` | 42 / 127 | 42 / 127 |
| `blockchain_tests_engine/for_amsterdam` | 3135 / 24948 | 3139 / 24993 |
| `blockchain_tests_engine/for_bpo2toamsterdamattime15k` | 43 / 129 | 43 / 129 |
| `state_tests/for_amsterdam` | 2758 / 15954 | 2759 / 15997 |

Every delta maps onto a release-notes bullet: +85 cases of `frontier/create/create_collision` (specs#3425) against −52 of `paris/eip7610_create_collision` and −2 ported-static Paris CREATE cases (specs#3417), with `collision_selfdestruct` moving from the EIP-7610 directory to EIP-6780; +3 `state_gas_reservoir/spilled_state_gas_consumed_across_halt_chain` ([specs#3423](https://github.com/ethereum/execution-specs/pull/3423), with the pinned-gas follow-ups [specs#3427](https://github.com/ethereum/execution-specs/pull/3427) and [specs#3383](https://github.com/ethereum/execution-specs/pull/3383)); +1 `bal_cross_tx_reverted_storage_reads` ([specs#3399](https://github.com/ethereum/execution-specs/pull/3399)); +8 EIP-7702 delegation-clearing and delegated-frame precompile cases ([specs#3420](https://github.com/ethereum/execution-specs/pull/3420), [specs#3421](https://github.com/ethereum/execution-specs/pull/3421)).

3126 of the 3133 pre-existing blockchain fixtures differ byte-for-byte, but that is `_info.url` commit-hash churn (`32f597f7` → `50a7f6ec`). Stripping `_info` leaves 15 genuinely changed fixtures, all under EIP-8037, and the change is the specs#3383 post-state-verification refactor moving beneficiaries off `0x…dead` onto generated addresses — not gas values. All 15 ran and passed.

The engine fixture format did not change. [specs#3294](https://github.com/ethereum/execution-specs/pull/3294) attaches an SSZ model to `FixtureExecutionPayload` and [specs#3219](https://github.com/ethereum/execution-specs/pull/3219) fixes Engine X fills on BAL forks, but `_SizedUint` subclasses `HexNumber`, so `Uint64`/`Uint256` serialize exactly as the `HexNumber` fields they replaced, and the fixture-JSON comparison above confirms it empirically: no pre-existing engine fixture changed shape.

**Which of the new coverage hive actually grades.** Worth stating explicitly, because "the pins match" does not mean "hive runs the new tests". The `Consume Engine Amsterdam` limit is an allowlist of devnet-8 EIP numbers, `.*(2780|7708|7732|7778|7843|7928|7954|7975|7976|7981|7997|8024|8037|8038|8045|8061|8070|8159|8246|8282).*`, matched against fixture case keys — it sweeps 3881 of the 24993 `for_amsterdam` engine cases. Matching it against the new coverage: the EIP-8037 `state_gas_reservoir` additions (123/123) and the EIP-7928 `bal_cross_tx_reverted_storage_reads` case (1/1) are graded by hive; the 85 `frontier/create/create_collision` cases (0/85), the EIP-7702 `set_code_tx` coverage and the relocated `collision_selfdestruct` case (0/1) are not, because their case keys carry no allowlisted EIP number. Everything hive misses here is still covered by the ef-test blockchain and engine suites, which run the whole bundle and are green on this branch.

The EIP-7702 figure deserves a footnote, because the naive count flatters it. Two of the 724 `set_code_tx` cases do match the limit — but incidentally, not deliberately: the limit is an unanchored substring match against the whole case key, and in those two the allowlisted number `8282` happens to be the last four hex digits of the system-contract address being tested (`…65613800e8282` and `…89300d8282`). Neither case has anything to do with EIP-8282. Masking long hex runs before re-matching puts 18 of the whole bundle's 3881 matches in that incidental category, so the 15% headline is sound but any per-EIP figure read off the raw match needs this check first.

This is not a regression from this bump — the tests the collision matrix replaces lived under `paris/eip7610_create_collision`, and 7610 is not in the allowlist either, so hive's collision coverage was zero before and after. Widening the limit would change CI scope and runtime and cannot be validated from this branch, so it is left alone and noted here for a follow-up rather than folded into a fixture bump.

**How to test**

```
make -C tooling/ef_tests/engine test
make -C tooling/ef_tests/state run-evm-ef-tests-ci   # heavy; ~21 min
```

Results on this branch:

| suite | result |
|---|---|
| blockchain | **14884 passed, 0 failed** |
| engine | **11024 passed, 0 failed** |
| state | **72625 passed, 0 failed** (`Summary: 72625/72625 (100.00%)`) |

The blockchain and engine suites register one Rust test per fixture file, so their counts are checkable straight against the tree: blockchain 14884 = 11756 EEST + 3128 legacytests, engine 11024 = its EEST file count. (The state suite counts cases rather than files, so it is reconciled separately below.) The EEST-side counts moved by exactly the bundle's net file delta — 11752 → 11756 and 11020 → 11024 — against the v8.1.1 figures reported in #7159, whose blockchain number covered the EEST subtree only.

The state suite's own report ends in `Summary: 72625/72625 (100.00%)`, and none of its 8980 per-directory summary lines is below 100%. Summing those lines per subtree puts `state_tests/for_amsterdam` at exactly **15997/15997** — the same case count the bundle table above predicts for v8.1.2, up from 15954. That is the check worth doing on every bump, because a suite that silently extracted nothing looks identical to a passing one.

One reporting gap found while doing it, not fixed here: `tooling/ef_tests/state/report.rs` hardcodes its per-fork rollup as Osaka, Prague, Cancun, Shanghai, Paris (lines 214-218 and 327-331), carrying the note "Uncomment the summaries if EF tests for those specific forks exist". Amsterdam's now exist and `runner/mod.rs:67` already defaults `--forks` to include Amsterdam, so the run executes 15997 Amsterdam cases and prints no `Amsterdam: n/n` line for them. It is only a reporting gap — `Summary` 72625 minus the five printed rollups is 15997, exactly the Amsterdam directory sum — but it means the count check above has to be done by hand. Left for its own PR rather than folded into a pin bump.

That the suites graded the new bundle at all, rather than a stale extraction, is checked directly: sampling 60 fixture files per suite and reading `_info.url` returns commit `50a7f6ec` — the v8.1.2 release commit — on every one, with no `32f597f7` survivors, and the `for_amsterdam` file counts on disk are 3137 / 3139 / 2759, exactly the v8.1.2 column of the table above.

The new coverage was confirmed executed rather than merely present: `bal_cross_tx_reverted_storage_reads`, `spilled_state_gas_consumed_across_halt_chain`, `cancun/eip6780_selfdestruct/collision_selfdestruct/selfdestruct_after_create2_collision`, all four `frontier/create/create_collision` files and the new EIP-7702 cases each appear once per suite log with `... ok`. No stale `eip7610_create_collision` directory survived inside `for_amsterdam` or `for_bpo2toamsterdamattime15k`; the EIP-7610 tests that still run come from the untouched mainnet `tests@v20.0.0` pin under `for_cancun`, `for_osaka` and `for_prague`.

No fixture was skipped, no exception mapping was touched, no assertion was weakened and no test was deleted. That the exception mapper needed no work is checked rather than assumed: `tooling/ef_tests/engine/src/exception_mapper.rs` is a hand-port of upstream's `client_clis/clis/ethrex.py` and warns in its header that drift between the two is "the main source of `validation_error` mismatch noise" — that Python file is byte-identical across `v8.1.1..v8.1.2`, and no exception-definition file changed either.

**Checklist**

- [x] No `Store` schema change, so `STORE_SCHEMA_VERSION` is untouched. The diff is two pinned fixture URLs, an `eels_commit` and two documentation files; no storage type, encoding or key layout is affected, so nothing here implies a re-sync.

Related: #6583, the Glamsterdam tracking issue the `glamsterdam-devnet-8` roadmap block points at, and upstream's release tracker [execution-specs#3452](https://github.com/ethereum/execution-specs/issues/3452). This does not close either.


ilitteri:
Closing: this was opened against `main` from a branch built on `glamsterdam-devnet-8`, so its diff shows that branch's whole divergence instead of the fixture bump. The v8.1.2 pins are already on `glamsterdam-devnet-8`; the `main` bump comes as its own change against `main`.

PR #7225, opened 2026-08-26
fix(l1): refresh the eth forkid ENR entry when the chain crosses a fork

ilitteri:
**Motivation**

`build_local_node_record` (`crates/networking/p2p/network.rs:134`) derives the EIP-2124 fork id once at startup, and nothing recomputes it. The moment the chain crosses a fork we keep advertising the old fork id, and geth's dial-candidate filter — `NewNodeFilter` in `eth/protocols/eth/discovery.go` — returns false when the `eth` entry does not match, so we are dropped as a dial candidate.

The existing comment there already describes the consequence: a peer "passes us over until we republish with it". Nothing republished.

Supersedes #7037.

**Description**

This is a rewrite, not a rebase of #7037. That PR read `self.store` inside the discovery actor, which #6449 deliberately removed — `network.rs:127` states the record is "Built here rather than inside discovery, which has no say in what this node announces about itself." Re-adding storage to the actor would revert that decision, so the fork id is **published to** discovery instead:

- the network layer owns a `tokio::sync::watch` sender and re-derives the fork id every 30s (`FORK_ID_REFRESH_INTERVAL`), so the ENR can lag a crossing by at most that long
- discovery holds the receiver and applies changes on the prune tick it already uses to republish after an IP change, next to `update_local_ip`

`send_if_modified` keeps the receiver's change flag clean while the chain stays on one fork, so a steady chain costs one flag check per tick and no signing. A transient storage error is logged at debug and retried on the next tick rather than logged every 30s.

Three review follow-ups tightened the startup path:

- **The channel is seeded with the fork id the initial ENR was actually built with**, not a second storage read. Two reads could straddle a fork crossing, leaving the ENR permanently stale: the receiver would already hold the new fork id as its initial value, so `has_changed` would never fire for it.
- **The seed no longer uses `futures::executor::block_on`** inside the async `start_network` chain. It only worked because `get_fork_id` happens to have no real await point today; the value now arrives as a parameter from `build_local_node_record`.
- **A transient failure of the initial read no longer disables refresh for the node's lifetime.** The channel carries `Option<ForkId>`: a failed startup read seeds `None` (announcing a record without an `eth` entry, as today), and the first successful derivation counts as a change that gives the ENR its `eth` entry.

**Tests**

`cargo clippy --workspace --all-targets -- -D warnings` and `--features l2` both clean, fmt clean, 106 `ethrex-p2p` tests and the 969-test integration suite pass.

What is **not** covered: nothing here proves the ENR is actually republished across a real fork crossing. The unit tests construct the actor with a dropped publisher (`watch::channel(None).1`), which exercises the no-op path only. Verifying the live behaviour needs a devnet that crosses a fork with a geth peer, watching whether geth dials us after the transition — that is the test that would have caught the original bug, and it does not exist in either version of this change.

### DIFF
--- a/crates/networking/p2p/discovery/server.rs
+++ b/crates/networking/p2p/discovery/server.rs
@@ -11,6 +11,7 @@ use crate::{
     types::{Node, NodeRecord},
 };
 use bytes::BytesMut;
+use ethrex_common::types::ForkId;
 use ethrex_common::utils::keccak;
 use futures::StreamExt;
 use secp256k1::SecretKey;
@@ -27,7 +28,7 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 use thiserror::Error;
 use tokio::net::UdpSocket;
 use tokio_util::udp::UdpFramed;
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info, trace, warn};
 
 use super::{DiscoveryConfig, codec::DiscriminatingCodec, lookup_interval_function};
 
@@ -90,6 +91,13 @@ pub struct DiscoveryServer {
     pub(crate) config: DiscoveryConfig,
     pub discv4: Option<Discv4State>,
     pub discv5: Option<Discv5State>,
+    /// Current EIP-2124 fork id, published by the network layer.
+    ///
+    /// Discovery does not read this from storage: the network layer decides what this
+    /// node announces about itself, and discovery only republishes. Without a refresh
+    /// the `eth` ENR entry goes stale the moment the chain crosses a fork, and geth's
+    /// dial-candidate filter (`NewNodeFilter`) drops us outright.
+    pub(crate) fork_id: Option<tokio::sync::watch::Receiver<ForkId>>,
 }
 
 impl std::fmt::Debug for DiscoveryServer {
@@ -106,6 +114,10 @@ impl std::fmt::Debug for DiscoveryServer {
 impl DiscoveryServer {
     /// Starts the discovery actor, advertising `local_node_record` as this
     /// node's ENR.
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "one more caller-supplied value; bundling them would only hide the call site"
+    )]
     pub async fn spawn(
         local_node: Node,
         local_node_record: NodeRecord,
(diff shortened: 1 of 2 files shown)


crates/networking/p2p/discovery/server.rs
```diff
@@ -90,6 +91,13 @@ pub struct DiscoveryServer {
     pub(crate) config: DiscoveryConfig,
     pub discv4: Option<Discv4State>,
     pub discv5: Option<Discv5State>,
+    /// Current EIP-2124 fork id, published by the network layer.
+    ///
+    /// Discovery does not read this from storage: the network layer decides what this
+    /// node announces about itself, and discovery only republishes. Without a refresh
+    /// the `eth` ENR entry goes stale the moment the chain crosses a fork, and geth's
+    /// dial-candidate filter (`NewNodeFilter`) drops us outright.
+    pub(crate) fork_id: Option<tokio::sync::watch::Receiver<ForkId>>,
```
MegaRedHand: We should import this instead of using the full path

crates/networking/p2p/discovery/server.rs
```diff
@@ -114,6 +126,7 @@ impl DiscoveryServer {
         peer_table: PeerTable,
         bootnodes: Vec<Node>,
         config: DiscoveryConfig,
+        fork_id: Option<tokio::sync::watch::Receiver<ForkId>>,
```
MegaRedHand: Same

crates/networking/p2p/network.rs
```diff
@@ -847,3 +855,42 @@ fn format_duration(duration: Duration) -> String {
...
+/// can lag a crossing by at most this long; cheap enough to keep tight.
+const FORK_ID_REFRESH_INTERVAL: std::time::Duration = std::time::Duration::from_secs(30);
+
+/// Publishes the current fork id, re-deriving it periodically.
+///
+/// Returns `None` when the chain cannot supply one at all, which keeps the existing
+/// behaviour of announcing a record without an `eth` entry rather than failing startup.
+fn spawn_fork_id_publisher(
+    context: &P2PContext,
+) -> Option<tokio::sync::watch::Receiver<ethrex_common::types::ForkId>> {
```
MegaRedHand: We should import these types

MegaRedHand (approved):
LGTM. Left some style comments

PR #7229, opened 2026-08-27
fix(l1): return the requested block from eth_getBlockByHash, not its canonical sibling

ilitteri:
## Motivation

`eth_getBlockByHash` resolved hash → number → **canonical** block, so asking for a
non-canonical hash answered with the canonical block at the same height — a different block
than requested, whose `hash` field did not even echo the query:

```bash
# on glamsterdam-devnet-8, requesting the orphaned slot-101721 payload:
curl -s -X POST -H 'Content-Type: application/json' localhost:8545 -d '{"jsonrpc":"2.0",
 "method":"eth_getBlockByHash",
 "params":["0xf34e7d4210b200725481828142eee19c89748cee993d78a5fdace17f4be21153",false],"id":1}'
# → returns block 0x0ad87652…  (the canonical block at the same height)
```

This breaks Prysm's ePBS payload-envelope reconstruction: serving
`execution_payload_envelopes_by_root` batch-fetches fork blocks by hash from its EL
(`ExecutionBlocksByHashes` → `eth_getBlockByHash`) and rejects responses whose hash does not
match the request — 54 `execution block hash mismatch` failures in 12 h on one
glamsterdam-devnet-8 node. A node that falls one envelope behind recovers through exactly
this path, so a broken by-hash read turns a one-envelope hiccup into a minutes-long stall;
this was the amplifier behind the slot-101721 stale-head proposal investigated on that
devnet.

Same pathology as the p2p serve path caught on devnet-5/6 (#6921, `GetBlockHeaders` by
non-canonical hash — on `glamsterdam-devnet-8`, PR to main still open), now caught on the
RPC path.

## Description

The store already reads header and body by the hash itself (`Store::get_block_by_hash`);
the handler now uses it instead of the hash → number → canonical resolution. Unknown hashes
keep returning `null` per the execution-apis `notFound` schema. The response's `hash` field
is the requested hash (no re-hash of the returned header).

## Testing

New `test/tests/rpc/get_block_by_hash_tests.rs`:

- store a canonical block and a non-canonical sibling at the same height, request the
  sibling's hash, and assert the sibling comes back. Without the fix this test fails by
  returning the canonical block — reproducing the live mismatch verbatim.
- unknown hash still returns `null`.

Verified: workspace clippy and `--features l2` clippy clean, fmt clean, 119 `ethrex-rpc`
unit tests and the full 983-test integration suite pass. The same commit was validated on
`glamsterdam-devnet-8` (cherry-pick + full rpc suite) and pushed there as `1c0d83a91`.


crates/networking/rpc/eth/block.rs
```diff
@@ -105,19 +105,15 @@ impl RpcHandler for GetBlockByHashRequest {
...
-        let body = storage.get_block_body(block_number).await?;
-        let (header, body) = match (header, body) {
-            (Some(header), Some(body)) => (header, body),
-            // Block not found
-            _ => return Ok(Value::Null),
+        // Read by the hash itself, never hash -> number -> canonical block: for a
+        // non-canonical hash that resolution answers with the canonical sibling at
+        // the same height — a different block than the caller asked for. Prysm's
+        // ePBS envelope reconstruction fetches fork blocks by hash and rejects the
+        // response when its hash does not echo the request.
```
MegaRedHand: This doesn't need a comment

```suggestion
```

MegaRedHand (approved):
LGTM

PR #7230, opened 2026-08-27
fix(l1): serve GetBlockHeaders by non-canonical hash so syncing peers can fetch their head

ilitteri:
## Motivation

`GetBlockHeaders::fetch_headers` translated a by-hash start to a block number whenever the
hash was known (`get_block_number(hash)`), then read by number. For a **non-canonical**
hash that returns a *different*, canonical block at the same height, so the response does
not start at the requested hash and the requesting peer rejects the whole batch. A syncing
peer asking for its own fork/sync head by hash — a head not yet canonical on the serving
node — therefore received no usable headers.

Same `hash → number → canonical` pathology as the RPC-path bug fixed in #7229
(`eth_getBlockByHash`); this is the p2p serve side. It surfaced originally on
glamsterdam-devnet-5/6 sync stalls.

## Description

Only translate a hash to a number when it is the canonical block at its height
(`get_canonical_block_hash(number) == hash`); otherwise keep the hash and serve the block
directly. For a reverse request, walk parent hashes so a non-canonical branch can be served
(ascending/skip by hash isn't representable, so a single header is returned there).

Redo of #6921: that branch drifted ~2 months / 110 files behind main and was unmergeable;
this reapplies the same fix cleanly on current main. **The fix is already present on
`glamsterdam-devnet-8`** (`783349968`), so this PR only brings it to `main`; #6921 can be
closed as superseded.

## Testing

`test/tests/p2p/rlpx/blocks_tests.rs`: build a canonical chain plus a fork head, request the
fork head by hash, and assert the fork branch comes back — a single header for the
non-canonical hash, and a reverse-by-hash walk down the fork's own parents. Without the fix
the request returns the canonical sibling, reproducing the peer-rejected response.

Verified: workspace and `--features l2` clippy clean, fmt clean, 106 `ethrex-p2p` unit + 144
p2p integration tests pass.


pablodeymo (approved):

pablodeymo:
The solution is correct. Just some comments:

1. Reverse-by-hash with skip > 0 returns only one header (blocks.rs:164). The Codex bot flagged this too, and it's technically right: it could be served by walking skip+1 parent links and discarding intermediates (what geth's GetAncestor does), so the comment "skipping by hash isn't representable" overstates. But it's not a regression — old code either served the wrong chain or a single header — and the spec allows returning fewer headers than requested. In practice, syncing peers use reverse-by-hash with skip == 0. I'd either soften the comment to "we don't support" or leave it for a follow-up; not worth blocking a fix that's already proven on devnet-8.
2. Buffer asymmetry in the canonicality check: get_block_number consults the in-memory write buffer, but get_canonical_block_hash reads only the backend (plus the latest-header fast path). A canonical-but-not-yet-flushed block below the head could be misjudged as non-canonical. The consequence is only degraded (fewer headers, or a parent-walk instead of a number walk), never wrong content — fine to leave as is.
3. Treating a storage error during the canonical lookup as "not canonical" (.ok().flatten()) is a safe fallback — correct choice here.

PR #7234, opened 2026-08-31
revert(l1): drop blockTimestamp from RPC log objects (#7142)

ilitteri:
**Motivation**

Reverts #7142.

The `blockTimestamp` key that #7142 added to RPC log objects makes eight hive `rpc-compat` cases fail: the simulation compares responses byte-exactly against a corpus pinned to execution-apis `d08382ae`, which predates the field. The pin cannot move forward (every revision whose fixtures carry `blockTimestamp` also carries a pre-merge test chain, which ethrex does not support), so the daily hive report has been stuck at `RPC API Compatibility: 88/96 (91.67%)` since the change landed. #7142 anticipated this and excluded the eight cases in `KNOWN_EXCLUDED_TESTS`, but the coverage loss is real — all four `eth_getLogs` cases are in the excluded set, leaving that method with no rpc-compat coverage at all.

**Description**

Removes `block_timestamp` from `RpcLog` (one struct change covers `eth_getLogs`, `eth_getTransactionReceipt` and `eth_getBlockReceipts`, which all build the same type), and drops the eight `KNOWN_EXCLUDED_TESTS` entries plus the `docs/known_issues.md` section that #7142 introduced.

Trade-off, kept explicit: log objects no longer carry `blockTimestamp`, diverging from the other clients that populate it. Hive conformance and full `eth_getLogs` rpc-compat coverage win over the convenience field.

**How to test**

```
cargo test -p ethrex-rpc --lib
```

End to end, validated locally with hive at the CI-pinned commit (`7c4c99e`) against a locally built image of this branch:

```
./hive --client-file ../fixtures/hive/clients.yaml --client ethrex \
  --sim ethereum/rpc-compat --sim.parallelism 4 --sim.loglevel 1 \
  --sim.buildarg "branch=d08382ae5c808680e976fce4b73f4ba91647199b"
```

Result: `tests=96 failed=0` — including all eight previously excluded cases (fixture diff analysis of the failing daily run showed the extra `blockTimestamp` key was the only difference in every case).

**Checklist**

- [x] No `Store` schema change, so `STORE_SCHEMA_VERSION` is untouched.

### DIFF
--- a/.github/scripts/check-hive-results.sh
+++ b/.github/scripts/check-hive-results.sh
@@ -59,23 +59,7 @@ mkdir -p "${failed_logs_root}"
 
 # Tests excluded from the failure count (substring match against test case
 # name).
-#
-# The rpc-compat cases below are the ones whose recorded response contains at
-# least one log object. That corpus is pinned to execution-apis d08382ae, which
-# predates `blockTimestamp` on log objects, and rpc-compat compares byte-exactly,
-# so emitting the field (as every other client does) cannot match. The pin cannot
-# move: the field only reaches those fixtures in revisions that also carry a
-# pre-merge test chain, which ethrex does not support by design. See
-# docs/known_issues.md.
 KNOWN_EXCLUDED_TESTS=(
-  "eth_getLogs/contract-addr"
-  "eth_getLogs/no-topics"
-  "eth_getLogs/topic-exact-match"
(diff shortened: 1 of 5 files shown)


MegaRedHand (approved):
LGTM. Let's revert this while we look for a workaround

PR #7236, opened 2026-08-31
fix(l1): return -32602 for eth_getLogs block ranges beyond head

MysticRyuujin:
**Motivation**

A cross-client survey in hive rpc-compat (ethereum/execution-apis#875) found ethrex diverges from the other five execution clients on eth_getLogs range validation. geth, Nethermind, Besu, Erigon, and reth return `-32602` for these requests. ethrex returns:

- `-32000 "Invalid params: Empty range"` when the resolved range is empty (reversed range, or fromBlock past the head with toBlock `latest`)
- `-32603 "Internal Error: Could not get header for block N"` when toBlock is a number past the head, because the per-block loop only fails when the header lookup misses

**Description**

- Add an `RpcErr::InvalidParams` variant that maps to code `-32602` with the given message.
- Validate the range up front in `fetch_logs_with_filter`: toBlock past the head returns `-32602 "block range extends beyond current head block"`, and fromBlock above toBlock returns `-32602 "invalid block range params"`. Both messages match go-ethereum.
- Add a test that covers the three shapes.

ethereum/execution-apis#875 also adds rpc-compat fixtures for these cases.

**Checklist**

### DIFF
--- a/crates/networking/rpc/eth/logs.rs
+++ b/crates/networking/rpc/eth/logs.rs
@@ -201,8 +201,16 @@ pub(crate) async fn fetch_logs_with_filter(
                 .resolve_block_number(&storage)
                 .await?
                 .ok_or(RpcErr::WrongParam("toBlock".to_string()))?;
-            if (from..=to).is_empty() {
-                return Err(RpcErr::BadParams("Empty range".to_string()));
+            let latest = storage.get_latest_block_number()?;
+            if to > latest {
+                return Err(RpcErr::InvalidParams(
+                    "block range extends beyond current head block".to_string(),
+                ));
+            }
+            if from > to {
+                return Err(RpcErr::InvalidParams(
+                    "invalid block range params".to_string(),
+                ));
             }
             // The idea here is to fetch every log and filter by address, if given.
             // For that, we'll need each block in range, and its transactions,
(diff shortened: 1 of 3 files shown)


ilitteri (commented):
Verified the error strings and code against geth (`eth/filters/api.go:39-40`, both `invalidParamsErr` -> `-32602`), and the test asserts the exact variant + message for all three shapes from the hive survey. Also checked the `pending` tag isn't regressed (it resolves to latest when no pending block exists).

Two non-blocking suggestions inline: one for byte-identical geth behavior when a request trips both checks, one doc nit on the new variant.

crates/networking/rpc/eth/logs.rs
```diff
@@ -201,8 +201,16 @@ pub(crate) async fn fetch_logs_with_filter(
...
+            if to > latest {
+                return Err(RpcErr::InvalidParams(
+                    "block range extends beyond current head block".to_string(),
+                ));
+            }
+            if from > to {
+                return Err(RpcErr::InvalidParams(
+                    "invalid block range params".to_string(),
+                ));
             }
```
ilitteri: geth validates in the opposite order: `from > to` first, then `to > head` (`eth/filters/filter.go:228-234`). For a request that trips both (e.g. `fromBlock=10, toBlock=5` on a 3-block chain), geth returns `invalid block range params` while this returns `block range extends beyond current head block`. Same `-32602` either way, so rpc-compat passes regardless, but swapping the checks makes the behavior byte-identical:

```suggestion
            if from > to {
                return Err(RpcErr::InvalidParams(
                    "invalid block range params".to_string(),
                ));
            }
            if to > latest {
                return Err(RpcErr::InvalidParams(
                    "block range extends beyond current head block".to_string(),
                ));
            }
```

crates/networking/rpc/utils.rs
```diff
@@ -32,6 +32,8 @@ pub enum RpcErr {
     WrongParam(String),
     #[error("Invalid params: {0}")]
     BadParams(String),
+    #[error("{0}")]
+    InvalidParams(String),
```
ilitteri, line 39: Nit: the enum now has three param-error variants (`WrongParam` and `BadParams` at `-32000`, this one at `-32602`). A short doc comment would keep the next contributor from picking the wrong one:

```suggestion
    /// Spec-compliant invalid params error (-32602). The message is used
    /// verbatim, with no "Invalid params: " prefix, so error strings can
    /// match other clients exactly. `BadParams`/`WrongParam` map to -32000.
    #[error("{0}")]
    InvalidParams(String),
```

crates/networking/rpc/eth/logs.rs
MysticRyuujin: Applied in e8a1f2fd6, thanks. Verified geth checks the reversed range first (eth/filters/filter.go:229-234), so this matches.

crates/networking/rpc/utils.rs
MysticRyuujin: Applied in e8a1f2fd6.

ilitteri (approved):

PR #7243, opened 2026-08-31
docs(l2): add the multiprover (SP1 GPU + TDX) release test

ilitteri:
**Motivation**

Every L2 check in the release process exercises a single prover against an `OnChainProposer` configured to require only that one. A production rollup runs with two, and nothing verified that a batch can satisfy both — so a regression in the two-prover verification path would ship unnoticed.

**Description**

Adds `docs/developers/l2/multiprover-test.md` and a release-process checklist entry for a deployment where `REQUIRE_SP1_PROOF` and `REQUIRE_TDX_PROOF` are both set, so `lastVerifiedBatch` only advances once a batch has satisfied both.

The TDX side runs with `ETHREX_TDX_DEV_MODE=false` on real TDX hardware, so the quote is checked on chain by `verifyAndAttestOnChain`. This is deliberately stronger than CI, which uses dev mode — there `TDXVerifier.register()` short-circuits, taking the signing address from the quote's first 20 bytes and never verifying anything. Dev mode is documented as the fallback when no TDX host is available.

**Validation**

Run against `v25.0.0-rc.2`, SP1 prover on `l2-gpu`, everything else on `ethrex-tdx-baremetal`:

| Check | Result |
| --- | --- |
| `TDX_VERIFIER_ADDRESS.isDevMode()` | `false` — real verification, not the dev short-circuit |
| `REQUIRE_SP1_PROOF()` / `REQUIRE_TDX_PROOF()` | both `true` |
| `lastVerifiedBatch` | advanced past zero with both flags set |
| `InvalidTdxProof()` reverts | none |
| `MRTD()` on chain | `0x4e2d467d…`, the measurement of the image under test |
| `authorizedSignature()` | set from the verified report, not from raw quote bytes |

`ProverSetup received for TDX` → `ProverSetupACK sent` with dev mode off is the verifier accepting a quote signed by that machine's silicon.

**Prerequisites the deploy does not handle**

Real attestation needed three things, each of which fails opaquely and none of which were documented:

| Blocker | What is actually wrong |
| --- | --- |
| `MRTD mismatch` | `_validateReport` compares MRTD/RTMR0-2 against constants compiled into `TDXVerifier.sol` that pin one historical image build; they must be pinned to the image being released |
| `TCB_Cert_Expired` (`0xea8cd522`) | nothing has expired — the deploy loads `CA.ROOT` and `CA.PLATFORM` but never `CA.SIGNING`, so the DAO cannot check collateral signatures |
| `Unsupported chain_id: 9` | `automata-dcap-qpl-tool` resolves addresses from a chain-id-keyed registry and cannot serve a dev chain; its exit status is discarded by `prepare_quote_prerequisites`, so the first symptom is an unrelated revert later |

Also recorded: TD_QE identity needs version 4 or 5, not the `version` inside the JSON (`Incorrect_Enclave_Id_Version`); these upserts must be sent as raw calldata because cast's tuple parser splits on the commas in the JSON payload; and `run-qemu` from `hypervisor.nix` cannot be detached because it hardcodes `-serial mon:stdio`.

**Environment pins**

Four more, each hit while bringing the test up:

| Pin | Symptom when wrong |
| --- | --- |
| solc exactly 0.8.31 | `Source file requires different compiler version` — `TDXVerifier.sol` pins the pragma |
| fresh chain per deploy | `deploy-p256` reverts with no reason string (CREATE2 addresses already populated) |
| `--watcher.watch-interval 1000` | every commit reverts `InvalidPrivilegedTransactionLogs()` (`0x9e6e5638`) on an instantly-mining dev L1 |
| both provers attached before the first batch; L1 and L2 reset together | a prover deadlocks on `No blocks to prove`, asking forever for a batch whose input was pruned, while `lastVerifiedBatch` cannot advance without it |

**Follow-ups this surfaced** (not in this PR)

- `prepare_quote_prerequisites` discards the qpl tool's exit status, so a failed collateral load only surfaces as an unrelated revert much later.
- The late-joining-prover deadlock is a coordinator robustness gap, not just a test-setup quirk: a prover that restarts after batches accumulate can never catch up.
- `TDXVerifier`'s measurements are storage variables with no setter, so authorising a new prover image means redeploying the verifier.

**Checklist**

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -122,6 +122,7 @@
   - [Integration tests](./developers/l2/integration-tests.md)
   - [Upgrade test](./developers/l2/upgrade-test.md)
   - [L2 integration tests with a SP1 GPU prover](./developers/l2/sp1-gpu-integration-test.md)
+  - [Multiprover test (SP1 GPU + TDX)](./developers/l2/multiprover-test.md)
   - [Running the Prover](./developers/l2/prover.md)
   - [Generate blobs for the state reconstruction test](./developers/l2/state-reconstruction-blobs.md)
 - [Prover](./prover/prover.md)
--- a/docs/developers/l2/multiprover-test.md
+++ b/docs/developers/l2/multiprover-test.md
@@ -161,6 +161,8 @@ Each response is `{"<object>": {...}, "signature": "<hex>"}`; the DAO wants the 
 
 The proof coordinator must bind `0.0.0.0` so the SP1 prover on the other host can reach it, and the qpl tool path must be passed even in dev mode.
 
+`--committer.commit-time` must not outpace proof generation. Both proofs are required per batch, so the pipeline advances at the slower prover's rate: an SP1 proof on `l2-gpu` measured a steady 106s, which is why 120s is used here. At 15s the committer outruns proving by about 7x, and the gap never closes — a 63-hour run reached batch 10983 committed against 2127 verified. Step 9 cannot pass in that state, however long it is left running.
+
 ```bash
 set -a; . ~/multiprover/$TAG/.env; set +a
 ~/multiprover/$TAG/ethrex-l2-linux-x86_64 l2 --no-monitor \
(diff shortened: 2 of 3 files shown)


ilitteri:
Superseded by #7284, which combines this with the other L2 release-test docs PR. Both extend the same release checklist, so they conflicted with each other; the combined branch carries every commit from this one.

PR #7249, opened 2026-09-01
feat(l1): upgrade stateless-validator guests to ERE v0.17.0 and pin ere by tag

ilitteri:
**Motivation**

ERE v0.17.0 is out (v0.16.3 fixed the OpenVM zkvm accelerator, v0.17.0 added `zkVMProver::execute_estimated_cost`).

Separately, the ere version is stated in the guest manifests, in `zkvm-version.sh` and in the release workflow's `ERE_TAG`, and nothing compares the last of those to the others. A bare rev is also not something the release workflow can act on: it pulls `ere-compiler-<zkvm>` and `ere-server-<zkvm>` by *image tag*, so the rev and the image tag are two independent statements of one fact. If they disagree, the guests are compiled against one ere and executed under another, which yields a plausible ELF rather than a failure.

**Description**

- Move all four stateless-validator manifests from `rev = "8961a4e7…"` to `tag = "v0.17.0"`, and regenerate the four lockfiles.
- `zkvm-version.sh` already declared the ere tag and rejected a guest manifest that had moved off it. It now also checks the **parent** manifest, not just `bin/<zkvm>` — the parent pins `ere-platform-core` while each bin pins its own `ere-platform-<zkvm>`, so a drift between them puts two ere versions in one guest, and checking only the bin being asked about let that through.
- Add `--ere-tag` to that script so the release workflow derives `ERE_TAG` from it instead of keeping a second copy of the version.
- Add `make check-ere-pins`, wired into the existing `check-cargo-locks` job, so a disagreement surfaces on the pull request rather than at release time.

No zkVM SDK moves in this range — OpenVM stays `v2.1.0-preview`, SP1 `v6.4.0`, ZisK `1.1.0-alpha` — so the version table and the release artifact names are unchanged.

Worth noting for reviewers: the only ere change in this range that touches a crate the guests link is `crates/platform/openvm/src/zkvm_accelerator.rs`, and that file is behind the `zkvm-accelerator` feature this build disables (we keep ethrex's own OpenVM crypto provider). The ELF bytes may still differ, since they are produced by a different compiler image. `build-stateless-validator-guest` runs on pushes to main and tags, not on pull requests, so the guest compile itself is exercised after merge.

**Checklist**

- [ ] Updated `STORE_SCHEMA_VERSION` (crates/storage/lib.rs) if the PR includes breaking changes to the `Store` requiring a re-sync. — not applicable, no `Store` changes.


.github/scripts/zkvm-version.sh
```diff
@@ -29,20 +36,37 @@ zkvm_version() {
...
+    "$BASE/bin/zisk/Cargo.toml"
+    "$BASE/bin/openvm/Cargo.toml"
+)
 
-# The table above is only valid for ERE_REV; refuse to guess if it has moved.
-if ! grep -q "rev = \"$ERE_REV\"" "$MANIFEST"; then
+# The table above is only valid for ERE_TAG; refuse to guess if anything moved.
+for manifest in "${MANIFESTS[@]}"; do
+    [[ -f $manifest ]] || { echo "no manifest at $manifest" >&2; exit 1; }
+    grep -q "tag = \"$ERE_TAG\"" "$manifest" && continue
```
MegaRedHand, line 56: We should match against `git = "https://github.com/eth-act/ere", tag = "$ERE_TAG"` here, in case there's another dependency tagged by tag and version as ERE

MegaRedHand (approved):

PR #7259, opened 2026-09-07
fix(l1): refuse a newer-schema database with a version error instead of a fake migration failure

ilitteri:
**Motivation**

When a datadir carries a schema version above the binary's, the store returned `MigrationFailed { from: v, to: STORE_SCHEMA_VERSION }` even though no migration ran. Its Display reads `Migration from v4 to v3 failed: DB version v4 is more recent than the client expects (v3)`, and the node wrapped it with "The database may be in an inconsistent state. Please erase your DB by running `ethrex removedb`". Both statements are false: the database is intact, and a binary that speaks the newer schema opens it as is.

This is not hypothetical. On a devnet-8 node the same datadir was opened by an image built from `main` (schema v4, which stamps the marker via the no-op `migrate_3_to_4`) and then handed back to the devnet-branch image (v3). The v3 binary crash-looped for ~9 hours with the message above, and the operator followed the advice and wiped a healthy 70 GB database — twice in one night — when switching images back would have resumed the node without a resync.

Related dead code made the situation worse: `IncompatibleDBVersion { found, expected }` existed with a dedicated startup message but was never constructed, and the "Database schema is at v{n}, ahead of this binary's… Upgrade the binary" warning in `run_pending_migrations` was unreachable because the store guard returns first.

**Description**

- `Store::new_with_config` guard: the "database ahead" and "v0 marker" arms return `IncompatibleDBVersion` instead of a fabricated `MigrationFailed`. `MigrationFailed` is now only produced by a migration step that actually failed, so its "may be in an inconsistent state" advice is accurate whenever it appears.
- `run_pending_migrations`: the unreachable warn-and-continue branch for `current_version > STORE_SCHEMA_VERSION` now returns `IncompatibleDBVersion`, so no caller can proceed into a format the binary cannot decode. The function's and `migrate_3_to_4`'s doc comments described the old (never-executed) behaviour and are corrected.
- Node startup (`cmd/ethrex/initializers.rs`): a newer-schema database gets its own message, naming the datadir and both versions, stating that the database has not been modified, that starting a build which supports that schema resumes without resyncing, and presenting `ethrex removedb` only as the way to deliberately abandon it. The pre-existing `removedb` advice is kept for `NotFoundDBVersion` and for a corrupt v0 marker, where it is correct.
- `docs/l1/fundamentals/databases.md`: documents the startup outcomes (same / older / newer / unversioned database) and the mixed-branch scenario that produces the newer case.
- Tests: `schema_version_guard_tests` (store guard: ahead → `IncompatibleDBVersion` with the marker left untouched; v0 → `IncompatibleDBVersion`) and `run_pending_migrations_rejects_a_database_ahead_of_the_binary` (migration runner).

Before / after, running a v4 binary against a datadir stamped v99:

```
Error: Migration from v99 to v4 failed: DB version v99 is more recent than the client expects (v4). Rolling back is not supported. The database may be in an inconsistent state. Please erase your DB by running `ethrex removedb` and restart node to resync.
```

```
Error: The database at <datadir> was written by a newer ethrex (schema v99) than this binary supports (schema v4). Downgrading a database is not supported and it has not been modified. Start an ethrex build that supports schema v99 or later and the node will resume without resyncing. Only if you intend to abandon this database, erase it with `ethrex removedb` and resync from scratch.
```

No schema or data-format change; `STORE_SCHEMA_VERSION` is unchanged.

**Checklist**

### DIFF
--- a/cmd/ethrex/initializers.rs
+++ b/cmd/ethrex/initializers.rs
@@ -867,6 +867,18 @@ pub async fn init_l1(
     };
     let store = match store_result {
         Ok(store) => store,
+        // Written by a newer ethrex. The data is intact and needs no resync; the
+        // fix is the binary, not the database. `removedb` is only mentioned as the
+        // deliberate way to abandon it.
+        Err(StoreError::IncompatibleDBVersion { found, expected }) if found > expected => {
+            return Err(eyre::eyre!(
+                "The database at {} was written by a newer ethrex (schema v{found}) than this binary supports (schema v{expected}). \
+                 Downgrading a database is not supported and it has not been modified. \
+                 Start an ethrex build that supports schema v{found} or later and the node will resume without resyncing. \
+                 Only if you intend to abandon this database, erase it with `ethrex removedb` and resync from scratch.",
+                datadir.display()
+            ));
+        }
         Err(err @ StoreError::IncompatibleDBVersion { .. })
         | Err(err @ StoreError::NotFoundDBVersion) => {
             return Err(eyre::eyre!(
--- a/crates/storage/migrations.rs
+++ b/crates/storage/migrations.rs
@@ -50,9 +50,9 @@ fn migration_for_version(version: u64) -> MigrationFn {
 /// list of `u32` offsets. Both forms are readable, so a v3 database needs no rewriting
 /// and this migration only moves the version marker.
 ///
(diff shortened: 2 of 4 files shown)


MegaRedHand (approved):
LGTM

